Issue #186, opened 2022-08-01
Add Envoy Service Support


danehans:
https://github.com/envoyproxy/gateway/pull/178 introduced the Kubernetes Infra Manager and manages the Envoy `ServiceAccount`. Support should be added for managing the Envoy `Service`.

PR #200, opened 2022-08-11
Adds Envoy Deployment Support

danehans:
Adds initial Envoy deployment support to Infra Manager.

Fixes: #185 

Signed-off-by: danehans <daneyonhansen@gmail.com>

PR #202, opened 2022-08-13
Adds Initial Envoy Service Support

danehans:
Adds initial support for provisioning the Envoy Service.

Requires: https://github.com/envoyproxy/gateway/pull/200

Fixes: https://github.com/envoyproxy/gateway/issues/186

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,6 +66,7 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	go.uber.org/atomic v1.7.0 // indirect
 	go.uber.org/multierr v1.6.0 // indirect
+	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e // indirect
 	golang.org/x/net v0.0.0-20220526153639-5463443f8c37 // indirect
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8 // indirect
 	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -197,13 +197,3 @@ func hostnameMatchesWildcardHostname(hostname, wildcardHostname string) bool {
 	wildcardMatch := strings.TrimSuffix(hostname, strings.TrimPrefix(wildcardHostname, "*"))
 	return len(wildcardMatch) > 0
 }
-
-// containsPortNum tells whether a contains b.
-func containsPortNum(a []v1beta1.PortNumber, b v1beta1.PortNumber) bool {
-	for _, n := range a {
-		if b == n {
-			return true
-		}
-	}
-	return false
-}
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -197,3 +197,13 @@ func hostnameMatchesWildcardHostname(hostname, wildcardHostname string) bool {
 	wildcardMatch := strings.TrimSuffix(hostname, strings.TrimPrefix(wildcardHostname, "*"))
 	return len(wildcardMatch) > 0
 }
+
+// containsPortNum tells whether a contains b.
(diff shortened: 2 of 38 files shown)


internal/gatewayapi/helpers.go
```diff
@@ -197,3 +197,13 @@ func hostnameMatchesWildcardHostname(hostname, wildcardHostname string) bool {
...
+
+// containsPortNum tells whether a contains b.
+func containsPortNum(a []v1beta1.PortNumber, b v1beta1.PortNumber) bool {
+	for _, n := range a {
+		if b == n {
+			return true
+		}
+	}
+	return false
+}
```
LukeShu, line 209: Pre-Go-1.18, this would have been a necessary function. But now we have [`golang.org/x/exp/slices.Contains`](https://pkg.go.dev/golang.org/x/exp/slices#Contains).

internal/infrastructure/kubernetes/service.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, listener := range infra.Proxy.Listeners {
+		for _, port := range listener.Ports {
+			// Set the target port based on the protocol of the IR port.
+			target := intstr.IntOrString{IntVal: envoyHTTPPort}
+			if port.Protocol == ir.HTTPSProtocolType {
+				target = intstr.IntOrString{IntVal: envoyHTTPSPort}
+			}
+			p := corev1.ServicePort{
+				Name:       port.Name,
+				Protocol:   "TCP",
```
LukeShu, line 76: Prefer to use constants like [`corev1.ProtocolTCP`](https://pkg.go.dev/k8s.io/api/core/v1#Protocol) instead of magic strings.

internal/infrastructure/kubernetes/service.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+const (
+	// envoyServiceHTTPPort is the HTTP port number of the Envoy service.
+	envoyServiceHTTPPort = int32(80)
+	// envoyServiceHTTPSPort is the HTTPS port number of the Envoy service.
+	envoyServiceHTTPSPort = int32(443)
```
LukeShu, line 20: Any reason to have these typed as `int32` rather than just letting them be [untyped constants](https://go.dev/blog/constants)?

(I prickle every time I see an port number typed as something other than the proper `uint16`; Kubernetes' excuse is that the Kube APIs were originally canonically specified as protobuf, and protobuf doesn't have 16-bit number types.)

LukeShu (approved):
Some minor nits that you can choose to disregard if you wish.

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -0,0 +1,218 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+
+	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+const (
+	// envoyContainerName is the name of the Envoy container.
+	envoyContainerName = "envoy"
+	// envoyNsEnvVar is the name of the contour namespace environment variable.
```
AliceProxy, line 21: Question: should this be `the name of the Envoy Gateway namespace environment variable` instead of `.... countour ....`?

AliceProxy (commented):
Had a question but overall looks good to me.
I would have to agree with Luke's nit about the port typed as something other than a `uint16`

AliceProxy (approved):
lgtm

go.mod
```diff
@@ -66,6 +66,7 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	go.uber.org/atomic v1.7.0 // indirect
 	go.uber.org/multierr v1.6.0 // indirect
+	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e // indirect
```
LukeShu, line 69: Weird that this ended up here; it's direct not indirect. Oh well, a future `go mod tidy` should move it.

LukeShu (approved):

PR #203, opened 2022-08-16
Initial docs framework.

kflynn:
Initial framework for docs.

This is pretty bare-bones stuff:

- Have a bare-bones docs directory
- Have make targets to install and run Sphinx to build HTML
- Have a CI workflow to run the docs build

Once this is getting some docs built at all in CI, next step
will be sorting out pushing them to the cloud for consumption
by the Envoy docs process.

Signed-off-by: Flynn <emissary@flynn.kodachi.com>


docs/get_involved.rst
```diff
@@ -0,0 +1,9 @@
+Getting Involved
```
phlax, line 1: you probably want to adopt a policy around title capitalization - in Envoy we are now moving towards "normal capitalization" - ie more like what you have below in:

```rst

Welcome to the Envoy Gateway documentation!
===========================================
```

ie - only the initial word and proper nouns are capitalized

doing what you have here "Where Every Word Is Capitalized" loses emphasis for proper nouns and throws up many more questions than it resolves - so my suggestion is to adopt normal caps

docs/conf.py
```diff
@@ -0,0 +1,21 @@
...
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'Envoy Gateway'
+copyright = '2022 Envoy Gateway Project Authors'
+author = 'Envoy Gateway Project Authors'
+
+version = '0.2.0-rc.1'
+release = '0.2.0-rc.1'
```
phlax, line 20: is it possible to get these versions from somewhere more authoritative ?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,23 @@
+name: Docs
+on:
+  push:
+    branches:
+      - "main"
```
phlax, line 5: it probably doesnt matter so much for gh actions yaml config - but in Envoy where we have a lot of complex envoy configs we now use "least indent" for yaml formatting - my ~strong suggestion would be to adopt that as a policy as it makes the yaml more readable - esp where people use both for yaml lists

docs/conf.py
```diff
@@ -0,0 +1,21 @@
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+    'sphinx.ext.duration',
+    'sphinx.ext.autosectionlabel',
+]
+
+html_theme = 'alabaster'
```
phlax, line 9: i dont think we need to worry too much at this stage - apart from anything the Envoy theme is pretty awful atm, but we may want to sync themes at some point so the website has some consistency

tools/make/docs.mk
```diff
@@ -0,0 +1,15 @@
+DOCS_DIR := docs
+DOCS_OUTPUT_DIR := docs/html
+
+.PHONY: docs
+docs: $(tools/sphinx-build)
+	tools/bin/sphinx-build -b html $(DOCS_DIR) $(DOCS_OUTPUT_DIR)
```
phlax, line 6: i would use multiproc for rendering if you can - ie `-j auto` - not all extensions are compatible with that but i would start with multiproc and review if its a problem down the road

phlax (commented):
@kflynn ive given this a quick review - hope that is helpful

im wondering if we can get a GCP/GCS account set up as a priority so that when the docs are rendered in CI reviewers can immediately view the output before approving etc

.github/workflows/docs.yaml
LukeShu: Is there a way to configure yamllint to check for that?

phlax: yep see here https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation

specifically with `indent-sequences` set to `false`

tools/make/docs.mk
LukeShu, line 6: use `$(tools/sphinx-build)` rather than hard-coding `tools/bin/sphinx-build`

.gitignore
```diff
@@ -1,2 +1,3 @@
 .DS_Store
 bin/
+docs/html
```
LukeShu, line 3: IMO this should either be rooted `/docs/html/` to ignore a specific directory, or `html/` to ignore any directory named "html".  "any file named 'html' in any directory named 'docs'" feels like a weird bad mixup of specific and vague to me.

docs/get_involved.rst
kflynn: My own preference is actually title case, but I think conforming to the Envoy style makes sense here.

.github/workflows/docs.yaml
kflynn: This is a great idea. The challenge is that there's a _ton_ of stuff in the Envoy Gateway project that fails lint when you do it. 😂 I'm going to hold off on this until we can discuss it.

tools/make/docs.mk
kflynn: 🤦‍♂️

.gitignore
kflynn: Switched to rooted.

tools/make/docs.mk
kflynn: 👍

docs/intro/index.rst
```diff
@@ -0,0 +1,15 @@
+Intro
+===========================================
```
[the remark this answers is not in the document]
kflynn: 👍

docs/conf.py
kflynn: I thought about that and decided that I didn't want to muck with themes right now. 😂 I'll take a look at the Envoy theme shortly.

docs/conf.py
kflynn: It needs to be made possible. 🙂

kflynn:
@phlax and @LukeShu, thanks for the feedback! I've addressed much of it and commented above where I've held out.

And @phlax, yeah, I'd love to get GCP/GPS set up ASAP. Who's the right person to talk to there?

kflynn:
Updated to use `goversion` for the version and to make `yamllint` pass with `indent-sequences: consistent`.

phlax:
> And @phlax, yeah, I'd love to get GCP/GPS set up ASAP. Who's the right person to talk to there?

apologies i missed this comment - @lizan do you what we would need to set up some storage to publish docs to from here?

.github/workflows/docs.yaml
kflynn: Resolving after switching to `indent-sequences: consistent`, per discussion with the group.

docs/conf.py
kflynn: Switched to `goversion` for this. 🙂

lizan:
I can't get GCS set up with GitHub Actions for PRs because they don't have secret in PR run. We might want to use [pull_request_target] trigger carefully though. Meanwhile we can just verify the build and push to https://github.com/envoyproxy/envoy-website after merging.

phlax (approved):
lgtm, thanks for following up @kflynn

kflynn:
Thanks @phlax! and @lizan. Looking forward to this getting landed. 🙂

danehans:
@LukeShu can you PTAL since you reviewed this PR previously?

tools/linter/yamllint/.yamllint
```diff
@@ -37,7 +37,7 @@ rules:
     max-spaces-after: 1
   indentation:
     spaces: 2
-    indent-sequences: whatever
+    indent-sequences: consistent
```
arkodg, line 40: nit: can you add a comment why this was needed ?

kflynn: It was desired rather than needed, FTR. 🙂 Done.

LukeShu (approved):

Issue #166, opened 2022-07-15
Manage ObservedGeneration of GatewayClass Status


danehans:
*Description*:
Pass spec.generation to newCondition(), so status.observedGeneration can be set.

[optional *Relevant Links*:]
Requires: https://github.com/envoyproxy/gateway/pull/150

Issue #165, opened 2022-07-15
Add Gateway Status Support 


danehans:
*Description*:
Add initial support for gateway status.

Depends on:

https://github.com/envoyproxy/gateway/pull/150
https://github.com/envoyproxy/gateway/pull/164

[optional *Relevant Links*:]
The implementation should follow [Contour's implementation](https://github.com/projectcontour/contour/blob/main/internal/status/gatewaystatus.go), and be consistent with the existing Kube provider [controllers](https://github.com/envoyproxy/gateway/tree/main/pkg/provider/kubernetes).

skriss:
Note that much of this is already implemented in https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi, but there may be some additional infra-related status items that need to be computed.

danehans:
All kube provider controllers need to write status back to kube API.

danehans:
Although the resource translator can handle status updates, something still needs to `Update()` using kube client.

PR #205, opened 2022-08-17
Adds Gateway Status Support

danehans:
Previously, the gateway controller would not update the status of gateways. This PR adds the `StatusUpdater` interface to the gateway controller and sets the "Scheduled=True" status condition for gateways that match the configured GatewayClass controllerName.

Fixes: https://github.com/envoyproxy/gateway/issues/165
Fixes: https://github.com/envoyproxy/gateway/issues/166


skriss (approved):
LGTM

Issue #201, opened 2022-08-12
Add Bootstrap Config to Envoy Deployment


danehans:
https://github.com/envoyproxy/gateway/pull/200 added initial support for deploying Envoy. However, this PR did not include creating Envoy's bootstrap config. A solution should be added for provisioning the bootstrap config as part of the Envoy deployment.

danehans:
@youngnick I see that Contour uses an init container to manage the bootstrap config. What is the reason why a configmap containing the bootstrap config can't simply be mounted to the Envoy container?

danehans:
xref https://github.com/projectcontour/contour/issues/1

youngnick:
I think that there were a couple of reasons:
- We had some information that needed the pod to be up before we could generate the bootstrap (although I can't recall what that is). I think this one is mitigated in the Envoy Gateway case by the fact that we're controlling the whole creation process.
- We wanted not to have any secret stuff in the bootstrap visible in the cluster. I think that there's actually not much that's very secret in the bootstrap config itself - all the secret stuff will be stored in Kubernetes Secrets (basically, the TLS details for connecting to the xDS server in EG).

In summary, I think that using a configmap for EG should be fine, if there are any problems with the approach, we can always change back later.

danehans:
Help wanted to either update take either of the following approaches to manage the bootstrap config:
1. Add the bootstrap config cli arg to the Envoy deployment.
2. Manage a configmap that holds the bootstrap config and mount the configmap to the envoy deployment.

PR #197, opened 2022-08-09
connect components using watchable

arkodg:
Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #206, opened 2022-08-17
Adds Envoy Bootstrap Config

danehans:
Adds support for managing Envoy's bootstrap config through a ConfigMap.

Fixes: #201 


internal/infrastructure/kubernetes/configmap.go
```diff
@@ -0,0 +1,179 @@
...
+  - connect_timeout: 1s
+    load_assignment:
+      cluster_name: xds_cluster
+      endpoints:
+      - lb_endpoints:
+        - endpoint:
+            address:
+              socket_address:
+                address: {{ .XdsServerAddress }}
+                port_value: 18000
```
arkodg, line 62: should be templated

internal/infrastructure/kubernetes/configmap.go
```diff
@@ -0,0 +1,179 @@
...
+	// envoyGatewayService is the name of the Envoy Gateway service.
+	envoyGatewayService = "envoy-gateway"
+)
+
+var envoyTmpl = template.Must(template.New(envoyCfgFileName).Parse(`
+admin:
+  access_log_path: /dev/null
+  address:
+    socket_address:
+      address: 127.0.0.1
```
arkodg, line 27: address and port value should be templated

internal/infrastructure/kubernetes/configmap.go
```diff
@@ -0,0 +1,179 @@
...
+	Envoy envoyConfig
+}
+
+// envoyConfig contains Envoy configuration parameters.
+type envoyConfig struct {
+	// XdsServerAddress is the address of the XDS Server that Envoy is managed by.
+	XdsServerAddress string
+}
+
+// createConfigMapIfNeeded creates a ConfigMap based on the provided infra, if
```
arkodg, line 93: afaik a pod doesn't restart if the config map is updated, we'd still need extra instrumentation to ensure that the pod restarts when the config map changes
a simpler solution might be to use the [config-yaml](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-config-yaml) argument and pass the config as a yaml string directly

danehans: @arkodg I refactored the PR to use `config-yaml` instead of a ConfigMap. @LukeShu @youngnick https://github.com/envoyproxy/gateway/pull/211 is a draft PR to use the configmap approach. This approach requires add'l refactoring that I'd prefer doing after #197. In the meantime, I would like to proceed with the [config-yaml](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-config-yaml) approach so we can begin e2e.

internal/infrastructure/kubernetes/configmap.go
danehans: https://github.com/envoyproxy/gateway/issues/212 tracks this requirement.

internal/infrastructure/kubernetes/configmap.go
danehans: https://github.com/envoyproxy/gateway/issues/212 tracks this requirement.

arkodg (approved):

AliceProxy (approved):
bootstrap looks good to me

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -27,13 +29,102 @@ const (
...
 	// envoyCfgFileName is the name of the Envoy configuration file.
-	envoyCfgFileName = "envoy.json"
+	envoyCfgFileName = "bootstrap.yaml"
 	// envoyHTTPPort is the container port number of Envoy's HTTP endpoint.
 	envoyHTTPPort = int32(8080)
 	// envoyHTTPSPort is the container port number of Envoy's HTTPS endpoint.
 	envoyHTTPSPort = int32(8443)
 )
 
+var bootstrapTmpl = template.Must(template.New(envoyCfgFileName).Parse(`
```
LukeShu, line 39: I'd consider putting this in a separate `bootstrap.yaml` file and saying 

```go
//go:embed bootstrap.yaml
const string bootstrapTmplStr

var bootstrapTmpl = template.Must(template.New(envoyCfgFileName).Parse(bootstrapTmplStr))
```
so that the file is easier to edit, with editors recognizing it as yaml.

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -27,13 +29,102 @@ const (
...
+            grpc_services:
+              envoy_grpc:
+                cluster_name: xds_cluster
+        name: runtime-0
+`))
+
+var (
+	// envoyGatewayService is the name of the Envoy Gateway service.
+	envoyGatewayService = "envoy-gateway"
+)
```
LukeShu, line 101: Surely this should be `const`?

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -27,13 +29,102 @@ const (
...
+	XdsServerAddress string
+}
+
+// render the stringified bootstrap config in yaml format.
+func (b *bootstrapConfig) render() error {
+	buf := new(bytes.Buffer)
+	if err := bootstrapTmpl.Execute(buf, b.parameters); err != nil {
+		return fmt.Errorf("failed to render bootstrap config: %v", err)
+	}
+	b.rendered = buf.String()
```
LukeShu, line 123: If we never need access it as a `[]byte` or mutate it (other than append), then it'd be more efficient to use a `strings.Builder` instead of a `bytes.Buffer`.

LukeShu (approved):
Some nits that you can choose to ignore. On the whole, lgtm.

internal/infrastructure/kubernetes/deployment.go
danehans: https://github.com/envoyproxy/gateway/pull/224

internal/infrastructure/kubernetes/deployment.go
danehans: https://github.com/envoyproxy/gateway/issues/225

danehans:
@LukeShu since we're getting close to the RC1 release, I created issues for most of your feedback and https://github.com/envoyproxy/gateway/pull/224 to use a const for the Envoy service name.

Issue #70, opened 2022-05-23
Control Plane (xDS Server - Envoy) Auth + Authz


arkodg:
This issue tracks how the the data plane Envoy Proxy authenticates and authorizes with the xDS Server within Envoy Gateway

danehans:
@LukeShu I assigned this issue to you based on the 5/24/22 community meeting [notes](https://docs.google.com/document/d/1leqwsHX8N-XxNEyTflYjRur462ukFxd19Rnk3Uzy55I/edit?usp=sharing). Please unassign yourself if that's not the case.

danehans:
xref https://github.com/envoyproxy/gateway/issues/97#issuecomment-1150602629 regarding how CP<>DP auth can be supported.

youngnick:
I think the relevant part of that comment is:

> In terms of authenticating between the control plane and Envoy, Contour has solved this problem in the simple case, using an on-disk SDS for hot-reload of certificates, an install process that generates one-year validity certs for both Contour and Envoy, and updates them on each upgrade. The update is seamless because the secrets change "on disk", Envoy sees the SDS change, and then control plane connection is rebuilt in less than 30 seconds. Notably, because it uses Kubernetes Secrets as intended and can handle rotations whenever you're ready, it plugs in easily to Vault or other secret-management tools - the Secret admin just has to ensure that the Secrets are updated for both Envoy and Contour around the same time.

I think that Contour's solution here is pretty great (maybe I'm biased, lol), and will work well to allow secure xds-server <-> Envoy comms no matter what provider is in use (although for things that use localhost only, we can probably skip TLS).

youngnick:
Not stale, just backlogged.

danehans:
FYI @arkodg I'm going to take this work on.

PR #208, opened 2022-08-18
Adds Support for Control Plane TLS

danehans:
- Adds a package for generating control plane auth certs.
- Adds a const to the EG config pkg to represent the default "envoy-gateway-system" ns.
- Updates `ObjectName()` method to get/set the default name of the ProxyInfra type.

fixes: #70 

### DIFF
--- a/internal/crypto/certgen.go
+++ b/internal/crypto/certgen.go
@@ -1,0 +1,278 @@
+package crypto
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha1" // nolint:gosec
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/envoyproxy/gateway/api/config/v1alpha1"
+	"github.com/envoyproxy/gateway/internal/envoygateway/config"
+)
+
+const (
+	// DefaultEnvoyGatewayDNSPrefix defines the default Envoy Gateway DNS prefix.
+	DefaultEnvoyGatewayDNSPrefix = config.EnvoyGatewayServiceName
+
+	// DefaultEnvoyDNSPrefix defines the default Envoy DNS prefix.
+	DefaultEnvoyDNSPrefix = config.EnvoyServiceName
+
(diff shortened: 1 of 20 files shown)


LukeShu:
To communicate: I haven't been looking at this because it is failing CI.

danehans:
@LukeShu I've been side tracked with other PR. I marked the PR as WIP.

internal/crypto/certgen_test.go
```diff
@@ -0,0 +1,133 @@
...
+	expiry := now.Add(24 * 365 * time.Hour)
+
+	caCert, caKey, err := newCA("envoy-gateway", expiry)
+	require.NoErrorf(t, err, "Failed to generate CA cert")
+
+	egCertReq := &certificateRequest{
+		caCertPEM:  caCert,
+		caKeyPEM:   caKey,
+		expiry:     expiry,
+		commonName: "envoy-gateway",
```
lizan, line 65: Common names are deprecated in RFC 2818 (in 2000) in favor of SANs. Major softwares (Golang, Chrome etc) stopped supporting them, so we should just issue certs without common names.

danehans:
xref #404 will causes this PR to support a TLS secret per managed Envoy deployment.

internal/crypto/certgen.go
```diff
@@ -0,0 +1,278 @@
...
+		DNSNames: request.altNames,
+	}
+	newCert, err := x509.CreateCertificate(rand.Reader, template, caCert, &newKey.PublicKey, caKey)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newKeyPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(newKey),
```
LukeShu, line 204: Getting in to bike-shed territory, but any reason to not use `"PRIVATE KEY"` and `x509.MarshalPKCS8PrivateKey`? It'd have fewer things to change if we ever want to switch away from RSA.

LukeShu (approved):
Some nit-picks, but on the whole LGTM.

LukeShu:
If anyone cares, here's some very similar code in Emissary: https://github.com/emissary-ingress/emissary/blob/master/cmd/apiext/ca.go

PR #216, opened 2022-08-22
Adds ProviderResource Nil Checks

### LINKED ISSUE #214: Runners Require Nil Check
#197 connects EG components using the [watchable](https://pkg.go.dev/github.com/telepresenceio/watchable) library. Since [ProviderResources](https://github.com/envoyproxy/gateway/blob/main/internal/message/types.go#L14-L22) contain pointer slices, the elements may be nil. Runners should check for nil pointers.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x21aff6d]

goroutine 153 [running]:
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate(0xc000a00e00, {0x29dd170, 0xc0000bfe40})
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:65 +0x12d
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:35 +0x165
```

danehans:
Adds nil checks for `ProviderResources` gatewayclasses, gateways, and httproutes.

Fixes #214 

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -64,15 +64,25 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		gatewayClasses := r.ProviderResources.GetGatewayClasses()
 		// Fetch the first gateway class since there should be only 1
 		// gateway class linked to this controller
-		t := &gatewayapi.Translator{
-			GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+		switch {
+		case gatewayClasses == nil:
+			// No need to translate, publish empty IRs, e.g. delete operation.
+			r.XdsIR.Delete(r.Name())
+			r.InfraIR.Delete(r.Name())
+		case gatewayClasses[0] == nil:
+			panic("gatewayclass is nil")
+		default:
+			// Translate and publish IRs.
+			t := &gatewayapi.Translator{
+				GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+			}
+			// Translate to IR
+			result := t.Translate(&in)
+			// Publish the IRs. Use the service name as the key
+			// to ensure there is always one element in the map
+			r.XdsIR.Store(r.Name(), result.XdsIR)
+			r.InfraIR.Store(r.Name(), result.InfraIR)
 		}
-		// Translate to the IRs
-		result := t.Translate(&in)
-		// Publish the IRs. Use the service name as the key to ensure there
-		// is always one element in the map.
-		r.XdsIR.Store(r.Name(), result.XdsIR)
-		r.InfraIR.Store(r.Name(), result.InfraIR)
 	}
 	r.Logger.Info("shutting down")
 }
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -112,6 +112,9 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
 	var relevant []*GatewayContext
 
 	for _, gateway := range gateways {
+		if gateway == nil {
+			panic("received nil gateway")
+		}
 		if gateway.Spec.GatewayClassName == t.GatewayClassName {
 			gc := &GatewayContext{
 				Gateway: gateway.DeepCopy(),
@@ -451,6 +454,9 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 	var relevantHTTPRoutes []*HTTPRouteContext
 
 	for _, h := range httpRoutes {
+		if h == nil {
+			panic("received nil httproute")
+		}
 		httpRoute := &HTTPRouteContext{HTTPRoute: h}
 
 		// Find out if this route attaches to one of our Gateway's listeners,


internal/gatewayapi/runner/runner.go
```diff
@@ -61,15 +61,17 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		gatewayClasses := r.ProviderResources.GetGatewayClasses()
 		// Fetch the first gateway class since there should be only 1
 		// gateway class linked to this controller
-		t := &gatewayapi.Translator{
-			GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+		if gatewayClasses[0] != nil {
```
arkodg, line 64: you probably also want to check if `gatewayClasses` is nil or not to handle the case when there are no more gateway classes left / the subscribe notification was triggered due to a delete

internal/gatewayapi/translator.go
```diff
@@ -112,7 +112,7 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
 	var relevant []*GatewayContext
 
 	for _, gateway := range gateways {
-		if gateway.Spec.GatewayClassName == t.GatewayClassName {
+		if gateway != nil && gateway.Spec.GatewayClassName == t.GatewayClassName {
```
arkodg, line 115: imho this should ideally not be happening, but rather than skip it, we should either
* log it or
* return early and empty
to surface the underlying issue faster

internal/gatewayapi/translator.go
```diff
@@ -451,6 +451,9 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 	var relevantHTTPRoutes []*HTTPRouteContext
 
 	for _, h := range httpRoutes {
+		if h == nil {
```
arkodg, line 454: same as https://github.com/envoyproxy/gateway/pull/216/files#r952032139

arkodg, line 454: same as https://github.com/envoyproxy/gateway/pull/216/files#r952032139

internal/gatewayapi/translator.go
youngnick: I agree we shouldn't be seeing `nil` gateways, so I'd actually be in favor of doing
```go
if gateway == nil {
  // Log some message here
  panic
}
```

internal/gatewayapi/translator.go
youngnick: Same for me, I'd rather see a `panic` than a `continue`.

youngnick (commented):
Thanks for finding, I think we should be brutal here and just `panic` if we get nil values, something has gone seriously wrong, and I'd rather find those cases now before v0.2.0.

danehans:
@arkodg @youngnick commit b35f07f implements your feedback. However, I am seeing nil gatewayclasses, gateways, and httproutes in my e2e testing so we have an issue elsewhere in the code that must be resolved.

internal/gatewayapi/runner/runner.go
```diff
@@ -61,15 +61,17 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		gatewayClasses := r.ProviderResources.GetGatewayClasses()
 		// Fetch the first gateway class since there should be only 1
 		// gateway class linked to this controller
-		t := &gatewayapi.Translator{
-			GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+		if gatewayClasses == nil || gatewayClasses[0] != nil {
```
arkodg, line 64: above line will panic on gateway class deletions
consider this change
if gatewayClasses == nil 
// no need to run translate, publish empty IRs
r.XdsIR.Store(r.Name(), nil)
r.InfraIR.Store(r.Name(), nil)
} else {
 Translate .....
 Store results......
}

internal/gatewayapi/runner/runner.go
```diff
@@ -61,15 +61,24 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		gatewayClasses := r.ProviderResources.GetGatewayClasses()
 		// Fetch the first gateway class since there should be only 1
 		// gateway class linked to this controller
-		t := &gatewayapi.Translator{
-			GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+		switch {
```
arkodg, line 64: nit: this is functionally correct, but I still prefer the if ... else if.... else style here, did`golangci-lint` suggest this style :) ?

internal/gatewayapi/runner/runner.go
```diff
@@ -61,15 +61,24 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
+		default:
+			// Translate and publish IRs.
+			t := &gatewayapi.Translator{
+				GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+			}
+			// Translate to IR
+			result := t.Translate(&in)
+			// Publish the IR. Use the service name as the key
+			// to ensure there is always one element in the map
+			r.XdsIR.Store(r.Name(), result.XdsIR)
```
arkodg, line 80: 🙈 I forgot to add the `r.InfraIR.Store(r.Name(), result.InfraIR)` here, I can add it in a future PR if it doesnt get in, in this PR

arkodg (approved):

internal/gatewayapi/runner/runner.go
danehans: Geez... I added it here in [L68](https://github.com/envoyproxy/gateway/pull/216/files#diff-0a92751feeb3e64824cb22f525cb08b27464d31c38cbeb309bb7f4d9e12e90c0R68) but forgot to add it here. I'll update and push a new commit.

internal/gatewayapi/runner/runner.go
```diff
@@ -61,15 +61,24 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		gatewayClasses := r.ProviderResources.GetGatewayClasses()
 		// Fetch the first gateway class since there should be only 1
 		// gateway class linked to this controller
-		t := &gatewayapi.Translator{
-			GatewayClassName: v1beta1.ObjectName(gatewayClasses[0].GetName()),
+		switch {
+		case gatewayClasses == nil:
+			// No need to translate, publish empty IRs, e.g. delete operation.
+			r.XdsIR.Store(r.Name(), nil)
+			r.InfraIR.Store(r.Name(), nil)
```
LukeShu, line 68: There is an actual `.Delete` method; why do we want these keys to stick around but be set to `nil`?

internal/gatewayapi/runner/runner.go
danehans: @arkodg https://github.com/envoyproxy/gateway/pull/224/ adds `r.InfraIR.Store()` as needed.

internal/gatewayapi/runner/runner.go
danehans: I find `switch` statements easier to read.

arkodg (dismissed):

danehans:
@arkodg I had to rebase. Please retag.

internal/gatewayapi/runner/runner.go
youngnick: I missed this before, but `switch` statements are generally more idiomatic in Go for more than one `if` statement.

arkodg (approved):

PR #217, opened 2022-08-22
tests for message pkg

arkodg:
* specify len and capacity seperately in `make` api
* gen DeepCopy for infra ir structs


internal/message/types_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+func TestProviderResources(t *testing.T) {
```
danehans, line 14: Can you add a test case where resources is nil?

xref: https://github.com/envoyproxy/gateway/pull/216

internal/message/types_test.go
```diff
@@ -0,0 +1,122 @@
...
+			Name:      "test1",
+			Namespace: "test",
+		},
+	}
+	r1 := &gwapiv1b1.HTTPRoute{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "route1",
+			Namespace: "test",
+		},
+	}
```
danehans, line 32: Can you add a test case where gatewayclasses, gateways, and httproutes are nil?

xref: https://github.com/envoyproxy/gateway/pull/216

danehans (changes requested):

danehans (approved):

PR #224, opened 2022-08-23
Adds xDS Server Configuration API 

danehans:
Adds xDS Server Configuration API and generates deepcopy funcs, manifests, etc.


arkodg:
isn't the comment referring to making https://github.com/envoyproxy/gateway/blob/9c34a8c5bd96e8dfb4fb34e0cea1995a951286f7/internal/infrastructure/kubernetes/deployment.go#L98 as a const.
If we plan on retrieving the xds address from the config package, it should at least be part of the config API imho, but for now I see most of the definitions in https://github.com/envoyproxy/gateway/blob/9c34a8c5bd96e8dfb4fb34e0cea1995a951286f7/internal/infrastructure/kubernetes/deployment.go#L20 which is fine for now

LukeShu:
What Arko said is how I intended it. I just saw a global value that didn't look like it should change declared as `var` when it could have been declared as `const`.

youngnick (approved):

youngnick:
Ah, sorry, I reviewed without reading the comments. I'll unapprove and leave y'all to sort this one out.

youngnick:
hmm, seems I can't unapprove, sorry. More caffeine required for me today.

api/config/v1alpha1/helpers.go
```diff
@@ -50,6 +42,29 @@ func DefaultProvider() *Provider {
 	}
 }
 
-func ProviderTypePtr(p ProviderType) *ProviderType {
```
danehans, line 53: __Note:__ This function was unused.

arkodg:
@danehans would prefer if you split up the work in this PR into multiple smaller PRs so it would be easier for reviewers to review this code, for e.g. would be great if there was 1 PR was just the API changes for introducing the `EnvoyProxy` Kind resource. TIA

danehans:
@arkodg most of the changes are related to implementing the API. I don't think pulling the API changes into a separate PR will make much of a difference in the size of the implementation PR, but I'll go ahead separate the two. The PR looks daunting mostly due to [testdata](https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata) updates.

arkodg:
thanks for just including the API changes in this PR ! reviewing rn ....

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -18,8 +23,21 @@ type EnvoyProxy struct {
 
 // EnvoyProxySpec defines the desired state of EnvoyProxy.
 type EnvoyProxySpec struct {
```
arkodg, line 25: Just refreshed my memory by going through https://github.com/envoyproxy/gateway/blob/main/docs/design/CONFIG_API.md#data-plane-api

based on the agreed upon design, the `EnvoyGateway` config, passed via the cmdline arg,  decides the configuration for the control plane components/runners/services
and the `EnvoyProxy` resources passed into K8s, linked to the `GatewayClass` decides the configuration of the dataplane Envoy proxy fleet.

The Xds Server address & port (not included here but is required) is needed by the proxy to connect to the server and the same address (optional, can bind to all addresses) and port is required by the server to listen for xds client connections, so such a configuration is needed in `EnvoyGateway` too and the values in both must match .
It has been hardcoded in the `xds-server` for now https://github.com/envoyproxy/gateway/blob/d99f32feef8ddd6ede45bd0c93b607586138d4c8/internal/xds/server/runner/runner.go#L57

danehans: @arkodg agreed that we should have logic to ensure the EnvoyProxy xds config aligns with how EG is running. IMO we need to be careful to not outsmart ourselves as different client/server config permutations will work. I think we track this as an issue to improve in the future. In the meantime, the API docs should suffice or I can elaborate on the docs if needed. THoughts @arkodg @LukeShu @youngnick @skriss

danehans:
Closing since we want don't want to expose config to users in v0.2.0. This can be reopendor used as a reference implementation of the EnvoyProxy API in the future.

PR #230, opened 2022-08-23
Adds Namespaces to Watchable ProviderResources

### LINKED ISSUE #229: Panic Due to ProviderResources Nil Namespaces
The translator panics when processing listeners allowed namespaces since the field is not set when initializing the gatewayapi runner.

```
$make run-kube-local
...
2022-08-23T10:11:15.490-0700	INFO	runner/runner.go:57	received a notification	{"runner": "gateway-api"}
goroutine 60 [running]:
github.com/envoyproxy/gateway/internal/gatewayapi.(*ListenerContext).AllowsNamespace(0xc0004f3590?, 0xc0004f32d0?)
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/contexts.go:147 +0xe5
github.com/envoyproxy/gateway/internal/gatewayapi.(*Translator).ProcessHTTPRoutes(0xc0007a3ed8?, {0xc000ace0e8, 0x1, 0x1?}, {0xc000ace100, 0x1, 0x1}, 0xc0005e1f18, 0xc000ac2720)
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/translator.go:481 +0x2973
github.com/envoyproxy/gateway/internal/gatewayapi.(*Translator).Translate(0xc0005e1ed8, 0xc0005e1f18)
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/translator.go:106 +0x1e5
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate(0xc0005b5780, {0x29e0c30, 0xc000382100})
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:68 +0x157
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:35 +0x165
```

danehans:
Adds namespace support to watchable ProviderResources.

Fixes #229 

### DIFF
--- a/internal/provider/kubernetes/httproute.go
+++ b/internal/provider/kubernetes/httproute.go
@@ -5,6 +5,7 @@ package kubernetes
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -12,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -121,17 +121,62 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
 
 	log.Info("reconciling httproute")
 
-	// Fetch the HTTPRoute from the cache.
-	httpRoute := &gwapiv1b1.HTTPRoute{}
-	err := r.client.Get(ctx, request.NamespacedName, httpRoute)
-	if errors.IsNotFound(err) {
-		log.V(2).Info("httproute not found, deleting it from the ResourceTable")
-		r.resources.HTTPRoutes.Delete(request.NamespacedName)
-		return reconcile.Result{}, nil
+	// Fetch all HTTPRoutes from the cache.
+	routeList := &gwapiv1b1.HTTPRouteList{}
+	if err := r.client.List(ctx, routeList); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error listing httproutes")
 	}
 
-	log.V(2).Info("adding httproute to the ResourceTable")
-	r.resources.HTTPRoutes.Store(request.NamespacedName, httpRoute.DeepCopy())
+	found := false
+	for i := range routeList.Items {
+		// See if this route from the list matched the reconciled route.
+		route := routeList.Items[i]
+		routeKey := NamespacedName(&route)
+		if routeKey == request.NamespacedName {
+			found = true
+		}
+
+		// Store the httproute in the resource map.
+		r.resources.HTTPRoutes.Store(routeKey, &route)
+		log.Info("added httproute to resource map")
+
+		// Get the route's namespace from the cache.
+		nsKey := types.NamespacedName{Name: route.Namespace}
+		ns := new(corev1.Namespace)
+		if err := r.client.Get(ctx, nsKey, ns); err != nil {
+			if errors.IsNotFound(err) {
+				// The route's namespace doesn't exist in the cache, so remove it from
+				// the namespace resource map if it exists.
+				if _, ok := r.resources.Namespaces.Load(nsKey.Name); ok {
+					r.resources.Namespaces.Delete(nsKey.Name)
+					log.Info("deleted namespace from resource map")
+				}
+			}
+			return reconcile.Result{}, fmt.Errorf("failed to get namespace %s", nsKey.Name)
+		}
+
+		// The route's namespace exists, so add it to the resource map.
+		r.resources.Namespaces.Store(nsKey.Name, ns)
+		log.Info("added namespace to resource map")
+	}
+
+	if !found {
+		// Delete the httproute from the resource map.
+		r.resources.HTTPRoutes.Delete(request.NamespacedName)
+		log.Info("deleted httproute from resource map")
+
+		// Delete the namespace from the resource map if no other routes exist in the namespace.
+		routeList = &gwapiv1b1.HTTPRouteList{}
+		if err := r.client.List(ctx, routeList, &client.ListOptions{Namespace: request.Namespace}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("error listing httproutes")
+		}
+		if len(routeList.Items) == 0 {
+			r.resources.Namespaces.Delete(request.Namespace)
+			log.Info("deleted namespace from resource map")
+		}
+	}
+
+	log.Info("reconciled httproute")
 
 	defer r.initializeOnce.Do(r.resources.Initialized.Done)
 	return reconcile.Result{}, nil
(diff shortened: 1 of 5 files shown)


internal/provider/kubernetes/httproute.go
```diff
@@ -127,11 +127,14 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
 	if errors.IsNotFound(err) {
 		log.V(2).Info("httproute not found, deleting it from the ResourceTable")
 		r.resources.HTTPRoutes.Delete(request.NamespacedName)
+		r.resources.Namespaces.Delete(request.Namespace)
```
arkodg, line 130: what happens if there other http routes in this namespace ?

internal/provider/kubernetes/httproute.go
```diff
@@ -127,11 +127,14 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
 		return reconcile.Result{}, nil
 	}
 
 	log.V(2).Info("adding httproute to the ResourceTable")
-	r.resources.HTTPRoutes.Store(request.NamespacedName, httpRoute.DeepCopy())
+	copy := httpRoute.DeepCopy()
+	r.resources.HTTPRoutes.Store(request.NamespacedName, copy)
+	r.resources.Namespaces.Store(request.Namespace, namespaceFromName(copy.Namespace))
```
arkodg, line 137: this line seems to store the namespace from the httpRoute's namespace
Im trying to understand why the namespace is relevant in the translator API if the namespace and http route namespace will always match making this check 
https://github.com/envoyproxy/gateway/blob/9c34a8c5bd96e8dfb4fb34e0cea1995a951286f7/internal/gatewayapi/translator.go#L478
always compute to true (which is the only place Namespaces seems to be used today)

skriss: The Namespace resources are needed because Gateway API gives Listeners the option to allow Routes to bind to them only if they're in a Namespace that matches a particular label selector (see https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.RouteNamespaces). Hence we need the Namespace, including its labels, to be able to check if it's allowed.

Because of that, I'm not sure that this approach of directly instantiating a Namespace will work - we really need to fetch it from the API server somehow, with its labels, in order for this part of the API to work.

skriss: Possibly you want to do this in the HTTPRoute controller - list all the HTTPRoutes (as we just discussed on the call), then get each one's associated Namespace from the client/cache, then deliver the list of all Routes and all Namespaces to the resource translator. You could also do Services there, since we'll need those as well.

danehans: > then get each one's associated Namespace from the client/cache

@skriss why do we need to get the httproute's associated ns object as opposed to building a list of local ns objects from the httproute's namespace? You mentioned something about checking ns labels, but I'm not following.

internal/provider/kubernetes/httproute.go
```diff
@@ -121,17 +121,61 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+						r.resources.Namespaces.Delete(nsKey.Name)
+						log.Info("deleted namespace from resource map")
+					}
+				}
+				return reconcile.Result{}, fmt.Errorf("failed to get namespace %s", nsKey.Name)
+			}
+
+			// The route's namespace exists, so add it to the resource map.
+			r.resources.Namespaces.Store(routeKey.Namespace, ns)
+			log.Info("added namespace to resource map")
```
skriss, line 157: Any reason not to run this logic for _all_ HTTPRoutes, not just the one that triggered the Reconcile? I think it'd be easier to reason about if we were processing all Resources, and likely helps avoid cold-start issues.

internal/provider/kubernetes/httproute.go
```diff
@@ -121,17 +121,61 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+	routeList := &gwapiv1b1.HTTPRouteList{}
+	if err := r.client.List(ctx, routeList); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error listing httproutes")
+	}
+
+	found := false
+	for i := range routeList.Items {
+		// Store the httproute in the resource map.
+		route := routeList.Items[i]
+		routeKey := NamespacedName(route.DeepCopy())
```
skriss, line 134: Don't think you need to do a DeepCopy here:
```suggestion
		routeKey := NamespacedName(&route)
```

internal/provider/kubernetes/httproute.go
```diff
@@ -121,17 +121,61 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+		routeList = &gwapiv1b1.HTTPRouteList{}
+		if err := r.client.List(ctx, routeList, &client.ListOptions{Namespace: request.Namespace}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("error listing httproutes")
+		}
+
+		if len(routeList.Items) == 0 {
+			r.resources.Namespaces.Delete(request.Namespace)
+			log.Info("deleted namespace from resource map")
+		}
 	}
```
skriss, line 176: I'm wondering if we need to do a full two-way reconcile here, where we go through all the items in the watchable map, ensure they still exist in the API server, and remove them if they don't, but for now this is probably fine, just something to keep in the back of our minds.

skriss (changes requested):

internal/provider/kubernetes/httproute.go
danehans: @skriss I created https://github.com/envoyproxy/gateway/issues/248 to track this feedback.

internal/provider/kubernetes/httproute.go
```diff
@@ -121,17 +121,58 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
-	err := r.client.Get(ctx, request.NamespacedName, httpRoute)
-	if errors.IsNotFound(err) {
-		log.V(2).Info("httproute not found, deleting it from the ResourceTable")
+	// Fetch all HTTPRoutes from the cache.
+	routeList := &gwapiv1b1.HTTPRouteList{}
+	if err := r.client.List(ctx, routeList); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error listing httproutes")
+	}
+
+	found := false
```
skriss, line 130: looks like this is never getting set to true, you do still need to keep track of if you find the item that triggered the Reconcile request so that the delete logic below only runs if that item no longer exists.

danehans: @skriss thanks for catching this. I need to slow down and double-check my work ;-) I also updated the Kube provider integration test to ensure the route's ns is being added to the watchable map.

internal/provider/kubernetes/kubernetes_test.go
```diff
@@ -57,10 +57,10 @@ func TestProvider(t *testing.T) {
 	}()
 
 	testcases := map[string]func(context.Context, *testing.T, *Provider, *message.ProviderResources){
-		"gatewayclass controller name": testGatewayClassController,
-		"gatewayclass accepted status": testGatewayClassAcceptedStatus,
-		"gateway scheduled status":     testGatewayScheduledStatus,
-		"httproute":                    testHTTPRoute,
+		//"gatewayclass controller name": testGatewayClassController,
+		//"gatewayclass accepted status": testGatewayClassAcceptedStatus,
+		//"gateway scheduled status":     testGatewayScheduledStatus,
```
skriss, line 62: guessing this was just for local testing and should be reverted?

skriss (approved):

Issue #225, opened 2022-08-23
Use strings.Builder to manage Envoy Bootstrap Config


danehans:
Use `strings.Builder` to manage the Envoy bootstrap config instead of `bytes.Buffer`

xref: https://github.com/envoyproxy/gateway/pull/206#discussion_r952197652

PR #232, opened 2022-08-24
Use strings.Builder to manage Envoy Bootstrap Config

zirain:

fixes: https://github.com/envoyproxy/gateway/issues/225


LukeShu (approved):
Thanks!

PR #234, opened 2022-08-24
update gitignore for GoLand and vscode

zirain:
update gitignore for GoLand and vscode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,10 @@
 .DS_Store
 bin/
+
+# Intellij
+*.iml
+.idea/
+.run/
+
+# Visual Studio Code
+.vscode/


LukeShu:
I guess, but if you use those, you really should be putting those in your global `~/.config/git/ignore` so you don't need to add them to every repo you contribute to.

basvanbeek:
> I guess, but if you use those, you really should be putting those in your global `~/.config/git/ignore` so you don't need to add them to every repo you contribute to.

While true, it is still a good idea to put in the repo as it will make contributing to this repo easier. Plenty of people do not have these global ignore settings and then pollute their PRs.

arkodg (approved):

skriss (approved):

Issue #235, opened 2022-08-24
Implement liveness and readiness probes on the Deployment


skriss:
There are liveness and readiness probes defined on the envoy-gateway Deployment: https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml#L40-L49

However, I don't see any code that implements these, and in my test cluster the pod is not responding to them and therefore not becoming ready / getting restarted regularly.

They need to either be implemented, or removed for now until we're ready to implement them so the pod can at least run.

skriss:
The probe definitions have been temporarily removed in #238 to unblock running EG in cluster but should eventually be re-added and implemented properly.

skriss:
> I can help on this, assign this to me plz

Great, thanks @Xunzhuo!

danehans:
@Xunzhuo are you still able to help with this issue?

PR #238, opened 2022-08-24
remove unimplemented liveness/readiness probes

skriss:
Just removing the probe definitions for now
since they're not implemented to unblock
running EG in-cluster, but we should circle
back and implement these eventually.

Updates #235.


arkodg (approved):
SGTM

Issue #241, opened 2022-08-24
Add Node Image Variable to Kind


danehans:
Add a $NODE_IMAGE variable so EG can be tested against multiple Kubernetes versions.

xref: https://github.com/envoyproxy/gateway/pull/240

PR #242, opened 2022-08-25
Add Node Image Variable to Kind

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/241

### DIFF
--- a/tools/hack/create-cluster.sh
+++ b/tools/hack/create-cluster.sh
@@ -2,17 +2,24 @@
 
 set -euo pipefail
 
-## Create kind cluster.
-tools/bin/kind create cluster \
-    --name envoy-gateway 
+# Setup default values
+CLUSTER_NAME=${CLUSTER_NAME:-"envoy-gateway"}
+METALLB_VERSION=${METALLB_VERSION:-"v0.12.1"}
+NODE_IMAGE=${NODE_IMAGE:-"kindest/node:v1.24.0"}
 
+## Create kind cluster.
+if [[ -z "${NODE_IMAGE}" ]]; then
+  tools/bin/kind create cluster --name "${CLUSTER_NAME}"
+else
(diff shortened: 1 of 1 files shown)


tools/hack/create-cluster.sh
```diff
@@ -2,17 +2,24 @@
 
 set -euo pipefail
 
-## Create kind cluster.
-tools/bin/kind create cluster \
-    --name envoy-gateway 
+# Setup default values
+CLUSTER_NAME=${CLUSTER_NAME:-"envoy-gateway"}
```
arkodg, line 6: we seem to have used `VAR ?= val` style in the Makefiles here e.g. https://github.com/envoyproxy/gateway/blob/d99f32feef8ddd6ede45bd0c93b607586138d4c8/tools/make/kube.mk#L2
prefer if we  use that style here as well

tools/hack/create-cluster.sh
```diff
@@ -2,17 +2,24 @@
...
+METALLB_VERSION=${METALLB_VERSION:-"v0.12.1"}
+NODE_IMAGE=${NODE_IMAGE:-"kindest/node:v1.24.0"}
```
arkodg, line 8: lets only make the `KIND_NODE_TAG` as the variable which corresponds to the kube version we'll be testing and fix the image to `kindest/node`

tools/hack/create-cluster.sh
zirain: 1. this lines have same style as `run-kube-local.sh`
2. is `?=` valid in shell?

tools/hack/create-cluster.sh
zirain: updated

tools/hack/create-cluster.sh
arkodg: oh yah my bad .....

arkodg (approved):

PR #245, opened 2022-08-25
Implements xDS Server Config API

danehans:
Manages Envoy xDS Server Address through the EnvoyProxy API.
- Updates the infra IR to use EnvoyProxySpec instead of EnvoyProxy.
- Updates the gatewayclass controller to manage parametersRef of accepted gatewayclass.
- Plumbs the xdsServerAddress throughout the system.

Requires: https://github.com/envoyproxy/gateway/pull/224


danehans:
Closing since we want don't want to expose config to users in v0.2.0. This can be reopendor used as a reference implementation of the EnvoyProxy API in the future.

Issue #246, opened 2022-08-25
Infra Manager Fails to Provision Infrastructure


danehans:
1. When a Kube infra type is created, the infra namespace is not set. This causes ns-scoped resources to be created without a ns, resulting in an error.
2. https://github.com/envoyproxy/gateway/pull/200 intended to mount the Envoy bootstrap config via a ConfigMap volume. However, https://github.com/envoyproxy/gateway/pull/206 added the bootstrap config using a CLI arg and did not remove the COnfigMap volume and `--config-path` arg that used the volume.

PR #247, opened 2022-08-25
Updates Kube Infra Namespace and Envoy Deployment

danehans:
- Set `Namespace` when creating a new Kube Infra manager.
- Removes the bootstrap config volume and references to the volume in the Envoy deployment.

Fixes: #246 

### DIFF
--- a/internal/infrastructure/manager.go
+++ b/internal/infrastructure/manager.go
@@ -34,6 +34,8 @@ func NewManager(cfg *config.Server) (*Manager, error) {
 		ns, found := os.LookupEnv("ENVOY_GATEWAY_NAMESPACE")
 		if found {
 			mgr.Infra.Namespace = ns
+		} else {
+			mgr.Infra.Namespace = "envoy-gateway-system"
 		}
 	} else {
 		// Kube is the only supported provider type.
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
-	"path/filepath"
 	"strings"
 	"text/template"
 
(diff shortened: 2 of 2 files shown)


internal/infrastructure/kubernetes/infra.go
```diff
@@ -35,6 +35,7 @@ func NewInfra(cli client.Client) *Infra {
 	return &Infra{
 		mu:        sync.Mutex{},
 		Client:    cli,
+		Namespace: "envoy-gateway-system",
```
arkodg, line 38: isnt this being set here https://github.com/envoyproxy/gateway/blob/2ba39476943984774a31715e7b37275b58a50534/internal/infrastructure/manager.go#L33

danehans: If the OS env is set to `ENVOY_GATEWAY_NAMESPACE`. I did not set it and we should use the default. Then the OS env can be used to override the default "envoy-gateway-system".

arkodg: sg, can we consolidate both the logic in one place using this PR

danehans: Done in https://github.com/envoyproxy/gateway/commit/80c28911a654cf93c8959eb1c128312be6174292

internal/infrastructure/manager.go
```diff
@@ -34,6 +34,8 @@ func NewManager(cfg *config.Server) (*Manager, error) {
 		ns, found := os.LookupEnv("ENVOY_GATEWAY_NAMESPACE")
 		if found {
 			mgr.Infra.Namespace = ns
```
arkodg, line 36: nit: would have preferred if this logic was living in `internal/infrastructure/kubernetes`

arkodg (approved):

skriss (approved):

Issue #159, opened 2022-07-15
add support for request redirect filter


skriss:
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter

Needed for core conformance.

AliceProxy:
I'd like to try my hand at taking this one on.

PR #249, opened 2022-08-25
Add support for redirect filters

AliceProxy:
Adds support for gateway API Redirect Filters.
This is my first PR to the repo so feedback is greatly appreciated .

Fixes: https://github.com/envoyproxy/gateway/issues/159


danehans:
@AliceProxy please rebase when you have a moment.

AliceProxy:
Sorry about the recent pushes. I thought I was done with this but realized there were some problems that I needed to fix and I added some better test coverage.

danehans:
@skriss do you mind reviewing since you have a good understanding of this code?

danehans:
@AliceProxy can you investigate the CI failure?

skriss:
Yep it's on my list to review, just need to find some time!

AliceProxy:
> @AliceProxy can you investigate the CI failure?

@danehans The most recent commit should resolve the CI failure.

internal/ir/xds.go
```diff
@@ -131,6 +131,10 @@ type HTTPRoute struct {
 	QueryParamMatches []*StringMatch
 	// Destinations associated with this matched route.
 	Destinations []*RouteDestination
+	// Redirections to be returned for this route. Takes precedence over Destinations
```
arkodg, line 134: can you update the order of the fields to reflect the priority mentioned in the comment

internal/ir/xds.go
```diff
@@ -175,6 +179,39 @@ type RouteDestination struct {
 	Weight uint32
 }
 
+// +k8s:deepcopy-gen=true
+type DirectResponse struct {
```
arkodg, line 183: can you add `Validate` methods for these new structs ?

internal/xds/translator/route.go
```diff
@@ -1,17 +1,27 @@
...
 
 import (
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
 
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
 func buildXdsRoute(httpRoute *ir.HTTPRoute) (*route.Route, error) {
```
arkodg, line 11: can you also add some tests here https://github.com/envoyproxy/gateway/blob/791408311703b662f56f26fcadf2b2e77154f34a/internal/xds/translator/translator_test.go#L26 for redirect and direct response ?

internal/gatewayapi/translator.go
```diff
@@ -546,11 +548,129 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
 
+					// TODO: implement header modifier filter
+					for _, filter := range rule.Filters {
+						if irRoute.DirectResponse != nil {
+							// If an invalid filter type has been configured then an error direct response must be returned
+							// so skip processing any other filters for this route
+							break
+						}
+						switch filter.Type {
+						case "RequestRedirect":
```
skriss, line 559: nit, there are constants for the filter types defined in Gateway API:

```suggestion
						case v1beta1.HTTPRouteFilterRequestRedirect:
```

internal/gatewayapi/translator.go
```diff
@@ -546,11 +548,129 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 						}
 					}
 
+					// TODO: implement header modifier filter
+					for _, filter := range rule.Filters {
```
skriss, line 552: It might make sense to move this out of the loop on `rule.Matches` since the same set of filters apply for all matches in a given Gateway API route rule, that way we can avoid processing the filters multiple times, once per match. But, not a big deal right now and we can always revisit.

internal/gatewayapi/translator.go
```diff
@@ -546,11 +548,129 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+							if redirect := filter.RequestRedirect; redirect != nil {
+								if redirect.Scheme != nil {
+									// Note that gateway API may support additional schemes in the future, but unknown values
+									// must result in an UnsupportedValue status
+									if *redirect.Scheme == "http" || *redirect.Scheme == "https" {
+										irRedirect.Scheme = redirect.Scheme
+									} else {
+										errMsg := fmt.Sprintf("Scheme: %s is unsupported, only 'https' and 'http' are supported", *redirect.Scheme)
+										parentRef.SetCondition(
+											v1beta1.RouteConditionResolvedRefs,
```
skriss, line 581: I see that the [Gateway API docs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter) reference setting an "Attached" condition in many of these error cases with a reason "UnsupportedValue", but that condition type doesn't exist for HTTPRoutes. I'm guessing that should actually be the "Accepted" condition, which is listed as having "UnsupportedValue" as a reason. (ref. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteConditionType).

We should open an issue/contribute a fix to Gateway API for those docs, and assuming that goes through, update these conditions to use "Accepted".

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
 	return false
 }
 
+// Checks if a hostname is valid according to RFC 1123 and gateway API's requirement that it not be an IP address
+func isValidHostname(hostname string) (valid bool, reason string) {
+	if len(hostname) == 0 || hostname == "" {
```
skriss, line 904: Contour's using https://github.com/projectcontour/contour/blob/e067e1c59b44cd7b859b6ca8b958bd0018edc789/internal/gatewayapi/listeners.go#L161-L179 to do the valid hostname check (plus the same IP address check you have here) - if that upstream function works, we could use that in place of the custom validation code here.

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
 	return false
 }
 
+// Checks if a hostname is valid according to RFC 1123 and gateway API's requirement that it not be an IP address
+func isValidHostname(hostname string) (valid bool, reason string) {
```
LukeShu, line 903: Any reason not to use [`k8s.io/apimachinery/pkg/util/validation.IsDNS1123Subdomain`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain)?

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
...
+
+	if hostname[0] == '-' || hostname[0] == '.' {
+		return false, "Hostname cannot start with '.' or '-'"
+	}
+	if hostname[len(hostname)-1] == '-' || hostname[len(hostname)-1] == '.' {
+		return false, "Hostname cannot end with '.' or '-'"
+	}
+
+	// The only allowed characters in hostnames are a-z, A-Z, 0-9, ., and -
+	var validChar = regexp.MustCompile(`[a-zA-Z0-9.-]`).MatchString
```
LukeShu, line 920: My gut reaction is that regexes are big and slow, and `strings.Contains("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-", char)` would be comparatively faster.

internal/gatewayapi/translator.go
```diff
@@ -546,11 +548,129 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+										continue
+									} else {
+										redirectHost := string(*redirect.Hostname)
+										irRedirect.Hostname = &redirectHost
+									}
+								}
+
+								if redirect.Path != nil {
+									switch redirect.Path.Type {
+									case "ReplaceFullPath":
```
skriss, line 607: nit, there are constants for the path type options upstream as well (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/httproute_types.go#L759-L775)

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
...
+	}
+	if hostname[len(hostname)-1] == '-' || hostname[len(hostname)-1] == '.' {
+		return false, "Hostname cannot end with '.' or '-'"
+	}
+
+	// The only allowed characters in hostnames are a-z, A-Z, 0-9, ., and -
+	var validChar = regexp.MustCompile(`[a-zA-Z0-9.-]`).MatchString
+
+	curLabel := 0
+	for i := 0; i < len(hostname); i++ {
```
LukeShu, line 923: Any reason not to use `for i, char := range hostname {`?

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
...
+						hostname,
+					)
+			}
+			curLabel = i + 1
+		}
+	}
+
+	// IP addresses are not allowed so parsing the hostname as an address needs to fail
+	if _, err := netip.ParseAddr(hostname); err == nil {
+		return false, fmt.Sprintf("Hostname: '%s' cannot be an ip address", hostname)
```
LukeShu, line 961: Prefer to use `%q` to handle quoting for you, instead of `'%s'`.

skriss (changes requested):
Thanks @AliceProxy, overall this looks great! Had a few initial comments, still looking through the test cases.

internal/gatewayapi/translator.go
```diff
@@ -777,6 +899,71 @@ func isValidCrossNamespaceRef(from crossNamespaceFrom, to crossNamespaceTo, refe
...
+				return false,
+					fmt.Sprintf(
+						"Label '%s' in hostname: '%s' cannot end with a '.' or '-'",
+						hostname[curLabel:i],
+						hostname,
+					)
+			}
+			curLabel = i + 1
+		}
+	}
```
LukeShu, line 957: So to sorta answer my earlier question: It looks like Kube's `IsDNS1123Subdomain` doesn't check the lengths of individual labels.  We should probably send kubenetes/kubernetes.git a PR to fix that.

But! It looks like yours doesn't quite to it right either.  If I'm reading this right, it'd allow the _last_ label to be longer than 63 chars.

internal/ir/xds.go
```diff
@@ -175,6 +179,39 @@ type RouteDestination struct {
...
+	// Body can actually support a config.core.v3.DataSource, but for now only an
+	// inline string is supported
```
LukeShu, line 200: Nit: These are complete sentences, so they should have punctuation. (I wouldn't bother with this comment on a plain comment in a function or something, but these are doc comments.)

skriss:
One other note, just for future reference - our initial goal for EG is to meet "core" Gateway API conformance. As such, I think it'd be fine to start by only implementing the parts of these filters that are required for core conformance, and save "extended" features for later (that's the approach that's been taken with most of the existing Gateway API code). It might help keep the PRs a bit smaller.

internal/gatewayapi/translator.go
```diff
@@ -546,11 +548,129 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+											v1beta1.RouteReasonUnsupportedValue,
+											errMsg,
+										)
+										continue
+									}
+								}
+
+								if redirect.Port != nil {
+									redirectPort := uint32(*redirect.Port)
+									irRedirect.Port = &redirectPort
```
LukeShu, line 650: We're ending up with a lot of silly code that looks like this. Perhaps we should add a utility function

```go
func ptrTo[T any](x T) *T {
    return &x
}
```

internal/gatewayapi/translator.go
AliceProxy: I'll fix the issue in my code and then also open a PR into kubernetes/kubetnetes.git, but for now I'll just leave a TODO comment saying that we should switch to `IsDNS1123Subdomain` once that submission lands.

internal/gatewayapi/translator.go
LukeShu: I sorta hate that this returns a `string` instead of an `error`.

AliceProxy: See my response to your other comment re: using `IsDNS1123Subdomain`

internal/gatewayapi/translator_test.go
```diff
@@ -85,6 +86,106 @@ func testName(inputFile string) string {
...
+			expected: false,
+			err:      "Invalid character: '.' in hostname: 'example..test.com' at offset: 8. hostname label cannot start with a '.'",
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := isValidHostname(tc.hostname)
+			assert.Equal(t, tc.expected, actual)
+			assert.Equal(t, tc.err, err)
```
LukeShu, line 184: At first I thought "that doesn't work, does it!? You need to use `assert.EqualError`, don't you?", then I noticed that `err` is actually a string.

arkodg:
@AliceProxy thanks for adding the tests in the IR and xDS libs ! LGTM from my end wrt code in those libs, hoping others can review the `gatewayapi/translator` diffs

internal/gatewayapi/translator.go
AliceProxy: Probably just had a brainfart thinking about indexes too much

internal/gatewayapi/translator.go
AliceProxy: Re Luke's comments about `IsDNS1123Subdomain` since it does not check label length I think PR a fix for that in `IsDNS1123Subdomain` and then switch to using that when the change is accepted would be appropriate here :thinking:

internal/gatewayapi/translator.go
arkodg: another option to consider https://github.com/istio/istio/blob/da6178604559bdf2c707a57f452d16bee0de90c8/pkg/config/labels/instance.go#L95

internal/gatewayapi/translator.go
skriss: I submitted https://github.com/kubernetes-sigs/gateway-api/pull/1382 for this

internal/gatewayapi/translator.go
AliceProxy: Actually I've switched to using `IsDNS1123Subdomain` along with the check for label length (and a test case for the last label as well)

internal/gatewayapi/translator.go
AliceProxy: Switched to `IsDNS1123Subdomain` which covers most of these checks.

internal/gatewayapi/translator.go
AliceProxy: I've actually switched to using `IsDNS1123Subdomain` along with Arko's suggestion to include a check for individual label length.

internal/gatewayapi/translator.go
AliceProxy: We don't need the wildcard check since that's invalid for a redirect, but the logic is now a combination of `IsDNS1123Subdomain` and a check on the individual label length.

internal/gatewayapi/translator.go
AliceProxy: Very true, but I kind of feel like this should be a separate PR after this just to keep the logical size smaller.

internal/gatewayapi/translator_test.go
AliceProxy: It's an `error` now :slightly_smiling_face:

internal/gatewayapi/translator.go
AliceProxy: Do you want me to change them all to `Accepted` in this PR? I'd be curious to see what the gateway API folks have to say about it first :thinking: 

I'm also wondering if `Accepted` with a status of `False` is the right move on this since the `HTTPRoute` itself would actually be accepted, just the filter may be invalid/ignored. Happy to change them in this PR if you anticipate the Gateway API folks agreeing.

skriss: Let's keep as-is for now, I'll create an issue to track the follow-up here, once we see what the Gateway API folks have to say about it. Ultimately we should do whatever is defined in the spec / the conformance tests, I think in this case there was just a typo.

skriss: https://github.com/envoyproxy/gateway/issues/357

internal/gatewayapi/testdata/httproute-with-redirect-filter-invalid-status.out.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - "*.envoyproxy.io"
+      routes:
+        - name: default-httproute-1-rule-0-match-0-gateway.envoyproxy.io
+          pathMatch:
+            prefix: "/"
+          headerMatches:
+            - name: ":authority"
+              exact: gateway.envoyproxy.io
+          # I believe the correct way to handle an invalid filter should be to allow the HTTPRoute to function
+          # normally but leave out the filter config and set the status, but this behaviour can be changed.
```
skriss, line 85: Yeah, I think the expected behavior here is unclear in the Gateway API spec -- we can drive some clarity upstream, but doesn't need to block this PR now.

skriss (approved):
LGTM! Will leave for other reviewers to merge when ready.

PR #250, opened 2022-08-25
Fix xds server port

arkodg:
* fix xds-server's port within the runner to 18000
* make sure the envoy bootstrap config within the infra library
uses the same constant from the xds server runner
* moved vars to const within the kube infra library and renamed some

### DIFF
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/utils/pointer"
 
 	"github.com/envoyproxy/gateway/internal/ir"
+	xdsrunner "github.com/envoyproxy/gateway/internal/xds/server/runner"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -35,24 +36,19 @@ const (
 	envoyHTTPPort = int32(8080)
 	// envoyHTTPSPort is the container port number of Envoy's HTTPS endpoint.
 	envoyHTTPSPort = int32(8443)
+	// envoyGatewayXdsServerHost is the address of the Xds Server within Envoy Gateway.
```
danehans, line 39: `envoyGatewayXdsServerHost is the address` seems confusing. I suggest explaining in the docs that it's either an IP address or DNS 1123 label name. We should also mention that the default value is the name of the Envoy Gateway Kubernetes Service.

danehans (changes requested):
I small nit.

internal/infrastructure/kubernetes/deployment.go
arkodg: since its set to https://github.com/envoyproxy/gateway/blob/13c1517bb07eb3c536fdd076bbd330a3521ce5ba/internal/infrastructure/kubernetes/bootstrap.yaml.tpl#L43 it must be a hostname, I'll update the comment

danehans (approved):

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -35,24 +36,20 @@ const (
 	envoyHTTPPort = int32(8080)
 	// envoyHTTPSPort is the container port number of Envoy's HTTPS endpoint.
 	envoyHTTPSPort = int32(8443)
+	// envoyGatewayXdsServerHost is the DNS name of the Xds Server within Envoy Gateway.
+	// It defaults to the Envoy Gateway Kubernetes service.
+	envoyGatewayXdsServerHost = "envoy-gateway"
+	// envoyAdminAddress is the listening address of the envoy admin interface.
+	envoyAdminAddress = "127.0.0.1"
```
zirain, line 43: should this be configurable, right now it's hard to debug via `config_dump`

danehans: @zirain we discussed making the bootstrap configurable post v0.2.0. Let me add it to the agenda for the community meeting.

PR #252, opened 2022-08-26
fix: sync unmatched eg namespace

Xunzhuo:
for `internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml`, 

ns and deployment are unmatched, it will cause deployment started failed.

and other files in `internal/provider/kubernetes/config/`, ns is not synced too, some called `system` and some called `envoy-gateway-system`.

Signed-off-by: Xunzhuo <mixdeers@gmail.com>

### DIFF
--- a/internal/provider/kubernetes/config/crd/patches/webhook_in_envoyproxies.yaml
+++ b/internal/provider/kubernetes/config/crd/patches/webhook_in_envoyproxies.yaml
@@ -9,7 +9,7 @@ spec:
     webhook:
       clientConfig:
         service:
-          namespace: system
+          namespace: envoy-gateway-system
           name: webhook-service
           path: /convert
       conversionReviewVersions:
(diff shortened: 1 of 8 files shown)


danehans:
@Xunzhuo please see how https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/config/default/kustomization.yaml properly sets the namespace.

Xunzhuo:
Yes I got it, but why we set these different namespaces in default?

danehans:
The [Kustomize default namespace ](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/config/default/kustomization.yaml#L2)sets the namespace of all Kube resources. If we were to accept this PR, the Kustomize default namespace should be removed. I prefer using the Kustomize default namespace b/c 1) keeps the logic consistent with upstream Kubebuilder and 2) fewer places to update the namespace if needed.

Xunzhuo:
Thanks @danehans, I will close this.

Issue #261, opened 2022-08-26
Envoy ServiceAccount Fails to Get Created


danehans:
Infra Manager fails to provision the Envoy ServiceAccount resource due to RBAC restrictions.

```
2022-08-26T19:47:09.161Z	ERROR	runner/runner.go:53	failed to create new infra	{"runner": "infrastructure", "error": "failed to get serviceaccount envoy-gateway-system/eg: serviceaccounts \"eg\" is forbidden: User \"system:serviceaccount:envoy-gateway-system:envoy-gateway\" cannot get resource \"serviceaccounts\" in API group \"\" in the namespace \"envoy-gateway-system\""}
```

PR #264, opened 2022-08-26
Updates RBAC for Infra Manager Permissions

danehans:
Updates RBAC for Infra Manager to manage Envoy resources.

Fixes: #261 

### DIFF
--- a/internal/provider/kubernetes/rbac.go
+++ b/internal/provider/kubernetes/rbac.go
@@ -3,4 +3,7 @@ package kubernetes
 // +kubebuilder:rbac:groups="gateway.networking.k8s.io",resources=gatewayclasses;gateways;httproutes;referencepolicies;referencegrants,verbs=get;list;watch
 // +kubebuilder:rbac:groups="gateway.networking.k8s.io",resources=gatewayclasses/status;gateways/status;httproutes/status,verbs=update
 
-// +kubebuilder:rbac:groups="",resources=secrets;services;namespaces,verbs=get;list;watch
+// RBAC for Infra Manager to manage Envoy.
+// +kubebuilder:rbac:groups="",resources=secrets;namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=services;serviceaccounts,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;delete
--- a/internal/provider/kubernetes/config/rbac/role.yaml
+++ b/internal/provider/kubernetes/config/rbac/role.yaml
@@ -10,10 +10,32 @@ rules:
   resources:
   - namespaces
   - secrets
-  - services
   verbs:
   - get
   - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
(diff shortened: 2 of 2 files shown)


Xunzhuo (dismissed):
LGTM

internal/provider/kubernetes/rbac.go
```diff
@@ -3,4 +3,7 @@ package kubernetes
 // +kubebuilder:rbac:groups="gateway.networking.k8s.io",resources=gatewayclasses;gateways;httproutes;referencepolicies;referencegrants,verbs=get;list;watch
 // +kubebuilder:rbac:groups="gateway.networking.k8s.io",resources=gatewayclasses/status;gateways/status;httproutes/status,verbs=update
 
-// +kubebuilder:rbac:groups="",resources=secrets;services;namespaces,verbs=get;list;watch
+// RBAC for Infra Manager to manage Envoy.
```
arkodg, line 6: should these live inside `internal/infrastructure/kubernetes` ?

skriss: Personally I like having them all in one place, but I can live with whatever.

internal/provider/kubernetes/rbac.go
```diff
@@ -3,4 +3,7 @@ package kubernetes
...
+// +kubebuilder:rbac:groups="",resources=secrets;namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=services;serviceaccounts,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;delete
```
skriss, line 9: Do we want to restrict some of these just to the default `envoy-gateway-system` namespace since that's where we're creating all infra? Feels better than giving CRUD access across all namespaces. It does mean that if a user wants to deploy EG into an alternate namespace, they'll need to customize the generated RBAC, but to me that's a better trade-off than not restricting them.

arkodg (dismissed):

danehans:
e2e testing fails with commit 4e2d2f79602316f742f31287f3d156e6b242fd27. Since RC1 is scheduled to release tomorrow and will be used for testing only, I'm going to revert this PR and have created https://github.com/envoyproxy/gateway/issues/291 to track this requirement for v0.2.0.

danehans:
@arkodg @skriss PTAL at https://github.com/envoyproxy/gateway/pull/264#issuecomment-1232021307. We need this PR merged for RC1.

internal/provider/kubernetes/rbac.go
danehans: https://github.com/envoyproxy/gateway/pull/292 takes the approach of splitting RBAC between Kube provider and Kube infra manager.

internal/provider/kubernetes/rbac.go
danehans: @skriss https://github.com/envoyproxy/gateway/pull/292 restricts Infra Manager RBAC to a Role within the envoy-gateway-system namespace instead of a ClusterRole. The Infra Manager RBAC is static but that should do for v0.2.0 unless someone wants to pick up https://github.com/envoyproxy/gateway/issues/289.

skriss (approved):
OK, let's just circle back and tackle #291 before 0.2.0.

danehans:
> OK, let's just circle back and tackle https://github.com/envoyproxy/gateway/issues/291 before 0.2.0.

Will do. I retagged #291 for rc2.

Issue #189, opened 2022-08-02
Add Service Watch to HTTPRoute Controller


arkodg:
*Description*:
Add a Service controller to the Kubernetes Provider package similar to the [HTTP Route Controller](https://github.com/envoyproxy/gateway/pull/179) that will be needed by the [Gateway API Translator](https://github.com/envoyproxy/gateway/blob/c83b7b04530f1cbdeb4e9596d0781eb8a2e1317b/internal/gatewayapi/translator.go#L23)

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

danehans:
The Kube provider should not need a service controller. Since Services are a dep of HTTPRoutes, the httproute controller should include a watch for services. When a service is CRUD'ed, an HTTPRoute reconciliation will be triggered by the httproute controller. The controller should get/list the necessary objects, e.g. Gateways, Services, etc required to call the translate func or pub the translate message. Post v0.2.0, we should consider an approach to filter the services to reduce the reconciliation churn.

arkodg:
reopened this issue since we haven't yet reconciled and stored the service resources in the provider resources message

skriss:
xref https://github.com/envoyproxy/gateway/pull/230#discussion_r953151075, need to decide on the pattern for this

PR #270, opened 2022-08-28
Adds HTTPRoute BackendRefs to Resource Map

danehans:
Adds HTTPRoute backendRefs to the watchable resource map.
- Adds backendRef validation to HTTPRoute controller. Note that only local Service object references are supported.
- Adds valid backendRefs to the watchable resource map.
- Updates Kube integration tests to ensure that valid backendRefs are added to the watchable resource map.
- Adds backendRefs, e.g. Services, to the gatewayapi runner.
- Adds  Services object type to ProviderResources and updates unit tests accordingly.

Fixes: https://github.com/envoyproxy/gateway/issues/189

### DIFF
--- a/internal/provider/kubernetes/httproute.go
+++ b/internal/provider/kubernetes/httproute.go
@@ -158,6 +158,37 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
 		// The route's namespace exists, so add it to the resource map.
 		r.resources.Namespaces.Store(nsKey.Name, ns)
 		log.Info("added namespace to resource map")
+
+		// Get the route's backendRefs from the cache. Note that a Service is the
+		// only supported kind.
+		for i := range route.Spec.Rules {
+			for j := range route.Spec.Rules[i].BackendRefs {
+				ref := route.Spec.Rules[i].BackendRefs[j]
+				if err := validateBackendRef(&ref); err != nil {
+					return reconcile.Result{}, fmt.Errorf("invalid backendRef: %w", err)
+				}
+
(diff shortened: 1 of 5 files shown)


internal/provider/kubernetes/httproute.go
```diff
@@ -158,21 +158,55 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+		// Delete the Namespace and Service from the resource maps if no other
+		// routes exist in the namespace.
 		routeList = &gwapiv1b1.HTTPRouteList{}
 		if err := r.client.List(ctx, routeList, &client.ListOptions{Namespace: request.Namespace}); err != nil {
 			return reconcile.Result{}, fmt.Errorf("error listing httproutes")
 		}
 		if len(routeList.Items) == 0 {
 			r.resources.Namespaces.Delete(request.Namespace)
 			log.Info("deleted namespace from resource map")
+			r.resources.Services.Delete(request.NamespacedName)
```
arkodg, line 208: if this backend is referenced by 2 HTTPRoutes and 1 HTTPRoute is deleted, this will cause the service to be deleted which is not what we want, same goes for the Namespace deletion in the above line

danehans: @arkodg the services and namespace maps are only deleted when no other routes exist.

arkodg: my bad, just saw the `if len(routeList.Items) == 0 {` check

arkodg (approved):

zirain:
ping @arkodg please Merge this to main

Issue #220, opened 2022-08-22
Fix flaky TestProvider testcase


arkodg:
*Description*:
Seeing `make go.test.coverage` sometimes fail in  `TestProvider` 

```
--- FAIL: TestProvider (20.30s)
[133](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:134)
    --- FAIL: TestProvider/gateway_of_gatewayclass (10.05s)
[134](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:135)
        kubernetes_test.go:200: 
[135](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:136)
            	Error Trace:	/home/runner/work/gateway/gateway/internal/provider/kubernetes/kubernetes_test.go:200
[136](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:137)
            	            				/home/runner/work/gateway/gateway/internal/provider/kubernetes/asm_amd64.s:1571
[137](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:138)
            	Error:      	Not equal: 
[138](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:139)
            	            	expected: 1
[139](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:140)
            	            	actual  : 2
[140](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:141)
            	Test:       	TestProvider/gateway_of_gatewayclass
[141](https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true#step:5:142)
        kubernetes_test.go:200:
```
Raising this issue to track and fix this

Failed run https://github.com/envoyproxy/gateway/runs/7962066333?check_suite_focus=true

danehans:
The tests are being run in parallel. Since each test creates a gatewayclass, more than one may exist in the map. I enabled logging and ran the tests:
```
2022-09-02T12:26:49.733-0700	INFO	manager/internal.go:585	Wait completed, proceeding to shutdown the manager
--- FAIL: TestProvider (10.33s)
    --- FAIL: TestProvider/gatewayclass_controller_name (0.03s)
        kubernetes_test.go:109: gatewayclass: httproute-test
[...]

PR #271, opened 2022-08-29
Adds Delete Support to Infra Manager

danehans:
Deletes managed proxy infra when `ir.Infra` is nil.

Fixes: #220 

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/gatewayapi"
+	"github.com/envoyproxy/gateway/internal/ir"
 	"github.com/envoyproxy/gateway/internal/message"
 )
 
@@ -69,11 +70,14 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		// gateway class linked to this controller
 		switch {
 		case gatewayClasses == nil:
+			// Envoy Gateway startup.
+			continue
+		case gatewayClasses[0] == nil:
 			// No need to translate, publish empty IRs, e.g. delete operation.
 			r.XdsIR.Delete(r.Name())
-			r.InfraIR.Delete(r.Name())
-		case gatewayClasses[0] == nil:
-			panic("gatewayclass is nil")
+			// A nil ProxyInfra tells the Infra Manager to delete the managed proxy infra.
+			infra := &ir.Infra{Proxy: new(ir.ProxyInfra)}
+			r.InfraIR.Store(r.Name(), infra)
 		default:
 			// Translate and publish IRs.
 			t := &gatewayapi.Translator{
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -76,8 +76,7 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 			// No need to translate, publish empty IRs, e.g. delete operation.
 			r.XdsIR.Delete(r.Name())
 			// A nil ProxyInfra tells the Infra Manager to delete the managed proxy infra.
-			infra := &ir.Infra{Proxy: new(ir.ProxyInfra)}
-			r.InfraIR.Store(r.Name(), infra)
+			r.InfraIR.Store(r.Name(), &ir.Infra{Proxy: nil})
 		default:
 			// Translate and publish IRs.
 			t := &gatewayapi.Translator{
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -19,6 +19,8 @@ import (
 )
 
 const (
+	// envoyDeploymentName is the name of the Envoy Deployment resource.
+	envoyDeploymentName = "envoy"
 	// envoyContainerName is the name of the Envoy container.
 	envoyContainerName = "envoy"
 	// envoyNsEnvVar is the name of the Envoy Gateway namespace environment variable.
@@ -93,7 +95,7 @@ func (b *bootstrapConfig) render() error {
 // createDeploymentIfNeeded creates a Deployment based on the provided infra, if
 // it doesn't exist in the kube api server.
 func (i *Infra) createDeploymentIfNeeded(ctx context.Context, infra *ir.Infra) error {
-	current, err := i.getDeployment(ctx, infra)
+	current, err := i.getDeployment(ctx)
 	if err != nil {
 		if kerrors.IsNotFound(err) {
 			deploy, err := i.createDeployment(ctx, infra)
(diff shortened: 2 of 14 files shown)


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -242,6 +242,24 @@ func (i *Infra) createDeployment(ctx context.Context, infra *ir.Infra) (*appsv1.
...
 }
 
+// deleteService deletes the Envoy Deployment in the kube api server, if it exists.
+func (i *Infra) deleteDeployment(ctx context.Context) error {
+	deploy := new(appsv1.Deployment)
+	key := types.NamespacedName{
+		Namespace: i.Namespace,
+		Name:      envoyDeploymentName,
+	}
+	err := i.Client.Get(ctx, key, deploy)
```
arkodg, line 252: thoughts on simplifying this to just calling `i.Client.Delete` ? if I had to guess, the server would return with a similar non found error, in case the deployment doesn't exist

internal/infrastructure/runner/runner.go
```diff
@@ -45,8 +45,10 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 		r.Logger.Info("received a notification")
 		in := r.InfraIR.Get()
 		if in == nil {
-			r.Logger.Info("ir is nil, skipping")
-			continue
```
arkodg, line 49: For the empty infra case, I think the IR producer should generate a key with nil value, rather than setting an empty map
This current implementation is going to result in this code trying to delete the infra during init/start which will fail because the deployment doesn't exist

internal/infrastructure/kubernetes/infra.go
```diff
@@ -103,3 +103,20 @@ func (i *Infra) CreateInfra(ctx context.Context, infra *ir.Infra) error {
...
+func (i *Infra) DeleteInfra(ctx context.Context, infra *ir.Infra) error {
+	if err := ir.ValidateInfra(infra); err != nil {
+		return err
+	}
+
+	if infra != nil {
+		return errors.New("infra ir is not nil")
+	}
+
+	if err := i.deleteServiceAccount(ctx); err != nil {
```
arkodg, line 117: guessing `deleteDeployment` & `deleteService` is required here

internal/provider/kubernetes/gatewayclass.go
```diff
@@ -114,9 +114,10 @@ func (r *gatewayClassReconciler) Reconcile(ctx context.Context, request reconcil
 		r.resources.GatewayClasses.Store(acceptedGC.GetName(), acceptedGC)
 	}
 
-	// no controlled gatewayclasses, trigger a delete
+	// No controlled gatewayclasses. Store a nil gatewayclass to trigger a delete.
 	if len(cc.matchedClasses) == 0 {
 		r.log.Info("failed to find gatewayclass", "name", request.Name)
+		r.resources.GatewayClasses.Store(request.Name, nil)
```
arkodg, line 120: if we are always going to set a key, should L-109 be deleted ?
```
	if !found {
		r.resources.GatewayClasses.Delete(request.Name)
	}
```

internal/gatewayapi/runner/runner.go
```diff
@@ -69,11 +70,14 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
 		case gatewayClasses == nil:
+			// Envoy Gateway startup.
+			continue
+		case gatewayClasses[0] == nil:
 			// No need to translate, publish empty IRs, e.g. delete operation.
 			r.XdsIR.Delete(r.Name())
-			r.InfraIR.Delete(r.Name())
-		case gatewayClasses[0] == nil:
-			panic("gatewayclass is nil")
+			// A nil ProxyInfra tells the Infra Manager to delete the managed proxy infra.
```
arkodg, line 78: the comment says `nil` ProxyInfra but the code checks for `nil` Listeners
```
if in.Proxy.Listeners != nil {
```
I prefer having a `nil` Proxy and checking for that in the next stage

internal/provider/kubernetes/gatewayclass.go
danehans: Removes in commit 47fba0d.

internal/gatewayapi/runner/runner.go
danehans: Updated in commit 47fba0d.

danehans:
@arkodg thanks for the review. I resolved your feedback in commit https://github.com/envoyproxy/gateway/commit/47fba0dcabe263ffa1369d852799f16952bc2bed, PTAL when you have a moment.

internal/infrastructure/kubernetes/service.go
```diff
@@ -116,3 +116,21 @@ func (i *Infra) createService(ctx context.Context, infra *ir.Infra) (*corev1.Ser
...
 }
+
+// deleteService deletes the Envoy Service in the kube api server, if it exists.
+func (i *Infra) deleteService(ctx context.Context) error {
+	svc := new(corev1.Service)
+	key := types.NamespacedName{
+		Namespace: i.Namespace,
+		Name:      envoyServiceName,
+	}
+	err := i.Client.Get(ctx, key, svc)
```
arkodg, line 127: do you want to upate this too to look like `deleteDeployment` which executes ` i.Client.Delete` and checks `kerrors.IsNotFound(err)` to decide ret err ?

internal/infrastructure/kubernetes/serviceaccount.go
```diff
@@ -86,3 +90,22 @@ func (i *Infra) createServiceAccount(ctx context.Context, infra *ir.Infra) (*cor
...
+
+// deleteServiceAccount deletes the Envoy ServiceAccount in the kube api server,
+// if it exists.
+func (i *Infra) deleteServiceAccount(ctx context.Context) error {
+	sa := new(corev1.ServiceAccount)
+	key := types.NamespacedName{
+		Namespace: i.Namespace,
+		Name:      envoyServiceAccountName,
+	}
+	err := i.Client.Get(ctx, key, sa)
```
arkodg, line 102: same as above

internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml
```diff
@@ -28,8 +28,8 @@ spec:
         runAsNonRoot: true
       containers:
       - name: envoy-gateway
-        image: envoyproxy/gateway-dev:latest
-        imagePullPolicy: Always
+        image: docker.io/envoyproxy/gateway-dev:981e90e
```
arkodg, line 31: lets revert ?

LukeShu:
Echoing Arko's comments. Otherwise, lgtm.

internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml
danehans: Will do. I created https://github.com/envoyproxy/gateway/issues/351 so this type of change doesn't get accidentally merged.

arkodg (approved):
LGTM, thanks for incorporating the comments

Issue #267, opened 2022-08-26
Envoy Gateway Fails to Add xDS Listener


danehans:
## Description
After running Envoy Gateway with the latest inflight PRs, Envoy Gateway properly runs and provisioning Envoy. However, after the initial xDS connection, Envoy Gateway fails to add the listener.

## EG Logs
```
2022-08-26T20:47:12.013Z	INFO	cache/logrwrapper.go:24	nodeID "envoy-eg-6f568884ff-5hs8l" requested type.googleapis.com/envoy.config.listener.v3.Listener[] and known map[]. Diff []	{"runner": "xds-server"}
2022-08-26T20:47:12.013Z	INFO	cache/logrwrapper.go:24	respond type.googleapis.com/envoy.config.listener.v3.Listener[] version "" with version "1"	{"runner": "xds-server"}
2022-08-26T20:47:12.013Z	INFO	cache/logrwrapper.go:24	Sending Response on stream 3 to node envoy-eg-6f568884ff-5hs8l	{"runner": "xds-server"}
2022-08-26T20:47:12.018Z	INFO	cache/logrwrapper.go:24	Got a new request, version_info , response_nonce 1105, nodeID envoy-eg-6f568884ff-5hs8l, node_version v1.24.0	{"runner": "xds-server"}
2022-08-26T20:47:12.018Z	INFO	cache/logrwrapper.go:24	handling v3 xDS resource request, version_info , response_nonce 1105, nodeID envoy-eg-6f568884ff-5hs8l, node_version v1.24.0, resource_names [], type_url type.googleapis.com/envoy.config.listener.v3.Listener, errorCode 13, errorMessage Error adding/updating listener(s) listener_default-eg-http_8080: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
	{"runner": "xds-server"}
```
## Envoy Logs
```
[2022-08-26 20:47:29.419][1][warning][config] [source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) listener_default-eg-http_8080: Didn't find a registered implementation for 'envoy.filters.http.router' with type URL: ''
```

## Reproducer
1. Create a topic branch with the latest inflight PRs.
2. Run `make create-cluster` to create a kind test cluster.
3. Run `make kube-install-image` to build a local EG image and install it in the Kind cluster.
[...]

PR #273, opened 2022-08-29
fix Envoy Gateway Fails to Add xDS Listener

zirain:
fixes: #267 

cc @danehans

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -5,6 +5,7 @@ import (
 
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	router "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/router/v3"
 	hcm "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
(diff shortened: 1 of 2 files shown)


internal/xds/translator/listener.go
```diff
@@ -30,7 +37,8 @@ func buildXdsListener(httpListener *ir.HTTPListener) (*listener.Listener, error)
 		},
 		// Use only router.
 		HttpFilters: []*hcm.HttpFilter{{
-			Name: wellknown.Router,
+			Name:       wellknown.Router,
```
lizan, line 40: in fact name can be omit because it isn't needed

lizan (approved):

internal/xds/translator/listener.go
zirain: yes, prefer to keep it for better reading

arkodg (approved):
🚀

internal/xds/translator/listener.go
```diff
@@ -30,7 +37,8 @@ func buildXdsListener(httpListener *ir.HTTPListener) (*listener.Listener, error)
...
+			ConfigType: &hcm.HttpFilter_TypedConfig{TypedConfig: routerAny},
```
arkodg, line 41: @lizan @zirain is this tied to a new API change or API implementation change ? I didn't need this earlier ?

zirain: IIRC, it's here https://www.envoyproxy.io/docs/envoy/v1.22.5/version_history/v1.22.0

image

Issue #278, opened 2022-08-30
XDS IR Logging


danehans:
It would be helpful for troubleshooting for the xds translator to log the input xds IR before translating.

PR #279, opened 2022-08-30
Adds Initial XDS IR Logging

danehans:
Fixes: #278 

### DIFF
--- a/internal/xds/translator/translator.go
+++ b/internal/xds/translator/translator.go
@@ -7,6 +7,7 @@ import (
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	"github.com/go-logr/logr"
 	"github.com/tetratelabs/multierror"
 
 	"github.com/envoyproxy/gateway/internal/ir"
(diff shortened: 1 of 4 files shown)


danehans:
Marked as draft to avoid merge conflicts with higher priority PRs.

internal/xds/translator/translator.go
```diff
@@ -7,17 +7,19 @@ import (
...
+	"github.com/go-logr/logr"
 	"github.com/tetratelabs/multierror"
 
 	"github.com/envoyproxy/gateway/internal/ir"
 	"github.com/envoyproxy/gateway/internal/xds/types"
 )
 
 // Translate translates the XDS IR into xDS resources
-func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
+func Translate(log logr.Logger, ir *ir.Xds) (*types.ResourceVersionTable, error) {
```
arkodg, line 18: instead of changing the API, the IRs (xds and infra) could be printed as a yaml string in the gateway api runner

danehans: I created #344 so we can make a decision on how to pass a logger.

arkodg: @danehans thanks for raising https://github.com/envoyproxy/gateway/issues/344 but its not related to my comment. I dont think we should be logging in libraries, instead we should we logging inside the clients consuming the libraries, in this case its the gateway-api runner (similar to what is done here https://github.com/envoyproxy/gateway/pull/302/commits/991f12caad34c6812188e1c85fe04aef8780b2d5)

danehans:
closed in favor of https://github.com/envoyproxy/gateway/pull/302.

Issue #282, opened 2022-08-30
Incorrect Envoy Deployment and Service Labels


danehans:
Currently, the Envoy Deployment and Service use the [same labels](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml#L5) as Envoy Gateway. This causes requests intended for Envoy to get kube-proxied between Envoy-gateway and envoy pods.

zirain:
I have same issue during debug https://github.com/envoyproxy/gateway/issues/267.

I add `component: controller` on envoy-gateway and `component: envoy` on envoy-eg, but I do not think it's a good practice.

IMHO, the naming of the two components are not that good,

PR #283, opened 2022-08-30
Updates Envoy Label K/V Pair

danehans:
Fixes: #282 

### DIFF
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -258,5 +258,5 @@ func EnvoyPodSelector() *metav1.LabelSelector {
 
 // envoyLabels returns the labels used for Envoy.
 func envoyLabels() map[string]string {
-	return map[string]string{"control-plane": "envoy-gateway"}
+	return map[string]string{"app": "envoy"}
 }


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -258,5 +258,5 @@ func EnvoyPodSelector() *metav1.LabelSelector {
 
 // envoyLabels returns the labels used for Envoy.
 func envoyLabels() map[string]string {
-	return map[string]string{"control-plane": "envoy-gateway"}
+	return map[string]string{"app": "envoy"}
```
arkodg, line 261: thoughts on `app : envoyproxy` for DP and `app: envoygateway` for CP ?

danehans: @arkodg @zirain both of you provide viable approaches to labeling. Since we are scheduled to release RC1 tomorrow and this fixes a blocking bug, I created https://github.com/envoyproxy/gateway/issues/290 to establish label conventions after the release.

arkodg (approved):
thanks for raising https://github.com/envoyproxy/gateway/issues/290
sgtm for now

PR #286, opened 2022-08-30
translator: listener hostnames unset

zirain:
fixes: #272
depends on: #276

### DIFF
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -692,6 +692,9 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 				}
 
 				irListener := xdsIR.GetListener(irListenerName(listener))
+				if len(irListener.Hostnames) == 0 {
+					irListener.Hostnames = hosts
+				}
 				irListener.Routes = append(irListener.Routes, perHostRoutes...)
 
 				// Theoretically there should only be one parent ref per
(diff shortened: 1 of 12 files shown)


internal/gatewayapi/translator.go
```diff
@@ -692,6 +692,9 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 				}
 
 				irListener := xdsIR.GetListener(irListenerName(listener))
+				if len(irListener.Hostnames) == 0 {
```
arkodg, line 695: unsure why this condition is not being met via https://github.com/envoyproxy/gateway/blob/a958828855fed980416acfb77a53c70b4cbcc67f/internal/gatewayapi/translator.go#L429
@skriss can you help review this one

skriss: Ah, in Gateway API the absense of any hostname means "match all hostnames" (see https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Listener -- "when unspecified, all hostnames are matched"). We should probably add an else clause after L429 to explicitly add the `*` to the IR Listener if no hostname is specified on the Gateway API listener.

arkodg: @zirain can you incorporate steve's comments (add `else` after L429) as well as add some comments in the code stating the WHY. TIA !

danehans: @skriss @arkodg since this PR is required for RC1, I created https://github.com/envoyproxy/gateway/issues/293 to track this issue. Hopefully, we can get the issue fixed in RC1, but e2e is functional without it.

danehans (approved):
@zirain thanks for the PR. If you have time please resolve https://github.com/envoyproxy/gateway/issues/293. Otherwise, I'll

Issue #280, opened 2022-08-30
Configurable Envoy Access Logging


danehans:
Currently, Envoy access logs are sent to `/dev/null`. Users should have the ability to configure Envoy access logging through the EnvoyProxy API type.

danehans:
In the meantime, do the following to enable access logging:
```
--- a/internal/infrastructure/kubernetes/bootstrap.yaml.tpl
+++ b/internal/infrastructure/kubernetes/bootstrap.yaml.tpl
@@ -1,5 +1,8 @@
 admin:
-  access_log_path: /dev/null
+  access_log:
+  - name: envoy.access_loggers.stdout
+    typed_config:
+      "@type": type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog
   address:
     socket_address:
       address: {{ .AdminServerAddress }}
```

arkodg:
@danehans imho EG should propose a sane default such as ^ and any mod should be addressed via https://github.com/envoyproxy/gateway/issues/31 / https://github.com/envoyproxy/gateway/issues/24

zirain:
for a short term, make bootstrap configurable seems enough.

for a long term, we may need something like [Telemetry](https://istio.io/latest/docs/reference/config/telemetry/) api in istio?

lizan:
@danehans that is only for the admin endpoint, the other logs is configured via xDS component.

zirain:
fixed by https://github.com/envoyproxy/gateway/pull/1407 and https://github.com/envoyproxy/gateway/pull/1464

PR #287, opened 2022-08-30
accesslog: make admin accesslog path configurable

zirain:
part of https://github.com/envoyproxy/gateway/issues/280

listener access log will be implement later


danehans:
@zirain since this is not an RC1 release blocker, I'm tagging as RC2. Please be patient with reviews until RC1 is released.

zirain:
I'm OK with that.

arkodg (approved):
🚀

Issue #297, opened 2022-08-30
Create Release GitHub Action


danehans:
Create a GH action that:

1. Pushes the tagged EG image to Docker hub.
2. Updates quickstart manifests, etc from main branch to release tag.

danehans:
@arkodg @AliceProxy can one of you please assign yourself and have a PR pushed by 8/31/22 noon PT?

Xunzhuo:
I can do this, you can assign this to me.

PR #299, opened 2022-08-30
feat(ci): add EG release workflows

Xunzhuo:
Fixes: #297

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,39 @@
+name: Release
+on:
+  create
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./tools/github-actions/setup-deps
+
+      # lint
+      - run: make lint-deps
+      - run: make -k lint
+
+  release:
+    runs-on: ubuntu-latest
+    needs: [ lint ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./tools/github-actions/setup-deps
+
+      # build
+      - run: make build-multiarch
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+
+      # test
+      - run: make go.test.unit
+
+      # push
+      - name: Login to DockerHub
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - run: make image-release
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,44 +1,32 @@
 name: Release
+
 on:
-  create
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:        
+      - v*
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: ./tools/github-actions/setup-deps
-
-      # lint
-      - run: make lint-deps
-      - run: make -k lint
-
   release:
     runs-on: ubuntu-latest
-    needs: [ lint ]
     steps:
       - uses: actions/checkout@v3
       - uses: ./tools/github-actions/setup-deps
 
-      - name: Extract Tag name
-        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+      - name: Extract Tag value
         id: tag_env
         shell: bash
         run: |
           echo "##[set-output name=version;]$(echo ${GITHUB_REF##*/})"
 
-      # test
-      - run: make go.test.unit
-
       # build and push image
       - name: Login to DockerHub
-        if: ${{ startsWith(github.ref, 'refs/tags/') }}
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
       - name: Build and Push EG Image
-        if: ${{ startsWith(github.ref, 'refs/tags/') }}
         env:
           EG_TAG_VERSION: ${{ steps.tag_env.outputs.version}}
         run: make image.push.multiarch TAG=${EG_TAG_VERSION} IMAGE_NAME=gateway PLATFORMS="linux_amd64 linux_arm64"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,8 +7,19 @@ on:
       - v*
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


danehans:
cc: @arkodg @AliceProxy

.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
+name: Release
```
arkodg, line 1: @lizan are you okay with this approach or prefer the release push to be part of https://github.com/envoyproxy/gateway/blob/main/.github/workflows/build_and_test.yaml to reduce dup ?

lizan: This is fine. Though we don't need to run all build steps, instead of that, just use some image tool (e.g. `skopeo` to copy `envoyproxy/gateway-dev:<git_sha>` to `envoyproxy/gateway:<git_tag>`, and upload manifests generated by #300 to github release.

arkodg: I prefer if we modify the existing workflow file and add another section similar to `push` in https://github.com/envoyproxy/gateway/blob/b77b8789518027308474898defa4227e562800d5/.github/workflows/build_and_test.yaml#L45 rather than pushing to the registry and pulling again (using an intermediate tool like `skopeo`) on a different runner.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - run: make go.test.unit
+
+      # push
+      - name: Login to DockerHub
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - run: make image-release
```
arkodg, line 37: prefer lesser targets to remember, can we use
```
TAG=${VERSION} make image.push.multiarch
```
instead ?

.github/workflows/release.yaml
lizan: It's not pushing to the registry and pull again, the build/push happens when a commit merged (which is the current build_and_test). A tag workflow should just do the tag.

.github/workflows/release.yaml
Xunzhuo: I am looking into this

.github/workflows/release.yaml
Xunzhuo: Yes, the release workflow is triggered by tag creating not commit merged. It makes sense to separate it with build_and_test workflow.

Xunzhuo: One more thing needs to confirm, do we need to keep make lint and make go unit test in releasing workflows? @lizan

Xunzhuo:
Workflow tested in https://github.com/Xunzhuo/gateway/runs/8104661961?check_suite_focus=true
Image build and push succesfully:
image

.github/workflows/release.yaml
```diff
@@ -0,0 +1,46 @@
+name: Release
+on:
```
arkodg, line 2:
```
on:
  push:
    # Sequence of patterns matched against refs/tags
    tags:        
      - v*
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  release:
+    runs-on: ubuntu-latest
+    needs: [ lint ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./tools/github-actions/setup-deps
+
+      - name: Extract Tag name
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
```
arkodg, line 24: if no longer needed if this GHA is only triggered on tag push

.github/workflows/release.yaml
danehans: I see the lint and unit test targets in the PR. I'm fine with keeping them.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      # build and push image
+      - name: Login to DockerHub
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Build and Push EG Image
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
```
arkodg, line 41: same as above

.github/workflows/release.yaml
arkodg: its a dup with build_and_test.yaml, but should be okay in short term and can be deduped later

Xunzhuo: Ok, let me remove this.

.github/workflows/release.yaml
Xunzhuo: Okay I will use this, and remove these if sentences below.

release/install.yaml
```diff
@@ -0,0 +1,342 @@
+apiVersion: v1
```
arkodg, line 1: releasing manifests by mistake ? :)

Xunzhuo: removed 😄

.github/workflows/release.yaml
```diff
@@ -0,0 +1,34 @@
+name: Release
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:        
+      - v*
+
+jobs:
```
arkodg, line 9: what I was suggesting earlier was to still call `lint` & `test` in the release flow but somehow not dedup the code/action (which was a secondary concern). we will still need to test the code/image before releasing it (may sound redundant since the pushed code should have passed CI, but sometimes ... )

Xunzhuo: Yes, this is what I thought before, we need to have some basic checks in case something weird happened.

arkodg: can you please add that lint and test back

Xunzhuo:
tested new changes in: https://github.com/Xunzhuo/gateway/actions/runs/2960956756
image is built normally.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      - name: Build and Push EG Image
+        env:
+          EG_TAG_VERSION: ${{ steps.tag_env.outputs.version}}
+        run: make image.push.multiarch TAG=${EG_TAG_VERSION} IMAGE_NAME=gateway PLATFORMS="linux_amd64 linux_arm64"
```
lizan, line 46: We really shouldn't rebuild the image but just re-tag (i.e. use what image that has been tested)

arkodg: makes sense, we can handle this in a follow up PR

arkodg (approved):
🚀

.github/workflows/release.yaml
arkodg: working on this rn

AliceProxy (commented):
Thanks! :slightly_smiling_face: :rocket:

PR #302, opened 2022-08-31
servicePort to containerPort translation

### LINKED ISSUE #192: mapping scheme from Service ports to ports that Envoy listens on
We'll need to sort out how ports on the service map to ports that Envoy listens on. Right now they're 1:1, but I imagine inside the Envoy container we'll want to listen on e.g. 8080/8443, and then have the Service map 80->8080 and 443->8443. However, since we need to handle arbitrary ports, not just those two, we'll need some kind of mapping scheme. I'll file an issue for this.

_Originally posted by @skriss in https://github.com/envoyproxy/gateway/pull/188#discussion_r937085989_

### LINKED ISSUE #278: XDS IR Logging

arkodg:
* Convert gateway listner ports into service ports and ephemeral
container ports
* Pass both port mappings in the Infra IR
* Pass the container port value to the Xds IR

Fixes: https://github.com/envoyproxy/gateway/issues/192

Also fixes https://github.com/envoyproxy/gateway/issues/278

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/yaml"
 
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/gatewayapi"
@@ -84,6 +85,10 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 			}
 			// Translate to IR
 			result := t.Translate(&in)
+			yamlXdsIR, _ := yaml.Marshal(&result.XdsIR)
+			r.Logger.Info(string(yamlXdsIR))
+			yamlInfraIR, _ := yaml.Marshal(&result.InfraIR)
+			r.Logger.Info(string(yamlInfraIR))
 
 			// Publish the IRs. Use the service name as the key
 			// to ensure there is always one element in the map.
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -30,9 +30,9 @@ const (
 
 	// minEphemeralPort is the first port in the ephemeral port range.
 	minEphemeralPort = 1024
-	// privilegedPortShift is the constant added to the privileged port
+	// wellKnownPortShift is the constant added to the well known port (1-1023)
 	// to convert it into an ephemeral port.
-	privilegedPortShift = 10000
+	wellKnownPortShift = 10000
 )
 
 // Resources holds the Gateway API and related
@@ -490,7 +490,7 @@ func servicePortToContainerPort(servicePort int32) int32 {
 	// This allows the container to bind to the port without needing a
 	// CAP_NET_BIND_SERVICE capability.
 	if servicePort < minEphemeralPort {
-		return servicePort + privilegedPortShift
+		return servicePort + wellKnownPortShift
 	}
 	return servicePort
 }
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -27,6 +27,12 @@ const (
 	// OwningGatewayNamespaceLabel is the owner reference label used for managed infra.
 	// The value should be the namespace of the Gateway used to CRUD the Service.
 	OwningGatewayNamespaceLabel = "gateway.envoyproxy.io/owning-gateway-namespace"
+
+	// minEphemeralPort is the first port in the ephemeral port range.
+	minEphemeralPort = 1024
+	// privilegedPortShift is the constant added to the privileged port
+	// to convert it into an ephemeral port.
+	privilegedPortShift = 10000
 )
 
 // Resources holds the Gateway API and related
@@ -219,7 +225,7 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR *ir.Xds,
 	}
 
 	// Infra IR proxy ports must be unique.
-	var foundPorts []v1beta1.PortNumber
+	var foundPorts []int32
 
 	// Iterate through all listeners to validate spec
 	// and compute status for each, and add valid ones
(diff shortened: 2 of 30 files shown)


danehans:
@arkodg this PR is not working as expected when I manually test. I see the listener get created on the targetPort of the Envoy service. I'm pushing this to RC2.

arkodg:
> @arkodg this PR is not working as expected when I manually test. I see the listener get created on the targetPort of the Envoy service. I'm pushing this to RC2.

if you are referring to envoy xds listener resource, that is the intent

danehans:
> if you are referring to envoy xds listener resource, that is the intent

Yes, I see the xds listener gets created in the envoy logs, but for some reason e2e is not working.

danehans (changes requested):
Fails manual e2e testing.

arkodg:
@danehans I spent some time manually testing this PR. This PR works fine but 
the workflow of updating a listener port in the Gateway spec doesnt work due to the infra manager's inability to update the service spec ports . Ive raised https://github.com/envoyproxy/gateway/issues/341 to track and fix it.

danehans:
@arkodg thanks for looking into this further. I prefer getting #341 fixed before merging this PR so we can test e2e. I'll pick up #341 if @Xunzhuo cannot work on it in the coming days.

arkodg:
@danehans I rebased this PR, tested it manually and it works
```
kubectl get gateway/eg -o yaml
kind: Gateway
metadata:
  annotations:
      {"apiVersion":"gateway.networking.k8s.io/v1alpha2","kind":"Gateway","metadata":{"annotations":{},"name":"eg","namespace":"default"},"spec":{"gatewayClassName":"eg","listeners":[{"name":"http","port":70,"protocol":"HTTP"}]}}
  creationTimestamp: "2022-09-02T23:00:52Z"
  generation: 1
  name: eg
  namespace: default
  resourceVersion: "788"
  uid: e152999b-977c-4e69-b4b4-4ec7961d541d
spec:
  gatewayClassName: eg
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 70
    protocol: HTTP
status:
  conditions:
  - lastTransitionTime: "2022-09-02T23:00:52Z"
    message: The Gateway has been scheduled by Envoy Gateway
    observedGeneration: 1
    reason: Scheduled
    status: "True"
    type: Scheduled

kubectl get svc -A
NAMESPACE              NAME                            TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)                  AGE
default                httpbin                         ClusterIP      10.108.10.70     <none>        80/TCP                   5d20h
default                kubernetes                      ClusterIP      10.96.0.1        <none>        443/TCP                  5d20h
envoy-gateway-system   envoy                           LoadBalancer   10.107.127.41    <pending>     70:30707/TCP             6m1s
envoy-gateway-system   envoy-gateway                   ClusterIP      10.105.104.57    <none>        18000/TCP                5d20h
envoy-gateway-system   envoy-gateway-metrics-service   ClusterIP      10.107.204.11    <none>        8443/TCP                 5d20h
gateway-system         gateway-api-admission-server    ClusterIP      10.110.193.193   <none>        443/TCP                  5d20h
kube-system            kube-dns                        ClusterIP      10.96.0.10       <none>        53/UDP,53/TCP,9153/TCP   5d20h

kubectl -n envoy-gateway-system port-forward service/envoy 8888:70 &
[1] 1199
🐳 ~/go-workspace/src/github.com/envoyproxy/gateway$ Forwarding from 127.0.0.1:8888 -> 10070
Forwarding from [::1]:8888 -> 10070

curl --verbose --header "Host: www.example.com" http://localhost:8888/get
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8888 (#0)
Handling connection for 8888
> GET /get HTTP/1.1
> Host: www.example.com
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< server: envoy
< date: Thu, 08 Sep 2022 19:37:30 GMT
< content-length: 238
< access-control-allow-origin: *
< x-envoy-upstream-service-time: 5
< 
{
  "args": {}, 
  "headers": {
    "Accept": "*/*", 
    "Host": "www.example.com", 
    "User-Agent": "curl/7.64.1", 
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000"
  }, 
  "origin": "10.1.0.23", 
  "url": "http://www.example.com/get"
}
* Connection #0 to host localhost left intact
* Closing connection 0
```

internal/gatewayapi/runner/runner.go
```diff
@@ -84,6 +85,10 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 			}
 			// Translate to IR
 			result := t.Translate(&in)
+			yamlXdsIR, _ := yaml.Marshal(&result.XdsIR)
+			r.Logger.Info(string(yamlXdsIR))
+			yamlInfraIR, _ := yaml.Marshal(&result.InfraIR)
+			r.Logger.Info(string(yamlInfraIR))
```
danehans, line 91: This was a surprise to see in the logs. If you want this to stay, can you please update the PR description? If this does stay, you can also add "Fixes #278" and I can close https://github.com/envoyproxy/gateway/pull/279.

internal/gatewayapi/translator.go
```diff
@@ -27,6 +27,12 @@ const (
 	// OwningGatewayNamespaceLabel is the owner reference label used for managed infra.
 	// The value should be the namespace of the Gateway used to CRUD the Service.
 	OwningGatewayNamespaceLabel = "gateway.envoyproxy.io/owning-gateway-namespace"
+
+	// minEphemeralPort is the first port in the ephemeral port range.
+	minEphemeralPort = 1024
+	// privilegedPortShift is the constant added to the privileged port
+	// to convert it into an ephemeral port.
```
danehans, line 34: > added to the privileged port

I don't see `privileged port` defined anywhere so it's difficult to understand this comment.

danehans (commented):
A few nits, otherwise /lgtm.

internal/gatewayapi/translator.go
arkodg: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports are also sometimes referred to as privileged ports because the user than can bind to such ports needs to have more privileges. I'll update this variable to `WellKnownPortShift` to keep it in line with the RFC https://datatracker.ietf.org/doc/html/rfc1340

danehans (approved):

PR #305, opened 2022-08-31
ci: use make targets instead actions

Xunzhuo:
Switch this to be a Make target
cc @LukeShu 


arkodg (approved):
🚀

PR #306, opened 2022-08-31
move infra kube logic into infra kube manager

arkodg:

### DIFF
--- a/internal/infrastructure/kubernetes/infra.go
+++ b/internal/infrastructure/kubernetes/infra.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -11,6 +12,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+const (
+	// envoyGatewayNamespace is the namespace where envoy-gateway is running.
+	envoyGatewayNamespace = "envoy-gateway-system"
 )
 
 // Infra holds all the translated Infra IR resources and provides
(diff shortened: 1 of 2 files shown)


internal/infrastructure/kubernetes/infra.go
```diff
@@ -32,11 +38,22 @@ type Resources struct {
...
 func NewInfra(cli client.Client) *Infra {
-	return &Infra{
+	infra := &Infra{
 		mu:        sync.Mutex{},
 		Client:    cli,
 		Resources: newResources(),
 	}
+
+	// Set the namespace used for the managed infra.
+	ns, found := os.LookupEnv("ENVOY_GATEWAY_NAMESPACE")
```
zirain, line 48: can use some tool like this https://github.com/istio/pkg/blob/8b4fffa5f37157b1dc5e3aa03b0f88c2a92f55e5/env/var.go#L128

arkodg: its eventually doing the same thing :) https://github.com/istio/pkg/blob/8b4fffa5f37157b1dc5e3aa03b0f88c2a92f55e5/env/var.go#L230
can you clarify how we gain by adopting something similar now ?

zirain: if there's more than one env vars to handle, the code will smell not good.

internal/infrastructure/kubernetes/infra.go
```diff
@@ -13,6 +14,11 @@ import (
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
+const (
+	// envoyGatewayNamespace is the namespace where envoy-gateway is running.
+	envoyGatewayNamespace = "envoy-gateway-system"
```
Xunzhuo, line 19: How about making this const exported? I think envoy-gateway-system will be used not only here in the future. It is like the root namespace of EG.

internal/infrastructure/kubernetes/infra.go
arkodg: yah agree, we should def find a way to reuse logic once we have more such env vars

internal/infrastructure/kubernetes/infra.go
arkodg: * currently the usage of `envoy-gateway-system` is only in YAML.
* I prefer making the const private now and changing it to visible outside the package once such a case exists

internal/infrastructure/kubernetes/infra.go
danehans: > if there's more than one env vars to handle, the code will smell not good.

@zirain feel free to create an issue and add it to the backlog milestone.

danehans (approved):

Issue #307, opened 2022-08-31
`image.multiarch.setup` and `image.multiarch.emulate` throw unexpected errors


Xunzhuo:
*Description*:
`image.multiarch.setup` and `image.multiarch.emulate` throw unexpected errors, even if buildx is installed.

*Logs*:

```
❯ make image.build.multiarch
/Library/Developer/CommandLineTools/usr/bin/make --warn-undefined-variables -f tools/make/common.mk image.build.multiarch
docker buildx --help | grep -qw "docker buildx" || { \
                echo "Cannot find `docker buildx`, please install first"; \
                exit 1; }
Usage of buildx:
      --builder string   
docker run --rm --privileged tonistiigi/binfmt --install linux/amd64 # Install QEMU emulator, the same emulator as the host will report an error but can safe ignore
installing: amd64 qemu-x86_64 already registered
{
  "supported": [
    "linux/arm64",
    "linux/amd64",
    "linux/riscv64",
    "linux/ppc64le",
    "linux/s390x",
    "linux/386",
    "linux/mips64le",
    "linux/mips64",
    "linux/arm/v7",
    "linux/arm/v6"
  ],
  "emulators": [
    "qemu-arm",
    "qemu-i386",
    "qemu-mips64",
    "qemu-mips64el",
    "qemu-ppc64le",
    "qemu-riscv64",
    "qemu-s390x",
    "qemu-x86_64"
  ]
}
```

PR #308, opened 2022-08-31
fix: unexpected verify errors

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/307

### DIFF
--- a/tools/make/image.mk
+++ b/tools/make/image.mk
@@ -67,15 +67,17 @@ image.push.%: image.build.%
 
 .PHONY: image.multiarch.verify
 image.multiarch.verify:
-	$(eval pass := $(shell ))
-	docker buildx --help | grep -qw "docker buildx" || { \
-		echo "Cannot find `docker buildx`, please install first"; \
-		exit 1; }
+	$(eval PASS := $(shell docker buildx --help | grep "docker buildx" ))
+	@if [ -z "$(PASS)" ]; then \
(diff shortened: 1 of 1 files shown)


tools/make/image.mk
```diff
@@ -67,15 +67,17 @@ image.push.%: image.build.%
...
+		echo "Cannot find docker buildx, please install first"; \
+		exit 1;\
+	fi
 
 .PHONY: image.multiarch.emulate $(EMULATE_TARGETS)
 image.multiarch.emulate: $(EMULATE_TARGETS)
 $(EMULATE_TARGETS): image.multiarch.emulate.%:
-	docker run --rm --privileged tonistiigi/binfmt --install linux/$* # Install QEMU emulator, the same emulator as the host will report an error but can safe ignore
+# Install QEMU emulator, the same emulator as the host will report an error but can safe ignore
+	docker run --rm --privileged tonistiigi/binfmt --install linux/$*
```
danehans, line 80: @Xunzhuo can you remove the extra whitespace here.

danehans (changes requested):
1 minor nit. Otherwise, /lgtm.

tools/make/image.mk
Xunzhuo: line 80? I do not think this is extra, command needs it in makefile.

danehans (approved):

PR #310, opened 2022-08-31
chore: restrict make build-multiarch to pull_request only

Xunzhuo:
Restrict make build-multiarch to pull_request only, since `make image.push.multiarch` will excute it again.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -39,6 +39,10 @@ jobs:
           name: codecov-envoy-gateway
           verbose: true
 
+      # build
+      - run: make build-multiarch
+        if: github.event_name == 'pull_request'
+
       # build and push image
       - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -42,6 +39,10 @@ jobs:
           name: codecov-envoy-gateway
           verbose: true
 
+      # build
+      - run: make build-multiarch
+        if: github.event_name == 'pull_request'
```
arkodg, line 44: nice, even though there is a build cache, this should save us some setup time for the push case

arkodg (approved):

.github/workflows/build_and_test.yaml
Xunzhuo: Yes, we should keep our workflows as clean as possible, do not run repeated and unuseful jobs.

PR #313, opened 2022-08-31
retag gateway-dev image in release GHA

arkodg:
* instead of rebuilding and retesting the image,
reuse/retag the envoyproxy/gateway-dev with the matching commit SHA

Addresses https://github.com/envoyproxy/gateway/pull/299#discussion_r959118354

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,7 +15,7 @@ jobs:
         id: vars
         shell: bash
         run: |
-          echo "##[set-output name=release_tag;]$(echo ${GITHUB_REF##*/})"
+          echo "::set-output name=release_tag::$(echo ${GITHUB_REF##*/})"
           echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
 
       - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -4,45 +4,35 @@ on:
...
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-
-      - name: Build and Push EG Image
-        env:
-          EG_TAG_VERSION: ${{ steps.tag_env.outputs.version}}
-        run: make image.push.multiarch TAG=${EG_TAG_VERSION} IMAGE_NAME=gateway PLATFORMS="linux_amd64 linux_arm64"
+      # pull, retag gateway-dev image and push
+      - name: Pull gateway-dev image
+        run: |
+          docker pull envoyproxy/gateway-dev:${{ steps.vars.outputs.sha_short }}
```
lizan, line 30: Use `skopeo`, `docker pull` only pulls system arch of the host machine, there's no way to use `docker` to retag a multiarch image.

arkodg: homer-disappear

danehans: @arkodg if you have no issues with `skopeo`, can you update the PR to resolve @lizan comments?

.github/workflows/release.yaml
```diff
@@ -5,50 +5,30 @@ on:
...
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-
-      - name: Build and Push EG Image
-        env:
-          EG_TAG_VERSION: ${{ steps.tag_env.outputs.version}}
-        run: make image.push.multiarch TAG=${EG_TAG_VERSION} IMAGE_NAME=gateway PLATFORMS="linux_amd64 linux_arm64"
+      - name: Retag and push existing gateway-dev image
+        run: |
+          skopeo copy --all docker://docker.io/envoyproxy/gateway-dev:${{ steps.vars.outputs.sha_short }} docker://docker.io/envoyproxy/gateway:${{ steps.vars.outputs.release_tag }}
```
Xunzhuo, line 28: I remembered that we are using envoyproxy/gateway-dev:latest as the dev image, but not building image with commit as the image tag. Is this way working as expectation, I guess it can not fetch image with commitID.

Xunzhuo: How about when building a dev image, we not only build it with tag `latest`, but also tag it with the commit id, using `git rev-parse --short HEAD`

arkodg: it pushes both https://hub.docker.com/r/envoyproxy/gateway-dev/tags

arkodg: Screen Shot 2022-09-01 at 12 28 22 PM

the skopeo cmd looks good, its failing because that SHA doesn't exist yet :)

Xunzhuo: Nope, I think dev workflow does not push image with commit id now, need to add this first. @arkodg

Xunzhuo (approved):

Xunzhuo (commented):
LGTM

danehans (approved):

lizan (approved):

PR #319, opened 2022-08-31
ci: fix manifest image tag

lizan:
the image tag was generated as:
`envoyproxy/gateway:[[ 2 == 2 ]]`

which was invalid


Xunzhuo (approved):
I have tested this, problem solved.

danehans (approved):

PR #320, opened 2022-08-31
fix: wrong image name in release manifests

Xunzhuo:
image

Fixed:

image

image tag of EG in release manifests is incorrect 😢, let us fix it.

### DIFF
--- a/tools/make/common.mk
+++ b/tools/make/common.mk
@@ -30,12 +30,6 @@ ifeq ($(origin OUTPUT_DIR),undefined)
 OUTPUT_DIR := $(ROOT_DIR)/bin
 endif
 
-# Set the version number. you should not need to do this
-# for the majority of scenarios.
-ifeq ($(origin VERSION), undefined)
-VERSION := $(shell git describe --abbrev=0 --dirty --always --tags | sed 's/-/./g')
-endif
-
 # REV is the short git sha of latest commit.
 REV=$(shell git rev-parse --short HEAD)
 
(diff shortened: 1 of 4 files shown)


tools/make/common.mk
```diff
@@ -33,7 +33,7 @@ endif
 # Set the version number. you should not need to do this
 # for the majority of scenarios.
 ifeq ($(origin VERSION), undefined)
-VERSION := $(shell git describe --abbrev=0 --dirty --always --tags | sed 's/-/./g')
+VERSION := $(shell git describe --abbrev=0 --dirty --always --tags)
```
arkodg, line 36: isn't TAG tied to REV ?
https://github.com/envoyproxy/gateway/blob/15859d119a4e73d452585b2f164e0256b7c11f08/tools/make/image.mk#L15

Xunzhuo: REV=$(shell git rev-parse --short HEAD)

REV is the latest commit not tag.

arkodg: and the release GHA is overriding the tag value here https://github.com/envoyproxy/gateway/blob/cc44487f4037b87a14f45597be5485ab3899d9b9/.github/workflows/release.yaml#L46

Xunzhuo: Right, we can just switch to TAG.

arkodg: can we just delete VERSION for now, since its not used anywhere and we are basing versioning off TAG

Xunzhuo: Yes, deleted.

arkodg (approved):
thanks for addressing this @Xunzhuo !

PR #321, opened 2022-08-31
Fixes Quickstart Guide URLs

danehans:
Removes `tree` from the URL path.


arkodg (approved):

arkodg:
lets merge once we release so we can test the URLs

Issue #295, opened 2022-08-30
Update Developer Guide


danehans:
Update [dev](https://github.com/envoyproxy/gateway/blob/main/DEVELOPER.md) guide or create a separate dev guide for walking contributors through the process of:

1. Hack EG.
2. Run tests.
3. Test against a kind cluster (create cluster, build/push EG image, etc.).
4. Other steps?

danehans:
Let's roll with the current contrib and dev guides for now and readdress this in the future based on feedback from the community.

PR #326, opened 2022-08-31
update DEVELOPER docs

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/295

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -26,8 +26,38 @@ Envoy Gateway is built using a [make][make]-based build system. Our CI is based 
   separately.
 
 ## Quick start
+* Run `make help` to see all the available targets to build, test and run `envoy-gateway`.
 
-Run `make help` to see all the available targets to build, test and run `envoy-gateway`.
+### Building the `envoy-gateway` binary
+* Run `make build` to build the binary that gets generated in the `bin/` directory
+
+### Running tests
+* Run `make test` to run the golang tests.
+
+### Running code linters
+* Run `make lint` to make sure your code passes all the linter checks.
+
+### Building and Pushing the Image
+* Run `IMAGE=foo/bar make image` to build the docker image.
+* Run `IMAGE=foo/bar make push-multiarch` to build and push the multi-arch docker image.
+
+**_NOTE:_**  Replace `IMAGE` with your registry's image name.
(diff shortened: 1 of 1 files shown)


DEVELOPER.md
```diff
@@ -26,8 +26,38 @@ Envoy Gateway is built using a [make][make]-based build system. Our CI is based
...
+* Run `make test` to run the golang tests.
+
+### Running code linters
+* Run `make lint` to make sure your code passes all the linter checks.
+
+### Building and Pushing the Image
+* Run `IMAGE=foo/bar make image` to build the docker image.
+* Run `IMAGE=foo/bar make push-multiarch` to build and push the multi-arch docker image.
+
+**_NOTE:_**  Replace `IMAGE` with your registry's image name.
```
danehans, line 44: Can you be more specific with this note, and/or provide an example above? `IMAGE` should be `REGISTRY/REPOSITORY/IMAGE_NAME`. For example:
```
IMAGE=docker.io/you/gateway-dev make push-multiarch
```

DEVELOPER.md
```diff
@@ -26,8 +26,38 @@ Envoy Gateway is built using a [make][make]-based build system. Our CI is based
...
+
+### Deploying Envoy Gateway in Kubernetes
```
danehans, line 46: I think most devs will want to use `make create-cluster` to dev against a Kind cluster. If you agree, can you add a step for this target and say that you can build/push a local EG image to the cluster with `make kube-install-image`?

DEVELOPER.md
```diff
@@ -26,8 +26,42 @@ Envoy Gateway is built using a [make][make]-based build system. Our CI is based
   separately.
 
 ## Quick start
+* Run `make help` to see all the available targets to build, test and run `envoy-gateway`.
```
danehans, line 29: @arkodg one last nit, do you mind reviewing all the bullets?

arkodg: this page already has bullets, prefer keeping the styling the same, similar to the rest of the docs such as [design](https://github.com/envoyproxy/gateway/blob/main/docs/design/SYSTEM_DESIGN.md) docs in this repo

danehans (approved):

Issue #290, opened 2022-08-30
Establish Label Conventions


danehans:
Currently, EG uses the labels established by controller-runtime. https://github.com/envoyproxy/gateway/pull/283 attempts to update the labels used by the managed proxy data-plane. We should establish label conventions for the control and data planes. Kube provides some guidance [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

danehans:
xref: https://github.com/envoyproxy/gateway/pull/283#discussion_r958666921

danehans:
cc: @arkodg @zirain

PR #328, opened 2022-09-01
infra: establish envoy labels

zirain:
fixes: #290

add `gatewyName` on envoy pod/service 


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -250,13 +252,16 @@ func (i *Infra) createDeployment(ctx context.Context, infra *ir.Infra) (*appsv1.
 //
 // TODO: Update k/v pair to use gatewayclass controller name to distinguish between
 //       multiple Envoy Gateways.
-func EnvoyPodSelector() *metav1.LabelSelector {
+func EnvoyPodSelector(gatewayName string) *metav1.LabelSelector {
```
danehans, line 255: The input name should be `className`, `gcName`, or something that represents a GatewayClass and not Gateway b/c the ProxyInfra name is established from the managed GatewayClass.

xref: https://github.com/envoyproxy/gateway/blob/v0.2.0-rc1/internal/gatewayapi/translator.go#L97

internal/infrastructure/kubernetes/deployment.go
```diff
@@ -250,13 +252,16 @@ func (i *Infra) createDeployment(ctx context.Context, infra *ir.Infra) (*appsv1.
...
 	return &metav1.LabelSelector{
-		MatchLabels: envoyLabels(),
+		MatchLabels: envoyLabels(gatewayName),
 	}
 }
 
 // envoyLabels returns the labels used for Envoy.
-func envoyLabels() map[string]string {
-	return map[string]string{"app": "envoy"}
+func envoyLabels(gatewayName string) map[string]string {
```
danehans, line 262: The input name should be `className`, `gcName`, or something that represents a GatewayClass and not Gateway b/c the ProxyInfra name is established from the managed GatewayClass.

xref: https://github.com/envoyproxy/gateway/blob/v0.2.0-rc1/internal/gatewayapi/translator.go#L97

danehans (changes requested):
Just a few small nits. Otherwise, /lgtm. If you have time, it would be great to add a test to `deployment_test.go` that checks for the existence of the labels and the selector.

zirain:
will add tests for this

internal/infrastructure/kubernetes/deployment.go
danehans: Yes, a GatewayClass is not 1:1 with a Gateway. Envoy Gateway supports multiple Gateways but only one GatewayClass that matches its configured controller name. Therefore, there is a 1:1 mapping of managed proxy infra with a GatewayClass. Envoy Gateway manages a single data plane, e.g. the Envoy Deployment. From a Kubernetes Infra Manager perspective, Gateways correlate to a Service. Therefore, it does make sense for Services to include a GatewayClass __and__ a Gateway label. This will allow users to understand what Gateway was used to create the Service. The selector for the Envoy Service should continue using the pod labels of the Envoy Deployment. [This section ](https://github.com/envoyproxy/gateway/blob/main/docs/design/SYSTEM_DESIGN.md#design-decisions)of the design doc provides additional background.

zirain: so, we should put both GatewayClass and Gateway's name?

danehans: @zirain I suggest we keep it simple and just use the GatewayCLass name for now.

zirain: updated
BTW, what's the plan about support multi gateway?

danehans (dismissed):
@zirain thanks for resolving the feedback.

zirain:
@danehans merge master to resolve conflicts

danehans (approved):

Issue #157, opened 2022-07-15
Run Gateway API conformance in CI


skriss:
Gateway API conformance tests should run as part of CI, to serve as our end-to-end testing. We should run the set of tests that are included in the Gateway API release that we're using (currently v0.5.0). Initially we'll want these to be non-blocking as we work towards a conformant implementation, but eventually they should be merge-blocking. 

xref. #65

skriss:
I've got much of this ready, but blocked for now until we have a functioning binary & YAML manifest.

skriss:
Still to do:
- add CI steps that run conformance
- when running for a PR, use the container image built for the PR rather than the default `envoyproxy/gateway-dev:latest`

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/425

PR #331, opened 2022-09-01
ci: add conformance tests in dev workflow

Xunzhuo:
xref: https://github.com/envoyproxy/gateway/issues/157

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -47,6 +47,10 @@ jobs:
     # build
     - run: make build-multiarch
 
+    # conformance
+    - name: Run Conformance Tests
+      run: make conformance
+
     # build and push image
     - name: Login to DockerHub
       if: github.event_name == 'push'


.github/workflows/build_and_test.yaml
```diff
@@ -43,6 +43,10 @@ jobs:
       - run: make build-multiarch
         if: github.event_name == 'pull_request'
 
+      # conformance
```
arkodg, line 46: this is def needed but we should add this once we have a better understanding on what works/doesnt
@skriss hoping you can guide on what are all the issues that need to be solved to pass some acceptable conformance (at par with the Quickstart guide)

Xunzhuo: Yes, I opened this PR for not merging this immediately(Some conformance is not acceptable yet).

danehans:
@Xunzhuo I converted the PR to "Draft" since it's a WIP.

Issue #334, opened 2022-09-02
 tech debt: refactor os.LookupEnv with default value


zirain:
xref: https://github.com/envoyproxy/gateway/pull/306#discussion_r960149240

https://github.com/envoyproxy/gateway/blob/29a89578d347db9611226125f6c808be27aff6b5/internal/infrastructure/kubernetes/infra.go#L48

PR #337, opened 2022-09-02
add utils/env package

zirain:
fixes: #334 

### DIFF
--- a/internal/utils/env/env.go
+++ b/internal/utils/env/env.go
@@ -1,0 +1,37 @@
+package env
+
+import (
+	"os"
+	"strconv"
+	"time"
+)
+
+type Var interface {
+	string | int | time.Duration
+}
+
+func Lookup[T Var](key string, defaultValue T) T {
+	value, ok := os.LookupEnv(key)
+	if !ok {
+		return defaultValue
+	}
+
+	var ret any
+	switch any(defaultValue).(type) {
+	case time.Duration:
+		d, err := time.ParseDuration(value)
+		if err != nil {
+			return defaultValue
+		}
+		ret = d
+	case string:
+		ret = value
+	case int:
+		i, err := strconv.ParseInt(value, 10, 0)
+		if err != nil {
+			return defaultValue
+		}
+		ret = int(i)
+	}
+	return ret.(T)
+}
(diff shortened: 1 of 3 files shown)


internal/utils/env/env.go
```diff
@@ -0,0 +1,37 @@
+package env
```
arkodg, line 1: nit: should we call the package `moreos` and API `LookupEnv` :) ?

internal/utils/env/env_test.go
```diff
@@ -0,0 +1,70 @@
...
+	got = Lookup("TEST_ENV_KEY", d)
+	require.Equal(t, got, time.Second*10)
+
+	os.Clearenv()
+	got = Lookup("TEST_ENV_KEY", d)
+	require.Equal(t, got, d)
+
+	os.Setenv("TEST_ENV_KEY", "1axx")
+	got = Lookup("TEST_ENV_KEY", d)
+	require.Equal(t, got, d)
```
arkodg, line 69: shouldnt this be `"1axx"`

internal/utils/env/env_test.go
```diff
@@ -0,0 +1,70 @@
...
+	require.Equal(t, got, 1000)
+
+	os.Clearenv()
+	got = Lookup("TEST_ENV_KEY", i)
+	require.Equal(t, got, i)
+
+	os.Clearenv()
+	os.Setenv("TEST_ENV_KEY", "10s")
+	got = Lookup("TEST_ENV_KEY", i)
+	require.Equal(t, got, i)
```
arkodg, line 49: shouldnt this be 10s ?

zirain: nop, `10s` is not a vaid `int`, backoff to defaultValue.

internal/utils/env/env_test.go
zirain: same as above

internal/utils/env/env.go
zirain: `moreos` sounds like handle the different between platform, like your guys do in [func-e](https://github.com/tetratelabs/func-e/tree/master/internal/moreos) project.

arkodg: cool, valid argument

internal/utils/env/env_test.go
arkodg: ah thanks for clarifying that, logic makes sense

arkodg (approved):
LGTM, thanks for adding this util !

Issue #335, opened 2022-09-02
Pin Envoy Image to v1.23 for v0.2.0


arkodg:
*Description*:

Pin the Envoy Proxy Image deployed by the Infra Manager to v1.23 for the v0.2.0 release

https://github.com/envoyproxy/gateway/blob/62bfc55332b0d1e47fad4f77e40713f3e52cb0b8/internal/ir/infra.go#L14

Related https://github.com/envoyproxy/gateway/issues/55#issuecomment-1234953351

PR #338, opened 2022-09-02
infra: Pin Envoy Image to v1.23

zirain:
fixes: #335

### DIFF
--- a/internal/gatewayapi/testdata/gateway-allows-same-namespace-with-allowed-httproute.out.yaml
+++ b/internal/gatewayapi/testdata/gateway-allows-same-namespace-with-allowed-httproute.out.yaml
@@ -71,7 +71,7 @@ xdsIR:
 infraIR:
   proxy:
     name: envoy-gateway-class
-    image: envoyproxy/envoy-dev:latest
+    image: envoyproxy/envoy:v1.23-latest
     listeners:
       - address: ""
         ports:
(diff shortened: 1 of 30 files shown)


arkodg (approved):
🚀

internal/ir/infra.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	DefaultProxyName  = "default"
-	DefaultProxyImage = "envoyproxy/envoy-dev:latest"
+	DefaultProxyImage = "envoyproxy/envoy:v1.23-latest"
```
danehans, line 14: @kflynn FYI for docs.

internal/gatewayapi/testdata/gateway-allows-same-namespace-with-allowed-httproute.out.yaml
```diff
@@ -71,7 +71,7 @@ xdsIR:
 infraIR:
   proxy:
     name: envoy-gateway-class
-    image: envoyproxy/envoy-dev:latest
+    image: envoyproxy/envoy:v1.23-latest
```
skriss, line 74: We should find a way to not have to bump these test cases every time we change the Envoy version, but fine to put on the backlog for now.

skriss (approved):

PR #350, opened 2022-09-07
Adds Initial Support for Infra IR Metadata

### LINKED ISSUE #173: Internal Metadata API
*Description*:
Design an internal metadata API used by all IR objects. Potential use cases:
1. Used to reference the external object(s) that were used to create the IR object.
2. Others?

[optional *Relevant Links*:]
https://github.com/envoyproxy/gateway/issues/119
https://github.com/envoyproxy/gateway/issues/32

danehans:
- Adds `metadata.labels` to the ProxyInfra IR. This allows the gatewayapi translator to plumb metadata from Gateway API resources, e.g. Gateway ns/name, into managed infra without polluting the IR with Gateway API-isms.
- Updates the gatewayapi translator to apply ProxyInfra metadata labels based upon managed Gateway ns and name.

__Note:__ `selfLink` is not used for metadata IR since it's deprecated and no longer set by Kubernetes.

Fixes #173 


internal/ir/infra.go
```diff
@@ -38,6 +37,14 @@ type ProxyInfra struct {
 	Listeners []ProxyListener
 }
 
+// InfraMetadata defines metadata for the managed proxy infrastructure.
+// +k8s:deepcopy-gen=true
+type InfraMetadata struct {
+	// Labels define a map of string keys and values that can be used to organize
+	// and categorize proxy infrastructure objects.
+	Labels map[string]string
```
arkodg, line 45: thoughts on `type InfraMetadata map[string]string` ?

danehans: IMHO the current approach makes it easier to introduce additional metadata types in the future, i.e. `ResourceVersion`, `CreationTimestamp`, etc.

LukeShu (approved):
I hadn't realized that selflink was deprecated. lgtm

Issue #318, opened 2022-08-31
Automate Release Docs Release Tags


danehans:
https://github.com/envoyproxy/gateway/pull/303 introduced a release-specific guide. The release references should be automatically updated when a release is cut.

- Update the `make release-manifests` Make target to `release artifacts`.
- Update `release-manifests.sh` to `release-artifacts.sh`
- Update the `release-artifacts.sh` to replace the previous release tag with the new release tag.

Xunzhuo:
Though I have some questions for this, but you can assign this to me. I will take a look into this issue, to find out the best practice for this.

Xunzhuo:
But I think this is in low priority, let us discuss this after releasing v0.2.0-rc1, no hurry for this:)

danehans:
Note that the quickstart from the `main` branch should not use release tags. This could cause considerable changes to the steps of the quickstart guide.

danehans:
Fixed by https://github.com/envoyproxy/gateway/pull/320

PR #352, opened 2022-09-07
Adds Support for Managing Gateway Status Addresses

### LINKED ISSUE #330: Set Gateway.Status.Addresses
EG needs to set the [Gateway.Status.Addresses field](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayStatus) with the address that is bound for the Gateway (i.e. the Envoy LB service address). This is necessary in order to make progress with Gateway API conformance.

A related question: do all Gateways use the same address? Does this have the potential to cause any kind of conflict across Gateways/Routes?

xref #65

danehans:
- Updates gateway controller to watch services and trigger a reconciliation for those with the EG Gateway owning labels.
- Moves label and label selector funcs to a common file, `labels.go`.
- Updates status manager to manage the Gateway "Ready" condition based on the availability of Gateway status addresses.
- Applies Gateway ns/name owning labels to the Envoy service.
- Updates the quickstart guide to use the Gateway status address instead of the Envoy service address.

Requires #350

Fixes #330

### DIFF
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -21,12 +21,9 @@ const (
 	KindService   = "Service"
 	KindSecret    = "Secret"
 
-	// OwningGatewayNameLabel is the owner reference label used for managed infra.
-	// The value should be the name of the Gateway used to CRUD the Service.
-	OwningGatewayNameLabel = "gateway.envoyproxy.io/owning-gateway-name"
-	// OwningGatewayNamespaceLabel is the owner reference label used for managed infra.
-	// The value should be the namespace of the Gateway used to CRUD the Service.
-	OwningGatewayNamespaceLabel = "gateway.envoyproxy.io/owning-gateway-namespace"
+	// OwningGatewayClassLabel is the owner reference label used for managed infra.
+	// The value should be the name of the accepted Envoy GatewayClass.
+	OwningGatewayClassLabel = "gateway.envoyproxy.io/owning-gatewayclass"
 
 	// minEphemeralPort is the first port in the ephemeral port range.
 	minEphemeralPort = 1024
@@ -110,15 +107,7 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
 
 	infraIR := ir.NewInfra()
 	infraIR.Proxy.Name = string(t.GatewayClassName)
-
-	// Apply Infra IR metadata.
-	if resources != nil {
-		for i := range resources.Gateways {
-			if resources.Gateways[i] != nil {
-				infraIR.Proxy.Metadata.Labels = GatewayOwnerLabels(resources.Gateways[i])
-			}
-		}
-	}
+	infraIR.Proxy.GetProxyMetadata().Labels = GatewayClassOwnerLabel(string(t.GatewayClassName))
 
 	// Get Gateways belonging to our GatewayClass.
 	gateways := t.GetRelevantGateways(resources.Gateways)
@@ -1005,10 +994,8 @@ func irTLSConfig(tlsSecret *v1.Secret) *ir.TLSListenerConfig {
 	}
 }
 
-// GatewayOwnerLabels returns owner labels for the provided Gateway.
-func GatewayOwnerLabels(gw *v1beta1.Gateway) map[string]string {
-	return map[string]string{
-		OwningGatewayNameLabel:      gw.Name,
-		OwningGatewayNamespaceLabel: gw.Namespace,
-	}
+// GatewayClassOwnerLabel returns the GatewayCLass Owner label using
+// the provided name as the value.
+func GatewayClassOwnerLabel(name string) map[string]string {
+	return map[string]string{OwningGatewayClassLabel: name}
 }
--- a/internal/infrastructure/kubernetes/infra_test.go
+++ b/internal/infrastructure/kubernetes/infra_test.go
@@ -19,9 +19,8 @@ import (
 func TestCreateInfra(t *testing.T) {
 	expected := ir.NewInfra()
 	// Apply the expected labels to the proxy infra.
-	expected.GetProxyInfra().GetProxyMetadata().Labels = envoyLabels()
-	expected.GetProxyInfra().GetProxyMetadata().Labels[gatewayapi.OwningGatewayNamespaceLabel] = "test-ns"
-	expected.GetProxyInfra().GetProxyMetadata().Labels[gatewayapi.OwningGatewayNameLabel] = "test-gw"
+	expected.GetProxyInfra().GetProxyMetadata().Labels = envoyAppLabel()
+	expected.GetProxyInfra().GetProxyMetadata().Labels[gatewayapi.OwningGatewayClassLabel] = "test-gc"
 
 	testCases := []struct {
 		name   string
--- a/internal/infrastructure/kubernetes/infra_test.go
+++ b/internal/infrastructure/kubernetes/infra_test.go
@@ -12,10 +12,17 @@ import (
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/envoyproxy/gateway/internal/envoygateway"
+	"github.com/envoyproxy/gateway/internal/gatewayapi"
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
 func TestCreateInfra(t *testing.T) {
+	expected := ir.NewInfra()
+	// Apply the expected labels to the proxy infra.
+	expected.GetProxyInfra().GetProxyMetadata().Labels = envoyLabels()
+	expected.GetProxyInfra().GetProxyMetadata().Labels[gatewayapi.OwningGatewayNamespaceLabel] = "test-ns"
+	expected.GetProxyInfra().GetProxyMetadata().Labels[gatewayapi.OwningGatewayNameLabel] = "test-gw"
+
 	testCases := []struct {
 		name   string
 		in     *ir.Infra
@@ -25,6 +32,25 @@ func TestCreateInfra(t *testing.T) {
 		{
 			name: "default infra",
 			in:   ir.NewInfra(),
+			out: &Resources{
+				ServiceAccount: &corev1.ServiceAccount{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       "ServiceAccount",
+						APIVersion: "v1",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace:       "default",
+						Name:            "envoy",
+						ResourceVersion: "1",
+					},
+				},
+			},
+			// Gateway owning labels are required to create the Envoy service.
+			expect: false,
+		},
+		{
+			name: "infra-with-expected-labels",
+			in:   expected,
 			out: &Resources{
 				ServiceAccount: &corev1.ServiceAccount{
 					TypeMeta: metav1.TypeMeta{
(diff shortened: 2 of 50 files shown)


internal/gatewayapi/translator.go
```diff
@@ -18,6 +18,13 @@ const (
 	KindHTTPRoute = "HTTPRoute"
 	KindService   = "Service"
 	KindSecret    = "Secret"
+
+	// OwningGatewayNameLabel is the owner reference label used for managed infra.
+	// The value should be the name of the Gateway used to CRUD the Service.
+	OwningGatewayNameLabel = "gateway.envoyproxy.io/owning-gateway-name"
+	// OwningGatewayNamespaceLabel is the owner reference label used for managed infra.
+	// The value should be the namespace of the Gateway used to CRUD the Service.
+	OwningGatewayNamespaceLabel = "gateway.envoyproxy.io/owning-gateway-namespace"
```
skriss, line 27: Doesn't the infra correspond to a GatewayClass, which could have >1 Gateways? I would think these labels should identify the GatewayClass.

internal/provider/kubernetes/gateway.go
```diff
@@ -65,6 +67,11 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
 	}
 	r.log.Info("watching gateway objects")
 
+	// Watch Services owned by managed Gateways.
```
skriss, line 70: We'll also need some code that takes the Service's LB IP (once present) and populates each Gateway's `status.addresses` field with it, I'm not seeing that in this iteration.

danehans:
Added `release-note` label to indicate that user-defined gateway addresses will not be supported in v0.2.0 unless https://github.com/envoyproxy/gateway/issues/360 is fixed.

internal/gatewayapi/translator.go
danehans: @skriss since a Gateway supports multiple addresses, each Gateway corresponds to a Service. Therefore, a Service name is constructed as `envoy-${GW_NAME}-${GW_NS}`, where the variables are derived from these labels.

internal/provider/kubernetes/gateway.go
danehans: @skriss understood. f8fd0fa adds the logic for populating `status.addresses` with ingress ip's/hostnames from the gateway's associated service.

docs/user/QUICKSTART.md
```diff
@@ -38,10 +38,11 @@ Create the HTTPRoute to route traffic through Envoy proxy to the example app:
 kubectl apply -f https://raw.githubusercontent.com/envoyproxy/gateway/v0.2.0-rc1/examples/kubernetes/httproute.yaml
 ```
 
-### Testing the configuration
+### Test the Configuration
+#### Option 1: Use Kubectl Port Forwarding
 Port forward to the Envoy service:
 ```shell
-kubectl -n envoy-gateway-system port-forward service/envoy-eg 8888:8080 &
+kubectl -n envoy-gateway-system port-forward service/envoy 8888:8080 &
```
arkodg, line 45: this is not really true, since the image points to v0.2.0 which creates `envoy-eg`

internal/infrastructure/kubernetes/infra_test.go
```diff
@@ -12,10 +12,17 @@ import (
...
 
 	"github.com/envoyproxy/gateway/internal/envoygateway"
+	"github.com/envoyproxy/gateway/internal/gatewayapi"
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
 func TestCreateInfra(t *testing.T) {
+	expected := ir.NewInfra()
+	// Apply the expected labels to the proxy infra.
+	expected.GetProxyInfra().GetProxyMetadata().Labels = envoyLabels()
```
arkodg, line 22: based on https://github.com/envoyproxy/gateway/issues/173#issuecomment-1238663746 I was under the assumption the metadata would be added as annotations. Can you please elaborate on why that changed ?

internal/provider/kubernetes/gateway.go
```diff
@@ -65,6 +68,11 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
...
+	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, r.enqueueRequestForOwningGateway()); err != nil {
```
arkodg, line 72: stared at this line for a while to understand why its needed - "trigger translation as well as updating the gateway status when the envoy service managed by the infra manager has changed". Would be great to add a similar comment in the code to make it easier for the next dev

internal/provider/kubernetes/gateway.go
```diff
@@ -135,24 +165,30 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 		r.resources.Gateways.Delete(request.NamespacedName)
 	}
 
-	// Set the "Scheduled" condition to true for all accepted gateways.
-	for _, gw := range acceptedGateways {
+	// Set status conditions for all accepted gateways.
+	for i := range acceptedGateways {
+		gw := acceptedGateways[i]
+		// Get the status address of the Gateway's associated Service.
+		svc, err := r.serviceForGateway(ctx, &gw)
```
arkodg, line 172: something to consider in the future - another way to solve this would have been for the infra manager to publish a `GatewayStatus` watchable message which would this provider would subscribe to, and update the status when it received a new one

internal/infrastructure/kubernetes/infra_test.go
arkodg: nvm, based on the current logic the labels are used to setup a watch in the gateway reconciler

internal/status/gateway.go
```diff
@@ -1,9 +1,51 @@
...
+	computeGatewayStatusAddrs(gw, svc)
+	gw.Status.Conditions = mergeConditions(gw.Status.Conditions,
+		computeGatewayScheduledCondition(gw, scheduled), computeGatewayReadyCondition(gw))
 	return gw
 }
+
+// computeGatewayStatusAddrs computes status addresses for the provided gateway
+// based on the status IP/Hostname of svc.
+func computeGatewayStatusAddrs(gw *gwapiv1b1.Gateway, svc *corev1.Service) {
+	var ingAddrs, ingHosts []string
```
arkodg, line 21: curious what the `ing` prefix is for ?

internal/infrastructure/kubernetes/infra_test.go
danehans: From the [Kube Annotations/Labels docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#:~:text=Labels%20can%20be%20used%20to,characters%20not%20permitted%20by%20labels.):

>Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects. The metadata in an annotation can be small or large, structured or unstructured, and can include characters not permitted by labels.

We are using the labels to __select objects__, e.g. select the Gateway to reconcile from the gateway controller Service watch. The PR's label-based approach appears to be consistent with the Kube docs.

@youngnick can you elaborate on why you believe Annotations should be used instead of Labels?

internal/provider/kubernetes/gateway.go
danehans: @arkodg how would ^ service controller know what Gateway to update status.addresses for?

internal/infrastructure/kubernetes/infra_test.go
danehans: @arkodg since this PR is blocking several other PRs, I created https://github.com/envoyproxy/gateway/issues/368 to resolve this issue. In the meantime, I would like to proceed with labels.

docs/user/QUICKSTART.md
danehans: Thanks for catching this. I removed the changes to the quickstart guide since it's pinned to rc1. I've added a note to https://github.com/envoyproxy/gateway/issues/318 so we can think about having a quickstart based on `main`.

internal/provider/kubernetes/gateway.go
arkodg: by building the `GatewayStatus` based on the IR metadata info. We might not need a controller as we start off,  just performing a `Get` post a Create/Update might be good enough

internal/status/gateway.go
danehans: It's short of `ingress`, since the addresses/hostnames are from `service.status.ingresses[0].ip`. I've removed `ing` from the var names to avoid confusion.

internal/provider/kubernetes/gateway.go
danehans: @arkodg ^ sounds like an interesting approach that requires additional thought and includes unknowns. Since we're close to RC2 and still have a ton of work, I prefer sticking with the current approach. I created https://github.com/envoyproxy/gateway/issues/371 to capture the potential for changing the approach to solving this problem.

arkodg:
@danehans curious if this PR unblocks the first few GatewayAPI conformance tests

danehans:
> @danehans curious if this PR unblocks the first few GatewayAPI conformance tests

@arkodg I have not looked into the tests so I'm unsure. From our discussions during the last few community meetings, I do know this PR will unblock our ability to __run__ conformance tests. This is becasue the test client uses `gateway.status.addresses` to connect to the Gateway(s) under test.

internal/provider/kubernetes/gateway.go
```diff
@@ -135,24 +166,30 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
...
```
arkodg, line 173: we should probably loop across all gateways and update status as a best effort rather than bail out early

arkodg:
> > @danehans curious if this PR unblocks the first few GatewayAPI conformance tests
> 
> @arkodg I have not looked into the tests so I'm unsure. From our discussions during the last few community meetings, I do know this PR will unblock our ability to **run** conformance tests. This is becasue the test client uses `gateway.status.addresses` to connect to the Gateway(s) under test.

cool, would be great if you can attach the Gateway output highlighting the updated status with the LB address, TIA

skriss (commented):
Mostly LGTM, one nit. I'm a bit unclear on what's going on with labels -- if we have a single service per GatewayClass, wouldn't we want to label the Service with the GatewayClass info? https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/translator.go#L114-L121 appears to only keep labels for the **last** Gateway that's seen since it's continually overwriting them for each Gateway, which seems weird. In the end I'm not sure how much it matters, since for every Reconcile call we're reconciling everything, but it may be confusing.

internal/provider/kubernetes/gateway.go
danehans: OK, now the controller logs the error and continues.

internal/provider/kubernetes/gateway.go
```diff
@@ -202,3 +239,20 @@ func gatewaysOfClass(gc *gwapiv1b1.GatewayClass, gwList *gwapiv1b1.GatewayList)
 	}
 	return ret
 }
+
+// serviceForGateway returns the service for the provided gateway, returning nil if
+// the service doesn't exist.
+func (r *gatewayReconciler) serviceForGateway(ctx context.Context, gw *gwapiv1b1.Gateway) (*corev1.Service, error) {
```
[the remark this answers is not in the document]
danehans: Removed.

danehans:
> I'm a bit unclear on what's going on with labels -- if we have a single service per GatewayClass, wouldn't we want to label the Service with the GatewayClass info?

@skriss thanks for the review. I have refactored the PR to use a GatewayClass owning label. The service watch event handler of the gateway controller now looks for the GatewayClass owning label. If found, it will load all the Gateways from the resource map and create a reconcile request for each Gateway that references a GatewayClass that matches the owning label.

skriss (approved):
This makes sense to me based on the current relationship between GatewayClass and infra/a Service, can always revisit if/when needed.

PR #353, opened 2022-09-07
fix: prevent panic when namespace is nil

Xunzhuo:
image

### DIFF
--- a/internal/gatewayapi/contexts.go
+++ b/internal/gatewayapi/contexts.go
@@ -133,6 +133,9 @@ func (l *ListenerContext) AllowsKind(kind v1beta1.RouteGroupKind) bool {
 }
 
 func (l *ListenerContext) AllowsNamespace(namespace *v1.Namespace) bool {
+	if namespace == nil {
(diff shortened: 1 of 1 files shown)


internal/gatewayapi/contexts.go
```diff
@@ -133,6 +133,9 @@ func (l *ListenerContext) AllowsKind(kind v1beta1.RouteGroupKind) bool {
 }
 
 func (l *ListenerContext) AllowsNamespace(namespace *v1.Namespace) bool {
+	if namespace == nil {
```
skriss, line 136: I think this is fine but in reality this situation should be pretty impossible -- if we're processing an HTTPRoute that's in a given Namespace, then that Namespace must by definition exist. I wonder if this is masking a more fundamental issue related to the cache not being properly filled.

Xunzhuo: Yes, but I think adding this check is to make codes safer.

skriss (approved):

PR #355, opened 2022-09-07
Delete xDs resources when GatewayClass is deleted

arkodg:
* Send an empty xDS IR in the `gateway-api` runner when there are no valid gateway classes

* Above operation should trigger a translation of empty xDs resources within the `xds-translator` runner


danehans:
@arkodg PTAL at the changes I made in https://github.com/envoyproxy/gateway/pull/352. Since each Gateway will have an associated Service, the Service name is derived from the Gateway's ns/name. The infra is now removed when the last Gateway of the managed GatewayClass is deleted. Otherwise, the Infra Manager is unable to delete the Envoy Service since no Gateways exist to construct the Service name. We can eventually separate the infra management, where the Service is CRUD'd with Gateways and all other infra resources, e.g. Deployment, are CRUD'd with the managed GatewayClass.

arkodg:
> @arkodg PTAL at the changes I made in #352. Since each Gateway will have an associated Service, the Service name is derived from the Gateway's ns/name. The infra is now removed when the last Gateway of the managed GatewayClass is deleted. Otherwise, the Infra Manager is unable to delete the Envoy Service since no Gateways exist to construct the Service name. We can eventually separate the infra management, where the Service is CRUD'd with Gateways and all other infra resources, e.g. Deployment, are CRUD'd with the managed GatewayClass.

If the delete logic for Gateway Class and Gateways is being refactored, prefer if it was part of a separate PR, and will hold off on this PR until then.

also im a little confused now because your comments in https://github.com/envoyproxy/gateway/issues/330#issuecomment-1238467328 suggest all gateways will use the same service.

danehans:
After discussing #330 during today's community call, supporting user-defined Gateway addresses is not required for Core conformance. #330 will be updated to revert to the current delete logic, e.g. GatewayClass-based.

danehans (approved):

Issue #158, opened 2022-07-15
add support for request header modifier filter


skriss:
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestHeaderFilter

Needed for core conformance.

AliceProxy:
I'll take this issue as well

PR #363, opened 2022-09-08
Add support for request modifier filters

AliceProxy:
Adds support for Gateway API's request modifier filters. Out of the fields `set/add/remove` none were listed as optional or core in the docs so I assumed they were all core.

Resolves: https://github.com/envoyproxy/gateway/issues/158


.gitignore
```diff
@@ -11,3 +11,7 @@ bin/
 
 # Release artifacts
 release-artifacts/
+
+# Outputs
+*.log
```
arkodg, line 16: curious who/what generates *.log from the repo

internal/ir/xds.go
```diff
@@ -181,6 +188,40 @@ func (h HTTPRoute) Validate() error {
 			errs = multierror.Append(errs, err)
 		}
 	}
+	if h.AddRequestHeaders != nil {
+		for i, header := range *h.AddRequestHeaders {
+			if err := header.Validate(); err != nil {
+				errs = multierror.Append(errs, err)
+			}
+			duplicate := false
```
arkodg, line 196: nit: using a map will be faster

internal/ir/xds.go
```diff
@@ -134,9 +137,13 @@ type HTTPRoute struct {
 	HeaderMatches []*StringMatch
 	// QueryParamMatches define the match conditions on the query parameters.
 	QueryParamMatches []*StringMatch
-	// Direct responses to be returned for this route. Takes precedence over Destinations and Redirect
+	// AddRequestHeaders defines header/value sets to be added to the headers of requests.
+	AddRequestHeaders *[]AddHeader
```
arkodg, line 141: this doesnt need to be ptr to a slice, using a slice directly should be fine

.gitignore
AliceProxy: Nothing in the repo generates a *.log, but I like to deploy the image to the cluster and check the logs sometimes so I frequently find myself committing a log file by accident then removing it from the commit.  

Happy to remove this change if you don't think it is necessary, but I figure that if there ever are .log files lying around then probably shouldn't be committed.

arkodg: yah lets rm it

arkodg (approved):
Thanks for adding this feature ! the xDS diffs LGTM, hoping @skriss can take a pass at the Gateway API translation

internal/gatewayapi/translator.go
```diff
@@ -657,6 +660,160 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 
 							redirectResponse = redir
 						}
+					case v1beta1.HTTPRouteFilterRequestHeaderModifier:
+						// Make sure the header modifier config actually exists
+						if headerModifier := filter.RequestHeaderModifier; headerModifier != nil {
+
```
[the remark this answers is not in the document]
LukeShu: Less preachy: It's so wide it makes the GitHub code-review webui hard to use.

internal/gatewayapi/translator.go
```diff
@@ -657,6 +660,160 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+							// Update the status if the filter failed to configure any valid headers to add/remove
+							if len(addRequestHeaders) == 0 && len(removeRequestHeaders) == 0 {
+								parentRef.SetCondition(
+									v1beta1.RouteConditionResolvedRefs,
+									metav1.ConditionFalse,
+									v1beta1.RouteReasonUnsupportedValue,
+									"RequestHeaderModifier Filter did not provide valid configuration to add/set/remove any headers",
+								)
+							}
+						}
```
LukeShu, line 816: This is inside of the `if headerModifier != nil {` block, which means it won't error for `requestHeaderModifier: null` but will error for `requestHeaderModifier: {}`, which just seems wrong to me.

My reading of the Gateway API spec is that an empty config is valid, and so we shouldn't error on it. (but maybe I missed something?)

internal/xds/translator/route.go
```diff
@@ -4,6 +4,7 @@ import (
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
+	wrappers "github.com/golang/protobuf/ptypes/wrappers"
```
LukeShu, line 7: `github.com/golang/protobuf` is deprecated, use `google.golang.org/protobuf` instead. (`google.golang.org/protobuf/types/known/wrapperspb` aught to be the specific package?)

Also, is there a reason to name the import instead of letting it take the default name?

LukeShu (commented):
I'm sorta requesting changes, but marking it as "comment" because for any of these if you say "there's actually a good reason", then I'll just say "ok" :)

"Request changes"
- The error on empty configs
- The protobuf package used

Nits that I wouldn't mark "request changes" for but if you're making changes anyway would be good to address:
- the error message typo
- the indentation
- adding the comment

LukeShu (approved):
Actually, since Alice is on PTO Thurs/Fri, let's merge this and I'll file a subsequent fixup PR.

PR #356, opened 2022-09-07
Update kube Infra resources if they already exist

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/341

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #365, opened 2022-09-08
Turn on the `scopelint` linter, fix complaints

LukeShu:
While reviewing https://github.com/envoyproxy/gateway/pull/356 I noticed a common mistake: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

Then a quick `grep` turned up several existing places that the mistake has already been made.  So I thought to myself "I bet there's a linter we can turn on to catch this."

As far as I can tell, the best match is `scopelint`, which is a little over-zealous.  If a sub-test doesn't call `t.Parallel()`, it isn't *really* necessary to address scopelint's complaint, but it doesn't hurt.

Signed-off-by: Luke Shumaker <lukeshu@datawire.io>


arkodg:
thanks for adding this @LukeShu , LGTM, I'll wait for this to merge before continuing with https://github.com/envoyproxy/gateway/pull/356

LukeShu:
Rebased on top of https://github.com/envoyproxy/gateway/pull/366, and fixed a pair of test failures.

arkodg (approved):
🚀

PR #369, opened 2022-09-08
update infra resources to nil when deleted

arkodg:


danehans:
@arkodg PTAL at https://github.com/envoyproxy/gateway/issues/370. I'm questioning whether we should continue using the Infra Manager Resource type.

danehans:
@arkodg do you have time to update this PR to "remove infra manager resources type" that fixes #370?

arkodg:
@danehans prefer closing this PR and reopening a new one

PR #373, opened 2022-09-09
Adds GatewayClass Finalizer

### LINKED ISSUE #358: Finalize the Accepted GatewayClass
Whenever one or more Gateways are using a GatewayClass, EG must add the `gateway-exists-finalizer.gateway.networking.k8s.io` finalizer on the associated GatewayClass. This ensures that a GatewayClass associated with a Gateway is not deleted while in use.

xref: https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gatewayclass_types.go#L45

danehans:
Fixes #358 

### DIFF
--- a/internal/provider/kubernetes/gateway.go
+++ b/internal/provider/kubernetes/gateway.go
@@ -160,20 +160,16 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 	if len(acceptedGateways) == 0 {
 		r.log.Info("No gateways found for accepted gatewayclass")
 		// If needed, remove the finalizer from the accepted GatewayClass.
-		if gatewayClassFinalized(acceptedClass) {
-			if err := r.removeFinalizer(ctx, acceptedClass); err != nil {
-				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from gatewayclass %s: %w",
-					acceptedClass.Name, err)
-			}
+		if err := r.removeFinalizer(ctx, acceptedClass); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from gatewayclass %s: %w",
+				acceptedClass.Name, err)
 		}
 	}
 
 	// If needed, finalize the accepted GatewayClass.
-	if !gatewayClassFinalized(acceptedClass) {
-		if err := r.addFinalizer(ctx, acceptedClass); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed adding finalizer to gatewayclass %s: %w",
-				acceptedClass.Name, err)
-		}
+	if err := r.addFinalizer(ctx, acceptedClass); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed adding finalizer to gatewayclass %s: %w",
+			acceptedClass.Name, err)
 	}
 
 	found := false
@@ -279,17 +275,7 @@ func (r *gatewayReconciler) serviceForGateway(ctx context.Context) (*corev1.Serv
 	return svc, nil
 }
 
-// gatewayClassFinalized returns true if the provided gc is finalized.
-func gatewayClassFinalized(gc *gwapiv1b1.GatewayClass) bool {
-	for _, f := range gc.Finalizers {
-		if f == gatewayClassFinalizer {
-			return true
-		}
-	}
-	return false
-}
-
-// addFinalizer adds the gatewayclass finalizer to the provided gc.
+// addFinalizer adds the gatewayclass finalizer to the provided gc, if it doesn't exist.
 func (r *gatewayReconciler) addFinalizer(ctx context.Context, gc *gwapiv1b1.GatewayClass) error {
 	if !slice.ContainsString(gc.Finalizers, gatewayClassFinalizer) {
 		updated := gc.DeepCopy()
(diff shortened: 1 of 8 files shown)


internal/provider/kubernetes/gateway.go
```diff
@@ -156,7 +159,23 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 	acceptedGateways := gatewaysOfClass(acceptedClass, allGateways)
 	if len(acceptedGateways) == 0 {
 		r.log.Info("No gateways found for accepted gatewayclass")
+		// If needed, remove the finalizer from the accepted GatewayClass.
+		if gatewayClassFinalized(acceptedClass) {
```
skriss, line 163: Your add and remove methods are already checking to see if the finalizer is there or not, this call to `gatewayClassFinalized` is redundant and can be removed.

internal/provider/kubernetes/gateway.go
```diff
@@ -156,7 +159,23 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
...
+			if err := r.removeFinalizer(ctx, acceptedClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from gatewayclass %s: %w",
+					acceptedClass.Name, err)
+			}
+		}
+	}
+
+	// If needed, finalize the accepted GatewayClass.
+	if !gatewayClassFinalized(acceptedClass) {
```
skriss, line 172: Your add and remove methods are already checking to see if the finalizer is there or not, this call to `gatewayClassFinalized` is redundant and can be removed.

internal/provider/kubernetes/gateway.go
```diff
@@ -156,7 +159,23 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
...
+		if err := r.addFinalizer(ctx, acceptedClass); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed adding finalizer to gatewayclass %s: %w",
+				acceptedClass.Name, err)
+		}
 	}
```
skriss, line 177: I think this should be in an `else` block of the above `if`? I.e. we should only add the finalizer if there are >0 accepted Gateways?

internal/provider/kubernetes/gateway.go
```diff
@@ -259,3 +278,37 @@ func (r *gatewayReconciler) serviceForGateway(ctx context.Context) (*corev1.Serv
...
+
+// gatewayClassFinalized returns true if the provided gc is finalized.
+func gatewayClassFinalized(gc *gwapiv1b1.GatewayClass) bool {
+	for _, f := range gc.Finalizers {
+		if f == gatewayClassFinalizer {
+			return true
+		}
+	}
+	return false
+}
```
skriss, line 290: per above comments, can drop this func.

internal/provider/kubernetes/gateway_test.go
```diff
@@ -289,3 +291,131 @@ func TestGatewaysOfClass(t *testing.T) {
 		})
 	}
 }
+
+func TestAddFinalizer(t *testing.T) {
```
skriss, line 295: In Contour we have some integration tests of `Reconcile` methods that use a fake client and don't require the whole `envtest` machinery -- I'm personally a fan of them since they test the whole set of logic rather than individual bits, but can be run simply without `envtest`. Might be a pattern we want to start using for our controllers here, to do better integration testing. See e.g. https://github.com/projectcontour/contour/blob/main/internal/provisioner/controller/gatewayclass_test.go. 

I imagine we'll punt this past our next milestone though.

internal/utils/slice/slice.go
```diff
@@ -0,0 +1,29 @@
...
+			continue
+		}
+		newSlice = append(newSlice, item)
+	}
+	if len(newSlice) == 0 {
+		// Sanitize for unit tests so we don't need to distinguish empty array
+		// and nil.
+		newSlice = nil
+	}
+	return newSlice
```
skriss, line 28: This can be simplified:
```suggestion
	var newSlice []string
	for _, item := range slice {
		if item == s {
			continue
		}
		newSlice = append(newSlice, item)
	}
	return newSlice
```

skriss (changes requested):

internal/provider/kubernetes/gateway_test.go
danehans: @skriss thanks for the review and feedback. I created https://github.com/envoyproxy/gateway/issues/394 to track this.

internal/provider/kubernetes/gateway.go
```diff
@@ -156,7 +159,19 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
...
+			return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from gatewayclass %s: %w",
+				acceptedClass.Name, err)
+		}
+	}
+
+	// If needed, finalize the accepted GatewayClass.
+	if err := r.addFinalizer(ctx, acceptedClass); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed adding finalizer to gatewayclass %s: %w",
+			acceptedClass.Name, err)
 	}
```
skriss, line 173: We should only try to add the finalizer if there's > 0 Gateways:

```suggestion
	} else {
	        // If needed, finalize the accepted GatewayClass.
	        if err := r.addFinalizer(ctx, acceptedClass); err != nil {
		        return reconcile.Result{}, fmt.Errorf("failed adding finalizer to gatewayclass %s: %w",
			        acceptedClass.Name, err)
	        }
	}
```

skriss: Note that the code actually works properly as-is, because when we reach the call to `addFinalizer`,  `acceptedClass` still has the finalizer since it's not being modified by the preceding `removeFinalizer`, a deep-copy of it is, therefore `addFinalizer` doesn't think it needs to add, but still the code as-is looks wrong so let's clear up the confusion.

skriss (approved):

PR #375, opened 2022-09-09
fix `make gen-check` error msg

arkodg:
* replace `` with '' to skip executing 'make manifests'

* rm @ in indented echo cmd within if statement


arkodg:
cc @zirain

skriss (approved):

PR #376, opened 2022-09-10
chore: ignore unfriendly msg in every cmds

Xunzhuo:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,6 @@
 # everything else run $(MAKE) directly, since that'd end up running
 # multiple sub-Makes if you give multiple targets on the CLI.
 _run:
-	$(MAKE) --warn-undefined-variables -f tools/make/common.mk $(MAKECMDGOALS)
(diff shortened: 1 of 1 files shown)


danehans (approved):

Makefile
```diff
@@ -15,6 +15,6 @@
 # everything else run $(MAKE) directly, since that'd end up running
 # multiple sub-Makes if you give multiple targets on the CLI.
 _run:
-	$(MAKE) --warn-undefined-variables -f tools/make/common.mk $(MAKECMDGOALS)
+	@$(MAKE) --warn-undefined-variables -f tools/make/common.mk $(MAKECMDGOALS)
```
arkodg, line 18: @LukeShu you had explicitly rm'd these, ptal
I personally like the commands being shown because I like to know what's being executed

arkodg: @LukeShu clarified that this would only rm the recursive make cmd being printed, so should be fine

Issue #323, opened 2022-08-31
Deploy custom image with the kube-deploy make target


arkodg:
*Description*:
`make kube-deploy` deploys the Gateway API CRDs as well as the EG resources into the K8s Cluster
It however deploys the `envoyproxy/gateway-dev:latest` image 
https://github.com/envoyproxy/gateway/blob/6fd629a98a5149e99f69e68337d1d8404ce35ebf/internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml#L31

Allowing the user to pass an ARG/ENV to `make kube-deploy` such as `IMAGE="foo/bar:latest" make kube-deploy` would allow devs to deploy their custom images in the cluster and test it out

danehans:
The kube-install-image Make target builds a local image and pushes it to the cluster.

arkodg:
> The kube-install-image Make target builds a local image and pushes it to the cluster.

this command builds & pushes into the kind cluster, this is very helpful for dev workflows
but also another case is to push to a non kind cluster.

TLDR we need to find a better way to `kustomize` image names & tag / maybe we can reuse https://github.com/envoyproxy/gateway/blob/c0a21c051488d90c291a4b75f7c5de07b62bd49a/tools/hack/release-manifests.sh#L34

Xunzhuo:
@arkodg /assign, I can work on this:)

PR #385, opened 2022-09-14
feat: support custom dev image in kube-deploy

Xunzhuo:
Fixes: #323 

+ Deploy custom EG image
image

+ Deploy latest EG image
image

+ Deploy released EG image
image

### DIFF
--- a/tools/make/common.mk
+++ b/tools/make/common.mk
@@ -30,6 +30,16 @@ ifeq ($(origin OUTPUT_DIR),undefined)
 OUTPUT_DIR := $(ROOT_DIR)/bin
 endif
 
+# Set Kubernetes Resources Directory Path
+ifeq ($(origin KUBE_DIR),undefined)
+KUBE_DIR := $(ROOT_DIR)/internal/provider/kubernetes/config
+endif
+
+# Set Infra Resources Directory Path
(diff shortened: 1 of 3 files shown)


tools/make/common.mk
```diff
@@ -30,6 +30,16 @@ ifeq ($(origin OUTPUT_DIR),undefined)
 OUTPUT_DIR := $(ROOT_DIR)/bin
 endif
 
+# Set Kubernetes Resources Directory Path
```
arkodg, line 33: * can we move these into kube.mk
* prefer if we named vars to KUBE_PROVIDER_DIR and KUBE_INFRA_DIR

tools/make/kube.mk
```diff
@@ -26,7 +26,11 @@ endif
 
 .PHONY: kube-install
 kube-install: manifests $(tools/kustomize) ## Install Envoy Gateway CRDs into the Kubernetes cluster specified in ~/.kube/config.
-	$(tools/kustomize) build internal/provider/kubernetes/config/crd | kubectl apply -f -
+	mkdir -pv $(OUTPUT_DIR)/kube
```
arkodg, line 29: should it be called `bin/kube` or `bin/manifests` ? 🤷

arkodg (approved):
LGTM  thanks !

Issue #387, opened 2022-09-14
Run `make run-kube-local` failed


Xunzhuo:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

make run-kube-local starts EG locally, but failed for couple of reasons:
1. Cannot find EG binary if we did not run make build manually before this:
<img width="813" alt="image" src="https://user-images.githubusercontent.com/48784001/190250033-3cf2a2a8-ee66-44e8-b0bc-0604f9935d8d.png">
2. multicluster kube-config will break the script:
<img width="800" alt="image" src="https://user-images.githubusercontent.com/48784001/190250270-b33457ae-08b3-4e50-bc50-bf1946ba3c89.png">

/assign

danehans:
@Xunzhuo even with https://github.com/envoyproxy/gateway/pull/388, this Make target will fail until https://github.com/envoyproxy/gateway/issues/256 is fixed, PTAL.

PR #388, opened 2022-09-14
fix: make run-kube-local failed

Xunzhuo:
Fixes: #387


tools/make/kube.mk
```diff
@@ -43,7 +43,7 @@ kube-undeploy: kube-uninstall ## Uninstall the Envoy Gateway controller into the
 	$(tools/kustomize) build internal/provider/kubernetes/config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f - 
 
 .PHONY: run-kube-local ## Run EG locally.
-run-kube-local: kube-install
+run-kube-local: build kube-install
```
arkodg, line 46: while you are here can you also move `## Run EG locally.` to this line, so it shows up in `make help`, TIA

Xunzhuo: Done.

arkodg (approved):
🚀

Issue #349, opened 2022-09-06
EG cannot collapse the base Gateways defined in the conformance tests


skriss:
Gateway API conformance defines a base set of 3 Gateways, all using the same GatewayClass: https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/base/manifests.yaml#L16-L62

EG's design is to be 1:1 with a GatewayClass, with a single address for all Gateways for that class. As such, all Listeners across all of the Gateways need to be compatible/collapsed, per https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/gateway_types.go#L72-L106.

Unfortunately, the Gateways defined for conformance are not collapsible: they each specify an HTTP Listener on port 80 with no specific hostname. EG is going to see those as conflicting, and not be able to program them. However, these Gateways are definitely valid, according to Gateway API. There's no requirement that Listeners across Gateways of a given GatewayClass be collapsible.

Seems like there are a few ways we could deal with this:
- consider these Gateways' Listeners compatible/collapsible -- this seems problematic, would mean all the routes across all the Gateways would get munged together 
- give each Gateway a unique address
- something else?

cc @youngnick, would appreciate your thoughts on this since I know you have historical context re: merging Listeners across Gateways.

danehans:
@youngnick can you please review to make sure ^ is accurate?

danehans:
@skriss thanks again for bringing this issue to light. IMHO this is a bug in the spec, so I created https://github.com/kubernetes-sigs/gateway-api/issues/1385. Let's see what the community has to say.

skriss:
The way I read it, the spec is fine -- it gives implementors the option to merge compatible Listeners across Gateways, but does not imply that they must, nor that Listeners across Gateways must be compatible. In this case they're not compatible, but they're still perfectly valid specs, so we need to handle them differently.

It seems to me like the best path forward is to go with a separate LB service per Gateway, each of which forwards to a different set of Envoy ports. That way, each Gateway has its own address/exposed ports, and that address only exposes the routes attached to that Gateway, but they all still share the same underlying data plane as originally designed.

[...]

Issue #382, opened 2022-09-13
Xds Translator Incorrectly Creates Listener


danehans:
*Description*:
The xds translator is creating a listener when listener creation is unneeded. Instead, the existing listener name and allowed hostnames should be updated.

*Repro steps*:
- Follow quickstart guide but update `gateway.listerns[].hostname: "*.example.com"`
- Create a 2nd gateway with a listen that uses a different hostname:
```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: eg-2
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      hostname: "*.foo.com"
      protocol: HTTP
      port: 8080
```
- Create a 2nd route that references the 2nd gateway:
```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: httpbin-2
spec:
  parentRefs:
    - name: eg-2
  hostnames:
    - "www.foo.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: httpbin
          port: 80
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /
```
- Tail the EG logs and observe:
```
2022-09-13T20:50:34.005Z	INFO	cache/logrwrapper.go:24	handling v3 xDS resource request, version_info 6, response_nonce 1975, nodeID envoy-9c56d94bc-6sjvz, node_version v1.23.1, resource_names [], type_url type.googleapis.com/envoy.config.listener.v3.Listener, errorCode 13, errorMessage Error adding/updating listener(s) listener_default-eg-2-http_8080: error adding listener: 'listener_default-eg-2-http_8080' has duplicate address '0.0.0.0:8080' as existing listener
```

The Gateway APi [spec](https://github.com/kubernetes-sigs/gateway-api/blob/v0.5.0/apis/v1beta1/gateway_types.go#L67-L113) states that the listeners should be merged.

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/404

Issue #380, opened 2022-09-13
Add support for multiple gateways


NomadXD:
*Description*:
I want to have several gateways(envoy proxies) under one envoy gateway and use `parentRef` in the `HttpRoute` to decide on which gateway the `HttpRoute` should get deployed. I tried the above scenario with the `v0.2.0-rc1` and couldn't make it work. 

### Steps I followed:

1. Followed the quick start guide and deployed 1 EG, 1 gateway, httpbin and 1 HttpRoute and verified the functionality. It works as expected.
2. Deployed a second gateway using the following CRD using `kubectl apply -f gateway.yaml`.

```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: egtest
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 8080
``` 

I'm getting the following error in envoy gateway container logs.

```
2022-09-13T07:05:50.939Z	ERROR	runner/runner.go:95	unable to validate infra ir, skipped sending it	{"runner": "gateway-api", "error": "listener ports field required", "errorCauses": [{"error": "listener ports field required", "errorCauses": [{"error": "listener ports field required"}]}]}
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate
	/home/runner/work/gateway/gateway/internal/gatewayapi/runner/runner.go:95
```

It says listener ports field required but I have provided it as you can see here.

skriss:
Note, this is similar to or maybe identical to #349 -- multiple Gateways, each with Listeners with identical hosts/ports/protocols.

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/404

PR #389, opened 2022-09-14
Updates Infra IR to Support Multiple Envoy Services

danehans:
Currently, EG creates a single Kube Service regardless of the number of Gateways. This causes issues with running conformance tests since the base manifests create three Gateways that cannot be collapsed into one. This PR updates the Infra IR to support managing multiple Kube Services by mapping a Gateway to Infra IR Listener and mapping a Gateway listener to Infra IR ListenerPort.

Partially fixes:
https://github.com/envoyproxy/gateway/issues/349
https://github.com/envoyproxy/gateway/issues/380
https://github.com/envoyproxy/gateway/issues/382

### DIFF
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -216,10 +216,17 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR *ir.Xds,
 	// Infra IR proxy ports must be unique.
 	var foundPorts []int32
 
+	// An Infra IR proxy listener is created per Gateway.
+	var infraListeners []ir.ProxyListener
+
 	// Iterate through all listeners to validate spec
 	// and compute status for each, and add valid ones
 	// to the Xds IR.
 	for _, gateway := range gateways {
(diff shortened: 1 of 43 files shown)


internal/gatewayapi/translator.go
```diff
@@ -459,16 +466,18 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR *ir.Xds,
 					proto = ir.HTTPSProtocolType
 				}
 				infraPort := ir.ListenerPort{
-					Name:          irInfraPortName(listener),
+					Name:          string(listener.Name),
 					Protocol:      proto,
 					ServicePort:   servicePort,
 					ContainerPort: containerPort,
```
arkodg, line 472: you'll also need to create a unique (servicePort, containerPort) tuple to ensure traffic doesnt get multiplexed in the envoy listener

danehans:
The community is deciding on the approach to supporting multiple gateways in https://github.com/envoyproxy/gateway/issues/349. I marked this as a draft until a decision is made.

PR #392, opened 2022-09-15
Add Gateway condition based on Envoy deployment status

### LINKED ISSUE #345: Surface Gateway Ready status based on Envoy Deployment Status
Currently, the "Ready" status condition is not being set for Gateways. When https://github.com/envoyproxy/gateway/issues/330 is fixed, the "Ready" condition will be based on the Envoy Service address being published. Additionally, the "Ready" condition should also consider the availability of the Envoy Deployment.

chauhanshubham:
This commit adds support for watching over envoy deployment updates, 
and capture the deployment status to eventually determine the gateway 
status condition. We mark the gateway as ready, if the envoy deployment
has no unavailable replicas.
Resolves #345 

### DIFF
--- a/internal/provider/kubernetes/gateway.go
+++ b/internal/provider/kubernetes/gateway.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -71,8 +72,11 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
 	}
 	r.log.Info("watching gateway objects")
 
-	// Trigger gateway reconciliation when the Envoy service has changed.
+	// Trigger gateway reconciliation when the Envoy Service or Deployment has changed.
 	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, r.enqueueRequestForOwningGatewayClass()); err != nil {
+		return err
+	}
+	if err := c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, r.enqueueRequestForOwningGatewayClass()); err != nil {
 		return err
 	}
 
@@ -105,7 +109,7 @@ func (r *gatewayReconciler) hasMatchingController(obj client.Object) bool {
 }
 
 // enqueueRequestForOwningGatewayClass returns an event handler that maps events with
-// the GatewayCLass owning label to Gateway objects.
+// the GatewayClass owning label to Gateway objects.
 func (r *gatewayReconciler) enqueueRequestForOwningGatewayClass() handler.EventHandler {
 	return handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
 		labels := a.GetLabels()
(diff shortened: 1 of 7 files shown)


internal/provider/kubernetes/gateway.go
```diff
@@ -172,12 +176,21 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 	// Set status conditions for all accepted gateways.
 	for i := range acceptedGateways {
 		gw := acceptedGateways[i]
-		// Get the status address of the Gateway's associated Service.
-		svc, err := r.serviceForGateway(ctx)
+
+		// Get the status of the Gateway's associated Envoy Deployment.
+		deployment, err := r.envoyDeploymentForGateway(ctx)
+		if err != nil {
+			r.log.Info("failed to get deployment for gatteway",
```
danehans, line 183: s/gatteway/gateway/

internal/provider/kubernetes/gateway.go
```diff
@@ -259,3 +272,19 @@ func (r *gatewayReconciler) serviceForGateway(ctx context.Context) (*corev1.Serv
 	}
 	return svc, nil
 }
+
+// envoyDeploymentForGateway returns the Envoy service, returning nil if the service doesn't exist.
```
danehans, line 276: s/service/deployment/ x2

internal/provider/kubernetes/gateway.go
danehans, line 183: s/gatteway/gateway/

internal/status/conditions.go
```diff
@@ -51,19 +52,25 @@ func computeGatewayScheduledCondition(gw *gwapiv1b1.Gateway, scheduled bool) met
...
 		return newCondition(string(gwapiv1b1.GatewayConditionReady), metav1.ConditionFalse,
 			string(gwapiv1b1.GatewayReasonAddressNotAssigned),
 			"No addresses have been assigned to the Gateway", time.Now(), gw.Generation)
-	default:
-		return newCondition(string(gwapiv1b1.GatewayConditionReady), metav1.ConditionTrue,
-			string(gwapiv1b1.GatewayReasonReady),
-			"Address assigned to the Gateway", time.Now(), gw.Generation)
 	}
+
+	// If there are any unavailable replicas for the Envoy Deployment, donot
```
danehans, line 63: Replace `donot` with don't or do not.

internal/status/conditions.go
```diff
@@ -51,19 +52,25 @@ func computeGatewayScheduledCondition(gw *gwapiv1b1.Gateway, scheduled bool) met
...
+	// mark the Gateway as ready yet.
+	if deployment.Status.UnavailableReplicas > 0 {
```
danehans, line 65: IMHO the Gateway should report "Ready=true" when `availableReplicas > 0`. Can you include the number of desired replicas (`spec.replicas) and the number of `availableReplicas` in the condition, e.g. `1/2 deployment replicas are available.

danehans (changes requested):
Thanks for the PR! I have a few minor changes to request.

internal/status/conditions.go
chauhanshubham: If 1/n replicas is available, do we want the gateway condition to be `ready=true`?
Based on the comment, I assume that we mark the gateway to be ready in this case.
But then do we still want the message to say the something like:
"Although just 1 replica is available, we're marking it `ready=true`, but  (n-1) replicas are still unavailable"?

chauhanshubham: I have taken care of this in the latest commits, could you PTAL and suggest changes if required. thanks

danehans (dismissed):
Thanks for updating the PR.

danehans (approved):

AliceProxy (approved):
Looks good to me

Issue #371, opened 2022-09-09
Should Infra Manager Watch Services and Update Status


danehans:
https://github.com/envoyproxy/gateway/pull/352 adds support for managing Gateway `status.addresses`. The approach in this PR adds a Service watch to the Kube provider gateway controller. When a Service is CRUD'd, the controller will reconcile the appropriate Gateway by parsing the "Gateway owning" label(s) from the Service. The appropriate Gateway `status.addresses` are then updated through the reconciliation process.

https://github.com/envoyproxy/gateway/pull/352#discussion_r966399688 suggests an alternative approach:

>something to consider in the future - another way to solve this would have been for the infra manager to publish a GatewayStatus watchable message which would this provider would subscribe to, and update the status when it received a new one

Should we stick with the current approach or shift to the approach from the above comment?

danehans:
From https://github.com/envoyproxy/gateway/pull/352#discussion_r966480558:
> by building the GatewayStatus based on the IR metadata info. We might not need a controller as we start off, just performing a Get post a Create/Update might be good enough

IMO this approach resonates if a Service controller is implemented in the Infra Manager. For example:
- The Service controller reconciles Services that include "Gateway owning" labels.
- The Service controller will publish a `GatewayStatus` message that includes the status ingress address/hostname for each Service that references the owning Gateway.
- Subscribers, e.g. the Kube provider gateway controller, will receive `GatewayStatus` messages and calculate the status for managed Gateways.

arkodg:
closing this issue for now since the Provider layer reconciles the Service resource and updates the Gateway Status, and this approach seems to be working fine.

PR #398, opened 2022-09-20
Updates Kube RBAC for Watching Envoy Deployment

### LINKED ISSUE #397: Controller Fails to List Deployments
https://github.com/envoyproxy/gateway/pull/392 added support for watching managing status based on the availability of the Envoy deployment. However, RBAC was not updated to allow the controller to list Deployments:

```
W0920 17:43:07.490750       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:envoy-gateway-system:envoy-gateway" cannot list resource "deployments" in API group "apps" at the cluster scope
```

danehans:
Fixes #397 

### DIFF
--- a/internal/provider/kubernetes/rbac.go
+++ b/internal/provider/kubernetes/rbac.go
@@ -5,3 +5,4 @@ package kubernetes
 
 // RBAC for watched resources of Gateway API controllers.
 // +kubebuilder:rbac:groups="",resources=secrets;services;namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
--- a/internal/provider/kubernetes/config/rbac/role.yaml
+++ b/internal/provider/kubernetes/config/rbac/role.yaml
@@ -16,6 +16,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - gateway.networking.k8s.io
   resources:
   - gatewayclasses


internal/provider/kubernetes/rbac.go
```diff
@@ -5,3 +5,4 @@ package kubernetes
 
 // RBAC for watched resources of Gateway API controllers.
 // +kubebuilder:rbac:groups="",resources=secrets;services;namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
```
arkodg, line 8: shouldnt this be limited to the `envoy-gateway-system` namespace ?

danehans: As with the other [watched resources](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/rbac.go#L6-L7), it's my understanding a controller [watch](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller#Controller.Watch) is across all namespaces and then an eventhandler/predicate are used to filter the sources, e.g. Deployment.

EG has RBAC config split between the Infra Manager and the Kube provider. In this case, the Kube provider controller is the component that requires a Deployment `watch` verb and not the Infra Manager.

@skriss thoughts?

danehans: When I add the `watch` and `list` verbes to the Infra Manager RBAC, I still see this bug:
```
$ kubectl get role/infra-manager -n envoy-gateway-system -o yaml
...
  resources:
  - deployments
  verbs:
  - create
  - get
  - list
  - watch
  - update
  - delete
```

EG Logs:
```
E0920 18:48:20.575121       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: Failed to watch *v1.Deployment: failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:envoy-gateway-system:envoy-gateway" cannot list resource "deployments" in API group "apps" at the cluster scope
```

@arkodg `... at the cluster scope` is key to the error message.

skriss: @danehans that sounds right to me, I don't think it's possible to constraint the informer for a single resource type to just one namespace.

skriss (approved):

internal/provider/kubernetes/rbac.go
arkodg: rbac can be kept within infra-manager when https://github.com/envoyproxy/gateway/issues/371 is worked on

Issue #391, opened 2022-09-15
Address comments from https://github.com/envoyproxy/gateway/pull/363


arkodg:
*Description*:
https://github.com/envoyproxy/gateway/pull/363 was merged, but raising this issue so comments raised by @LukeShu can be considered

arkodg:
cc @AliceProxy

PR #399, opened 2022-09-20
Address comments from PR#363

AliceProxy:
There were a few outstanding comments from https://github.com/envoyproxy/gateway/pull/363 that were not addressed when it was merged. This PR adds the requested changes and resolves issue https://github.com/envoyproxy/gateway/issues/391.

Changes:
- Fixed a typo in one of the request header modifier filter error statuses.
- Reduced un-necessary indentation
- Added comments with reason behind one of the error cases (Gateway API specification)
- Replaced a deprecated dependency

### DIFF
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-no-headers.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-no-headers.out.yaml
@@ -49,14 +49,9 @@ httpRoutes:
       filters:
       - type: RequestHeaderModifier
         requestHeaderModifier:
-          set: 
-          - name: "example:1"
-            value: "some-value"
-          - name: "good-header"
-            value: "some-value"
           add:
-          - name: "example/2"
-            value: "some-value"
+          set:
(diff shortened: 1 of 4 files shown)


LukeShu (approved):
Thanks!

internal/gatewayapi/testdata/httproute-with-header-filter-no-headers.out.yaml
```diff
@@ -0,0 +1,98 @@
...
+          add:
+          set:
+          remove:
+  status:
+    parents:
+    - parentRef:
+        namespace: envoy-gateway
+        name: gateway-1
+        sectionName: http
+      # controllerName: envoyproxy.io/gateway-controller
```
arkodg, line 61: once you rebase, this test will fail
this needs to be fixed to `controllerName: gateway.envoyproxy.io/gateway-controller`

arkodg (approved):

danehans (approved):

PR #400, opened 2022-09-20
Pull Envoy and Gateway API versions out of code.

kflynn:
Include them dynamically in the docs index, and manually in a barebones compatibility matrix.

Reviewers: It is very much worth discussing whether the approach here of pulling versions out
of the code is the Right Way to Do Things™.

Signed-off-by: Flynn emissary@flynn.kodachi.com

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -2,22 +2,52 @@ name: Docs
 on:
   push:
     branches:
-    - "main"
-    - "release-v*"
+      - "main"
+      - "release-v*"
     paths-ignore:
-    - "**/*.png"
-  pull_request:
-    branches:
-    - "main"
-    - "release-v*"
-    paths-ignore:
-    - "**/*.png"
+      - "**/*.png"
+  # pull_request:
+  #   branches:
+  #     - "main"
+  #     - "release-v*"
+  #   paths-ignore:
+  #     - "**/*.png"
 jobs:
-  docs:
+  docs-build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - uses: ./tools/github-actions/setup-deps
+      - uses: actions/checkout@v3
+      - uses: ./tools/github-actions/setup-deps
 
-    # docs
-    - run: make docs
+      # docs
+      - run: make docs
+
+      # Upload docs for GitHub Pages    
+      - name: Upload GitHub Pages artifact
+        uses: actions/upload-pages-artifact@v1.0.4
+        with:
+          # Path of the directory containing the static assets.
+          path: docs/html
+          # Duration after which artifact will expire in days.
+          # retention-days: # optional, default is 1
+
+  # This workflow contains a single job called "build"
+  docs-publish:
+    runs-on: ubuntu-latest
+    needs: docs-build
+
+    # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
+    permissions:
+      pages: write      # to deploy to Pages
+      deployments: write
+      id-token: write   # to verify the deployment originates from an appropriate source
+
+    # Deploy to the github-pages environment
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v1.2.1
--- a/internal/cmd/versions.go
+++ b/internal/cmd/versions.go
@@ -9,19 +9,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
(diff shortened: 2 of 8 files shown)


internal/cmd/versions.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	cmd.PersistentFlags().BoolVarP(&envOutput, "env", "e", false,
+		"If set, output as environment variable settings.")
+
+	return cmd
+}
+
+// versions shows the versions of the Envoy Gateway.
+func versions() error {
```
LukeShu, line 35: Globals are bad. Can we scope `envOutput`? Something like this:

```go
// getVersionsCommand returns the server cobra command to be executed.
func getVersionsCommand() *cobra.Command {
	// envOutput determines whether to output as environment settings
	var envOutput bool

	cmd := &cobra.Command{
		Use:     "versions",
		Aliases: []string{"version"},
		Short:   "Show versions",
		RunE: func(cmd *cobra.Command, args []string) error {
			return versions(envOutput)
		},
	}

	cmd.PersistentFlags().BoolVarP(&envOutput, "env", "e", false,
		"If set, output as environment variable settings.")e do this as

	return cmd
}

// versions shows the versions of the Envoy Gateway.
func versions(envOutput bool) error {
```
?

internal/cmd/versions.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+
+			foundGatewayAPI = true
+			break
+		}
+	}
+
+	if !foundGatewayAPI {
+		return fmt.Errorf("could not find Gateway API version")
+	}
```
LukeShu, line 66: It's kinda weird to me that it reports the Envoy version and the Gateway API version, but not its own version.

If this isn't intended to be a user-facing command, and just used internally to build the docs, then I guess that's fine.

In `ocibuild` what I did for such commands that were only used to build the docs is to put them behind a `//go:build aux` comment, so that I could `go run -tags=aux` them, but they wouldn't be included in the compiled binary. https://github.com/datawire/ocibuild/blob/master/main_aux.go

LukeShu: If this _is_ meant to be a user-facing thing, then it should probably use Cobra's built-in version stuff; set `cmd.Version` on the root command, and adjust `cmd.SetVersionTemplate()` as necessary.

danehans:
@kflynn when you have a moment, can you fix the merge conflict and resolve the above feedback?

internal/cmd/versions.go
kflynn: Good point; done.

danehans:
@kflynn consider using https://pages.github.com/ to host the project docs.

internal/cmd/versions.go
kflynn: @LukeShu We discussed this at the Envoy Gateway meeting today, and landed on the side of "it should be user-visible". You OK approving this version and letting a subsequent PR switch to the Cobra stuff? I'd like to kick around the best way to do that with you.

danehans:
@kflynn can you fix the CI failure, create a v0.2.0 issue to track https://github.com/envoyproxy/gateway/pull/400#discussion_r984153656 and then I'm fine merging?

kflynn:
@danehans Fixed the lint failure and created issue #498.

internal/cmd/versions.go
LukeShu: Sure.

.github/workflows/docs.yaml
```diff
@@ -2,22 +2,52 @@ name: Docs
 on:
   push:
     branches:
-    - "main"
-    - "release-v*"
+      - "main"
+      - "release-v*"
```
LukeShu, line 6: I'm confused that `yamllint` isn't complaining about this?

LukeShu, line 6: I'm confused that `yamllint` isn't complaining about this?

tools/make/docs.mk
```diff
@@ -3,7 +3,7 @@ DOCS_OUTPUT_DIR := docs/html
 
 .PHONY: docs
 docs: $(tools/sphinx-build) $(tools/goversion)
-	BUILD_VERSION=$(shell $(tools/goversion)) $(tools/sphinx-build) -j auto -b html $(DOCS_DIR) $(DOCS_OUTPUT_DIR)
+	env BUILD_VERSION=$(shell $(tools/goversion)) $(shell go run ./cmd/envoy-gateway versions --env) $(tools/sphinx-build) -j auto -b html $(DOCS_DIR) $(DOCS_OUTPUT_DIR)
```
LukeShu, line 6: What's that leading `env` doing here?

LukeShu (approved):
I couple of points of confusion, but lgtm.

Issue #301, opened 2022-08-31
Gateway API Controllers Should Watch Other Gateway API Resources


danehans:
Currently, Gateway API resources must be installed in a specific order for e2e functionality. A Gateway API controller should watch for changes to dependent resources. For example, the httproute controller should trigger an httproute reconciliation whenever a Gateway resource is CRUD'd.

danehans:
For those interested in helping with this issue, here is the use case that must be resolved.

- A user creates an HTTPRoute- Which causes the httproute controller to reconcile the HTTPRoute, adds the HTTPRoute and associated service/namespace to the watchable map, and writes status (once #329 is fixed). Since a Gateway doesn't exist, the HTTPRoute status will not be written (even with #329 fixed). Since the HTTPRoute references a Gateway, e.g. `spec.parentRefs[]`, that doesn't exist, I question whether the HTTPRoute should be written to the watchable map. Maybe the httproute controller should check for the availability of each `spec.parentRefs[]` and not process the route if the ref does not exist? Thoughts @skriss @LukeShu @arkodg?
- A user creates a Gateway- Which causes the gateway controller to reconcile the Gateway. The Gateway doesn't get added to the watchable map because it's invalid (references a GatewayClass that doesn't exist).
-  A user creates a GatewayClass- Which causes the gatewayclass controller to reconcile the GatewayClass. The GatewayClass is added to the watchable map, status is set to "Accepted=true" but "Ready=true" is not set since the proxy infra has not been provisioned by the Infra Manager due to the invalid Gateway. A notification is not sent over the watchable channel since all 3 resources are required. Now that the GatewayClass has been created, the Gateway is no longer invalid. Since the gateway controller is not watching GatewayClasses, the GatewayClass does not get reconciled until the next periodic reconciliation.

All 3 resources exist but were not processed in the necessary order so the Infra and Xds IRs do not exist until either the next periodic reconciliation or the Gateway is recreated.

If the Gateway changes, e.g. address, status, deleted, etc. HTTPRoute reconciliation should be triggered to update HTTPRoute status.

arkodg:
if there is a parent->child relationship b/w Gateway Class and Gateway and Gateway and HTTPRoute and order or accepted GatewayClass or Gateway matters/is sequential, maybe each reconciler could
* Update itself to save the current accepted resource
* force a reconcile (maybe by just calling child.Reconcile()) of the child resource
* store the accepted resource in the watchable map

youngnick:
[...]

PR #401, opened 2022-09-21
Adds HTTPRoute ParentRef Validation

danehans:
Adds support for validating `parentRefs` of an HTTPRoute. Previously, the httproute controller would process an HTTPRoute even if it references a gateway that doesn't. This causes EG to perform unnecessary processing. An HTTPRoute does not contain `status` so an error is surfaced.

Partially Fixes: #301

### DIFF
--- a/internal/provider/kubernetes/httproute.go
+++ b/internal/provider/kubernetes/httproute.go
@@ -33,9 +33,10 @@ const (
 )
 
 type httpRouteReconciler struct {
-	client        client.Client
-	log           logr.Logger
-	statusUpdater status.Updater
+	client          client.Client
+	log             logr.Logger
+	statusUpdater   status.Updater
+	classController gwapiv1b1.GatewayController
 
 	initializeOnce sync.Once
 	resources      *message.ProviderResources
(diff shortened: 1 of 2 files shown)


internal/provider/kubernetes/httproute.go
```diff
@@ -276,3 +292,46 @@ func (r *httpRouteReconciler) subscribeAndUpdateStatus(ctx context.Context) {
...
+
+	var ret []gwapiv1b1.Gateway
+	for i := range route.Spec.ParentRefs {
+		ref := route.Spec.ParentRefs[i]
+		if *ref.Kind != "Gateway" {
+			return nil, fmt.Errorf("invalid Kind %q", *ref.Kind)
+		}
+		if *ref.Group != gwapiv1b1.GroupName {
+			return nil, fmt.Errorf("invalid Group %q", *ref.Group)
+		}
```
skriss, line 311: should probably nil-check these before dereferencing

internal/provider/kubernetes/httproute.go
```diff
@@ -145,6 +147,20 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
...
+				return reconcile.Result{}, fmt.Errorf("invalid parentRefs for httproute %s/%s: %v",
+					routeKey.Namespace, routeKey.Name, err)
+			}
+			if len(gws) == 0 {
+				// Remove the route from the watchable map since it doesn't reference
+				// a managed Gateway.
+				r.resources.HTTPRoutes.Delete(routeKey)
+				return reconcile.Result{}, nil
+			}
+			log.Info("validated httproute parentRefs")
```
skriss, line 163: Since we're iterating over every route and adding them all to the watchable map, wouldn't we want to do this check for every route, not *just* the one that triggered the reconcile?

internal/provider/kubernetes/httproute.go
```diff
@@ -145,6 +147,20 @@ func (r *httpRouteReconciler) Reconcile(ctx context.Context, request reconcile.R
 		routeKey := utils.NamespacedName(&route)
 		if routeKey == request.NamespacedName {
 			found = true
+			gws, err := r.validateParentRefs(ctx, &route)
+			if err != nil {
+				// Remove the route from the watchable map since it's invalid.
+				r.resources.HTTPRoutes.Delete(routeKey)
+				return reconcile.Result{}, fmt.Errorf("invalid parentRefs for httproute %s/%s: %v",
```
skriss, line 154: returning an error will re-enqueue the object for reconciliation again, I don't think we want to do that here, we can just re-reconcile if/when it's changed in the future.

skriss (approved):

PR #395, opened 2022-09-16
send resource statuses using watchable

arkodg:
* Allow multiple runners (such as gatewayapi) to update the status of resources such as gateway and httproute by publishing the updated status of the resource on a watchable map.
* The provider runner subscribes to these status updates and updates the resource status in the API Server

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #402, opened 2022-09-21
TLS Passthrough support

### LINKED ISSUE #168: add support for TLSRoute / TLS listener protocol / TLS passthrough
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ProtocolType
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.TLSModeType
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.TLSRoute

chauhanshubham:
This commit adds a tlsroute controller which is further used to
configure tls passthrough in envoy.

**Testing Done**
1. Deployed the nginx application as mentioned here: https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/

2. Applied gateway configs as follows
```yaml
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: tls
      protocol: TLS
      tls:
        mode: Passthrough
      port: 8080
```

```yaml
kind: TLSRoute
metadata:
  name: tlsroute
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "nginx.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: my-nginx
          port: 443
          weight: 1
```

3. To test traffic
```bash
$ kubectl -n envoy-gateway-system port-forward service/envoy-default-eg 8888:8080
```

```bash
$ curl -v --resolve nginx.example.com:8888:127.0.0.1 https://nginx.example.com:8888 --cacert testing/tls-passthrough/example.com.crt
* Added nginx.example.com:8888:127.0.0.1 to DNS cache
* Hostname nginx.example.com was found in DNS cache
*   Trying 127.0.0.1:8888...
* Connected to nginx.example.com (127.0.0.1) port 8888 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: personal/tls-passthrough/example.com.crt
*  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* ... **collapsed for brevity**
* TLSv1.2 (IN), TLS handshake, Finished (20):
* Server certificate:
*  subject: CN=nginx.example.com; O=some organization
*  start date: Oct  8 05:48:33 2022 GMT
*  expire date: Oct  8 05:48:33 2023 GMT
*  common name: nginx.example.com (matched)
*  issuer: O=example Inc.; CN=example.com
*  SSL certificate verify ok.
> GET / HTTP/1.1
> Host: nginx.example.com:8888
> User-Agent: curl/7.77.0
> Accept: */*
>
... **collapsed for brevity**
<title>Welcome to nginx!</title>
... **collapsed for brevity**
* Connection #0 to host nginx.example.com left intact
```

Fixes #168 


internal/ir/xds.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net"
 
 	"github.com/tetratelabs/multierror"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
```
arkodg, line 8: this package should be free of any Gateway API imports

internal/ir/xds.go
```diff
@@ -108,6 +109,9 @@ func (h HTTPListener) Validate() error {
 // TLSListenerConfig holds the configuration for downstream TLS context.
 // +k8s:deepcopy-gen=true
 type TLSListenerConfig struct {
+	// TLSMode could either be Terminate or Passthrough. If the TLS mode
+	// is Passthrough, certificate and private key information is not required.
+	TLSMode v1beta1.TLSModeType
```
arkodg, line 114: I think TLS passthrough warrants its own Listener e.g.
```

type Xds struct {
	// HTTP listeners exposed by the gateway.
	HTTP []*HTTPListener
	// TLS passthrough listener exposed by the gateway
	TLS []*TLSListener
}
```

chauhanshubham: I started off with that, wasn't too sure about it so just added the tls mode in HTTPListener, assuming HTTPListener could just be called and used as a generic Listener (similar to the gateway object's listener).

maybe adding a different tls route makes more sense?
```
type HTTPListener struct {  // rename to "Listener"
    Name, Address, Port, Hostnames, TLS
    
    Routes  []*HTTPRoute
    TLSRoutes []*TLSRoute
}
```

internal/xds/translator/translator.go
```diff
@@ -54,14 +63,14 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
 
 		for _, httpRoute := range httpListener.Routes {
 			// 1:1 between IR HTTPRoute and xDS config.route.v3.Route
-			xdsRoute, err := buildXdsRoute(httpRoute)
+			xdsRoute, err := buildXdsRoute(httpRoute, tlsPassthrough)
```
arkodg, line 66: this doesnt look right, recommend looking at https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/tls#step-5-test-proxying-https-passthrough to see how its using a `tcp_proxy` filter

internal/xds/translator/translator.go
```diff
@@ -23,13 +25,20 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
 
 	for _, httpListener := range ir.HTTP {
 		// 1:1 between IR HTTPListener and xDS Listener
-		xdsListener, err := buildXdsListener(httpListener)
+		var xdsListener *listener.Listener
+		var err error
+		tlsPassthrough := httpListener.TLS != nil && httpListener.TLS.TLSMode == v1beta1.TLSModePassthrough
```
arkodg, line 30: preferred if this was another loop
```
for _, httpListener := range ir.TLS {
```
so the logic is less cyclomatic

internal/xds/translator/testdata/out/xds-ir/tls-route-passthrough.listeners.yaml
```diff
@@ -0,0 +1,42 @@
+- address:
+    socketAddress:
+      address: 0.0.0.0
+      portValue: 10080
+  filterChains:
+  - filters:
+    - name: envoy.filters.network.http_connection_manager
```
arkodg, line 7: why is this needed, shouldnt this be `tcp_proxy` instead ?

internal/ir/xds.go
arkodg: I prefer if you created a new one because the fields within `HTTPRoute` dont apply to TLS passthrough

chauhanshubham: done. thanks

chauhanshubham:
This requires two more things
1. adding tlsroute status updates to be implemented, similar to httproutes in #395 
2. update tlsroute controller to listen and validate gateway parent #403 
I can take it up separately if this PR is getting too long

danehans (changes requested):
@chauhanshubham the Kube provider [RBAC](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/rbac.go) needs to be updated for TLSRoutes.

internal/xds/translator/translator.go
```diff
@@ -78,6 +78,46 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
...
 	}
 
+	for _, tlsListener := range ir.TLS {
+		// 1:1 between IR TLSListener and xDS Listener
+		xdsListener, err := buildXdsPassthroughListener(tlsListener)
+		if err != nil {
+			return nil, multierror.Append(err, errors.New("error building xds listener"))
+		}
+
+		// Allocate virtual host for this tlsListener.
```
arkodg, line 88: are you sure we need a route resource ?
looking at https://github.com/envoyproxy/envoy/pull/13975 you probably need a filter chain match to match the SNI
```
    - filter_chain_match:
         server_names:
         - domain3.example.com
```
and directly assign a `cluster` to the filter
cc @lizan

danehans (commented):
One nit. Otherwise, can you create tlsroute integration test similar to [this](https://github.com/envoyproxy/gateway/blob/v0.2.0-rc2/internal/provider/kubernetes/kubernetes_test.go#L64)?

internal/provider/kubernetes/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// validateParentRefs validates parentRefs for the provided route, returning the
```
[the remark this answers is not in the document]
chauhanshubham: updated the comment. I think the requested tlsroute integration test is already in place https://github.com/envoyproxy/gateway/pull/402/files#diff-5f29f313b6f74b9836b294d6fe03a909e7e023a42655873eb1a13b316e624e7dR66

Had also added the rbac for tlsroute.

internal/xds/translator/translator.go
chauhanshubham: addressed. could you PTAL. thanks

internal/xds/translator/listener.go
```diff
@@ -71,6 +73,61 @@ func buildXdsListener(httpListener *ir.HTTPListener) (*listener.Listener, error)
 	}, nil
 }
 
+func buildXdsPassthroughListener(clusterName string, tcpListener *ir.TCPListener) (*listener.Listener, error) {
```
arkodg, line 76: nit: buildXdsTCPListener

internal/xds/translator/listener.go
```diff
@@ -71,6 +73,61 @@ func buildXdsListener(httpListener *ir.HTTPListener) (*listener.Listener, error)
...
+		StatPrefix: "passthrough",
+		ClusterSpecifier: &tcp.TcpProxy_Cluster{
+			Cluster: clusterName,
+		},
+	}
+	mgrAny, err := anypb.New(mgr)
+	if err != nil {
+		return nil, err
+	}
+
```
arkodg, line 91: not a blocker but should `tcpListener.SNIs` be replaced with `tcpListener.TLS.SNIs`, so the code under this can include the TLS inspector if `tcpListener.TLS != nil`

internal/ir/xds.go
```diff
@@ -48,6 +50,24 @@ func (x Xds) Validate() error {
 	return errs
 }
 
+func (x Xds) GetListener(name string) *HTTPListener {
```
arkodg, line 53: why not `GetHTTPListener` ?

arkodg:
thanks for addressing the comments ! the xds config looks good, have some minor suggestions, mainly to simplify the life of the dev adding TCPRoute support, but not blockers.
@skriss @AliceProxy can you please review the Gateway API Translator changes ?

arkodg:
@chauhanshubham when possible can you also attach a working example, there doesnt seem to be an existing conformance test for TLS passthrough yet

internal/ir/xds.go
```diff
@@ -384,3 +395,41 @@ func (s StringMatch) Validate() error {
...
+	Name string
+	// Address that the listener should listen on.
+	Address string
+	// Port on which the service can be expected to be accessed by clients.
+	Port uint32
+	// Server names that are compared against the server names of a new connection.
+	// Wildcard hosts are supported in the prefix form. Partial wildcards are not
+	// supported, and values like *w.example.com are invalid.
+	// SNIs are used only in case of TLS Passthrough.
+	SNIs []string
```
arkodg, line 412: if we added SNIs under a TLS struct, we could validate it to ensure that SNIs is not empty

internal/xds/translator/listener.go
```diff
@@ -71,6 +73,74 @@ func buildXdsListener(httpListener *ir.HTTPListener) (*listener.Listener, error)
...
+					PortSpecifier: &core.SocketAddress_PortValue{
+						PortValue: tcpListener.Port,
+					},
+				},
+			},
+		},
+		FilterChains: []*listener.FilterChain{filterChain},
+	}
+
+	if tcpListener.TLS != nil {
```
arkodg, line 126: thanks for addressing this !

internal/provider/kubernetes/tlsroute.go
```diff
@@ -0,0 +1,323 @@
...
+			key := update.Key
+			val := update.Value
+			r.statusUpdater.Send(status.Update{
+				NamespacedName: key,
+				Resource:       new(gwapiv1a2.TLSRoute),
+				Mutator: status.MutatorFunc(func(obj client.Object) client.Object {
+					if _, ok := obj.(*gwapiv1a2.TLSRoute); !ok {
+						panic(fmt.Sprintf("unsupported object type %T", obj))
+					}
+					return val
```
arkodg, line 317: recommend only updating the required field similar to https://github.com/envoyproxy/gateway/blob/34afca0f0ad17e4caed3cda7a1ac2a1a1658ab11/internal/provider/kubernetes/httproute.go#L338

internal/provider/kubernetes/tlsroute.go
```diff
@@ -0,0 +1,323 @@
...
+		r.resources.TLSRoutes.Delete(request.NamespacedName)
+		log.Info("deleted tlsroute from resource map")
+
+		// Delete the Namespace and Service from the resource maps if no other
+		// routes exist in the namespace.
+		routeList = &gwapiv1a2.TLSRouteList{}
+		if err := r.client.List(ctx, routeList, &client.ListOptions{Namespace: request.Namespace}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("error listing tlsroutes")
+		}
+		if len(routeList.Items) == 0 {
```
arkodg, line 266: is this safe ? what if HTTPRoute is accessing the same namespace and service ?

internal/gatewayapi/translator_test.go
```diff
@@ -76,6 +77,19 @@ func TestTranslate(t *testing.T) {
 
 			got := translator.Translate(resources)
 
+			envoyGatewayNsName := "envoy-gateway-gateway-1"
```
arkodg, line 80: this shouldnt be needed, we address the order issue here https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/sort.go

internal/provider/kubernetes/tlsroute.go
arkodg: if not ? I think it should be okay to handle this by raising a GH issue and fixing it in a follow up PR

internal/provider/kubernetes/tlsroute.go
```diff
@@ -0,0 +1,323 @@
...
+		log.Info("added tlsroute to resource map")
+
+		// Get the route's namespace from the cache.
+		nsKey := types.NamespacedName{Name: route.Namespace}
+		ns := new(corev1.Namespace)
+		if err := r.client.Get(ctx, nsKey, ns); err != nil {
+			if errors.IsNotFound(err) {
+				// The route's namespace doesn't exist in the cache, so remove it from
+				// the namespace resource map if it exists.
+				if _, ok := r.resources.Namespaces.Load(nsKey.Name); ok {
```
arkodg, line 211: same comment as below, is it safe to delete ?

internal/provider/kubernetes/tlsroute.go
chauhanshubham: This should not be deleted, we'll have list httproutes too at this point to check whether any httproutes exist in the namespace. Same in httproute controller.
I'll fix this.

internal/gatewayapi/runner/runner.go
```diff
@@ -3,8 +3,8 @@ package runner
 import (
 	"context"
 
+	"gopkg.in/yaml.v2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
-	"sigs.k8s.io/yaml"
```
arkodg, line 7: can you revert this ?

internal/gatewayapi/sort.go
```diff
@@ -50,6 +50,8 @@ func sortXdsIRMap(xdsIR XdsIRMap) {
 			// descending order
 			sort.Sort(sort.Reverse(XdsIRRoutes(http.Routes)))
 		}
+
+		sort.SliceStable(ir.TCP, func(i, j int) bool { return ir.TCP[i].Name < ir.TCP[j].Name })
```
arkodg, line 54: I think https://github.com/envoyproxy/gateway/pull/535 should handle order and rm the need for this

internal/provider/kubernetes/tlsroute.go
```diff
@@ -0,0 +1,326 @@
...
+		// Delete the Namespace and Service from the resource maps if no other
+		// routes (TLSRoute or HTTPRoute) exist in the namespace.
+		found, err := isRoutePresentInNamespace(ctx, r.client, request.NamespacedName.Namespace)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !found {
+			r.resources.Namespaces.Delete(request.Namespace)
+			log.Info("deleted namespace from resource map")
+			r.resources.Services.Delete(request.NamespacedName)
```
chauhanshubham, line 269: This attempts to delete a possible non-existent Service, and not the relevant ones. It is introducing an issue in the tlsroute controller, which was also present for httproute controller too. I have filed an issue for this https://github.com/envoyproxy/gateway/issues/536

arkodg (approved):
LGTM, thanks for spending the extra cycles to get TLS Passthrough support in for 0.2.0 !

danehans (approved):

PR #403, opened 2022-09-21
Adds Gateway Watch to HTTPRoute Controller

danehans:
Adds a watch for Gateway objects to the httproute controller. Valid HTTPRoutes that reference gateways managed by EG will be enqueued for reconciliation.

Partially Fixes: #301

### DIFF
--- a/internal/provider/kubernetes/httproute.go
+++ b/internal/provider/kubernetes/httproute.go
@@ -92,6 +92,14 @@ func newHTTPRouteController(mgr manager.Manager, cfg *config.Server, su status.U
 		return err
 	}
 
+	// Watch Gateway CRUDs and reconcile affected HTTPRoutes.
+	if err := c.Watch(
+		&source.Kind{Type: &gwapiv1b1.Gateway{}},
+		handler.EnqueueRequestsFromMapFunc(r.getHTTPRoutesForGateway),
+	); err != nil {
+		return err
+	}
+
 	// Watch Service CRUDs and reconcile affected HTTPRoutes.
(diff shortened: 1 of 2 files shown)


danehans:
cd01a94 rebases to pull in CI conformance test coverage.

internal/provider/kubernetes/httproute.go
```diff
@@ -92,6 +92,14 @@ func newHTTPRouteController(mgr manager.Manager, cfg *config.Server, su status.U
 		return err
 	}
 
+	// Watch Gateway CRUDs and reconcile affected HTTPRoutes.
+	if err := c.Watch(
+		&source.Kind{Type: &gwapiv1b1.Gateway{}},
+		handler.EnqueueRequestsFromMapFunc(r.getHTTPRoutesForGateway),
```
arkodg, line 98: curious how dequeue/deregister works

skriss (dismissed):
I'm OK with this in the interest of getting a demo release out, but I definitely think we need to look at https://github.com/envoyproxy/gateway/issues/413 sooner than later. After merging this, if we have a Gateway with 10 HTTPRoutes attached, then any time the Gateway changes, we'll end up running at least 11 full translation loops, and likely more due to other inter-related watches. This is obviously not going to be scaleable.

danehans:
Rebased to include #404.

danehans:
@skriss I added https://github.com/envoyproxy/gateway/issues/413 to v0.3.0-rc1.

skriss (approved):

skriss (approved):

PR #404, opened 2022-09-21
envoy per gateway

arkodg:
* create a xdsIR per gateway

* create a infraIR per gateway

* use the gateway namespace-name as the key for above IRs

* populate the envoy bootstrap config with a node id that matches the IR key

* populate the xds server snapshot for each xds request based on the node id

Fixes: https://github.com/envoyproxy/gateway/issues/349

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #412, opened 2022-09-22
Adds GatewayClass Watch to Gateway Controller

danehans:
Adds a watch for GatewayClass objects to the gateway controller. Gateways that reference the managed GatewayClass will be enqueued for reconciliation.

Partially Fixes: #301

### DIFF
--- a/internal/provider/kubernetes/gateway.go
+++ b/internal/provider/kubernetes/gateway.go
@@ -73,6 +73,12 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
 	}
 	r.log.Info("watching gateway objects")
 
+	// Trigger Gateway reconciliation when a managed GatewayClass has changed.
+	if err := c.Watch(
+		&source.Kind{Type: &gwapiv1b1.GatewayClass{}}, r.enqueueRequestForGatewaysOfClass()); err != nil {
+		return err
+	}
+
 	// Trigger gateway reconciliation when the Envoy Service or Deployment has changed.
 	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, r.enqueueRequestForOwningGateway()); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


danehans:
5c3adb6 rebases to pull in CI conformance test coverage.

danehans:
Rebased to include #404.

skriss:
CI failing here

danehans:
@skriss the CI failure is not related to this PR. #447 fixed the CI issue. 666968b rebases the PR.

danehans:
@skriss yeah, https://github.com/envoyproxy/gateway/pull/404 introduced a regression that was fixed by https://github.com/envoyproxy/gateway/pull/403.

internal/provider/kubernetes/gateway.go
```diff
@@ -67,12 +67,26 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
...
+		predicate.NewPredicateFuncs(r.gatewayHasMatchingController),
 	); err != nil {
 		return err
 	}
 	r.log.Info("watching gateway objects")
 
+	// Only enqueue GatewayClass objects that match this Envoy Gateway's controller name.
+	if err := c.Watch(
+		&source.Kind{Type: &gwapiv1b1.GatewayClass{}},
+		&handler.EnqueueRequestForObject{},
```
skriss, line 79: I *think* you want to use r.getGatewaysForClass here as a map func -- so watch GatewayClasses with a matching controller string, then trigger reconcile requests for the Gateways of that class.

Otherwise, getGatewaysForClass isn't being used anywhere

internal/provider/kubernetes/gateway.go
```diff
@@ -67,12 +67,26 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
...
+		predicate.NewPredicateFuncs(r.classHasMatchingController),
+	); err != nil {
+		return err
+	}
+	r.log.Info("watching gatewayclass objects")
+
+	// Trigger Gateway reconciliation when a managed GatewayClass has changed.
+	if err := c.Watch(&source.Kind{Type: &gwapiv1b1.GatewayClass{}}, r.enqueueRequestForOwningGatewayClass()); err != nil {
+		return err
+	}
```
skriss, line 89: Can you explain why this is needed in addition to the above watch on GatewayClasses? I'd think the first one would cover it (provided the comment is addressed).

internal/provider/kubernetes/gateway.go
```diff
@@ -84,9 +98,42 @@ func newGatewayController(mgr manager.Manager, cfg *config.Server, su status.Upd
...
+			r.log.Error(err, "failed to list gateways")
+			return false
+		}
+		for i := range gateways.Items {
+			if gateways.Items[i].Spec.GatewayClassName == gwapiv1b1.ObjectName(gc.Name) {
+				log.Info("enqueueing gatewayclass")
+				return true
+			}
+		}
+		log.Info("no managed gateways; bypassing reconciliation")
```
skriss, line 126: I would leave this up to r.getGatewaysForClass -- if there are no Gateways for the GatewayClass, then no reconcile requests will be triggered from the mapping func. Having it here seems redundant.

skriss (changes requested):

internal/provider/kubernetes/gateway.go
LukeShu, line 89: I'm confused by this.

It's my understanding of controller-runtime that it's basically only ever correct to use `EnqueueRequestForObject` for one type within a single controller, but here you're adding using it for `GatewayClass` in addition to the existing use of it for `Gateway`.

And why are we setting up two separate watches for `GatewayClass` here?

And in that last one, isn't the relationship backwards? That it should be a `r.enqueueRequestsForOwnedGateways()` that takes a `GatewayClass` and lists `Gateways` that are associated with it?

danehans:
@skriss @LukeShu https://github.com/envoyproxy/gateway/commit/049cd68a27be4e632abb74ab50256be7a574f008 removes the duplicate watch.

danehans:
I am seeing https://github.com/envoyproxy/gateway/issues/464 when I create an httproute, gateway, and then gatewayclass with this PR.

danehans:
Closing in favor of fixing the issue as part of https://github.com/envoyproxy/gateway/issues/413.

PR #425, opened 2022-09-26
Update Conformance Tests

danehans:
Updates conformance tests to:
- Run in CI
- Run one test, `HTTPRouteSimpleSameNamespace` instead of the entire suite.
- Update Gateways to use different port numbers.
- Adds sets additional test flags to make target.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -47,6 +47,10 @@ jobs:
     # build
     - run: make build-multiarch
 
+    # conformance
+    - name: Run Conformance Tests
+      run: make conformance
+
     # build and push image
     - name: Login to DockerHub
       if: github.event_name == 'push'
--- a/tools/hack/kind-load-image.sh
+++ b/tools/hack/kind-load-image.sh
@@ -39,16 +39,20 @@ fi
 
 # Update the image pull policy in the Envoy Gateway deployment manifest so
 # the image is served by the kind cluster.
-echo "setting \"imagePullPolicy: IfNotPresent\" for Envoy Gateway deployment"
-run::sed \
-  "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
-  "./internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml"
+for file in internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml internal/provider/kubernetes/config/envoy-gateway/certgen-job.yaml ; do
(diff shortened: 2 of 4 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -47,6 +47,10 @@ jobs:
     # build
     - run: make build-multiarch
 
+    # conformance
+    - name: Run Conformance Tests
+      run: make conformance
```
arkodg, line 52: since no command above is pushing this image to the hub, will this command auto load the built image (default IMAGE=envoyproxy/gateway-dev TAG=commit-short-sha) into the cluster ?

danehans: The certgen job was not creating the secret that EG requires at startup. 741b1efd31e3578e1168094a4be05ecff32cb3fb updates the `kube-install-image` target so the certgen job pulls the built EG image from kind nodes instead of the EG docker repo.

tools/hack/kind-load-image.sh
```diff
@@ -39,16 +39,20 @@ fi
 
 # Update the image pull policy in the Envoy Gateway deployment manifest so
 # the image is served by the kind cluster.
-echo "setting \"imagePullPolicy: IfNotPresent\" for Envoy Gateway deployment"
-run::sed \
-  "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
-  "./internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml"
+for file in internal/provider/kubernetes/config/envoy-gateway/deploy_and_ns.yaml internal/provider/kubernetes/config/envoy-gateway/certgen-job.yaml ; do
```
arkodg, line 42: nit: can we dedup the file names to create `for file in files` ?

tools/make/kube.mk
```diff
@@ -79,7 +79,7 @@ run-conformance: ## Run Gateway API conformance.
 	kubectl wait --timeout=5m -n gateway-system deployment/gateway-api-admission-server --for=condition=Available
 	kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
 	kubectl apply -f internal/provider/kubernetes/config/samples/gatewayclass.yaml
-	go test -tags conformance ./test/conformance --gateway-class=envoy-gateway
+	go test -tags conformance ./test/conformance --gateway-class=envoy-gateway --debug=true --cleanup-base-resources=false
```
arkodg, line 82: can you add `-v` so the CI logs are more verbose ?

arkodg (approved):

Issue #423, opened 2022-09-26
Handle IR deletes


arkodg:
*Description*:
With https://github.com/envoyproxy/gateway/pull/404, the gateway-api translator now creates an IR per Gateway resource. This change in logic means that the gateway-api runner must also trigger an IR delete when a gateway resource has been deleted.

arkodg:
one way to handle this would be 
1)  the runner would delete/reset all IR keys
2) repopulate with new keys
3) the infra and xds runner would subscribe for updates and not the complete state so can delete the translated ir resource based on `delete` notifications
4) must ensure 1) doesnt trigger multiple notifications and the end result of 2) is sent as a single notification

arkodg:
hey @LukeShu is there a way to pause watchable notifications so interim loads and deletes dont trigger notifications or would we need to use an external mutex/condition variable construct.

PR #442, opened 2022-09-28
Handle IR Deletes

arkodg:
* Add a function that computes which keys to delete in the gateway-api runner

* Delete above keys from the Infra and Xds IR maps

* Update infra and xds runners to trigger deletes when a delete notification is recieved.

Fixes: https://github.com/envoyproxy/gateway/issues/423


LukeShu (approved):

AliceProxy (approved):
lgtm

arkodg:
https://github.com/telepresenceio/watchable/pull/6 can be used in the future to improve perf since we wouldn't need to loop across the IR keys multiple times

Issue #162, opened 2022-07-15
return 500's for invalid HTTPRoute backend refs


skriss:
There is a set of rules around how to handle 1+ invalid backend refs on an HTTPRoute that generally involve returning 500 responses for any invalid backends. This needs to be implemented properly.

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteRule (see backendRefs).

Needed for core conformance.

AliceProxy:
I'll take this issue

danehans:
@AliceProxy FYI I'm seeing this issue with the following conformance tests:

https://github.com/envoyproxy/gateway/issues/431
https://github.com/envoyproxy/gateway/issues/432
https://github.com/envoyproxy/gateway/issues/433

Please see these issues for additional details.

AliceProxy:
> @AliceProxy FYI I'm seeing this issue with the following conformance tests:
> 
> #431 #432 #433
> 
> Please see these issues for additional details.

:+1: The changes I was working on at the end of last week didn't properly handle the weighted responses so I switched over to a solution similar to how Istio is doing it. I'm trying to get a PR up for the gateway meeting in a few hours, but I'll keep working on this until it's done either way today.

Issue #432, opened 2022-09-27
HTTPRouteInvalidBackendRefUnknownKind Conformance Test Fails


danehans:
The [HTTPRouteInvalidBackendRefUnknownKind](https://pkg.go.dev/sigs.k8s.io/gateway-api/conformance/tests#HTTPRouteInvalidBackendRefUnknownKind) conformance test is failing.

Test Logs:
```
...
http.go:159: Response expectation failed, not ready yet: expected status code to be 500, got 503
Sending Request:
< GET /v2 HTTP/1.1
< Host: 127.0.0.1:80
< User-Agent: Go-http-client/1.1
< Accept-Encoding: gzip
<
<

Received Response:
< HTTP/1.1 503 Service Unavailable
< Content-Length: 19
< Content-Type: text/plain
< Date: Tue, 27 Sep 2022 17:21:31 GMT
< Server: envoy
<
< no healthy upstream

    http.go:159: Response expectation failed, not ready yet: expected status code to be 500, got 503
    http.go:140: timeout while waiting after 30 attempts, 0/3 sucessess
--- FAIL: TestGatewayAPIConformance (53.61s)
    --- FAIL: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind (32.20s)
        --- PASS: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind/HTTPRoute_with_Invalid_Kind_has_a_ResolvedRefs_Condition_with_status_False_and_Reason_InvalidKind (0.09s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind/HTTP_Request_to_invalid_backend_with_invalid_Kind_receives_a_500 (30.00s)
FAIL
```

EG Logs:
```
2022-09-27T17:21:08.584Z	ERROR	controller/controller.go:326	Reconciler error	{"runner": "provider", "controller": "httproute", "object": {"name":"invalid-backend-ref-unknown-kind","namespace":"gateway-conformance-infra"}, "namespace": "gateway-conformance-infra", "name": "invalid-backend-ref-unknown-kind", "reconcileID": "5326eb1b-705b-4bab-9273-fa806b3b2403", "error": "invalid backendRef: invalid group; must be nil or empty string"}
```

[...]

Issue #431, opened 2022-09-27
HTTPRouteInvalidNonExistentBackendRef Conformance Test Fails


danehans:
The [HTTPRouteInvalidNonExistentBackendRef](https://pkg.go.dev/sigs.k8s.io/gateway-api/conformance/tests#HTTPRouteInvalidNonExistentBackendRef) conformance test is failing.

Test Logs:
```
...
http.go:159: Response expectation failed, not ready yet: expected status code to be 500, got 503
Sending Request:
< GET / HTTP/1.1
< Host: 127.0.0.1:80
< User-Agent: Go-http-client/1.1
< Accept-Encoding: gzip
<
<

Received Response:
< HTTP/1.1 503 Service Unavailable
< Content-Length: 19
< Content-Type: text/plain
< Date: Tue, 27 Sep 2022 17:05:46 GMT
< Server: envoy
<
< no healthy upstream

    http.go:159: Response expectation failed, not ready yet: expected status code to be 500, got 503
    http.go:140: timeout while waiting after 30 attempts, 0/3 sucessess
...
--- FAIL: TestGatewayAPIConformance (52.53s)
    --- FAIL: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef (31.13s)
        --- PASS: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef/HTTPRoute_with_only_a_nonexistent_BackendRef_has_a_ResolvedRefs_Condition_with_status_False_and_Reason_BackendNotFound (0.00s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef/HTTP_Request_to_invalid_nonexistent_backend_receive_a_500 (30.00s)
FAIL
```

Envoy is returning a 503 instead of a 500. This may be due to the httproute controller not adding the route to the watchable map since it's considered invalid.

EG Logs:
```
...
2022-09-27T17:05:36.839Z	ERROR	controller/controller.go:326	Reconciler error	{"runner": "provider", "controller": "httproute", "object": {"name":"invalid-nonexistent-backend-ref","namespace":"gateway-conformance-infra"}, "namespace": "gateway-conformance-infra", "name": "invalid-nonexistent-backend-ref", "reconcileID": "2b953b24-718c-4319-9658-789d9d9cb9cc", "error": "invalid backendRef: invalid namespace; must be nil"}
```

[...]

PR #443, opened 2022-09-28
Return 500 responses for invalid backends

AliceProxy:
When a HTTPRoute has invalid backends we need to return 500 responses for gateway API conformance. If there are no valid backends at all then a direct response is issued with status 500, if there is a mix of valid and invalid backends then a weighted cluster is used. 

resolves https://github.com/envoyproxy/gateway/issues/162

### DIFF
--- a/internal/gatewayapi/testdata/httproute-with-some-invalid-backend-refs-no-service.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-some-invalid-backend-refs-no-service.out.yaml
@@ -1,0 +1,99 @@
+gateways:
+- apiVersion: gateway.networking.k8s.io/v1beta1
+  kind: Gateway
+  metadata:
+    namespace: envoy-gateway
+    name: gateway-1
+  spec:
+    gatewayClassName: envoy-gateway-class
+    listeners:
+    - name: http
+      protocol: HTTP
+      port: 80
+      allowedRoutes:
+        namespaces:
+          from: All
+  status:
+    listeners:
+    - name: http
+      supportedKinds:
+      - group: gateway.networking.k8s.io
+        kind: HTTPRoute
+      attachedRoutes: 1
+      conditions:
+      - type: Ready
+        status: "True"
+        reason: Ready
+        message: Listener is ready
+httpRoutes:
+- apiVersion: gateway.networking.k8s.io/v1beta1
+  kind: HTTPRoute
+  metadata:
+    namespace: default
+    name: httproute-1
+  spec:
+    parentRefs:
+    - namespace: envoy-gateway
+      name: gateway-1
+    rules:
+    - matches:
+      - path:
+          type: Exact
+          value: "/exact"
+      backendRefs:
+      - name: service-that-doesnt-exist
+        port: 8080
+      - name: service-that-doesnt-exist-2
+        port: 8080
+      - name: service-1
+        port: 8080
+  status:
+    parents:
+    - parentRef:
+        namespace: envoy-gateway
+        name: gateway-1
+      controllerName: gateway.envoyproxy.io/gatewayclass-controller
+      conditions:
+      - type: Accepted
+        status: "True"
+        reason: Accepted
+        message: Route is accepted
+      - type: ResolvedRefs
+        status: "False"
+        reason: BackendNotFound
+        message: "Service default/service-that-doesnt-exist-2 not found"
+xdsIR:
+  envoy-gateway-gateway-1:
+    http:
+    - name: envoy-gateway-gateway-1-http
+      address: 0.0.0.0
+      port: 10080
+      hostnames:
+      - "*"
+      routes:
+      - name: default-httproute-1-rule-0-match-0-*
+        pathMatch:
+          exact: "/exact"
+        backendWeights: 
+          invalid: 2  # The weighted clusters for invalid backend refs aren't part of the IR
+          valid: 1
+        destinations:
+        - host: 7.7.7.7
+          port: 8080
+          weight: 1
+infraIR:
+  envoy-gateway-gateway-1:
+    proxy:
+      metadata:
+        labels:
+          gateway.envoyproxy.io/owning-gateway: gateway-1
+      name: envoy-gateway-gateway-1
+      image: envoyproxy/envoy:v1.23-latest
+      listeners:
+      - address: ""
+        ports:
+        - name: envoy-gateway-gateway-1
+          protocol: "HTTP"
+          containerPort: 10080
+          servicePort: 80
+
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -505,6 +505,111 @@ func servicePortToContainerPort(servicePort int32) int32 {
 	return servicePort
 }
 
+// buildRuleRouteDest takes a backendRef and translates it into a destination or sets error statuses and
+// returns the weight for the backend so that 500 error responses can be returned for invalid backends in
+// the same proportion as the backend would have otherwise received
+func buildRuleRouteDest(backendRef v1beta1.HTTPBackendRef,
+	parentRef *RouteParentContext,
+	httpRoute *HTTPRouteContext,
+	resources *Resources) (destination *ir.RouteDestination, backendWeight uint32) {
+
+	weight := uint32(1)
+	if backendRef.Weight != nil {
+		weight = uint32(*backendRef.Weight)
+	}
+
+	if backendRef.Group != nil && *backendRef.Group != "" {
+		parentRef.SetCondition(
+			v1beta1.RouteConditionResolvedRefs,
+			metav1.ConditionFalse,
+			v1beta1.RouteReasonInvalidKind,
+			"Group is invalid, only the core API group (specified by omitting the group field or setting it to an empty string) is supported",
+		)
+		return nil, weight
+	}
+
+	if backendRef.Kind != nil && *backendRef.Kind != KindService {
+		parentRef.SetCondition(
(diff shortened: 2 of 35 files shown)


internal/gatewayapi/translator.go
```diff
@@ -901,108 +1004,35 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
-							continue
+					for _, route := range ruleRoutes {
+						// If the route already has a direct response or redirect configured, then it was from a filter so skip
+						// processing any destinations for this route.
+						if route.DirectResponse == nil && route.Redirect == nil {
+							if destination != nil {
+								route.Destinations = append(route.Destinations, destination)
+								continue
+							} else {
+								route.InvalidBackends += 1
```
arkodg, line 1018: should `InvalidBackends` be a field within `DirectResponse` ?

internal/gatewayapi/translator.go
```diff
@@ -901,108 +1004,35 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+				}
 
-					var portFound bool
-					for _, port := range service.Spec.Ports {
-						if port.Port == int32(*backendRef.Port) {
-							portFound = true
-							break
+				// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
+				for _, ruleRoute := range ruleRoutes {
+					if ruleRoute.InvalidBackends > 0 && len(ruleRoute.Destinations) == 0 {
```
arkodg, line 1027: why not set the DirectResponse status in L1018 ?

internal/xds/translator/route.go
```diff
@@ -135,6 +140,31 @@ func buildXdsRouteAction(routeName string) *route.RouteAction {
 	}
 }
 
+func buildXdsWeightedRouteAction(httpRoute *ir.HTTPRoute) *route.RouteAction {
+	invalidBackendWeight := uint32(100 * float64(httpRoute.InvalidBackends) / float64(len(httpRoute.Destinations)+httpRoute.InvalidBackends))
```
arkodg, line 144: shouldnt the weight take into consideration the weight of each destination ?

internal/xds/translator/route.go
```diff
@@ -135,6 +140,31 @@ func buildXdsRouteAction(routeName string) *route.RouteAction {
...
+	destinationWeight := 100 - invalidBackendWeight
+	clusters := []*route.WeightedCluster_ClusterWeight{
```
arkodg, line 146: do we really need weighted clusters ? its another level of complexity or can we implement the same with https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-msg-config-endpoint-v3-lbendpoint

arkodg: cc @lizan

AliceProxy: Where would the endpoint point to though?

arkodg: looks like this was already raised https://github.com/envoyproxy/envoy/issues/21930 and fixed via this route action https://github.com/envoyproxy/envoy/pull/22148 :)

internal/gatewayapi/translator.go
AliceProxy: No because it's used to calculate the weights. If we're using `DirectResponse` then we don't do anything with the weight, but I'm going to need to rework that field anyways to properly calculate the weight.

internal/xds/translator/route.go
AliceProxy: It absolutely does need to, good catch :+1:

internal/gatewayapi/translator.go
AliceProxy: If `DirectResponse` did support weighted matches alongside regular clusters that would be great since I wouldn't have to setup the weighted clusters.

internal/gatewayapi/translator.go
AliceProxy: Because at that point we've not figured out if the backends are valid or not yet so we can't determine if the `DirectResponse` will be used at all.

internal/gatewayapi/translator.go
arkodg: if the below approach of using the basic `cluster` works, can we reuse `RouteDestination` https://github.com/envoyproxy/gateway/blob/54b26e2ef5d2bb6d536539acf0cbbfa5cdc31bea/internal/ir/xds.go#L219 and explicitly set an `Invalid` bool in there ?

internal/xds/translator/route.go
AliceProxy: Yeah, thats exactly what I'm doing below. I updated the go control plane just so I could use that. I don't think it applies to endpoints though so thats why I went for the weighted cluster approach after reading that issue.

lizan: My first impression here is do we really need to send invalid backends to Envoy? Is that required by Gateway API conformance?

internal/xds/translator/route.go
```diff
@@ -135,6 +140,31 @@ func buildXdsRouteAction(routeName string) *route.RouteAction {
 	}
 }
 
+func buildXdsWeightedRouteAction(httpRoute *ir.HTTPRoute) *route.RouteAction {
+	invalidBackendWeight := uint32(100 * float64(httpRoute.BackendWeights.Invalid) / float64(httpRoute.BackendWeights.Valid+httpRoute.BackendWeights.Invalid))
```
lizan, line 144: No need those calculation, you can just use the values and total_weights is optional in xDS.

internal/xds/translator/route.go
youngnick: Yes, it's required for conformance that if input config is incorrect in certain ways, then 500 error codes (exactly) are returned for the traffic that matches that config, and if there are multiple match conditions, and one has an error, the proportions need to be correct, as @AliceProxy says above.

The idea here is that if you ask for an invalid config, that's what you get - otherwise you wouldn't be able to detect, for example, that a canary had a typo in its Service name until you flipped the weight to 100 - if we just silently dropped invalid config, it could get really weird.

lizan: That makes sense if we don't have other way to give obvious feedback to user. In that case weighted cluster sounds reasonable to me.

arkodg: tried out empty endpoints
```
      - lb_endpoints:
        - endpoint:
            address: {}      
        loadBalancingWeight: 99 
        locality: {} 
``` 
which fails validation so looks like `this the way` !

internal/gatewayapi/testdata/httproute-with-some-invalid-backend-refs-no-service.out.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - name: envoy-gateway-gateway-1-http
+      address: 0.0.0.0
+      port: 10080
+      hostnames:
+      - "*"
+      routes:
+      - name: default-httproute-1-rule-0-match-0-*
+        pathMatch:
+          exact: "/exact"
+        BackendWeights:
```
arkodg, line 77: nit: can we make `BackendWeights`  and other newly added fields camelCase to make it uniform across all yamls

arkodg:
can you also add the appropriate conformance test to https://github.com/envoyproxy/gateway/blob/3da08fe21c27bf70b47c22971b51e69480c92fcd/test/conformance/conformance_test.go#L41

arkodg:
> can you also add the appropriate conformance test to
> 
> https://github.com/envoyproxy/gateway/blob/3da08fe21c27bf70b47c22971b51e69480c92fcd/test/conformance/conformance_test.go#L41

 https://github.com/envoyproxy/gateway/issues/431 and https://github.com/envoyproxy/gateway/issues/432 should have the names of the tests

AliceProxy:
> > can you also add the appropriate conformance test to
> > https://github.com/envoyproxy/gateway/blob/3da08fe21c27bf70b47c22971b51e69480c92fcd/test/conformance/conformance_test.go#L41
> 
> #431 and #432 should have the names of the tests

That all I needed to do? Looks like they ran and passed. I checked the logs to confirm they really ran.
```
 --- PASS: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef (8.04s)

 --- PASS: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef/HTTPRoute_with_only_a_nonexistent_BackendRef_has_a_ResolvedRefs_Condition_with_status_False_and_Reason_BackendNotFound (0.00s)

--- PASS: TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef/HTTP_Request_to_invalid_nonexistent_backend_receive_a_500 (0.01s)

--- PASS: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind (7.06s)

--- PASS: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind/HTTPRoute_with_Invalid_Kind_has_a_ResolvedRefs_Condition_with_status_False_and_Reason_InvalidKind (0.00s)

--- PASS: TestGatewayAPIConformance/HTTPRouteInvalidBackendRefUnknownKind/HTTP_Request_to_invalid_backend_with_invalid_Kind_receives_a_500 (0.00s)
```

arkodg (approved):
LGTM, thanks for adding this tricky feature !

AliceProxy:
> LGTM, thanks for adding this tricky feature !

Thanks for the many reviews after all my changes :sweat_smile:

lizan (approved):

Issue #413, opened 2022-09-23
Consider Merging Controllers


danehans:
Since not all Gateway API resources have a parent-child relationship, consider merging controllers into a single reconciler.

xref: https://github.com/envoyproxy/gateway/pull/412
xref: https://github.com/envoyproxy/gateway/pull/403

arkodg:
I prefer if we implement this sooner than later and here's why
* performance
  * adding multiple watches for multiple resource controller means we'll be multiple notifications if a single resource changes, which will trigger multiple redundant translations (due to updating the watchable map)
* code complexity
  * handling deletes - since each resource type is part of a separate watchable map, the logic to perform deletes is non trivial at the moment
  * knowing which resource to to reconcile when another resource is updated is non trivial and error prone

chauhanshubham:
+1
I think we could consider Gateway object as a root object, add a Gateway controller which watches over httproute, tlsroute, services, refgrants. and push these objects into the resource table as the updates come in.

danehans:
> I think we could consider Gateway object as a root object
A GatewayClass must be the root object. All others are objects are dependents of a GatewayClass.

skriss:
+1 to this, I think it will also help down on a lot of code duplication that we're starting to get between the reconcilers and the Gateway API translator, around things like mapping routes to gateways, validating parent refs, etc.

chauhanshubham:
I can take this up, have some psuedo code ready.

Xunzhuo:
Thanks @chauhanshubham

PR #447, opened 2022-09-28
rm provider resource wait during init logic

arkodg:
* removes all the waitgroup logic associated with the provider resources (gatewayclass, gateway, httproute) that was added to ensure that all resources within the watchable map was comepletely initialized before the gateway api translated the resources

* this logic is being removed because the individual waitgroups for each resource are causing a sort of deadlock within the gateway-api runner and not allowing it to process new notifications

* removing this logic does mean that there might some churn in the dataplane during EG reboots. This should get fixed with https://github.com/envoyproxy/gateway/issues/413 which plans on reconciling all resources within a single controller so we cont need to synchonize across multiple controllers using multiple waitgroups

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -22,7 +22,6 @@ type Config struct {
 
 type Runner struct {
 	Config
-	xdsIRReady bool
 }
 
 func New(cfg *Config) *Runner {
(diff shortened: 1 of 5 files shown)


LukeShu (dismissed):
I sorta hate it because now startup is potentially racy, but it's good enough for now.

skriss (approved):
LGTM, I'll file an issue to track reimplementing this in some form for 0.3

PR #456, opened 2022-09-29
set Gateway namespace and name labels on dependent resources

### LINKED ISSUE #452: fix issues with enqueueRequestForOwningGatewayClass method
https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/gateway.go#L112-L136

Looks like this method needs some further updates for the switch to having an Envoy service/deployment per Gateway -- the name is now inaccurate, and we're comparing the value of the label (which should be a Gateway name) to the GatewayClassName field on the Gateway, which seems off. It's probably also not necessary to loop through Gateways - just return a reconcile request for the owning Gateway's namespace/name, and let the Gateway reconciler handle figuring out if it's relevant or not.

Impact is presumably that when Envoy services/deployments change, they're not triggering their owning Gateway to re-reconcile, but I haven't tested to verify impact / if it's masked by other things.

### LINKED ISSUE #453: labels on Gateway-owned resources (Envoy service, deployment) should include namespace
Currently we set a single label on any resources owned by a Gateway (e.g. the envoy service and deployment) that contains the Gateway's name. This is going to be problematic if you have Gateways in two different namespaces that have the same name. We should either put namespace/name in the label value, or add a second label containing the namespace, to allow differentiating.

A quick look indicates that maybe there's a similar problem with generating names for infra resources, but this needs further investigation.

xref #452 (fixing 452 probably requires addressing this issue as well since otherwise we don't know what the namespace of the Gateway is to enqueue a reconcile request for).

skriss:
Switches from a single label that only specifies
the Gateway name to two labels, one for namespace
and one for name, on Gateway dependent resources.
This avoids conflicts between Gateways in different 
namespaces with the same names.

Also fixes mapping from dependent resources to
the owning Gateway in the Gateway reconciler.

Closes #452.
Closes #453.

Note, I'll tackle labeling the ServiceAccount and ConfigMap in a separate PR.

### DIFF
--- a/internal/gatewayapi/testdata/gateway-allows-same-namespace-with-allowed-httproute.out.yaml
+++ b/internal/gatewayapi/testdata/gateway-allows-same-namespace-with-allowed-httproute.out.yaml
@@ -74,7 +74,8 @@ infraIR:
     proxy:
       metadata:
         labels:
-          gateway.envoyproxy.io/owning-gateway: gateway-1
+          gateway.envoyproxy.io/owning-gateway-namespace: envoy-gateway
+          gateway.envoyproxy.io/owning-gateway-name: gateway-1
       name: envoy-gateway-gateway-1
       image: envoyproxy/envoy:v1.23-latest
       listeners:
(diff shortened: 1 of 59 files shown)


arkodg (approved):
🚀

danehans (approved):

PR #458, opened 2022-09-29
Use Delta xDS by default

arkodg:


arkodg:
This seems to pass traffic conformance tests in [CI](https://github.com/envoyproxy/gateway/actions/runs/3153363561/jobs/5129749442#step:7:1) as well as looks good from local testing.

arkodg:
@danehans any specific reason this wasn't added in https://github.com/envoyproxy/gateway/pull/206 ?

AliceProxy (approved):

danehans:
No reason why Delta was not originally included in the bootstrap.

danehans (approved):

PR #460, opened 2022-09-29
Fix flow for gatewayclass deletes

arkodg:
* Delete the gatewayclass from the watchable map if it has been scheduled for deletion

* We always end up storing the accepted gatewayclass in the same controller in the same function

* Clean up the gateway-api runner code which earlier expected a gatewayclass slice with a `nil` element


skriss (approved):
:+1:

Issue #457, opened 2022-09-29
add Gateway owner labels to ServiceAccount and ConfigMap


skriss:
Currently the Deployment and Service created per Gateway are labeled with Gateway owner labels, but the ServiceAccount and ConfigMap are not. They should be labeled the same way for consistency.

PR #461, opened 2022-09-29
internal/infrastructure/kubernetes: drop Resources struct

### LINKED ISSUE #370: Should Infra Manager Resources Be Removed
Currently, Infra Manager resources are being maintained but not used for any purpose, adding another layer of state. Should this type be removed and have Infra Manage continue working with objects from the controller-runtime cache?

skriss:
Drops the Resources struct as it was only
being used to record state of the created
Kubernetes resources for testing purposes.
Replaces it with calls to the fake kube
client to fetch the resources.

Closes #370.

This will make implementing #457 easier.

cc @danehans - let me know if I misunderstood the purpose of the existing structure.

### DIFF
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -292,10 +292,6 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, infra *ir.Infra) e
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


arkodg (approved):
LGTM, this should also fix https://github.com/envoyproxy/gateway/issues/370 !

Issue #467, opened 2022-09-29
EG Panics When Removing GatewayClass


danehans:
To reproduce:
- Install a cluster.
- Install EG
- Install the quickstart.
- Uninstall the quickstart.
- EG panics.

EG Logs:
```
...
2022-09-29T20:20:57.489Z	INFO	envoy-gateway	kubernetes/gatewayclass.go:80	enqueueing gatewayclass	{"runner": "provider"}
2022-09-29T20:20:57.489Z	INFO	envoy-gateway	kubernetes/gatewayclass.go:89	reconciling gatewayclass	{"runner": "provider"}
2022-09-29T20:20:57.489Z	INFO	kubernetes/gatewayclass.go:121	failed to find an accepted gatewayclass	{"runner": "provider"}
2022-09-29T20:20:57.489Z	INFO	runner/runner.go:60	received a notification	{"runner": "gateway-api"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x15cd38c]

goroutine 93 [running]:
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate(0xc0002a2e00, {0x1e0b5b0, 0xc000370580})
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:76 +0x34c
created by github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).Start
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/runner/runner.go:37 +0x165
```

PR #468, opened 2022-09-29
Dont store nil resources in watchable

arkodg:
The arch has been revised to handle watchable stores and deletes

Fixes: https://github.com/envoyproxy/gateway/issues/467


danehans:
I confirmed that this PR fixes $467.

danehans (approved):

Issue #407, opened 2022-09-22
Improve Provider resource reconciliation


arkodg:
*Description*:
We can further improve the performance of the system by ensuring that the provider reconcilers (for GatewayClass, Gateway and HTTPRoute) reconcile only when the resource spec has changed.

Today, when the Status Updater updates the status of a resource, we receive another reconcile notification which also triggers a redundant translation in the the gateway-api layer which triggers another redundant translation in the infra and xds layers.

.

youngnick:
If we were okay with keeping some light state, we could keep the last seen `metadata.generation` for each resource, which would allow us to drop updates that don't change the spec, since `metadata.generation` is autoincremented by the apiserver each time the resource is changed.

arkodg:
> If we were okay with keeping some light state, we could keep the last seen `metadata.generation` for each resource, which would allow us to drop updates that don't change the spec, since `metadata.generation` is autoincremented by the apiserver each time the resource is changed.

ha we should be able to implement this logic, we should have some state in the watchable map !

also in the near future this enhancement will be needed since https://github.com/envoyproxy/gateway/issues/413 will convert everything into a single object and leverage watchable (using `snapshot.Updates` rather than `snapshot.State`) to inform the next layer whether anything changed or not

PR #480, opened 2022-09-30
provider: only store resource if spec has changed

arkodg:
Leverage the metadata.Generation field to consider whether to update the newly reconciled resource into the watchable map which will trigger translations in the backend.

Fixes: https://github.com/envoyproxy/gateway/issues/407

### DIFF
--- a/internal/provider/kubernetes/gateway.go
+++ b/internal/provider/kubernetes/gateway.go
@@ -202,14 +202,23 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 		// update address field and ready condition
 		status.UpdateGatewayStatusReadyCondition(&gw, svc, deployment)
 
-		// publish status
-		// do it inline and not using the status updater which does not update
-		// the Status.Addresses field.
-		if err := r.client.Status().Update(ctx, &gw); err != nil {
-			r.log.Error(err, "error updating status of gateway", "name", gw.Name)
-		}
-
 		key := utils.NamespacedName(&gw)
+		// publish status
+		// do it inline since this code flow updates the
+		// Status.Addresses field whereas the message bus / subscriber
+		// does not.
+		r.statusUpdater.Send(status.Update{
+			NamespacedName: key,
+			Resource:       new(gwapiv1b1.Gateway),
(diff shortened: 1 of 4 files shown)


chauhanshubham:
I remember that if we `Patch` the subresource the resource version didn't use to change (as compared to `Update`).
So I think if we consider using a `Patch` then that too can prevent the endless loop of reconcile -> status update -> reconcile.

I tried by replacing the Status update with 
```go
u.client.Status().Patch(context.Background(), newObj, client.MergeFrom(newObj))
```

and it seems to be working fine for me.

chauhanshubham:
Also, just checking if the `spec` has changed, sometimes doesn't go well if users are changing certain metadata fields like `annotations` or `labels`, or if a finalizer was added/removed (also deletionTimestamp) let's say.

danehans:
@arkodg now that #508 and #516 have merged, are you able to update this PR?

arkodg:
> I remember that if we `Patch` the subresource the resource version/generation doesn't change (as compared to using `Update`). So I think if we consider using a `Patch` then that too can prevent the endless loop of reconcile -> status update -> reconcile.
> 
> I tried by replacing the Status update with
> 
> ```go
> u.client.Status().Patch(context.Background(), newObj, client.MergeFrom(newObj))
> ```
> 
> and it seems to be working fine for me.

does `Patch` not trigger a reconcile ?

arkodg:
> @arkodg now that #508 and #516 have merged, are you able to update this PR?

now its blocked on https://github.com/envoyproxy/gateway/pull/535 due to Changs introduced in https://github.com/envoyproxy/gateway/pull/518

arkodg:
> Also, just checking if the `spec` has changed, sometimes doesn't go well if users (or other systems) are changing certain metadata fields like `annotations` or `labels`, or if a finalizer was added/removed (also deletionTimestamp) let's say.

for translation, the spec affects a translation change (for now) so imho this approach should be safe for now

danehans (approved):

danehans:
@arkodg should the same be done for HTTPRoutes?

arkodg:
@danehans the diffs in this commit apply to HTTPRoutes as well

PR #490, opened 2022-10-03
Delete `make lint-deps`

arkodg:
Since its not needed, the binary for each lint target is installed as part of the dependancy target

### DIFF
--- a/tools/make/lint.mk
+++ b/tools/make/lint.mk
@@ -9,20 +9,15 @@ GITHUB_ACTION ?=
 .PHONY: lint
 lint: ## Run all linter of code sources, including golint, yamllint, whitenoise lint and codespell.
 
-PHONY: lint-deps
-lint-deps: ## Everything necessary to lint (useful to separate out in the logs)
-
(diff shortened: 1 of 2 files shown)


danehans (approved):

tools/make/lint.mk
```diff
@@ -9,28 +9,22 @@ GITHUB_ACTION ?=
 .PHONY: lint
 lint: ## Run all linter of code sources, including golint, yamllint, whitenoise lint and codespell.
 
-PHONY: lint-deps
-lint-deps: ## Everything necessary to lint (useful to separate out in the logs)
```
LukeShu, line 13: IMO, when looking at CI logs it's useful to have the linter output not interspersed with ~100 `go: downloading gopkg.in/ini.v1 v1.67.0` lines as things get installed. And so that's why I created a `lint-deps` target.  If you think that value isn't worth the hassle of maintaining the `lint-deps` dependency list, then OK.

arkodg: ah cool, makes sense, I'll close this one

PR #495, opened 2022-10-04
Updates ParentRef Status Conditions

### LINKED ISSUE #494: Improper Use of UnsupportedValue HTTPRoute Condition Reason
Currently, the [RouteReasonUnsupportedValue](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1beta1@v0.5.1#RouteReasonUnsupportedValue) reason is being used with the "ResolvedRefs" condition type. In [v0.5.1](https://github.com/kubernetes-sigs/gateway-api/pull/1382), this reason should be used with the “Accepted” condition.

xref: https://github.com/envoyproxy/gateway/pull/249#discussion_r964126956

danehans:
Updates the `Accepted` condition type to use [RouteReasonUnsupportedValue](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1beta1@v0.5.1#RouteReasonUnsupportedValue) instead of the `ResolvedRefs` condition type.

Fixes #494 

### DIFF
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
@@ -71,10 +71,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: Accepted
-        message: Route is accepted
-      - type: ResolvedRefs
-        status: "False"
         reason: UnsupportedValue
         message: "RequestHeaderModifier Filter already configures request header: add-header-1 to be added, ignoring second entry"
 xdsIR:
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
@@ -71,8 +71,8 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: UnsupportedValue
-        message: "RequestHeaderModifier Filter already configures request header: add-header-1 to be added, ignoring second entry"
+        reason: Accepted
+        message: Route is accepted
 xdsIR:
   envoy-gateway-gateway-1:
     http:
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
@@ -81,9 +81,8 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: UnsupportedValue
-        # Currently only one invalid value status will be set. If there are multiple, then only the latest is displayed until that issue is resolved.
-        message: "RequestHeaderModifier Filter already configures request header: set-header-4 to be added/set, ignoring second entry"
+        reason: Accepted
+        message: Route is accepted
 xdsIR:
   envoy-gateway-gateway-1:
     http:
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
@@ -81,10 +81,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: Accepted
-        message: Route is accepted
-      - type: ResolvedRefs
-        status: "False"
         reason: UnsupportedValue
         # Currently only one invalid value status will be set. If there are multiple, then only the latest is displayed until that issue is resolved.
         message: "RequestHeaderModifier Filter already configures request header: set-header-4 to be added/set, ignoring second entry"
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
@@ -67,8 +67,8 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: UnsupportedValue
-        message: "RequestHeaderModifier Filter already configures request header: rem-header-1 to be removed, ignoring second entry"
+        reason: Accepted
+        message: Route is accepted
 xdsIR:
   envoy-gateway-gateway-1:
     http:
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
@@ -67,10 +67,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
-        reason: Accepted
-        message: Route is accepted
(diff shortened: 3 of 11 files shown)


danehans:
@skriss with this PR an `Accepted: False` route is still considered an attached route from the parent Gateway. I would expect  `Accepted: False` routes to not be attached. Is that correct?

danehans:
@arkodg from my understanding of https://github.com/envoyproxy/gateway/pull/516, this PR is still required.

skriss:
> @skriss with this PR an `Accepted: False` route is still considered an attached route from the parent Gateway. I would expect `Accepted: False` routes to not be attached. Is that correct?

Agree, I'd expect an `Accepted: False` route to not be attached to its parent(s).

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
```diff
@@ -71,10 +71,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
```
skriss, line 73: I'd think this should be `status: "False"` if there's a reason of UnsupportedValue?

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
```diff
@@ -81,10 +81,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
```
skriss, line 83: ditto

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
```diff
@@ -67,10 +67,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
```
skriss, line 69: ditto

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-removes.out.yaml
```diff
@@ -62,10 +62,6 @@ httpRoutes:
       conditions:
       - type: Accepted
         status: "True"
```
skriss, line 64: ditto

internal/gatewayapi/translator.go
```diff
@@ -839,8 +839,8 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 
 								if !canAddHeader {
 									parentRef.SetCondition(
-										v1beta1.RouteConditionResolvedRefs,
-										metav1.ConditionFalse,
+										v1beta1.RouteConditionAccepted,
+										metav1.ConditionTrue,
 										v1beta1.RouteReasonUnsupportedValue,
 										fmt.Sprintf("RequestHeaderModifier Filter already configures request header: %s to be added, ignoring second entry", headerKey),
 									)
```
skriss, line 846: From my read of the spec, we should either be ignoring the duplicate header, or appending its value:

- https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPHeaderFilter
- https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPHeader

But it doesn't seem like we need to be setting a condition here.

Regardless, the combo of `Accepted: true` with reason `UnsupportedValue` does not seem valid -- if we're using the `UnsupportedValue` reason, then the status should be `False`.

internal/gatewayapi/translator.go
```diff
@@ -895,8 +895,8 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 								}
 								if !canAddHeader {
 									parentRef.SetCondition(
-										v1beta1.RouteConditionResolvedRefs,
-										metav1.ConditionFalse,
+										v1beta1.RouteConditionAccepted,
+										metav1.ConditionTrue,
 										v1beta1.RouteReasonUnsupportedValue,
 										fmt.Sprintf("RequestHeaderModifier Filter already configures request header: %s to be added/set, ignoring second entry", headerKey),
 									)
```
skriss, line 902: ditto, probably shouldn't be setting a condition here

internal/gatewayapi/translator.go
```diff
@@ -939,8 +939,8 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 								}
 								if !canRemHeader {
 									parentRef.SetCondition(
-										v1beta1.RouteConditionResolvedRefs,
-										metav1.ConditionFalse,
+										v1beta1.RouteConditionAccepted,
+										metav1.ConditionTrue,
 										v1beta1.RouteReasonUnsupportedValue,
 										fmt.Sprintf("RequestHeaderModifier Filter already configures request header: %s to be removed, ignoring second entry", removedHeader),
 									)
```
skriss, line 946: ditto

internal/gatewayapi/translator.go
LukeShu: /cc @AliceProxy

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-add-multiple-filters.out.yaml
danehans: @skriss commit e14991a updates the PR to only set the "UnsupportedValue" reason when `Accepted=false`.

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-adds.out.yaml
danehans: @skriss commit e14991a updates the PR to only set the "UnsupportedValue" reason when `Accepted=false`.

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-removes.out.yaml
danehans: Commit e14991a updates the PR to only set the "UnsupportedValue" reason when `Accepted=false`.

internal/gatewayapi/testdata/httproute-with-header-filter-duplicate-remove-multiple-filters.out.yaml
danehans: Commit e14991a updates the PR to only set the "UnsupportedValue" reason when `Accepted=false`.

internal/gatewayapi/translator.go
danehans: > From my read of the spec, we should either be ignoring the duplicate header, or appending its value:

The code currently ignores duplicate headers from the add/remove operations. Commit e14991a no longer sets a status condition for these cases and therefore an `Accepted=true` condition is surfaced.

internal/gatewayapi/translator.go
danehans: https://github.com/envoyproxy/gateway/pull/495#discussion_r992829787

internal/gatewayapi/translator.go
danehans: https://github.com/envoyproxy/gateway/pull/495#discussion_r992829787

skriss (approved):
Looks reasonable to me, and we'll likely revisit some of this with upcoming Gateway API spec changes around conditions.

Issue #74, opened 2022-05-25
enforce code coverage thresholds


skriss:
The project should have code coverage reporting and enforce certain levels of code coverage for PRs. https://about.codecov.io/ is a good option that integrates well with GitHub and is free for open-source projects.

ethanhanjiahao:
/accept

I want try to resolve this issue

danehans:
@ethanhanjiahao thanks for your willingness to help. Please assign yourself the issue if you're going to work on it.

ethanhanjiahao:
> @ethanhanjiahao thanks for your willingness to help. Please assign yourself the issue if you're going to work on it.

yeah, but i cannot assign to me. Can you assign the issue to me. I will work on it

skriss:
Still to do: add PR checks that enforce required code coverage #'s.

danehans:
Moving this to backlog since codecov was added but coverage #s are not required.

arkodg:
we are currently at ~60% [code coverage](https://app.codecov.io/gh/envoyproxy/gateway) and this is a decent `threshold` to start off with imho.
based on https://github.com/codecov/codecov-action/issues/554 I'm unsure if `threshold` is supported by the GHA or not. 
Hey @ethanhanjiahao is this something you could further look into and enhance in our CI ?

ethanhanjiahao:
> we are currently at ~60% [code coverage](https://app.codecov.io/gh/envoyproxy/gateway) and this is a decent `threshold` to start off with imho. based on [codecov/codecov-action#554](https://github.com/codecov/codecov-action/issues/554) I'm unsure if `threshold` is supported by the GHA or not. Hey @ethanhanjiahao is this something you could further look into and enhance in our CI ?

yes， i can enhance it

danehans:
@ethanhanjiahao can you please confirm if you can fix this issue by the [v0.2.0 milestone](https://github.com/envoyproxy/gateway/milestone/6)?

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/513

arkodg:
reopening this issue since a drop a coverage didn't manage to fail the CI check (seen https://github.com/envoyproxy/gateway/pull/728#discussion_r1042709751)

ethanhanjiahao:
> reopening this issue since a drop a coverage didn't manage to fail the CI check (seen [#728 (comment)](https://github.com/envoyproxy/gateway/pull/728#discussion_r1042709751))

[...]

PR #513, opened 2022-10-06
Update codecov.yml for threshold check

ethanhanjiahao:
* add coverage threshold ci test, from base commit coverage to compare against. and we not allow coverage drop. ref: <https://docs.codecov.com/docs/commit-status>

Signed-off-by: ethan <hanjiahao@corp.netease.com>
issue: https://github.com/envoyproxy/gateway/issues/74

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -3,6 +3,6 @@ coverage:
     project:
       default:
         # basic
-        target: auto
-        threshold: 0%
+        target: 60%
+        threshold: 5%
         if_ci_failed: error
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: auto
+        threshold: 0%
+        if_ci_failed: error


arkodg:
hey @ethanhanjiahao looks like lint is failing, you can run `make lint` locally to better understand the error

.github/codecov.yml
```diff
@@ -0,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: auto
+        threshold: 0%
```
arkodg, line 7: can the config here be updated to ensure that total coverage never drops below 60% and CI fails if PR reduces coverage by 5%

arkodg (approved):
🚀  LGTM thanks for enhancing the CI !

.github/codecov.yml
ethanhanjiahao: cc @arkodg ，set target: 60%, threshold: 5%  is ok

arkodg:
reopening this issue since a drop a coverage didn't manage to fail the CI check (seen [here](https://github.com/envoyproxy/gateway/pull/728#discussion_r1042709751))

PR #514, opened 2022-10-06
[docs] add certgen install/uninstall in quickstart, other minor updates

chauhanshubham:


docs/user/QUICKSTART.md
```diff
@@ -12,6 +12,11 @@ Install the Gateway API CRDs:
 kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/v0.2.0-rc2/gatewayapi-crds.yaml
 ```
 
+Run the certgen Job required for Envoy Gateway:
```
arkodg, line 15: certgen should get installed as part of `install.yaml`, is that not happening ?

chauhanshubham: right, I think this happened for me because I was running the eg binary locally.

danehans: @chauhanshubham you are correct, the issue is related to running EG locally. I would like us to develop a solution to properly run EG locally after the v0.2.0. I think it's a good idea to comment out the run-kube-local target until EG is fully functional when running locally.

docs/design/gatewayapi-translator.md
```diff
@@ -1,13 +1,13 @@
...
 ## Assumptions
-- initially target core conformance features only, to be followed by extended conformance features
+- Initially target core conformance features only, to be followed by extended conformance features
 
 ## Inputs and Outputs
 
 The main inputs to the Gateway API translator are:
 - the GatewayClass to process
-- Gateways, HTTPRoutes, Services, Secrets
+- Gateways, HTTPRoutes, Services, Secrets, ReferenceGrants
```
danehans, line 10: @chauhanshubham TLSRoutes should be added as part of your TLS pass through work.

Issue #415, opened 2022-09-23
HTTPRoute Status Not Updated After Adding Backend


danehans:
I create a gatewayclass, gateway, and httproute. EG surfaces an error:
```
2022-09-23T16:51:34.608Z	ERROR	controller/controller.go:326	Reconciler error	{"runner": "provider", "controller": "httproute", "object": {"name":"httpbin","namespace":"default"}, "namespace": "default", "name": "httpbin", "reconcileID": "1d30e643-0896-4171-9323-7fc6aa75e865", "error": "failed to get service default/httpbin"}
```

I then create the backend service but the route's status is never updated:
```
...
  status:
    parents:
    - conditions:
      - lastTransitionTime: "2022-09-23T16:51:30Z"
        message: Route is accepted
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      - lastTransitionTime: "2022-09-23T16:51:30Z"
        message: Service default/httpbin not found
        observedGeneration: 1
        reason: BackendNotFound
        status: "False"
        type: ResolvedRefs
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
      parentRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
```

cc: @arkodg @skriss

danehans:
I'm seeing this issue even when the backend is created before the httproute.

danehans:
I was able to work around the issue by deleting the httproute, gateway, and gatewayclass. After recreating the gatewayclass, gateway, and httproute (after creating the backend) the httproute surfaced the proper status conditions:
```
  status:
    parents:
    - conditions:
      - lastTransitionTime: "2022-09-23T17:10:28Z"
        message: Route is accepted
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
      parentRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
```

arkodg:
sounds like an issue in the provider reconciler, maybe another reason to implement https://github.com/envoyproxy/gateway/issues/413 sooner than later

arkodg:
[...]

Issue #448, opened 2022-09-28
on startup, don't start translation until all resources have been loaded


skriss:
xref #447
xref #413
xref https://github.com/telepresenceio/watchable/pull/6 

On startup/restart, EG should not start translating/sending xDS traffic until all resources have been loaded, to avoid issues with incomplete inputs.

danehans:
Note that this fix is needed to ensure `lastTransitionTime` is not updated when EG is restarted.

danehans:
This issue should be resolved when #413 is fixed.

PR #516, opened 2022-10-06
Reset status conditions in GatewayAPI translator

arkodg:
* reset the conditions field within the listeners field (Gateway status) and the the parents field (HTTPRoute Status) before further setting any newer conditions so older conditons that are no longer true can be reset.

Fixes: https://github.com/envoyproxy/gateway/issues/415


danehans (approved):
Note that `lastTransitionTime` for HTTPRoutes and Gateway listeners will still update when EG is restarted until https://github.com/envoyproxy/gateway/issues/448 is fixed.

Issue #314, opened 2022-08-31
Add Shellcheck Linter


danehans:
Scripts continue to be added to the project, mainly to support Make targets. Shellcheck should be used to for static analysis and lint tool, for these scripts.

danehans:
Reopening since I see the following error when running `make lint` locally:
```
...
Running Shellcheck linter ...
/bin/sh: shellcheck: command not found
make[1]: *** [lint.shellcheck] Error 127
```

I understand that spellchecker is installed in GitHub Actions by default but Devs should be able to run the target locally without error.

Xunzhuo:
Right, shelchecker is pre-installed in GA, we can add it in local.

danehans:
> ... we can add it in local.

@Xunzhuo are you able to submit a PR to add local support of `spellcheck`?

Xunzhuo:
Ok, sure. @danehans

PR #521, opened 2022-10-08
chore: add local shellcheck tools

Xunzhuo:
Resolve: https://github.com/envoyproxy/gateway/issues/314

### DIFF
--- a/tools/make/golang.mk
+++ b/tools/make/golang.mk
@@ -12,7 +12,7 @@ GO_VERSION = $(shell grep -oE "^go [[:digit:]]*\.[[:digit:]]*" go.mod | cut -d' 
 # Build the target binary in target platform.
 # The pattern of build.% is `build.{Platform}.{Command}`.
 # If we want to build envoy-gateway in linux amd64 platform, 
-# just execute make build.linux_amd64.envoy-gateway.
+# just execute make go.build.linux_amd64.envoy-gateway.
 .PHONY: go.build.%
 go.build.%:
 	$(eval COMMAND := $(word 2,$(subst ., ,$*)))
--- a/tools/make/lint.mk
+++ b/tools/make/lint.mk
@@ -58,12 +58,13 @@ lint.whitenoise: $(tools/whitenoise)
 	@echo Running WhiteNoise linter ...
 	$(tools/whitenoise)
 
-# GitHub has shellcheck pre-installed
+
 .PHONY: lint.shellcheck
 lint: lint.shellcheck
-lint.shellcheck:
+lint-deps: $(tools/shellcheck)
+lint.shellcheck: $(tools/shellcheck)
 	@echo Running Shellcheck linter ...
-	@shellcheck tools/hack/*.sh
+	$(tools/shellcheck) tools/hack/*.sh
 
 .PHONY: gen-check
 gen-check: generate manifests
--- a/tools/make/tools.mk
+++ b/tools/make/tools.mk
@@ -33,3 +33,23 @@ $(tools.bindir)/%.d/venv: $(tools.srcdir)/%/requirements.txt
 	$@/bin/pip3 install -r $< || (rm -rf $@; exit 1)
 $(tools.bindir)/%: $(tools.bindir)/%.d/venv
 	ln -sf $*.d/venv/bin/$* $@
+
+ifneq ($(GOOS),windows)
+# Shellcheck
+# ==========
+#
+tools/shellcheck = $(tools.bindir)/shellcheck
+SHELLCHECK_VERSION=0.8.0
+SHELLCHECK_ARCH=$(shell uname -m)
+# shellcheck uses the same binary on Intel and Apple Silicon Mac.
+ifeq ($(GOOS),darwin)
+SHELLCHECK_ARCH=x86_64
+endif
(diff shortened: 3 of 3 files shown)


danehans (commented):
LGTM. @arkodg @lizan WDYT?

tools/src/shellcheck.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+SHELLCHECK_VERSION=v0.8.0
+GOARCH=$(go env GOARCH)
+GOOS=$(go env GOOS)
+
+cd tools/bin && wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.${GOOS}.x86_64.tar.xz" | tar -xJv
```
zirain, line 6: do not `wget` every time run this script?

Xunzhuo: Thanks for raising this. Yes, we should check if binary existed.

tools/src/shellcheck.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+SHELLCHECK_VERSION=v0.8.0
+
+GOARCH=$(go env GOARCH)
+GOOS=$(go env GOOS)
+
+RELEASE_URL="https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.${GOOS}.x86_64.tar.xz"
```
arkodg, line 7: will not work for arm machines

tools/src/shellcheck.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+SHELLCHECK_VERSION=v0.8.0
+
+GOARCH=$(go env GOARCH)
+GOOS=$(go env GOOS)
+
+if [[ "$GOARCH" == "amd64" || "$GOOS" == "darwin" ]]; then
```
arkodg, line 8: why is the `darwin` check needed ? I think above `if` will be true for M1 Macs (`darwin/arm64`) which we dont want

Xunzhuo: In shell release page, there is only x64 binary for darwin

arkodg: shellcheck is supported for `darwin/arm64` via homebrew https://github.com/koalaman/shellcheck/issues/2229#issuecomment-858753596
if we merge this, we'll drop support for `darwin/arm64`, since the M1 Mac user cannot use their local `shellcheck` neither can they download it

api/config/v1alpha1/groupversion_info.go
```diff
@@ -1,7 +1,7 @@
 // Package v1alpha1 contains API Schema definitions for the config v1alpha1 API group.
 //
-//+kubebuilder:object:generate=true
-//+groupName=config.gateway.envoyproxy.io
+// +kubebuilder:object:generate=true
+// +groupName=config.gateway.envoyproxy.io
```
LukeShu, line 4: This is the result of `gofmt` 1.19, which is different than the `gofmt` 1.18 that we're still using. This should be in a separate Go 1.19 PR.

internal/status/status.go
```diff
@@ -140,9 +140,10 @@ func (u *UpdateWriter) Send(update Update) {
 // isStatusEqual checks if two objects have equivalent status.
 //
 // Supported objects:
-//  GatewayClasses
-//  Gateway
-//  HTTPRoute
+//
+//	GatewayClasses
+//	Gateway
+//	HTTPRoute
```
LukeShu, line 146: This is the result of `gofmt` 1.19, which is different than the `gofmt` 1.18 that we're still using. This should be in a separate Go 1.19 PR.

Plus, this should be formatted as a bullet-list, not a code-block (godoc 1.19 added support for bullet lists).

tools/src/shellcheck.sh
LukeShu: I don't think I like having a wrapper shellscript do tis.  `tools.mk` should just know how to download shellcheck and install that binary to `tools/bin/shellcheck`.

Here's Telepresence's rule for doing that: https://github.com/telepresenceio/telepresence/blob/4c6fc8953adc9012a3a5b85856a9802c51e84cac/build-aux/tools.mk#L80-L97

LukeShu (changes requested):

tools/src/shellcheck.sh
LukeShu: IIRC, despite "x86_64" being in the filename, the `darwin.x86_64` file is actually a "fat" binary with compiled code for both x86_64 and arm64.

arkodg: while @Xunzhuo addresses @LukeShu 's suggestions, @chauhanshubham can you help verify these PR changes on a `darwin/arm64` machine ?

Xunzhuo: I am using M1 Mac, it works fine.

api/config/v1alpha1/groupversion_info.go
Xunzhuo: Thanks, I will remove this.

Xunzhuo:
Thanks @LukeShu, updated based on your comments.

LukeShu (approved):
Thanks!

PR #524, opened 2022-10-09
feat: implement liveness and readiness probes

Xunzhuo:
Resolves: https://github.com/envoyproxy/gateway/issues/235

Tested locally and works well:

image

image


internal/provider/kubernetes/kubernetes.go
```diff
@@ -60,6 +62,17 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 	}, nil
 }
 
+// ServeHealthProbes starts serving health probes, including health check and ready check.
```
arkodg, line 65: curious why not add this into New() ?

Xunzhuo: Right, we can add to New(), no need to add another func.

test/conformance/conformance_test.go
```diff
@@ -50,7 +50,6 @@ func TestGatewayAPIConformance(t *testing.T) {
 		tests.HTTPRouteInvalidNonExistentBackendRef,
 		tests.HTTPRouteInvalidBackendRefUnknownKind,
 		tests.HTTPRouteInvalidCrossNamespaceBackendRef,
-		tests.HTTPRouteHeaderMatching,
```
arkodg, line 53: why is this removed ?

Xunzhuo: Looks like this is duplicated. I did not notice that you already added this.

arkodg (dismissed):

arkodg (approved):

PR #529, opened 2022-10-09
Add authn policy design with JWT only

lizan:


docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+  jwt:
+    iss: https://www.okta.com
+    aud: bookinfo.com
+    jwksUri: https://bookinfo.com/jwt/public-key/jwks.json
+  targetRef:
+    kind: HTTPRoute
+    name: httpbin
+```
+
+JWT Bearer token will be translate to Envoy's JWT authentication filter. The JWKS URI need to be translated to a separate cluster for JWKS fetch and refersh.
```
arkodg, line 40:
```suggestion
JWT Bearer token will be translate to Envoy's JWT authentication filter. The JWKS URI need to be translated to a separate cluster for JWKS fetch and refresh.
```

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+A JWT Bearer Token authentication policy will look like the following:
+
+```
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Authentication
+metadata:
+  name: productpage
+spec:
+  type: jwt
+  jwt:
```
arkodg, line 31: if multiple issuers exist, does the user create multiple authn policies or will this policy allow the user to specify a list of issuers and associated jwksUri ?

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+    iss: https://www.okta.com
+    aud: bookinfo.com
+    jwksUri: https://bookinfo.com/jwt/public-key/jwks.json
+  targetRef:
+    kind: HTTPRoute
```
arkodg, line 36: will Gateway `targetRef` be supported ?

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
+# Envoy Gateway Authentication Policy API
+
+## Overview
+
+This authen policy is to declare the authentication mechanisms, to be enforce on connection and request going though Envoy Gateway. This includes the credential (X.509, JWT, etc), parameters (cipher suites, key algorithms)
```
arkodg, line 5:
```suggestion
This authentication policy declares the authentication mechanisms, to be enforced on connection and request going though Envoy Gateway. This includes the credential (X.509, JWT, etc), parameters (cipher suites, key algorithms)
```

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Authentication
+metadata:
+  name: productpage
+spec:
+  type: jwt
+  jwt:
+    iss: https://www.okta.com
+    aud: bookinfo.com
+    jwksUri: https://bookinfo.com/jwt/public-key/jwks.json
```
arkodg, line 34: should we support custom `jwks` string ? I prefer if we defer that to later

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+```
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Authentication
+metadata:
+  name: productpage
+spec:
+  type: jwt
+  jwt:
+    iss: https://www.okta.com
+    aud: bookinfo.com
```
arkodg, line 33: should this be a list ?
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#remote-jwks-config-example

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+
+A JWT Bearer Token authentication policy will look like the following:
+
+```
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Authentication
+metadata:
+  name: productpage
+spec:
+  type: jwt
```
arkodg, line 30: are we going to expose `from_headers` ? I'd vote for not now, and prefer to use the defaults in Envoy 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider

danehans:
@lizan as we discussed during today's Envoy Gateway community meeting, this design should be reviewed with the Gateway API GAMMA team. It would be beneficial for the extension to be supported by both projects.

docs/design/authentication_policy.md
```diff
@@ -0,0 +1,40 @@
...
+This policy should support the following authentication mechanisms:
+- JWT Bearer Token
+- mutualTLS (client certificate)
+- OAuth2
+- OIDC
+- External authentication
+
+In the first phase, Envoy Gateway will implement JWT Bearer Token authentication.
+
+In general those policy translates into Envoy HTTP filters at HTTP connection manager level, and route specific settings will be applied for each route. These APIs are expressed in a Policy CRD and attached to Gateway API resources with [Policy Attachement](https://gateway-api.sigs.k8s.io/references/policy-attachment/).
```
sunjayBhatia, line 18: re: `route specific settings will be applied for each route`

The Gateway API Policy `targetRef` field does not yet have a way to specify a sub-section of a resource, so applying a Policy to an individual route *rule* is not "directly" possible, unless you require users to separate that route rule into its own HTTPRoute resource (in this case you might have a Policy target a Gateway and another target an attached HTTPRoute that specifies some overrides just for it)

Am I understanding the intent here properly? Might be good to flesh out some more examples on how this might work

docs/design/authentication_policy.md
sunjayBhatia: yeah agreed, it would be good to clarify what resources will be allowed to be targeted by the `Authentication` Policy and further, what sort of override/defaulting behavior should be expected, re this section of the reference: https://gateway-api.sigs.k8s.io/references/policy-attachment/#hierarchy

sunjayBhatia (commented):
Great to see some Gateway API extension CRDs being worked on, we're doing the same over at Contour, still in early staged trying to get settled how we will approach the extension mechanisms generally

Left a few comments and questions

If you like we can talk more broadly about how to approach the extension mechanisms, I've started some notes for Contour specifically here: https://github.com/projectcontour/contour/pull/4749 but there is some background info that is more widely applicable as well

Also got some example spikes on how one might implement local rate limiting with Filters vs. Policies here:
- https://github.com/projectcontour/contour/pull/4775
- https://github.com/projectcontour/contour/pull/4776

arkodg:
@lizan any new insights wrt this PR ?
this PR / Issue will be setting the standard for other advanced features [scheduled in this release](https://github.com/envoyproxy/gateway/blob/main/docs/design/roadmap.md#v030-drive-advanced-features-through-extension-mechanisms) and should be the reference pattern for those features such as [ratelimiting](https://github.com/envoyproxy/gateway/issues/670), [cors](https://github.com/envoyproxy/gateway/issues/567), etc

The one thing I'd add is to consider limiting the Policy CRDs to 1 or 2 i.e.

1. one generic policy CRD with features as top level fields
```
 kind: Policy
 .....
 spec:
   authentication:
   .....
   ratelimiting:
   ....
 ```
2. one generic policy CRD with `networking`  (shapes traffic) and `security` (authenticates & authorizes traffic) as top fields
```
 kind: Policy
 .....
 spec:
   security:
     authentication:
     ....
   networking:
     ratelimiting:
     .....
 ```
3. 2 Policy CRDs - one for `networking` and the other for `security`
```
 kind: SecurityPolicy
 ....
 ....
 spec:
   authentication:
```   

```
 kind: NetworkingPolicy
 ....
 spec:
   ratelimiting:
   ....
```

docs/design/authentication_policy.md
sunjayBhatia: see also: https://github.com/kubernetes-sigs/gateway-api/discussions/1489

arkodg: thanks for writing down that in depth analysis ! 
and surfacing some limitations with the PolicyAttachment that exist today.
Hoping those get incorporated soon, @youngnick has already acknowledged this !

Until then, since we plan on adding new features such as `authentication` in our next v0.3.0 release, imho we should peruse the existing spec for now

danehans:
During today's EG community meeting, a consensus was achieved for using separate API types to express request authentication and other advanced Envoy features, e.g. #671, #670, etc.

arkodg:
> During today's EG community meeting, a consensus was achieved for using separate API types to express request authentication and other advanced Envoy features, e.g. #671, #670, etc.

@danehans raised https://github.com/envoyproxy/gateway/issues/677 so all our votes are recorded and maintainers who weren't able to attend can also vote.

docs/design/authentication_policy.md
danehans: +1

docs/design/authentication_policy.md
danehans: From the Gateway API [Policy Attachment spec](https://gateway-api.sigs.k8s.io/references/policy-attachment/):

> Although Policy attachment can be used to target an entire Route or Backend, it cannot currently be used to target specific Route rules or backend references.

Based on this information, it appears that an authentication API fits better as an [implementation-specific filter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter) since HTTPRoute rules may require different authentication specs. In the future, EG can add GWAPI Policy Attachment support for authentication defaulting and overrides.

docs/design/authentication_policy.md
danehans: Why is a mutual binding required? A user should be able to create the ns-scoped authentication resource specified by this PR that describes details, e.g. type of authentication. Any resource in the ns that supports `kind: OurAuthKind`, e.g. HTTPRoute, should be able to reference it.

docs/design/authentication_policy.md
sunjayBhatia: HTTPRoute Filters themselves also have some drawbacks as they may not directly translate to Envoy HTTP Filters, given the scope they are configurable at.

HTTPRoute Filters can be configured per-Route Rule or per-Backend and Envoy HTTP Filters mostly have the bulk of their configuration as part of the Filter in a per-virtualhost HTTPConnectionManager FilterChain. I say "mostly" here because many filters enable per-Route enablement/disablement, but most of the configuration (e.g. JWT provider specification: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-jwtauthentication) lives at the higher level Filter object.

We may be able to have per-route JWT providers we collect from HTTPRoute Filter objects and then configure on the Envoy JWT HTTP Filter, but I haven't actually tried it so not sure how complex that will get.

Contour implemented this recently in HTTPProxy here:
- https://github.com/projectcontour/contour/blob/4badccc7267aeec2644f9122ed7ca1547934a5b5/design/jwt-verification-design.md
- https://projectcontour.io/docs/v1.23.0/config/jwt-verification/

Curious to see your thoughts @lizan if you've gotten a chance to spike this out as a Policy vs. a Filter

docs/design/authentication_policy.md
lizan: It is not a mutual binding, this is a Policy Attachement and not referenced from HTTPRoute or Gateway.

danehans: Thanks for the clarification. Based on this draft [doc](https://docs.google.com/document/d/1p_kmAOmqebX8zyHnXW5I58cD8ZhNX4QNSa_gw8uAYxw/edit?usp=sharing) from @youngnick, the policy object should not refer to a specific instance of a supported kind. Additionally, the Policy Attachment should be used to set defaults and overrides and not attaching policy.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+Authentication should support additional authentication types in the future, for example:
+- mutualTLS (client certificate)
+- OAuth2
+- OIDC
+- External authentication
+
+## Outstanding Questions
+
+- If Envoy Gateway owns the Authentication API, is an xDS IR equivalent needed?
+- Should local JWKS be implemented before remote JWKS?
```
lizan, line 371: Prefer remote JWKS first, local JWKS could be later.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+- How should Envoy obtain the trusted CA for a remote JWKS?
```
lizan, line 372: Default Envoy image includes system CA and that should work for the most cases.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+- Should HTTPS be the only supported scheme for remote JWKS?
```
lizan, line 373: Ideally yes, but for some local network fetch, HTTP might be also supported.

lizan (commented):
Merged @danehans 's refactor to this PR to move discussion here. Answering some outstanding questions for discussion.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+
+[Issue 336][] specifies the need for exposing a user-facing API to configure request authentication. Request
+authentication is defined as an authentication mechanism to be enforced by Envoy on a per-request basis. A connection
+will be rejected if it contains invalid authentication information, based on the `Authentication` API type proposed in
+this design document.
+
+Envoy Gateway leverages [Gateway API][] for configuring managed Envoy proxies. Gateway API defines core, extended, and
+implementation-specific API [support levels][] for implementors such as Envoy Gateway to expose features. Since
+implementing request authentication is not covered by `Core` or `Extended` APIs, an `Implementation-specific` API will
+be created for this purpose.
```
danehans, line 13: The key point of this design is the Authentication API is meant to be an __implementation-specific filter__. The API does not follow the PolicyAttachment specification. The Authentication API defined by this design will allow App Devs to easily attach authentication to API endpoints. At a future date, I expect Envoy Gateway to support an AuthenticationPolicy API that follows the PolicyAttachment spec and allow Infra Admins to override or set defaults for fields defined by the Authentication API. The direction of this design will change if maintainers agree via #675 that PolicyAttachment is the preferred method for allowing App Devs to attach authentication to API endpoints.

cc: @youngnick @sunjayBhatia @skriss @AliceProxy @Xunzhuo

docs/design/request-authentication.md
danehans: @lizan why do you prefer local over remote? It seems that local is easier to implement.

docs/design/request-authentication.md
youngnick: Okay, I posted an update to #675 (https://github.com/envoyproxy/gateway/issues/675#issuecomment-1309767753), that I think should help understand this, but in short:
- I think Policy Attachment is _preferred overall_, **but**
- in this case, I think a custom filter is okay for now.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+	//
+	//   <INSERT_LINK>
+	//
+	// +optional
+	Foo *FooAuthentication `json:"foo"`
+}
+```
+
+Authentication should support additional authentication types in the future, for example:
+- mutualTLS (client certificate)
```
sunjayBhatia, line 363: does this mean mutual TLS with the app backend or client mutual TLS with the Envoy proxy?

tbh regardless, should we exclude this one from this level of configuration for now, since it is a connection level mechanism? This seems like something for the former that we will be able to use the "Backend Properties" GEP to do and for the latter something we need to configure on a Gateway Listener

lizan: This is to support client certificate to Envoy Gateway, so nothing to do with gateway to app backend (i.e. "Backend Properties").

This statements is actually meant to exclude mTLS from this design so deferred to future.

docs/design/request-authentication.md
```diff
@@ -0,0 +1,384 @@
...
+        - path:
+            type: PathPrefix
+            value: /foo
+      filters:
+        - type: ExtensionRef
+          extensionRef:
+            group: security.gateway.envoyproxy.io
+            kind: Authentication
+            name: example-jwt
+      backendRefs:
```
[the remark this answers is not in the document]
danehans: @sunjayBhatia thanks for pointing out the [BackendRefs](https://gateway-api.sigs.k8s.io/concepts/api-overview/?h=extension#extension-points) extension point. I suggest we support using the Authentication API as an HTTPRoute filter rule. I'll update the doc so this is clear.

docs/design/request-authentication.md
lizan: Because remote JWKS is used more widely and it is critical for OIDC. The OIDC Discovery document has `jwks_uri` for fetching JWKS.

danehans:
Based on https://github.com/envoyproxy/gateway/issues/675#issuecomment-1316024761, maintainers have agreed to use an HTTPRoute implementation-specific filter to implement request authentication. The design doc currently follows this approach.

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,537 @@
...
+exhaustive list of features for authentication support in Envoy Gateway.
+
+As a Service Producer, I need the ability to:
+* Authenticate a request before forwarding it to a backend service.
+* Have different authentication mechanisms per route rule.
+* Choose from different authentication mechanisms supported by Envoy, e.g. OIDC.
+
+### Security API Group
+
+A new API group, `security.gateway.envoyproxy.io` is introduced to group security-related APIs. This will allow security
```
arkodg, line 42: curious why a new group is needed ?
why not graduate individual extensions from `v1alpha1` to `v1beta1` similar to what Gateway API does https://github.com/kubernetes-sigs/gateway-api/tree/main/apis

danehans: A group is not needed per se. The reason for the group is documented. I don't have a strong opinion on the topic. We grouped the config APIs, so we've already established a pattern.

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,537 @@
...
+}
+```
+
+Authentication should support additional authentication types in the future, for example:
+- mutualTLS (client certificate)
+- OAuth2
+- OIDC
+- External authentication
+
+## Outstanding Questions
```
danehans, line 518: I would appreciate any feedback to help resolve these outstanding questions.

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,537 @@
...
+
+- If Envoy Gateway owns the Authentication API, is an xDS IR equivalent needed?
+- Should local JWKS be implemented before remote JWKS?
+- How should Envoy obtain the trusted CA for a remote JWKS?
+- Should HTTPS be the only supported scheme for remote JWKS?
+- Should OR'ing JWT providers be supported?
+- Should Authentication provide status?
```
arkodg, line 525: would be good to add, but imho doesnt need to be added right away

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,537 @@
...
+	// +optional
+	Foo *FooAuthentication `json:"foo"`
+}
+```
+
+Authentication should support additional authentication types in the future, for example:
+- mutualTLS (client certificate)
+- OAuth2
+- OIDC
+- External authentication
```
arkodg, line 516: prefer if we removed `external authentication` for now, this falls under `expert-mode`, and could be achieved by configuring xds resources directly

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,537 @@
...
+
+```yaml
+configs:
+...
+dynamic_listeners:
+  - name: default-eg-http
+    ...
+        default_filter_chain:
+          filters:
+          - name: envoy.filters.network.http_connection_manager_1
```
lizan, line 338: This won't work, HCM is a terminal filter, so any filter after that won't be effect and be rejected.

docs/design/authentication_policy.md
danehans: https://github.com/envoyproxy/gateway/pull/733 provides details. Multiple `jwtProviders` can be specified in an Authentication. The JWT auth passes if any one of the providers responds with a success. If an HTTPRoute references multiple Authentications, then at least one `jwtProviders` for each Authentication must pass for JWT auth to succeed.

docs/design/authentication_policy.md
danehans: Considering this resolved since we agreed to take a route filter approach for authN.

docs/design/authentication_policy.md
danehans: #733 specifies `JwtProviders` as a list.

docs/design/authentication_policy.md
danehans: #733 specifies `RemoteJWKS` as a user-defined type with `URI` as a string to fetch the JWKS.

docs/design/authentication_policy.md
danehans: #733 does not specify `from_headers`. As you mention, this can be introduced in the future.

docs/design/authentication_policy.md
danehans: > HTTPRoute Filters can be configured per-Route Rule or per-Backend

#733 implements authN as an [HTTPRoute filter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).

> We may be able to have per-route JWT providers we collect from HTTPRoute Filter objects and then configure on the Envoy JWT HTTP Filter, but I haven't actually tried it so not sure how complex that will get.

I am going to test this as an approach for implementing authN. @lizan please comment if you have any suggestions.

docs/design/request-authentication.md
danehans: Considering this resolved due to https://github.com/envoyproxy/gateway/pull/529#issuecomment-1299355946.

docs/design/request-authentication.md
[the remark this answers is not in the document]
danehans: The PR has been updated to state authN is implemented as an [HTTPRoute filter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).

docs/latest/design/request-authentication.md
danehans: @lizan I have updated the JWT xDS configuration example, PTAL.

docs/latest/design/request-authentication.md
danehans: Done in commit [8c2fde2](https://github.com/envoyproxy/gateway/pull/529/commits/8c2fde2f616a38b96cf602e0d40d5c6ebc2bbe3e).

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,507 @@
...
+                            cluster: default-example2-example2-jwt
+                    rules:
+                      - match:
+                          exact: /foo
+                        requires:
+                          provider_name: default-example1-example1
+                      - match:
+                          exact: /bar
+                        requires:
+                          provider_name: default-example2-example2
```
lizan, line 377: This works, though we might end up with [per route config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-perrouteconfig) and requirement map when this grows.

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,507 @@
...
+```
+
+Authentication should support additional authentication types in the future, for example:
+- OAuth2
+- OIDC
+
+## Outstanding Questions
+
+- If Envoy Gateway owns the Authentication API, is an xDS IR equivalent needed?
+- Should local JWKS be implemented before remote JWKS?
```
kflynn, line 491: Given the next two questions about remote JWKS, I'd definitely vote in favor of local first.

docs/latest/design/request-authentication.md
kflynn: Definitely worth considering what status updates would make sense.

kflynn (approved):
Overall I think this looks really good. As discussed in the 29 November Envoy Gateway meeting, I think it's important that this stay at alpha until we have real-world usage to back it up.

docs/latest/design/request-authentication.md
```diff
@@ -0,0 +1,507 @@
...
+* Authenticate a request before forwarding it to a backend service.
+* Have different authentication mechanisms per route rule.
+* Choose from different authentication mechanisms supported by Envoy, e.g. OIDC.
+
+### Security API Group
+
+A new API group, `security.gateway.envoyproxy.io` is introduced to group security-related APIs. This will allow security
+APIs to be versioned, changed, etc. over time.
+
+### Authentication API Type
```
youngnick, line 45: I'd recommend we make clear that this is for use as a Filter by calling this `AuthenticationFilter`.

docs/latest/design/request-authentication.md
lizan: Not sure why my previous comment is not shown anymore https://github.com/envoyproxy/gateway/pull/529#discussion_r1018478689

Remote JWKS first is strongly preferred as for most of public JWT issuer, keys are rotated quite frequently. (e.g. Google OAuth2 public key https://www.googleapis.com/oauth2/v3/certs rotates daily IIRC).

Rest of questions answered there as well.

danehans:
> Definitely worth considering what status updates would make sense.

@kflynn I agree. I would like to add status as a follow-on after implementing the API and gaining an understanding of how it's used.

danehans:
> Overall I think this looks really good. As discussed in the 29 November Envoy Gateway meeting, I think it's important that this stay at alpha until we have real-world usage to back it up.

@kflynn agreed and why I created https://github.com/envoyproxy/gateway/issues/714.

docs/latest/design/request-authentication.md
danehans: Responded here: https://github.com/envoyproxy/gateway/pull/529#issuecomment-1332825708

danehans:
commit [947359c](https://github.com/envoyproxy/gateway/commit/947359c71dae95a0e725262014dc039cc21083f9) removes the security API group and renames the API to `AuthenticationFilter`.

cc: @arkodg @youngnick

docs/latest/design/request-authentication.md
danehans: Renames in [947359c](https://github.com/envoyproxy/gateway/commit/947359c71dae95a0e725262014dc039cc21083f9)

docs/latest/design/request-authentication.md
danehans: +1 on prioritizing remote jwks.

arkodg (approved):

Issue #182, opened 2022-07-29
Convert Manager Struct to Interface


danehans:
Convert the infra `Manager` struct into an interface. This should be done when additional infra providers are supported. Have all providers implement the interface.

xref: https://github.com/envoyproxy/gateway/pull/178#discussion_r932701218

PR #530, opened 2022-10-10
feat: set mgr to interface

Xunzhuo:
Resolve: https://github.com/envoyproxy/gateway/issues/182

Since this is a backlog issue, I will work on this after releasing v0.2.0.

### DIFF
--- a/internal/infrastructure/manager.go
+++ b/internal/infrastructure/manager.go
@@ -1,6 +1,7 @@
 package infrastructure
 
 import (
+	"context"
 	"fmt"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


LukeShu:
Looks OK to me, but I know that @danehans was hesitant to create an `interface` until we had a second provider to have an idea of what things might need to be different between providers.

Xunzhuo:
Thanks for this comment @LukeShu, yes, so this can wait until we got another provider, and we can do better abstraction.

internal/infrastructure/manager.go
```diff
@@ -10,26 +11,30 @@ import (
 	"github.com/envoyproxy/gateway/internal/envoygateway"
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/infrastructure/kubernetes"
+	"github.com/envoyproxy/gateway/internal/ir"
 )
 
+var _ Manager = (*kubernetes.Infra)(nil)
```
arkodg, line 17: why is this needed ?

arkodg:
should be fine to get this in, and iterate, thanks for adding this @Xunzhuo

internal/infrastructure/manager.go
Xunzhuo: Just make sure kubernetes.Infra have implemented the interface.:)

arkodg (approved):

Issue #441, opened 2022-09-28
Hash the Names of Managed Infra Resources


danehans:
EG currently generates names of managed envoy infra. The names are generally constructed of gateway namespace, gateway name, etc. The names can be long and potentially exceed the allowed length (63 chars) of a Kube name. Instead, and similar to Kube, the name should be generated from a hash.

xref: https://github.com/envoyproxy/gateway/pull/404

danehans:
Added `bug` label due to #434.

danehans:
When using hashed names for managed resources, users will need to use ` --selector` to correlate managed resources to a Gateway. For example:
```

$ kubectl get deploy -n envoy-gateway-system --selector=gateway.envoyproxy.io/owning-gateway-name=eg,gateway.envoyproxy.io/owning-gateway-namespace=default
NAME               READY   UP-TO-DATE   AVAILABLE   AGE
envoy-default-eg   1/1     1            1           2m14s
```

It might make debugability a bit easier if we had one label that was `namespace/name`. @arkodg @skriss @LukeShu @youngnick thoughts?

danehans:
A similar approach will be needed for the ports of the managed Envoy serivce.

arkodg:
I prefer if we used a hybrid approach in naming - `envoy-<40-50 char namespaced name>-<short-hash>` to improve debug speed.

AliceProxy:
I'll take this one on. Should have some bandwidth in the coming week.

PR #531, opened 2022-10-10
fix: incorrect eg svc name of `kube-demo`

Xunzhuo:
EG Service New Pattern is `EnvoyServicePrefix-GatewayNamespace-GatewayName`

### DIFF
--- a/tools/make/kube.mk
+++ b/tools/make/kube.mk
@@ -61,10 +61,10 @@ kube-undeploy: kube-uninstall ## Uninstall the Envoy Gateway controller into the
 
 .PHONY: kube-demo
 kube-demo: ## Deploy a demo backend service, gatewayclass, gateway and httproute resource and test the configuration.
-	kubectl apply -f examples/kubernetes/httpbin.yaml
-	kubectl apply -f examples/kubernetes/gatewayclass.yaml
-	kubectl apply -f examples/kubernetes/gateway.yaml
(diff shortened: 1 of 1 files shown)


tools/make/kube.mk
```diff
@@ -66,7 +66,7 @@ kube-demo: ## Deploy a demo backend service, gatewayclass, gateway and httproute
 	kubectl apply -f examples/kubernetes/gateway.yaml
 	kubectl apply -f examples/kubernetes/httproute.yaml
 	@echo "\nPort forward to the Envoy service using the command below"
-	@echo "kubectl -n envoy-gateway-system port-forward service/envoy-default-eg 8888:8080 &"
```
arkodg, line 69: the existing name is correct, the svc is running in the `envoy-gateway-system` system  (specified with `-n envoy-gateway-system`) with a name `envoy-<gw NS>-<gw Name>` , not the service name will soon change with https://github.com/envoyproxy/gateway/issues/441

Xunzhuo: The gw NS is envoy-gateway-system, so the svc name should not be envoy-envoy-gateway-system-eg?

arkodg: the gw NS here is the namespace where the gateway resource (`gateway.networking.k8s.io/v1beta1`) resides in / created in, not where the envoy gateway deployment/svc is running in

Issue #430, opened 2022-09-27
HTTPRouteHeaderMatching Conformance Test Fails


danehans:
The [HTTPRouteHeaderMatching](https://pkg.go.dev/sigs.k8s.io/gateway-api/conformance/tests#HTTPRouteHeaderMatching) conformance test fails.

Test logs:
```
http.go:159: Response expectation failed, not ready yet: expected pod name to start with infra-backend-v1, got infra-backend-v2-5cbc7f98db-hwzhm
Sending Request:
< GET / HTTP/1.1
< Host: 127.0.0.1:80
< User-Agent: Go-http-client/1.1
< Color: orange
< Version: two
< Accept-Encoding: gzip
<
<

Received Response:
< HTTP/1.1 200 OK
< Content-Length: 523
< Content-Type: application/json
< Date: Tue, 27 Sep 2022 16:55:11 GMT
< Server: envoy
< X-Content-Type-Options: nosniff
< X-Envoy-Upstream-Service-Time: 0
<
< {
<  "path": "/",
<  "host": "127.0.0.1:80",
<  "method": "GET",
<  "proto": "HTTP/1.1",
<  "headers": {
<   "Accept-Encoding": [
<    "gzip"
<   ],
<   "Color": [
<    "orange"
<   ],
<   "User-Agent": [
<    "Go-http-client/1.1"
<   ],
<   "Version": [
<    "two"
<   ],
<   "X-Envoy-Expected-Rq-Timeout-Ms": [
<    "15000"
<   ],
<   "X-Forwarded-Proto": [
<    "http"
<   ],
<   "X-Request-Id": [
<    "237162ca-ac38-4011-98aa-0c3e9c49225c"
<   ]
<  },
<  "namespace": "gateway-conformance-infra",
<  "ingress": "",
<  "service": "",
<  "pod": "infra-backend-v2-5cbc7f98db-hwzhm"
< }

    http.go:159: Response expectation failed, not ready yet: expected pod name to start with infra-backend-v1, got infra-backend-v2-5cbc7f98db-hwzhm
    http.go:140: timeout while waiting after 30 attempts, 0/3 sucessess
...
        --- FAIL: TestGatewayAPIConformance/HTTPRouteHeaderMatching/2_request_to_/_with_headers_should_go_to_infra-backend-v1 (30.00s)
```

Reproducer:
- Run `make kube-deploy`.
[...]

PR #532, opened 2022-10-10
fix(conformance): add HTTPRouteHeaderMatching case

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/430

Resolved by: https://github.com/envoyproxy/gateway/pull/508


arkodg (approved):
🚀

Issue #454, opened 2022-09-29
Run Conformance Tests with Identical Listener ports across Gateways


arkodg:
*Description*:

* Currently the Conformance Tests run with Gateway Listeners with unique ports due to the config specified here https://github.com/envoyproxy/gateway/blob/b10f2623e5d67af2d96d6f127649dfb7d6bcac55/test/conformance/conformance_test.go#L32
* This helps developers (like me) run `make run-conformance` on K8s / Docker Desktop on my Mac which has a limitation of supplying a single LoadBalancer Address `localhost`, making it not possible to create unique addresses when creating identical listeners.
* We would get more coverage if we ran Conformance Tests with identical ports.

arkodg:
thinking out load, we could run the CI command as `CONFORMANCE_UNIQUE_PORTS=false make-conformance` and use the env var value in https://github.com/envoyproxy/gateway/blob/b10f2623e5d67af2d96d6f127649dfb7d6bcac55/test/conformance/conformance_test.go#L32 to make a decision

Xunzhuo:
Anyone working on this?

PR #534, opened 2022-10-10
chore: run conformance tests with identical ports

Xunzhuo:
Resolve: #454

Run conformance tests with identical ports across gateways.

```
CONFORMANCE_UNIQUE_PORTS=false make run-conformance
```


test/conformance/conformance_test.go
```diff
@@ -24,18 +26,25 @@ func TestGatewayAPIConformance(t *testing.T) {
 
 	require.NoError(t, v1alpha2.AddToScheme(client.Scheme()))
 
+	setUniquePorts := env.Lookup("CONFORMANCE_UNIQUE_PORTS", "true")
```
arkodg, line 29: would `flag.Parse()` be a cleaner solution here ?

```
useUniquePorts = flag.Bool("use-unique-ports", true, "Whether to cleanup base test resources after the run")
func TestMain(m *testing.M) {
    flag.Parse()
    os.Exit(m.Run())
}

TestGatewayAPIConformance(t *testing.T) {
 if *useUniquePorts {
  ....
 } else {
 ....
 }
....
}
```

it would need us to create another make target similar to https://github.com/envoyproxy/gateway/blob/fcdb9b168c823d2fd47dd94c3edfc4c5db984f42/tools/make/kube.mk#L96 that gets called in CI
on the other hand the env approach would only require a change here https://github.com/envoyproxy/gateway/blob/fcdb9b168c823d2fd47dd94c3edfc4c5db984f42/.github/workflows/build_and_test.yaml#L52 but we lose out on the discoverability of the flag

Xunzhuo: Thanks for this comment, will update this later:)

arkodg (approved):
🚀

arkodg (approved):

Issue #533, opened 2022-10-10
Translate httproute with N listeners Order failed


Xunzhuo:
*Description*:

Occasionally occurs: https://github.com/envoyproxy/gateway/actions/runs/3213492844/jobs/5253203220

When we translate Gateway CR Listener to ListenerContext, types change from array to map, which makes order may not follow the Gateway CR Listeners. The port name may not follow the origin order, which cause the translation failed.

PR #535, opened 2022-10-10
fix: set correct listenerContext order

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/533

### DIFF
--- a/internal/gatewayapi/contexts.go
+++ b/internal/gatewayapi/contexts.go
@@ -17,16 +17,18 @@ import (
 type GatewayContext struct {
 	*v1beta1.Gateway
 
-	listeners map[v1beta1.SectionName]*ListenerContext
+	listeners []*ListenerContext
 }
 
 func (g *GatewayContext) GetListenerContext(listenerName v1beta1.SectionName) *ListenerContext {
 	if g.listeners == nil {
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/contexts.go
```diff
@@ -17,16 +17,18 @@ import (
...
 
 func (g *GatewayContext) GetListenerContext(listenerName v1beta1.SectionName) *ListenerContext {
 	if g.listeners == nil {
-		g.listeners = make(map[v1beta1.SectionName]*ListenerContext)
+		g.listeners = make([]*ListenerContext, 0)
 	}
 
-	if ctx := g.listeners[listenerName]; ctx != nil {
-		return ctx
+	for _, l := range g.listeners {
```
arkodg, line 28: if we take this approach of implicitly relying on input listener order, we'll no longer need to sort the listener Names in the o/p, can you also rm https://github.com/envoyproxy/gateway/blob/34afca0f0ad17e4caed3cda7a1ac2a1a1658ab11/internal/gatewayapi/sort.go#L48

arkodg (approved):
LGTM, thanks for fixing this issue ! this does make listener retrieval a little slower but it easier code to manage rather than add sort functions in the o/p

Issue #520, opened 2022-10-08
Multiple HTTPS Listeners Fails


danehans:
TLS termination fails for one listener when multiple HTTPS listeners are configured.

__Reproducer:__
Create the TLS assets for both listeners.
```
# Create root cert and key.
openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -subj '/O=example Inc./CN=example.com' -keyout example.com.key -out example.com.crt

# Create cert/key for listener 1
openssl req -out foo.example.com.csr -newkey rsa:2048 -nodes -keyout foo.example.com.key -subj "/CN=foo.example.com/O=httpbin organization"
openssl x509 -req -days 365 -CA example.com.crt -CAkey example.com.key -set_serial 0 -in foo.example.com.csr -out foo.example.com.crt


# Create cert/key for listener 2
openssl req -out bar.example.com.csr -newkey rsa:2048 -nodes -keyout bar.example.com.key -subj "/CN=bar.example.com/O=httpbin organization"
openssl x509 -req -days 365 -CA example.com.crt -CAkey example.com.key -set_serial 0 -in bar.example.com.csr -out bar.example.com.crt
```

Store the certs/keys in a Secret:
```
kubectl create secret tls foo-cert --key=foo.example.com.key --cert=foo.example.com.crt
kubectl create secret tls bar-cert --key=bar.example.com.key --cert=bar.example.com.crt
```

Create the example backend app:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: httpbin
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
    service: httpbin
spec:
  ports:
    - name: http
      port: 80
      targetPort: 80
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
    spec:
      serviceAccountName: httpbin
      containers:
        - image: docker.io/kennethreitz/httpbin
          imagePullPolicy: IfNotPresent
          name: httpbin
          ports:
            - containerPort: 80
```

Create the GatewayClass:
```
cat <<EOF | kubectl apply -f -
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: GatewayClass
metadata:
  name: eg
spec:
[...]

Issue #493, opened 2022-10-04
test EG against multiple Kubernetes versions


skriss:
We should test EG against multiple Kubernetes versions -- IMO at least the latest 3 minor versions.

skriss:
Assigning myself, will pick this up soon.

PR #542, opened 2022-10-11
Adds Secure Gateway User Doc

danehans:
Adds a user doc for configuring Gateway TLS termination.

### DIFF
--- a/docs/user/SECURE_GATEWAY.md
+++ b/docs/user/SECURE_GATEWAY.md
@@ -1,0 +1,136 @@
+## Introduction
+This guide will help you get started with using secure Gateways. This document uses a self-signed CA, so it should be
+used for testing and demonstration purposes only.
+
+## Prerequisites
+- A Kubernetes cluster with `kubectl` context configured for the cluster.
+- OpenSSL to generate TLS assets.
+
+__Note:__ Envoy Gateway is tested against Kubernetes v1.24.0.
+
+## Installation
+Install the Gateway API CRDs:
+```shell
+kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/v0.2.0/gatewayapi-crds.yaml
+```
+
+Run Envoy Gateway:
+```shell
+kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/v0.2.0/install.yaml
+```
+
+Run the example app:
+```shell
+kubectl apply -f https://raw.githubusercontent.com/envoyproxy/gateway/v0.2.0/examples/kubernetes/httpbin.yaml
+```
+
+### TLS Assets
+
+Generate TLS assets used by the Gateway.
+
+For macOS users, verify curl is compiled with the LibreSSL library:
+```shell
+curl --version | grep LibreSSL
+curl 7.54.0 (x86_64-apple-darwin17.0) libcurl/7.54.0 LibreSSL/2.0.20 zlib/1.2.11 nghttp2/1.24.0
+```
+
+Create a root certificate and private key to sign certificates:
+```shell
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -subj '/O=example Inc./CN=example.com' -keyout example.com.key -out example.com.crt
+```
+
+Create a certificate and a private key for `www.example.com`:
+```shell
+openssl req -out www.example.com.csr -newkey rsa:2048 -nodes -keyout www.example.com.key -subj "/CN=www.example.com/O=httpbin organization"
+openssl x509 -req -days 365 -CA example.com.crt -CAkey example.com.key -set_serial 0 -in www.example.com.csr -out www.example.com.crt
+```
+
+Store the cert/keys in A Secret:
+```shell
+kubectl create secret tls example-cert --key=www.example.com.key --cert=www.example.com.crt
+```
+
+The Gateway API resources must be created in the following order. First, create the GatewayClass:
+```shell
+kubectl apply -f https://raw.githubusercontent.com/envoyproxy/gateway/v0.2.0/examples/kubernetes/gatewayclass.yaml
+```
+
+Create the Gateway:
+```shell
+kubectl apply -f https://raw.githubusercontent.com//envoyproxy/gateway/v0.2.0/examples/kubernetes/gateway.yaml
+```
+
+Create the HTTPRoute to route traffic through Envoy proxy to the example app:
+```shell
+kubectl apply -f https://raw.githubusercontent.com/envoyproxy/gateway/v0.2.0/examples/kubernetes/httproute.yaml
+```
+
+### Testing the configuration
+Port forward to the Envoy service:
+```shell
+kubectl -n envoy-gateway-system port-forward service/envoy-default-eg 8043:8443 &
+```
+
+Curl the example app through Envoy proxy:
+```shell
+curl -v -HHost:www.example.com --resolve "www.example.com:8043:127.0.0.1" \
+--cacert example.com.crt https://www.example.com:8043/get
+```
+You can replace `get` with any of the supported [httpbin methods][httpbin_methods].
+
+### For clusters with External Loadbalancer support
+You can also test the same functionality by sending traffic to the External IP of the Gateway:
+```shell
+export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
+```
+
+Curl the example app through the Gateway, e.g. Envoy proxy:
+```shell
+curl -v -HHost:www.example.com --resolve "www.example.com:8443:${GATEWAY_HOST}" \
+--cacert example.com.crt https://www.example.com:8443/get
+```
+You can replace `get` with any of the supported [httpbin methods][httpbin_methods].
+
+## Clean-Up
+Use the steps in this section to uninstall everything from the quickstart guide.
+
+Delete the HTTPRoute:
+```shell
+kubectl delete httproute/httpbin
+```
+
+Delete the Gateway:
+```shell
+kubectl delete gateway/eg
+```
+
+Delete the Secret:
+```shell
+kubectl delete secret/example-cert
+```
+
+Delete the GatewayClass:
+```shell
+kubectl delete gc/eg
+```
+
+Uninstall the example app:
+```shell
+kubectl delete -f https://raw.githubusercontent.com/envoyproxy/gateway/v0.2.0-rc2/examples/kubernetes/httpbin.yaml
+```
+
+Uninstall Envoy Gateway:
+```shell
+kubectl delete -f https://github.com/envoyproxy/gateway/releases/download/v0.2.0-rc2/install.yaml
+```
+
+Uninstall Gateway API CRDs:
+```shell
+kubectl delete -f https://github.com/envoyproxy/gateway/releases/download/v0.2.0-rc2/gatewayapi-crds.yaml
+```
+
+## Next Steps
+Checkout the [Developer Guide](../../DEVELOPER.md) to get involved in the project.
+
+[kind]: https://kind.sigs.k8s.io/
+[httpbin_methods]: https://httpbin.org/#/HTTP_Methods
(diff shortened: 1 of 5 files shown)


chauhanshubham (dismissed):
lgtm

Xunzhuo (dismissed):

danehans:
Removing "Draft" since https://github.com/envoyproxy/gateway/pull/571 adds .md support.

danehans:
CI failure is due to Codecov upload, rerunning.

danehans:
Commit 813d589 is a rebase since #571 added markdown linting.

docs/user/SECURE_GATEWAY.md
```diff
@@ -0,0 +1,190 @@
...
+kubectl create secret tls example-cert --key=www.example.com.key --cert=www.example.com.crt
+```
+
+## Testing
+
+### Clusters without External Loadbalancer Support
+
+Port forward to the Envoy service:
+```shell
+kubectl -n envoy-gateway-system port-forward service/envoy-default-eg 8043:8443 &
```
skriss: Now that the name hashing is in, I'm seeing `envoy-default-eg-64656661` for the service name

examples/kubernetes/quickstart.yaml
```diff
@@ -15,6 +15,15 @@ spec:
...
       port: 8080
+    - name: https
+      protocol: HTTPS
+      port: 8443
+      tls:
+        mode: Terminate
+        certificateRefs:
+          - kind: Secret
+            group: ""
+            name: example-cert
```
skriss, line 26: Just noting here that users who are just using the regular quickstart will see this Listener marked as invalid in their Gateway conditions since we're not creating a TLS cert in the quickstart so the ref is invalid -- HTTP routing still works fine, just may be confusing. You could use a separate example YAML for the secure gateway guide, or add a kubectl command to add this listener in the secure gateway guide, or leaving as-is is probably fine too for now.

skriss (changes requested):

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+Store the previously created cert/key in Secret "example-cert-2":
+
+```shell
+kubectl create secret tls example-cert-2 --key=www.example.com.key --cert=www.example.com.crt
+```
+
+Follow the steps in the [Testing section](#testing) to test connectivity to the backend app through both Gateway
+listeners.
+
+## Cross Namespace Certificate References
```
skriss, line 129: Couple notes here, I'll see if there are existing issues for these in a sec:
- it seems like if you create the ReferenceGrant as the last step, it doesn't take effect until an EG restart or something else kicks the Gateway to re-reconcile
- Once everything's working, if you then delete the ReferenceGrant, the Listener continues to be "Ready" and still serves traffic with the TLS cert that it shouldn't be allowed to use.

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+  - group: ""
+    kind: Secret
+EOF
+```
+
+Delete the previously created Secret:
+
+```shell
+kubectl delete secret/example-cert
+```
```
skriss, line 159: It appears that right after this step, the Listener is still "Ready" and still serves traffic using the now-deleted TLS cert -- possibly fixed by https://github.com/envoyproxy/gateway/pull/579?

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+
+You can replace `get` with any of the supported [httpbin methods][httpbin_methods].
+
+### Clusters with External Loadbalancer Support
+
+Get the External IP of the Gateway:
+```shell
+export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
+```
```
LukeShu, line 72: The linter is upset that there's no blank line after "Get the External IP of the Gateway:"

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+
+This guide will help you get started using secure Gateways. The guide uses a self-signed CA, so it should be used for
+testing and demonstration purposes only.
+
+## Prerequisites
+
+- A Kubernetes cluster with `kubectl` context configured for the cluster.
+- OpenSSL to generate TLS assets.
+
+__Note:__ Envoy Gateway is tested against Kubernetes v1.24.
```
LukeShu, line 11: Getting a little off-topic, but: Should we figure out a matrix of Kube versions that we want to test against?

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+## Installation
+
+Follow the steps from the [Quickstart Guide](QUICKSTART.md) to install Envoy Gateway and the example manifest.
+Before proceeding, you should be able to curl the example backend using HTTP.
+
+## TLS Certificates
+
+Generate the certificates and keys used by the Gateway to terminate client TLS connections.
+
+For macOS users, verify curl is compiled with the LibreSSL library:
```
LukeShu, line 22: Why does that matter?

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+
+```shell
+curl --version | grep LibreSSL
+curl 7.54.0 (x86_64-apple-darwin17.0) libcurl/7.54.0 LibreSSL/2.0.20 zlib/1.2.11 nghttp2/1.24.0
+```
```
LukeShu, line 27: For snippets that include both the command and its output, we should mark these as `console` instead of `shell`, and give a leading `$ ` on the command lines; like

    ```console
    $ curl --version | grep LibreSSL
    curl 7.54.0 (x86_64-apple-darwin17.0) libcurl/7.54.0 LibreSSL/2.0.20 zlib/1.2.11 nghttp2/1.24.0
    ```

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+
+Store the cert/key in a Secret:
+
+```shell
+kubectl create secret tls example-cert --key=www.example.com.key --cert=www.example.com.crt
+```
+
+## Testing
+
+### Clusters without External Loadbalancer Support
```
LukeShu, line 50: The "B" in "Loadbalancer" should probably be capitalized.

precedent: `Service.spec.type=LoadBalancer`

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,214 @@
...
+
+Port forward to the Envoy service:
+
+```shell
+kubectl -n envoy-gateway-system port-forward service/envoy-default-eg 8043:8443 &
+```
+
+Curl the example app through Envoy proxy:
```
LukeShu, line 58: I have weird feelings about using "curl" as a verb. It feels slang-y/informal. Perhaps "Query the example app..."?

docs/user/secure-gateways.md
danehans: Slated for v0.3.0, [xref](https://github.com/envoyproxy/gateway/issues/493).

docs/user/secure-gateways.md
danehans: I referenced this from https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#before-you-begin. I'll remove.

docs/user/secure-gateways.md
danehans: ack

docs/user/secure-gateways.md
danehans: Thanks for finding this bug. It appears that ^ PR does not resolve this issue. I'm working through a fix that I will send a separate PR to fix.

docs/user/secure-gateways.md
danehans: Sam feedback as https://github.com/envoyproxy/gateway/pull/542/files#r997713038.

examples/kubernetes/quickstart.yaml
danehans: I removed this listener from the quickstart manifest and update the example gateway in this doc.

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,237 @@
...
+
+```shell
+kubectl create secret tls example-cert --key=www.example.com.key --cert=www.example.com.crt
+```
+
+Update the Gateway from the Quickstart guide to include an HTTPS listener that listens on port `8443` and references the
+`example-cert` Secret:
+
+```console
+$ cat <<EOF | kubectl apply -f -
```
chauhanshubham, line 45: a `patch` to the listener on an exisiting quickstart Gateway might be better.

chauhanshubham: did this for the tls-passthrough docs PR https://github.com/envoyproxy/gateway/pull/575/files#diff-c9d477931ae33d7a5bcbf71cf85dd9ca1d8fbb09e13367765df70fa00c5dc182R58

```console
$ kubectl patch gateway eg --type=json --patch '[{
   "op": "add",
   "path": "/spec/listeners/-",
   "value": {
      "name": "tls",
      "protocol": "TLS",
      "hostname": "passthrough.example.com",
      "tls": {"mode": "Passthrough"}, 
      "port": 6443,
    },
}]'
```

docs/user/SECURE_GATEWAY.md
danehans: Fixed.

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,233 @@
...
+Query the example app through the Gateway:
+
+```shell
+curl -v -HHost:www.example.com --resolve "www.example.com:8443:${GATEWAY_HOST}" \
+--cacert example.com.crt https://www.example.com:8443/get
+```
+
+## Multiple HTTPS Listeners
+
+Due to [Issue 520][], multiple HTTP listeners must use different port numbers. For example:
```
skriss, line 104: Was this fixed by https://github.com/envoyproxy/gateway/pull/574?

danehans: Good catch, let me verify.

danehans: @skriss @arkodg #574 did not fix this issue. See https://github.com/envoyproxy/gateway/issues/520#issuecomment-1283089969 for details.

danehans: @skriss this is now fixed and doc updated.

docs/user/secure-gateways.md
danehans: This is now fixed and doc updated.

docs/user/secure-gateways.md
danehans: This is now fixed and doc updated.

danehans:
Moving to v0.3.0-rc.1 since docs can be added post-v0.2.0.

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,260 @@
...
+
+```shell
+kubectl -n envoy-gateway-system port-forward service/${ENVOY_SERVICE} 8043:8443 &
+```
+
+Query the example app through Envoy proxy:
+
+```shell
+curl -v -HHost:www.example.com --resolve "www.example.com:8043:127.0.0.1" \
+--cacert example.com.crt https://www.example.com:8043/get
```
skriss, line 90: No need to explicitly provide the `Host` header since you're using `--resolve` so the URL being curl'ed already has the host:
```suggestion
curl -v --resolve "www.example.com:8043:127.0.0.1" \
--cacert example.com.crt https://www.example.com:8043/get
```

docs/user/secure-gateways.md
```diff
@@ -0,0 +1,260 @@
...
+
+```shell
+export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
+```
+
+Query the example app through the Gateway:
+
+```shell
+curl -v -HHost:www.example.com --resolve "www.example.com:8443:${GATEWAY_HOST}" \
+--cacert example.com.crt https://www.example.com:8443/get
```
skriss, line 105: Ditto: 
```suggestion
curl -v --resolve "www.example.com:8443:${GATEWAY_HOST}" \
--cacert example.com.crt https://www.example.com:8443/get
```

skriss (approved):
I was able to run through this all successfully.

Two little nits that you can choose to address or ignore.

Issue #540, opened 2022-10-11
Update Example Manifests API Group


danehans:
The [example manifests](https://github.com/envoyproxy/gateway/tree/main/examples) should use the `v1beta1` instead of `v1alpha2` API group.

PR #544, opened 2022-10-11
Update example manifest apiversion

AliceProxy:
Signed-off-by: AliceProxy <alicewasko@datawire.io>

Updates the `apiVersion` of the gateway API resources in `quickstart.yaml`. Resolves https://github.com/envoyproxy/gateway/issues/540


arkodg (approved):
🚀

PR #546, opened 2022-10-11
Moves Changelogs to Release Notes

danehans:
- Moves changelogs to release-notes since GitHub is auto-generating the changelog per release.
- Adds `generate-artifacts` target to add release notes to the release GA workflow.


arkodg:
might make sense to rename to `somewhere/<version>/release-notes` to so uploading the file is easier here https://github.com/envoyproxy/gateway/blob/5d5c28eb0911968b82f355376767e97fd89dcdf7/.github/workflows/release.yaml#L38

danehans:
@arkodg commit d030bb8 resolves your feedback above, PTAL.

cc: @Xunzhuo

Xunzhuo (approved):
LGTM

arkodg (approved):

youngnick:
Does this mean that the Github release will have the details from the YAML file on disk? That's probably fine for now, but I think that the Github Release page should have a proper explanation of the changes as far as we can at some point in the future.

danehans:
> Does this mean that the Github release will have the details from the YAML file on disk? 

Yes

>That's probably fine for now, but I think that the Github Release page should have a proper explanation of the changes as far as we can at some point in the future.

Maybe we add a step in our release guide that adds the explanations from the release-nodes.yaml into the release markdown page?

PR #547, opened 2022-10-12
use RefNotPermitted reason for invalid cross-namespace TLS cert ref

### LINKED ISSUE #538: Gateway Listener Conditions are Not Updated Properly for an Invalid or Missing ReferenceGrant
https://github.com/envoyproxy/gateway/pull/519 adds support for TLS termination. As part of this PR, initial ReferenceGrant support was added to allow Gateways to reference Secrets in a different namespace. However, listener status is not being updated properly for the following conditions:

1. A Gateway listener should fail to become ready if the Gateway has a `certificateRef` for a Secret in a different namespace and a ReferenceGrant exists but does not grant permission to that specific Secret.
2. A Gateway listener should fail to become ready if the Gateway has a `certificateRef` for a Secret in a different namespace and a ReferenceGrant granting permission to the Secret does not exist.

skriss:
Closes #538.

Leaving as draft for now since #441 is also a blocker for this issue.

### DIFF
--- a/internal/gatewayapi/testdata/gateway-with-listener-with-invalid-tls-configuration-secret-in-other-namespace.out.yaml
+++ b/internal/gatewayapi/testdata/gateway-with-listener-with-invalid-tls-configuration-secret-in-other-namespace.out.yaml
@@ -28,7 +28,7 @@ gateways:
           conditions:
             - type: ResolvedRefs
               status: "False"
-              reason: InvalidCertificateRef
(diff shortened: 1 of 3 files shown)


danehans:
@skriss #441 has merged and is no longer a blocker.

skriss:
I'll reopen a replacement for this shortly, accidentally created the feature branch in the upstream repo.

Issue #536, opened 2022-10-10
Fix logic for deleting Services from ProviderResources


chauhanshubham:
*Description*
The current logic attempts to Delete Services from the ProviderResources map when no Routes are found on the cluster, for a given namespace.

https://github.com/envoyproxy/gateway/blob/81eb8178550e7dfe111caf329390e5554c61d0c8/internal/provider/kubernetes/httproute.go#L284

_Snippet_

```
// Delete the Namespace and Service from the resource maps if no other
// routes exist in the namespace.
routeList = &gwapiv1b1.HTTPRouteList{}
if err := r.client.List(ctx, routeList, &client.ListOptions{Namespace: request.Namespace}); err != nil {
	return reconcile.Result{}, fmt.Errorf("error listing httproutes")
}
if len(routeList.Items) == 0 {
	r.resources.Namespaces.Delete(request.Namespace)
	log.Info("deleted namespace from resource map")
	r.resources.Services.Delete(request.NamespacedName)
	log.Info("deleted service from resource map")
}
```

This tries deleting a Service with namespace/name that matches the HTTPRoute which came in to be Reconciled. This will not be the case most of the times.
We must delete the Services which were referred to by the HTTPRoute that has been now deleted.

danehans:
@chauhanshubham thanks for catching this issue. The httproute/tlsroute controllers should get the HTTPRoute from the resource map before deleting it and then iterate through the Service backends, deleting each from the map.

danehans:
xref #539

chauhanshubham:
Does having a separate cache for these mappings make sense? Something like
```
type ProviderCache struct {
   lock
   routeToServiceMappings map[string][]string
   serviceToRouteMappings map[string]string
}
```
Or we'll probably have to iterate through the ProviderResource cache multiple times
once to find out which services were referred by the deleted route
and second time to find whether the services are referred by other routes currently present.
^ both the steps must be performed for every kind of route in the ProviderResources

Issue #539, opened 2022-10-10
A ReferenceGrant Should Be Removed From Watchable Map When UnReferenced


danehans:
#471 added initial support for ReferenceGrants by allowing Gateways to reference Secrets in a different namespace. The gateway controller adds the ReferenceGrant and the ReferenceGrant namespace to the provider watchable map when:

1. A Gateway listener references a Secret in a different namespace.
2. A ReferenceGrant in the above namespace exists and permits the Gateways `from` the Gateway's namespace `to` Secrets.

When the Gateway is deleted, the ReferenceGrant and ReferenceGrant namespace should be deleted from the provider's watchable map if no other Gateway references the ReferenceGrant.

skriss:
Note https://github.com/envoyproxy/gateway/pull/580#pullrequestreview-1144310617, when this is fixed the two invalid/missing referencegrant conformance tests should be able to be uncommented.

arkodg:
I prefer if we move this to 0.3.0 . Based on current logic there is no easy way to handle deletes for RefGrants, easier to implement this with https://github.com/envoyproxy/gateway/issues/413 where we dont need to handle deletes, just start from scratch with every reconcile.

danehans:
Being noted in the v0.2.0 release notes and moving to v0.3.0-rc.1.

chauhanshubham:
https://github.com/envoyproxy/gateway/pull/702 fixes this issue and uncomment's the two conformance tests.

PR #549, opened 2022-10-12
Fix Service object remove from resource map from route controller

chauhanshubham:
This commit fetches the services that were referenced by the route which has now been deleted, and removes those objects from the resource map. In order to facilitate this, the commit introduces a provider cache for kubernetes that stores mappings between kubernetes objects - for now it stores the map between [tls/http]route -> backend service references.
So that once the [tls/http]route is deleted we have the services that it referenced.
Fixes: #536


arkodg:
thanks for working on this ! 
this functionality will most likely not be needed once https://github.com/envoyproxy/gateway/issues/413 is in since it will always build out SoTW from scratch, using the controller runtime cache,  but it might make sense to introduce this PR if we want to support RefereneGrant in 0.2.0 since https://github.com/envoyproxy/gateway/issues/539 (a much more severe consequence of not removing the resource) will leverage the same library introduced in this PR

chauhanshubham:
Will there not be a need to store Service objects any more in the resource map post 413?
Maybe we should try fetching the Services from the controller cache right now itself? doesn't seem to depend on 413. Maybe I'm missing something.
We'll have to pass on the client to the gatewayapi/translator for fetching these objects. 

I'm looking at https://github.com/envoyproxy/gateway/blob/0be1aef9248c54e16836d0f3a960df147bc09f11/internal/gatewayapi/translator.go#L66

The `GetNamespace`, `GetService`, and `GetSecret` functions here, that try fetching it from the Resources struct, instead we could read it from the controller cache perhaps? We can add a similar `GetReferenceGrant` too.

arkodg:
> Will there not be a need to store Service objects any more in the resource map post 413? Could we try fetching the Services from the controller cache right now itself? doesn't seem to depend on 413. Maybe I'm missing something. We'll have to pass on the client to the gatewayapi/translator for fetching these objects.
> 
> I'm looking at
> 
> https://github.com/envoyproxy/gateway/blob/0be1aef9248c54e16836d0f3a960df147bc09f11/internal/gatewayapi/translator.go#L66
> 
> The `GetNamespace`, `GetService`, and `GetSecret` functions here, that try fetching it from the Resources struct, instead we could read it from the controller cache perhaps? We can add a similar `GetReferenceGrant` too.

I was referring to the fact that we won't need to handle `delete` logic anymore with #413 .
* Any changes to gatewayclasses, gateways, httproutes, services .... will trigger a reconcile
* the function would build all the `resource` state from scratch starting from the accepted GatewayClass
```
type resource struct {
GatewayClass string
Gateways []*v1b1.Gateway
HTTPRoutes []*v1b1.HTTPRoute
....
```
and update the watchable map with the same key, so we dont need to deal with deletes in the provider anymore
* if we want to add watches for specific services, namespaces etc. we might need to handle delete logic there

arkodg (dismissed):
🚀

internal/provider/kubernetes/cache.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// providerCache maintains additional mappings related to Kubernetes provider
+// resources. The mappings are regularly updated from the reconcilers based
+// on the existence of the object in the Kubernetes datastore.
+type providerCache struct {
```
LukeShu, line 12: Calling this a "cache" is misleading. It isn't really a cache, it's a reference-counting machine.

internal/provider/kubernetes/cache.go
```diff
@@ -0,0 +1,66 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for _, svcs := range p.routeToServicesMappings {
+		if svcs.Has(service) {
+			return true
+		}
+	}
+	return false
+}
```
LukeShu, line 66: This is O(n), and that makes me uneasy. Would it make sense to have the primary key be the service, rather than the route?

LukeShu (approved):
There are some things I think should be changed, but if we want to get this in to 0.2.0, I'm OK merging as-is and cleaning it up later.

internal/provider/kubernetes/cache.go
chauhanshubham: It'll get uglier if we use the reverse mapping with the service as P.K.
As we need to query this primarily - `Find all services for a Route r1 which was deleted`, while reconciling the deleted Route. With service as the P.K. we'll have to traverse all the Services (and find a route match) which will probably be more than the number of TLS/HTTP Routes. (Note that with route as the P.K we are filtering based on kind too)

I started off by adding multiple mappings like `route-to-services` and `service-to-routes`, but seemed like we're optimizing at the cost of double the memory, so stuck with the `route-to-services` mapping only.

arkodg:
hey @chauhanshubham can you rm the `docs/user/tls-passthrough.md` changes from this PR :)

LukeShu (approved):
Thanks!

Issue #550, opened 2022-10-12
Allow release-manifests Target to Specify Image


danehans:
Add `$(IMAGE)` to the `release-manifests` target. This will allow e2e testing of a release prior to cutting the actual release. For example, @Xunzhuo created [this](https://github.com/Xunzhuo/gateway/releases/tag/v0.2.0) test v0.2.0 release but the EG deployment manifest uses a non-existent image.

```
$ kubectl get deploy/envoy-gateway -n envoy-gateway-system -o yaml | grep "image: "
        image: envoyproxy/gateway:v0.2.0
```

danehans:
@arkodg thoughts?

Xunzhuo:
OK, ping me if this is needed, I have already implemented this locally.
<img width="819" alt="image" src="https://user-images.githubusercontent.com/48784001/195400782-f796635a-ed78-4ef2-bf78-211ed6609b01.png">

arkodg:
yah this sounds good, thanks for picking this up @Xunzhuo !

PR #553, opened 2022-10-12
chore: set IMAGE to release-manifests

Xunzhuo:
Fixes: #550

image
 


tools/make/kube.mk
```diff
@@ -101,4 +103,20 @@ delete-cluster: $(tools/kind) ## Delete kind cluster.
...
+	@curl -sLo release-artifacts/gatewayapi-crds.yaml ${GATEWAY_RELEASE_URL}
+	@echo "Added: release-artifacts/gatewayapi-crds.yaml"
+
+	@mkdir -pv $(OUTPUT_DIR)/manifests/provider
+	@cp -r $(KUBE_PROVIDER_DIR) $(OUTPUT_DIR)/manifests/provider
+	@mkdir -pv $(OUTPUT_DIR)/manifests/infra
+	@cp -r $(KUBE_INFRA_DIR) $(OUTPUT_DIR)/manifests/infra
+	@cd $(OUTPUT_DIR)/manifests/provider/config/envoy-gateway && $(ROOT_DIR)/$(tools/kustomize) edit set image envoyproxy/gateway-dev=$(IMAGE):$(TAG)
+	@$(tools/kustomize) build $(OUTPUT_DIR)/manifests/provider/config/default > release-artifacts/envoy-gateway.yaml
+	@$(tools/kustomize) build $(OUTPUT_DIR)/manifests/infra/config/rbac > release-artifacts/infra-manager-rbac.yaml
```
arkodg, line 116: should these intermediate YAML be deleted so they dont get uploaded ?

tools/make/kube.mk
```diff
@@ -101,4 +103,20 @@ delete-cluster: $(tools/kind) ## Delete kind cluster.
...
+	@$(tools/kustomize) build > release-artifacts/install.yaml
+	@echo "Generated: release-artifacts/install.yaml"
+
+	@cp examples/kubernetes/quickstart.yaml release-artifacts/quickstart.yaml
```
arkodg, line 120: lets rm `@` from all non echo commands to improve debuggability

tools/make/kube.mk
```diff
@@ -101,4 +103,20 @@ delete-cluster: $(tools/kind) ## Delete kind cluster.
 
 .PHONY: release-manifests
 release-manifests: $(tools/kustomize) ## Generate Kubernetes release manifests.
-	tools/hack/release-manifests.sh $(GATEWAY_API_VERSION) $(TAG)
+	@mkdir -p release-artifacts/
+	@curl -sLo release-artifacts/gatewayapi-crds.yaml ${GATEWAY_RELEASE_URL}
+	@echo "Added: release-artifacts/gatewayapi-crds.yaml"
+
+	@mkdir -pv $(OUTPUT_DIR)/manifests/provider
```
arkodg, line 110: can you try and reuse the previous work you did in https://github.com/envoyproxy/gateway/blob/2fb3ca8954705dabbcfa08007b658655fcb60f46/tools/make/kube.mk#L41

tools/make/kube.mk
Xunzhuo: Okay

tools/make/kube.mk
Xunzhuo: https://github.com/envoyproxy/gateway/blob/main/.github/workflows/release.yaml#L37 we specify the release file, so no need to worry about this.

tools/make/kube.mk
Xunzhuo: That will apply the outputs, but right, we can make this more common used.

.github/workflows/build_and_test.yaml
```diff
@@ -64,9 +64,9 @@ jobs:
     - name: Build and Push EG Commit Image
       if: github.event_name == 'push'
       # tag is set to the short SHA of the commit
-      run: make image.push.multiarch PLATFORMS="linux_amd64 linux_arm64"
+      run: make image.push.multiarch PLATFORMS="linux_amd64 linux_arm64" IMAGE=envoyproxy/gateway-dev
```
arkodg, line 67: is this change needed or just added it here to make it more explicit ?

Xunzhuo: Not needed, but to make it more explicit:), used for people who want to have a pre release

arkodg (approved):
🚀

PR #557, opened 2022-10-13
Fix intermittent failures

### LINKED ISSUE #504: Fix flaky xds Translator TestRunner
*Description*:
Fix and re enable flaky Xds [TestRunner](https://github.com/envoyproxy/gateway/blob/5bf94c11e16a5c6bda2938de70feb15d2b60e626/internal/xds/translator/runner/runner_test.go#L16) 

[optional *Relevant Links*:]
Intermittent failures seen in https://github.com/envoyproxy/gateway/actions/runs/3184962312/jobs/5193973518#step:4:110

LukeShu:
Fixes #504

The XDS translator's startup is racy; it needs to be changed to handle anything that's already in the `watchable.Map` when it calls `.Subscribe`. Since this turned out to be a common pattern I created a utility function, and moved everything over to use it.

Also, a different test is broken with any `-count` > 1, because `controller-runtime.SetupSignalHandler()` is designed to panic if it is called more than once.

I used `go test -v -count=300 ./...` to validate that these changes make the intermittent test failures to go away.

Also, I added `vendor/` to `.gitignore` so that I could run `go mod vendor` and then add log statements to 3rd-party code while debugging.

### DIFF
--- a/internal/xds/server/runner/runner.go
+++ b/internal/xds/server/runner/runner.go
@@ -10,12 +10,15 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/telepresenceio/watchable"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/message"
+	"github.com/envoyproxy/gateway/internal/utils/watchutil"
 	"github.com/envoyproxy/gateway/internal/xds/cache"
+	xdstypes "github.com/envoyproxy/gateway/internal/xds/types"
 	controlplane_service_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/service/cluster/v3"
 	controlplane_service_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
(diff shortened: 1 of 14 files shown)


Xunzhuo (commented):
Looks good, we can fix it with this way, but I wonder if we can make the first time creation, to be a update? The create/update/delete, I think they should be notified.

internal/xds/server/runner/runner.go
```diff
@@ -112,10 +115,8 @@ func registerServer(srv controlplane_server_v3.Server, g *grpc.Server) {
 
 func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 	// Subscribe to resources
-	for snapshot := range r.Xds.Subscribe(ctx) {
-		r.Logger.Info("received a notification")
-		// Load all resources required for translation
-		for _, update := range snapshot.Updates {
+	watchutil.HandleSubscription(r.Xds.Subscribe(ctx),
+		func(update watchable.Update[string, *xdstypes.ResourceVersionTable]) {
```
arkodg, line 119: any way to move the watchable defs into the callee lib ?

LukeShu: Unfortunately,
```
internal/message/watchutil.go:7:6: generic type cannot be alias
```
however, since `Update` has no methods it isn't so important that it be an alias, it can just be a new child type, so I've now done that.

arkodg (approved):
LGTM, thanks for catching the race and refactoring how snapshot updates are consumed !

Issue #501, opened 2022-10-04
docs: User-facing tutorial examples


kflynn:
We should probably follow https://gateway-api.sigs.k8s.io/guides/getting-started/ here:

- [x] [Simple Gateway](https://gateway-api.sigs.k8s.io/guides/simple-gateway/)
- [x] [HTTP Routing](https://gateway-api.sigs.k8s.io/guides/http-routing/)
- [x] [HTTP Redirects](https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/), but note that we think that rewrites aren't working yet
- [x] [HTTP Traffic Splitting](https://gateway-api.sigs.k8s.io/guides/traffic-splitting/)
- [ ] [Routing Across Namespaces](https://gateway-api.sigs.k8s.io/guides/multiple-ns/)
- [x] [TLS](https://gateway-api.sigs.k8s.io/guides/tls/) may not be supported
- [x] TLS passthrough

Note that [TCP Routing](https://gateway-api.sigs.k8s.io/guides/tls/) is not currently supported, and we should explicitly call that out.

danehans:
All but one of the tasks was completed in v0.2.0... yay!!! Moving to 0.3.0-rc.1 to resolve the last task.

PR #558, opened 2022-10-13
Adds HTTPRouting User Doc

danehans:
xref: #501


fredzach:
Cannot properly replicate arm issues on my m1 mac now, likely due to rosetta. Will run another round of tests on the pis after the weekend.

@danehans should this PR also update the quickstart with the k8s-staging-ingressconformance/echoserver, replacing httpbin?

danehans:
> @danehans should this PR also update the quickstart with the k8s-staging-ingressconformance/echoserver, replacing httpbin?

@fredzach I want to keep this PR focused on the HTTP routing user doc changes. Would you like to submit a PR for updating the Quickstart image?

danehans:
Removing "Draft" since https://github.com/envoyproxy/gateway/pull/571 adds .md support.

arkodg (approved):
🚀

docs/user/HTTP_ROUTING.md
```diff
@@ -0,0 +1,98 @@
...
+```
+The status should reflect "Ready=True", indicating the Envoy proxy infrastructure has been provisioned. The status also
+provides the address of the Gateway. This address is used later in the guide to test connectivity to proxied backend
+services.
+
+The three HTTPRoute resources create routing rules on the Gateway. In order to receive traffic from a Gateway, 
+an HTTPRoute must be configured with `parentRefs` which reference the parent Gateway(s) that it should be attached to.
+An HTTPRoute can match against a [single set of hostnames][spec]. These hostnames are matched before any other matching
+within the HTTPRoute takes place. Since `example.com`, `foo.example.com`, and `bar.example.com` are separate hosts with
+different routing requirements, each is deployed as its own HTTPRoute - `example-route, ``foo-route`, and `bar-route`.
```
youngnick, line 39:
```suggestion
different routing requirements, each is deployed as its own HTTPRoute - `example-route`, `foo-route`, and `bar-route`.
```

youngnick (approved):
LGTM with one minor formtting nit.

PR #559, opened 2022-10-13
Hash ir/infra resources

AliceProxy:
Changes some infra names so that everything is hashed now (service/deployment/configmap/serviceaccount). It preserves up to 48 characters of the original resource's name and then adds an 8 character hash (48 chosen because when appending the 8 character hash and the envoy prefix("envoy") it adds up to 63 characters) .

The labels for the owning gateway name and namespace have not been altered.

Should resolve https://github.com/envoyproxy/gateway/issues/441


internal/provider/utils/utils.go
```diff
@@ -12,3 +16,14 @@ func NamespacedName(obj client.Object) types.NamespacedName {
...
 	}
 }
+
+// Returns a partially hashed name for the string if it is more than 60 chars. Otherwise returns the original string
+func GetHashedName(name string) string {
+	if len(name) > 60 {
+		h := sha256.New() // Using sha256 instead of sha1 due to Blocklisted import crypto/sha1: weak cryptographic primitive (gosec)
+		hsha := h.Sum([]byte(name))
+		hashedName := strings.ToLower(fmt.Sprintf("%x", hsha))
+		return fmt.Sprintf("%s-%s", name[0:32], hashedName[0:16])
```
arkodg, line 26: 7/8 nibbles should be for the short hash imho https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1

internal/provider/utils/utils.go
```diff
@@ -12,3 +16,14 @@ func NamespacedName(obj client.Object) types.NamespacedName {
 		Name:      obj.GetName(),
 	}
 }
+
+// Returns a partially hashed name for the string if it is more than 60 chars. Otherwise returns the original string
+func GetHashedName(name string) string {
+	if len(name) > 60 {
```
arkodg, line 22: might make sense to make this hashing consistent for all strings and not just the ones above 60 char

internal/provider/kubernetes/gateway.go
```diff
@@ -651,9 +651,12 @@ func (r *gatewayReconciler) subscribeAndUpdateStatus(ctx context.Context) {
 }
 
 func infraServiceName(gateway *gwapiv1b1.Gateway) string {
-	return fmt.Sprintf("%s-%s-%s", config.EnvoyServicePrefix, gateway.Namespace, gateway.Name)
+
+	infraName := utils.GetHashedName(fmt.Sprintf("%s-%s", gateway.Namespace, gateway.Name))
+	return fmt.Sprintf("%s-%s", config.EnvoyDeploymentPrefix, infraName)
```
arkodg, line 656:
```suggestion
	return fmt.Sprintf("%s-%s", config.EnvoyServicePrefix, infraName)
```

internal/ir/infra.go
```diff
@@ -209,5 +210,5 @@ func (p *ProxyInfra) ObjectName() string {
 	if len(p.Name) == 0 {
```
arkodg, line 210: I dont think `ObjectName` is called by anyone, can we delete it ?
cc @danehans

internal/infrastructure/kubernetes/serviceaccount.go
```diff
@@ -9,16 +9,14 @@ import (
...
-)
-
-const (
-	envoyServiceAccountPrefix = "envoy"
+	"github.com/envoyproxy/gateway/internal/provider/utils"
 )
 
 func expectedServiceAccountName(proxyName string) string {
-	return fmt.Sprintf("%s-%s", envoyServiceAccountPrefix, proxyName)
+	return utils.GetHashedName(fmt.Sprintf("%s-%s", config.EnvoyConfigMapPrefix, proxyName))
```
arkodg, line 19:
```suggestion
	return utils.GetHashedName(fmt.Sprintf("%s-%s", config.EnvoyServicePrefix, proxyName))
```

arkodg, line 19: lets either rename all to `EnvoyProxyPrefix` / `EnvoyPrefix` or add another `EnvoyServiceAccountPrefix`

internal/ir/infra.go
AliceProxy: Yeah I started here but didn't see it called anywhere so I wasn't sure.

internal/infrastructure/kubernetes/serviceaccount.go
AliceProxy: Ah good point, good point

internal/provider/utils/utils.go
AliceProxy: Ok, everything is now hashed. It now takes up to 32 characters (arbitrary number I chose that seemed reasonable) and then adds the 8 character hash. I can increase/change how much of the original name we keep if you think we should increase it or decrease it further.

internal/gatewayapi/translator.go
```diff
@@ -1519,7 +1520,7 @@ func isValidHostname(hostname string) error {
 }
 
 func irStringKey(gateway *v1beta1.Gateway) string {
-	return fmt.Sprintf("%s-%s", gateway.Namespace, gateway.Name)
+	return utils.GetHashedName(fmt.Sprintf("%s-%s", gateway.Namespace, gateway.Name))
```
arkodg, line 1523: any specific reason to add this ?

internal/provider/utils/utils.go
arkodg: thanks, the more chars the better imho, I think we should be able to do 48 chars without breaking anything (48 + 8, 5 for envoy, 2 for the `-`)

api/config/v1alpha1/groupversion_info.go
```diff
@@ -1,7 +1,7 @@
 // Package v1alpha1 contains API Schema definitions for the config v1alpha1 API group.
 //
-//+kubebuilder:object:generate=true
-//+groupName=config.gateway.envoyproxy.io
+// +kubebuilder:object:generate=true
```
danehans, line 3: You must be running go 1.19 since gofmt now adds a space to doc comments. Since the project still uses 1.18, this change should be reverted.

internal/gatewayapi/translator.go
danehans: @arkodgis ^ b/c we should only be hashing the ns/name at the Infra Manager?

internal/provider/utils/utils.go
```diff
@@ -12,3 +16,16 @@ func NamespacedName(obj client.Object) types.NamespacedName {
 		Name:      obj.GetName(),
 	}
 }
+
+// Returns a partially hashed name for the string if it is more than 60 chars. Otherwise returns the original string
```
danehans, line 20: s/60/32/

internal/provider/utils/utils.go
```diff
@@ -12,3 +16,16 @@ func NamespacedName(obj client.Object) types.NamespacedName {
...
+func GetHashedName(name string) string {
+
+	h := sha256.New() // Using sha256 instead of sha1 due to Blocklisted import crypto/sha1: weak cryptographic primitive (gosec)
+	hsha := h.Sum([]byte(name))
+	hashedName := strings.ToLower(fmt.Sprintf("%x", hsha))
+
+	if len(name) > 32 {
+		return fmt.Sprintf("%s-%s", name[0:32], hashedName[0:8])
+	}
```
danehans, line 29: Would it be more efficient if the func first checks the provided name is less or equal to 32 chars and then return early with the provided name?

internal/status/status.go
```diff
@@ -141,10 +141,11 @@ func (u *UpdateWriter) Send(update Update) {
...
 // Supported objects:
-//  GatewayClasses
-//  Gateway
-//  HTTPRoute
-//  TLSRoute
+//
+//	GatewayClasses
+//	Gateway
+//	HTTPRoute
+//	TLSRoute
```
danehans, line 148: You must be running go 1.19 since gofmt now adds a space to doc comments. Since the project still uses 1.18, this change should be reverted.

danehans (changes requested):

internal/status/status.go
AliceProxy: Ah yep that would be it. Switched to go 1.19 for some other work and didn't notice.

internal/gatewayapi/translator.go
arkodg: yeah, my comment is not a blocker, but I see the need to hash as something thats being done to be conformant within k8s (specific platform), not affecting any other components

internal/provider/utils/utils.go
AliceProxy: Sounds good. 48 it is now which puts us right at the 63 character mark

internal/ir/infra.go
AliceProxy: Removed the change here since it wasn't needed

internal/gatewayapi/translator.go
AliceProxy: Removed this bit and now only the service/deployment/configmap/serviceaccount are hashed.

arkodg (approved):
LGTM, thanks !

internal/provider/utils/utils.go
AliceProxy: This bit has been updated to use up to 48 characters so that with the `"envoy-` and the 8 character hash we can go all the way up to 63 characters max. Hashing still occurs on all resources regardless of initial name length at @arkodg's  suggestion but I'm good with doing it either way.

danehans (approved):

Issue #562, opened 2022-10-13
HTTPRoute with URL Rewrite Produces Accepted=True HTTPRoute Status Condition


danehans:
## Reproducer
- Install EG and quickstart guide
- Create an HTTPRoute that rewrites the URL:
```
cat <<EOF | kubectl apply -f -
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: http-filter-rewrite
spec:
  parentRefs:
    - name: eg
  hostnames:
    - rewrite.example
  rules:
    - filters:
      - type: URLRewrite
        urlRewrite:
          hostname: www.example.com
      backendRefs:
      - name: httpbin
        port: 80
EOF
```

HTTPRoute status:
```
- apiVersion: gateway.networking.k8s.io/v1beta1
  kind: HTTPRoute
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"gateway.networking.k8s.io/v1beta1","kind":"HTTPRoute","metadata":{"annotations":{},"name":"http-filter-rewrite","namespace":"default"},"spec":{"hostnames":["rewrite.example"],"parentRefs":[{"name":"eg"}],"rules":[{"backendRefs":[{"name":"httpbin","port":80}],"filters":[{"type":"URLRewrite","urlRewrite":{"hostname":"www.example.com"}}]}]}}
    creationTimestamp: "2022-10-13T19:16:07Z"
    generation: 1
    name: http-filter-rewrite
    namespace: default
    resourceVersion: "49003"
    uid: 45a55677-f725-4e77-9d58-d69e9418b1c6
  spec:
    hostnames:
    - rewrite.example
    parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: eg
    rules:
    - backendRefs:
      - group: ""
        kind: Service
        name: httpbin
        port: 80
        weight: 1
      filters:
      - type: URLRewrite
        urlRewrite:
          hostname: www.example.com
      matches:
      - path:
          type: PathPrefix
          value: /
  status:
    parents:
    - conditions:
      - lastTransitionTime: "2022-10-13T19:16:07Z"
        message: Route is accepted
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
      parentRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
[...]

PR #563, opened 2022-10-13
Fixes Calculating ParentRef Status Conditions

danehans:
Previously, `ProcessHTTPRoutes()` of the gatewayapi translator would set "Accepted=True" as the default status condition. With #516 the status conditions get reset for every HTTPRoute reconciliation and "Accepted=True" should only be set if no status conditions for the parentRef exist.

Fixes: #562

### DIFF
--- a/internal/gatewayapi/testdata/httproute-with-header-filter-empty-headers.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-header-filter-empty-headers.out.yaml
@@ -66,9 +66,9 @@ httpRoutes:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
       conditions:
       - type: Accepted
-        status: "True"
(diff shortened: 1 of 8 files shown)


AliceProxy (dismissed):
Looks good to me. Thanks for the fix.

chauhanshubham:
@danehans I think this will be required for TLSRoute as well, could you please add a similar logic there as well. thanks.

danehans:
> I think this will be required for TLSRoute as well, could you please add a similar logic there too. thanks.

@chauhanshubham I added a condition length check to `ProcessTLSRoutes()` before setting the default "Accepted=True" condition. Otherwise, TLSRoutes have no filters so we're good.

chauhanshubham (approved):

youngnick (approved):

PR #565, opened 2022-10-14
Adds HTTPRoute Redirect Docs

danehans:
Adds a user guide for configuring HTTPRoute redirects.

xref: #501 

### DIFF
--- a/docs/user/HTTP_REDIRECT.md
+++ b/docs/user/HTTP_REDIRECT.md
@@ -1,0 +1,107 @@
+# HTTP Redirects
+
+The [HTTPRoute][] resource can issue redirects to clients or rewrite paths sent upstream using filters. Note that
+HTTPRoute rules cannot use both filter types at once. Currently, Envoy Gateway only supports core [HTTPRoute filters][]
+which consist of `RequestRedirect` and `RequestHeaderModifier`. To learn more about HTTP routing, refer to the
+[Gateway API documentation][].
+
+Follow the steps from the [Secure Gateways](SECURE_GATEWAY.md) to install Envoy Gateway and the example manifest. Do not
+proceed until you can curl the example backend from the Quickstart guide using HTTPS.
+
+## Redirects
+Redirects return HTTP 3XX responses to a client, instructing it to retrieve a different resource. A
+[`RequestRedirect` filter][req_filter] instructs Gateways to emit a redirect response to requests that match the rule.
+For example, to issue a permanent redirect (301) from HTTP to HTTPS, configure `requestRedirect.statusCode=301` and
+`requestRedirect.scheme="https"`:
+```shell
+cat <<EOF | kubectl apply -f -
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: http-to-https-filter-redirect
+spec:
+  parentRefs:
+    - name: eg
+  hostnames:
+    - redirect.example
+  rules:
+    - filters:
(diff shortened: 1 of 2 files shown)


arkodg (approved):
🚀

youngnick (approved):

danehans:
Rebasing since #571 added markdown linting.

docs/user/HTTP_REDIRECT.md
```diff
@@ -0,0 +1,107 @@
...
+kubectl get httproute/http-to-https-filter-redirect -o yaml
+```
+
+Curl'ing `redirect.example/get` should result in a `301` response from the example Gateway and redirecting to the
+configured redirect hostname.
+```shell
+$ curl -L -vvv --header "Host: redirect.example" "http://${GATEWAY_HOST}:8080/get"
+...
+< HTTP/1.1 301 Moved Permanently
+< location: https://www.example.com:8443/get
```
LukeShu, line 51: Blocks that contain the shell command and its output should be marked `console`, not `shell`. (There are a few such code-blocks in this PR.)

docs/user/HTTP_REDIRECT.md
```diff
@@ -0,0 +1,107 @@
...
+  parentRefs:
+    - name: eg
+  hostnames:
+    - redirect.example
+  rules:
+    - filters:
+      - type: RequestRedirect
+        requestRedirect:
+          scheme: https
+          statusCode: 301
```
AliceProxy, line 32: As a heads up, if you don't set `statusCode` at all it will default to 301 (because in the xds if you don't configure a specific status code then Envoy will use 301 by default. Gateway API only specifies support for `301` and `302` here so really you only need to include `statusCode: 302`, otherwise you can leave it out.

That much detail isn't necessary for the docs, but might be nice to call out that if you don't configure the `statusCode` then it has a default.

danehans: @AliceProxy thanks for the review and feedback. We're setting default values in other docs. For example, see the Gateway listener in the Quickstart Guide. I'll leave the example as-is and add a note below it.

docs/user/http-redirect.md
```diff
@@ -0,0 +1,117 @@
...
+
+```shell
+kubectl get httproute/http-to-https-filter-redirect -o yaml
+```
+
+Curl'ing `redirect.example/get` should result in a `301` response from the example Gateway and redirecting to the
+configured redirect hostname.
+
+```shell
+$ curl -L -vvv --header "Host: redirect.example" "http://${GATEWAY_HOST}:8080/get"
```
Xunzhuo, line 54: DO we need to let user know how to configure GATEWAY_HOST?

like 
```
export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
```

AliceProxy: I don't think its totally necessary, but I think its a good idea. I think one of the assumptions we make for all these specific docs pages is that the user has already gone through the quickstart, so they should know how to get the address, but eliminating friction anywhere we can sounds like a good idea to me.

danehans: This doc does state to complete the Quickstart but it doesn't hurt to add the export. I'll update and push a new commit.

danehans: Resolved feedback in 97e26df.

danehans:
Codecov upload failed. Rerunning CI.

LukeShu (approved):

Issue #566, opened 2022-10-14
Introduce canary release of EG.


Xunzhuo:
*Description*:

This proposal is about adding a canary release of EG. This canary version will be pre-released everytime we get PR merged. Take this as an example: https://github.com/fermyon/spin/releases/tag/canary.

We can always get the latest pre-release artifacts, like install.yaml. quickstart.yaml, etc. 

1. This is good for developers who want to try the latest dev version of EG, not the official release(We know that the duration of a release is long, so if people need some features which get merged but not release, it makes them hard to try.)

2. This is good for release-manager who wants to have a pre-release test of a offical release, like what I did in https://github.com/Xunzhuo/gateway/releases/tag/v0.5.0, but in the fork repo, there are many places need to be configured like canceling the coverage uploading / set up github pages / config the personal docker registry.. This is complex. By adding a canary release of EG, it makes everything easy.

Xunzhuo:
See: https://github.com/Xunzhuo/gateway/releases and https://github.com/Xunzhuo/gateway/actions/runs/3247935540.

danehans:
+1 for supporting an EG canary release. @arkodg @youngnick @skriss @LukeShu thoughts?

arkodg:
+1

LukeShu:
Calling this "canary" doesn't quite mesh with my use of the word "canary", but that's not real important.

+1 to doing these builds

PR #569, opened 2022-10-14
feat: support canary release of EG

Xunzhuo:
Resolves: #566 

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -30,12 +30,14 @@ jobs:
 
   build-and-test:
     runs-on: ubuntu-latest
+    needs: [lint, gen-check]
     steps:
     - uses: actions/checkout@v3
     - uses: ./tools/github-actions/setup-deps
 
     # test
-    - run: make go.test.coverage
+    - name: Run Coverage Tests
+      run: make go.test.coverage
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v2
       with:
@@ -45,7 +47,8 @@ jobs:
         verbose: true
 
     # build
-    - run: make build-multiarch
+    - name: Build Multiarch EG Binaries
+      run: make build-multiarch PLATFORMS="linux_amd64 linux_arm64"
 
     # conformance
     - name: Run Conformance Tests
@@ -59,7 +62,8 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
-    - run: make image.multiarch.setup
+    - name: Setup Multiarch Environment
+      run: make image.multiarch.setup
 
     - name: Build and Push EG Commit Image
       if: github.event_name == 'push'
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -1,0 +1,53 @@
+name: Canary Release
+
+on:
+  push:
+    branches:
+    - "main"
+    paths-ignore:
+    - "**/*.png"
+
+jobs:
+  canary-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Generate Release Manifests
+      run: make generate-manifests IMAGE=envoyproxy/gateway-dev TAG=latest OUTPUT_DIR=release-artifacts
+
+    # Ignore the error from the first time canary release deletion.
+    # We do not have the release at first, after that, the error will not appear again.
+    - name: Delete Canary Release
+      continue-on-error: true
+      run: |
+        gh release delete latest --repo $GITHUB_REPOSITORY
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_REPOSITORY: ${{ github.repository_owner }}/${{ github.event.repository.name }}
+
+    # Same as above, ignore the error from the first time canary tag deletion.
+    # We do not have the tag at first, after that, the error will not appear again.
+    - name: Delete Canary Tag
+      continue-on-error: true
+      run:
+        gh api --method DELETE /repos/$GITHUB_REPOSITORY/git/refs/tags/latest
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_REPOSITORY: ${{ github.repository_owner }}/${{ github.event.repository.name }}
+
+    - name: Recreate Canary Release and Tag
+      uses: softprops/action-gh-release@v1
+      with:
+        draft: false
+        prerelease: true
+        tag_name: latest
+        files: |
+          release-artifacts/install.yaml
+          release-artifacts/quickstart.yaml
+        body: |
+          This is a "Canary Release" of **Envoy Gateway**, which contains the most recent commits on our main branch.
+          
+          Canary is **not stable**.
+
+          It is only intended for developers wishing to try out the latest features in Envoy Gateway, some of which may not be fully implemented.


.github/workflows/build_and_test.yaml
```diff
@@ -70,3 +74,33 @@ jobs:
...
+    if: github.ref == 'refs/heads/main'
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./tools/github-actions/setup-deps
+
+    - name: Generate Release Manifests
+      run: make generate-manifests IMAGE=envoyproxy/gateway-dev TAG=latest OUTPUT_DIR=release-artifacts
+
+    - name: Delete canary tag
+      uses: dev-drprasad/delete-tag-and-release@v0.2.0
```
arkodg, line 91: is this a popular active action
ive used this in the past
```
    # Upload artifacts to the `edge` release
    - uses: pyTooling/Actions/releaser@r0
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
        files: |
          ...
        tag: 'edge'
```

.github/workflows/build_and_test.yaml
```diff
@@ -70,3 +74,33 @@ jobs:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: canary
```
arkodg, line 95: should this be edge or canary or pre release ?

.github/workflows/build_and_test.yaml
```diff
@@ -70,3 +74,33 @@ jobs:
       if: github.event_name == 'push' && github.ref == 'refs/heads/main'
       # tag is set to `latest` when pushing to main branch
       run: make image.push.multiarch TAG=latest PLATFORMS="linux_amd64 linux_arm64" IMAGE=envoyproxy/gateway-dev
+
+  canary-release:
```
arkodg, line 78: can you move this to a pre-release.yaml file mimicking https://github.com/envoyproxy/gateway/blob/main/.github/workflows/release.yaml

Xunzhuo: I think this is a part a build-and-test, and this requires build and tests passed correctly, moving this to another workflow makes it hard to know if build and test passed.

.github/workflows/build_and_test.yaml
Xunzhuo: let us call it canary, is it OK?

.github/workflows/build_and_test.yaml
arkodg: you can use the merge to main trigger to trigger the GHA, if something has been merged, it must have built and passed successfully

Xunzhuo: Yes, I think this makes sense, I will separate them.

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,36 @@
+name: Canary Release
+
+on:
+  workflow_run:
+    workflows: ["Build and Test"]
```
arkodg, line 5: prefer if this was GH trigger based and not based on the workflow names, and causes uneeded coupling

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,36 @@
...
+      run: make generate-manifests IMAGE=envoyproxy/gateway-dev TAG=latest OUTPUT_DIR=release-artifacts
+
+    - name: Delete canary tag
+      uses: dev-drprasad/delete-tag-and-release@v0.2.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: canary
+
+    - name: Recreate canary tag and release
```
arkodg, line 27: why not use `softprops/action-gh-release`, with `prerelease:true` does it delete older files ?

.github/workflows/pre_release.yaml
Xunzhuo: This is used for triggering canary-release after completing the build-and-test workflow, when pushing to the main branch, in case that canary has been released but the build and test workflow failed. This is just my concern, because build-and-test workflow often failed after pushing to main, even this commit has passed when sending a PR.

.github/workflows/pre_release.yaml
Xunzhuo: I will have a test.

Xunzhuo: This can work, updated.

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  canary-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Generate Release Manifests
+      run: make generate-manifests IMAGE=envoyproxy/gateway-dev TAG=latest OUTPUT_DIR=release-artifacts
+
+    - name: Delete canary tag
```
arkodg, line 19: is delete needed ?

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - completed
+
+jobs:
+  canary-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Generate Release Manifests
+      run: make generate-manifests IMAGE=envoyproxy/gateway-dev TAG=latest OUTPUT_DIR=release-artifacts
```
arkodg, line 17: tag here should match the tag name below

.github/workflows/pre_release.yaml
arkodg: I agree with your point that a broken `Build and Test` will result in a stale `edge` artifact, but there are many ways to ensure that doesnt happen.
triggering workflows from UI is common so we dont want this action to get triggered for those cases

.github/workflows/pre_release.yaml
Xunzhuo: This tag is the image tag, built by build-and-test workflow, we do not have canary tag, we just have the latest tag and a commit id tag.

arkodg: yeah im recommending they both be the same

.github/workflows/pre_release.yaml
Xunzhuo: Yes, otherwise the git tag is already existed, we cannot create a new one, then we always have a static tag without newest commits.

.github/workflows/pre_release.yaml
Xunzhuo: Let us call them both 'latest'? Tag name is latest too.

Xunzhuo: I do not suggest building an image with tag called canary, people often use latest instead, if we want them to be same, let us call them both latest?

.github/workflows/pre_release.yaml
Xunzhuo: Thanks, I will change this.

.github/workflows/pre_release.yaml
arkodg: yeah `latest` is a good start and it won't break anything if `Build and Test` fails :trollface:

Xunzhuo: Thanks, updated.

.github/workflows/pre_release.yaml
arkodg: can we use an action thats actively used by the community ? we need to be careful here

Xunzhuo: Any suggestions?

Xunzhuo: I did not find other better options to do this, deleting tags and release. 😄

arkodg: not atm :),  let me dig around a little, ive seen some repos do this by hand (`git tag -d`)
apart from this, this PR looks good ! this will be super useful for downstream projects and users testing regularly

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,41 @@
+name: Canary Release
+
+on:
+  push:
+    branches:
+    - "main"
+    - "release-v*"
```
danehans, line 7: Shouldn't the canary release only be built from the main branch?

Xunzhuo: Right, update it later. Thanks!

danehans (approved):

Xunzhuo:
Tested at: https://github.com/Xunzhuo/gateway/actions/runs/3258312065/jobs/5350296738
Canary release: https://github.com/Xunzhuo/gateway/releases/tag/latest

Everything works fine now, and we get rid of third party of GH actions to do deletion operation of tag and release.
We use GH CLI to do this, make everything under control, in case something bad happens when using the third party action, which have write access to our repo.

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,53 @@
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_REPOSITORY: ${{ github.repository_owner }}/${{ github.event.repository.name }}
+
+    # Same as above, ignore the error from the first time canary tag deletion.
+    # We do not have the tag at first, after that, the error will not appear again.
+    - name: Delete Canary Tag
+      continue-on-error: true
+      run:
+        gh api --method DELETE /repos/$GITHUB_REPOSITORY/git/refs/tags/latest
```
arkodg, line 34: thanks for making this change and reusing the `gh` cli in the workflow.

arkodg: `latest` is a special keyword in GH https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases, can we make sure deleting `latest` will delete the `latest` tag and not the latest `vX.Y.Z` tag

Xunzhuo: This is the tag name here:)

Xunzhuo: git/tags/ref/{git-tag-name}

arkodg (approved):
LGTM, thanks for the multiple iterations and addressing the comments !

AliceProxy (approved):
lgtm

LukeShu (commented):
Lotsa tools tend to assume that Git tags are immutable.  `git fetch` won't update a tag that changed. SaaS tools tend to cache them forever.

Is there a way we can do this without creating a Git tag?

.github/workflows/pre_release.yaml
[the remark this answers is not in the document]
Xunzhuo: Running `git push --delete origin refs/tags/latest` in GA will fail, the GA cannot directly write to the repo. So we can just do this by Github API:)

LukeShu (approved):
Like I said, I'd prefer to avoid creating a Git tag at all. But that's not a blocker.

Issue #437, opened 2022-09-27
HTTPRouteHostnameIntersection Conformance Test Fails


danehans:
The [HTTPRouteHostnameIntersection](https://pkg.go.dev/sigs.k8s.io/gateway-api/conformance/tests#HTTPRouteHostnameIntersection) conformance test is failing.

Test Logs:
```
Sending Request:
< GET /s5 HTTP/1.1
< Host: wildcard.io
< User-Agent: Go-http-client/1.1
< Accept-Encoding: gzip
<
<

Received Response:
< HTTP/1.1 404 Not Found
< Date: Tue, 27 Sep 2022 19:13:44 GMT
< Server: envoy
< Content-Length: 0
<
<

    http.go:165: Request passed
Received Response:
< HTTP/1.1 404 Not Found
< Date: Tue, 27 Sep 2022 19:13:44 GMT
< Server: envoy
< Content-Length: 0
<
<

=== CONT  TestGatewayAPIConformance/HTTPRouteHostnameIntersection/HTTPRoutes_that_do_not_intersect_with_listener_hostnames/0_request_to_specific.but.wrong.com/s5_should_go_to_
    http.go:165: Request passed
=== CONT  TestGatewayAPIConformance/HTTPRouteHostnameIntersection
    apply.go:204: Deleting no-intersecting-hosts HTTPRoute
    apply.go:204: Deleting wildcard-host-matches-listener-wildcard-host HTTPRoute
    apply.go:204: Deleting wildcard-host-matches-listener-specific-host HTTPRoute
    apply.go:204: Deleting specific-host-matches-listener-wildcard-host HTTPRoute
    apply.go:204: Deleting specific-host-matches-listener-specific-host HTTPRoute
    apply.go:204: Deleting httproute-hostname-intersection Gateway
=== CONT  TestGatewayAPIConformance
    apply.go:160: Deleting certificate
    apply.go:204: Deleting web-backend Deployment
    apply.go:204: Deleting web-backend Service
    apply.go:204: Deleting gateway-conformance-web-backend Namespace
    apply.go:204: Deleting app-backend-v2 Deployment
    apply.go:204: Deleting app-backend-v2 Service
    apply.go:204: Deleting app-backend-v1 Deployment
    apply.go:204: Deleting app-backend-v1 Service
    apply.go:204: Deleting gateway-conformance-app-backend Namespace
    apply.go:204: Deleting infra-backend-v3 Deployment
[...]

Issue #434, opened 2022-09-27
HTTPRouteListenerHostnameMatching Conformance Test Fails


danehans:
The [HTTPRouteListenerHostnameMatching](https://pkg.go.dev/sigs.k8s.io/gateway-api/conformance/tests#HTTPRouteListenerHostnameMatching) conformance test is failing.

Test Logs:
```
http.go:159: Response expectation failed, not ready yet: expected status code to be 200, got 404
Sending Request:
< GET / HTTP/1.1
< Host: multiple.prefixes.foo.com
< User-Agent: Go-http-client/1.1
< Accept-Encoding: gzip
<
<

Received Response:
< HTTP/1.1 404 Not Found
< Date: Tue, 27 Sep 2022 17:42:46 GMT
< Server: envoy
< Content-Length: 0
<
<

=== CONT  TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/3_request_to_boo.bar.com/_should_go_to_infra-backend-v3
    http.go:159: Response expectation failed, not ready yet: expected status code to be 200, got 404
...
--- FAIL: TestGatewayAPIConformance (57.92s)
    --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching (6.92s)
        --- PASS: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/6_request_to_foo.com/_should_go_to_ (0.01s)
        --- PASS: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/7_request_to_no.matching.host/_should_go_to_ (0.01s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/4_request_to_multiple.prefixes.bar.com/_should_go_to_infra-backend-v3 (30.00s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/0_request_to_bar.com/_should_go_to_infra-backend-v1 (30.00s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/1_request_to_foo.bar.com/_should_go_to_infra-backend-v2 (30.00s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerHostnameMatching/2_request_to_baz.bar.com/_should_go_to_infra-backend-v3 (30.00s)
[...]

PR #574, opened 2022-10-16
Merge ir listeners into single xds listener

arkodg:
* Reuse an xds listener if one already exists for the same addr:port
* Add filter chain matches with a unique filter chain and route config for each of these ir listeners, except for http listeners which share the same DefaultFilterChain and the same route config
* removed the xds resource naming for now since it wasnt adding much value

Fixes: https://github.com/envoyproxy/gateway/issues/437
Fixes: https://github.com/envoyproxy/gateway/issues/434


arkodg:
with this PR, we might be able to collapse all Gateway resources into a single Envoy deployment, assuming there is no intersection of Hostnames across the Gateways

AliceProxy (approved):
This all seems fine to me. Nice going knocking out the last two conformance issues for 0.2.0! :rocket: :1st_place_medal:

danehans (approved):

PR #581, opened 2022-10-17
feat(telemetry): init metrics observability

Xunzhuo:
image

image


arkodg:
thanks for kickstarting this work @Xunzhuo . Before implementing this, would be good to track the intent (what kind of metrics and from whom (Envoy Gateway and/or Envoy Proxy)) are we monitoring with a GH issue, and what is a good way to start achieving it.

Xunzhuo:
close this, and will raise a desgin first

PR #590, opened 2022-10-18
Update licensing

### LINKED ISSUE #92: Finalize License Header Management
Kubernetes and many other projects include a license header in each project code file while Envoy only requires license headers at the project root. With https://github.com/envoyproxy/gateway/pull/84, EG will follow the Kubernetes approach. Determine if this approach is needed, and if not, follow Envoy's approach to license header management.

cc: @lizan @mattklein123

youngnick:
This PR adds license headers and license checks as discussed in #92.
It also adds a CI check for license headers.

It's a pretty big PR, sorry, but I had to touch every single Go file in the repo.

I've left it to only Go files for now, the tool _can_ do YAML files as well, but I figured this was big enough already.

Closes #92.


api/config/v1alpha1/groupversion_info.go
```diff
@@ -1,7 +1,11 @@
+// Copyright 2022 Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
 // Package v1alpha1 contains API Schema definitions for the config v1alpha1 API group.
 //
-//+kubebuilder:object:generate=true
-//+groupName=config.gateway.envoyproxy.io
+// +kubebuilder:object:generate=true
+// +groupName=config.gateway.envoyproxy.io
```
danehans, line 8: This change is related to gofmt changes in 1.19. Since the project still uses 1.18, please revert.

tools/boilerplate/boilerplate.generatego.txt
```diff
@@ -0,0 +1,4 @@
+// Copyright Envoy Gateway Authors
```
danehans, line 1: It would be nice if the boilerplate template ended with a blank line so a space exists before the `package` declaration in go files.

tools/boilerplate/boilerplate.go.txt
```diff
@@ -0,0 +1,4 @@
+// Copyright YEAR Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
```
danehans, line 4: It would be nice if the boilerplate template ended with a blank line so a space exists before the `package` declaration in go files.

internal/provider/kubernetes/kubernetes.go
```diff
@@ -1,3 +1,7 @@
+// Copyright 2022 Envoy Gateway Authors
```
danehans, line 1: Should we remove the year from the boilerplate so we don't have to maintain it?

xref: https://github.com/kubernetes-sigs/gateway-api/commit/afbd2cc7db6cf4a08e9b3bc969baca9f83a85171

tools/make/tools.mk
```diff
@@ -13,6 +13,7 @@ $(tools.bindir)/%: $(tools.srcdir)/%.sh
 # ====================
 #
 tools/controller-gen = $(tools.bindir)/controller-gen
+# tools/controller-gen = /home/youngnick.linux/go/bin/controller-gen
```
danehans, line 16: Should this be removed?

tools/src/controller-gen/go.mod
```diff
@@ -6,7 +6,7 @@ require sigs.k8s.io/controller-tools v0.9.2
 
 require (
```
danehans, line 7: Why is a mod bump needed?

danehans (changes requested):
@youngnick thanks for making this happen. I have a few nits throughout.

tools/src/controller-gen/go.mod
youngnick: It shouldn't be - I thought I had reverted this, sigh.

tools/boilerplate/boilerplate.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python
```
arkodg, line 1: any way to reuse rather than copy code over and mantain ?

.github/workflows/build_and_test.yaml
```diff
@@ -28,6 +28,13 @@ jobs:
     - uses: ./tools/github-actions/setup-deps
     - run: make -k gen-check
 
+  license-check:
```
arkodg, line 31: nit: if we have more of these, we should consider creating `check` and add `licencse-check` and `gem-check` under it

tools/boilerplate/boilerplate.py
youngnick: Not easily, no. This should need minimal maintenance though, since it's only using core python features, and already does everything we need.

.github/workflows/build_and_test.yaml
youngnick: Yes, I went back and forth on breaking this out, but decided three should be the critical number. Does that seem okay?

tools/src/golangci-lint/pin.go
```diff
@@ -1,6 +1,11 @@
...
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
 //go:build pin
 // +build pin
 
 package ignore
 
-import "github.com/golangci/golangci-lint/cmd/golangci-lint"
+import _ "github.com/golangci/golangci-lint/cmd/golangci-lint"
```
youngnick, line 11: Adding these makes these imports not get removed by `gofmt`, so auto-format-on-save won't break them.

.github/workflows/build_and_test.yaml
arkodg: sg thanks for sharing the thought process

tools/src/golangci-lint/pin.go
LukeShu: `gofmt` won't remove them, but `goimports` (which is mostly a drop-in replacement for `gofmt`, and is what your auto-format-on-save is likely using) will.

api/config/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,13 @@
...
 // +build !ignore_autogenerated
 
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+
+
 // Code generated by controller-gen. DO NOT EDIT.
```
LukeShu, line 11: Surely the linter is complaining about all these extra newlines? (Because `gofmt` would remove them)

internal/status/status.go
```diff
@@ -141,10 +151,11 @@ func (u *UpdateWriter) Send(update Update) {
...
 // Supported objects:
-//  GatewayClasses
-//  Gateway
-//  HTTPRoute
-//  TLSRoute
+//
+//	GatewayClasses
+//	Gateway
+//	HTTPRoute
+//	TLSRoute
```
LukeShu, line 158: This is due to a change in gofmt from 1.18 to 1.19. This should be in a separate Go 1.19 PR.

.github/workflows/build_and_test.yaml
LukeShu: On the other side, I'd argue that this should be part of the existing `make lint`.

tools/src/golangci-lint/pin.go
LukeShu: But this is also going to require a change to the regex in `tools.mk`

```patch
-	cd $(<D) && GOOS= GOARCH= go build -o $(abspath $@) $$(sed -En 's,^import "(.*)".*,\1,p' pin.go)
+	cd $(<D) && GOOS= GOARCH= go build -o $(abspath $@) $$(sed -En 's,^import _ "(.*)".*,\1,p' pin.go)
```

tools/boilerplate/verify-boilerplate.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
LukeShu, line 1: The existing `golangci-lint` can do this too, https://golangci-lint.run/usage/linters/#goheader no need to introduce a bunch more Python and Bash.

tools/src/golangci-lint/pin.go
youngnick: Oh, that'll teach me to work on this late. Adding the underscore breaks the whole tooling, Go doesn't consider the import relevant any more and so can't do `go build`, it seems. Sigh, I'll roll it back, no editing `pin.go` without using a separate editor for me.

tools/boilerplate/verify-boilerplate.sh
youngnick: Done, converted. There are now two places to update the boilerplate though, `tools/linter/golangci-lint/.golangci.yml` (golangci-lint config) and `tools/boilerplate.txt` (controller-gen template). The former expects no comment marks, while the latter requires them.

internal/status/status.go
youngnick: Reverted.

api/config/v1alpha1/zz_generated.deepcopy.go
youngnick: This was an artifact of how controller-gen was working, fixed.

.github/workflows/build_and_test.yaml
youngnick: `boilerplate.py` has been removed, so this target has also been removed.

tools/boilerplate/verify-boilerplate.sh
youngnick: After doing this and rebasing, there were two files missing headers, but the golangci-lint header check doesn't tell you the names, just that there are files missing headers, when it runs in Github Actions, anway. I spent about an hour poking at it, but I've reverted to `boilerplate.py`, since that actually has useful error messages.

If we want to change it later, that's fine, but someone else will need to have a crack.

.github/workflows/build_and_test.yaml
youngnick: I'm going to leave this as-is for now, so that we can have proper licensing for our first big public release. I'm happy for targets to move around at a later date though.

danehans: Since we're at the v0.2.0 release date, we can iterate to improve in v0.3.0.

danehans (approved):

Issue #591, opened 2022-10-18
Querying the Quickstart Echoserver Image Does Not Provide Expected Response


danehans:
https://github.com/envoyproxy/gateway/pull/578 updated example images to use the kube ingress conformance echoserver. However, it did not set the necessary pod environment variables so calling `/get` does not produce the expected results. For example:

```
curl -v -HHost:www.example.com --resolve "www.example.com:8080:${GATEWAY_HOST}" http://www.example.com:8080/get
* Added www.example.com:8080:172.18.255.200 to DNS cache
* Hostname www.example.com was found in DNS cache
*   Trying 172.18.255.200:8080...
* Connected to www.example.com (172.18.255.200) port 8080 (#0)
> GET /get HTTP/1.1
> Host:www.example.com
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
...
 },
 "namespace": "",
 "ingress": "",
 "service": "",
 "pod": ""
```

See the example [HTTP routing manifest](https://github.com/envoyproxy/gateway/blob/main/examples/kubernetes/http-routing.yaml#L40-L71) for details.

Xunzhuo:
Hi, I can fix this, plz assign this to me:)

danehans:
Fixed by #592

PR #592, opened 2022-10-18
fix: add pod / ns env to echo server

Xunzhuo:

### DIFF
--- a/examples/kubernetes/quickstart.yaml
+++ b/examples/kubernetes/quickstart.yaml
@@ -59,6 +59,15 @@ spec:
           name: backend
           ports:
             - containerPort: 3000
+          env:
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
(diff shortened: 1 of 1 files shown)


examples/kubernetes/quickstart.yaml
```diff
@@ -59,6 +59,15 @@ spec:
           name: backend
           ports:
             - containerPort: 3000
+          env:
```
arkodg, line 62: can you share the impact - before/after ?

arkodg: nvm just saw this https://github.com/envoyproxy/gateway/issues/591

arkodg: can you please attach a log :)

Xunzhuo: before:
image

after:

image

arkodg (approved):
🚀

PR #595, opened 2022-10-18
docs: update quickstart to use latest release

Xunzhuo:
Resolves: https://github.com/envoyproxy/gateway/issues/594

Signed-off-by: bitliu <bitliu@tencent.com>

PR #603, opened 2022-10-19
Updates Sphinx Docs

danehans:
- Adds markdown docs to Sphinx index.
- Lowercases docs names.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -20,8 +20,9 @@ standalone or Kubernetes-based application gateway.
 
   intro/index
   intro/compatibility
-  user/QUICKSTART
+  user/quickstart
   user/http-routing
+  user/http-request-headers
   design/system-design
   design/ROADMAP
   design/gatewayapi-translator
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,5 @@
-Envoy Gateway: envoyproxy + Gateway API
-=======================================
+Envoy Gateway
+=============
 
 Release |version| (Envoy |envoyVersion|, Gateway API |gatewayAPIVersion|)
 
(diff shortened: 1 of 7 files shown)


danehans:
Marking as draft until other docs get merged. I'll then update the index and push a new commit.

docs/index.rst
```diff
@@ -20,7 +20,9 @@ standalone or Kubernetes-based application gateway.
 
   intro/index
   intro/compatibility
-  user/QUICKSTART
```
Xunzhuo, line 23: This should be careful, if you only change QUCkSTART.md to quickstart.md, no changes will be made to git workspace

Xunzhuo (changes requested):

docs/index.rst
Xunzhuo: Sphinx will not find user/quickstart, so it will disappear..

docs/index.rst
```diff
@@ -20,7 +20,9 @@ standalone or Kubernetes-based application gateway.
...
+  user/quickstart
+  user/http-routing
+  user/http-request-headers
   design/system-design
   design/ROADMAP
```
Xunzhuo, line 27: Here too, if you want everything in lowercase.

docs/index.rst
danehans: I thought I updated all QUICKSTART.md references. Am I missing any?

Xunzhuo: Yes, you have updated all QUICKSTART.md referances, but the file name is still QUICKSTART.md not quickstart.md

Xunzhuo: So this ` user/quickstart` will be not found.

Xunzhuo: Or you should configure `git config core.ignorecase false` ?

danehans:
https://github.com/envoyproxy/gateway/pull/595 added most of the changes in this PR. Add the missing http-request-headers.md to the index.

danehans:
This PR is no longer needed since the changes were implemented in #614.

Issue #512, opened 2022-10-05
Add support for HTTPQueryParamMatch in Gateway API Translator


arkodg:
*Description*:
Add support for translating [HTTPQueryParamMatch](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPQueryParamMatch) into the [XDS IR](https://github.com/envoyproxy/gateway/blob/c059da0e582fc39ae49c0f9a9cc5f028f9792675/internal/ir/xds.go#L145) field in the Gateway API translator

zhaohuabing:
@arkodg I would like to take this one if no one else is currently working on it.

PR #606, opened 2022-10-19
Add support for HTTPQueryParamMatch in Gateway API Translator

zhaohuabing:
fix https://github.com/envoyproxy/gateway/issues/512

### DIFF
--- a/internal/gatewayapi/testdata/httproutes-with-multiple-matches.out.yaml
+++ b/internal/gatewayapi/testdata/httproutes-with-multiple-matches.out.yaml
@@ -187,10 +187,9 @@ xdsIR:
       - name: envoy-gateway-httproute-2-rule-0-match-0-example.com
         pathMatch:
           prefix: "/v1/example"
-        # Remove comments once https://github.com/envoyproxy/gateway/issues/512 is fixed
-        # queryParamMatches:
-        # - name: "debug"
-        #   exact: "yes"
+        queryParamMatches:
+        - name: "debug"
+          exact: "yes"
         headerMatches:
         - name: :authority
           exact: example.com
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -1083,6 +1083,14 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 							})
 						}
 					}
+					for _, queryParamMatch := range match.QueryParams {
+						if QueryParamMatchTypeDerefOr(queryParamMatch.Type, v1beta1.QueryParamMatchExact) == v1beta1.QueryParamMatchExact {
+							irRoute.QueryParamMatches = append(irRoute.QueryParamMatches, &ir.StringMatch{
+								Name:  queryParamMatch.Name,
+								Exact: StringPtr(queryParamMatch.Value),
+							})
+						}
+					}
 
 					// Add the redirect filter or direct response that were created earlier to all the irRoutes
 					if redirectResponse != nil {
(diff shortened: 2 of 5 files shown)


arkodg:
looks good @zhaohuabing !
can you also add a test to https://github.com/envoyproxy/gateway/blob/4d722b22593daa5e735751e29670804e8606af66/internal/xds/translator/translator_test.go#L26 which confirms that there IR -> Xds support exists as well, TIA

internal/gatewayapi/translator.go
```diff
@@ -1083,6 +1083,14 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
 							})
 						}
 					}
+					for _, queryParamMatch := range match.QueryParams {
+						if QueryParamMatchTypeDerefOr(queryParamMatch.Type, v1beta1.QueryParamMatchExact) == v1beta1.QueryParamMatchExact {
+							irRoute.QueryParamMatches = append(irRoute.QueryParamMatches, &ir.StringMatch{
+								Name:  queryParamMatch.Name,
+								Exact: StringPtr(queryParamMatch.Value),
+							})
```
zhaohuabing, line 1091: I didn't handle the regex match type here because the path and header neither did than. But I don't know why we don't hanle regex.

arkodg: can you raise an issue to track this
cc @skriss

zhaohuabing: https://github.com/envoyproxy/gateway/issues/608

internal/gatewayapi/testdata/httproutes-with-multiple-matches.out.yaml
```diff
@@ -187,10 +187,9 @@ xdsIR:
       - name: envoy-gateway-httproute-2-rule-0-match-0-example.com
         pathMatch:
           prefix: "/v1/example"
-        # Remove comments once https://github.com/envoyproxy/gateway/issues/512 is fixed
-        # queryParamMatches:
-        # - name: "debug"
-        #   exact: "yes"
+        queryParamMatches:
+        - name: "debug"
+           exact: "yes"
```
Xunzhuo, line 192: nit: remove the extra space.

zhaohuabing:
> looks good @zhaohuabing ! can you also add a test to
> 
> https://github.com/envoyproxy/gateway/blob/4d722b22593daa5e735751e29670804e8606af66/internal/xds/translator/translator_test.go#L26
> 
> which confirms that there IR -> Xds support exists as well, TIA

Test added for QueryParams. I added tests for other match conditions as well.

arkodg:
although this is slated for 0.3.0, if  the HTTPRouteQueryParamMatching test https://github.com/kubernetes-sigs/gateway-api/blob/9e2805c66d603d2be9817e6209cb4bb0f2496562/conformance/tests/httproute-query-param-matching.go#L30 can be adding to our conformance CI
https://github.com/envoyproxy/gateway/blob/main/test/conformance/conformance_test.go
should be okay to merge this today and make it part of 0.2.0 :)

Xunzhuo:
> although this is slated for 0.3.0, if the HTTPRouteQueryParamMatching test https://github.com/kubernetes-sigs/gateway-api/blob/9e2805c66d603d2be9817e6209cb4bb0f2496562/conformance/tests/httproute-query-param-matching.go#L30 can be added to our conformance CI https://github.com/envoyproxy/gateway/blob/main/test/conformance/conformance_test.go should be okay to merge this today and make it part of 0.2.0 :)

I think we already got HTTPRouteQueryParamMatching in? 

See: https://github.com/envoyproxy/gateway/blob/main/test/conformance/conformance_test.go#L55

zhaohuabing:
@arkodg Is anything more I can do on this PR, or we just need to wait after 0.2.0?

Xunzhuo (approved):
Thanks for adding this support, v0.2.0 has been released, I think we can get this in :)

PR #607, opened 2022-10-19
Add a Printable method to Xds IR

arkodg:
* Printable returns a safe copy that can be printed/logged

* Reintroduce printing xds ir in the gateway api runner


AliceProxy (approved):
lgtm

Issue #619, opened 2022-10-19
EG Panics When Removing Gateway


danehans:
EG panics after removing a working Gateway. Follow the quickstart, add HTTPS listeners based on the draft secure gateways guide, delete Gateway:
```
2022-10-19T17:07:26.800Z	INFO	controller/controller.go:117	Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference	{"runner": "provider", "controller": "gateway", "object": {"name":"eg","namespace":"default"}, "namespace": "default", "name": "eg", "reconcileID": "e949958a-6c54-45ea-9a8a-1f962fb09b6d"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x15d3a19]

goroutine 501 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/Users/daneyonhansen/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:118 +0x1f4
panic({0x18ee5a0, 0x2c65220})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/envoyproxy/gateway/internal/provider/kubernetes.(*gatewayReconciler).Reconcile(0xc00010dd10, {0x1e3b8b8, 0xc0005a9da0}, {{{0xc000799798?, 0x10?}, {0xc000799796?, 0x40d7e7?}}})
	/Users/daneyonhansen/go/src/github.com/envoyproxy/gateway/internal/provider/kubernetes/gateway.go:380 +0xed9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x1e3b810?, {0x1e3b8b8?, 0xc0005a9da0?}, {{{0xc000799798?, 0x1aba7a0?}, {0xc000799796?, 0x4041f4?}}})
	/Users/daneyonhansen/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000221680, {0x1e3b810, 0xc000153ac0}, {0x197d0e0?, 0xc0004bc520?})
[...]

PR #622, opened 2022-10-19
fix: panic when removing gateway

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/619

### DIFF
--- a/internal/provider/kubernetes/gateway.go
+++ b/internal/provider/kubernetes/gateway.go
@@ -377,7 +377,12 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 	if !found {
 		gw, ok := r.resources.Gateways.Load(request.NamespacedName)
 		if !ok {
-			r.log.Info("failed to find gateway in the watchable map", "namespace", gw.Namespace, "name", gw.Name)
+			if gw != nil {
+				r.log.Info("failed to find accepted gateway in the watchable map", "namespace", gw.Namespace, "name", gw.Name)
+			} else {
+				r.log.Info("failed to find gateway in the watchable map")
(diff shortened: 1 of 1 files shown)


internal/provider/kubernetes/gateway.go
```diff
@@ -377,7 +377,12 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
 	if !found {
 		gw, ok := r.resources.Gateways.Load(request.NamespacedName)
 		if !ok {
-			r.log.Info("failed to find gateway in the watchable map", "namespace", gw.Namespace, "name", gw.Name)
```
arkodg, line 380: uggh 🙈 , can you just print `request.NamespacedName.String` ?

internal/provider/kubernetes/gateway.go
```diff
@@ -377,7 +377,12 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, request reconcile.Req
...
+			if gw != nil {
```
arkodg, line 380: not needed, since its already `!ok`

Xunzhuo: So I think we should return directly here?

internal/provider/kubernetes/gateway.go
Xunzhuo: Yep

internal/provider/kubernetes/gateway.go
Xunzhuo: When we can load gw from watchable map, we should not let logic to go down, which will cause more panics, I think.

arkodg (approved):
thanks for the fix !

Issue #551, opened 2022-10-12
Per HttpRoute(API) cluster creation


NomadXD:
*Description*:
With the current implementation a cluster is created per each of the routes. For example in the following HttpRoute, 3 clusters will be created one for each "/abc", "/def" and "/ghi".

```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: httpbin
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: httpbin
          port: 80
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /abc
        - path:
            type: PathPrefix
            value: /def
        - path:
            type: PathPrefix
            value: /ghi
```

Envoy proxy uses a cluster to represent a backend service (host and port) and there's NO point of creating multiple clusters if each route of the same HttpRoute points to the same cluster. **The current approach is not very scalable if Envoy proxy is used for an API gateway** with several thousands APIs deployed as memory usage increases with the number of clusters created. If we can have an implementation where we can choose to create clusters per HttpRoute, then we can support a wide range of API gateway related use cases. For example the a swagger / OpenAPI 3 can be directly converted to a HttpRoute and deployed in Envoy Proxy where only 1 cluster is created per API. 

I also know that Gateway API is supposed to be a super set of Ingress API and basically targeted at improving the Ingress API with additional capabilities. **But IMO in order to fully utilize the features of Envoy proxy, we should consider the use cases related to API gateways too.**

Any plans to support these API gateway related use cases in future ? Or is this fully focused on ingress related use cases ?

arkodg:
thanks for raising this issue, adding this to the 0.3.0 milestone
in case someone would like to pick this one, here's one suggestion on how to dedup clusters
* naming the RouteDestination IR https://github.com/envoyproxy/gateway/blob/2fb3ca8954705dabbcfa08007b658655fcb60f46/internal/ir/xds.go#L239 or creating a wrapper struct around the list
https://github.com/envoyproxy/gateway/blob/2fb3ca8954705dabbcfa08007b658655fcb60f46/internal/ir/xds.go#L169
* reusing the named RouteDestination in the Gateway API translator
https://github.com/envoyproxy/gateway/blob/2fb3ca8954705dabbcfa08007b658655fcb60f46/internal/gatewayapi/translator.go#L630
[...]

PR #623, opened 2022-10-19
xds: create clusters per HttpRoute (API)

NomadXD:
Deduplicate cluster creation by creating a cluster per HttpRoute (API). With the current implementation we use the `backendRefs` list in the `spec.rules` and all the backends listed are added as load balance endpoints in the cluster. 

Fixes: #551 

Signed-off-by: Lahiru De Silva [lahirude@wso2.com](mailto:lahirude@wso2.com)


internal/ir/xds.go
```diff
@@ -176,7 +177,7 @@ type HTTPRoute struct {
 	// Redirections to be returned for this route. Takes precedence over Destinations.
 	Redirect *Redirect
 	// Destinations associated with this matched route.
-	Destinations []*RouteDestination
+	DestinationCluster *DestinationCluster
```
arkodg, line 180: would it be easier to guarantee reuse as well as uniqueness if we added a 
`Destination string` here and created a top level `Destinations map[string][]*RouteDestination` within `Xds` ?

danehans:
@NomadXD can you get CI to pass and fix the merge conflicts?

danehans:
@NomadXD can you get CI to pass and fix the merge conflicts?

internal/ir/xds.go
NomadXD: yeah that's a better approach, I will update the PR.

Issue #624, opened 2022-10-19
Add Reference to Gateway API Implementations Page


danehans:
Add a reference to the [Gateway API implementations page](https://gateway-api.sigs.k8s.io/implementations/#envoy-gateway) from Sphinx docs.

PR #627, opened 2022-10-20
docs: add reference to gateway-api and sort index

Xunzhuo:
Resolves: #624

image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,8 +15,8 @@ standalone or Kubernetes-based application gateway.
    This project is under active development. Many, many features are not
    complete. We would love for you to :doc:`get involved<get_involved>`.
 
-   This project is now listed in official 
-   `Gateway API implementations <https://gateway-api.sigs.k8s.io/implementations/#envoy-gateway>`_ .
+   This project is now listed as an official
+   `Gateway API implementation <https://gateway-api.sigs.k8s.io/implementations/#envoy-gateway>`_ .
 
 .. toctree::
   :maxdepth: 1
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -15,6 +15,9 @@ standalone or Kubernetes-based application gateway.
    This project is under active development. Many, many features are not
    complete. We would love for you to :doc:`get involved<get_involved>`.
 
+   This project is now listed in official
```
danehans, line 18: s/in official/as an official/

IMO this information is better conveyed on the compatibility page. Thoughts @arkodg @LukeShu @skriss @AliceProxy?

docs/index.rst
```diff
@@ -15,6 +15,9 @@ standalone or Kubernetes-based application gateway.
    This project is under active development. Many, many features are not
    complete. We would love for you to :doc:`get involved<get_involved>`.
 
+   This project is now listed in official 
+   `Gateway API implementations <https://gateway-api.sigs.k8s.io/implementations/#envoy-gateway>`_ .
```
danehans, line 19: s/implementations/implementation/

arkodg:
imho if we we are linking back to the Gateway API docs, there should be a good reason why it matters to the user
* either we are encouraging them to check out other implementations :)
* or the Gateway API docs has stated the level of conformance/support which Envoy Gateway provides (currently `core` conformance) reassuring the user of feature support.
I think we should highlight the latter point, raised https://github.com/kubernetes-sigs/gateway-api/pull/1475 to highlight that EG passes `core` conformance tests and we should include that verbiage here as well.

danehans:
@Xunzhuo do you mind resolving the review feedback or closing this PR?

Xunzhuo:
Yes @danehans,  I think we can close this PR.

Issue #632, opened 2022-10-20
Support Versioned Docs


danehans:
Currently, docs runs on the main branch. Docs should be versioned, where the default version is the latest supported version, e.g. `v0.2.0`.

Xunzhuo:
I will work on this in v0.3.0-rc1.

PR #639, opened 2022-10-21
feat: support versioned docs

Xunzhuo:
Resolves: #632 

Take this as an example: https://gateway.merbridge.cn

+ For updates in DOC.md can read: https://gateway.merbridge.cn/v0.2.0/dev/DOCS.html
+ For updates in releasing.md can read: https://gateway.merbridge.cn/v0.2.0/dev/releasing.html

---

We can access the current release version of EG by https://gateway.merbridge.cn.

And we can access the latest version of EG by https://gateway.merbridge.cn/latest.

Or by clicking the latest docs at the footer of the pages, which redirects to the /latest pages of the site:

image

We can come back to the current release version of EG by clicking the index title, which redirects to the root page of the site:

image

Checkout the quickstart guides in two different versions:
+ Latest: https://gateway.merbridge.cn/latest/user/quickstart.html
+ Release v0.2.0: https://gateway.merbridge.cn/user/quickstart.html


phlax:
slightly related to this - one problem we had with Envoy docs was that changelogs contained links to content that no longer existed - breaking sphinx in different ways or causing links to be updated to approximately related content

we resolved this by using intersphinx so that all the links in the changelogs are versioned and can point to the docs version where the change occurred.

its a bit of work to setup/maintain but it resolves the problem - not sure if you will have changelogs in docs - but if you will, its probably worth thinking about

Xunzhuo:
Thanks for the comments and suggestions @phlax! I think at least for now, we did not meet such issues, and this PR is to init the versioned docs, which we do not support yet. I think maybe we can make it work at first, and optimize it later if we meet issues like you pointed.

danehans:
@phlax https://github.com/envoyproxy/gateway/pull/639#issuecomment-1287699146 should not be an issue since we use GitHub-generated changelogs and have release-specific release notes. See the [v0.2 release announcement](https://gateway.envoyproxy.io/releases/v0.2.html) as an example.

.github/workflows/docs.yaml
```diff
@@ -3,15 +3,11 @@ on:
   push:
     branches:
       - "main"
-      # Uncomment here until https://github.com/envoyproxy/gateway/issues/632 has been fixed.
-      # - "release/v*"
```
arkodg, line 7: so the GHA file will live within the release branch which will be edited to run that release specific docs pipeline ?

docs/v0.2.0/about_docs.rst
```diff
@@ -0,0 +1,9 @@
+About the documentation
```
arkodg, line 1: so now the `main` branch will contain `docs` dir per release (0.2.0, latest etc) ?

Xunzhuo: Yes, like what envoy-website does too, this is how versioned docs work.

https://github.com/envoyproxy/envoy-website/tree/main/docs/envoy

Xunzhuo: I provided a make target to make this run easily.

.github/workflows/docs.yaml
arkodg: can infer from your below comment that this point is moot now, everything happens on `main`

tools/make/docs.mk
```diff
@@ -1,14 +1,36 @@
-DOCS_DIR := docs
-DOCS_OUTPUT_DIR := docs/html
+DOCS_DIR := docs/latest
```
arkodg, line 1: curious why this is special, and we cannot use `latest` as a `RELEASE_VERSION` ?

Xunzhuo: Thanks for pointing this out, I have updated.

docs/index.html
```diff
@@ -0,0 +1,5 @@
+<html>
+  <head>
+	  <meta http-equiv="refresh" content="0; url=/v0.2.0">
```
arkodg, line 3: is this generated or will someone need to update this as well as the `VERSION` file value during every release ?

docs/v0.2.0/dev/DOCS.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+make docs
+```
+
+This will create `docs/html` with the built HTML pages. You can view the docs
+either simply by pointing a web browser at the `file://` path to your
+`docs/html/latest`, or by firing up a static webserver from that directory, e.g.
+
+``` shell
+cd docs/html/latest ; python3 -m http.server
```
danehans, line 42: This should be `docs/html` to support versioned docs. Otherwise, `code 404, message File not found` when trying to access a versioned doc.

danehans (changes requested):

danehans:
This PR versions __all__ the docs. Do all docs need to be versioned or just the user docs? THoughts @LukeShu @arkodg @Xunzhuo @skriss @AliceProxy @youngnick

docs/index.html
Xunzhuo: The VERSION is generated too, so I think maybe we can generate this file too. Or we can just point to the latest not the current release

docs/v0.2.0/dev/DOCS.md
Xunzhuo: Thanks, I forgot to update this, Will be updated later.

Xunzhuo: image

Test locally and works fine.

docs/index.html
danehans: Can we just take [this approach](https://github.com/envoyproxy/gateway/pull/639/files#diff-b5b4b7207c61dcd99aa1033537dd2f0a81bb0aef3dd79d663718c67ba396c1a8R9) to update this value?

Xunzhuo:
> I run `make docs-release TAG=v0.3.0` and see the v0.3.0 docs content is generated:
> 
> ```
> 
> $ cat VERSION
> 
> v0.3.0
> 
> 
> 
> $ ls -al docs
> 
> total 8
> 
> ...
> 
> drwxr-xr-x  17 daneyonhansen  staff   544 Nov  1 08:23 latest
> 
> drwxr-xr-x  17 daneyonhansen  staff   544 Nov  1 08:23 v0.2.0
> 
> drwxr-xr-x  17 daneyonhansen  staff   544 Nov  1 08:24 v0.3.0
> 
> ```
> 
> 
> 
> However, the docs still default to v0.2.0 when I run locally.

Thanks @danehans, right, we need to update the redirect index.html too. I will update this ASAP.

danehans:
@Xunzhuo can you rebase and include https://github.com/envoyproxy/gateway/pull/665 in this PR?

Issue #641, opened 2022-10-21
Add support for UDPRoute API


arkodg:
*Description*:
Add Support for the UDP Route API defined [here](https://github.com/kubernetes-sigs/gateway-api/blob/21bba43681b3db3a62d7ca9f108bf3be4c0cea71/apis/v1alpha2/udproute_types.go#L33)

Related https://github.com/envoyproxy/gateway/issues/47

arkodg:
we should wait on https://github.com/envoyproxy/gateway/issues/413 which is refactor the provider before kickstarting this

zhaohuabing:
@arkodg May I take this one? It would be a relatively easy start for me to dig into the core mechanism of EG since I can refer to the implementation of the current HTTRoute. If this goes well, I would like to take https://github.com/envoyproxy/gateway/issues/642 as well.

arkodg:
go for it @zhaohuabing . suggesting raising incremental PRs in the below order to increase PR review speeds and reduce merge conflicts :)
* [Xds IR](https://github.com/envoyproxy/gateway/blob/d572c238cfde8623046de85c06b3139479f1d7aa/internal/ir/xds.go#L41)
* [Xds Translator](https://github.com/envoyproxy/gateway/blob/d572c238cfde8623046de85c06b3139479f1d7aa/internal/xds/translator/translator.go#L107)
* [Gateway API Translator](https://github.com/envoyproxy/gateway/blob/d572c238cfde8623046de85c06b3139479f1d7aa/internal/gatewayapi/translator.go#L51)
* Provider (which is currently being refactored)

youngnick:
I'd like to see some work done on how the UDP flows will actually work. It's pretty straightforward to set up a UDP listener for Envoy, but it's very difficult to preserve the client address at the same time.

Basically, you have two options:
- Use Envoy in non-transparent proxy mode, which means that the client address will be seen at the backend pods as Envoy's address. As I said, most UDP protocols don't like this.
- Use Envoy in transparent proxy mode, which means that packets will be forwarded to the backend pods without changing the client IP, *but* the server IP will also not be updated to Envoy's, which means that the return path *from* the backend back to the client will both not go via Envoy, and will be from a different IP address than what the packet was sent to (since it will be the backend's address). Most firewalls will drop UDP traffic like this, that both doesn't come from and address traffic is already heading to, and (most likely) comes from an RFC1918 address.

[...]

PR #646, opened 2022-10-26
add ir for udp route

zhaohuabing:
Add ir for udp route #641 


Xunzhuo (dismissed):
LGTM, thanks!

internal/ir/xds.go
```diff
@@ -471,3 +492,36 @@ func (t TLSInspectorConfig) Validate() error {
...
+// UDPListener holds the UDP listener configuration.
+// +k8s:deepcopy-gen=true
+type UDPListener struct {
+	// Name of the UDPListener
+	Name string
+	// Address that the listener should listen on.
+	Address string
+	// Port on which the service can be expected to be accessed by clients.
+	Port uint32
+	// Destinations associated with TCP traffic to the service.
```
alex-berger, line 505: Did you mean `UDP` instead of `TCP` here?

arkodg:
apart from the TCP typo, looks good @zhaohuabing !
thanks for also adding the TCP Validate calls which was missed earlier :)

youngnick:
Yeah, I'm sorry to be a blocker, but I think we need to write down a design for what UDP options we'll make available, with some explanation of the tradeoffs, before we go too much further here.

danehans:
>  I think we need to write down a design for what UDP options we'll make available, with some explanation of the tradeoffs..

@zhaohuabing do you mind creating a .md for this and submitting a separate PR?

zhaohuabing:
Okay, I'll do that.

zhaohuabing:
> Okay, I'll do that.

https://github.com/envoyproxy/gateway/pull/685

arkodg:
thanks for adding the design doc with https://github.com/envoyproxy/gateway/pull/685, imho this PR should be good to go

arkodg (approved):

PR #653, opened 2022-10-28
internal/gatewayapi: decouple test cases from default proxy image

### LINKED ISSUE #346: internal/gatewayapi: decouple test cases from default Envoy image
We should find a way to not have to bump these test cases every time we change the Envoy version, but fine to put on the backlog for now.

_Originally posted by @skriss in https://github.com/envoyproxy/gateway/pull/338#discussion_r964106420_

skriss:
Adds an optional proxy image override to the
translator so that test cases can be decoupled
from the default image, which will change over
time.

Closes #346.


arkodg (approved):
neat 🚀

PR #654, opened 2022-10-28
update Envoy to v1.24

skriss:

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -136,4 +136,4 @@ There are many other endpoints on the [Envoy admin interface][] that may be help
 [Kubernetes support]: https://docs.docker.com/desktop/kubernetes/
 [gateway-dev]: https://hub.docker.com/r/envoyproxy/gateway-dev/tags
 [mac_connect]: https://github.com/chipmk/docker-mac-net-connect
-[Envoy admin interface]: https://www.envoyproxy.io/docs/envoy/v1.23.0/operations/admin#operations-admin-interface
+[Envoy admin interface]: https://www.envoyproxy.io/docs/envoy/latest/operations/admin#operations-admin-interface
--- a/internal/ir/infra.go
+++ b/internal/ir/infra.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	DefaultProxyName  = "default"
-	DefaultProxyImage = "envoyproxy/envoy:v1.23-latest"
+	DefaultProxyImage = "envoyproxy/envoy:v1.24-latest"
 )
 
 // Infra defines managed infrastructure.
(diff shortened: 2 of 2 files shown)


arkodg:
thanks for bumping this, any specific reason apart from running envoy with the latest image features and fixes ?
took a peek into the release notes https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.24/v1.24.0 and the only thing that stood out was `total_weight` being deprecated (all we'll need to do is delete the line)
https://github.com/envoyproxy/gateway/blob/6f540b51db45fefd95ca13f6f9ec9f835c3c3f7d/internal/xds/translator/route.go#L165

skriss:
> thanks for bumping this, any specific reason apart from running envoy with the latest image features and fixes ?

Just that; I figure as an Envoy-aligned project we'll want to keep pretty up-to-date with the latest and greatest. It also helps us stay on top of deprecations/etc.

> the only thing that stood out was total_weight being deprecated (all we'll need to do is delete the line)

Something to consider here is that if we immediately drop that deprecated field, then it's possible it could cause issues during upgrades from previous versions of EG to this new version -- the EG control plane gets upgraded, and before all the data planes have been upgraded, it sends out an xDS config that does not contain this field, which is invalid for older versions of Envoy. In Contour we tend to upgrade Envoy in one release, and then address deprecations / start making use of new features from that Envoy version in the next Contour release, which tends to avoid this issue. Maybe it's too soon to worry about this for EG, but wanted to raise it for consideration. The other thing we need to think about is EG's release cadence relative to Envoy's -- if we're only releasing every 6mo, while Envoy is every quarter, then we may need to do skip some minor versions of Envoy, which could be tricky. IMO we should consider shipping quarterly as well to align with Envoy, but maybe 2-4 weeks after new Envoy minor versions come out so we can pick them up.

arkodg:
thanks for raising those issues, agree with the depreciation strategy
also agree we consider we should consider quarterly releases, something we can discuss in the next community meeting

danehans:
I agree that EG should coincide with Envoy. If so, we need to review the milestones, roadmap, etc and realign.

danehans:
@skriss do you mind creating issues or agenda topics to address the release cadence and depreciation strategy?

lizan:
> Something to consider here is that if we immediately drop that deprecated field, then it's possible it could cause issues during upgrades from previous versions of EG to this new version -- the EG control plane gets upgraded, and before all the data planes have been upgraded, it sends out an xDS config that does not contain this field, which is invalid for older versions of Envoy.

total_weights has been optional for quite a while so I believe it is safe to remove in this case.

arkodg (approved):

danehans:
> do you mind creating issues or agenda topics to address the release cadence and depreciation strategy?

xref https://github.com/envoyproxy/gateway/issues/663

PR #660, opened 2022-10-28
fix version to v0.2.0 in tls passthrough docs

arkodg:


danehans:
@arkodg can you update the target in tools/make/kube.mk to cover this doc so we don't hit the same issue in the next release?

Xunzhuo (approved):

Xunzhuo:
The versioned docs PR has updated image tag in all .md files under user/, I think we dont need to update it here.

PR #665, opened 2022-10-31
Adds Release Details Doc

### LINKED ISSUE #663: Document Release Schedule
Currently, GitHub milestones are used to document the release schedule. As with [Envoy Proxy](https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#major-release-schedule), the release schedule should be formally documented.

danehans:
Adds a doc that provides EG release details.

Fixes #663 

### DIFF
--- a/docs/releases/README.md
+++ b/docs/releases/README.md
@@ -1,0 +1,41 @@
+# Release Details
+
+This document provides details for Envoy Gateway releases. Envoy Gateway follows the Semantic Versioning [v2.0.0 spec][]
+for release versioning. Since Envoy Gateway is a new project, minor releases are the only defined releases. Envoy
+Gateway maintainers will establish additional release details, e.g. patch releases, at a future date.
+
+## Stable Releases
+
+Stable releases of Envoy Gateway include:
+
+* Minor Releases- A new release branch and corresponding tag are created from the `main` branch. A minor release
+  is supported for 6 months following the release date. As the project matures, Envoy Gateway maintainers will reassess
+  the support timeframe.
+
+Minor releases happen quarterly and follow the schedule below.
+
+## Release Management
+
+Minor releases are handled by a designated Envoy Gateway maintainer. This maintainer is considered the Release Manager
+for the release. The details for creating a release are outlined in the [release guide][].  The Release Manager is
+responsible for coordinating the overall release. This includes identifying issues to be fixed in the release,
+communications with the Envoy Gateway community, and the mechanics of the release.
+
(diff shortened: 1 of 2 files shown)


AliceProxy (approved):
Looks good to me but I think we need to define what "supported" means for a release? I assume security vulnerabilities, but we should have a police about bugfixes and what kind of security patches (high severity / all) would require us to backport fixes to an old release.

arkodg (approved):

danehans:
@AliceProxy thanks for the review and feedback. IMHO "supported" means that we'll answer questions, reproduce, fix, etc. reported issues. [This](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1666990049906419) Slack thread is a good example of what "support" means. I specifically did not include patch or security releases under the `Stable Releases` section (but mentioned it) since we discussed avoiding this level of support until the project matures. Users that require a fix can use the [latest](https://github.com/envoyproxy/gateway/releases/tag/latest) release until we cut the next stable release. I'm open to if/how we provide these additional details.

Xunzhuo (approved):

youngnick (approved):

docs/releases/README.md
```diff
@@ -0,0 +1,41 @@
...
+communications with the Envoy Gateway community, and the mechanics of the release.
+
+| Quarter |                        Release Manager                         |
+|:-------:|:--------------------------------------------------------------:|
+| 2022 Q4 |    Daneyon Hansen ([danehans](https://github.com/danehans))    |
+| 2023 Q1 |                              TBD                               |
+
+## Release Schedule
+
+In order to align with the Envoy Proxy [release schedule][], Envoy Gateway releases are produced on a fixed schedule
```
skriss, line 31: WDYT about moving the target release date back another ~week to the last day of each month? That would give us two weeks between Envoy's target date and EG's, which would make it less likely that an Envoy release slip would affect EG's timeline.

skriss (approved):
LGTM, one suggestion on timelines but fine with whatever the group wants to do.

docs/releases/README.md
LukeShu: Not a strong opinion, but: That sounds like a good idea to me.

danehans: @skriss I'm going to merge since the PR has approvals. I created https://github.com/envoyproxy/gateway/issues/672 to discuss the idea of tracking [envoy-dev](https://hub.docker.com/r/envoyproxy/envoy-dev/tags) to help achieve the 7-day release target, PTAL.

Issue #642, opened 2022-10-21
Add support for the GRPCRoute API


arkodg:
*Description*:

Add support for the GRPCRoute API defined [here](https://github.com/kubernetes-sigs/gateway-api/blob/21bba43681b3db3a62d7ca9f108bf3be4c0cea71/apis/v1alpha2/grpcroute_types.go#L64)

arkodg:
we should wait on https://github.com/envoyproxy/gateway/issues/413 which will refactor the provider before starting this issue

arkodg:
I can commit to delivering this for v0.3.0
@zhaohuabing if you have cycles as well, can split this issue into smaller units so we can tag team this one in

arkodg:
blocked on https://github.com/envoyproxy/gateway/issues/716 since the API is not available in v0.5

arkodg:
waiting on other work items to complete in the gateway-api library before rebasing and completing this work item

danehans:
Docs are needed.

PR #666, opened 2022-10-31
Enhance HTTP IR to support GRPCRoute

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/642

### DIFF
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -102,6 +102,8 @@ type HTTPListener struct {
 	TLS *TLSListenerConfig
 	// Routes associated with HTTP traffic to the service.
 	Routes []*HTTPRoute
+	// IsHTTP2 is set if the upstream client as well as the downstream server are configured to serve HTTP2 traffic.
+	IsHTTP2 bool
 }
 
 // Validate the fields within the HTTPListener structure


danehans:
@arkodg can you provide additional context on how this change supports GRPCRoute?

From Envoy's [gRPC docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/grpc):

> At the transport layer it uses HTTP/2 or above for request/response multiplexing.

Does this mean gRPC-Web will not be supported for GRPCRoute?

arkodg:
> @arkodg can you provide additional context on how this change supports GRPCRoute?
> 
> From Envoy's [gRPC docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/grpc):
> 
> > At the transport layer it uses HTTP/2 or above for request/response multiplexing.
> 
> Does this mean gRPC-Web will not be supported for GRPCRoute?
* this knob could be used to filter for `h2` and `h2c` upstream connections in the listener and will mainly be used to add some more fields in the `cluster` so the upstream connection is http2
* this sadly won't support [grpc-web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#grpc-web), I think of that feature as a seperate `transcoding` feature, 
we might be able to bundle it up and also enable grpc-web on the same downstream connection if the [grpc web filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/grpc_web/v3/grpc_web.proto#envoy-v3-api-msg-extensions-filters-http-grpc-web-v3-grpcweb) allows grpc traffic to flow through seamlessly
@lizan do have any datapoints around this ?

internal/ir/xds.go
```diff
@@ -102,6 +102,8 @@ type HTTPListener struct {
 	TLS *TLSListenerConfig
 	// Routes associated with HTTP traffic to the service.
 	Routes []*HTTPRoute
+	// IsHTTP2 is set if the upstream client as well as the downstream server are configured to serve HTTP2 traffic.
+	IsHTTP2 bool
```
danehans, line 106: @arkodg please take a moment to review the Kube API [conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) regarding bool fields.

arkodg: ha thanks for nudging me to think twice :), here's why I chose a `bool` for now
* most the HTTP logic in `gatewayapi` and `xds` can be reused between HTTP and GRPC APIs, reducing code complexity
* this is an internal API, and should be able to be changed easily w/o user intervention in case the `HTTPRoute` and `GRPCRoute` APIs completely diverge

danehans (approved):

Xunzhuo (approved):

PR #681, opened 2022-11-02
run conformance tests on three Kubernetes versions

### LINKED ISSUE #493: test EG against multiple Kubernetes versions

skriss:
Closes #493.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -37,7 +37,7 @@ jobs:
 
   build-and-test:
     runs-on: ubuntu-latest
-    needs: [lint, gen-check, license-check]
+    needs: [lint, gen-check]
     steps:
     - uses: actions/checkout@v3
     - uses: ./tools/github-actions/setup-deps
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -35,12 +35,14 @@ jobs:
     - uses: ./tools/github-actions/setup-deps
     - run: make -k licensecheck
 
-  unit-tests:
+  build-and-test:
     runs-on: ubuntu-latest
-    needs: [lint, gen-check]
+    needs: [lint, gen-check, license-check]
     steps:
     - uses: actions/checkout@v3
     - uses: ./tools/github-actions/setup-deps
+
+    # test
     - name: Run Coverage Tests
       run: make go.test.coverage
     - name: Upload coverage to Codecov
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -52,16 +50,33 @@ jobs:
         files: ./coverage.xml
         name: codecov-envoy-gateway
         verbose: true
+  
+  conformance-tests:
+    runs-on: ubuntu-latest
+    needs: [unit-tests]
+    strategy:
+      matrix:
+        kind_node_tag: ["v1.24.0", "v1.23.6", "v1.22.9"]
```
arkodg, line 59: this will increase PR time by 6min, I think thats fine for now

skriss: They run in parallel though

.github/workflows/build_and_test.yaml
```diff
@@ -52,16 +50,33 @@ jobs:
...
+
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    needs: [conformance-tests]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./tools/github-actions/setup-deps
 
-    # build
     - name: Build Multiarch EG Binaries
```
arkodg, line 77: I think order matters here, we build the binary and then copy it into the image

.github/workflows/build_and_test.yaml
arkodg: ah amazing !

.github/workflows/build_and_test.yaml
skriss: I'm pretty sure that `make conformance` ends up rebuilding the binary even if it already exists locally, but I'll have to check again tmrw.

arkodg: nvm it should be fine

.github/workflows/build_and_test.yaml
```diff
@@ -52,16 +50,33 @@ jobs:
         files: ./coverage.xml
         name: codecov-envoy-gateway
         verbose: true
+  
+  conformance-tests:
+    runs-on: ubuntu-latest
```
arkodg, line 55: we ideally want unit-tests, conformance and push to be on the same runner

skriss: AFAIK there's no way to run steps within a job in parallel, hence splitting conformance out into its own matrix job so it's not too slow. We can use upload/download artifact steps to share the binaries between jobs (this is the gist of https://github.com/envoyproxy/gateway/issues/683). Are there other reasons for wanting everything on the same runner?

arkodg: the only one is to make sure the same binary is used between all steps, so recommend
* running the jobs sequentially in this PR and then
* #683 can be used to enhance the CI to be able to share builds across runners

arkodg (approved):

PR #687, opened 2022-11-03
fix: panic when getting envoy version

hexiaodai:
Signed-off-by: hexiaodai <jianmin.he@daocloud.io>

### DIFF
--- a/internal/cmd/versions.go
+++ b/internal/cmd/versions.go
@@ -36,7 +36,11 @@ func getVersionsCommand() *cobra.Command {
 
 // versions shows the versions of the Envoy Gateway.
 func versions(envOutput bool) error {
-	envoyVersion := strings.Split(ir.DefaultProxyImage, ":")[1]
+	parts := strings.Split(ir.DefaultProxyImage, ":")
+	if len(parts) < 2 {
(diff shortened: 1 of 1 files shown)


internal/cmd/versions.go
```diff
@@ -36,7 +36,11 @@ func getVersionsCommand() *cobra.Command {
 
 // versions shows the versions of the Envoy Gateway.
 func versions(envOutput bool) error {
-	envoyVersion := strings.Split(ir.DefaultProxyImage, ":")[1]
+	parts := strings.Split(ir.DefaultProxyImage, ":")
```
arkodg, line 39: curious when will we get here ? isn't `DefaultProxyImage` always a sane `constant` ?

Xunzhuo: I can understand the PR tries to check the slice lower bounds of the default proxy image. But honestly saying, this does not make more sence, because the default image is always naming with a `:` to sparate the image tag.

If we support custom proxy image, which is not a const but a string parsing from configs, this kind of checks will make more meaningful. But as for the default image, this is no needed for now.

hexiaodai: That's right. I agree with you completely.

Issue #677, opened 2022-11-02
decision: Segmenting API Gateway features into PolicyAttachment CRDs


arkodg:
*Description*:

Adding options for how API Gateway features such as `ratelimiting` can be expressed as PolicyAttachment CRDs. Please dont focus on the names, just added placeholders.

1. one generic policy CRD with features as top level fields
```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: Policy
 .....
 spec:
   authentication:
   .....
   ratelimiting:
   ....
 ```
2. one generic policy CRD with `networking`  (shapes traffic) and `security` (authenticates & authorizes traffic) as top fields
```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: Policy
 .....
 spec:
   security:
     authentication:
     ....
   networking:
     ratelimiting:
     .....
 ```
3. 2 Policy CRDs - one for `networking` and the other for `security`
```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: SecurityPolicy
 ....
 ....
 spec:
   authentication:
```   

```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: NetworkingPolicy
 ....
 spec:
   ratelimiting:
   ....
```   

4.  1 CRD per API Gateway feature

```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: Authentication
 ....
 spec:
   jwt:
   ....
```   

```
apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: Ratelimiting
 ....
 spec:
   ....
```

arkodg:
@Xunzhuo @danehans @youngnick @skriss @LukeShu @AliceProxy can you please vote on this issue

Xunzhuo:
I prefer 3 or 4, they have more independence and loose coupling between features.

arkodg:
I vote for 2. or 3. and here's why
* 1 or 2 CRDs are easier to maintain for the project than multiple
* The user can express their intent in 1 instantiated resource (useful when team is small) or multiple ( useful when multiple teams author intent)

AliceProxy:
[...]

Issue #675, opened 2022-11-01
Gateway API extensions: Usage of Policy Attachment vs. Filters


sunjayBhatia:
*Description*:
[Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) and [HTTPRouteFilters](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter) represent two different extension points in Gateway API to implement features outside the core/extended APIs provided by the upstream project.

Envoy Gateway is moving forward with implementing more advanced features on top of the core Gateway API and part of this is developing guidelines for implementers/users on how the project will use Policies and Filters.

In this document I wrote up a design/discussion for Contour on differences between using Policies and Filters: https://github.com/projectcontour/contour/pull/4749 which should be relevant to this project as well. In particular the differences between implementing the case-study spikes for rate limiting as a Policy vs. Filter.

Some other key points to discuss that may or may not be included in the above document:
- The Policy Attachment spec has content about Policies and Filters serving different purposes here: https://gateway-api.sigs.k8s.io/references/policy-attachment/#interaction-with-custom-route-filters (particular about them not being used to set the same configuration fields)
  - If we want to use Policies to set defaults or overrides for Filters, we need to address this
- How users know what configuration is actually applied/in effect on their data-plane is an open issue, especially for Policy Attachment
  - See: https://github.com/kubernetes-sigs/gateway-api/discussions/1498#discussioncomment-4030909
  - While we can create a powerful override/default hierarchy to allow flexible configuration, if users do not know what configuration is actually in effect, that is not a great experience
  - Adding Policies influencing how Filters work adds another dimension to this, possibly convoluted user experience
- Mentioned in the document linked above, but Filters in the HTTPRoute sense to me do not directly translate to Envoy HTTP Filters, as they are configured at a much different scope
  - They may work in some cases where there are per-Route incarnations of these filters in Envoy, but that does not exist for every filter, rather most filters are created per HTTPConnectionManager FilterChain

arkodg:
@sunjayBhatia  thanks for sharing all the research on this topic !
The EG community too needs a decide on which extension point to use for the supporting advanced API Gateway features.
I'm sharing some options here, so the community can voice their preference.
1. Use HTTPRoute Filters for all API Gateway features
```
kind: RateLimit
 apiVersion: gateway.envoyproxy.io/v1alpha1
 metadata:
  name: rl-example
 spec:
[...]

Issue #671, opened 2022-11-01
Add Support for WAF


danehans:
*Description*:

Add WAF support. The implementation should focus on [coraza-proxy-wasm](https://github.com/corazawaf/coraza-proxy-wasm) but allow other WAF solutions in the future.
- [ ] Create a design doc for the user-facing API. See #529 (draft) as an example.
- [ ] Update xds IR.
- [ ] Update the Kube provider to reconcile the WAF Custom Resource.
- [ ] Update the resource translator to translate the WAF Custom Resource into IRs.
- [ ] Update the xds translator to translate the xds IR into xds APIs, e.g. ECDS.
- [ ] Add a user-facing doc and link into Sphinx docs, e.g. `user_docs.rst`.
- [ ] Others?

~This issue depends on #71 and/or #24.~

danehans:
@nacx thanks for providing input on this issue. To make sure I understand your comment, are you suggesting that this feature be implemented as an [ExternalProcessor](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter) filter instead of a Wasm filter?

nacx:
My bad. I was mixing the runtime with the control plane, so the suggestion won't apply beyond the fact that coraza is fully written in go and that the project could leverage that, if needed. But at runtime, the wasm module should be the way to go.

arkodg:
this issue shouldnt depend on https://github.com/envoyproxy/gateway/issues/71, both the issues are orthogonal, #71 is about exposing a user facing API for WASM, this issue might use WASM which is an internal implementation detail

Xunzhuo:
> this issue shouldnt depend on https://github.com/envoyproxy/gateway/issues/71, both the issues are orthogonal, #71 is about exposing a user facing API for WASM, this issue might use WASM which is an internal implementation detail

Yes, I am willing to design and implement #71 and this issue, since I am familiar with proxy-wasm. But not sure for the issues priorities, maybe this can be done in v0.3.0, but I can not promise this, this depends on the status of issues I already assigned in v0.3.0 and the priority of wasm extension support.

anuraaga:
Hi all - copying some points from the email thread on this.

[...]

Issue #670, opened 2022-11-01
Add Global Rate Limiting Support


danehans:
*Description*:

Add support for [global rate limiting](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting).
- [ ] Create a design doc for the user-facing API. See #529 (draft) as an example.
- [ ] Update IRs. Maybe we should initially focus on the xds IR and require the global rate limiting infra to be pre-provisioned?
- [ ] Update the Kube provider to reconcile the global rate limiting Custom Resource.
- [ ] Update the resource translator to translate the global rate limiting Custom Resource into IRs.
- [ ] If global rate limiting infra will be supported, update the Kube Infra Manager to provision global rate limiting infra. We should remember that 1) some users may want to use existing global rate limiting infra and 2) we should allow integration with different global rate limiting backends (gRPC required). Use [ratelimit](https://github.com/envoyproxy/ratelimit) as the first, and potentially only, supported implementation.
- [ ] Update the xds translator to translate the xds IR into xds APIs, e.g. `type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit`.
- [ ] Add a user-facing doc and link into Sphinx docs, e.g. `user_docs.rst`.
- [ ] Others?

danehans:
From the Gateway API [Policy Attachment spec](https://gateway-api.sigs.k8s.io/references/policy-attachment/):

> Although Policy attachment can be used to target an entire Route or Backend, it cannot currently be used to target specific Route rules or backend references.

Based on this information, it appears that global rate limiting fits better as an [implementation-specific filter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter) since HTTPRoute rules may require different rate limits. Thoughts @youngnick @skriss?

arkodg:
I can find cycles to work on this

PR #695, opened 2022-11-03
Update roadmap for v0.3.0

arkodg:

### DIFF
--- a/docs/latest/design/roadmap.md
+++ b/docs/latest/design/roadmap.md
@@ -32,10 +32,11 @@ contributing to the project.
 - Achieve Gateway API conformance (e.g. routing, LB, Header transformation, etc.)- [Issue #65][65].
 - Setup a CI/CD pipeline- [Issue #63][63].
 
-### [v0.3.0][v0.3.0]: Driving Advanced Features through Extension Mechanisms
+### [v0.3.0][v0.3.0]: Drive Advanced Features through Extension Mechanisms
 
 - Support extended Gateway API fields.
 - Support experimental Gateway APIs such as TCPRoute [Issue #643][643], UDPRoute [Issue #641][641] and GRPCRoute [Issue #642][642].
+- Establish guidelines for leveragaing Gateway API extensions [Issue #675][675].
 - Rate Limiting [Issue #670][670].
 - Authentication [Issue #336][336].
 
@@ -69,3 +70,4 @@ contributing to the project.
 [642]: https://github.com/envoyproxy/gateway/issues/642
 [643]: https://github.com/envoyproxy/gateway/issues/643
 [670]: https://github.com/envoyproxy/gateway/issues/670
+[675]: https://github.com/envoyproxy/gateway/issues/675
--- a/docs/latest/design/roadmap.md
+++ b/docs/latest/design/roadmap.md
@@ -22,7 +22,7 @@ Roadmap features and timelines may change based on feedback, community contribut
 roadmap item, you're encouraged to attend a community meeting to discuss the details, or help us deliver the feature by
 contributing to the project.
 
-`Last Updated: October 2022`
+`Last Updated: November 2022`
 
 ### [v0.2.0][v0.2.0]: Establish a Solid Foundation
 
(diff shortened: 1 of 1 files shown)


arkodg:
ptal @danehans @LukeShu @AliceProxy @Xunzhuo @skriss @youngnick

AliceProxy (dismissed):
Looks fine to me

Xunzhuo (approved):
🚀

docs/latest/design/roadmap.md
```diff
@@ -32,13 +32,18 @@ contributing to the project.
...
-### [v0.3.0][v0.3.0]: Drive Advanced Features through Extension Mechanisms
+### [v0.3.0][v0.3.0]: Driving Advanced Features through Extension Mechanisms
 
-- Global Rate Limiting
-- AuthN/AuthZ- [Issue #336][336].
-- Lets Encrypt Integration
+- Support extended Gateway API fields.
+- Support experimental Gateway APIs such as TCPRoute [Issue #643][643], UDPRoute [Issue #641][641] and GRPCRoute [Issue #642][642].
+- Rate Limiting [Issue #670][670].
+- Authentication [Issue #336][336].
```
danehans, line 40: IMHO this is overpromising. I would like to understand which maintainers will support these features. We should not depend on the community to deliver roadmap features. I'm in the process of updating #336, but this only covers the AuthN design. From recent feedback (#675 #677) and past experience, the AuthN design alone will take time. Since AuthN is laying the foundation for our approach to Gateway API extensions, this may cause a delay in starting #670. What was the conclusion for delaying RL for RLQS?

docs/latest/design/roadmap.md
```diff
@@ -32,13 +32,18 @@ contributing to the project.
...
 
-### [v0.4.0][v0.4.0]: Manageability and Scale
+### [v0.4.0][v0.4.0]: More Advanced Features through Extension Mechanisms
+
+- Allow users to configure xDS Resources [Issue #24][24].
```
danehans, line 44: I suggest moving some of the v0.3 deliverables to v0.4. Note that the original v0.3 deliverables were based on a 6-month release cycle.

danehans (changes requested):

docs/latest/design/roadmap.md
arkodg: @danehans I assigned #670 to myself, I can design many aspects of it while waiting on the extension decision.
reg RLQS
* we dont have any info on when it will release
* will it be a better fit than Envoy's native RLS for the Gateway case ?
* will it be more stable  ?
so will start the initial design with Envoy RLS, and can raise a follow up issue to consider moving to RLQS when we
have enough data to answer above questions

docs/latest/design/roadmap.md
```diff
@@ -32,13 +32,18 @@ contributing to the project.
 - Achieve Gateway API conformance (e.g. routing, LB, Header transformation, etc.)- [Issue #65][65].
 - Setup a CI/CD pipeline- [Issue #63][63].
 
-### [v0.3.0][v0.3.0]: Drive Advanced Features through Extension Mechanisms
+### [v0.3.0][v0.3.0]: Driving Advanced Features through Extension Mechanisms
```
skriss, line 35: Related to @danehans's comment, maybe worth calling out the deliverable of establishing a pattern for all extensions, since this is a big part of what we're actually doing with the authn design.

skriss (approved):
one suggestion, but LGTM overall

docs/latest/design/roadmap.md
arkodg: Good point, will add

docs/latest/design/roadmap.md
```diff
@@ -34,11 +34,17 @@ contributing to the project.
 
 ### [v0.3.0][v0.3.0]: Drive Advanced Features through Extension Mechanisms
 
-- Global Rate Limiting
-- AuthN/AuthZ- [Issue #336][336].
-- Lets Encrypt Integration
+- Support extended Gateway API fields.
```
Xunzhuo, line 37: Can we link this to a specific issue like other goals in v0.3.0? This will be more clear for a roadmap.

danehans: It should be linked to https://github.com/envoyproxy/gateway/pull/529 and  https://github.com/envoyproxy/gateway/issues/670. IMO https://github.com/envoyproxy/gateway/issues/671 is a stretch for v0.3, so feel free to add it to v0.4 if you feel confident in delivering the feature.

arkodg: I think @Xunzhuo was referring to create a specific epic, created this one https://github.com/envoyproxy/gateway/issues/707 and will link it

danehans:
Codecov upload [failure](https://github.com/envoyproxy/gateway/actions/runs/3413608671/jobs/5680572431), rerunning CI.

docs/latest/design/roadmap.md
Xunzhuo: This is my concern too.

Xunzhuo (commented):
https://github.com/envoyproxy/gateway/issues/657 https://github.com/envoyproxy/gateway/issues/656 https://github.com/envoyproxy/gateway/issues/564

BTW, need we add these three high-priority filters implementations into roadmap?

docs/latest/design/roadmap.md
```diff
@@ -34,11 +34,17 @@ contributing to the project.
...
 ### [v0.3.0][v0.3.0]: Drive Advanced Features through Extension Mechanisms
 
-- Global Rate Limiting
-- AuthN/AuthZ- [Issue #336][336].
-- Lets Encrypt Integration
+- Support extended Gateway API fields [Issue #707][707].
+- Support experimental Gateway APIs such as TCPRoute [Issue #643][643], UDPRoute [Issue #641][641] and GRPCRoute [Issue #642][642].
+- Establish guidelines for leveragaing Gateway API extensions [Issue #675][675].
+- Rate Limiting [Issue #670][670].
+- Authentication [Issue #336][336].
```
danehans, line 41: IMO we're still too aggressive with our v0.3 deliverables. For example, #642 should not be on the roadmap until someone is assigned and has confirmed delivering the feature for the milestone.

arkodg:
> #656

@Xunzhuo these have been updated in the top level epic https://github.com/envoyproxy/gateway/issues/707

Xunzhuo (approved):
LGTM:)

danehans (approved):
Since a majority of maintainers have approved, I will not block. I am concerned that we are overcommitted for v0.3 but we can update the roadmap if needed.

PR #702, opened 2022-11-06
[provider] refactoring kubernetes provider to single reconciler

### LINKED ISSUE #413: Consider Merging Controllers

### LINKED ISSUE #539: A ReferenceGrant Should Be Removed From Watchable Map When UnReferenced

chauhanshubham:
This commit removes separate controllers for gateway API objects, and 
works on only a single controller that now watches for *all* related objects.
Prior to this we had `n` controllers each watching `>=1` resources and pushing
according to the resource map. These pushes happened from these `n` controllers.
For instance; 
- HTTPRoute controller pushing httproute, service and namespace (for refgrant processing) to the resource map
- every route controller duplicating this process
- Gateway controller pushing gateways, referencegrants, namespace, secret to the resource map
- essentially every controller pushing it's namespace to the resource map

Now:
We have a single gatewayAPIController that has a single gatewayAPIReconciler, which reconciles
the gatewayClass object while watching for *ALL* gateway API objects. Any CRUD in these related
objects, triggers the *same* reconciliation logic, where the resource structure is created from scratch,
and pushed as is to the resource map.
Any error during reconciliation - particularly around any object not being found (non-existant/deleted resource),
skips the reconciliation around that particular update, and therefore does not lead to a push to
the resource map. Therefore, the push *ONLY* happens when all related objects exist - `once per reconciliation`.

A few tests are not required so those have been removed - we could work towards adding more
meaningful ones subsequently. Conformance tests remain as is, with two new additions as part of
this PR (commented out reference grant ones)

Resolves #413 
Resolves #539


internal/provider/kubernetes/controller.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	// The gatewayclass was already deleted/finalized and there are stale queue entries.
+	acceptedGC := cc.acceptedClass()
+	if acceptedGC == nil {
+		r.log.Info("failed to find an accepted gatewayclass")
+		return reconcile.Result{}, nil
+	}
+
+	// Store the accepted gatewayclass in the watchable map.
```
arkodg, line 177: this logic will need to change.
you will need to refactor https://github.com/envoyproxy/gateway/blob/54d44d57df53bd9ea9dec77f08a51fdbb0b28ed9/internal/message/types.go#L20 so it looks more like https://github.com/envoyproxy/gateway/blob/54d44d57df53bd9ea9dec77f08a51fdbb0b28ed9/internal/gatewayapi/translator.go#L51
```
import github.com/envoyproxy/gateway/internal/gatewayapi
```
	watchable.Map[string, *gatewayapi.Resources]
}
```

this will allow you to populate the a single struct `gatewayapi.Resources` as you parse the reconciled resources and the last step should be to perform a `Store` into the watchable map

Comment (arkodg) [reply internal/provider/kubernetes/controller.go:177]: with this logic you create a new struct every time the controller reconciles, and use the same key, moving the `diff`
complexity into the watchable layer which will notify the next layer (gateway api) if anything has changed or not

File: internal/provider/kubernetes/route.go [hunk 2]
```diff
@@ -0,0 +1,421 @@
...
+		requests = append(requests, r.processGateway(gateways[j].DeepCopy())...)
+	}
+
+	return requests
+}
+
+// addHTTPRouteIndexers adds indexing on HTTPRoute, for Service objects that are
+// referenced in HTTPRoute objects via `.spec.rules.backendRefs`. This helps in
+// querying for HTTPRoutes that are affected by a particular Service CRUD.
+func addHTTPRouteIndexers(ctx context.Context, mgr manager.Manager) error {
```
arkodg, line 138: trying to understand how does the controller watch for specific Objects i.e. at `t0` when an `HTTPRoute`
references Secret `secret-a`, the controller also watches for changes to `secret-a`. However at `t1` when the
same `HTTPRoute` references `secret-b` instead of `secret-a`, does the logic ensure that we stop watching `secret-a`
and watch `secret-b` instead ?

chauhanshubham:
oh god all the tests passed :')

chauhanshubham:
so brute force single controller is working (all based on existing tests)
On to a few more manual tests and @arkodg 's review comments.
Manual tests would mostly cover experimenting with obj creation ordering.

internal/provider/kubernetes/controller.go
chauhanshubham: it makes sense yes. I will load the obj into the resource table only at the gatewayclass reconcile level, and not at individual watch handlers.

arkodg: you would need to perform a `Store` at the end of reconciling all resources, so rather than having individual controllers, I recommend having one controller 
```
func gatewayAPIController() error {
   var res resources
   // Populate matched gateway class
   res.GatewayClass = gwClass
   // Populate matching gateways
   res.Gateways = append(res.Gateways, gw)
   // Populate Secrets
   .....
   pResources.Store("provider", res)
}
```

we need to perform the `Store` at the end to not cause any temporary churn in the data plane during restarts.

arkodg:
hey @chauhanshubham did you get a chance to review the comments ?

internal/gatewayapi/translator.go
```diff
@@ -58,6 +58,32 @@ type Resources struct {
 	Secrets         []*v1.Secret
 }
 
+func (r *Resources) DeepCopy() *Resources {
```
arkodg, line 61: instead of hand writing this, have you tried adding the `// +k8s:deepcopy-gen=true` annotation above the struct and running `make generate` ?

internal/status/status.go
```diff
@@ -151,10 +151,11 @@ func (u *UpdateWriter) Send(update Update) {
 // isStatusEqual checks if two objects have equivalent status.
 //
 // Supported objects:
-//  GatewayClasses
-//  Gateway
-//  HTTPRoute
-//  TLSRoute
+//
+//	GatewayClasses
```
arkodg, line 155: from go 1.19 ?

internal/gatewayapi/translator.go
chauhanshubham: hey, thanks. I was looking for it, but ended up writing it all and pushing. will do this.

internal/status/status.go
chauhanshubham: yes

internal/provider/kubernetes/route.go
chauhanshubham: I don't think dynamically changing the watch scope would be possible. Although in general, the logic takes care when the service/secret or any other object reference changes, and adds corresponding objects to the resource Map accordingly.

internal/provider/kubernetes/helpers.go
```diff
@@ -61,21 +70,152 @@ func validateParentRefs(ctx context.Context, client client.Client, namespace str
...
+	for _, cond := range gc.Status.Conditions {
+		if cond.Type == string(gwapiv1b1.GatewayClassConditionStatusAccepted) && cond.Status == metav1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+// gatewaysOfClass returns a list of gateways that reference gc from the provided gwList.
+func gatewaysOfClass(gc *gwapiv1b1.GatewayClass, gwList *gwapiv1b1.GatewayList) []gwapiv1b1.Gateway {
```
danehans, line 159: Many of these helper funcs had unit tests in the {controller_name}_test.go. Can you either add these tests or create an issue and link to this PR so they can be readded in a separate PR?

danehans (changes requested):
@chauhanshubham thanks for all your work on this PR. I have a few notes and questions that I would like your feedback on.

internal/provider/kubernetes/controller.go
```diff
@@ -0,0 +1,866 @@
...
+		TLSRoutes:       []*gwapiv1a2.TLSRoute{},
+		Services:        []*corev1.Service{},
+		Secrets:         []*corev1.Secret{},
+		ReferenceGrants: []*gwapiv1a2.ReferenceGrant{},
+		Namespaces:      []*corev1.Namespace{},
+	}
+
+	// Add objects' namespaces into the map.
+	// Make sure to add only those objects' namespaces, that exist.
+	allAssociatedNamespaces := map[string]struct{}{}
```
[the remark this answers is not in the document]
chauhanshubham: having it as a slice i.e. `[]client.Object` might lead to duplicates so a map is preferable.
for `map[string]client.Object` I am not fetching the namespace (or service) while processing the gateway listeners. I store it in these maps and once I'm done processing all listeners, only then I proceed with fetch the entire objects, and storing in the resource map.

internal/provider/kubernetes/controller.go
```diff
@@ -0,0 +1,866 @@
...
+		&source.Kind{Type: &gwapiv1b1.GatewayClass{}},
+		&handler.EnqueueRequestForObject{},
+		predicate.NewPredicateFuncs(r.hasMatchingController),
+	); err != nil {
+		return err
+	}
+
+	// Watch Gateway CRUDs and reconcile affected GatewayClass.
+	if err := c.Watch(&source.Kind{Type: &gwapiv1b1.Gateway{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
```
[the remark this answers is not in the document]
chauhanshubham: yes, I missed these filters as part of this controller. will be adding this gateway filter so that not all gateway updates end up in the reconciliation loop. similarly I will address this comment too https://github.com/envoyproxy/gateway/pull/702#discussion_r1033977585
thanks

skriss (commented):
Thanks @chauhanshubham for taking this on, just a couple comments/questions so far. Still working my way through things.

internal/message/types.go
```diff
@@ -7,120 +7,42 @@ package message
...
-	GatewayClasses watchable.Map[string, *gwapiv1b1.GatewayClass]
-	Gateways       watchable.Map[types.NamespacedName, *gwapiv1b1.Gateway]
-	HTTPRoutes     watchable.Map[types.NamespacedName, *gwapiv1b1.HTTPRoute]
-	TLSRoutes      watchable.Map[types.NamespacedName, *gwapiv1a2.TLSRoute]
-	Namespaces     watchable.Map[string, *corev1.Namespace]
-	Services       watchable.Map[types.NamespacedName, *corev1.Service]
-	Secrets        watchable.Map[types.NamespacedName, *corev1.Secret]
-
-	ReferenceGrants watchable.Map[types.NamespacedName, *gwapiv1a2.ReferenceGrant]
+	GatewayAPIResources watchable.Map[string, *gatewayapi.Resources]
```
skriss, line 21: It'd be helpful to add a godoc comment indicating the key to this map is a GatewayClass name.

internal/provider/kubernetes/controller.go
```diff
@@ -0,0 +1,867 @@
...
+							Namespace: backendNamespace,
+							Name:      string(backendRef.Name),
+						}] = struct{}{}
+
+						if backendNamespace != httpRoute.Namespace {
+							from := ObjectKindNamespacedName{kind: gatewayapi.KindHTTPRoute, namespace: httpRoute.Namespace, name: httpRoute.Name}
+							to := ObjectKindNamespacedName{kind: gatewayapi.KindService, namespace: backendNamespace, name: string(backendRef.Name)}
+							refGrant, err := r.findReferenceGrant(ctx, from, to)
+							if err != nil {
+								r.log.Error(err, "unable to find ReferenceGrant that links the Service to TLSRoute")
```
skriss, line 371:
```suggestion
								r.log.Error(err, "unable to find ReferenceGrant that links the Service to HTTPRoute")
```

internal/provider/kubernetes/controller.go
```diff
@@ -0,0 +1,867 @@
...
+		return err
+	}
+	if err := addTLSRouteIndexers(ctx, mgr); err != nil {
+		return err
+	}
+
+	// Watch Service CRUDs and process affected *Route objects.
+	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, handler.EnqueueRequestsFromMapFunc(r.processServiceForOwningGateway)); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
chauhanshubham: you're right! addressed. thanks

internal/provider/kubernetes/predicates.go
```diff
@@ -0,0 +1,206 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// validateServiceForReconcile tries finding the owning Gateway of the Service
+// if it exists, finds the Gateway's Deployment, and further updates the Gateway
+// status Ready condition. All Services are pushed for reconciliation.
```
danehans, line 149: The only services that should trigger reconciliation are:

- Services referenced by managed routes, e.g. HTTPRoute, should trigger reconciliation.
- The Service that front-ends the managed Envoy fleet, e.g. contain  the managed gateway ns/name owning labels (currently implemented in `validateServiceForReconcile()`)

chauhanshubham: yes, I wanted to check the tests behvaior so pushed early.

danehans:
@chauhanshubham I found a regression when testing this PR. `gateway.status.listeners[].attachedRoutes` is not being calculated properly.

Reproducer:
1. Create a cluster.
2. Install EG using your branch.
3. Install quickstart (verify that you can curl the example app). Check the Gateway status and you should see `attachedRoutes: 1`.
4. Create a 2nd listener.
5. Check the Gateway status again. You should see 2 listeners, each with `attachedRoutes: 2` when each listener should reflect `attachedRoutes: 1` (I confirmed this is the case using the `latest` release).

danehans (approved):
I have tested this PR locally and all looks good. I see #773 and #774 in my testing, but these bugs are not associated with this PR.

danehans:
Merging since this PR is blocking several other PRs.

PR #704, opened 2022-11-07
translator: add accesslog

zirain:
fixes: #684

output as follow:
```
[2022-11-07T01:44:50.341Z] "GET /get HTTP/1.1" 200 - 0 426 0 0 "-" "curl/7.58.0" "f5dec879-4c62-4a7d-a555-d7c044c9d5f0" "www.example.com" "10.96.77.203:3000"
```

this's not for https://github.com/envoyproxy/gateway/issues/699

PR #709, opened 2022-11-08
xds translator for udp route

zhaohuabing:
xds translator for udp route
This PR is a part of https://github.com/envoyproxy/gateway/issues/641


internal/xds/translator/listener.go
```diff
@@ -254,3 +257,65 @@ func buildXdsDownstreamTLSSecret(listenerName string,
 		},
 	}, nil
 }
+
+func buildXdsUDPListener(clusterName string, udpListener *ir.UDPListener) (*listener.Listener, error) {
+	if udpListener == nil {
+		return nil, errors.New("udp listener is nil")
+	}
+
+	statPrefix := "udp"
```
danehans, line 266: From the [UDP Stats docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/udp_proxy#statistics):
> The UDP proxy filter emits both its own downstream statistics as well as many of the [cluster upstream statistics](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#config-cluster-manager-cluster-stats) where applicable. The downstream statistics are rooted at udp.<stat_prefix>.

With that said, `statPrefix` will cause stats to be generated with a prefix of `udp.udp`. Maybe use `service` as the value? Thoughts @arkodg @lizan?

internal/xds/translator/translator.go
```diff
@@ -113,29 +114,57 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
...
-		xdsListener := findXdsListener(tCtx, tcpListener.Address, tcpListener.Port)
+		xdsListener := findXdsListener(tCtx, tcpListener.Address, tcpListener.Port, core.SocketAddress_TCP)
 		if xdsListener == nil {
-			xdsListener = buildXdsListener(tcpListener.Name, tcpListener.Address, tcpListener.Port)
+			xdsListener = buildXdsTCPListener(tcpListener.Name, tcpListener.Address, tcpListener.Port)
 			tCtx.AddXdsResource(resource.ListenerType, xdsListener)
 		}
-
+		// todo: huabing zhao we should check the conflicted listeners: multiple TCP listeners or TCP and HTTP listeners
+		// should not listen on the same port
```
danehans, line 123: Can you create an issue and xref in your comment?

internal/xds/translator/translator.go
```diff
@@ -113,29 +114,57 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
...
 		if err := addXdsTCPFilterChain(xdsListener, tcpListener, xdsCluster.Name); err != nil {
 			return nil, err
 		}
 	}
+
+	for _, udpListener := range ir.UDP {
+		// 1:1 between IR UDPListener and xDS Cluster
```
danehans, line 130: Is the 1:1 b/c UDPRoute does not support any of the [UDP matcher rules](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api#extension-category-envoy-matching-network-input), e.g. source IP?

internal/xds/translator/listener.go
```diff
@@ -254,3 +257,65 @@ func buildXdsDownstreamTLSSecret(listenerName string,
...
+
+	route := &udp.Route{
+		Cluster: clusterName,
+	}
+	routeAny, err := anypb.New(route)
+	if err != nil {
+		return nil, err
+	}
+
+	udpProxy := &udp.UdpProxyConfig{
```
arkodg, line 276: thoughts on keeping this simpler ?
```
              - name: envoy.filters.udp_listener.udp_proxy
                typed_config:
                  '@type': type.googleapis.com/envoy.extensions.filters.udp.udp_proxy.v3.UdpProxyConfig
                  stat_prefix: service
                  cluster: service_udp
                  upstream_socket_config:
                    max_rx_datagram_size: 9000
   ```

File: internal/xds/translator/listener.go (hunk 1 above)
Comment (arkodg) [reply internal/xds/translator/listener.go:266]: good point @danehans , I think we followed a similar pattern for other protocols too, will raise a issue to track and fix for all

File: internal/xds/translator/translator.go (hunk 2 above)
Comment (arkodg) [reply internal/xds/translator/translator.go:123]: the xds code is complicated as is, prefer if we moved all these validations checks to higher levels such as IR or Gateway - API

File: internal/xds/translator/translator.go [hunk 5]
```diff
@@ -113,29 +114,57 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
...
+		xdsCluster, err := buildXdsCluster(udpListener.Name, udpListener.Destinations)
+		if err != nil {
+			return nil, multierror.Append(err, errors.New("error building xds cluster"))
+		}
+		tCtx.AddXdsResource(resource.ClusterType, xdsCluster)
+
+		// Search for an existing listener, if it does not exist, create one.
+		xdsListener := findXdsListener(tCtx, udpListener.Address, udpListener.Port, core.SocketAddress_UDP)
```
arkodg, line 138: why is the `find` needed here, is there any case where we want to support `udp` and another protocol on the same addr:port ?

internal/xds/translator/translator.go
zhaohuabing: @danehans Envoy can route UDP diagrams by their soure IP, but the G: ateway API doesn't support that.

Envoy UDP route: https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/udp_filters/udp_proxy#routing

Gateway UDPRoute API: https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.UDPRouteSpec

internal/xds/translator/listener.go
zhaohuabing: [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/udp/udp_proxy/v3/udp_proxy.proto#envoy-v3-api-msg-extensions-filters-udp-udp-proxy-v3-udpproxyconfig) in favor of matcher since cluster has been deprecated.

internal/xds/translator/translator.go
zhaohuabing: No, but we need to throw an error in case there are two listeners on the same UDP ports.

internal/xds/translator/listener.go
arkodg: ah good point, lets keep this newer API then :)

internal/xds/translator/translator.go
arkodg: i think this case should be handled in the Gateway API translator itself, which has enough info to detect this as well as signal back to the user using the `UDPRoute` `Status` sub resource, prefer if this code is simpler

arkodg:
hey @zhaohuabing can you also enhance this PR with access logs for UDP Proxy similar to whats done in https://github.com/envoyproxy/gateway/pull/704

internal/xds/translator/translator.go
zhaohuabing: > u also enhance this PR with ac

sure

zhaohuabing: > i think this case should be handled in the Gateway API translator itself, which has enough info to detect this as well as signal back to the user using the `UDPRoute` `Status` sub resource, prefer if this code is simpler

OK, we can remove the check here if Gateway API translator can detect it in the first place.

internal/xds/translator/translator.go
zhaohuabing: @danehans @arkodg So we agreed that all the validation should be in Gateway translator and assume the configuration passed down to XDS translator is correct?

arkodg: yeah I prefer that because that layer has the ability to signal back to the user using the `Status` and it performs a lot of the port conflict checks already today
https://github.com/envoyproxy/gateway/blob/9eecce661715577f9a95b0ec28d9a7450716d715/internal/gatewayapi/translator.go#L226

arkodg (approved):
LGTM, looks good, thanks !

PR #710, opened 2022-11-08
opt: run conformance test concurrently in multi-kube-version

Xunzhuo:
Run conformance test concurrently in multi kube-version.

Roughly calculated that the new mode will save about half the time compared to the previous one.

As we add more kube versions to run conformance tests, the running time of build-and-test workflow will nearly not increase. But the former way will cause the time to grow exponentially.

https://github.com/envoyproxy/gateway/actions/runs/3425534566

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -35,9 +35,8 @@ jobs:
     - uses: ./tools/github-actions/setup-deps
     - run: make -k licensecheck
 
-  build-and-test:
+  coverage-test:
     runs-on: ubuntu-latest
-    needs: [lint, gen-check, license-check]
     steps:
     - uses: actions/checkout@v3
     - uses: ./tools/github-actions/setup-deps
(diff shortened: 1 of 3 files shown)


danehans (commented):
/lgtm

.github/workflows/build_and_test.yaml
```diff
@@ -53,29 +53,34 @@ jobs:
...
-    - name: Run Conformance Tests (v1.22.9)
+    - name: Run Conformance Tests 
       env:
-        KIND_NODE_TAG: v1.22.9
+        KIND_NODE_TAG: ${{ matrix.version }}
         CONFORMANCE_UNIQUE_PORTS: false
       run: make conformance
 
+  build-and-publish:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
arkodg: but a different binary is used to run conformance tests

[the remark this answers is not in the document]
Xunzhuo: But for conformance tests, the binaries are different too in the former approach. make conformance will run make image.build every time.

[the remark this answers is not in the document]
arkodg: we need to fix that part first (build once), tests multiple times, push (same build) once

.github/workflows/build_and_test.yaml
```diff
@@ -53,29 +52,68 @@ jobs:
...
+    - uses: ./tools/github-actions/setup-deps
+
+    - name: Download EG Binaries
+      uses: actions/download-artifact@v3
+      with:
+        name: envoy-gateway
+        path: bin/
     
-    - name: Run Conformance Tests (v1.22.9)
+    - name: Give Privileges To EG Binaries
```
arkodg, line 87: should this be moved into the `build` stage ?

.github/workflows/build_and_test.yaml
```diff
@@ -53,29 +52,68 @@ jobs:
...
-    # build
-    - name: Build Multiarch EG Binaries
+  build:
+    runs-on: ubuntu-latest
+    needs: [lint, gen-check, license-check, coverage-test]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./tools/github-actions/setup-deps
+
+    - name: Build EG Multiarch Binaries
```
arkodg, line 62: great stuff, thoughts on uploading the container image rather than binary itself, so we dont need to build the image 3 times before running each conformance test ?
this doesnt need to be tackled in this PR, but would be good to track it with a GH issue

.github/workflows/build_and_test.yaml
Xunzhuo: Nope, the build stage works fine. The problem is that the upload/download action will break binaries privileges, so we can just give them privileges after download them from GitHub action,

.github/workflows/build_and_test.yaml
Xunzhuo: This could be tracked in a new issue, thanks for the tips.

arkodg (approved):
LGTM, thanks !

Issue #672, opened 2022-11-01
Add Support for Main to use envoy-dev Image


danehans:
#665 updates EG releases to coincide with Envoy (1 week after). This may not be enough time for the community to test and fix potential bugs since `main` tracks a released version of Envoy. Instead, have `main` track [envoy-dev](https://hub.docker.com/r/envoyproxy/envoy-dev/tags).  The Envoy image can then be updated to a released version as part of the release workflow.

xref: https://github.com/envoyproxy/gateway/pull/654

arkodg:
this is a great idea in theory, but it also means that PRs raised in the project might flake or fail due to some intermittent issue in the upstream `main`

danehans:
@arkodg Understood, but that could also benefit us by discovering breaking changes early. As EG matures, I could see the benefits being reciprocated to Envoy. In the future, I could also foresee releasing EG on the same day as Envoy which would be a competitive advantage for the project. Both approaches have tradeoffs. Thoughts @LukeShu @youngnick @skriss @Xunzhuo @AliceProxy?

youngnick:
I'm +1 on this, it makes sense for Gateway dev to be linked to Envoy dev. I think that the risks of having flakes because of Envoy `main` having problems is probably a net win, because we'll be further exercising envoy-dev.

Xunzhuo:
I am +1 on this, dev to dev, make sense in main branch, it makes a good collaboration with EG and upstream Envoy.

Xunzhuo:
Wait for others comments, if most think this is OK to process, I would like to assign this.

skriss:
:+1: I think this makes sense

danehans:
xref https://github.com/envoyproxy/gateway/pull/654 for bumping the envoy image.

PR #712, opened 2022-11-08
feat: set envoyproxy image to envoy-dev latest in main

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/672

### DIFF
--- a/internal/ir/infra.go
+++ b/internal/ir/infra.go
@@ -16,7 +16,7 @@ import (
 
 const (
 	DefaultProxyName  = "default"
-	DefaultProxyImage = "envoyproxy/envoy:v1.24-latest"
+	DefaultProxyImage = "envoyproxy/envoy-dev:latest"
 )
 
 // Infra defines managed infrastructure.
@@ -36,7 +36,7 @@ type ProxyInfra struct {
 	// Config defines user-facing configuration of the managed proxy infrastructure.
 	Config *v1alpha1.EnvoyProxy
 	// Image is the container image used for the managed proxy infrastructure.
-	// If unset, defaults to "envoyproxy/envoy:v1.24-latest".
+	// If unset, defaults to "envoyproxy/envoy-dev:latest".
 	Image string
 	// Listeners define the listeners exposed by the proxy infrastructure.
 	Listeners []ProxyListener


api/config/v1alpha1/envoygateway_types.go
```diff
@@ -52,6 +54,12 @@ type Gateway struct {
 	//
 	// +optional
 	ControllerName string `json:"controllerName,omitempty"`
+
+	// ProxyImage is the optional proxy image to use in
+	// the Infra IR to deploy EnvoyProxy. If unspecified, the default proxy
+	// image will be used.
+	// +optional
```
danehans, line 61: - Since this is a user-facing API, I think it's better to s/in the Infra IR to deploy EnvoyProxy/for the managed Envoy proxies/
- IMO this field does not belong in the `Gateway` struct` and I'm unsure if we want to expose the Envoy version to users. Making the version configurable can create a support nightmare of neverending EG<>Envoy combinations. IMO we release EG with the latest Envoy release and users upgrade EG if they need a newer version of Envoy. Thoughts @arkodg @LukeShu @skriss @youngnick @AliceProxy?
- Add an empty line before `+optional`.

danehans (commented):
[decision]: Do we expose the Envoy image version to users?

api/config/v1alpha1/envoygateway_types.go
arkodg: +1 to delaying this API change until users really need it. Post v0.3.0 an exercise we will need to do is
* what level of customizations do we need to provide for
  * xds
  * EG platform/k8s deployment/service
  * Envoy platform/k8s deployment/service
* how do we expose these customizations

danehans: > I think it should be updated in the configuration...

If we expose the Envoy image here, then users can change the image. I don't think we should do this due to the concerns I stated in https://github.com/envoyproxy/gateway/pull/712#discussion_r1016856370.

Xunzhuo: Yes, I understand that @danehans. But for resolving https://github.com/envoyproxy/gateway/issues/672, how do we do it without exposing configuration? By just changing the default image const? We keep the const to the envoy-dev:latest in main. Everytime we release, we update the const to the appropriate envoy release version.

danehans: > We keep the const to the envoy-dev:latest in main. Everytime we release, we update the const to the appropriate envoy release version.

Yes.

Xunzhuo: Okay, if so, this is just a easy-update. I have updated. PTAL @danehans @arkodg.

arkodg (approved):

Issue #184, opened 2022-08-01
Add IPv4/IPv6 Dual Stack Support


danehans:
Currently, EG only supports IPv4. For example, an Infra IR proxy listen address represents an IPv4 address. The code should be audited for IPv4-ism's and updated to support a dual-stack IPv4/IPv6 environment properly.

xref: https://github.com/envoyproxy/gateway/pull/183#issuecomment-1201507246

https://kubernetes.io/docs/concepts/services-networking/dual-stack/

Baenimyr:
On my IPv6 cluster, the readiness probe fails.
> Readiness probe failed: Get "http://[_IPv6_]:19001/ready": dial tcp [_IPv6_]:19001: connect: connection refused

# Test
I ran the test from https://gateway.envoyproxy.io/latest/user/http-routing/

# Fix
I can patch the deployment to use "::" as the listening address (solution discussed [here](https://github.com/envoyproxy/gateway/pull/1568#pullrequestreview-1495217290)), but the envoygateway controller tries to restore it.
The fixed replicaset is not deleted because the restored value is never "Ready".

I can replace the bootstrap values thanks to an [envoy proxy](https://gateway.envoyproxy.io/latest/user/customize-envoyproxy/#customize-envoyproxy-bootstrap-config) but I must rewrite the whole value.

# Complete solution
Can you give me a EnvoyProxy patch to replace only the wrong values.
It would be nice if the listening address could use a "any IPv4 **or IPv6**" syntax.

owenhaynes:
Yeah the `KubernetesServiceSpec` is a bit limited as you can not even configure the `ipFamilyPolicy` let alone additional labels so you can not make the exposed enovy service a dual stack.

Uburro:
> Yeah the `KubernetesServiceSpec` is a bit limited as you can not even configure the `ipFamilyPolicy` let alone additional labels so you can not make the exposed enovy service a dual stack.

i have same problem

Uburro:
just for guys who wanna make ipv6 with envoy-gateway and found out this issue
pathing  Envoy

`apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  annotations:
  name: only-patch
  namespace: envoy-gateway-system
spec:
  jsonPatches:
  - name: envoy-gateway-system/${GATEWAY_NAME}/${GATEWAY_NAME_PORT}
    operation:
      op: replace
      path: /address
      value:
        socket_address:
[...]

Issue #24, opened 2022-04-28
Allow users to configure xDS resources


arkodg:
Users might want to directly configure [xDS Resources](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-types) in case Envoy Gateway's API doesn't support their use case.
They might want to add/merge/remove/replace xDS resources that are pushed by Envoy Gateway into Envoy Proxy.
Projects such as Istio provide APIs such as [EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/) to fulfill these use cases.

Outlining work items

- [x]  Define and Design API https://github.com/envoyproxy/gateway/pull/1410
- [x] Implement API in the Xds Translator https://github.com/envoyproxy/gateway/pull/1606
- [x] Implement support in `egctl` for offline verification https://github.com/envoyproxy/gateway/pull/1682
- [x] Add status to Policy to show runtime status https://github.com/envoyproxy/gateway/pull/1691
- [x] E2E https://github.com/envoyproxy/gateway/pull/1738
- [x] Docs https://github.com/envoyproxy/gateway/pull/1733

NomadXD:
@youngnick What Policy APIs you are referring here ? Kubernetes Gateway API policy attachment ? https://gateway-api.sigs.k8s.io/references/policy-attachment/

youngnick:
Sorry @NomadXD, I missed this.

There are both extension points (usually using an `extensionRef` field that is a reference to an arbitrary CRD, to be defined by the implementation), and Policy Attachment, which is intended for settings that can affect multiple resources (like setting a default for HTTPRoute timeout at the Gateway level by attaching a TimeoutPolicy to the Gateway).

zhaohuabing:
Although it could potentially cause unintended changes in the generated xds and break the whole system, EnvoyFilter is really a powerful API. I think the best use case for EnvoyFilter is it allows other extenstions servers to generate patch for xds. In Aeraki Mesh project, we use EnvoyFilter API to support non-HTTP protocols in Istio. IMHO, EnvoyFilter could also be benificial for use cases like that in Envoy Gateway.

[...]

PR #726, opened 2022-11-11
Global RateLimit Xds translation

arkodg:
* Enhance `XdsIR` with `RateLimit` to hold rate limiting config.

* Translate IR field into route level rate limit actions

* Add `BuildRateLimitServiceConfig` which translates the XdsIR into configuration for the envoy rate limit service.


arkodg:
this PR tries to keep all API Gateway functionality in a separate file, to simplify the core xds translation logic using the pattern of patching xds resources, might be useful to peruse in https://github.com/envoyproxy/gateway/pull/813 @danehans

internal/ir/xds.go
```diff
@@ -494,6 +498,12 @@ func (s StringMatch) Validate() error {
 	if s.SafeRegex != nil {
 		matchCount++
 	}
+	if s.Distinct {
```
zirain, line 501:
```golang
if s.Name == "" {
			errs = multierror.Append(errs, ErrStringMatchNameIsEmpty)
		}else{
		matchCount++
}
```

internal/xds/translator/ratelimit.go
```diff
@@ -0,0 +1,229 @@
...
+func getRateLimitDescriptorKey(prefix string, ruleIndex, matchIndex int) string {
+	return prefix + "-key-rule-" + strconv.Itoa(ruleIndex) + "-match-" + strconv.Itoa(matchIndex)
+}
+
+func getRateLimitDescriptorValue(prefix string, ruleIndex, matchIndex int) string {
+	return prefix + "-value-rule-" + strconv.Itoa(ruleIndex) + "-match-" + strconv.Itoa(matchIndex)
+}
+
+func getRateLimitServiceClusterName() string {
+	return "ratelimit_cluster"
```
zirain, line 219: is this a static cluster? if not, does EG allow end users use multiple RLSs?

internal/ir/xds.go
arkodg: I dont think `else` is right here, we are appending errors, so both cases (name is empty and match is set) need to be considered

internal/xds/translator/translator.go
```diff
@@ -147,6 +159,22 @@ func findXdsListener(tCtx *types.ResourceVersionTable, address string, port uint
 	return nil
 }
 
+// findXdsCluster finds a xds cluster with the same name, and returns nil if there is no match.
+func findXdsCluster(tCtx *types.ResourceVersionTable, name string) *cluster.Cluster {
+	if tCtx == nil || tCtx.XdsResources == nil || tCtx.XdsResources[resource.ClusterType] == nil {
+		return nil
+	}
+
+	for _, r := range tCtx.XdsResources[resource.ClusterType] {
```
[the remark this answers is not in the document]
arkodg: sure, will add a TODO comment in there

internal/xds/translator/ratelimit.go
arkodg: for the default RL feature, EG will build a static cluster, user can connect other RLS and build their own RL filters using https://github.com/envoyproxy/gateway/issues/24

internal/xds/translator/ratelimit.go
```diff
@@ -0,0 +1,229 @@
...
+				headerMatcher := &route.HeaderMatcher{
+					Name: match.Name,
+					HeaderMatchSpecifier: &route.HeaderMatcher_StringMatch{
+						StringMatch: buildXdsStringMatcher(match),
+					},
+				}
+				action := &route.RateLimit_Action{
+					ActionSpecifier: &route.RateLimit_Action_HeaderValueMatch_{
+						HeaderValueMatch: &route.RateLimit_Action_HeaderValueMatch{
+							DescriptorValue: descriptorVal,
```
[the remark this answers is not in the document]
arkodg: cool, thanks for sharing, will add a custom key in here as well

internal/xds/translator/ratelimit.go
```diff
@@ -0,0 +1,229 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+		Http2ProtocolOptions: &core.Http2ProtocolOptions{},
+		DnsRefreshRate:       durationpb.New(30 * time.Second),
+		RespectDnsTtl:        true,
+		DnsLookupFamily:      cluster.Cluster_V4_ONLY,
```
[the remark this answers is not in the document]
arkodg: yes, https://github.com/envoyproxy/gateway/issues/184 has been raised to track support for ipv6

internal/ir/xds.go
```diff
@@ -473,6 +477,9 @@ type StringMatch struct {
 	Suffix *string
 	// SafeRegex match condition.
 	SafeRegex *string
+	// Distinct match condition.
+	// Used to match any and all possible unique values encountered within the Name field.
+	Distinct bool
```
danehans, line 482: The godoc for `StringMatch` should be updated accordingly.

internal/xds/translator/testdata/out/xds-ir/ratelimit.clusters.yaml
```diff
@@ -0,0 +1,71 @@
...
+  lbPolicy: RANDOM
+  loadAssignment:
+    clusterName: ratelimit_cluster
+    endpoints:
+    - lbEndpoints:
+      - endpoint:
+          address:
+            socketAddress:
+              address: TODO
+              portValue: 0
```
danehans, line 68: This will be the address of the RLS endpoint, correct?

internal/xds/translator/translator.go
```diff
@@ -93,6 +94,18 @@ func Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
 		}
 
 		xdsRouteCfg.VirtualHosts = append(xdsRouteCfg.VirtualHosts, vHost)
+
+		// TODO: Make this into a generic interface for API Gateway features
```
danehans, line 98: It would be nice if the TODO references a GH issue.

arkodg: sure will raise a GH issue once this PR is merged and link it

internal/ir/xds.go
arkodg: thanks for catching this, done

internal/ir/xds.go
```diff
@@ -211,6 +212,9 @@ type HTTPRoute struct {
 	Destinations []*RouteDestination
 	// Rewrite to be changed for this route.
 	URLRewrite *URLRewrite
+	// RateLimit defines the more specific match conditions as well as limits for ratelimiting
+	// the requests on this route.
+	RateLimit *RateLimit
```
danehans, line 217: @arkodg I don't understand why RateLimit doesn't use any of the RateLimitFilter fields. For example, #813 uses [JwtAuthenticationFilterProvider](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/authenticationfilter_types.go#L61) instead of recreating the type in the IR. Since we own extension APIs, e.g. RateLimitFilter, we should leverage these types as much as possible.

arkodg: creating a decoupling from Front End as well as Back End API, that was the main goal of the IR, which is why we created custom structures even for known structures such as `StringMatch`

arkodg: as you mentioned, since the project owns the structure, should be safe to reuse it, I can update the PR with this change

danehans:
@arkodg regarding https://github.com/envoyproxy/gateway/pull/726#issuecomment-1370376387, thanks for the suggestion. #813 focuses solely on the xds IR changes. I plan to submit a follow-on PR that adds AuthenticationFilter support to the xds translator. I'll look at following the pattern here for the follow-on PR.

zirain (approved):

internal/xds/translator/translator.go
arkodg: raised https://github.com/envoyproxy/gateway/issues/882 to track this

Issue #336, opened 2022-09-02
Design Authentication Policy as a GatewayAPI Extension


arkodg:
*Description*:

This issue tracks the design of the authentication policy APIs for Envoy Gateway leveraging the extension points within the [Gateway API](https://gateway-api.sigs.k8s.io)

Design docs for extensions are [here](https://docs.google.com/document/d/12jEb81Hf40UeMqXtnGObowB3PzyB21VrF2m4dXGYnls/edit?usp=sharing) & specifically authentication policy are [here](https://docs.google.com/document/d/1TlQjBy1utEwgrxE_HVT4-EHpVJ51hgnfMuAh0Q_uNoE/edit)

lizan:
@danehans do we want to have the docs translated into markdown into https://github.com/envoyproxy/gateway/tree/main/docs/design ?

danehans:
Yes, design docs should be in markdown under `./docs/design`. Refer to the other merged design docs as a reference model for this design doc.

lizan:
Sure, will open an PR for discussion this week.

arkodg:
Design PR for JWT is [here](https://github.com/envoyproxy/gateway/pull/529)

PR #733, opened 2022-11-16
Adds Request Authentication API

danehans:
Adds request authentication API with `JWT` being the first supported provider.

Requires: #529

xref: #336 


arkodg:
thanks @lizan @danehans ! the design LGTM. 
I do prefer if we didn't have to create a new `security` subgroup just yet, but this is not a blocking comment

Issue #643, opened 2022-10-21
Add Support for TCPRoute API


arkodg:
*Description*:
Add support for TCPRoute API defined [here](https://github.com/kubernetes-sigs/gateway-api/blob/21bba43681b3db3a62d7ca9f108bf3be4c0cea71/apis/v1alpha2/tcproute_types.go#L33)

Related https://github.com/envoyproxy/gateway/issues/47

arkodg:
we should wait on https://github.com/envoyproxy/gateway/issues/413 which will refactor the provider before starting this issue

danehans:
@arkodg this seems like a dupe of https://github.com/envoyproxy/gateway/issues/47.

arkodg:
@danehans I broke up #47 into individual issues and linked it back to #47  (one for TCP and other for UDP)

Xunzhuo:
I am assigning myself, but if this is required #413, I think the finished time cannot be measured correctly. Not sure if this can be done in v0.3.0-rc1, but let us see the progress.

youngnick:
Similarly to UDPRoute(#641), I think we need some design docs around the tradeoffs we'll need to make between transparent mode and proxy mode for TCP streams. The choice is basically between:

- transparent mode: client address available, routing changes required somehow to avoid return path check failure in transit.
- proxy mode: client address will appear to be the Envoy IP.

arkodg:
@youngnick we are already using the the [tcp_proxy filter](https://github.com/envoyproxy/gateway/blob/d572c238cfde8623046de85c06b3139479f1d7aa/internal/xds/translator/listener.go#L147) for TLS Passthrough which works well, and envoy **only supports proxy mode** for TCP (docs are [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto#tcp-proxy-proto)) . 
Imho this should unblock the implementation for TCPRoute in the Provider and Gateway API Translator.
To highlight this design decision of only supporting proxy mode for TCP, I can update the decision [here](https://github.com/envoyproxy/gateway/blob/main/docs/design/system-design.md#design-decisions). Are there any other considerations/decisions we should make before continuing with this implementation ?

youngnick:
I think it's fine as long as we call out the limitation. I wouldn't be surprised if people end up wanting us to put in something about PROXY mode eventually, but that seems fine to me.

Xunzhuo:
[...]

PR #737, opened 2022-11-17
feat: add support for TCP Route

Xunzhuo:
Resolves: https://github.com/envoyproxy/gateway/issues/643

Progress:

- [x] Xds IR
- [x] Xds Translator
- [x] Gateway API Translator
- [x] Provider
- [x] Docs


danehans:
@Xunzhuo are you able to complete this PR and the additional work to add TCPRoute support for v0.3.0-rc.1?

Xunzhuo:
Yes, it could be done possibly, I am quite busy this week, will come back to feature implementations this weekend.

docs/latest/user/tcp-routing.md
```diff
@@ -0,0 +1,283 @@
...
+
+Before testing, please get the tcp-gateway Gateway's address first:
+
+```shell
+export GATEWAY_HOST=$(kubectl get gateway/tcp-gateway -o jsonpath='{.status.addresses[0].value}')
+```
+
+You can try to use nc to test the TCP connections of envoy gateway with different ports, and you can see them succeeded:
+
+```shell
```
arkodg, line 212: can you expand a little more for this example

docs/latest/user/tcp-routing.md
```diff
@@ -0,0 +1,283 @@
...
+nc -zv ${GATEWAY_HOST} 8088
+
+nc -zv ${GATEWAY_HOST} 8089
+```
+
+You can try to send request to envoy gateway with different ports and get responses like below:
```
arkodg, line 218:
```suggestion
You can also try to send requests to envoy gateway and get responses as shown below:
```

internal/gatewayapi/contexts.go
```diff
@@ -446,6 +446,88 @@ func (u *UDPRouteContext) GetRouteParentContext(forParentRef v1beta1.ParentRefer
...
+	*v1alpha2.TCPRoute
+
+	parentRefs map[v1beta1.ParentReference]*RouteParentContext
+}
+
+func (u *TCPRouteContext) GetRouteType() string {
+	return KindTCPRoute
+}
+
+// GetHostnames return empty string array because TCPRoute has no hostnames
```
arkodg, line 461: does this method have to be added ?

internal/gatewayapi/helpers.go
```diff
@@ -305,19 +305,24 @@ func hostnameMatchesWildcardHostname(hostname, wildcardHostname string) bool {
 
 func containsPort(ports []*protocolPort, port *protocolPort) bool {
 	for _, protocolPort := range ports {
-		if protocolPort.port == port.port && layer4Protocol(protocolPort) == layer4Protocol(port) {
+		oldProtocol, oldLevel := layer4Protocol(protocolPort)
```
arkodg, line 308: curious why you chose to use `old` and `new` prefix here, doesnt seem to add up here
wouldn't `my` and `cur` be more apt ?

internal/gatewayapi/contexts.go
Xunzhuo: Same as:

```go
// GetHostnames return empty string array because UDPRoute has no hostnames
func (u *UDPRouteContext) GetHostnames() []string {
	return []string{""}
}
```

Xunzhuo: If we need to remove this, we should clean UDPRoute method too.

Xunzhuo: And this is a RouteContext interface, if we do not implement it, there will be reported errors:

```go
// RouteContext represents a generic Route object (HTTPRoute, TLSRoute, etc.)
// that can reference Gateway objects.
type RouteContext interface {
	client.Object

	// GetRouteType returns the Kind of the Route object, HTTPRoute,
	// TLSRoute, TCPRoute, UDPRoute etc.
	GetRouteType() string

	// TODO: [v1alpha2-v1beta1] This should not be required once all Route
	// objects being implemented are of type v1beta1.
	// GetHostnames returns the hosts targeted by the Route object.
	GetHostnames() []string

	// TODO: [v1alpha2-v1beta1] This should not be required once all Route
	// objects being implemented are of type v1beta1.
	// GetParentReferences returns the ParentReference of the Route object.
	GetParentReferences() []v1beta1.ParentReference

	// GetRouteParentContext returns RouteParentContext by using the Route
	// objects' ParentReference.
	GetRouteParentContext(forParentRef v1beta1.ParentReference) *RouteParentContext
}
```

zirain: need to open an ticket track this

Xunzhuo: We should consider if we need to remove this in RouteContext. It seems like a common method for L7 Routes but not for L4 Routes.

```go
	// TODO: [v1alpha2-v1beta1] This should not be required once all Route
	// objects being implemented are of type v1beta1.
	// GetHostnames returns the hosts targeted by the Route object.
	GetHostnames() []string
```

internal/gatewayapi/contexts.go
```diff
@@ -516,6 +599,25 @@ func (r *RouteParentContext) SetCondition(route RouteContext, conditionType v1be
 		} else {
 			r.tlsRoute.Status.Parents[r.routeParentStatusIdx].Conditions = append(r.tlsRoute.Status.Parents[r.routeParentStatusIdx].Conditions, cond)
 		}
+	case KindTCPRoute:
+		for i, existing := range r.tcpRoute.Status.Parents[r.routeParentStatusIdx].Conditions {
+			if existing.Type == cond.Type {
```
zirain, line 604: this seems a little hard to understand

Xunzhuo: Agree, need to do some abstractions to this, but this need to refactor other Routes too.

docs/latest/user/tcp-routing.md
Xunzhuo: Can we expand this with an additional docs PR?

internal/gatewayapi/contexts.go
Xunzhuo: I am removing this.

docs/latest/user/tcp-routing.md
arkodg: sure np

arkodg (approved):
LGTM, thanks for adding this feature in @Xunzhuo 🚀  !

Issue #738, opened 2022-11-17
Wildcard in HttpRoute hostnames generates exact match


paveliak:
*Description*:
When `listeners` spec in the Gateway does not specify hosts and HttpRoute spec specifies _wildcard_ host envoy-gateway generates Exact match for the `:authority` that results in matching failures and hence 404.

E.g. for HttpRoute host name `*.example.com` it generates 
```
HeaderMatches:
      - Exact: '*.example.com'
        Name: :authority
        Prefix: null
        SafeRegex: null
```   

The problem seems to be in [this code](https://github.com/envoyproxy/gateway/blob/cede77ebbecf25fbe8994cfd7643e8295bf75712/internal/gatewayapi/translator.go#L1166) that compares host to `*` while it probably needs to consider `*.example.com` cases

**Expected:**
Regex match is generated instead of Exact match and request passes the filter

*Repro steps*:
Apply this Gateway manifest
```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
```
Apply this HttpRoute manifest:
```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "*.example.com"
  rules:
    - backendRefs:
        - name: paveliak-playground-1
          port: 8080
      matches:
        - path:
            type: PathPrefix
            value: /
```
*Environment*:
- Minikube 1.27.1
- K8s 1.25.2
- envoy-gateway (tried both `v0.2.0` and `main`)

*Logs*:
```
  HTTP:
  - Address: 0.0.0.0
    Hostnames:
    - '*'
    Name: default-eg-http
    Port: 10080
    Routes:
    - AddRequestHeaders: null
      BackendWeights:
        Invalid: 0
        Valid: 0
      Destinations:
      - Host: 10.100.121.40
        Port: 8080
        Weight: 1
      DirectResponse: null
      HeaderMatches:
      - Exact: '*.example.com'
        Name: :authority
        Prefix: null
        SafeRegex: null
      Name: default-backend-rule-0-match-0-*.example.com
      PathMatch:
        Exact: null
        Name: ""
        Prefix: /
        SafeRegex: null
      QueryParamMatches: null
      Redirect: null
      RemoveRequestHeaders: null
[...]

PR #739, opened 2022-11-17
Support suffix matches for headers

arkodg:
* The gateway API allows wildcard hostnames such as `*.example.com` to be used in HTTPRoutes https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Listener

* Added a suffix match in the Xds IR as well as the xds matcher translation to support this case

Fixes: https://github.com/envoyproxy/gateway/issues/738


arkodg:
picked suffix instead of regex, since `suffix` matches are faster

danehans (approved):

PR #741, opened 2022-11-21
add -e to echo command

thredreams:
To solve the problem of invalid escape characters, add the - e option to the echo command in the ". mk" files in the gateway/tool/make directory

### DIFF
--- a/tools/make/common.mk
+++ b/tools/make/common.mk
@@ -14,6 +14,8 @@
 # https://news.ycombinator.com/item?id=16486331
 .SECONDARY:
 
+SHELL:=/bin/bash
+
 # ====================================================================================================
 # ROOT Options:
 # ====================================================================================================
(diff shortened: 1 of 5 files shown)


Xunzhuo:
Plz sign your commit first, and describe the details of the issue you pointed out.

tools/make/common.mk
```diff
@@ -112,6 +112,6 @@ export USAGE_OPTIONS
 .PHONY: help
 help:
 	@echo "Envoy Gateway is an open source project for managing Envoy Proxy as a standalone or Kubernetes-based application gateway\n"
```
Xunzhuo, line 114: Add -e to here too.

Xunzhuo (changes requested):
You need to add `SHELL=/bin/bash` to common.mk, to specify the shell which makefile is using.

thredreams:
> You need to add `SHELL=/bin/bash` to common.mk, to specify the shell which makefile is using.
> 
> Others LGTM. Thank you @thredreams.

In fact, I didn't do this at the beginning because I was worried about whether there were some commands in the file that should not be executed in bash

Xunzhuo:
> In fact, I didn't do this at the beginning because I was worried about whether there were some commands in the file that should not be executed in bash

Not for now : )

Xunzhuo:
Thanks @thredreams, after CI passed, I am going to merge this.

PR #744, opened 2022-11-22
ci: add go mod linter to lint target

Xunzhuo:

After #743.


danehans:
@Xunzhuo #743 has merged so feel free to update this PR.

tools/make/golang.mk
```diff
@@ -78,8 +80,8 @@ test: ## Run all Go test of code sources.
 test: go.test.unit
 
 .PHONY: format
-format: ## Update dependences with mod tidy.
-format: go.tidy
+format: ## Update and check dependences with go mod tidy.
+format: go.mod.lint
```
danehans, line 84: Why do we need the `format` target if it's doing the same thing as the `lint` target?

Xunzhuo: `format` just does the go mod tidy locally, but `lint` includes other linter checks. We can add more format things like goimport under format, which can format our codes.

danehans:
@Xunzhuo LGTM after a rebase.

arkodg (approved):

PR #748, opened 2022-11-22
chore: sync targets logs

Xunzhuo:
+ add LOG_TARGET to all make targets
+ sync and wrap all logs with log func
+ add errorlog func, and wrap all error log with it


Xunzhuo:
Take this as an example: https://github.com/envoyproxy/gateway/actions/runs/3523015932/jobs/5906760579#step:6:1

danehans (approved):

Issue #327, opened 2022-09-01
make envoy-eg pod able to debug


zirain:
correct me if I'm wrong

Right now, there's no debug tool (`curl` etc) in `envoyproxy/envoy`, admin port is listen on `127.0.0.1`, we need provider a way to access `config_dump` for debug?

cc @danehans @arkodg WDYT?

arkodg:
hey @zirain what about port forwarding to the envoy pod ?

zirain:
> hey @zirain what about port forwarding to the envoy pod ?

that need open more than one shell, once the pod changed, you need port forward again.

I want just run one line script like following:
```
kubectl exec -it <pod_name> -nenvoy-gateway-system -- curl 127.0.0.1:19000/config_dump
```

arkodg:
I understand your pain :). I think we might have to continue with above workflow until
* we have a debug subcommand such as `envoy-gateway collect` to collect `config_dump` and much more for offline triaging (we could target this for v0.2.0)
* or we support BYO Envoy Bootstrap config (scheduled for post v0.2.0)

zirain:
Actually I have some workarounds now, just open a ticket to track this.

IMO, it's necessary, but not that urgent.

here's my proposal:
- change the base-images like [this](https://github.com/istio/istio/blob/master/docker/Dockerfile.base)
- add a subcommand like `istioctl pc all`(what we do in istio project) if we have tools like `egctl`

zirain:
replaced by https://github.com/envoyproxy/gateway/issues/794

PR #749, opened 2022-11-24
infra: support custom base image

zirain:

fixes: https://github.com/envoyproxy/gateway/issues/327

this PR make folks can custom base image for better debugging by running following command:
```console
IMAGE_BASE=<your_base_image:-ubuntu:jammy> make kube-install-image
```

### DIFF
--- a/tools/docker/envoy-gateway/Dockerfile
+++ b/tools/docker/envoy-gateway/Dockerfile
@@ -1,6 +1,5 @@
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+ARG IMAGE_BASE
+FROM ${IMAGE_BASE}
 ARG TARGETPLATFORM
 COPY $TARGETPLATFORM/envoy-gateway /usr/local/bin/
 
--- a/tools/make/image.mk
+++ b/tools/make/image.mk
@@ -11,6 +11,9 @@ REGISTRY ?= docker.io/envoyproxy
 IMAGE_NAME ?= gateway-dev
 # IMAGE is the image URL for build and push image targets.
 IMAGE ?= ${REGISTRY}/${IMAGE_NAME}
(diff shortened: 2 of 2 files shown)


danehans (changes requested):
FYI from https://github.com/envoyproxy/gateway/issues/327#issuecomment-1233643635, a command is desired for #327 to be closed.

cc: @arkodg

zirain:
`collect` is not enough, dev users also need change logging level on the fly by `logging`, or check stats from `/stats`
BTW, `envoy-gateway` need a long way to go.

danehans:
I would like to get input from @arkodg before merging this PR due to https://github.com/envoyproxy/gateway/issues/327#issuecomment-1233643635.

arkodg:
since the goal is to improve debuggablity, would prefer EG provided commands, reports directly instead of the user swapping out the container image to something that contains more utility commands in it

danehans:
@zirain is https://github.com/envoyproxy/gateway/pull/749#issuecomment-1337876893 something you can take on? It would be a big help to the project.

arkodg:
> I do not think it's a good idea to provider feature on `eg`, maybe a tools call `egctl`?

yeah `egctl` makes a lot of sense here, which can be the entry point to  `analyze` / `report` / `collect` subcmd ?

zirain:
> > I do not think it's a good idea to provider feature on `eg`, maybe a tools call `egctl`?
> 
> yeah `egctl` makes a lot of sense here, which can be the entry point to `analyze` / `report` / `collect` subcmd ?

yes, that's what I think.

zirain:
@danehans @arkodg let's close this first, then I will raise a proposal about new binary named `egctl`, WDYT?

arkodg:
> @danehans @arkodg let's close this first, then I will raise a proposal about new binary named `egctl`, WDYT?

sounds good, thanks !

zirain:
close this for https://github.com/envoyproxy/gateway/issues/794

Issue #705, opened 2022-11-07
Github action `build-and-test` fails randomly


zirain:
- fail on `make go.test.coverage`

https://github.com/envoyproxy/gateway/actions/runs/3406968389/jobs/5666166782

- fail on `make conformance` v1.23.6

https://github.com/envoyproxy/gateway/actions/runs/3407232578/jobs/5666696862

- fail on `make conformance` v1.24

https://github.com/envoyproxy/gateway/actions/runs/3407315316/jobs/5666855486

zirain:
```
Error from server (InternalError): error when creating "internal/provider/kubernetes/config/samples/gatewayclass.yaml": Internal error occurred: failed calling webhook "validate.gateway.networking.k8s.io": failed to call webhook: Post "[https://gateway-api-admission-server.gateway-system.svc:443/validate?timeout=10s](https://gateway-api-admission-server.gateway-system.svc/validate?timeout=10s)": x509: certificate signed by unknown authority
```

Xunzhuo:
> I think we should wait job completed, but not 100%

Probably right, the cert might not be generated, so the auth is failed.

PR #750, opened 2022-11-24
infra: wait for job succeeded

zirain:

should fix: #705

### DIFF
--- a/tools/make/kube.mk
+++ b/tools/make/kube.mk
@@ -82,6 +82,7 @@ kube-install-image: image.build $(tools/kind) ## Install the EG image to a kind 
 run-conformance: ## Run Gateway API conformance.
 	kubectl wait --timeout=5m -n gateway-system deployment/gateway-api-admission-server --for=condition=Available
 	kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
+	kubectl wait --timeout=5m -n gateway-system pod -l job-name=gateway-api-admission --for=jsonpath='{.status.phase}'=Succeeded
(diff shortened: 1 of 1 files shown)


tools/make/kube.mk
```diff
@@ -82,6 +82,7 @@ kube-install-image: image.build $(tools/kind) ## Install the EG image to a kind
 run-conformance: ## Run Gateway API conformance.
 	kubectl wait --timeout=5m -n gateway-system deployment/gateway-api-admission-server --for=condition=Available
 	kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
+	kubectl wait --timeout=5m -n gateway-system pod -l job-name=gateway-api-admission --for=jsonpath='{.status.phase}'=Succeeded
```
danehans, line 85: Any specific reason for using `jsonpath` over `--for=condition=complete`?

```
$ kubectl get job/gateway-api-admission -n gateway-system -o yaml
...
status:
  completionTime: "2022-11-28T19:27:06Z"
  conditions:
  - lastProbeTime: "2022-11-28T19:27:06Z"
    lastTransitionTime: "2022-11-28T19:27:06Z"
    status: "True"
    type: Complete
  ready: 0
  startTime: "2022-11-28T19:26:57Z"
  succeeded: 1
```

`Succeeded` specifies the number of pods that reached phase "Succeeded". If the number of pods is increased for the Job and one fails, the wait will succeed.

From [Kube API docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#jobstatus-v1-batch):
> When a Job is completed, one of the conditions will have the type "Complete" and status true.

zirain: updated

danehans (approved):

Issue #757, opened 2022-11-30
Update Deprecated http2_protocol_options


danehans:
From Envoy logs:
```
[2022-11-30 00:14:22.851][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.config.cluster.v3.Cluster Using deprecated option 'envoy.config.cluster.v3.Cluster.http2_protocol_options' from file cluster.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

```
$ grep -ri http2_protocol_options internal/infrastructure
internal/infrastructure/kubernetes/bootstrap.yaml.tpl:    http2_protocol_options: {}
```

PR #770, opened 2022-12-02
fix deprecated http2_protocol_options

zirain:

fix: https://github.com/envoyproxy/gateway/issues/757


arkodg (approved):
🚀

PR #780, opened 2022-12-08
Run all core and extended conformance tests

arkodg:
Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #793, opened 2022-12-09
Add ReferenceGrant Support for Routes

### LINKED ISSUE #545: finish implementing ReferenceGrant for HTTPRoute -> Service references
#519 completed the missing piece of implementing ReferenceGrant for Listener -> Secret references. Something similar needs to be done for HTTPRoute -> Service references. The Gateway API translator already has logic to check cross-namespace references (https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/translator.go#L660-L683), but the provider / HTTPRoute reconciler needs to pass the relevant ReferenceGrants into the translator layer.

May also be simpler to wait for https://github.com/envoyproxy/gateway/issues/413 before doing this.

xref. #411.

### LINKED ISSUE #788: backendref service not found causes xds no message received
*Description*:
case: backendref service not found
1、this case ，**xds no message received** 
2、log level error, unnecessary
```
2022-12-09T17:59:13.737+0800    ERROR   kubernetes/controller.go:332    unable to find associated Services      {"runner": "provider", "error": "Service \"test-ad-service\" not found"}
github.com/envoyproxy/gateway/internal/provider/kubernetes.(*gatewayAPIReconciler).Reconcile
        /Users/zsw/Documents/workspace/gateway/envoy-gateway/internal/provider/kubernetes/controller.go:332
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
        /Users/zsw/Documents/workspace/gateway/envoy-gateway/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:121
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /workspace/envoy-gateway/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:320
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /workspace/envoy-gateway/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /workspace/envoy-gateway/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234
```
-----
version:  commit/7e176dec6cdfdee58a940e9c4b969349ced1b0c8

### LINKED ISSUE #792: A BackendRef Should Be Allowed to Include a Namespace
An HTTPRoute backendRef should be allowed to include a namespace. If a ReferenceGrant does not exist that allows the reference, the HTTPRoute should surface an `Accepted=False` condition. If a ReferenceGrant does exist that allows the reference, the HTTPRoute should surface an `Accepted=True` condition.

danehans:
Previously, a route, e.g. HTTPRoute, could not reference a backend in a different namespace due to `validateBackendRef()` returning an error if a BackendRef included the `namespace` field.

- Updates `Reconcile()` to only add a RefereneGrant to the resource tree if the RefereneGrant can be found and only return an error if an error was received when listing RefereneGrants.
- Updates `Reconcile()` to only add s Service to the resource tree if the Service can be found and only return an error if an error was received when Getting the Service.
- Updates `validateBackendRef()` to no longer return an error if the `namespace` field is set for a backendRef.
- Adds the `HTTPRouteReferenceGrant` conformance test.

Fixes #792
Fixes #788
Fixes #545 

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -290,15 +290,29 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 						r.log.Info("processing Secret", "namespace", secretNamespace, "name", string(certRef.Name))
 
 						if secretNamespace != gtw.Namespace {
-							from := ObjectKindNamespacedName{kind: gatewayapi.KindGateway, namespace: gtw.Namespace, name: gtw.Name}
-							to := ObjectKindNamespacedName{kind: gatewayapi.KindSecret, namespace: secretNamespace, name: string(certRef.Name)}
-							refGrant, err := r.findReferenceGrant(ctx, from, to)
-							if err != nil {
-								r.log.Error(err, "unable to find ReferenceGrant that links the Secret to Gateway")
-								continue
+							from := ObjectKindNamespacedName{
+								kind:      gatewayapi.KindGateway,
+								namespace: gtw.Namespace,
+								name:      gtw.Name,
 							}
-
-							resourceMap.allAssociatedRefGrants[utils.NamespacedName(refGrant)] = refGrant
+							to := ObjectKindNamespacedName{
+								kind:      gatewayapi.KindSecret,
+								namespace: secretNamespace,
+								name:      string(certRef.Name),
+							}
+							refGrant, err := r.findReferenceGrant(ctx, from, to)
+							switch {
+							case err != nil:
+								r.log.Error(err, "failed to find ReferenceGrant")
+							case refGrant == nil:
+								r.log.Info("no matching ReferenceGrants found", "from", from.kind,
+									"from namespace", from.namespace, "target", to.kind, "target namespace", to.namespace)
+							default:
+								// RefGrant found
+								resourceMap.allAssociatedRefGrants[utils.NamespacedName(refGrant)] = refGrant
+								r.log.Info("added ReferenceGrant to resource map", "namespace", refGrant.Namespace,
+									"name", refGrant.Name)
+							}
 						}
 
 						resourceMap.allAssociatedNamespaces[secretNamespace] = struct{}{}
@@ -329,15 +343,14 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 		service := new(corev1.Service)
 		err := r.client.Get(ctx, serviceNamespaceName, service)
 		if err != nil {
-			r.log.Error(err, "unable to find associated Services")
-			if kerrors.IsNotFound(err) {
-				return reconcile.Result{}, nil
-			}
-			return reconcile.Result{}, err
+			r.log.Error(err, "failed to get Service", "namespace", serviceNamespaceName.Namespace,
+				"name", serviceNamespaceName.Name)
+		} else {
+			resourceMap.allAssociatedNamespaces[service.Namespace] = struct{}{}
+			resourceTree.Services = append(resourceTree.Services, service)
+			r.log.Info("added Service to resource tree", "namespace", serviceNamespaceName.Namespace,
+				"name", serviceNamespaceName.Name)
 		}
-
-		resourceMap.allAssociatedNamespaces[service.Namespace] = struct{}{}
-		resourceTree.Services = append(resourceTree.Services, service)
 	}
 
 	// Add all ReferenceGrants to the resourceTree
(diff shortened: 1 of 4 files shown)


danehans:
@sunjayBhatia @youngnick @arkodg PTAL at the associated issue and failed run. I think this may be an upstream bug in the `HTTPRouteInvalidNonExistentBackendRef` test.

danehans:
xref https://github.com/envoyproxy/gateway/pull/780 for adding add'l conformance tests.

internal/provider/kubernetes/controller.go
```diff
@@ -290,15 +290,23 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
...
+							to := ObjectKindNamespacedName{
+								kind:      gatewayapi.KindSecret,
+								namespace: secretNamespace,
+								name:      string(certRef.Name),
+							}
 							refGrant, err := r.findReferenceGrant(ctx, from, to)
 							if err != nil {
-								r.log.Error(err, "unable to find ReferenceGrant that links the Secret to Gateway")
-								continue
+								return reconcile.Result{}, err
```
arkodg, line 305: this doesnt look right, we are returning early here, skipping processing and sending other correct info

internal/provider/kubernetes/controller.go
```diff
@@ -329,15 +337,17 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 		service := new(corev1.Service)
 		err := r.client.Get(ctx, serviceNamespaceName, service)
 		if err != nil {
-			r.log.Error(err, "unable to find associated Services")
 			if kerrors.IsNotFound(err) {
-				return reconcile.Result{}, nil
+				r.log.Info("failed to find Service", "namespace", serviceNamespaceName.Namespace,
+					"name", serviceNamespaceName.Name, "error", err)
+			} else {
+				return reconcile.Result{}, err
```
arkodg, line 344: same as above

internal/provider/kubernetes/controller.go
danehans: @arkodg this error is due to the client not being able to list ReferenceGrants. Do we want to recompute the resource tree under this circumstance and potentially cause routes to be inadvertently CRUD'd?

internal/provider/kubernetes/controller.go
danehans: Same response as https://github.com/envoyproxy/gateway/pull/793#discussion_r1049034624.

internal/provider/kubernetes/controller.go
arkodg: this error looks like its for a specific refgrant object
we have two options here 
1) do nothing - which will result in user not being able to apply any new intent
2) skip processing bad object but process the rest - which will result in translating all readable intent
2 seems more correct as well as intuitive than 1

skriss:
@danehans sounds like this closes https://github.com/envoyproxy/gateway/issues/545? If so, can you add that to the "Fixes" list?

internal/provider/kubernetes/controller.go
danehans: Agreed that #2 is more intuitive than #1. I'll update the PR.

arkodg: thanks, I think if the `return` statement is removed, it should satisfy 1)

danehans: Resolved by commit 753e8c7.

internal/provider/kubernetes/controller.go
danehans: Resolved by commit 753e8c7.

internal/provider/kubernetes/controller.go
```diff
@@ -290,15 +290,25 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
...
+							to := ObjectKindNamespacedName{
+								kind:      gatewayapi.KindSecret,
+								namespace: secretNamespace,
+								name:      string(certRef.Name),
+							}
 							refGrant, err := r.findReferenceGrant(ctx, from, to)
 							if err != nil {
-								r.log.Error(err, "unable to find ReferenceGrant that links the Secret to Gateway")
-								continue
+								r.log.Error(err, "failed to find ReferenceGrant")
```
arkodg, line 305: nit, can we add some more info here to difference which exact refgrant was not found

internal/provider/kubernetes/controller.go
```diff
@@ -329,15 +339,17 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 		service := new(corev1.Service)
 		err := r.client.Get(ctx, serviceNamespaceName, service)
 		if err != nil {
-			r.log.Error(err, "unable to find associated Services")
 			if kerrors.IsNotFound(err) {
-				return reconcile.Result{}, nil
+				r.log.Info("failed to find Service", "namespace", serviceNamespaceName.Namespace,
+					"name", serviceNamespaceName.Name, "error", err)
+			} else {
+				return reconcile.Result{}, err
```
arkodg, line 346: shouldnt this `return` be removed as well ?

internal/provider/kubernetes/controller.go
arkodg: nvm, please ignore, looks like the callee function is printing out the details so prob not needed here

internal/provider/kubernetes/controller.go
danehans: Good catch. Yes, commit 365610b  fixes this.

arkodg (approved):

Xunzhuo (approved):

Issue #794, opened 2022-12-10
proposal new tool named `egctl`


zirain:
tracking for https://github.com/envoyproxy/gateway/pull/749#issuecomment-1343762241, will update details later.

danehans:
@zirain I can move the Milestone to v0.3.0-rc.1 if you expect to have a PR ready in time.

zirain:
@danehans let's get design doc merged first.

zirain:
fixed by https://github.com/envoyproxy/gateway/pull/797

PR #797, opened 2022-12-12
`egctl` proposal

zirain:
xref: https://github.com/envoyproxy/gateway/issues/794

cc @arkodg @danehans

### DIFF
--- a/docs/latest/design/egctl.md
+++ b/docs/latest/design/egctl.md
@@ -5,7 +5,6 @@
 EG should provide a command line tool with following capabilities:
 
 - Collect configuration from envoy proxy and gateway
-- Update admin options such as changing log level
 - Analyse system configuration to diagnose any issues in envoy gateway
 
 This tool is named `egctl`.
@@ -39,7 +38,6 @@ The following table includes short descriptions and the general syntax for all t
 | --------- | -------------------------------- | --------------------------------------------------------------------------- |
 | `version` | `egctl version`                  | Prints out build version information.                                       |
 | `config`  | `egctl config ENTITY`            | Retrieve information about proxy configuration from envoy proxy and gateway |
-| `log`     | `egctl log ENTITY --level trace` | Change envoy proxy's log level                                              |
 | `analyze` | `egctl analyze`                  | Analyze EG configuration and print validation messages                      |
 
 ## Examples
@@ -55,7 +53,4 @@ egctl config envoy-proxy listener <instance_name>
 
 # Retrieve information about envoy gateway
 egctl config envoy-gateway
-
-# Change log level of envoy proxy
-egctl log envoy-proxy <instance_name> --level trace
 ```


docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
+# Introduce egctl
+
+## Motivation
+
+EG should provide a command line tool with following capabilities:
+
+- One-stop installation/uninstallation experience
+- Collect configuration form data plane
```
arkodg, line 8:
```suggestion
- Collect configuration from data plane
```

arkodg: can you expand this to "configuration and logs"

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
+# Introduce egctl
+
+## Motivation
+
+EG should provide a command line tool with following capabilities:
+
+- One-stop installation/uninstallation experience
```
arkodg, line 7: did you mean one step here ?

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+- Collect configuration form data plane
+- Change envoy logging configuration on the fly
```
arkodg, line 9: can we expand this to "update admin options such as changing log level"

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+- Analyse system configuration
```
arkodg, line 10: can we expand this to mention "to diagnose any issues in envoy-gateway"

docs/latest/design/egctl.md
arkodg, line 8: can we add another similar bullet for control plane (to collect logs and config)

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+## Operation
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation      | Syntax                          | Description                                               |
+| -------------- |---------------------------------|-----------------------------------------------------------|
+| `version`      | `egctl version`                 | Prints out build version information.                     |
+| `install`      | `egctl install -f FILENAME`     | Install or reconfigure EG on a cluster.                   |
+| `uninstall`    | `egctl uninstall`               | Uninstall EG from a cluster                               |
+| `proxy-config` | `egctl proxy-config SUBCOMMAND` | Retrieve information about proxy configuration from envoy |
```
arkodg, line 42: prefer having two subcommands here - `envoy-gateway` and `envoy-proxy` to add more features within each (data plane and control plane)

arkodg: thinking again, should this be replaced with `collect` instead, so
```
egctl collect all
```
wil collect everything (logs and config) for data plane and control plane

```
egctl collect config --envoy-proxy
```

will collect config associated with envoy proxy

docs/latest/design/egctl.md
zirain: sorry for the nit

docs/latest/design/egctl.md
zirain: what should `egctl` collect from envoy-gateway, is there any debug endpoint on `envoy-gateway`?

arkodg: nothing today, but as you hinted, will need a similar admin endpoint like envoy to toggle log level, get EG config, etc, so we can start building with keeping that in mind, but it shouldnt block this proposal

zirain: my original think was adding a subcommand like `egctl gateway-config` or else when we have that debug endpoint.

zirain: gentle ping @arkodg

arkodg: yeah so the discussion is around what the parent noun should be , imho top level should be "what the user is trying to achieve" - like `install` or `collect` (info) or `analyze` 
@danehans @AliceProxy @youngnick @Xunzhuo @skriss can you share your thoughts as well

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+If you need help, run `kubectl help` from the terminal window.
+
+## Operation
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation      | Syntax                          | Description                                               |
+| -------------- |---------------------------------|-----------------------------------------------------------|
+| `version`      | `egctl version`                 | Prints out build version information.                     |
+| `install`      | `egctl install -f FILENAME`     | Install or reconfigure EG on a cluster.                   |
```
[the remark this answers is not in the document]
zirain: my original thoughts were only the `control-plane` part, I used to think `GatewayClass` and `Gateway` are part of quickstart, not installation。

docs/latest/design/egctl.md
danehans: > yeah so the discussion is around what the parent noun should be , imho top level should be "what the user is trying to achieve" - like install or collect (info) or analyze

Agreed. If the parent noun can be used for both control and data planes, then that noun should expose `control-plane` and `data-plane` flags are exposed to filter the action. See https://github.com/envoyproxy/gateway/pull/797/files#r1050247359 for a detailed example.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+| `uninstall`    | `egctl uninstall`               | Uninstall EG from a cluster                               |
+| `proxy-config` | `egctl proxy-config SUBCOMMAND` | Retrieve information about proxy configuration from envoy |
+| `analyze`      | `egctl analyze`                 | Analyze configuration and print validation messages       |
+
+## Examples
+
+Use the following set of examples to help you familiarize yourself with running the commonly used `egctl` operations:
+
+```console
+# Install EG using the definition in custom-values.yaml
```
danehans, line 50: As I mention here, will the format of `custom-values.yaml` follow the [EnvoyGateway](https://github.com/envoyproxy/gateway/blob/main/api/config/v1alpha1/envoygateway_types.go) spec?

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+egctl install -f custom-values.yaml
+
+# Uninstall EG
+egctl uninstall
```
danehans, line 54: Thinking about future requirements, we should consider an `import` subcommand to start managing unmanaged proxy infra.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Retrieve all information about proxy configuration from envoy
+egctl proxy-config all <pod_name>
```
danehans, line 57: I prefer `config` over the `proxy-config` noun.

To pull config from the managed proxies:
```
egctl get config --data-plane
```

To pull config from EG:
```
egctl get config --control-plane
```

When the `--control-plane` or `--data-plane` are not specified, egctl gets config from control and data planes.

danehans (changes requested):
Specific feedback is provided in separate comments.

docs/latest/design/egctl.md
zirain: Inspired by `istioctl`, `proxy-config` can filter xds by specific type(`endpoint`, `ecds`, `cluster` etc.)

Also, I think we don't have a debug endpoint in `control-plane`.

docs/latest/design/egctl.md
zirain: what's `unmanaged proxy infra` mean, can you share more details?

docs/latest/design/egctl.md
arkodg: @zirain you could achieve the same by `egctl config envoy-proxy cluster` . I prefer this over `egctl envoy-proxy-config cluster` because former maintains a more consistent hierarchy starting from `verb` user is trying to perform

docs/latest/design/egctl.md
zirain: will EG provide operator in the future? if so, it's better to use a CRD named `EnvoyGatewayOperator` for installation,WDYT?

docs/latest/design/egctl.md
zirain: what about `control-plane` with `egctl config` in your mind, `ectl config control-plane`?

arkodg: I prefer `envoy-gateway` and `envoy-proxy` over `control-plane` and `data-plane` but this naming is minor for me.
we can have
```
egctl config envoy-proxy
egctl config envoy-gateway
egctl config envoy-rate-limit-service
egctl config all
```
and we can start with `egctl config envoy-proxy`

arkodg: if you want the `config` subcommand to only display config, then we can create another subcommand called `collect` which can be similar to `bug-report`

zirain: I have no strong option on naming, will update the doc later.

zirain: updated, PTAL

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,67 @@
...
+egctl uninstall
+
+# Retrieve all information about proxy configuration from envoy
+egctl config envoy-proxy all <pod_name>
+
+# Retrieve listener information about proxy configuration from envoy 
+egctl config envoy-proxy listener <pod_name>
+
+# Change log level of envoy proxy
+egctl config envoy-proxy log <pod_name> --level trace
```
arkodg, line 63: prefer
```suggestion
egctl log envoy-proxy <pod_name> --level trace
```

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,67 @@
...
+| `config`    | `egctl config SUBCOMMAND`        | Retrieve information about proxy configuration from envoy proxy and gateway |
+| `analyze`   | `egctl analyze`                  | Analyze configuration and print validation messages                         |
+
+## Examples
+
+Use the following set of examples to help you familiarize yourself with running the commonly used `egctl` operations:
+
+```console
+# Install EG using the definition in custom profile file
+egctl install -f custom-profile.yaml
```
arkodg, line 51: there is no notion of custom profile in EG but a config file, also can we follow same convention of using `-c` https://github.com/envoyproxy/gateway/blob/46b7b7523cc8d8167fd999c9cdecdca2ffb9fd71/internal/cmd/server.go#L36

arkodg:
thanks for the doc update @zirain , it mostly looks good, added some more comments

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,67 @@
...
+
+```console
+egctl [command] [subcommand] [name] [flags]
+```
+
+where `command`, `name`, and `flags` are:
+
+* `command`: Specifies the operation that you want to perform on one or more resources,
+  for example `install`, `uninstall`, `config`, `version`.
+
```
arkodg, line 26: can you add a explanation for `subcommand` or replace it with another term such as `entity` or `resource` : Specifies the entity the operation is being performed on such as `envoy-proxy` or `envoy-gateway`

docs/latest/design/egctl.md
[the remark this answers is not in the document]
danehans: @zirain `egctl install` can start off by only installing the control-plane, but we should consider how the tool can install control and data planes in the future. The control-plane install should include the GatewayClass with `spec.conrtrollerName: gateway.envoyproxy.io/gatewayclass-controller`. A Gateway resource is used to manage the data plane. Something like `egctl install data-plane --namespace foo` will install the data-plane in namespace foo by creating a Gateway resource like:
```
kind: Gateway
metadata:
  name: eg
  namespace: foo
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
```

We also should keep in mind that we would like EG to support additional providers in the future. If `egctl` doesn't;t specify --provider`, then the default Kubernetes provider should be used.

docs/latest/design/egctl.md
danehans: EnvoyGateway is the spec for configuring Envoy Gateway, so egtcl will need to support the spec for the `-f` option. When egtl supports installing the data-plane, it will need to parse the yaml file into both EnvoyGateway and EnvoyProxy objects.

docs/latest/design/egctl.md
danehans: Unmanaged means that a user may have Envoy proxies that are not managed by EG. `egctl import` would pull those proxies into EG's management. The details are TBD, I just want to throw out ideas for us to consider during the initial design.

docs/latest/design/egctl.md
danehans: +1 ^

docs/latest/design/egctl.md
danehans: Not a blocker but I prefer `control-plane` and `data-plane` over `envoy-gateway` and `envoy-proxy`.

arkodg: @danehans using actual names (`envoy-gateway`, `envoy-proxy`) allows us to fetch info from specific resources and can be extended to other resources such as `envoy-ratelimit-service` and other auxiliary control planes and data planes

docs/latest/design/egctl.md
danehans: Note that if we intend on egctl eventually installing the data-plane, the `-c` will need to support the EnvoyGateway and EnvoyProxy APIs. For `egctl install -c example.yml`, example.yml will be as follows:
```
kind: EnvoyGateway
provider:
  type: Kubernetes
gateway:
---
kind: EnvoyProxy
metadata:
  name: example-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        replicas: 3
```

__Notes:__
- The example does not specify `control-plane` or `data-plane` so both will be installed.
- If the yaml referenced by `-c` does not specify EnvoyProxy, then the Envoy proxy infra is created with default settings.
- A GatewayClass with a `spec.controllerName` value that matches EG's controllerName should be created when the `control-plane` is installed.
- A Gateway with a `spec.gatewayClassName` value that matches the above GatewayClass should be created when the `data-plane` is installed.

As I mentioned previously, `egctl` can start off by only managing the control-plane. I provide the data-plane examples and details as design considerations.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,67 @@
...
+* `name`: Specifies the name of the specified pod.
```
danehans, line 27: Currently, EG only supports the Kubernetes provider. However, we do intend on supporting other providers, e.g. File. We should consider how egctl will support different providers. For example, when the `--provider` flag is unspecified, egctl will assume the Kubernetes provider. With that said, we should avoid Kube'isms such as `pod`.

danehans: > using actual names (envoy-gateway, envoy-proxy) allows us to fetch info from specific resources

@arkodg can you provide an example to help me understand what you mean? Since egctl should interact with different EG instances, i.e. control/data planes, maybe we should consider supporting a `config` subcommand similar to `kubectl`. A user can set the credentials, EG endpoint, etc. to interact with different EG control/data planes.

arkodg: `egctl config envoy-proxy` gets envoy specific config (LDS, RDS etc)
`egctl config envoy-ratelimit-service` gets ratelimit service config

docs/latest/design/egctl.md
zirain: we can update `import` later.

docs/latest/design/egctl.md
```diff
@@ -16,17 +16,19 @@ This tool is named `egctl`.
 Use the following syntax to run `egctl` commands from your terminal window:
 
 ```console
-egctl [command] [subcommand] [name] [flags]
+egctl [command] [entry] [name] [flags]
```
arkodg, line 19: did you mean `entity` ?

zirain: yes, typo.

arkodg (dismissed):
LGTM ! thanks @zirain

docs/latest/design/egctl.md
arkodg: I think it does make sense here to not explicitly call this `name` of a `pod`, since this project plans on supporting multiple providers, instead call it `name` of the runnable entity (for e.g. pod for the kubernetes provider)

zirain:
cc @danehans PTAL

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+* `command`: Specifies the operation that you want to perform on one or more resources,
+  for example `install`, `uninstall`, `config`, `version`.
+
+* `entity`: Specifies the entity the operation is being performed on such as `envoy-proxy` or `envoy-gateway`.
+
+* `name`: Specifies the name of the specified instance. 
+
+* `flags`: Specifies optional flags. For example, you can use the `-c` or `--config` flags to specify the values for installing.
+
+If you need help, run `kubectl help` from the terminal window.
```
danehans, line 33: s/kubectl/egctl/

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Operation
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation   | Syntax                           | Description                                                                 |
+| ----------- | -------------------------------- | --------------------------------------------------------------------------- |
+| `version`   | `egctl version`                  | Prints out build version information.                                       |
+| `install`   | `egctl install -c CUSTOMPROFILE` | Install or reconfigure EG on a cluster.                                     |
```
danehans, line 42: I believe `-c` should reference a config file in [EnvoyGateway](https://github.com/envoyproxy/gateway/blob/main/api/config/v1alpha1/envoygateway_types.go) format.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+| `uninstall` | `egctl uninstall`                | Uninstall EG from a cluster                                                 |
+| `config`    | `egctl config ENTITY`            | Retrieve information about proxy configuration from envoy proxy and gateway |
+| `log`       | `egctl log ENTITY --level trace` | Change envoy proxy's log level                                              |
```
danehans, line 45: I would expect `log` to also support changing EG's log level. I also prefer `logs` over `log`. Can `log` be used to tail the logs of EG or an Envoy Proxy instance? For example:
```
egctl log envoy-proxy <name> -f
```

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+## Operation
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation   | Syntax                           | Description                                                                 |
+| ----------- | -------------------------------- | --------------------------------------------------------------------------- |
+| `version`   | `egctl version`                  | Prints out build version information.                                       |
+| `install`   | `egctl install -c CUSTOMPROFILE` | Install or reconfigure EG on a cluster.                                     |
+| `uninstall` | `egctl uninstall`                | Uninstall EG from a cluster                                                 |
+| `config`    | `egctl config ENTITY`            | Retrieve information about proxy configuration from envoy proxy and gateway |
```
danehans, line 44: Since `config` supports EG and proxy, s/proxy configuration/configuration/

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+| `log`       | `egctl log ENTITY --level trace` | Change envoy proxy's log level                                              |
+| `analyze`   | `egctl analyze`                  | Analyze configuration and print validation messages                         |
```
danehans, line 46: Does `analyze` only support EG config analysis? If so, please update the description.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Examples
+
+Use the following set of examples to help you familiarize yourself with running the commonly used `egctl` operations:
+
+```console
+# Install EG using the definition in custom profile file
```
danehans, line 53: > custom profile file

A profile is a new concept for EG that is not defined in this design doc. Profiles should be defined in this doc or in a separate design doc and referenced.

docs/latest/design/egctl.md
danehans: Based on the description of `config`, IMO `get` is a better-suited verb. For example:

- To list all running Envoy proxy instances (pods for the Kube provider):  `egctl get envoy-proxy`.
- To list a specific running Envoy proxy instance (pod for the Kube provider):  `egctl get envoy-proxy <name>`.
- To get the runtime config of a specific Envoy Proxy instance (pod for the Kube provider):  `egctl get envoy-proxy <name> --runtime-config`. This is similar to `127.0.0.1:19000/config_dump`.
- To get the bootstrap config of a specific Envoy Proxy instance (pod for the Kube provider):  `egctl get envoy-proxy <name> --bootstrap-config`.

__Note:__ The `--namespace` flag is not specified so egctl will use the default `envoy-gateway-system` namespace.

Additionally, since egctl should support interacting with multiple EGs, `config` should be designated to configure egctl. For example:

- To display egctl's current config context: `egctl config get-contexts`
- To set a context entry in the egctl config file: `egctl config set-context`
- etc.

danehans (changes requested):
@zirain I've been on holiday leave so I apologize for the delay. I appreciate all your hard work on the design and I think it's close to being complete. Please review my review feedback and let me know your thoughts.

docs/latest/design/egctl.md
zirain: Actually, `-c` is alias for `--kubeconfig` in kubernetes, should `EG` keep constaint with others?

docs/latest/design/egctl.md
zirain: `log` only change EG's log level, did `envoy-gateway` support change log level on the fly?

BTW, I donot think we should support `-f`, you can use `kubectl logs` in kubernetes world, we may extend when other provider is supported, but not now.

docs/latest/design/egctl.md
zirain: as you said get seems a little odd, but `get` is not a good choice also.
what about change `config` to `config-dump` or `dump`?

danehans: > as you said get seems a little odd, but get is not a good choice also.
When the description of `config` is `Retrieve information about proxy`, why is `get` not appropriate? Retrieve is a synonym of `get`.

> what about change config to config-dump or dump?
If we proceed with using `config` to configure egctl, i.e. tell the tool what instance of EG to communicate with, then a user may build a mental model around `config` being associated with the egctl config. Additionally, EG has multiple types of config (startup, bootstrap, etc).

docs/latest/design/egctl.md
danehans: > log only change EG's log level, did envoy-gateway support change log level on the fly?

Changing the EG log level will cause a rolling restart.

> BTW, I donot think we should support -f, you can use kubectl logs in kubernetes world, we may extend when other provider is supported, but not now.

I'm fine using kubectl until EG supports other providers.

I have given additional thought to supporting the `log` subcommand. IMHO we should drive all config changes through the EnvoyProxy API. Otherwise, we set a precedent for supporting additional config knobs in the future which increases the support burden. Additionally, having one way to manage config provides a simple mental model for users.

If a user would like to change the log level, we should support:
```
cat <<EOF | egctl apply -f -
kind: EnvoyProxy
metadata:
  name: example-config
spec:
  logging:
    ...
EOF
```

docs/latest/design/egctl.md
danehans: Yes, if kubectl uses `-c` to specify the kubeconfig, then `-c` should be used to specify the egconfig file. The egconfig file will be similar to the kubeconfig, i.e. contain EG endpoints, credentials, etc. Let's pick a different short name for specifying the EnvoyGateway/EnvoyProxy config used with the `install` subcommand.

docs/latest/design/egctl.md
zirain: A rolling restart with control plane(envoy-gateway) is ok in mostly usecases, but it's a little danger on data plane(envoy-proxy), especially TCP.

After you change the config CRD, you can apply changes with `install`.

danehans: I agree that a rolling restart of EG is a non-issue. However, changing Envoy's log level will cause a restart since the Envoy deployment `--log-level` CLI argument will be changed. Can Envoy avoid the restart if the log-level configuration was accomplished without using the CLI argument?

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,73 @@
...
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation   | Syntax                                 | Description                                                                 |
+| ----------- | -------------------------------------- | --------------------------------------------------------------------------- |
+| `version`   | `egctl version`                        | Prints out build version information.                                       |
+| `install`   | `egctl install ENTITY -c EGCONFIGFILE` | Install or reconfigure EG on a cluster.                                     |
+| `uninstall` | `egctl uninstall`                      | Uninstall EG from a cluster                                                 |
+| `get`       | `egctl get ENTITY`                     | Retrieve information about proxy configuration from envoy proxy and gateway |
+| `log`       | `egctl log ENTITY --level trace`       | Change envoy proxy's log level                                              |
```
danehans, line 45: As I mention in other comments, I suggest we remove the `log` subcommand and drive all config management through the EnvoyGateway and EnvoyProxy APIs. See #842 for the initial EnvoyProxy logging config API.

docs/latest/design/egctl.md
zirain: sure, see https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging

docs/latest/design/egctl.md
arkodg: imho `config` seems clearer than the rest, the user will need to specify a specific id/name for the runnable entity any case
`get` (and `set`) is implied in most of the commands outlined in the doc (such as _get_ `version`)

docs/latest/design/egctl.md
zirain: envoy can change log level on the fly, it's very useful in real world.

zirain:
@danehans @arkodg I remove the unclear part(install/uninstall), which will be added back in the future.

After a long think, prefer to use `config` than `get`, PTAL.

danehans:
@zirain I prefer `config` is used to manage the egctl configuration (similar to kubectl config). We can go with `egctl get proxy bootstrap|cluster|route|all <pod-name[.namespace]>` to retrieve the relevant proxy configuration. The same functionality can be supported __in the future__ for the control plane by introducing the `control-plane` subcommand (`cp` for shorthand). 

With the istioctl command structure aside, why do you prefer `egctl proxy-config bootstrap` over `egctl get proxy bootstrap`? I prefer the `egctl get proxy ...` command line structure since it reads naturally, making it easier to remember. However, I’m open to `egctl proxy-config ...`  if you can provide justification on why you prefer `egctl proxy-config bootstrap` over `egctl get proxy bootstrap`.

arkodg:
> @zirain I prefer `config` is used to manage the egctl configuration (similar to kubectl config). We can go with `egctl get proxy bootstrap|cluster|route|all <pod-name[.namespace]>` to retrieve the relevant proxy configuration. The same functionality can be supported **in the future** for the control plane by introducing the `control-plane` subcommand (`cp` for shorthand).
> 
> With the istioctl command structure aside, why do you prefer `egctl proxy-config bootstrap` over `egctl get proxy bootstrap`? I prefer the `egctl get proxy ...` command line structure since it reads naturally, making it easier to remember. However, I’m open to `egctl proxy-config ...` if you can provide justification on why you prefer `egctl proxy-config bootstrap` over `egctl get proxy bootstrap`.

+1 on eliminating `get` since its inferred, and doesnt need to be explicitly typed. I believe @zirain has already changed the design doc from `proxy-config` to `config`

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,61 @@
...
+If you need help, run `egctl help` from the terminal window.
+
+## Operation
+
+The following table includes short descriptions and the general syntax for all the `egctl` operations:
+
+| Operation | Syntax                           | Description                                                                 |
+| --------- | -------------------------------- | --------------------------------------------------------------------------- |
+| `version` | `egctl version`                  | Prints out build version information.                                       |
+| `config`  | `egctl config ENTITY`            | Retrieve information about proxy configuration from envoy proxy and gateway |
```
danehans, line 41: As previously mentioned, `config` should be used to manage the egctl config and not the proxy configuration.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,61 @@
...
+| `log`     | `egctl log ENTITY --level trace` | Change envoy proxy's log level                                              |
```
danehans, line 42: Proxy configuration changes should be accomplished through a configuration manifest and not through an egctl subcommand. For example- `kubectl apply -f envoyproxy.yaml`. Xref https://github.com/envoyproxy/gateway/pull/842 for an example on configuring proxy logging through the EnvoyProxy API.

docs/latest/design/egctl.md
```diff
@@ -0,0 +1,61 @@
...
+| `analyze` | `egctl analyze`                  | Analyze EG configuration and print validation messages                      |
```
danehans, line 43: Can you help me better understand the `analyze` subcommand. It might be better to remove this subcommand for now and have egctl focus on `proxy-config` to retrieve the proxy bootstrap|cluster|route|etc configs and `version`.

danehans (changes requested):

docs/latest/design/egctl.md
arkodg: the PR shares some examples on how `config` is used
```
# Retrieve all information about proxy configuration from envoy
 egctl config envoy-proxy all <instance_name>

 # Retrieve listener information about proxy configuration from envoy 
 egctl config envoy-proxy listener <instance_name>

 # Retrieve information about envoy gateway
 egctl config envoy-gateway
 ```

docs/latest/design/egctl.md
arkodg: this is inspired from istio where `analyze` is used to collect information from multiple sources and surface any runtime issues to the end user
https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/

arkodg (dismissed):

AliceProxy (dismissed):

docs/latest/design/egctl.md
danehans: @arkodg this is where I disagree. The `config` subcommand should be used to configure egctl (similar to `kubectl config`). To manage the proxy config, I prefer `egctl get proxy bootstrap|clusters|routes|etc` but I'm open to `egctl proxy-config bootstrap|clusters|routes|etc`. To support similar functionality for Envoy Gateway in the future, the `egctl gateway-config or control-plane-config` subcommand can be introduced. If you don't agree with this proposal, how do you suggest managing the egctl configuration, i.e. changing between EG endpoints?

arkodg: flags such as `--context` (or `kubeconfig`) can be used to change the server endpoint

docs/latest/design/egctl.md
zirain: will envoy proxy rolling after applying changes to `envoyproxy.yaml`? if so, the `log` will change proxy without a rolling.
Technically, it will port forward admin port and send a post request to [logging endpoint](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging)

danehans: The current implementation uses the `--log-level` CLI argument. Therefore managing logging verbosity through the `log` subcommand or the config file will cause a rolling restart without changing the implementation per https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging.

zirain:
> @zirain I am ready to /approve after the `log` subcommand and all references to it are removed. As I previously mentioned, proxy config changes should be accomplished through the EnvoyProxy API.

done

danehans (approved):
@zirain thanks for all your work on this PR.

Issue #790, opened 2022-12-09
Add Request Authentication Support


danehans:
- [x] Design user-facing API for authentication requests (#336).
- [x] Add API support to Kube provider.
- [x] Add API support to gatewayapi translator.
- [x] Add API support to IR.
- [x] Add API support to xds translator and server.
- [x] Update user docs with config example.
- [ ] E2E testing.

zhshw:
envoy auth filiter need define request body buffer **MaxRequestBytes**

danehans:
@zhshw does a default value exist for MaxRequestBytes? If not, what is an appropriate default value? We should also consider the min/max values too.

arkodg:
@danehans any idea on how E2E will be implemented for this feature ?

danehans:
@arkodg my thoughts are:
- Use the kind cluster from the conformance tests for running e2e testing.
- Mock a JWKS in our e2e testing package.
- Create an HTTP request that contains a JWT.
- Create a test backend app, Gateway, HTTPRoute, AuthenticationFilter, etc.
- Send the request to the Gateway.

danehans:
This feature is shipping in v0.3 with manual testing (see [authn](https://gateway.envoyproxy.io/latest/user/authn.html) user doc). Moving to v0.4.0-rc.1 to add e2e testing.

PR #804, opened 2022-12-13
Adds AuthenticationFilter Support to Resource Translator

danehans:
Adds AuthenticationFilter support to the resource translator.

xref:https://github.com/envoyproxy/gateway/issues/790

### DIFF
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -25,7 +25,7 @@ type HTTPFiltersTranslator interface {
 	processRedirectFilter(filter v1beta1.HTTPRouteFilter, filterContext *HTTPFiltersContext)
 	processRequestHeaderModifierFilter(filter v1beta1.HTTPRouteFilter, filterContext *HTTPFiltersContext)
 	processResponseHeaderModifierFilter(filter v1beta1.HTTPRouteFilter, filterContext *HTTPFiltersContext)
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/translator.go
```diff
@@ -1175,18 +1176,37 @@ func (t *Translator) ProcessHTTPRoutes(httpRoutes []*v1beta1.HTTPRoute, gateways
...
+							}
+						}
+						if !found && valid {
+							errMsg = fmt.Sprintf("Reference not found for filter type: %s", filter.Type)
+						}
+						if !valid || !found {
+							parentRef.SetCondition(httpRoute,
+								v1beta1.RouteConditionAccepted,
+								metav1.ConditionFalse,
+								v1beta1.RouteReasonUnsupportedValue,
```
skriss, line 1203: It's not 100% clear in the Gateway API spec, but I think if the filter ref can't be resolved (i.e. the referent doesn't exist), then setting a `ResolvedRefs: false` condition would be more appropriate than `Accepted: false (UnsupportedValue)`.

skriss: (For unknown filter types, what we're doing here still makes sense though)

danehans: @skriss thanks for the review. Commit 07ec1b4 resolves your above comment.

arkodg (approved):

Issue #736, opened 2022-11-17
Refactory: gateway api translator duplication


zhaohuabing:
The gateway API translator has grown too big(nearly 2000 lines now) and has a large amount of duplicated code in it, I'd like to clear it to make it easier to read and maintain.

zhaohuabing:
I would like to take this one.

danehans:
@zhaohuabing Thanks for taking this on. For planning purposes, when do you think you'll be ready to submit a PR?

zhaohuabing:
@danehans I can start to work on this after https://github.com/envoyproxy/gateway/pull/728 has been merged. If everything goes well, I think I can submit a PR for review next week.

danehans:
Sounds great. Thanks for the update.

PR #807, opened 2022-12-14
refactory gateway api translator

zhaohuabing:
extract methods to remove duplication.

https://github.com/envoyproxy/gateway/issues/736

### DIFF
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -693,88 +693,10 @@ func buildRuleRouteDest(backendRef v1beta1.HTTPBackendRef,
 		weight = uint32(*backendRef.Weight)
 	}
 
-	if backendRef.Group != nil && *backendRef.Group != "" {
-		parentRef.SetCondition(httpRoute,
-			v1beta1.RouteConditionResolvedRefs,
-			metav1.ConditionFalse,
-			v1beta1.RouteReasonInvalidKind,
-			"Group is invalid, only the core API group (specified by omitting the group field or setting it to an empty string) is supported",
-		)
-		return nil, weight
-	}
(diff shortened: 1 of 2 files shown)


danehans:
@zhaohuabing thanks for the PR. Can you fix the linter error when you have a moment?

zhaohuabing:
> @zhaohuabing thanks for the PR. Can you fix the linter error when you have a moment?

Sure. It's been fixed.

internal/gatewayapi/translator.go
```diff
@@ -1721,91 +1565,11 @@ func (t *Translator) ProcessUDPRoutes(udpRoutes []*v1alpha2.UDPRoute, gateways [
...
-			if !portFound {
-				parentRef.SetCondition(udpRoute,
-					v1beta1.RouteConditionResolvedRefs,
-					metav1.ConditionFalse,
-					"PortNotFound",
-					fmt.Sprintf("UDP Port %d not found on service %s/%s", *backendRef.Port, serviceNamespace,
-						string(backendRef.Name)),
-				)
+			if !checkBackendRef(&backendRef, parentRef, udpRoute, resources, serviceNamespace, KindUDPRoute) {
 				continue
```
zirain, line 1573: need a warn log, a metric will be better

zhaohuabing: A condition will be set to the route if the backendRef can't pass validation.  Furthermore, if we want to add log to translator, we need to do it in a consistent way, which means we may want to add log to all the checks, not only here. It would be better in another PR if we think this is necessary.

zirain: reasonable, got it.

danehans (approved):

PR #813, opened 2022-12-15
Adds AuthenticationFilter Support to IR.

danehans:
Adds AuthenticationFilter support to IR:
- `api/v1alpha1/validation/authenticationfilter.go`: Adds validation package for validating the AuthenticationFilter type.
- `internal/ir/xds.go`: Adds internal Request Authentication types with initial support for JWT.

Requires #804

xref: #790


api/v1alpha1/authenticationfilter_types.go
```diff
@@ -9,6 +9,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// AuthenticationFilterKind is the name of the AuthenticationFilter kind.
+	AuthenticationFilterKind = "AuthenticationFilter"
```
chauhanshubham, line 14: nit: I think the general format is Kind...
and in our code as well https://github.com/envoyproxy/gateway/blob/46b7b7523cc8d8167fd999c9cdecdca2ffb9fd71/internal/gatewayapi/translator.go#L24

api/v1alpha1/validation/authenticationfilter.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// ValidateJwtProvider validates the provided JWT authentication filter provider.
+func ValidateJwtProvider(jwt *egv1a1.JwtAuthenticationFilterProvider) error {
+	var errs []error
+
+	switch {
+	case len(jwt.Name) == 0:
+		errs = append(errs, errors.New("name must be set for jwt provider"))
+	case len(jwt.RemoteJWKS.URI) == 0:
```
chauhanshubham, line 67: should we add validation logic for `URI` format as well?

internal/provider/kubernetes/routes.go
```diff
@@ -110,6 +113,28 @@ func (r *gatewayAPIReconciler) processHTTPRoutes(ctx context.Context, gatewayNam
...
+
+			for i := range rule.Filters {
+				filter := rule.Filters[i]
+				if err := gatewayapi.ValidateHTTPRouteFilter(&filter); err != nil {
+					r.log.Error(err, "bypassing filter rule", "index", i)
+					continue
+				}
+
+				if filter.Type == gwapiv1b1.HTTPRouteFilterExtensionRef {
+					authenFilter, err := r.getAuthenticationFilter(ctx, httpRoute.Namespace, string(filter.ExtensionRef.Name))
```
chauhanshubham, line 125: It might be so, that this call is triggered multiple times for the same authentication filter, as multiple httproutes could use the same authn filter.

If we just store authentication filter NamespacedName as key in resourceMap then after processing all httproutes we can do the `getAuthenticationFilter` call for unique filters.

internal/gatewayapi/testdata/httproute-with-valid-authenfilter.in.yaml
```diff
@@ -0,0 +1,55 @@
...
+      backendRefs:
+      - name: service-1
+        port: 8080
+      filters:
+      - type: ExtensionRef
+        extensionRef:
+          group: gateway.envoyproxy.io
+          kind: AuthenticationFilter
+          name: test
+AuthenFilters:
```
chauhanshubham, line 43: nit; wondering why `AuthenFilters` is supposed to start with uppercase char.

danehans: Commit [07e3398](https://github.com/envoyproxy/gateway/pull/813/commits/07e339888bc4aac8dfc1497d4dddca8b6902d934) resolves this comment.

internal/provider/kubernetes/routes.go
danehans: Commit [07e3398](https://github.com/envoyproxy/gateway/pull/813/commits/07e339888bc4aac8dfc1497d4dddca8b6902d934) resolves this comment. I took a slightly different approach than what you suggested, PTAL.

api/v1alpha1/validation/authenticationfilter.go
danehans: Commit [07e3398](https://github.com/envoyproxy/gateway/pull/813/commits/07e339888bc4aac8dfc1497d4dddca8b6902d934) resolves this comment.

api/v1alpha1/authenticationfilter_types.go
danehans: Commit [07e3398](https://github.com/envoyproxy/gateway/pull/813/commits/07e339888bc4aac8dfc1497d4dddca8b6902d934) resolves this comment.

internal/ir/xds.go
```diff
@@ -125,6 +141,63 @@ type HTTPListener struct {
 	Routes []*HTTPRoute
 	// IsHTTP2 is set if the upstream client as well as the downstream server are configured to serve HTTP2 traffic.
 	IsHTTP2 bool
+	// RequestAuthentication defines the schema for authenticating HTTP requests.
+	RequestAuthentication *RequestAuthentication
```
arkodg, line 145: shouldnt this be per HTTPRoute instead of per HTTPListener

danehans: JWT is configured as an Envoy listener filter, specifically `envoy.filters.http.jwt_authn`. The request authen design [doc](https://github.com/envoyproxy/gateway/blob/main/docs/latest/design/request-authentication.md) provides examples on how the AuthentictionFilter<>HTTPRoutre resource get translated into an xDS config. The AuthenticationFilter spec gets translated into [jwt requirements](https://github.com/envoyproxy/gateway/pull/813/files#diff-b0330312d3f47f14bcdc8e1b589d84001a1969564c86ff59347bd56795db7ccaR177) and the `match` of the HTTPRoute rule with a `filter` of type AuthenticationFilter gets translated into [jwt matches](https://github.com/envoyproxy/gateway/pull/813/files#diff-b0330312d3f47f14bcdc8e1b589d84001a1969564c86ff59347bd56795db7ccaR175). The [JWT filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extension-envoy-filters-http-jwt-authn) processes the request before routing.

arkodg: would recommend looking into the route config here https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-route which allows you to the same thing using the `typed_per_filter_config` field within `Route`. This eliminates the need to duplicate the info in `httproute.spec.rules[].matches[]` since it is already applicable to the route

danehans: @arkodg thanks for the review and pointer. Commit cf98e95 updates RequestAuthentication to be a field of HTTPRoute instead of HTTPListener.

arkodg: thanks for incorporating this, this will continue to keep the mapping of GAPI resources to xDS resources easy to maintain

internal/provider/kubernetes/routes.go
```diff
@@ -152,6 +151,17 @@ func (r *gatewayAPIReconciler) processGRPCRoutes(ctx context.Context, gatewayNam
...
 	resourceMap *resourceMappings, resourceTree *gatewayapi.Resources) error {
 	httpRouteList := &gwapiv1b1.HTTPRouteList{}
+
+	// An HTTPRoute may reference an AuthenticationFilter, so add them to the resource map first (if they exist).
+	filters, err := r.getAuthenticationFilters(ctx)
+	if err != nil {
+		return err
+	}
+	for i := range filters {
+		filter := filters[i]
```
arkodg, line 161: curious why this step is needed

internal/ir/xds.go
```diff
@@ -7,9 +7,20 @@ package ir
...
 	"net"
 
 	"github.com/tetratelabs/multierror"
+
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
+	"github.com/envoyproxy/gateway/api/v1alpha1/validation"
+)
+
+const (
+	maxJwtRules          = 4
```
arkodg, line 20: if these are user facing limits, shouldnt they be enforced in the provider or resource translator instead ?

internal/provider/kubernetes/routes.go
danehans: Do to implicit memory aliasing in for loop:
```
$ make lint
===========> Running go.mod.lint ...
===========> Go module looks clean!
===========> Running lint.golint ...
tools/bin/golangci-lint run  --build-tags=e2e --config=tools/linter/golangci-lint/.golangci.yml
WARN [runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref.
internal/provider/kubernetes/routes.go:161:50: G601: Implicit memory aliasing in for loop. (gosec)
		resourceMap.authenFilters[utils.NamespacedName(&filter)] = &filter
		                                               ^
make[1]: *** [lint.golint] Error 1
```

internal/ir/xds.go
danehans: @arkodg these limits, e.g. validation rules, are specified in the [AuthenticationFilter API](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/authenticationfilter_types.go) and will be enforced by the kube-api server. The validation rules are added to the xds IR for consistency and additional protection. I can remove them if you have a strong opinion against this approach.

arkodg (dismissed):

danehans:
@arkodg since this PR depends on and includes https://github.com/envoyproxy/gateway/pull/804, can you /approve that PR as well?

arkodg (approved):

PR #816, opened 2022-12-16
bump gateway api to v0.6.0-rc2

arkodg:
Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #791, opened 2022-12-09
Adds AuthenticationFilter Support to Kubernetes Provider

danehans:
Adds AuthenticationFilter API support to the Kubernetes provider:
- `api/v1alpha1/authenticationfilter_types.go`: Adds missing Authentication -> AuthenticationFilter changes from #733.
- `api/v1alpha1/zz_generated.deepcopy.go`: Regenerates DeepCopy methods due to AuthenticationFilter API changes.
- `internal/envoygateway/scheme.go`: Adds the `v1alpha1`, e.g. AuthenticationFilter, group to the scheme.
- `internal/gatewayapi/helpers.go`: Adds `ValidateAuthenticationFilterRef()` helper function to validate AuthenticationFilter references from an HTTPRoute.
- `internal/gatewayapi/helpers_test.go`: Adds `ValidateAuthenticationFilterRef()` unit tests.
- `internal/gatewayapi/translator.go`:  Adds AuthenticationFilter to the gatewayapi Resources type. Adds `NewResources()` helper function.
- `internal/gatewayapi/zz_generated.deepcopy.go`: Regenerates DeepCopy methods due to API changes.
- `internal/provider/kubernetes/config/crd/kustomization.yaml`: Adds AuthenticationFilter CRD to the list of installed CRDs.
- `internal/provider/kubernetes/config/rbac/role.yaml`: Generated by `make manifests`, updating EG RBAC for AuthenticationFilter.
- `internal/provider/kubernetes/controller.go`: Add a watch for instances of AuthenticationFilter. Updates the resourceMappings type to map HTTPRoutes<>AuthenticationFilters. Adds AuthenticationFilters referenced by managed HTTPRoutes to the resourceTree.
- `internal/provider/kubernetes/rbac.go`: Updates EG RBAC kubebuilder tags for managing instances of AuthenticationFilter.
- ~`internal/provider/kubernetes/helpers.go`: Removes namespace check from `validateBackendRef()`. Now that #702 adds full ReferenceGrant support, an HTTPRoute backend should be allowed to reference a Service in a different namespace if the necessary ReferenceGrant exists ([xref](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/routes.go#L101-L111)).~
- `internal/provider/kubernetes/kubernetes_test.go`: Adds AuthenticationFilter test case to Kube provider integration tests.
- `internal/provider/kubernetes/predicates.go` Adds predicate to filter AuthenticationFilter instancess from the controller's watch.
[...]

PR #821, opened 2022-12-19
Fix package import for HTTPRouteFilterExtensionRef

arkodg:
Caused by merging https://github.com/envoyproxy/gateway/pull/791 without rebasing https://github.com/envoyproxy/gateway/pull/816


arkodg:
looks like the tests also need to be updated to use the indexes in the tests

danehans (approved):
Thanks for catching and fixing this issue @arkodg.

PR #842, opened 2023-01-02
Adds Logging API Type to EnvoyProxy

danehans:
Adds Logging API type to support logging configuration of managed proxies.

Supports #797 


api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -74,6 +80,31 @@ type EnvoyDeployment struct {
...
+}
+
+// LogLevel defines a log level for system logs.
+//
+// +kubebuilder:validation:Enum=Debug;Info;Error
+type LogLevel string
+
+const (
+	// LogLevelDebug defines the "Debug" logging level.
+	LogLevelDebug LogLevel = "Debug"
```
zirain, line 99: `debug` is not enough for envoy in real world.

also, need support `http:trace,ratelimit:debug` likely.

danehans: @zirain I agree that we will need to support additional log levels. This PR is meant to be small to __introduce__ the logging config API. We can iterate on the API, including verbosity levels, in future PRs.

zirain: at least the API should support different loglevels on different components/modules

danehans: @zirain commit [8b75fed](https://github.com/envoyproxy/gateway/pull/842/commits/8b75fedcd36e39bf04b8b95bbedd3fdfddffcc34) adds component-level logging. I did not include every component since this PR is meant to introduce logging and additional components can be added in the future, PTAL.

zirain (approved):
lgtm, it's a good start.

Issue #839, opened 2022-12-30
Bump Kube Provider Test Data Manifests to v0.6.0


danehans:
Currently, the Kube provider [test data manifests](https://github.com/envoyproxy/gateway/tree/main/internal/provider/kubernetes/testdata/in) are based on GAPI v0.6.0-rc1. They should be updated to v0.6.0 when https://github.com/envoyproxy/gateway/pull/828 is merged.

Issue #716, opened 2022-11-09
Update to the v0.6.0 Gateway API release 


arkodg:

arkodg:
currently blocked on the Gateway API release itself

Xunzhuo:
I am assigning this to bump GWAPI to v0.6.0, since some tasks of extended filters are not covered in v0.5.1.

arkodg:
this is blocking https://github.com/envoyproxy/gateway/pull/717

arkodg:
hey @Xunzhuo looks like https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.6.0-rc1 is released, this tag should unblock us

Xunzhuo:
Let us open this issue for a while, need to bump when v0.6.0 released.

arkodg:
update currently blocked due to the failure of these tests
https://github.com/envoyproxy/gateway/pull/828#issuecomment-1361668873
needs to be triaged

PR #853, opened 2023-01-04
feat: bump gwapi to v0.6.0

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/839
Fixes: https://github.com/envoyproxy/gateway/issues/716

Closes: https://github.com/envoyproxy/gateway/pull/828


internal/provider/kubernetes/controller.go
```diff
@@ -495,6 +495,14 @@ func (r *gatewayAPIReconciler) statusUpdateForGateway(gtw *gwapiv1b1.Gateway, sv
 			}
 			gCopy := g.DeepCopy()
 			gCopy.Status.Conditions = status.MergeConditions(gCopy.Status.Conditions, gtw.Status.Conditions...)
+			for index := range gCopy.Status.Conditions {
```
arkodg, line 498: thanks for fixing the issue @Xunzhuo !
I have two questions
* why are we updating all the conditions here ?
isn't this already done correctly in https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/status/conditions.go#L116
* shouldn't this logic be living in https://github.com/envoyproxy/gateway/blob/main/internal/status/conditions.go
cc @skriss  who was also looking into this

arkodg:
looks like the core issue is interpretation of mapping of Observed Generation to Status Conditions
The new GAPI tests assert that the new Observed Generation value be asserted on all Status Conditions (https://github.com/kubernetes-sigs/gateway-api/pull/1586) but Envoy Gateway does not update the status conditions for all older conditions if the core values have not changed https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/status/conditions.go#L127 . Imho the latter approach seems more logical. Is there any spec governing this mapping @youngnick

sunjayBhatia:
> looks like the core issue is interpretation of mapping of Observed Generation to Status Conditions The new GAPI tests assert that the new Observed Generation value be asserted on all Status Conditions ([kubernetes-sigs/gateway-api#1586](https://github.com/kubernetes-sigs/gateway-api/pull/1586)) but Envoy Gateway does not update the status conditions for all older conditions if the core fields (such as `Status`, `Reason` and `Message`) have not changed
> 
> https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/status/conditions.go#L127
> 
> and
> https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/gatewayapi/contexts.go#L484
> 
> . Imho the latter approach seems more logical. Is there any spec governing this mapping @youngnick @skriss

from the Condition spec: https://github.com/kubernetes/apimachinery/blob/6c409361e35e40e38c4056ba0b86647d4244c047/pkg/apis/meta/v1/types.go#L1480-L1485

arkodg:
thanks for sharing that @sunjayBhatia ! , the spec makes it clear that the condition is considered out of date if the observed generation is older
hey @Xunzhuo if you can enhance these snippets (https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/status/conditions.go#L127 and https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/gatewayapi/contexts.go#L484) with an extra check to consider a condition unchanged or changed if the observed generation is also unchanged or changed respectively, should fix the issue, thanks !

internal/provider/kubernetes/controller.go
arkodg: please ignore this in favor of https://github.com/envoyproxy/gateway/pull/853#issuecomment-1371423935

Xunzhuo:
> even if the type/reason/status haven't changed, the observed generation should be updated to make clear the status condition applies to the latest version of the resource
From this comment, do we still need to add an extra check to consider a condition unchanged or changed if the observed generation is also unchanged or changed respectively ?

Xunzhuo:
> thanks for sharing that @sunjayBhatia ! , the spec makes it clear that the condition is considered out of date if the observed generation is older hey @Xunzhuo if you can enhance these snippets (
> 
> https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/status/conditions.go#L127
> 
> and
> https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/gatewayapi/contexts.go#L484
> 
> ) with an extra check to consider a condition unchanged or changed if the observed generation is also unchanged or changed respectively, should fix the issue, thanks !

Done.

internal/gatewayapi/contexts.go
```diff
@@ -494,7 +494,7 @@ func (r *RouteParentContext) SetCondition(route RouteContext, conditionType v1be
 	idx := -1
 	routeStatus := route.GetRouteStatus()
 	for i, existing := range routeStatus.Parents[r.routeParentStatusIdx].Conditions {
-		if existing.Type == cond.Type {
+		if existing.Type == cond.Type && existing.ObservedGeneration == cond.ObservedGeneration {
```
arkodg, line 497: lines 499-501 need to be changed to
```
if existing.Status == cond.Status &&
 				existing.Reason == cond.Reason &&
				existing.Message == cond.Message &&
				existing.ObservedGeneration == cond.ObservedGeneration {
```

internal/provider/kubernetes/controller.go
```diff
@@ -495,6 +495,14 @@ func (r *gatewayAPIReconciler) statusUpdateForGateway(gtw *gwapiv1b1.Gateway, sv
...
+				gCopy.Status.Conditions[index].ObservedGeneration = gtw.Generation
```
arkodg, line 499: this loop needs to be removed, this should happen automatically

internal/provider/kubernetes/controller.go
```diff
@@ -894,6 +902,15 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
 						}
 						gCopy := g.DeepCopy()
 						gCopy.Status.Listeners = val.Status.Listeners
+
+						for index := range gCopy.Status.Conditions {
```
arkodg, line 906: this loop needs to be removed as well

internal/provider/kubernetes/controller.go
```diff
@@ -922,6 +939,12 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
 						}
 						hCopy := h.DeepCopy()
 						hCopy.Status.Parents = val.Status.Parents
+
+						for parentIndex, parent := range hCopy.Status.Parents {
```
arkodg, line 943: same as above

arkodg:
@Xunzhuo you will also need to edit this code to
```
func conditionChanged(a, b metav1.Condition) bool {
	return a.Status != b.Status || a.Reason != b.Reason || a.Message != b.Message || || a.ObservedGeneration != b.ObservedGeneration
}
```

arkodg:
@Xunzhuo you will also need to rm `ObservedGeneration` from here https://github.com/envoyproxy/gateway/blob/97032bee2db5ec73027b172e24b03d4dfe98ac3a/internal/status/status.go#L164 since it is now a comparable field

internal/provider/kubernetes/controller.go
Xunzhuo: After removing them, CI fails : (

arkodg:
hey @Xunzhuo looks like there was an issue where the gateway API status updater code path would try and update `Status.Listeners` without updating the ObservedGeneraton with `Status.Conditions`  causing the test to fail. Ive updated the logic here https://github.com/envoyproxy/gateway/pull/874

PR #858, opened 2023-01-04
Add details on Global Rate Limit implementation

arkodg:

### DIFF
--- a/docs/latest/design/ratelimit.md
+++ b/docs/latest/design/ratelimit.md
@@ -260,6 +260,29 @@ The two levels of selectors/matches allow for flexibility and aim to hold match 
 of each configuration to be different. It also allows the `clientSelectors` field within the RateLimitFilter to be enhanced with other matchable
 attribute such as [IP subnet][] in the future that are not relevant in the [HTTPRoute][] API.
 
+## Implementation Details
+
+### Global Rate limiting
+
+*  [Global rate limiting][] in Envoy Proxy can be achieved using the following -
+  * [Actions][] can be conifgured per [xDS Route][].
+  * If the match criteria defined within these actions is met for a specific HTTP Request, a set of key value pairs called [descriptors][]
+  defined within the above actions is sent to a remote [rate limit service][], whose configuration (such as the URL for the rate limit service) is defined 
+  using a [rate limit filter][].
+  * Based on information received by the rate limit service and its programmed configuration, a decision is computed, whether to rate limit
+  the HTTP Request or not, and is sent back to Envoy, which enforces this decision on the data plane.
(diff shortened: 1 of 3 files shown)


docs/latest/design/ratelimit.md
```diff
@@ -260,10 +260,42 @@ The two levels of selectors/matches allow for flexibility and aim to hold match
...
+  * If the match criteria defined within these actions is met for a specific HTTP Request, a set of key value pairs called [descriptors][]
+  defined within the above actions is sent to a remote [rate limit service][], whose configuration (such as the URL for the rate limit service) is defined 
+  using a [rate limit filter][].
+  * Based on information received by the rate limit service and its programmed configuration, a decision is computed, whether to rate limit
+  the HTTP Request or not, and is sent back to Envoy, which enforces this decision on the data plane.
+* Envoy Gateway will leverage this Envoy Proxy feature by -
+  * Translating the user facing RateLimitFilter API into Rate limit [Actions][] as well as Rate limit service configuration to implement
+  the desired API intent.
+  * Envoy Gateway will use the existing [reference implementation][] of the rate limit service.
+    * The Infrastructure administrator will need to enable the rate limit service using new settings that will be defined in the [EnvoyGateway][] config API.
```
danehans, line 278: Why use EnvoyGateway instead of the EnvoyProxy API? A couple of points of note:

- Making a change to EnvoyGateway will cause an Envoy Gateway restart.
- EnvoyProxy was designed to manage the data plane and auxiliary control planes ([xref](https://github.com/envoyproxy/gateway/blob/main/api/config/v1alpha1/envoyproxy_types.go#L40-L60)). If the auxiliary control plane infra is managed through the EnvoyGateway API, the godocs for EnvoyProxy should be updated.

arkodg: thanks for raising these points !
so cons of adding it in `EnvoyProxy`-
* the name of the API makes it sound very specific to the envoy proxy or related data plane infra. The rate limit service is a control plane entity

the cons of adding it in `EnvoyGateway` -
* Will require an edit of the Envoy Gateway which will require a restart of envoy gateway
raised https://github.com/envoyproxy/gateway/issues/864 to track this issue, but as we are starting off, I wouldn't consider this as a blocker 

the pros of adding it in `Envoygateway` - 
* keeping control plane config in `EnvoyGateway` and data plane config `EnvoyProxy` sounds more consistent, which is why **im leaning towards this option**

arkodg: ive update the EnvoyProxy docs in this PR, ptal

danehans (approved):

PR #859, opened 2023-01-04
add RouteStatus API to RouteContext interface

arkodg:
Deduplicates the code associated with route status conditions


arkodg:
> Should unit tests be added?

this code should be covered by the [top level translator test](https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/gatewayapi/translator_test.go#L30) which checks status as well

danehans (approved):

PR #861, opened 2023-01-05
Adds EnvoyProxy Support to Kubernetes Provider

danehans:
Adds EnvoyProxy support to the Kubernetes provider.
- Updates status pkg to support GatewayClass [InvalidParamtersRef](https://github.com/kubernetes-sigs/gateway-api/blob/v0.6.0/apis/v1beta1/gatewayclass_types.go#L172) status condition.
- Updates the Kube controller to watch for EnvoyProxy API objects and trigger GatewayClass reconciliation. This supports processing the paramtersRef and updating GatewayClass status conditions.
- Adds unit and integration tests.

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -26,6 +26,8 @@ import (
 	gwapiv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
+	egcfgv1a1 "github.com/envoyproxy/gateway/api/config/v1alpha1"
+	"github.com/envoyproxy/gateway/api/config/v1alpha1/validation"
 	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/gatewayapi"
@@ -150,7 +152,8 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// Update status for all gateway classes
 	for _, gc := range cc.notAcceptedClasses() {
-		if err := r.gatewayClassUpdater(ctx, gc, false); err != nil {
+		if err := r.gatewayClassUpdater(ctx, gc, false, string(status.ReasonOlderGatewayClassExists),
+			status.MsgOlderGatewayClassExists); err != nil {
 			r.resources.GatewayAPIResources.Delete(acceptedGC.Name)
 			return reconcile.Result{}, err
 		}
@@ -201,7 +204,18 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 		resourceTree.Namespaces = append(resourceTree.Namespaces, namespace)
 	}
 
-	if err := r.gatewayClassUpdater(ctx, acceptedGC, true); err != nil {
+	// Process the parametersRef of the accepted GatewayClass.
+	if acceptedGC.Spec.ParametersRef != nil && acceptedGC.DeletionTimestamp == nil {
+		if err := r.processParamsRef(ctx, acceptedGC, resourceTree); err != nil {
+			msg := fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err)
+			if err := r.gatewayClassUpdater(ctx, acceptedGC, false, string(gwapiv1b1.GatewayClassReasonInvalidParameters), msg); err != nil {
+				r.log.Error(err, "unable to update GatewayClass status")
+			}
+			r.log.Error(err, "failed to process parametersRef for gatewayclass", "name", acceptedGC.Name)
+		}
+	}
+
+	if err := r.gatewayClassUpdater(ctx, acceptedGC, true, string(gwapiv1b1.GatewayClassReasonAccepted), string(status.MsgValidGatewayClass)); err != nil {
 		r.log.Error(err, "unable to update GatewayClass status")
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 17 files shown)


internal/provider/kubernetes/controller.go
```diff
@@ -1014,3 +1073,63 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
...
+
+	if !valid {
+		return fmt.Errorf("invalid gatewayclass %s: %v", gc.Name, validationErr)
+	}
+
+	return nil
+}
+
+// classRefsEnvoyProxy returns true if the provided GatewayClass references the provided EnvoyProxy.
+func (r *gatewayAPIReconciler) classRefsEnvoyProxy(gc *gwapiv1b1.GatewayClass, ep *egcfgv1a1.EnvoyProxy) bool {
```
zirain, line 1127: It doesn't seem necessary to be a method of `gatewayAPIReconciler`

internal/provider/kubernetes/controller.go
```diff
@@ -196,7 +207,41 @@ func newGatewayAPIController(mgr manager.Manager, cfg *config.Server, su status.
 	return nil
 }
 
+func (r *gatewayAPIReconciler) enqueueManagedClass(obj client.Object) []reconcile.Request {
+	ep, ok := obj.(*egcfgv1a1.EnvoyProxy)
+	if !ok {
+		panic(fmt.Sprintf("unsupported object type %T", obj))
+	}
+
+	// The EnvoyProxy must be in the same namespace as EG.
```
zirain, line 216: it's a bit confusing here
`EnvoyProxy` is namespaced resource, but required to same namespace as EG, then namespace is ignored after enqueued.

danehans:
@zirain thanks for the review. Commit [2e42dcd](https://github.com/envoyproxy/gateway/pull/861/commits/2e42dcd9fe38ca6230906705d81ef38f06953959) resolves your comments, PTAL.

internal/provider/kubernetes/controller.go
```diff
@@ -196,7 +207,45 @@ func newGatewayAPIController(mgr manager.Manager, cfg *config.Server, su status.
 	return nil
 }
 
+func (r *gatewayAPIReconciler) enqueueManagedClass(obj client.Object) []reconcile.Request {
```
arkodg, line 210: curious why this function Is needed instead of just capturing the gateway class and the envoy proxy associated with it here https://github.com/envoyproxy/gateway/blob/50e0bcb936884195ec1ce9d03bd2503bcbad7c7a/internal/provider/kubernetes/controller.go#L250

internal/provider/kubernetes/controller.go
```diff
@@ -280,7 +330,7 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 
 	// Update status for all gateway classes
 	for _, gc := range cc.notAcceptedClasses() {
-		if err := updater(gc, false); err != nil {
+		if err := updater(gc, false, string(status.ReasonOlderGatewayClassExists), status.MsgOlderGatewayClassExists); err != nil {
```
arkodg, line 333: the status says `ReasonOlderGatewayClassExists`, how are we so sure that these gateway classes were not accepted because they were younger than the accepted gateway class, maybe they are linked to a different controller ?

internal/provider/kubernetes/controller.go
```diff
@@ -426,7 +476,20 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
...
 	}
 
-	if err := updater(acceptedGC, true); err != nil {
+	// Process the parametersRef of the accepted GatewayClass.
+	if acceptedGC.Spec.ParametersRef != nil && acceptedGC.DeletionTimestamp == nil {
+		if err := r.processParamsRef(ctx, acceptedGC, resourceMap, resourceTree); err != nil {
+			msg := fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err)
+			if err := updater(acceptedGC, false, string(gwapiv1b1.GatewayClassReasonInvalidParameters), msg); err != nil {
+				r.log.Error(err, "unable to update GatewayClass status")
+				return reconcile.Result{}, err
```
arkodg, line 485: we can't return early, we need to process other resources

internal/provider/kubernetes/controller.go
```diff
@@ -426,7 +476,20 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
...
+			}
+			r.log.Info("failed to process parametersRef for gatewayclass", "name", acceptedGC.Name, "error", err)
+			return reconcile.Result{}, nil
```
arkodg, line 488: same as above, we can't return early, instead the log should be an error log, not a info log

Xunzhuo (commented):
Mostly LGTM, but can this be merged after [TCPRoute Implementation](https://github.com/envoyproxy/gateway/pull/737) get merged? Or there will be more conflicts I need to resolve.

danehans:
@arkodg @Xunzhuo thanks for the review. I've been busy with other responsibilities over the past several days. I'll respond and resolve review comments shortly.

internal/provider/kubernetes/controller.go
danehans: `cc` and `acceptedGC` are created during a `Reconcile()`. This method creates the reconcile request for the GC that ref's the watched EnvoyProxy. EnvoyProxy does not bind/ref other resources, e.g. parentRef, so the method must list all GCs to see which one references the watched EnvoyProxy. This is similar to listing all HTTPRoutes for a Gateway CRUD to understand which routes are used to calculate listener conditions.

internal/provider/kubernetes/controller.go
danehans: `notAcceptedClasses()` iterates through all __matched__ GCs and only returns GCs that are not the oldest, e.g. managed, GC.

internal/provider/kubernetes/controller.go
danehans: Ack

internal/provider/kubernetes/controller.go
danehans: Ack

danehans:
@arkodg I commented/resolved your review feedback, PTAL.

arkodg (approved):

zirain (approved):

Xunzhuo (approved):

Issue #836, opened 2022-12-27
support websocket 


qicz:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Xunzhuo:
Plz provide more infos, then we can have a deep understanding on the background of this issue.

qicz:
the current eg version cannot support WebSocket requests that are missing upgrade configs for this. just this. @Xunzhuo

danehans:
Moving to the backlog since https://github.com/kubernetes-sigs/gateway-api/issues/205 must be resolved before this feature can be implemented.

zhshw:
@danehans  HTTPRoute add annations ,like this can imp websocket、timeout、retry......
`
 route.Annotations["envoy-gateway/upgrade-websocket"]
`

routeAction setting UpgradeConfig
`
if isWebsocket {
  routeAction.UpgradeConfigs = []*route.RouteAction_UpgradeConfig{{
	  UpgradeType: "websocket",
  }}
}

`

arkodg:
@zhshw annotations lack structure needed to correctly validate intent, we prefer adding support using explicit fields instead

zhshw:
@arkodg 
Explicit fields may not be necessary, but annotations are the simplest and most practical way to be compatible with HTTPRoute.

Backendref serivce can be an associated Envoy route kind. Use the Envoy xds route field as a richer configuration, rather than losing a large number of useful configurations for the translator

luofish:
@arkodg  it's a feature request, maybe we can use anno instead before explicit fields be designed

crain-cn:
@arkodg it's a feature request, maybe we can use anno instead before explicit fields be designed

arkodg:
it looks like envoy allows web socket upgrade https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/websocket which should be easy to achieve using this config in HCM
```          
upgrade_configs:
          - upgrade_type: websocket
```

and it looks like istio enables this by default, so leaning towards enabling this by default for HTTP 1.1 (HTTPRoute), upgrades are disallowed for HTTP 2.0 (GRPCRoute)

zhshw:
@arkodg This is a crude way to violate the semantics that it is an optional value. Most projects are set up on demand。
This default value, istio is not worth learning

PR #867, opened 2023-01-06
Allow upgrade to websocket for HTTP 1.1 listeners

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/836

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -92,6 +92,17 @@ func addXdsHTTPFilterChain(xdsListener *listener.Listener, irListener *ir.HTTPLi
 		}},
 	}
 
+	// Allow websocket upgrades for HTTP 1.1
+	// Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
+	if !irListener.IsHTTP2 {
+		mgr.UpgradeConfigs = []*hcm.HttpConnectionManager_UpgradeConfig{
+			{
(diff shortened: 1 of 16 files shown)


arkodg:
@skriss @AliceProxy does Contour and Emissary have this upgrade enabled by default as well, similar to Istio ?

Xunzhuo (commented):
Or should we use configuration to determine if we use websocket or not?

arkodg:
> Or should we use configuration to determine if we use websocket or not?

@Xunzhuo can you share any reason why a user might have web sockets enabled in the app backend but would like to force disable it in the proxy ?
imho this commit enables web socket upgrade proxying based on client-server web socket capability, and enabling it by default is helping the HTTPRoute owner (app developer) set one less thing to get their apps connected to the outside world.

Xunzhuo (dismissed):
Sorry maybe I overthink it, LGTM, thanks!

AliceProxy:
Emissary will not upgrade to websockets unless explicitly configured by the user

skriss:
Contour's is also opt-in on a per-route basis.

arkodg:
thanks @AliceProxy and @skriss for sharing Emissary's and Contour's implementation. 
I proposed enabling it by default for now (and changing this behavior in the future in case the upstream Gateway API is explicit about this feature being opt-in) in the community meeting today and did not receive any objection

Xunzhuo (approved):

PR #868, opened 2023-01-06
gatewayapi: validate tls secret in listener certificateRef

### LINKED ISSUE #783: Enable GatewayInvalidTLSConfiguration conformance test

LanceEa:
Previously, the tls secret referenced in a listener certifcateRef would only validate that data existed within the `tls.key` and `tls.cert` fields. This adds additional validation to ensure the data in the `tls.cert` is a valid pem encoded cert and  the `tls.key` contains a valid pem encoded private key. The GatewayInvalidTLSConfiguration conformance test can now be enabled and passes.

Fixes #783


LanceEa:
Ok, sorry for all the pushed commits and fixups. I should finish my coffee before doing early morning PR's 😄. 

I fixed up the lint issue reported by CI and the broken gatewayapi tests. With the new validation logic a few of the tests with TLS now require valid tls secrets for the inputs and expected outputs so I have adjusted those test so they match the new tls validation behavior.

LanceEa:
Looks like my fixup commits didn't get the DCO. Let me know if you need me to squash those or I think the project squash merges so that should take care of them.

danehans:
@LanceEa each commit needs to be signed for the DCO job to pass.

internal/gatewayapi/tls.go
```diff
@@ -0,0 +1,55 @@
...
+
+// validateTLSSecretData ensures the cert and key provided in a secret
+// is not malformed and can be properly parsed.
+func validateTLSSecretData(secret *corev1.Secret) error {
+
+	certData := secret.Data[corev1.TLSCertKey]
+
+	certBlock, _ := pem.Decode(certData)
+	if certBlock == nil {
+		return fmt.Errorf("unable to parse pem data in %s", corev1.TLSCertKey)
```
arkodg, line 24: dont think we should be logging the key value

internal/gatewayapi/tls.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	_, err := x509.ParseCertificate(certBlock.Bytes)
+	if err != nil {
+		return fmt.Errorf("unable to parse certificate in %s: %w", corev1.TLSCertKey, err)
```
arkodg, line 29: same as above

internal/gatewayapi/tls.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	keyData := secret.Data[corev1.TLSPrivateKeyKey]
+
+	keyBlock, _ := pem.Decode(keyData)
+	if keyBlock == nil {
+		return fmt.Errorf("unable to decode pem data in %s", corev1.TLSCertKey)
```
arkodg, line 36: same as above, lets not log the secret content, but maybe the name of the secret and unparsable attribute

internal/gatewayapi/tls.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
```
arkodg, line 38: instead of parsing all types, you could switch on `keyBlock.Type` and use the specific parsing function instead

internal/gatewayapi/tls.go
haq204: `corev1.TLSCertKey` is an alias to the key name inside the secret https://github.com/kubernetes/api/blob/master/core/v1/types.go#L6372. 

But moreover, do we want to enforce that secrets be of type `kubernetes.io/tls`? I don't think the conformance test actually enforces this so it would be a restriction we would be apply additionally applying

arkodg: @haq204 I think its okay to start with this strict check since that is the only type we support today, we can loosen it as we start supporting other types which the API supports
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig

internal/gatewayapi/tls.go
arkodg: 🙈

internal/gatewayapi/tls.go
arkodg: 🙈

internal/gatewayapi/tls.go
LanceEa: @haq204 - This internal helper function is currently written with the assumption that the secret is valid because it is done in the main validate function as seen here: https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/validate.go#L314 

@arkodg - That is just the key so that it gives the user an additional indication of which part of the secret is bad...is the cert or the key. Do you still want me to remove it and make it generic or hard code the key in place?

arkodg: @LanceEa it should be perfectly fine to keep the error strings as is, was a mistake on my end

LanceEa:
Sorry for the pain with the forced push,  I went ahead and squashed the fixup commits to address the DCO and addressed feedback. I will be sure to get the DCO applied to fixup commits too, in the future.

arkodg (approved):

PR #870, opened 2023-01-06
gateway-api: fix supported route kind validation

### LINKED ISSUE #782: Enable GatewayInvalidRouteKind conformance test
*Description*:

LanceEa:
The logic for validating `allowedRoute.kinds` of a listener was overwriting the SupportedKinds when multiple kinds are configured. This caused the `GatewayInvalidRouteKind` conformance test to fail. 

Items addressed:
- refactored `validateAllowedRoutes` to reduce nesting
- fixed `validateAllowedRoutes` logic so  it collects all of the SupportedKinds before setting them on the listener
- provide additional context to user to help identify which RouteKind is invalid on a listener
- updated unit tests for translator based on new expected output
- added additional unit test case for translator to test for when multiple Kinds are added which includes valid and invalid Kinds.

Fixes #782

### DIFF
--- a/internal/gatewayapi/testdata/gateway-with-listener-with-invalid-allowed-routes-group.out.yaml
+++ b/internal/gatewayapi/testdata/gateway-with-listener-with-invalid-allowed-routes-group.out.yaml
@@ -19,6 +19,7 @@ gateways:
     status:
       listeners:
         - name: http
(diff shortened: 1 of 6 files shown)


arkodg (approved):
Good catch, LGTM ! 🚀

PR #871, opened 2023-01-06
Rate Limit Service config in EnvoyGateway API

arkodg:
* Defines the configuration needed to enable as well as connect the rate limit service to a redis database backend
* This will be required to be set by the infrastructure admin if the app developers want the global rate limiting feature using `RateLimitFilter`


kflynn (dismissed):
Looks pretty reasonable here, FWIW. 🙂

danehans:
An appropriate RateLimitFilter status condition if EG is not configured per this PR. Note that I don't expect this point regarding status to be part of this PR but it should be tracked by a separate issue @arkodg.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -86,6 +94,46 @@ type FileProvider struct {
 	// TODO: Add config as use cases are better understood.
 }
 
+// RateLimit defines the configuration associated with the Rate Limit Service
+// using for Global Rate Limiting.
```
danehans, line 98: s/using/used/

danehans, line 98: s/using/used/

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -86,6 +94,46 @@ type FileProvider struct {
...
+type RateLimit struct {
+	// Backend holds the configuration associated with the
+	// database backend used by the rate limit service to store
+	// state associated with global ratelimiting.
+	Backend *RateLimitDatabaseBackend `json:"backend"`
+}
+
+// RateLimitDatabaseBackend defines the configuration associated with
+// the database backend used by the rate limit service.
```
danehans, line 107: `// +union` is required since this is a union type.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -40,6 +40,14 @@ type EnvoyGatewaySpec struct {
 	//
 	// +optional
 	Provider *Provider `json:"provider,omitempty"`
+
+	// RateLimit defines the configuration associated with the Rate Limit service
+	// deployed by Envoy Gateway required to implement the Global Rate limiting
+	// functionality. The specific rate limit service used here is the reference
+	// implementation in Envoy. For more details visit https://github.com/envoyproxy/ratelimit.
+	//
```
danehans, line 48: Although the godocs state "Global Rate Limiting", consider adding a line that states this does not apply for "local" rate limiting and reference https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/ratelimitfilter_types.go for additional details.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -86,6 +94,46 @@ type FileProvider struct {
...
+type RateLimitDatabaseBackendType string
+
+const (
+	// RedisBackendType uses a redis database for the rate limit service.
+	RedisBackendType RateLimitDatabaseBackendType = "Redis"
+)
+
+// RateLimitRedisSettings defines the configuration for connecting to
+// a Redis database.
+type RateLimitRedisSettings struct {
```
danehans, line 132: Why is this a string pointer? This seems liek a required field. If not, keey the pointer and add a godoc that states the default value when unspecified and the appropriate kubebuilder `// +optional` tag.

danehans (changes requested):

api/config/v1alpha1/envoygateway_types.go
arkodg: it is a required field, pointers help differentiate between nil and empty strings, not very useful for this case, can move it to `string` if that is your preference

danehans: Please do.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -86,6 +95,47 @@ type FileProvider struct {
 	// TODO: Add config as use cases are better understood.
 }
 
+// RateLimit defines the configuration associated with the Rate Limit Service
+// used for Global Rate Limiting.
+type RateLimit struct {
+	// Backend holds the configuration associated with the
+	// database backend used by the rate limit service to store
+	// state associated with global ratelimiting.
+	Backend *RateLimitDatabaseBackend `json:"backend"`
```
danehans, line 104: This should not be required since it's required. Xref the [Kube API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) if needed.

danehans (changes requested):

api/config/v1alpha1/envoygateway_types.go
arkodg: thanks for sharing the link, fixed

danehans (approved):

PR #874, opened 2023-01-09
Bump to gateway api v0.6.0

arkodg:
* update go module
* copy over newer experimental CRDs into testdata
* incorporate ObservedGeneration into status condition comparisons
* ensure that the two gateway status update code paths (one when a service or deployment linked to a gateway changes and the second where the gateway-api translator has recomputed the status) use the same status updater function, ensuring that the ObservedGeneration is always up to date in all status fields

Fixes: https://github.com/envoyproxy/gateway/issues/716 Fixes: https://github.com/envoyproxy/gateway/issues/839


Xunzhuo (approved):
lgtm

PR #878, opened 2023-01-10
feat: support configure envoyproxy image from env

Xunzhuo:
When generate manifests, we can specify the envoyproxy image, this is used for release manager/release workflow or end-user to specify the envoyproxy version.

The current envoyproxy image is hardcode and it is the latest dev version of envoyproxy, if the release manager is going to release, he/she will need to change the default dev version to the compatible stable version of envoyproxy:

```shell
make generate-manifests IMAGE=envoyproxy/gateway TAG=v0.3.0 PROXY_IMAGE=envoyproxy/envoy:v1.23.3 OUTPUT_DIR=release-artifacts
```

When make kube-deploy, we can also specify the envoyproxy image to test or dev:

```shell
TAG=latest IMAGE=envoyproxy/gateway-dev PROXY_IMAGE=envoyproxy/envoy:v1.23.3 make kube-deploy
```

### DIFF
--- a/internal/ir/infra.go
+++ b/internal/ir/infra.go
@@ -8,6 +8,7 @@ package ir
 import (
 	"errors"
 	"fmt"
+	"os"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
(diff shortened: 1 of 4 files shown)


arkodg:
@Xunzhuo can you please share motivation behind this ?

If there is enough +1s for this, this should probably be driven from https://github.com/envoyproxy/gateway/blob/89ae35388ec6e0134beb32efbba693d68c685ba8/api/config/v1alpha1/shared_types.go#L22 instead

arkodg:
based on the comments from https://github.com/envoyproxy/gateway/issues/672 it looks like the envoy version needs to be updated in the release branch right before cutting the release.
looks like 1.25 has not been released yet, my suggestion would be to raise a draft PR when it does to proactively catch any issue

danehans:
> my suggestion would be to raise a draft PR when it does to proactively catch any issue
Let's make sure this draft PR

PR #885, opened 2023-01-10
implement parent ref port matching when specified

### LINKED ISSUE #784: Enable HTTPRouteInvalidParentRefNotMatchingListenerPort conformance test
*Description*:

skriss:
When port is specified in a parentRef, only select listeners if they have a matching port, and set
conditions appropriately if not.

Closes #784.

### DIFF
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -159,7 +159,7 @@ func GetReferencedListeners(parentRef v1beta1.ParentReference, gateways []*Gatew
 
 		// The parentRef may be to the entire Gateway, or to a specific listener.
 		for _, listener := range gateway.listeners {
-			if parentRef.SectionName == nil || *parentRef.SectionName == listener.Name {
+			if (parentRef.SectionName == nil || *parentRef.SectionName == listener.Name) && (parentRef.Port == nil || *parentRef.Port == listener.Port) {
 				referencedListeners = append(referencedListeners, listener)
 			}
 		}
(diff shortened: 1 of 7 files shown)


skriss:
Conformance test now running & passing: https://github.com/envoyproxy/gateway/actions/runs/3888096972/jobs/6635148298#step:6:12686

internal/gatewayapi/helpers.go
```diff
@@ -159,7 +159,7 @@ func GetReferencedListeners(parentRef v1beta1.ParentReference, gateways []*Gatew
 
 		// The parentRef may be to the entire Gateway, or to a specific listener.
 		for _, listener := range gateway.listeners {
-			if parentRef.SectionName == nil || *parentRef.SectionName == listener.Name {
+			if (parentRef.SectionName == nil || *parentRef.SectionName == listener.Name) && (parentRef.Port == nil || *parentRef.Port == listener.Port) {
```
arkodg, line 162: guessing this check was added to make it tighter based on API spec https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ParentReference

arkodg (approved):
🚀

internal/gatewayapi/helpers.go
skriss: Yeah, I think `port` was added to the ParentReference struct more recently so we didn't have support for it yet.

arkodg: Ah thanks for confirming

Xunzhuo (approved):

Issue #879, opened 2023-01-10
How configure to expose grpc to grpc-web to FE call


quyenhoang96:
*Description*:
>Describe the issue.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
thanks for raising this issue @quyenhoang96 . The two things that need to be answered here is
* can the [GRPC Route API](https://gateway-api.sigs.k8s.io/api-types/grpcroute/) be implictly used to ingress in [grpc-web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) traffic
* if we implictly allow grpc-web to ingress in, can we auto enable the [grpc-web envoy filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/grpc_web/v3/grpc_web.proto#envoy-v3-api-msg-extensions-filters-http-grpc-web-v3-grpcweb) for all gRPC Route listeners

sunjayBhatia:
> * if we implictly allow grpc-web to ingress in, can we auto enable the [grpc-web envoy filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/grpc_web/v3/grpc_web.proto#envoy-v3-api-msg-extensions-filters-http-grpc-web-v3-grpcweb) for all gRPC Route listeners

contour does this, but we've more recently received requests to disable this behavior for some use cases, in case that is relevant

arkodg:
> > * if we implictly allow grpc-web to ingress in, can we auto enable the [grpc-web envoy filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/grpc_web/v3/grpc_web.proto#envoy-v3-api-msg-extensions-filters-http-grpc-web-v3-grpcweb) for all gRPC Route listeners
> 
> contour does this, but we've more recently received requests to disable this behavior for some use cases, in case that is relevant

thanks for sharing Contour's behavior @sunjayBhatia , is the `grpc-web` envoy filter enabled by default  for`GRPCRoute` or just the native Contour API ?

sunjayBhatia:
we haven't done GRPCRoute yet, but the filter is enabled on all filterchains

arkodg:
> we haven't done GRPCRoute yet, but the filter is enabled on all filterchains

thanks for sharing, inclined to do something similar (enable for GRPC listeners) for this project.

PR #805, opened 2022-12-13
Gateway API Translator support for GRPCRoute

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/642

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #886, opened 2023-01-10
Enable envoy grpc-web filter for HTTP2 listener

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/879


Xunzhuo (approved):

arkodg:
yeah @Xunzhuo also hoping to discuss this in the next community meeting

HoangTheQuyen:
Is it possible to add documentation such as a [quickstart](https://github.com/envoyproxy/gateway/blob/main/examples/kubernetes/quickstart.yaml) or [tcp-routing](https://github.com/envoyproxy/gateway/blob/main/examples/kubernetes/tcp-routing.yaml) item?

arkodg:
@HoangTheQuyen will make sure to add docs as part of https://github.com/envoyproxy/gateway/issues/642, since some work items are still WIP

HoangTheQuyen:
> 

@arkodg Is it this PR: [Here](https://github.com/envoyproxy/gateway/pull/805)?

internal/xds/translator/listener.go
```diff
@@ -92,6 +93,21 @@ func addXdsHTTPFilterChain(xdsListener *listener.Listener, irListener *ir.HTTPLi
 		}},
 	}
 
+	// Enable grpc-web filter for HTTP2
+	if irListener.IsHTTP2 {
```
zirain, line 97: always enables this for a http2 listener?
what about use an annotation to enable or close this

sunjayBhatia, line 97: from the godoc on `IsHTTP2`:

```
```

this is a little misleading i think? since the codec type is always set to "auto" and will infer the downstream protocol which could be HTTP/1 and upgraded when talking with the backend to gRPC now that this filter is applied

HoangTheQuyen: > 

How should I configure it so that it can touch it ? @sunjayBhatia

arkodg: hoping this project stays away from annotations and env variables due to the lack to discoverability and validation

reg auto enable of grpc-web for HTTP2 listener, we have two options
1. wait for upstream gateway-api to dictate whether grpc-web can be implicitly assumed for GRPCRoute or add an opt-in knob in the API. This process is bound to take a much longer time since its a common API across multiple implementations.
2. we auto enable it for GRPCRoute based listeners (which will have `IsHTTP2` set), if users want to only select pure grpc, they could do so by matching on `Content-Type: application/grpc`

I prefer 2. over 1.

arkodg: good point @sunjayBhatia, we could limit `IsHTTP2` to downstreams speaking HTTP/2, thereby limiting `grpc-web` for HTTP/2 only clients

arkodg: updated the PR with the codec type change from auto to http2

sunjayBhatia: seems useful for clients to be able to use gRPC-Web over HTTP/1 -> Envoy -> gRPC backend over HTTP/2 as well! but seems like that could be a separate feature if desired

zirain (commented):
lgtm

Xunzhuo (approved):

PR #889, opened 2023-01-11
update http2 for grpc web

quyenhoang96:


internal/provider/kubernetes/config/envoy-gateway/certgen-job.yaml
```diff
@@ -43,8 +43,8 @@ spec:
     spec:
       containers:
       - name: envoy-gateway-certgen
-        image: envoyproxy/gateway-dev:latest
-        imagePullPolicy: Always
+        image: docker.io/envoyproxy/gateway-dev:7f3c9b4
```
zirain, line 46: please revert unnecessary changes

internal/xds/translator/listener.go
```diff
@@ -103,6 +104,21 @@ func addXdsHTTPFilterChain(xdsListener *listener.Listener, irListener *ir.HTTPLi
 
 	}
 
+	// Enable grpc-web filter for HTTP2
+	if irListener.IsHTTP2 {
```
zirain, line 108: should EG enabled this for all http2 listener?
cc @danehans @arkodg

Issue #656, opened 2022-10-28
support HTTPRequestMirror filter (extended conformance)


skriss:
*Description*:
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter (HTTPRequestMirrorFilter)

There is no upstream conformance test for this yet.

arkodg:
cc @AliceProxy

danehans:
@AliceProxy this issue is tracked for the v0.3.0-rc.1 milestone. Checking to see if you still plan on fixing this issue.

AliceProxy:
@danehans Yep I'm actively working on this now

danehans:
Great. Thanks for the update.

arkodg:
looks like this is needed for [GRPCRoute](https://github.com/kubernetes-sigs/gateway-api/blob/e7260d356a01afc973772610b6223686a971b43e/apis/v1alpha2/grpcroute_types.go#L445) as well

arkodg:
just linking the relevant envoy filter here https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routeaction-requestmirrorpolicy

PR #892, opened 2023-01-12
Add HTTPRoute support for request mirror filters

AliceProxy:
Adds support for request mirror filters for HTTPRoutes. 

Resolves https://github.com/envoyproxy/gateway/issues/656 but as noted on that issue, the request mirrors will need to be added for gRPCRoutes as well.

if there is anything left that needs to be done for request mirror filters on gRPCRoutes once support for them lands from https://github.com/envoyproxy/gateway/pull/805 then I can take that as a separate/follow-up issue.

### DIFF
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -212,6 +212,8 @@ type HTTPRoute struct {
 	DirectResponse *DirectResponse
 	// Redirections to be returned for this route. Takes precedence over Destinations.
 	Redirect *Redirect
+	// Destinations that requests to this HTTPRoute will be mirrored to
+	Mirrors []*RouteDestination
 	// Destinations associated with this matched route.
 	Destinations []*RouteDestination
 	// Rewrite to be changed for this route.
@@ -319,6 +321,11 @@ func (h HTTPRoute) Validate() error {
 	}
 	if h.URLRewrite != nil {
 		if err := h.URLRewrite.Validate(); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+	for _, mirror := range h.Mirrors {
+		if err := mirror.Validate(); err != nil {
 			errs = multierror.Append(errs, err)
 		}
 	}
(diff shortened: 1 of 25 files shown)


internal/ir/xds.go
```diff
@@ -470,6 +481,15 @@ func (r DirectResponse) Validate() error {
 	return errs
 }
 
+// Direct response holds the details for mirroring requests to a second cluster
+// +k8s:deepcopy-gen=true
+type RouteMirror struct {
+	// The name of the route mirror is used to create a new cluster
+	ClusterName string
```
arkodg, line 488: is the `ClusterName` field needed ? can we create a suffixed `-mirror` cluster instead ?

arkodg: if this is reg optimization, we can optimize cluster creation in https://github.com/envoyproxy/gateway/issues/551

AliceProxy: Yeah, I'm cool with getting rid of it and just doing `routename-mirror-0`/`routename-mirror-1` based on the order they appear.

arkodg: sg, thanks @AliceProxy !

internal/ir/xds.go
```diff
@@ -34,6 +34,7 @@ var (
 	ErrRedirectUnsupportedScheme     = errors.New("only http and https are supported for the scheme in redirect filters")
 	ErrHTTPPathModifierDoubleReplace = errors.New("redirect filter cannot have a path modifier that supplies both fullPathReplace and prefixMatchReplace")
 	ErrHTTPPathModifierNoReplace     = errors.New("redirect filter cannot have a path modifier that does not supply either fullPathReplace or prefixMatchReplace")
+	ErrRequestMirrorNoClustername    = errors.New("request mirror filter must have a cluster name")
```
arkodg, line 37: needs to removed

arkodg:
one minor comment, but overall LGTM @AliceProxy !

internal/ir/xds.go
AliceProxy: Whoopsies :open_mouth:

arkodg (approved):
LGTM, thanks !

Xunzhuo:
I will take a quick review, if nothing is blocking I will merge it. Thanks!

Xunzhuo (approved):
LGTM. Thanks @AliceProxy, don`t forget an end user doc : )

PR #893, opened 2023-01-13
Build rate limit service config

arkodg:
* Translates Xds IR into ratelimit service config https://github.com/envoyproxy/ratelimit#configuration

### DIFF
--- a/internal/xds/translator/translator_test.go
+++ b/internal/xds/translator/translator_test.go
@@ -11,14 +11,15 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/envoyproxy/gateway/internal/ir"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	ratelimitserviceconfig "github.com/envoyproxy/ratelimit/src/config"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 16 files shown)


danehans (commented):
@arkodg LGTM after rebasing.

internal/xds/translator/translator_test.go
```diff
@@ -11,14 +11,15 @@ import (
...
 	"testing"
 
+	"github.com/envoyproxy/gateway/internal/ir"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	ratelimitserviceconfig "github.com/envoyproxy/ratelimit/src/config"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
+	goyaml "gopkg.in/yaml.v3"
```
zirain, line 21: use `sigs.k8s.io/yaml` instead

danehans:
@arkodg LGTM after you resolve https://github.com/envoyproxy/gateway/pull/893#discussion_r1070948747.

internal/xds/translator/translator_test.go
arkodg: tried, doesnt work with it, the tags are `yaml` tags, which is why I add to add this package

zirain: merge this first, I will review depguard settings.

zirain (approved):

Issue #883, opened 2023-01-10
Enable GatewayClassObservedGenerationBump conformance test


arkodg:
*Description*:
Seeing this conformance test fail in in https://github.com/envoyproxy/gateway/pull/780

Once fixed it should be enabled via https://github.com/envoyproxy/gateway/blob/main/test/conformance/conformance_test.go

danehans:
This is the last 0.6.0 conformance test that is failing.

Xunzhuo:
See: https://kubernetes.slack.com/archives/CR0H13KGA/p1673599220709819.

danehans:
Waiting for gw api v0.6.1 to drop.

PR #896, opened 2023-01-13
conformance: enable GatewayClassObservedGenerationBump

Xunzhuo:
Fixes: https://github.com/envoyproxy/gateway/issues/883

### DIFF
--- a/test/conformance/conformance_test.go
+++ b/test/conformance/conformance_test.go
@@ -60,6 +60,7 @@ func TestGatewayAPIConformance(t *testing.T) {
 			suite.SupportHTTPResponseHeaderModification,
 			suite.SupportHTTPRouteMethodMatching,
 			suite.SupportRouteDestinationPortMatching,
+			suite.SupportGatewayClassObservedGenerationBump,
 		),
 	})
 	cSuite.Setup(t)
@@ -86,6 +87,7 @@ func TestGatewayAPIConformance(t *testing.T) {
 		tests.GatewaySecretInvalidReferenceGrant,
 		tests.GatewayInvalidTLSConfiguration,
 		tests.GatewayInvalidRouteKind,
+		tests.GatewayClassObservedGenerationBump,
 		tests.HTTPRouteReferenceGrant,
 		tests.HTTPRoutePartiallyInvalidViaInvalidReferenceGrant,
 		tests.HTTPRouteInvalidParentRefNotMatchingListenerPort,


arkodg:
@danehans from this test https://github.com/kubernetes-sigs/gateway-api/blob/release-0.6/conformance/tests/gatewayclass-observed-generation-bump.go it looks like its updating the description of the same GatewayClass, not a new one (`NamespacedName` used is same)

Xunzhuo:
The GC name is different, so I think it's created a new one. 

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/gatewayclass-observed-generation-bump.yaml

@arkodg

arkodg:
ah thanks for the clarification @Xunzhuo 
looking at https://github.com/envoyproxy/gateway/pull/861/, it updates status for GCs with matching controller name but not accepted, will that fix this test ?

Xunzhuo:
@arkodg I might not, see: https://github.com/kubernetes-sigs/gateway-api/blob/release-0.6/conformance/tests/gatewayclass-observed-generation-bump.go#L65. It cannot process util GC is accepted.

Xunzhuo:
See: https://kubernetes.slack.com/archives/CR0H13KGA/p1673599220709819

danehans:
@Xunzhuo thanks for discussing this issue with the Gateway API community. I have confirmed that EG is handling observedGeneration properly for GCs with a matching controllerName. I performed the following to reproduce:

- Install latest EG: `kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/latest/install.yaml`
- Install latest quickstart: `kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/latest/quickstart.yaml`
- Confirm quickstart resources are reporting the expected status conditions.
- Create a 2nd GC, `eg2`, with a controllerName that matches EG:
  ```
  kind: GatewayClass
  metadata:
    name: eg2
  spec:
  ```
- Verify the 2nd GC reports the expected "Accepted=False" status condition and an observedGeneration that matches `spec.generation`, e.g. `1`.
- Update the 2nd GC to bump `spec.generation`:
  ```
  kind: GatewayClass
  metadata:
    name: eg2
  spec:
    parametersRef:
      group: config.gateway.envoyproxy.io
      kind: EnvoyProxy
      name: eg
  ```
- Verify the 2nd GC, `eg2`, still surfaces the `Accepted=False` status condition and the observedGeneration matches `spec.generation`, e.g. `2`.

Instead of changing the GC name as done by this PR, I suggest that we comment-out this test with a link to https://github.com/kubernetes-sigs/gateway-api/issues/1650. Thoughts @arkodg @skriss @youngnick @AliceProxy @zirain ?

arkodg:
yah @danehans I agree with commenting out this extended test (`GatewayClassObservedGenerationBump`) . In the initial design, the project decided to support a single accepted GatewayClass to not have to deal with multi tenancy (multiple accepted GatewayClass) within a single process (users should instead create a new Envoy Gateway controller). 
The design can def be changed in the future, but imho we will incur more complexity and community cycles without adding much value prop if we take on this design change to make above test pass

arkodg:
based on @youngnick 's suggestion in https://kubernetes.slack.com/archives/CR0H13KGA/p1673836949189379?thread_ts=1673599220.709819&cid=CR0H13KGA EG should be able to support the test if the conformance test is edited to enforce an `Accepted` condition rather than a `Accpeted=True` condition .

arkodg:
> based on @youngnick 's suggestion in https://kubernetes.slack.com/archives/CR0H13KGA/p1673836949189379?thread_ts=1673599220.709819&cid=CR0H13KGA EG should be able to support the test if the conformance test is edited to enforce an `Accepted` condition rather than a `Accpeted=True` condition .

raised https://github.com/kubernetes-sigs/gateway-api/pull/1655 to address this.

danehans:
Waiting for gw api v0.6.1.

arkodg:
can confirm via https://github.com/envoyproxy/gateway/pull/780 that this test has been fixed, we should use this PR to insert `GatewayClassObservedGenerationBump` into the list of conformance tests that ned to be run

danehans:
@Xunzhuo now that gw api has been bumped to v0.6.1, can you please rebase and remove draft?

arkodg (approved):
🚀

Issue #898, opened 2023-01-13
UDP service can't not be accessed through gateway cluster IP/external IP


zhaohuabing:
*Description*:
UDP service can't not be accessed through gateway cluster IP/external IP.

It's because the protocol of the service is wrong. The current protocol is TCP, it should be UDP instead.

```yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2022-12-23T06:14:57Z"
  labels:
    app.gateway.envoyproxy.io/name: envoy
    gateway.envoyproxy.io/owning-gateway-name: eg
    gateway.envoyproxy.io/owning-gateway-namespace: default
  name: envoy-default-eg-64656661
  namespace: envoy-gateway-system
  resourceVersion: "808984"
  uid: 53a22ef1-7262-4f9f-9469-40356a7ab8d7
spec:
  clusterIP: 10.96.7.8
  clusterIPs:
  - 10.96.7.8
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: udp
    port: 5300
    protocol: TCP   //The protocol should be UDP
    targetPort: 5300
```

zhaohuabing:
Please assign it to me.

PR #899, opened 2023-01-13
fix UDP service can't not be accessed through gateway cluster IP/external IP

zhaohuabing:

fixes: https://github.com/envoyproxy/gateway/issues/898


internal/infrastructure/kubernetes/service.go
```diff
@@ -33,9 +33,13 @@ func (i *Infra) expectedService(infra *ir.Infra) (*corev1.Service, error) {
 	for _, listener := range infra.Proxy.Listeners {
 		for _, port := range listener.Ports {
 			target := intstr.IntOrString{IntVal: port.ContainerPort}
+			protocol := corev1.ProtocolTCP
```
zirain, line 36: can you address a comment that we need E2E in the future?

Xunzhuo: This can be converted into an issue, and it's not blocking. I will approve it.

Xunzhuo (approved):

PR #902, opened 2023-01-14
fix typo

zirain:

I think this's a typo mistake, cc @arkodg


arkodg (approved):
thanks for catching this !

PR #914, opened 2023-01-16
egctl: add `version` command

zirain:
xref: https://github.com/envoyproxy/gateway/issues/915


Xunzhuo:
Can you sparate the design docs into issues in case we can better track it or make contributors/maintainers can help implement it? https://github.com/envoyproxy/gateway/blob/main/docs/latest/design/egctl.md @zirain

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+		Short:   "Show versions",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return versions(cmd.OutOrStdout(), remote, output)
+		},
+	}
+
+	flags := versionCommand.Flags()
+	options.AddKubeConfigFlags(flags)
+	versionCommand.PersistentFlags().BoolVar(&remote, "remote", false, "if true, retrieve server version from cluster")
+	versionCommand.PersistentFlags().StringVarP(&output, "output", "o", "json", "One of 'yaml' or 'json'")
```
danehans, line 41: Since this is user-facing, I prefer yaml as the default over json.

```
$ egctl version --remote -o yaml
client: v0.2.0
servers:
  envoy-gateway-system/envoy-gateway-84f8b94f-8tqq7:
    envoyGatewayVersion: v0.2.0
    envoyVersion: latest
    gatewayAPIVersion: v0.6.0
    gitCommitID: 6f8b6fbeab204f9d5b334fb1e1cb6fbdc8b33112
```

Can you also update the following to the version output:
- Have the version command return `server` instead of `servers.
- Update `envoyVersion` to `envoyProxyVersion`.
- Parse `envoy-gateway-system/envoy-gateway-84f8b94f-8tqq7` to provide additional context in the output. For example:
```
deploymentName: envoy-gateway
```

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return fmt.Errorf("list EG pods fail: %w", err)
+		}
+
+		for _, pod := range pods.Items {
+			nn := types.NamespacedName{
+				Namespace: pod.Namespace,
+				Name:      pod.Name,
+			}
+			stdout, _, err := c.PodExec(nn, "envoy-gateway", "envoy-gateway version -ojson")
```
danehans, line 77: The EG namespace should be exposed as a flag that defaults to `envoy-gateway`.

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+	options.AddKubeConfigFlags(flags)
+	versionCommand.PersistentFlags().BoolVar(&remote, "remote", false, "if true, retrieve server version from cluster")
+	versionCommand.PersistentFlags().StringVarP(&output, "output", "o", "json", "One of 'yaml' or 'json'")
+
+	return versionCommand
+}
+
+type VersionInfo struct {
+	ClientVersion  string                   `json:"client"`
+	ServerVersions map[string]*version.Info `json:"servers,omitempty"`
```
danehans, line 48: `ServerVersions` should be updated if the `remote` flag is removed and `egctl version` always returns the server version.

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+		ServerVersions: map[string]*version.Info{},
+	}
+}
+
+func versions(w io.Writer, remote bool, output string) error {
+	v := Get()
+
+	c, err := kube.NewCLIClient(options.DefaultConfigFlags.ToRawKubeConfigLoader())
+	if err != nil {
+		return fmt.Errorf("build CLI client fail: %w", err)
```
danehans, line 63: s/build CLI client fail/failed to build Kubernetes client/

cmd/egctl/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/envoyproxy/gateway/internal/ctl/app"
```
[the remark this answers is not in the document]
danehans: Is the `app` directory for `egctl` being used to separate `egctl` and `envoy-gateway` builds? Would it be better to reorganize and use the current `internal/cmd` structure? For example:
```
internal/cmd/version
internal/cmd/envoy-gateway
internal/cmd/egctl
```

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		for _, pod := range pods.Items {
+			nn := types.NamespacedName{
+				Namespace: pod.Namespace,
+				Name:      pod.Name,
+			}
+			stdout, _, err := c.PodExec(nn, "envoy-gateway", "envoy-gateway version -ojson")
+			if err != nil {
+				return fmt.Errorf("pod exec on %s fail: %w", nn, err)
```
danehans, line 79: s/fail/failed/

Other parts of the code use `%s/%s` with nn.Namespace and nn.Name.

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+
+			info := &version.Info{}
+			if err := json.Unmarshal([]byte(stdout), info); err != nil {
+				return fmt.Errorf("unmarshall pod %s exec result fail: %w", nn, err)
```
danehans, line 84: Other parts of the code use `%s/%s` with nn.Namespace and nn.Name.

internal/utils/kube/client.go
```diff
@@ -0,0 +1,170 @@
...
+	GatewayScheme = gatewayScheme()
+)
+
+func gatewayScheme() *runtime.Scheme {
+	scheme := runtime.NewScheme()
+	utilruntime.Must(kubescheme.AddToScheme(scheme))
+	utilruntime.Must(apiextv1.AddToScheme(scheme))
+	return scheme
+}
+
```
danehans, line 39: Why not use the existing `GetScheme()`?

internal/utils/kube/client.go
```diff
@@ -0,0 +1,170 @@
+// Copyright Envoy Gateway Authors
```
danehans, line 1: IMO this should be part of a new pkg `internal/kubernetes`.

internal/utils/kube/client.go
```diff
@@ -0,0 +1,170 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
+	kubescheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/remotecommand"
```
danehans, line 25: Why not use client-go?

danehans (changes requested):

internal/ctl/app/version.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func NewVersionsCommand() *cobra.Command {
+	var (
+		output string
+		remote bool
+	)
+
+	versionCommand := &cobra.Command{
+		Use:     "versions",
```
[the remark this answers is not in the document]
zirain: same as https://github.com/envoyproxy/gateway/blob/2479b9391055d4b0958e8f25b9966c968d8d75a8/internal/cmd/versions.go#L19

should we change both?

internal/ctl/app/version.go
zirain: here is not namespace, it's container name.
BTW, it should support customization.

internal/utils/kube/client.go
zirain: I don't understand your point.

internal/ctl/app/version.go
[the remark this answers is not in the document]
zirain: /hold https://github.com/envoyproxy/gateway/pull/956

internal/cmd/egctl/version.go
```diff
@@ -0,0 +1,126 @@
...
+func NewVersionsCommand() *cobra.Command {
+	var (
+		output          string
+		egContainerName string
+	)
+
+	versionCommand := &cobra.Command{
+		Use:     "version",
+		Aliases: []string{"versions", "v"},
+		Short:   "Show versions",
```
danehans, line 37: /versions/version/

cmd/egctl/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+func rootCommand() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:               "egctl",
+		Long:              "A command line utility for operating Envoy Gateway",
+		SilenceUsage:      true,
+		DisableAutoGenTag: true,
+	}
+
+	rootCmd.AddCommand(egctl.NewVersionsCommand())
```
danehans, line 32: s/NewVersionsCommand/NewVersionCommand/

internal/cmd/egctl/version.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/envoyproxy/gateway/internal/cmd/options"
+	"github.com/envoyproxy/gateway/internal/cmd/version"
+	kube "github.com/envoyproxy/gateway/internal/kubernetes"
+)
+
+const (
+	yamlOutput = "yaml"
+)
+
+func NewVersionsCommand() *cobra.Command {
```
danehans, line 28: s/NewVersionsCommand/NewVersionCommand/

internal/cmd/egctl/version.go
```diff
@@ -0,0 +1,126 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return versions(cmd.OutOrStdout(), egContainerName, output)
+		},
+	}
+
+	flags := versionCommand.Flags()
+	options.AddKubeConfigFlags(flags)
+
+	versionCommand.PersistentFlags().StringVarP(&output, "output", "o", yamlOutput, "One of 'yaml' or 'json'")
+	versionCommand.PersistentFlags().StringVar(&egContainerName, "eg-container-name", "envoy-gateway", "Name of the Envoy Gateway container")
```
danehans, line 47: I don't expect the EG container name to change so I'm not sure why the container name needs to be exposed to a user. However, I can see EG being run in a different ns than `envoy-gateway-system` and expose the ns to the user while defaulting to "envoy-gateway-system".

danehans (changes requested):

internal/cmd/egctl/version.go
zirain: `version` will search all namespace, expect vendors may use EG multi-tenancy with custom container name？

zirain:
output of running `make build && bin/linux/amd64/egctl v`

```
client: v0.2.0
server:
- Name: envoy-gateway-6cbd765bdc-ddqpm
  Namespace: envoy-gateway-system
  envoyGatewayVersion: v0.2.0
  envoyProxyVersion: latest
  gatewayAPIVersion: v0.6.0
  gitCommitID: 54fdd5bc185d8fa57e7afa7ec5f53bb1466757e0
```

internal/cmd/egctl/version.go
danehans: I'm not following your point here about the EG container name. For multitenancy, a user would install EG in different namespaces with each EG using a different `controllerName`. A user would then create a GatewayClass per EG and one or more Gateways that reference the GatewayClass. By default, egctl should expect EG in the `envoy-gateway-system` namespace with the Deployment name of `envoy-gateway`. If egctl needs to communicate with an EG in a different namespace, a flag such as `--namespace` should be surfaced. For security purposes, we should avoid providing egctl access across all namespaces.

danehans:
I just tested egctl and it's showing the incorrect client version. Maybe this PR needs to be rebased?
```
$ ./bin/darwin/amd64/egctl version
client: v0.2.0
server:
- Name: envoy-gateway-7fd4c7b9cf-jpjrt
  Namespace: envoy-gateway-system
  envoyGatewayVersion: v0.3.0
  envoyProxyVersion: v1.25-latest
  gatewayAPIVersion: v0.6.1
  gitCommitID: f735041d6ce31e104bf00f9a96f47d327634a923
```

In the future, EG could be running multiple replicas. In this case, which pod name will be displayed? It might be better to surface the deployment name instead.

internal/cmd/egctl/version.go
zirain: let me remove this part first.

zirain:
> I just tested egctl and it's showing the incorrect client version. Maybe this PR needs to be rebased?
> 
> ```
> $ ./bin/darwin/amd64/egctl version
> client: v0.2.0
> server:
> - Name: envoy-gateway-7fd4c7b9cf-jpjrt
>   Namespace: envoy-gateway-system
>   envoyGatewayVersion: v0.3.0
>   envoyProxyVersion: v1.25-latest
>   gatewayAPIVersion: v0.6.1
>   gitCommitID: f735041d6ce31e104bf00f9a96f47d327634a923
> ```
> 
> In the future, EG could be running multiple replicas. In this case, which pod name will be displayed? It might be better to surface the deployment name instead.

after rebasing from main, it's correct now.

danehans:
@zirain thanks for rebasing. I have confirmed that the client/server versions now match. As a follow-on to https://github.com/envoyproxy/gateway/pull/914#issuecomment-1426105378, EG can run multiple replicas in the future. In this case, which pod name will be displayed? I think it's better to surface the EG deployment name instead of the pod name. Thoughts?

zirain:
> In the future, EG could be running multiple replicas. In this case, which pod name will be displayed? It might be better to surface the deployment name instead.

for now, envoy pod will displayed.

danehans:
Why display the Pod name instead of the Deployment name?

zirain:
> Why display the Pod name instead of the Deployment name?

if you look at the implement, `egctl` list pods by selector, then gather server information via `PodExec`.
BTW, I have no strong option about change to display deployment name.

danehans:
Yes, I understand the implementation. I just don't understand the purpose of displaying the EG pod name instead of the deployment name. I'm fine with resolving this in the future when EG supports multiple replicas.

danehans:
@zirain I can approve after you resolve the merge conflicts.

danehans (approved):

Xunzhuo:
> Yes, I understand the implementation. I just don't understand the purpose of displaying the EG pod name instead of the deployment name. I'm fine with resolving this in the future when EG supports multiple replicas.

I prefer deployment name too, but let us get this in first. We can make it later.

Xunzhuo (approved):
Thanks @zirain !

PR #916, opened 2023-01-16
golangci: add depguard

zirain:

### DIFF
--- a/tools/linter/golangci-lint/.golangci.yml
+++ b/tools/linter/golangci-lint/.golangci.yml
@@ -4,18 +4,24 @@ run:
 linters:
   enable:
     - bodyclose
+    - depguard
+    - exportloopref
     - gofmt
     - goimports
-    - revive
-    - gosec
-    - misspell
-    - exportloopref
-    - unconvert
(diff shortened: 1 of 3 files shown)


tools/linter/golangci-lint/.golangci.yml
```diff
@@ -4,18 +4,24 @@ run:
...
     - misspell
-    - exportloopref
+    - revive
     - unconvert
     - unparam
-    - goheader
-    - gocritic
 
 linters-settings:
+  depguard:
```
arkodg, line 20: is this needed, I dont think we use `gogo` in this project ?
and reg `gopkg.in/yaml.v3` - we actually need to use this package in https://github.com/envoyproxy/gateway/pull/893

zirain: prevent use `gogo` in the future.

zirain: updated with `nolint` for special usecase.

arkodg (approved):
LGTM thanks !

PR #924, opened 2023-01-17
Adds EnvoyProxy to Translator and Kube Infra Manager

### LINKED ISSUE #713: Support configuring number of replicas of Envoy Proxy
Currently EG deploys only a single instance of Envoy Proxy. For scaling to high number of requests it will be useful to be able to increase number of replicas of Envoy Proxy to distribute load. It can be used for the following scenario -
- Deploy EG with one replica of Envoy Proxy
- As the traffic increases, metrics report high CPU usage of Envoy Proxy pod
- Increase the number of replicas of Envoy Proxy to distribute the load and reduce CPU usage per pod

This property can be exposed through [Envoy Proxy Config API](https://github.com/envoyproxy/gateway/blob/v0.2.0/api/config/v1alpha1/envoyproxy_types.go) - #703

danehans:
Adds initial EnvoyProxy support to Gateway API Translator and Kube Infra Manager packages.

Fixes #713 

### DIFF
--- a/api/config/v1alpha1/helpers.go
+++ b/api/config/v1alpha1/helpers.go
@@ -55,9 +55,66 @@ func DefaultProvider() *Provider {
 	}
 }
 
+// GetProvider returns the Provider of EnvoyGateway or a default Provider if unspecified.
 func (e *EnvoyGateway) GetProvider() *Provider {
 	if e.Provider != nil {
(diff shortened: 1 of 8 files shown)


zirain (approved):
LGTM

Xunzhuo (approved):

PR #929, opened 2023-01-17
bugfix: remove stale listener condition

### LINKED ISSUE #773: Stale listener condition present in Gateway for a non-existing Listener
*Description*:
Following the reproduction steps, it is seen that the Gateway is left with a stale listener condition, for a listener
that no longer exists.

*Repro steps*:
1. apply `examples/kubernetes/quickstart.yaml` - ensure that Gateway has proper Status conditions, with 1 listener condition
showing `attachedRoutes: 1`.
2. Add another listener in the gateway, for the listener spec to look something like this
```yaml
listeners:
- allowedRoutes:
     namespaces:
       from: Same
     name: http
     port: 8080
     protocol: HTTP
- allowedRoutes:
     namespaces:
       from: Same
     name: http2
     port: 8081
     protocol: HTTP 
```
This should create 2 listener conditions, both specifying `attachedRoutes: 1`.
3. Remove the single HTTPRoute that was created by the `quickstart.yaml`. Both existing listener conditions are updated correctly, specifying `attachedRoutes: 0`.
4. Re-apply the `quickstart.yaml` which will "recreate" the HTTPRoute, and "update" the Gateway to have only one listener.
5. The bug is that - in the Gateway status still shows the stale listener Condition corresponding to listener `http2` with `attachedRoutes: 0`
The spec and status thus looks like this
```yaml
 spec:
    gatewayClassName: eg
    listeners:
    - allowedRoutes:
        namespaces:
          from: Same
      name: http
      port: 8080
      protocol: HTTP
  status:
    ....
    conditions:
    ....
    listeners:
    - attachedRoutes: 1
      conditions:
      - lastTransitionTime: "2022-12-03T06:58:58Z"
        message: Listener is ready
        observedGeneration: 3
        reason: Ready
        status: "True"
        type: Ready
      name: http
      supportedKinds:
        kind: HTTPRoute
    - attachedRoutes: 0
      conditions:
      - lastTransitionTime: "2022-12-03T06:58:48Z"
        message: Listener is ready
        observedGeneration: 2
        reason: Ready
        status: "True"
        type: Ready
      name: http2
      supportedKinds:
        kind: HTTPRoute
```

haq204:
Previously the listener status in the Gateway resource could be left stale if one of the listeners is removed.

We add new method to GatewayContext to reset the listener status and listener contexts and is called before re-computing

Testing:
- Add unit test for new method
- Add a couple of new testcases to the Translation unit tests to check for stale listener conditions and to ensure the status didn't change if the condition did not change.

Fixes #773

### DIFF
--- a/internal/gatewayapi/translator.go
+++ b/internal/gatewayapi/translator.go
@@ -124,6 +124,8 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
 	return newTranslateResult(gateways, httpRoutes, tlsRoutes, tcpRoutes, udpRoutes, xdsIR, infraIR)
 }
 
+// GetRelevantGateways returns GatewayContexts, containing a copy of the original
+// Gateway with the Listener statuses reset.
 func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*GatewayContext {
 	var relevant []*GatewayContext
 
@@ -136,14 +138,7 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
 			gc := &GatewayContext{
 				Gateway: gateway.DeepCopy(),
 			}
-
-			for _, listener := range gateway.Spec.Listeners {
-				l := gc.GetListenerContext(listener.Name)
-				// Reset conditions and attached route count
-				// since it will be recomputed during translation.
-				l.ResetConditions()
-				l.ResetAttachedRoutes()
-			}
+			gc.ResetListeners()
 
 			relevant = append(relevant, gc)
 		}
--- a/internal/gatewayapi/contexts.go
+++ b/internal/gatewayapi/contexts.go
@@ -74,6 +74,23 @@ func (g *GatewayContext) GetListenerContext(listenerName v1beta1.SectionName) *L
 	return ctx
 }
 
+// ResetListeners resets the listener statuses and re-generates the GatewayContext
+// ListenerContexts from the Gateway spec.
+func (g *GatewayContext) ResetListeners() {
+	numListeners := len(g.Spec.Listeners)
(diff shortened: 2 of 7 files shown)


internal/gatewayapi/translator.go
```diff
@@ -136,14 +138,7 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
...
 			}
-
-			for _, listener := range gateway.Spec.Listeners {
-				l := gc.GetListenerContext(listener.Name)
-				// Reset conditions and attached route count
-				// since it will be recomputed during translation.
-				l.ResetConditions()
-				l.ResetAttachedRoutes()
-			}
+			gc.ResetListeners()
```
arkodg, line 141: * can we delete the `ResetConditions` and `ResetAttachedRoutes` since they are no longer used
* the code within `GetListenerContext` and `ResetListeners` is very similar, can we break up `GetListenerContext` to only perform a Get and use Reset to reset the context ?

haq204: yea forgot to bring that up but also wanted to ask if we could get rid `ResetConditions` and `ResetAttachedRoutes`

arkodg: yah please delete, also `GetListenerContext` can be deleted too I think
cc @skriss

haq204: Now that I think about it, `GetListenerContext` is also not used anymore. Wondering if we should remove that method too?

arkodg: sure go for it

haq204: @arkodg removed unused methods

arkodg (approved):
LGTM thanks !

AliceProxy (approved):
lgtm, thanks

PR #933, opened 2023-01-21
RateLimitFilter: Gateway API to Xds IR translaton

arkodg:
* Also rm'ed `ResetConditon` and instead set `Accepted=False` for negative cases for ExtensionFilter cases

### DIFF
--- a/internal/gatewayapi/testdata/httproute-with-invalid-ratelimitfilter.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-with-invalid-ratelimitfilter.out.yaml
@@ -50,7 +50,7 @@ httpRoutes:
       - type: ExtensionRef
         extensionRef:
           group: gateway.envoyproxy.io
-          kind: AuthenticationFilter
+          kind: RateLimitFilter
           name: test 
   status:
     parents:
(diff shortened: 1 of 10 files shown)


internal/gatewayapi/testdata/httproute-with-invalid-ratelimitfilter.out.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - parentRef:
+        namespace: envoy-gateway
+        name: gateway-1
+        sectionName: http
+      controllerName: gateway.envoyproxy.io/gatewayclass-controller
+      conditions:
+      - type: Accepted
+        status: "False"
+        reason: UnsupportedValue
+        message: "Unable to translate RateLimitFilter: default/test"
```
danehans, line 66: We should iterate on this status message to provide more actionable feedback to the user.

danehans (approved):

internal/gatewayapi/testdata/httproute-with-invalid-ratelimitfilter.out.yaml
arkodg: raised https://github.com/envoyproxy/gateway/issues/938 to track this

PR #939, opened 2023-01-26
Translate RateLimitInfra into K8s resources

arkodg:

### DIFF
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -290,7 +290,6 @@ func (i *Infra) createOrUpdateProxyDeployment(ctx context.Context, infra *ir.Inf
 	if err != nil {
 		return err
 	}
-
 	return i.createOrUpdateDeployment(ctx, deploy)
 }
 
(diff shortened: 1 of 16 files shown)


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -50,6 +50,13 @@ const (
 	envoyAdminPort = 19000
 	// envoyAdminAccessLogPath is the path used to expose admin access log.
 	envoyAdminAccessLogPath = "/dev/null"
+
+	// rateLimitInfraName is the name for rate-limit resources.
+	rateLimitInfraName = "envoy-ratelimit"
+	// rateLimitInfraHTTPPort is the http port that the rate limit service listens on.
+	rateLimitInfraHTTPPort = 8080
+	// rateLimitInfraImage is the container image for the rate limit service.
+	rateLimitInfraImage = "envoyproxy/ratelimit:latest"
```
zirain, line 59: make this configurable in `EnvoyGateway.RateLimit`

internal/infrastructure/kubernetes/service.go
```diff
@@ -107,15 +107,98 @@ func (i *Infra) createOrUpdateService(ctx context.Context, infra *ir.Infra) erro
...
+}
+
+func (i *Infra) createOrUpdateService(ctx context.Context, svc *corev1.Service) error {
+	current := &corev1.Service{}
+	key := types.NamespacedName{
+		Namespace: svc.Namespace,
+		Name:      svc.Name,
+	}
+
+	if err := i.Client.Get(ctx, key, current); err != nil {
```
zirain, line 180: can we make `L180-L197` reusable?

arkodg: looks like we'd move the complexity over to the util func -  we could  pass objects as`Client.Object` to it, which is fine, but we also do an equality check on `obj.Spec`, so we'd have to typecast it back to the orig obj type

internal/infrastructure/kubernetes/deployment.go
arkodg: sure, rather than add a TODO comment in code, should be raised as a GH issue explaining the use cases where this is needed

internal/infrastructure/kubernetes/service.go
zirain: address and refactor this later

zirain (approved):
LGTM

PR #940, opened 2023-01-26
Switch the Quickstart to use port 80 instead of 8080.

### LINKED ISSUE #644: switch to Listener ports 80/443 instead of 8080/8443 for the quickstart & secure Gateway guides
Using the standard ports 80 & 443 would be more realistic, and  would simplify various commands that would no longer need to explicitly specify ports. It would also eliminate the need to explicitly specify `Host` headers in curl commands using `--resolve` (that seems to include the port # in the Host header by default if using non-standard ports).

kflynn:
Switch quickstart.yaml to use port 80, and update quickstart.md in the docs.

Fixes #644

Signed-off-by: Flynn <emissary@flynn.kodachi.com>

### DIFF
--- a/docs/latest/user/http-redirect.md
+++ b/docs/latest/user/http-redirect.md
@@ -35,7 +35,7 @@ spec:
           scheme: https
           statusCode: 301
           hostname: www.example.com
-          port: 8443
+          port: 443
       backendRefs:
       - name: backend
         port: 3000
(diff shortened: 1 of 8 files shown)


arkodg:
hey @kflynn you will also need to change the docs for all http features relying on that port for e.g. https://github.com/envoyproxy/gateway/blob/main/docs/latest/user/http-urlrewrite.md

also if we moving to `80` from `8080` can you also move from`8443` for `443` for https, TIA !

kflynn:
Augh! Thanks much, @arkodg, that was just foolish of me. Back to draft this goes!

danehans:
@kflynn are you able to update this PR in time for the 0.3.0-rc.1 milestone?

kflynn:
If anyone finds docs I (somehow) missed, let me know!

zirain (commented):
LGTM

PR #943, opened 2023-01-30
ratelimit: support distinct header match

zirain:

cc @arkodg

### DIFF
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -682,8 +682,6 @@ func (t *Translator) processExtensionRefHTTPFilter(filter v1beta1.HTTPRouteFilte
 								}
 								rules[i].HeaderMatches = append(rules[i].HeaderMatches, m)
 							case *header.Type == egv1a1.HeaderMatchDistinct && header.Value == nil:
-								fallthrough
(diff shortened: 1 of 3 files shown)


internal/gatewayapi/filters.go
```diff
@@ -681,6 +681,8 @@ func (t *Translator) processExtensionRefHTTPFilter(filter v1beta1.HTTPRouteFilte
 									SafeRegex: header.Value,
 								}
 								rules[i].HeaderMatches = append(rules[i].HeaderMatches, m)
+							case *header.Type == egv1a1.HeaderMatchDistinct && header.Value == nil:
```
arkodg, line 684: thanks for catching this, the bug is an unfortunate typo :( 
```
*header.Type == egv1a1.HeaderMatchRegularExpression && header.Value == nil
```
should be changed to 
```
*header.Type == egv1a1.HeaderMatchDistinct && header.Value == nil
```
can you also enhance https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/testdata/httproute-with-valid-ratelimitfilter.in.yaml#L52 by adding another `rule` for distinct

zirain: updated

arkodg (approved):

Issue #852, opened 2023-01-04
Ratelimit based on IP Subnet


arkodg:
*Description*:
Support ratelimiting client based on their IP addresses.
Relates to https://github.com/envoyproxy/gateway/issues/670

zirain:
xref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html?highlight=maskedremoteaddress#config-route-v3-ratelimit-action-maskedremoteaddress

zirain:
sorry, I forgot this, will do this ASAP.

PR #946, opened 2023-01-30
ratelimit(api): support SourceIP

zirain:

xref: #852

cc @arkodg


api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -99,6 +99,23 @@ type RateLimitSelectCondition struct {
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
 	Headers []HeaderMatch `json:"headers,omitempty"`
+
+	// SourceIP defines the match attributes for source IP addresses from “X-Forwarded-For“ header.
+	// +optional
+	SourceIP *SourceIPMatch `json:"sourceIP,omitempty"`
```
arkodg, line 105: thoughts on alternate names -  `SourceIPSubnet` or `SourceIPCIDR` ?

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -99,6 +99,23 @@ type RateLimitSelectCondition struct {
...
+}
+
+// SourceIPMatch defines the match attributes for source IP addresses from “X-Forwarded-For“ header.
+type SourceIPMatch struct {
```
arkodg, line 109: prefer if this was a `string` , internally the system should be able to validate a IPv4/IPv6 CIDR using something like https://pkg.go.dev/net#ParseCIDR

arkodg:
thanks for kickstarting the design for ratelimiting based on source IP,  added comments

zirain:
> thanks for kickstarting the design for ratelimiting based on source IP, added comments

updated

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -99,6 +99,11 @@ type RateLimitSelectCondition struct {
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
 	Headers []HeaderMatch `json:"headers,omitempty"`
+
+	// SourceIPCIDR defines the match attributes for masked source IP addresses from `X-Forwarded-For` header.
+	// For example, `192.168.0.1/32`, `192.168.0.0/24`, `001:db8::/64`.
+	// +optional
+	SourceIPCIDR *string `json:"sourceIPCIDR,omitempty"`
```
arkodg, line 106: sorry for the churn here, but I think the field name `SourceIP` is more apt, and the doc string can highlight `subnet` or `cidr`.

zirain: done

docs/latest/design/ratelimit.md
```diff
@@ -165,6 +165,51 @@ spec:
...
+### Rate limit per source IP
+
+* Here is an example of a rate limit implemented by the application developer that limits the total requests made
+to a specific route by matching on source IP. In this case, requests from `x.x.x.x` will be rate limited at 10 requests/hour.
+
+```yaml
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: RateLimitFilter
+metadata:
+  name: ratelimit-per-user
```
arkodg, line 177:
```suggestion
  name: ratelimit-per-ip
```

arkodg:
this design looks good @zirain , lgtm from my side, lets merge post 0.3.0 so the unimplemented API field doesnt show up in docs ?

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -99,6 +99,11 @@ type RateLimitSelectCondition struct {
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
 	Headers []HeaderMatch `json:"headers,omitempty"`
+
+	// SourceIP is IP CIDR from `X-Forwarded-For` header.
```
arkodg, line 103: want to ensure that edge proxy case is taken care of, will the internal implementation add the downstream's source IP into `X-Forwarded-For` by setting `use_remote_address` ?https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for

zirain: I do not think, there's something special.

arkodg: imho this would need to be applied internally else the user might not be able to ratelimit based on the  L3 Source IP

zirain: you may need implement a new action in Envoy?

arkodg: why not peruse existing mechanism highlighted above ? setting `use_remote_address` which ensures XFF has remote/client/source IP set

arkodg: will raise a GH issue once this PR is in, to ensure `use_remote_address` is set

zirain: > will raise a GH issue once this PR is in, to ensure `use_remote_address` is set

xref: https://github.com/envoyproxy/gateway/issues/998

arkodg: thanks for raising it, we should also reframe the current doc string into - 
```
// SourceIP is the IP CIDR that represents the range of Source IP Addresses of the client. These could also be the intermediate addresses through which the request has flown through and is part of the  `X-Forwarded-For` header.
```

arkodg:
thanks for updating the doc string ! 
LGTM from my side, hoping this API PR gets another review merged it gets merged it
cc @Xunzhuo @danehans @youngnick @skriss @AliceProxy

Xunzhuo (approved):
LGTM, side comment, this defines the sourceIP api field in RateLimitFilter, but have we implemented in EG?

zirain:
> LGTM, side comment, this defines the sourceIP api field in RateLimitFilter, but have we implemented it in EG?

will implement it later.

PR #955, opened 2023-02-01
Support HTTPRoute with empty matches

arkodg:
* Add a Path Prefix match on "/" which matches every request when the matches field is not set

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule


zirain (approved):
LGTM

PR #960, opened 2023-02-02
fix: gen docs failed

Xunzhuo:


docs/v0.2.0/conf.py
```diff
@@ -20,7 +20,7 @@
 
 # General information about the project.
 version = os.environ["BUILD_VERSION"]
-envoyVersion = os.environ["ENVOY_VERSION"]
```
zirain, line 23: why build doc for a released version?

Xunzhuo: Changes could be backport to a released version, it is common for building versioned docs.

PR #974, opened 2023-02-03
xds: Deprecated http2_protocol_options

zirain:

fixes: #968


internal/xds/translator/authentication.go
```diff
@@ -219,11 +219,11 @@ func buildClusterFromJwks(jwks *jwksCluster) (*cluster.Cluster, error) {
 				},
 			},
 		},
-		Http2ProtocolOptions: &core.Http2ProtocolOptions{},
```
arkodg, line 222: unrelated to this PR, but curious why the auth cluster is HTTP2 @danehans ?

arkodg (approved):
LGTM thanks !

internal/xds/translator/authentication.go
danehans: @arkodg I was following the rate limit cluster [specs](https://github.com/envoyproxy/gateway/blob/v0.3.0-rc.1/internal/xds/translator/ratelimit.go#L295). If this is unneeded, can you submit an issue and assign me or create a PR to fix so it can be in the v0.3 release?

arkodg: for RL, we know for sure that the endpoint is a gRPC endpoint supporting HTTP2, for this case we cannot be sure, will raise a PR to change it

PR #980, opened 2023-02-04
Adds API Docs Tooling

### LINKED ISSUE #965: HTML generated docs for Gateway API Extensions
*Description*:

Starting with v0.3.0, Envoy Gateway will support a custom set of APIs / GatewayAPI based Extensions. The complete API reference should be published on the docs website

https://github.com/envoyproxy/gateway/tree/main/api

danehans:
Uses `github.com/elastic/crd-ref-docs` to generate API docs.

Fixes #965


danehans:
I'm not 100% happy with the generated output, but this can serve as a starting point for the API docs.

arkodg (dismissed):
thanks for building this 
the code looks fine, SGTM
here's another tool https://github.com/ahmetb/gen-crd-api-reference-docs that seems to be used by upstream https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/api-docs/generate.sh to generate api docs

zirain:
> thanks for building this the code looks fine, SGTM here's another tool https://github.com/ahmetb/gen-crd-api-reference-docs that seems to be used by upstream https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/api-docs/generate.sh to generate api docs

I tried with that last night, almost completed.

danehans:
> here's another tool https://github.com/ahmetb/gen-crd-api-reference-docs that seems to be used by upstream https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/api-docs/generate.sh to generate api docs

@arkodg I looked at ^ tool first but settled on crd-reference-documentation-generator since it seemed better supported and for [these reasons](https://github.com/elastic/crd-ref-docs#crd-reference-documentation-generator). Although the generator is fast, I'm not too impressed by the outcome and I still don't understand why the generator is having issues with the `HeaderMatch` type.

arkodg:
> > here's another tool https://github.com/ahmetb/gen-crd-api-reference-docs that seems to be used by upstream https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/api-docs/generate.sh to generate api docs
> 
> @arkodg I looked at ^ tool first but settled on crd-reference-documentation-generator since it seemed better supported and for [these reasons](https://github.com/elastic/crd-ref-docs#crd-reference-documentation-generator). Although the generator is fast, I'm not too impressed by the outcome and I still don't understand why the generator is having issues with the `HeaderMatch` type.

SGTM thanks for explaining, the lint error is due to https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md051---link-fragments-should-be-valid
and it looks like there should be a valid title for `headermatch` which doesnt exist

zirain:
FYI, https://github.com/ahmetb/gen-crd-api-reference-docs generate a bad output when using type like `map[LogComponent]LogLevel`

danehans:
@zirain thanks for the information. The lint issue with this PR is due to the `HeaderMatch` struct which is not a map.

zirain:
send a PR with my work on api  doc https://github.com/envoyproxy/gateway/pull/996

Xunzhuo (approved):
I prefer this one, LGTM, but need to resolve this lint error, or just try to ignore it.

zirain (dismissed):
this one is better than https://github.com/envoyproxy/gateway/pull/996

zirain:
image

this part is better!

zirain (approved):
lgtm, thanks!

PR #983, opened 2023-02-04
Plug in rate limit service URL into xds cluster

arkodg:

### DIFF
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -401,7 +401,7 @@ func expectedRateLimitContainers(infra *ir.RateLimitInfra) []corev1.Container {
 				},
 				{
 					Name:  "LOG_LEVEL",
-					Value: "debug",
+					Value: "info",
 				},
 				{
 					Name:  "USE_STATSD",
--- a/internal/infrastructure/kubernetes/deployment.go
+++ b/internal/infrastructure/kubernetes/deployment.go
@@ -53,10 +53,10 @@ const (
 
 	// rateLimitInfraName is the name for rate-limit resources.
 	rateLimitInfraName = "envoy-ratelimit"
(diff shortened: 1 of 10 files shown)


internal/infrastructure/kubernetes/deployment.go
```diff
@@ -401,7 +401,7 @@ func expectedRateLimitContainers(infra *ir.RateLimitInfra) []corev1.Container {
 				},
 				{
 					Name:  "LOG_LEVEL",
-					Value: "info",
+					Value: "debug",
```
danehans, line 404: Why is this log level now debug?

arkodg: adds verbosity for this new feature, which can help triaging future issues, can move it to `info` if that is the preference for default

arkodg: reverted to `info`

internal/infrastructure/kubernetes/service.go
```diff
@@ -151,6 +152,11 @@ func (i *Infra) expectedRateLimitService(_ *ir.RateLimitInfra) *corev1.Service {
 	return svc
 }
 
+// GetRateLimitServiceURL returns the URL for the rate limit service.
+func GetRateLimitServiceURL(namespace string) string {
+	return "grpc://" + rateLimitInfraName + "." + namespace + ".svc.cluster.local:" + strconv.Itoa(rateLimitInfraGRPCPort)
```
zirain, line 157: prefer to user `fmt.Printf`

arkodg: ack

internal/infrastructure/kubernetes/deployment.go
danehans: Thanks for reverting to info. I think it makes sense to have a consistent logging level across EG components.

danehans (approved):

Issue #995, opened 2023-02-08
Enable HTTPRouteMatching conformance test


arkodg:
*Description*:
Seeing this fail in https://github.com/envoyproxy/gateway/pull/780

Once fixed they should be enabled via https://github.com/envoyproxy/gateway/blob/main/test/conformance/conformance_test.go

danehans:
I see this test pass with GW API v0.6.0 when run locally.

danehans:
I see this test fail in v0.6.1 due to these test [changes](https://github.com/kubernetes-sigs/gateway-api/commit/070a6b65bcdc6fbc85f69b4882de2569c9e477b7) upstream.

PR #1004, opened 2023-02-08
Use `path_separated_prefix` Route match

arkodg:
We were using `prefix` Route match in xds for
the `PathPrefix` Gateway API match defined here
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPPathMatch

The match translates to a `path_separated_prefix` instead https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routematch

This should fix the `HTTPRouteMatching` conformance test where requests with `/v2example` should match `/` and not `/v2`

Fixes: https://github.com/envoyproxy/gateway/issues/995


zirain (approved):
lgtm

PR #1015, opened 2023-02-09
translator: refactor authn and ratelimit to use buildXdsCluster

### LINKED ISSUE #979: Refactor authn and ratelimit features to reuse buildXdsCluster
*Description*:
The ratelimit feature creates an xds cluster using https://github.com/envoyproxy/gateway/blob/4a04915e3f8b95ac54baeed6d42503324fb7af9b/internal/xds/translator/ratelimit.go#L267 and the authn feature creates an xds cluster using https://github.com/envoyproxy/gateway/blob/4a04915e3f8b95ac54baeed6d42503324fb7af9b/internal/xds/translator/authentication.go#L413

These features should instead be using the existing method to build an xds cluster defined here
https://github.com/envoyproxy/gateway/blob/4a04915e3f8b95ac54baeed6d42503324fb7af9b/internal/xds/translator/cluster.go#L22 so its easier to manage the intended cluster field values

[optional *Relevant Links*:]

sbko:
Fixes #979

### DIFF
--- a/internal/xds/translator/authentication.go
+++ b/internal/xds/translator/authentication.go
@@ -15,8 +15,8 @@ import (
 	"time"
 
 	cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
-	endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	routev3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	jwtext "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/jwt_authn/v3"
@@ -195,34 +195,25 @@ func buildClusterFromJwks(jwks *jwksCluster) (*cluster.Cluster, error) {
 		return nil, errors.New("jwks is nil")
 	}
 
-	endpoints, err := jwks.toLbEndpoints()
+	destinations, err := jwks.toRouteDestinations()
 	if err != nil {
 		return nil, err
 	}
+
+	cluster := buildXdsCluster(jwks.name, destinations, false /*isHTTP2 */)
+	cluster.ConnectTimeout = durationpb.New(10 * time.Second)
+	cluster.DnsRefreshRate = durationpb.New(30 * time.Second)
+	cluster.LbPolicy = clusterv3.Cluster_RANDOM
+	cluster.RespectDnsTtl = true
 
 	tSocket, err := buildXdsUpstreamTLSSocket()
 	if err != nil {
 		return nil, err
 	}
 
-	return &cluster.Cluster{
-		Name:                 jwks.name,
-		ClusterDiscoveryType: &cluster.Cluster_Type{Type: cluster.Cluster_STATIC},
-		ConnectTimeout:       durationpb.New(10 * time.Second),
-		LbPolicy:             cluster.Cluster_RANDOM,
-		LoadAssignment: &endpoint.ClusterLoadAssignment{
-			ClusterName: jwks.name,
-			Endpoints: []*endpoint.LocalityLbEndpoints{
-				{
-					LbEndpoints: endpoints,
-				},
-			},
-		},
-		DnsRefreshRate:  durationpb.New(30 * time.Second),
-		RespectDnsTtl:   true,
-		DnsLookupFamily: cluster.Cluster_V4_ONLY,
-		TransportSocket: tSocket,
-	}, nil
+	cluster.TransportSocket = tSocket
+
+	return cluster, nil
 }
 
 // buildXdsUpstreamTLSSocket returns an xDS TransportSocket that uses envoyTrustBundle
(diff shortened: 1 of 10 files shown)


internal/xds/translator/authentication.go
```diff
@@ -15,8 +15,8 @@ import (
 	"time"
 
 	cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
```
danehans, line 18: This pkg is already being imported as `cluster`.

internal/xds/translator/authentication.go
```diff
@@ -195,34 +195,25 @@ func buildClusterFromJwks(jwks *jwksCluster) (*cluster.Cluster, error) {
...
+	destinations, err := jwks.toRouteDestinations()
 	if err != nil {
 		return nil, err
 	}
 
+	cluster := buildXdsCluster(jwks.name, destinations, false /*isHTTP2 */)
+	cluster.ConnectTimeout = durationpb.New(10 * time.Second)
+	cluster.DnsRefreshRate = durationpb.New(30 * time.Second)
+	cluster.LbPolicy = clusterv3.Cluster_RANDOM
+	cluster.RespectDnsTtl = true
```
danehans, line 207: I believe RespectDnsTtl can be dropped since the cluster is static.

As a follow-on PR, it would be nice to have a type that represents cluster config params, e.g. connection timeout, destinations, etc., that can be instantiated and passed into buildXdsCluster().

internal/xds/translator/ratelimit.go
```diff
@@ -269,39 +268,19 @@ func (t *Translator) buildRateLimitServiceCluster(irListener *ir.HTTPListener) *
...
-		DnsRefreshRate:                durationpb.New(30 * time.Second),
-		RespectDnsTtl:                 true,
-		DnsLookupFamily:               cluster.Cluster_V4_ONLY,
-	}
+	var routeDestinations []*ir.RouteDestination
+	routeDestinations = append(routeDestinations, &ir.RouteDestination{
+		Host: host,
+		Port: uint32(port),
+	})
+
```
danehans, line 276:
```
routeDestinations := []*ir.RouteDestination{ir.NewRouteDest(host, port)}
```
```
pkg ir

func NewRouteDest(host string, port int) *ir.RouteDestination {
return &ir.RouteDestination{
		Host: host,
		Port: uint32(port),
	}
}

File: internal/xds/translator/authentication.go [hunk 4]
```diff
@@ -195,34 +195,25 @@ func buildClusterFromJwks(jwks *jwksCluster) (*cluster.Cluster, error) {
...
 
-	endpoints, err := jwks.toLbEndpoints()
+	destinations, err := jwks.toRouteDestinations()
 	if err != nil {
 		return nil, err
 	}
 
+	cluster := buildXdsCluster(jwks.name, destinations, false /*isHTTP2 */)
+	cluster.ConnectTimeout = durationpb.New(10 * time.Second)
+	cluster.DnsRefreshRate = durationpb.New(30 * time.Second)
```
danehans, line 205: I believe this field can be removed since the cluster type is static or you can try moving the JWKS cluster type to DNS and remove the code that converts the JWKS URL hostname to IP addresses.

internal/xds/translator/authentication.go
```diff
@@ -195,34 +195,25 @@ func buildClusterFromJwks(jwks *jwksCluster) (*cluster.Cluster, error) {
...
 	}
 
-	endpoints, err := jwks.toLbEndpoints()
+	destinations, err := jwks.toRouteDestinations()
 	if err != nil {
 		return nil, err
 	}
 
+	cluster := buildXdsCluster(jwks.name, destinations, false /*isHTTP2 */)
+	cluster.ConnectTimeout = durationpb.New(10 * time.Second)
```
danehans, line 204: @arkodg I followed the rate limit cluster connection timeout settings for the JWKS cluster. Should both clusters stick with `ConnectTimeout` setting from buildXdsCluster()?

danehans (changes requested):

danehans:
Moving this to 0.4 since it's not fixing a v0.3 release blocker bug.

internal/xds/translator/authentication.go
arkodg: @sbko can all common settings such as `ConnectTimeout`  live inside `buildXdsCluster` ? the goal is to not have the caller set these fields unless they want to override it

sbko: @arkodg `ConnectTimeout` already set inside `buildXdsCluster`, but default value is 5 seconds. should I change default value to 10 seconds?

arkodg: I think 5 is fine. I dont think the ratelimit or auth cluster callers need to perform any overrides on the returned cluster val.
Another thing you will need to handle is
```
ClusterDiscoveryType: &cluster.Cluster_Type{Type: cluster.Cluster_STRICT_DNS}
```
if the endpoints are not static IPs we need to set above field along with other DNS settings mentioned in this PR
the easiest way to do it now would be to extend `buildXdsCluster` and add a `bool isStatic` for now, and set that for the auth and ratelimit callers.

internal/xds/translator/authentication.go
arkodg: thinking about this again, if the API supports an IP addr as a Hostname then maybe this layer can decide based on the str and create a static or dynamic cluster, dont think this needs to be part of this PR since its out of scope from the issue this PR is fixing

sbko: changed JWKS cluster type to STRICT_DNS

arkodg: I dont think STRICT_DNS is right for this case (sorry if I said the opposite before)
from the API spec https://gateway.envoyproxy.io/v0.3.0/api/extension_types.html#remotejwks
the hostname section within the URI could be an IP Addr or a FQDN. 
https://github.com/envoyproxy/gateway/blob/17d325d3ff59e04a045fff49a1edaafe3f633314/internal/xds/translator/authentication.go#L480
current converts hostname into IP string, so this cluster must be STATIC

In the future this code case could be enhanced to eliminate resolveHostname and decide discovery type based on Hostname type

arkodg (approved):
LGTM, thanks for addressing the comments !

arkodg:
@danehans can you ptal ?

danehans (approved):

PR #1019, opened 2023-02-09
remove empty route error check in auth xds logic

arkodg:
the Xds translator should not error out when the IR routes are empty

### DIFF
--- a/internal/xds/translator/authentication.go
+++ b/internal/xds/translator/authentication.go
@@ -48,10 +48,6 @@ func patchHCMWithJwtAuthnFilter(mgr *hcm.HttpConnectionManager, irListener *ir.H
 		return errors.New("ir listener is nil")
 	}
 
-	if len(irListener.Routes) == 0 {
-		return errors.New("ir listener contains no routes")
-	}
-
 	if !listenerContainsJwtAuthn(irListener) {
 		return nil
 	}


arkodg:
impact of keeping this check is, the Xds translator errors out when the IR routes are empty. This might alarm the user
but the routes are empty because they might not be available yet or because they are invalid are not set by the gateway api translator layer
```
failed to translate xds ir	{"runner": "xds-translator", "error": "ir listener contains no routes"}
```

arkodg:
@skriss @AliceProxy @danehans @youngnick @Xunzhuo @zirain can you ptal, I think we should add this into v0.3.0

internal/xds/translator/authentication.go
```diff
@@ -48,10 +48,6 @@ func patchHCMWithJwtAuthnFilter(mgr *hcm.HttpConnectionManager, irListener *ir.H
 		return errors.New("ir listener is nil")
 	}
 
-	if len(irListener.Routes) == 0 {
-		return errors.New("ir listener contains no routes")
```
danehans, line 52: Instead of returning an error, the func should just return early since other parts of the code path are expecting the listener to contain routes. Otherwise, please update the other authn funcs to check for `irListener.Routes` and return early.

arkodg: looks like its being handled already https://github.com/envoyproxy/gateway/blob/38a6dddd21efd0edb1deade116d771f8252e1ad7/internal/xds/translator/authentication.go#L515

danehans (approved):

PR #1026, opened 2023-02-10
Updates Gateway API Support Doc

danehans:
Updates the Gateway API support doc.


danehans:
Marking as "draft" to wait for the v0.3 docs to be published so the content can be added to the versioned docs.

zirain (approved):

PR #1028, opened 2023-02-10
fix: validate user config

tony612:
Signed-off-by: Bing Han <h.bing612@gmail.com>

Move the cfg.Validate below to validate user config. Without this fix, the test will fail:
```
--- FAIL: TestValidateUserConfig (0.00s)
    server_test.go:33:
        	Error Trace:	/Users/hanbing/repo/go/envoyproxy-gateway/internal/cmd/server_test.go:33
        	Error:      	An error is expected but got nil.
        	Test:       	TestValidateUserConfig
FAIL
FAIL	github.com/envoyproxy/gateway/internal/cmd	0.601s
FAIL
```

### DIFF
--- a/internal/cmd/certgen.go
+++ b/internal/cmd/certgen.go
@@ -35,7 +35,7 @@ func getCertGenCommand() *cobra.Command {
 
 // certGen generates control plane certificates.
 func certGen() error {
-	cfg, err := getConfig(cfgPath)
+	cfg, err := getConfig()
 	if err != nil {
 		return err
 	}
--- a/internal/cmd/certgen.go
+++ b/internal/cmd/certgen.go
@@ -35,7 +35,7 @@ func getCertGenCommand() *cobra.Command {
 
 // certGen generates control plane certificates.
 func certGen() error {
-	cfg, err := getConfig()
+	cfg, err := getConfig(cfgPath)
 	if err != nil {
 		return err
 	}
--- a/internal/cmd/server_test.go
+++ b/internal/cmd/server_test.go
@@ -12,23 +12,46 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	invalidGatewayConfig = `
+kind: EnvoyGateway
+apiVersion: config.gateway.envoyproxy.io/v1alpha1
+gateway: {}
+`
+)
+
 func TestGetServerCommand(t *testing.T) {
 	got := getServerCommand()
 	assert.Equal(t, "server", got.Use)
 }
 
 func TestGetConfigValidate(t *testing.T) {
-	file, err := os.CreateTemp("", "config")
-	assert.NoError(t, err)
-	defer os.Remove(file.Name())
+	tests := []struct {
+		name   string
+		input  string
+		errors []string
+	}{
+		{
+			name:   "invalid gateway",
+			input:  invalidGatewayConfig,
+			errors: []string{"is unspecified"},
+		},
+	}
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			file, err := os.CreateTemp("", "config")
+			assert.NoError(t, err)
+			defer os.Remove(file.Name())
 
-	_, err = file.Write([]byte(`
-kind: EnvoyGateway
-apiVersion: config.gateway.envoyproxy.io/v1alpha1
-gateway: {}
-`))
-	assert.NoError(t, err)
+			_, err = file.Write([]byte(test.input))
+			assert.NoError(t, err)
 
-	_, err = getConfigByPath(file.Name())
-	assert.Error(t, err)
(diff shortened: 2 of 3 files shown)


internal/cmd/certgen.go
```diff
@@ -35,7 +35,7 @@ func getCertGenCommand() *cobra.Command {
 
 // certGen generates control plane certificates.
 func certGen() error {
-	cfg, err := getConfig()
+	cfg, err := getConfig(cfgPath)
```
arkodg, line 38: looks like cfgPath is a global var in the `cmd` package, dont understand the issue here

tony612: In getConfig test, I need to create a temp file and use the file path to get the config. I add a argument to `getConfig` so that I can pass the temp file path to `getConfig`. In this way, the global `cfgPath` won't be polluted by the test.

tony612: Or maybe I can define a new function like `getConfigByPath()` and call it in `getConfig` with the global var. Then I can test `getConfigByPath` in test and the cmd package can still use `getConfig`.

tony612: @arkodg Any idea?

arkodg: can you help understand what this PR is `fixing`, or is it just increasing test coverage for the package ?

tony612: `cfg.Validate()` is called just after `config.New()` before this fix and the user config won't be validated if `cfgPath` is provided via `--config-path`. After the fix, `getConfig` will return an error if user uses an invalid config.

arkodg: why not leave the current API as is, and just set the global var `cfgPath` in the Test ?

internal/cmd/server_test.go
```diff
@@ -15,3 +16,19 @@ func TestGetServerCommand(t *testing.T) {
 	got := getServerCommand()
 	assert.Equal(t, "server", got.Use)
 }
+
+func TestValidateUserConfig(t *testing.T) {
```
arkodg, line 20: would be great if the test looks like https://github.com/envoyproxy/gateway/blob/3c69c36f6de255d0532cb6a1ca245a4494ff4124/internal/ir/xds_test.go#L570 so its easier to add more test cases in the future

internal/cmd/certgen.go
tony612: > why not leave the current API as is, and just set the global var `cfgPath` in the Test ?

I just don't want to pollute the global var in the tests(maybe due to my unimportant habit in functional languages XD), and this may cause other tests to fail if `cfgPath` is also used in other tests in the future. Yeah, I can do this and recover it after the test, this is fine in theory.

arkodg: can you delete the global var `cfgPath`, make it local to the func and instead pass by value everywhere ?

tony612: I think we can still keep it global now because it's used as a flag and it will be hard to use in non-test code. We can do this after some refactor like wrapping all flags in a struct then change functions like `server()` to its methods. For this PR, I think we can just do what I said before:

> Or maybe I can define a new function like getConfigByPath() and call it in getConfig with the global var. Then I can test getConfigByPath in test and the cmd package can still use getConfig.

btw, I found another small thing: `getConfig` is used in `certGen`, but `cfgPath` is not registered as a flag. I'm not sure if this is a problem.

tony612: I updated the code as I said. Please let me know if there's anything you feel not good. Thanks!

internal/cmd/server_test.go
```diff
@@ -6,12 +6,52 @@
...
 }
+
+func TestGetConfigValidate(t *testing.T) {
+	tests := []struct {
+		name   string
+		input  string
+		errors []string
+	}{
+		{
+			name:   "invalid gateway",
```
arkodg, line 35: can you also add a good case ?
here's an example
https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/config/envoy-gateway/envoy-gateway.yaml

internal/cmd/certgen.go
arkodg: the changes are not ideal, but fine for this PR (would have ideally liked to get rid of the global var so it doesnt confuse the next dev)

arkodg: but looks like the way the code is today, with `certGen` calling `getConfig`, its probably easier to keep it as a global var.
thanks again for taking the time to explain the fix and addressing comments

internal/cmd/server_test.go
tony612: Done. Thanks for reviewing!

arkodg (approved):

arkodg:
LGTM, thanks !

Issue #998, opened 2023-02-08
ensure `use_remote_address` is set


zirain:
follow up: https://github.com/envoyproxy/gateway/pull/946#discussion_r1093556536

PR #1029, opened 2023-02-10
ensure `use_remote_address` is set

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/998

cc @arkodg

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -22,6 +22,7 @@ import (
 	tls "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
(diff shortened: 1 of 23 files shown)


internal/xds/translator/listener.go
```diff
@@ -86,6 +87,7 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listener.Listener, irLis
 				RouteConfigName: irListener.Name,
 			},
 		},
+		UseRemoteAddress: &wrappers.BoolValue{Value: true},
```
arkodg, line 90: can we add this link https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for in the code for reference ?

arkodg (approved):
LGTM thanks ! left a nit

internal/xds/translator/listener.go
zirain: done

danehans (approved):

arkodg (approved):

Issue #1034, opened 2023-02-11
test flaky: DATA RACE


zirain:
https://github.com/envoyproxy/gateway/actions/runs/4149211743/jobs/7177957701

PR #1039, opened 2023-02-13
dont run k8s infra tests in parallel

arkodg:
* the k8s infra tests are performing CRUD ops on the same k8s env/cluster. They shouldnt be run in parallel as each test might affect the other.

* tested using `go test ./... -count 50 -race -covermode=atomic` which fails on the current `main` branch

Fixes: https://github.com/envoyproxy/gateway/issues/1034


zirain (approved):
LGTM, thanks!

Issue #31, opened 2022-04-28
Override default Envoy Proxy Bootstrap Configuration


arkodg:
https://github.com/envoyproxy/gateway/issues/24 highlights the need for users to override xDS resources configured by
Envoy Gateway. Raising this issue to track how do you express this intent for the [Envoy Proxy Bootstrap Config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#config-bootstrap-v3-bootstrap) using the[ Gateway API](https://gateway-api.sigs.k8s.io)

here's how popular OSS projects do it
* Istio - [this](https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter-ApplyTo) & [this](https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter-ApplyTo)
* Contour allows the user to [attach a new bootstrap file](https://projectcontour.io/docs/v1.20.1/configuration/#bootstrap-config-file)
* Emissary allows the user to override a [subset of the bootstrap resources](https://www.getambassador.io/docs/emissary/2.2/topics/running/statistics/envoy-statsd/#envoy-statistics-with-statsd) 

Updated
- [x] Design doc https://github.com/envoyproxy/gateway/pull/1070
- [x] API Design https://github.com/envoyproxy/gateway/pull/1070
- [x] EnvoyProxy Bootstrap validation https://github.com/envoyproxy/gateway/pull/1109
- [x] Implement API https://github.com/envoyproxy/gateway/pull/1094
- [x] Update GatewayClass Status  based on EnvoyProxy validation https://github.com/envoyproxy/gateway/pull/1107
- [x] Implement `egctl` subcommand to generate default `bootstrap` https://github.com/envoyproxy/gateway/pull/1281
- [x] Docs https://github.com/envoyproxy/gateway/pull/1333

danehans:
> Raising this issue to track how do you express this intent for the [Envoy Proxy Bootstrap Config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#config-bootstrap-v3-bootstrap) using the[ Gateway API](https://gateway-api.sigs.k8s.io/)

[...]

PR #1047, opened 2023-02-15
egctl x translate

arkodg:
* add an `experimental` subcommand called `translate` allowing users to observe the translated xds output when gateway api resources are inputted

* this should help users understand the mapping between gatewapi resources and xds resources which will help with https://github.com/envoyproxy/gateway/issues/24 and https://github.com/envoyproxy/gateway/issues/31


internal/cmd/egctl/experimental.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewExperimentalCommand() *cobra.Command {
+
+	experimentalCommand := &cobra.Command{
+		Use:     "experimental",
+		Aliases: []string{"x"},
+		Short:   "Experimental features",
```
zirain, line 17: great! 

can you update the `egctl` doc, `Experimental features do not guarantee compatibility and stability.`

arkodg: sure will do, will also add a user facing `egctl` doc along with this PR once ready

Xunzhuo (commented):
This is great, I like this feature.

Xunzhuo (approved):
LGTM! Thanks!

PR #1050, opened 2023-02-16
gatewayapi: adds matching on GRPCRoute method and service

### LINKED ISSUE #963: Add support for service method match in GRPCRoute
*Description*:
Support the service method match defined in the GRPCRoute
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1alpha2.GRPCMethodMatch

Should be fixed here https://github.com/envoyproxy/gateway/blob/f3b77142752fd485747c4684984d99e82dd8831d/internal/gatewayapi/route.go#L404

sbko:
Closes #963

### DIFF
--- a/docs/latest/user/grpc-routing.md
+++ b/docs/latest/user/grpc-routing.md
@@ -1,6 +1,6 @@
 # GRPC Routing
 
-The [GRPCRoute][] resource allows users to configure gRPC routing by matching HTTP/2 traffic and forwarding it to backend gRPC servers. 
+The [GRPCRoute][] resource allows users to configure gRPC routing by matching HTTP/2 traffic and forwarding it to backend gRPC servers.
 To learn more about gRPC routing, refer to the [Gateway API documentation][].
 
 ## Prerequisites
@@ -88,6 +88,52 @@ Envoy Gateway also supports [gRPC-Web][] requests for this configuration. The be
 
 ```shell
 curl --http2-prior-knowledge -s ${GATEWAY_HOST}:80/yages.Echo/Ping -H 'Host: grpc-example.com'   -H 'Content-Type: application/grpc-web-text'   -H 'Accept: application/grpc-web-text' -XPOST -d'AAAAAAA=' | base64 -d
+```
+
+## GRPCRoute Match
+The `matches` field can be used to restrict the route to a specific set of requests based on GRPC's service and method names.
+The following example shows how to match requests for any method in the `yages.Echo` service and requests for the `ServerReflectionInfo` method in any service.
+
+```shell
+cat <<EOF | kubectl apply -f -
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: GRPCRoute
+metadata:
+  name: yages
+  labels:
+    example: grpc-routing
+spec:
+  parentRefs:
+    - name: example-gateway
+  hostnames:
+    - "grpc-example.com"
+  rules:
+    - matches:
+      - method:
+          method: ServerReflectionInfo
+          type: RegularExpression
+      - method:
+          service: yages.Echo
+          type: Exact
+      backendRefs:
+        - group: ""
+          kind: Service
+          name: yages
+          port: 9000
+          weight: 1
+EOF
+```
+
+Verify the GRPCRoute status:
+
+```shell
+kubectl get grpcroutes --selector=example=grpc-routing -o yaml
+```
+
+Test GRPC routing to the `yages` backend using the [grpcurl][] command.
+
+```shell
+grpcurl -plaintext -authority=grpc-example.com ${GATEWAY_HOST}:80 yages.Echo/Ping
 ```
 
 
--- a/docs/latest/user/grpc-routing.md
+++ b/docs/latest/user/grpc-routing.md
@@ -91,8 +91,9 @@ curl --http2-prior-knowledge -s ${GATEWAY_HOST}:80/yages.Echo/Ping -H 'Host: grp
 ```
 
 ## GRPCRoute Match
-The `matches` field can be used to restrict the route to a specific set of requests based on GRPC's service and method names.
-The following example shows how to match requests for any method in the `yages.Echo` service and requests for the `ServerReflectionInfo` method in any service.
+The `matches` field can be used to restrict the route to a specific set of requests based on GRPC's service and/or method names.
+The following example shows how to match a request based on the service and method names for `grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo`,
+as well as a regular expression match for all services with a method name `Ping` which matches `yages.Echo/Ping` in our deployment.
 
 ```shell
 cat <<EOF | kubectl apply -f -
@@ -111,10 +112,11 @@ spec:
     - matches:
       - method:
           method: ServerReflectionInfo
-          type: RegularExpression
-      - method:
-          service: yages.Echo
+          service: grpc.reflection.v1alpha.ServerReflection
           type: Exact
+      - method:
+          method: Ping
+          type: RegularExpression
       backendRefs:
         - group: ""
           kind: Service
--- a/docs/latest/user/grpc-routing.md
+++ b/docs/latest/user/grpc-routing.md
@@ -93,7 +93,9 @@ curl --http2-prior-knowledge -s ${GATEWAY_HOST}:80/yages.Echo/Ping -H 'Host: grp
 ## GRPCRoute Match
 The `matches` field can be used to restrict the route to a specific set of requests based on GRPC's service and/or method names.
 The following example shows how to match a request based on the service and method names for `grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo`,
-as well as a regular expression match for all services with a method name `Ping` which matches `yages.Echo/Ping` in our deployment.
+as well as a match for all services with a method name `Ping` which matches `yages.Echo/Ping` in our deployment.
+
+Current implementation supports only `Exact` match. `RegularExpression` match will be supported after https://github.com/kubernetes-sigs/gateway-api/issues/1746 is resolved.
 
 ```shell
 cat <<EOF | kubectl apply -f -
@@ -113,10 +115,8 @@ spec:
       - method:
           method: ServerReflectionInfo
           service: grpc.reflection.v1alpha.ServerReflection
-          type: Exact
       - method:
           method: Ping
-          type: RegularExpression
       backendRefs:
         - group: ""
           kind: Service
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
 		}
 
 		if match.Method != nil {
-			if match.Method.Method != nil {
-				irRoute.HeaderMatches = append(irRoute.HeaderMatches, &ir.StringMatch{
-					Name:  ":method",
-					Exact: match.Method.Method,
-				})
+			// GRPC's path is in the form of "/<service>/<method>"
+			//nolint:gocritic
+			switch GRPCMethodMatchTypeDerefOr(match.Method.Type, v1alpha2.GRPCMethodMatchExact) {
+			case v1alpha2.GRPCMethodMatchExact:
+				if match.Method.Service != nil && match.Method.Method != nil {
+					irRoute.PathMatch = &ir.StringMatch{
+						Exact: StringPtr(fmt.Sprintf("/%s/%s", *match.Method.Service, *match.Method.Method)),
+					}
+				} else if match.Method.Service != nil {
+					// Use prefix match if only the service name is specified
+					irRoute.PathMatch = &ir.StringMatch{
+						Prefix: StringPtr(fmt.Sprintf("/%s/", *match.Method.Service)),
+					}
+				} else if match.Method.Method != nil {
+					// Use regex match if only the method name is specified to match any service
+					irRoute.PathMatch = &ir.StringMatch{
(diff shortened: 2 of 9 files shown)


internal/gatewayapi/route.go
```diff
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
-			if match.Method.Method != nil {
-				irRoute.HeaderMatches = append(irRoute.HeaderMatches, &ir.StringMatch{
-					Name:  ":method",
-					Exact: match.Method.Method,
-				})
-			}
-			/* TODO
-			if match.Method.Service != nil {
+			// GRPC's path is in the form of "/<service>/<method>"
+			//nolint:gocritic
```
arkodg, line 394: why was the nolint needed ?

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+			switch GRPCMethodMatchTypeDerefOr(match.Method.Type, v1alpha2.GRPCMethodMatchExact) {
+			case v1alpha2.GRPCMethodMatchExact:
+				if match.Method.Service != nil && match.Method.Method != nil {
```
arkodg, line 397: this code can be simplified with 
```
if match.Method.Method != nil {
  pathMatch += XX
}
if match.Method.Service != nil {
  pathMatch += XX
}

irRoute.PathMatch = pathMatch
```

arkodg:
hey @sbko thanks for picking this up, can you incorporate one of these matches into the user doc https://github.com/envoyproxy/gateway/blob/main/docs/latest/user/grpc-routing.md, maybe add a service match for `yages.Ping`

internal/gatewayapi/route.go
arkodg: or maybe it can't because of the regex needed for `Service` ?

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+					}
+				} else if match.Method.Service != nil {
+					// Use prefix match if only the service name is specified
+					irRoute.PathMatch = &ir.StringMatch{
+						Prefix: StringPtr(fmt.Sprintf("/%s/", *match.Method.Service)),
+					}
+				} else if match.Method.Method != nil {
+					// Use regex match if only the method name is specified to match any service
+					irRoute.PathMatch = &ir.StringMatch{
+						SafeRegex: StringPtr(fmt.Sprintf("/[.a-zA-Z0-9-]+/%s", *match.Method.Method)),
```
arkodg, line 409: are digits allowed, shouldnt it be `^[a-zA-Z]+$` so it starts and ends with a char ?
also would help to get this tested with a `saferegex` tool

internal/gatewayapi/route.go
sbko: @arkodg we need to use regex if only method is specified, so this wouldn't work.

internal/gatewayapi/route.go
sbko: add nolint to ignore 'Error: ifElseChain: rewrite if-else to switch statement (gocritic)'

internal/gatewayapi/route.go
sbko: @arkodg digits are allowed. I checked with saferegex there is an issue in current state. I can fix it like this "^/[.a-zA-Z0-9-]+/%s". what do you think?

arkodg: why is there a `-` in the end ?

sbko: It's a typo, should be `_`. I'm going to push a change based on what K8s is doing to validate those fields.

sbko:
> hey @sbko thanks for picking this up, can you incorporate one of these matches into the user doc https://github.com/envoyproxy/gateway/blob/main/docs/latest/user/grpc-routing.md, maybe add a service match for `yages.Ping`

done

docs/latest/user/grpc-routing.md
```diff
@@ -90,6 +90,52 @@ Envoy Gateway also supports [gRPC-Web][] requests for this configuration. The be
...
+    example: grpc-routing
+spec:
+  parentRefs:
+    - name: example-gateway
+  hostnames:
+    - "grpc-example.com"
+  rules:
+    - matches:
+      - method:
+          method: ServerReflectionInfo
```
arkodg, line 113: shouldnt this be `Ping` ?

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+					}
+				} else if match.Method.Service != nil {
+					// Use prefix match if only the service name is specified
+					irRoute.PathMatch = &ir.StringMatch{
+						Prefix: StringPtr(fmt.Sprintf("/%s", *match.Method.Service)),
+					}
+				} else if match.Method.Method != nil {
+					// Use regex match if only the method name is specified to match any service
+					irRoute.PathMatch = &ir.StringMatch{
+						SafeRegex: StringPtr(fmt.Sprintf("^/(?i)\\.?[a-z_][a-z_0-9]*(\\.[a-z_][a-z_0-9]*)*/%s$", *match.Method.Method)),
```
arkodg, line 409: lets make this into a constant, so it can be reused, so will need to spend some cycles to ensure this regex looks right

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,40 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+					}
+				}
+			case v1alpha2.GRPCMethodMatchRegularExpression:
+				if match.Method.Service != nil && match.Method.Method != nil {
+					irRoute.PathMatch = &ir.StringMatch{
+						SafeRegex: StringPtr(fmt.Sprintf("^/%s/%s", *match.Method.Service, *match.Method.Method)),
```
arkodg, line 415: why is `^` needed ?

sbko: added for consistency between regexes. maybe slightly faster(?) :)

docs/latest/user/grpc-routing.md
sbko: in this case I'm matching for another service/method - grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo, but I see how this can be confused as an example. I'll change it to match for method and service

arkodg: if the example worked with existing config and a grpc call to `yages.Echo/Ping`, it makes me think that the regex string config  is incorrect

sbko: existing example matches a request to any service that has method `ServerReflectionInfo`(will not match `yages.Echo/Ping` ) OR a request to any method in a service `yages.Echo` which matches ` yages.Echo/Ping`. We need to match both because without matching ServerReflectionInfo you will get `Error invoking method "yages.Echo/Ping": failed to query for service descriptor "yages.Echo": server does not support the reflection API`. Probably cleaner way will be to exact match both methods and services, but I wanted to showcase Exact and RegularExpression types of matches.

arkodg: thanks for clarifying that ! the `exact` match makes more sense here imho

internal/gatewayapi/route.go
sbko: made it constant, refactored `if else if ` so no need for `nolint:gocritic` anymore. please let me know if you prefer previous structure.

internal/gatewayapi/route.go
```diff
@@ -26,6 +26,11 @@ type RoutesTranslator interface {
 	ProcessUDPRoutes(udpRoutes []*v1alpha2.UDPRoute, gateways []*GatewayContext, resources *Resources, xdsIR XdsIRMap) []*UDPRouteContext
 }
 
+const (
+	GRPCServiceRegexPattern = "(?i)\\.?[a-z_][a-z_0-9]*(\\.[a-z_][a-z_0-9]*)*"
```
arkodg, line 30: why are there two different types here, isnt the only use case needed is when we want to match on method but service is empty ?

internal/gatewayapi/route.go
```diff
@@ -26,6 +26,11 @@ type RoutesTranslator interface {
 	ProcessUDPRoutes(udpRoutes []*v1alpha2.UDPRoute, gateways []*GatewayContext, resources *Resources, xdsIR XdsIRMap) []*UDPRouteContext
 }
 
+const (
```
arkodg, line 29: @lizan is a smarter way to do path segment matching in envoy rather than relying on regex matching ?

internal/gatewayapi/route.go
sbko: we need `GRPMethodRegexPattern` to support a use case when `service` is specified without `method` and `type: RegularExpression`. envoy's prefix is used when `type` is empty or `type: Exact`

internal/gatewayapi/route.go
```diff
@@ -390,16 +395,33 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+			switch {
+			case methodType == v1alpha2.GRPCMethodMatchRegularExpression, match.Method.Service == nil:
+				// Use regex when the type is set to RegularExpression or when the service name is not specified
+				irRoute.PathMatch = &ir.StringMatch{
+					SafeRegex: StringPtr(fmt.Sprintf("^/%s/%s$", service, method)),
+				}
+			case match.Method.Method == nil:
+				// Use prefix match if only the service name is specified
+				irRoute.PathMatch = &ir.StringMatch{
+					Prefix: StringPtr(fmt.Sprintf("/%s", service)),
```
lizan, line 417: use path_separated_prefix or add `/` at the end of prefix?

internal/gatewayapi/route.go
lizan: From [GEP-1016](https://gateway-api.sigs.k8s.io/geps/gep-1016/#matcher-types) we can use suffix match (with header matcher)?

internal/gatewayapi/route.go
sbko: it uses path_separated_prefix

docs/latest/user/grpc-routing.md
```diff
@@ -90,6 +90,54 @@ Envoy Gateway also supports [gRPC-Web][] requests for this configuration. The be
...
+## GRPCRoute Match
+The `matches` field can be used to restrict the route to a specific set of requests based on GRPC's service and/or method names.
+The following example shows how to match a request based on the service and method names for `grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo`,
+as well as a match for all services with a method name `Ping` which matches `yages.Echo/Ping` in our deployment.
+
+Current implementation supports only `Exact` match. `RegularExpression` match will be supported after https://github.com/kubernetes-sigs/gateway-api/issues/1746 is resolved.
+
+```shell
+cat <<EOF | kubectl apply -f -
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
arkodg, line 102: due to this bug https://github.com/kubernetes-sigs/gateway-api/issues/1751, the default config will no longer work now that we are parsing `method`
can you replace the original config in https://github.com/envoyproxy/gateway/blob/main/examples/kubernetes/grpc-routing.yaml with this one (which has the explicit method match)

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,23 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
 		}
 
 		if match.Method != nil {
-			if match.Method.Method != nil {
+			// GRPC's path is in the form of "/<service>/<method>"
+			// TODO: support regex match type after https://github.com/kubernetes-sigs/gateway-api/issues/1746 is resolved
```
arkodg, line 394: can you raise a GH issue within EG so its easier for this project to track it

internal/gatewayapi/route.go
```diff
@@ -390,16 +390,23 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+			switch {
+			case match.Method.Service != nil && match.Method.Method != nil:
+				irRoute.PathMatch = &ir.StringMatch{
+					Exact: StringPtr(fmt.Sprintf("/%s/%s", *match.Method.Service, *match.Method.Method)),
+				}
+			case match.Method.Method != nil:
 				irRoute.HeaderMatches = append(irRoute.HeaderMatches, &ir.StringMatch{
-					Name:  ":method",
-					Exact: match.Method.Method,
+					Name:   ":path",
```
arkodg, line 402: why not use `irRoute.PathMatch` here ?

sbko: `irRoute.PathMatch` doesn't support `Suffix` match

docs/latest/user/grpc-routing.md
sbko: default config works as expected. we're matching everything when neither service or method are specified.

internal/gatewayapi/route.go
sbko: https://github.com/envoyproxy/gateway/issues/1071

internal/gatewayapi/route.go
arkodg: sgtm can you add a comment in here mentioning ^ ?
cc @lizan

sbko: @arkodg done

arkodg (approved):
LGTM thanks !

Issue #1052, opened 2023-02-17
`egctl version` should not error out if server pod not present


arkodg:
*Description*:
```
go run cmd/egctl/main.go version
Error: pod exec on envoy-gateway-system/envoy-gateway-578c86dc47-2dpls failed: error exec into envoy-gateway-system/envoy-gateway-578c86dc47-2dpls container envoy-gateway: unable to upgrade connection: container not found ("envoy-gateway")
pod exec on envoy-gateway-system/envoy-gateway-578c86dc47-2dpls failed: error exec into envoy-gateway-system/envoy-gateway-578c86dc47-2dpls container envoy-gateway: unable to upgrade connection: container not found ("envoy-gateway")
exit status 1
```

arkodg:
cc @zirain

zirain:
can you show me the output of `kubectl get po -n envoy-gateway-system envoy-gateway-578c86dc47-2dpls -oyaml`

arkodg:
> kubectl get po -n envoy-gateway-system envoy-gateway-578c86dc47-2dpls -oyaml

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubectl.kubernetes.io/default-container: envoy-gateway
  creationTimestamp: "2023-02-17T00:54:16Z"
  generateName: envoy-gateway-578c86dc47-
  labels:
    control-plane: envoy-gateway
    pod-template-hash: 578c86dc47
  name: envoy-gateway-578c86dc47-2dpls
  namespace: envoy-gateway-system
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: envoy-gateway-578c86dc47
    uid: cb22a03e-6ddd-453c-a613-fc13d46dd26b
  resourceVersion: "39962"
  uid: d578850c-a7f9-4923-96a6-4aa5acf626ef
spec:
  containers:
  - args:
    - server
    - --config-path=/config/envoy-gateway.yaml
    env:
    - name: ENVOY_GATEWAY_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
    image: arkodg/gateway-dev:9785b41
    imagePullPolicy: Always
    livenessProbe:
      failureThreshold: 3
      httpGet:
        path: /healthz
        port: 8081
        scheme: HTTP
      initialDelaySeconds: 15
      periodSeconds: 20
[...]

PR #1053, opened 2023-02-17
egctl: only exec on running pod

zirain:
fixes: #1052 #1054 

cc @arkodg

### DIFF
--- a/internal/cmd/egctl/version.go
+++ b/internal/cmd/egctl/version.go
@@ -11,9 +11,11 @@ import (
 	"io"
 	"sort"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/yaml"
 
 	"github.com/envoyproxy/gateway/internal/cmd/options"
@@ -35,8 +37,8 @@ func NewVersionCommand() *cobra.Command {
 		Use:     "version",
 		Aliases: []string{"versions", "v"},
 		Short:   "Show version",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return versions(cmd.OutOrStdout(), egContainerName, output)
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(versions(cmd.OutOrStdout(), egContainerName, output))
 		},
 	}
 
@@ -75,7 +77,7 @@ func versions(w io.Writer, containerName, output string) error {
 
 	pods, err := c.PodsForSelector(metav1.NamespaceAll, "control-plane=envoy-gateway")
 	if err != nil {
-		return fmt.Errorf("list EG pods failed: %w", err)
+		return errors.Wrap(err, "list EG pods failed")
 	}
 
 	for _, pod := range pods.Items {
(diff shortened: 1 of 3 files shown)


internal/cmd/egctl/version.go
```diff
@@ -79,6 +79,12 @@ func versions(w io.Writer, containerName, output string) error {
 	}
 
 	for _, pod := range pods.Items {
```
arkodg, line 81: also consider looking into line 76, it should be completely fine to run `egctl version` w/o getting server output

zirain: L76 will fail when `kube-apiserver` is unreachable, what's your concern?

arkodg: it is preferable to show unavailable output in a better way rather than erroring out
here's how `docker` does it
```
docker version
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Client:
 Cloud integration: v1.0.29
 Version:           20.10.22
 API version:       1.41
 Go version:        go1.18.9
 Git commit:        3a2c30b
 Built:             Thu Dec 15 22:28:41 2022
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true
 ```
 and here's how `kubectl` does it
```
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"clean", BuildDate:"2022-11-09T13:36:36Z", GoVersion:"go1.19.3", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
The connection to the server kubernetes.docker.internal:6443 was refused - did you specify the right host or port?
```

Comment (zirain) [reply internal/cmd/egctl/version.go:81]: sound good, will take a look later.

Comment (zirain) [reply internal/cmd/egctl/version.go:81]: open a ticket for that, https://github.com/envoyproxy/gateway/issues/1054

Comment (zirain) [reply internal/cmd/egctl/version.go:83]: https://github.com/envoyproxy/gateway/pull/1053/commits/a4316a0afd86c8624255620141a0907166433611 maybe help this

File: internal/cmd/egctl/version.go [hunk 2]
```diff
@@ -75,10 +77,16 @@ func versions(w io.Writer, containerName, output string) error {
 
 	pods, err := c.PodsForSelector(metav1.NamespaceAll, "control-plane=envoy-gateway")
 	if err != nil {
-		return fmt.Errorf("list EG pods failed: %w", err)
+		return errors.Wrap(err, "list EG pods failed")
 	}
 
 	for _, pod := range pods.Items {
+		if pod.Status.Phase != "Running" {
```
arkodg, line 84: what about the case when no pods are running ? we will just skip this entire section, shouldnt we add some helpful prompt for the user when such a case happens ?

zirain: I think we should print some warning message in that case.

arkodg: this is fine for now, but in the future we should be try and remove any kubeism from the command

internal/cmd/egctl/version.go
```diff
@@ -75,10 +77,16 @@ func versions(w io.Writer, containerName, output string) error {
...
+
+			fmt.Fprintf(w, "WARN: pod %s/%s is not running, skip", pod.Namespace, pod.Name)
```
arkodg, line 86:
```suggestion
			fmt.Fprintf(w, "WARN: pod %s/%s is not running, skipping it.", pod.Namespace, pod.Name)
```

arkodg (approved):

Issue #915, opened 2023-01-16
Implement `egctl`


zirain:
- [x] [design doc](https://github.com/envoyproxy/gateway/blob/main/docs/latest/design/egctl.md)
- [x] `version` command #914 
- [x] `config` command
  - [x] `config all` https://github.com/envoyproxy/gateway/pull/1060
  - [x] `config cluster` https://github.com/envoyproxy/gateway/pull/1083
  - [x] `config listener` https://github.com/envoyproxy/gateway/pull/1083
  - [ ] ?
- [ ] `analyze` command
- [ ] E2E tests
- [ ] Others?

Xunzhuo:
Design: https://github.com/envoyproxy/gateway/blob/main/docs/latest/design/egctl.md

Xunzhuo:
I see `egctl analyze` in desgin doc, what is the status of it?

arkodg:
@zirain do you think we should address this in 0.4.0 or 0.5.0 (I prefer latter) ?

zirain:
plan to support `version` and `config` in v0.4

Xunzhuo:
How about adding support for:
+ `egctl install`
+ `egctl uninstall`
+ `egctl manifests`

After #651 gets merged?

zirain:
IMO, we should consider add those once the install/uninstall became stable, support helm chart is enough for now.

BTW, setup e2e framework is more urgent.

arkodg:
@zirain if all v0.4 items are tackled, can we move this issue to v0.5 ?

arkodg:
closing in favor of https://github.com/envoyproxy/gateway/issues/2028

PR #1060, opened 2023-02-18
egctl: implement `config envoy-proxy all` command

zirain:
xref: #915


arkodg:
shouldnt the title be `egctl config envoy-proxy all <instance_name>`

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	cfgCommand.AddCommand(proxyCommand())
+
+	flags := cfgCommand.Flags()
+	options.AddKubeConfigFlags(flags)
+
+	cfgCommand.PersistentFlags().StringVarP(&output, "output", "o", "json", "One of 'yaml' or 'json'")
+	cfgCommand.PersistentFlags().StringVar(&containerName, "container-name", "envoy", "Name of container where envoy proxy is installed.")
+	cfgCommand.PersistentFlags().Int32Var(&adminPort, "admin-port", 19000, "Admin port of envoy proxy.")
```
arkodg, line 45: lets rm this until we make it configurable ?

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,158 @@
...
+		Long:    "Retrieve information about proxy configuration from envoy proxy and gateway.",
+	}
+
+	cfgCommand.AddCommand(proxyCommand())
+
+	flags := cfgCommand.Flags()
+	options.AddKubeConfigFlags(flags)
+
+	cfgCommand.PersistentFlags().StringVarP(&output, "output", "o", "json", "One of 'yaml' or 'json'")
+	cfgCommand.PersistentFlags().StringVar(&containerName, "container-name", "envoy", "Name of container where envoy proxy is installed.")
```
arkodg, line 44: the original design doc had `instance-name`, why is `container-name` needed, can we do without adding this arg ?

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,158 @@
...
+func allConfigCmd() *cobra.Command {
+
+	allConfigCmd := &cobra.Command{
+		Use:   "all <pod-name>",
+		Short: "Retrieves all configuration for the Envoy in the specified pod",
+		Long:  `Retrieve information about all configuration for the Envoy instance in the specified pod.`,
+		Example: `  # Retrieve summary about all configuration for a given pod from Envoy.
+  egctl config envoy-proxy all <pod-name> -n <pod-namespace>
+
+  # Retrieve full cluster dump as YAML
```
arkodg, line 72: why `cluster` ?

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,158 @@
...
+  egctl config envoy-proxy all <pod-name> -n <pod-namespace> -o yaml
+
+  # Retrieve full cluster dump with short syntax
```
arkodg, line 75: same as above

internal/cmd/egctl/config.go
zirain: this's my fault, I test this on an istio pod locally.

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,159 @@
...
+	c.AddCommand(allConfigCmd())
+
+	return c
+}
+
+func allConfigCmd() *cobra.Command {
+
+	allConfigCmd := &cobra.Command{
+		Use:   "all <pod-name>",
+		Short: "Retrieves all configuration for the Envoy in the specified pod",
```
arkodg, line 68:
```suggestion
		Short: "Retrieves all Envoy xDS resources from the specified pod",
```

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,159 @@
...
+		Long:  `Retrieve information about all configuration for the Envoy instance in the specified pod.`,
```
arkodg, line 69:
```suggestion
		Long:  `Retrieves information about all Envoy xDS resources from the specified pod.`,
```

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	c.AddCommand(allConfigCmd())
+
+	return c
+}
+
+func allConfigCmd() *cobra.Command {
+
+	allConfigCmd := &cobra.Command{
```
arkodg, line 66: thoughts on adding `-A` to get information from all pods in all namespaces ?

zirain: I never thought of such a user case, can you share more details?
it would be better to track that with a ticket.

Xunzhuo: I think it makes sense, `-A` allows users to get info from all pods in all namespaces, like `kubectl get <resources> -A`.

internal/cmd/egctl/config.go
```diff
@@ -0,0 +1,159 @@
...
+		Short: "Retrieves all Envoy xDS resources from the specified pod",
+		Long:  `Retrieves information about all Envoy xDS resources from the Envoy instance in the specified pod.`,
+		Example: `  # Retrieve summary about all configuration for a given pod from Envoy.
+  egctl config envoy-proxy all <pod-name> -n <pod-namespace>
+
+  # Retrieve full configuration dump as YAML
+  egctl config envoy-proxy all <pod-name> -n <pod-namespace> -o yaml
+
+  # Retrieve full configuration dump with short syntax
+  egctl c proxy a <pod-name> -n <pod-namespace>
```
Xunzhuo, line 77: I feel this is a bit weird here, The short name `a` here means `all`, but in an english word, it represents a single odd number.

internal/cmd/egctl/config.go
arkodg: yah helps users like me who dont want to run an extra `kubectl get po -A` cmd :)  to get pod name

internal/cmd/egctl/config.go
arkodg: agreed, we dont need a short name for `all`

zirain: reverted

arkodg:
thanks @zirain this looks like a great start, hoping you can add some test cases since the current diff coverage for this PR is 0%

internal/cmd/egctl/config_test.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+		{
+			output:   "yaml",
+			expected: "out.yaml",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.output, func(t *testing.T) {
+			got, err := extractConfigDump(fw, tc.output)
```
Xunzhuo, line 95: Thanks for your effort to put UT around this feature. But in my view, this kind of testes could be better to add as e2e tests. 

In this approach, we put configDumps into fake server, and request to it and get configDumps back, and compare them. It make less sence I think 😄. And we will contain long test jsons/yamls to mimic real envoyproxy configs. cc @arkodg WDYT.

zirain: it's useful for `config envoy-proxy cluster`

arkodg: @Xunzhuo yeah e2e would be ideal, maybe we can raise a GH issue to track it, but for now, this approach add coverage for a sufficient amount of code in this PR, which Im fine with

tools/linter/codespell/.codespell.skip
```diff
@@ -12,3 +12,4 @@
 go.mod
 go.sum
 bin
+testdata
```
arkodg, line 15: this isn't right, we were linting `testdata` earlier

internal/cmd/egctl/config_test.go
Xunzhuo: Thanks @zirain, @arkodg. It makes sense.

tools/linter/codespell/.codespell.skip
zirain: let me ignore special file, configdump file will not pass codespell check.

zirain: done

Xunzhuo (dismissed):
Thanks @zirain

tools/linter/codespell/.codespell.skip
arkodg: I pulled in your branch and deleted the `tracer` that was causing the issue
```
diff --git a/internal/cmd/egctl/testdata/config/out.json b/internal/cmd/egctl/testdata/config/out.json
index d7bc3d3..1bd6f16 100644
--- a/internal/cmd/egctl/testdata/config/out.json
+++ b/internal/cmd/egctl/testdata/config/out.json
@@ -1316,10 +1316,6 @@
                                 "envoy.extensions.filters.network.thrift_proxy.router.v3.Router"
                             ]
                         },
-                        {
-                            "name": "envoy.dynamic.ot",
-                            "category": "envoy.tracers"
-                        },

```
this fixed `codespell`

arkodg (approved):
LGTM, thanks !

PR #1066, opened 2023-02-21
feat: make egctl x translate output struct

Xunzhuo:
Resolved: #1055

Issue #50, opened 2022-05-02
Define Initial APIs


danehans:
https://github.com/envoyproxy/gateway/pull/16 specifies the need to define APIs, e.g. bootstrap config. To begin developing these APIs, the community should agree on:

1. How these APIs will defined, protos vs Go types. [Emissary](https://github.com/emissary-ingress/emissary) uses protos to define [API types](https://github.com/emissary-ingress/emissary/blob/master/api/agent/director.proto) whereas [Contour](https://github.com/projectcontour/contour) uses Go types.
2. How these APIs will documented.
3. How these APIs will be maintained.
4. Others?

danehans:
@LukeShu per the 5/24/22 meeting notes, please document the rationale for using kube-builder/controller-gen over protobuf.

LukeShu:
Sorry, I didn't get to this on Wednesday, and have been on vacation since then. Will get that documentation up soon.

danehans:
https://github.com/envoyproxy/gateway/issues/51 defines the initial config APIs for control and data planes.

PR #1070, opened 2023-02-22
xds Bootstrap Design and API

arkodg:
relates to https://github.com/envoyproxy/gateway/issues/31


arkodg:
cc @mattklein123 @lizan

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -43,6 +43,15 @@ type EnvoyProxySpec struct {
...
 	Logging ProxyLogging `json:"logging,omitempty"`
+
+	// Bootstrap defines the Envoy Bootstrap as a YAML string.
+	// Visit https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-bootstrap
+	// to learn more about the syntax.
+	// Some fields within the Bootstrap that are required to communicate with the xDS Server (Envoy Gateway) and receive xDS resources
+	// from it are not configurable and will result in the `EnvoyProxy` resource being rejected.
+	// We strongly recommend using `egctl x translate` to generate a `EnvoyProxy` resource with the `Bootstrap` field set to the default
+	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
+	Bootstrap string `json:"bootstrap,omitempty"`
```
zirain, line 54: IMO, a string is easy to use, but hard to validate.

arkodg: we will need to unmarshal it to https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#bootstrap-proto and run the object's `Validate` along with any other custom validations we want to do

arkodg: if there are failures, we will highlight those in the `Status` field of `EnvoyProxy` resource

docs/latest/design/bootstrap.md
```diff
@@ -0,0 +1,374 @@
...
+}
+```
+
+## Tooling
+
+A CLI tool `egctl x translate` will be provided to the user to help generate a working Bootstrap configuration.
+Here is an example where a user inputs a `GatewayClass` and the CLI generates the `EnvoyProxy` resource with the `Bootstrap` field populated.
+
+```
+cat <<EOF | egctl x translate --from gateway-api --to gateway-api -f -
```
zirain, line 48: prefer to `egctl x translate bootstrap`

Xunzhuo: In https://github.com/envoyproxy/gateway/pull/1066, we use `egctl x translate --from gateway-api --to gateway-api -t bootstrap -f xxx` to generate boostrap config of input.

Also we can use:
+ `egctl x translate --from gateway-api --to gateway-api -t cluster`
+ `egctl x translate --from gateway-api --to gateway-api -t listener`
+ `egctl x translate --from gateway-api --to gateway-api -t route`

To get other resource type configs.

arkodg: @Xunzhuo  imho `-t bootstrap` doesnt make sense when output type is `gateway-api`

arkodg: @zirain I would like to continue to leverage `egctl x translate --from gateway-api --to gateway-api` so that in the future we can use this for #24 where we can translate Gateway API resources into Gateway API resources with xDS Filter Extensions

api/config/v1alpha1/envoyproxy_types.go
arkodg: this was raised in the EG Community meeting today. Even if we used a strongly typed field here such as envoy's `v3.Bootstrap`, we would need to build extra tooling to convert the imported structure into CRD schema validations which would help validating early and synchronously during a `kubectl apply`.

docs/latest/design/bootstrap.md
Xunzhuo: @arkodg Right, only works when --to=xds for now.

api/config/v1alpha1/envoyproxy_types.go
zirain: prefer to use a good webhook, not status to tell user what's wrong, not strong option.

arkodg: I tried to use v3.Bootstrap but `controller-gen` was unable to generate a CRD for it.

zirain: `v3.Bootstrap` came from protobuf, there's a lot of magic.

arkodg: yes, it was decided early on to stick to `controller-gen` tooling rather than protobuf tooling ( which would have helped with this issue using https://pkg.go.dev/istio.io/tools/cmd/cue-gen)

Xunzhuo:
Need to run `make manifests` @arkodg I think.

arkodg:
> Need to run `make manifests` @arkodg I think.

thanks and pushed !

docs/latest/design/bootstrap.md
```diff
@@ -0,0 +1,374 @@
...
+* Provide tooling to allow the user to generate the default Bootstrap configuration
+  as well as validate their custom Bootstrap.
+
+## Non Goals
+
+* Allow user to configure only a section of the Bootstrap
+
+## API
+
+Leverage the existing [EnvoyProxy][] resource which can be attached to the [GatewayClass][] using
```
Xunzhuo, line 23: I wonder if EnvoyProxy Config can be attached to `Gateway`, I think maybe `GatewayClass` is effected all Gateways, right? Can end user just override single Gateway?

docs/latest/design/bootstrap.md
```diff
@@ -0,0 +1,374 @@
...
+	// from it are not configurable and will result in the `EnvoyProxy` resource being rejected.
+	// We strongly recommend using `egctl x translate` to generate a `EnvoyProxy` resource with the `Bootstrap` field set to the default
+	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
+	Bootstrap string `json:"bootstrap,omitempty"`
+}
+```
+
+## Tooling
+
+A CLI tool `egctl x translate` will be provided to the user to help generate a working Bootstrap configuration.
```
Xunzhuo, line 44: I think maybe this is a bit not clear here, I think how about --from gateway-api --to envoy-gateway ?

arkodg: I would like to reuse this CLI for #24 i.e. EG with Envoy Extensions so favoring `--to gateway-api`

api/config/v1alpha1/envoyproxy_types.go
arkodg: https://github.com/envoyproxy/gateway/issues/50 has more details @zirain

docs/latest/design/bootstrap.md
Xunzhuo: Make sences, thanks @arkodg

Xunzhuo:
Thanks for starting this up @arkodg. Need another look from @zirain

zirain (approved):
LGTM, thanks!

Issue #71, opened 2022-05-24
WASM Plugin Design and Implementation


rahulanand16nov:
Hi guys, we have been following this project closely because it overlaps with some of the use cases [Kuadrant](https://github.com/Kuadrant) solves using Istio, but we feel we can also extend support with Envoy Gateway as well, given there is a way to inject WebAssembly modules from an image registry.

Currently, we use [WasmPlugin](https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/) resource from Istio, but I think it would be very beneficial to allow something similar to increase the extensibility of Envoy Gateway project.

arkodg:
hi @rahulanand16nov are you referring to extending data plane/Envoy using WASM or control plane (Envoy Gateway using WASM, raised https://github.com/envoyproxy/gateway/issues/20 to track this) using WASM ?

rahulanand16nov:
@arkodg There is already a way of extending data plane/Envoy using WASM (filter and singleton service), but that works only when you're manually configuring it or a control plane provides a CRD to configure that for you (Istio provides WasmPlugin as linked above). I believe Envoy Gateway's control plane can also offer a similar CRD to inject WASM modules but through an OCI image. 

In Envoy, you can either specify the local path or remote URI for the .wasm file. Istio's WasmPlugin extends fetching from the image registry by deploying a local cache in the control plane, which will be very beneficial for the Envoy Gateway project users.

danehans:
@rahulanand16nov thanks for the additional context. I have set this issue to the backlog as the upcoming v0.2.0 will focus on core gateway/ingress functionality. This could be a feature that's considered for a follow-on release. See https://github.com/envoyproxy/gateway/issues/11#issuecomment-1126340096 for additional details. Is this a feature that you're willing to contribute in the future?

danehans:
@Xunzhuo thanks for picking this issue up. I assigned the issue to you and included it in the v0.3.0-rc.1 milestone. I suggest starting off with a high-level design doc. You can use the system design doc as a reference.

danehans:
@Xunzhuo thoughts on you assisting with #24 first?

Xunzhuo:
Sorry, I missed these comments, yes, this should be after #24.

Xunzhuo:
I will still assign myself to this, this is not in v0.3.0, but in my schedule in the next releases.

arkodg:
closing in favor of https://github.com/envoyproxy/gateway/issues/2025

PR #1074, opened 2023-02-23
Add design doc for extending Envoy Gateway

AliceProxy:
## Summary

This is the initial design and proposal for a method to allow vendors to extend Envoy Gateway and provide additional configuration and functionality to their users. The primary focus of the extension supported by this proposal is to provide a powerful and extensive means of changing how Envoy Gateway operates and the usage of these extensions is targeted at mainly vendors. This does not cater to the use-case of a super-user wanting to make some modifications Envoy Gateway. Super-users would better be served by additional means of extending Envoy Gateway statically such as issues (TODO: link other issues here).

My objective here is to get feedback and conversation going on the design laid out in the user-docs and after the design is agreed upon and approved then I will provide the implementation of these features for the 0.4.0 release of Envoy Gateway.

The design is a group effort between @LanceEa, @haq204, and myself so they will be involved in the discussion here as well as discussion on the implementation when we get to that stage.

We have a [branch in our fork of Envoy Gateway](https://github.com/datawire/envoy-gateway/tree/extensions) where we have built out a proof of concept using the design laid out here. For now I have left the protos and specifics of the hook calls out of the user-doc since I am expecting to have some discussion and iteration around the implementation before adding additional user-facing documentation at the end.

## Considerations and Open Questions

- We considered having a more dynamic registration system for extension servers but leaned towards a bootstrap since it was faster to implement. If there are alternative suggestions here that eliminate some friction for the users then I would be happy to revise it now. Alternatively, we could implement a more dynamic registration if needed in the future.
- This proposal allows for running multiple extensions but in practice, we are not sure if that is something developers or users would want because it would become hard to reason about two extensions modifying the same components, and being unable to guarantee compatibility between extensions is an open concern. A possible outcome from this could be to prevent more than a single extension server from being registered.
- Versioning of apiGroups and the namespaces of an `extensionRef` have been left out of this design because the Gateway API currently doesn't allow for it. We may want to propose that addition to the Gateway API maintainers.


arkodg:
@skriss  @sunjayBhatia is something Contour would be interested in ?
@Xunzhuo can you also ptal, I know you were interested in leveraging EG in other projects

arkodg:
hey @AliceProxy trying to understand the use case, please correct me If I wrong -
Extension developers / vendors would like to expose new functionality (not exposed using existing Gateway API fields or existing Envoy Gateway Extensions) , using ExtensionFilters (or PolicyAttachment) without having to implement the feature the inside Envoy Gateway. 
* this would require some entity to reconcile the new feature ExtensionFilter
* as well as require another entity to translate the filter into xDS resources and then patch it into the right
location within the main xDS resources generated by EG

What are you thoughts on building a controller on top of Envoy Gateway (rather than beside it) which
* Reconciles the ExtensionFilter
* Mutates the HTTPRoute with the ExtensionFilter with the actual xDS resource. 

Here's one example

```
apiVersion: gateway.unicorn.io/v1alpha1
kind: LuaFilter
metadata:
  name: some-lua-filter
spec:
  script: "blah"
---

kind: HTTPRoute
metadata:
  name: http-with-lua
spec:
  parentRefs:
  - name: eg
  hostnames:
  - lua.example 
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    filters:
    - type: ExtensionRef
      extensionRef:
        group: gateway.unicorn.io
        kind: LuaFilter
        name: some-lua-filter
    backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
```

The Extension Controller can reconcile the ` LuaFilter` as well as the `HTTPRoute` and then mutate the 
HTTPRoute with a new ExtensionRef called `EnvoyFIlter`
```
kind: HTTPRoute
metadata:
  name: http-with-lua
spec:
  parentRefs:
  - name: eg
  hostnames:
  - lua.example 
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    filters:
    - type: ExtensionRef
      extensionRef:
        group: gateway.unicorn.io
        kind: LuaFilter
        name: some-lua-filter
     - type: ExtensionRef
        extensionRef:
          group:   gateway.envoyproxy.io
          kind: EnvoyFilter
          name: http-with-lua-xds-httproute
    backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000

kind: EnvoyFilter
metadata:
  name: http-with-lua-xds-httproute
spec:
   type: Route
   value : |
    - match:
         headers:
          - name: :authority
             stringMatch:
                exact: www.example.com
            prefix: /
        route:
          cluster: default-backend-rule-0-match-0-www.example.com
          typed_per_filter_config:
            envoy.filters.http.lua:
              "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.LuaPerRoute
              name: hello.lua
```

Above is one direction where we build an `EnvoyFilter` API (as part of #24) and use [egctl x translate](https://gateway.envoyproxy.io/latest/user/egctl.html#egctl-experimental-translate) to help with this translation. This would not be a patch API but an all or nothing API for a specific xDS Resource (we would  have to come up with way to map a Gateway listener to an xds Listener, `BackendRef` to a xds Cluster etc)

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+as means to add larger features that require dynamic user-configuration.
+
+As an example, an extension developer may wish to provide their own out of the box authentication filters that require configuration from the end-user. This is a scenario where the ability to introduce
+custom resources and attach them to [HTTPRoute][]s as an [ExtensionRef][] is necessary. Providing the same feature through a series of xDS patch resources would be too cumbersome for many end-users that want to avoid
+that level of complexity when managing their clusters.  
+
+## Goals
+
+- Provide a foundation for extending the Envoy Gateway control plane
+- Allow Extension Developers to introudce their own cusom resources for extending the Gateway-API via [ExtensionRefs][], [policyAttachments][] (future) and [backendRefs][] (future).
```
Xunzhuo, line 18: typo: introudce -> introduce

zirain: why this pass spellcheck CI?

AliceProxy:
Hey @arkodg thanks for the idea. First I have a couple questions just to make sure I understand your suggestion.

> What are you thoughts on building a controller on top of Envoy Gateway (rather than beside it)

Can you elaborate on specifics of what you envision by `on top of` instead of `beside it`? Do you mean the extension "extends" EG's controller and essentially takes over all controller functions for EG or something along those lines?

> The Extension Controller can reconcile the  LuaFilter as well as the HTTPRoute and then mutate the
HTTPRoute with a new ExtensionRef called EnvoyFIlter

Is this something you see happening in memory only or do you mean the extension server would inject this into the HTTPRoute in the API server? If so then I think this would be a big problem for ArgoCD users since the extension will make a bunch of modifications to their HTTPRoutes and cause Argo to complain about sync differences between the repo and the cluster. While you can configure Argo to ignore sync differences for certain changes that you expect, that might not be feasible since these would be dynamic transformations and not something like a static line you expect to be injected the same way every time. 


> Above is one direction where we build an EnvoyFilter API (as part of https://github.com/envoyproxy/gateway/issues/24) and use [egctl x translate](https://gateway.envoyproxy.io/latest/user/egctl.html#egctl-experimental-translate) to help with this translation (which provides the translation context to the extension developer without having to adds hooks in the code)

I think trying to fulfill both the use-cases of advanced users wanting to patch xDS resources and vendors needing to extend Envoy Gateway in the same feature might be missing the mark. I definitely like this as a proposal for fulfilling #24 on it's own, but vendors need a lot more business logic than applying a simply patch/override. 

As an example, if I wanted to insert my own Authentication filter that uses ext_authz I would want to inject a new cluster, potentially make changes to a bunch of existing routes, look at several parts of the existing xDS first to determine what I need to change/modify. So that all begins to get pretty complex with the number of xDS resources an extension needs to have the ability to look at, inject, and modify. Plus I think it would be a bit confusing for end-users to see all this stuff pop-up in their cluster and HTTPRoutes (assuming this is not all just in-memory as I was curious about) but that's just a nit-pick.

arkodg:
* The problem statement of vendor extensions could be reduced to providing a way to patch specific Envoy resources.  At Tetrate, we build **on top of** Istio using [EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/) which provides a way to patch envoy resources. This is how #24 may provide a way **to patch from the outside rather than the inside**. 

* Instead of patching, complete xDS resources can be provided by the vendor, which #24 could help with.

* Your point around mutating HTTPRoute and causing issues with Argo is a valid one. This knob in Argo might help ignore the filter field https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#respect-ignore-difference-configs

* I agree with your observation that patching from the outside might alarm the user, saving the extensions/policies in a extension namespace could help make them less visible

arkodg:
hey @AliceProxy can you also elaborate on the API between EG and the Extension Service and how do the `hooks` work  ?

arkodg:
I like @sunjayBhatia's suggestion of some version of Gateway API under the hood, maybe we shouldnt expose #24 so soon, instead implement it as a special API that can be applied using a special/non k8s provider (like a REST endpoint in EG), so its not exposed to end users right away, and vendors/extension devs are taking the onus of xDS compatibility for now

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+inject the authorization service as a sidecar to the Envoy deployments created by Envoy Gateway rather than deploy the authorization service in the Kubernetes cluster separately. Doing so would help to minimize the 
+latency between Envoy and the authentication service and also allows the end user to spend less time worrying about how things are deployed in the cluster.
+
+![Control Plane Extension Example](../images/extension-auth-example.png)
+
+## Known Challenges
+
+Extending Envoy Gateway by using an external extension server which makes use of hook points in Envoy Gateway does come with some know issues that will have to be worked around.
+
+- Extension server handling it's own state such as in the scenario where an extension establishes watches its own CRDs in the cluster can lead to a scenario where
```
zhaohuabing, line 151: Is it possible to let Envoy Gateway wrap the custom CRD resources in the context when calling the extenstion server hooks?

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+ensuring that the extension and Envoy Gateway are in-sync becomes a challenge.
+  - A potential workaround to this issue would be to have the extension use hook points in Envoy Gateway to replace Envoy Gateway's Kubernetes provider (the component that watches resources such as the
```
zhaohuabing, line 153: Dose this mean that Envoy Gateway watches the custom CRD resources and wrap the custom CRD resources in the RPC context when calling the extenstion server webhooks?

Does this mean that the Envoy Gateway watches the CRD resources for extensions and includes them in the RPC context when calling the extension server webhooks?

AliceProxy:
> hey @AliceProxy can you also elaborate on the API between EG and the Extension Service and how do the `hooks` work ?

Sure thing. [Here is the example protos from our fork](https://github.com/datawire/envoy-gateway/tree/extensions/proto/extension) that we used to create a POC running in-cluster. I refrained from specifying the API in the user-doc at this time since I figured that the specific internals of the hooks was more of an implementation detail and if we go ahead and move forward with this design then I would add user-docs for the finalized version of the hooks API when we merged the implementation of the design. I also figured that depending on the ideas discussed here that we may make changes to what hooks we were expecting to create/need. I think an excellent example of this would be the idea of an Extenal provider that wasn't part of my initial design but rather a great idea that came out of discussions around this design and solving some of the open questions that I listed when I opened the PR.

arkodg:
> > hey @AliceProxy can you also elaborate on the API between EG and the Extension Service and how do the `hooks` work ?
> 
> Sure thing. [Here is the example protos from our fork](https://github.com/datawire/envoy-gateway/tree/extensions/proto/extension) that we used to create a POC running in-cluster. I refrained from specifying the API in the user-doc at this time since I figured that the specific internals of the hooks was more of an implementation detail and if we go ahead and move forward with this design then I would add user-docs for the finalized version of the hooks API when we merged the implementation of the design. I also figured that depending on the ideas discussed here that we may make changes to what hooks we were expecting to create/need. I think an excellent example of this would be the idea of an Extenal provider that wasn't part of my initial design but rather a great idea that came out of discussions around this design and solving some of the open questions that I listed when I opened the PR.

great thanks for sharing this, super helpful in understanding what is happening, here are some thoughts 
* looks like the extension provider is updating all xDS resources (`Listener`, `RouteConfiguration`) in 1 single request/response callback, which means it needs to send out all xDS resources as well as entire context in 1 payload, thoughts on splitting up callbacks per xDS resource ?
* the context / back pointer points to a few of the fields within Gateway API resources today. Is it easier to build out a front pointer instead (mapping Gateway API resources to xDS resources)
* Same concern around scale wrt context size, which boils down to how to split up callbacks per resources, or maybe this is premature optimization that can be improved later with newer callbacks.
The onus of managing scale falls on the extension controller, not EG :)

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+        group: example.myextension.io
+        kind: OAuth2Filter
+        name: oauth2-filter
+    backendRefs:
+    - name: backend
+      port: 3000
+```
+
+In order to enable the translation of new resources introduced by extensions, Envoy Gateway provides hook points within the translation pipeline, where it calls out to extension services registered in the [EnvoyGateway config][] 
+if they specify an `apiGroup` that matches the `group` of an `ExtensionRef` filter. The extension will then be able to modify the xDS that Envoy Gateway generated and send back the
```
muyuan0, line 124: Does it mean that the extension will work in the xDS Translator layer?
Envoy gateway will send the xDS IR to the extension and the extension send the xDS patch back to the Envoy gateway.

docs/latest/design/extending-envoy-gateway.md
muyuan0, line 18:
```suggestion
- Allow Extension Developers to introduce their own custom resources for extending the Gateway-API via [ExtensionRefs][], [policyAttachments][] (future) and [backendRefs][] (future).
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+- Allow Extension Developers to introudce their own cusom resources for extending the Gateway-API via [ExtensionRefs][], [policyAttachments][] (future) and [backendRefs][] (future).
+- Extension developers should **NOT** have to maintain a custom fork of Envoy Gateway
+- Provide a system for extending Envoy Gateway which allows extension projects to ship updates independent of Envoy Gateway's release schedule
+- Extensions can hook into both the Envoy configuration pipeline and the infra manager pipeline
+- Modify the generated Envoy xDS and Envoy Gateway infra config (either directly via an xDS hook or earlier in the pipeline or by dynamically adding resources)
+- Setup a foundation for the initial iteration of Extending Envoy Gateway
+
+## Non-Goals
+
+- The initial deisgn does not capture every hook that Envoy Gateway will eventually support.
```
muyuan0, line 27:
```suggestion
- The initial design does not capture every hook that Envoy Gateway will eventually support.
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+- Extend [Gateway API Policy Attachments][]. At some point these will be addressed using this extension system, but the initial implementation omits these.
+- Guarantee that multiple extensions will work together without conflicts
+
+## Overview
+
+Envoy Gateway can be extended by vendors by means of an extension server developed by the vendor and deployed alongside Envoy Gateway.
+An extension server can make use of one or more pre/post hooks inside Envoy Gateway before and after it's major components (translator, etc.) to allow the extension to modify the data going into or coming out of these components.
```
muyuan0, line 34:
```suggestion
An extension server can make use of one or more pre/post hooks inside Envoy Gateway before and after its major components (translator, etc.) to allow the extension to modify the data going into or coming out of these components.
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+An extension can be created external to Envoy Gateway as it's own Kubernetes deployment or loaded as a sidecar. gRPC is used for the calls between Envoy Gateway and an extension. In the hook call, Envoy Gateway sends data as well
```
muyuan0, line 35:
```suggestion
An extension can be created external to Envoy Gateway as its own Kubernetes deployment or loaded as a sidecar. gRPC is used for the calls between Envoy Gateway and an extension. In the hook call, Envoy Gateway sends data as well 
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Extending Gateway API and the Data Plane
+
+Envoy Gateway manages [Envoy][] deployments, which act as the data plane that handles actual user traffic. Users configure the data plane using the K8s Gateway API resources which Envoy 
+Gateway converts into [Envoy specific configuration (xDS)][] to send over to Envoy.
+
+Gateway API offers [ExtensionRef filters][] and [Policy Attachments][] as extension points for implementers to use. Envoy Gateway extends the Gateway API using these extension points to provide support for [rate limiting][] 
+and [authentication][]. The initial design of Envoy Gateway extensions will primarily focus on ExtensionRef filters, where users can reference a custom resource as an HTTP Filter.
+
+When Envoy Gateway encounters an [HTTPRoute][] or [GRPCRoute][] that has an `ExtensionRef` `filter` with an `apiGroup` and `kind` that Envoy Gateway does not supoport, it will first
```
muyuan0, line 84:
```suggestion
When Envoy Gateway encounters an [HTTPRoute][] or [GRPCRoute][] that has an `ExtensionRef` `filter` with an `apiGroup` and `kind` that Envoy Gateway does not support, it will first
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,176 @@
...
+In the example above where an extension developer wants to create a new `ExtensionRef` filter that provides authentication, they would likely deploy their own authorization service into the cluster
+and make use of the extension hooks to configure Envoy to use that service to authenticate requests. By also using extension points in the control plane, the extension can dynamically
+inject the authorization service as a sidecar to the Envoy deployments created by Envoy Gateway rather than deploy the authorization service in the Kubernetes cluster separately. Doing so would help to minimize the 
+latency between Envoy and the authentication service and also allows the end user to spend less time worrying about how things are deployed in the cluster.
+
+![Control Plane Extension Example](../images/extension-auth-example.png)
+
+## Known Challenges
+
+Extending Envoy Gateway by using an external extension server which makes use of hook points in Envoy Gateway does come with some know issues that will have to be worked around.
```
muyuan0, line 149:
```suggestion
Extending Envoy Gateway by using an external extension server which makes use of hook points in Envoy Gateway does come with some known issues that will have to be worked around.
```

docs/latest/design/extending-envoy-gateway.md
muyuan0, line 151:
```suggestion
- Extension server handling its own state such as in the scenario where an extension establishes watches its own CRDs in the cluster can lead to a scenario where
```

docs/latest/design/extending-envoy-gateway.md
AliceProxy: Correct. For the initial implementation of this design, there will likely only be a single hook point where an extension has control over the xDS resources right before Envoy Gateway updates them in the xDS server to be consumed by Envoy Proxy. Still, the framework of the extensions system is meant to be a foundation where more fine-grained hooks can be introduced into the future for use cases where the extension does not need such broad control.

AliceProxy:
> the context / back pointer points to a few of the fields within Gateway API resources today. Is it easier to build out a front pointer instead (mapping Gateway API resources to xDS resources)

Not entirely sure I'm thinking of the same thing you are here. Do you mean the `ExtensionContext *HTTPListenerExtensionContext` field in the example I linked that is used to pass context? If so then yes those are passed to the extension to let it know which `HTTPRoutes` referenced `extensionRef` objects with an `apiGroup/Kind` that the extension owns. This helps the extension know which xDS routes came from `HTTPRoutes` it should care about and which ones it should ignore. I would definitely love to hear more about any ideas for how we can make it easier for the extension to know about which resources it can ignore and which ones it needs to look at. 

> Same concern around scale wrt context size, which boils down to how to split up callbacks per resources, or maybe this is premature optimization that can be improved later with newer callbacks.
The onus of managing scale falls on the extension controller, not EG :)

1. Yes the responsibility of managing the performance and scale of any computation that EG would be calling out to an extension to perform is absolutely on the extension developer to optimize for and is worth explicitly mentioning in the doc.
2. Also yes, wrt a single gRPC call in the example I provided, we chose to go with a single one to prove out the concept in a simple way that would be easy to understand. While single gRPC calls should have a max size somewhere around 4MB if I recall correctly which might be fine for most setups, this might need to be implemented as a stream or something similar to what you suggested for performance improvements and ensuring the system will work even with very large xDS configs for the actual implementation.

docs/latest/design/extending-envoy-gateway.md
muyuan0: Thanks for your reply.

zhaohuabing:
I'm thinking that  Webhook and EnvoyFilter extension mechanisms could be used in different scenarios:

* Webhook is best suited for situations where consistency is highly required, as it patches xds resoureces in a synchronized manner before sending them to the Envoy. However, directly binding to Webhooks can introuduce addtional latency. 
* EnvoyFilter is ideal for scenarios where eventual consistency is acceptable.  This mechanism gives more flexibilities since  extension controllers can watch extension CRDs, reconcile them, and generate EnvoyFilters, which can later be mergered into XDS by Envoy Gateway and send to the Envoy.

zirain:
> You can use Webhook and EnvoyFilter in different scenarios.
> 
> * Webhook is best suited for situations where consistency is highly required, as it patches xds resoureces in a synchronized manner before sending them to the Envoy. However, directly binding to Webhooks can introuduce addtional latency.
> * EnvoyFilter is ideal for scenarios where eventual consistency is acceptable.  This mechanism gives more flexibilities since  extension controllers can watch extension CRDs, reconcile them, and generate EnvoyFilters, which can later be mergered into XDS by Envoy Gateway and send to the Envoy.

IMO, `EnvoyFilter` is a kind of patching, `EG` should provide a way like `EnvoyFilter`.
Also if there's a way to control the whole xDS, that would be great.

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -149,9 +149,13 @@ latency between Envoy and the authentication service and also allows the end use
 Extending Envoy Gateway by using an external extension server which makes use of hook points in Envoy Gateway does come with some known issues that will have to be worked around.
 
 - Extension server handling its own state such as in the scenario where an extension establishes watches its own CRDs in the cluster can lead to a scenario where
-ensuring that the extension and Envoy Gateway are in-sync becomes a challenge.
-  - A potential workaround to this issue would be to have the extension use hook points in Envoy Gateway to replace Envoy Gateway's Kubernetes provider (the component that watches resources such as the 
-Gateway API CRDs) with a custom provider that can enable Envoy Gateway to be a single source of truth for configuration coming from the cluster.
+ensuring that the extension and Envoy Gateway are in-sync becomes a challenge. There are two potential solutions to this problem.
+  - Option 1: Create a new External Provider mechanism where an extension can take over all provider responsibility for Envoy Gateway. The current functionality of the provider would need to become a Library that EG would make use of
```
arkodg, line 153: I think the option 1 suggested on the call today was - EG reconciles/watches only EG resources, and the extension watches the Extension resources

docs/latest/design/extending-envoy-gateway.md
zhaohuabing: I think it may be too much to ask extensions to take over all provider responsibilities. Is it feasible for an extension to only monitor its own resources? If consistency is a concern, I feel that the approach of 'the provider listens to all resources and passes them back to the extension as needed' sounds more reasonable.

docs/latest/design/extending-envoy-gateway.md
zhaohuabing: I agree this is a better way to seperates concerns. However, if keeping EG and Extensions in sync poses a problem, I prefer 'the provider listens to all resources and passes them back to the extension as needed' approach. I think we should put too much responsibility on the Extension side.

AliceProxy:
I just pushed a new commit to address the following takeaways from last week's Envoy Gateway meeting where we all agreed that this design proposal needed some more context added and to make a decision about how custom resources introduced by extensions would be watched.

- Split up the idea of a single xDS hook into multiple more granular hooks
  - Helps reduce the size of a single gRPC call
  - An extension doesn't need as much context now that there are multiple Hook calls to modify a few different xDS resources (Route, VirtualHost, Listener to start with). This means that 
  - Instead of needing to search through a whole Listener to make changes at the Route level, an extension can instead modify Routes as they come in. This means processing on the extension side can be much simpler and faster.
- Added example API for the various hooks 
- Made the decision to not support multiple extensions at the same time
  - Too difficult to make sure extensions wouldn't step on each other's toes or conflict
  - Realistically probably very few people would be running multiple extensions at the same time anyways. Smaller extensions to Envoy Gateway can be accommodated later via #24 and #71 which can co-exist with Envoy Gateway extensions.
 - Made the decision to have Envoy Gateway register new watches on resources introduced by extensions and pass them back to the extension via a hook call when the resources are referenced as an `externalRef` in a `HTTPRoute`.
  - This solves the question about how new resources get watched and solves the potential problem of race conditions and reconcile loops if the extension server and Envoy Gateway both established independent watches.
  - There are tradeoffs with all the proposed ideas for watching new resources, but I think that this solution keeps things simple, and doesn't incur any questions/problems about networking if the extension server were to take over watching all resources by replacing Envoy Gateway's provider.

AliceProxy:
I'd love to get any final reviews from the folks who've chimed in on this issue previously. @Xunzhuo since Arko mentioned that you were interested in leveraging EG in other projects I'd love to hear if you see this design for the foundation of building extensions on top of Envoy Gateway being useful to you.

Xunzhuo:
Hey @AliceProxy, this is great and thanks for pinging a couple of times, and sorry for late reply.

I have read through the design docs and comments, I think currently direction is good.

> Made the decision to not support multiple extensions at the same time

This is definitely correct, we need to start it with a promising way, we can sparate some extensionbilities with xDS patches like EnvoyFilter support, the EnvoyFilter is quite important for super user if they want to customize EG deeper. And also I am in favor of supporting some higher APIs compared with EnvoyFilter like WASM Plugin and Telemetry APIs if possible in the future.

> Split up the idea of a single xDS hook into multiple more granular hooks

Yep, this is reasonable, we need to sparate it into small hooks, like the Pull Push and the Delta Push in xDS protocol. Make it into small parts can help we have more detailed control of changes.

Overall, thanks for spending a lot on designs and pushing to solve lots of blockers. And we can have a general agreement in the end.

Thanks @AliceProxy !

Xunzhuo (dismissed):
Thanks! 🚀

Xunzhuo:
Side comment, I think one of the attractive things EG has should be the extension abilities. So EG should support as many reasonable extending ways as possible to highlight the envoyproxy's advantages on extension abilities. This helps the gateway vendors or the teams try to integrate/build an gateway with envoyproxy. We just need to have a good and clean boundary of each extending ways.

BTW after extending EG, another main focus should be the observability, since this is one of the advantages in envoyproxy too. This is in the roadmap after v0.4.0

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+- Extension service connection info (host, port, TLS settings, etc.)
+
+An example configuration:
+
+```yaml
+apiVersion: config.gateway.envoyproxy.io/v1alpha1
+kind: EnvoyGateway
+extension:
+  apiGroups:
+  - apiGroup: example.myextension.io
```
arkodg, line 59:
```suggestion
  - group: example.myextension.io
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+    apiVersion: v2
```
arkodg, line 60:
```suggestion
    version: v2
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+- The `Kind`s supported under each API group
+- Extension service connection info (host, port, TLS settings, etc.)
+
+An example configuration:
+
+```yaml
+apiVersion: config.gateway.envoyproxy.io/v1alpha1
+kind: EnvoyGateway
+extension:
+  apiGroups:
```
arkodg, line 58:
```suggestion
  extensionRefs:
```

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+  - apiGroup: example.myextension.io
+    apiVersion: v2
+    kinds:
```
arkodg, line 61: reusing similar language as https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.LocalObjectReference

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+    - OAuth2Filter
+  service:
+    host: my-extension.example
+    port: 443
+    tls:
```
arkodg, line 66: can this look similar to https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayTLSConfig ?

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+
+This configuration is required to be provided at bootstrap and modifying the registered extension during runtime is not currently supported.
+Envoy Gateway will keep track of the registered extension and its `apiGroups` and `kinds` when processing Gateway API resources.
+
+## Extending Gateway API and the Data Plane
+
+Envoy Gateway manages [Envoy][] deployments, which act as the data plane that handles actual user traffic. Users configure the data plane using the K8s Gateway API resources which Envoy 
+Gateway converts into [Envoy specific configuration (xDS)][] to send over to Envoy.
+
+Gateway API offers [ExtensionRef filters][] and [Policy Attachments][] as extension points for implementers to use. Envoy Gateway extends the Gateway API using these extension points to provide support for [rate limiting][]
```
arkodg, line 81: how can we better frame this to highlight 2 types of extensions
* Native to the project
* Extending the control plane by the vendor/extension dev

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+
+## Control Plane Extensions
+
+To provide extensions the flexibility to do many things such as introduce a new authentication filter that introduces
+its own new [external authorization][] server, it is also necessary to be able to extend the Control plane.
+
+To power these use-cases, Envoy Gateway provides additional hook entrypoints to the Infra Manager which is responsible for dynamically creating the
+instances of Envoy Proxy. These hook points empower an extension to modify the Envoy Proxy deployment and service dynamically.
+
+In the example above where an extension developer wants to create a new `ExtensionRef` filter that provides authentication, they would likely deploy their own authorization service into the cluster
```
arkodg, line 145: is extending infra manager part of the proposal as well ?

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+The Route Hook provides a way for an extension to modify a route generated by Envoy Gateway before it is finalized.
+Doing so allows the extension to configure/modify route fields configured by Envoy Gateway and also to configure the
+Route's TypedPerFilterConfig which may be desirable to do things such as pass settings and information to
+ext_authz filters. The Route Hook also passes a list of Unstructured data for the externalRefs owned by the extension on the HTTPRoute that created this xDS route. The Route Hook will only be executed if an extension is loaded and only on Routes that were generated from an HTTPRoute that uses extension resources as externalRef filters.
+
+```go
+// RouteModifyRequest sends over the Route that was generated by Envoy Gateway along with some context about how it was generated
+// to assist the extension in making any necessary changes to the route.
+message RouteModifyRequest {
+    envoy.config.route.v3.Route route = 1;
```
arkodg, line 175: what links a `route` to a `route_context` ?

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+    // hostnames are the fully qualified domain names attached to the HTTPRoute
+    repeated string hostnames = 3;
+}
+
+// ExtensionResource stores the data for a K8s API object referenced in an HTTPRouteFilter
+// extensionRef. It is constructed from an unstructured.Unstructured marshaled to JSON. An extension
+// can marshal the bytes from this resource back into an unstructured.Unstructured and then 
+// perform type checking to obtain the resource.
+message ExtensionResource {
+    bytes unstructured_bytes = 1;
```
arkodg, line 199: lets add some structure to this

docs/latest/design/extending-envoy-gateway.md
```diff
@@ -0,0 +1,302 @@
...
+
+### HTTP Listener Modification Hook
+
+The HTTP Listener modification hook is the broadest xDS modification Hook available and allows an extension to make many changes at a higher level. The HTTP Listener Hook allows an extension to make changes to a Listener after it was generated by Envoy Gateway. It is always executed when an extension is loaded. An extension may return an unmodified version of the Listener it
+received in order to not make any changes to it, or return nil to cause the Listener to be discarded.
+
+```protobuf
+// HTTPListenerModifyRequest sends over the Listener that was generated by Envoy Gateway so that an extension can modify it
+message HTTPListenerModifyRequest {
+    envoy.config.listener.v3.Listener listener = 1;
```
arkodg, line 238: no context needed here ?

arkodg:
hey @AliceProxy thanks for another iteration. I think it might take a while for this PR to get merged if we continue to address the API level pieces in this PR. I'd be okay if we deferred the APIs / hooks into a future PR and used this design doc to highlight the intent of extending EG control plane, but I'll leave that decision to you

arkodg (approved):

Xunzhuo (approved):

Xunzhuo:
Thanks @AliceProxy !

Issue #1069, opened 2023-02-22
`egctl config` support `-A` flag


zirain:
https://github.com/envoyproxy/gateway/pull/1060#discussion_r1113455823

cc @arkodg , can you share more details here?

arkodg:
completed with https://github.com/envoyproxy/gateway/pull/1385

PR #1082, opened 2023-02-28
feat: add namespaced all proxies config dump support

Xunzhuo:
xref: https://github.com/envoyproxy/gateway/issues/1069

### DIFF
--- a/internal/cmd/egctl/config.go
+++ b/internal/cmd/egctl/config.go
@@ -6,6 +6,7 @@
 package egctl
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -17,10 +18,12 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/yaml"
 
 	"github.com/envoyproxy/gateway/internal/cmd/options"
+	"github.com/envoyproxy/gateway/internal/infrastructure/kubernetes/proxy"
 	kube "github.com/envoyproxy/gateway/internal/kubernetes"
 )
 
@@ -29,6 +32,7 @@ var (
 	podName        string
 	podNamespace   string
 	labelSelectors []string
+	allNamespaces  bool
 )
 
 const (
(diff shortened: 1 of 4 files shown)


Xunzhuo:
cc @zirain

Xunzhuo:
cc @arkodg, please take a look too.

internal/cmd/egctl/config.go
```diff
@@ -75,6 +79,12 @@ func allConfigCmd() *cobra.Command {
 
   # Retrieve full configuration dump with short syntax
   egctl c proxy all <pod-name> -n <pod-namespace>
+
+  # Retrieve all proxies configurations in a certain namespace
+  egctl config envoy-proxy all --proxies -n <pod-namespace>
```
arkodg, line 84: `proxies` seems redundant since the command already has `envoy-proxy` in it.
How about `-A/--all` which collects from all namespaces ? this doesnt solve the use case of all proxies in a namespace though :(  , lets find a solution to address both the use cases

AliceProxy: @Xunzhuo Any time to look at this recently and share your thoughts? :slightly_smiling_face:

kflynn: Maybe

```shell
egctl config envoy-proxy all -n $namespace
```

to get config from all the proxies in the namespace, and

```shell
egctl config envoy-proxy all --all-namespaces
```

to get from all proxies in all namespaces? That parallels `kubectl`...

Xunzhuo: @zirain Hi, do you think we should continue to process this PR?

arkodg: +1 on the suggestion from @kflynn 

* get from a specific pod
```
egctl config envoy-proxy all -n <namespace> <pod>
```
* get from all pods in a namespace
```
egctl config envoy-proxy all -n <namespace>
```
* get from all pods in all namespaces
```
egctl config envoy-proxy all --all-namespaces/-A
```

Xunzhuo:
image


image

image

image

Xunzhuo:
When replicas is 2

image

image

image

image

internal/cmd/egctl/config.go
Xunzhuo: Thanks,done

arkodg (approved):
🚀 ❤️

internal/cmd/egctl/config.go
```diff
@@ -113,10 +113,28 @@ func retrieveConfigDump(args []string, includeEds bool, configType envoyConfigTy
 // fetchRunningEnvoyPods gets the Pods, either based on the NamespacedName or the labelSelectors.
 // It further filters out only those Pods that are in "Running" state.
 // labelSelectors, if provided, take precedence over the pod NamespacedName.
-func fetchRunningEnvoyPods(c kube.CLIClient, nn types.NamespacedName, labelSelectors []string) ([]types.NamespacedName, error) {
+func fetchRunningEnvoyPods(c kube.CLIClient, nn types.NamespacedName, labelSelectors []string, allNamespaces bool) ([]types.NamespacedName, error) {
```
qicz, line 116: IMO `nn` is not good, prefer `name` or `namespacedName`

qicz (approved):
LGTM
some nit

internal/cmd/egctl/config.go
Xunzhuo: I did not change the previous vars, I think this is nit and not a blocker for this PR.

Issue #1045, opened 2023-02-15
Add GH workflow for `egctl`


Xunzhuo:
*Description*:

We should add `egctl` to build and test workflow, and release/(latest-release) workflow.

Subtasks:
- [x] toolings
- [x] build and test
- [x] release
- [x] latest-release
- [x] get-egctl.sh and hosted on [gateway.envoyproxy.io](https://gateway.envoyproxy.io/)

AliceProxy:
IMO default install to `/usr/local/bin` should be ok assuming it is easy and obvious although I also really like the idea of installing to the local directory and then allowing the user to move the binary to their chosen location.

arkodg:
hey @Xunzhuo can we get this into 0.4.0-rc1 ?

Xunzhuo:
@arkodg sure :)

PR #1085, opened 2023-02-28
feat: add hack and docs for installing egctl

Xunzhuo:
part of: https://github.com/envoyproxy/gateway/issues/1045

### DIFF
--- a/docs/latest/user/egctl.md
+++ b/docs/latest/user/egctl.md
@@ -2,6 +2,45 @@
 
 `egctl` is a command line tool to provide additional functionality for Envoy Gateway users.
 
+## Installing egctl
+
+This guide shows how to install the egctl CLI. egctl can be installed either from source, or from pre-built binary releases.
+
+### From The Envoy Gateway Project
+
+The Envoy Gateway project provides two ways to fetch and install egctl. These are the official methods to get egctl releases. Installation through those methods can be found below the official methods.
+
+### From the Binary Releases
+
+Every [release](https://github.com/envoyproxy/gateway/releases) of egctl provides binary releases for a variety of OSes. These binary versions can be manually downloaded and installed.
+
+1. Download your [desired version](https://github.com/envoyproxy/gateway/releases)
+2. Unpack it (tar -zxvf egctl_latest_linux_amd64.tar.gz)
+3. Find the egctl binary in the unpacked directory, and move it to its desired destination (mv bin/linux/amd64/egctl /usr/local/bin/egctl)
+
+From there, you should be able to run: `egctl help`.
+
+### From Script
+
+`egctl` now has an installer script that will automatically grab the latest release version of egctl and install it locally. (You can only run this script when v0.4.0 gets released, or please pass `-v latest` to specify the latest development version of egctl binary, otherwise you can not get the binary correctly.)
+
+You can fetch that script, and then execute it locally. It's well documented so that you can read through it and understand what it is doing before you run it.
+
+```shell
+ curl -fsSL -o get-egctl.sh https://raw.githubusercontent.com/envoyproxy/gateway/main/tools/hack/get-egctl.sh
+ chmod 700 get-egctl.sh
+# get help info of the 
+  bash get-egctl.sh --help
+# install the latest development version of egctl
+  bash get-egctl.sh -v latest
+```
+
+Yes, you can just execute below command if you want to live on the edge.
+
+```shell
+ curl https://raw.githubusercontent.com/envoyproxy/gateway/main/tools/hack/get-egctl.sh | bash 
+```
+
 ## egctl experimental translate
 
 This subcommand allows users to translate from an input configuration type to an output configuration type.
@@ -255,8 +294,7 @@ configs:
 resourceType: all
 ```  
 
-You can also use the `--type`/`-t` flag to retrieve specific output types. In the below example, we will translate the 
-Kubernetes resources (including the Gateway API resources) into xDS `route` resources.
+You can also use the `--type`/`-t` flag to retrieve specific output types. In the below example, we will translate the Kubernetes resources (including the Gateway API resources) into xDS `route` resources.
 
 ```shell
 cat <<EOF | egctl x translate --from gateway-api --to xds -t route -f -
--- a/docs/latest/user/egctl.md
+++ b/docs/latest/user/egctl.md
@@ -22,23 +22,26 @@ From there, you should be able to run: `egctl help`.
 
 ### From Script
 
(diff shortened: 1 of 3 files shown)


docs/latest/user/egctl.md
```diff
@@ -2,6 +2,45 @@
...
+
+### From Script
+
+`egctl` now has an installer script that will automatically grab the latest release version of egctl and install it locally. (You can only run this script when v0.4.0 gets released, or please pass `-v latest` to specify the latest development version of egctl binary, otherwise you can not get the binary correctly.)
+
+You can fetch that script, and then execute it locally. It's well documented so that you can read through it and understand what it is doing before you run it.
+
+```shell
+ curl -fsSL -o get-egctl.sh https://raw.githubusercontent.com/envoyproxy/gateway/main/tools/hack/get-egctl.sh
+ chmod 700 get-egctl.sh
```
zirain, line 31: chmod +x ?

docs/latest/user/egctl.md
```diff
@@ -2,6 +2,45 @@
...
+From there, you should be able to run: `egctl help`.
+
+### From Script
+
+`egctl` now has an installer script that will automatically grab the latest release version of egctl and install it locally. (You can only run this script when v0.4.0 gets released, or please pass `-v latest` to specify the latest development version of egctl binary, otherwise you can not get the binary correctly.)
+
+You can fetch that script, and then execute it locally. It's well documented so that you can read through it and understand what it is doing before you run it.
+
+```shell
+ curl -fsSL -o get-egctl.sh https://raw.githubusercontent.com/envoyproxy/gateway/main/tools/hack/get-egctl.sh
```
arkodg, line 30: this should ideally be `https://gateway.envoyproxy.io/get-egctl.sh`

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,208 @@
...
+    '--version'|-v)
+       shift
+       if [[ $# -ne 0 ]]; then
+           export DESIRED_VERSION="${1}"
+       else
+           echo -e "Please provide the desired version. e.g. --version v3.0.0 or -v canary"
+           exit 0
+       fi
+       ;;
+    '--no-sudo')
```
arkodg, line 185: why is this flag needed ?

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,208 @@
...
+    exit 1
+  fi
+  set -e
+}
+
+# help provides possible cli installation arguments
+help () {
+  echo "Accepted cli arguments are:"
+  echo -e "\t[--help|-h ] ->> prints this help"
+  echo -e "\t[--version|-v <desired_version>] . When not defined it fetches the latest release from GitHub"
```
arkodg, line 153: thoughts on making VERSION also an env var ?
e.g. how Istio does it
```
curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.17.1 TARGET_ARCH=x86_64 sh -
```

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,208 @@
...
+initOS() {
+  OS="$(uname|tr '[:upper:]' '[:lower:]')"
+
+  case "$OS" in
+    # Minimalist GNU for Windows
+    mingw*|cygwin*) OS='windows';;
+  esac
+}
+
+# runs the given command as root (detects if we are root already)
```
arkodg, line 36: I dont think we should be doing this, lets install the binary somewhere else

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env bash
+
+: "${BINARY_NAME:="egctl"}"
+: "${USE_SUDO:="true"}"
+: "${EGCTL_INSTALL_DIR:="/usr/local/bin"}"
```
arkodg, line 5: lets maybe install the binary in `~/.envoy-gateway/bin` ?
and then provide instructions to the user to add this to path to their `PATH` env var ?

arkodg:
thanks for picking this issue up @Xunzhuo !

docs/latest/user/egctl.md
Xunzhuo: Sounds good.

tools/hack/get-egctl.sh
Xunzhuo: image
I think this is more convenient, just one command, user can install eg quickly. Don`t need them to add it their `PATH`.

tools/hack/get-egctl.sh
Xunzhuo: Sounds good. Done.

docs/latest/user/egctl.md
Xunzhuo: Done.

tools/hack/get-egctl.sh
Xunzhuo: Removed.

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
+
+: "${BINARY_NAME:="egctl"}"
+: "${EGCTL_INSTALL_DIR:="/usr/local/bin"}"
+
+export EG_VERSION
```
arkodg, line 6: prefer `VERSION` to `EG_VERSION`

tools/hack/get-egctl.sh
arkodg: doesnt this require root permission ? 
how about making it explicit ? using a `-b` flag , which by default is set to
`~/.envoy-gateway/bin` and can be explicitly set to `/usr/local/bin`
example on https://func-e.io

tools/hack/get-egctl.sh
```diff
@@ -0,0 +1,171 @@
...
+  fi
+}
+
+# checkDesiredVersion checks if the desired version is available.
+checkDesiredVersion() {
+  if [ "$EG_VERSION" == "" ]; then
+    # Get tag from release URL
+    local latest_release_url="https://github.com/envoyproxy/gateway/releases"
+    if [ "${HAS_CURL}" == "true" ]; then
+      TAG=$(curl -Ls $latest_release_url | grep 'href="/envoyproxy/gateway/releases/tag/v[0-9]*.[0-9]*.[0-9]*\"' | sed -E 's/.*\/envoyproxy\/gateway\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)
```
arkodg, line 72: is a new var `TAG` needed here, or can we reuse `VERSION` ?
trying to reduce number of nouns in this script

Xunzhuo: TAG is used for auto upgrading egctl, compare eg current version with latest version.

tools/hack/get-egctl.sh
Xunzhuo: image
This can automatically ask end users to type pwd if has not root permission.

I have tried func-e, it will not ask permission from users, and it will fail.
image

arkodg (approved):

PR #1089, opened 2023-03-01
feat: add egctl only client version

Xunzhuo:
xref: #1085

### DIFF
--- a/internal/cmd/egctl/version.go
+++ b/internal/cmd/egctl/version.go
@@ -32,15 +32,24 @@ const (
 func NewVersionCommand() *cobra.Command {
 	var (
 		output string
-		client bool
+		remote bool
 	)
 
 	versionCommand := &cobra.Command{
 		Use:     "version",
 		Aliases: []string{"versions", "v"},
 		Short:   "Show version",
+		Example: `  # Show versions of both client and server.
+  egctl version
+
+  # Show versions of both client and server in Json format.
+  egctl version --output=json
+
+  # Show version of client without server.
+  egctl version --remote=false
+	  `,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(versions(cmd.OutOrStdout(), egContainerName, output, client))
+			cmdutil.CheckErr(versions(cmd.OutOrStdout(), egContainerName, output, remote))
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


Xunzhuo:
cc @zirain

internal/cmd/egctl/version.go
```diff
@@ -48,6 +49,8 @@ func NewVersionCommand() *cobra.Command {
 
 	versionCommand.PersistentFlags().StringVarP(&output, "output", "o", yamlOutput, "One of 'yaml' or 'json'")
 
+	versionCommand.PersistentFlags().BoolVarP(&client, "client", "c", false, "If only output client version")
```
zirain, line 52: `client` is not a good name, specifically should not use `c` for short.

what about add `--remote` with default value `true`?

Xunzhuo: Good idea. Thanks!

internal/cmd/egctl/version.go
```diff
@@ -48,6 +58,8 @@ func NewVersionCommand() *cobra.Command {
 
 	versionCommand.PersistentFlags().StringVarP(&output, "output", "o", yamlOutput, "One of 'yaml' or 'json'")
 
+	versionCommand.PersistentFlags().BoolVarP(&remote, "remote", "r", true, "If retrieve version from remote apiserver.")
```
zirain, line 61:
```suggestion
	versionCommand.PersistentFlags().BoolVarP(&remote, "remote", "r", true, "If true, retrieve version from remote apiserver.")
```

internal/cmd/egctl/version.go
```diff
@@ -32,14 +32,24 @@ const (
...
 	)
 
 	versionCommand := &cobra.Command{
 		Use:     "version",
 		Aliases: []string{"versions", "v"},
 		Short:   "Show version",
+		Example: `  # Show versions of both client and server.
+  egctl version
+
+  # Show versions of both client and server in Json format.
```
zirain, line 45:
```suggestion
  # Show versions of both client and server in JSON format.
```

zirain (dismissed):
lgtm, with small nits.

internal/cmd/egctl/version.go
Xunzhuo: Done. Thanks!

internal/cmd/egctl/version.go
Xunzhuo: Done. Thanks!

zirain (approved):
lgtm, thanks!

Issue #1090, opened 2023-03-01
support auto release charts and host eg charts on GitHub Repo


Xunzhuo:
*Description*:

1. support auto release eg charts
2. host Helm Charts via GitHub Pages and Releases

subtask of #1072.

Blocker:

chart-releaser-action needs to use github pages via branch not github action. Our site for now is using github action. We need to refactor our versioned docs to transfer into `gh-pages` branch, instead of github actions.

[optional *Relevant Links*:]

Use: https://github.com/helm/chart-releaser-action
Ref: https://github.com/helm/chart-releaser

arkodg:
hey @Xunzhuo can you highlight the pros and cons of hosting in GH vs something like Docker Hub which supports OCI Artifacts now

zirain:
+1 for using OCI if possible.

Xunzhuo:
I think using DockerHub is a good option too. I am very sure of the differences between them.

Let us hear some ideas from @AliceProxy.

kflynn:
+1 for DockerHub, after reading over the Docker doc above. 🙂

LanceEa:
Adding to the +1 on DockerHub for charts.

PR #1091, opened 2023-03-01
ci: transfer docs to gh-pages

Xunzhuo:
xref: #1090


zirain (changes requested):
/hold

Xunzhuo:
close in favor of helm oci support

PR #1092, opened 2023-03-01
ci: support automation helm chart release

Xunzhuo:
Resolves: #1090


haq204 (approved):
lgtm

AliceProxy (approved):

zirain (changes requested):
/hold

IMO, it's not a good idea adding chart as part of Github release artifacts. 
Hold this before we make a decision about using of OCI.

Xunzhuo:
Right, I am thinking use OCI too. I will update the PRs.

zirain:
```
helm package charts/eg/
helm push eg-0.1.0.tgz oci://docker.io/envoy
```

helm install eg oci://docker.io/envoy/eg --version 0.1.0

arkodg:
sidenote @Xunzhuo  why is the chart called `eg` and not `envoy-gateway` ?

Xunzhuo:
@arkodg This can be changed since we haven`t release helm yet. Any ideas which is better?

.github/workflows/latest_release.yaml
```diff
@@ -62,6 +62,12 @@ jobs:
 
           It is only intended for developers wishing to try out the latest features in Envoy Gateway, some of which may not be fully implemented.
 
+          We use `0.0.0` as the latest chart version to install latest envoy-gateway:
```
arkodg, line 65: is this common convention used elsewhere ?

tools/make/image.mk
```diff
@@ -118,3 +118,24 @@ push: image.push
 .PHONY: push-multiarch
 push-multiarch: ## Push docker images for multiple platforms to registry.
 push-multiarch: image.multiarch.setup go.build.multiarch image.push.multiarch
+
+
+OCI_REGISTRY ?= oci://docker.io/envoyproxy
```
arkodg, line 123: can we add these to `helm.mk` ?

tools/make/image.mk
```diff
@@ -118,3 +118,24 @@ push: image.push
...
+CHART_NAME ?= envoy-gateway
+CHART_VERSION ?= 0.1.0
```
arkodg, line 125: should this be `$(REV)` instead of defaulting to 0.1.0 ?

tools/make/image.mk
arkodg: or `kube.mk`

zirain (approved):
LGTM

PR #1094, opened 2023-03-03
Implement Bootstrap API

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/31

### DIFF
--- a/docs/latest/design/bootstrap.md
+++ b/docs/latest/design/bootstrap.md
@@ -38,7 +38,9 @@ type EnvoyProxySpec struct {
 	// Backward compatibility across minor versions is not guaranteed.
 	// We strongly recommend using `egctl x translate` to generate a `EnvoyProxy` resource with the `Bootstrap` field set to the default
 	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
-	Bootstrap *string `json:"bootstrap,omitempty"
+	//
+	// +optional
+	Bootstrap *string `json:"bootstrap,omitempty"`
 }
 ```
 
--- a/docs/latest/design/bootstrap.md
+++ b/docs/latest/design/bootstrap.md
@@ -35,7 +35,7 @@ type EnvoyProxySpec struct {
 	// from it are not configurable and will result in the `EnvoyProxy` resource being rejected.
 	// We strongly recommend using `egctl x translate` to generate a `EnvoyProxy` resource with the `Bootstrap` field set to the default
 	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
-	Bootstrap string `json:"bootstrap,omitempty"`
+	Bootstrap *string `json:"bootstrap,omitempty"`
 }
 ```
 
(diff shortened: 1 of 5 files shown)


docs/latest/design/bootstrap.md
```diff
@@ -35,7 +35,7 @@ type EnvoyProxySpec struct {
 	// from it are not configurable and will result in the `EnvoyProxy` resource being rejected.
 	// We strongly recommend using `egctl x translate` to generate a `EnvoyProxy` resource with the `Bootstrap` field set to the default
 	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
-	Bootstrap string `json:"bootstrap,omitempty"`
+	Bootstrap *string `json:"bootstrap,omitempty"`
```
zirain, line 38: why need a point here? Is there difference between nil and empty string?

arkodg: yes, nil is a better check than empty string. This is just updating docs, this change was already made to the API

zirain: Can you elaborate?

arkodg: yes thats right

zirain: so what's the difference between nil and empty string？

why nil is better？

arkodg: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required recommends pointer type for optional fields

internal/infrastructure/kubernetes/proxy_deployment.go
```diff
@@ -133,9 +133,19 @@ func expectedProxyContainers(infra *ir.Infra) ([]corev1.Container, error) {
...
 
-	cfg, err := bootstrap.GetRenderedBootstrapConfig()
-	if err != nil {
-		return nil, err
+	var cfg string
+	// Get Bootstrap from EnvoyProxy API if set by the user
+	// The config should have been validated already
+	if infra.Proxy.Config != nil &&
+		infra.Proxy.Config.Spec.Bootstrap != nil {
+		cfg = *infra.Proxy.Config.Spec.Bootstrap
```
Xunzhuo, line 141: I think we should unmarshal it to bootstrap struct to validate it if the bootstrap user provided is invalid.

internal/infrastructure/kubernetes/proxy_deployment_test.go
```diff
@@ -152,6 +153,35 @@ func TestExpectedProxyDeployment(t *testing.T) {
...
+	cli := fakeclient.NewClientBuilder().WithScheme(envoygateway.GetScheme()).WithObjects().Build()
+	kube := NewInfra(cli, svrCfg)
+	infra := ir.NewInfra()
+
+	infra.Proxy.GetProxyMetadata().Labels[gatewayapi.OwningGatewayNamespaceLabel] = "default"
+	infra.Proxy.GetProxyMetadata().Labels[gatewayapi.OwningGatewayNameLabel] = infra.Proxy.Name
+
+	// Set a custom bootstrap config into EnvoyProxy API and ensure the same
+	// value is set as the container arg.
+	bstrap := "blah"
```
Xunzhuo, line 168: This is obvious a invalid bootstrap config I think? There should be an error reported.

internal/infrastructure/kubernetes/proxy_deployment.go
arkodg: the validation will happen in the validation webhook, and the data should be valid by the time it reaches this component

internal/infrastructure/kubernetes/proxy_deployment_test.go
arkodg: as I mentioned in the prevous comment, the validation will happen earlier. Just used a dummy string to test in this component.

Xunzhuo (approved):
LGTM

docs/latest/design/bootstrap.md
zirain: make sense

PR #1095, opened 2023-03-04
use ADS for all xDS resources

arkodg:
Referred to https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#aggregated-discovery-service


arkodg:
@lizan can you ptal

zirain (approved):

PR #1104, opened 2023-03-06
chore: unify the way to create RouteDestination

### LINKED ISSUE #786: ir.RouteDestination weight  ?
*Description*:

```go
type RouteDestination struct {
	// Host refers to the FQDN or IP address of the backend service.
	Host string
	// Port on the service to forward the request to.
	Port uint32
	// Note: Weight is not used in UDP route.
	Weight uint32
}
```

In fact, there are 2 cases, 0 and nil
-  Weight 0, which is actually a legal input
-  Weight == nil ，no setting required

muyuan0:
Fix #786 
When taking a look at issue 786, I found that different places use different way to create RouteDestination.
This commit unifies the behavior.


muyuan0:
I would like to put each field into private, but it will break the yaml marshal used in the test (in/out diff). However, we can solve it if this type of refactoring is acceptable.

internal/gatewayapi/route.go
```diff
@@ -684,10 +684,12 @@ func (t *Translator) processUDPRouteParentRefs(udpRoute *UDPRouteContext, resour
 		}
 
 		// weight is not used in udp route destinations
-		routeDestinations = append(routeDestinations, &ir.RouteDestination{
-			Host: service.Spec.ClusterIP,
-			Port: uint32(*backendRef.Port),
-		})
+		weight := uint32(0)
```
arkodg, line 687: lets rm this, since it wasn't part of the orig logic

internal/gatewayapi/route.go
```diff
@@ -809,10 +811,12 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
 		}
 
 		// weight is not used in tcp route destinations
-		routeDestinations = append(routeDestinations, &ir.RouteDestination{
-			Host: service.Spec.ClusterIP,
-			Port: uint32(*backendRef.Port),
-		})
+		weight := uint32(0)
```
arkodg, line 814: same comment as above

arkodg:
thanks @muyuan0 , added some comments

internal/ir/xds.go
```diff
@@ -393,14 +394,24 @@ func (r RouteDestination) Validate() error {
...
+		Port: port,
 	}
 }
 
+func NewRouteDestWithWeight(host string, port uint32, weight uint32) *RouteDestination {
+	rd := &RouteDestination{
+		Host: host,
+		Port: port,
+	}
+	if weight != 0 {
```
arkodg, line 409: imo this check isn't needed, what if the caller wants to explicitly set the weight to zero ?

muyuan0: Addressed in the new commit.

arkodg (approved):
LGTM, thanks !

PR #1111, opened 2023-03-08
update deprecated RegexMatcher

### LINKED ISSUE #1056: update deprecated RegexMatcher
Seeing this warning message in logs when using regex for GRPCRoute match

[2023-02-17 21:11:06.049][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.type.matcher.v3.RegexMatcher Using deprecated option 'envoy.type.matcher.v3.RegexMatcher.google_re2' from file regex.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.

muyuan0:
Fix #1056

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -108,8 +108,7 @@ func buildXdsRouteMatch(pathMatch *ir.StringMatch, headerMatches []*ir.StringMat
 		} else if pathMatch.SafeRegex != nil {
 			outMatch.PathSpecifier = &routev3.RouteMatch_SafeRegex{
 				SafeRegex: &matcherv3.RegexMatcher{
-					EngineType: &matcherv3.RegexMatcher_GoogleRe2{},
-					Regex:      *pathMatch.SafeRegex,
+					Regex: *pathMatch.SafeRegex,
 				},
 			}
 		}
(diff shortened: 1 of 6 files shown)


zirain (approved):
lgmt, thanks!

internal/xds/translator/route.go
```diff
@@ -108,8 +108,7 @@ func buildXdsRouteMatch(pathMatch *ir.StringMatch, headerMatches []*ir.StringMat
 		} else if pathMatch.SafeRegex != nil {
 			outMatch.PathSpecifier = &routev3.RouteMatch_SafeRegex{
 				SafeRegex: &matcherv3.RegexMatcher{
-					EngineType: &matcherv3.RegexMatcher_GoogleRe2{},
-					Regex:      *pathMatch.SafeRegex,
+					Regex: *pathMatch.SafeRegex,
```
arkodg, line 111: shouldn this be `GoogleRe2` instead ?
https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/regex.proto#envoy-v3-api-msg-type-matcher-v3-regexmatcher

muyuan0: According to https://github.com/envoyproxy/envoy/issues/24256#issuecomment-1333007691, 
> google_re2 is required for engine_type while engine_type is optional for RegexMatcher

This PR removes the optional engine_type.

According to https://github.com/envoyproxy/gateway/blob/main/docs/latest/intro/compatibility.rst, we only need to support Envoy 1.25+ so we don't need to worry the old Envoy behavior described in https://github.com/envoyproxy/envoy/issues/24256#issuecomment-1333078240

PR #1113, opened 2023-03-08
fix(egctl): don't skip resource caused by ns mismatch

muyuan0:
Here I use the quickstart.yaml of this project as the test case.
Without this commit, the HTTPRoute in quickstart.yaml will be skipped
because it doesn't have a namespace.

### DIFF
--- a/internal/cmd/egctl/translate.go
+++ b/internal/cmd/egctl/translate.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/envoyproxy/gateway/internal/envoygateway"
+	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/gatewayapi"
 	infra "github.com/envoyproxy/gateway/internal/infrastructure/kubernetes"
 	"github.com/envoyproxy/gateway/internal/xds/bootstrap"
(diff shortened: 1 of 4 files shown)


internal/cmd/egctl/translate.go
```diff
@@ -389,6 +391,13 @@ func kubernetesYAMLToResources(str string) (string, *gatewayapi.Resources, error
 		un := unstructured.Unstructured{Object: obj}
 		gvk := un.GroupVersionKind()
 		name, namespace := un.GetName(), un.GetNamespace()
+		if namespace == "" {
```
arkodg, line 394: so the Gateway API translator doesnt like the fact that the resource `Namespace` field is unset or that the `default` Namespace resource is not part of the input ?

arkodg: another thought, rather than automagaically inserting resources by default, how about adding a flag called `--add-missing-resources` which explicitly does this on behalf on the user

muyuan0: This PR intends to emulate the behavior of kubectl, which uses the current namespace if the resources in the applied yaml don't have one.
With this change, a bundle of resources used by the kubectl to setup the environment (here the quickstart.yaml in this project) can also be accepted by the egctl.

As for `--add-missing-resources`, IMO maybe we can make it the default. Therefore the user can specify a single HTTPRoute and let egctl provide a default gateway/gatewayclass/service and so on. It would be very user-friendly if we can get rid of the boilerplate. What about discussing it in another issue?

arkodg: sure sgtm

arkodg: raised https://github.com/envoyproxy/gateway/issues/1122

arkodg (approved):
SGTM, thanks !

Issue #787, opened 2022-12-09
EDS + endpoint slow start Support


zhshw:
*Description*:
There are many types of programs that start slowly and require the flow control of the Pod Endpoint.
envoy supports slow start config feature

arkodg:
@zhshw can you please elaborate on why EDS is needed for the slow start feature ?
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig

We build static clusters in this project, and when a service endpoint changes, it will be reconciled by the controller, translated into an xds cluster and then, only the updated cluster will be pushed to envoy via delta xds

zhshw:
> 
@arkodg  For example, many Java applications start very slowly, requiring a flow warm-up process.
This is a problem that many people will face when practicing cloud native process, so aws ALb takes the lead in supporting and envoy implements Istio also supports configuration changes
service -> Pod endpoints -> lbEndpoints , envoy lbEndpoint support slow start

arkodg:
there is no strong use case for [EDS](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/dynamic_configuration#eds) yet

but the [slow start](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig) feature is a valid use case that I believe is separate. It is however blocked by https://github.com/envoyproxy/gateway/issues/35

arkodg:
reopening this issue, as the use case for incorporating EDS is to seamlessly update endpoints without disrupting existing upstream connections associated with the Cluster.

zirain:
FYI, `EDS` will be helpful on case like https://github.com/istio/istio/issues/43575

PR #1118, opened 2023-03-09
Add EDS for static endpoints

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/787


internal/xds/translator/translator.go
```diff
@@ -229,6 +218,18 @@ func findXdsCluster(tCtx *types.ResourceVersionTable, name string) *clusterv3.Cl
 	return nil
 }
 
+func addXdsCluster(tCtx *types.ResourceVersionTable, name string, destinations []*ir.RouteDestination, tSocket *corev3.TransportSocket, isHTTP2 bool, isStatic bool) {
```
zirain, line 221: IMO, the last two parameters(`isHTTP2` `isStatic`) has a smell of bad code, so you need add comments everytime you call `addXdsCluster`.

arkodg: agree, we should add a `addClusterArgs{}` in a future PR

zirain: can you open a ticket to follow up?

arkodg: raised https://github.com/envoyproxy/gateway/issues/1141

PR #1119, opened 2023-03-09
fix crash in serveXdsServer when failed to listen

muyuan0:
The serveXdsServer will crash if the net.Listen failed, because of:
1. the log call uses an odd number of arguments in argument pair
2. the code continues to execute and will access the nil listener.

This PR fixes the behavior.

### DIFF
--- a/tools/make/golang.mk
+++ b/tools/make/golang.mk
@@ -41,12 +41,13 @@ go.build.multiarch: $(foreach p,$(PLATFORMS),$(addprefix go.build., $(addprefix 
 
 .PHONY: go.test.unit
 go.test.unit: ## Run go unit tests
-	go test ./...
+	# The gcflags here ensure gomonkey always work, see https://github.com/agiledragon/gomonkey#notes
+	go test -gcflags=all=-l ./...
 
 .PHONY: go.test.coverage
 go.test.coverage: $(tools/setup-envtest) ## Run go unit and integration tests in GitHub Actions
 	@$(LOG_TARGET)
-	KUBEBUILDER_ASSETS="$(shell $(tools/setup-envtest) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... --tags=integration -race -coverprofile=coverage.xml -covermode=atomic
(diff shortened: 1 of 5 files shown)


Xunzhuo:
Hey @muyuan0 Thanks for the works.

Just one suggestion. Before getting started, could you raise an issue first, in case maintainers can have a good review to understand what we are solving now?

muyuan0:
> Hey @muyuan0 Thanks for the works.
> 
> Just one suggestion. Before getting started, could you raise an issue first, in case maintainers can have a good review to understand what we are solving now?

Thanks. I will raise an issue next time.

muyuan0:
As for this case, I think the test case is already clear enough to show my intention. That's why I didn't open an issue first.

zirain:
an issue or PR description is needed.

muyuan0:
> an issue or PR description is needed.

Updated the PR description.

tools/make/golang.mk
```diff
@@ -41,12 +41,13 @@ go.build.multiarch: $(foreach p,$(PLATFORMS),$(addprefix go.build., $(addprefix
 
 .PHONY: go.test.unit
 go.test.unit: ## Run go unit tests
-	go test ./...
+	# The gcflags here ensure gomonkey always work, see https://github.com/agiledragon/gomonkey#notes
+	go test -gcflags=all=-l ./...
```
arkodg, line 45: im reluctant to add this,
am I'm fine with the code w/o the test (with monkey patch to mock it)

muyuan0: The gomonkey can work without the no-inline flag in this case.
Could you tell me the reason why the `gcflags` is not welcome here? I always add `gcflags="all=-N -l"` with `go build` to make the debugger work usefully so I am comfortable with new gcflags...

arkodg (approved):
LGTM, thanks for fixing this and also adding the test case to validate it !

muyuan0:
Sorry for the rebase by accident!

AliceProxy (approved):
lgtm, thanks for fixing this

arkodg (approved):

Issue #1116, opened 2023-03-09
make `internal/xds/extensions/extensions.go` auto generated


zirain:
follow up: https://github.com/envoyproxy/gateway/pull/1109#discussion_r1130252198

PR #1126, opened 2023-03-10
auto generate extensions.go

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/1116


go.mod
```diff
@@ -87,6 +87,7 @@ require (
 	github.com/sirupsen/logrus v1.8.1 // indirect
 	github.com/tsaarni/x500dn v0.0.0-20210331182804-14283c7f5a16 // indirect
 	github.com/xlab/treeprint v1.1.0 // indirect
+	go.opentelemetry.io/proto/otlp v0.19.0 // indirect
```
arkodg, line 90: why is this added ?

tools/make/golang.mk
```diff
@@ -32,7 +32,7 @@ go.build.%:
 
 # Build the envoy-gateway binaries in the hosted platforms.
 .PHONY: go.build
-go.build: $(addprefix go.build., $(addprefix $(PLATFORM)., $(BINS)))
+go.build: go.generate $(addprefix go.build., $(addprefix $(PLATFORM)., $(BINS)))
```
arkodg, line 35: this shouldnt be dependancy imo

tools/make/golang.mk
```diff
@@ -67,6 +67,11 @@ go.mod.lint:
 		$(call log, "Go module looks clean!"); \
    	fi
 
+.PHONY: go.generate
+go.generate: ## Generate code from templates
```
arkodg, line 71: needs to be linked to `make generate`

internal/xds/extensions/extensions.go
```diff
@@ -3,143 +3,17 @@
...
-)
+// nolint: lll
+//
+//go:generate sh -c "echo '// Copyright Envoy Gateway Authors' > extensions.gen.go"
+//go:generate sh -c "echo '// SPDX-License-Identifier: Apache-2.0' >> extensions.gen.go"
+//go:generate sh -c "echo '// The full text of the Apache license is available in the LICENSE file at' >> extensions.gen.go"
+//go:generate sh -c "echo '// the root of the repo.\n' >> extensions.gen.go"
+//go:generate sh -c "echo '// GENERATED FILE -- DO NOT EDIT\n' >> extensions.gen.go"
+//go:generate sh -c "echo 'package extensions\n\nimport (' >> extensions.gen.go"
+//go:generate sh -c "go list github.com/envoyproxy/go-control-plane/... | grep 'v[2-9]' | grep -v /pkg/ | xargs -I{} echo '\t_ \"{}\"' >> extensions.gen.go"
```
arkodg, line 18: 🚀  thanks for automating this !

tools/make/golang.mk
zirain: you should run `go generate` before `go build`

tools/make/golang.mk
zirain: IMO, linked to gen-check is enough.

go.mod
zirain: because of `github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/open_telemetry/v3`

tools/make/golang.mk
arkodg: we already generate a lot of go code today 
https://github.com/envoyproxy/gateway/blob/630cae98c9fb6e66f303ef9174876aa6d21186b6/tools/make/kube.mk#L32
I suggest we move this target into `common.mk` and also add a `go generate` step in there

go.mod
arkodg: ah thanks for the clarification !

tools/make/golang.mk
zirain: yeah, that make sense

zirain: updated, PTAL

arkodg: thanks, it looks good, can you confirm that `make geo-check` works with no hiccups ?
can you also update the print statement in https://github.com/envoyproxy/gateway/blob/630cae98c9fb6e66f303ef9174876aa6d21186b6/tools/make/lint.mk#L72 ?

tools/make/common.mk
```diff
@@ -113,6 +113,9 @@ Options:
 endef
 export USAGE_OPTIONS
 
+.PHONY: generate
+generate: kube-generate go.generate
```
arkodg, line 117: can you add a helper string that will show up with `make help` - `## Generate go code from templates and tags`

tools/make/golang.mk
zirain: the output looks expected to me
```
make[1]: Entering directory '/root/go/src/github.com/envoyproxy/gateway'
===========> Running kube-generate ... 
tools/bin/controller-gen object:headerFile="/root/go/src/github.com/envoyproxy/gateway/tools/boilerplate/boilerplate.generatego.txt",year=2023 paths="{/root/go/src/github.com/envoyproxy/gateway/api/...,/root/go/src/github.com/envoyproxy/gateway/internal/ir/...,/root/go/src/github.com/envoyproxy/gateway/internal/gatewayapi/...}"
===========> Running go.generate ... 
go generate ./...
===========> Running manifests ... 
tools/bin/controller-gen rbac:roleName=envoy-gateway-role crd webhook paths="./..." output:crd:artifacts:config=charts/eg/crds/generated output:rbac:artifacts:config=charts/eg/templates/generated/rbac output:webhook:artifacts:config=charts/eg/templates/generated/webhook
===========> Running gen-check ... 
make[1]: Leaving directory '/root/go/src/github.com/envoyproxy/gateway'
```

tools/make/common.mk
zirain: done

arkodg (approved):
LGTM thanks for ensuring there is no drift in the future !

PR #1129, opened 2023-03-11
Implement source IP RateLimit

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/852

### DIFF
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -706,7 +706,7 @@ type IPMatch struct {
 	MaskLen int
 }
 
-func (r *RateLimitRule) MatchSetted() bool {
+func (r *RateLimitRule) IsMatchSet() bool {
 	return len(r.HeaderMatches) != 0 || r.IPMatch != nil
 }
 
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -694,20 +694,20 @@ type GlobalRateLimit struct {
 type RateLimitRule struct {
 	// HeaderMatches define the match conditions on the request headers for this route.
 	HeaderMatches []*StringMatch
-	// IPMatch define the match conditions on the source IP for this route.
-	IPMatch *IPMatch
+	// CIDRMatch define the match conditions on the source IP's CIDR for this route.
+	CIDRMatch *CIDRMatch
 	// Limit holds the rate limit values.
 	Limit *RateLimitValue
 }
 
-type IPMatch struct {
+type CIDRMatch struct {
 	CIDR    string
 	IPv6    bool
 	MaskLen int
 }
 
 func (r *RateLimitRule) IsMatchSet() bool {
-	return len(r.HeaderMatches) != 0 || r.IPMatch != nil
+	return len(r.HeaderMatches) != 0 || r.CIDRMatch != nil
 }
 
 type RateLimitUnit egv1a1.RateLimitUnit
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -694,8 +694,20 @@ type GlobalRateLimit struct {
 type RateLimitRule struct {
 	// HeaderMatches define the match conditions on the request headers for this route.
 	HeaderMatches []*StringMatch
+	// IPMatch define the match conditions on the source IP for this route.
+	IPMatch *IPMatch
 	// Limit holds the rate limit values.
 	Limit *RateLimitValue
+}
+
+type IPMatch struct {
+	CIDR    string
+	IPv6    bool
+	MaskLen int
+}
+
+func (r *RateLimitRule) MatchSetted() bool {
+	return len(r.HeaderMatches) != 0 || r.IPMatch != nil
 }
 
 type RateLimitUnit egv1a1.RateLimitUnit
--- a/internal/xds/translator/ratelimit.go
+++ b/internal/xds/translator/ratelimit.go
@@ -147,9 +147,26 @@ func buildRouteRateLimits(descriptorPrefix string, global *ir.GlobalRateLimit) [
 			}
 		}
 
+		if rule.IPMatch != nil {
+			// Setup MaskedRemoteAddress action
+			mra := &routev3.RateLimit_Action_MaskedRemoteAddress{}
+			maskLen := &wrapperspb.UInt32Value{Value: uint32(rule.IPMatch.MaskLen)}
+			if rule.IPMatch.IPv6 {
+				mra.V6PrefixMaskLen = maskLen
+			} else {
+				mra.V4PrefixMaskLen = maskLen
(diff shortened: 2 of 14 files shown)


internal/ir/xds.go
```diff
@@ -694,10 +694,22 @@ type GlobalRateLimit struct {
 type RateLimitRule struct {
 	// HeaderMatches define the match conditions on the request headers for this route.
 	HeaderMatches []*StringMatch
+	// IPMatch define the match conditions on the source IP for this route.
+	IPMatch *IPMatch
```
arkodg, line 698: why not list ?

internal/xds/translator/ratelimit.go
```diff
@@ -147,9 +147,26 @@ func buildRouteRateLimits(descriptorPrefix string, global *ir.GlobalRateLimit) [
 			}
 		}
 
+		if rule.IPMatch != nil {
+			// Setup MaskedRemoteAddress action
```
arkodg, line 151: from the spec it looks like the masked remote address is being set to RLS https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action-maskedremoteaddress
how are individual IPs getting rate limited in a unique way ?

internal/ir/xds.go
```diff
@@ -694,10 +694,22 @@ type GlobalRateLimit struct {
...
 	Limit *RateLimitValue
 }
 
+type IPMatch struct {
+	CIDR    string
+	IPv6    bool
+	MaskLen int
+}
+
+func (r *RateLimitRule) MatchSetted() bool {
```
arkodg, line 709:
```suggestion
func (r *RateLimitRule) IsMatchSet() bool {
```

internal/xds/translator/ratelimit.go
arkodg: this boils down to a top level API question, when a user specifies `192.168.1.0/24`and a limit of 2 requests/hourdoes this mean `192.168.1.1` and `192.168.1.2` are treated as the same client so limit is reached when `192.168.1.1` sends 1 request and `192.168.1.2` sends another request within the same hour ? 
Or
is `192.168.1.1` and `192.168.1.2` allowed 2 requests each ?

zirain: in that case you should use `192.168.1.1/32` and `192.168.1.1/32`

internal/ir/xds.go
zirain: https://github.com/envoyproxy/gateway/blob/9c4a15053ef0c645591795f4bbf597e2411cc81f/api/v1alpha1/ratelimitfilter_types.go#L105

you want to combine rules here?

internal/xds/translator/ratelimit.go
arkodg: thanks for clarifying , we should clarify this in the API https://github.com/envoyproxy/gateway/blob/27d14f8e9c2babe1a7c9f441114bf59b031b1b6d/api/v1alpha1/ratelimitfilter_types.go#L101 i.e. `All IP Addresses within the specified CIDR are treated as a single client selector and share the rate limit value`

In the future, we could add a `Distinct` match type so each IP is treated in a unique way

internal/ir/xds.go
```diff
@@ -694,10 +694,22 @@ type GlobalRateLimit struct {
 type RateLimitRule struct {
 	// HeaderMatches define the match conditions on the request headers for this route.
 	HeaderMatches []*StringMatch
+	// IPMatch define the match conditions on the source IP for this route.
+	IPMatch *IPMatch
 	// Limit holds the rate limit values.
 	Limit *RateLimitValue
 }
 
+type IPMatch struct {
```
arkodg, line 703: nit: `CIDRMatch` ?

internal/ir/xds.go
arkodg: I think this is non needed, my bad

internal/xds/translator/ratelimit.go
zirain: `MaskedRemoteAddress` is usefull for both Global and Local ratelimit.
In istio, we want to user local ratelimit between pod and pod, pod ip is short live, `marked_remote_address` will be very useful in this scenario.

arkodg:
hey @zirain has this been tested end to end ? would be great to track it with a GH issue so we track docs completion for this feature

internal/xds/translator/testdata/in/ratelimit-config/masked-remote-address-match.yaml
```diff
@@ -0,0 +1,22 @@
...
+- "*"
+routes:
+- name: "first-route"
+  rateLimit:
+    global:
+      rules:
+      - cidrMatch:
+          cidr: "192.168.0.0/16"
+          ipv6: false
+          ones: 16
```
arkodg, line 14: ones ?

zirain: come from:
```
// Size returns the number of leading ones and total bits in the mask.
// If the mask is not in the canonical form--ones followed by zeros--then
// Size returns 0, 0.
func (m IPMask) Size() (ones, bits int) {
	ones, bits = simpleMaskLength(m), len(m)*8
	if ones == -1 {
		return 0, 0
	}
	return
}
```

zirain: maybe change it to `maskLen`?

zirain:
open https://github.com/envoyproxy/gateway/issues/1144

internal/xds/translator/testdata/in/ratelimit-config/masked-remote-address-match.yaml
arkodg: yeah, isn't this the struct
```
type CIDRMatch struct {
 	CIDR    string
 	IPv6    bool
 	MaskLen int
 }
 ```
 so the input should be `maskLen`
 
 rateLimitConfig doesnt use `maskLen` so this test is passing :)

zirain: yeah, I found it.

arkodg (approved):
LGTM, thanks for building out this useful feature !

internal/xds/translator/ratelimit.go
arkodg: raised https://github.com/envoyproxy/gateway/pull/1145 to clarify docs

Xunzhuo (approved):
LGTM, thanks! Can you update the ratelimit doc to describe this?

PR #1130, opened 2023-03-12
fix auto generate extensions kube.mk

qicz:
@zirain maybe #1126 Miss updating. pls check and review this.


zirain:
`generate` include `kube-generate` and `go.generate`, what's the real problem?

qicz:
> `generate` include `kube-generate` and `go.generate`, what's the real problem?

my mistake, found generate in common.mk

PR #1133, opened 2023-03-13
fix: rename chart to gateway and reuse gateway dockerhub repo

Xunzhuo:
replace:
```
helm install eg oci://docker.io/envoyproxy/envoy-gateway -n envoy-gateway-system --create-namespace
```

with:
```
helm install eg oci://docker.io/envoyproxy/gateway -n envoy-gateway-system --create-namespace
```

1. reuse gateway repo
2. fix chart release failed


zirain (changes requested):
this will confict with image tag?

PR #1134, opened 2023-03-13
Revert "ci: add helm login to prevent authorization failed"

Xunzhuo:
Reverts envoyproxy/gateway#1131

After docker login, we do not need to helm login anymore.

Need to create a repo called envoy-gateway in dockerhub to release helm chart.

We cannot reuse gateway / gateway-dev , preventing conflicts and overrides the same tag.

Hold after creating a new repo in dockerhub.


Xunzhuo:
image

It is maybe the current EG using dockerhub account id/password does not have the priviledge to create a repo when pushing and found repo is not existed.

zirain (approved):

PR #1125, opened 2023-03-10
fix PathSeparatedPrefix & optimize the prefix ends with / logic.

qicz:
fix & support #1044 

the upstream receive double "/" needs envoy fix , issue https://github.com/envoyproxy/envoy/issues/26024

PR #1143, opened 2023-03-14
fix error UT

zirain:
main is broken after https://github.com/envoyproxy/gateway/pull/1118


Xunzhuo (approved):

arkodg (approved):
thanks !

arkodg:
this happened because #1118 did not rebase after https://github.com/envoyproxy/gateway/pull/1125 was merged, we should make sure GH catches this

zirain:
return `build and test` or all actions before merging?

Issue #1151, opened 2023-03-14
decision: Helm Docker Repository name and Chart versioning scheme


arkodg:
*Description*:
1. It would be ideal if the Application Image and Application Helm Chart could live in the same repository https://hub.docker.com/r/envoyproxy/gateway/tags but since you cannot have two artifacts in the same repository with the same tag (version is the tag), we will need to create a separate repository to store Helm Charts.
Brainstorming some names - `gateway-helm` , `gateway-chart`
Here's an example https://hub.docker.com/r/apache/skywalking-helm

2. The Helm Chart version/tag must be semVar based, so we need a way to represent the latest (bleeding edge) Helm Chart version that is linked to the latest release https://github.com/envoyproxy/gateway/releases/tag/latest
Suggesting `v0.0.0-latest` here, this is used in other projects such as Apache SkyWalking.

3. Do we need another repo similar to [envoyproxy/gateway-dev](https://hub.docker.com/r/envoyproxy/gateway-dev/tags) to save Helm Charts tied to every merged commit ?

arkodg:
ptal @zirain @Xunzhuo @skriss @youngnick @AliceProxy

arkodg:
@AliceProxy @kflynn and I discussed this issue in the community meeting today and this is our preference
1) Use a separate repo called `gateway-helm` to store Helm charts
2) Use `v0.0.0-latest` as the Chart version for the `latest` release
3) There is no strong use case to create a `gateway-dev-helm` repo at this moment

zirain:
copy that, No objection.

Xunzhuo:
+1 for this. Need someone to create the repo for us.

haq204:
Also +1

arkodg:
great thanks for the comments everyone, looks like we have majority consensus
@mattklein123 can you please help create a repository in Docker Hub called `gateway-helm` under `envoyproxy`, TIA !

Xunzhuo:
@mattklein123 told me that @phlax has all the creds for docker, etc.

Maybe phlax can help create one repo called `gateway-helm` for us?

cc @phlax

youngnick:
+1 from me.

arkodg:
thanks @Xunzhuo for driving this change
Here is the repo https://hub.docker.com/r/envoyproxy/gateway-helm/tags

PR #1155, opened 2023-03-15
chore: rename charts and add CI for release

Xunzhuo:
ref: https://github.com/envoyproxy/gateway/issues/1151


zirain (commented):
LGTM, @arkodg PTAL.

Xunzhuo:
I think we should merge it after `gateway-helm` is created in docker hub and make sure current 
`DOCKERHUB_USERNAME`/`DOCKERHUB_PASSWORD` can access that repo.

charts/gateway-helm/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-name: envoy-gateway
+name: gateway-helm
```
arkodg, line 2: this can still be `envoy-gateway`, right ?

charts/gateway-helm/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can also install the helm chart from the source code:
+
+To install the eg chart along with Gateway API CRDs and Envoy Gateway CRDs:
+
+``` shell
+    helm install eg --create-namespace charts/gateway-helm -n envoy-gateway-system
+```
+
+## Skip install CRDs
```
arkodg, line 24: something for the future, would be good to only install EG CRDs and skip GAPI CRDs

internal/provider/kubernetes/kubernetes_test.go
```diff
@@ -87,7 +87,7 @@ func TestProvider(t *testing.T) {
 func startEnv() (*envtest.Environment, *rest.Config, error) {
 	log.SetLogger(zap.New(zap.WriteTo(os.Stderr), zap.UseDevMode(true)))
 	gwAPIs := filepath.Join(".", "testdata", "in")
-	egAPIs := filepath.Join("..", "..", "..", "charts", "envoy-gateway", "crds", "generated")
```
arkodg, line 90: same as above, is there a good reason to keep chart name and chart repo same ?

tools/make/helm.mk
```diff
@@ -3,7 +3,7 @@
 # All make targets related to helmß are defined in this file.
 
 OCI_REGISTRY ?= oci://docker.io/envoyproxy
-CHART_NAME ?= envoy-gateway
+CHART_NAME ?= gateway-helm
```
arkodg, line 6: same as above

charts/gateway-helm/Chart.yaml
arkodg: Okay

arkodg (approved):

charts/gateway-helm/Chart.yaml
Xunzhuo: This is me, sorry forks.. Not being hacked...

AliceProxy (approved):
Lgtm, thank you!

Issue #1122, opened 2023-03-09
Allow the user to not have to specify all K8s resources in `egctl x translate --to=xds`


arkodg:
*Description*:
Today the user must specify all K8s resources when translating from `gateway-api` to `xds`. Support an argument such as `--add-missing-resources` where the user only specifies the Gateway API resources and `egctl` adds the rest (Namespaces, Services, Secrets) as dummy resources.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

muyuan0:
I would like to split this task into three steps and finish them one by one:
- [ ] provide default Namespace
- [ ] provide default GatewayClass and Gateway
- [ ] provide default Service

arkodg:
hey @muyuan0 thanks for implementing the first part for `namespaces`. 
curious if you are still looking into issue ? if not, will add the `help-wanted` label so another contributor can help out.

muyuan0:
Sorry for missing this! I have been super busy recently, so I can't give any time to implement the default service. But I will find some time to do it this weekend. Or maybe the next week.

arkodg:
no rush @muyuan0, thanks for the comment

PR #1158, opened 2023-03-15
feat(egctl): provide dummy Namespace resource

muyuan0:
Ref https://github.com/envoyproxy/gateway/issues/1122

### DIFF
--- a/internal/cmd/egctl/translate.go
+++ b/internal/cmd/egctl/translate.go
@@ -486,7 +486,19 @@ func kubernetesYAMLToResources(str string, addMissingResources bool) (string, *g
 		}
 	}
 
-	if useDefaultNamespace || addMissingResources {
+	if useDefaultNamespace {
+		if _, found := providedNamespaceMap[config.DefaultNamespace]; !found {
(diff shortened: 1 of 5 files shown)


internal/cmd/egctl/translate.go
```diff
@@ -477,23 +486,17 @@ func kubernetesYAMLToResources(str string) (string, *gatewayapi.Resources, error
 		}
 	}
 
-	if useDefaultNamespace {
-		var found bool
-		for _, ns := range resources.Namespaces {
-			if ns.GetName() == config.DefaultNamespace {
-				found = true
-				break
+	if useDefaultNamespace || addMissingResources {
```
arkodg, line 489: this logic is not quite right, if `useDefaultNamespace` it might trigger creation of multiple dummy namespaces which might not be the actual intent since `addMissingResources` is `false`.
You can either account for this complex logic or rm `useDefaultNamespace` and request the user to set `addMissingResources` to also add `default` namespace if its not included

zirain (approved):

PR #1161, opened 2023-03-16
ci: add sleep to latest release

Xunzhuo:


zirain:
you should verify it's deleted via GH cli/api?

zirain:
like a retry until?

Xunzhuo:
It could, I think the issue here may be solved by easier way, just after calling gh apis, the wait sometime to make it take effect.

.github/workflows/latest_release.yaml
```diff
@@ -29,15 +29,17 @@ jobs:
       continue-on-error: true
       run: |
         gh release delete latest --repo $GITHUB_REPOSITORY
+        sleep 3
```
arkodg, line 32: should we start with sleep only on L32 first ?

arkodg: also would be good to add a comment explaining the WHY

Xunzhuo: I think all sleeps matters. GH APIs take sometime to make effect, we should make sure before `Recreate the Latest Release and Tag` tag and release all get deleted.

And after `Recreate the Latest Release and Tag`, we need to sleep some time to wait new tag and release are created before checking it.

arkodg: Okay can you add this info as comments above each sleep ?

Xunzhuo: Yes.

arkodg (approved):
yolo

Issue #20, opened 2022-04-25
Extend EnvoyGateway Control Plane


arkodg:
Consider defining a [WASM](https://www.tetrate.io/blog/wasm-modules-and-envoy-extensibility-explained-part-1/) ABI spec for Control Plane APIGateways similar to [proxy-wasm](https://github.com/proxy-wasm/spec) so the components defined in https://github.com/envoyproxy/gateway/pull/16 can be extended by users to define new control plane APIGateway functionality . [wazero](https://github.com/tetratelabs/wazero) might be an apt runtime to use.

arkodg:
cc @AliceProxy

danehans:
xref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/config

Xunzhuo:
Do we have a more detailed proposal or design doc for it?

cc @AliceProxy  @danehans

danehans:
@AliceProxy has been assigned so I'll let her respond.

PR #1163, opened 2023-03-16
api: add extension configuration to EnvoyGateway

haq204:
Companion to https://github.com/envoyproxy/gateway/pull/1074. This adds the configuration necessary in `EnvoyGateway` to register an extension

Relates to https://github.com/envoyproxy/gateway/issues/20


api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +156,56 @@ type RateLimitRedisSettings struct {
 	URL string `json:"url"`
 }
 
+// Extension defines the configuration for registering an extension to
+// the Envoy Gateway control plane.
+type Extension struct {
```
arkodg, line 161: does this need to be further split into which pre/post hooks the extension is registering for 
* provider
* gateway-api
* xds-server
* infra-manager
?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +156,56 @@ type RateLimitRedisSettings struct {
...
+	//
+	// +unionDiscriminator
+	Type TLSType `json:"type"`
+
+	// Secret defines which K8s secret to load the TLS certificate from.
+	//
+	// +optional
+	Secret *TLSSecret `json:"secret,omitempty"`
+
+	// File defines the configuration for loading the TLS certificate from the filesystem.
```
arkodg, line 203: is this needed as we start off ?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +156,56 @@ type RateLimitRedisSettings struct {
...
+
+	// TLS defines TLS configuration for communication between Envoy Gateway and
+	// the extension service.
+	//
+	// +optional
+	TLS *ExtensionTLS `json:"tls,omitempty"`
+}
+
+// ExtensionTLS defines the TLS configuration when connecting to an extension service.
+type ExtensionTLS struct {
```
arkodg, line 190: can this be defined similar to https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayTLSConfig

api/config/v1alpha1/shared_types.go
```diff
@@ -10,6 +10,14 @@ const (
 	DefaultEnvoyReplicas = 1
 )
 
+// GroupVersionKind unambiguously identifies a Kind.
+// It can be converted to k8s.io/apimachinery/pkg/runtime/schema.GroupVersionKind
+type GroupVersionKind struct {
```
arkodg, line 15: can this be reused from upstream ?

haq204: The main addition here is that it is JSON serializable which the upstream version afaik isn't. 
And I don't think Go always adding additional methods (like a `MarshallJSON` method) to types from other packages.

api/config/v1alpha1/envoygateway_types.go
haq204: `GatewayTLSConfig` can only fetch tls certs from secrets. where `ExtensionTLS` makes it possible to fetch from other sources like files or remote secret stores in the future. This would particularly handy for our more security conscious users who don't want to place certs in K8s secrets.

Also certain fields in `GatewayTLSConfig` kinda don't make sense in this context such as `TLSModeType`.

api/config/v1alpha1/envoygateway_types.go
haq204: for the initial iteration I would say no. I was planning to update the `config.Validate` method to flag if this is set and say it's not supported yet similar to what we do now for `FileProvider`

api/config/v1alpha1/envoygateway_types.go
arkodg: I was alluding to the `certRefs` style , later a `File` kind could be supported, wdyt ?

api/config/v1alpha1/envoygateway_types.go
haq204: Personally, I don't think that is something a user configuring this really needs to know about since that to me is an implementation detail.

One idea I was thinking is on startup, EG would reach out to the extension service to ask what hooks is it going to support and can mark that in its internal extension registry in memory.

But @AliceProxy might have other thoughts on that.

api/config/v1alpha1/envoygateway_types.go
haq204: making `ExtensionTLS.Secret` `[]SecretObjectReference` which would be the same as `certificateRefs` I think is fine.

api/config/v1alpha1/envoygateway_types.go
arkodg: I personally dont see a user managing this EG, but rather a vendor, who would take on this responsibility of bootstrapping EG

arkodg: this would look similar to envoy
* there is a filter that is registered (or a hook) e.g. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/ratelimit/v3/rls.proto#rate-limit-service-proto
* there is spec defined for request/response e.g. https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto

haq204: I think that's fair - though what are your thoughts and making this a separate config file that would be managed by a vendor to clearly separate what parts of the config are intended to be managed by users and what parts are managed by extension vendors?

api/config/v1alpha1/envoygateway_types.go
haq204: I've updated so that `ExtensionTLS` now contains a `certificateRef` field similar to `GatewayTLSConfig`.

For now it would only accept K8s Secret references (which the default anyways) but we can hijack `SecretObjectReference.Kind` to also refer to `File` type when we want to support that. i.e to configure setting the tls config to point to a certificate on the filesystem inside the container (e.g. attached as a volume) it can be configured as 
```
tls:
  certificateRef:
    kind: File
    object: path/to/cert/on/filesystem
```

haq204: It's also just `SecretObjectReference` and not `[]SecretObjectReference` since I can't think of a reason why you want to refer to multiple certificate references in this context.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +141,58 @@ type RateLimitRedisSettings struct {
...
+	// Resources defines the set of K8s resources the extension will handle.
+	//
+	// +optional
+	Resources []GroupVersionKind `json:"resources,omitempty"`
+
+	// Hooks defines the set of hooks the extension supports
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	Hooks []ExtensionHook `json:"hooks,omitempty"`
```
arkodg, line 156: should this be tiered - as in define `XDSVirtualHost` under `post` under `xdsServer` ?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +141,58 @@ type RateLimitRedisSettings struct {
...
+
+	// Service defines the configuration of the extension service that the Envoy
+	// Gateway Control Plane will call through extension hooks.
+	//
+	// +kubebuilder:validation:Required
+	Service *ExtensionService `json:"service,omitempty"`
```
arkodg, line 162: should `service` be defined per hook or maybe this way makes it less repetitive , and in the future extension can be made plural

haq204: we only allow one extension to be registered so don't see a real benefit in that

api/config/v1alpha1/envoygateway_types.go
haq204: if you think that's more readable then sure though I think the only benefit in that is a very tiny small bump in look ups? but I can't imagine the number of eventual hooks to be so large it would make a meaningful difference

haq204: @arkodg so looking into this further, making it tiered is sorta more cumbersome to do validation on the hooks that are registered and I'm not sure what significant benefit we get from making it tiered but let me know what you think.

arkodg: the way I was thinking about this was -
* EG has a few components/runners (provider, xds-server etc)
* Each one subscribes and publishes a data type
* hooks can be added before or after to modify the data right before its published for the next stage (post) or right after subscription before ingesting the data (pre)

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +141,76 @@ type RateLimitRedisSettings struct {
...
+}
+
+// ExtensionHooks defines extension hooks across all supported runners
+type ExtensionHooks struct {
+	// XDS defines all the supported extension hooks for the XDS runner
+	XDS *XDSHooks `json:"xds,omitempty"`
```
arkodg, line 167: is this tied to the [xds-server](https://github.com/envoyproxy/gateway/blob/main/internal/xds/server/runner/runner.go) or [xds-translator](https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/runner/runner.go)

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -135,6 +141,76 @@ type RateLimitRedisSettings struct {
...
+}
+
+// RunnerHooks is a generic type that contains all the pre and post hook for a runner.
+// We disable DeepCopy generation and instead concretely instantiate types so that controller-gen
```
arkodg, line 171: this is because controller-gen can't deal with generics ?

haq204: yea when it generates the DeepCopy methods it doesn't include a type instantiation leading to a compile error

api/config/v1alpha1/envoygateway_types.go
haq204: XDS translator. I'll update to make it more clear

arkodg (approved):
LGTM, thanks for addressing all the comments, hoping this API unlocks a wider set of use cases for the ecosystem !

Xunzhuo (approved):
Looks good

arkodg:
> just one small question, should `Extension` be a slice?

good point, either we make it `Extensions` and add a limit of 1 or deprecate this field in favor of `Extensions` when we support multiple

zirain:
ok, let's review that later, LGTM for now.

zirain (approved):

haq204:
As stated in https://github.com/envoyproxy/gateway/pull/1074 we made the decision to not support multiple extensions at the same time:

- Too difficult to make sure extensions wouldn't step on each other's toes or conflict
- Realistically few people would be running multiple extensions at the same time. Smaller extensions to Envoy Gateway can be accommodated later via https://github.com/envoyproxy/gateway/issues/24 and https://github.com/envoyproxy/gateway/issues/71 which can co-exist with this form of extensions

arkodg:
> @arkodg should get #1074 merged?

we discussed this in the community meeting, and @AliceProxy and @haq204 mentioned they will raise some PRs for the Hooks API to make sure it is in line with the design doc contents

PR #1179, opened 2023-03-18
ci: don't run build & test for docs

### LINKED ISSUE #1159: don't run build & test for docs in CI
*Description*:
would be used.

Hi, folks. I notice that even a typo fix in the doc will trigger the conformance test, which is not so necessary. Therefore I refactor the workflow:
* move the common checks to a separate workflow
* add path scope for the workflows so they will be triggered on demand.

Here is the output:
https://github.com/muyuan0/gateway/pull/1 (coverage test doesn't pass because I haven't setup the codecov).

If permitted, I would submit a PR with it.

[optional *Relevant Links*:]

muyuan0:
Fix #1159

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -5,7 +5,8 @@ on:
     - "main"
     - "release/v*"
     paths-ignore:
-    - "**/*.png"
+    - 'docs/**'
+    - '**/*.md'
   pull_request:
     branches:
     - "main"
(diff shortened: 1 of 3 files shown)


arkodg (changes requested):
lets keep all the `steps` in this file, its easier to reason about it

.github/workflows/build_and_test.yaml
```diff
@@ -5,36 +5,22 @@ on:
     - "main"
     - "release/v*"
     paths-ignore:
-    - "**/*.png"
+    - 'docs/**'
```
arkodg, line 8: as I mentioned earlier, we should always run the workflow on a `push`, and not ignore any paths, 
the `docs` path can be ignored on a `pull_request`

arkodg:
rethinking, we can't ignore the `docs` path for pull requests since we need to run lint/spell checks on it.
I prefer if we left things as is

.github/workflows/build_and_test.yaml
muyuan0: Sorry by editing the wrong fields!

PR #1183, opened 2023-03-19
refactor: egctl root cmd & config commands.

qicz:
- refactor egctl root cmd with the same way like eg
- refactor egctl config commands
- remove unnecessary new line


Xunzhuo (approved):
Generally LGTM, thanks!

cc @zirain

zirain (approved):

PR #1186, opened 2023-03-19
ci: don't run build & test for docs in CI

### LINKED ISSUE #1159: don't run build & test for docs in CI

muyuan0:
Fix #1159

I find a way to skip the tests when the change is doc only, with minimal change on the current workflow.

The effect of doc only change can be seen at https://github.com/muyuan0/gateway/pull/3/
The effect of doc&code change can be seen at https://github.com/muyuan0/gateway/pull/4/

And now the push event will trigger the coverage-test whether it is doc only change or not.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -35,7 +35,22 @@ jobs:
     - uses: ./tools/github-actions/setup-deps
     - run: make -k licensecheck
 
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          code:
+            - '!docs/**'
+
   coverage-test:
+    needs: changes
+    if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3


zirain:
cc @Xunzhuo

arkodg:
im a -1 on this change atm, we are adding complexity (running an extra user based GH and adding checks) to ensure that we dont run tests (which adds a few more min of GH runner time), and I would prefer to keep things simple rn.

kflynn (changes requested):
I'm not sure if this change will cause `build` (which depends on `coverage-test`) to be skipped, or to fail -- has that been checked? 🙂

muyuan0:
> I'm not sure if this change will cause `build` (which depends on `coverage-test`) to be skipped, or to fail -- has that been checked? 🙂

It will be skipped, you can see the output of https://github.com/muyuan0/gateway/pull/3

muyuan0:
> im a -1 on this change atm, we are adding complexity (running an extra user based GH and adding checks) to ensure that we dont run tests (which adds a few more min of GH runner time), and I would prefer to keep things simple rn.

What about keeping it open in case it's in need in the future?

arkodg:
> > im a -1 on this change atm, we are adding complexity (running an extra user based GH and adding checks) to ensure that we dont run tests (which adds a few more min of GH runner time), and I would prefer to keep things simple rn.
> 
> What about keeping it open in case it's in need in the future?

sure, sgtm

AliceProxy:
We're going to close this one for now but if this is desired in the future then we can re-open.

kflynn:
Thanks much for working on this!

PR #1187, opened 2023-03-20
docs: update latest user guide with helm

Xunzhuo:


zirain (approved):
LGTM
after switch to Hugo, we can reuse content with `boierplates`.

arkodg (approved):

PR #1199, opened 2023-03-21
feat(egctl): support showing processed k8s resources

### LINKED ISSUE #1120: Improve the observability of the egctl
*Description*:

When I played with egctl's translate feature with https://github.com/envoyproxy/gateway/blob/main/examples/kubernetes/quickstart.yaml, I found it doesn't generate the correct RDS. At first glance, I suspect it's Envoy gateway's bug when translating xDS. However, the same configuration works with Envoy gateway well when I applied it via kubectl.

After hours of debugging, I found that the HTTPRoute is skipped silently because its namespace is empty, and there is a namespace check in the Envoy gateway. It wasn't an issue when the same configuration is applied by kubectl because kubectl will fix the empty namespace with the current namespace.

Considering the resources in almost every yaml example in https://github.com/envoyproxy/gateway/tree/main/examples/kubernetes don't have an explicit namespace, I thought it would be better to let egctl work like kubectl, and submitted a PR for it: https://github.com/envoyproxy/gateway/pull/1113

There is another thing I want to discuss here egctl doesn't have enough logs when a resource is rejected/skipped. It would be great to provide details 
when generating the xDS so people don't need to spend hours debugging it. In k8s, people can know a resource's state by using the corresponding method to query it. However, there can't be a similar way to do so when we use egctl to deal with the resources. 

My premature thought is to add a log in https://github.com/envoyproxy/gateway/blob/d70e2c6e7abd5018b79bdd6f2e2247aab3cea4cb/internal/gatewayapi/contexts.go#L83 and other similar places when the code is called in the egctl mode.

What about your opinions?

[optional *Relevant Links*:]

muyuan0:
Fix #1120


internal/cmd/egctl/translate.go
```diff
@@ -195,58 +209,139 @@ func translate(w io.Writer, inFile, inType, outType, output, resourceType string
...
+			results = append(results, resources)
+		}
+
+		if toXds {
+			keys := []string{}
+			for key := range gRes.XdsIR {
+				keys = append(keys, key)
+			}
+			// Make output stable
+			sort.Strings(keys)
```
arkodg, line 246: imo it should be inherently stable, we have `sort` inside the gateway-api lib, is that not the case ?

internal/cmd/egctl/testdata/translate/out/echo-gateway-api.cluster.yaml
```diff
@@ -0,0 +1,55 @@
+HTTPRoutes:
```
arkodg, line 1: can this be small case ?

internal/cmd/egctl/translate_test.go
```diff
@@ -26,6 +98,7 @@ func TestTranslate(t *testing.T) {
 		resourceType string
 		extraArgs    []string
 		expect       bool
+		filterFunc   func(string) string
```
arkodg, line 101: these filter make it very hard for me to understand what the final output will look like

internal/cmd/egctl/translate.go
```diff
@@ -195,58 +209,139 @@ func translate(w io.Writer, inFile, inType, outType, output, resourceType string
...
+
+		// Translate from Gateway API to Xds IR
+		gTranslator := &gatewayapi.Translator{
+			GatewayClassName:       v1beta1.ObjectName(gcName),
+			GlobalRateLimitEnabled: true,
+		}
+		gRes := gTranslator.Translate(resources)
+
+		results := []interface{}{}
+		if echoGatewayAPI {
```
arkodg, line 236: code looks good !

internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
```diff
@@ -1,3 +1,4 @@
+[
```
arkodg, line 1: its an array since we have multiple `configKey`s ?

internal/cmd/egctl/testdata/translate/out/echo-gateway-api.cluster.yaml
muyuan0: Do you mean `HttpRoutes`? It seems the official name is HTTPRoute: https://gateway-api.sigs.k8s.io/concepts/api-overview/#httproute

internal/cmd/egctl/translate.go
muyuan0: The map iteration after the gateway-api translation doesn't have a stable order:
https://github.com/envoyproxy/gateway/blob/384577334dea2c0ba28800643a2b94fea4eed244/internal/cmd/egctl/translate.go#L220

internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
muyuan0: Excuse me, I could not quite understand the words. It is already an array.

internal/cmd/egctl/translate_test.go
muyuan0: I would like to remove it, but unfortunately, there are some timestamp fields like `lastTransitionTime: "2023-03-21T12:03:11Z"` which need to be stripped from the expected data.

arkodg: we skip timestamp in many places, here's an example
https://github.com/envoyproxy/gateway/blob/384577334dea2c0ba28800643a2b94fea4eed244/internal/gatewayapi/translator_test.go#L103

internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
arkodg: yah so this was a bug, you fixed by adding the `[` and `]` ?

internal/cmd/egctl/translate.go
arkodg: thanks for clarifying that, can you add that as a comment - `Make output stable since XdsIR is a map`

internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
muyuan0: Yes. The previous output is:
```
{}
{}
```
Now it become:
```
[
{},
{}
]
```

internal/cmd/egctl/testdata/translate/out/echo-gateway-api.cluster.yaml
arkodg: I meant `httpRoutes`

internal/cmd/egctl/translate_test.go
arkodg: thanks for the clarification, I understand the issue better now, do you prefer this approach instead of creating some typed field instead of `[]interface{}{}` ?

muyuan0: Yes, I will remove `[]interface{}{}` from the filterFunc.

arkodg (approved):
LGTM thanks for introducing this feature, will be super useful for end users to debug their config even before they apply it !
 I prefer if we were using stronger types in the output, but that can be taken care of in a future PR.

zirain (approved):

Issue #1198, opened 2023-03-21
Conformance tests failing on main


arkodg:
*Description*:
Seeing conformance tests fail on main. Here is the snippet of the error logs
```
/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v0.6.2/conformance/utils/suite/suite.go:262
[422](https://github.com/envoyproxy/gateway/actions/runs/4475406797/jobs/7864924009?pr=1153#step:6:422)
        	            				/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v0.6.2/conformance/utils/suite/suite.go:220
[423](https://github.com/envoyproxy/gateway/actions/runs/4475406797/jobs/7864924009?pr=1153#step:6:423)
        	Error:      	Received unexpected error:
[424](https://github.com/envoyproxy/gateway/actions/runs/4475406797/jobs/7864924009?pr=1153#step:6:424)
        	            	timed out waiting for the condition
[425](https://github.com/envoyproxy/gateway/actions/runs/4475406797/jobs/7864924009?pr=1153#step:6:425)
        	Test:       	TestGatewayAPIConformance/HTTPExactPathMatching
[426](https://github.com/envoyproxy/gateway/actions/runs/4475406797/jobs/7864924009?pr=1153#step:6:426)
        	Messages:   	error waiting for Gateway to have at least one IP
```

PR #1157, opened 2023-03-15
Add dynamically setup containerPorts in proxy infrastructure

ShawnHXH:
Add dynamically setup containerPorts in proxy infrastructure

refer to #1032

PR #1200, opened 2023-03-21
Fix `invalid: protocol: Unsupported value: "HTTP"` error

quyenhoang96:
When apply config as bellow:
```
kind: GatewayClass
metadata:
  name: gateway-class
spec:
---
kind: Gateway
metadata:
  name: shared-gw
  namespace: qat-infra
spec:
  gatewayClassName: gateway-class
  listeners:
    - name: http
      protocol: HTTP
      port: 8080
      allowedRoutes:
        namespaces:
          from: All
```

It throws an error and does not create envoy instance:

`2023-03-22T01:02:38.760+0700    ERROR   runner/runner.go:68     failed to create new infra      {"runner": "infrastructure", "error": "failed to update deployment envoy-gateway-system/envoy-qat-infra-shared-gw-7161742d: Deployment.apps \"envoy-qat-infra-shared-gw-7161742d\" is invalid: spec.template.spec.containers[0].ports[0].protocol: Unsupported value: \"HTTP\": supported values: \"SCTP\", \"TCP\", \"UDP\""}
`

### DIFF
--- a/internal/infrastructure/kubernetes/proxy_deployment.go
+++ b/internal/infrastructure/kubernetes/proxy_deployment.go
@@ -127,13 +127,25 @@ func (i *Infra) expectedProxyDeployment(infra *ir.Infra) (*appsv1.Deployment, er
 }
 
 func expectedProxyContainers(infra *ir.Infra) ([]corev1.Container, error) {
+	// Define slice to hold container ports
 	var ports []corev1.ContainerPort
+
+	// Iterate over listeners and ports to get container ports
 	for _, listener := range infra.Proxy.Listeners {
 		for _, p := range listener.Ports {
+			var protocol corev1.Protocol
+			switch p.Protocol {
+			case ir.HTTPProtocolType, ir.HTTPSProtocolType, ir.TLSProtocolType, ir.TCPProtocolType:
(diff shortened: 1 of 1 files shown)


arkodg:
thanks for fixing this, this regression is caused by https://github.com/envoyproxy/gateway/pull/1157
prefer if the same style of code as https://github.com/envoyproxy/gateway/blob/a4afb8cc8dc08ebf0110f5ce11a747d8cb647614/internal/infrastructure/kubernetes/proxy_service.go#L26
was used here as well

arkodg:
also it looks like the test only checks for port and not protocol, would be great if that can be enhanced as part of this PR, TIA !
https://github.com/envoyproxy/gateway/blob/a4afb8cc8dc08ebf0110f5ce11a747d8cb647614/internal/infrastructure/kubernetes/proxy_deployment_test.go#L89

arkodg:
this PR fixes https://github.com/envoyproxy/gateway/issues/1198, thanks @quyenhoang96 !

arkodg (approved):
approving this for now since it makes `main` green again.
please consider adding the suggestions in a follow up PR

Issue #1202, opened 2023-03-22
Support Managed Envoy Service Type


qicz:
support `ClusterIP` for cluster-internal proxy: just for cluster-internal
support `Loadbalancer` for edge proxy: can work for internal with eg svc and for internet directly
but doesn't support `NodePort`

ref: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types

- [x] provide custom proxy service spec API 
- [x] API doc
- [x] user face doc

qicz:
pls assign to me

arkodg:
@qicz can you elaborate on which specific fields in the Envoy Service you would like to customize and why ?

qicz:
> @qicz can you elaborate on which specific fields in the Envoy Service you would like to customize and why ?

clusterip or loadbalancer

arkodg:
+1 for `ClusterIP` and `Loadbalancer` `type` https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types. Can you please update the GH issue to be specific about this feature .

sidenote - I dont think we can support `Nodeport` unless we are allowed to specify `ip:port` in Status.Addresses. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayAddress
would this be okay @youngnick  ?

qicz:
> +1 for `ClusterIP` and `Loadbalancer` `type` https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types. Can you please update the GH issue to be specific about this feature .
> 
> sidenote - I dont think we can support `Nodeport` unless we are allowed to specify `ip:port` in Status.Addresses. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayAddress would this be okay @youngnick ?

@arkodg updated, can you review again?

arkodg:
went ahead and also updated title, thanks in advance for working on this @qicz !

zirain:
should EG provide `ClusterIP` and `Loadbalancer` and same time? or multi `Loadbalancer` service?

one case is provide internal lb and internet lb at the same time.

qicz:
> should EG provide `ClusterIP` and `Loadbalancer` and same time? or multi `Loadbalancer` service?
> 
> one case is provide internal lb and internet lb at the same time.

imo, the two type don't have to exist at the same time. 
clusterip, just for internal
[...]

PR #1205, opened 2023-03-23
feat: support managed envoy service type Loadbalancer & ClusterIP.

qicz:
support #1202

support `ClusterIP` for cluster-internal proxy: just for cluster-internal
support `Loadbalancer` for edge proxy: can work for internal with eg svc and for internet directly
but doesn't support `NodePort`


api/config/v1alpha1/shared_types.go
```diff
@@ -67,6 +81,13 @@ type KubernetesServiceSpec struct {
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// Type determines how the Service is exposed. Defaults to LoadBalancer.
+	// Valid options are ClusterIP and LoadBalancer.
+	// "LoadBalancer" means a service will be exposed via an external load balancer (if the cloud provider supports it).
+	// "ClusterIP" means a service will only be accessible inside the cluster, via the cluster IP.
+	// +optional
+	Type ServiceType `json:"type,omitempty"`
```
zirain, line 89: use `ClusterIP` as default value

// +kubebuilder:default:="ClusterIP
// +kubebuilder:validation:Enum=ClusterIP;LoadBalancer

internal/infrastructure/kubernetes/proxy_service.go
```diff
@@ -46,11 +46,13 @@ func (i *Infra) expectedProxyService(infra *ir.Infra) (*corev1.Service, error) {
...
-	svcCfg := provider.GetKubeResourceProvider().EnvoyService
-	if svcCfg != nil {
-		annotations = svcCfg.Annotations
-
+	envoyServiceConfig := provider.GetKubeResourceProvider().EnvoyService
+	if envoyServiceConfig.Annotations != nil {
+		annotations = envoyServiceConfig.Annotations
 	}
+	serviceSpec := serviceSpecByServiceType(envoyServiceConfig.Type)
+	serviceSpec.Ports = ports
```
zirain, line 54: move to `serviceSpecByServiceType`

qicz: i think its unnecessary,  `serviceSpecByServiceType` func just do different by type

api/config/v1alpha1/shared_types.go
qicz: The current version uses `LoadBalancer ` by default, IMO change this is not friendly to users who already use it.

qicz: @arkodg Any other suggestion for this?

internal/xds/translator/cluster.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
-func buildXdsCluster(routeName string, tSocket *corev3.TransportSocket, protocol ProtocolType, c EndpointType) *clusterv3.Cluster {
+func buildXdsCluster(routeName string, tSocket *corev3.TransportSocket, protocol ProtocolType, endpointType EndpointType) *clusterv3.Cluster {
```
arkodg, line 23: unrelated to this PR, please raise another PR to add this

qicz: done

qicz:
@arkodg pls review again, thanks

api/config/v1alpha1/validation/envoyproxy.go
```diff
@@ -41,17 +41,41 @@ func validateEnvoyProxySpec(spec *egcfgv1a1.EnvoyProxySpec) error {
 	if spec == nil {
 		errs = append(errs, errors.New("spec is nil"))
 	}
-	if spec != nil && spec.Provider != nil && spec.Provider.Type != egcfgv1a1.ProviderTypeKubernetes {
-		errs = append(errs, fmt.Errorf("unsupported provider type %v", spec.Provider.Type))
+
+	// validate provider
+	providerErrs := validateProvider(spec)
+	if spec != nil && spec.Provider != nil && providerErrs != nil {
```
arkodg, line 47: is this check needed here, can it be moved within `validateProvider` ?

arkodg: so this becomes
```
if len(providerErrs) != 0 {
    append.....
}
```

api/config/v1alpha1/validation/envoyproxy.go
```diff
@@ -41,17 +41,41 @@ func validateEnvoyProxySpec(spec *egcfgv1a1.EnvoyProxySpec) error {
...
 }
 
+func validateProvider(spec *egcfgv1a1.EnvoyProxySpec) []error {
+	if spec != nil && spec.Provider != nil {
+		var errs []error
+		if spec.Provider.Type != egcfgv1a1.ProviderTypeKubernetes {
+			errs = append(errs, fmt.Errorf("unsupported provider type %v", spec.Provider.Type))
+		}
+		if spec.Provider.Kubernetes != nil && spec.Provider.Kubernetes.EnvoyService != nil {
+			serviceType := spec.Provider.Kubernetes.EnvoyService.Type
```
arkodg, line 67: this will not be needed once its a pointer

internal/envoygateway/config/config.go
```diff
@@ -92,14 +92,14 @@ func (s *Server) Validate() error {
 		}
 
 		if s.EnvoyGateway.Extension.Service.TLS != nil {
```
arkodg, line 94: please remove this diff from this PR since its not related

internal/infrastructure/kubernetes/proxy_service.go
arkodg: +1 either move all service spec code to one func (`expectedProxyServiceSpec()`) or leave them here (and rm `serviceSpecXX` ), else it makes it very hard to understand final intent

arkodg:
@qicz we will also need to fetch the right IP address and set it in the status here, for this implementation type to be conformant
https://github.com/envoyproxy/gateway/blob/384577334dea2c0ba28800643a2b94fea4eed244/internal/status/gateway.go#L32
and ensure conformance tests pass
https://github.com/envoyproxy/gateway/blob/main/docs/latest/dev/README.md#run-gateway-api-conformance-tests
can be done in this PR or a follow up (if so lets track it with an issue)

api/config/v1alpha1/validation/envoyproxy.go
```diff
@@ -41,17 +41,41 @@ func validateEnvoyProxySpec(spec *egcfgv1a1.EnvoyProxySpec) error {
...
+			if serviceType == "" {
```
zirain, line 68: will this happen when crd has a default value?

api/config/v1alpha1/validation/envoyproxy.go
```diff
@@ -41,17 +41,41 @@ func validateEnvoyProxySpec(spec *egcfgv1a1.EnvoyProxySpec) error {
...
+				errs = append(errs, errors.New("envoy service type can not be empty"))
+			} else if serviceType != egcfgv1a1.ServiceTypeLoadBalancer && serviceType != egcfgv1a1.ServiceTypeClusterIP {
```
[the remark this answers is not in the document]
zirain: I think api-server will do this check?

qicz:
> @qicz we will also need to fetch the right IP address and set it in the status here, for this implementation type to be conformant
> 
> https://github.com/envoyproxy/gateway/blob/384577334dea2c0ba28800643a2b94fea4eed244/internal/status/gateway.go#L32
> 
> 
> and ensure conformance tests pass
> https://github.com/envoyproxy/gateway/blob/main/docs/latest/dev/README.md#run-gateway-api-conformance-tests
> can be done in this PR or a follow up (if so lets track it with an issue)

I will follow up with a new issue

qicz:
@arkodg all case resolved pls review again

api/config/v1alpha1/validation/envoyproxy.go
```diff
@@ -41,17 +41,47 @@ func validateEnvoyProxySpec(spec *egcfgv1a1.EnvoyProxySpec) error {
...
+		}
+	}
+	return errs
+}
+
+func validateServiceType(spec *egcfgv1a1.EnvoyProxySpec) []error {
+	var errs []error
+	if spec.Provider.Kubernetes != nil && spec.Provider.Kubernetes.EnvoyService != nil {
+		serviceType := spec.Provider.Kubernetes.EnvoyService.Type
+		if *serviceType != egcfgv1a1.ServiceTypeLoadBalancer && *serviceType != egcfgv1a1.ServiceTypeClusterIP {
```
zirain, line 78: trying to apply following configuration:
```
kind: EnvoyProxy
metadata:
  name: example-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        type: NodePort
```
you will get following message: `The EnvoyProxy "example-config" is invalid: spec.provider.kubernetes.envoyService.type: Unsupported value: "NodePort": supported values: "LoadBalancer", "ClusterIP"`
IMO, this check is redundant。

qicz: so you mean the API-server has checked this?

qicz: But I think the validation here is necessary for other modules inside eg

api/config/v1alpha1/shared_types.go
```diff
@@ -67,6 +82,14 @@ type KubernetesServiceSpec struct {
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// Type determines how the Service is exposed. Defaults to LoadBalancer.
+	// Valid options are ClusterIP and LoadBalancer.
+	// "LoadBalancer" means a service will be exposed via an external load balancer (if the cloud provider supports it).
+	// "ClusterIP" means a service will only be accessible inside the cluster, via the cluster IP.
+	// +kubebuilder:default:="LoadBalancer"
+	// +optional
+	Type *ServiceType `json:"type,omitempty"`
```
zirain, line 91: if `ServiceType` is optional, default value will not work, after following CR applied
```console
kind: EnvoyProxy
metadata:
  name: example-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService: {}
```

you can check EnvoyProxy by running:
```
k get envoyproxy -oyaml example-config 
kind: EnvoyProxy
metadata:
  annotations:
      {"apiVersion":"config.gateway.envoyproxy.io/v1alpha1","kind":"EnvoyProxy","metadata":{"annotations":{},"name":"example-config","namespace":"default"},"spec":{"provider":{"kubernetes":{"envoyService":{}},"type":"Kubernetes"}}}
  creationTimestamp: "2023-03-24T03:09:19Z"
  generation: 1
  name: example-config
  namespace: default
  resourceVersion: "236531"
  uid: 81deaa89-0f6f-41b4-a17c-dba8a70ae834
spec:
  logging:
    level:
      system: info
  provider:
    kubernetes:
      envoyService:
        type: LoadBalancer
    type: Kubernetes
```

qicz: the default value is `LoadBalancer ` and after following CR applied the type same as the default. 

what's the problem?

api/config/v1alpha1/validation/envoyproxy.go
zirain: > so you mean the API-server has checked this?

yes

> But I think the validation here is necessary for other modules inside eg, like test and controller etc

EG watch it through client-go?

api/config/v1alpha1/shared_types.go
zirain: my mistake, tested with a wrong commit.

api/config/v1alpha1/validation/envoyproxy.go
arkodg: yes this check is not needed since the kube builder annotation bakes it into the CRD that is used by the API server to validate the data

arkodg: but we probably need a `nil` check since its a pointer, right ?

qicz: > but we probably need a `nil` check since its a pointer, right ?

nil will use the default, but i think double check is necessary? @arkodg

arkodg:
> > @qicz we will also need to fetch the right IP address and set it in the status here, for this implementation type to be conformant
> > https://github.com/envoyproxy/gateway/blob/384577334dea2c0ba28800643a2b94fea4eed244/internal/status/gateway.go#L32
> > 
> > and ensure conformance tests pass
> > https://github.com/envoyproxy/gateway/blob/main/docs/latest/dev/README.md#run-gateway-api-conformance-tests
> > can be done in this PR or a follow up (if so lets track it with an issue)
> 
> I will follow up with a new issue but there is no API about `ClusterIP` @arkodg image

you can probably set the `ClusterIP` value

arkodg (approved):
LGTM with a minor suggestion reg validation check, thanks for adding this enhancement !

zirain (approved):

PR #1208, opened 2023-03-23
provider: watch for resources from extensions

haq204:
Updates Kubernetes provider to watch for resources registered in the extensions config

This PR makes the assumption that registered resources are filters to be referenced in `HTTPRouteFIlter.extensionRef`. Policy attachments from extensions can be handled in a future iteration when we decide on a way to handle PolicyAttachements in general for EG

Relates to:
- https://github.com/envoyproxy/gateway/issues/20
- https://github.com/envoyproxy/gateway/pull/1074
- https://github.com/envoyproxy/gateway/pull/1163

### DIFF
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
@@ -187,7 +188,7 @@ func HasReadyListener(listeners []*ListenerContext) bool {
 }
 
 // ValidateHTTPRouteFilter validates the provided filter within HTTPRoute.
-func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter) error {
+func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter, extGKs ...schema.GroupKind) error {
 	switch {
 	case filter == nil:
 		return errors.New("filter is nil")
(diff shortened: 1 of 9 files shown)


internal/gatewayapi/helpers.go
```diff
@@ -187,7 +188,7 @@ func HasReadyListener(listeners []*ListenerContext) bool {
 }
 
 // ValidateHTTPRouteFilter validates the provided filter within HTTPRoute.
-func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter) error {
+func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter, addlGKs ...schema.GroupKind) error {
```
arkodg, line 191:
```suggestion
func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter, extGKs ...schema.GroupKind) error {
```

internal/gatewayapi/helpers.go
```diff
@@ -208,11 +209,17 @@ func ValidateHTTPRouteFilter(filter *v1beta1.HTTPRouteFilter) error {
 		case string(filter.ExtensionRef.Kind) == egv1a1.KindRateLimitFilter:
 			return nil
 		default:
+			for _, gk := range addlGKs {
+				if filter.ExtensionRef.Group == v1beta1.Group(gk.Group) &&
```
arkodg, line 213: validation is essentially being skipped, which is not ideal

internal/gatewayapi/helpers.go
```diff
@@ -232,14 +239,25 @@ func IsRateLimitHTTPFilter(filter *v1beta1.HTTPRouteFilter) bool {
 }
 
 // ValidateGRPCRouteFilter validates the provided filter within GRPCRoute.
-func ValidateGRPCRouteFilter(filter *v1alpha2.GRPCRouteFilter) error {
+func ValidateGRPCRouteFilter(filter *v1alpha2.GRPCRouteFilter, addlGKs ...schema.GroupKind) error {
```
arkodg, line 242:
```suggestion
func ValidateGRPCRouteFilter(filter *v1alpha2.GRPCRouteFilter, extGKs ...schema.GroupKind) error {
```

internal/provider/kubernetes/routes.go
```diff
@@ -155,8 +161,8 @@ func (r *gatewayAPIReconciler) processHTTPRoutes(ctx context.Context, gatewayNam
 	resourceMap *resourceMappings, resourceTree *gatewayapi.Resources) error {
 	httpRouteList := &gwapiv1b1.HTTPRouteList{}
 
-	// An HTTPRoute may reference an AuthenticationFilter or RateLimitFilter,
-	// so add them to the resource map first (if they exist).
+	// An HTTPRoute may reference an AuthenticationFilter. RateLimitFilter, or a filter managed
```
arkodg, line 164:
```suggestion
	// An HTTPRoute may reference an AuthenticationFilter, RateLimitFilter, or a filter managed
```

internal/gatewayapi/helpers.go
arkodg: yeah most validation is usually handled either pre provider (validation webhook) or post provider (related to gateway api logic and status), there are some validations like this one in the provider and we should surface these in the status as well

arkodg: nvm, all the current validations in the provider for external filters will be moved into the validation webhook soon

arkodg (approved):

AliceProxy (approved):

PR #1209, opened 2023-03-24
Add extension hooks implementation

AliceProxy:
This is a PR working towards the goal of https://github.com/envoyproxy/gateway/issues/20 and follows much of the design of https://github.com/envoyproxy/gateway/pull/1074

Configuration for resisting extensions is already supported. This PR builds on that by introducing an API that allows extension developers to write a service that makes use of gRPC hook calls to modify xDS resources.

Along with introducing the API, the xDS translation layer has been updated to call out to a registered extension. 

The remaining pieces of functionality needed to fully support the initial version of the extensions system are [Envoy Gateway watches extension resources](https://github.com/envoyproxy/gateway/pull/1208) and after that, the translation layer will need to be updated so that these resources are actually passed along and sent to extensions in the hook call(s).

The current implementation only includes Post xDS modification hooks. In the future, more hooks will be supported such as Pre xDS modification hooks, and one or more hooks in the infra manager system so that extensions can dynamically create/remove deployments along with the lifecycle of EnvoyProxy instances that are created and managed by Envoy Gateway.

Relates to:
- https://github.com/envoyproxy/gateway/issues/20
- https://github.com/envoyproxy/gateway/pull/1074
- https://github.com/envoyproxy/gateway/pull/1163
- https://github.com/envoyproxy/gateway/pull/1208


buf.gen.yaml
```diff
@@ -0,0 +1,9 @@
+version: v1
```
arkodg, line 1: can all this live inside `tools/src/buf` ?

tools/make/tools.mk
```diff
@@ -12,15 +12,25 @@ $(tools.bindir)/%: $(tools.srcdir)/%.sh
...
+tools/kind               = $(tools.bindir)/kind
+tools/setup-envtest      = $(tools.bindir)/setup-envtest
+tools/crd-ref-docs = $(tools.bindir)/crd-ref-docs
+tools/buf                = $(tools.bindir)/buf
+tools/protoc-gen-go      = $(tools.bindir)/protoc-gen-go
+tools/protoc-gen-go-grpc = $(tools.bindir)/protoc-gen-go-grpc
 $(tools.bindir)/%: $(tools.srcdir)/%/pin.go $(tools.srcdir)/%/go.mod
 	cd $(<D) && GOOS= GOARCH= go build -o $(abspath $@) $$(sed -En 's,^import "(.*)".*,\1,p' pin.go)
 
+# Let these use the main Envoy Gateway go.mod instead of having their own go.mod to ensure runtime
```
arkodg, line 27: not sure I understand the WHY

internal/xds/translator/translator.go
```diff
@@ -34,14 +38,14 @@ type GlobalRateLimitSettings struct {
 }
 
 // Translate translates the XDS IR into xDS resources
-func (t *Translator) Translate(ir *ir.Xds) (*types.ResourceVersionTable, error) {
+func (t *Translator) Translate(ir *ir.Xds, extManager extensionTypes.Manager) (*types.ResourceVersionTable, error) {
```
arkodg, line 41: you probably want `extensionTypes.Manager` to be set within the `Translator` context instead of an arg

internal/xds/translator/translator.go
```diff
@@ -53,10 +57,41 @@ func (t *Translator) Translate(ir *ir.Xds) (*types.ResourceVersionTable, error)
 		return nil, err
 	}
 
+	// If there is a loaded extension that wants to inject clusters/secrets, then call it
```
arkodg, line 60: can we move all extension code into a separate file similar to what's done for ratelimit

internal/provider/runner/runner.go
```diff
@@ -20,6 +22,7 @@ import (
 type Config struct {
 	config.Server
 	ProviderResources *message.ProviderResources
+	ExtensionManager  extension.Manager
```
arkodg, line 25: shouldnt this already be part of `config.Server` ?

internal/provider/kubernetes/kubernetes_test.go
```diff
@@ -53,7 +54,28 @@ func TestProvider(t *testing.T) {
 	svr, err := config.New()
 	require.NoError(t, err)
 	resources := new(message.ProviderResources)
-	provider, err := New(cliCfg, svr, resources)
+	ext := egcfgv1a1.Extension{
```
arkodg, line 57: can this be added aa a separate test ?

buf.gen.yaml
AliceProxy: I tried to make this change, but there seems to be a lack of functionality in `buf` to support this. While `buf generate` supports setting a different location for the `buf.yaml` and `buf.gen.yaml` using the `--template` and `--config` flags, it seems to always look for the `buf.lock` file in the root directory. `buf` also expects the `buf.yaml` files to be located alongside the `buf.lock` file so I guess those flags are for use-cases where the `buf` command is not being invoked from the root dir but that you need to point `buf` at where the root dir is. Based on the `buf` docs, does seem like this is known and intended design, but if you figure out a way to make this work then lmk.

internal/xds/translator/runner/runner_test.go
```diff
@@ -24,10 +26,12 @@ func TestRunner(t *testing.T) {
 	xds := new(message.Xds)
 	cfg, err := config.New()
 	require.NoError(t, err)
+	em := testutils.NewManager(egv1a1cfg.Extension{})
```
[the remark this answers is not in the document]
AliceProxy: I did create a separate test for the xDS `translate_test.go`, but this line is just passing in an empty extension manager with no extension. It doesn't do anything at all here.

internal/provider/kubernetes/kubernetes_test.go
AliceProxy: removed a bunch of the extension manager passing around since #1208 handles the validation of HTTPRoutes/GRPCRoutes.

AliceProxy:
In last Tuesday's meeting we discussed potentially changing the hook execution logic to run entirely after the `xds/translator/translator.go` instead of happening inside of `translator.go` after each resource is generated. I gave it some though and discussed things with my team and we felt like we'd be adding additional performance overhead for not any tangible benefit. 

An extension provider will already have to pay the performance cost of executing all the hooks to modify xDS resources and we really don't want to have to also re-iterate over the resource table after it's been built the first time. Listeners wouldn't be so bad since there are so few, but we have users running environments with the equivalent of nearly a hundred Gateways and countless routes. Going through all the resources again to pick out the routes, and figure out which ir.HTTPRoute generated that route so we can pass in the `extensionRef` resources seems pointlessly costly when we can just modify the route in-line after it is generated, and any users of Envoy Gateway that aren't using an extension wouldn't have any degraded performance from this way since the logic only happens when an extension is loaded and subscribed to modify the resource in question.

AliceProxy:
Per the meeting discussion, I'm going to try out changing the proto for sending over extension resources from `[]byte` to `proto.Any`. In a separate PR (not targeting rc.1) I'll work on creating an "example" extension and user-doc for writing and building extensions. This can be further used in the future for creating E2E tests of extensions.

buf.gen.yaml
arkodg: lets keep buf.lock in the root (and add a comment highlighting the limitation), and move others into src  ?

go.sum
```diff
@@ -671,6 +671,8 @@ google.golang.org/grpc v1.40.0/go.mod h1:ogyxbiOoUXAkP+4+xa6PZSE9DZgIHtSpzjDTB9K
 google.golang.org/grpc v1.42.0/go.mod h1:k+4IHHFw41K8+bbowsex27ge2rCb65oeWqe4jJ590SU=
 google.golang.org/grpc v1.54.0 h1:EhTqbhiYeixwWQtAEZAxmV9MGqcjEU2mFx52xCzNyag=
 google.golang.org/grpc v1.54.0/go.mod h1:PUSEXI6iWghWaB6lXM4knEgpJNu2qUcKfDtNci3EC2g=
+google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.2.0 h1:TLkBREm4nIsEcexnCjgQd5GQWaHcqMzwQV0TX9pq8S0=
```
arkodg, line 674: can we rm this ?

internal/cmd/server.go
```diff
@@ -98,6 +99,12 @@ func setupRunners(cfg *config.Server) error {
 	// https://github.com/envoyproxy/gateway/issues/43
 	ctx := ctrl.SetupSignalHandler()
 
+	// Setup the Extension Manager
```
arkodg, line 102: would be good to add a check here
var extMgr *pointer
```
if cfg.Extension. != nil {
   extMgr = 
```

and you can use this value nil check during cleanup

internal/extension/registry/extension_manager.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}]}`
+
+var _ extTypes.Manager = (*Manager)(nil)
+
+type Manager struct {
+	k8sClient     k8scli.Client
+	namespace     string
+	extension     v1alpha1.Extension
+	hookConnCache map[extTypes.ExtensionXDSHookType]*grpc.ClientConn
```
arkodg, line 51: shouldnt this a xdsHookConnCache ?
how would pre & post hook state be stored ?
would a separate cache be created for each hook type ?

proto/extension/service.proto
```diff
@@ -0,0 +1,93 @@
...
+
+
+// PostHTTPListenerModifyResponse is the expected response from an extension and contains a modified version of the Listener that was sent
+// An extension may choose to return a nil Listener to cause Envoy Gateway to discard it
+message PostHTTPListenerModifyResponse {
+    envoy.config.listener.v3.Listener listener = 1;
+}
+
+
+// PostTranslationInsertRequest currently sends only context to an extension
```
arkodg, line 83: can we only have modify ?

proto/extension/context.proto
```diff
@@ -0,0 +1,55 @@
...
+message PostXDSInsertExtensionContext {
+
+}
+
+
+// ExtensionResource stores the data for a K8s API object referenced in an HTTPRouteFilter
+// extensionRef. It is constructed from an unstructured.Unstructured marshalled to JSON. An extension
+// can marshal the bytes from this resource back into an unstructured.Unstructured and then 
+// perform type checking to obtain the resource.
+message ExtensionResource {
```
arkodg, line 53: should this be called `GatewayAPIResource` instead ?

internal/xds/translator/translator.go
```diff
@@ -53,10 +58,28 @@ func (t *Translator) Translate(ir *ir.Xds) (*types.ResourceVersionTable, error)
...
+
 	return tCtx, nil
 }
 
 func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersionTable, httpListeners []*ir.HTTPListener) error {
+
+	// See if any extensions are loaded that want to modify xDS VirtualHosts or Routes
+	// If so, then store the clients for them up-front to prevent needing to grab the client multiple times
+	// in the loops below
+	var extensionRouteClient extensionTypes.XDSHookClient
```
arkodg, line 75: can this logic be moved into the extension file ?

internal/xds/translator/translator.go
```diff
@@ -113,10 +136,20 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 			protocol = HTTP2
 		}
 
+		// Check if an extension is loaded that wants to modify xDS Routes after they have been generated
 		for _, httpRoute := range httpListener.Routes {
 			// 1:1 between IR HTTPRoute and xDS config.route.v3.Route
 			xdsRoute := buildXdsRoute(httpRoute, xdsListener)
-			vHost.Routes = append(vHost.Routes, xdsRoute)
+
```
arkodg, line 143: can we achieve this without affecting the core translator logic ?

internal/xds/translator/runner/runner.go
```diff
@@ -53,7 +55,7 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				r.Xds.Delete(key)
 			} else {
 				// Translate to xds resources
-				t := &translator.Translator{}
+				t := &translator.Translator{ExtensionManager: &r.ExtensionManager}
```
arkodg, line 58: can we follow the code style similar to L60, where we set the value if its not nil ?

internal/ir/xds.go
```diff
@@ -223,6 +225,20 @@ type HTTPRoute struct {
 	RateLimit *RateLimit
 	// RequestAuthentication defines the schema for authenticating HTTP requests.
 	RequestAuthentication *RequestAuthentication
+	// ExtensionRefs holds unstructured resources that were introduced by an extension and used on the HTTPRoute as extensionRef filters
+	// TODO: Envoy Gateway is not currently watching resources introduced by an extension.
+	// When that functionality is introduced then this field will be used to pass the resources through the xDS layer
+	// and on to the extension
+	ExtensionRefs []*UnstructuredRef
```
arkodg, line 232: so were only passing the filter ? not the `HTTPRoute` ?

proto/extension/service.proto
arkodg: was recommending to use `modify` so the end returned result can be an addition or deletion of a list of resources

internal/xds/translator/translator.go
AliceProxy: As much of it as possible has been pulled out of the core translator logic without entirely re-processing things "post-translator.go"

tools/make/tools.mk
AliceProxy: moved things to use their own go.mod

AliceProxy:
> Per the meeting discussion, I'm going to try out changing the proto for sending over extension resources from []byte to proto.Any. In a separate PR (not targeting rc.1) I'll work on creating an "example" extension and user-doc for writing and building extensions. This can be further used in the future for creating E2E tests of extensions.

Something I didn't remember when we discussed making this change was the fact that you can't create a `proto.Any` with anything that isn't a `proto.message` or an `interface` which `unstructured.Unstructured` is not, so going to stick with the byte array unless a workable alternative is suggested.

internal/extension/registry/extension_manager.go
AliceProxy: Yeah I think you're right and we don't need a separate cache for each hook since there is only a single extension and the connection info is the same regardless of what hook is being called.

internal/cmd/server.go
AliceProxy: The `extensionregistry.NewManager()` `func` now performs that check and will create an empty manager that does nothing in the case that `cfg.EnvoyGateway` or `cfg.EnvoyGateway.Extension` is `nil`. The only time the func returns `nil, error` now is when `*rest.Config for talking to a Kubernetes apiserver` cannot be created. That is an error where we quit early rather than continue and can never happen when not trying to use an extension.

internal/xds/translator/extension.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/envoyproxy/gateway/internal/ir"
+	"github.com/envoyproxy/gateway/internal/xds/types"
+
+	"github.com/envoyproxy/gateway/api/config/v1alpha1"
+	extensionTypes "github.com/envoyproxy/gateway/internal/extension/types"
+)
+
+func processExtensionPostRouteHook(route *routev3.Route, vHost *routev3.VirtualHost, irRoute *ir.HTTPRoute, em *extensionTypes.Manager) (*routev3.Route, error) {
+	if em != nil && len(irRoute.ExtensionRefs) > 0 {
```
arkodg, line 29: can all the functions return early
```
if em == nil {
 return nil
}
```

internal/xds/translator/translator.go
```diff
@@ -113,9 +124,20 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
 		}
 
+		// Check if an extension is loaded that wants to modify xDS Routes after they have been generated
 		for _, httpRoute := range httpListener.Routes {
 			// 1:1 between IR HTTPRoute and xDS config.route.v3.Route
 			xdsRoute := buildXdsRoute(httpRoute, xdsListener)
+
+			// Check if an extension want to modify the route we just generated
+			// If no extension exists (or it doesn't subscribe to this hook) then this is a quick no-op.
+			if modifiedRoute, err := processExtensionPostRouteHook(xdsRoute, vHost, httpRoute, t.ExtensionManager); err == nil {
```
arkodg, line 134: `xdsRoute` is already a pointer being passed, and is a scalar quantity, why do we need to return it as a return value ?

internal/xds/translator/translator.go
```diff
@@ -146,6 +168,15 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 			}
 		}
 
+		// Check if an extension want to modify the Virtual Host we just generated
+		// If no extension exists (or it doesn't subscribe to this hook) then this is a quick no-op.
+		if modifiedVHost, err := processExtensionPostVHostHook(vHost, t.ExtensionManager); err == nil {
```
arkodg, line 173: same comment as above

internal/xds/translator/translator.go
AliceProxy: I was originally going to have `processExtensionPostRouteHook` overwrite the passed in route and only return an error like so:

```
	if modifiedRoute != nil {
		route = modifiedRoute
	}
```

But lint fails since there is an assignment to `route` and it is never used. I figured that returning the route we want to use would be simpler than going through and doing `route.Name = modifiedRoute.Name` etc. for all the fields, but I can make that change if you think it's preferable.

arkodg: yeah prefer if `processExtensionPostRouteHook` passed the route as a pointer, and updated it/or didnt, and there was no logic around `modifiedRoute` route here

buf.gen.yaml
arkodg: yeah lets stick to 2. for now than, thanks for investing cycles to find a cleaner solution.
would be good to follow up with this in `buf` upstream

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):
Looks good overall ! Thanks : )

Issue #1196, opened 2023-03-20
set envoy proxy & ratelimit container resource requests


qicz:
set the envoy proxy container the minimum amount of computing resources required.

- [x] envoy proxy container resources settings
- [x] ratelimit container resources settings
- [x] resources settings api docs
- [x] envoy proxy setting user face docs by #1203 
- [x] ratelimit  setting user face docs

arkodg:
@qicz is this issue related to changing the default resource settings https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/proxy_deployment.go
or adding an API to change the defaults https://github.com/envoyproxy/gateway/blob/c459f655f971f3c3ef986f1fe2a16ee65a906f2c/api/config/v1alpha1/envoyproxy_types.go#L82 ?
(both are valid and needed :) )

qicz:
> 

@arkodg yes, I would like to do this, pls assign to me

qicz:
> @qicz is this issue related to changing the default resource settings https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/proxy_deployment.go or adding an API to change the defaults
> 
> https://github.com/envoyproxy/gateway/blob/c459f655f971f3c3ef986f1fe2a16ee65a906f2c/api/config/v1alpha1/envoyproxy_types.go#L82
> 
> ?
> (both are valid and needed :) )

@arkodg  yes, I would like to do this, pls assign to me

qicz:
support by #1197 #1218

PR #1218, opened 2023-03-26
feat: support envoy ratelimit pod & container resources & securitycontext settings.

qicz:
part support https://github.com/envoyproxy/gateway/issues/1196

- support ratelimit pod annotation and securitycontext setting
- support ratelimit container resources and securitycontext setting

```yaml
kind: EnvoyGateway
gateway:
provider:
  kubernetes:
    rateLimitDeployment:
      replicas: 1
      container:
        resources:
          requests:
            cpu: 100m
            memory: 512Mi
        securityContext:
          runAsUser: 2000
      pod:
        annotations:
          key1: val1
          key2: val2
        securityContext:
          runAsUser: 1000
          runAsGroup: 3000
          fsGroup: 2000
rateLimit:
  backend:
    type: Redis
    redis:
      url: localhost:6379
```


api/config/v1alpha1/envoygateway_types.go
```diff
@@ -107,6 +108,11 @@ type RateLimit struct {
 	// database backend used by the rate limit service to store
 	// state associated with global ratelimiting.
 	Backend RateLimitDatabaseBackend `json:"backend"`
+
+	// DeploymentResources required by the ratelimit deployment container.
```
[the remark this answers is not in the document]
qicz: I did a version like this 

```golang
// RateLimit defines the configuration associated with the Rate Limit Service
// used for Global Rate Limiting.
type RateLimit struct {
	// Backend holds the configuration associated with the
	// database backend used by the rate limit service to store
	// state associated with global ratelimiting.
	Backend RateLimitDatabaseBackend `json:"backend"`
	// Deployment defines the desired state of the Envoy ratelimit deployment resource.
	// If unspecified, default settings for the manged Envoy ratelimit deployment resource
	// are applied.
	//
	// +optional
	Deployment *KubernetesDeploymentSpec `json:"deployment,omitempty"`
}
```

but I finally changed my mind, I think the replicas setting for the ratelimit is unnecessary, just set the `REDIS_TYPE` to `cluster`, redis to handle more traffic.

I'll wait for the conclusion of the community meeting and refactor this.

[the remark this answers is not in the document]
arkodg: chatted with @haq204 @AliceProxy in the community meeting today and we prefer
```
 provider:
    kubernetes:
      rateLimitDeployment:
        replicas: 2
```
to reduce the number of nested structures

but prefer adding this if actually needed by real users today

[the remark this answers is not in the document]
qicz: I will go on after #1232 merged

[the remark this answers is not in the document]
qicz: > @qicz are you using ratelimiting ? if so are you hitting scale issues ?

It is not used now, but it will be used in the future. and hope it can be configured like envoy deployment, auth can also be configured

arkodg (changes requested):
lets not change any existing names yet
also I think we should add modifications to ratelimit service when users have shared limitations with default, would prefer if we delayed this PR until then

qicz:
> lets not change any existing names yet also I think we should add modifications to ratelimit service when users have shared limitations with default, would prefer if we delayed this PR until then

@arkodg 

In fact, changing the entity name does not affect crds. They were modified because many of the definitions of envoyproxy and gateway are too similar to be easily ambiguous. Of course, if such ambiguity is not a concern, I will revert to the original name, although I suggest modifying them.

qicz:
@arkodg another, maybe changing ratelimit service is unnecessary. IMO the ratelimit service should be `ClusterIP` because just used for Cluster-Internel, and annotations use the default settings.

arkodg:
> @arkodg another, maybe changing ratelimit service is unnecessary. IMO the ratelimit service should be `ClusterIP` because just used for Cluster-Internel, and annotations use the default settings.

agree @qicz , and suggest opening a separate PR for just renaming the provider structures, its easier to mantain and track PRs/commits that solve 1 thing instead of many

qicz:
> > @arkodg another, maybe changing ratelimit service is unnecessary. IMO the ratelimit service should be `ClusterIP` because just used for Cluster-Internel, and annotations use the default settings.
> 
> agree @qicz , and suggest opening a separate PR for just renaming the provider structures, its easier to mantain and track PRs/commits that solve 1 thing instead of many

 ok, I always revert the change and will open another pr do this. 

pls review again @arkodg

qicz:
> > > > @arkodg another, maybe changing ratelimit service is unnecessary. IMO the ratelimit service should be `ClusterIP` because just used for Cluster-Internel, and annotations use the default settings.
> > > 
> > > 
> > > agree @qicz , and suggest opening a separate PR for just renaming the provider structures, its easier to mantain and track PRs/commits that solve 1 thing instead of many
> > 
> > 
> > ok, I already revert the change and will open another pr do this.
> > pls review again @arkodg
> 
> can you please update the PR title and description to reflect the changes and why its needed so it helps other reviewers understand the why quickly, before diving into the changes

update the pr description, pls review again @arkodg

arkodg:
@qicz looks like there might have been misunderstanding.
I assumed you agreed as well that adding resource knob or securityContext knobs to rate limit is premature and not needed right now, which may be added later when there is strong user interest.

Instead, you were going to raise a separate PR or modify this one to only add the diff involving updating the `Provider` struct names for adding more clarity to the project

arkodg:
@xuaning67 will this PR support your use case ?

xuaning67:
> @xuaning67 will this PR support your use case ?

Thank you for your reply. I think it is helpful for me to support my use case.

arkodg:
@zirain are you okay if we let this feature in and then rebase and refactor https://github.com/envoyproxy/gateway/pull/1259 ?

zirain:
I'm OK.

qicz:
> > @xuaning67 will this PR support your use case ?
> 
> @arkodg Thank you for your reply. I think it is helpful for me to support my use case.

sorry for the late reply, I'm glad this pr can help you @xuaning67 

@arkodg I have already rebased the main branch and refactored.

internal/ir/ratelimit_infra.go
```diff
@@ -5,13 +5,17 @@
...
 // +k8s:deepcopy-gen=true
 type RateLimitInfra struct {
-	// Rate limit service configuration
-	Configs []*RateLimitServiceConfig
-	// Backend holds configuration associated with the backend database.
-	Backend *RateLimitDBBackend
+	// ServiceConfigs for Rate limit service configuration.
+	ServiceConfigs []*RateLimitServiceConfig
+	// Config ratelimit configuration.
+	Config *v1alpha1.EnvoyGateway
```
arkodg, line 18: do we need to pass entire EnvoyGateway config here, or just  `rateLimitDeployment` ?

arkodg: since EnvoyGateway is a startup config, it exists in every runner, so maybe https://github.com/envoyproxy/gateway/blob/6fc9f8dfe8883df61343b133da694257eb313f69/internal/infrastructure/kubernetes/infra.go#L19 can be enhanced to store envoy gateway config

qicz: > do we need to pass entire EnvoyGateway config here, or just `rateLimitDeployment` ?

using `EnvoyGateway` is convinced for other extensions, like supporting different `Backend`.

qicz: > since EnvoyGateway is a startup config, it exists in every runner, so maybe
> 
> https://github.com/envoyproxy/gateway/blob/6fc9f8dfe8883df61343b133da694257eb313f69/internal/infrastructure/kubernetes/infra.go#L19
> 
> can be enhanced to store envoy gateway config

store envoy gateway config to `internal/infrastructure/kubernetes/infra.go:27`, and assign the config when the subscription changes.

internal/infrastructure/runner/runner.go
```diff
@@ -79,6 +79,9 @@ func (r *Runner) subscribeToRateLimitInfraIR(ctx context.Context) {
 		func(update message.Update[string, *ir.RateLimitInfra]) {
 			val := update.Value
 
+			// loading EnvoyGateway Configurations.
+			val.Config = r.EnvoyGateway
```
arkodg, line 83: why is this needed, `EnvoyGateway` is a static config, set during startup, and it doesnt change, so it shouldnt be part the pub-sub messages

arkodg (approved):
LGTM, thanks for addressing the comment !

zirain (approved):

Issue #1212, opened 2023-03-25
Populate Gateway Status address for ClusterIP type Envoy Service


arkodg:
*Description*:
Relates to https://github.com/envoyproxy/gateway/pull/1205#issuecomment-1483594260

qicz:
pls assgin to me

PR #1219, opened 2023-03-26
Populate Gateway Status address for ClusterIP type Envoy Service.

qicz:
support #1212


internal/status/gateway.go
```diff
@@ -40,24 +40,35 @@ func UpdateGatewayStatusProgrammedCondition(gw *gwapiv1b1.Gateway, svc *corev1.S
 			}
 		}
 
-		var gwAddrs []gwapiv1b1.GatewayAddress
+		if svc.Spec.ClusterIP != "" {
+			addresses = append(addresses, svc.Spec.ClusterIP)
+		}
+		if len(svc.Spec.ClusterIPs) != 0 {
```
zirain, line 46: redundant check?

qicz: One is clusterip, another is clusterips.

internal/status/gateway.go
```diff
@@ -40,24 +40,35 @@ func UpdateGatewayStatusProgrammedCondition(gw *gwapiv1b1.Gateway, svc *corev1.S
 			}
 		}
 
-		var gwAddrs []gwapiv1b1.GatewayAddress
+		if svc.Spec.ClusterIP != "" {
```
arkodg, line 43: shouldnt a check be added to make sure we are only doing this for svc type `ClusterIP` ? else we'll end up adding the `ClusterIP` for `LoadBalancer` type as well which we dont want

internal/status/gateway.go
qicz: done

internal/status/gateway.go
qicz: done

internal/status/gateway.go
```diff
@@ -40,24 +40,32 @@ func UpdateGatewayStatusProgrammedCondition(gw *gwapiv1b1.Gateway, svc *corev1.S
 			}
 		}
 
-		var gwAddrs []gwapiv1b1.GatewayAddress
+		if svc.Spec.Type == corev1.ServiceTypeClusterIP && len(svc.Spec.ClusterIPs) != 0 {
```
arkodg, line 43: nit: len check not needed :)

arkodg: can you add a similar check on L29 (svc.Spec.Type == Loadbalancer) ?

qicz: did it

arkodg (approved):
LGTM, with minor nit !

Xunzhuo (approved):

Issue #1071, opened 2023-02-22
Support GRPCRoute regex type match


sbko:
*Description*:

support for https://gateway-api.sigs.k8s.io/geps/gep-1016/#type-regularexpression is blocked by https://github.com/kubernetes-sigs/gateway-api/issues/1746

add regex support [here](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/route.go#L391) after upstream issue is resolved.

arkodg:
should be ready to be worked on since https://github.com/kubernetes-sigs/gateway-api/issues/1746 has been fixed and brought into this project as part of the v0.6.2 bump

shawnh2:
hi, I'm willing to take this, assign me please

Xunzhuo:
Close in favor of: https://github.com/envoyproxy/gateway/pull/1240

PR #1240, opened 2023-04-01
Add support for GRPCRoute regex type match

shawnh2:
refer to #1071

### DIFF
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -413,22 +413,11 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
 
 		if match.Method != nil {
 			// GRPC's path is in the form of "/<service>/<method>"
-			// TODO: support regex match type after https://github.com/kubernetes-sigs/gateway-api/issues/1746 is resolved
-			switch {
-			case match.Method.Service != nil && match.Method.Method != nil:
-				irRoute.PathMatch = &ir.StringMatch{
-					Exact: StringPtr(fmt.Sprintf("/%s/%s", *match.Method.Service, *match.Method.Method)),
-				}
-			case match.Method.Method != nil:
-				// Use a header match since the PathMatch doesn't support Suffix matching
-				irRoute.HeaderMatches = append(irRoute.HeaderMatches, &ir.StringMatch{
-					Name:   ":path",
-					Suffix: StringPtr(fmt.Sprintf("/%s", *match.Method.Method)),
-				})
-			case match.Method.Service != nil:
-				irRoute.PathMatch = &ir.StringMatch{
-					Prefix: StringPtr(fmt.Sprintf("/%s", *match.Method.Service)),
-				}
+			switch GRPCMethodMatchTypeDerefOr(match.Method.Type, v1alpha2.GRPCMethodMatchExact) {
+			case v1alpha2.GRPCMethodMatchExact:
+				t.processGRPCRouteMethodExact(match.Method, irRoute)
+			case v1alpha2.GRPCMethodMatchRegularExpression:
+				t.processGRPCRouteMethodRegularExpression(match.Method, irRoute)
 			}
 		}
 
@@ -436,6 +425,46 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
 		applyHTTPFiltersContextToIRRoute(httpFiltersContext, irRoute)
 	}
 	return ruleRoutes
+}
+
+func (t *Translator) processGRPCRouteMethodExact(method *v1alpha2.GRPCMethodMatch, irRoute *ir.HTTPRoute) {
+	switch {
+	case method.Service != nil && method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			Exact: StringPtr(fmt.Sprintf("/%s/%s", *method.Service, *method.Method)),
+		}
+	case method.Method != nil:
+		// Use a header match since the PathMatch doesn't support Suffix matching
+		irRoute.HeaderMatches = append(irRoute.HeaderMatches, &ir.StringMatch{
+			Name:   ":path",
+			Suffix: StringPtr(fmt.Sprintf("/%s", *method.Method)),
+		})
+	case method.Service != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			Prefix: StringPtr(fmt.Sprintf("/%s", *method.Service)),
+		}
+	}
+}
+
+func (t *Translator) processGRPCRouteMethodRegularExpression(method *v1alpha2.GRPCMethodMatch, irRoute *ir.HTTPRoute) {
+	switch {
+	case method.Service != nil && method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/%s", *method.Service, *method.Method)),
+		}
+	case method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/.+/%s", *method.Method)),
+		}
+	case method.Service != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/.+", *method.Service)),
+		}
+	default:
+		irRoute.PathMatch = &ir.StringMatch{
+			Prefix: StringPtr("/"),
+		}
+	}
 }
 
 func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, routeRoutes []*ir.HTTPRoute, parentRef *RouteParentContext, xdsIR XdsIRMap) bool {


internal/gatewayapi/route.go
```diff
@@ -438,6 +427,46 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+		}
+	case method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/.+/%s", *method.Method)),
+		}
+	case method.Service != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/.+", *method.Service)),
+		}
+	default:
```
arkodg, line 463: shouldnt default be an invalid case?

internal/gatewayapi/route.go
```diff
@@ -438,6 +427,46 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/%s", *method.Service, *method.Method)),
+		}
+	case method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/.+/%s", *method.Method)),
+		}
+	case method.Service != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/.+", *method.Service)),
```
arkodg, line 461: why is `/.+` needed here ?

internal/gatewayapi/route.go
```diff
@@ -438,6 +427,46 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+
+func (t *Translator) processGRPCRouteMethodRegularExpression(method *v1alpha2.GRPCMethodMatch, irRoute *ir.HTTPRoute) {
+	switch {
+	case method.Service != nil && method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/%s/%s", *method.Service, *method.Method)),
+		}
+	case method.Method != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			SafeRegex: StringPtr(fmt.Sprintf("/.+/%s", *method.Method)),
```
arkodg, line 457: thinking out loud, it is okay to use `.+` here (accepts one or more of any char) because the field has already been validated in the validation webhook, before it reaches here ?

internal/gatewayapi/route.go
shawnh2: sure this is an invalid case, i'll rm it

internal/gatewayapi/route.go
shawnh2: actually this field never be validated by validation webhook when `MatchType=RegularExpression`, see https://github.com/kubernetes-sigs/gateway-api/blob/8dde50805b8010f6ec52ae52e7d47d799cdcaabf/apis/v1alpha2/validation/grpcroute.go#L80

so we need `/.+/%s` to make sure that the `service` has been matched at right position, for example, not `//method` nor `/method`, but `/x/method`.

and so does `/%s/.+`, making sure the `method` has been matched at right position.

arkodg: thanks for checking upstream, in that case should this be `[A-Za-z0-9]+` instead of `.+` so we are validating char type

arkodg: sure that should work for both, thanks !
can you please test this out as well, here is a good starting point https://gateway.envoyproxy.io/latest/user/grpc-routing.html

shawnh2:
hi, a little problem here.

when I deploy Envoy Gateway on my k8s cluster following the guide here: https://gateway.envoyproxy.io/latest/user/grpc-routing.html, I keep failing to get ip address for the `gateway`.

then I found in `make create-cluster`, a MetalLB has been deployed, 

https://github.com/envoyproxy/gateway/blob/bd74367f131e5ded0609851cda9392e1e20336d6/tools/hack/create-cluster.sh#L17-L51

so I'm curious is MetalLB has to be deployed? or should we point it out in the doc.

Xunzhuo:
EG deploys with LB mode in default, you can set it with clusterIP type if you want to use it as the InCluster Gateway.

shawnh2:
> EG deploys with LB Service mode in default, you can set it with clusterIP type if you want to use it as the InCluster Gateway.
> 
> Refer to #1202

thanks, it helps a lot

internal/gatewayapi/route.go
shawnh2: hi, I have run some test on `/[validServiceName]/%s` and `/%s/[validMethodName]` case, and it works well.

here is a snapshot about `/[validServiceName]/%s` case:

Snipaste_2023-04-08_10-43-58

internal/gatewayapi/testdata/grpcroute-with-service-match.out.yaml
```diff
@@ -79,16 +79,16 @@ xdsIR:
       - "*"
       isHTTP2: true
       routes:
-      - name: default-grpcroute-1-rule-0-match-0-*
+      - name: default-grpcroute-1-rule-0-match-1-*
```
arkodg, line 82: why is the order swapped here ?

internal/gatewayapi/route.go
arkodg: can you try by using a regex value for service such as `*.Echo` and leaving `method` to be `nil` ?

shawnh2: cannot do this now, present gateway-api v0.6.2 will determine this as an error: https://github.com/kubernetes-sigs/gateway-api/blob/7042cd866bc69b160ab6d04b51676e14b487505d/apis/v1alpha2/grpcroute_types.go#L329

the changes in https://github.com/kubernetes-sigs/gateway-api/issues/1746 will be introduced in gateway-api v0.7.0

shawnh2: can be tested until this https://github.com/envoyproxy/gateway/issues/1267 solved

internal/gatewayapi/testdata/grpcroute-with-service-match.out.yaml
shawnh2: becasue in here, the routes has been `reversed` before sorting
https://github.com/envoyproxy/gateway/blob/78b62e403b1941b00ecd341e55b0f7e13594e325/internal/gatewayapi/sort.go#L50-L57
and the sorting rule is first matching path, then matching header:
https://github.com/envoyproxy/gateway/blob/78b62e403b1941b00ecd341e55b0f7e13594e325/internal/gatewayapi/sort.go#L18-L30

arkodg (dismissed):
LGTM, thanks !
would prefer to hold this until GW API 0.7.0 is brought in and is tested with it

internal/gatewayapi/route.go
```diff
@@ -444,6 +437,42 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
 	return ruleRoutes
 }
 
+func (t *Translator) processGRPCRouteMethodExact(method *v1alpha2.GRPCMethodMatch, irRoute *ir.HTTPRoute) {
+	switch {
```
kflynn, line 441: I know you're just moving code here, but not gonna lie: this feels like a more natural candidate for `if` than `switch` -- it took me a minute to sort out what was up since there's no default case in the switch... 🙂

internal/gatewayapi/route.go
```diff
@@ -444,6 +437,42 @@ func (t *Translator) processGRPCRouteRule(grpcRoute *GRPCRouteContext, ruleIdx i
...
+		})
+	case method.Service != nil:
+		irRoute.PathMatch = &ir.StringMatch{
+			Prefix: StringPtr(fmt.Sprintf("/%s", *method.Service)),
+		}
+	}
+}
+
+func (t *Translator) processGRPCRouteMethodRegularExpression(method *v1alpha2.GRPCMethodMatch, irRoute *ir.HTTPRoute) {
+	switch {
```
kflynn, line 460: Same here.

kflynn (approved):
Two minor nits -- looks like it oughta work. 🙂

internal/gatewayapi/route.go
shawnh2: after thinking about it, `if` could never be better, thanks.

arkodg:
hey @shawnh2 now that we have upgraded to gateway api v0.7.0-rc1, can you rebase the PR and confirm that these changes work end to end for regex match ?

internal/gatewayapi/route.go
shawnh2: fixed

shawnh2:
hi, @arkodg , the regex type match has been tested out, and it works well.

and I also added documentation about it. PTAL

<details>
  <summary>test case here</summary>
  <pre><code>/[validServiceName]/Ping, /yages.Echo/[validMethodName], /[validServiceName]/Pin.+, /.*.Echo/Ping, /.*.Echo/Pin.+ </code></pre>
</details>

internal/gatewayapi/route.go
shawnh2: the lint suggests `switch` stmt, so I kept it

internal/gatewayapi/route.go
```diff
@@ -18,8 +18,8 @@ import (
 
 var (
 	_                RoutesTranslator = (*Translator)(nil)
-	validServiceName                  = `^(?i)\.?[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*$`
-	validMethodName                   = `^[A-Za-z_][A-Za-z_0-9]*$`
+	validServiceName                  = `(?i)\.?[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*`
```
arkodg, line 21: curious why the `^` and `$` was removed

shawnh2: cause if we do not remove `^` and `$`, the final regex pattern will be like `/^...$/method`, and it does not match any request.

arkodg: Thanks for the clarification

arkodg (approved):

internal/gatewayapi/route.go
kflynn: Silly linter. 😂

PR #1241, opened 2023-04-02
support for multiple certificates http listener

### LINKED ISSUE #393: Support for multiple TLS certificates for a single host
*Description*:
LB vendors provide the ability to configure multiple certificates for a single fqdn.
The usecase is to have multiple certificates with different signing algorithms (RSA and/or ECC).
The server, based on the clients' SSL capabilities decides to send the relevant certificate,
preferably ECC signed cert (if configured).
So customers might want to configure two certificates for a single host - one for each algorithm RSA / ECC.

[optional *Relevant Links*:]
> https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_certificate
> https://www.ibm.com/docs/en/i/7.2?topic=ssl-multiple-certificate-selection
> https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#certificate-selection

Came across this - so probably EG needs to not have this limitation, (and additionally some other checks and configs?)
https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/translator.go#L321

chauhanshubham:
Fixes #393 

Provide the ability to configure multiple certificates for a single fqdn.
The usecase is to have multiple certificates with different signing algorithms (RSA and/or ECC).
The server, based on the clients' SSL capabilities decides to send the relevant certificate,
preferably ECC signed cert (if configured).

More details in the issue #393 
and in the doc created in the PR


chauhanshubham:
I'll be adding a doc for this here, as I did not find any straightforward doc around generating certs and testing this feature. The code is ready

internal/gatewayapi/helpers.go
```diff
@@ -409,13 +409,18 @@ func routeName(route RouteContext, ruleIdx, matchIdx int) string {
...
 		return nil
 	}
 
-	return &ir.TLSListenerConfig{
-		ServerCertificate: tlsSecret.Data[v1.TLSCertKey],
-		PrivateKey:        tlsSecret.Data[v1.TLSPrivateKeyKey],
+	tlsListenerConfigs := make([]*ir.TLSListenerConfig, len(tlsSecrets))
+	for i, tlsSecret := range tlsSecrets {
+		tlsListenerConfigs[i] = &ir.TLSListenerConfig{
+			SecretName:        tlsSecret.Namespace + "-" + tlsSecret.Name,
```
arkodg, line 420: nit: good to add a helper function that can be reused later for the name
https://github.com/envoyproxy/gateway/blob/40830e86066e5729ea5719d1561b2e8cafe2dd39/internal/gatewayapi/helpers.go#L392

internal/gatewayapi/sort.go
```diff
@@ -52,6 +60,7 @@ func sortXdsIRMap(xdsIR XdsIRMap) {
 		for _, http := range irItem.HTTP {
 			// descending order
 			sort.Sort(sort.Reverse(XdsIRRoutes(http.Routes)))
+			sort.Sort(sort.Reverse(XdsIRSecrets(http.TLS)))
```
arkodg, line 63: why is the sort needed ?

arkodg: isn't there ordering since the `certRefs` is already a list

internal/gatewayapi/validate.go
```diff
@@ -240,111 +241,126 @@ func (t *Translator) validateTLSConfiguration(listener *ListenerContext, resourc
...
-				"Listener's TLS certificate ref group must be unspecified/empty.",
-			)
-			break
-		}
+			if certificateRef.Kind != nil && string(*certificateRef.Kind) != KindSecret {
+				listener.SetCondition(
+					v1beta1.ListenerConditionResolvedRefs,
+					metav1.ConditionFalse,
+					v1beta1.ListenerReasonInvalidCertificateRef,
+					fmt.Sprintf("Listener's TLS certificate ref kind must be %s.", KindSecret),
```
arkodg, line 271: maybe also add an index into this string so the user knows which certRef is invalid

internal/ir/xds.go
```diff
@@ -165,6 +167,8 @@ func (h HTTPListener) Validate() error {
 // TLSListenerConfig holds the configuration for downstream TLS context.
 // +k8s:deepcopy-gen=true
 type TLSListenerConfig struct {
+	// SecretName of the Secret object.
+	SecretName string
```
arkodg, line 171: nit: just `Name` maybe ?

internal/gatewayapi/validate.go
```diff
@@ -240,111 +241,126 @@ func (t *Translator) validateTLSConfiguration(listener *ListenerContext, resourc
...
 
-		if len(secret.Data[v1.TLSCertKey]) == 0 || len(secret.Data[v1.TLSPrivateKeyKey]) == 0 {
-			listener.SetCondition(
-				v1beta1.ListenerConditionResolvedRefs,
-				metav1.ConditionFalse,
-				v1beta1.ListenerReasonInvalidCertificateRef,
-				fmt.Sprintf("Secret %s/%s must contain %s and %s.", listener.gateway.Namespace, certificateRef.Name, v1.TLSCertKey, v1.TLSPrivateKeyKey),
-			)
-			break
+			// Check whether the public key algorithm in the referenced secrets are unique.
```
arkodg, line 348: do we need to enforce this ?
what about the case when `algo` is same but the `SNI` is different

internal/gatewayapi/validate.go
chauhanshubham: This is before we're checking the namespace name. Would an index be alright here?

internal/gatewayapi/validate.go
chauhanshubham: I haven't covered multiple certs and SNI based matching in this PR. Maybe we can keep that as separate? 
Earlier we had a hard enforcement of 1 cert per listener, which this PR relaxes, and allows for 2 now - with varying algos.

internal/gatewayapi/sort.go
chauhanshubham: my bad yea, I think earlier I saw some UT fail, but that was not because of the missing sort.

internal/gatewayapi/validate.go
chauhanshubham: IIUC right now, people can configure different hosts in different listeners, with specific tls termination configs? and then with this PR they'll be able to configure dual certs for those hosts

internal/gatewayapi/tls.go
```diff
@@ -14,26 +14,29 @@ import (
 )
 
 // validateTLSSecretData ensures the cert and key provided in a secret
-// is not malformed and can be properly parsed.
-func validateTLSSecretData(secret *corev1.Secret) error {
-
+// is not malformed and can be properly parsed. It also returns the public key
+// encryption algorithm used to encrypt secret information.
+func validateTLSSecretData(secret *corev1.Secret) (string, error) {
+	var publicKeyAlgorithm string
```
arkodg, line 20: this API feels a little overloaded, but guessing `publicKeyAlgorithm` won't be needed in the future
can you raise a GH issue to support multiple certs for SNI based selection

internal/gatewayapi/validate.go
arkodg: yes if there are two invalid `Kind`s, it will help decipher which one the status is referring to

internal/gatewayapi/validate.go
arkodg: cool, as suggested can you raise a GH issue to relax this for multi cert SNI selection ?

arkodg: seems like envoy should support it based on https://github.com/envoyproxy/envoy/issues/24824

arkodg (approved):
some non blocking comments, that can can hopefully be addressed in a follow up PR

Xunzhuo (approved):
LGTM. Thanks. I will merge it after CI passed.

chauhanshubham:
thanks @arkodg @Xunzhuo i'll track multiple cert support based on SNI selection separately 👍 
I'm guessing it'll be a comparatively smaller PR with more focus on validation / status updates / tests.

PR #1246, opened 2023-04-03
ci: generate should include docs-api

zirain:
auto update references


.github/markdown_lint_config.json
```diff
@@ -48,5 +48,6 @@
   "MD047": false,
   "MD048": false,
   "MD049": false,
-  "MD050": false
+  "MD050": false,
+  "MD051": false
```
zirain, line 52: temporarily hack, xref: #1246

tools/src/crd-ref-docs/go.mod
```diff
@@ -2,10 +2,11 @@ module local
 
 go 1.20
 
+replace github.com/elastic/crd-ref-docs => github.com/zirain/crd-ref-docs v0.0.0-20230403084159-e8e051398dc2
```
zirain, line 5: xref: https://github.com/elastic/crd-ref-docs/pull/37

arkodg:
thanks for raising this @zirain 
looks like the core fix tied to the PR title is
```
generate: kube-generate go.generate docs-api
```

would it be possible to split this up into 2 PRs so the other useful fix of linking upstream Gateway API docs can be brought in later once the PR you raised has made its way into the upstream repo ?

arkodg:
you can circumvent the situation by ignoring the field for now
https://github.com/envoyproxy/gateway/blob/793595957fb9348fd64960aa018e1e20a390bcbf/tools/crd-ref-docs/config.yaml#L4

.github/markdown_lint_config.json
zirain, line 52: add this line to circumvent lint error

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

PR #1269, opened 2023-04-08
move to under tools

zirain:


zirain:
prefer to https://github.com/envoyproxy/gateway/pull/1271

PR #1271, opened 2023-04-08
optimize helm valuese generate

misstick86:
As I know, `helm.go` file  function is to replace the variables in the `values.tmpl.yaml` file with environment variables, and then generate `values.yaml` file. This function can be implemented using the `envsubst` command.


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -44,4 +44,4 @@ envoyGatewayMetricsService:
     protocol: TCP
     targetPort: https
 
-kubernetesClusterDomain: cluster.local
+kubernetesClusterDomain: cluster.local
```
zirain, line 47: add new line at end.

zirain (commented):
LGTM

arkodg:
guessing this works for Mac and Linux but will fail for Windows ?

zirain:
use https://github.com/drone/envsubst ?

tools/make/helm.mk
```diff
@@ -32,4 +32,4 @@ helm-release: helm-generate
 	helm package charts/${CHART_NAME} --app-version ${TAG} --version ${CHART_VERSION} --destination ${OUTPUT_DIR}/charts/
 
 helm-generate:
-	IMAGE=${IMAGE} RELEASE_TAG=${TAG} go run helm/*
+	ImageRepository=${IMAGE} ImageTag=${TAG} envsubst < charts/gateway-helm/values.tmpl.yaml > ./charts/gateway-helm/values.yaml
```
Xunzhuo, line 35: I am sure the Env Name is correct. When release Helm chart, we use IMAGE and RELEASE_TAG from release workflow. ImageRepository and ImageTag are not set in release workflow.

misstick86: @Xunzhuo oh, I see what you mean. when helm chart releasing, you'll set a `TAG` environment variable. [release](https://github.com/envoyproxy/gateway/blob/main/.github/workflows/release.yaml#L34) but it's helm tag(may be it called helm app version), not *gateway* image tag. Should we set two different variable names to differentiate this?

Xunzhuo: Plz add changes like shown above.

misstick86: added, please check again.

charts/gateway-helm/values.tmpl.yaml
misstick86: Shall we need this? according to this [pr](https://github.com/envoyproxy/gateway/pull/1277/files), this is removed.

Xunzhuo: We can remove it for now.

misstick86: done

Xunzhuo (approved):
Looks good to me. Thanks

zirain (approved):

misstick86:
there are a little bit weird.  It's total fine When I execute this testing method alone. Shall I do something about it now?

Issue #866, opened 2023-01-06
docs: allow users to configure custom certs for control plane auth


arkodg:
*Description*:
https://github.com/envoyproxy/gateway/pull/208 added support for control plane auth in the project, by introducing a job to generate self signed certs and configure them in envoy gateway and envoy proxy.

However users might want to create and plug in their own custom certs for control plane auth (between envoy gateway and envoy proxy)

Relates to https://github.com/envoyproxy/gateway/issues/862

zirain:
what about support [cert-manager](https://cert-manager.io/docs/)?

Xunzhuo:
Should we close this by already adding https://gateway.envoyproxy.io/latest/user/security/tls-cert-manager ?

Xunzhuo:
cc @arkodg

arkodg:
this is control plane certs, we still need a user doc for this

PR #1273, opened 2023-04-10
helm: integrates with cert-manager

zirain:
xref: https://github.com/envoyproxy/gateway/issues/866

### DIFF
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -1,3 +1,4 @@
+{{ if not .Values.certManager.enabled }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -41,3 +42,4 @@ subjects:
 - kind: ServiceAccount
   name: '{{ include "eg.fullname" . }}-certgen'
   namespace: '{{ .Release.Namespace }}'
+{{ end }}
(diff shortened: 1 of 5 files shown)


arkodg:
thanks for looking into this @zirain 
* for custom certs, shouldnt it be an API change in `EnvoyGateway` CR specifying the location/name of the cert/secret ?
* The user can leverage `cert-manager` to generate that certs, but imo `cert-manager` shouldnt be part of `helm` install, just docs (highlighting demo)

zirain:
> thanks for looking into this @zirain
> 
> * for custom certs, shouldnt it be an API change in `EnvoyGateway` CR specifying the location/name of the cert/secret ?
> * The user can leverage `cert-manager` to generate that certs, but imo `cert-manager` shouldnt be part of `helm` install, just docs (highlighting demo)

I'm thinking let's cert-manager handle all custom certs requirements, for both `ControlPlane` and `DataPlane`
Could you bring up this topic on the meeting?

arkodg:
we ( including @AliceProxy @kflynn @haq204 ) discussed this issue in the community meeting today and prefer not to add any cert-manager details such as `Certificate` to the helm install. Instead
* introduce an API field (like `controlPlaneAuth`) allowing the user to specify the `certRef`/`Secret` 
* add docs demonstrating above API using `cert-manager` to create the generate the cert and save it as a `Secret`

PR #1277, opened 2023-04-10
Remove unused helm values

arkodg:


arkodg:
hey @Xunzhuo should we disable metrics service and kube-rbac-proxy for 0.4.0 and enable it later in a future release ?

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
```diff
@@ -69,7 +69,7 @@ spec:
         - --v=0
         env:
         - name: KUBERNETES_CLUSTER_DOMAIN
-          value: {{ .Values.kubernetesClusterDomain }}
```
zirain, line 72: why remove this?

arkodg: Imo we need to treat helm values as APi fields and expose them when needed, else we cause breaking changes, the goal of this PR is to expose minimum helm values needed for 0.4.0

zirain: but this need be configurable in the future, xref: https://github.com/envoyproxy/gateway/issues/1280

Xunzhuo (approved):

Xunzhuo:
Thoughts on https://github.com/envoyproxy/gateway/pull/1277#discussion_r1162220737 ?
cc @arkodg, we can remove it for now

Xunzhuo:
Merge it since I am adding docs for values

PR #1279, opened 2023-04-10
refactor egctl translate

arkodg:
* Use a TranslationResult struct to hold output


AliceProxy (approved):
I don't see any issues here :+1:

qicz (approved):
LGTM

PR #1285, opened 2023-04-11
Extensions: pass watched resources to IR

AliceProxy:
Work has already landed to support Envoy Gateway watching resources from extensions, and also for passing those resources back to the extension whenever the Route level xDS hook is called. This is the last piece of the work that updates the gateway API translator to store these watched resources in the IR so that they can be consumed later.

### DIFF
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -80,7 +80,7 @@ func (t *Translator) ProcessHTTPFilters(parentRef *RouteParentContext,
 		if httpFiltersContext.DirectResponse != nil {
 			break
 		}
-		if err := ValidateHTTPRouteFilter(&filter); err != nil {
+		if err := ValidateHTTPRouteFilter(&filter, t.ExtensionGroupKinds...); err != nil {
 			t.processInvalidHTTPFilter(string(filter.Type), httpFiltersContext, err)
 			break
 		}
@@ -759,6 +759,30 @@ func (t *Translator) processExtensionRefHTTPFilter(extFilter *v1beta1.LocalObjec
 
 				}
 				filterContext.HTTPFilterIR.RateLimit = rateLimit
+				return
+			}
+		}
+	}
+
+	// This list of resources will be empty unless an extension is loaded (and introduces resources)
+	for _, res := range resources.ExtensionRefFilters {
+		if res.GetKind() == string(extFilter.Kind) && res.GetName() == string(extFilter.Name) && res.GetNamespace() == filterNs {
+			apiVers := res.GetAPIVersion()
+
+			// To get only the group we cut off the version.
+			// This could be a one liner but just to be safe we check that the APIVersion is properly formatted
+			idx := strings.IndexByte(apiVers, '/')
+			if idx == -1 {
+				errMsg := fmt.Sprintf("Unable to translate APIVersion for Extension Filter: kind: %s, %s/%s", res.GetKind(), filterNs, extFilter.Name)
+				t.processUnresolvedHTTPFilter(errMsg, filterContext)
+				return
+			}
+			group := apiVers[:idx]
+			if group == string(extFilter.Group) {
+				resource := res // Capture loop variable
+				filterContext.ExtensionRefs = append(filterContext.ExtensionRefs, &ir.UnstructuredRef{
+					Object: &resource,
+				})
 				return
 			}
 		}
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -8,6 +8,7 @@ package runner
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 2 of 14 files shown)


Xunzhuo:
Can you rebase with main to fix the CI failure.

internal/gatewayapi/runner/runner.go
```diff
@@ -61,6 +62,15 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				GatewayClassName:       v1beta1.ObjectName(update.Key),
 				GlobalRateLimitEnabled: r.EnvoyGateway.RateLimit != nil,
 			}
+
+			// If an extension is loaded, pass its supported groups/kinds to to translator
```
arkodg, line 66:
```suggestion
			// If an extension is loaded, pass its supported groups/kinds to the translator
```

internal/ir/xds.go
```diff
@@ -230,7 +230,6 @@ type HTTPRoute struct {
 	// RequestAuthentication defines the schema for authenticating HTTP requests.
 	RequestAuthentication *RequestAuthentication
 	// ExtensionRefs holds unstructured resources that were introduced by an extension and used on the HTTPRoute as extensionRef filters
-	// TODO: (aliceproxy) in a follow-up PR, update the translator to store the watched resources in this IR
 	ExtensionRefs []*UnstructuredRef
```
arkodg, line 233: this field is lo longer needed, right ?

internal/gatewayapi/translator.go
```diff
@@ -65,6 +66,11 @@ type Translator struct {
 	// GlobalRateLimitEnabled is true when global
 	// ratelimiting has been configured by the admin.
 	GlobalRateLimitEnabled bool
+
+	// ExtensionGroupKinds stores the group/kind for all resources
+	// introduced by an Extension so that the translator can
+	// store referenced resources in the IR for later use.
+	ExtensionGroupKinds []schema.GroupKind
```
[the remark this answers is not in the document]
AliceProxy: The `TranslationResult.Resources` only has `ExtensionRefFilters   []unstructured.Unstructured` which are just the resources from an extension that we're watching. I added this so that in the gateway API translator, we have a list of the Extension's supported groups/kinds for the call to `ValidateHTTPRouteFilter`. The `gatewayAPIReconciler` has `extGVKs   []schema.GroupVersionKind` for the same reason, but `ValidateHTTPRouteFilter` is also called from the gateway API translator which doesn't have that info currently.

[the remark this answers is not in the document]
arkodg: is there a need to validate twice, once in provider and another time in gateway-api translator ?

internal/gatewayapi/filters.go
```diff
@@ -764,6 +764,30 @@ func (t *Translator) processExtensionRefHTTPFilter(extFilter *v1beta1.LocalObjec
 		}
 	}
 
+	// This list of resources will be empty unless an extension is loaded (and introduces resources)
+	for _, res := range resources.ExtensionRefFilters {
```
arkodg, line 768: imo validation in the gateway API layer is useful to check whether the extFilter within a `HTTPRoute` points to an existing extension resource. This code loops over `resources.ExtensionRefFilters` rather than checking whether the filter is an extension type and whether the typed instance exists or not

AliceProxy: Let me know if my following understanding sounds correct to you or if I missed your point here and need to fix something.

So there are two different groups for the "validity" of extensionRef filters. The first is whether the group/kind is supported or not:
1. Group/Kind is natively supported (Authentication/RateLimit)
2. Group/Kind is supported by extensions
3. Group/Kind is not supported

The second is whether the resource exists or not (we don't get to this check unless the group/kind was already confirmed to be supported):
1. Reference is found
2. Reference is not found

>This code loops over resources.ExtensionRefFilters rather than checking whether the filter is an extension type and whether the typed instance exists or not

This logic in `processExtensionRefHTTPFilter()` is taking a `LocalObjectReference` and the list of loaded resources that EG is watching to see if the referenced object exists or not, so the section I added should be doing a similar job to the bits of code above it that check the Auth / Ratelimit resources.

`ValidateHTTPRouteFilter()` is what performs the check you mentioned on whether the group/kind is supported or not and it gets executed before we run `processExtensionRefHTTPFilter()`. If the group/kind is not supported then we break out early and do not execute `processExtensionRefHTTPFilter()`.

`resources.ExtensionRefFilters` only contains resources that we are watching for an extension, so it should never contain unsupported resources, now we just need to confirm if the resource that was referenced exists in there or not if it isn't an Auth/Ratelimit filter that is natively supported.

For Auth/Ratelimit we can first just check if the `extFilter`'s type matches the Auth/Ratelimit types since they are static and we know what they are so we can just check `resources.AuthenticationFilters` and `resources.RateLimitFilters` respectively. For extension resources, we don't know what all the types stored in `resources.ExtensionRefFilters` are, but we do know for certain that `resources.ExtensionRefFilters` only contains resources owned by the extension. So that's why I had it iterate over the resources first and just make sure that the name/namespace/group/kind match the `LocalObjectReference`.

I added another test case file just in case to make sure that even if `resources.extensionRefFilters` contains a referenced resource that is not supported by extensions (it never should since it only contains watched extension resources), we still set `UnsupportedValue`: `Invalid filter ExtensionRef: unknown kind foo.example.io/Unsupported` rather than `ResolvedRefs`: `Reference default/example not found for filter type: Foo`.

zhaohuabing (commented):
It's just a question rather than a comment: When the extension fitlers are finally translated to xds, we need to call a third-party webhook, right?

AliceProxy:
> It's just a question rather than a comment: When the extension fitlers are finally translated to xds, we need to call a third-party webhook, right?

Yes, at a high level:
1. You configure whether an extension is being used or not in the `EnvoyGateway` config. You can supply the connection information to talk to the extension (hostname/port/tls), then you can also supply a list of group/version/kinds for custom resources that the extension is introducing. These resources can be used on `HTTPRoute`/`GRPCRoute` objects as `externalRef` filters. 
2. There are currently only implementations for "post xDS" hook points where after Envoy Gateway generates an xDS resource such as a route for example, it will call out to the extension via a gRPC hook and allow the Extension to modify the route before it is finalized. The route level post xDS hook also sends these resources over to the extension.

You can review the implemented changes in this merged PR https://github.com/envoyproxy/gateway/pull/1209

arkodg (approved):
LGTM thanks for the clarification on comments and adding the test cases

zhaohuabing (approved):
LGTM, great job!

Issue #1230, opened 2023-03-27
Merge Gateways on to a single EnvoyProxy fleet


arkodg:
*Description*:
* Today we map the GatewayAPI `Gateway` resource on to a unique Envoy Proxy fleet (https://github.com/envoyproxy/gateway/pull/404 has more info) , creating a new Envoy Proxy infrastructure for every `Gateway` resource created. This design does not efficiently utilize the infra resources in the clusters
* This approach was needed because the Gateway API spec does allow a Listener (Port, Protocol, Hostname) tuple to be same across Gateway resources.

Raising this issue to create a way for Gateway resources to be merged on to the same Envoy Proxy Infrastructure by the user signaling this intent - one suggestion is to define a field in the GatewayClass paramsRef resource (`EnvoyProxy`) called
`mergeGateways`. Setting this field would merge all Gateway Listeners but would also ensure that this tuple (Port, Protocol, Hostname) is unique across all Listeners on all Gateways (if not, the newest conflicting listener would be rejected)

cnvergence:
I can take a look at this one :) @arkodg

arkodg:
great @cnvergence thanks !
recommend starting off with an API PR, since this is a API change, building consensus might get take time, will help out to ensure this work moves in a positive direction.

Another thing to note, that there is an alternative proposal upstream https://github.com/kubernetes-sigs/gateway-api/pull/1863 which is taking the approach of linking child (Gateway) to parent (Gateway) thereby merging with it, rather than what is being suggested here - parent (GatewayClass) enforces merge semantics and children (Gateway) are merged into 1. I personally prefer what is suggested in this issue, but would like the community to chime in on which approach to take

tamalsaha:
We are looking to use Envoy Gateway to expose databases using TCP connections. We would like users to be able to define their own listener per database (not edit the same Gateway object). This feature will help us make sure one IP can expose many different databases different TCP ports. 

My understanding is that this is a supported model for the service gateway project. We would appreciate support for this in Envoy gateway. 
https://kubernetes.slack.com/archives/CR0H13KGA/p1691411234021039?thread_ts=1691400241.997149&cid=CR0H13KGA

arkodg:
if we do decide to go ahead and not wait for upstream support, imo there are two options here
* add `mergeGateways` boolean field to `EnvoyGateway` API
  * cons -
    * affects all `GatewayClasses`, so the user loses out on granularity, today we support only 1, but we may support multiple in the future
[...]

PR #1289, opened 2023-04-11
api: add mergeGateways field to EnvoyProxy spec

cnvergence:
Extend API to provide a way to merge gateways

See also:
- https://github.com/envoyproxy/gateway/issues/1230


api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -57,6 +57,12 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Bootstrap *string `json:"bootstrap,omitempty"`
+
+	// MergeGateways defines if Gateway resources should be merged onto the same Envoy Proxy Infrastructure.
+	// Setting this field to true would merge all Gateway Listeners under the parent Gateway Class.
```
arkodg, line 62:
```suggestion
	// Setting this field to true would merge all Gateway Listeners under the parent Gateway Class.
	// This means that the port, protocol and hostname tuple must be unique for every listener.
	// If a duplicate listener is detected, the newer listener (based on timestamp) will be rejected and its status will be updated with a "Accepted=False" condition..
```

arkodg:
hey @cnvergence thanks for raising this PR, lets wait a bit to see if upstream plans on introducing a native API for this in v0.8.0, else we can continue with this approach.

cnvergence:
sure @arkodg, let's put it on hold for now then

cnvergence:
@arkodg PTAL once again

arkodg (approved):
LGTM thanks !

AliceProxy (approved):

Issue #1290, opened 2023-04-11
Flaky test TestTranslate/gateway-with-listener-with-valid-multiple-tls-configuration


arkodg:
*Description*:
Seeing the test fail
```
--- FAIL: TestTranslate (0.57s)
    --- FAIL: TestTranslate/gateway-with-listener-with-valid-multiple-tls-configuration (0.13s)
        translator_test.go:107: 
            	Error Trace:	/Users/arkodebdasgupta/go-workspace/src/github.com/envoyproxy/gateway/internal/gatewayapi/translator_test.go:107
            	Error:      	Should be empty, but was   &gatewayapi.TranslateResult{
[...]

Issue #1275, opened 2023-04-10
Pick TLS Cert based on SNI


arkodg:
*Description*:
Relates to https://github.com/envoyproxy/gateway/pull/1241#discussion_r1161427156

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
cc @chauhanshubham

chauhanshubham:
thanks @arkodg will take this up.

arkodg:
reopening this issue to tracks completion of docs for SNI based cert selection

Issue #1268, opened 2023-04-07
CI is flaky & failing during helm install


arkodg:
*Description*:

See CI sometimes fail on `main` 
https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333

```
===========> Running kube-deploy ... 
[256](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:257)
helm install eg charts/gateway-helm --set deployment.envoyGateway.imagePullPolicy=IfNotPresent -n envoy-gateway-system --create-namespace
[257](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:258)
Error: INSTALLATION FAILED: failed pre-install: timed out waiting for the condition
[258](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:259)
make[1]: *** [tools/make/kube.mk:62: kube-deploy] Error 1
[259](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:260)
make[1]: Leaving directory '/home/runner/work/gateway/gateway'
[260](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:261)
make: *** [Makefile:18: _run] Error 2
[261](https://github.com/envoyproxy/gateway/actions/runs/4639838003/jobs/8211280333#step:6:262)
Error: Process completed with exit code 2.
```

arkodg:
another run where this failed
https://github.com/envoyproxy/gateway/actions/runs/4639339239/jobs/8210216144
cc @Xunzhuo

Xunzhuo:
We have a pre-install hook in resource: certgen job. It looks like it is time-out sometimes.

PR #1291, opened 2023-04-11
Relax Listener TLS CertRef constaints

arkodg:
* Fixes https://github.com/envoyproxy/gateway/issues/1275
* Also fixes https://github.com/envoyproxy/gateway/issues/1290 since we use a list instead of a map to generate a deterministic output


arkodg:
docs for #1275 can be added in a follow up PR

qicz (approved):
LGTM

zhaohuabing (approved):
LGTM

BTW, why there was this one secret/cert per alog constraint in the first place?

Xunzhuo (dismissed):
LGTM

Xunzhuo:
Conformance is failed cc @arkodg

arkodg:
@Xunzhuo its failing because of https://github.com/envoyproxy/gateway/issues/1268

chauhanshubham (approved):
thanks @arkodg 
In addition to this, I was also thinking if we could add status updates or validations for the following two scenarios
- single fqn has conflicting certs configured with same algo (the removed test basically)
- no valid cert is found for an fqn
also discussed here https://github.com/envoyproxy/envoy/issues/21739#issuecomment-1158688548 and https://github.com/envoyproxy/envoy/issues/21739#issuecomment-1176086430 in the same issue

arkodg:
> thanks @arkodg In addition to this, I was also thinking if we could add status updates or validations for the following two scenarios
> 
> * single fqn has conflicting certs configured with same algo (the removed test basically)
> * no valid cert is found for an fqn
>   also discussed here [SNI-based cert selection in TLS transport socket envoy#21739 (comment)](https://github.com/envoyproxy/envoy/issues/21739#issuecomment-1158688548) and [SNI-based cert selection in TLS transport socket envoy#21739 (comment)](https://github.com/envoyproxy/envoy/issues/21739#issuecomment-1176086430) in the same issue

thanks for highlighting the edge cases, let me raise an issue for this

Issue #1217, opened 2023-03-26
Support Envoy RateLimit extension settings


qicz:
support more ratelimit settings

- [ ] redis type ref: [https://github.com/envoyproxy/ratelimit#redis-type](https://github.com/envoyproxy/ratelimit#redis-type)
- [ ] redis auth ref: [https://github.com/envoyproxy/ratelimit#redis](https://github.com/envoyproxy/ratelimit#redis)

qicz:
pls assign to me

arkodg:
hi @qicz, would also help if you shared why you needed these settings to be exposed

qicz:
> hi @qicz, would also help if you shared why you needed these settings to be exposed

if the redis server needs auth, we should set REDIS_AUTH.

arkodg:
yes redis auth is useful, thinking out loud
```
tls: {}
```
enables `tls`

```
tls:
  auth (or password):
  certificateRef: 
```
for specifying password and certs

mingshun:
Is there any progress on redis auth? I need it on public cloud redis cache.

arkodg:
the final piece left in this epic is to support `auth` and `ca.cert` .
Here is an API I'm proposing 
```
+++ b/api/v1alpha1/envoygateway_types.go
@@ -356,7 +356,11 @@ const (
 // RedisTLSSettings defines the TLS configuration for connecting to redis database.
 type RedisTLSSettings struct {
        // CertificateRef defines the client certificate reference for TLS connections.
-       // Currently only a Kubernetes Secret of type TLS is supported.
+       // Here are the valid (optional) keys that can be defined
+       // `tls.key` - TLS Key
+       // `tls.crt` - TLS Certificate
+       // `ca.crt` - CA Certificate
+       // `auth` - Authentication string either in `<password>` or `<username>:<password>` format
        // +optional
        CertificateRef *gwapiv1.SecretObjectReference `json:"certificateRef,omitempty"`
 }
```

This will hopefully require us to selectively select env vars based on whether the secret key exists or not
@qicz can you help look into this one ?

until this feature is implanted, @mingshun a workaround is to use https://gateway.envoyproxy.io/v0.6.0/api/extension_types/#envoygatewaykubernetesprovider to manually setup the `REDIS_AUTH` env var and fetch the value from a secret

mingshun:
[...]

Issue #988, opened 2023-02-06
global-ratelimiting: add ability to configure TLS from rate limit service to Redis backend


arkodg:
*Description*:
Envoy-ratelimiting service supports configuring TLS between the service (client) and the Redis backend
https://github.com/envoyproxy/ratelimit#redis
This should be exposed in[ EnvoyGateway](https://github.com/envoyproxy/gateway/blob/fadc1429e503fc8d70101233e15c0e4810fb9cb7/api/config/v1alpha1/envoygateway_types.go#L134) so the admin can configure this.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

danehans:
@Xunzhuo can you call out this Rate Limit issue in the v0.3 release notes? @arkodg you may want to not this in the RL user guide too.

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/1294

PR #1294, opened 2023-04-12
feat: support envoy ratelimit TLS  settings.

qicz:
part support #1217 

configuration refrence

- `tlsCertificateRef` the client certificate reference secret for TLS connections.
- `env` defines the extension env configuration for connecting to Redis database. reference https://github.com/envoyproxy/ratelimit/blob/main/src/settings/settings.go

```yaml
kind: EnvoyGateway
gateway:
provider:
  type: Kubernetes
  kubernetes:
    rateLimitDeployment:
      replicas: 1
      container:
        env:
        - name: REDIS_POOL_SIZE
          value: 10
        - name: REDIS_PERSECOND
          value: true
        resources:
          requests:
            cpu: 100m
            memory: 512Mi
        securityContext:
          runAsUser: 2000
      pod:
        annotations:
          key1: val1
          key2: val2
        securityContext:
          runAsUser: 1000
          runAsGroup: 3000
          fsGroup: 2000
rateLimit:
  backend:
    type: Redis
    redis:
      url: localhost:6379
      tls
         certificateRef:
             name: secret_name
```

cc @arkodg


qicz:
ptal @arkodg

arkodg:
@qicz can this API PR be broken up into two parts
* one that exposes TLS config b/w RL Service and Redis https://github.com/envoyproxy/gateway/issues/988
* other than allows user to configure environment variables for the container . Would prefer only exposing `env` now, and making each config a first class citizen based on user requests

qicz:
> @qicz can this API PR be broken up into two parts
> 
> * one that exposes TLS config b/w RL Service and Redis [global-ratelimiting: add ability to configure TLS from rate limit service to Redis backend #988](https://github.com/envoyproxy/gateway/issues/988)
> * other than allows user to configure environment variables for the container . Would prefer only exposing `env` now, and making each config a first class citizen based on user requests

My plan is also to divide into two PRS to complete this feature when the api like PR description confirmed and merged. So current api structure is ok or not?

arkodg:
> > @qicz can this API PR be broken up into two parts
> > 
> > * one that exposes TLS config b/w RL Service and Redis [global-ratelimiting: add ability to configure TLS from rate limit service to Redis backend #988](https://github.com/envoyproxy/gateway/issues/988)
> > * other than allows user to configure environment variables for the container . Would prefer only exposing `env` now, and making each config a first class citizen based on user requests
> 
> My plan is also to divide into two PRS to complete this feature when the api like PR description confirmed and merged. So current api structure is ok or not?

would prefer if the 2 API pieces were added as separate PRs because they are different features
high level comments on existing API
* `env` should live inside `rateLimitDeployment.container`
* `tls` style looks mostly good
* I dont think we should introduce `type` so soon , and let the user configure that field using `env`

qicz:
> > > @qicz can this API PR be broken up into two parts
> > > 
> > > * one that exposes TLS config b/w RL Service and Redis [global-ratelimiting: add ability to configure TLS from rate limit service to Redis backend #988](https://github.com/envoyproxy/gateway/issues/988)
> > > * other than allows user to configure environment variables for the container . Would prefer only exposing `env` now, and making each config a first class citizen based on user requests
> > 
> > 
> > My plan is also to divide into two PRS to complete this feature when the api like PR description confirmed and merged. So current api structure is ok or not?
> 
> would prefer if the 2 API pieces were added as separate PRs because they are different features high level comments on existing API
> 
> * `env` should live inside `rateLimitDeployment.container`
> * `tls` style looks mostly good
> * I dont think we should introduce `type` so soon , and let the user configure that field using `env`

I will do these after #1259

qicz:
@arkodg i am testing the reivewers requests but looks missing some permission

qicz:
ptal @arkodg

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,28 @@ const (
...
 )
 
-// RateLimitRedisSettings defines the configuration for connecting to
-// a Redis database.
+// RedisTLSSettings defines the TLS configuration for connecting to redis database.
+type RedisTLSSettings struct {
+	// Auth defines auth configuration to enable password-only authentication to the redis host.
+	//
+	// +optional
+	Auth string `json:"auth,omitempty"`
```
arkodg, line 147: this feels like something that should be part of a secret and not directly in the API
then the question is, should be be part a different `Opaque` secret from the `TLS` secrets for the certs or the same one ?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,28 @@ const (
...
+
+	// CertificateRef defines the client certificate reference kubernetes secret for TLS connections.
+	//
+	// +optional
```
arkodg, line 151: optional fields are always pointer based

api/config/v1alpha1/envoygateway_types.go
arkodg: or should we rm the `auth` field for now and ask users to use
```
kind: EnvoyGateway
    provider:
      type: Kubernetes
      kubernetes:
        rateLimitDeployment:
            env:
            - name: REDIS_AUTH
               valueFrom:
                 secretKeyRef:
                   name: redis-passwordd
                   key: password
```

api/config/v1alpha1/envoygateway_types.go
qicz: from k8s.io
```golang
// A single application container that you want to run within a pod.
type Container struct {
	// Name of the container specified as a DNS_LABEL.
	// Each container in a pod must have a unique name (DNS_LABEL).
	// Cannot be updated.
	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
	// Container image name.
	// More info: https://kubernetes.io/docs/concepts/containers/images
	// This field is optional to allow higher level config management to default or override
	// container images in workload controllers like Deployments and StatefulSets.
	// +optional
	Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
```
Is pointer unnecessary?

api/config/v1alpha1/envoygateway_types.go
qicz: I will rm `auth` and `RedisTLSSettings`, using `TLSCertificateRef` in `RateLimitRedisSettings` directly


```suggestion
	// RateLimitRedisSettings defines the configuration for connecting to redis database.
type RateLimitRedisSettings struct {
	// URL of the Redis Database.
	URL string `json:"url"`

	// TLSCertificateRef defines the client certificate reference kubernetes secret for TLS connections.
	//
	// +optional
	TLSCertificateRef string `json:"tlsCertificateRef,omitempty"`
}
```

api/config/v1alpha1/envoygateway_types.go
arkodg: this might be because `name=image`

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,15 @@ const (
...
-// a Redis database.
+// RateLimitRedisSettings defines the configuration for connecting to redis database.
 type RateLimitRedisSettings struct {
 	// URL of the Redis Database.
 	URL string `json:"url"`
+
+	// TLSCertificateRef defines the client certificate reference kubernetes secret for TLS connections.
+	//
+	// +optional
+	TLSCertificateRef string `json:"tlsCertificateRef,omitempty"`
```
arkodg, line 150: I preferred the previous definition which had a top level `tls`. It allows you to add more fields in there in the future like CA cert

arkodg: https://github.com/envoyproxy/gateway/blob/ea2a688f61451b5cfebf2e23fd6a85efc9d4ea69/api/config/v1alpha1/envoygateway_types.go#L197 is a good reference

qicz: > I preferred the previous definition which had a top level `tls`. It allows you to add more fields in there in the future like CA cert

agree, but more setting can be override by env, the CA needs different config? Imo, maybe unnecessary?

If make sure to do this, i refactor.tomorrow

qicz: > https://github.com/envoyproxy/gateway/blob/ea2a688f61451b5cfebf2e23fd6a85efc9d4ea69/api/config/v1alpha1/envoygateway_types.go#L197
> 
> is a good reference

IMO, the `ratelimit` tls and other more fileds are different from `ExtensionTLS `, the `ratelimit` tls secret from same ns. 
so, I suggest to use the `RedisTLSSettings` directly for  the `ratelimit` extension settings.

api/config/v1alpha1/envoygateway_types.go
qicz: using pointer

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,23 @@ const (
...
 )
 
-// RateLimitRedisSettings defines the configuration for connecting to
-// a Redis database.
+// RedisTLSSettings defines the TLS configuration for connecting to redis database.
+type RedisTLSSettings struct {
+	// CertificateRef defines the client certificate reference kubernetes secret for TLS connections.
+	//
+	// +optional
+	CertificateRef *string `json:"certificateRef,omitempty"`
```
arkodg, line 147: as shared before, lets keep the `tls` APIs identical
https://github.com/envoyproxy/gateway/blob/main/api/config/v1alpha1/envoygateway_types.go
```
	// CertificateRef contains a references to objects (Kubernetes objects or otherwise) that
	// contains a TLS certificate and private keys. These certificates are used to
	// establish a TLS handshake to the redis server.
	//
	// CertificateRef can only reference a Kubernetes Secret at this time.
	//
	// +kubebuilder:validation:Required
	CertificateRef gwapiv1b1.SecretObjectReference `json:"certificateRef"`
```

qicz: done

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,24 @@ const (
 	RedisBackendType RateLimitDatabaseBackendType = "Redis"
 )
 
-// RateLimitRedisSettings defines the configuration for connecting to
-// a Redis database.
+// RedisTLSSettings defines the TLS configuration for connecting to redis database.
+type RedisTLSSettings struct {
+	// CertificateRef defines the client certificate reference kubernetes secret for TLS connections.
+	//
+	// +optional
```
arkodg, line 146: since `validation:Required` is added, lets rm the optional tag ?

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -6,18 +6,40 @@
...
+	// RedisTLSClientKeyEnvVar is the redis tls client key.
+	RedisTLSClientKeyEnvVar = "REDIS_TLS_CLIENT_KEY"
+	// RedisTLSClientKeyFilename is the redis client key file.
+	RedisTLSClientKeyFilename = "/certs/tls.key"
+	// RedisTLSCaCertEnvVar is the redis tls ca cert.
+	RedisTLSCaCertEnvVar = "REDIS_TLS_CACERT"
+	// RedisTLSCaCertFilename is the redis tls ca cert file.
+	RedisTLSCaCertFilename = "/certs/ca.crt"
+	// RedisTLSSkipHostNameVerificationEnvVar is the redis skip hostname verification.
+	RedisTLSSkipHostNameVerificationEnvVar = "REDIS_TLS_SKIP_HOSTNAME_VERIFICATION"
```
arkodg, line 42: we shouldnt allow this by default, lets rm this, use can explicitly set this with `env` if they want to

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -6,18 +6,40 @@
...
+	"github.com/envoyproxy/gateway/internal/provider/kubernetes"
 )
 
 const (
 	// RedisSocketTypeEnvVar is the redis socket type.
 	RedisSocketTypeEnvVar = "REDIS_SOCKET_TYPE"
 	// RedisURLEnvVar is the redis url.
 	RedisURLEnvVar = "REDIS_URL"
+	// RedisAuthEnvVar is the redis auth.
+	RedisAuthEnvVar = "REDIS_AUTH"
```
arkodg, line 26: lets rm this, since its not used

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -87,16 +103,68 @@ func expectedRateLimitContainers(ratelimit *egcfgv1a1.RateLimit, rateLimitDeploy
...
+			Name:      "certs",
+			MountPath: "/certs",
+			ReadOnly:  true,
+		})
+	}
+
+	return volumeMounts
+}
+
+// exceptedDeploymentVolumes returns expected rateLimit deployment volumes.
```
arkodg, line 128:
```suggestion
// expectedDeploymentVolumes returns expected rateLimit deployment volumes.
```

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -124,21 +192,60 @@ func expectedRateLimitContainerEnv(ratelimit *egcfgv1a1.RateLimit, rateLimitDepl
...
 }
+
+// Validate the ratelimit tls secret validating.
+func Validate(ctx context.Context, client client.Client, gateway *egcfgv1a1.EnvoyGateway, namespace string) error {
+	if gateway.RateLimit.Backend.Redis.TLS == nil {
+		return nil
+	}
+
+	certificateRef := gateway.RateLimit.Backend.Redis.TLS.CertificateRef
+	if certificateRef == nil {
```
arkodg, line 244: this should never be nil, since its required, lets make this a non pointer 🙈  ?

qicz:
ptal @arkodg all comments resolved

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -139,11 +139,23 @@ const (
 	RedisBackendType RateLimitDatabaseBackendType = "Redis"
 )
 
-// RateLimitRedisSettings defines the configuration for connecting to
-// a Redis database.
+// RedisTLSSettings defines the TLS configuration for connecting to redis database.
+type RedisTLSSettings struct {
+	// CertificateRef defines the client certificate reference kubernetes secret for TLS connections.
+	//
+	// +kubebuilder:validation:Required
```
arkodg, line 146: Have you tried using this config ?

I feel like there are two cases for TLS
1. simple TLS where only TLS needs to be set (no certs need to be specified), and the system CA Cert is used to validate server chain
2. mutual TLS where where certs + CA cert needs to be specified since it's likely that the CA cert for this case is a custom one.

qicz: user can set tls and env replace some field. Using tls.key tls.crt by default.

qicz: In fact, ca.crt is needs but the user can also replace it, so i rm it for your suggestion

arkodg: my bad
can we add
```
CACertificateRef *gwapiv1b1.SecretObjectReference `json:"caCertificateRef,omitempty"`
```
as well

arkodg (approved):

internal/kubernetes/secret.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	k8smachinery "k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
```
zirain, line 16: redunent empty line

zirain (approved):

zirain (approved):

PR #1203, opened 2023-03-23
feat: expose envoyproxy image by proxyConfig

Xunzhuo:
1. add doc `Customize EnvoyProxy`
2. add support to customize envoyproxy image

Issue #1296, opened 2023-04-13
followup: expose container image


qicz:
*Description*:
>Describe the issue.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #1297, opened 2023-04-13
fix expose ratelimit image.

qicz:
support #1296 

resolve #1203 review case

cc @arkodg @Xunzhuo

### DIFF
--- a/api/config/v1alpha1/shared_types.go
+++ b/api/config/v1alpha1/shared_types.go
@@ -8,14 +8,16 @@ package v1alpha1
 import corev1 "k8s.io/api/core/v1"
 
 const (
-	// DefaultEnvoyReplicas is the default number of Envoy replicas.
-	DefaultEnvoyReplicas = 1
+	// DefaultDeploymentReplicas is the default number of deployment replicas.
+	DefaultDeploymentReplicas = 1
 	// DefaultDeploymentCPUResourceRequests for deployment cpu resource
 	DefaultDeploymentCPUResourceRequests = "100m"
 	// DefaultDeploymentMemoryResourceRequests for deployment memory resource
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
 	DefaultEnvoyProxyImage = "envoyproxy/envoy-dev:latest"
+	// DefaultRateLimitImage is the default image used by ratelimit.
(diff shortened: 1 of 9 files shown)


Xunzhuo (dismissed):

arkodg:
does this need to go into 0.4.0 ?
cc @AliceProxy

Xunzhuo:
I think this is a follow-up of #1203 and #1218, we probably need it in v0.4.0

Xunzhuo (approved):

api/config/v1alpha1/shared_types.go
```diff
@@ -8,14 +8,16 @@ package v1alpha1
...
+	// DefaultDeploymentReplicas is the default number of deployment replicas.
+	DefaultDeploymentReplicas = 1
 	// DefaultDeploymentCPUResourceRequests for deployment cpu resource
 	DefaultDeploymentCPUResourceRequests = "100m"
 	// DefaultDeploymentMemoryResourceRequests for deployment memory resource
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
 	DefaultEnvoyProxyImage = "envoyproxy/envoy-dev:latest"
+	// DefaultRateLimitImage is the default image used by ratelimit.
+	DefaultRateLimitImage = "envoyproxy/ratelimit:f28024e3"
```
kflynn, line 20: I kind of feel like either this shouldn't be pinned, or the Envoy version should be -- having one pinned to a specific version and the other not feels weird to me. How will this get updated?

docs/latest/user/rate-limit.md
```diff
@@ -143,6 +143,7 @@ data:
         rateLimitDeployment:
           replicas: 1
           container:
+            image: envoyproxy/ratelimit:latest
```
kflynn, line 146: This makes me feel especially weird about the default being pinned. 🙂

kflynn (changes requested):
Looks fine except that having the default ratelimit image pinned to a specific version feels a little odd to me?

api/config/v1alpha1/shared_types.go
qicz: ref  ratelimit_deployment.go

qicz:
> Looks fine except that having the default ratelimit image pinned to a specific version feels a little odd to me?

`Default` image just eg initial config, can change this by `RateLimitDeployment`

PR #1305, opened 2023-04-13
0.4.0-rc.1: Pin Envoy Proxy and Envoy Ratelimit for release

AliceProxy:
542a6047 is the most recent ratelimit tag as of this PR other than `master`


arkodg:
0.4.0 should be based off envoy 1.26,  hoping that is part of a follow up PR

arkodg (approved):

AliceProxy:
> 0.4.0 should be based off envoy 1.26, hoping that is part of a follow up PR

Yep. Since 1.26 isn't out yet I figure we will pin the RC to 1.25-latest for now and then when 1.26 is released we can pin that version for 0.4.0 final.

Issue #1048, opened 2023-02-15
Ensure Envoy recommended edge proxy settings are set by default


arkodg:
*Description*:
Ensure we are setting Envoy Edge Proxy settings by default to the values specified in https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy

TCP proxies 

- [x] restrict access to the admin endpoint, https://github.com/envoyproxy/gateway/pull/206
- [x] [overload_manager](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager#config-overload-manager),
- [x] [listener buffer limits](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes) to 32 KiB, https://github.com/envoyproxy/gateway/pull/1306
- [x] [cluster buffer limits](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-per-connection-buffer-limit-bytes) to 32 KiB. https://github.com/envoyproxy/gateway/pull/1325

HTTP proxies:

- [x] [use_remote_address](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-use-remote-address) to true  (to avoid consuming HTTP headers from external clients, see [HTTP header sanitizing](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_sanitizing#config-http-conn-man-header-sanitizing) for details) https://github.com/envoyproxy/gateway/pull/1029,
- [ ] [connection and stream timeouts](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#faq-configuration-timeouts),
[...]

PR #1306, opened 2023-04-14
feat: set tcp proxy listener buffer

Xunzhuo:
Part of: #1048

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -24,8 +24,14 @@ import (
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+const (
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes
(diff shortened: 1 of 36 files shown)


internal/xds/translator/listener.go
```diff
@@ -39,6 +40,7 @@ func buildXdsTCPListener(name, address string, port uint32) *listenerv3.Listener
 				Filter:     listenerAccessLogFilter,
 			},
 		},
+		PerConnectionBufferLimitBytes: wrapperspb.UInt32(32768),
```
zirain, line 43: should EG support to custom this?

arkodg: lets raise a issue for this ? #24 is always a back up option

arkodg, line 43: would be great if this is a magic number const, with a link to https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes

Xunzhuo: If so, we need to add new apis in envoyproxy too, let us keep it in this way, unless someone requests for it.

Xunzhuo: Done

arkodg (approved):
🚀

zirain (approved):

haorenfsa:
Em... I just tried on v1.1.0, seems not work for me. I add a `BackendTrafficPolicy` but the `per_connection_buffer_limit_bytes` is still 32KB

arkodg:
> Add a knob to make `per_connection_buffer_limit_bytes` configurable #2600

cc @yaelSchechter

Issue #1150, opened 2023-03-14
Support Distinct Match Type for Source IP based Ratelimiting 


arkodg:
*Description*:
* Today [the SourceIP client selector](https://github.com/envoyproxy/gateway/blob/29619902a75ecc8bd454d82727f22a955eda1d83/api/v1alpha1/ratelimitfilter_types.go#L107) allows the user to specify a IP subnet. All IPs within the subnet are treated as the same client / part of the same rate limit bucket i.e. if `SourceIP` is `192.168.1.0/24` then requests from `192.168.1.1` and `192.168.1.2` count towards the same rate limit.

Another use case is, user might want to ratelimit all IPs equally within an IP Subnet i.e. if `SourceIP` is `192.168.1.0/24` then `192.168.1.1` and `192.168.1.2` should be allowed X requests/unit individually. Suggest using a match type called `Distinct` to represent this. Without this feature a user has to specify multiple entries such as `192.168.1.1/32` and `192.168.1.2/32` which can be tedious and hard to manage when these IPs are in the order of hundreds or thousands.

This feature _might_ need upstream work to
* Send Client Address (and not masked address) as a descriptor value from the Rate Limit filter to Rate Limit Service
* Pass a unique key from Rate Limit Filter
* No new work will be needed in `envoyproxy/ratelimit` since it already supports [this](https://github.com/envoyproxy/ratelimit#example-3)

zirain:
IIRC, the result `remote_addres` is IP string without CIDR.

arkodg:
@zirain , the top level descriptor
```
  - key: masked_remote_address //catch all the source IPs inside a CIDR
    value: 192.168.0.0/16
```
is added to separate this match from another RL match

so on the envoy side both https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action-remoteaddress & https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action-maskedremoteaddress need to be set

I think this is genius @zhaohuabing 🚀

zhaohuabing:
> @zirain , the top level descriptor
> ```
>   - key: masked_remote_address //catch all the source IPs inside a CIDR
>     value: 192.168.0.0/16
> ```
> is added to separate this match from another RL match
> 
[...]

PR #1308, opened 2023-04-14
Change ratelimit API to support distinct match

zhaohuabing:
This PR is the API change for https://github.com/envoyproxy/gateway/issues/1150. Once we agreed on API, the implementation will be submitted in another PR.


qicz:
PR title spell error

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -98,13 +98,33 @@ type RateLimitSelectCondition struct {
...
-	// All IP Addresses within the specified SourceIP CIDR are treated as a single client selector and share the same rate limit bucket.
-	//
+	// Deprecated: Use RemoteAddress instead.
 	// +optional
 	SourceIP *string `json:"sourceIP,omitempty"`
+
+	// RemoteAddress is the remote IP Address to match.
+	//
+	// +optional
+	RemoteAddress *RemoteAddress `json:"remoteAddress,omitempty"`
```
arkodg, line 108: prefer `SourceIP` or `SourceIPCIDR` to `RemoteAddress` .

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -98,13 +98,33 @@ type RateLimitSelectCondition struct {
...
+}
+
+type RemoteAddressMatchType string
+
+const (
+	// RemoteAddressMatchCIDR All IP Addresses within the specified SourceIP CIDR are treated as a single client selector
+	// and share the same rate limit bucket.
+	RemoteAddressMatchCIDR RemoteAddressMatchType = "CIDR"
```
arkodg, line 116: thoughts on using `Exact` here ?

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -98,13 +98,33 @@ type RateLimitSelectCondition struct {
...
+	RemoteAddressMatchDistinct RemoteAddressMatchType = "Distinct"
+)
+
+type RemoteAddress struct {
+	Type *RemoteAddressMatchType `json:"type,omitempty"`
+
+	// Address is the IP CIDR that represents the range of Source IP Addresses of the client.
+	// These could also be the intermediate addresses through which the request has flown through and is part of the  `X-Forwarded-For` header.
+	// For example, `192.168.0.1/32`, `192.168.0.0/24`, `001:db8::/64`.
+	Address string `json:"address"`
```
arkodg, line 127: should this be called `value` instead ?

api/v1alpha1/ratelimitfilter_types.go
zhaohuabing: I also prefer SourceIP as it's more precise.  But reusing it will break the existing API. SourceIPCIDR is a good - - alternative.

zhaohuabing: I also prefer SourceIP as it's more precise.  But reusing it will break the existing API. SourceIPCIDR is a good alternative.

api/v1alpha1/ratelimitfilter_types.go
zhaohuabing: Exact sounds not clear enough for me. How about Masked or Subnet if you think CIDR is not suitable here.

AliceProxy (approved):
I like this change and I think it's definitely much needed for common ratelimit use-cases. Looks like a few conformance tests are failing but it could just be because of the pre-install flake we were seeing. I'l restart them real quick.

arkodg:
hey @zhaohuabing wanted to share some more suggestions, hoping @envoyproxy/gateway-maintainers can chime in
1. Name of the new field `SourceCIDR` vs `SourceIPCIDR` vs `SourceIPSubnet`
I vote for `SourceCIDR`
2. Name of the match type
```
sourceCIDR:
  type: Exact
  value: "192.168.1.0/24"
```
vs 

```
sourceCIDR:
  type: CIDR
  value: "192.168.1.0/24"
```

I prefer `type: Exact` because we have the `Exact` match type for header match, so its reusing the noun (one less thing to remember for the user) and `CIDR` as a field name + match type feel repetitiive

zhaohuabing:
sourceCIDR:
  type: Exact
  value: "192.168.1.0/24"


@arkodg Exact is a bit vague. How about Masked or Subnet？

arkodg:
@zhaohuabing `Masked` and `Subnet` sound the same as `CIDR`, my argument is the `value` is already a `CIDR` type so the match of type `Exact` would mean matching on the CIDR value. Hoping this discussion gets more eyes so others can also share their preferences.

zhaohuabing:
I got it. I'm ok with Exact unless we find a better name.

chauhanshubham:
no strong opinions around this but for the current struct
I feel calling out `match` explicitly in the `type` field makes sense.

| current | |
| --- | --- |
| `sourceCIDR: {type: "" value: ""}` | `sourceCIDR: {matchType: "" value: ""}` |

Just so it reads `source CIDR match type` and `source CIDR value`.

`source CIDR type` felt like there's a IP type aka ipv4/ipv6 distinction in value.
Thanks

arkodg:
hey @chauhanshubham this is what it would look like 
```
global:
    rules:
    - clientSelectors:
      - headers:
        - name: x-user-id
           type: Exact
           value: one
        sourceCIDR:
          type: Exact
          value: "192.168.1.0/24"
```      
the word `clientSelectors` is used to describe `matches` but specifically match attributes related to the client (request headers and source IP in this case).
This has the same look and feel as the `matches` field within `HTTPRoute`
```
- matches:
    - headers:
      - type: Exact
        name: magic
        value: foo
```

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -98,13 +98,34 @@ type RateLimitSelectCondition struct {
...
+}
+
+type SourceMatchType string
+
+const (
+	// SourceMatchExact All IP Addresses within the specified Source IP CIDR are treated as a single client selector
+	// and share the same rate limit bucket.
+	SourceMatchExact SourceMatchType = "Exact"
+	// SourceMatchDistinct Each IP Address within the specified Source IP CIDR is treated as a distinct client selector
+	// and use a separate rate limit bucket.
```
arkodg, line 118:
```suggestion
	// and uses a separate rate limit bucket/counter.
```

api/v1alpha1/ratelimitfilter_types.go
```diff
@@ -98,13 +98,34 @@ type RateLimitSelectCondition struct {
...
-	//
+	// Deprecated: Use SourceCIDR instead.
 	// +optional
 	SourceIP *string `json:"sourceIP,omitempty"`
+
+	// SourceCIDR is the client IP Address to match.
+	//
+	// +optional
+	SourceCIDR *SourceMatch `json:"sourceIPCIDR,omitempty"`
+}
```
qicz, line 109: Json flag incorrect

AliceProxy (approved):
These changes work for me

arkodg (approved):

arkodg:
merging, thanks @zhaohuabing !

PR #1311, opened 2023-04-14
Changes HTTPRoute hostname matching to ignore trailing port numbers.

tommie:
This is useful for testing behind NAT, e.g. when deploying in a VM.

HTTP authority is e.g. 'example.com:1234', but the Gateway API restricts the values of Hostnames to domain names [1]. I see two interpretations: (1) port numbers are not allowed or (2) only the host portion should be matched. (1) is not a problem if the cluster owner only uses the standard ports, so (2) seems like a more useful interpretation.

This cannot use a simple `contains_match`, since the separating colon is optional. For performance, perhaps exact matches should be special-cased somehow: that ignoring the port is only done for routes with a certain annotation.

1. https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Hostname

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -319,7 +319,8 @@ xds:
                 headers:
                 - name: :authority
                   stringMatch:
-                    exact: www.example.com
+                    safeRegex:
+                      regex: "www\\.example\\.com(?::\\d+)?"
                 prefix: /
               name: default-backend-rule-0-match-0-www.example.com
               route:
--- a/internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
+++ b/internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
@@ -585,7 +585,9 @@
 									"headers": [{
 										"name": ":authority",
 										"stringMatch": {
-											"exact": "www.example.com"
+											"safeRegex": {
+												"regex": "www\\.example\\.com(?::\\d+)?"
+											}
 										}
 									}],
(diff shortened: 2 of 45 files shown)


arkodg:
thanks for raising this issue @tommie . The issue as you pointed out is with upstream validation
due to this regex check https://github.com/kubernetes-sigs/gateway-api/blob/b751bbdce6245c24556ac8ee13b6de0797515bb8/apis/v1beta1/shared_types.go#L381
that allows `example.com` but not `example.com:1234`.
Imo the fix would be to tweak the validation regex to support this use case since its a valid hostname
can you raise this issue upstream ?

tommie:
Yes, I'll ask for clarification.

arkodg:
thanks for raising the upstream issue, to unblock yourself for now, you would circumvent the issue and ensure tighter authz using the `match` field
```
    kind: HTTPRoute
    metadata:
      namespace: default
      name: httproute-1
    spec:
      parentRefs:
        - namespace: envoy-gateway
          name: gateway-1
      rules:
        - matches:
              headers:
                - name: ":authority"
                  value: "example.com:1234"
```

tommie:
> to unblock yourself for now

Oh, I didn't even think of looking for that option. Thank you!

arkodg:
also thinking out, if the implementation had to solve this issue w/o any upstream changes, rather than making it a regex match (which are slower, and wider (allows all valid ports )), prefer if the gateway-api layer added two exacts matches
* one for `hostname`
* other for `hostname:listerner-port`
having said that it would be good to wait for guidance from upstream on this

thanks again @tommie

arkodg:
closing this PR in favor of the proposed change in https://github.com/envoyproxy/gateway/issues/1368

PR #1323, opened 2023-04-17
Extension: fix pointer error

AliceProxy:
previously I was under the false impression that despite the linter warning about an ineffectual assignment that things were working as intended. This was because the test cases directly operated on passed-in pointers instead of ensuring that the returned pointer was being used.

I added a new `func copyPtr()` `func` to assist with the copying of xDS resource pointer fields instead of doing the following because of the mutex copy.

```
*x = *y
```

I don't like how `copyPtr()` exists twice in two different files and would like to move it to a `helpers`/`utils` file/package. All the existing helpers/utils files seem pretty package specific so I didn't think it would fit in any of them. I might just make a more general/top-level utils package and move it there so let me know if you're a +1/-1 on that plan or if you have a better idea for where this helper `func` can live.

### DIFF
--- a/internal/extension/testutils/hooks.go
+++ b/internal/extension/testutils/hooks.go
@@ -16,6 +16,7 @@ import (
 	listenerV3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	routeV3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	tlsV3 "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+	"github.com/golang/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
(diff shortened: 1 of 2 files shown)


internal/xds/translator/extension.go
```diff
@@ -81,9 +81,7 @@ func processExtensionPostVHostHook(vHost *routev3.VirtualHost, em *extensionType
 
 	// If the extension returned a modified Virtual Host, then copy its to the one that was passed in as a reference
 	if modifiedVH != nil {
-		// Overwrite the pointer for the original virtual host.
-		// Uses nolint because of the ineffectual assignment check
-		vHost = modifiedVH //nolint
+		copyPtr(modifiedVH, vHost)
```
arkodg, line 84: can we use https://pkg.go.dev/github.com/golang/protobuf/proto#Clone here instead ?

internal/xds/translator/extension.go
```diff
@@ -53,9 +55,7 @@ func processExtensionPostRouteHook(route *routev3.Route, vHost *routev3.VirtualH
 
 	// If the extension returned a modified Route, then copy its to the one that was passed in as a reference
 	if modifiedRoute != nil {
-		// Overwrite the pointer for the original route.
-		// Uses nolint because of the ineffectual assignment check
-		route = modifiedRoute //nolint
```
Xunzhuo, line 58: why not just 

```
*route = *modifiedRoute
```

And 

```
*vHost = *modifiedVH
```

internal/extension/testutils/hooks.go
```diff
@@ -161,3 +177,14 @@ func (c *XDSHookClient) PostTranslateModifyHook(clusters []*clusterV3.Cluster, s
 
 	return clusters, secrets, nil
 }
+
+func copyPtr(src interface{}, dest interface{}) {
+	srcVal := reflect.ValueOf(src)
+	if srcVal.Kind() == reflect.Ptr {
```
Xunzhuo, line 183: why just check src Kind, and not check dest Kind ?

internal/xds/translator/extension.go
AliceProxy: This is what I initially wanted to do, but unfortunately it causes a lint error because all the xDS resources have mutexes in the message state fields :sweat_smile:

```
copylocks: assignment copies lock value to *route: github.com/envoyproxy/go-control-plane/envoy/config/route/v3.Route contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
```

internal/xds/translator/extension.go
AliceProxy: I used it to get rid of the duplicated function in the `testutils` functions since those setup new variables, but it won't work for this file since it returns a value, so you end up in the same scenario I described where `*x = *y` and `*x = *proto.Clone(y)` cause lint errors because of the mutex inside all the xDS resource messages, and doing `x = proto.Clone(y)` results in x not really getting updated outside of the function since it's a local variable.

I was looking at `proto.Merge` as a potential solution, but unfortunately that won't work either since if you don't `.Reset()` the initial `Route/VirtualHost` first, then `Merge()` results in all the fields that are lists (like add request headers) getting duplicate entries. If you do call `.Reset()` on the `Route/VirtualHost` first, then `proto.Merge()` does nothing since it only merges fields that are set on the source and dest.

arkodg (approved):

AliceProxy:
@arkodg I know you approved, but since codecov/patch failed I'm happy to add individual coverage for the `func deepCopyPtr()` I added by moving it into the `utils` package at `/internal/xds/utils/utils.go` (I didn't initially do that since I don't think it's really an "xDS themed" util function, nor do I think it needs to be exported) or by adding tests in `translator_test.go` (wouldn't really fit the theme with testing translation bits). Lmk if you have any thoughts.

arkodg:
I'm fine with this going in, I believe there are tests testing this diff

Xunzhuo (approved):

PR #1324, opened 2023-04-17
feat(egctl): provide default services

### LINKED ISSUE #1122: Allow the user to not have to specify all K8s resources in `egctl x translate --to=xds`

muyuan0:
Fix #1122


internal/cmd/egctl/translate.go
```diff
@@ -644,6 +735,46 @@ func kubernetesYAMLToResources(str string, addMissingResources bool) (*gatewayap
 				resources.Namespaces = append(resources.Namespaces, namespace)
 			}
 		}
+
+		requiredServiceMap := map[string]map[string]*v1.Service{}
```
arkodg, line 739: thoughts on using 2 state entities (existingServiceMap, resources.Services)
existingServiceMap := map[string]bool where the key is the namespacedName of the service

So the logic would be - if a backend doesnt exist in map, add it to the map as well as to `resources`

muyuan0: Thanks for your suggestion. It indeed simplifies the logic. There is a need to merge multiple ports of the service in the same name:
```
			inserted := false
			for _, port := range service.Spec.Ports {
				if port.Protocol == servicePort.Protocol && port.Port == servicePort.Port {
					inserted = true
					break
				}
			}

			if !inserted {
				service.Spec.Ports = append(service.Spec.Ports, servicePort)
			}
```
So we still need to track the v1.Service.

arkodg: good catch, agree this is needed

internal/cmd/egctl/translate.go
```diff
@@ -644,6 +730,35 @@ func kubernetesYAMLToResources(str string, addMissingResources bool) (*gatewayap
...
+			addMissingServices(requiredServiceMap, route)
+		}
+
+		providedServiceMap := map[string]bool{}
+		for _, service := range resources.Services {
+			providedServiceMap[service.Namespace+"/"+service.Name] = true
+		}
+
+		for key, service := range requiredServiceMap {
+			if _, found := providedServiceMap[key]; !found {
```
arkodg, line 757: this logic is not quite right, you are building out an accurate `requiredServiceMap` and inserting missing service as well as service ports, but in this line you are not adding the service if the service exists (but not checking if the ports exist)
what are you thoughts on this suggestion
1. pass an existing `serviceMap` (prepopulated with `resources.Services`) to `addMissingServices`
2. rebuild/rewrite `resources.Services` based on `serviceMap`

muyuan0: I would like to rewrite the providedServiceMap (changing the value from bool to v1.Service first), and then rebuild the `resources.Services`. Therefore we don't need to add more arguments to addMissingServices.
What do you think about it?

muyuan0: I would like to rewrite the providedServiceMap (changing the value from bool to v1.Service first), and then rebuild the `resources.Services`. Therefore we don't need to add more arguments to addMissingServices.
What do you think about it?

arkodg: yeah that works, you would need to prepopulate `providedServiceMap` before  with existing `resources.Services`

arkodg (approved):
LGTM, thanks !

AliceProxy (approved):

PR #1325, opened 2023-04-18
feat: set tcp proxy cluster buffer

Xunzhuo:
**What type of PR is this?**

feat: set tcp proxy cluster buffer 

set tcp proxy cluster buffer [regarding Configuring Envoy as an edge proxy](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy)

Part of #1048

### DIFF
--- a/internal/cmd/egctl/config_test.go
+++ b/internal/cmd/egctl/config_test.go
@@ -94,7 +94,7 @@ func TestExtractAllConfigDump(t *testing.T) {
 	}
 
 	for _, tc := range cases {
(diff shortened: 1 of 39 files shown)


arkodg:
@Xunzhuo shouldnt this as well as https://github.com/envoyproxy/gateway/pull/1306 be specific for TCPProxy i.e. xds Listener and xds Cluster associated with Gateway API TCPRoute case ?

arkodg (approved):

zirain (approved):

Issue #1307, opened 2023-04-14
The new `helm template` in `generate-manifests` doesn't create namespace


tommie:
*Description*:

When running `make generate-manifests` in main, the generated `bin/install.yaml` cannot be installed, because `namespace/envoy-gateway-system` is not created. `namespace/gateway-system` from upstream CRDs is created. It doesn't look like this was intentional.

*Repro steps*:

```sh
git clone ...
make generate-manifests
grep -A4 'kind: Namespace' bin/install.yaml

# Compare to
curl -L https://github.com/envoyproxy/gateway/releases/download/v0.3.0/install.yaml | grep -A4 'kind: Namespace'
```

# Related

* https://github.com/helm/helm/issues/10737
* https://github.com/helm/helm/issues/9813

PR #1332, opened 2023-04-19
fix: add the namespace resource within helm templates

arkodg:
This is an unfortunate workaround due the difference in UX between `helm template` and `helm install`
The project recommends `helm install` as a way to install EG which supports a `--create-namespace` flag to create a namespace However we also generate a static YAML using `helm template` as part of the release artifact so a user can install the YAML directly using `kubectl` instead of `helm` . The issue here is `helm template` does not support `--create-namespace`, so instead this commit adds a knob called `createNamespace` to the Helm chart which is `false` by default, but turned on during `make generate-manifests`

Fixes: https://github.com/envoyproxy/gateway/issues/1307

### DIFF
--- a/charts/gateway-helm/templates/namespace.yaml
+++ b/charts/gateway-helm/templates/namespace.yaml
@@ -1,0 +1,6 @@
+{{ if .Values.createNamespace }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: '{{ .Release.Namespace }}' 
+{{ end }}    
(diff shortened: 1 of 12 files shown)


charts/gateway-helm/templates/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+{{ if .Values.createNamespace }}
```
zirain, line 1: maybe I'm out of date, I recall it's not a good practice to put namespace in a helm chart.

arkodg: it is not, but we are using `helm` as the source of truth for manifest templates, in the previous release it was `customize`
* for the common case using `helm install` we skip this file (`createNamespace` is `false`)
* for the manifest generation case to support install via 1 single YAML, we rely on it

zirain (approved):
/lgtm for now

we need review the helm chart in v0.5, this and the `crds` folder.

Xunzhuo (approved):

Xunzhuo:
Thanks @arkodg

PR #1341, opened 2023-04-23
feat: set path normalization settings

tanujd11:
**What type of PR is this?**
 Feature to set defaults in http connection manager

Sets path normalization settings according to the best practices https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge

Fixes part of #1048

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -101,6 +101,11 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 			Name:       wellknown.Router,
 			ConfigType: &hcmv3.HttpFilter_TypedConfig{TypedConfig: routerAny},
 		}},
+		// normalize paths according to RFC 3986
+		NormalizePath: &wrapperspb.BoolValue{Value: true},
+		// merge adjacent slashes in the path
+		MergeSlashes:                 true,
+		PathWithEscapedSlashesAction: hcmv3.HttpConnectionManager_UNESCAPE_AND_REDIRECT,
 	}
 
 	if irListener.IsHTTP2 {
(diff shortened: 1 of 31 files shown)


tanujd11:
@Xunzhuo could you please rerun the CI. I guess there was some issue in uploading the Codecov report in the last run. Thanks

internal/xds/translator/listener.go
```diff
@@ -101,6 +101,11 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 			Name:       wellknown.Router,
 			ConfigType: &hcmv3.HttpFilter_TypedConfig{TypedConfig: routerAny},
 		}},
+		// normalize paths according to RFC 3986
+		NormalizePath: &wrapperspb.BoolValue{Value: true},
+		// merge adjacent slashes in the path
+		MergeSlashes:                 true,
+		PathWithEscapedSlashesAction: hcmv3.HttpConnectionManager_UNESCAPE_AND_REDIRECT,
```
arkodg, line 108: this one is tricky
```
UNESCAPE_AND_REDIRECT if downstream client supports HTTP redirect (i.e. a browser). This option minimizes possibility of path confusion by forcing request to be re-issued with the same path across all parties: downstream client, Envoy and upstream server. Note that gRPC requests will still be rejected with the INTERNAL (13) error code, as gRPC clients do not support redirect.
```
so ideally it should be enabled for HTTP and disabled for gRPC.
Can this one be rm'd from this PR and solved in a follow up

tanujd11: So the default best practice for gRPC is to REJECT_REQUEST 

```REJECT_REQUEST if dowstream clients are expected to use [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) compliant normalized paths (i.e. gRPC clients).```

And the UNESCAPE_AND_REDIRECT will also reject the request right? So won't this cover both the http and gRPC? We can take this in next iteration but is there any gap in my understanding here?

arkodg (approved):
🚀

AliceProxy (approved):
lgtm, thanks!

PR #1342, opened 2023-04-23
fix: #1340 not compare all svc.spec for user modified scene

### LINKED ISSUE #1340: envoyproxy instance svc update error after gateway restart
*Description*:
>What issue is being seen?
after gateway restart, gateway will update the exist envoy proxy instance, so there will encounter an error, seem like https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/infra.go#L74 miss add the original info to new svc.

*Repro steps*:
> apply an gateway cr and envoyproxy cr, check an envoy proxy instance whether created in envoy-gateway-system, then restart the gateway.

*Environment*:
> the main branch code; 
k8s server: v1.19.8

*Logs*:
>2023-04-23T11:00:59.473+0800    ERROR   runner/runner.go:68     failed to create new infra      {"runner": "infrastructure", "error": "failed to update service envoy-gateway-system/envoy-default-eg-64656661: Service \"envoy-default-eg-64656661\" is invalid: spec.clusterIP: Invalid value: \"\": field is immutable"}
github.com/envoyproxy/gateway/internal/infrastructure/runner.(*Runner).subscribeToProxyInfraIR.func1
        gateway/internal/infrastructure/runner/runner.go:68
gateway/internal/message.HandleSubscription[...]
        gateway/internal/message/watchutil.go:28
github.com/envoyproxy/gateway/internal/infrastructure/runner.(*Runner).subscribeToProxyInfraIR
        gateway/internal/infrastructure/runner/runner.go:57

spwangxp:
**What type of PR is this?**
fix: #1340 

if not, the ir will encounter an error when update svc.
for example, user A deployed an envoyproxy, and set the svc type to NodePort instead of LoadBalancer, then the gateway restart, the check will always be error. 

Fixes #1340


internal/infrastructure/kubernetes/applier/apply.go
```diff
@@ -122,7 +122,13 @@ func (i *Instance) CreateOrUpdateService(ctx context.Context, svc *corev1.Servic
...
 	} else {
 		// Update if current value is different.
-		if !reflect.DeepEqual(svc.Spec, current.Spec) {
+		// Only compare the selector and ports(not include nodePort) in case user have modified for some scene.
+		var currentPortsWithoutNP []corev1.ServicePort
+		for _, port := range current.Spec.Ports {
+			port.NodePort = 0
+			currentPortsWithoutNP = append(currentPortsWithoutNP, port)
+		}
+		if !reflect.DeepEqual(svc.Spec.Selector, current.Spec.Selector) && !reflect.DeepEqual(svc.Spec.Selector, currentPortsWithoutNP) {
```
zirain, line 131: I'm thinking use `github.com/google/go-cmp/cmp` with ignore options.
cc @arkodg WDYT?

internal/infrastructure/kubernetes/applier/apply.go
```diff
@@ -122,7 +124,9 @@ func (i *Instance) CreateOrUpdateService(ctx context.Context, svc *corev1.Servic
 		}
 	} else {
 		// Update if current value is different.
-		if !reflect.DeepEqual(svc.Spec, current.Spec) {
+		// Only compare the selector and ports(not include nodePort) in case user have modified for some scene.
+		if !cmp.Equal(svc.Spec.Selector, current.Spec.Selector) &&
```
zirain, line 128: can you move these to a single function for better extension in the future?

spwangxp: ok

internal/infrastructure/kubernetes/utils/utils_test.go
```diff
@@ -76,3 +78,101 @@ func TestGetSelector(t *testing.T) {
...
 	}
 }
+
+func TestCompareSvc(t *testing.T) {
+	cases := []struct {
+		ExpectRet   bool
+		NewSvc      *corev1.Service
+		OriginalSvc *corev1.Service
+	}{
+		{
```
zirain, line 88: can you also add a testcase about what happen in #1340, `clusterip` is updated.

zirain (commented):
LGTM

cc @arkodg for second review

internal/infrastructure/kubernetes/applier/apply.go
```diff
@@ -122,7 +124,7 @@ func (i *Instance) CreateOrUpdateService(ctx context.Context, svc *corev1.Servic
 		}
 	} else {
 		// Update if current value is different.
-		if !reflect.DeepEqual(svc.Spec, current.Spec) {
+		if !utils.CompareSvc(svc, current) {
```
arkodg, line 127: there are two bugs associated with this logic
1. we are comparing the entire `Spec` here, instead  we should be comparing _all_ the fields the infra layer sets or skip the fields the infra layer does not set https://pkg.go.dev/k8s.io/api/core/v1#ServiceSpec.

2. when updating the service, we need to set immutable fields like `ClusterIP` and `ClusterIPs`
```
// clusterIP is the IP address of the service and is usually assigned
	// randomly. If an address is specified manually, is in-range (as per
	// system configuration), and is not in use, it will be allocated to the
	// service; otherwise creation of the service will fail. This field may not
	// be changed through updates unless the type field is also being changed
	// to ExternalName (which requires this field to be blank) or the type
	// field is being changed from ExternalName (in which case this field may
	// optionally be specified, as describe above).  Valid values are "None",
	// empty string (""), or a valid IP address. Setting this to "None" makes a
	// "headless service" (no virtual IP), which is useful when direct endpoint
	// connections are preferred and proxying is not required.  Only applies to
	// types ClusterIP, NodePort, and LoadBalancer. If this field is specified
	// when creating a Service of type ExternalName, creation will fail. This
	// field will be wiped when updating a Service to type ExternalName.
	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies
	// +optional
	ClusterIP [string](https://pkg.go.dev/builtin#string) `json:"clusterIP,omitempty" protobuf:"bytes,3,opt,name=clusterIP"`
```

internal/infrastructure/kubernetes/utils/utils.go
```diff
@@ -30,3 +32,9 @@ func ExpectedServiceSpec(serviceType *egcfgv1a1.ServiceType) corev1.ServiceSpec
 	}
 	return serviceSpec
 }
+
+// CompareSvc Only compare the selector and ports(not include nodePort) in case user have modified for some scene.
+func CompareSvc(newSvc, originalSvc *corev1.Service) bool {
+	return cmp.Equal(newSvc.Spec.Selector, originalSvc.Spec.Selector) &&
```
arkodg, line 38: the infra layer also sets `Type` and `SessionAffinity`
https://github.com/envoyproxy/gateway/blob/690a045b4c54bd7a9acbdd4544f8575be99e2e8a/internal/infrastructure/kubernetes/utils/utils.go#L23

arkodg: something for the future, we should find a sane way to ensure we check what we set

arkodg:
thanks for highlighting this bug @spwangxp, added some comments

arkodg:
also @spwangxp if you are using service type `NodePort`, can you please raise a issue to add support for it ?

internal/infrastructure/kubernetes/applier/apply.go
spwangxp: 1, ok, i'll add imutable fields compare we set. 
2, when updating the svc, is these be better that just patch what we need rather than set everything to the new svc?
3, actually，when we use loadbalacer by default, svc.Spec.ports[*].nodePort will be set an random number by k8s, so it's not just about user modified the svc type.

internal/infrastructure/kubernetes/utils/utils.go
spwangxp: is there any need to update svc type rather than follow user's setting?

internal/infrastructure/kubernetes/applier/apply.go
arkodg: Reg 2. `Patch` is fine, my only request is to keep the code simple so it can be enhanced with more fields in the future
Reg 3. its okay to undo what user did, to solve this, we must
* support `Nodeport` natively
* support BYO Envoy Svc and Envoy Deployment and disable managed infra

internal/infrastructure/kubernetes/utils/utils.go
arkodg: added some comments above, `Patch` is fine as long as its easy to mantain

internal/infrastructure/kubernetes/applier/apply.go
spwangxp: raise another pr for support `NodePort` natively might be a better way?

arkodg: sure please go ahead @spwangxp

arkodg:
@qicz can you take a first pass ? with the recent refactor want to make sure the logic lives in the right place, TIA

internal/infrastructure/kubernetes/applier/apply.go
qicz, line 1: this file has been deleted at go head

qicz: Please rebase main branch and go on this job

qicz (dismissed):
LGTM, pls approve ci @arkodg

Xunzhuo:
I got this @qicz

qicz (approved):

zirain (approved):
@arkodg for final review.

internal/infrastructure/kubernetes/resource/resource.go
```diff
@@ -31,3 +33,9 @@ func ExpectedServiceSpec(serviceType *egcfgv1a1.ServiceType) corev1.ServiceSpec
 	}
 	return serviceSpec
 }
+
+// CompareSvc Only compare the selector and ports(not include nodePort) in case user have modified for some scene.
+func CompareSvc(currentSvc, originalSvc *corev1.Service) bool {
+	return cmp.Equal(currentSvc.Spec.Selector, originalSvc.Spec.Selector) &&
```
arkodg, line 39: what about service Type, since we support configuring it, should we also ensure it is not being changed

spwangxp: yes, i'll made svc'type compare if my [question](https://github.com/envoyproxy/gateway/issues/1373) solved.

qicz: #1373 I will fix it ASAP

arkodg (dismissed):

qicz (dismissed):
LGTM, thanks

spwangxp:
> /hold can we compare the entire `Spec` here, and ignore any fields (`NodePort`) . This approach will be easier to maintain in the future
> 
> I've raised two issues to track your use case @spwangxp
> 
> * Support NodePort [Support NodePort Type Envoy Proxy service #1378](https://github.com/envoyproxy/gateway/issues/1378)
> * Support user provisioned envoy fleet [Support User Provisioned Envoy Proxy fleet #1379](https://github.com/envoyproxy/gateway/issues/1379)

1, actual `gw` just need ensure the field unchanged we set before.
2, field should exclude more than need compare, compare entire spec is harder to maintain.

arkodg:
> > /hold can we compare the entire `Spec` here, and ignore any fields (`NodePort`) . This approach will be easier to maintain in the future
> > I've raised two issues to track your use case @spwangxp
> > 
> > * Support NodePort [Support NodePort Type Envoy Proxy service #1378](https://github.com/envoyproxy/gateway/issues/1378)
> > * Support user provisioned envoy fleet [Support User Provisioned Envoy Proxy fleet #1379](https://github.com/envoyproxy/gateway/issues/1379)
> 
> 1, actual `gw` just need ensure the field unchanged we set before. 2, field should exclude more than need compare, compare entire spec is harder to maintain.

every time an API change is made to the `KubernetesServiceSpec`, the reviewer/maintainer will need to ensure that the comparison field is added in the infra layer which is extra cognitive load for the reviewer, else it will cause a bug
Moving the compare to `Spec` is simpler to maintain long term

spwangxp:
> > > /hold can we compare the entire `Spec` here, and ignore any fields (`NodePort`) . This approach will be easier to maintain in the future
> > > I've raised two issues to track your use case @spwangxp
> > > 
> > > * Support NodePort [Support NodePort Type Envoy Proxy service #1378](https://github.com/envoyproxy/gateway/issues/1378)
> > > * Support user provisioned envoy fleet [Support User Provisioned Envoy Proxy fleet #1379](https://github.com/envoyproxy/gateway/issues/1379)
> > 
> > 
> > 1, actual `gw` just need ensure the field unchanged we set before. 2, field should exclude more than need compare, compare entire spec is harder to maintain.
> 
> every time an API change is made to the `KubernetesServiceSpec`, the reviewer/maintainer will need to ensure that the comparison field is added in the infra layer which is extra cognitive load for the reviewer, else it will cause a bug Moving the compare to `Spec` is simpler to maintain long term

check and remove ignored field should de done when `KubernetesServiceSpec` changed, so there is no difference.
or u hava a ignore field list already？

arkodg:
The ignore list shouldnt change in the codebase, but should only only increase whenever upstream K8s API adds a dynamic field into the `Spec`, and shouldnt change whenever a contributor updates the internal `KubernetesServiceSpec` struct in this project

spwangxp:
1, add some comment on `KubernetesServiceSpec` to notice that.
2, pls help check the below list, will use cmp.Equal(newSvc.Spec, oldSvc.Spec, cmpopts.IgnoreFields(ignore list)) to compare two svc.
ignore list: 
- Ports.NodePort
- ClusterIP
- ClusterIPs
- ExternalIPs
- SessionAffinity
- LoadBalancerIP
- LoadBalancerSourceRanges
- ExternalTrafficPolicy
- HealthCheckNodePort
- SessionAffinityConfig
- IPFamilies
- IPFamilyPolicy
- AllocateLoadBalancerNodePorts
- LoadBalancerClass
- InternalTrafficPolicy

check list:
- Ports
- Selector
- Type
- ExternalName
- PublishNotReadyAddresses

> The ignore list shouldnt change in the codebase, but should only only increase whenever upstream K8s API adds a dynamic field into the `Spec`, and shouldnt change whenever a contributor updates the internal `KubernetesServiceSpec` struct in this project

arkodg:
afaik we should only ignore
```
Ports.NodePort
ClusterIP
ClusterIPs
```
 since they are generated by K8s
 `LoadBalancerIP` has been deprecated, so we can ignore it

AliceProxy:
>/hold can we compare the entire Spec here, and ignore any fields (NodePort) . This approach will be easier to maintain in the future

> afaik we should only ignore
> 
> ```
> Ports.NodePort
> ClusterIP
> ClusterIPs
> ```
> 
> since they are generated by K8s `LoadBalancerIP` has been deprecated, so we can ignore it

+1

qicz:
pls @Xunzhuo @zirain @arkodg approve the ci

Xunzhuo:
done

Xunzhuo (dismissed):
/lgtm
/approve

Thanks for working on it @spwangxp

arkodg (approved):
LGTM thanks for considering all the suggestions !

PR #1363, opened 2023-04-25
feat: support envoyproxy & ratelimit deployment container env configu…

qicz:
relates: https://github.com/envoyproxy/gateway/pull/1294#issuecomment-1509435506


internal/infrastructure/kubernetes/ratelimit_deployment_test.go
```diff
@@ -88,6 +90,14 @@ func TestCreateOrUpdateRateLimitDeployment(t *testing.T) {
 				},
 			}
 			require.NoError(t, kube.Client.Get(context.Background(), client.ObjectKeyFromObject(actual), actual))
+
+			sortEnv := func(env []corev1.EnvVar) {
+				sort.Slice(env, func(i, j int) bool {
+					return env[i].Name > env[j].Name
+				})
+			}
+			sortEnv(tc.want.Spec.Template.Spec.Containers[0].Env)
```
arkodg, line 99: env is a list, why do we need to sort

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -5,10 +5,117 @@
 
 package ratelimit
 
-import "fmt"
+import (
+	"fmt"
```
arkodg, line 9: why is this new file created ? another refactor ?

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -5,10 +5,117 @@
...
+		if defaultEnvMapping[envVar.Name] != "" {
+			delete(defaultEnvMapping, envVar.Name)
+		} else {
+			env = append(env, envVar)
+		}
+	}
+
+	// append default env that extension env does not provide.
+	for key, value := range defaultEnvMapping {
+		env = append(env, corev1.EnvVar{Name: key, Value: value})
```
qicz, line 114: @arkodg there use the map so sort env, to make sure the test data is.right

internal/infrastructure/kubernetes/ratelimit/resource.go
qicz: Like proxy with the same structure and logic, the `resource` is the current module public resource logic.

internal/infrastructure/kubernetes/ratelimit/resource.go
qicz: there is a 🐛, i will fix it today

arkodg: cool, will review again after

internal/infrastructure/kubernetes/proxy/resource_provider_test.go
```diff
@@ -128,6 +202,14 @@ func TestDeployment(t *testing.T) {
 			expected, err := loadDeployment(tc.caseName)
 			require.NoError(t, err)
 
+			sortEnv := func(env []corev1.EnvVar) {
```
arkodg, line 205: this shouldnt be needed, since its a list
```
if extensionEnv := deploymentConfig.Container.Env; extensionEnv != nil {
 		env = append(env, extensionEnv...)
 	}
```

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -5,10 +5,120 @@
...
+	}
+
+	return containers
+}
+
+// expectedRateLimitContainerEnv returns expected ratelimit container envs.
+func expectedRateLimitContainerEnv(ratelimit *egcfgv1a1.RateLimit, rateLimitDeployment *egcfgv1a1.KubernetesDeploymentSpec) []corev1.EnvVar {
+	env := []corev1.EnvVar{}
+
+	defaultEnvMapping := map[string]string{
```
arkodg, line 93: the issue is with this logic, I think we should stay away from `maps`, the number of elements here is small,
so the improvement in time complexity is minimal.

I propose this logic
* add default env vars into `env` list
* iterate through user defined env var
    * find and replace if exists
* lets not do anything special like skip `RedisURLEnvVar`, if the user
wants to override it, we can let them, they are on their own

qicz: When the user setting more env than default env, we should merge these env, override default env by the user setting and keep another default env, imo if the user can override default env , the default env is unnecessary in eg and use the user setting env override `ratelimit` default env settings.

internal/infrastructure/kubernetes/proxy/resource_provider_test.go
qicz: current logic using `map`, so needs to sort and make sure test data.

internal/infrastructure/kubernetes/ratelimit/resource_provider_test.go
```diff
@@ -252,6 +324,15 @@ func TestDeployment(t *testing.T) {
 			expected, err := loadDeployment(tc.caseName)
 			require.NoError(t, err)
 
+			sortEnv := func(env []corev1.EnvVar) {
+				sort.Slice(env, func(i, j int) bool {
+					return env[i].Name > env[j].Name
+				})
+			}
+
+			sortEnv(dp.Spec.Template.Spec.Containers[0].Env)
```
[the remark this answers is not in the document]
qicz: ref https://github.com/envoyproxy/gateway/pull/1363#discussion_r1177129324

internal/infrastructure/kubernetes/ratelimit_deployment_test.go
qicz: ref https://github.com/envoyproxy/gateway/pull/1363#discussion_r1177129324

internal/infrastructure/kubernetes/ratelimit/resource.go
arkodg: * so reiterating that using a `map` is not a good idea, ordering within a map is unstable, so every time the map is translated into a list, the order may change causing churn in the K8s API server and will probably also end up restarting the envoy pod
* regarding overriding backend settings - the backend setting is the common case to enable rate limiting, editing the `env` is an edge case which very few users might configure, and if they choose to override ` RedisURLEnvVar`, we should probably let them because that is the user intent. In the future, if we want to add guardrails, we should consider adding validation early on in the pipeline during EG startup

qicz: done, pls review again @arkodg

arkodg (approved):
LGTM, thanks for addressing all the comments ! this feature
should unblock users who have more advanced use cases wrt
the `envoyproxy/ratelimit` service

zirain (approved):

Issue #1292, opened 2023-04-12
Add validations for TLS Certs for a specific listener


arkodg:
*Description*:
Validate TLS Certs for a specific listener and update status
relates to https://github.com/envoyproxy/gateway/pull/1291#pullrequestreview-1380403980

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

cnvergence:
I can take a look at this in the meantime 🙂

arkodg:
thanks @cnvergence suggesting to keep all validation logic for multiple secrets within https://github.com/envoyproxy/gateway/blob/6baaaf1df622a295f93e23c4fa8b15c16cf2f9da/internal/gatewayapi/tls.go#L18 so its easier to maintain

PR #1377, opened 2023-04-28
Add validations for TLS Certs for a specific listener

cnvergence:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/1292


internal/gatewayapi/validate.go
```diff
@@ -332,8 +333,12 @@ func (t *Translator) validateTLSConfiguration(listener *ListenerContext, resourc
 				)
 				break
 			}
+			var hostname string
+			if listener.Hostname != nil {
+				hostname = string(*listener.Hostname)
+			}
 
-			err := validateTLSSecretData(secret)
```
arkodg, line 336: thoughts on moving all the logic into the caller method so this API looks like `err := validateTLSSecretsData(secrets)` ?

arkodg:
the CI failure / conformance test failures seem to be legit

internal/gatewayapi/validate.go
cnvergence: Right, this is a good idea to move them all inside tls.go.

internal/gatewayapi/tls.go
```diff
@@ -54,8 +67,95 @@ func validateTLSSecretData(secret *corev1.Secret) error {
...
+	}
+	return fqdn
+}
+
+// https://golang.google.cn/doc/go1.15#commonname
+func verifyHostname(cert *x509.Certificate, host string) error {
+	lowered := toLowerCaseASCII(host)
+	if len(cert.DNSNames) > 0 {
+		for _, match := range cert.DNSNames {
+			if matchHostnames(toLowerCaseASCII(match), lowered) {
```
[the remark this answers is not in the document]
cnvergence: I don't think so, as this is checking for a hostname in the x509.Certificate, first checking if SAN extension is present, then checking  Subject CommonName field.

I think computeHosts would be valuable if we wanted to get hosts from both listener and routes instead of this: 
https://github.com/envoyproxy/gateway/blob/1ff177fd3bde7ccb31cb559c0350d7cdc84de3da/internal/gatewayapi/validate.go#L340

internal/gatewayapi/tls.go
```diff
@@ -9,53 +9,169 @@ import (
...
+
+		if domain != "" {
+			err = verifyHostname(cert, domain)
+			if err != nil {
+				return fmt.Errorf("hostname %s does not match Common Name or DNS Names in the certificate %s", domain, corev1.TLSCertKey)
+			}
+		}
+		certFQDN = getFQDN(cert)
+		// Check whether the public key algorithm in the referenced secrets are unique.
+		if certFQDN, ok := pkaSecretSet[publicKeyAlgorithm]; ok {
```
arkodg, line 55: I think it should be possible to have multiple secrets/certs with the same `algo` and different CN/SAN
https://gateway.envoyproxy.io/v0.4.0/user/secure-gateways.html#sni-based-certificate-selection

internal/gatewayapi/tls.go
```diff
@@ -9,53 +9,169 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("unable to parse certificate in %s: %w", corev1.TLSCertKey, err)
+	return parseErr
+}
+
+func getFQDN(cert *x509.Certificate) string {
+	fqdn := ""
+	if len(cert.DNSNames) > 0 {
+		for _, name := range cert.DNSNames {
+			fqdn += name + ","
```
arkodg, line 90: should this be a []string instead ?

internal/gatewayapi/tls.go
```diff
@@ -9,53 +9,169 @@ import (
...
-	case "EC PRIVATE KEY":
-		_, err := x509.ParseECPrivateKey(keyBlock.Bytes)
-		if err != nil {
-			parseErr = fmt.Errorf("unable to parse EC formatted private key in %s", corev1.TLSPrivateKeyKey)
+	}
+
+	return true
+}
+
+// toLowerCaseASCII returns a lower-case version of in. See RFC 6125 6.4.1. We use
```
arkodg, line 147: trying to better understand why a custom fun is needed instead of using `strings.ToLower` ?

internal/gatewayapi/tls.go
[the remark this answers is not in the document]
arkodg: I meant as a replacement for `matchHostnames` , they both seem to doing the same thing, ensuring that the list of strings in ARG1, must be a subset of ARG2

internal/gatewayapi/tls.go
cnvergence: This will catch any secrets with the same algo and the same CN/SAN

internal/gatewayapi/tls.go
cnvergence: so basically I wanted to use this [function](https://pkg.go.dev/crypto/x509#Certificate.VerifyHostname) to validate hostnames in x509.Certificate.
But since Go 1.9 CommonName is deprecated for validation in the x509.Certificate
https://golang.google.cn/doc/go1.15#commonname

I wanted to restore this behavior to support certificates without SAN and wanted to copy all logic from crypto/x509.
But I would also be okay with simplifying this function only to our needs.

cnvergence:
> the CI failure / conformance test failures seem to be legit

I need to fix this PR, it seems that the wildcard certificate is still breaking it

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/suite.go#L158

internal/gatewayapi/tls.go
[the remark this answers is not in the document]
cnvergence: I will check it then

internal/gatewayapi/tls.go
arkodg: but you're only keying in the `algo` and not checking the CN/SAN
you probably want to do 
`if gotCertFQDN, ok := pkaSecretSet[publicKeyAlgorithm]; ok && certFQDN == gotCertFQDN {`
even this logic can be improved if `fqdn` is a []string
so it looks more like `if gotCertFQDN, ok := pkaSecretSet[publicKeyAlgorithm]; ok && len(intersection(certFQDNs,gotCertFQDNs)) > 0 {`

internal/gatewayapi/tls.go
arkodg: ah either works, rewriting to use `strings.Lower` or adding a link to the orig code and updating the file with the copyright

internal/gatewayapi/tls.go
cnvergence: Ah, you are right

internal/gatewayapi/tls.go
cnvergence: I have added the comment already, but will also link orginal code to make it more visible

internal/gatewayapi/validate.go
```diff
@@ -333,7 +335,12 @@ func (t *Translator) validateTLSConfiguration(listener *ListenerContext, resourc
 				break
 			}
 
-			err := validateTLSSecretData(secret)
+			var hostname string
+			if listener.Hostname != nil {
+				hostname = string(*listener.Hostname)
+			}
+
+			err := validateTLSSecretsData(secrets, hostname)
```
arkodg, line 343: @cnvergence you are passing an empty string when `listener.Hostname` is nil , but you ideally want to pass `*`, this might be why CI is failing :)
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Listener

internal/gatewayapi/tls.go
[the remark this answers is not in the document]
cnvergence: Thanks for pointing me to this function :)

internal/gatewayapi/validate.go
cnvergence: It was the issue with the wildcard certificate, it seems that x509.verifyHostname is not working with those certificates.
I have removed usage of it.

internal/gatewayapi/validate.go
```diff
@@ -303,7 +304,7 @@ func (t *Translator) validateTerminateModeAndGetTLSSecrets(listener *ListenerCon
 			break
 		}
 
-		err := validateTLSSecretData(secret)
+		err := validateTLSSecretsData(secrets, listener.Hostname)
```
arkodg, line 307: is this always a non `nil` / empty value ?

internal/gatewayapi/tls.go
```diff
@@ -11,51 +11,100 @@ import (
...
+		if certFQDN, ok := pkaSecretSet[pkaSecretKey]; ok {
+			return fmt.Errorf("secret %s/%s public key algorithm must be unique. Cerificate FQDN %s has a conficting algorithm [%s]",
+				secret.Namespace, secret.Name, certFQDN, publicKeyAlgorithm)
+
 		}
-	case "EC PRIVATE KEY":
-		_, err := x509.ParseECPrivateKey(keyBlock.Bytes)
-		if err != nil {
-			parseErr = fmt.Errorf("unable to parse EC formatted private key in %s", corev1.TLSPrivateKeyKey)
+		pkaSecretSet[pkaSecretKey] = getCertFQDN(cert)
```
arkodg, line 57: can we do without `getCertFQDN` ?
and instead use
```
pkaSecretSet[pkaSecretKey] = matchedFQDN
```

to reduce code, improve coverage ?

cnvergence: Hmm, wanted to keep it for logging purposes, but we can simplify it and use only matchedFQDN

internal/gatewayapi/validate.go
cnvergence: I think computeHosts got covered all of the cases, so it will default to [*]

arkodg (approved):
LGTM, thanks @cnvergence

LanceEa (commented):
looks good, just a couple of small questions/items

internal/gatewayapi/tls.go
```diff
@@ -11,51 +11,88 @@ import (
...
+			return fmt.Errorf("unable to decode pem data in %s", corev1.TLSPrivateKeyKey)
+		}
 
-	switch keyBlock.Type {
-	case "PRIVATE KEY":
-		_, err := x509.ParsePKCS8PrivateKey(keyBlock.Bytes)
+		matchedFQDN, err := verifyHostname(cert, host)
 		if err != nil {
-			parseErr = fmt.Errorf("unable to parse PKCS8 formatted private key in %s", corev1.TLSPrivateKeyKey)
+			return fmt.Errorf("hostname %s does not match Common Name or DNS Names in the certificate %s", string(*host), corev1.TLSCertKey)
```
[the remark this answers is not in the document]
LanceEa: Yes, but with the change to the function signature for `validateTLSSecretsData` there is no guarantee that the error returned is associated with the secret for that listener. 

For example, lets say there are 3 `secrets` passed in and the first secret is bad then this function would always return the error for the first secret. When in theory the 3 secret is the one associated to the line you shared. 

My guess is the reason for the function signature change is so that you can add additional validation for checking for overlapping certs? Can this maybe be done independent of changing this func?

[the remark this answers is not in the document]
cnvergence: Ah, I think you are right
Earlier after review, I did change `validateTLSSecretsData` func just to keep all the logic inside of this function.
Also in between there was added another TLS validation, so it looks a bit different now.
We could extract overlapping certs validation again, WDYT @arkodg ?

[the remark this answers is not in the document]
arkodg: looks like I missed this earlier - `validateTLSSecretsData` needs to be called outside the certificateRefs loop after collecting all `secrets` & you can embed the secret name/namespace within the err string within `validateTLSSecretsData`

arkodg (approved):
LGTM, thanks for addressing the comments

zirain (approved):

internal/gatewayapi/tls.go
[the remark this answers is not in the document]
LanceEa: Sorry I'm on vacation so was slow to re-review. The final changes lgtm and it looks like it was already merged. Thanks 😊

Issue #1367, opened 2023-04-26
Pass xRoute Status object instead of entire xRoute object using watchable


arkodg:
*Description*:
The status for xRoute resources is computed in the `gateway-api` layer and published, and subscribed by the `provider` layer using watchable https://github.com/envoyproxy/gateway/blob/4adbc2a97065bbb997aaf2850e43224bcdf4ffab/internal/message/types.go#L25
This structure can be optimized to only publish the status object such as https://github.com/kubernetes-sigs/gateway-api/blob/964e20fbdfafbcbf410d774419fec13ddaae1304/apis/v1beta1/httproute_types.go#L1015 instead of passing the entire xRoute object.

This should reduce memory use in the system

Relates to https://github.com/envoyproxy/gateway/issues/1365

PR #1384, opened 2023-05-01
fix: Pass Status instead of entire obj in watchable

arkodg:
Fixes https://github.com/envoyproxy/gateway/issues/1367


chauhanshubham (approved):
thanks. lgtm

AliceProxy (approved):
branch needs update but lgtm

PR #1385, opened 2023-05-02
egctl: add label selector for egctl config dump of Pod

chauhanshubham:
While using `egctl config envoy-proxy` CLI, there is an additional step
to fetch the Pod name which can then be used in the egctl command. 
This commit adds support to select a Pod based on label selectors, so that 
it's easier to query for the envoy-proxy config of that Pod with something like

```shell
egctl config envoy-proxy cluster -l gateway.envoyproxy.io/owning-gateway-name=eg -l gateway.envoyproxy.io/owning-gateway-namespace=default
```

where we can use the envoy gateway owner labels to fetch the Pod. This
eliminates the need to get the Pod name specifically.

Right now this commit adds a limitation that the label selectors must select one and only one Pod.
If there are more than one pods selected, there is a requirement to change the output format, 
something like
```
{
  "<pod-name>": {
    "config": {}
  }
}
```
which is still open for discussion, as part of issue https://github.com/envoyproxy/gateway/issues/1069


internal/cmd/egctl/config.go
```diff
@@ -66,21 +70,43 @@ func retrieveConfigDump(args []string, includeEds bool) (*adminv3.ConfigDump, er
...
+	if len(labelSelectors) > 0 {
+		podList, err := c.PodsForSelector(nn.Namespace, labelSelectors...)
+		if err != nil {
+			return nil, fmt.Errorf("get pod %s fail: %w", nn, err)
+		}
+
+		if len(podList.Items) == 0 {
+			return nil, fmt.Errorf("no Pods found for label selectors %+v", labelSelectors)
+		}
+		if len(podList.Items) > 1 {
```
arkodg, line 90: curious why the 1 pod limitation is added ?

arkodg: ah got it, yah a map works, thats what we recently did with `translate`
https://github.com/envoyproxy/gateway/blob/f998eacfaf54e1bfc18bb93db2074121e686ae90/internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.yaml#L2
could be done here or lets raise a issue to this can be tracked 
cc @zirain

chauhanshubham: wouldn't `<namespace>-<name>` be tricky to interpret given that both of them can have `-` in their names?

chauhanshubham: I think we need some spec here, maybe something like this?
```
{
    ns-1: {
        pod-name-1: {....},
        pod-name-2: {....},
    },
}
```

arkodg: we are using `ns-name` across the project as a `key` for config e.g https://github.com/envoyproxy/gateway/blob/f998eacfaf54e1bfc18bb93db2074121e686ae90/internal/gatewayapi/helpers.go#L388
```
 ns-1: {
        pod-name-1: {....},
```
gets tricky because it the definition more Kubernetesy which we should ideally avoid, if we want to support other environments

arkodg: thinking again, something like
```
 ns-1: {
        pod-name-1: {....},
```
is more readable for the end user, which might make sense

AliceProxy: > thinking again, something like
> 
> ```
>  ns-1: {
>         pod-name-1: {....},
> ```
> 
> is more readable for the end user, which might make sense

+1 for this style for visibility

qicz: > thinking again, something like
> 
> ```
>  ns-1: {
>         pod-name-1: {....},
> ```
> 
> is more readable for the end user, which might make sense

+1 for this

chauhanshubham: nice. will output in this format then, and update the PR

qicz (approved):
LGTM, thanks

zirain (approved):

Issue #1368, opened 2023-04-26
Dealing with `host:port` in Host header


arkodg:
*Description*:
We need a way to support the use case when the `Host` header contains a port e.g. `example.com:1234` which is valid according to the [RFC](https://www.rfc-editor.org/rfc/rfc3986#section-3.2). This gets tricky because the Gateway API does not intend to allow 
(more in https://github.com/kubernetes-sigs/gateway-api/issues/1936) specifying `port` in the `Hostnames` field with `Gateway` and `HTTPRoute` .

3 options I can think of as a workaround
1) Strip the port from the `Host` header using a knob in Envoy https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.
This gets tricky because incoming service port and container port might differ so we might need to use `strip_any_host_port` instead of `strip_matching_host_port`.
2) Create another xds route with a match to `hostname:listener-port`
3) Circumvent the issue by not specifying any `Hostnames` in `Gateway` or `HTTPRoute` and instead relying 
on a header match to the `:authority` header
https://github.com/envoyproxy/gateway/pull/1311#issuecomment-1511980948

Relates to https://github.com/envoyproxy/gateway/pull/1311

arkodg:
ptal @envoyproxy/gateway-maintainers

arkodg:
vote for 1. [strip_any_host_port](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto) since its less complex, and easier to maintain, although its a wider match, the user can add an extra `match` to tighten it. e.g.
```
      rules:
        - matches:
              headers:
                - name: ":authority"
                  value: "example.com:1234"
```

AliceProxy:
Prefer setting `strip_any_host_port` automatically whenever we have an `HTTPRoute/etc.` that sets a hostname. CIrcumventing it via the provided workaround will also enable users who may not want to have the port stripped from the header for whatever reason.

jackkleeman:
```
      rules:
        - matches:
              headers:
                - name: ":authority"
                  value: "example.com:1234"
```

This doesn't work actually due to CRD valuation on the `name` field which doesn't allow :

PR #1389, opened 2023-05-03
feat: Strip port from host if user specifies a Hostname

arkodg:
* This allows us to ignore the port section from the Host header when matching incoming requests based
on the Host/Authority header and the Hostname specified by the user

Fixes: https://github.com/envoyproxy/gateway/issues/1368


AliceProxy (approved):
lgtm

zirain (approved):

PR #1392, opened 2023-05-03
allow exposing gateway on nodeport service - status updates

### LINKED ISSUE #1378: Support NodePort Type Envoy Proxy service
*Description*:
Tracks support for `NodePort` type service
https://github.com/envoyproxy/gateway/blob/29995bd22b39b80aa14e31f9a7afaa47807c250a/api/config/v1alpha1/shared_types.go#L108

chauhanshubham:
This PR allows for users to expose Gateways on a service of type NodePort.
If the service of type NodePort, the exposed Gateway gets the addresses of the Nodes, in the Gateway status section.
In any case a Node is created/updated/deleted, the Gateway addresses status must be updated, hence the additional watch and cache logic.

Resolves #1378


chauhanshubham:
I'll add more tests if we're good with the general implementation of this.
I'm not sure whether we should be adding the node-port value to the gateway `.status.addresses`
I believe it makes sense for it to be a part of the status in case of nodeport service.

internal/provider/utils/store.go
```diff
@@ -0,0 +1,66 @@
...
+	externalIP string
+}
+
+type providerStore struct {
+	// nodes holds information required for updating Gateway status with the Node
+	// addresses, in case the Gateway is exposed by a Service of type NodePort.
+	nodes map[string]nodeDetails
+}
+
+var providerStoreInstance *providerStore
```
arkodg, line 25: instead of global, should be part of https://github.com/envoyproxy/gateway/blob/19eed9dc1c0d2ac49b10380f5ac91c2f58e56326/internal/provider/kubernetes/controller.go#L59

internal/provider/utils/store.go
```diff
@@ -0,0 +1,66 @@
...
+	var internalIP, externalIP string
+	for _, addr := range n.Status.Addresses {
+		if addr.Type == corev1.NodeExternalIP {
+			externalIP = addr.Address
+		}
+		if addr.Type == corev1.NodeInternalIP {
+			internalIP = addr.Address
+		}
+	}
+	if externalIP != "" {
```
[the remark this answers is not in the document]
chauhanshubham: yea, specifically - in my kind cluster there was no externalIP on the node obj, only internalIP. So I converted internal to external.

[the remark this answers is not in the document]
arkodg: yeah so if we can convert that assumption - `use internalIP when external is empty, which might be routable from client to the gateway` 
and should be field be called `nodeIP` or just `address` instead ?

internal/status/gateway.go
```diff
@@ -50,6 +50,10 @@ func UpdateGatewayStatusProgrammedCondition(gw *gwapiv1b1.Gateway, svc *corev1.S
 			}
 		}
 
+		if svc.Spec.Type == corev1.ServiceTypeNodePort {
+			addresses = nodeAddresses
```
arkodg, line 54: should we also append the NodePort to this address, and expose it in the status ?

internal/provider/kubernetes/predicates.go
```diff
@@ -288,3 +288,25 @@ func (r gatewayAPIReconciler) findOwningGateway(ctx context.Context, labels map[
...
+	node, ok := obj.(*corev1.Node)
+	if !ok {
+		r.log.Info("unexpected object type, bypassing reconciliation", "object", obj)
+		return false
+	}
+
+	key := types.NamespacedName{Name: node.Name}
+	if err := r.client.Get(ctx, key, node); err != nil {
+		if kerrors.IsNotFound(err) {
+			utils.ProviderStore().RemoveNode(node)
```
arkodg, line 303: so the way to know the obj was deleted, is to perform a `GET` and see if the obj exists or not ?

internal/status/gateway.go
arkodg: looks like we cannot expose the port in a clean way, because there will be a unique node port per listener port

arkodg (dismissed):
LGTM thanks

charts/gateway-helm/templates/generated/rbac/roles.yaml
```diff
@@ -8,6 +8,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
```
arkodg, line 11: `geo-check` is failing because this also need to be part of `rbac.go`

internal/provider/kubernetes/predicates.go
chauhanshubham: yes, we don't get the event type in reconciler.

arkodg (approved):

api/config/v1alpha1/validate.go
```diff
@@ -74,7 +74,9 @@ func validateServiceType(spec *EnvoyProxySpec) []error {
 	var errs []error
 	if spec.Provider.Kubernetes != nil && spec.Provider.Kubernetes.EnvoyService != nil {
 		if serviceType := spec.Provider.Kubernetes.EnvoyService.Type; serviceType != nil {
-			if *serviceType != ServiceTypeLoadBalancer && *serviceType != ServiceTypeClusterIP {
+			if *serviceType != ServiceTypeLoadBalancer &&
```
zirain, line 77: can you simply this lines with a function in a followup

zirain (approved):

api/config/v1alpha1/validate.go
chauhanshubham: sure

Issue #1379, opened 2023-04-28
Support User Provisioned Envoy Proxy fleet


arkodg:
*Description*:
It is hard to support all Kubernetes knobs for Deployment and Service within our  `EnvoyProxy` API to customize EnvoyProxy Fleet resources, so its easier to allow users to create and update the envoy proxy fleet themselves, and use Envoy Gateway only to retrieve Xds Configuration

Here are the following things we need to do if this is something we want to support
* Disable the Infra Manager and support User Provisioned Envoy Proxy fleet
* Provide helm charts to create the Envoy Fleet resources (Service, Deployment)
* Add helm knobs to specify
   * Xds Server URL
   * TLS Credentials
   * Labels for Envoy Deployment and Service

qicz:
IMO, supporting user-provisioned fleet, it is hard to config the XDS server and it would be inconvenient to extend more features. how we can support Wasm?  

the infra manager can manage the envoy proxy easily.

support all Kubernetes knobs for Development and Service is hard but we can keep the frequently-used knobs only.

arkodg:
this is something not ideal from the project perspective since its another layer of customizations/support burden but it seems to be something many users want

qicz:
my project about "Manages Envoy Proxy as a standalone or Kubernetes-based application gateway", I think we should manage the proxy easily. user just needs to manage the gateway, the proxy could auto manage by default setting.

if the user needs other knobs, update `EnvoyProxy`.

from user-end thinking, my company is running `gateway`.

zhaohuabing:
To ensure we're prepared to support additional features in K8s resource spec - such as HPA, VPA, volumes, volumeMounts and more - I suggest we establish a clear design beforehand. 

In my research, I came across how Istio handles Ingress Gateway customization. It's a complex process but could be an excellent reference for our EG implementation. You can find more information on KubernetesResourcesSpec at the following link:

https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

arkodg:
with this approach, some features such as `EnvoyProxy.Spec.Telemetry.Metrics` would need to be caveated since those fields edit the bootstrap file in Envoy https://github.com/envoyproxy/gateway/blob/main/internal/xds/bootstrap/bootstrap.yaml.tpl which EG might not have access to for this BYO case

yongzhang:
[...]

PR #1398, opened 2023-05-04
feat: support envoy & ratelimit pod affinity and tolerations.

### LINKED ISSUE #1396: Support Node Affinity and Tolerations for EnvoyProxy Pods
#### *Description*:

I am trying to run a set of proxy on a restricted node pool. The pod needs `affinity` and `tolerations` set to be correctly scheduled.

I would assume it would look something like this:
``` yaml
kind: EnvoyProxy
metadata:
  name: router-public
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        replicas: 2
        pod:
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                  - matchExpressions:
                      - key: cloud.google.com/gke-nodepool
                        operator: In
                        values:
                          - router-node
          tolerations:
            - effect: NoSchedule
              key: node-type
              operator: Exists
              value: "router"
``` 
Where `affinity` and `tolerations`  are propagated to the pod spec in the deployment.

In general, would it be better to allow any fields under `spec.provider.pod` but exclude anything eg controls? i.e. A user can add anything there and the will be merged to the deployment, but certain fields (such as `containers`, `initContainer`) will not be allowed. 

#### *Relevant Links*:
- https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/
- https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

qicz:
Fixes #1396 

support envoy & ratelimit pod affinity and tolerations.

```yaml
kind: EnvoyProxy
metadata:
  name: test
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        type: LoadBalancer
      envoyDeployment:
        replicas: 2
        container:
          env:
          - name: env_a
            value: env_a_value
          - name: env_b
            value: env_b_name
          image: "envoyproxy/envoy-dev:latest"
          resources:
            requests:
              cpu: 100m
              memory: 512Mi
          securityContext:
            runAsUser: 2000
        pod:
          annotations:
            key1: val1
            key2: val2
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                - matchExpressions:
                  - key: cloud.google.com/gke-nodepool
                    operator: In
                    values:
                    - router-node
          tolerations:
          - effect: NoSchedule
            key: node-type
            operator: Exists
            value: "router"
          securityContext:
            runAsUser: 1000
            runAsGroup: 3000
            fsGroup: 2000
```


qicz:
when #1294 merged, will append more unit tests.

qicz:
another choice about #1379 is that we can provide the API `corev1.PodSpec` for `Deployment`with some default settings, such as certs, and volumes. the user can override the default settings.

api/config/v1alpha1/shared_types.go
```diff
@@ -75,6 +75,21 @@ type KubernetesPodSpec struct {
...
+	// +mapType=atomic
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// If specified, the pod's scheduling constraints.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty" protobuf:"bytes,18,opt,name=affinity"`
+
+	// If specified, the pod's tolerations.
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty" protobuf:"bytes,22,opt,name=tolerations"`
```
arkodg, line 92: same as above, is the protobuf str needed ?

api/config/v1alpha1/shared_types.go
```diff
@@ -75,6 +75,21 @@ type KubernetesPodSpec struct {
 	//
 	// +optional
 	SecurityContext *corev1.PodSecurityContext `json:"securityContext,omitempty"`
+
+	// NodeSelector is a selector which must be true for the pod to fit on a node.
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	// +mapType=atomic
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
arkodg, line 84: can we do without `NodeSelector` and rely on `Affinity` ?

qicz: `corev1.PodSpec` define them as different property, so I think we should keep this.

zhaohuabing:
To ensure we're prepared to support additional features in K8s resource spec - such as HPA, VPA, volumes, volumeMounts and more - I suggest we establish a clear design beforehand.

In my research, I came across how Istio handles Ingress Gateway customization. It's a complex process but could be an excellent reference for our EG implementation. You can find more information on KubernetesResourcesSpec at the following link:

https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

zhaohuabing (commented):
To ensure we're prepared to support additional features in K8s resource spec - such as HPA, VPA, volumes, volumeMounts and more - I suggest we establish a clear design beforehand.

In my research, I came across how Istio handles Ingress Gateway customization. Its configuration is quite complex but could be an excellent reference for our EG implementation. You can find more information on KubernetesResourcesSpec at the following link:

https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

api/config/v1alpha1/shared_types.go
arkodg: reframing, which use case does `NodeSelector` enable that `Affinity` cannot do today ?

arkodg:
> To ensure we're prepared to support additional features in K8s resource spec - such as HPA, VPA, volumes, volumeMounts and more - I suggest we establish a clear design beforehand.
> 
> In my research, I came across how Istio handles Ingress Gateway customization. Its configuration is quite complex but could be an excellent reference for our EG implementation. You can find more information on KubernetesResourcesSpec at the following link:
> 
> https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

hey @zhaohuabing if you are referring to [overlays](https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#K8sObjectOverlay) in Istio, im a -1 on it because as a end user, if found it extremely hard to generate the right config
regarding additional features/fields we can raise them as individual issues and debate about the pros/cons of adding it. Since we dont have a clear owner for the entire feature, we haven't enforced a design doc, just accepting/enabling k8s fields based on use cases

api/config/v1alpha1/shared_types.go
qicz: xref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector

they are similar, but should we keep the same for the k8s API?

zhaohuabing:
> arding additional features/fields we can raise them as individual issues and de

I mean the overall structure of the K8s resource： https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

But I agree that we can add new fields when users ask for them. The current structure of envoyDeployment looks good to me.

zhaohuabing (dismissed):
LGTM

api/config/v1alpha1/shared_types.go
arkodg: I would vote to rm `nodeSelector` from this PR, since afaik the use cases can be solved with `affinity` and even the user has not requested this yet

qicz: cc @AliceProxy @zirain @Xunzhuo

qicz:
ptal @arkodg

internal/infrastructure/kubernetes/ratelimit/testdata/deployments/nodeselector.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
arkodg, line 3: is this file needed ?

zhaohuabing (approved):

internal/infrastructure/kubernetes/ratelimit/testdata/deployments/nodeselector.yaml
qicz: deleted

arkodg (approved):

PR #1408, opened 2023-05-08
Added http best practices to hcm

tanujd11:
**What type of PR is this?**

Feature to set defaults http options in http connection manager

Sets HTTP protocol settings according to the best practices https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge

Fixes part of https://github.com/envoyproxy/gateway/issues/1048

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -106,6 +106,14 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 		// merge adjacent slashes in the path
 		MergeSlashes:                 true,
 		PathWithEscapedSlashesAction: hcmv3.HttpConnectionManager_UNESCAPE_AND_REDIRECT,
+		Http2ProtocolOptions: &corev3.Http2ProtocolOptions{
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
+			InitialStreamWindowSize:     wrapperspb.UInt32(65536),   // 64 KiB
+			InitialConnectionWindowSize: wrapperspb.UInt32(1048576), // 1 MiB
+		},
+		CommonHttpProtocolOptions: &corev3.HttpProtocolOptions{
+			HeadersWithUnderscoresAction: corev3.HttpProtocolOptions_REJECT_REQUEST,
+		},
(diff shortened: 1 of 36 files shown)


internal/xds/translator/listener.go
```diff
@@ -106,6 +106,14 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 		// merge adjacent slashes in the path
 		MergeSlashes:                 true,
 		PathWithEscapedSlashesAction: hcmv3.HttpConnectionManager_UNESCAPE_AND_REDIRECT,
+		Http2ProtocolOptions: &corev3.Http2ProtocolOptions{
```
arkodg, line 109: should this be set in L125 where we do specific things only for `isListener.IsHTTP2` ?

internal/xds/translator/listener.go
```diff
@@ -106,6 +106,14 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
...
+			MaxConcurrentStreams:        wrapperspb.UInt32(100),
```
arkodg, line 110: lets define the magic numbers as a `const` at the top of the file

internal/xds/translator/listener.go
```diff
@@ -106,6 +106,14 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
...
+			InitialStreamWindowSize:     wrapperspb.UInt32(65536),   // 64 KiB
+			InitialConnectionWindowSize: wrapperspb.UInt32(1048576), // 1 MiB
+		},
+		CommonHttpProtocolOptions: &corev3.HttpProtocolOptions{
```
arkodg, line 114: this should probably continue to live here since its common for http 1.1 and http2 clients

internal/xds/translator/listener.go
tanujd11: Both should be fine I guess. This should only work with h2 anyways. But I could move it to the if conditional to prevent this config unnecessarily coming on the listener of non-h2 type.

zirain (approved):

PR #1410, opened 2023-05-09
feat: EnvoyPatchPolicy API

arkodg:


arkodg:
cc @howardjohn @costinm in case you see any resuse/overlap w Istio w Gateway API

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+    - type: "type.googleapis.com/envoy.config.listener.v3.Listener"
+      # The listener name is of the form <GatewayNamespace>-<GatewayName>-<GatewayListenerName>
+      name: default-eg-gw-example
+      operation:
+        op: add
+        path: "/filter_chains/0/filters/0/http_filters/0"
+        value: |
+          name: "envoy.filters.http.ratelimit"
+          typed_config:
+            "@type": "type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit"
```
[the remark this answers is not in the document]
arkodg: yeah we autogenerate and import today in the client side (`egctl`)
with this PR we will need to do it in server side too
https://github.com/envoyproxy/gateway/blob/main/internal/xds/extensions/extensions.gen.go

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+* Offline - Leverage [egctl x translate][] to ensure that the `EnvoyPatchPolicy` can be successfully applied and the desired
+output xDS is created.
+* Runtime - Use the `Status` field within `EnvoyPatchPolicy` to highlight whether the patch was applied successfully or not.
+
+## State of the World
+* Istio - Supports the [EnvoyFilter][] API which allows users to customize the output xDS using patches and proto based merge
+semantics.
+
+## Design Decisions
+* This API will only support a single `targetRef` and can bind to only a `Gateway` resource. This simplifies reasoning of how
```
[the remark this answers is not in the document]
arkodg: ah good point thanks for raising, will add `priority` / ordering into open questions

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Goals
+* Add an API allowing users to modify the generated xDS Configuration
+
+## Non Goals
+* Support multiple patch mechanisims
+
+## Implementation
+`EnvoyPatchPolicy` is a [Direct Policy Attachment][] type API that can be used to extend [Gateway API][]
+Modifications to the generated xDS configuration can be provided as a JSON Patch which is defined in
```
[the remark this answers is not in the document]
arkodg: it won't be each resource, `name` of the resource here is mandatory, so we would need to find (guessing its O(N) based on https://github.com/envoyproxy/gateway/blob/0f710c430e29c741d376f799c101a2f11a058fc4/internal/xds/types/resourceversiontable.go#L16) the  specific resource and marshal + patch + unmarshal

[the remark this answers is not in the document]
arkodg: picked json over proto because it has an RFC and more commonly used in the ecosystem (k8s, kustomize) and REST APIs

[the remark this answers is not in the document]
arkodg: thanks for outlining the CPU cost, we should quantity this performance hit and highlight it in docs

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Overview
+
+This design introduces the `EnvoyPatchPolicy` API allowing users to modify the generated Envoy xDS Configuration
+that Envoy Gateway generates before sending it to Envoy Proxy.
+
+Envoy Gateway allows users to configure networking and security intent using the
+upstream [Gateway API][] as well as implementation specific [Extension APIs][] defined in this project
+to provide a more batteries included experience for application developers.
+* These APIs are an abstracted version of the underlying Envoy xDS API to provide a a better user experience for the application developer, exposing and setting only a subset of the fields for a specific feature, sometimes in a opinionated way (e.g [RateLimit][])
```
Xunzhuo, line 11: nit: repeated `a a`.

docs/latest/design/envoy-patch-policy.md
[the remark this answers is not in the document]
howardjohn: :+1: one minor suggestion: don't import the v2 stuff. One less headache to deal with down the line... and in only sort of works IME

howardjohn:
> cc @howardjohn @costinm in case you see any reuse/overlap w Istio w Gateway API

I don't personally. Patching APIs are impossible to support, I don't think anything like this should be standardized in GW API

docs/latest/design/envoy-patch-policy.md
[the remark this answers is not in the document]
arkodg: thanks,  will rm them

arkodg:
> > cc @howardjohn @costinm in case you see any reuse/overlap w Istio w Gateway API
> 
> I don't personally. Patching APIs are impossible to support, I don't think anything like this should be standardized in GW API

@howardjohn I was referring to a common Gateway Extension API across the projects.
Since Istio already has EnvoyFIlter, that should work along side Gateway APIs, patching in Istio is already covered :) but this API could move the spec into Gateway API nouns

qicz (approved):
looks good to me

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+* Support multiple patch mechanisims
+
+## Implementation
+`EnvoyPatchPolicy` is a [Direct Policy Attachment][] type API that can be used to extend [Gateway API][]
+Modifications to the generated xDS configuration can be provided as a JSON Patch which is defined in
+[RFC 6902][]. This patching mechanism has been adopted in [Kubernetes][] as well as [Kustomize][] to update
+resource objects.
+
+### Example
+Here is an example highlighting how a user can configure global ratelimiting using an external rate limit service using this API.
```
kflynn, line 30: Given that we can already configure rate limiting using the native Envoy Gateway ratelimit API, I'd like to see some other examples:

1. You have three different upstream services, each of which needs to use a specific certificate to originate TLS.
   - The current Gateway API doesn't have a lot of support for upstream TLS origination. There's ongoing work to support it natively, but until that's accepted, this would be a reasonable use for an extension mechanism.
   - Obviously you don't know the routes involved, you just know that you happen to have three upstream services that need certificates.

1. You need to allow the HTTP listener, but not the HTTPS listener, to route traffic to the path prefix `/.well-known/acme-challenge/` over to the `my-acme-server` Service. (Other traffic may arrive on the HTTP listener too, but that traffic should be unaffected.)
   - This is the path used by the HTTP-01 ACME challenge, which always arrives over HTTP and must not be redirected to HTTPS. If a vendor wants to do custom ACME work, this might be a thing they'd need to do.

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+## State of the World
+* Istio - Supports the [EnvoyFilter][] API which allows users to customize the output xDS using patches and proto based merge
+semantics.
+
+## Design Decisions
+* This API will only support a single `targetRef` and can bind to only a `Gateway` resource. This simplifies reasoning of how
+patches will work.
+* This API will always be an experimental API and cannot be graduated into a stable API because Envoy Gateway cannot garuntee
+  * that the naming scheme for the generated resources names will not change across releases
+  * that the underlying Envoy Proxy API will not change across releases
```
kflynn, line 128: This was pretty fascinating to read. I agree with the sentiment and I feel compelled to note that if even the author acknowledges that it must forever remain experimental because it's too fragile to promote to stable, we should all but certainly not implement this in the first place unless there's a **truly** compelling reason to do so. I haven't seen that yet.

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Open Questions
+* Should this be an Opt In API disabled by default ?
```
kflynn, line 131: 100%. One should *definitely* have to explicitly opt in to using fragile, experimental APIs.

kflynn:
@howardjohn writes:
> I don't personally [see any overlap with utility for Istio]. Patching APIs are impossible to support, I don't think anything like this should be standardized in GW API

It's kind of rare for me to unequivocally agree with John (😉), but I do here. I'm extremely concerned with how support for this API would play out in the long term. Emissary used to offer a much more constrained capability to do something similar, and it caused us more than a few problems.

kflynn (changes requested):
GitHub says I'm reviewing rather than just commenting, so, uh, OK. 🙂 I definitely want to see some more example uses, and I think we should probably further discuss in the Envoy Gateway meetings before proceeding.

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+	Status EnvoyPatchPolicyStatus `json:"status"`
+}
+
+// EnvoyPatchPolicySpec defines the desired state of RateLimitFilter.
+// +union
+type EnvoyPatchPolicySpec struct {
+	// Type decides the type of patch.
+	// Valid EnvoyPatchType values are "JSONPatch".
+	//
+	// +unionDiscriminator
```
zirain, line 39: this something like `oneOf` in protobuf?

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+// EnvoyJSONPatchConfig defines the configuration for patching a Envoy xDS Resource
+// using JSONPatch semantic
+type EnvoyJSONPatchConfig struct {
+	// Type is the typed URL of the Envoy xDS Resource
+	Type EnvoyResourceType `json:"type"`
+	// Name is the name of the resource
+	Name string `json:"name"`
+	// Patch defines the JSON Patch Operation
+	Operation JSONPatchOperation `json:"operation"`
```
zhaohuabing, line 72: Does this imply that in order to apply a patch, I need to know the absolute path of a listener/route/cluster from the configuration root?

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// JSONPatchOperation defines the JSON Patch Operation as defined in
+// https://datatracker.ietf.org/doc/html/rfc6902
+type JSONPatchOperation struct {
+	// Op is the type of operation to perform
+	Op string `json:"op"`
+	// Path is the location of the target document/field where the operation will be performed
+	// Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
+	Path string `json:"path"`
```
zhaohuabing, line 97: Does this imply that in order to apply a patch, I need to know the absolute path of a listener/route/cluster from the configuration root?

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+const (
+	// KindEnvoyPatchPolicy is the name of the EnvoyPatchPolicy kind.
+	KindEnvoyPatchPolicy = "EnvoyPatchPolicy"
+)
+
+// +kubebuilder:object:root=true
+
+// EnvoyPatchPolicy allows the user to modify the generated Envoy xDS
+// resources by Envoy Gateway using this patch API
+type EnvoyPatchPolicy struct {
```
zhaohuabing, line 22: Why name it a "Policy" instead of just a "Patch"?

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the desired state of EnvoyPatchPolicy.
+	Spec EnvoyPatchPolicySpec `json:"spec"`
+
+	// Status defines the current status of EnvoyPatchPolicy.
+	Status EnvoyPatchPolicyStatus `json:"status"`
+}
+
+// EnvoyPatchPolicySpec defines the desired state of RateLimitFilter.
```
zhaohuabing, line 33: typo: RateLimitFilter

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -0,0 +1,124 @@
...
+	//
+	// +optional
+	JSONPatches []EnvoyJSONPatchConfig `json:"jsonPatches,omitempty"`
+	// TargetRef is the name of the Gateway API resource this policy
+	// is being attached to.
+	// Currently only attaching to Gateway is supported
+	// This Policy and the TargetRef MUST be in the same namespace
+	// for this Policy to have effect and be applied to the Gateway
+	// TargetRef
+	TargetRef gwapiv1a2.PolicyTargetReference `json:"targetRef"`
```
zhaohuabing, line 51: Can I have a "all" target? There might be scenarios multiple gateway resources need to be apply the same EnvoyFilters.

docs/latest/design/envoy-patch-policy.md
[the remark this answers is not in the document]
zhaohuabing: I guess we can still use customized extensions, right? It's really important because most EG deployments would have some sort of customized extensions which don't exist in the Envoy upstream.

I have done similar things in Istio with “typed_config". Below is an example:

```json
{
  "name":"envoy.filters.network.meta_protocol_proxy",
  "typed_config":{
    "@type":"type.googleapis.com/udpa.type.v1.TypedStruct",
    "type_url":"type.googleapis.com/aeraki.meta_protocol_proxy.v1alpha.MetaProtocolProxy",
     ....
  }
}

Since it's wrapped in a typed_config, Even Istio doesn't know about the MetaProtocolProxy API, it won't complain about it.
```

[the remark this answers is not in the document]
arkodg: raised https://github.com/envoyproxy/gateway/pull/1442 to rm v2 imports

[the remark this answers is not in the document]
arkodg: @zhaohuabing I dont think we'll be able to handle your use case if we purely relied on `protojson` tooling

api/v1alpha1/envoypatchpolicy_types.go
arkodg: prefer if we started off with 1:1 mappings, since policy attachment UX is already perceived as complex, the use case can be satisfied today by creating another identical PolicyAttachment and applying it to another `TargetRef`. In the future we could make `TargetRef` plural, note this definition is taken from upstream

api/v1alpha1/envoypatchpolicy_types.go
arkodg: thanks

api/v1alpha1/envoypatchpolicy_types.go
arkodg: previous reviewers for the RateLimit feature had requested RateLimit to suffix with Filter so we could differentiate Filter and Policy

api/v1alpha1/envoypatchpolicy_types.go
arkodg: yes, they could easily add to the beginning using `0` or end using `-`.If they want to insert somewhere in the middle , they would need to rely on `egctl` to make sure their config works correctly. This experience is identical to `kubectl patch` e.g. https://gateway.envoyproxy.io/v0.4.0/user/secure-gateways.html#tls-certificates. This experience might not be extremely use friendly, but its based on a known RFC/spec and is not meant to be a common case

api/v1alpha1/envoypatchpolicy_types.go
arkodg: yup

docs/latest/design/envoy-patch-policy.md
arkodg: raised https://github.com/envoyproxy/gateway/issues/1468 to track this

docs/latest/design/envoy-patch-policy.md
arkodg: can you share some raw xds Config for the examples ?

docs/latest/design/envoy-patch-policy.md
arkodg: thanks !

qicz (approved):
LGTM

qicz (approved):
LGTM

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,151 @@
...
+* Offline - Leverage [egctl x translate][] to ensure that the `EnvoyPatchPolicy` can be successfully applied and the desired
+output xDS is created.
+* Runtime - Use the `Status` field within `EnvoyPatchPolicy` to highlight whether the patch was applied successfully or not.
+
+## State of the World
+* Istio - Supports the [EnvoyFilter][] API which allows users to customize the output xDS using patches and proto based merge
+semantics.
+
+## Design Decisions
+* This API will only support a single `targetRef` and can bind to only a `Gateway` resource. This simplifies reasoning of how
```
Xunzhuo, line 124: A EnvoyPatchPolicy can be only binded to one Gateway, right ? But can a Gateway be binded with multiple different EnvoyPatchPolicy APIs ?

arkodg: yes, ive added a `priority` field with `PolicyAttachment` so user can optionally order the Policiies that are being bind to a Gateway

docs/latest/design/envoy-patch-policy.md
```diff
@@ -0,0 +1,151 @@
...
+patches will work.
+* This API will always be an experimental API and cannot be graduated into a stable API because Envoy Gateway cannot garuntee
+  * that the naming scheme for the generated resources names will not change across releases
+  * that the underlying Envoy Proxy API will not change across releases
+* This API needs to be explicitly enabled using the [EnvoyGateway][] API
```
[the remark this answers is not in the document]
arkodg: ive raised a separate issue to track this https://github.com/envoyproxy/gateway/issues/1468

docs/latest/design/envoy-patch-policy.md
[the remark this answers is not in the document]
zhaohuabing: It would be a headache for any user who adds their own custom filters to envoy since EG can't recognize the API of these filters.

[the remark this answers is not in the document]
arkodg: I understand your sentiment, and it is something we should support, which can be tracked with a GH issue, for now, it doesnt require any changes to the API, or am I missing something ?

Xunzhuo (approved):
lgtm!

docs/latest/design/envoy-patch-policy.md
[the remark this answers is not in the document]
zhaohuabing: Yes, it's an implementation concern rather than API.

zhaohuabing (approved):
LGTM

zirain (approved):

PR #1411, opened 2023-05-09
feat(api): Support custom control plane cert configuration

### LINKED ISSUE #866: docs: allow users to configure custom certs for control plane auth

dboslee:
This PR adds support for user specified certificates. This allows a user to configure kubernetes secrets to use for control plane communication for both envoy-gateway and the envoy proxy deployment.

**Current State:**

I have only included API changes to get some consensus from the community on what this API will look like. Once there is some agreement I will proceed with adding the entire implementation to this PR.

The main piece of configuration proposed here is the `TLSConfig` which allows users to specify the paths to the tls cert, private key, and ca cert.

This is passed to the xds server via a `XDSServer` field added to the `EnvoyGatewaySpec`. This currently only contains a `TLSConfig` but could be extended in the future to expose additional xDS specific configuration.

It also adds a `KubernetesTLSConfigSecret` to the `EnvoyProxyKubernetesProvider` which wraps a `TLSConfig` with a reference to the secret name.

The xds server does not need a reference to the secret name because that will need to be passed in and added to the envoy-gateway deployment through helm.

TODO:
- [ ] Get concensus on API changes
- [ ] Implement API changes
- [ ] Update helm
- [ ] Update docs

Fixes #866

### DIFF
--- a/api/config/v1alpha1/envoygateway_types.go
+++ b/api/config/v1alpha1/envoygateway_types.go
@@ -42,6 +42,11 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	Provider *EnvoyGatewayProvider `json:"provider,omitempty"`
 
+	// XDSServer defines the desired xds server configuration.
+	//
+	// +optional
+	XDSServer *XDSServer `json:"xdsServer,omitempty"`
+
 	// RateLimit defines the configuration associated with the Rate Limit service
 	// deployed by Envoy Gateway required to implement the Global Rate limiting
 	// functionality. The specific rate limit service used here is the reference
@@ -103,6 +108,14 @@ type EnvoyGatewayKubernetesProvider struct {
 // EnvoyGatewayFileProvider defines configuration for the File provider.
 type EnvoyGatewayFileProvider struct {
 	// TODO: Add config as use cases are better understood.
+}
+
+// XDSServer defines configuration for the xds server.
+type XDSServer struct {
+	// TLSConfig configures the file paths to the tls cert, private key, and ca cert.
+	//
+	// +optional
+	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
 }
 
 // RateLimit defines the configuration associated with the Rate Limit Service
--- a/api/config/v1alpha1/envoyproxy_types.go
+++ b/api/config/v1alpha1/envoyproxy_types.go
@@ -93,6 +93,14 @@ type EnvoyProxyKubernetesProvider struct {
 	//
 	// +optional
 	EnvoyService *KubernetesServiceSpec `json:"envoyService,omitempty"`
+
+	// TLSConfigSecret is a reference to the secret containing the TLS certificate,
+	// certificate authority certificate, and private key used to secure Envoy's
+	// client connection to the control plane. Defaults to {Name: "envoy",
+	// Certificate: "tls.crt", PrivateKey: "tls.key", CACertificate: "ca.crt"}.
+	//
+	// +optional
+	TLSConfigSecret *KubernetesTLSConfigSecret `json:"tlsConfigSecret,omitempty"`
 }
 
 // ProxyLogging defines logging parameters for managed proxies. This type is not
--- a/api/config/v1alpha1/shared_types.go
+++ b/api/config/v1alpha1/shared_types.go
@@ -14,6 +14,14 @@ const (
 	DefaultDeploymentCPUResourceRequests = "100m"
 	// DefaultDeploymentMemoryResourceRequests for deployment memory resource
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
+	// DefaultTLSConfigSecretName is the default secret used for the Envoy deployment tls config.
+	DefaultTLSConfigSecretName = "envoy"
+	// DefaultTLSConfigCertificate is the default path used for a TLSConfig certificate.
+	DefaultTLSConfigCertificate = "tls.crt"
+	// DefaultTLSConfigPrivateKey is the default path used for a TLSConfig private key.
+	DefaultTLSConfigPrivateKey = "tls.key"
+	// DefaultTLSConfigCACertificate is the default path used for a TLSConfig CA cert.
+	DefaultTLSConfigCACertificate = "ca.crt"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
 	DefaultEnvoyProxyImage = "envoyproxy/envoy-dev:latest"
 	// DefaultRateLimitImage is the default image used by ratelimit.
@@ -101,6 +109,24 @@ type KubernetesContainerSpec struct {
 	//
 	// +optional
 	Image *string `json:"image,omitempty"`
+}
+
+// KubernetesTLSConfigSecret is a reference to a secret containing a TLSConfig.
+type KubernetesTLSConfigSecret struct {
+	// Name is the name of the secret.
+	Name string `json:"name"`
+	// TLSConfig contains the file paths to a certificate, private key, and ca cert.
+	TLSConfig `json:",inline"`
+}
+
+// TLSConfig contains the file paths to a certificate, private key, and ca cert.
+type TLSConfig struct {
+	// Certificate is the path key where the certificate is stored.
+	Certificate string `json:"certificate"`
+	// PrivateKey is the path where the private key is stored.
+	PrivateKey string `json:"privateKey"`
+	// CACertificate is the path where CA certificate is stored.
+	CACertificate string `json:"caCertificate"`
 }
 
 // ServiceType string describes ingress methods for a service


api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -93,6 +93,14 @@ type EnvoyProxyKubernetesProvider struct {
...
 	// +optional
 	EnvoyService *KubernetesServiceSpec `json:"envoyService,omitempty"`
+
+	// TLSConfigSecret is a reference to the secret containing the TLS certificate,
+	// certificate authority certificate, and private key used to secure Envoy's
+	// client connection to the control plane. Defaults to {Name: "envoy",
+	// Certificate: "tls.crt", PrivateKey: "tls.key", CACertificate: "ca.crt"}.
+	//
+	// +optional
+	TLSConfigSecret *KubernetesTLSConfigSecret `json:"tlsConfigSecret,omitempty"`
```
dboslee, line 103: I considered putting this in `EnvoyProxyKubernetesProvider.EnvoyDeployment  *KubernetesDeploymentSpec` but this is shared with `EnvoyGatewayKubernetesProvider.RateLimitDeployment` so I wasn't certain if it made sense for both to have the option to configure a `KubernetesTLSConfigSecret`.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -42,6 +42,11 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	Provider *EnvoyGatewayProvider `json:"provider,omitempty"`
 
+	// XDSServer defines the desired xds server configuration.
+	//
+	// +optional
+	XDSServer *XDSServer `json:"xdsServer,omitempty"`
```
arkodg, line 48: prefer `ControlPlaneAuth *ControlPlaneTLSConfig` over `xdsServer` because the ca cert needs to be programed in xds client / envoy proxy, but hoping others chime in here

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -105,6 +110,14 @@ type EnvoyGatewayFileProvider struct {
 	// TODO: Add config as use cases are better understood.
 }
 
+// XDSServer defines configuration for the xds server.
+type XDSServer struct {
+	// TLSConfig configures the file paths to the tls cert, private key, and ca cert.
+	//
+	// +optional
+	TLSConfig *TLSConfig `json:"tlsConfig,omitempty"`
```
arkodg, line 118: recommend looking at existing APIs and reusing them
* https://github.com/envoyproxy/gateway/blob/ea2a688f61451b5cfebf2e23fd6a85efc9d4ea69/api/config/v1alpha1/envoygateway_types.go#L197
* https://github.com/envoyproxy/gateway/pull/1294

The only addition that will be needed here is a new field called `CACertificateRef` which stores the CA Cert as a secret as `tls.crt` in the TLS Type Secret

api/config/v1alpha1/envoygateway_types.go
arkodg: looks like this field needs to exist in `EnvoyGateway` and `EnvoyProxy` since we have certs/keys for server and client
https://github.com/envoyproxy/gateway/blob/0f710c430e29c741d376f799c101a2f11a058fc4/internal/crypto/certgen.go#L141

api/config/v1alpha1/envoyproxy_types.go
arkodg: for the name of the field, I vote for `controlPlaneAuth` or something that keeps it similar to what is defined in `EnvoyGateway`
reg the datatype https://github.com/envoyproxy/gateway/pull/1411#discussion_r1189172867 should be relevant

arkodg:
thanks for submitting this API change @dboslee, added some comments

api/config/v1alpha1/envoygateway_types.go
dboslee: Open to any naming suggestions 😄. But I do want to clarify that the `XDSServer` is for configuring the server side tls config not the client side. The client side would currently be configured through the EnvoyProxy CR under `EnvoyProxyKubernetesProvider.TLSConfigSecret`.

api/config/v1alpha1/envoygateway_types.go
dboslee: Thanks, I will take a look.

api/config/v1alpha1/envoygateway_types.go
arkodg: thinking about this again, rather than adding it in two places, does it make sense to just add it here with two different top level fields, keeping all the control plane auth config in one place ?

arkodg: thoughts on this style
```
controlPlaneAuth:
  envoyGatewayCertRef:
  envoyProxyCertRef:
  caCertRef:
```

api/config/v1alpha1/envoygateway_types.go
dboslee: While I think it does make sense to use the ExtensionTLS (and underlying [SecretObjectReference](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.SecretObjectReference)) for configuring the envoy proxy tls configuration I am not sure it make sense for the envoy gateway tls certs. The envoy gateway certs are mounted as a volume by helm or manually by the user so I am not sure they need to be a reference to a secret but rather the paths they are expected to be mounted at. What do you think?

arkodg: I see what you mean https://github.com/envoyproxy/gateway/blob/d801f1c7c5152ac68e2c013e9d7f3eab13731d11/charts/gateway-helm/templates/envoy-gateway-deployment.yaml#L93
I prefer if we could find creative ways to leverage using and mounting this new secret that the user provided and under the hood we did the right thing (new crt can be accessed via `/certs/tls.crt`)

dboslee: Just to clarify, you are thinking rather than use the volume mount, a user would configure a `SecretObjectReference` that envoy gateway would then watch/get the latest version via the kube api?

arkodg: close ish, still thinking :)
So I like the idea of a user providing certs/creds as a Secret for K8s, but I also like the idea of the internal process using the creds from the file system, which allows EG to be used in non K8s envs easily.
So should we just add a knob in helm charts that disabled the cert job and instead add docs on how users can add creds as a Secret with a specific name (`envoy-gateway` and `envoy`) in the controller namespace ?
That is what [Istio ](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) does today.

dboslee: Got it, that is pretty close to what I was going for originally. So I think I will move the Envoy Proxy `TLSConfig` over to a `TLSExtention`/`SecretObjectReference`. This will let someone using kubernetes configure a secret for envoy gateway to mount to the Envoy Proxy deployment pods for control plane auth.  Then I will keep something similar to the `TLSConfig` for configuring Envoy Gateway's tls certs. This will allow a user to specify the paths to the tls certs without envoy gateway needing to know how they got there.

I don't think the secret name is import, the user can supply the name of the secret to helm. Although we can default to the current (envoy-gateway and envoy) for backwards compatibility or when the user wants to use the certgen job.

arkodg: Can we start off w/o an APi change and just docs, telling the user to create a specific secret with a specific name, and disable cert job, so when envoy gateway starts, the secret exists  and is mounted in ?

arkodg: Hoping @envoyproxy/gateway-maintainers can share their thoughts as well

dboslee: Even simpler I like it. Yep I can do that 😄

arkodg: Hey @dboslee we discussed this in the community call today with @AliceProxy @haq204 @kflynn and we are in agreement reg 
1. disabling the cert gen job using a helm chart knob
2. documenting how a user can bring their own certs by 
  * creating secrets with specific names in the EG namespace
  * installing EG with the cert gen job disabled

dboslee: Great I will put something together.

arkodg: awesome, thanks

arkodg: @dboslee raised https://github.com/envoyproxy/gateway/issues/1471 to highlight that cert-gen should not overwrite certs if they are already present in the namespace, so the helm knob will likely not be needed

dboslee: Got it so sounds like we just need docs on how to use cert manager?

arkodg: yeah thats right

PR #1414, opened 2023-05-10
feat: optimize edsClusterConfig for the association between cluster resources and endpoint resources in envoy's configdump?include_eds interface

tmsnan:
In the `config_dump?inclueds_eds` interface of the envoy of the gateway, there is no association information between the endpoints resource and the cluster resource, which is inconvenient for us to troubleshoot, as shown in the following figure
4231c518e0472eea72e2e81658663b40
After optimization, the picture is as follows:
0f54296543a303df65e78bf33f79c080

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -530,6 +530,7 @@ xds:
           connectTimeout: 10s
           dnsLookupFamily: V4_ONLY
           edsClusterConfig:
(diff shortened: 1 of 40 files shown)


arkodg:
sounds good @tmsnan this PR looks fine to me
recommend running `make test` locally to fix the errors, most of the tests are YAML based and you will need to insert the new cluster field into the YAML testdata

tmsnan:
done

arkodg (approved):
LGTM, thanks for your contribution !

qicz (approved):
lgtm

arkodg:
@zirain looks like RL E2E is flaky

PR #1416, opened 2023-05-11
feat: switch global ratelimit config provider to xds-grpc-sotw based server

### LINKED ISSUE #1254: ratelimit: switch to XdsGrpcSotwProvider
For now, we use [FileProvider](https://github.com/envoyproxy/ratelimit/blob/main/src/provider/file_provider.go#L16) in RLS

After https://github.com/envoyproxy/ratelimit/pull/373, [XdsGrpcSotwProvider](https://github.com/envoyproxy/ratelimit/blob/main/src/provider/xds_grpc_sotw_provider.go#L31) can be used to speed config propagation. 

cc @arkodg

shawnh2:
**What type of PR is this?**

feat(global-ratelimit): switch config provider to xds-grpc-sotw based

Fixes #1254


internal/globalratelimit/runner/runner.go
```diff
@@ -36,62 +57,123 @@ func New(cfg *Config) *Runner {
 // Start starts the infrastructure runner
 func (r *Runner) Start(ctx context.Context) error {
 	r.Logger = r.Logger.WithValues("runner", r.Name())
+
+	// Set up the gRPC server for RateLimit xDS Config.
+	r.grpc = grpc.NewServer()
```
shawnh2, line 62: do we need to enable a TLS connection with the xDS confi server?

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -34,6 +36,17 @@ const (
...
 	InfraGRPCPort = 8081
+	// ConfigType is the configuration loading method for ratelimit.
+	ConfigType = "CONFIG_TYPE"
+	// ConfigGrpcXdsServerURL is the url of ratelimit config xds server.
+	ConfigGrpcXdsServerURL = "CONFIG_GRPC_XDS_SERVER_URL"
+	// ConfigGrpcXdsNodeID is the id of ratelimit node.
+	ConfigGrpcXdsNodeID = "CONFIG_GRPC_XDS_NODE_ID"
+
+	// XdsGrpcSotwConfigServerPort is the listening port of the RateLimit xDS config server.
+	XdsGrpcSotwConfigServerPort = 18001
```
shawnh2, line 47: I haven't expose this port in envoy-gateway service yet. see if is ok to open this port for xds config server? @arkodg

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -122,6 +135,18 @@ func expectedRateLimitContainerEnv(ratelimit *egcfgv1a1.RateLimit, rateLimitDepl
 			Name:  UseStatsdEnvVar,
 			Value: "false",
 		},
+		{
+			Name:  ConfigType,
+			Value: "GRPC_XDS_SOTW",
```
shawnh2, line 140: for now it passes the test using cases in https://gateway.envoyproxy.io/v0.4.0/user/rate-limit.html, but still need to test `ConfigType: FILE` mode out, in case someone is tampering with this variable

internal/globalratelimit/runner/runner.go
arkodg: yes, can be done as a follow up

internal/infrastructure/kubernetes/ratelimit/resource.go
arkodg: sure since 
```
DefaultXdsServerPort = 18000
```

arkodg:
that was quick @shawnh2 ! will take another look when its out of draft

internal/infrastructure/kubernetes/ratelimit/resource.go
shawnh2: okay, sure

internal/infrastructure/kubernetes/ratelimit/resource.go
shawnh2: if so, do we need to prevent someone is changing the `ConfigType` to `FILE`, maybe add some validations?

zirain (commented):
LGTM, defer to @arkodg

internal/globalratelimit/runner/runner.go
```diff
@@ -36,62 +57,123 @@ func New(cfg *Config) *Runner {
...
-		}
+func (r *Runner) serverXdsConfigServer(ctx context.Context) {
+	addr := net.JoinHostPort(XdsGrpcSotwConfigServerAddress, strconv.Itoa(ratelimit.XdsGrpcSotwConfigServerPort))
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		r.Logger.Error(err, "failed to listen on address", "address", addr)
+		return
+	}
+	err = r.grpc.Serve(l)
+	if err != nil {
```
Xunzhuo, line 88:
```
if err := r.grpc.Serve(l); err != nil {}
```

internal/globalratelimit/runner/runner.go
```diff
@@ -36,62 +57,123 @@ func New(cfg *Config) *Runner {
...
-		// Translate to ratelimit infra IR
-		result, err := r.translate(xdsIRs)
-		if err != nil {
-			r.Logger.Error(err, "failed to translate xds ir")
-		} else {
-			if result == nil {
-				r.RateLimitInfraIR.Delete(r.Name())
+			if update.Delete {
+				err := r.addNewSnapshot(ctx, nil)
+				if err != nil {
```
Xunzhuo, line 105:
```
if err := r.addNewSnapshot(ctx, nil) ; err != nil {
}
```

internal/globalratelimit/runner/runner.go
```diff
@@ -36,62 +57,123 @@ func New(cfg *Config) *Runner {
...
+					r.Logger.Error(err, "failed to update the config snapshot")
+				}
 			} else {
-				// Publish ratelimit infra IR
-				r.RateLimitInfraIR.Store(r.Name(), result)
+				// Translate to ratelimit infra IR and ratelimit xDS Config.
+				rlIR, rvt, err := r.translate(update.Value)
+				if err != nil {
+					r.Logger.Error(err, "failed to translate xds ir and config")
+				} else {
```
Xunzhuo, line 113: From line 113 to 131, can you do some opts on them ? Too many if else and make it less readable.

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -182,6 +195,18 @@ func expectedRateLimitContainerEnv(rateLimit *egcfgv1a1.RateLimit, rateLimitDepl
...
+		},
+		{
+			Name:  ConfigGrpcXdsServerURL,
+			Value: net.JoinHostPort(XdsGrpcSotwConfigServerHost, strconv.Itoa(XdsGrpcSotwConfigServerPort)),
```
Xunzhuo, line 204: Do we need to check if `XdsGrpcSotwConfigServerPort` is a valid port ?

Xunzhuo (commented):
Some nits, generally LGTM.

internal/globalratelimit/runner/runner.go
shawnh2: wrapped config cache updating logic into a method, should be better now.

internal/infrastructure/kubernetes/ratelimit/resource.go
shawnh2: right now the `DefaultXdsServerPort` does not have a check either.

https://github.com/envoyproxy/gateway/blob/89b8b398b7c2bf2caa8452f4c72a6228feb2db30/internal/xds/bootstrap/bootstrap.go#L29-L30

maybe we can make these two ports configureable, and done as a follow-up?

zirain:
@shawnh2 I think this will speed up `e2e` test, can you try to change following configuration, let's see what will happen.
https://github.com/envoyproxy/gateway/blob/89b8b398b7c2bf2caa8452f4c72a6228feb2db30/test/e2e/tests/ratelimit.go#L38

internal/infrastructure/kubernetes/ratelimit/resource.go
arkodg: maybe the user want want to override to `FILE` manually, its fine to ignore that case imo

internal/globalratelimit/runner/runner.go
```diff
@@ -36,62 +57,126 @@ func New(cfg *Config) *Runner {
...
+
+			str, err := translator.GetRateLimitServiceConfigStr(cfg)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to get rate limit config string: %w", err)
 			}
+			c := &ir.RateLimitServiceConfig{
+				Name:   listener.Name,
+				Config: str,
+			}
+			ratelimitInfra.ServiceConfigs = append(ratelimitInfra.ServiceConfigs, c)
```
arkodg, line 149: curious why is the service config needed now that we are using xds ?

arkodg:
PR looks good ! just added some minor comments

internal/globalratelimit/runner/runner.go
shawnh2: because for `ratelimitInfra`, we only use it to update the configMap `envoy-ratelimit`:

https://github.com/envoyproxy/gateway/blob/a9a85281f03cada583f65dca61c111e7c36e10d9/internal/infrastructure/kubernetes/ratelimit/resource_provider.go#L48-L50

that configMap `envoy-ratelimit` is used when we run ratelimit provider in FILE mode.

like the discuss above, someone may want to override the XDS mode to FILE manually. so if people does that, we can still make sure that ratelimit provider is working.

shawnh2:
> @shawnh2 I think this will speed up ratelimit `e2e` test, can you try to change following configuration, let's see what will happen.
> 
> https://github.com/envoyproxy/gateway/blob/89b8b398b7c2bf2caa8452f4c72a6228feb2db30/test/e2e/tests/ratelimit.go#L38

@zirain hi, I shrink the sleep time down to 30s now, to see how it goes.

i notice that there're still some TODO in ratelimit e2e test, i can help with that as follow-up.
https://github.com/envoyproxy/gateway/blob/a9a85281f03cada583f65dca61c111e7c36e10d9/test/e2e/tests/ratelimit.go#L36

internal/globalratelimit/runner/runner.go
arkodg: imo we should remove all the configmap logic, its more maintenance burden  for this project to support multiple config methods. afaik there is no use case to support both, users would like to configure ratelimiting and are not concerned with how ratelimiting is being implemented

shawnh2: agree, I've removed the `FILE` mode.

internal/ir/ratelimit_infra.go
```diff
@@ -8,16 +8,6 @@ package ir
...
-// RateLimitServiceConfig holds the rate limit service configurations
-// defined here https://github.com/envoyproxy/ratelimit#configuration-1
-// +k8s:deepcopy-gen=true
-type RateLimitServiceConfig struct {
-	// Name of the config file.
-	Name string
-	// Config contents saved as a YAML string.
-	Config string
+	// ServiceNames for Rate limit service configuration name.
+	ServiceNames []string
```
arkodg, line 12: can you elaborate why this is needed ?

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -43,11 +43,6 @@ func (r *ResourceRender) Name() string {
 // ConfigMap returns the expected ConfigMap based on the provided infra.
 func (r *ResourceRender) ConfigMap() (*corev1.ConfigMap, error) {
```
arkodg, line 44: can we delete the entire ConfigMap ? dont think its needed anymore

internal/ir/ratelimit_infra.go
shawnh2: since we don't use RateLimitServiceConfig anymore, I rm it.

https://github.com/envoyproxy/gateway/blob/334422166f208a03440fcec1534c8fe4a2aab3ef/internal/ir/ratelimit_infra.go#L18-L23

but we do need to keep RateLimitInfra, 

https://github.com/envoyproxy/gateway/blob/334422166f208a03440fcec1534c8fe4a2aab3ef/internal/ir/ratelimit_infra.go#L10-L13

so when we Subscribe the changes from xdsIR channel, like someone enable the global ratelimit:

https://github.com/envoyproxy/gateway/blob/334422166f208a03440fcec1534c8fe4a2aab3ef/internal/globalratelimit/runner/runner.go#L46

the RateLimitInfraIR should be informed, but how can we tell whether is delete/create? so I use `ServiceNames []string`, it's just derived from `[]*RateLimitServiceConfig` but only contains `name`.

```go
if rlIR == nil {
	r.RateLimitInfraIR.Delete(r.Name())
} else {
	r.RateLimitInfraIR.Store(r.Name(), rlIR)
}
```

and infra Subscribe to the RateLimitInfraIR, so it can create/delete the service and deployment of ratelimit.

https://github.com/envoyproxy/gateway/blob/334422166f208a03440fcec1534c8fe4a2aab3ef/internal/infrastructure/runner/runner.go#L76-L79

arkodg: ah thanks for flagging this, I think more deletions are needed :)
* we need to get rid of https://github.com/envoyproxy/gateway/blob/main/internal/ir/ratelimit_infra.go since there is no useful info being passed around from the rl-service to the infra-manager
* we need to get rid of the message https://github.com/envoyproxy/gateway/blob/554176b8cbbc5ffa627dc7109e1b3d2b59474e99/internal/message/types.go#L72 since its no longer useful
* the subscriber in the infra-manager needs to be deleted
https://github.com/envoyproxy/gateway/blob/554176b8cbbc5ffa627dc7109e1b3d2b59474e99/internal/infrastructure/runner/runner.go#L76 since its no longer useful
* the infra manager should deploy a RL service based on whether rate limit is configured or not
https://github.com/envoyproxy/gateway/blob/554176b8cbbc5ffa627dc7109e1b3d2b59474e99/internal/infrastructure/runner/runner.go#L48
it doesnt need to be dynamic

All these pieces can be handled in this PR or a follow up PR is fine, dont want to block this PR, since its tremendous improves the time for ratelimit configuration

arkodg (approved):

qicz (approved):
great job

zirain (approved):

zirain (approved):

internal/ir/ratelimit_infra.go
shawnh2: sure, the more deletions will be done as follow-up

internal/globalratelimit/runner/runner.go
arkodg: raised https://github.com/envoyproxy/gateway/issues/1454 to track it

internal/ir/ratelimit_infra.go
arkodg: resolved with https://github.com/envoyproxy/gateway/pull/1451

PR #1425, opened 2023-05-16
feat: add spew dump for envoy-gateway config when starting up

Xunzhuo:
**What type of PR is this?**

feat: add spew dump for envoy-gateway config when starting up, as the envoy gateway config grows bigger and more complex.

we should provide a direct way in log to see how the config recieved by eg looks like.

image


qicz (approved):
LGTM

zirain:
could the output format be optimized, json or yaml is more readable.

zirain (approved):

PR #1426, opened 2023-05-16
fix: use better log for no gc

Xunzhuo:
**What type of PR is this?**

fix: use better log for no gc

`failed` always means something wrong happened, like an error, but after deploying eg, and I did not create gatewayClass, I see this log which makes me feel weird.


zirain (approved):

Xunzhuo:
cc @qicz

zirain (approved):

Issue #700, opened 2022-11-04
Observability: Envoy Gateway


arkodg:
*Description*:
This epic tracks the design and implementation for surfacing logs, metrics and traces generated by Envoy Gateway that are relevant to the user managing Envoy Gateway

Xunzhuo:
Sure

jcchavezs:
QQ: what is the current status here?

arkodg:
closing in favor of https://github.com/envoyproxy/gateway/issues/701

arkodg:
hey @Xunzhuo we discussed this issue in the community, would it be possible to prioritize logging over metrics over tracing for control plane telemetry ?

arkodg:
since its unlikely this will make it into v0.5.0, moving this to the v0.6.0 milestone

arkodg:
closing as completed, individual issues have been created to track remaining work

arkodg:
nvm, reopening since we haven't wrapped up `tracing` in EG

Xunzhuo:
close for duplicated with https://github.com/envoyproxy/gateway/issues/701

PR #1429, opened 2023-05-18
feat(observability): add pprof debug support to eg

Xunzhuo:
**What type of PR is this?**

feat(observability): add pprof debug support to eg

Add pprof debug support to eg, this is a part of https://github.com/envoyproxy/gateway/issues/700

PROFILE	| FUNCTION
-- | --
/debug/pprof/allocs | Returns a sampling of all past memory allocations.
/debug/pprof/block | Returns stack traces of goroutines that led to blocking on sychronization primitives.
/debug/pprof/cmdline | Returns the command line that was invoked by the current program.
/debug/pprof/goroutine | Returns stack traces of all current goroutines.
/debug/pprof/heap | Returns a sampling of memory allocations of live objects.
/debug/pprof/mutex | Returns stack traces of goroutines holding contended mutexes.
/debug/pprof/profile | Returns pprof-formatted cpu profile. You can specify the duration using the seconds GET parameter. The default duration is 30 seconds.
/debug/pprof/symbol | Returns the program counters listed in the request.
/debug/pprof/threadcreate | Returns stack traces that led to creation of new OS threads.
/debug/pprof/trace | Returns the execution trace in binary form. You can specify the duration using the seconds GET parameter. The default duration is 1 second.

Part of https://github.com/envoyproxy/gateway/issues/700

image


api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+type EnvoyGatewayObservability struct {
+
+	// Debug defines Debug Server configuration of Envoy Gateway
+	//
+	// +optional
+	Debug *DebugConfig `json:"debug,omitempty"`
+}
+
+// DebugConfig defines the Envoy Gateway Debug Server configuration.
+type DebugConfig struct {
```
zirain, line 333: enable pprof in production env is dangerous
maybe EG should support enable debug but disabled pprof

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	// Port defines the port the debug server is exposed on.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=8899
+	Port int `json:"port,omitempty"`
+	// Host defines the debug server hostname.
+	//
+	// +optional
+	// +kubebuilder:default="0.0.0.0"
```
alpha-baby, line 348: Directly exposing golang pprof to 0.0.0.0 has security risks. Exposing it to 127.0.0.1 by default may be a better option.

api/config/v1alpha1/envoygateway_types.go
qicz: Agree

Xunzhuo:
Hi @AliceProxy, I saw emissary expose pprof in 0.0.0.0 and enable it in default, can you share some ideas about reviews from others?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	Debug *DebugConfig `json:"debug,omitempty"`
+}
+
+// DebugConfig defines the Envoy Gateway Debug Server configuration.
+type DebugConfig struct {
+	// IPNetworkFamily defines the debug server IPNetworkFamily.
+	//
+	// +optional
+	// +kubebuilder:default=tcp
+	IPNetworkFamily string `json:"ipNetworkFamily,omitempty"`
```
arkodg, line 338: imo we should only expose `Port` for now, and internally bind to `tcp:127.0.0.1:<port>`

api/config/v1alpha1/envoygateway_types.go
arkodg: agree this should be opt in as we start

haq204:
> Hi @AliceProxy, I saw emissary expose pprof in 0.0.0.0 and enable it in default, can you share some ideas about reviews from others?

I wasn't around for when that decision was made but my guess is that so it can be easily routable from outside the pod but when in the same network i.e I can run telepresence to connect into the cluster network and hit the debug endpoint from there. If it's on 127.0.0.1 you're only option is `kubectl port-forward`

Also it's fairly normal to bind to 0.0.0.0 in container environments unless there's a reason to keep it private within the pod.

Overall, I take the stance of defaulting to 127.0.0.1 but exposing the `IPNetworkFamily` option for those that do want to expose on 0.0.0.0.

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -42,6 +42,13 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	Provider *EnvoyGatewayProvider `json:"provider,omitempty"`
 
+	// Observability defines the desired observability related abilities.
+	// If unspecified, the Observability is used with default configuration
+	// parameters.
+	//
+	// +optional
+	Observability *EnvoyGatewayObservability `json:"observability,omitempty"`
```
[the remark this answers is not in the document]
zirain: perfer to `admin`

arkodg:
hey @Xunzhuo, @AliceProxy @haq204 @kflynn and I discussed this PR in the community meeting today and here are some suggestions for the PR
* Renaming the top level field to `Admin` instead of `Observability`
* Only exposing the `Port` within `Address` and binding to `127.0.0.1` for now
* Making `debug` opt-in with a boolean flag field.

Xunzhuo:
Do we need to implement things like https://github.com/emissary-ingress/emissary/tree/master/pkg/debug ?

And expose it in eg debug server too ? cc @AliceProxy

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	//
+	// +optional
+	// +kubebuilder:default=false
+	Enable bool
+	// Port defines the port the debug server is exposed on.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=8899
+	Port int `json:"port,omitempty"`
```
arkodg, line 344: imo, `port` should live at the top level under `Admin` , because it is the port of the `Admin` Server. maybe even
a field like `Address` makes sense, in the future we can also can `IP`/`Host` in case we want to expose bind addr
```
admin:
  address:
    port: 8080
 ```

File: api/config/v1alpha1/envoygateway_types.go [hunk 6]
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	Debug *DebugConfig `json:"debug,omitempty"`
+}
+
+// DebugConfig defines the Envoy Gateway Debug Server configuration.
+type DebugConfig struct {
+	// Enable defines if enable the debug server.
+	//
+	// +optional
+	// +kubebuilder:default=false
+	Enable bool
```
arkodg, line 338: thoughts on making this a top level bool field called `debug` so it looks like
```
admin:
  # this field enables the /debug endpoint
  debug: true
```

arkodg:
> Do we need to implement things like https://github.com/emissary-ingress/emissary/tree/master/pkg/debug ?
> 
> And expose it in eg debug server too, if so, this can be a follow-up ? cc @AliceProxy

imo, should probably wait for user asks before adding more debug features

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	Debug bool `json:"debug,omitempty"`
+}
+
+// EnvoyGatewayAdminAddress defines the Envoy Gateway Admin Address configuration.
+type EnvoyGatewayAdminAddress struct {
+	// Port defines the port the admin server is exposed on.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=8899
```
zirain, line 343: is `8899` a magic number?

arkodg (dismissed):
LGTM thanks !

api/config/v1alpha1/helpers.go
```diff
@@ -248,3 +258,13 @@ func (r *EnvoyGatewayProvider) GetEnvoyGatewayKubeProvider() *EnvoyGatewayKubern
 
 	return r.Kubernetes
 }
+
+func DefaultEnvoyGatewayAdmin() *EnvoyGatewayAdmin {
+	return &EnvoyGatewayAdmin{
+		Debug: false,
+		Address: &EnvoyGatewayAdminAddress{
+			Port: 8899,
```
qicz, line 266: IMO should use constant instead of magic

api/config/v1alpha1/envoygateway_types.go
qicz: prefer to use constant

zirain:
we're using squash merge mode, do not force push for better reviews.

api/config/v1alpha1/helpers.go
Xunzhuo: Done

api/config/v1alpha1/envoygateway_types.go
Xunzhuo: Done

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -313,6 +320,35 @@ type ExtensionTLS struct {
...
+	Port int `json:"port,omitempty"`
+	// Host defines the admin server hostname.
+	//
+	// +optional
+	// +kubebuilder:default="127.0.0.1"
+	Host string `json:"host,omitempty"`
```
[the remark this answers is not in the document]
zirain: from std method `net.JoinHostPort`, I think so

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -15,6 +15,10 @@ const (
 	KindEnvoyGateway = "EnvoyGateway"
 	// GatewayControllerName is the name of the GatewayClass controller.
 	GatewayControllerName = "gateway.envoyproxy.io/gatewayclass-controller"
+	// GatewayAdminPort is the port which envoy gateway admin server is listening on.
+	GatewayAdminPort = 8899
```
zirain, line 19: change this to 19000, same as dataplane?

api/config/v1alpha1/envoygateway_types.go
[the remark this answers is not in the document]
arkodg: discussed this in the community meeting today and majority preferred `Host`, resolving this one

api/config/v1alpha1/envoygateway_types.go
arkodg: we discussed this in the community meeting today and majority agreed to reuse the same port number as the admin server used in envoy, which is `19000`, ( one less thing to remember :) )

Xunzhuo: Done

arkodg (approved):
LGTM thanks !

qicz (approved):
LGTM

Issue #1420, opened 2023-05-12
flaky: TestProvider race


zirain:
https://github.com/envoyproxy/gateway/actions/runs/4957252173/jobs/8868696936?pr=1407

Xunzhuo:
close it via #1430

PR #1430, opened 2023-05-19
chore: use patch for finalizer updates on GC

chauhanshubham:
For issue https://github.com/envoyproxy/gateway/issues/1420
in the corresponding failed pipeline: https://github.com/envoyproxy/gateway/actions/runs/4957252173/jobs/8868696936?pr=1407
The error occured when the first Update failed, and was retried later, while trying to remove Finalizer from GatewayClass.
Instead of doing a Get and Update, in loop, this PR updates the logic to use a single Patch call.

### DIFF
--- a/charts/gateway-helm/templates/generated/rbac/roles.yaml
+++ b/charts/gateway-helm/templates/generated/rbac/roles.yaml
@@ -58,8 +58,8 @@ rules:
   verbs:
   - get
   - list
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):
LGTM, defer to @zirain @qicz

zirain (approved):

PR #1442, opened 2023-05-24
rm v2 xds imports

arkodg:
* not needed since EG doesnt need to maintain compatiblity with xds v2


internal/xds/extensions/extensions.gen.go
```diff
@@ -34,121 +34,28 @@ import (
...
 	_ "github.com/envoyproxy/go-control-plane/envoy/config/overload/v3"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/ratelimit/v2"
 	_ "github.com/envoyproxy/go-control-plane/envoy/config/ratelimit/v3"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/rbac/v2"
 	_ "github.com/envoyproxy/go-control-plane/envoy/config/rbac/v3"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/resource_monitor/fixed_heap/v2alpha"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/resource_monitor/injected_resource/v2alpha"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/retry/omit_canary_hosts/v2"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/retry/omit_host_metadata/v2"
-	_ "github.com/envoyproxy/go-control-plane/envoy/config/retry/previous_hosts/v2"
```
zirain, line 134: this lines seem not that right

arkodg: can you share why

zirain: I'm thinking if there's a v3 api, it's safe to remove v2 version.

zirain: ignore this, it's ok now, let's add them back if there's ture problem

arkodg: yeah I think its moved to another place
https://github.com/envoyproxy/go-control-plane/blob/main/envoy/extensions/retry/priority/previous_priorities/v3/previous_priorities_config.pb.go

zirain (approved):

Xunzhuo (approved):
/lgtm

PR #1445, opened 2023-05-24
fix: update gateway api crds in testdata for v0.7.0

arkodg:


zirain (approved):
/lgtm

should make these as part of in `make generate`?

arkodg:
> /lgtm
> 
> should make these as part of in `make generate`?

good point

Xunzhuo (approved):

PR #1451, opened 2023-05-26
follow-up: removal of the ratelimit ir

shawnh2:
**What type of PR is this?**

this PR will remove all logic related to ratelimit ir.

we don't need it anymore since the ratelimit have been switched to XDS mode, and the ratelimit can be created by config statically.

related to #1416


internal/cmd/server.go
```diff
@@ -175,15 +173,19 @@ func setupRunners(cfg *config.Server) error {
...
 
-	// Start the global rateLimit runner if it has been enabled through the config
+	// Start the global rateLimit if it has been enabled through the config
 	if cfg.EnvoyGateway.RateLimit != nil {
-		// Start the Global RateLimit Runner
-		// It subscribes to the xds Resources and translates it to Envoy Ratelimit Service
-		// infrastructure and configuration.
+		// Create the global ratelimit infra statically.
+		mgr := infraRunner.Manager()
+		if err := mgr.CreateOrUpdateRateLimitInfra(ctx); err != nil {
```
qicz, line 180: prefer to  move to `runner`

arkodg: +1 , prefer if this lives inside the infra-manager runner
Also you can block on `<-ctx.Done()` in https://github.com/envoyproxy/gateway/blob/56edabaac958d34254028b8fb9f0f68ed3c1ac9d/internal/infrastructure/runner/runner.go#L51
and delete the RL service if it was created before

shawnh2: sounds really great

qicz (dismissed):
LGTM

arkodg (approved):
LGTM, thanks for the clean up

PR #1453, opened 2023-05-26
feat: support custom log level for per runner components. 

### LINKED ISSUE #1452: EG logging supports custom log level
*Description*:
> current log uses the default `zap.NewDevelopmentConfig()` and can not customize the level and development mode.

https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/internal/log/log.go#L14-L20

I think we should provide this config like

``` go 
func NewLogger(level string) (logr.Logger, error)
```

and level from the EG config.

qicz:
Fixes #1452

### DIFF
--- a/api/config/v1alpha1/envoygateway_types.go
+++ b/api/config/v1alpha1/envoygateway_types.go
@@ -79,7 +79,7 @@ type EnvoyGatewayLogComponent string
 
 const (
 	// LogComponentGatewayDefault defines the "default"-wide logging component. When specified,
-	// all other logging components are ignored except "gateway-api" which is set to "warn".
+	// all other logging components are ignored.
 	LogComponentGatewayDefault EnvoyGatewayLogComponent = "default"
 
 	// LogComponentProviderRunner defines the "provider" runner component.
--- a/api/config/v1alpha1/helpers.go
+++ b/api/config/v1alpha1/helpers.go
@@ -63,12 +63,8 @@ func DefaultEnvoyGatewayLogging() *EnvoyGatewayLogging {
 }
 
 // DefaultEnvoyGatewayLoggingLevel returns a new EnvoyGatewayLogging with default configuration parameters.
-// When v1alpha1.LogComponentGatewayDefault specified, all other logging components are ignored except "gateway-api" which is set to "warn".
-func DefaultEnvoyGatewayLoggingLevel(component string, level LogLevel) LogLevel {
-	if string(LogComponentGatewayApiRunner) == component && level == LogLevelWarn {
-		return LogLevelWarn
-	}
-
+// When v1alpha1.LogComponentGatewayDefault specified, all other logging components are ignored.
+func DefaultEnvoyGatewayLoggingLevel(level LogLevel) LogLevel {
 	if level != "" {
 		return level
 	}
--- a/api/config/v1alpha1/validate_test.go
+++ b/api/config/v1alpha1/validate_test.go
@@ -300,7 +300,7 @@ func TestDefaultEnvoyGatewayLoggingLevel(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := DefaultEnvoyGatewayLoggingLevel(tt.args.component, tt.args.level); got != tt.want {
+			if got := DefaultEnvoyGatewayLoggingLevel(tt.args.level); got != tt.want {
 				t.Errorf("defaultLevel() = %v, want %v", got, tt.want)
 			}
 		})
--- a/internal/logging/log.go
+++ b/internal/logging/log.go
@@ -50,7 +50,7 @@ func DefaultLogger(level v1alpha1.LogLevel) Logger {
 // more information).
 func (l Logger) WithName(name string) Logger {
 	logLevel := l.logging.Level[v1alpha1.EnvoyGatewayLogComponent(name)]
-	logger := initZapLogger(v1alpha1.DefaultEnvoyGatewayLoggingLevel(name, logLevel))
+	logger := initZapLogger(logLevel)
 
 	return Logger{
 		Logger:        zapr.NewLogger(logger).WithName(name),
@@ -89,7 +89,7 @@ func (l Logger) Sugar() *zap.SugaredLogger {
 }
 
 func initZapLogger(level v1alpha1.LogLevel) *zap.Logger {
-	parseLevel, _ := zapcore.ParseLevel(string(level))
+	parseLevel, _ := zapcore.ParseLevel(string(v1alpha1.DefaultEnvoyGatewayLoggingLevel(level)))
 	core := zapcore.NewCore(zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig()), zapcore.AddSync(os.Stdout), zap.NewAtomicLevelAt(parseLevel))
 
 	return zap.New(core, zap.AddCaller())


arkodg:
@qicz I think we need an API change in [EnvoyGateway ](https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/api/config/v1alpha1/envoygateway_types.go#L31)to be able to configure https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/internal/envoygateway/config/config.go#L42
There is already an API in [EnvoyProxy](https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/api/config/v1alpha1/envoyproxy_types.go#L45) to change the logging level in Envoy, is this not working ?

qicz:
> @qicz I think we need an API change in [EnvoyGateway ](https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/api/config/v1alpha1/envoygateway_types.go#L31)to be able to configure
> 
> https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/internal/envoygateway/config/config.go#L42
> 
> 
> There is already an API in [EnvoyProxy](https://github.com/envoyproxy/gateway/blob/1b565db339c474f5aa87846e7622fcade86592ae/api/config/v1alpha1/envoyproxy_types.go#L45) to change the logging level in Envoy, is this not working ?

yep, but the proxylogging doesn't use in proxy and imo EG logging is different. 

zap log provides development mode and level config. We can move to EnvoyGateway config API, but should use different logging with development mode and level.

arkodg:
* yeah +1 to adding a `Log`/`Logging` field within `EnvoyGateway` API - we would need to define different components - here are some thoughts, creating log components per runner - `all`, `provider`, `gateway-api`, `xds-translator`, `xds-server`, `infrastructure`
* Do we need to add a `development` knob ?  is it something that can be programmatically done in code to improve debugging, rather than present it as an API option ?

qicz:
> * yeah +1 to adding a `Log`/`Logging` field within `EnvoyGateway` API - we would need to define different components - here are some thoughts, creating log components per runner - `all`, `provider`, `gateway-api`, `xds-translator`, `xds-server`, `infrastructure`
> * Do we need to add a `development` knob ?  is it something that can be programmatically done in code to improve debugging, rather than present it as an API option ?

currently, there is one zap log instance in EG, I cannot get your means why we needs different components? I think we just config this single log instance 

`development` works for zap with different encode  `console` for development or `json` for production. I think we need this knob and using the zap recommend config and extends some.

internal/cmd/server.go
```diff
@@ -38,6 +42,10 @@ func getServerCommand() *cobra.Command {
 	}
 	cmd.PersistentFlags().StringVarP(&cfgPath, "config-path", "c", "",
 		"The path to the configuration file.")
+	cmd.PersistentFlags().BoolVarP(&development, "development", "d", true,
```
zirain, line 45: this looks a bit abstract, maybe `log-development-mode` is better?
and you should mention it's somehting form zap log in description.

qicz:
ptal @arkodg

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -57,6 +63,47 @@ type EnvoyGatewaySpec struct {
...
+type EnvoyGatewayLogging struct {
+	// Level is the logging level. If unspecified, defaults to "Info".
+	//
+	// +kubebuilder:default={system: info}
+	Level map[EnvoyGatewayLogComponent]LogLevel `json:"level,omitempty"`
+
+	// DevelopmentMode the log development mode or not for EnvoyGateway running, If unspecified, defaults to true.
+	//
+	// +kubebuilder:default=true
+	DevelopmentMode bool `json:"developmentMode,omitempty"`
```
arkodg, line 76: why is development mode needed, why not use `debug` level logging instead ?

qicz: https://github.com/envoyproxy/gateway/blob/6dc6e1da07596f1279835060af626e8ce92b47d0/internal/logging/log.go#L66-L77

we can use different zap config

arkodg: how is this helping the user

qicz: user can use different log encoder

```go
// NewProductionEncoderConfig returns an opinionated EncoderConfig for
// production environments.
func NewProductionEncoderConfig() zapcore.EncoderConfig {
	return zapcore.EncoderConfig{
		TimeKey:        "ts",
		LevelKey:       "level",
		NameKey:        "logger",
		CallerKey:      "caller",
		FunctionKey:    zapcore.OmitKey,
		MessageKey:     "msg",
		StacktraceKey:  "stacktrace",
		LineEnding:     zapcore.DefaultLineEnding,
		EncodeLevel:    zapcore.LowercaseLevelEncoder,
		EncodeTime:     zapcore.EpochTimeEncoder,
		EncodeDuration: zapcore.SecondsDurationEncoder,
		EncodeCaller:   zapcore.ShortCallerEncoder,
	}
}
```

```go
// NewDevelopmentEncoderConfig returns an opinionated EncoderConfig for
// development environments.
func NewDevelopmentEncoderConfig() zapcore.EncoderConfig {
	return zapcore.EncoderConfig{
		// Keys can be anything except the empty string.
		TimeKey:        "T",
		LevelKey:       "L",
		NameKey:        "N",
		CallerKey:      "C",
		FunctionKey:    zapcore.OmitKey,
		MessageKey:     "M",
		StacktraceKey:  "S",
		LineEnding:     zapcore.DefaultLineEnding,
		EncodeLevel:    zapcore.CapitalLevelEncoder,
		EncodeTime:     zapcore.ISO8601TimeEncoder,
		EncodeDuration: zapcore.StringDurationEncoder,
		EncodeCaller:   zapcore.ShortCallerEncoder,
	}
}
```

of course, we can use the NewDevelopmentEncoderConfig directly and remove developmentMode

qicz: I will remove this, and use the development config .

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -57,6 +63,44 @@ type EnvoyGatewaySpec struct {
 	Extension *Extension `json:"extension,omitempty"`
 }
 
+// EnvoyGatewayLogging defines logging for Envoy Gateway.
+type EnvoyGatewayLogging struct {
+	// Level is the logging level. If unspecified, defaults to "Info".
+	// EnvoyGatewayLogComponent options: system/provider/gateway-api/xds-translator/xds-server/infrastructure/global-ratelimit.
+	// LogLevel options: debug/info/error/warn.
+	//
+	// +kubebuilder:default={system: info}
```
arkodg, line 72: thanks for introducing the log API.
here are some questions I had
1. should it be called `all` instead of `system` for enabling all component logging ?
2. instead of `all/system` , should the user simply specify the level such as `info` and we can enable logging for all components ?

qicz: sure, should call `all`, but I think `system` is more representative of all runner in EG.

for all components needs update `log.defaultLevel` when `all/system` present works for it.

qicz: for point 1 did it: 

https://github.com/envoyproxy/gateway/blob/d45112a0d09078b58c4a23cf6bafbfd3dd552097/internal/logging/log.go#L90-L108

arkodg: hey @qicz we discussed this in the meeting today, and the preference was to use the noun `default` to represent default log level for all components. This way user can configure
```
logging:
  level:
    default: debug
    gateway-api: warn
```
to enable debug logging for all components except for `gateway-api` which is set to `warn`

qicz: @arkodg resolved, ptal

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -42,6 +42,12 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	Provider *EnvoyGatewayProvider `json:"provider,omitempty"`
 
+	// Logging defines logging parameters for Envoy Gateway.
+	//
+	// +optional
+	// +kubebuilder:default={default: debug, gateway-api: warn}
```
zirain, line 48: too much noise for debug, prefer to `info`

api/config/v1alpha1/envoygateway_types.go
qicz: > hey @qicz we discussed this in the meeting today, and the preference was to use the noun `default` to represent default log level for all components. This way user can configure
> 
> ```
> logging:
>   level:
>     default: debug
>     gateway-api: warn
> ```
> 
> to enable debug logging for all components except for `gateway-api` which is set to `warn`

@zirain

arkodg: Hey @qicz this was an example illustration :), not meant to be a default

api/config/v1alpha1/envoygateway_types.go
arkodg: +1

qicz: main branch:

https://github.com/envoyproxy/gateway/blob/7f8a4913b78b492c192b85262344dce4f4f64314/internal/log/log.go#L6-L20
image

api/config/v1alpha1/envoygateway_types.go
qicz: ok, the pre-version uses the `info`, and the current main branch version uses the `debug` so change to the `debug`.

I change to info, itis better

zirain (approved):

api/config/v1alpha1/helpers.go
```diff
@@ -51,6 +53,36 @@ func DefaultGateway() *Gateway {
...
+		Level: map[EnvoyGatewayLogComponent]LogLevel{
+			LogComponentGatewayDefault: LogLevelInfo,
+		},
+	}
+}
+
+// DefaultEnvoyGatewayLoggingLevel returns a new EnvoyGatewayLogging with default configuration parameters.
+// When v1alpha1.LogComponentGatewayDefault specified, all other logging components are ignored except "gateway-api" which is set to "warn".
+func DefaultEnvoyGatewayLoggingLevel(component string, level LogLevel) LogLevel {
+	if string(LogComponentGatewayApiRunner) == component && level == LogLevelWarn {
```
arkodg, line 68: the gateway-api logic here need to be removed

internal/xds/cache/snapshotcache.go
```diff
@@ -55,7 +56,7 @@ type snapshotCache struct {
 	streamIDNodeInfo nodeInfoMap
 	snapshotVersion  int64
 	lastSnapshot     snapshotMap
-	log              *LogrWrapper
+	log              *zap.SugaredLogger
```
arkodg, line 59: curious, why ?

internal/log/log.go
```diff
@@ -1,20 +0,0 @@
...
-
-package log
-
-import (
-	"github.com/go-logr/logr"
-	"github.com/go-logr/zapr"
-	"go.uber.org/zap"
-)
-
-func NewLogger() (logr.Logger, error) {
```
arkodg, line 14: why delete this ?

api/config/v1alpha1/helpers.go
qicz: for your meeting feedback

- to enable debug logging for all components except for gateway-api which is set to warn

so, change to use `default` enable for all?

arkodg: Yah

internal/xds/cache/snapshotcache.go
qicz: LogrWrapper and zap.SugaredLogger have same api like `Debugf` ,`Infof` etc.

We.can use `zap.SugaredLogger` direct ly

internal/log/log.go
qicz: instead of another `NewLogger` with level that the user setted.

api/config/v1alpha1/helpers.go
qicz: change it?

arkodg (approved):
LGTM, thanks for adding this to the API

zirain (approved):

Issue #1370, opened 2023-04-27
Pass xRoute Spec object instead of entire xRoute object using watchable 


arkodg:
*Description*:
The xRoutes are passed from the `provider` to the `gateway-api` layer using the watchable messaging system.
https://github.com/envoyproxy/gateway/blob/a5d3f36227b4c48f9dd70a60b213bd5278991e1f/internal/message/types.go#L23
Instead of passing the [entire resources](https://github.com/envoyproxy/gateway/blob/a5d3f36227b4c48f9dd70a60b213bd5278991e1f/internal/gatewayapi/resource.go#L25) we can only pass the [Spec](https://github.com/kubernetes-sigs/gateway-api/blob/16c54cdd2546bce917c1d045a5dc6a49e51625ae/apis/v1beta1/httproute_types.go#L56) which is the required for translation.

This should reduce memory consumption as well as improve caching hits, since any field related to `time` that is part of the meta resource or status will not affect the object comparison and will reduce object churn in the system

Relates to https://github.com/envoyproxy/gateway/issues/1365

haq204:
We  use `metadata.name` on some resources for xDS names

arkodg:
> We use `metadata.name` on some resources for xDS names https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/route.go#L23

good point, we'll then need to find to only share the Namespace, Name and Spec

PR #1455, opened 2023-05-26
Discard Status before publishing in provider

arkodg:
* Discard the Status of the Gateway and xRoute resources before saving them in the watchable layer. This reduces the memory consumption within watchable / lesser things to cache. Its safe to discard the status since it is always recomputed by the gateway-api layer, and only written back into the K8s API Server if something, other than the `LastTransitionTime` has changed.

Fixes: https://github.com/envoyproxy/gateway/issues/1370


zirain (approved):
/lgtm

it would be better if there's a benchmark test.

chauhanshubham (approved):
thanks

haq204 (approved):

Issue #1035, opened 2023-02-11
Add support deployment on bare metal or VM


lizj3624:
*Description*:
I hope gateway support deployment on bare metal or VM

arkodg:
+1 this can be supported using a [File based Provider](https://github.com/envoyproxy/gateway/issues/1001) or a Docker/Container Provider

lizj3624:
+1

arkodg:
@tommie would running EnvoyProxy as a docker container satisfy your use case ?

tommie:
Preferably not. I'm using docker-compose and a cron'd update now, but my (wet) dream is to have a proper orchestrator where I can stop treating my two servers as precious. Trying to build a bottlerocket+kubeadm+flannel+envoyproxy(?)+... solution that can go from bootstrapping on a single cheap node, while also being easy to scale.

I could do something special for ingress, but since K8s is supposed to abstract that, I'd love to avoid it. And since I'm just starting to play with K8s, using Gateway seems preferable over Ingress, muddying the waters with LBs, IIUC.

When I tried patching in `externalIPs` into the Gateway's Service, that seemed to work until it was "healed", so I think I'm in favor of #360. It requires me to maintain a list of ingress machines, and I'd probably want to abstract that with node labels, but that's much nicer than having to run stuff outside of K8s.

Since Hetzner's cheap servers don't give me a controllable SDN, I don't think MetalLB will work for me (based on a casual investation.) I would basically just move the `externalIPs` assignment problem one layer down the stack.

tommie:
Follow-up on #1322, and looking for input on next steps.

The generated gateway services set `externalTrafficPolicy: Local`, with a comment saying it's to preserve the client address. However, this causes Kubernetes to generate iptables rules like

    -A KUBE-EXTERNAL-SERVICES -d 10.0.2.16/32 -p tcp -m comment --comment "envoy-gateway-system/envoy-default-http-64656661:http has no local endpoints" -m tcp --dport 80 -j DROP

on nodes that don't have such a pod running. Probably from here: https://github.com/kubernetes/kubernetes/blob/52353a2e4ac121ccbfc0fc243e5738888286cc96/pkg/proxy/iptables/proxier.go#L1070

Since I'm trying to run Envoy without an external LB, I think I need a way for Gateway to run a pod on (at least) each node that is referenced in `gateway.spec.addresses`. Currently, it seems replica count is static, and there is no node affinity that could link the assigned external IPs to the replicas.

[...]

PR #1457, opened 2023-05-28
feat: Add EnvoyProxySpec.EnvoyDaemonSet.

tommie:

Allow using a DaemonSet instead of Deployment for EnvoyProxy.

Making EnvoyDeployment and EnvoyDaemonSet mutually exclusive.

Also adds KubernetesPodSpec.NodeSelector.

This is useful when there is no separate load-balancer or floating IPs routing to the cluster. Instead, some nodes are designated as public gateways. This allows for a fully self-contained Kubernetes cluster.

Part of #1035


tommie:
Talking to @arkodg, DaemonSet might be the wrong approach. Though I still think there is a place for having the `nodeSelector`. Being discussed.

tommie:
Pulling the plug on this. Probably not needed. See #1035.

Issue #699, opened 2022-11-04
Observability: Envoy Proxy


arkodg:
*Description*:
This epic tracks the design and implementation for surfacing `logs`, `metrics` and `traces` generated by Envoy Proxy that are relevant to the user personas

arkodg:
related PR for metrics https://github.com/envoyproxy/gateway/pull/581/files

arkodg:
related PR for traces https://github.com/envoyproxy/gateway/pull/697

zirain:
best practices form prometheus: https://prometheus.io/docs/practices/naming/#metric-and-label-naming

zirain:
accesslog: I think should focus on File(send /dev/stdout) and OpenTelemetry(I believe it's the future)
traces: only support OpenTelemetry is an options

danehans:
@Xunzhuo unless you're actively working on this issue, can you unassign yourself and add the help-wanted label?

arkodg:
should we configure envoy so it sends OLTP format to otel-collector and the user configures EG API to specify the  Otel-collector endpoint (user (or docs) brings their otel-collector)
There seems to be support for [access log](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/open_telemetry/v3/logs_service.proto) [and traces](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/opentelemetry.proto.html) and see a [PR for metrics support](https://github.com/envoyproxy/envoy/pull/26620)

arkodg:
are there any existing grafana dashboards we can reuse to showcase metrics ?

zirain:
I made two proposals before:
1. via EnvoyProxy https://github.com/envoyproxy/gateway/pull/697
2. via PolicyAttachment https://github.com/envoyproxy/gateway/pull/1121

so I think this time just add configuration in EnvoyGateway?

cc @arkodg @AliceProxy @kflynn @Xunzhuo

arkodg:
@zirain I vote for defining Data Plane Observability/Telemetry within the `EnvoyProxy` resource .  Your PRs gain should gain more traction and reviews in the coming weeks, now that Observability is the theme for the next release.
Also request to add a design doc highlighting the - why, what and how.
Tackling one signal at a time (access logging first, then the rest) should also speed up the review process

zirain:
ok, I will move forward https://github.com/envoyproxy/gateway/pull/697

[...]

PR #1464, opened 2023-05-30
feat: implement accesslog

zirain:
**What type of PR is this?**

part of https://github.com/envoyproxy/gateway/issues/699

Fixes #


internal/gatewayapi/listener.go
```diff
@@ -144,3 +155,67 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 		}
 	}
 }
+
+func processAccessLog(envoyproxy *configv1a1.EnvoyProxy) *ir.AccessLog {
+	if envoyproxy != nil && len(envoyproxy.Spec.AccessLoggings) > 0 {
+		irAccessLog := &ir.AccessLog{}
+		// translate the access log configuration to the IR
```
arkodg, line 162: no default case ?

zirain: IMO, disabled accesslog by default.
The user should explicitly set accesslog configuration, especially for a Edge gateway.

arkodg: enabling access logs by default especially some format like `field:value`, adds a lot of value on top on Envoy. This does however make the API complex e.g how to disable default

arkodg: thoughts on this approach
* if accessLogs is null , we add a default text(or json) + file based Access log
* if Len(acesslogs) is 0 `accessLogs: []` , we dont add any defaults
* else if the user has set `accessLogs` in the API, we use that, and dont append any default
?

internal/gatewayapi/listener.go
```diff
@@ -11,9 +11,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 
+	configv1a1 "github.com/envoyproxy/gateway/api/config/v1alpha1"
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
+const (
```
arkodg, line 18: imo this code is specific to xds, and should live in the XDS translator, we probably should follow the same style
used in Infra IR where the entire `EnvoyProxy` is passed into the IR and then parsed in the infra layer
https://github.com/envoyproxy/gateway/blob/22bd06b33f914e406bba9ce9690e92cc3f5c3226/internal/ir/infra.go#L36

internal/gatewayapi/listener.go
qicz: +1, I think needs default case.

zirain: for me, it's odd to have configuration like `[]` to disable somehting
```
kind: EnvoyProxy
metadata:
  name: disable-access-logging
spec:
  accessLoggings: []
```

internal/gatewayapi/listener.go
zirain: make sense

internal/xds/translator/accesslog.go
```diff
@@ -6,19 +6,218 @@
...
+					TextFormatSource: &cfgcore.DataSource{
+						Specifier: &cfgcore.DataSource_InlineString{
+							InlineString: format,
+						},
+					},
+				},
+			},
+		}
+
+		accesslogAny, _ := anypb.New(filelog)
```
Xunzhuo, line 76: why omit the error check here?

internal/xds/translator/accesslog.go
```diff
@@ -6,19 +6,218 @@
...
+			AccessLogFormat: &fileaccesslog.FileAccessLog_LogFormat{
+				LogFormat: &cfgcore.SubstitutionFormatString{
+					Format: &cfgcore.SubstitutionFormatString_JsonFormat{
+						JsonFormat: jsonFormat,
+					},
+				},
+			},
+		}
+
+		accesslogAny, _ := anypb.New(filelog)
```
Xunzhuo, line 113: ditto

internal/xds/translator/accesslog.go
```diff
@@ -6,19 +6,218 @@
...
+			al.Body = &otlpcommonv1.AnyValue{
+				Value: &otlpcommonv1.AnyValue_StringValue{
+					StringValue: format,
+				},
+			}
+		}
+
+		al.Attributes = convertToKeyValueList(otel.Attributes)
+
+		accesslogAny, _ := anypb.New(al)
```
Xunzhuo, line 154: ditto

Xunzhuo (commented):
Thanks for working on it. Some nits.

internal/gatewayapi/listener.go
arkodg: we discussed this in the community meeting and all agreed that having defaults is very useful.
But using `[]` is awkward and not an obvious way to disable `accessLogs`, one way to do this would be to create an explicit `disabled` field
```
accessLogs:
  disabled: true
  settings:
  - format: 
    sinks:
```

api/config/v1alpha1/accesslogging_types.go
```diff
@@ -5,32 +5,41 @@
...
+	// +optional
+	Settings []ProxyAccessLogSetting `json:"settings,omitempty"`
+}
+
+type ProxyAccessLogSetting struct {
+	// Format defines the format of accesslog.
+	Format ProxyAccessLogFormat `json:"format"`
+	// Sinks defines the sinks of accesslog.
 	// +kubebuilder:validation:MinItems=1
 	Sinks []ProxyAccessLoggingSink `json:"sinks"`
```
arkodg, line 22: should this be `ProxyAccessLogSink` as well ?

internal/gatewayapi/listener.go
```diff
@@ -144,3 +147,73 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 		}
 	}
 }
+
+func processAccessLog(envoyproxy *configv1a1.EnvoyProxy) *ir.AccessLog {
+	if envoyproxy == nil ||
+		envoyproxy.Spec.Telemetry.AccessLog == nil ||
```
arkodg, line 153:
```suggestion
                envoyproxy.Spec.Telemetry == nil ||
		envoyproxy.Spec.Telemetry.AccessLog == nil ||
		
```

arkodg (dismissed):
LGTM, thanks !

api/config/v1alpha1/accesslogging_types.go
zirain: yes, thanks for catching this

internal/xds/translator/accesslog.go
zirain: addressed, will handle later

arkodg (approved):

qicz (approved):
LGTM

Issue #1463, opened 2023-05-29
Using NodePort leaves proxy service externalIPs empty


tommie:
*Description*:
I'm trying to run Gateway on bare metal with just public-IP nodes.
I.e. no load balancer or floating IPs. Part of #1035.

When creating a Gateway without `spec.addresses` and using NodePort for proxy services, I would expect the rendered service to have `externalIPs` set to the external IPs of the nodes where the pods are currently running. Right now, they are empty, likely because [1] and [2] means they are only set from `GatewaySpec.addresses` (this was recently added, as a start.) This should make`GatewayStatus.addresses` empty, but it's actually populated with *all* node addresses [3]? This leads to ExternalDNS adding nodes to RRs, that have nothing to do with the Gateway.

**Edit:** If `GatewaySpec.addresses` are provided, the IP addresses in `GatewayStatus.addresses` end up with duplicates, because of [3] and then also adding the Service externalIPs. But until I understand [3], I don't know what that function should look like.

In the other direction, for NodePort, it would make sense if setting `GatewaySpec.addresses` caused the proxy pods to be scheduled on nodes that match those external IPs. Or rejecting that configuration.

*Repro steps*:

1. Create an `EnvoyProxy` with `NodePort`.
2. Create a Gateway without `spec.addresses`.
3. Look at the generated Service.

---

1. https://github.com/envoyproxy/gateway/blob/b1a752080adf25f92375f31c3bf8d708250b911c/internal/infrastructure/kubernetes/proxy/resource_provider.go#L100
2. https://github.com/envoyproxy/gateway/blob/b1a752080adf25f92375f31c3bf8d708250b911c/internal/gatewayapi/address.go#L30
3. https://github.com/envoyproxy/gateway/blob/b1a752080adf25f92375f31c3bf8d708250b911c/internal/status/gateway.go#L54

arkodg:
What is happening
* All the Node Addresses (Internal and External) are being appended to `Gateway.Status.Addresses`
* The `Gateway.Spec.Addresses` is also being appended to `Gateway.Status.Addresses`

What should happen
* We cannot tie location of where the Envoy pod is running to the IP that should be used to reach it. Kubernetes allows the Pod to be reachable from any node using iptables magic
* We currently append Internal IP of the node to `Gateway.Status.Addresses` if External IP is not present. It is debatable whether this is a good idea or not, and whether we should only limit it to External IP of the node ?
[...]

PR #1465, opened 2023-05-30
fix: only set gw.Spec.Addresses in gw.Status.Addresses

arkodg:
Dont append, just set/override the gw.Status.Addresses with the values from gw.Spec.Addresses (which eventually get set in svc.Spec.ExternalIPs)

Relates to https://github.com/envoyproxy/gateway/issues/1463

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway


arkodg:
cc @tommie hoping this can solve your use case of explicitly setting the gw.Spec.Addresses with public IPs of specific nodes and using ExternalDNS integration to build out the DNS Records

qicz (approved):

zirain (approved):

PR #1469, opened 2023-05-31
complement for RL e2e test

shawnh2:
**What type of PR is this?**

remove the two TODO lines in RL e2e test.

- as for first TODO, wait for deployment/envoy-gateway to finish restart, so the deployment/envoy-ratelimit can be created, and the `Sleep` can be removed
- as for second TODO, make the resp and req more explicit

### DIFF
--- a/test/e2e/tests/ratelimit.go
+++ b/test/e2e/tests/ratelimit.go
@@ -33,8 +33,7 @@ var RateLimitTest = suite.ConformanceTest{
 			gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
 			gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
 
-			// should just send exactly 4 requests, and expect 429
-			firstThreeExpResp := http.ExpectedResponse{
+			expectOkResp := http.ExpectedResponse{
 				Request: http.Request{
 					Path: "/",
 				},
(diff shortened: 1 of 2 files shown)


shawnh2:
it's wired that the request to gateway may `connect: connection refused` few times and then got `200`

test/e2e/tests/ratelimit.go
```diff
@@ -33,20 +33,48 @@ var RateLimitTest = suite.ConformanceTest{
...
+			}
+			lastFourthReq := http.MakeRequest(t, &lastFourthExpResp, gwAddr, "HTTP", "http")
+			if err := GotNTimesExpectedResponse(t, 1, suite.RoundTripper, lastFourthReq, lastFourthExpResp); err != nil {
+				t.Errorf("fail to get expected response at last fourth request: %v", err)
+			}
 		})
 	},
 }
+
+func GotNTimesExpectedResponse(t *testing.T, n int, r roundtripper.RoundTripper, req roundtripper.Request, resp http.ExpectedResponse) error {
```
arkodg, line 68: instead of doing an exact check of `n` requests, you might need to use some until/eventual  logic and wait for some timeout duration until all `n` requests ended up generating the expected response / status code.
because the e2e test might start sending traffic before the route was configured, etc

tools/make/kube.mk
```diff
@@ -102,12 +102,15 @@ install-ratelimit:
 	@$(LOG_TARGET)
 	kubectl apply -f examples/redis/redis.yaml
 	kubectl rollout restart deployment envoy-gateway -n envoy-gateway-system
+	kubectl rollout status --watch --timeout=5m -n envoy-gateway-system deployment/envoy-gateway
```
arkodg, line 105: why is this needed ?

arkodg: the next line checks for `condition=Available`

shawnh2: because the `rollout restart` does not affect the condition of deployments.

even through its replicaset is not ready yet, the condition of deployments is still `Available`. so the `rollout status` here watches till its replicaset gets ready.

arkodg: thanks for the clarification, seems weird that it doesnt reset the status

test/e2e/tests/ratelimit.go
shawnh2: thanks, right now i wait till get the first 200, and then fire the rest three requests

arkodg: thanks, re-read the logic and it looks good !

arkodg (approved):

zirain (approved):

PR #1473, opened 2023-06-01
docs: fix user/tls-cert-manager Issuer/ClusterIssuer

tommie:
Based on feedback from @irbekrm of cert-manager.


docs/latest/user/tls-cert-manager.md
```diff
@@ -67,7 +67,7 @@ EOF
 
 The `selfSigned` issuer type is namespaced, which means it will use the Issuer resource kind.
 It can only issue certificates for Secrets in the same namespace, here "default."
-We will later see ClusterIssuer used, which can deal with Secrets (and hence Gateways) in any namespace.
+You could also use [ClusterIssuer](https://cert-manager.io/docs/concepts/issuer/), which can be used from any namespace.
```
irbekrm, line 70: +1

Suggestion: also change line 68 to something like `Here we create `Issuer` kind which is a namespaced resource` or something similar.

irbekrm (changes requested):
Thanks for writing the awesome tutorial and thanks for making the changes @tommie !

I've added a nit.

Apart from that I have two optional suggestions:
- we really don't recommend using self-signed issuer for any other case except for quick testing of things and for bootstrapping a root CA cert. In our own example setups we typically use a selfsigned issuer to issue a CA cert for a CA issuer that can then be used to issue certs for applications like [so](https://github.com/cert-manager/csi-driver/blob/main/deploy/example/example-app.yaml#L6-L35). I understand that you may not want to make the setup too complex here, but perhaps would be nice to add an additional warning that the self signed issuer should not be used beyond a quick test of things

- In the garbage collection section it mentions that `Certificate`s will be garbage collected when `Secret`s are removed - this is not quite the case, for gateway-shim/ingress-shim they will be deleted because if the relevant TLS block has been removed from `Gateway` or `Ingress`. If a `Secret` was deleted (i.e by a user) that is referenced by a `Certificate`, cert-manager would actually re-create the `Secret`

docs/latest/user/tls-cert-manager.md
tommie: Maybe I should just change all of them to `ClusterIssuer` and just mention `Issuer` at the end, next to GC.

tommie:
> * we really don't recommend using self-signed issuer for any other case except for quick testing of things and for bootstrapping a root CA cert. In our own example setups we typically use a selfsigned issuer to issue a CA cert for a CA issuer that can then be used to issue certs for applications like [so](https://github.com/cert-manager/csi-driver/blob/main/deploy/example/example-app.yaml#L6-L35). I understand that you may not want to make the setup too complex here, but perhaps would be nice to add an additional warning that the self signed issuer should not be used beyond a quick test of things

Yes, I want something that is simple to start with. selfSigned has no configuration, and doesn't require access to external services, which makes it an ideal thing to start with. Difficult to screw up. I've added a note.

> * In the garbage collection section it mentions that `Certificate`s will be garbage collected when `Secret`s are removed - this is not quite the case, for gateway-shim/ingress-shim they will be deleted because if the relevant TLS block has been removed from `Gateway` or `Ingress`. If a `Secret` was deleted (i.e by a user) that is referenced by a `Certificate`, cert-manager would actually re-create the `Secret`

Nice catch. Updated.

docs/latest/user/tls-cert-manager.md
tommie: Changed.

irbekrm:
Thanks for making the changes! Looks great from cert-manager perspective 😄 

/lgtm

arkodg (approved):
thanks for the updates @tommie and @irbekrm for the review

Issue #1256, opened 2023-04-05
EndpointSlice Support


pgold30:
*Description*:
EndpointSlice is now enabled by default in Kubernetes and offers some solid performance benefits for clusters with large Endpoint sets. It might be worthwhile to support consuming EndpointSlice, at least as optional.

Envoy is designed for scale and the use case for endpointslice also. 

The lack of EndpointSlices implementation is unfortunately impacting production for us now. Since Kubernetes v1.22, Services that exceed 1000 Pods/network endpoints, Endpoints are now being truncated to a maximum of 1000 items. 
Im sure there are many other people with the same issue. (same [here](https://github.com/kubernetes/ingress-nginx/issues/6017#issuecomment-1170285796) in Countour)

Outlining the pieces needed to implement this issue @arkodg  pointed [here](https://github.com/envoyproxy/gateway/issues/35#issuecomment-1387344863) 

Work Items:

- [x]  Add EndpointSlice support in the Kubernetes Provider https://github.com/envoyproxy/gateway/pull/1474
- [x] Parse EndpointSlice support to Xds IR Route Destinations in the Gateway API Translator https://github.com/envoyproxy/gateway/pull/1494
- [ ] #1900

[*Relevant Links*:]
[Similar to Countour open issue](https://github.com/projectcontour/contour/issues/2696).
This was shortly discussed [here](https://github.com/envoyproxy/gateway/issues/35) and [this explanation](https://github.com/envoyproxy/gateway/issues/35#issuecomment-1396422011)
This was fixed on Feb 15 for nginx-ingress [here](https://github.com/kubernetes/ingress-nginx/pull/8890)

Xunzhuo:
Thanks for raising it @pgold30, I will take this in v0.5.0.

arkodg:
thanks for picking this one up @Xunzhuo  !

arkodg:
sidenote - this might also require an `EnvoyProxy` API change to add a field to disable endpoint resolution / fall back to service VIP, in case the user prefers routing/load balancing via kube-proxy

arkodg:
moving this into the backlog for now, since this is not a must have for v0.5.0

Xunzhuo:
Hey @arkodg, since you have started working on this issue, let me transfer this to you, thanks!

PR #1474, opened 2023-06-01
feat: Support EndpointSlice in Kubernetes Provider

arkodg:
* Reconcile relavant EndpointSlices associated with a Service that is referenced in a xRoute
* Add the EndpointSlices to the provider message

Relates to https://github.com/envoyproxy/gateway/issues/1256

### DIFF
--- a/charts/gateway-helm/templates/generated/rbac/roles.yaml
+++ b/charts/gateway-helm/templates/generated/rbac/roles.yaml
@@ -34,6 +34,14 @@ rules:
   - update
   - watch
 - apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - gateway.envoyproxy.io
   resources:
   - authenticationfilters
--- a/internal/gatewayapi/resource.go
+++ b/internal/gatewayapi/resource.go
@@ -7,6 +7,7 @@ package gatewayapi
 
 import (
 	v1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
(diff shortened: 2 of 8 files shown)


zirain:
this's only part of https://github.com/envoyproxy/gateway/issues/1256?

arkodg:
> this's only part of #1256?

yes, we need to take this data, and parse it in gateway api library which I plan on doing in a follow up PR

zirain:
> > this's only part of #1256?
> 
> yes, we need to take this data, and parse it in gateway api library which I plan on doing in a follow up PR

what about `endpoint`, I know it will EOL soon, but in real world people always running old version kubernetes.

arkodg:
> > > this's only part of #1256?
> > 
> > 
> > yes, we need to take this data, and parse it in gateway api library which I plan on doing in a follow up PR
> 
> what about `endpoint`, I know it will EOL soon, but in real world people always running old version kubernetes.

currently we only support last 3 k8s versions 
https://gateway.envoyproxy.io/v0.4.0/intro/compatibility.html#compatibility-matrix
by the time we GA, my guess is most users will be using k8s versions with endpointslice enabled

zirain:
> > > > this's only part of #1256?
> > > 
> > > 
> > > yes, we need to take this data, and parse it in gateway api library which I plan on doing in a follow up PR
> > 
> > 
> > what about `endpoint`, I know it will EOL soon, but in real world people always running old version kubernetes.
> 
> currently we only support last 3 k8s versions https://gateway.envoyproxy.io/v0.4.0/intro/compatibility.html#compatibility-matrix by the time we GA, my guess is most users will be using k8s versions with endpointslice enabled

I don't think that's the real world.

arkodg:
> > > > > this's only part of #1256?
> > > > 
> > > > 
> > > > yes, we need to take this data, and parse it in gateway api library which I plan on doing in a follow up PR
> > > 
> > > 
> > > what about `endpoint`, I know it will EOL soon, but in real world people always running old version kubernetes.
> > 
> > 
> > currently we only support last 3 k8s versions https://gateway.envoyproxy.io/v0.4.0/intro/compatibility.html#compatibility-matrix by the time we GA, my guess is most users will be using k8s versions with endpointslice enabled
> 
> I don't think that's the real world.

users can always revert to the `ClusterIP` LB policy  https://github.com/envoyproxy/gateway/pull/1477

zirain (approved):

Issue #1471, opened 2023-05-31
Add functionality in cert-gen to skip creating certs or rewrite existing certs


arkodg:
*Description*:
* By default cert-gen should not create new certs, if they already exist in the same namespace, e.g. during upgrades
* A new arg such as `--rewrite-certs` should be added if we want cert-gen to delete existing certs and create new ones

kflynn:
This sounds good to me, yeah. The default for the update case should be to reuse your existing certs.

PR #1476, opened 2023-06-02
Update control plane certs only if user opts in

arkodg:
* Make sure certgen does not update the control plane cert secrets if they already exist. This ensures that if a user creates certs as part of https://github.com/envoyproxy/gateway/issues/866 they are not overriden by certgen.

* If `overwriteControlPlaneCerts` is set in the `EnvoyGateway` config, certgen can be used to update the secrets with newer certs

Fixes: https://github.com/envoyproxy/gateway/issues/1471

### DIFF
--- a/api/config/v1alpha1/envoygateway_types.go
+++ b/api/config/v1alpha1/envoygateway_types.go
@@ -160,6 +160,8 @@ type EnvoyGatewayKubernetesProvider struct {
 	// should be deployed
 	// +optional
 	Deploy *KubernetesDeployMode `json:"deploy,omitempty"`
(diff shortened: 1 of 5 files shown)


AliceProxy (approved):
Changes look fine to me. Do we want to also update the quickstart to have users set this to true and explain the field?

arkodg:
@AliceProxy I'll add a sub task to add docs and we can use the orig issue to track it to completion

AliceProxy:
> @AliceProxy I'll add a sub task to add docs and we can use the orig issue to track it to completion

Cool, yeah that doesn't need to happen in this PR

zirain (approved):

Issue #1105, opened 2023-03-06
proposal: LoadbalancingPolicy


arkodg:
*Description*:
Allow the user to specify Load Balancing policies to load balance requests from the gateway to the upstream.

[optional *Relevant Links*:]
Relates to https://github.com/kubernetes-sigs/gateway-api/issues/1778

Currently the requests are being load balanced at an L3 / kube-proxy level since requests are being sent to the Cluster IP of the Service. We need to route to the endpoints directly with https://github.com/envoyproxy/gateway/issues/35

qicz:
should we provide the global locadbalancing policy in envoyproxy API?

arkodg:
+1 to adding a global policy in `EnvoyProxy`, it probably needs to be done in a way that it can be overridden per Backend

PR #1477, opened 2023-06-03
API for LoadBalancing Strategy in EnvoyProxy API

arkodg:
* Allows the user to configure a global load balancing strategy for the envoy proxy data plane

```
kind: EnvoyProxy
metadata:
  name: set-lb-strategy
spec:
  traffic:
    loadBalancer:
      strategy: LeastRequest
```
* Sets the default load balancing strategy to `LeastRequest` which will internally set the `WeightedLeastRequest` knob in Envoy

* Relates to https://github.com/envoyproxy/gateway/issues/1256 allowing the user to fall back to `ClusterIP` style loadbalancing

* Relates to https://github.com/envoyproxy/gateway/issues/1105


zirain:
does gateway API support config this per route/backend?

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -164,6 +167,49 @@ const (
...
+	LoadBalancerStrategyRoundRobin LoadBalancerStrategy = "RoundRobin"
+
+	// LoadBalancerStrategyRandom defines the "Random" load balancing strategy.
+	// In this mode, the upstream host is chosen randomly.
+	LoadBalancerStrategyRandom LoadBalancerStrategy = "Random"
+
+	// LoadBalancerStrategyClusterIP defines the "ClusterIP" load balancing strategy.
+	// In this mode, EnvoyProxy routes requests to the ClusterIP of the service and
+	// relies on another entity such as kube-proxy to perform load balancing to the upstream hosts.
+	LoadBalancerStrategyClusterIP LoadBalancerStrategy = "ClusterIP"
```
zirain, line 210: does this mean, eds will just send ClusterIP?

arkodg: yes, this is what is happening as of v0.4.0

arkodg:
> does gateway API support config this per route/backend?

no, another way to expose this is to create a hierarchical policy called `TrafficPolicy` that can be applied to Gateway and to HTTPRoute, I think either should work

api/config/v1alpha1/envoyproxy_types.go
zirain: correct me if I'm wrong, only `ClusterIP` will change the result of `EDS`, other types only change `LbPolicy` on envoy cluster?

arkodg: yeah thats right :)

zirain: this's a little odd to me, Is there any way to make them more consistent?

arkodg: can you share why its odd, this API is meant for end users (platform owners / app owners) to configure LB policy, so they can
* Use LB Policies provided at a L7 request level by Envoy (round-robin, least request, etc) or
* LB Policies provided at a L3/L4 flow level by another intermediate data plane entity such as kube-proxy

arkodg:
closing in favor of https://github.com/envoyproxy/gateway/issues/1492

PR #1478, opened 2023-06-05
feat: bump gwapi to v0.7.1

Xunzhuo:
**What type of PR is this?**

feat: bump gwapi to v0.7.1

bump gwapi to v0.7.1

### DIFF
--- a/internal/xds/translator/translator.go
+++ b/internal/xds/translator/translator.go
@@ -104,7 +104,8 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 		// Create a route config if we have not found one yet
 		if xdsRouteCfg == nil {
 			xdsRouteCfg = &routev3.RouteConfiguration{
-				Name: httpListener.Name,
+				IgnorePortInHostMatching: true,
+				Name:                     httpListener.Name,
 			}
 			tCtx.AddXdsResource(resourcev3.RouteType, xdsRouteCfg)
 		}
--- a/internal/xds/translator/translator_test.go
+++ b/internal/xds/translator/translator_test.go
@@ -6,7 +6,10 @@
 package translator
 
 import (
+	"bufio"
 	"embed"
+	"flag"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -28,6 +31,10 @@ var (
 	outFiles embed.FS
 	//go:embed testdata/in/*
 	inFiles embed.FS
+)
+
+var (
+	overrideTestData = flag.Bool("override-testdata", false, "if override the test output data.")
 )
 
 func TestTranslateXds(t *testing.T) {
@@ -163,12 +170,21 @@ func TestTranslateXds(t *testing.T) {
 			routes := tCtx.XdsResources[resourcev3.RouteType]
 			clusters := tCtx.XdsResources[resourcev3.ClusterType]
 			endpoints := tCtx.XdsResources[resourcev3.EndpointType]
+			if *overrideTestData {
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, listeners), "xds-ir", tc.name+".listeners.yaml")
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, routes), "xds-ir", tc.name+".routes.yaml")
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, clusters), "xds-ir", tc.name+".clusters.yaml")
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, endpoints), "xds-ir", tc.name+".endpoints.yaml")
+			}
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".listeners.yaml"), requireResourcesToYAMLString(t, listeners))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".routes.yaml"), requireResourcesToYAMLString(t, routes))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".clusters.yaml"), requireResourcesToYAMLString(t, clusters))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".endpoints.yaml"), requireResourcesToYAMLString(t, endpoints))
 			if tc.requireSecrets {
 				secrets := tCtx.XdsResources[resourcev3.SecretType]
+				if *overrideTestData {
+					overrideTestDataOutFile(t, requireResourcesToYAMLString(t, secrets), "xds-ir", tc.name+".secrets.yaml")
+				}
 				require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".secrets.yaml"), requireResourcesToYAMLString(t, secrets))
 			}
 		})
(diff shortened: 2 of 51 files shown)


arkodg:
looks like `IgnorePortInHostMatching` https://pkg.go.dev/github.com/envoyproxy/go-control-plane/envoy/config/route/v3#RouteConfiguration needs to be set instead of stripping the port from the host (i.e. revert https://github.com/envoyproxy/gateway/pull/1389)
thanks to @sunjayBhatia & @howardjohn's triage in https://github.com/kubernetes-sigs/gateway-api/issues/2091

qicz (approved):

Xunzhuo:
@arkodg so we need to revert https://github.com/envoyproxy/gateway/pull/1389 ?

arkodg:
yah @Xunzhuo , we'll need to 
* revert https://github.com/envoyproxy/gateway/pull/1389
* add `IgnorePortInHostMatching` in every RouteConfiguration
* then bump to v0.7.1

can be done in this PR or we can raise issues for each

arkodg:
looks like `HTTPRouteRequestMirror` will need to be skipped, can you also raise a issue to track it

internal/xds/translator/translator_test.go
```diff
@@ -163,12 +165,17 @@ func TestTranslateXds(t *testing.T) {
 			routes := tCtx.XdsResources[resourcev3.RouteType]
 			clusters := tCtx.XdsResources[resourcev3.ClusterType]
 			endpoints := tCtx.XdsResources[resourcev3.EndpointType]
+			overrideTestDataOutFile(t, requireResourcesToYAMLString(t, listeners), "xds-ir", tc.name+".listeners.yaml")
```
arkodg, line 168: why is this override needed ?

internal/xds/translator/translator.go
```diff
@@ -104,7 +104,8 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 		// Create a route config if we have not found one yet
 		if xdsRouteCfg == nil {
 			xdsRouteCfg = &routev3.RouteConfiguration{
-				Name: httpListener.Name,
+				IgnorePortInHostMatching: true,
+				Name:                     httpListener.Name,
```
arkodg, line 108: trying to better understand how the yaml already contained the `name` before this addition, is it also set somewhere else in the codebase ?

Xunzhuo: We had `Name` set in RouteConfiguration before, we just added `IgnorePortInHostMatching` here.

internal/xds/translator/translator_test.go
Xunzhuo: Trying to update the tons of json/yaml output files automatically by running `make test`, otherwise I need to manually update all of them manually (40+ files).

internal/xds/translator/translator.go
arkodg: my bad 🙈

internal/xds/translator/translator_test.go
arkodg: I love the idea, can we use a seperate `make` target to achieve this, as well as in a follow up PR
running `make test` to check errors and fix errors feels fishy, how about `make testdata` to perform this specific task.

Xunzhuo: Done : )

internal/xds/translator/translator_test.go
```diff
@@ -163,12 +170,21 @@ func TestTranslateXds(t *testing.T) {
...
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, endpoints), "xds-ir", tc.name+".endpoints.yaml")
+			}
+
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".listeners.yaml"), requireResourcesToYAMLString(t, listeners))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".routes.yaml"), requireResourcesToYAMLString(t, routes))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".clusters.yaml"), requireResourcesToYAMLString(t, clusters))
 			require.Equal(t, requireTestDataOutFile(t, "xds-ir", tc.name+".endpoints.yaml"), requireResourcesToYAMLString(t, endpoints))
 			if tc.requireSecrets {
 				secrets := tCtx.XdsResources[resourcev3.SecretType]
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, secrets), "xds-ir", tc.name+".secrets.yaml")
```
arkodg, line 187: guessing these need to be deleted ?

internal/xds/translator/translator_test.go
```diff
@@ -294,12 +310,17 @@ func TestTranslateXdsWithExtension(t *testing.T) {
 				routes := tCtx.XdsResources[resourcev3.RouteType]
 				clusters := tCtx.XdsResources[resourcev3.ClusterType]
 				endpoints := tCtx.XdsResources[resourcev3.EndpointType]
+				overrideTestDataOutFile(t, requireResourcesToYAMLString(t, listeners), "extension-xds-ir", tc.name+".listeners.yaml")
```
arkodg, line 313: same as above

internal/xds/translator/translator_test.go
Xunzhuo: Updated.

internal/xds/translator/translator_test.go
Xunzhuo: Updated.

internal/xds/translator/translator_test.go
arkodg: missed 
```
if *overrideTestData {
```
here

Xunzhuo: Done, thanks : )

arkodg (approved):
LGTM, thanks for also improving dev velocity by introducing `make testdata`

howardjohn:
> I think IgnorePortInHostMatching won't work for :authority exact/suffix header matching, which is set up when we have hosts defined in httproutes and not in listeners

Istio handles this by matching in `domains` rather than a `:authority` header match

jackkleeman:
ok, maybe i can pr a fix to use that approach for the routes instead of :authority

jackkleeman:
hm, the problem here is that domains is listener-scoped, but i am talking about the case where httproute is more specific on hostname than listeners - in that case we are currently creating a authority header match. maybe that header match just needs to be more port agnostic

howardjohn:
I mean here:
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-virtualhost

jackkleeman:
no indeed, but doesnt virtualhost map to gateway listener?

howardjohn:
Listener: match whatever
Routes:
```
virtual_hosts:
- domains: [*.example.com]
- domains: [bar.example.com]
```

jackkleeman:
sure, but envoy/gateway currently creates only 1 virtual host per HttpListener as defined in the gateway object:

```
// Allocate virtual host for this httpListener.
		// 1:1 between IR HTTPListener and xDS VirtualHost
		vHost := &routev3.VirtualHost{
			Name:    httpListener.Name,
			Domains: httpListener.Hostnames,
		}
```

So we would need to change that to be 1-to-many irght?

howardjohn:
Yes

howardjohn:
TBH I did not realize I was commenting in envoyproxy/gateway repo :slightly_smiling_face:

jackkleeman:
hah! no worries

arkodg:
hey @jackkleeman thanks for surfacing this issue, created a newer issue to track this limitation https://github.com/envoyproxy/gateway/issues/1687

Issue #1409, opened 2023-05-09
ratelimit: unmanged mode


zirain:
Recently, there's a lot of PRs about how to managed a ratelimit backend. 
For a long term, EG should not put all properties into KubernetesDeploymentSpec or KubernetesServiceSpec

Alternatives:
1. EG should make https://github.com/envoyproxy/gateway/issues/1254 as default
2. Support unmanged mode ratelimit backend

With unmanged mode, end-users should take care of the all k8s resources(e.g. deployment, service, configmap) themself.
EG will only send configuration via gRPC.

@arkodg @Xunzhuo @qicz  WDYT?

qicz:
xref #1379 IMO,  EG should manage envoy eco components easily for k8s resources, envoy xds and ratelimit configurations. 

As an end-user, its more convenient for me.

arkodg:
yeah this is fine, should follow the same flow/API as #1379

zirain:
I'd like close this first.

PR #1490, opened 2023-06-05
[WIP] ratelimit: support external backend

zirain:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/1409


arkodg:
@zirain this means, that the external RL service MUST adhere to [this RateLimit config](https://github.com/envoyproxy/go-control-plane/blob/main/ratelimit/config/ratelimit/v3/rls_conf.pb.go) spec, and this must be mentioned in the API field
Can you please add WHY this is needed in the Rate Limit design doc as well, thanks

PR #1493, opened 2023-06-06
chore: remove unused files

zirain:
**What type of PR is this?**

Fixes #


charts/eg/crds/generated/config.gateway.envoyproxy.io_envoyproxies.yaml
```diff
@@ -1,239 +0,0 @@
----
```
arkodg, line 1: can you please remind me why this  isn't needed anymore

zirain: I think it's moved to charts/gateway-helm and will not be update by `make generate manifests`

arkodg: good catch, so `charts/eg` is not longer needed ? 
cc @Xunzhuo

Xunzhuo (approved):
good catch!

Issue #1492, opened 2023-06-05
proposal: TrafficPolicy


arkodg:
*Description*:
Introduce a [hierarchical policy attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) called `TrafficPolicy` that can be applied to the `Gateway` and `HTTPRoute` / `xRoute` resources to further configure the traffic ingressing in which can be applied either for the `downstream` connection or `upstream` connection.

```
kind: TrafficPolicy
spec:
  default:
    downstream:
      keepalive: .....
    upstream:
      keepalive: .....
      cors:
      loadBalancer:
      tls:
  targetRef:
    kind: Gateway
    name: example
```    

Relates to

* https://github.com/envoyproxy/gateway/issues/1105
* https://github.com/envoyproxy/gateway/issues/1031
* https://github.com/envoyproxy/gateway/issues/877
* https://github.com/envoyproxy/gateway/issues/837
* https://github.com/envoyproxy/gateway/issues/731
* https://github.com/envoyproxy/gateway/issues/567
* https://github.com/envoyproxy/gateway/issues/671
* https://github.com/envoyproxy/gateway/issues/492
* https://github.com/envoyproxy/gateway/issues/1328

arkodg:
ptal @envoyproxy/gateway-maintainers

kflynn:
@AliceProxy and I have done some preliminary work on how Gateway API compares to the extant Emissary configuration language [here](https://docs.google.com/document/d/1Yrm13z0NBFIEM0_xQazZnbVF1h7Mte1vcPdtqDBodoM/edit#heading=h.pxkddc96i5m1), which might be relevant to this discussion.

arkodg:
in the community meeting yesterday, @AliceProxy brought up a great way to split up this API
* A Direct DownstreamTrafficPolicy` API which can be applied only to the `Gateway` resource and holds downstream specific information applicable for the downstream listener
* A `UpstreamTrafficPolicy` API which can be applied to the `Gateway` resource as well as the `xRoute` resource.

Im a big +1 on this approach, here are some open issues that need more debate
* how do we deal with `defaults` and `overrides` when an `UpstreamTrafficPolicy` is applied to a `Gateway` as well a `HTTPRoute` that is attached to the same `Gateway`, which policy field wins ? Making this policy type a [hierarchical policy attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) should solve this issue
[...]

PR #1494, opened 2023-06-06
EndpointSlice to IR Route Destinations

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1256


arkodg:
this issue is blocked on https://github.com/envoyproxy/gateway/issues/1105 which is blocked on https://github.com/envoyproxy/gateway/issues/1492, because we would like to provide a way to fallback to ClusterIP Loadbalancing support for users who are running on systems with EndpointSlice disabled or prefer L3 Loadbalancing via ClusterIP / kube-proxy

internal/gatewayapi/route.go
```diff
@@ -980,28 +982,83 @@ func (t *Translator) processDestEndpoints(backendRef v1beta1.BackendRef,
...
+								if routeType == KindTCPRoute ||
+									routeType == KindUDPRoute {
+									ep = ir.NewDestEndpoint(
+										address,
+										uint32(servicePort.TargetPort.IntVal))
+								} else {
+									ep = ir.NewDestEndpointWithWeight(
+										address,
+										uint32(servicePort.TargetPort.IntVal),
+										weight)
```
zhaohuabing, line 1038: I think the `weight` is not correct here.

Assume we have a route rule consisting of two backends:

```yaml
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: foo
          port: 3000
          weight: 50
        - group: ""
          kind: Service
          name: bar
          port: 3000
          weight: 50
```

The "foo" service has 100 endpoints and the "bar" service has 500 endpoints. The actual traffic ratio between foo and bar is 2:1 instead of the intended 1:1.

internal/cmd/egctl/translate.go
```diff
@@ -262,9 +262,10 @@ func translateGatewayAPIToGatewayAPI(resources *gatewayapi.Resources) (gatewayap
 
 	// Translate from Gateway API to Xds IR
 	gTranslator := &gatewayapi.Translator{
-		GatewayControllerName:  egv1alpha1.GatewayControllerName,
-		GatewayClassName:       v1beta1.ObjectName(resources.GatewayClass.Name),
-		GlobalRateLimitEnabled: true,
+		GatewayControllerName:   egv1alpha1.GatewayControllerName,
+		GatewayClassName:        v1beta1.ObjectName(resources.GatewayClass.Name),
+		GlobalRateLimitEnabled:  true,
+		EndpointRoutingDisabled: true,
```
zhaohuabing, line 268: I understand that we want the routing behavior to be backward compatible, but I think using endpoints as default to route traffic allows EG to bypass kube-proxy's overhead and fully leverage Envoy's advanced capabilities: more LB algorithms, session sticky, etc.

internal/gatewayapi/route.go
arkodg: thanks for catching this, fixing this rn

internal/cmd/egctl/translate.go
arkodg: I have disabled this in `egctl x translate`, because if we enable it, it will be mandatory, and all translations will fail, I was planning on raising a separate GH issue on dealing with this convenience / correctness dilemma

zhaohuabing (approved):

zhaohuabing (approved):
LGTM. Just need to fix the lint and test.

internal/xds/translator/ratelimit.go
```diff
@@ -429,7 +430,10 @@ func (t *Translator) createRateLimitServiceCluster(tCtx *types.ResourceVersionTa
 	clusterName := getRateLimitServiceClusterName()
 	// Create cluster if it does not exist
 	host, port := t.getRateLimitServiceGrpcHostPort()
-	endpoints := []*ir.DestinationEndpoint{ir.NewDestEndpoint(host, uint32(port))}
+	ds := &ir.DestinationSetting{
+		Weight:    ptr.To(uint32(1)),
+		Endpoints: []*ir.DestinationEndpoint{ir.NewDestEndpoint(host, uint32(port))},
+	}
```
zirain, line 436: this can be simplified to a function

arkodg: can this refactor be done in a follow up ?
this PR is changing 255 files most of which are YAML, and will affect other PRs

zirain (approved):

Issue #1454, opened 2023-05-26
Control Plane Auth for Rate Limit Service


arkodg:
*Description*:
Ensures all communication with the Rate Limit Service is encrypted & authenicated.

- [x]  Generate keys/certs for the Rate Limit Service in the Cert-gen job https://github.com/envoyproxy/gateway/blob/80db45944e010a5ba8eb2636201450ffb1a59f8e/internal/crypto/certgen.go#L141 https://github.com/envoyproxy/gateway/pull/1499
- [x]  Setup TLS b/w EG RL xDS Server and RL Service
  - [x] xds Server  https://github.com/envoyproxy/gateway/blob/80db45944e010a5ba8eb2636201450ffb1a59f8e/internal/globalratelimit/runner/runner.go#L80 https://github.com/envoyproxy/gateway/pull/1523
  - [x] RLS https://github.com/envoyproxy/ratelimit#xds-management-server-based-configuration-loading https://github.com/envoyproxy/gateway/pull/1523
- [x] Setup TLS b/w Envoy Proxy and RLS Service 
  - [x] RL Filter https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/ratelimit/v3/rls.proto#config-ratelimit-v3-ratelimitserviceconfig https://github.com/envoyproxy/gateway/pull/1499
  - [x] RLS  https://github.com/envoyproxy/ratelimit#mtls https://github.com/envoyproxy/gateway/pull/1499
 
[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

oowl:
Hi, I can work on this feature, Please assign to me if it is Okay.

zirain:
@oowl assign this to me

arkodg:
Thanks for adding this feature @oowl !

PR #1499, opened 2023-06-06
feat(crypto): control plane TLS support for Rate Limit Service phase I

oowl:
**What type of PR is this?**

add TLS Support between the envoy and ratelimted services

Relates to #1454

### DIFF
--- a/internal/crypto/certgen.go
+++ b/internal/crypto/certgen.go
@@ -70,11 +70,13 @@ const (
 // Certificates contains a set of Certificates as []byte each holding
 // the CA Cert along with Envoy Gateway & Envoy certificates.
 type Certificates struct {
-	CACertificate           []byte
-	EnvoyGatewayCertificate []byte
-	EnvoyGatewayPrivateKey  []byte
-	EnvoyCertificate        []byte
-	EnvoyPrivateKey         []byte
+	CACertificate             []byte
+	EnvoyGatewayCertificate   []byte
+	EnvoyGatewayPrivateKey    []byte
+	EnvoyCertificate          []byte
+	EnvoyPrivateKey           []byte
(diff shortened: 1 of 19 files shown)


internal/crypto/certgen.go
```diff
@@ -70,11 +73,15 @@ const (
...
-	EnvoyGatewayCertificate []byte
-	EnvoyGatewayPrivateKey  []byte
-	EnvoyCertificate        []byte
-	EnvoyPrivateKey         []byte
+	CACertificate             []byte
+	EnvoyGatewayCertificate   []byte
+	EnvoyGatewayPrivateKey    []byte
+	EnvoyCertificate          []byte
+	EnvoyPrivateKey           []byte
+	RateLimitCertificate      []byte
```
arkodg, line 81: curious why we need ratelimit and envoyratelimit, we just need one for the rl service xds client, right ?

oowl: Yeah, we can reuse the existing cert, I have been undecided about this before, so I will change it.

internal/infrastructure/kubernetes/ratelimit/resource_provider_test.go
```diff
@@ -16,6 +16,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/pointer"
+
+	// "k8s.io/apimachinery/pkg/api/resource"
```
arkodg, line 20: delete these instead of commenting it out ?

arkodg:
looks really good @oowl, added one comment, thanks for working on this feature !

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1502, opened 2023-06-07
refactor: dedup the route context code using reflection

### LINKED ISSUE #1252: Dedup code pertaining to xRoute using reflection
*Description*:
There is a lot of code in https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/contexts.go and https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/route.go that could be deduplicated using reflection https://pkg.go.dev/reflect

shawnh2:
**What type of PR is this?**

Fixes #1252


arkodg:
thanks for picking this refactor up @shawnh2 !

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,100 @@ type TCPRouteContext struct {
...
+}
+
+// GetRouteType returns the Kind of the Route object, HTTPRoute,
+// TLSRoute, TCPRoute, UDPRoute etc.
+//
+// This function use the typename of RouteContext and its return is
+// corresponding to the const defined in translator, like KindHTTPRoute,
+// KindGRPCRoute, KindTLSRoute, KindTCPRoute, KindUDPRoute
+func GetRouteType(route RouteContext) v1beta1.Kind {
+	rv := reflect.ValueOf(route)
```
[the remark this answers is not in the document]
shawnh2: thanks for pointing it out, but i think https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field is not suitable for our case.

despite the fact that this package provides several powerful methods, but its intention is for validations only. in our case, we need more than that, we need to write some values for some fields in one struct as well, but this package does not have that kind of ability.

i thought about whether we could replace current validations with the one provided in that package, and turns out we may not. [for example](https://github.com/kubernetes-sigs/gateway-api/blob/d4347bfa43866525e57770210487e22e6abb2102/apis/v1beta1/validation/httproute.go#L160), the `strings.HasPrefix(*path.Value, "/")` use the variable `path` directly, which has explicit type `*gatewayv1b1.HTTPPathMatch`. however, in our case, the only variable we can manipulate with is `route`, which is a interface `RouteContext`. so we need the ability provided in package `reflect`, to play with the real type behinds whatever it represents.

so basically, i think `relfect` is inevitable.

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,100 @@ type TCPRouteContext struct {
...
-	for i, p := range t.Spec.ParentRefs {
-		p := UpgradeParentReference(p)
-		if reflect.DeepEqual(p, forParentRef) {
-			upgraded := UpgradeParentReference(t.Spec.ParentRefs[i])
-			parentRef = &upgraded
+	specParentRefs := rv.FieldByName("Spec").FieldByName("ParentRefs")
+	for i := 0; i < specParentRefs.Len(); i++ {
+		p := specParentRefs.Index(i).Interface().(v1beta1.ParentReference)
+		up := p
+		if !isHTTPRoute {
```
arkodg, line 296: curious why this custom `HTTPRoute` logic is needed

shawnh2: becasue there are some differences between the implementation of `HTTPRoute` and other Routes in `GetRouteParentContext`

for example, `HTTPRoute` is like:

https://github.com/envoyproxy/gateway/blob/7f8a4913b78b492c192b85262344dce4f4f64314/internal/gatewayapi/contexts.go#L220-L226

and other Route is like:

https://github.com/envoyproxy/gateway/blob/7f8a4913b78b492c192b85262344dce4f4f64314/internal/gatewayapi/contexts.go#L471-L479

so add a `if` logic to deal with `HTTPRoute`

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,100 @@ type TCPRouteContext struct {
...
+// GetRouteType returns the Kind of the Route object, HTTPRoute,
+// TLSRoute, TCPRoute, UDPRoute etc.
+//
+// This function use the typename of RouteContext and its return is
+// corresponding to the const defined in translator, like KindHTTPRoute,
+// KindGRPCRoute, KindTLSRoute, KindTCPRoute, KindUDPRoute
+func GetRouteType(route RouteContext) v1beta1.Kind {
+	rv := reflect.ValueOf(route)
+	rt := rv.Type().String()
+	return v1beta1.Kind(strings.TrimSuffix(rt, "Context")[strings.LastIndex(rt, ".")+1:])
```
arkodg, line 224: this looks a little intimidating to me, can we use `FieldByName` here ?

shawnh2: sure we can use `FieldByName`, in that case, we need to add one more field, like `type string` in struct HTTP/GRPC/.../Context, indicating the type of this struct, and all the value of field `type` comes from:

https://github.com/envoyproxy/gateway/blob/4dd4f87b8b5b83673e97ee0b77ce372cd90ce273/internal/gatewayapi/translator.go#L17-L22

@arkodg see if this is ok?

arkodg: or you could also use `GetKind()` ?
see it being used in other places in the codebase
https://github.com/envoyproxy/gateway/blob/4dd4f87b8b5b83673e97ee0b77ce372cd90ce273/internal/gatewayapi/filters.go#L769

shawnh2: thanks for the heads up!

now we can use `FieldByName("Kind")`, BTW, I add some `TypeMeta` for xRoute resources in `translate` func, nor it will keep getting `""`.

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,95 @@ type TCPRouteContext struct {
...
 
-func (t *TCPRouteContext) GetRouteType() v1beta1.Kind {
-	return KindTCPRoute
+// TODO: [v1alpha2-v1beta1] This should not be required once all Route
+// objects being implemented are of type v1beta1.
+// GetHostnames returns the hosts targeted by the Route object.
+func GetHostnames(route RouteContext) []string {
+	rv := reflect.ValueOf(route).Elem()
+	rt := rv.FieldByName("Kind").String()
+	if strings.Contains(rt, "TCP") || strings.Contains(rt, "UDP") {
```
arkodg, line 228: can we use existing constants here https://github.com/envoyproxy/gateway/blob/ca36e3f0b11f9d742796f136ad0853295b37352d/internal/gatewayapi/translator.go#L20

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,95 @@ type TCPRouteContext struct {
...
-	for idx, p := range t.Spec.ParentRefs {
-		parentReferences[idx] = UpgradeParentReference(p)
+// TODO: [v1alpha2-v1beta1] This should not be required once all Route
+// objects being implemented are of type v1beta1.
+// GetParentReferences returns the ParentReference of the Route object.
+func GetParentReferences(route RouteContext) []v1beta1.ParentReference {
+	rv := reflect.ValueOf(route).Elem()
+	rt := rv.FieldByName("Kind").String()
+	pr := rv.FieldByName("Spec").FieldByName("ParentRefs")
+	if strings.Contains(rt, "HTTP") || strings.Contains(rt, "GRPC") {
```
arkodg, line 247: can we use existing constants here  https://github.com/envoyproxy/gateway/blob/ca36e3f0b11f9d742796f136ad0853295b37352d/internal/gatewayapi/translator.go#L20

internal/gatewayapi/contexts.go
```diff
@@ -527,40 +209,95 @@ type TCPRouteContext struct {
...
 	}
 
-	if ctx := t.parentRefs[forParentRef]; ctx != nil {
+	if p := pr.MapIndex(reflect.ValueOf(forParentRef)); p.IsValid() && !p.IsZero() {
+		ctx := p.Interface().(*RouteParentContext)
 		return ctx
 	}
 
+	isHTTPRoute := false
+	if strings.Contains(rv.Type().String(), "HTTP") {
```
arkodg, line 282: same as above

arkodg:
thanks @shawnh2 for addressing the comments, adding a minor comment, else LGTM !

internal/gatewayapi/contexts.go
shawnh2: sure, i'm missing that details

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1508, opened 2023-06-11
e2e: add accesslog test

zirain:


arkodg:
this looks great @zirain, in the future, can we reuse `Loki` and also deploy `otel-collector` to ensure access logs are sent to `loki` via `otel-collector` ?

test/e2e/tests/accesslog.go
```diff
@@ -0,0 +1,133 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	httputils "sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, AccessLogTest)
```
arkodg, line 31: lets sprinkle in `File` or `StdOut` prefix to test names to differentiate it from  `OpenTelemetry` AccessLog tests ?

zirain:
> this looks great @zirain, in the future, can we reuse `Loki` and also deploy `otel-collector` to ensure access logs are sent to `loki` via `otel-collector` ?

sure, the orginal plan is reuse fluent-bit for `opentelemety` and `stdout`, but currently grpc listner is not supported, upstream issue: https://github.com/fluent/fluent-bit/issues/7228

test/e2e/tests/accesslog.go
zirain: make sense, will update later.

examples/fluent-bit/fluent-bit.yaml
```diff
@@ -0,0 +1,227 @@
+---
```
arkodg, line 1: should we use helm chart instead https://docs.fluentbit.io/manual/installation/kubernetes#installing-with-helm-chart ?
which is easier to maintain

zirain: yes, actually these lines are generated by `helm template`.

examples/loki/loki.yaml
```diff
@@ -0,0 +1,282 @@
...
+metadata:
+  name: loki
+  labels:
+    helm.sh/chart: loki-4.8.0
+    app.kubernetes.io/name: loki
+    app.kubernetes.io/instance: loki
+    app.kubernetes.io/version: "2.7.3"
+    app.kubernetes.io/managed-by: Helm
+spec:
+  type: LoadBalancer
```
zirain, line 150: we need a LB type sercie in e2e, the latest loki chart does suppport change it with custom values yaml file

https://github.com/grafana/loki/blob/8391e557b1b9e36c11a1c9458de3d8e28f56b16e/production/helm/loki/templates/single-binary/service.yaml#L12

will move to use helm chart when it's possible to do.

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

PR #1521, opened 2023-06-14
feat: support deployment strategy settings.

### LINKED ISSUE #1520: Support EnvoyProxy&RateLimit deployment strategy settings
xref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy

qicz:
Fixes #1520 

For more others settings, users can use the deployment patch yaml set them. like follows

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: envoy
    gateway.envoyproxy.io/owning-gateway-name: default
  name: envoy-default-eg-63636f73
spec:
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 50%
      maxUnavailable: 70%
```

### DIFF
--- a/api/config/v1alpha1/helpers.go
+++ b/api/config/v1alpha1/helpers.go
@@ -6,6 +6,7 @@
 package v1alpha1
 
 import (
+	appv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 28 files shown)


arkodg (approved):
LGTM thanks !

zirain:
I think we should provide an common way like Kustomize to support things like this, rather than adding every single fields from Deployment/Service to EG's API.

qicz:
> I think we should provide an common way like Kustomize to support things like this, rather than adding every single fields from Deployment/Service to EG's API.

IMO, some common fields we should intervene in, another field can be set like PR description using the patch yaml or Kustomize. this `strategy` setting is the basic for deploy.

Xunzhuo (approved):
/lgtm

zirain (approved):

PR #1523, opened 2023-06-14
feat(crypto): add ratelimit xds tls function

oowl:
**What type of PR is this?**

add TLS Support between the gateway and ratelimted services

Relates to https://github.com/envoyproxy/gateway/issues/1454

### DIFF
--- a/internal/infrastructure/kubernetes/ratelimit/resource.go
+++ b/internal/infrastructure/kubernetes/ratelimit/resource.go
@@ -46,16 +46,24 @@ const (
 	GRPCServerUseTLSEnvVar = "GRPC_SERVER_USE_TLS"
 	// GRPCServerTLSCertEnvVar is the grpc server tls cert.
 	GRPCServerTLSCertEnvVar = "GRPC_SERVER_TLS_CERT"
-	// GRPCServerTLSClientCertFilename is the GRPC tls cert file.
-	GRPCServerTLSCertFilename = "/certs/tls.crt"
+	// GRPCTLSCertFilename is the GRPC tls cert file.
+	GRPCTLSCertFilename = "/certs/tls.crt"
 	// GRPCServerTLSKeyEnvVarEnvVar is the grpc server tls key.
 	GRPCServerTLSKeyEnvVarEnvVar = "GRPC_SERVER_TLS_KEY"
-	// GRPCServerTLSKeyFilename is the grpc server key file.
-	GRPCServerTLSKeyFilename = "/certs/tls.key"
+	// GRPCTLSKeyFilename is the grpc server key file.
+	GRPCTLSKeyFilename = "/certs/tls.key"
 	// GRPCServerTLSCACertEnvVar is the grpc server tls ca cert.
 	GRPCServerTLSCACertEnvVar = "GRPC_SERVER_TLS_CA_CERT"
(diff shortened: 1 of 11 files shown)


qicz (approved):

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -46,16 +46,24 @@ const (
 	GRPCServerUseTLSEnvVar = "GRPC_SERVER_USE_TLS"
 	// GRPCServerTLSCertEnvVar is the grpc server tls cert.
 	GRPCServerTLSCertEnvVar = "GRPC_SERVER_TLS_CERT"
-	// GRPCServerTLSClientCertFilename is the GRPC tls cert file.
-	GRPCServerTLSCertFilename = "/certs/tls.crt"
+	// GRPCTLSCertFilename is the GRPC tls cert file.
+	GRPCTLSCertFilename = "/certs/tls.crt"
```
qicz, line 50: prefer to

oowl: I did this because this cert not only be used by GRPC Service between rate limit services and envoy but also in rate limited service and gateway.

internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -46,16 +46,24 @@ const (
...
 	// GRPCServerTLSKeyEnvVarEnvVar is the grpc server tls key.
 	GRPCServerTLSKeyEnvVarEnvVar = "GRPC_SERVER_TLS_KEY"
-	// GRPCServerTLSKeyFilename is the grpc server key file.
-	GRPCServerTLSKeyFilename = "/certs/tls.key"
+	// GRPCTLSKeyFilename is the grpc server key file.
```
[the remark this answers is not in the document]
oowl: ditto

internal/infrastructure/kubernetes/ratelimit/resource.go
qicz: then move thems to `Kubernetes/resource/resource.go`. The  commom resources and constants have defined here

qicz: IMO, maybe needs different certs.

oowl: https://github.com/envoyproxy/gateway/pull/1499#discussion_r1220097937 
I think we do not need to add another TLS cert, client and server cert are the same in another way

qicz: two parts needs certs
- EG to ratelimit
- ratelimit to redis.server

I don't think we should use the same certs. Because the redis backend is not standard, the users can design it by themselves. 

cc @arkodg

arkodg: @qicz there are 3 connections in play
1) ratelimit to xds server/EG
2) ratelimit to envoy
3) ratelimit to redis

this PR only solves 1, and has solved 2 already, 3 will always be user input

arkodg (approved):
LGTM thanks !

qicz (approved):

zirain (approved):

Issue #865, opened 2023-01-05
Finalize EnvoyProxy


danehans:
When https://github.com/envoyproxy/gateway/pull/861 merges, the EnvoyProxy custom resource can be used to customize the managed proxy infrastructure through `gatewayclass.spec.parametersRef`. Deleting the EnvoyProxy referenced by a managed GatewayClass can result in significant unintended consequences. Should EG finalize an EnvoyProxy referenced by a managed GatewayClass to avoid this potential issue?

AliceProxy:
+1 For finalizing the EnvoyProxy resource after startup.

cnvergence:
I would like to take a look at this one :)

zirain:
@cnvergence assign this to you

AliceProxy:
completed via #1534

PR #1534, opened 2023-06-15
provider: finalize EnvoyProxy referenced by a managed GatewayClass

cnvergence:
**What type of PR is this?**

Add support for envoy proxy finalizers  managed by GatewayClass

Fixes:
- https://github.com/envoyproxy/gateway/issues/865


internal/provider/kubernetes/controller_test.go
```diff
@@ -89,6 +88,115 @@ func TestAddGatewayClassFinalizer(t *testing.T) {
...
+			ep: &egcfgv1a1.EnvoyProxy{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "test-ep",
+					Finalizers: []string{"fooFinalizer"},
+				},
+			},
+			expect: []string{"fooFinalizer"},
+		},
+		{
+			name: "envoyproxy with existing gatewayclass finalizer",
```
arkodg, line 172: isn't this case same as L152 ?

arkodg:
sorry for the delay in review @cnvergence, added some comments, can you please rebase as well, thanks !

internal/provider/kubernetes/controller_test.go
cnvergence: right, I will fix it

internal/provider/kubernetes/controller.go
```diff
@@ -249,7 +249,7 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.
 	}
 
 	// Process the parametersRef of the accepted GatewayClass.
-	if acceptedGC.Spec.ParametersRef != nil && acceptedGC.DeletionTimestamp == nil {
+	if acceptedGC.Spec.ParametersRef != nil {
```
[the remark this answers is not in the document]
cnvergence: I have added the check inside of processParamsRef func to remove finalizer if gatewayClass will be terminated.
https://github.com/envoyproxy/gateway/blob/5797f6731a4846438a1372dfd78ca397ed36f36c/internal/provider/kubernetes/controller.go#L1344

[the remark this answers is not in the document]
arkodg: the issue is at the end of this `if` where it performs a `r.resources.GatewayAPIResources.Store(acceptedGC.Name, resourceTree)`
we dont want to insert a deleted GC into the resource tree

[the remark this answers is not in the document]
cnvergence: This `if` is used only for the `processParamsRef` function or am I missing something there?
```go
if acceptedGC.Spec.ParametersRef != nil {	
   if err := r.processParamsRef(ctx, acceptedGC, resourceTree); err != nil {
       msg := fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err)
       if err := r.gatewayClassUpdater(ctx, acceptedGC, false, string(gwapiv1b1.GatewayClassReasonInvalidParameters), msg); err != nil {
          r.log.Error(err, "unable to update GatewayClass status")
       }
       r.log.Error(err, "failed to process parametersRef for gatewayclass", "name", acceptedGC.Name)
       return reconcile.Result{}, err
   }
}
```
I moved the `acceptedGC.DeletionTimestamp` check inside this function, so if gateway class is being deleted `processParamsRef` will remove the finalizer from EnvoyProxy and return nil (and not add anything to the resourceTree).

arkodg (approved):
LGTM thanks !

internal/provider/kubernetes/controller.go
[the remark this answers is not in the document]
arkodg: yup you're right, this flow is fine, the `acceptedGC` at this point is not deleted since it has been checked earlier when computing `acceptedGC`, thanks for the correction

cnvergence:
@arkodg Yes, we should, but currently if GatewayClass stops referencing a EnvoyProxy, the EnvoyProxy resource will remain on the cluster.
I have opened up a follow-up issue for this during the work on this.
- https://github.com/envoyproxy/gateway/issues/1575

arkodg:
@cnvergence if we merge this PR, it feels like a regression for this case
* user creates EnvoyProxy E-A
* user attaches E-A to GatewayClass G-C, which in turn adds a finalizer on E-A
* user clears E-A from G-C
* since the finalizer is still on E-A, it cannot be deleted by user, which will have a negative impact on the user's experience

cnvergence:
@arkodg I have changed `processParamsRef` function to support this case

cnvergence:
@arkodg understood, I have added this scenario to the integration test [here](https://github.com/envoyproxy/gateway/pull/1534/files#diff-5f29f313b6f74b9836b294d6fe03a909e7e023a42655873eb1a13b316e624e7dR212)
I have modified `processParamsRef` function so that whenever it won't match GatewayClass with EnvoyProxy https://github.com/envoyproxy/gateway/blob/e78e2dbb19cf140650ea38c5d668ba54c794177f/internal/provider/kubernetes/controller.go#L1464
and EnvoyProxy will have finalizers, they will be removed
https://github.com/envoyproxy/gateway/blob/e78e2dbb19cf140650ea38c5d668ba54c794177f/internal/provider/kubernetes/controller.go#L1487

cnvergence:
@arkodg understood, I have added this scenario to the kubernetes_test.go.
I have modified `processParamsRef` function so that whenever it won't match EnvoyProxy with GatewayClass and that EnvoyProxy will have finalizers, they will be removed.
https://github.com/envoyproxy/gateway/blob/e78e2dbb19cf140650ea38c5d668ba54c794177f/internal/provider/kubernetes/controller.go#L1464

internal/provider/kubernetes/controller.go
```diff
@@ -1432,30 +1458,40 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
...
+					ep.Name))
+				return err
+			}
+
 			resourceTree.EnvoyProxy = &ep
 			break
 		}
-	}
 
-	if !found {
```
arkodg, line 1453: why have these been removed ?
these help build status messages that surface in the GatewayClass

internal/provider/kubernetes/controller.go
```diff
@@ -1432,30 +1458,40 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
...
-		return fmt.Errorf("failed to find envoyproxy referenced by gatewayclass: %s", gc.Name)
-	}
-
-	if !valid {
```
arkodg, line 1457: same for this

internal/provider/kubernetes/controller.go
cnvergence: Right, I thought this change was needed and wanted to discuss it.
I have added it again and tested it out, it seems I was wrong earlier.
Thanks for catching this!

internal/provider/kubernetes/controller.go
```diff
@@ -298,15 +298,13 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
 	// Process the parametersRef of the accepted GatewayClass.
-	if acceptedGC.Spec.ParametersRef != nil && acceptedGC.DeletionTimestamp == nil {
-		if err := r.processParamsRef(ctx, acceptedGC, resourceTree); err != nil {
-			msg := fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err)
-			if err := r.gatewayClassUpdater(ctx, acceptedGC, false, string(gwapiv1b1.GatewayClassReasonInvalidParameters), msg); err != nil {
-				r.log.Error(err, "unable to update GatewayClass status")
-			}
-			r.log.Error(err, "failed to process parametersRef for gatewayclass", "name", acceptedGC.Name)
-			return reconcile.Result{}, err
+	if err := r.processParamsRef(ctx, acceptedGC, resourceTree); err != nil {
```
arkodg, line 301: `if acceptedGC.Spec.ParametersRef != nil` is removed from here, can we add it back ?

arkodg (approved):
LGTM thanks @cnvergence !

AliceProxy (approved):

PR #1543, opened 2023-06-18
refactor: optimize dump the EG config logic in debug mode.

qicz:
optimize the dump the EG config logic just needs in debug mode.

cc @Xunzhuo @arkodg

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -57,9 +57,11 @@ func server() error {
 		return err
 	}
 
-	config := spew.NewDefaultConfig()
-	config.DisableMethods = true
-	config.Dump(cfg)
+	if cfg.EnvoyGateway.Admin.Debug {
+		spewConfig := spew.NewDefaultConfig()
+		spewConfig.DisableMethods = true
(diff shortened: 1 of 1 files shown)


zirain (approved):

internal/cmd/server.go
```diff
@@ -57,9 +57,11 @@ func server() error {
 		return err
 	}
 
-	config := spew.NewDefaultConfig()
-	config.DisableMethods = true
-	config.Dump(cfg)
+	if cfg.EnvoyGateway.Admin.Debug {
```
arkodg, line 60:
```suggestion
	if cfg.EnvoyGateway && cfg.EnvoyGateway.Admin && cfg.EnvoyGateway.Admin.Debug {
```

qicz: `Admin` has been initialized in API helper. maybe just init it when configpath="" and Admin is nil

arkodg: Is that's the case then this is fine

arkodg (approved):

PR #1544, opened 2023-06-18
refactor: move envoy proxy logging default logic to api helper.

qicz:
move the EnvoyProxy Logging `Default` logic to helper

cc @zirain


zirain (changes requested):
IMO, we should not put functions like this to API package

qicz:
> IMO, we should not put functions like this to API package

It's part of `ProxyLogging`, just `helper` for this

zirain:
prefer not bring too much OOP style into an golang project if not unnecessaary, should keep api package small 

that's my opinion

PR #1548, opened 2023-06-19
helm: move admission out of crds folder

### LINKED ISSUE #1506: flaky: make kube-deploy fails
I saw these many times, in different cases
https://github.com/envoyproxy/gateway/actions/runs/5225886301/jobs/9436049835

zirain:

this PR introduce a new tool named gtwapi-manifests, which will parse manifests from upstream release.

Fixes #1506

Alternatives: move Gatway API project resources out of EG's chart,users mange resources themself.


charts/gateway-helm/crds/gatewayapi-crds.yaml
```diff
@@ -1,24 +1,3 @@
-# Copyright 2023 The Kubernetes Authors.
```
zirain, line 1: find a way to keep this

tools/src/gtwapi-manifests/pin.go
```diff
@@ -0,0 +1,140 @@
...
+		Factory: testFactory.WithNamespace("default"),
+		Log:     nopLogger,
+	}
+
+	return c
+}
+
+func main() {
+	outputDir := env("OUTPUT_DIR", "charts/gateway-helm")
+	gtwAPIVer := env("GATEWAY_API_VERSION", "v0.7.1")
```
qicz, line 57: auto read current gwapi version?

zirain: it's passed from Makefile

https://github.com/envoyproxy/gateway/blob/76f66f4a37899af358fe294b41f67e6b9dd6a027/tools/make/kube.mk#L38

AliceProxy:
IMO we shouldn't move this stuff into the main chart directory since removing these items on a `helm uninstall` is not desirable and cleaning them up manually when you really want to fully uninstall EG from your cluster is pretty easy. In multitenancy situations where you have multiple instances of EG per-cluster, you wouldn't want these resources getting removed when you `helm uninstall` only one instance so I think it makes sense for them to remain in the `crds` directory.

zirain:
there's two concerns:
1. install CRDs have const timeout(60s) which will cause flaky, see https://github.com/envoyproxy/gateway/pull/1548
2. IIRC, it's not a good pratiace to put namespace resources in an helm chart

what about move these out of EG chart?

AliceProxy:
> IIRC, it's not a good pratiace to put namespace resources in an helm chart

Generally speaking IMO it's better to `helm install example --namespace foo --create-namespace` so that users know what namespace everything is going into and have control over it unless you're using the `default` ns. Putting anything other than CRDs inside Helm's `crds` directory is probably "bad practice" but I can think of many legitimate reasons why you might need to. I think the reason why we have a `Namespace` object is also why we have this stuff in the `crds` directory is so that Helm won't try to install them again if you're running multiple instances per cluster. Likewise you won't want these objects/namespace removed when doing a `helm uninstall`.

> what about move these out of EG chart?

That would probably be better practice than any of the above and I'd be a +1 for that.

arkodg:
Will creating a sub chart for gateway-api help here ?

zirain:
> Will creating a sub chart for gateway-api help here ?

I think that chart should be provided by gateway-api project.

arkodg:
> > Will creating a sub chart for gateway-api help here ?
> 
> I think that chart should be provided by gateway-api project.

theoretically I agree with you :)

arkodg:
@AliceProxy and I discussed this issue in the community meeting today and we were +1 on moving the gateway-api resources into a separate chart with chart values such as
```
gateway-api
  enabled: true
```

or
```
gateway-api:
  crds:
    enabled: true
  admissionServer:
    enabled: true
``` 

this would imply not adding the `crds` in a `/crds` dir so they can be easily opted-in and out during install and uninstall

AliceProxy:
> @AliceProxy and I discussed this issue in the community meeting today and we were +1 on moving the gateway-api resources into a separate chart with chart values such as
> 
> ```
> gateway-api
>   enabled: true
> ```
> 
> or
> 
> ```
> gateway-api:
>   crds:
>     enabled: true
>   admissionServer:
>     enabled: true
> ```
> 
> this would imply not adding the `crds` in a `/crds` dir so they can be easily opted-in and out during install and uninstall

+1, I would prefer two charts (1 for CRDs +  admission server and one for the controller). However, a sub chart with the installation of the admission server and CRDs being opt-in would be an acceptable solution as well, that IMO is better than the current way we're handling this.

zirain:
I'd like close this now, will handle this in v0.6.

PR #1564, opened 2023-06-19
chore: update dependabot

zirain:
reduce noise, see https://github.com/envoyproxy/gateway/pull/1557


.github/dependabot.yml
```diff
@@ -13,6 +13,11 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      # Ignore the following modules as they are updated by k8s.io/kubectl
```
arkodg, line 17: imo, this makes sense if we can prove that all the releases for these modules are ALWAYS triggered together

zirain: we cannot, but in most cases, you don't need to update them immediately after release.

PR #1566, opened 2023-06-20
accesslog: always send k8s.namespace.name and k8s.pod.name

zirain:
With fluent-bit, your can get these informations from [Kubernetes Filter](https://docs.fluentbit.io/manual/pipeline/filters/kubernetes); with OpenTelemetry sink, we need send these manually.

### DIFF
--- a/api/config/v1alpha1/accesslogging_types.go
+++ b/api/config/v1alpha1/accesslogging_types.go
@@ -60,7 +60,8 @@ const (
 	// ProxyAccessLogSinkTypeFile defines the file accesslog sink.
 	ProxyAccessLogSinkTypeFile ProxyAccessLogSinkType = "File"
 	// ProxyAccessLogSinkTypeOpenTelemetry defines the OpenTelemetry accesslog sink.
-	// EnvoyGateway always sends `k8s.namespace.name` and `k8s.pod.name` as attributes.
+	// When the provider is Kubernetes, EnvoyGateway always sends `k8s.namespace.name`
+	// and `k8s.pod.name` as additional attributes.
 	ProxyAccessLogSinkTypeOpenTelemetry ProxyAccessLogSinkType = "OpenTelemetry"
 )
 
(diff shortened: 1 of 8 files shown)


api/config/v1alpha1/accesslogging_types.go
```diff
@@ -60,6 +60,7 @@ const (
 	// ProxyAccessLogSinkTypeFile defines the file accesslog sink.
 	ProxyAccessLogSinkTypeFile ProxyAccessLogSinkType = "File"
 	// ProxyAccessLogSinkTypeOpenTelemetry defines the OpenTelemetry accesslog sink.
+	// EnvoyGateway always sends `k8s.namespace.name` and `k8s.pod.name` as attributes.
```
arkodg, line 63:
```suggestion
	// When the provider is Kubernetes, EnvoyGateway always sends `k8s.namespace.name` and `k8s.pod.name` as additional attributes.
```

internal/xds/translator/accesslog.go
```diff
@@ -171,13 +171,38 @@ func buildXdsAccessLog(al *ir.AccessLog, forListener bool) []*accesslog.AccessLo
...
+	maxLen := len(attributes)
+	if additionalLabels {
+		maxLen += 2
+	}
 	keyValueList := &otlpcommonv1.KeyValueList{
-		Values: make([]*otlpcommonv1.KeyValue, 0, len(attributes)),
+		Values: make([]*otlpcommonv1.KeyValue, 0, maxLen),
+	}
+
+	// always set the k8s namespace and pod name for better UX
```
arkodg, line 189: Im unsure where this should live exactly, here or in the gateway-api layer because this is `provider` specific, but this is fine for now

arkodg (approved):

internal/xds/translator/accesslog.go
zirain: address it with TODO.

zirain: FYI: I'm trying to remove some redunent info in https://github.com/envoyproxy/envoy/pull/28046

Xunzhuo (approved):

PR #1568, opened 2023-06-20
feat(infrastructure/proxy): add readiness probes in envoy proxy deployments

### LINKED ISSUE #1567: Add readinessProbes for envoy proxy deployment
*Description*:
Envoy-gateway should add readiness probes in envoy proxy deployments to allow in-place upgrades(and proxy configurations) without traffic drop. Without any readiness probes, the proxy is immediately marked ready for traffic. This leads to traffic drops in case of in-place upgrades or proxy configurations.

Steps to reproduce:
- Install EG with quickstart yaml and configure multiple replicas using EnvoyProxy configuration:
```
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: config.gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
---
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        replicas: 2
        pod:
          annotations:
            custom: deploy-annotation
```
- Check that pods are ready and traffic is getting routed
- Generate continuous traffic for quickstart backend service. 
- Change the annotations in EnvoyProxy configuration to `custom1: deploy-annotation1`

Expected Behaviour:
No traffic drop as the gateway has replicas, and rolloutStrategy is configured on gateway deployment.

Current Behaviour:
Traffic drops due to pods not being ready to receive traffic.

psinghal20:
**What type of PR is this?**

This PR adds readiness probes in envoy proxy deployments to prevent pods in a not-ready state to receive traffic. 

This allows envoy proxy configurations and in-place upgrades without traffic drops using the configured rollout strategy.

Fixes #1567 

To access the admin APIs through the readiness probe, I had to change the listener admin address to `0.0.0.0` from `127.0.0.1` to allow the envoy to listen on other network interfaces as well. Let me know if this needs to be changed.


internal/xds/bootstrap/bootstrap.go
```diff
@@ -20,7 +20,7 @@ const (
 	// It defaults to the Envoy Gateway Kubernetes service.
 	envoyGatewayXdsServerHost = "envoy-gateway"
 	// envoyAdminAddress is the listening address of the envoy admin interface.
-	envoyAdminAddress = "127.0.0.1"
+	envoyAdminAddress = "0.0.0.0"
```
zirain, line 23: you should not change this for security reason

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -154,6 +159,14 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egcfgv1a1.K
 			VolumeMounts:             expectedContainerVolumeMounts(deploymentConfig),
 			TerminationMessagePolicy: corev1.TerminationMessageReadFile,
 			TerminationMessagePath:   "/dev/termination-log",
+			ReadinessProbe: &corev1.Probe{
+				ProbeHandler: corev1.ProbeHandler{
+					HTTPGet: &corev1.HTTPGetAction{
+						Path: envoyReadinessPath,
```
arkodg, line 165: looks like `Host` option is also available which can be set to `localhost`

zirain: IIRC, that only work for hostnetwork?

internal/xds/bootstrap/bootstrap.go
zirain: yes, maybe use exec, I don't think we should bring thing like pilot-agent now.

psinghal20:
Hi @arkodg @zirain, I have updated the PR. I have opted for a similar approach to emissary-ingress [here](https://github.com/emissary-ingress/emissary/pull/4715/files#diff-b05b9c5111c538de6527cbf67e2069e54d7dc8394fc80e1b54321966ec2b1844). 

I have added a new listener on `0.0.0.0` interface on `19001` port. Let me know if any change is required for the ready port or listener address.

arkodg (approved):
LGTM, thanks !

internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -23,6 +23,33 @@ dynamic_resources:
...
     resource_api_version: V3
 static_resources:
+  listeners:
+  - name: envoy-gateway-proxy-ready-{{ .ReadyServer.Address }}-{{ .ReadyServer.Port }}
+    address:
+      socket_address:
+        address: {{ .ReadyServer.Address }}
+        port_value: {{ .ReadyServer.Port }}
+        protocol: TCP
+    filter_chains:
```
arkodg, line 33: lets also add a text based access log that outputs to stdout https://github.com/envoyproxy/gateway/blob/75e7ada61fe1db9455481f2321309e099f3d8977/internal/xds/translator/listener.go#LL57C19-L57C19

arkodg: as well as the per connection setting https://github.com/envoyproxy/gateway/blob/75e7ada61fe1db9455481f2321309e099f3d8977/internal/xds/translator/listener.go#L61

zirain: that will be too noise

arkodg: @psinghal20 can you incorporate the `PerConnectionBufferLimitBytes` setting into the config, to reduce buffer space, tia !

internal/xds/bootstrap/bootstrap.go
```diff
@@ -28,6 +28,10 @@ const (
 
 	// DefaultXdsServerPort is the default listening port of the xds-server.
 	DefaultXdsServerPort = 18000
+
+	envoyReadinessAddress = "0.0.0.0"
```
LanceEa, line 32: @arkodg - I'm not sure it matters for this PR but does EG at this point claim to work in an IPV6 only environment?

If so then I don't believe `0.0.0.0` would work and I think its something like `::` for IPV6 (_I'm definitely not an IPv6 expert so could be wrong_).

Maybe we just need to capture it as a separate issue but thought I would call it out just in case.

LanceEa (commented):
small comment, other than that it looks good to me.

internal/xds/bootstrap/bootstrap.go
zirain: FYI: https://github.com/envoyproxy/gateway/issues/184

LanceEa: @zirain - cool thanks for pointing out! It looks my comment is already captured so no worries from me.

zirain (approved):

qicz (approved):
LGTM

Xunzhuo (approved):

PR #1576, opened 2023-06-22
chore: rm .golangci.yml file in root dir

arkodg:
`tools/linter/golangci-lint/.golangci.yml` is used as the config file for `golangci-lint`
```
===========> Running lint.golint ...
tools/bin/golangci-lint run  --build-tags=e2e --config=tools/linter/golangci-lint/.golangci.yml
```


zirain (dismissed):

LanceEa:
@arkodg - +1 on making sure there is only a single source of truth.

Do we have any where that we outline developer tool setup? It might be good to add the fact that it is no longer at the root and guide folks to make sure they configure their IDE.

As far as I can tell, IDE's like VSCode would assume it is in the root and if not it would probably require setting the `-c` flag to the lint flags that points at the `tools/linter/golangci-lint/.golangci.yml` file.

https://golangci-lint.run/usage/integrations/#go-for-visual-studio-code

arkodg:
> @arkodg - +1 on making sure there is only a single source of truth.
> 
> Do we have any where that we outline developer tool setup? It might be good to add the fact that it is no longer at the root and guide folks to make sure they configure their IDE.
> 
> As far as I can tell, IDE's like VSCode would assume it is in the root and if not it would probably require setting the `-c` flag to the lint flags that points at the `tools/linter/golangci-lint/.golangci.yml` file.
> 
> https://golangci-lint.run/usage/integrations/#go-for-visual-studio-code

hey @LanceEa updated the docs with the actual `golangci-lint` config file, ptal

zirain (approved):

LanceEa (approved):
Lgtm, thanks!

Issue #1577, opened 2023-06-23
Deployment of ratelimit hard code is LoadBalancer


quyenhoang96:
Deployment of ratelimit hard code is LoadBalancer
But I want to change its type of Service. How to configure it ?

zirain:
@qicz this seems related to reusage of `DefaultKubernetesServiceType`, can you take a look?

zirain:
Thanks for catching this

cnvergence:
I can take a look at this one

cnvergence:
As per https://github.com/envoyproxy/gateway/pull/1578#discussion_r1239995329, I will change the default value to Cluster.

zirain:
fixed by https://github.com/envoyproxy/gateway/pull/1578

PR #1578, opened 2023-06-23
ratelimit: Change rate limit service type to cluster

cnvergence:

Fixes 
- https://github.com/envoyproxy/gateway/issues/1577


internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -54,8 +54,11 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 	}
 
 	labels := rateLimitLabels()
+	if r.rateLimitDeployment.ServiceType == nil {
+		r.rateLimitDeployment.ServiceType = egcfgv1a1.DefaultKubernetesServiceType()
+	}
 
-	serviceSpec := resource.ExpectedServiceSpec(egcfgv1a1.DefaultKubernetesServiceType())
+	serviceSpec := resource.ExpectedServiceSpec(egcfgv1a1.GetKubernetesServiceType(*r.rateLimitDeployment.ServiceType))
```
arkodg, line 61: I dont think we need to define an API here, should be okay to hardcode to `Cluster`
the Rate Limit Service is internal to the Envoys living in the same cluster (and same namespace), so imo there is no need
to expose it to clients outside the cluster using `NodePort` or `LoadBalancer`

cnvergence: This makes sense to me, I will write in the original issue that there will be just a change of the default setting.

arkodg (changes requested):
this doesn't look right, we dont want to change the default API value, which is used for the Envoy (proxy) service, which should default to LoadBalancer.
we need to somehow change the service type within the rate limit infra code to always be set to `Cluster`

arkodg (approved):
LGTM thanks, would be good if you could also add a comment highlighting the "why" - its internal to the envoys running in the cluster

zirain (approved):

PR #1579, opened 2023-06-23
Fix proto tooling

arkodg:
* Add `protoc-gen-go-grpc` & `protoc-gen-go` dep to `make protos`

* init go mod for `protoc-gen-go`


tools/src/protoc-gen-go/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/envoyproxy/gateway/tools/src/protoc-gen-go
```
arkodg, line 1: hey @AliceProxy any idea why this wasn't added in the orig PR
https://github.com/envoyproxy/gateway/pull/1209/

arkodg: cc @haq204

AliceProxy: Huh... I remember seeing that when I was messing around with stuff... maybe I forgot to push it or something when I was trying to figure out a cleaner way to handle the buf files?

AliceProxy (dismissed):

ZackButcher:
@arkodg I think you should also update `buf.gen.yaml` in this PR to output the `go` files in a relative dir as well:
```yaml
version: v1
plugins:
  - plugin: go
    out: ./
    path: tools/bin/protoc-gen-go
    opt: paths=source_relative
  - plugin: go-grpc
    out: ./
    path: tools/bin/protoc-gen-go-grpc
    opt: paths=source_relative
```

Right now it's only set for gRPC.

tools/src/protoc-gen-go/go.mod
arkodg: cool thanks for confirming !

arkodg:
> @arkodg I think you should also update `buf.gen.yaml` in this PR to output the `go` files in a relative dir as well:
> 
> ```yaml
> version: v1
> plugins:
>   - plugin: go
>     out: ./
>     path: tools/bin/protoc-gen-go
>     opt: paths=source_relative
>   - plugin: go-grpc
>     out: ./
>     path: tools/bin/protoc-gen-go-grpc
>     opt: paths=source_relative
> ```
> 
> Right now it's only set for gRPC.

thanks @ZackButcher , done

AliceProxy (approved):

zirain (approved):

PR #1580, opened 2023-06-24
Observability: init tracing API

zirain:
part of https://github.com/envoyproxy/gateway/issues/699


api/config/v1alpha1/tracing_types.go
```diff
@@ -0,0 +1,88 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+type ProxyTracing struct {
+	// RandomSamplingPercentage controls the rate at which traffic will be
+	// selected for tracing if no prior sampling decision has been made.
+	// Defaults to 0, valid values [0-100]. 100 indicates 100% sampling.
```
arkodg, line 11: should this default to 100 instead ?

api/config/v1alpha1/tracing_types.go
```diff
@@ -0,0 +1,88 @@
...
+	// CustomTags defines the custom tags to add to each span.
+	// If provider is kubernetes, pod name and namespace are added by default.
+	CustomTags map[string]CustomTag `json:"customTags,omitempty"`
+	// Provider defines the tracing provider.
+	// Only OpenTelemetry is supported currently.
+	Provider TracingProvider `json:"provider"`
+}
+
+type TracingProvider struct {
+	// Host define the provider service hostname.
```
arkodg, line 25: needs another level here with `Type: OpenTelemetry`, so other types can be added in the future

api/config/v1alpha1/tracing_types.go
```diff
@@ -0,0 +1,88 @@
...
+package v1alpha1
+
+type ProxyTracing struct {
+	// RandomSamplingPercentage controls the rate at which traffic will be
+	// selected for tracing if no prior sampling decision has been made.
+	// Defaults to 0, valid values [0-100]. 100 indicates 100% sampling.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	// +optional
+	RandomSamplingPercentage *uint32 `json:"randomSamplingPercentage,omitempty"`
```
arkodg, line 15: thoughts on calling it `samplingRate` ?

api/config/v1alpha1/tracing_types.go
zirain: I think we make an agreement on the WG meeting about just support OpenTelemetry.

api/config/v1alpha1/tracing_types.go
zirain: Actually, tracing may cause performance issue, particularly on edge proxy with high load, that's why I prefer to disabled it by default.
 @envoyproxy/gateway-maintainers WDYT?

api/config/v1alpha1/tracing_types.go
zirain: make sense

api/config/v1alpha1/tracing_types.go
arkodg: since tracing is disabled by default, I think defaulting sampling rate to 100 is okay when tracing is enabled, imo

api/config/v1alpha1/tracing_types.go
arkodg: I do agree that starting off with `OpenTelemetry` is a good idea, however we could leave room for expansion

api/config/v1alpha1/tracing_types.go
AliceProxy: imo 100 by default is the least confusing. Just need to make sure we call this out in the docs in a way that is very visible so that users can turn this down if they want to.

AliceProxy (dismissed):

api/config/v1alpha1/tracing_types.go
zirain: yes, I will update the default value and add a note in the doc.

api/config/v1alpha1/tracing_types.go
zirain: updated

zirain: all envoy support tracing backend require `host` and `port`, it's not as same as accesslog.

arkodg:
```
type: OpenTelemetry
openTelemetry:
  host:
  port:
  ```
 adds room for future expansion for open telemetry fields that are exposed by envoy in https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/opentelemetry.proto.html#envoy-v3-api-msg-config-trace-v3-opentelemetryconfig

Comment (arkodg) [reply api/config/v1alpha1/tracing_types.go:37]: related to this, an improvement in the API can be to replace all the `Host` & `Port` fields with an embedded `ExternalService` struct which can be defined once and used everywhere and we can include additional fields such as `tls` within it

Comment (zirain) [reply api/config/v1alpha1/tracing_types.go:37]: this's a trade off between:
```
type: OpenTelemetry
openTelemetry:
  host:
  port:
zipkin:
  host:
  port:
```

vs

```
type: OpenTelemetry
host:
port:
openTelemetry:
  otelXXXX:
zipkin:
  zipikinXXX:
```

Comment (arkodg) [reply api/config/v1alpha1/tracing_types.go:37]: thanks for raising this, ptal @envoyproxy/gateway-maintainers

Comment (AliceProxy) [reply api/config/v1alpha1/tracing_types.go:37]: I prefer this as it currently is:

```
type: OpenTelemetry/Zipkin
host:
port:
```

Fields that are shared by more than one subtype can be top level, fields specific to the type such as zipkin/otel go under their structs. If we're currently only adding support for host/port which are shared by both zipkin/otel then I'd vote that we accept this as it currently is and then when we do want to introduce fields specific to one type like @arkodg mentioned, then we will add the `openTelemetry/etc.` sub object at that time like @zirain's example:

```
type: OpenTelemetry
host: 
port:
openTelemetry:
  otelXXXX:
zipkin:
  zipikinXXX:
```

arkodg (approved):
LGTM thanks

AliceProxy (approved):

PR #1584, opened 2023-06-25
chore: fix ci publish order

### LINKED ISSUE #1583: CI: publish should be after both conformance and e2e tests
*Description*:
returned, etc.

Before:

image

After:

image

*Repro steps*:

sharing.

*Environment*:

*Logs*:

Xunzhuo:
**What type of PR is this?**

Fixes #1583


zirain (approved):

qicz (approved):
LGTM

PR #1592, opened 2023-06-26
feat: add testdata to egctl/gwapi/xds lib

Xunzhuo:
**What type of PR is this?**

feat: add testdata to egctl/gwapi/xds lib

introduce `make testdata` to manage tons of json/yaml files of eg.


tools/make/golang.mk
```diff
@@ -48,7 +48,10 @@ go.test.unit: ## Run go unit tests
 go.testdata.complete: ## Override test ouputdata
 	go test -timeout 30s -run ^TestTranslateXds github.com/envoyproxy/gateway/internal/xds/translator --override-testdata=true
 	go test -timeout 30s -run ^TestTranslate github.com/envoyproxy/gateway/internal/cmd/egctl --override-testdata=true
-
+	go test -timeout 30s -run ^TestExtractSubResourcesConfigDump github.com/envoyproxy/gateway/internal/cmd/egctl --override-testdata=true
```
arkodg, line 51: this makes the target faster but also requires the developer to remember to add the Test name in here, can we just run this on all tests ?

internal/cmd/egctl/testdata/config/out/out.all.json
```diff
@@ -1,1513 +1,2316 @@
 {
-	"default": {
```
arkodg, line 2: why are these deleted

internal/cmd/egctl/config_test.go
```diff
@@ -251,6 +260,18 @@ func readOutputConfig(filename string) ([]byte, error) {
 	return b, nil
 }
 
+func overrideOutputConfig(t *testing.T, data string, name ...string) {
```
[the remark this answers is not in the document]
Xunzhuo: Done : )

tools/make/golang.mk
Xunzhuo: Yep, that is right.

internal/cmd/egctl/testdata/config/out/out.all.json
Xunzhuo: It is not deleted, just change the indent, we write it by hand before, now it is overidden by codes.

internal/ir/xds.go
```diff
@@ -49,19 +49,19 @@ var (
 // +k8s:deepcopy-gen=true
 type Xds struct {
 	// AccessLog configuration for the gateway.
-	AccessLog *AccessLog
```
arkodg, line 52: needs `omitempty` tags as well since they show up in the o/p as `null` atm

Xunzhuo: Done

Xunzhuo:
Some additional Updates:

1. Set fixed `LastTransitionTime` with the first commit time, otherwise each time we run `make testdata` will update over 100 files the `LastTransitionTime` with the current time, the changes should not be expected:
```
commit b86296d61439d115b288f01979e82865c8621f44
Author: Matt Klein <mattklein123@gmail.com>
Date:   Mon Apr 11 16:34:39 2022 -0600

    Initial commit
```
2. Add `make testdata` to `gen-check` in CI, so it will check if PR changes IR and test configs are updated with `make testdata` correctly.

cc @arkodg

arkodg:
@Xunzhuo thoughts on setting any timestamp to `null` instead which is easier to comprehend, debug and fix in the testdata by the user, in case they are handwriting test files instead of using `make testdata`

arkodg:
also if we add it to CI, it might cause a failure when a user adds the right content, in wrong order (for fields)

Xunzhuo:
> @Xunzhuo thoughts on setting any timestamp to `null` instead which is easier to comprehend, debug and fix in the testdata by the user, in case they are handwriting test files instead of using `make testdata` ?

Right, I have updated it :)

Xunzhuo:
> also if we add it to CI, it might cause a failure when a user adds the right content (who is unaware of `make testdata`) in the testdata yaml, in wrong order (for fields)

I think we should always suggest devs using make testdata instead of handwriting. Handwriting codes is a nightmare for everyone I think, and even if content is correct, but handwriting cannot promise format or indent is correct in json/yaml.

arkodg (approved):
LGTM thanks @Xunzhuo !
this should increase velocity of development by automating adding testdata with `make testdata`

qicz (commented):
LGTM

Issue #703, opened 2022-11-06
HorizontalPodAutoscaler for envoy proxy


gitanuj:
*Description*:
Does envoy gateway have any design to scale out envoy proxy pods? Are there any best practices on how to handle increased number of requests? I'm mostly looking at deploying this to clouds like AWS/GCP/Azure which will allocate application/network Load Balancers but then how can I make sure that envoy proxy pod is not the bottleneck in the gateway and can scale out when needed.

danehans:
HPA is currently not supported by Envoy Gateway. It's a feature that would be beneficial to the project. I have added the feature to the backlog. Let me know if you're interested in working on it and I can assign it to you.

> I'm mostly looking at deploying this to clouds like AWS/GCP/Azure which will allocate application/network Load Balancers...

Envoy Gateway currently creates an ELB when running in AWS.  #648 intends to design an API for managing the Envoy service, PTAL, and comment as needed.

gitanuj:
1. Is there any design doc explaining which pods are involved in the data path for EG? I'm trying to understand if there are any other components other that EnvoyProxy which need to be scaled out as well.
2. Before implementing HPA I wanted to understand if I can simply update the number of replicas [here](https://github.com/envoyproxy/gateway/blob/7c6c37ad72a6fb2403f7ea85b94732c0e14906f7/internal/infrastructure/kubernetes/deployment.go#L131). Is there an API through which this can be exposed?

danehans:
>Is there any design doc explaining which pods are involved in the data path for EG? I'm trying to understand if there are any other components other than EnvoyProxy which need to be scaled out as well.

Only the Envoy proxy deployment needs to be scaled. The [design docs](https://gateway.envoyproxy.io/v0.2.0/design_docs.html) are the best place for this type of info, specifically the system design and config API design docs.

> Before implementing HPA I wanted to understand if I can simply update the number of replicas [here](https://github.com/envoyproxy/gateway/blob/7c6c37ad72a6fb2403f7ea85b94732c0e14906f7/internal/infrastructure/kubernetes/deployment.go#L131). Is there an API through which this can be exposed?

You can update this field but EG will revert it back to the desired state. This field should be configurable using the [EnvoyProxy config API](https://github.com/envoyproxy/gateway/blob/v0.2.0/api/config/v1alpha1/envoyproxy_types.go). If you would like to make this field configurable, please create an issue that describes your use case and xref this issue.

arkodg:
[...]

PR #1594, opened 2023-06-26
refactor: set defaults in Deployment, else k8s sets them for you, creating infinite reconciliation loop

### LINKED ISSUE #1590: Proxy always terminating
*Description*:
>after install eg and proxy, when I add or delete a httproute, proxy will new pods then terminating every time, as shown

*Environment*:
>the latest version

*Logs*:
>
image

qicz:
fixes the envoy proxy resources update bug, the resourceProvider's provides resources have more different fields from applied.

Fixes 
- #1590


arkodg:
@qicz can you please elaborate on the bug

qicz:
> @qicz can you please elaborate on the bug ?

the `Controller` watch more `Resource`, when some `Resource` changed the `r.InfraIR`  will Store new `Infra` https://github.com/envoyproxy/gateway/blob/4e6786a5dc9103fc155ad24b340ba43bc5fdb537/internal/gatewayapi/runner/runner.go#L91-L97

and the `r.InfraIR` will `Subscribe` the change and  do `CreateOrUpdateProxyInfra`
https://github.com/envoyproxy/gateway/blob/4e6786a5dc9103fc155ad24b340ba43bc5fdb537/internal/infrastructure/runner/runner.go#L56-L75

https://github.com/envoyproxy/gateway/blob/4e6786a5dc9103fc155ad24b340ba43bc5fdb537/internal/infrastructure/kubernetes/infra_resource.go#L63-L78

when creating the deployment, some filed will fill default values like `DefaultMode` 420, `DeprecatedServiceAccount` the sa name, `RevisionHistoryLimit` &10, `ProgressDeadlineSeconds` &600, and the `HPA` will change the `Replicas` when occurring the AutoScaler. we should align these fields to ensure the `updateChecker()` works right, otherwise will update the deployment all the time. 

https://github.com/envoyproxy/gateway/blob/328a7dc95624c4dc7b58e8172e1c665bd5b53635/internal/infrastructure/kubernetes/infra_resource.go#L75-L84

https://github.com/envoyproxy/gateway/blob/328a7dc95624c4dc7b58e8172e1c665bd5b53635/internal/infrastructure/kubernetes/infra_client.go#L39-L45

image

qicz:
maybe we should watch the `Resource` by per controller @arkodg

qicz:
@zirain the e2e needs review, sometime works fine on v1.27.0, or v1.25.8 or v1.26.3 but can not works fine all.

zirain:
> @zirain the e2e needs review, sometime works fine on v1.27.0, or v1.25.8 or v1.26.3 but can not works fine all.

see https://github.com/envoyproxy/gateway/pull/1599

qicz:
> should the logic be reverted back to update only the fields that EG cares about instead of updating entire object ? or should we skip comparison of some fields (that are changed by other controllers) ?

maybe needs a controller for `envoyproxy`, then we just needs compare enovyproxy. 

At present, we only compare the fields that EG cares about or ignore some fields for subsequent feature development, which is easy to cause comparative logic omissions.

At the same time, when the HPA is triggered, it will also be affected, for example, the original number of replicas is 2, the HPA is changed to 3, at this time we modify the `envoyproxy`, changing it to 4 will not take effect.  We needs compare the hard deployment and k8s `current` deployment replicas when hard > `current` fire updating.

internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -73,6 +73,14 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, r ResourceRender)
 	}
 
 	return i.Client.CreateOrUpdate(ctx, key, current, deployment, func() bool {
+		// applied to k8s the "DeprecatedServiceAccount" will fill it.
+		deployment.Spec.Template.Spec.DeprecatedServiceAccount = current.Spec.Template.Spec.DeprecatedServiceAccount
+		// applied to k8s the "SecurityContext" will fill it with default settings.
```
arkodg, line 78: aren't we just overriding security context here, so the user defined values will be lost ?

internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -73,6 +73,14 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, r ResourceRender)
...
+		deployment.Spec.Template.Spec.SecurityContext = current.Spec.Template.Spec.SecurityContext
+		// adapter the hpa updating and envoyproxy updating.
+		if *deployment.Spec.Replicas < *current.Spec.Replicas {
```
arkodg, line 81: thanks for raising this issue, imo we should either
* skip comparing replica if unset (lets make sure its a pointer in the API) and highlight in HPA docs to do so
* if replica is set in our API, we recommend not using HPA, and use that value for comparison

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -158,10 +158,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egcfgv1a1.K
...
 				ProbeHandler: corev1.ProbeHandler{
 					HTTPGet: &corev1.HTTPGetAction{
-						Path: bootstrap.EnvoyReadinessPath,
-						Port: intstr.IntOrString{Type: intstr.Int, IntVal: bootstrap.EnvoyReadinessPort},
+						Path:   bootstrap.EnvoyReadinessPath,
+						Port:   intstr.IntOrString{Type: intstr.Int, IntVal: bootstrap.EnvoyReadinessPort},
+						Scheme: corev1.URISchemeHTTP,
 					},
 				},
+				TimeoutSeconds:   1,
```
arkodg, line 166: any reason why this is being set, can this be moved into a separate PR, tia

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -203,6 +203,8 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 					Volumes:                       expectedDeploymentVolumes(r.infra.Name, deploymentConfig),
 				},
 			},
+			RevisionHistoryLimit:    pointer.Int32(10),
```
arkodg, line 206: why are these field being set ?

internal/infrastructure/kubernetes/infra_resource.go
qicz: > skip comparing replica if unset (lets make sure its a pointer in the API) and highlight in HPA docs to do so

unset the k8s will fill the default value "1", then `current` `replicas` is not equal to `deployment` will fire an update.

> if the replica is set in our API, we recommend not using HPA, and using that value for comparison

IMO, we can compare the replicas from envoyporoxy or k8s like the current compare

internal/infrastructure/kubernetes/proxy/resource.go
qicz: this belongs to this fix.

the `TimeoutSecounds` will fill the default value, and compare the `current` from k8s and `deployment` from code, it never equal forever will fire current bug : more terminating pod and deployment generation will increment all the time

internal/infrastructure/kubernetes/proxy/resource_provider.go
qicz: ditto

internal/infrastructure/kubernetes/proxy/resource.go
qicz: image

qicz:
apply the service account resource has a bug with the same situation, fixes #1604

zirain:
can you simply this PR with `go-cmp`, ignore fileds EG don't care.

internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -75,7 +75,7 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, r ResourceRender)
 	}
 
 	return i.Client.CreateOrUpdate(ctx, key, current, deployment, func() bool {
-		return !reflect.DeepEqual(deployment.Spec, current.Spec)
+		return !resource.CompareDeployment(current, deployment)
```
arkodg, line 78: why are we coming non `Spec` fields here ?

internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -56,7 +56,7 @@ func (i *Infra) createOrUpdateConfigMap(ctx context.Context, r ResourceRender) e
 	}
 
 	return i.Client.CreateOrUpdate(ctx, key, current, cm, func() bool {
-		return !reflect.DeepEqual(cm.Data, current.Data)
+		return !reflect.DeepEqual(cm.Data, current.Data) || !reflect.DeepEqual(cm.BinaryData, current.BinaryData)
```
arkodg, line 59: why do we need to compare `BinaryData` here ?

internal/infrastructure/kubernetes/resource/resource.go
```diff
@@ -41,6 +44,24 @@ func CompareSvc(currentSvc, originalSvc *corev1.Service) bool {
 		cmpopts.IgnoreFields(corev1.ServiceSpec{}, "ClusterIP", "ClusterIPs"))
 }
 
+// CompareDeployment compare the current from the k8s and deployment from the resource_provider.
+func CompareDeployment(current, deployment *appv1.Deployment) bool {
+	// applied to k8s the "DeprecatedServiceAccount" will fill it.
+	deployment.Spec.Template.Spec.DeprecatedServiceAccount = current.Spec.Template.Spec.DeprecatedServiceAccount
```
arkodg, line 50: why not ignore this field using `cmp.Diff` ??

arkodg:
hey @qicz seeing a lot of churn in this library, thoughts on refactoring this lib to do the following
1. create resource if it doesnt exist
2. if exists, taking the `current` (in k8s API server), and modify some of the field based on `EnvoyProxy` settings ? and update `current` . Today we are trying to update based on `proposed` which is causing issues imo

qicz:
> hey @qicz seeing a lot of churn in this library, thoughts on refactoring this lib to do the following
> 
> 1. create resource if it doesnt exist
> 2. if exists, taking the `current` (in k8s API server), and modify some of the field based on `EnvoyProxy` settings ? and update `current` . Today we are trying to update based on `proposed` which is causing issues imo

yes, if exists, modify some of the field based on `deployment` settings from `EnvoyProxy` settings .

arkodg:
> > hey @qicz seeing a lot of churn in this library, thoughts on refactoring this lib to do the following
> > 
> > 1. create resource if it doesnt exist
> > 2. if exists, taking the `current` (in k8s API server), and modify some of the field based on `EnvoyProxy` settings ? and update `current` . Today we are trying to update based on `proposed` which is causing issues imo
> 
> yes, if exists, modify some of the field based on `deployment` settings from `EnvoyProxy` settings .

yes, today we compare all fields which is causing this issue

internal/infrastructure/kubernetes/resource/resource.go
qicz: too many fields need to ignore, and when new feature needs to be paid attention to here, and In order to maintain consistency with the following

internal/infrastructure/kubernetes/infra_resource.go
qicz: Users can set the cm with binary data

qicz:
> > > hey @qicz seeing a lot of churn in this library, thoughts on refactoring this lib to do the following
> > > 
> > > 1. create resource if it doesnt exist
> > > 2. if exists, taking the `current` (in k8s API server), and modify some of the field based on `EnvoyProxy` settings ? and update `current` . Today we are trying to update based on `proposed` which is causing issues imo
> > 
> > 
> > yes, if exists, modify some of the field based on `deployment` settings from `EnvoyProxy` settings .
> 
> yes, today we compare all fields which is causing this issue

So what should I do? IMO i am compared the fields from deployment from envoyproxy setting s and some fields with Default value from k8s.

arkodg:
hi @qicz this changes in this PR doesnt quite look right to me.
@AliceProxy @kflynn and I discussed this issue in the community meeting today and we prefer the following
* resource like `Deployment` generated by EG is the source of truth. If another user changes it at runtime, EG will eventually reconcile and override those changes
* above implies users should not make changes to above resource, if they want to do so, the project must support a seperate static deployment of `Deployment` that EG does not generate or reconcile tracked with https://github.com/envoyproxy/gateway/issues/1379
* this means all intent should be generated by EG including HPA


Can you highlight the specific issue here - is OpenShift editing the Deployment spec, if so, is there any way to ask the platform to not do so ?

qicz:
> hi @qicz this changes in this PR doesnt quite look right to me. @AliceProxy @kflynn and I discussed this issue in the community meeting today and we prefer the following
> 
> * resource like `Deployment` generated by EG is the source of truth. If another user changes it at runtime, EG will eventually reconcile and override those changes
> * above implies users should not make changes to above resource, if they want to do so, the project must support a seperate static deployment of `Deployment` that EG does not generate or reconcile tracked with [Support User Provisioned Envoy Proxy fleet #1379](https://github.com/envoyproxy/gateway/issues/1379)
> * this means all intent should be generated by EG including HPA
> 
> Can you highlight the specific issue here - is OpenShift editing the Deployment spec, if so, is there any way to ask the platform to not do so ?


Maybe, I need to reorganize the description.

First of all, for the update of envoy deployment, the data source of deployment is two parts, one part source is the change of EnvoyProxy resource, and the other is EG based on resource_provider, which is actually the structure of k8s API, when based on API construction, many default values, if not provided, then after the deployment is created, the next time you go to controller tuning, it will start the comparison between the existing deployment in k8s and the deployment constructed by EG, and at this time, the deployment in k8s has already filled in the default values, but not in EG, so the comparison is bound to be unequal. The deployment will be updated again, but in fact, the difference between the two is that the deployment constructed by EG does not carry the default value field that should be modified, which is why this PR modifies the value of the default field. Because the replicas field will be modified by HPA, special processing is done for replicas when comparing deployments.

In addition, why the comparison of binary data is added to the comparison of configmap, because configmap not only provides the configuration method based on map[string]string, but also supports the method of map[string]byte[], and users can configure it in this way. We think we need to compare such possibilities in this way.

In general, in the first case, not only on openshift, but also on k8s, we need to set these default values. At the same time, when there are new features related to deployment changes later, it is also necessary to pay attention to whether there are new default values that need to be set in the comparison place, otherwise, it will still cause inconsistencies between the deployments that already exist in k8s and the EG structure.

@arkodg @AliceProxy @kflynn

arkodg:
* for hpa, since EG is the source of truth, the hpa resource should be generated by EG by exposing it as an API field in `EnvoyProxy` resource and generating it in the infra layer (https://github.com/envoyproxy/gateway/issues/703)
* for ConfigMap BinaryData, if EG doesnt generate BinaryData, we dont need to check it, other controllers shouldnt be editing EG's resources

qicz:
> 

- cm binary data cmp logic has been removed
- expose hpa，maybe raise a new PR？

qicz:
> * for hpa, since EG is the source of truth, the hpa resource should be generated by EG by exposing it as an API field in `EnvoyProxy` resource and generating it in the infra layer ([docs: HorizontalPodAutoscaler for envoy proxy #703](https://github.com/envoyproxy/gateway/issues/703))
> * for ConfigMap BinaryData, if EG doesnt generate BinaryData, we dont need to check it, other controllers shouldnt be editing EG's resources

@arkodg 

- cm binary data cmp logic has been removed
- expose hpa， maybe raise a new PR？

arkodg:
sg thanks @qicz 
from this PR I think `CompareDeployment` should be removed/reverted

qicz:
@arkodg ptal, has been updated by offline communication

arkodg (dismissed):
LGTM thanks !

zirain (approved):

zirain:
cc @arkodg need cherrypick to release-0.5?

Xunzhuo (approved):
Good catch and thanks for digging into the problem : )

Xunzhuo:
https://github.com/envoyproxy/gateway/actions/runs/5675991681/job/15382049734 
cc @zirain

zirain:
have no idea about the fail, maybe something underlying changed?

PR #1596, opened 2023-06-26
refactor: optimize client-go performance by protobuf context-type.

qicz:
**What type of PR is this?**
using `protobuf` context-type to optimize client-go performance.
image

using the `json` needs more memory and time
image

### DIFF
--- a/internal/infrastructure/manager.go
+++ b/internal/infrastructure/manager.go
@@ -10,13 +10,13 @@ import (
 	"fmt"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


zirain:
please guard this with an env/cmd args, make it avaliable to fallback.

qicz:
> please guard this with an env/cmd args, make it avaliable to fallback.

I will do this PR at next version

arkodg:
@qicz is this still needed ?

qicz:
xref: https://github.com/envoyproxy/gateway/pull/4282#issuecomment-2378285969

PR #1597, opened 2023-06-26
Fix `make conformance` on Mac

arkodg:
* Use `buildx` for all builds

* Adds the `--load` flag to `docker buildx build` to save
the image in the local image store.


qicz (approved):

zirain (approved):
will this work on m1/m2?

Issue #1142, opened 2023-03-13
Support GRPCRoute and HTTPRoute on the same Gateway listener


arkodg:
*Description*:

* Do we need to support GRPCRoute and HTTPRoute attached the same Gateway Listener ?

If we do we need to refactor the [xDS HCM Logic](https://github.com/envoyproxy/gateway/blob/f219904808675376a5b1ba445cd08b01a36b3a73/internal/xds/translator/listener.go#L99) that changes the codec type as well as includes web socket options based on whether the listener is linked to HTTP2 or not

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

muyuan0:
> Do we need to support GRPCRoute and HTTPRoute attached the same Gateway Listener ?

There are a few real-world cases of this. For example, etcd will expose port 2379 for both gRPC and HTTP API.

arkodg:
thanks for pointing that out @muyuan0, can you also link this to the etcd docs as well

arkodg:
great thanks for sharing these links @muyuan0, we can prioritize this issue once this becomes more relevant for users

quyenhoang96:
> 

Do you mean itself detect protocol to upgrade ?

muyuan0:
Yes. Envoy will do auto detection if the codeType is AUTO, see

https://github.com/envoyproxy/envoy/blob/6c702f45fadf4abd2f24605e3d9e81c748b13f85/source/common/http/conn_manager_utility.cc#L56-L76

PR #1605, opened 2023-06-29
Always set codec type to AUTO in the xds listener

arkodg:
There is no good reason to pin to HTTP2 and disable HTTP1.1 when a GRPCRoute and HTTPRoute are attached to the same listener.

More info on codec type in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto

Fixes: https://github.com/envoyproxy/gateway/issues/1142

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -110,9 +110,6 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	}
 
 	if irListener.IsHTTP2 {
-		// Set codec to HTTP2
-		mgr.CodecType = hcmv3.HttpConnectionManager_HTTP2
-
 		mgr.HttpFilters = append(mgr.HttpFilters, xdsfilters.GRPCWeb)
 		// always enable grpc stats filter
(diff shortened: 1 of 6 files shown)


zirain (approved):
LGTM

internal/xds/translator/listener.go
```diff
@@ -110,9 +110,6 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	}
 
 	if irListener.IsHTTP2 {
```
zhaohuabing, line 112: Do we also need to remove ```IsHTTP2``` flag in the irListener? 

Since a listener can have both grpcroutes and httproutes, a listener should not have a protocol property.

arkodg: its still used to attach proto specific filters such as `grpc-web`, which we'd ideally not like to include by default

zhaohuabing: Makes sense.

Maybe rename it to HasHTTP2Route to make the name more precise.

zhaohuabing (approved):

PR #1606, opened 2023-06-29
Implement JSON Patch in Xds Translator

arkodg:


arkodg:
trying to fight one last issue wrt unmarshalling
```
address:
  socket_address:
    address: 0.0.0.0
    port_value: 10080
default_filter_chain:
  filters:
    typed_config:
      '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
      common_http_protocol_options:
        headers_with_underscores_action: REJECT_REQUEST
      http_filters:
      - |
        name: "envoy.filters.http.ratelimit"
        typed_config:
          "@type": "type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit"
          domain: "eg-ratelimit"
          failure_mode_deny: true
          timeout: 1s
          rate_limit_service:
            grpc_service:
              envoy_grpc:
                cluster_name: rate-limit-cluster
            transport_api_version: V3'
      - name: envoy.filters.http.router
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
      http2_protocol_options:
        initial_connection_window_size: 1048576
        initial_stream_window_size: 65536
        max_concurrent_streams: 100
      merge_slashes: true
      normalize_path: true
      path_with_escaped_slashes_action: UNESCAPE_AND_REDIRECT
      rds:
        config_source:
          ads: {}
          resource_api_version: V3
        route_config_name: first-listener
      stat_prefix: http
      upgrade_configs:
      - upgrade_type: websocket
      use_remote_address: true
name: first-listener
per_connection_buffer_limit_bytes: 32768
```

looks like the  `|` header is getting literally translated instead of being interpreted as a block of YAML string

zirain:
> 

why there's a `|`?

arkodg:
managed to solve the issue by setting the value of `Value` to `interface{}` but now `controller-gen` `DeepCopy()` is failing, might need to handwrite it myself

internal/ir/xds.go
```diff
@@ -818,5 +818,5 @@ type JSONPatchOperation struct {
 	// Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
 	Path string `json:"path"`
 	// Value is the new value of the path location.
-	Value string `json:"value"`
+	Value interface{} `json:"value"`
```
zirain, line 821: try `apiextensionsv1.JSON` ?

zirain: apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"

arkodg: it worked, thanks !

internal/ir/xds.go
```diff
@@ -792,3 +796,28 @@ type OpenTelemetryAccessLog struct {
...
+	// Patch defines the JSON Patch Operation
+	Operation JSONPatchOperation `json:"operation"`
+}
+
+// JSONPatchOperation defines the JSON Patch Operation as defined in
+// https://datatracker.ietf.org/doc/html/rfc6902
+// +k8s:deepcopy-gen=true
+type JSONPatchOperation struct {
+	// Op is the type of operation to perform
+	Op string `json:"op"`
```
zirain, line 817: use full name here?

internal/xds/translator/jsonpatch.go
```diff
@@ -0,0 +1,279 @@
...
+	for _, p := range jsonPatches {
+		var (
+			listener     *listenerv3.Listener
+			routeConfig  *routev3.RouteConfiguration
+			cluster      *clusterv3.Cluster
+			endpoint     *endpointv3.ClusterLoadAssignment
+			resourceJSON []byte
+			err          error
+		)
+		m := protojson.MarshalOptions{
```
zirain, line 38: global?

internal/ir/xds.go
zirain: or rename to `Type`, and change to Enum?

arkodg: this nameing is tied to JSONPatch spec, since we marshal this and directly pass it to https://github.com/evanphx/json-patch#create-and-apply-a-json-patch

zirain: `Operation.Op` that's odd.

internal/xds/translator/translator.go
```diff
@@ -253,8 +258,8 @@ func processUDPListenerXdsTranslation(tCtx *types.ResourceVersionTable, udpListe
 
 }
 
-// findXdsListener finds a xds listener with the same address, port and protocol, and returns nil if there is no match.
-func findXdsListener(tCtx *types.ResourceVersionTable, address string, port uint32,
+// findXdsListenerByAddress finds a xds listener with the same address, port and protocol, and returns nil if there is no match.
+func findXdsListenerByAddress(tCtx *types.ResourceVersionTable, address string, port uint32,
```
zirain, line 262:
```suggestion
func findXdsListenerByAddress(tCtx *types.ResourceVersionTable, host string, port uint32,
```
address = schema+host+port

internal/ir/xds.go
qicz: +1

internal/xds/translator/jsonpatch.go
```diff
@@ -0,0 +1,280 @@
...
+			routeConfig  *routev3.RouteConfiguration
+			cluster      *clusterv3.Cluster
+			endpoint     *endpointv3.ClusterLoadAssignment
+			resourceJSON []byte
+			err          error
+		)
+
+		// If Path is "" and op is "add", unmarshal and add the patch as a complete
+		// resource
+		if p.Operation.Op == "add" && p.Operation.Path == "" {
```
qicz, line 45: prefer using  constant

arkodg: sg

internal/xds/translator/translator.go
arkodg: how about naming  it to `findXdsListenerByHostPort` ?

internal/ir/xds.go
arkodg: its part of the spec :)
https://datatracker.ietf.org/doc/html/rfc6902/#section-4

internal/xds/translator/translator.go
zirain: I'm ok with the name, cause most of time you don't care scheme.

internal/ir/xds.go
zhaohuabing: I know it looks a bit awkward, but as Arko pointed out, it aligns with the RFC standard.

Below description is extracted from the RFC:

Operation objects MUST have exactly one "op" member, whose value indicates the operation to perform.  Its value MUST be one of "add", "remove", "replace", "move", "copy", or "test"; other values are  errors.  

The following examples are from the same RFC.
```
   { "op": "add", "path": "/a/b/c", "value": "foo" }
   { "path": "/a/b/c", "op": "add", "value": "foo" }
   { "value": "foo", "path": "/a/b/c", "op": "add" }
 ```
 
 So I think we can settle with the current implementation so we won't block vendor extensions replying on the EnvoyPatchPolicy. It's internal to EG anyway,  we can always go back and change it once we figure out a better way to handle it.

zhaohuabing (approved):
LGTM.

As I mentioned in the other comment, the Operation structure issue is not blocking and could be solved in the futurue.

zirain (approved):

PR #1608, opened 2023-06-30
fix(install/helm): envoy gateway configuration propagation from values file to config map

### LINKED ISSUE #1607: Pass envoy gateway config from values.yaml to helm chart configmap
*Description*:
returned, etc.

The values configured in values.yaml for configuring envoy gateway config are not propagated to `envoy-gateway-config.yaml`.

*Repro steps*:

Install helm chart using values.yaml with envoyGateway configuration as below:
```yaml
config:
  envoyGateway:
    gateway:
    provider:
      type: Kubernetes
      kubernetes:
        rateLimitDeployment:
          replicas: 2
    rateLimit:
      backend:
        type: Redis
        redis:
          url: redis.redis-system.svc.cluster.local:6379
```

The expected behavior is the propagation of the above configuration to `envoy-gateway-config.yaml` config map but only the controllerName and provider.type are propagated to the install config map manifest.

sharing.

*Environment*:

*Logs*:

psinghal20:
This PR updates the config map manifest in the helm chart to propagate the envoyGateway config map from values file.

Fixes #1607

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-config.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-config.yaml
@@ -8,20 +8,5 @@ metadata:
 data:
   envoy-gateway.yaml: |
     apiVersion: config.gateway.envoyproxy.io/v1alpha1
-    gateway:
-      controllerName: {{ .Values.config.envoyGateway.gateway.controllerName
-        | quote }}
     kind: EnvoyGateway
-    provider:
-      {{- toYaml .Values.config.envoyGateway.provider | nindent 6 }}
-    {{- if .Values.config.envoyGateway.rateLimit }}
-    rateLimit:
-      {{- toYaml .Values.config.envoyGateway.rateLimit | nindent 6 }}
-    {{- end }}
-    {{- if .Values.config.envoyGateway.extension }}
-    extension:
-      {{- toYaml .Values.config.envoyGateway.extension | nindent 6 }}
-    {{- end }}
-    logging:
-      level:
-        default: info
+    {{- toYaml .Values.config.envoyGateway | nindent 4 }}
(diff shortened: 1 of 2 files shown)


charts/gateway-helm/templates/envoy-gateway-config.yaml
```diff
@@ -13,7 +13,15 @@ data:
         | quote }}
     kind: EnvoyGateway
     provider:
-      type: {{ .Values.config.envoyGateway.provider.type | quote }}
+      {{- toYaml .Values.config.envoyGateway.provider | nindent 6 }}
```
zirain, line 16: My concern is everytime there's an new field introduced, you need update these lines too.
Is it possible render `Values.config.envoyGateway` directly?

psinghal20: Hi @zirain, I have updated the manifest file to render the complete `values.config.envoyGateway` configuration and moved the logging levels to the values file as well.

zirain: would you like to add an e2e test to ensure this work as expected?

psinghal20: Sure, I would love to. 

WDYT about updating the current ratelimit tests manifests [here](https://github.com/envoyproxy/gateway/blob/main/examples/redis/redis.yaml#L52-L69) with a Kubernetes provider deployment spec(something like configuring cache key prefix) and checking the desired behavior in ratelimit e2e tests?

zirain: you can reuse that manifests configuration, but in a sepreated test case.

arkodg (approved):
LGTM thanks !

zirain (approved):
LGTM, a e2e in following PR will be better.

Issue #1117, opened 2023-03-09
Configure Envoy Gateway in different Deployment modes


arkodg:
*Description*:
Today by default, Envoy Gateway `watches` for resources (Gateways, HTTPRoutes) in all namespaces, and `deploys` managed resources (Envoy Deployment, Service) in a single namespace (where EG is running e.g. `envoy-gateway-system`)
It would be great if the project supported other modes such as 
* watch for resources in a single namespace and/or

- [x] API https://github.com/envoyproxy/gateway/pull/1413
- [x] Implement https://github.com/envoyproxy/gateway/pull/1610
- [x]  Docs

* deploy in multiple namespaces (where `Gateway` is created)

- [ ] API
- [ ] Implement
- [ ] Docs

This intent could be defined in the [EnvoyGateway](https://github.com/envoyproxy/gateway/blob/f76ebe01d265d60e9e703d7069005c49861d11c9/api/config/v1alpha1/envoygateway_types.go#L88) config

[optional *Relevant Links*:]
Relates to https://github.com/envoyproxy/gateway/issues/1067

arkodg:
thinking out loud, potential API could look like
```
deploymentMode:
  watch:
    namespaces: .........
  deploy:
    envoy:
      namespaceType: GatewayNamespace  
```

* By default (when this field is unset), EG watches all resources in all namespaces, and depoys infra in the namespace where EG is running (`ControllerNamespace`)
* `watch.namespaces` can be used to set the specific namespaces to watch resources (Gateway, xRoute, Service etc)
* `deploy.envoy.namespaceType` can be used to specify to EG to deploy proxy infra to deploy envoy deployments in the namespace where the Gateway is created in. This option will need a helm knob too, to configure added permissions to be able to create deployments and services in other namespaces

PR #1610, opened 2023-06-30
Implement Watch Namespaces

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1117

### DIFF
--- a/internal/provider/kubernetes/kubernetes.go
+++ b/internal/provider/kubernetes/kubernetes.go
@@ -40,6 +40,11 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 		LeaderElectionID:       "5b9825d2.gateway.envoyproxy.io",
 		MetricsBindAddress:     ":8080",
 	}
+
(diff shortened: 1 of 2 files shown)


internal/provider/kubernetes/kubernetes.go
```diff
@@ -40,6 +40,14 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 		LeaderElectionID:       "5b9825d2.gateway.envoyproxy.io",
 		MetricsBindAddress:     ":8080",
 	}
+
+	if svr.EnvoyGateway.Provider != nil &&
+		svr.EnvoyGateway.Provider.Kubernetes != nil &&
+		(svr.EnvoyGateway.Provider.Kubernetes.Watch != nil) &&
+		(len(svr.EnvoyGateway.Provider.Kubernetes.Watch.Namespaces) > 0) {
+		mgrOpts.Cache.Namespaces = svr.EnvoyGateway.Provider.Kubernetes.Watch.Namespaces
```
zirain, line 48: will EG only get resources from `WatchNamespaces`?

arkodg: yes if this field is set, the test case reflects this

zirain (approved):

qicz (approved):
it's great, it is better for EG memory occupied.

LGTM

Issue #1613, opened 2023-07-03
EG high memory usage


qicz:
*Version*

the main branch

*Resources Settings*

limits: memory 1Gi,  CPU 500m

*Platform*

Openshift

*Description*:
The current EG implementation watches a lot of resources, such as HTTPRoute, GPCRoute, EnvoyProxy, etc., and a copy of these resources will be reserved in the EG memory, causing the EG memory usage to be very high. When the user's HTTPRoute and other resources continue to increase, the memory of EG will take up more, and the user needs to adjust the resource limits dynamically ?

![image](https://github.com/envoyproxy/gateway/assets/2174082/34d964e8-23b9-4945-8e5c-bf4872747ffe)

When any resource changes, EG will trigger the behavior of these changes, such as translation and infrastructure reconstruction. 

https://github.com/envoyproxy/gateway/blob/6af516d9b36b237c7da44a087eafc57bf3d7d9a2/internal/provider/kubernetes/controller.go#L287-L290

https://github.com/envoyproxy/gateway/blob/6af516d9b36b237c7da44a087eafc57bf3d7d9a2/internal/gatewayapi/runner/runner.go#L51-L66

These are actually unnecessary. Is it possible to separate watch for related resources or separate store and subscribe when resources really change?

like this:

```go
if val, ok := r.resources.GatewayAPIHTTPRoutesResources.Load(acceptedGC.Name); !ok || cmp.Diff(val, httpRoutes) != "" {
   r.resources.GatewayAPIHTTPRoutesResources.Store(acceptedGC.Name, httpRoutes)
}
```

qicz:
> thanks for raising this issue @qicz . Can you also highlight
> 
> 1. Number of resources you have running - Gateways, HTTPRoutes, etc ?
> 2. Version of EG you are running
> 
> Why is the memory increasing when the number of resources are increasing ?
> 
> * As you highlighted above, EG store relevant resources in the `watchable` store, so increase in resources increases the memory usage of the EG.
> * This is an intentional design decision to
>   
>   * Reduce CPU - Caching helps us avoid translations if input resources haven't changed
>   * Reduce I/O - Caching helps us avoid writes to Envoy (xds) or K8s API Server (status) if output resources haven't changed
>     which comes at a cost of memory usage.
> 
> How can we deal with this memory usage at high scale ?
> 
[...]

Issue #1365, opened 2023-04-25
Envoy Gateway performance at scale


arkodg:
*Description*:
This issue tracks the performance (Throughput, Latency) of Envoy Gateway control plane and data plane at scale (Service, xRoutes, Gateway, Client Connections) . 
  * The output of this issue should be a document that can be presented to the user to better understand what is the performance of Envoy Gateway by default and how can this be increased if there is an increase in scale.
  * Other issues might need to be created (based on testing) and linked to this issue 

[optional *Relevant Links*:]
Emissary: https://www.getambassador.io/docs/emissary/latest/topics/running/scaling
Contour: https://github.com/projectcontour/contour-perf / https://projectcontour.io/guides/resource-limits/
Istio: https://istio.io/v1.16/docs/ops/deployment/performance-and-scalability/

arkodg:
related https://www.envoyproxy.io/docs/envoy/latest/faq/performance/how_to_benchmark_envoy

kflynn:
Good first target: 1GB RAM usage at 1000 HTTPRoutes. (We'd probably be OK at 2GB, but let's go for 1GB.)

Xunzhuo:
Do we run performance test in GitHub CI ? I do not know if GitHub CI provides enough resources to run large scale eg tests.

arkodg:
> Do we run performance test in GitHub CI ? I do not know if GitHub CI provides enough resources to run large scale eg tests.

it provides 7GB RAM https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources, we might be able to do without a self hosted runner, since we are shooting for a usage of 1-2GB

haq204:
Current statistics

Notes:
- This is for a vanilla install of Envoy Gateway. No extensions
- HTTPRoutes do not use any filters
- HTTPRoutes are routed to the same service just with different paths (e.g `/get<n>`)
- It is against a single Gateway & GatewayClass

| cpu | mem | # simultaneously applied HTTPRoutes before OOMing |
| ----- | -------| --------------------- |
| 1000m | 1Gb | 400-500 |
| 1000m | 2Gb | 700 |
| 1000m | 4Gb | 1000 |

arkodg:
looks like envoy proxy is building a framework / test suite for perf testing
[...]

PR #1617, opened 2023-07-03
bump resource limits for Envoy Gateway deployment

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1613

Must be revisited again when https://github.com/envoyproxy/gateway/issues/1365 is completed


zirain (approved):

qicz:
ref https://github.com/envoyproxy/gateway/issues/1613#issuecomment-1619288673

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -7,10 +7,10 @@ deployment:
     resources:
       limits:
         cpu: 500m
-        memory: 128Mi
+        memory: 1024Mi
       requests:
-        cpu: 10m
-        memory: 64Mi
+        cpu: 100m
```
qicz, line 12: the `requests` does not need to increase. EG can run on cpu 10m and mem 64Mi. for the limits IMO can't fix memory hign usage. https://github.com/envoyproxy/gateway/issues/1613#issuecomment-1619288673

qicz: I see, but request changing is unnecessary for my production env.

arkodg: I feel like `10m` is too low to start off with

Xunzhuo (approved):

PR #1618, opened 2023-07-04
Improve EnvoyPatchPolicy CRD Validations

arkodg:
* using kube-builder annotations

### DIFF
--- a/api/v1alpha1/envoypatchpolicy_types.go
+++ b/api/v1alpha1/envoypatchpolicy_types.go
@@ -94,15 +94,20 @@ const (
 	ClusterLoadAssignmentEnvoyResourceType EnvoyResourceType = "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment"
 )
 
+// JSONPatchOperationType specifies the JSON Patch operations that can be performed.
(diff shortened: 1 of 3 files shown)


api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -94,15 +94,20 @@ const (
 	ClusterLoadAssignmentEnvoyResourceType EnvoyResourceType = "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment"
 )
 
+// JSONPatchOperationType specifies the JSON Patch operations that can be performed.
+// +kubebuilder:validation:Enum=add;remove;replace;move;copy;test
+type JSONPatchOperationType string
```
zirain, line 99: better to add some const values?

arkodg: I can, but the comments will end up linking back to the orig spec, trying to keep the JSONPatch education away from this API, user will have to learn by looking at the spec or going through our docs, and keeping validation here for failing early with client side validations

zirain (approved):

AliceProxy (approved):

PR #1628, opened 2023-07-05
Add EnvoyPatchPolicy support to Gateway API lib

arkodg:


api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -107,8 +108,7 @@ type JSONPatchOperation struct {
 	// Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
 	Path string `json:"path"`
 	// Value is the new value of the path location.
-	// +kubebuilder:validation:MinLength=1
-	Value string `json:"value"`
+	Value apiextensionsv1.JSON `json:"value"`
```
zhaohuabing, line 111: I prefer the old string value because it's easier troubleshooting.  String is a human readable format but  apiextensionsv1.JSON is not.

arkodg: im unable to unmarshal into `string` which is why I had to set it to `apiextensionsv1.JSON` in https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/ir/xds.go#L826

due to the above constraint I had to also set the API field to `apiextensionsv1.JSON` so could copy the field values

zhaohuabing: Let's go with byte[] and try to find a solution to change it to string in a subsequent pr.

zhaohuabing (approved):
LGTM. 

We will try to solve the json patch value issue in a subsequent PR.

zirain (approved):

PR #1629, opened 2023-07-06
tracing: correct service name

zirain:
use EnvoyProxy name as service name


internal/gatewayapi/listener.go
```diff
@@ -218,10 +218,13 @@ func processAccessLog(envoyproxy *configv1a1.EnvoyProxy) *ir.AccessLog {
...
 
-func processTracing(envoyproxy *configv1a1.EnvoyProxy) *configv1a1.ProxyTracing {
+func processTracing(gw *v1beta1.Gateway, envoyproxy *configv1a1.EnvoyProxy) *ir.Tracing {
 	if envoyproxy == nil || envoyproxy.Spec.Telemetry.Tracing == nil {
 		return nil
 	}
 
-	return envoyproxy.Spec.Telemetry.Tracing
+	return &ir.Tracing{
+		ServiceName:  gw.Name + "." + gw.Namespace,
```
arkodg, line 227: lets use this helper func here instead ?
https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/gatewayapi/helpers.go#L389

zirain: sure, let me find a right package

quyenhoang96:
Is it ready to run?

arkodg (approved):

AliceProxy (approved):

internal/utils/naming/naming.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package naming
+
+import "k8s.io/apimachinery/pkg/types"
+
+func ServiceName(nn types.NamespacedName) string {
+	return nn.Name + "." + nn.Namespace
```
arkodg, line 11: should this be `namespace/name`, now that his how we are naming things ?

zirain: align to what istio does, also same as the fqdn format in Kubernetes, e.g. `nginx.default`, `foo.bar`

arkodg (approved):

Issue #1620, opened 2023-07-04
support for x-ratelimit-* response headers in ratelimit 


tmsnan:
*Description*:
> support for x-ratelimit-* response headers in ratelimit (easy to check current limit status）
[`XRateLimitHeadersRFCVersion` ](https://github.com/envoyproxy/envoy/blob/4e51ba92fe9effd943f68568762c3b213ec818bd/api/envoy/extensions/common/ratelimit/v3/ratelimit.proto#L20C1-L35C2)

[optional *Relevant Links*:]
>[draft RFC Version 03 ](https://datatracker.ietf.org/doc/id/draft-polli-ratelimit-headers-03.html)

tmsnan:
If accepted, I can complete this。

arkodg:
+1

quyenhoang96:
If no one has done it, you can give it to me

PR #1630, opened 2023-07-06
feat:ratelimit support xRateLimitHeaders

tmsnan:
ratelimit support xRateLimitHeaders
Relates to #1620

### DIFF
--- a/api/config/v1alpha1/envoygateway_types.go
+++ b/api/config/v1alpha1/envoygateway_types.go
@@ -265,14 +265,25 @@ type RateLimit struct {
 	XRateLimitHeadersRFCVersion XRateLimitHeadersRFCVersion `json:"xRateLimitHeadersRFCVersion"`
 }
 
+// XRateLimitHeadersRFCVersion specifies the types of X-RateLimit headers
+// to be used by the ratelimit.
+// +kubebuilder:validation:Enum=draft_version_03
 type XRateLimitHeadersRFCVersion string
 
 const (
 	// X-RateLimit headers disabled.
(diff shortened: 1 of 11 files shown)


api/config/v1alpha1/envoygateway_types.go
```diff
@@ -258,8 +258,34 @@ type RateLimit struct {
 	// database backend used by the rate limit service to store
 	// state associated with global ratelimiting.
 	Backend RateLimitDatabaseBackend `json:"backend"`
+
+	// XRateLimitHeadersRFCVersion defines the version of the standard to use for X-RateLimit headers.
```
arkodg, line 262: I would enable this feature by default without adding any API field.

arkodg:
thanks for adding this feature @tmsnan . 
Im in favor of enabling this field and adding the Ratelimit response headers without explicitly adding a API field for this, since this is the default/common case.
cc @envoyproxy/gateway-maintainers as well for additional input

zirain:
+1 enable this by default without API for now.

api/config/v1alpha1/envoygateway_types.go
tmsnan: If there is no extreme performance requirement, it is also acceptable to enable it by default.

arkodg: go for it !

tmsnan: done

arkodg (approved):

zirain (approved):

PR #1631, opened 2023-07-06
feat: add support for custom envoy proxy pod labels configuration

### LINKED ISSUE #1626: Support Envoy Proxy pod labels configuration
*Description*:
>Be able to set custom labels on the Envoy Proxy pod

```
kind: EnvoyProxy
metadata:
  name: envoy-proxy
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        replicas: 1
        pod:
          labels:
            key1: val1
            key2: val2
```

shawnh2:
**What type of PR is this?**

Fixes #1626


api/config/v1alpha1/shared_types.go
```diff
@@ -77,6 +77,12 @@ type KubernetesPodSpec struct {
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// Labels are the labels that should be tagged to the pods.
```
arkodg, line 80:
```suggestion
	// Labels are the additional labels that should be tagged to the pods.
```

api/config/v1alpha1/shared_types.go
```diff
@@ -77,6 +77,12 @@ type KubernetesPodSpec struct {
...
+	// By default, no pod labels are tagged.
```
arkodg, line 81:
```suggestion
```

api/config/v1alpha1/shared_types.go
```diff
@@ -77,6 +77,12 @@ type KubernetesPodSpec struct {
...
+	//
```
arkodg, line 82: made above comments since we do have default labels https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/infrastructure/kubernetes/proxy/resource.go#L62, we dont want to the user to rm these

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -170,6 +170,17 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		annotations = deploymentConfig.Pod.Annotations
 	}
 
+	// Get custom and selector labels for pod
```
arkodg, line 173: we should perform a `append` here, not replace

quyenhoang96:
For now, My gateway on the production of my company needs to this feature. Maybe, I can become the first person to test this feature on real project for you.

internal/infrastructure/kubernetes/proxy/resource_provider.go
qicz: +1, we should append some new labels and keep the default label

shawnh2: if append the custom labels in this func:
https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/infrastructure/kubernetes/proxy/resource_provider.go#L160
the custom labels will be tagged successfully on the pod:
https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/infrastructure/kubernetes/proxy/resource_provider.go#L188-L190
but these custom labels will also be tagged on its deployment:
https://github.com/envoyproxy/gateway/blob/773bdfbefc769f51779d5056547e88d12e1d9508/internal/infrastructure/kubernetes/proxy/resource_provider.go#L178-L181

is this a reasonable behavior?

arkodg: probably call the func twice, and assign it to `deploymentLabels` and `podLabels` fields, and reuse them ?

shawnh2: great idea

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -77,12 +77,14 @@ func EnvoyAppLabelSelector() []string {
 }
 
 // envoyLabels returns the labels, including extraLabels, used for Envoy resources.
-func envoyLabels(extraLabels map[string]string) map[string]string {
+func envoyLabels(extraLabels, customLabels map[string]string) map[string]string {
```
arkodg, line 80: this is not a great API imo
the caller should be merging this maybe using https://pkg.go.dev/golang.org/x/exp/maps#Copy before passing it to `envoyLabels`

shawnh2: fixed

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

Issue #1468, opened 2023-05-31
Opt in / Opt Out of Extension APIs


arkodg:
*Description*:
Add knobs in the `EnvoyGateway` config to limit the features/APIs in the Envoy Gateway controller to just the APIs exposed by the upstream Gateway API project and not any extension APIs supported by Envoy Gateway (such as RateLimitFilter or EnvoyPatchPolicy)

PR #1632, opened 2023-07-06
Add opt-in knob to enable EnvoyPatchPolicy

arkodg:
Fixes https://github.com/envoyproxy/gateway/issues/1468


AliceProxy (dismissed):
Yeah I think opting-in explicitly to "experimental" features is fine

zirain:
can EG reuse `feautreGates` like controller-runtime?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -120,6 +120,19 @@ type Gateway struct {
 	//
 	// +optional
 	ControllerName string `json:"controllerName,omitempty"`
+
+	// ExtensionAPIs defines the settings related to specific Gateway API Extensions
```
zhaohuabing, line 124: Is this comment correct?

arkodg: any suggestions for improving it ?

zhaohuabing: The comment (and the name ExtensionAPIs as well) looks vague to me. Did you intentionally make it this way because this API will also serve other purposes other just than disabling/enabling specific features?

arkodg: yeah sorta, trying to use a high level noun that can encapsulate an entire class of settings, so chose `ExtensionAPIs` to signify all settings pertaining to it

arkodg:
> can EG reuse `feautreGates` like controller-runtime?

@zirain can you elaborate on this

api/config/v1alpha1/envoygateway_types.go
zhaohuabing: EnvoyPatchPolicy is not part of the Gateway API. That's why I feel confused when I saw the location of the ExtensionAPIs(it's part of the Gateway Structure) and the comment “ ExtensionAPIs defines the settings related to specific Gateway API Extensions implemented by Envoy Gateway”.

arkodg: any suggestions on what might be a better home ?

zhaohuabing: For any features we're consider dangerous of not stable, we can pass them as command line parameters like [K8s feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) does.

--feature-gates=...,EnvoyPatchPolicy=true

zirain: @zhaohuabing yes, that's what I'm thinking.

On the other hand, EnvoyPatch will before there's a CR applied, so maybe it's unneccessary to do this for EnvoyPatch?

arkodg: based on this discussion I believe there are 3 approaches
1. add a `--feature-flags` flag to the EG CLI
2. add a top level `FeatureFlags` struct to the `EnvoyGateway` config
```
featureFlags:
  enableEnvoyPatchPolicy: true
```  
3. Continue with the above approach
```
gateway:
  extenstionAPIs:
    enableEnvoyPatchPolicy: true
```

I'm a -1 on 1. since I'd prefer if we kept all our config knobs inside the `EnvoyGateway` config instead of creating  many CLI flags and its easier to manage doc generation with the `EnvoyGateway` config.

Fine with 2. or 3. , 2. is a flat approach and might end up having many unrelated knobs living next to each other, but that might be tolerable.

@zhaohuabing @zirain @AliceProxy any thoughts ?

zhaohuabing: I agree with your reasons on 3, but still the location and the name of the structure `extenstionAPIs ` seem not quite aligned with its actual function. But I can't figure out a better name and location either.

zhaohuabing: Maybe we can hold this until we have other settings that we want to put in the `ExtensionAPIs`?  By that time, we can revisit this matter and potentially come up with a better way to handle it.

arkodg: we will need to fix this before v0.5.0, since we want `EnvoyPatchPolicy` to be opt in

arkodg: @kflynn @AliceProxy and I discussed this issue in the community meeting today, and favor 3.
Although `ExtentionAPIs` is not the best name, we cannot come up with a better name at this point so prefer sticking to it for now,

zhaohuabing: > we will need to fix this before v0.5.0, since we want `EnvoyPatchPolicy` to be opt in

@arkodg @AliceProxy is this a must-have in v0.5.0? As @zirain said, if no EnvoyPatchPolicies are created, it has no impact at all.

arkodg: chatted with @zhaohuabing a little more, and he shared his hesitation with
```
gateway:
  extensionAPIs:
    ...
```
where `extensionAPIs` are not part of the Gateway API, which is valid.

another option is to reuse the top level `extension` field, that is used by the extension service
```
extension:
  apis:
    enableEnvoyPatchPolicy: true
  controlPlane:
    hooks:
    service:
```

zhaohuabing (approved):
LGTM.

I would feel more comfortable if the word "Gateway" is removed from the comment of the Gateway struct :-)

"Gateway defines the desired API configuration of Envoy Gateway." 

Because it's a container of the configuration of the whole Envoy Gateway API, not only limited to Gateway API.

api/config/v1alpha1/envoygateway_types.go
zhaohuabing: Gateway can be the container of the settings of the whole EG, not just Gateway API. 

I suggest a minor change to the comment to address this:

Change 

"Gateway defines the desired Gateway API configuration of Envoy Gateway."

To

"Gateway defines the desired API configuration of Envoy Gateway."

arkodg:
hey @AliceProxy @kflynn @haq204 , in the latest commit I moved the `extensionApis` field to the top level, which forced me rename the `extension` field which was used for the extension manager control plane settings to `extensionManager` , this is a breaking change, but it keeps all the extension configuration at the same level, which might be better for long term, wdyt

zhaohuabing (approved):
LGTM

I'm ok with either way, slightly prefer placing the  `extensionApis` at the top level.

zirain (commented):
lgtm, @AliceProxy ptal

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -67,10 +67,16 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	RateLimit *RateLimit `json:"rateLimit,omitempty"`
 
-	// Extension defines an extension to register for the Envoy Gateway Control Plane.
+	// ExtensionManager defines an extension manager to register for the Envoy Gateway Control Plane.
 	//
 	// +optional
-	Extension *Extension `json:"extension,omitempty"`
+	ExtensionManager *ExtensionManager `json:"extensionManager,omitempty"`
```
AliceProxy, line 73: lgtm even though this is a breaking change for Envoy Gateway extensions. If possible we should draw extra attention to this in the release notes with a :warning: emoji or something.

AliceProxy (approved):

api/config/v1alpha1/envoygateway_types.go
arkodg: will do, adding the release notes label to ensure we handle ^

PR #1636, opened 2023-07-08
fix json patch issue

zhaohuabing:
fix https://github.com/envoyproxy/gateway/pull/1628#discussion_r1255191062

It's a bit tricky because Marshal method actually converts the json in the operation value to a string literal by adding escape characters. The workaround is manually creating the json patch.


zhaohuabing:
why coverage failed?

internal/xds/translator/testdata/in/xds-ir/jsonpatch.yaml
```diff
@@ -21,15 +21,15 @@ jsonPatches:
...
+    value: |
+      - actions:
+        - remote_address: {}
 - type: "type.googleapis.com/envoy.config.cluster.v3.Cluster"
   name: rate-limit-cluster
   operation:
     op: add
     path: ""
-    value:
+    value: |
```
zirain, line 32: is the '|' required? if so, I don't think this is better than before.

zhaohuabing: The  "|" means that the value is a multi-line string, that's it.

zirain: It's ok for a configmap.
I think here we want some like `pb.Struct` or `map[string]interface{}`, so I prefer `apiext.JSON` than `string`.

zhaohuabing: The value of the Patch is a string. Besides, it's very hard to troubleshooting if we define the API as a byte array.

zhaohuabing: I does look a bit weird. It's better to just treat the value as a yaml/json structure.

PR #1431, opened 2023-05-19
support tls termination for tcp traffic

tanujd11:
**What type of PR is this?**
 Feature to provide TLS termination with TCP traffic

**What this PR does / why we need it**:

Supports the following configuration:
Listener Protocol = TLS
TLS Mode = Terminate
Route Type =TCPRoute

**Which issue(s) this PR fixes**:

Fixes #1391

PR #1643, opened 2023-07-10
docs: missing docs for tls termination for tcp

### LINKED ISSUE #1642: tls termination for tcp docs missing on latest
*Description*:
TLS termination was added in this [PR](https://github.com/envoyproxy/gateway/pull/1431) and docs was also added but it was not added to `user_docs.rst` file so it is not visible on the latest docs.

slayer321:
**What type of PR is this?**
docs: missing docs for tls termination for tcp

TLS termination was added in this [PR](https://github.com/envoyproxy/gateway/pull/1431) and docs was also added but it was not added to user_docs.rst file so it is not visible on the latest docs.

Fixes #1642

### DIFF
--- a/docs/latest/user_docs.rst
+++ b/docs/latest/user_docs.rst
@@ -17,6 +17,7 @@ Learn how to deploy, use, and operate Envoy Gateway.
   user/secure-gateways
   user/tls-cert-manager
   user/tls-passthrough
+  user/tls-termination
   user/tcp-routing
   user/udp-routing
   user/grpc-routing


docs/latest/user_docs.rst
```diff
@@ -17,6 +17,7 @@ Learn how to deploy, use, and operate Envoy Gateway.
   user/secure-gateways
   user/tls-cert-manager
   user/tls-passthrough
```
arkodg, line 19: good catch, can you also update https://github.com/envoyproxy/gateway/blob/main/docs/v0.4.0/user_docs.rst

arkodg: if its relevant to this release (0.4.0), would be good to add it ^ as well

slayer321: As `0.4.0` version don't have that [file](https://github.com/envoyproxy/gateway/tree/main/docs/v0.4.0/user). So I think there is no need to add it to `0.4.0` version.

arkodg (approved):
LGTM, thanks !

Xunzhuo (approved):

Issue #1633, opened 2023-07-07
Support Envoy Gateway pod labels and annotations configuration in Helm


rkosyk:
*Description*:
>Be able to set custom labels and annotations on the Envoy Gateway pod using Helm
   
   
Helm values.yaml file
```
deployment:
  envoyGateway:
  kubeRbacProxy:
  ports:
  replicas: 1
  pod:
    annotations:
      key1: val1
      key2: val2
    labels:
      key1: val1
      key2: val2
```

shawnh2:
cc @arkodg can be done as a follow-up of #1631

PR #1644, opened 2023-07-10
Support_Envoy_Gateway_pod_labels_and_annotations

rkosyk:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/1633


arkodg:
should we directly expose `deployment.metdata` & `deployment.spec` as helm values so we dont need to keep adding helm templating for each sub field ?

rkosyk:
> should we directly expose `deployment.metdata` & `deployment.spec` as helm values so we dont need to keep adding helm templating for each sub field ?

It is not recommended, as users don't need to have in-depth knowledge of Kubernetes object structures and syntax.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1651, opened 2023-07-12
docs: highlight helm chart customizations in installation page

### LINKED ISSUE #1611: Highlight Helm Chart customizations in the Installation Page
*Description*:
Highlight the many ways the Envoy Gateway can be customized using helm, shown here
https://github.com/envoyproxy/gateway/tree/main/charts/gateway-helm#values to the user
in the docs website, in the installation page https://github.com/envoyproxy/gateway/blob/main/docs/latest/user/installation.md

[optional *Relevant Links*:]

slayer321:
**What type of PR is this?**
docs: highlight helm chart customizations in installation page

Here we are highlighting helm customizations in  installation page.
Using `helm-docs` cli  to sync the the api.md file.
Added new make command for helm .. now whenever we run `make docs` command helm will also be updated.  

Fixes #1611

PS : One concern that I had is though we have added the `helm-docs` cli to update the `helm/api.md` file . But our README.md file still don't get updated . If there are any changes. We can update that as well using `helm-docs` but it will remove some section like Usage from README.md file.


tools/hack/helm-readme-gen.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+gitroot=$(git rev-parse --show-toplevel)
+#Install latest helm-docs cli
+go install github.com/norwoodj/helm-docs/cmd/helm-docs@latest
```
zirain, line 9: normally, eg does not use this way, please refer `tools.mk`.
BTW, you should pin to a stable version rather than `latest`.

arkodg (approved):
LGTM thanks !

zirain (approved):
Thanks！

PR #1656, opened 2023-07-13
helm: apply Role and RoleBinding for envoy gateway in namespaced mode

### LINKED ISSUE #1619: Apply EG Role and RoleBinding when running in namespaced mode
*Description*:
Envoy Gateway recently added support to allow users to limit the namespaces Envoy Gateway watches to gather resources in https://github.com/envoyproxy/gateway/issues/1117

This also means Envoy Gateway can use namespaced `Role` & `RoleBinding` resources instead of `ClusterRole` https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/generated/rbac/roles.yaml reducing the permissions needed to operate.

shawnh2:
**What type of PR is this?**

Fixes #1619


arkodg:
thanks for working on this @shawnh2, sharing some thoughts, @Xunzhuo would also love your thoughts on this
* maybe its time to stop auto generating the `ClusterRole` https://github.com/envoyproxy/gateway/blob/2edcafe430a2ac1e61ccb4d1a0f5a5ce00ad5a08/tools/make/kube.mk#L34 and just manage it ourselves. This allows us to use your current logic to decide if we need to create a `ClusterRole` or a `Role`
* I think we need `ClusterRole` for `GatewayClass` even for the namespaced mode, so we can split up roles into 2 resources, one for the gateway class cluster role and the other for everything else (cluster role for regular mode and role for namespaced mode)
*  Can we use `{{ include "eg.fullname" $ }}-envoy-gateway-role` instead of `{{ include "eg.fullname" $ }}-{{ $ns }}-envoy-gateway-role` since they are created in different namespaces so the `ns` separation shouldnt matter ?
* Would be good to test this out with the steps highlighted in https://gateway.envoyproxy.io/v0.4.0/user/deployment-mode.html#multi-tenancy and see that it works

shawnh2:
1. sure we can mange `ClusterRole` and `Role` ourselves
2. +1, we should create `ClusterRole` and `Role` resources no matter what, but create `RoleBinding` if we run in namespaced mode, otherwise create `ClusterRoleBinding`
3. thanks for pointing it out, missing that oops
4. thanks for heads up, i will test it out

charts/gateway-helm/templates/envoy-gateway-rbac.yaml
```diff
@@ -1,3 +1,123 @@
+{{ $allNamespaces := "*" }}
```
arkodg, line 1: pretty close, this looks good 
what I was referring to earlier is you will need to create a `ClusterRole` + `ClusterRoleBinding` for a `GatewayClass` by default / always since GatewayClass is always cluster scoped 
https://gateway-api.sigs.k8s.io/api-types/gatewayclass/

arkodg: do you see the same error if you replace `Role` with `ClusterRole` , but use `RoleBinding` ?

shawnh2: yes, the error is same

arkodg: hmm any ideas @envoyproxy/gateway-reviewers & @envoyproxy/gateway-maintainers ?

chauhanshubham: just thinking out loud, whether it has something to do with the scope of the [manager](https://github.com/envoyproxy/gateway/blob/7c6aa9ea849667d40d0c74c2acf37c8664068329/internal/provider/kubernetes/kubernetes.go#L51). not sure though. The `at the cluster scope` pointed me in that direction.

shawnh2: it has been tested out using https://gateway.envoyproxy.io/v0.4.0/user/deployment-mode.html#multi-tenancy, and it works great!

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -185,7 +185,7 @@ type KubernetesWatchMode struct {
 	// GatewayClass that it is linked to.
 	// By default, when this field is unset or empty, Envoy Gateway will watch for input namespaced resources
 	// from all namespaces.
-	Namespaces []string
+	Namespaces []string `json:"namespaces"`
```
arkodg, line 188: is `omitempty` needed here ?

shawnh2: of course

shawnh2:
> LGTM thanks @shawnh2, since this PR makes an API change (and affects the docs dir), lets merge this after v0.5 releases. In the interim would be great if you can share the rbac o/p for default case as well as namespaces case in this PR, thanks !

sure, i added some usage in https://gateway.envoyproxy.io/latest/user/deployment-mode.html

docs/latest/user/deployment-mode.md
```diff
@@ -29,6 +31,12 @@ by the `marketing` and `product` teams in separate namespaces.
 helm install --set config.envoyGateway.gateway.controllerName=gateway.envoyproxy.io/marketing-gatewayclass-controller eg-marketing oci://docker.io/envoyproxy/gateway-helm --version v0.0.0-latest -n marketing --create-namespace
 ```
 
+* If you want to deploy Envoy Gateway in Namespaced mode, and watches resources only in the `marketing` namespace, you can specify a namespace by running:
```
arkodg, line 34: lets edit existing `helm` cmd instead of adding one

shawnh2: updated

docs/latest/user/deployment-mode.md
```diff
@@ -29,6 +31,12 @@ by the `marketing` and `product` teams in separate namespaces.
...
+
+```
+helm install --set config.envoyGateway.gateway.controllerName=gateway.envoyproxy.io/marketing-gatewayclass-controller --set config.envoyGateway.provider.kubernetes.watch.namespaces={marketing} eg-marketing oci://docker.io/envoyproxy/gateway-helm --version v0.0.0-latest -n marketing --create-namespace
```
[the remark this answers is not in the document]
shawnh2: updated

arkodg (approved):
LGTM thanks ! lets wait for v0.5 to land before merging this one

zirain (approved):

zirain:
@arkodg ready to merge？

PR #1661, opened 2023-07-16
feat: watch resource with selectors

### LINKED ISSUE #1622: watch resource with selector
*Description*:
in order to optimize EG occupy high memory,  #1610 supports EG watching the spec namespace. but the users need to list all namespaces that need watching.  It's not convenient, we can watch the namespace that has some label.  IMO should watch the namespace with the same label, like `shared-gateway-access`  https://gateway-api.sigs.k8s.io/guides/multiple-ns/

xref: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.0/pkg/cache#Options

cc @arkodg

den3tsou:
This PR is to provides a handy way for users to watch namespaces with certain labels. Envoy Gateway currently only supports specifying namespaces to watch which is not convenient for users.

Add namespaceSelectors watch mode for Kubernetes provider. Users will need to specify `EnvoyGateway.Provider.Type` and precisely one of `EnvoyGateway.Provider.Kubernetes.Wach.Namespaces` and `EnvoyGateway.Provider.Kubernetes.Wach.NamespaceSelectors` to set the KuberNtes wach mode.

The namespaceSelectors doesn't change the namespace informers watch. The informer still watches all namespaces. The events which have Objects that are not under namesapces with the labels set by `NamespaceSelectors` are filtered out.

Fixes #1622


internal/provider/kubernetes/kubernetes.go
```diff
@@ -41,10 +42,15 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 		MetricsBindAddress:     ":8080",
 	}
 
-	if svr.EnvoyGateway.Provider != nil &&
+	// TODO: implment config validatoin on the watch mode config
```
den3tsou, line 45: There doesn't seem to be any validation in place. Do I need to add it in this PR?

internal/provider/kubernetes/predicates.go
```diff
@@ -42,6 +42,48 @@ func (r *gatewayAPIReconciler) hasMatchingController(obj client.Object) bool {
...
+		ns := &corev1.Namespace{}
+		if err := r.client.Get(
+			context.Background(),
+			client.ObjectKey{
+				Namespace: "", // Namespace object should have empty Namesapce
+				Name:      nsString,
+			},
+			ns,
+		); err != nil {
+			// Question: what's the common practice to handle error in predicate?
```
den3tsou, line 59: Could anyone help me to understand the common practices in Enovy Gateway to handle errors in predicate?

internal/provider/kubernetes/predicates_test.go
```diff
@@ -59,6 +62,75 @@ func TestGatewayClassHasMatchingController(t *testing.T) {
 	}
 }
 
+// TestGatewayClassHasMatchingNamespaceLabels tests the hasMatchingNamespaceLabels
+// predicate function.
+func TestGatewayClassHasMatchingNamespaceLabels(t *testing.T) {
```
den3tsou, line 67: Is this the only place to add tests? Do I need to add tests anywhere else?

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -162,15 +162,34 @@ type EnvoyGatewayKubernetesProvider struct {
 	Deploy *KubernetesDeployMode `json:"deploy,omitempty"`
 }
 
+const (
+	// KubernetesWatchModeTypeNamespaces indicates that the namespace watch mode is used.
+	KubernetesWatchModeTypeNamespaces = "namespace"
+
+	// KubernetesWatchModeTypeNamespaceSelectors indicates that namespaceSelectors watch
+	// mode is used.
+	KubernetesWatchModeTypeNamespaceSelectors = "namespaceSelectors"
```
arkodg, line 171:
```suggestion
	KubernetesWatchModeTypeNamespaceSelectors = "NamespaceSelectors"
```

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -162,15 +162,34 @@ type EnvoyGatewayKubernetesProvider struct {
 	Deploy *KubernetesDeployMode `json:"deploy,omitempty"`
 }
 
+const (
+	// KubernetesWatchModeTypeNamespaces indicates that the namespace watch mode is used.
+	KubernetesWatchModeTypeNamespaces = "namespace"
```
arkodg, line 167:
```suggestion
	KubernetesWatchModeTypeNamespaces = "Namespaces"
```

internal/provider/kubernetes/predicates.go
arkodg: logging an error and returning `false` is fine

internal/provider/kubernetes/predicates_test.go
arkodg: a good end to end test to add would be to implement this example https://gateway-api.sigs.k8s.io/guides/multiple-ns/ in our e2e https://github.com/envoyproxy/gateway/tree/main/test/e2e/tests

den3tsou: Hi @arkodg, would you please help me understand how to test the predicate behaviour with e2e tests?  Is there any API I could hit in order to know that certain events are filtered out?

arkodg: based on https://gateway-api.sigs.k8s.io/guides/multiple-ns/#shared-gateway you could ensure e2e routing works when `default` ns has label `shared-gateway-access: "true"` set and fails when it is not ns

den3tsou: Hi @arkodg, this is doable but difficult. There is no easy way to reconfig Envoy Gateway in the current testing structure inlcuding 1) apply new ConfigMap 2) restart Envoy Gateway which is needed for this test. Adding those changes will make this PR too big. What's your thoughts on this?

arkodg: okay, then can we add some tests in https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/kubernetes_test.go that highlight that when we are in this special mode only specific resources are reconciled ?

den3tsou: Hey @arkodg , do you mind elaborating more on what you meant? I've gone through the source code to see if there are any interfaces I could use in order to test the behaviour. Unfortunately, the config go straight into controller.controller.Watch which is implementation details which is not possible to test. Do I miss anything here?

arkodg: hope this test helps https://github.com/envoyproxy/gateway/blob/07297df653e96751b5d1adffdc06c00b92aa431a/internal/provider/kubernetes/kubernetes_test.go#L1669

den3tsou: This test is a bit different from what I am doing because the other watch mode changes the client, so using the client to query can test the behavior. But the one I am currently working on doesn't change the behavior of the client. It changes the behavior of informers. I guess I miss something here. I will have a look again. If this doesn't work, I think the only solution is to change the test structure of E2E tests so that updating Envoy Gateway is possible

arkodg: for the verification steps, instead of using the client you can peek into the `ProviderResources` message to see if the resource should to shouldnt exist

den3tsou: After the investigation, I realise that adding predicate will not help in this case because reconciling will try to obtain everything for whatever event comes in. The direction of this PR is not correct. I will dig into more to see what I need to do in order to implement this feature within reconciling.

den3tsou: Hi @arkodg , I would like to get your opinion before proceeding. After the investigation, addtional logic is needed to be added into functions including `processGateways`, `processTLSRoutes`, `processHTTPRoutes`, `processGRPCRoutes`, `processTCPRoutes` and `processUDPRoutes` to have additional check to make sure that only resources in the namespaces with certain labels are processed. I can see that this will increase the complexity of the codebase. I just wonder if you are happy with me going head? or you have other solutions in mind.

arkodg: can we do it in a way that doesnt add too much complexity ? but the goal is what you described - `make sure that only resources in the namespaces with certain labels are processed`

den3tsou: This is a bit tricky. I will share what I've found, so you could help me to brainstorm some ideas.

There are two approach available at this moment

-  Initialise the k8s client to obtain all namespaces with certain labels
    * Pros:
       * No complex change for reconciler because configuring Manager is enough
    * Cons:
       * Users need to restart Envoy Gateway controller every time they add update the namespaces unless there is another goroutine or anything similar to check the namespace regularly and update the Manger config accordingly
- Use predicate and update all of logic inside reconciler to check if the objects' namespaces have certain labels
    * Pros:
       * Users don't need to redeploy Envoy Gateway controller every time they update namespaces
    * Cons:
       * All logic inside reconciler uses manager.client needs to have additional logic to check if the objects are under namepsace with certain labels. This may cause performance issue because every object will need to be checked.

These two solutions are what I've found so far. I've checked the `controller-runtime` to see if it has any features that I could workaround. Unfortunately, I coudln't find any solutions yet. Do you have any ideas?

den3tsou: Hi @arkodg  This is a bit tricky. I will share what I've found, so you could help me to brainstorm some ideas.

There are two approach available at this moment

-  Initialise the k8s client to obtain all namespaces with certain labels
    * Pros:
       * No complex change for reconciler because configuring Manager is enough
    * Cons:
       * Users need to restart Envoy Gateway controller every time they add update the namespaces unless there is another goroutine or anything similar to check the namespace regularly and update the Manger config accordingly
- Use predicate and update all of logic inside reconciler to check if the objects' namespaces have certain labels
    * Pros:
       * Users don't need to redeploy Envoy Gateway controller every time they update namespaces
    * Cons:
       * All logic inside reconciler uses manager.client needs to have additional logic to check if the objects are under namepsace with certain labels. This may cause performance issue because every object will need to be checked.

These two solutions are what I've found so far. I've checked the `controller-runtime` to see if it has any features that I could workaround. Unfortunately, I coudln't find any solutions yet. Do you have any ideas?

den3tsou: Hi @arkodg , any update on this. I am fine to decline this PR and work on other tasks. I've spent fair bit of time finding other solution. But this is more related to how reconciler works, there is no work around for it. All of code related to pull data from api server will need to be updated.

den3tsou: Thanks 👍 . I will start to implement it :-D

api/config/v1alpha1/envoygateway_types.go
```diff
@@ -177,15 +177,37 @@ type EnvoyGatewayKubernetesProvider struct {
...
+
+// KubernetesWatchModeType defines the type of KubernetesWatchMode
+type KubernetesWatchModeType string
+
 // KubernetesWatchMode holds the configuration for which input resources to watch and reconcile.
 type KubernetesWatchMode struct {
+	// Type indicates what watch mode to use. KubernetesWatchModeTypeNamespaces and
+	// KubernetesWatchModeTypeNamespaceSelectors are currently supported
+	// By default, when this field is unset or empty, Envoy Gateway will watch for input namespaced resources
+	// from all namespaces.
```
arkodg, line 197: needs a default tag here similar to https://github.com/envoyproxy/gateway/blob/91c0888ad63631440286264f18f1bd8d2ccab4cd/api/config/v1alpha1/envoyproxy_types.go#L169

internal/provider/kubernetes/routes.go
```diff
@@ -497,7 +586,26 @@ func (r *gatewayAPIReconciler) processUDPRoutes(ctx context.Context, gatewayName
 		return err
 	}
 
-	for _, udpRoute := range udpRouteList.Items {
+	udpRoutes := udpRouteList.Items
```
arkodg, line 589: can all this code be consolidated by using a helper func with a client.Object signature ?

internal/provider/kubernetes/predicates.go
```diff
@@ -42,6 +42,62 @@ func (r *gatewayAPIReconciler) hasMatchingController(obj client.Object) bool {
...
 
+// hasMatchingNamespaceLabels returns true if the namespace of provided object has
+// the provided labels or false otherwise.
+func (r *gatewayAPIReconciler) hasMatchingNamespaceLabels(labels []string) func(client.Object) bool {
+	return func(obj client.Object) bool {
+		ns := obj.GetNamespace()
+		ok, err := r.checkNamespaceLabels(ns)
+		if err != nil {
+			r.log.Error(
+				err, "fail to get Namespace",
```
arkodg, line 53:
```suggestion
				err, "failed to get Namespace labels",
```

internal/provider/kubernetes/controller.go
```diff
@@ -1165,20 +1226,30 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	}
 
 	// Only enqueue EnvoyProxy objects that match this Envoy Gateway's GatewayClass.
+	epPredicates := []predicate.Predicate{
+		predicate.ResourceVersionChangedPredicate{},
+		predicate.NewPredicateFuncs(r.hasManagedClass),
+	}
+	if len(ls) != 0 {
```
arkodg, line 1233: why is `ls` needed here, can't you just use `r.namespaceLabels` as a way to check

api/config/v1alpha1/envoygateway_types.go
den3tsou: Do I miss anything here? This is an optional field. It has to be defined with either `Namespaces` or `NamespaceSelectors`

internal/provider/kubernetes/routes.go
den3tsou: ahhh. Good catch. I didn't realise that `client.Object` can be used!

internal/provider/kubernetes/controller.go
den3tsou: good catch. updated

internal/provider/kubernetes/predicates.go
den3tsou: Fixed

internal/provider/kubernetes/routes.go
den3tsou: Unfortunately, `AuthenticaionFilter`, RateLimitFilter` and `unstructured.Unstructured` don't implment the common interface. The alternative I could do is to add a NamespaceGetter interface to simplify the code.

arkodg: you could use existing code for them, and add a `TODO`

api/config/v1alpha1/envoygateway_types.go
den3tsou: Hi @arkodg, do you mind helping me understand this?

den3tsou:
Hi @arkodg , I finally complete the integration tests. Would you please have a look again?

api/config/v1alpha1/envoygateway_types.go
arkodg: this should be `// +kubebuilder:default=Namespaces`

arkodg (approved):
one nit, else looks good, thanks @den3tsou !

arkodg:
lets merge this one after https://github.com/envoyproxy/gateway/pull/1534 has merged to reduce conflicts

AliceProxy (approved):
lgtm. @arkodg I see no conflicts after merging #1534 so I'm gonna go ahead and merge this now to prevent additional churn on this PR. Any outstanding nits can be resolved via a follow-up PR.

Issue #1647, opened 2023-07-11
Run `ValidateAll()` for Xds Resources before adding them to Xds Resource Table


arkodg:
*Description*:

Run the `ValidateAll()` method associated with each resource (https://github.com/envoyproxy/go-control-plane/blob/25fcba6ee652be7c29ee36eb1b7cdbe0fd0e3b43/envoy/config/endpoint/v3/endpoint.pb.validate.go#L49) before adding them into the Xds Resource Table https://github.com/envoyproxy/gateway/blob/ab7aed6ad997216853325a64bdb708064e6de594/internal/xds/types/resourceversiontable.go#L66

Here is an example where it should be added 
https://github.com/envoyproxy/gateway/blob/ab7aed6ad997216853325a64bdb708064e6de594/internal/xds/translator/translator.go#L88

It is done in some functions such as for the EnvoyPatchPolicy https://github.com/envoyproxy/gateway/blob/ab7aed6ad997216853325a64bdb708064e6de594/internal/xds/translator/jsonpatch.go#L73
but should be done everywhere

Ronnie-personal:
@arkodg  Could you assign the issue to me? I have created the PR.

Thanks!

PR #1662, opened 2023-07-16
chore: Validate before adding Xds resource

Ronnie-personal:
**What type of PR is this?**  Chore: Validate before adding Xds resource

**What this PR does / why we need it**: This PR calls Validate() before adding the resource to Xds resource table.

**Which issue(s) this PR fixes**: #1647 

Fixes #


Ronnie-personal:
> thanks for raising this PR @Ronnie-personal . Imo attaching a `ValidateAll()` right before inserting the resource into `tCtx.AddXdsResource` is something the developer might forget to do, instead should we add `ValidateAll()` inside the `tCtx.AddXdsResource` and change the signature of it to return a error code instead ?

How about this code change?
func (t *ResourceVersionTable) AddXdsResource(rType resourcev3.Type, xdsResource types.Resource) error {
	// Perform type switch to handle different types of xdsResource
	switch r := xdsResource.(type) {
	case *listenerv3.Listener:
		// Handle Type specific operations
		if resourceOfType, ok := xdsResource.(*listenerv3.Listener); ok {
			if err := resourceOfType.ValidateAll(); err != nil {
				return fmt.Errorf("validation failed for xds resource %+v, err:%v", r, err)
			}
		} else {
			return fmt.Errorf("failed to cast xds resource %+v to Listener type", r)
		}
	case *routev3.RouteConfiguration:
		// Handle Type specific operations
		if resourceOfType, ok := xdsResource.(*routev3.RouteConfiguration); ok {
			if err := resourceOfType.ValidateAll(); err != nil {
				return fmt.Errorf("validation failed for xds resource %+v, err:%v", r, err)
			}
		} else {
			return errors.New("failed to cast xds resource to RouteConfiguration type", r)
		}

	case *Type3:
		// Handle Type3 specific operations

	// Add more cases for other types as needed
	default:
		// Handle the case when the type is not recognized or supported
		fmt.Errorf("Unsupported resource type: %v", rType)
	}

	if t.XdsResources == nil {
		t.XdsResources = make(XdsResources)
	}
	if t.XdsResources[rType] == nil {
		t.XdsResources[rType] = make([]types.Resource, 0, 1)
	}

	t.XdsResources[rType] = append(t.XdsResources[rType], xdsResource)
	return nil
}

arkodg:
yeah @Ronnie-personal something similar to above logic is needed, you can probably use `rType resourcev3.Type` as in input to `switch` and then typecast from resource to the specific type before running `ValidateAll`

Ronnie-personal:
@arkodg 
The e2e ratelimit test failed. Please find detailed error at https://github.com/envoyproxy/gateway/actions/runs/5633849037/job/15263275737?pr=1662#step:6:1913.

This PR does not change e2e test code. Do you think the error is caused by any code change in this pull request?

Ronnie-personal:
@arkodg 

The existing tests failed.
I suspect the failure is caused by the ValidateAll() on the ClusterType.

Is it worth the effort to revert back the changes on ClusterType and test again?
Can you please advise on the next step?

internal/xds/types/resourceversiontable.go
```diff
@@ -68,7 +74,60 @@ func (t *ResourceVersionTable) GetXdsResources() XdsResources {
...
+		if resourceOfType, ok := xdsResource.(*tlsv3.Secret); ok {
+			if err := resourceOfType.ValidateAll(); err != nil {
+				return fmt.Errorf("validation failed for xds resource %+v, err:%v", xdsResource, err)
+			}
+		} else {
+			return fmt.Errorf("failed to cast xds resource %+v to RouteConfiguration type", xdsResource)
+		}
+
+	case resourcev3.EndpointType:
+		// TBD - ValidateAll() breaks existing test internal/cmd/egctl/translate_test
```
arkodg, line 110: thanks for catching this, lets raise a issue to track this

internal/xds/types/resourceversiontable.go
```diff
@@ -68,7 +74,60 @@ func (t *ResourceVersionTable) GetXdsResources() XdsResources {
...
+		if resourceOfType, ok := xdsResource.(*clusterv3.Cluster); ok {
+			if err := resourceOfType.ValidateAll(); err != nil {
+				return fmt.Errorf("validation failed for xds resource %+v, err:%v", xdsResource, err)
+			}
+		} else {
+			return fmt.Errorf("failed to cast xds resource %+v to RouteConfiguration type", xdsResource)
+		}
+	case resourcev3.RateLimitConfigType:
+		// Handle specific operations
+		// cfg resource from runner.go is the RateLimitConfig type from "github.com/envoyproxy/go-control-plane/ratelimit/config/ratelimit/v3", which does have validate function.
```
arkodg, line 124: this also raise a issue to track this

internal/xds/types/resourceversiontable_test.go
```diff
@@ -233,7 +278,7 @@ func TestAddOrReplaceXdsResource(t *testing.T) {
 			name: "inject-new-listener",
 			tableIn: &ResourceVersionTable{
 				XdsResources: XdsResources{
```
arkodg, line 280: was this an existing bug ?

arkodg:
LGTM @Ronnie-personal, this PR looks good, just added some minor comments, thanks again !

internal/xds/types/resourceversiontable_test.go
Ronnie-personal: Thanks for the review.
I believe the code was carried over from the main branch when I created my branch.
I cannot recall making any changes to it.

But I agree that 'resourcev3.ListenerType: []types.Resource{},' makes more sense. We need to test adding a new listener to the table, so we wouldn't need to initialize 'tableIn' with the 'testListener'.

arkodg (approved):

internal/xds/types/resourceversiontable.go
Ronnie-personal: Sure, I have opened https://github.com/envoyproxy/gateway/issues/1742. I will investigate further.

Ronnie-personal:
@arkodg 
Sorry, I pushed a minor change to fix the typo.
Here are the changes https://github.com/Ronnie-personal/gateway/commit/d886fe3f053d70cd49d37e8ace62370affcf0a01

arkodg (approved):

zirain (approved):

internal/xds/types/resourceversiontable.go
Ronnie-personal: Sure, I have opened https://github.com/envoyproxy/gateway/issues/1742. Here is the PR to fix the issue https://github.com/envoyproxy/gateway/pull/1758

Xunzhuo (approved):

PR #1663, opened 2023-07-17
Fix json patch issue

zhaohuabing:
Use string as JSONPatch value and override the default Marshal and UnMarshal functions.


internal/gatewayapi/testdata/envoypatchpolicy-valid.out.yaml
```diff
@@ -52,9 +52,48 @@ xdsIR:
...
+        value:
+          name: envoy.filters.http.oauth2
+          typed_config:
+            "@type": type.googleapis.com/envoy.extensions.filters.http.oauth2.v3.OAuth2Config
+            auth_scopes:
+            - openid
+            auth_type: BASIC_AUTH
+            authorization_endpoint: https://accounts.google.com/o/oauth2/v2/auth
+            credentials:
+              client_id: 250344188863-lddmgbasbdom9qpt1mr0rkln62281s6d.apps.googleusercontent.com
```
zirain, line 66: you'd better not use a real client_id, even it's test only.

PR #1674, opened 2023-07-19
fix: make envoyPatchPolicy status field optional

arkodg:


api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -28,7 +28,7 @@ type EnvoyPatchPolicy struct {
 	Spec EnvoyPatchPolicySpec `json:"spec"`
 
 	// Status defines the current status of EnvoyPatchPolicy.
-	Status EnvoyPatchPolicyStatus `json:"status"`
```
zirain, line 31: so this should be an pointer?

arkodg: good point, followed upstream APIs
https://github.com/kubernetes-sigs/gateway-api/blob/12dae809795674dfbca7b7637892a2f4027f6e80/apis/v1beta1/httproute_types.go#L43
but asked this question in the upstream slack channel, will get back on this ....

zhaohuabing (approved):

zhaohuabing (approved):
LGTM

Thanks for fixing this.

api/v1alpha1/envoypatchpolicy_types.go
zhaohuabing: I guess it's not a pointer because status is not needed when creating a new resource but it's still a necessary part of the resource.

zirain: this shoule not be a block, we can fix it later.

zirain (approved):

PR #1691, opened 2023-07-22
feat: EnvoyPatchPolicy Status

arkodg:
* compute the `Accepted` condition in the `gateway-api` layer

* compute the `Programmed` condition in the `xds-translator` layer

* Enhance Xds IR to hold the EnvoyPatchPolicy status

* Publish the EnvoyPatchPolicy status from the `xds-translator` layer

* `provider` subscribes to the status and writes it into the API server

Relates to #24


arkodg:
hey @Xunzhuo `gen-check` keeps failing because `make testdata` is unable to set the `LastTransitionTime` to `null`, any idea what is required to get this to work ?

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -17,6 +17,7 @@ const (
 )
 
 // +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
```
AliceProxy, line 20: Not something I'd want to hold up this PR for, but along with this addition, I'd love to see something along the lines of:
```
// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Programmed")].reason 
```
so you can see if your patch policy applied successfully at a glance. That's just an example, but you could tweak the selector to your liking to showcase the other possible states of the resource.
On that topic, resource age is also nice to have for the custom resources we're introducing. 
```
// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```

internal/gatewayapi/envoypatchpolicy.go
```diff
@@ -39,7 +46,54 @@ func (t *Translator) ProcessEnvoyPatchPolicies(envoyPatchPolicies []*egv1a1.Envo
...
+		policyIR := ir.EnvoyPatchPolicy{}
+		policyIR.Name = policy.Name
+		policyIR.Namespace = policy.Namespace
+		policyIR.Status = &policy.Status
+
+		// Append the IR
+		gwXdsIR.EnvoyPatchPolicies = append(gwXdsIR.EnvoyPatchPolicies, &policyIR)
+
+		// Ensure policy can only target a Gateway
+		if policy.Spec.TargetRef.Group != gwv1b1.GroupName || policy.Spec.TargetRef.Kind != KindGateway {
```
AliceProxy, line 73: Do we not want to add this check before adding it to the IR so that we can short-circuit?

internal/gatewayapi/envoypatchpolicy.go
```diff
@@ -39,7 +46,54 @@ func (t *Translator) ProcessEnvoyPatchPolicies(envoyPatchPolicies []*egv1a1.Envo
...
+				gwv1a2.PolicyConditionAccepted,
+				metav1.ConditionFalse,
+				gwv1a2.PolicyReasonInvalid,
+				message,
+			)
+			continue
+		}
+
+		// Ensure Policy and target Gateway are in the same namespace
+		if policy.Namespace != string(*targetNs) {
```
AliceProxy, line 87: Same question as the above comment for this one

internal/utils/ptr/ptr.go
```diff
@@ -0,0 +1,10 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package ptr
+
+func To[T any](in T) *T {
```
AliceProxy, line 8: We should use this to clean up a bunch of the misc test files at some point that declare a ton of ints just to pass the pointer when declareing xDS in-line.

AliceProxy (commented):
Small question but otherwise lgtm

api/v1alpha1/envoypatchpolicy_types.go
arkodg: thanks, added

internal/gatewayapi/envoypatchpolicy.go
arkodg: its post adding it to the IR w/o actually attaching a patch into the IR, so the xds-translator layer can publish the status w/o computing the patch

internal/gatewayapi/envoypatchpolicy.go
arkodg: same as https://github.com/envoyproxy/gateway/pull/1691#discussion_r1274218967

internal/utils/ptr/ptr.go
arkodg: sg, will raise a issue for this

internal/gatewayapi/envoypatchpolicy.go
AliceProxy: ahhh gotcha, thanks for the clarification

AliceProxy (approved):

qicz (approved):
LGTM

PR #1693, opened 2023-07-23
fix: the imported packages in pin.go are removed by goland

zhaohuabing:
Add a [blank](https://go.dev/ref/spec#Blank_identifier) identifier to pin.go so the imported tool packages won't be automatically deleted by  Goland.


zirain (approved):

Xunzhuo (approved):
Thanks

PR #1702, opened 2023-07-24
fix: Ignore ports in host headers when matching against HTTPRoute hostnames

### LINKED ISSUE #1687: Traffic flows with `Host: <host>:<port>` not allowed when hostname is set in HTTPRoute not Listener
*Description*:

Related https://github.com/envoyproxy/gateway/pull/1478#issuecomment-1644126796

jackkleeman:
**What type of PR is this?**

fix

We need to be able to match on hostnames in httproute (that are more specific than hostnames in the listener) while also ignoring ports in host headers. Currently we do not ignore ports, and there is no workaround because hostnames may not contain ports, and nor can headermatch values. There are two approaches to solving this:
1. We could change the authority header match that currently exists to be some kind of regex which allows ports
2. We could create xds virtualhosts for each of the hostnames in the httproutes

This PR is the second approach

NB: This PR does **not** allow 'fallthrough' from one route hostname to a less specific one (eg from google.com to *.com) when there is no match for the first hostname, as envoy does not fall through after a virtual host is selected. This **is a breaking change of behaviour** as currently the less specific matcher will simply come later in the list of routes under the same virtual host. However, this is moving in line with the spec as discussed in https://github.com/kubernetes-sigs/gateway-api/discussions/2294

Fixes #1687

### DIFF
--- a/internal/extension/testutils/hooks.go
+++ b/internal/extension/testutils/hooks.go
@@ -80,7 +80,7 @@ func (c *XDSHookClient) PostRouteModifyHook(route *routeV3.Route, routeHostnames
 func (c *XDSHookClient) PostVirtualHostModifyHook(vh *routeV3.VirtualHost) (*routeV3.VirtualHost, error) {
 	// Only make the change when the VirtualHost's name matches the expected testdata
 	// This prevents us from having to update every single testfile.out
-	if vh.Name == "extension-post-xdsvirtualhost-hook-error" {
+	if vh.Name == "extension-post-xdsvirtualhost-hook-error-*" {
 		return nil, fmt.Errorf("extension post xds virtual host hook error")
 	} else if vh.Name == "extension-listener" {
 		// Setup a new VirtualHost to avoid operating directly on the passed in pointer for better test coverage that the
--- a/internal/extension/testutils/hooks.go
+++ b/internal/extension/testutils/hooks.go
@@ -80,7 +80,7 @@ func (c *XDSHookClient) PostRouteModifyHook(route *routeV3.Route, routeHostnames
 func (c *XDSHookClient) PostVirtualHostModifyHook(vh *routeV3.VirtualHost) (*routeV3.VirtualHost, error) {
 	// Only make the change when the VirtualHost's name matches the expected testdata
 	// This prevents us from having to update every single testfile.out
-	if vh.Name == "extension-post-xdsvirtualhost-hook-error-*" {
+	if vh.Name == "extension-post-xdsvirtualhost-hook-error/*" {
 		return nil, fmt.Errorf("extension post xds virtual host hook error")
 	} else if vh.Name == "extension-listener" {
 		// Setup a new VirtualHost to avoid operating directly on the passed in pointer for better test coverage that the
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -488,26 +488,6 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 
 		var perHostRoutes []*ir.HTTPRoute
 		for _, host := range hosts {
-			var headerMatches []*ir.StringMatch
-
-			// If the intersecting host is more specific than the Listener's hostname,
-			// add an additional header match to all of the routes for it
-			if host != "*" && (listener.Hostname == nil || string(*listener.Hostname) != host) {
-				// Hostnames that are prefixed with a wildcard label (*.)
-				// are interpreted as a suffix match.
-				if strings.HasPrefix(host, "*.") {
-					headerMatches = append(headerMatches, &ir.StringMatch{
-						Name:   ":authority",
-						Suffix: StringPtr(host[2:]),
-					})
-				} else {
-					headerMatches = append(headerMatches, &ir.StringMatch{
-						Name:  ":authority",
-						Exact: StringPtr(host),
-					})
-				}
-			}
-
 			for _, routeRoute := range routeRoutes {
 				// If the redirect port is not set, the final redirect port must be derived.
 				if routeRoute.Redirect != nil && routeRoute.Redirect.Port == nil {
@@ -529,8 +509,9 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 
 				hostRoute := &ir.HTTPRoute{
 					Name:                  fmt.Sprintf("%s-%s", routeRoute.Name, host),
+					Hostname:              host,
 					PathMatch:             routeRoute.PathMatch,
-					HeaderMatches:         append(headerMatches, routeRoute.HeaderMatches...),
+					HeaderMatches:         routeRoute.HeaderMatches,
 					QueryParamMatches:     routeRoute.QueryParamMatches,
 					AddRequestHeaders:     routeRoute.AddRequestHeaders,
 					RemoveRequestHeaders:  routeRoute.RemoveRequestHeaders,
@@ -544,6 +525,9 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 					RequestAuthentication: routeRoute.RequestAuthentication,
 					RateLimit:             routeRoute.RateLimit,
 					ExtensionRefs:         routeRoute.ExtensionRefs,
+				}
+				if hostRoute.Hostname == "*" {
+					hostRoute.Hostname = ""
 				}
 				// Don't bother copying over the weights unless the route has invalid backends.
 				if routeRoute.BackendWeights.Invalid > 0 {
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -526,9 +526,6 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 					RateLimit:             routeRoute.RateLimit,
 					ExtensionRefs:         routeRoute.ExtensionRefs,
 				}
-				if hostRoute.Hostname == "*" {
-					hostRoute.Hostname = ""
-				}
 				// Don't bother copying over the weights unless the route has invalid backends.
 				if routeRoute.BackendWeights.Invalid > 0 {
 					hostRoute.BackendWeights = routeRoute.BackendWeights
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -30,6 +30,7 @@ var (
 	ErrTLSServerCertEmpty            = errors.New("field ServerCertificate must be specified")
 	ErrTLSPrivateKey                 = errors.New("field PrivateKey must be specified")
 	ErrHTTPRouteNameEmpty            = errors.New("field Name must be specified")
+	ErrHTTPRouteHostnameEmpty        = errors.New("field Hostname must be specified")
 	ErrHTTPRouteMatchEmpty           = errors.New("either PathMatch, HeaderMatches or QueryParamMatches fields must be specified")
 	ErrRouteDestinationHostInvalid   = errors.New("field Address must be a valid IP address")
 	ErrRouteDestinationPortInvalid   = errors.New("field Port specified is invalid")
@@ -237,7 +238,7 @@ type HTTPRoute struct {
 	// Name of the HTTPRoute
 	Name string `json:"name" yaml:"name"`
 	// Hostname that the route matches against
-	Hostname string `json:"hostname,omitempty" yaml:"hostname,omitempty"`
+	Hostname string `json:"hostname" yaml:"hostname,omitempty"`
 	// PathMatch defines the match conditions on the path.
 	PathMatch *StringMatch `json:"pathMatch,omitempty" yaml:"pathMatch,omitempty"`
 	// HeaderMatches define the match conditions on the request headers for this route.
@@ -307,6 +308,9 @@ func (h HTTPRoute) Validate() error {
 	var errs error
 	if h.Name == "" {
 		errs = multierror.Append(errs, ErrHTTPRouteNameEmpty)
+	}
+	if h.Hostname == "" {
+		errs = multierror.Append(errs, ErrHTTPRouteHostnameEmpty)
 	}
 	if h.PathMatch == nil && (len(h.HeaderMatches) == 0) && (len(h.QueryParamMatches) == 0) {
 		errs = multierror.Append(errs, ErrHTTPRouteMatchEmpty)
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -236,6 +236,8 @@ type BackendWeights struct {
 type HTTPRoute struct {
 	// Name of the HTTPRoute
 	Name string `json:"name" yaml:"name"`
+	// Hostname that the route matches against
+	Hostname string `json:"hostname,omitempty" yaml:"hostname,omitempty"`
 	// PathMatch defines the match conditions on the path.
 	PathMatch *StringMatch `json:"pathMatch,omitempty" yaml:"pathMatch,omitempty"`
 	// HeaderMatches define the match conditions on the request headers for this route.
--- a/internal/xds/translator/translator.go
+++ b/internal/xds/translator/translator.go
@@ -145,19 +145,35 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 			}
 		}
 
-		// Allocate virtual host for this httpListener.
-		// 1:1 between IR HTTPListener and xDS VirtualHost
-		vHost := &routev3.VirtualHost{
-			Name:    httpListener.Name,
-			Domains: httpListener.Hostnames,
-		}
 		protocol := DefaultProtocol
 		if httpListener.IsHTTP2 {
 			protocol = HTTP2
 		}
 
+		// store virtual hosts by domain
+		vHosts := map[string]*routev3.VirtualHost{}
+		// keep track of order by using a list as well as the map
+		var vHostsList []*routev3.VirtualHost
+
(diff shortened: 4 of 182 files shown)


jackkleeman:
cc @arkodg

internal/cmd/egctl/testdata/translate/in/envoy-patch-policy.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: gateway.networking.k8s.io/v1beta1
 kind: GatewayClass
 metadata:
   name: eg
+  namespace: default
```
jackkleeman, line 5: this is needed for tests to pass 🤷

internal/cmd/egctl/translate_test.go
```diff
@@ -185,6 +185,7 @@ func TestTranslate(t *testing.T) {
 			output:       yamlOutput,
 			resourceType: string(AllEnvoyConfigType),
 			expect:       true,
+			extraArgs:    []string{"--add-missing-resources"},
```
jackkleeman, line 188: this is needed for tests to pass 🤷

internal/gatewayapi/route.go
```diff
@@ -486,53 +484,37 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
...
-						Name:   ":authority",
-						Suffix: StringPtr(host[2:]),
-					})
-				} else {
-					headerMatches = append(headerMatches, &ir.StringMatch{
-						Name:  ":authority",
-						Exact: StringPtr(host),
-					})
-				}
+		for _, routeRoute := range routeRoutes {
```
arkodg, line 493: looks like `perHostRoutes` is no longer needed, if thats the case, will it be simpler if the code looked lik
```
for _, routeRoute := range routeRoutes {
    routeRoute.Hostnames = hosts
```

internal/xds/translator/translator.go
```diff
@@ -125,65 +126,150 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
+				for _, hostname := range httpRoute.Hostnames {
+					routesByHostname[hostname] = append(routesByHostname[hostname], httpRoute)
+				}
 			}
+		}
+		hostnames := make([]string, 0, len(routesByHostname))
+		for hostname := range routesByHostname {
+			hostnames = append(hostnames, hostname)
+		}
+		// Sort according to rules in https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteSpec
```
arkodg, line 151: all sorting should happen before hand
https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/sort.go

internal/xds/translator/translator.go
```diff
@@ -125,65 +126,150 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
-			Name:    httpListener.Name,
-			Domains: httpListener.Hostnames,
-		}
 		protocol := DefaultProtocol
 		if httpListener.IsHTTP2 {
 			protocol = HTTP2
 		}
 
-		// Check if an extension is loaded that wants to modify xDS Routes after they have been generated
+		routesByHostname := make(map[string][]*ir.HTTPRoute, len(httpListener.Routes))
```
arkodg, line 134: trying to understand why we need this map (sidenote - maps lack order, making any sort of comparison really hard)
instead why not just
```
 		// 1:1 between IR HTTPRoute and xDS VirtualHost
 		vHost := &routev3.VirtualHost{
 			Name:    httpRoute. Name,
 			Domains: httpRoute.Hostnames, // this should be an intersection of listener and route hostnames from the gateway api layer
 		}
```

jackkleeman: a virtual host needs to contain all routes that match into it eg if the vhost domain is *.google.com we need routes defined for a.google.com and b.google.com to be duplicated in here - see line 192. as such, virtual hosts domains need to be pretty specific 

if the host had two route hostnames *.google.com and example.com - this would not be an appropriate domain for the virtual host, as a.google.com should not go in this virtual host.

arkodg: there is currently a order defined in envoy
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost
virtual host with domains with exact match such as `a.example.com` will be selected over another virtual domain with a domain set to `*.example.com`

These domains should hopefully be computed correctly in the gateway api layer.
Lets say we have 1 gateway with hostnames set to - `*.example.com`
And 3 HTTPRoutes set to it each with hostnames set to `nil`, `a.example.com`, `b.example.com` respectively
For this case the gateway api should populate the IRRoute.Hostnames to `*.example.com`, `a.example.com` & `b.example.com` for each IR Route respectively.
When these map to xds VirtualHost, Envoy will perform the best match/longest match.

So I guess this boils down to, is https://github.com/envoyproxy/gateway/blob/8b52159c21b1c64ee42ed222bba6e91a343019ab/internal/gatewayapi/helpers.go#L314 doing the right thing ?

internal/cmd/egctl/translate_test.go
arkodg: ummm have you tried running `make testdata` , in auto generates the golden testdata 😄

internal/xds/translator/translator.go
jackkleeman: let me try and give a better example of the case i am concerned about. maybe i am getting myself mixed up, but let me know what you think

lets say i have a route `first-route` with domain `*` and no other matches
and another route `second-route` with domain `foo.example.com` which matches path prefix `/bar`

under the domain order rules, `second-route` takes precedence over `first-route` as it has a non wildcard domain. however, that doesn't mean that `first-route` can never match on host `foo.example.com`! if i request `foo.example.com/baz` then i should not match against `second-route`, and i need to instead match `first-route`. The way this is achieved in the current implementation with one virtual host is that there is a virtual host for the listener ie `*` containing `second-route` and `first-route`, and then `second-route` has an authority check so that even though it is ordered before `first-route`, it wont match unintended domains

in a world with multiple virtual hosts, envoy will match against the `foo.example.com` virtual host, because, as you mentioned, it is more specific. in the `/baz` request case, it will not match against `second-route`, and if `first-route` is not present, it will 404. we cannot fall through between virtual hosts, to my understanding - this is the critical difference from the current approach. once a virtual host is selected, all possible routes need to be present in that virtual host. this is why we need to duplicate `first-route` into the `foo.example.com` virtual host and the `*` virtual host (which should only contain `first-route` in this case)

arkodg: Ah, I see the issue, your statement
```
under the domain order rules, second-route takes precedence over first-route as it has a non wildcard domain. however, that doesn't mean that first-route can never match on host foo.example.com! if i request foo.example.com/baz then i should not match against second-route, and i need to instead match first-route. 
```

does match the intended precedence based outcome defined in the spec, 
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute
rather than a `hostname` ownership approach
```
In the event that multiple HTTPRoutes specify intersecting hostnames (e.g. overlapping wildcard matching and exact matching hostnames), precedence must be given to rules from the HTTPRoute with the largest number of:

Characters in a matching non-wildcard hostname.
Characters in a matching hostname.
If ties exist across multiple Routes, the matching precedence rules for HTTPRouteMatches takes over.
```

a. have we crosschecked if xds virtualHost cannot fall back to a less specific domain match, after a missing route match ?

this logic inside the xds layer seems complex, and will be hard to maintain imo, we've managed to deal with all the complexity in the gateway-api and IR layers and keep xds as simple as possible.
is it possible to move the route duplication logic into gateway api layer if a. is not possible ?

arkodg: @zirain @howardjohn im guessing istio is going to have this same problem

howardjohn: Hmm never considered it...

howardjohn: Actually re-reading it I think Istio/envoy handles it correctly. You are not supposed to "fall back to a less specific domain match, after a missing route match ?", and Envoy will not

arkodg: @howardjohn from https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute the Gateway API spec does define HTTPRoute hostname matching in terms of `precedence`, not `exclusivity`

howardjohn: To me its ambiguous. I read it as once you match you should not backtrack, but I can see how you could intepret it otherwise.

This is what leads me to think that:

> If ties exist across multiple Routes, the matching precedence rules for HTTPRouteMatches takes over.

Which implies you only do this on ties, to me. Also I think most LBs work this way so I made that assumptions.

Lets clear it up in the spec and conformance?

arkodg: sure, raising an upstream issue, thanks for cross checking

arkodg: raised https://github.com/kubernetes-sigs/gateway-api/discussions/2294

jackkleeman: oh, this is interesting! i agree that things would be a lot easier if we had exclusivity on domains like envoy essentially intends. this precedence things is a bit of a pain - but its the current implementation's behaviour so I felt best to keep it

jackkleeman: ok fab! ill update this

jackkleeman: thinking a bit more @arkodg , i think as long as there can be a 'tie' between different httproutes that eg both match on google.com or both match on *, then we need to:
1. merge routes with the same domain into one virtual host, with the routes in the correct order within that VH to reflect the httproutematch ordering rules - i think we can probably do this as we convert into IR, so a bit earlier than in this PR
2. a virtual host per domain, not just per route (with multiple domains) because the above merging needs to be done per conflicting domain

internal/extension/testutils/hooks.go
```diff
@@ -80,7 +80,7 @@ func (c *XDSHookClient) PostRouteModifyHook(route *routeV3.Route, routeHostnames
 func (c *XDSHookClient) PostVirtualHostModifyHook(vh *routeV3.VirtualHost) (*routeV3.VirtualHost, error) {
 	// Only make the change when the VirtualHost's name matches the expected testdata
 	// This prevents us from having to update every single testfile.out
-	if vh.Name == "extension-post-xdsvirtualhost-hook-error" {
+	if vh.Name == "extension-post-xdsvirtualhost-hook-error-*" {
```
arkodg, line 83: cc @AliceProxy

internal/gatewayapi/route.go
```diff
@@ -488,26 +488,6 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 
 		var perHostRoutes []*ir.HTTPRoute
 		for _, host := range hosts {
```
arkodg, line 490: is this per host loop needed anymore ? imo, dont think so

internal/gatewayapi/route.go
```diff
@@ -529,8 +509,9 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 
 				hostRoute := &ir.HTTPRoute{
 					Name:                  fmt.Sprintf("%s-%s", routeRoute.Name, host),
```
arkodg, line 511: is this per host name needed anymore ? imo, dont think so

internal/gatewayapi/route.go
```diff
@@ -545,6 +526,9 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 					RateLimit:             routeRoute.RateLimit,
 					ExtensionRefs:         routeRoute.ExtensionRefs,
 				}
+				if hostRoute.Hostname == "*" {
```
arkodg, line 529: is this needed ?

internal/gatewayapi/route.go
```diff
@@ -529,8 +509,9 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 
 				hostRoute := &ir.HTTPRoute{
 					Name:                  fmt.Sprintf("%s-%s", routeRoute.Name, host),
+					Hostname:              host,
```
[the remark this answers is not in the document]
arkodg: rethinking this piece again, we could assign `HTTPRoute.Spec.Hostnames` to `ir.HTTPRoute.Hostnames` and map this exact value to the xdsVirtualHost.Domains

* the `Gateway.Spec.Listener[0].Hostnames` (which maps to `ir.HTTPListener.Hostnames`) can be only used to set the exact filter chain match (SNI match) in the xds listener . For HTTP, we can ignore this step, since our logic guarantees that the route hostnames is a subset of listener hostnames.

* we dont need to deal with any sorting here, since xds/envoy has inbuilt rules for specific and wildcard matches which align with the gateway api spec

internal/gatewayapi/route.go
jackkleeman: not necessarily

internal/gatewayapi/route.go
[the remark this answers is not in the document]
arkodg: I dont think `httproutes that eg both match on google.com or both match on *, then we need to` is possible. 

the spec here https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute
only talks about about overlap in terms of a specific match `foo.example.com` and wildcard match `*.example.com`, which is being handled by xds vhost semantics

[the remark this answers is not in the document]
jackkleeman: from that spec:
>If ties exist across multiple Routes, the matching precedence rules for HTTPRouteMatches takes over.

how can a tie exist where both hostnames match without them being the same hostname:
1. if they are both non wildcard, and both match the domain, both hostnames must be the domain
2. if they are both wildcard, and both match the domain, and both have the same number of characters, then they must also be the same, because a valid wilidcard is '*<suffix>', and you cannot have two different matching suffixes with the same number of characters

so, my reading of the spec is that the fact that it considers ties must mean it is valid to have two routes with the same hostname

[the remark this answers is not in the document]
arkodg: based on https://github.com/kubernetes-sigs/gateway-api/discussions/2294#discussioncomment-6727871, I dont think the ties you are describing are possible, will ask the q in the upstream discussion thread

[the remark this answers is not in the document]
jackkleeman: ok, based on my reading of the discussion there, the approach here is right?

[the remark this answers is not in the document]
arkodg: yeah, you're right on the logic, if the hostnames are same, we need to merge the rules

internal/xds/translator/translator.go
```diff
@@ -145,19 +145,35 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
 
+		// store virtual hosts by domain
+		vHosts := map[string]*routev3.VirtualHost{}
+		// keep track of order by using a list as well as the map
+		var vHostsList []*routev3.VirtualHost
+
 		// Check if an extension is loaded that wants to modify xDS Routes after they have been generated
 		for _, httpRoute := range httpListener.Routes {
+			hostname := httpRoute.Hostname
+			if hostname == "" {
```
arkodg, line 161: if we rm https://github.com/envoyproxy/gateway/pull/1702/files#r1294798367 guessing this won't be needed as well ?

internal/xds/translator/translator.go
```diff
@@ -145,19 +145,35 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
+				hostname = "*"
+			}
+
+			// 1:1 between IR HTTPRoute Hostname and xDS VirtualHost.
+			vHost := vHosts[hostname]
+			if vHost == nil {
+				// Allocate virtual host for this httpRoute.
+				vHost = &routev3.VirtualHost{
+					Name:    fmt.Sprintf("%s-%s", httpListener.Name, hostname),
```
arkodg, line 170: we've moved naming separators to `/` instead of `-` , can we change to `/` here

arkodg:
few minor comments, overall LGTM, thanks for hanging in there, and bringing clarity on precedence matching !

internal/gatewayapi/route.go
arkodg: still needed, we are building perHostRoutes for now

internal/ir/xds.go
```diff
@@ -206,6 +206,8 @@ type BackendWeights struct {
 type HTTPRoute struct {
 	// Name of the HTTPRoute
 	Name string `json:"name" yaml:"name"`
+	// Hostname that the route matches against
```
arkodg, line 209: since this is critical, lets enhance the `Validate` func to check for `len(Hostname) > 1`

internal/ir/xds.go
```diff
@@ -206,6 +206,8 @@ type BackendWeights struct {
...
+	Hostname string `json:"hostname,omitempty" yaml:"hostname,omitempty"`
```
arkodg, line 210: `omitempty` is not needed here since its a required field

internal/xds/translator/translator.go
jackkleeman: yeah

arkodg:
waiting for a LGTM for another reviewer

zirain (approved):

Issue #9, opened 2022-04-14
Support Multi-Cluster Services using ServiceImport as a BackendRef


danehans:
As a Kubernetes cluster administrator, I would like Gateway to support Kubernetes services that run in multiple clusters. Gateway should support multi-cluster service networking methods defined by [sig-multicluster](https://github.com/kubernetes/community/blob/master/sig-multicluster/README.md) but not preclude other methods.

xref: https://github.com/envoyproxy/gateway/pull/7

arkodg:
linking relevant upstream Gateway API design proposal on using `ServiceImport` as a valid `BackendRef`
https://docs.google.com/document/d/1akwzBKtMKkkUV8tX-O7tPcI4BPMOLZ-gmS7Iz-7AOQE/edit#

arkodg:
supporting  `ServiceImport` in Gateway API has transitioned to experimental https://gateway-api.sigs.k8s.io/geps/gep-1748/ and this project should implement this API soon

Xunzhuo:
https://github.com/kubernetes-sigs/gateway-api/pull/1922

Make it extented now

arkodg:
assigning this to @tanujd11 who's proactively picked this one up, thanks !

tanujd11:
Sure @arkodg, thanks

arkodg:
reopening this issue to track completion of docs

tanujd11:
Hey @arkodg can this be safely closed with the docs PR being merged?

PR #1705, opened 2023-07-25
Feat/service import backend support

tanujd11:
**What type of PR is this?**
 Feature

Adds serviceImport Backend for Routes.

Fixes https://github.com/envoyproxy/gateway/issues/9


arkodg:
thanks for picking this feature up @tanujd11 ! would be great if this land in v0.6
quick q - isn't this feature blocked on https://github.com/envoyproxy/gateway/pull/1494 which is adding EndpointSlice support ?

tanujd11:
@arkodg , I think this can go in with using ClusterSetIP type in multi cluster service import. EndpointSlice can be used once that PR gets in. It will be similar to service backend.

arkodg:
Rethinking, you're right @tanujd11, this PR can go in by itself.
Will review this soon once we start working on v0.6 features.

charts/gateway-helm/crds/multicluster-svc.yaml
```diff
@@ -0,0 +1,151 @@
+apiVersion: apiextensions.k8s.io/v1
```
arkodg, line 1: hey lets rm the CRD from here, and add it as part of docs

internal/gatewayapi/translator_test.go
```diff
@@ -94,6 +95,21 @@ func TestTranslate(t *testing.T) {
 				},
 			)
 
+			resources.ServiceImports = append(resources.ServiceImports,
```
arkodg, line 98: prefer if we dont add this by default here, but in the YAML testdata input, since this resource is not the common case

internal/provider/kubernetes/kubernetes_test.go
```diff
@@ -88,8 +88,10 @@ func startEnv() (*envtest.Environment, *rest.Config, error) {
 	log.SetLogger(zap.New(zap.WriteTo(os.Stderr), zap.UseDevMode(true)))
 	gwAPIs := filepath.Join("..", "..", "..", "charts", "gateway-helm", "crds", "gatewayapi-crds.yaml")
 	egAPIs := filepath.Join("..", "..", "..", "charts", "gateway-helm", "crds", "generated")
+	mcsAPIs := filepath.Join("..", "..", "..", "charts", "gateway-helm", "crds", "multicluster-svc.yaml")
```
arkodg, line 91: this could live under `internal/provider/kubernetes/testdata/mcs-crds/`

arkodg:
did a first pass, it looks good ! added some minor comments

Xunzhuo (commented):
This is awesome, but there are some conflcts : )

It would be a nice feat in v0.6.0.

arkodg:
ptal @envoyproxy/gateway-maintainers, a decision also needs be made here on whether we want the ServiceImport API to be opt in or not

tanujd11:
@arkodg, if we are removing the serviceImport CRD from EG chart, then it automatically becomes the a Opt-in feature where user needs to install the CRD to use this feature. I can add a validation to check if the CRD exist before someone uses serviceImport as part of the backend. for example if the user has mentioned serviceImport as backendRef and the CRD does not exist, we can reject the config with status `backendRef kind does not exist`

arkodg:
> @arkodg, if we are removing the serviceImport CRD from EG chart, then it automatically becomes the a Opt-in feature where user needs to install the CRD to use this feature. I can add a validation to check if the CRD exist before someone uses serviceImport as part of the backend. for example if the user has mentioned serviceImport as backendRef and the CRD does not exist, we can reject the config with status `backendRef kind does not exist`

great idea, +1 to this, but from an implementation perspective, won't the `provider` layer error out if it tries to `Watch` the `ServiceImport` resource w/o have a CRD installed ?

arkodg (approved):
LGTM, thanks for adding this !

tanujd11:
> great idea, +1 to this, but from an implementation perspective, won't the `provider` layer error out if it tries to `Watch` the `ServiceImport` resource w/o have a CRD installed ?

You are right, the current one is throwing error. Should we implement a way to check the presence of CRD for watching also. Or is there any other way in your mind to tackle this

arkodg:
@tanujd11 can you share the error you are seeing ?

one way to solve the opt in problem is add a knob in[ EnvoyGateway.ExtensionApis](https://github.com/envoyproxy/gateway/blob/b08d33579f80f0acf8c4e03febe0b39112cc611f/api/config/v1alpha1/envoygateway_types.go#L132)

Xunzhuo:
LGTM after conflicts are resolved and CI passed.

tanujd11:
@arkodg , can you take a relook at this work? I have added the knob at the place suggested by you. I will work on the doc in a subsequent PR

zirain (dismissed):

arkodg:
had a chat with @tanujd11 who is trying to implement the functionality w/o having to add an explicit API knob

Xunzhuo (dismissed):
Looks like this PR has been rebased a couple of times, maybe we should get this in higher priority, in case more rebases. Thanks @tanujd11, mind having another rebase to fix conflicts?

internal/provider/kubernetes/controller.go
```diff
@@ -1155,6 +1180,7 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
 
 // watchResources watches gateway api resources.
 func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.Manager, c controller.Controller) error {
+	log.SetLogger(zap.New(zap.WriteTo(os.Stderr), zap.UseDevMode(true)))
```
arkodg, line 1183: is this needed ? can we use `r.log` directly ?

internal/provider/kubernetes/controller.go
```diff
@@ -1422,3 +1464,26 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
 
 	return nil
 }
+
+// mcsServiceImportFound checks for the existence of the ServiceImport CRD in k8s APIServer before watching it
+func (r *gatewayAPIReconciler) mcsServiceImportFound(mgr manager.Manager) bool {
```
arkodg, line 1469: nit: ServiceImportCRDExists

internal/provider/kubernetes/controller.go
```diff
@@ -1249,6 +1275,22 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
+	serviceImportFound := r.mcsServiceImportFound(mgr)
+	if !serviceImportFound {
+		r.log.Info("ServiceImport CRD not found, skipping ServiceImport watch")
+	}
+
+	// Watch ServiceImport CRUDs and process affected *Route objects.
+	if serviceImportFound {
+		if err := c.Watch(
+			source.Kind(mgr.GetCache(), &mcsapi.ServiceImport{}),
+			&handler.EnqueueRequestForObject{},
```
arkodg, line 1287: lets use `handler.EnqueueRequestsFromMapFunc(r.enqueueClass)` here instead, similar to what is done for all other `Watch` calls, might require a `rebase`, since this change was done recently

arkodg:
hey @tanujd11 just took a peek at your PR, the approach of using the o/p. of `discoveryClient.ServerResources()` to check if the service import API exists in the API server is a great way to decide on whether to watch the resource or not.

arkodg:
seeing some test yaml fail, recommend running `make testdata` to fix it

internal/provider/kubernetes/controller.go
tanujd11: removed it, this is not required.

internal/provider/kubernetes/controller.go
```diff
@@ -1249,6 +1271,22 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
+	}
+
+	// Watch ServiceImport CRUDs and process affected *Route objects.
+	if serviceImportCRDExists {
+		if err := c.Watch(
+			source.Kind(mgr.GetCache(), &mcsapi.ServiceImport{}),
+			handler.EnqueueRequestsFromMapFunc(r.enqueueClass),
+			predicate.NewPredicateFuncs(r.validateServiceImportForReconcile)); err != nil {
+			// ServiceImport is not available in the cluster, skip the watch and not throw error.
+			r.log.Info("Unable to watch not ServiceImport: %s", err.Error())
```
arkodg, line 1286:
```suggestion
			r.log.Info("Unable to watch ServiceImport: %s", err.Error())
```

internal/provider/kubernetes/controller.go
```diff
@@ -1422,3 +1460,26 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
...
+
+// serviceImportCRDExists checks for the existence of the ServiceImport CRD in k8s APIServer before watching it
+func (r *gatewayAPIReconciler) serviceImportCRDExists(mgr manager.Manager) bool {
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	if err != nil {
+		r.log.Error(err, "Failed to create discovery client")
+	}
+	apiResourceList, err := discoveryClient.ServerPreferredResources()
+	if err != nil {
+		r.log.Error(err, "Failed to get API resource list")
```
arkodg, line 1472:
```suggestion
		r.log.Error(err, "failed to get API resource list")
```

internal/provider/kubernetes/controller.go
```diff
@@ -1422,3 +1460,26 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
 
 	return nil
 }
+
+// serviceImportCRDExists checks for the existence of the ServiceImport CRD in k8s APIServer before watching it
+func (r *gatewayAPIReconciler) serviceImportCRDExists(mgr manager.Manager) bool {
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	if err != nil {
+		r.log.Error(err, "Failed to create discovery client")
```
arkodg, line 1468:
```suggestion
		r.log.Error(err, "failed to create discovery client")
```

internal/provider/kubernetes/controller.go
arkodg, line 1286:
```suggestion
			r.log.Info("unable to watch not serviceImport: %s", err.Error())
```

arkodg, line 1286:
```suggestion
			r.log.Info("unable to watch ServiceImport: %s", err.Error())
```

arkodg (approved):
LGTM, added some minor nits around error string
thanks for adding this !

zirain:
can you add a user face doc in following PR.

arkodg (approved):

PR #1714, opened 2023-07-26
update cherry-pick GHA for v0.5

arkodg:

### DIFF
--- a/.github/workflows/cherrypick.yaml
+++ b/.github/workflows/cherrypick.yaml
@@ -6,23 +6,23 @@ on:
     types: ["closed"]
 
 jobs:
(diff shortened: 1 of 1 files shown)


zirain (approved):

Xunzhuo (approved):
If so, we can not cp commits into v0.4 and `cherrypick/release-v0.4` won`t work anymore ?

Thoughts on adding instead of replacing ?

zirain:
maybe we change upstream tool, make it more generic?

arkodg:
@Xunzhuo, in the future once we support patch versions we will need to keep steps for each minor release, right now we are only supporting v0.x.0 releases, I think it's safe to replace

Issue #1718, opened 2023-07-27
Additional proxy level configuration options


ardikabs:
*Description*:

Currently, my company is interested to spike the Envoy Gateway as our Ingress Controller replacement, therefore we do some load testing and apparently need to configure additional options to the Envoy Proxy.
So, I would like to know the chance of providing additional Envoy proxy options such as `--concurrency`, referring to the Istio implementation such as [ProxyConfig](https://istio.io/latest/docs/reference/config/networking/proxy-config/).

I would appreciate it if you could inform me whether this enhancement is already planned for the next development phase. If not, I am eager to contribute to the project and would like to get in touch for further collaboration.

zirain:
thanks @ardikabs , this's something EG must supported.

zirain:
it's ok, let get it step by step.

arkodg:
@ardikabs curious why you need to tune `concurrency`, by default `concurrency` is equal to the number of `hardware threads` https://www.envoyproxy.io/docs/envoy/latest/operations/cli
are you trying to reduce this number for high core machines ?

ardikabs:
@arkodg
We are under a circumstance where everything needs to be controlled, including the resources that have been taken by the workload. Thus, when we are setting it with the necessary limits such as 4 cores, Envoy won't run with that value according to the `--concurrency` default value which will preserve all hardware threads in the machine. Following https://www.envoyproxy.io/docs/envoy/latest/faq/performance/how_to_benchmark_envoy, we need to set the concurrency to match the number of cores/threads that are being reserved.

arkodg:
@ardikabs have you tried to limit resources at a container level first, you can tweak the container resources for the EnvoyProxy https://gateway.envoyproxy.io/v0.4.0/api/config_types.html#kubernetescontainerspec ?

arkodg:
@ardikabs if im understanding correctly, the container cpu requests/limits are set to 4, but the hardware threads are much higher so `concurrency` is needed to improve thread scheduling/performance of Envoy ?

ardikabs:
@arkodg 
> have you tried to limit resources at a container level first, you can tweak the container resources for the EnvoyProxy https://gateway.envoyproxy.io/v0.4.0/api/config_types.html#kubernetescontainerspec ?

We have done this, but apparently, during the performance test, we got slightly different results, with `concurrency`, the result was slightly better in terms of utilization, given that the same RPM was reached but the utilization was slightly lower.
[...]

PR #1719, opened 2023-07-27
feat(proxy): introduce Concurrency under EnvoyProxy resource

ardikabs:
**What type of PR is this?**
Introduce an additional field under `EnvoyProxy` for additional proxy-level configuration options. The idea should be similar to Istio's [ProxyConfig](https://istio.io/latest/docs/reference/config/networking/proxy-config/).

**What this PR does / why we need it**: Our use case demands configuring the non-default concurrency for the Envoy proxy which currently is not supported by Envoy Gateway.

**Which issue(s) this PR fixes**: #1718 

Fixes #


ardikabs:
cc: @zirain

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -57,6 +57,19 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Bootstrap *string `json:"bootstrap,omitempty"`
+
+	// Config defines proxy level configuration options
+	//
+	// +optional
+	Config *ProxyConfig `json:"config,omitempty"`
```
arkodg, line 64: imo lets rm the top level `Config` field and directly add `Concurrency` here

ardikabs: @arkodg actually, I'm fine either moving Concurrency to the top level or keeping it within the Config field as both satisfy my use case, but should we also take future extensibility into account to accommodate other proxy-level options when needed, wdyt?

arkodg: this field lives inside `EnvoyProxy.Spec` so we already have a envoy proxy specific home created, other fields at the same level are `logging` & `bootstrap` which are also envoy proxy cmdline args

ardikabs: make sense, fixed.

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -57,6 +57,12 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Bootstrap *string `json:"bootstrap,omitempty"`
+
+	// Concurrency defines the number of worker threads to run. If unset, the default back to
```
arkodg, line 61:
```suggestion
	// Concurrency defines the number of worker threads to run. If unset, it defaults to
```

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -57,6 +57,12 @@ type EnvoyProxySpec struct {
...
+	// Envoy default value which the number of hardware threads on the machine.
```
arkodg, line 62:
```suggestion
	// the the number of cpuset threads on the platform.
```

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -143,10 +143,16 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egcfgv1a1.K
 		fmt.Sprintf("--log-level %s", logLevel),
 		"--cpuset-threads",
 	}
+
 	if componentLogLevel := componentLogLevelArgs(proxyLogging.Level); componentLogLevel != "" {
 		args = append(args, fmt.Sprintf("--component-log-level %s", componentLogLevel))
 	}
 
+	if infra.Config != nil &&
```
arkodg, line 151: can we keep this closer to L144 near `cpuset-threads` ?

ardikabs: okay, done!

ardikabs:
@arkodg yes, i missed the `make generate`. Fixed!

arkodg (approved):
LGTM thanks !
lets wait until v0.5 lands to merge this PR, since this PR introduces doc changes

zirain (approved):

Issue #1715, opened 2023-07-26
Seeing continous provider and gateway-api logs in stdout


arkodg:
*Description*:

Even when no new resources are applied, seeing the provider continuously reconciling and publishing resources to the gateway-api layer which is recomputing the IRs
```
023-07-26T21:58:47.477Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "", "name": "eg"}
2023-07-26T21:58:47.477Z	INFO	provider.eg	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:58:47.477Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "backend"}
2023-07-26T21:58:47.477Z	INFO	provider.backend.default	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:58:47.477Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "eg"}
2023-07-26T21:58:47.477Z	INFO	provider.eg.default	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:58:47.491Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "eg"}
2023-07-26T21:58:47.492Z	INFO	provider.eg.default	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:58:47.500Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "eg"}
2023-07-26T21:58:47.500Z	INFO	provider.eg.default	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:58:47.511Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "eg"}
2023-07-26T21:58:47.511Z	INFO	provider.eg.default	status/status.go:92	status unchanged, bypassing update	{"runner": "provider"}
2023-07-26T21:59:18.416Z	INFO	docker-desktop	kubernetes/controller.go:148	reconciling object	{"namespace": "", "name": "docker-desktop"}
2023-07-26T21:59:18.416Z	INFO	provider	kubernetes/controller.go:437	processing Gateway	{"runner": "provider", "namespace": "default", "name": "eg"}
[...]

PR #1723, opened 2023-07-28
DeepCopy resources that require status updates

arkodg:
* Was seeing constant churn between provider runner publishing resources and gateway-api runner receiving them.

* Tried to debug it by printing the o/p of `cmp.Diff` between current and previous values
```
diff --git a/internal/gatewayapi/runner/runner.go b/internal/gatewayapi/runner/runner.go
index 050394ba..50d09f6f 100644
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -8,6 +8,7 @@ package runner
 import (
        "context"

+       "github.com/google/go-cmp/cmp"
        "k8s.io/apimachinery/pkg/runtime/schema"
        "sigs.k8s.io/yaml"
@@ -49,6 +50,7 @@ func (r *Runner) Start(ctx context.Context) error {
 }

+       prev := &gatewayapi.Resources{}
        message.HandleSubscription(r.ProviderResources.GatewayAPIResources.Subscribe(ctx),
                func(update message.Update[string, *gatewayapi.Resources]) {
                        val := update.Value
@@ -56,6 +58,9 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
                        if update.Delete || val == nil {
                                return
                        }
+                       diff := cmp.Diff(prev, val)
+                       r.Logger.WithValues("output", "diff").Info(diff)
+                       prev = val.DeepCopy()

                        // Translate and publish IRs.
                        t := &gatewayapi.Translator{
```

Here's the o/p and its empty
```
2023-07-27T23:55:29.795Z	INFO	gateway-api	runner/runner.go:62		{"runner": "gateway-api", "output": "diff"}
```

* Using a DeepCopy for resources that were updating the `Status` subresource seems to have solved the issue, which implies that watchable doesnt like clients to mutate the value, even though they are meant to be a `DeepCopy`

Fixes: https://github.com/envoyproxy/gateway/issues/1715


arkodg:
cc @LukeShu

AliceProxy (approved):

zirain (approved):

internal/gatewayapi/runner/runner.go
```diff
@@ -51,6 +51,8 @@ func (r *Runner) Start(ctx context.Context) error {
 func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 	message.HandleSubscription(r.ProviderResources.GatewayAPIResources.Subscribe(ctx),
 		func(update message.Update[string, *gatewayapi.Resources]) {
+			r.Logger.Info("received an update")
```
zirain, line 54: this should be a debug log

PR #1734, opened 2023-08-01
fix: misc bug fixes for EnvoyPatchPolicy

arkodg:
* Only add EnvoyPatchPolicy to provider resources if enabled in EnvoyGateway API

* Add `omitempty` tag to `priority` field to make it optional


arkodg:
closing and fixing changes as part of https://github.com/envoyproxy/gateway/pull/1738

PR #1737, opened 2023-08-01
Add request mirroring doc

AliceProxy:
Adds user-doc for request mirroring via HTTPRoutes


docs/latest/user/http-request-mirroring.md
```diff
@@ -0,0 +1,237 @@
...
+Get the Gateway's address:
+
+```shell
+export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
+```
+
+Querying `backends.example/get` should result in a `200` response from the example Gateway and the output from the
+example app should indicate which pod handled the request. There is only one pod in the deployment for the example app
+from the quickstart, so it will be the same on all subsequent requests.
+
```
arkodg, line 124: any way to highlight that `backend-2` received the request ?

docs/latest/api/deprecation-policy.md
```diff
@@ -0,0 +1,60 @@
+# Envoy Gateway API Deprecation Policy
```
arkodg, line 1: can we split this up in a separate PR

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1738, opened 2023-08-01
e2e & misc fixes for EnvoyPatchPolicy

arkodg:
misc fixes:
* Only add EnvoyPatchPolicy to provider resources if enabled in
  EnvoyGateway API
* Add `omitempty` tag to `priority` field to make it optional

e2e
* Use LocalReplyConfig to return a custom status code `406` when there is no valid route match

### DIFF
--- a/api/v1alpha1/envoypatchpolicy_types.go
+++ b/api/v1alpha1/envoypatchpolicy_types.go
@@ -59,7 +59,7 @@ type EnvoyPatchPolicySpec struct {
 	// the priority i.e. int32.min has the highest priority and
 	// int32.max has the lowest priority.
 	// Defaults to 0.
-	Priority int32 `json:"priority"`
+	Priority int32 `json:"priority,omitempty"`
 }
 
 // EnvoyPatchType specifies the types of Envoy patching mechanisms.
--- a/examples/redis/redis.yaml
+++ b/examples/redis/redis.yaml
@@ -62,6 +62,8 @@ data:
       type: Kubernetes
     gateway:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
+    extensionApis:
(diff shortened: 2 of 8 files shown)


api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -59,7 +59,7 @@ type EnvoyPatchPolicySpec struct {
 	// the priority i.e. int32.min has the highest priority and
 	// int32.max has the lowest priority.
 	// Defaults to 0.
-	Priority int32 `json:"priority"`
+	Priority int32 `json:"priority,omitempty"`
```
zirain, line 62: this's OK to me, but I recall the api guide said should be a pointer?

examples/redis/redis.yaml
```diff
@@ -62,6 +62,8 @@ data:
       type: Kubernetes
     gateway:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
+    extensionApis:
```
zirain, line 65: this seems not related to an `redis`example, can you address some comment?

api/v1alpha1/envoypatchpolicy_types.go
arkodg: its still required internally, `omitempty` allows the user to skip setting it when creating the resource and use the default value / `0` instead

examples/redis/redis.yaml
arkodg: this is the yaml used in e2e to update the startup config, I can start renaming and breaking things in separate files, would you prefer that ?

zirain: yes, we can do that later

zirain (approved):

qicz (approved):

Issue #714, opened 2022-11-08
Document API Deprecation Policy


danehans:
A deprecation policy should be documented for EG APIs. I suggest EG follow the [Kube API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).

PR #1739, opened 2023-08-01
add deprecation policy doc

Alice-Lilith:
Adds a doc for a deprecating policy regarding features/fields/APIs

### DIFF
--- a/docs/latest/api/deprecation-policy.md
+++ b/docs/latest/api/deprecation-policy.md
@@ -1,0 +1,60 @@
+# Envoy Gateway API Deprecation Policy
+
+This document details the deprecation policy for various APIs supported by Envoy Gateway.
+
+**Note**: Envoy Gateway is currently in the pre-GA (General Availability) phase and is not yet considered fully stable. Therefore, this deprecation policy will only be applicable after the product reaches GA status. During the pre-GA phase, breaking changes may occur without following this deprecation policy.
+
+## Overview
+
+The deprecation policy is designed to provide a structured approach to managing changes to APIs and other features in Envoy Gateway to minimize disruption to users while introducing new functionalities.
+General Guidelines
+
+This policy is only applicable to the Envoy Gateway APIs ([EnvoyGateway][], [EnvoyProxy][],
+[RateLimitFilter][], [AuthenticationFilter][], and [EnvoyPatchPolicy][]). The authors of Envoy Gateway do not have control over APIs of dependencies of Envoy Gateway (namely [Envoy Proxy][] and [Gateway API][]). Any upgrades to these dependencies that would deprecate features or cause breaking changes will be clearly communicated by Envoy Gateway using the below deprecation policy and will only occur in a major version release of Envoy Gateway.
+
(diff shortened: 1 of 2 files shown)


docs/latest/api/deprecation-policy.md
```diff
@@ -0,0 +1,60 @@
...
+
+Deprecation Levels
+
+There are two levels of deprecation that can be applied to features, fields, or APIs within Envoy Gateway.
+
+- **Soft Deprecation**: Features, fields, or APIs marked with soft deprecation will continue to function as expected but are marked for future removal. During the deprecation period, users will be encouraged to migrate to alternative solutions, but the feature will remain operational.
+
+- **Hard Deprecation**: Features, fields, or APIs marked with hard deprecation will still be present but with reduced functionality or limited support. Hard deprecation indicates that the feature will be removed entirely in a future release. Users are strongly advised to migrate to alternative solutions during the deprecation period.
+
+## Deprecation Process
```
arkodg, line 37: would be good to add timelines here similar to what EnvoyProxy has
https://github.com/envoyproxy/envoy/blob/main//CONTRIBUTING.md#breaking-change-policy

arkodg, line 37: Some open questions - 
* is the policy same for v1 / stable APIs as well as v1alpha1 APIs ?
* how do we represent a alpha field within a stable API

AliceProxy: I'm not seeing a timeline there, or do you mean adding an item like:

```
All deprecations/breaking changes will be clearly listed in the [version history](https://github.com/envoyproxy/envoy/blob/main/docs/root/version_history).
```

AliceProxy: Originally that was my plan yes. I feel like having two separate policies for v1alpha1 and stable APIs, or having "alpha fields" within stable APIs is too confusing for end-users who and they don't care about our internal semantics for what version an API / field is. 

Users only care about whether or not the next upgrade has any chance of breaking them or not. I feel like whether an API or field is "alpha" can just be a way that we communicate to users how confident we are that the API will not need any breaking changes for the foreseeable future, but it is always possible that something will come up that requires us to make a breaking change to a "stable" API. The important thing is how we handle those changes, communicate those changes, and version our releases.

arkodg: something like this
```
This policy means that organizations deploying main should have some time to get ready for breaking changes at the next major API version. This is typically a window of at least 12 months or until the organization moves to the next major API version.
```

arkodg: would prefer if the recommendations in the doc are more definitive so the only decision to make in a PR / GH issue is
* can & should the field or API be deprecated
* and based on the deprecation guidelines, implementation is standardized
* as well as transition time from deprecation to removal
* and communication of the above in a release note

zirain:
related to https://github.com/envoyproxy/gateway/issues/714 ?
kindly ping @AliceProxy @arkodg

arkodg:
closing this PR due to inactivity, feel free to reopen if you're still working on it

PR #1744, opened 2023-08-02
Remove youngnick from maintainers

youngnick:
This PR removes me from Envoy Gateway maintainers. I haven't been able to contribute anything worth while for some time, so I think that this is really updating to reflect the actual state.

I've really enjoyed working with everyone on this project, and I really wish the project and everyone involved all the best.


zirain (approved):
Thanks @youngnick !

PR #1747, opened 2023-08-02
Add XDS keepalives

### LINKED ISSUE #1746: XDS connections need TCP keepalives
*Description*:
It is possible for XDS connections to half close, leaving envoy pods without config. This is well documented in other XDS implementations and is generally resolved with keepalives:
https://github.com/envoyproxy/envoy/issues/6323
https://github.com/projectcontour/contour/pull/1746

jackkleeman:
Fixes #1746

### DIFF
--- a/internal/xds/bootstrap/bootstrap.yaml.tpl
+++ b/internal/xds/bootstrap/bootstrap.yaml.tpl
@@ -125,6 +125,11 @@ static_resources:
           http2_protocol_options: {}
     name: xds_cluster
     type: STRICT_DNS
+    upstream_connection_options:
+      tcp_keepalive:
+        keepalive_probes: 3
+        keepalive_time: 30
+        keepalive_interval: 5
     transport_socket:
       name: envoy.transport_sockets.tls
       typed_config:
--- a/internal/xds/bootstrap/bootstrap.yaml.tpl
+++ b/internal/xds/bootstrap/bootstrap.yaml.tpl
@@ -125,11 +125,10 @@ static_resources:
           http2_protocol_options: {}
     name: xds_cluster
     type: STRICT_DNS
-    upstream_connection_options:
-      tcp_keepalive:
-        keepalive_probes: 3
-        keepalive_time: 30
-        keepalive_interval: 5
+    http2_protocol_options:
+      connection_keepalive:
+        interval: 30s
+        timeout: 5s
     transport_socket:
       name: envoy.transport_sockets.tls
       typed_config:
--- a/internal/cmd/egctl/testdata/translate/out/authn-single-route-single-match-to-xds.all.json
+++ b/internal/cmd/egctl/testdata/translate/out/authn-single-route-single-match-to-xds.all.json
@@ -62,6 +62,12 @@
               "clusters": [
                 {
                   "connectTimeout": "10s",
+                  "http2ProtocolOptions": {
+                    "connectionKeepalive": {
+                      "interval": "30s",
+                      "timeout": "5s"
+                    }
+                  },
                   "loadAssignment": {
                     "clusterName": "xds_cluster",
                     "endpoints": [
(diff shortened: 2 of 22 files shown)


internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -125,6 +125,11 @@ static_resources:
           http2_protocol_options: {}
     name: xds_cluster
     type: STRICT_DNS
+    upstream_connection_options:
+      tcp_keepalive:
+        keepalive_probes: 3
```
zirain, line 130: where are these number came from?

jackkleeman: I used the same ones as Contour, but we can use anything really

zirain: it's OK now, just want to clearify.

arkodg: since we control the server (xds server https://github.com/envoyproxy/gateway/blob/72bfe9bde95015b6731857f52ea30d27ee85145d/internal/xds/server/runner/runner.go#L94C12-L94C22 &  https://pkg.go.dev/net#ListenConfig) and client (this envoy cluster), we can use higher values to reduce cpu ?

arkodg:
hey @jackkleeman can you also force push with a signed commit since DCO is failing

internal/xds/bootstrap/bootstrap.yaml.tpl
jackkleeman: how does controlling the server help? the goal is for the client detect half closed situation fairly quickly. i think the server will detect this anyway whenever it sends an update. i dont feel strongly about what values we use though, if you have a preference i am all ears!

jackkleeman: i think either is fine personally

jackkleeman:
done @arkodg

internal/xds/bootstrap/bootstrap.yaml.tpl
arkodg: here are the comments in the `bootstrap.yaml` in https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#aggregated-discovery-service
```
   # It is recommended to configure either HTTP/2 or TCP keepalives in order to detect
    # connection issues, and allow Envoy to reconnect. TCP keepalive is less expensive, but
    # may be inadequate if there is a TCP proxy between Envoy and the management server.
    # HTTP/2 keepalive is slightly more expensive, but may detect issues through more types
    # of intermediate proxies.
```

so boils down to, if we believe there might ever be an proxy b/w the xds server and client (like a sidecar :) ) , it might make more sense to stick with HTTP/2 keepalives

arkodg: looks like even contour eventually moved to HTTP2 keep alives eventually https://github.com/projectcontour/contour/issues/3882

jackkleeman: ok, i can do this :)

jackkleeman: done

arkodg: awesome thanks for doing this in the client/envoy, this will also need to be done in the peer/server
```
grpc.KeepaliveParams
```

https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters

jackkleeman: This is for server detecting client going away, right? I feel its a separate discussion, I suppose the goal would be saving some resources on server side, but it won't matter for clients, right?

arkodg: for grpc keep alive the server needs to be set up with similar grpc keep alive settings, else the client might receive `GO_AWAY ENHANCE_YOUR_CALM`
https://grpc.io/docs/guides/keepalive/#background

jackkleeman: done :)

zirain (approved):
lgtm, thanks!

arkodg (approved):
🚀

jackkleeman:
I can't merge - please go ahead when happy to

PR #1757, opened 2023-08-03
rm deprecated SourceIP field

arkodg:
Relates to https://github.com/envoyproxy/gateway/pull/1308


zirain (approved):
we should do this before v1.0

Xunzhuo (approved):

Issue #1669, opened 2023-07-19
followup: support more options from proxy metric


zirain:
1. stats_matcher https://github.com/envoyproxy/gateway/pull/1653#issuecomment-1633593610
2. HistogramBucketSettings https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto.html#config-metrics-v3-histogrambucketsettings

tmsnan:
hi, i can help with that, plz assign

zirain:
go for it @tmsnan

arkodg:
moving this to v0.6.0-rc1

tmsnan:
3. stats_tags https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#envoy-v3-api-msg-config-metrics-v3-tagspecifier，provide available tag extraction rules. releated #727

zirain:
close this now, more features should be tracked one by one.

PR #1761, opened 2023-08-03
api: add stats_matcher for envoy's stats

tmsnan:
api: add stats_matcher  for envoy's stats
Relates to  #1669


zirain:
please add api first

tmsnan:
> please add api first

done

api/config/v1alpha1/metric_types.go
```diff
@@ -10,6 +10,13 @@ type ProxyMetrics struct {
 	Prometheus *PrometheusProvider `json:"prometheus,omitempty"`
 	// Sinks defines the metric sinks where metrics are sent to.
 	Sinks []MetricSink `json:"sinks,omitempty"`
+	// ProxyStatMatcher defines configuration for reporting custom Envoy stats.
+	// To redurece memory and CPU overhead from Envoy stats system.
```
arkodg, line 14:
```suggestion
	// To reduce memory and CPU overhead from Envoy stats system.
```

api/config/v1alpha1/metric_types.go
```diff
@@ -10,6 +10,13 @@ type ProxyMetrics struct {
 	Prometheus *PrometheusProvider `json:"prometheus,omitempty"`
 	// Sinks defines the metric sinks where metrics are sent to.
 	Sinks []MetricSink `json:"sinks,omitempty"`
+	// ProxyStatMatcher defines configuration for reporting custom Envoy stats.
```
arkodg, line 13: * what is the default behavior,  is it all stats ? if so can we add a note around that ?
* what happens when the user specifies a match string that doesnt match any metric ?
* should we combine `ProxyStatMatcher` & `HistogramBucketSettings` into one setting ?
    * for a specific `match` of a stat, use this bucket, if bucket unspecified, use default bucket setting 
 * `proxyStatsMatcher` seems too wordy, thoughts on `matches` or `statsSelectors`    
 * can we use the existing style of `Type` used in many places in this project and upstream for `InclusionPrefixes` etc e.g.
 ```
 matches:
 - type: "Prefix"
    value: "cluster_upstream_rq"
 ```

tmsnan: - The current default behavior is to output metrics prefixed with `cluster_manager, listener_manager, server, cluster.xds-grpc`, I think these are the most basic metrics. Default behaviors need to be discussed，releated  https://github.com/envoyproxy/envoy/issues/28179
- The current metric generation behavior is to only generate metrics matching default rules and matching custom rules. If no custom rules are matched, only the default behavioral metrics are generated.
- ProxyStatMatcher & HistogramBucketSettings can be used to reduce memory and CPU overhead from envoy stats system. Maybe combine them into `customStats` field.
-  For [StatsMatcher](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#config-metrics-v3-statsmatcher) ，we actually only used the inclusion_list (No stats are enabled except for those matching one of the supplied StringMatcher). It may be clearer to use InclusionXX here

arkodg: @AliceProxy @kflynn and I discussed this in the meeting today are here are some thoughts
* lets create a separate issue for Histogram buckets and better understand the type of buckets users wants to configure
* the current API is not very intuitive
```
metrics:
  customMetricScale:
    statsMatcher:
      inclusionMatches:
      - type:  ...
        value: ...

```

there seem be multiple layers of `match` which feels cumbersome

Instead can we stick to the earlier suggestion
```
metrics:
  matches:
  - type: ...
     value: ...
```

and lets document which metrics/stats are computed when
* when `matches` is `nil`
* when `matches` is not empty

tmsnan: create a issue for this: https://github.com/envoyproxy/gateway/issues/1786

tmsnan: > @AliceProxy @kflynn and I discussed this in the meeting today are here are some thoughts
> 
> * lets create a separate issue for Histogram buckets and better understand the type of buckets users wants to configure
> * the current API is not very intuitive
> 
> ```
> metrics:
>   customMetricScale:
>     statsMatcher:
>       inclusionMatches:
>       - type:  ...
>         value: ...
> ```
> 
> there seem be multiple layers of `match` which feels cumbersome
> 
> Instead can we stick to the earlier suggestion
> 
> ```
> metrics:
>   matches:
>   - type: ...
>      value: ...
> ```
> 
> and lets document which metrics/stats are computed when
> 
> * when `matches` is `nil`
> * when `matches` is not empty

done,using `matches` directly seems convenient to use.

- When Matches is nil(not set), gateway by default create and expose only a subset of Envoy stats with prefixes(cluster_manager,listener_manager,server,cluster.xds-grpc) expressions match on the name of stats.

- When Matches is not empty, this option is to control creation of additional Envoy stats with prefix, suffix, and regex expressions match on the name of the stats.

api/config/v1alpha1/metric_types.go
```diff
@@ -10,6 +10,14 @@ type ProxyMetrics struct {
 	Prometheus *PrometheusProvider `json:"prometheus,omitempty"`
 	// Sinks defines the metric sinks where metrics are sent to.
 	Sinks []MetricSink `json:"sinks,omitempty"`
+	// Matches defines configuration for metric matching rules
```
arkodg, line 13:
```suggestion
	// Matches defines configuration for selecting specific metrics instead of generating all metrics stats that are enabled by default. This helps reduce CPU and memory overhead in Envoy.
```

api/config/v1alpha1/metric_types.go
```diff
@@ -29,6 +37,19 @@ type MetricSink struct {
 	OpenTelemetry *OpenTelemetrySink `json:"openTelemetry,omitempty"`
 }
 
+type Match struct {
```
arkodg, line 40: // Match defines the stats match configuration.

api/config/v1alpha1/metric_types.go
```diff
@@ -29,6 +37,19 @@ type MetricSink struct {
...
+	Type  MatcherType `json:"type"`
+	Value string      `json:"value"`
+}
+
```
arkodg, line 44: // MatcherType defines the stats matcher type
// 
// +kubebuilder:validation:Enum=RegularExpression;Prefix;Suffix

api/config/v1alpha1/metric_types.go
```diff
@@ -29,6 +37,19 @@ type MetricSink struct {
...
+type MatcherType string
+
+const (
+	Prefix MatcherType = "Prefix"
+	Regex  MatcherType = "Regex"
```
arkodg, line 49: lets call this `RegularExpression` keeping it inline with upstream https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HeaderMatchType

arkodg:
thanks for incorporating the suggestions @tmsnan, had some more minor suggestions around comments and doc string, overall LGTM

api/config/v1alpha1/metric_types.go
```diff
@@ -10,6 +10,14 @@ type ProxyMetrics struct {
 	Prometheus *PrometheusProvider `json:"prometheus,omitempty"`
 	// Sinks defines the metric sinks where metrics are sent to.
 	Sinks []MetricSink `json:"sinks,omitempty"`
+	// Matches defines configuration for selecting specific metrics instead of generating all metrics stats
+	// that are enabled by default. This helps reduce CPU and memory overhead in Envoy.
+	// When Matches is nil(not set), gateway by default create and expose only a subset of
```
arkodg, line 15: lines 15-19 can be removed, imo
also because its hard to define in docs what exactly the default stats are, for e.g. we currently also output `http` stats

zirain (approved):

arkodg (approved):
LGTM thanks !

PR #1762, opened 2023-08-03
feat: add ratelimitfilter support for grpcroute

### LINKED ISSUE #1676: Support GRPCRoute + RateLimitFilter
*Description*:

Similar changes needed as https://github.com/envoyproxy/gateway/issues/1675

shawnh2:
**What type of PR is this?**

Fixes #1676

### DIFF
--- a/internal/gatewayapi/testdata/grpcroute-with-valid-ratelimitfilter.in.yaml
+++ b/internal/gatewayapi/testdata/grpcroute-with-valid-ratelimitfilter.in.yaml
@@ -1,0 +1,55 @@
+gateways:
+- apiVersion: gateway.networking.k8s.io/v1beta1
+  kind: Gateway
+  metadata:
+    namespace: envoy-gateway
+    name: gateway-1
+  spec:
+    gatewayClassName: envoy-gateway-class
+    listeners:
+    - name: http
+      protocol: HTTP
+      port: 80
+      allowedRoutes:
+        namespaces:
+          from: All
+grpcRoutes:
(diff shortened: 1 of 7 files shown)


shawnh2 (commented):
should be tested out using https://gateway.envoyproxy.io/v0.5.0/user/grpc-routing.html and https://gateway.envoyproxy.io/v0.5.0/user/rate-limit.html

internal/gatewayapi/testdata/grpcroute-with-valid-ratelimitfilter.in.yaml
```diff
@@ -0,0 +1,55 @@
+gateways:
```
shawnh2, line 1: should we also add some samples in `example/kubernetes`? and also add some doc for this?

internal/gatewayapi/testdata/grpcroute-with-valid-ratelimitfilter.out.yaml
```diff
@@ -0,0 +1,124 @@
+gateways:
```
shawnh2, line 1: should be tested out using https://gateway.envoyproxy.io/v0.5.0/user/grpc-routing.html and https://gateway.envoyproxy.io/v0.5.0/user/rate-limit.html

arkodg (approved):
LGTM thanks !
would be great to add an example in the docs as a follow up PR

zirain (approved):
LGTM, would be good to add an e2e test.

shawnh2:
sure, the `doc` and `e2e` test will be added in the follow-up PR

PR #1766, opened 2023-08-04
follow-up(doc, fix): support for grpcroute ratelimitfilter

shawnh2:
**What type of PR is this?**

fix: change the logic when dealing with authen and ratelimit filters

refactor: move `getRateLimitFilters` func into filters.go for consistency

doc: add doc for the usage of grpcroute ratelimitfilter

follow-up of #1762


docs/latest/user/rate-limit.md
```diff
@@ -13,14 +13,15 @@ i.e. if the data plane has 2 replicas of Envoy running, and the rate limit is 10
...
+can be linked to a [HTTPRoute][] or [GRPCRoute][] resource using an [ExtensionRef][] filter.
 
 ## Prerequisites
 
 ### Install Envoy Gateway
 
-* Follow the steps from the [Quickstart Guide](quickstart.md) to install Envoy Gateway and the example manifest.
-Before proceeding, you should be able to query the example backend using HTTP.
+* Follow the steps from the [Quickstart Guide](quickstart.md) to install Envoy Gateway and the HTTPRoute example manifest,
+or install the GRPCRoute example from [GRPC Routing](grpc-routing.md).
```
arkodg, line 23: maybe move this under `### GRPCRoute` because thats not the common case

internal/provider/kubernetes/routes.go
```diff
@@ -177,16 +177,16 @@ func (r *gatewayAPIReconciler) processGRPCRoutes(ctx context.Context, gatewayNam
 					authFilter, ok := resourceMap.authenFilters[key]
```
arkodg, line 177: this logic doesnt look right, shouldnt it look more like https://github.com/envoyproxy/gateway/blob/a731995bb3b56366a98eb0b256cc3bec3726e46f/internal/provider/kubernetes/routes.go#L355 which is based on `switch`.

tldr - its not an actual error if the filter is not a AuthenticationFilter and we are trying to fetch something that doesnt exist

shawnh2: fixed

docs/latest/user/rate-limit.md
shawnh2: updated

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

PR #1601, opened 2023-06-28
conformance: enable HTTPRouteRedirectPortAndScheme

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

HTTPRouteRedirectPortAndScheme conformance test is related to https://github.com/kubernetes-sigs/gateway-api/pull/1880

**Which issue(s) this PR fixes**:

Fixes #1441

PR #1784, opened 2023-08-10
Make jwt fetch more reliable

jackkleeman:
Async is useful because it will not block requests with jwt fetches, will not need to jwt fetch per thread Retrying once after a failed fetch is also useful


arkodg (approved):
LGTM thanks !

arkodg:
probably blocked on https://github.com/envoyproxy/gateway/pull/1601 which fixes some issues in the conformance tests in the newer v0.8.0-rc1 version that was recently added

zirain (approved):

Issue #1703, opened 2023-07-25
follow up: make port 19001 configurable


zirain:
EG use 0.0.0.0:19001 for health check and prometheus metric, should make this configuration.

arkodg:
+1 to making it configurable, that is something that can be done in the next release in v0.6, but shouldnt we have a dedicated port for prometheus, maybe `19002` by default ?

zirain:
> +1 to making it configurable, that is something that can be done in the next release in v0.6, but shouldnt we have a dedicated port for prometheus, maybe `19002` by default ?

why not share the same port with health check?

arkodg:
good point, if worker thread is different for different requests / starvation is unlikely, this is fine, until we add an API to make it configurable

arkodg:
yah both issues make sense @shawnh2

Issue #244, opened 2022-08-25
Add support to configure xDS server and port through EnvoyGateway API type


NomadXD:
*Description*:
With the current implementation it's not possible to configure xDS host and port. https://github.com/envoyproxy/gateway/blob/d99f32feef8ddd6ede45bd0c93b607586138d4c8/internal/xds/server/runner/runner.go#L58

We need to expose these configs through [EnvoyGateway](https://github.com/envoyproxy/gateway/blob/main/api/config/v1alpha1/envoygateway_types.go) type.

[optional *Relevant Links*:]
https://github.com/envoyproxy/gateway/issues/212

NomadXD:
@danehans I think the host and port config fields should go inside the `Gateway` struct. We can either put the fields there straightway or wrap it with another struct (something like `Listener`) and add it there. WDYT ?  

https://github.com/envoyproxy/gateway/blob/d99f32feef8ddd6ede45bd0c93b607586138d4c8/api/config/v1alpha1/envoygateway_types.go#L41

PR #1787, opened 2023-08-11
feat: add api for configure prometheus metric port

shawnh2:
**What type of PR is this?**

related to #1703 , its implementation will be done in next PR.

### DIFF
--- a/api/config/v1alpha1/metric_types.go
+++ b/api/config/v1alpha1/metric_types.go
@@ -44,4 +44,12 @@ type OpenTelemetrySink struct {
 }
 
 type PrometheusProvider struct {
+	// Port defines the port for Prometheus metrics.
+	// It shares the same port as EnvoyProxy readiness port by default.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	// +kubebuilder:default=19001
+	Port int32 `json:"port,omitempty"`
 }
--- a/charts/gateway-helm/crds/generated/config.gateway.envoyproxy.io_envoyproxies.yaml
+++ b/charts/gateway-helm/crds/generated/config.gateway.envoyproxy.io_envoyproxies.yaml
@@ -3929,6 +3929,16 @@ spec:
                       prometheus:
                         description: Prometheus defines the configuration for Admin
                           endpoint `/stats/prometheus`.
+                        properties:
+                          port:
+                            default: 19001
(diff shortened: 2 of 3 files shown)


api/config/v1alpha1/metric_types.go
```diff
@@ -44,4 +44,12 @@ type OpenTelemetrySink struct {
...
 
 type PrometheusProvider struct {
+	// Port defines the port for Prometheus metrics.
+	// It shares the same port as EnvoyProxy readiness port by default.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	// +kubebuilder:default=19001
+	Port int32 `json:"port,omitempty"`
```
zirain, line 54: TBH, I don't think this's right place to put, actually prometheus reuse the port for readiness, and I think this's the right way.

shawnh2: thanks, i thought we only need to make metric port configurable, and this port happen to be the same as readiness port by default.

if metric port reuse readiness port, and i will make readiness port configurable.

zirain: I'm looking forward an API named without readiness or metric. EG maybe extends the usage in the future.

arkodg: we may need to account for both cases - metrics port and readiness port is same or different

zirain: > we may need to account for both cases - metrics port and readiness port is same or different

I cannot image a use case for different port, it's a kind of waste.

shawnh2: just thinking out loud, if we need to take both future-extensible and all-ports-are-configurable into consideration, i suppose we could make this api looks more like:
```yaml
ports:
- type: EnvoyReadiness
  name: xxx
  port: 19001  # 19001 is the default port, the user can reassign it themselves
- type: EnvoyMetrics
  name: xxx
  port: 19001 # ditto
- type: SomePortTypeWillBeUsedInTheFuture
  name: xxx
  port: xxx
```

all we need to do is to maintain the type of port.

shawnh2: in this case, the xds server port (#244) can also be configured in this way (like `type: XdsServer`)

any thoughts? cc @arkodg @zirain

shawnh2:
closing this for now, since all the conflicts. can raise a new PR if we figure out a proper name for metric port.

Issue #1785, opened 2023-08-10
Prometheus stats not working with EG generated cluster names


psinghal20:
*Description*:
>What issue is being seen? 

Incorrect Prometheus metric names being generated by Envoy because of cluster names having the hostnames.

EG generates cluster names by appending hostnames(exposed in the HTTPRoute). For example, Cluster names in EG look like this: `default/backend/rule/0/match/0-www.example.com`.

When generating cluster metrics, Envoy uses these cluster names. The problem comes when Envoy tries to convert them to Prometheus expected format. Envoy generates the metrics in the format `cluster.default/backend/rule/0/match/0-www.example.com.upstream_cx_total: 12` which, when get converted to Prometheus metrics, are converted to `envoy_cluster_example_com_upstream_cx_total{cluster_name="default/backend/rule/0/match/0-www"}`.  Notice the incorrect metric name and cluster name in the above exposed metric in prom stats.

The incorrect translation is due to Envoy assuming that cluster names will not have `.`. Envoy breaks the initial metric name at the first period and takes the first part as the cluster name and the rest as the metric name. This leads to the cluster name being incorrectly parsed as `default/backend/rule/0/match/0-www` and the metric name as `example_com_upstream_cx_total`.

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

Instal EG and enable Prometheus by adding an Envoy Proxy config provided here: https://raw.githubusercontent.com/envoyproxy/gateway/latest/examples/kubernetes/metric/prometheus.yaml

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

Latest master

*Logs*:
>Include the access logs and the Envoy logs.

arkodg:
thanks for finding this issue, let me look into this

tmsnan:
https://github.com/envoyproxy/gateway/issues/1669#issuecomment-1653011042 。We need some constraints, set a custom regular extractor.

arkodg:
@tmsnan  the issue is we rely on IR Route Name (which appends host)
https://github.com/envoyproxy/gateway/blob/c0a619ca85c916658424d6b74870744bf14a4f1d/internal/gatewayapi/route.go#L513 which is in turn used as the xds cluster name https://github.com/envoyproxy/gateway/blob/c0a619ca85c916658424d6b74870744bf14a4f1d/internal/xds/translator/translator.go#L177

[...]

PR #1788, opened 2023-08-11
Add a Name field to a IR Route Destination

arkodg:
* Allows us to reduce and reuse the number of xds clusters created which are per xRoute per rule 
Fixes: https://github.com/envoyproxy/gateway/issues/551

* Removes the hostname from the xds cluster name since the xds cluster name is no longer tied to the ir route name
Fixes: https://github.com/envoyproxy/gateway/issues/1785

* Also append `Kind` (`httproute` / `grpcroute`) to the IR Route and IR RouteDestination names so the generated xds routes and clusters are unique across different GAPI resource types and dont collide


arkodg:
looks like I broke the Redirect conformance tests
```
2023-08-16T22:10:53.4357382Z     --- FAIL: TestGatewayAPIConformance/HTTPRouteRedirectHostAndStatus (60.04s)
2023-08-16T22:10:53.4358026Z     --- FAIL: TestGatewayAPIConformance/HTTPRouteRedirectPath (60.02s)
2023-08-16T22:10:53.4358767Z     --- FAIL: TestGatewayAPIConformance/HTTPRouteRedirectPortAndScheme (60.08s)
2023-08-16T22:10:53.4361651Z     --- FAIL: TestGatewayAPIConformance/HTTPRouteRedirectPort (60.02s)
2023-08-16T22:10:53.4362251Z     --- FAIL: TestGatewayAPIConformance/HTTPRouteRedirectScheme (60.02s)
```

looking into this rn

internal/xds/translator/cluster.go
```diff
@@ -69,28 +69,28 @@ func buildXdsCluster(routeName string, tSocket *corev3.TransportSocket, protocol
 	return cluster
```
zhaohuabing, line 69: Maybe also change the parameter `routeName` to `clusterName`?

```
func buildXdsCluster(routeName string, tSocket *corev3.TransportSocket, protocol ProtocolType, endpointType EndpointType) 
```

zhaohuabing (commented):
Please correct me if I am wrong: 

This PR creates one cluster per route rule, which removes the duplicated clusters when a route is attached to multiple gateway listeners. But if a backend service is used in multiple routes, there still are duplicated clusters for that service. Should we consider creating one cluster per Backend service?

arkodg:
> Please correct me if I am wrong:
> 
> This PR creates one cluster per route rule, eliminating the duplicated clusters when a route is attached to multiple gateway listeners. However, if a backend service is referenced in more than one route, there still are duplicated clusters created for that service. Should we consider creating one cluster per backend service?

good point, here's the thing, the xds cluster also contains other attributes such as loadbalancing and resiliency (outlier detection) that may be applied at a per xRoute per Rule level via a PolicyAttachment, so we need to create different clusters for those cases :)

zhaohuabing (approved):

AliceProxy (approved):
Looks good to me and the xDS names are clean

zirain (dismissed):

arkodg:
lets wait another day for https://github.com/envoyproxy/gateway/pull/1702 to go in, before this PR goes in, 1702 has already had to rebase a few times

zirain (approved):

Issue #1685, opened 2023-07-20
Override part of Envoy bootstrap config


zhaohuabing:
Users should be able to just override part of the Envoy bootstrap config. For example,  a user may only want to customize the accessLog or add a static cluster and doesn't care about other parts.  

The current implementation requires users to write the whole EG bootstrap configuration in the  `bootstrap` field of the `EnvoyProxy` resource,  which exposes internal details such as the static cluster for xds server to users.

zirain:
EG can use `-c` and `--config-yaml` at the same time.

zhaohuabing:
Do you mean I can just specify the part I need to override in the bootstrap field of the EnvoyProxy CRD?

zhaohuabing:
looks like it doesn't.

```
➜  ~ k -n gateway-system exec envoy-default-eg-64656661-dcbc78886-s5tbj -- ps -ef|cat 
UID          PID    PPID  C STIME TTY          TIME CMD
[...]

PR #1791, opened 2023-08-16
feat: support merging user and default bootstrap configurations

zhaohuabing:
Related issue: https://github.com/envoyproxy/gateway/issues/1685

Since this is a breaking change, please consider including the following notification, or something similar,  in the release note to remind the users to update their `EnvoyProxy` configuration. 

**Important** The `Booststrap` field of `EnvoyProxy` API has been changed from a string to a struct. Please ensure you adjust your `EnvoyProxy` configurations accordingly. Here is an example:

Before:

```
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  bootstrap: |
    admin:
      access_log: // truncated for brevity
```

After:
```
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  bootstrap:
    type: Replace
    bootstrap: |
      admin:
        access_log: // truncated for brevity
```


api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -161,6 +161,29 @@ const (
 	LogComponentRuntime LogComponent = "runtime"
 )
 
+// ProxyBootstrap defines Envoy Bootstrap configuration.
+type ProxyBootstrap struct {
+	// Type is the type of the bootstrap configuration, it should be either Merge or Replace.
```
arkodg, line 166: lets default to `Replace`

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -161,6 +161,29 @@ const (
 	LogComponentRuntime LogComponent = "runtime"
 )
 
+// ProxyBootstrap defines Envoy Bootstrap configuration.
```
arkodg, line 164: do we want to handle to list of `merges ?

arkodg:
thanks for picking this up @zhaohuabing, can you split up API and implementation into 2 PRs, which may help with faster reviews, thanks !

zhaohuabing:
> thanks for picking this up @zhaohuabing, can you split up API and implementation into 2 PRs, which may help with faster reviews, thanks !

@arkodg Since this is a breaking change, it won't pass the build if just changing the API.

arkodg:
> > thanks for picking this up @zhaohuabing, can you split up API and implementation into 2 PRs, which may help with faster reviews, thanks !
> 
> @arkodg Since this is a breaking change, it won't pass the build if just changing the API.

ah yah, sure lets stick with this for now

api/config/v1alpha1/envoyproxy_types.go
zhaohuabing:
```
        // Merge merges the provided bootstrap with the default one. The provided bootstrap can add or override a value
	// within a map, or add a new value to a list.
	// Please note that the provided bootstrap can't override a value within a list.
	BootstrapTypeMerge BootstrapType = "Merge"
```

zhaohuabing: I'm not sure whether there're any use cases that need a list of `Merge`.

zhaohuabing:
@arkodg Should I add a notice in the release note so that users won't be surprised by this breaking change? If yes, do you know where to add this note?

arkodg:
> @arkodg Should I add a notice in the release note so that users won't be surprised by this breaking change? If yes, do you know where to add this note?

added the release note label, hopefully this reminds the release manager to add this note, we can work on proactively adding release notes too, maybe in the PR description that the release manager can copy into the release notes

api/config/v1alpha1/envoyproxy_types.go
arkodg: cool, yah, most likely not needed

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -124,18 +125,29 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egcfgv1a1.K
...
 	if infra.Config != nil &&
 		infra.Config.Spec.Bootstrap != nil {
-		bootstrapConfigurations = *infra.Config.Spec.Bootstrap
-	} else {
-		var err error
-		// Use the default Bootstrap
-		bootstrapConfigurations, err = bootstrap.GetRenderedBootstrapConfig(proxyMetrics)
-		if err != nil {
-			return nil, err
+		userBootstrapConfigurations := infra.Config.Spec.Bootstrap.Value
```
arkodg, line 139: seeing this style used in many places, can we add a `util` in `internal/xds/bootstrap` to improve reuse ?

internal/utils/yaml/yaml.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"reflect"
+
+	"sigs.k8s.io/yaml"
+)
+
+// MergeYAML merges two yaml files. The second yaml file will override the first one if the same key exists.
+// This method can add or override a value within a map, or add a new value to a list.
+// Please note that this method can't override a value within a list.
+func MergeYAML(base, override string) (string, error) {
```
arkodg, line 17: can we use https://github.com/evanphx/json-patch instead, i.e. convert the YAMLs into JSON using `sigs.k8s.io/yaml` and then use `jsonpatch.MergePatch` and convert it back to YAML ?

arkodg:
can we also update the user docs https://gateway.envoyproxy.io/v0.5.0/user/customize-envoyproxy.html#customize-envoyproxy-bootstrap-config in this PR or as a follow up, thanks !

api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -56,7 +56,7 @@ type EnvoyProxySpec struct {
 	// Bootstrap configuration used. You can edit this configuration, and rerun `egctl x translate` to ensure there are no validation errors.
 	//
 	// +optional
-	Bootstrap *string `json:"bootstrap,omitempty"`
```
zirain, line 59: this's a breaking change?

internal/utils/yaml/yaml.go
zhaohuabing: `jsonpatch.MergePatch` can't be used to merge two JSON strings, and there are no other methods to merge JSON strings either.

See https://github.com/evanphx/json-patch#create-and-apply-a-merge-patch

api/config/v1alpha1/envoyproxy_types.go
zhaohuabing: Yes, it's a breaking change. We can make it non-breaking by leaving the current `Booststrap` as a string and adding a new field, but I can't think of a good name for the new field. Do you have any idea?

zirain: I'm OK with this, but must highlight in release-notes.

arkodg: yes, added the release-notes tag

zirain (dismissed):
LGTM, thanks @zhaohuabing !

arkodg (approved):
LGTM thanks !

PR #1795, opened 2023-08-17
fix: implement comparable interface for ir.Xds to skip unnecessary updates

### LINKED ISSUE #1634: watchable: skip an update if the elements in slice have not changed, but order has
*Description*:
* [Here](https://github.com/envoyproxy/gateway/blob/708287e53af5146873c91c9a4cda2a9ed8126f03/internal/gatewayapi/resource.go#L26) is Gateway API resources list that contains many fields that hold slice values
* This structure is reused as a [watchable message](https://github.com/envoyproxy/gateway/blob/708287e53af5146873c91c9a4cda2a9ed8126f03/internal/message/types.go#L20) and published from the `provider` to the `gateway-api` subscriber.
* There is no guarantee that the provider will insert elements into the slice as the same order it did the last time, since the controller-runtime reconciler/cache does not guarantee this. This might cause elements to be inserted in the list out of order
* This will cause the watchable layer to send an update to the `subscriber` since it thinks there is a change in the message. It relies on [reflect.DeepCopy](https://github.com/telepresenceio/watchable/blob/9bb86f92afa73d51a5e3e97aaf2959282d21818f/valtypes.go#L98) today for comparison.
* We can probably do better by adding a custom types that implements the `Equal` method using cmp.Diff https://pkg.go.dev/github.com/google/go-cmp/cmp#Option and a custom sort option.
* This increasing cpu needed for sorting but eliminates re translation in other layers such as `gateway-api`
 
 Relates to https://github.com/envoyproxy/gateway/issues/1503#issuecomment-1625932031 
 & https://github.com/envoyproxy/gateway/issues/1365

dboslee:
This change implements the comparable interface used by watchable to sort ir.Xds listeners when checking for differences. This reduces the number of xds updates that get pushed to envoy during re-syncs of k8s resources.

fixes #1634

### DIFF
--- a/internal/ir/xds.go
+++ b/internal/ir/xds.go
@@ -8,8 +8,10 @@ package ir
 import (
 	"errors"
 	"net"
+	"reflect"
 
 	"github.com/tetratelabs/multierror"
+	"golang.org/x/exp/slices"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


arkodg:
there's a lint error
```
Error: ST1016: methods on the same type should have the same receiver name (seen 1x "x1", 6x "x") (stylecheck)
```

probably can be recreated by running `make lint` locally

arkodg (dismissed):
LGTM thanks for adding this
would also be great to see if `watchable` is actually leveraging this `Equal` method with a test in 
https://github.com/envoyproxy/gateway/blob/main/internal/message/watchutil_test.go

dboslee:
> LGTM thanks for adding this would also be great to see if `watchable` is actually leveraging this `Equal` method with a test in https://github.com/envoyproxy/gateway/blob/main/internal/message/watchutil_test.go

Sure thing. I will add a test for this.

arkodg (approved):
awesome, thanks !

AliceProxy (approved):
Looks good, thanks for contributing 🚀

PR #1799, opened 2023-08-18
fix: ensure all updates are processed in xds ir update test

dboslee:
TestXdsIRUpdates added in #1795 is flaky. This adds a sleep to reduce the chance of failing. Open to suggestions to avoid the sleep but this pattern is also used in another test here https://github.com/envoyproxy/gateway/blob/080674a93bc42504575d7c0c83d21300abbb6e00/internal/message/watchutil_test.go#L43

### DIFF
--- a/internal/message/watchutil_test.go
+++ b/internal/message/watchutil_test.go
@@ -113,6 +113,7 @@ func TestXdsIRUpdates(t *testing.T) {
 				for _, x := range tc.xx {
 					m.Store("test", x)
 				}
+				time.Sleep(100 * time.Millisecond)
 				m.Close()
 			}()
 
--- a/internal/message/watchutil_test.go
+++ b/internal/message/watchutil_test.go
@@ -109,17 +109,26 @@ func TestXdsIRUpdates(t *testing.T) {
 			m := new(message.XdsIR)
 
 			snapshotC := m.Subscribe(ctx)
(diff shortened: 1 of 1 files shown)


arkodg:
how about using `require.Eventually` e.g. in [here](https://github.com/envoyproxy/gateway/blob/080674a93bc42504575d7c0c83d21300abbb6e00/internal/xds/translator/runner/runner_test.go#L82) ?

arkodg:
would be good to test this with `go test ./... -count=20`

dboslee:
> how about using `require.Eventually` e.g. in [here](https://github.com/envoyproxy/gateway/blob/080674a93bc42504575d7c0c83d21300abbb6e00/internal/xds/translator/runner/runner_test.go#L82) ?

@arkodg updated in https://github.com/envoyproxy/gateway/pull/1799/commits/5fcae5d4990ff97ae63da3a550ad0d173c3d854c

require.Eventually didn't exactly fit here. Checking that the number of updates is eventually equal to the desired value could work but could also give a false positive where the number of updates would have been incremented past the desired value otherwise.

This updates version of the test uses the `start` and `end` keys to indicate the first and last updates so we know when to close the map and ensure that all events have been processed.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1808, opened 2023-08-21
fix: can't show debug level logging message

zhaohuabing:
**What type of PR is this?**

Fix: EG can't show debug level logging message because the sugar logger created in the WithName function is incorrect.

### DIFF
--- a/internal/logging/log.go
+++ b/internal/logging/log.go
@@ -55,7 +55,7 @@ func (l Logger) WithName(name string) Logger {
 	return Logger{
 		Logger:        zapr.NewLogger(logger).WithName(name),
 		logging:       l.logging,
-		sugaredLogger: l.sugaredLogger,
+		sugaredLogger: logger.Sugar(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


arkodg (approved):

arkodg:
ptal @qicz

internal/logging/log.go
```diff
@@ -55,7 +55,7 @@ func (l Logger) WithName(name string) Logger {
 	return Logger{
 		Logger:        zapr.NewLogger(logger).WithName(name),
 		logging:       l.logging,
-		sugaredLogger: l.sugaredLogger,
+		sugaredLogger: logger.Sugar(),
```
qicz, line 58: inited

https://github.com/envoyproxy/gateway/blob/4285f4e269ff309e81ec24b2f80fa9b5343f6a8f/internal/logging/log.go#L25-L44

zhaohuabing: @qicz The issue with the current implementation `sugaredLogger: l.sugaredLogger` is that the logging level is incorrect. Please see the test `TestLoggerWithName` in this PR.

```
logger.Sugar().Debugf("debug message")
...
assert.Contains(t, capturedOutput, "debug message")
```

qicz: misunderstand it

qicz (approved):
LGTM

PR #1809, opened 2023-08-21
feat: support to get coredump

### LINKED ISSUE #1765: Support enableCoreDump knobs for enabling core dumps in envoy
*Description*:
>If set, when envoy crash can be captured core dumps.

[optional *Relevant Links*:]

tmsnan:
Support enableCoreDump knobs for getting core dumps in envoy 

Fixes #1765

### DIFF
--- a/api/config/v1alpha1/envoyproxy_types.go
+++ b/api/config/v1alpha1/envoyproxy_types.go
@@ -63,6 +63,13 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
+
+	// Enables core dumps for gateway.
+	//
+	// If set, gateway will have core dumps enabled.
+	//
+	// +optional
+	EnableCoreDump bool `json:"enableCoreDump,omitempty"`
 }
 
 type ProxyTelemetry struct {
--- a/api/config/v1alpha1/envoyproxy_types.go
+++ b/api/config/v1alpha1/envoyproxy_types.go
@@ -64,9 +64,10 @@ type EnvoyProxySpec struct {
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
 
-	// Enables core dumps for gateway.
+	// Enables core dumps for the managed Envoy Proxy fleet.
 	//
-	// If set, gateway will have core dumps enabled.
+	// If set, the managed Envoy Proxy fleet will generate core dumps
+	// when they crash or experience a segmentation fault.
 	//
 	// +optional
 	EnableCoreDump bool `json:"enableCoreDump,omitempty"`
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -38,6 +38,8 @@ const (
 	envoyNsEnvVar = "ENVOY_GATEWAY_NAMESPACE"
 	// envoyPodEnvVar is the name of the Envoy pod name environment variable.
 	envoyPodEnvVar = "ENVOY_POD_NAME"
+	// initContainerName is the name of the init container.
+	initContainerName = "enable-core-dump"
 )
 
 var (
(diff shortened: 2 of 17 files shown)


api/config/v1alpha1/envoyproxy_types.go
```diff
@@ -63,6 +63,13 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
+
+	// Enables core dumps for gateway.
+	//
+	// If set, gateway will have core dumps enabled.
```
arkodg, line 69:
```suggestion
	// If set, the managed Envoy Proxy fleet will generate core dumps when they crash or experience a segmentation fault.
```

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -192,6 +194,29 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egcfgv1a1.K
...
+		"sysctl -w kernel.core_pattern=/tmp/core-%e-%p-%t && ulimit -c unlimited",
+	}
+	containers := []corev1.Container{
+		{
+			Name:            initContainerName,
+			Image:           *pointer.String(image),
+			ImagePullPolicy: corev1.PullIfNotPresent,
+			Command:         []string{"/bin/sh"},
+			Args:            args,
+			SecurityContext: &corev1.SecurityContext{
```
arkodg, line 209: can we do w/o setting the `SecurityContext` ?

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -38,6 +38,8 @@ const (
 	envoyNsEnvVar = "ENVOY_GATEWAY_NAMESPACE"
 	// envoyPodEnvVar is the name of the Envoy pod name environment variable.
 	envoyPodEnvVar = "ENVOY_POD_NAME"
+	// initContainerName is the name of the init container.
+	initContainerName = "enable-core-dump"
```
arkodg, line 42:
```suggestion
	initContainerName = "configure-core-dump"
```

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -204,13 +229,18 @@ func expectedContainerVolumeMounts(deploymentSpec *egcfgv1a1.KubernetesDeploymen
 			Name:      "sds",
 			MountPath: "/sds",
 		},
+		{
```
arkodg, line 232: should be optional imo, if `r.infra.Config.Spec.EnableCoreDump`

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -204,13 +229,18 @@ func expectedContainerVolumeMounts(deploymentSpec *egcfgv1a1.KubernetesDeploymen
...
+			Name:      "coredump",
+			MountPath: "/tmp/",
```
arkodg, line 234: instead of `/tmp` , should we use `/cores` ?

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -243,6 +273,15 @@ func expectedDeploymentVolumes(name string, deploymentSpec *egcfgv1a1.Kubernetes
 				},
 			},
 		},
+		{
+			Name: "coredump",
```
arkodg, line 277: should be optional `if r.infra.Config.Spec.EnableCoreDump`

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -243,6 +273,15 @@ func expectedDeploymentVolumes(name string, deploymentSpec *egcfgv1a1.Kubernetes
...
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/data/log",
```
arkodg, line 280: `/data/log` ????

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -229,5 +229,10 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		},
 	}
 
+	if r.infra.Config.Spec.EnableCoreDump {
+		initcontainers := expectedInitContainers(*deploymentConfig.Container.Image)
```
arkodg, line 233:
```suggestion
		coreDumpInitContainer := expectedCoreDumpInitContainer(*deploymentConfig.Container.Image)
```

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -229,5 +229,10 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
...
+		deployment.Spec.Template.Spec.InitContainers = initcontainers
```
arkodg, line 234:
```suggestion
		deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, coreDumpInitContainer)
```

arkodg:
thanks for building this out @tmsnan, would be great to show an example / enhance docs of how a user can copy out the core dump file and analyze with `gdb` or something similar, and this can be orachestrated by maybe sending a `SIGABRT` to the envoy process

tmsnan:
> thanks for building this out @tmsnan, would be great to show an example / enhance docs of how a user can copy out the core dump file and analyze with `gdb` or something similar, and this can be orchestrated by maybe sending a `SIGABRT` to the envoy process

ok,I will add this part in a later commit.

internal/infrastructure/kubernetes/proxy/resource.go
tmsnan: Considering whether to mount a general directory to meet other requirements later, not just core files. So in the above design, /tmp was mounted by default.

arkodg: its due to `/tmp` that we need to add a security context right ? im trying to remove this constraint

tmsnan: Modifying the sysctl kernel parameters requires permission, maybe the securityContext permission can be set smaller, let me check it.

tmsnan:
It seems that envoy is used as an init process, causing the core file to be incorrectly generated by sending SEGV signals directly. I will continue to study this issue.

arkodg:

PR #1810, opened 2023-08-21
chore: bump k8s.io packages to v0.28.0

arkodg:
* Had to manually do this because there's a change in a doc string that affects the CRD
which was causing `make gen-check` to fail when
`dependabot` was trying to automate this in
https://github.com/envoyproxy/gateway/pull/1806


arkodg:
closed in favor of https://github.com/envoyproxy/gateway/pull/1817

PR #1813, opened 2023-08-21
fix: enqueue a single request for all resources

### LINKED ISSUE #1797: Reduce resource utilization during startup and resyncs
*Description*:

Currently with a large number of CRs the gateway controller's resource utilization is very spiky at startup and at re-syncs when resources are periodically re-reconciled. This requires high resource limits to be configured in order to handle these short infrequent bursts.

Ideally the resource usage would be lower and more consistent during large re-sync events.

This could be accomplished by compacting many reconcile calls into a single run of the main reconciliation loop rather then wasting resources re-building the state of the world for each CR. external-dns does something similar [here](https://github.com/kubernetes-sigs/external-dns/blob/master/controller/controller.go).

The minimum interval the reconciliation loop runs could be configurable with a sensible default (something like 1-5 seconds). The behavior should allow the controller to react immediately to a single event but then wait the the configured interval before running again.

Here is the behavior observed at startup where cpu and memory usage is high while the controller tries to reconcile ~600 TLSRoutes.
image

dboslee:

The current behavior of the controller is to run `Reconcile` for each resource that is updated. This gets quite expensive when you have 100s-1000s of resources.

This change allows for multiple resource updates to result in a single run of `Reconcile`.

This greatly reduces the cpu/memory consumption at startup and during periodic re-syncs of all resources.

Fixes #1797

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -147,7 +147,7 @@ func newResourceMapping() *resourceMappings {
 }
 
 func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	r.log.WithName(request.Name).Info("reconciling gateways", "namespace")
+	r.log.WithName(request.Name).Info("reconciling gateways")
 
 	var gatewayClasses gwapiv1b1.GatewayClassList
 	if err := r.client.List(ctx, &gatewayClasses); err != nil {
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -146,8 +146,11 @@ func newResourceMapping() *resourceMappings {
 	}
 }
 
-func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	r.log.WithName(request.Name).Info("reconciling gateways")
+// Reconcile handles reconciling all resources in a single call. Any resource event should enqueue the
+// same reconcile.Request containing the gateway controller name. This allows multiple resource updates to
+// be handled by a single call to Reconcile. The reconcile.Request DOES NOT map to a specific resource.
+func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	r.log.Info("reconciling gateways")
 
 	var gatewayClasses gwapiv1b1.GatewayClassList
 	if err := r.client.List(ctx, &gatewayClasses); err != nil {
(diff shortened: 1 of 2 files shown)


internal/provider/kubernetes/controller.go
```diff
@@ -1296,15 +1297,15 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
+		handler.EnqueueRequestsFromMapFunc(r.enqueueClass),
 		predicate.NewPredicateFuncs(r.httpRoutesForAuthenticationFilter)); err != nil {
 		return err
 	}
 
 	// Watch RateLimitFilter CRUDs and enqueue associated HTTPRoute objects.
 	if err := c.Watch(
 		source.Kind(mgr.GetCache(), &egv1a1.RateLimitFilter{}),
-		&handler.EnqueueRequestForObject{},
+		handler.EnqueueRequestsFromMapFunc(r.enqueueClass),
```
arkodg, line 1308: can we also do this for L1318 ?

arkodg:
this is a neat idea, thanks for implementing this @dboslee, trying to understand this better, by using `EnqueueRequestsFromMapFunc` we've managed to create a specific `Reconcile` request with the same name for all `Watch` cases, which get coalesced / rate limited by the controller (since the name is same ) so we end up getting 1 reconcile request within a specific duration ?

dboslee:
> this is a neat idea, thanks for implementing this @dboslee, trying to understand this better, by using `EnqueueRequestsFromMapFunc` we've managed to create a specific `Reconcile` request with the same name for all `Watch` cases, which get coalesced / rate limited by the controller (since the name is same ) so we end up getting 1 reconcile request within a specific duration ?

I think you've got it more or less. There isn't a specific duration here other than the rate limiting provided by the controller. Here are a few examples of describing the behavior.

At startup a single call to reconcile is executed since the controller will wait for all caches to initialized and events are coalesced.

In general, during a re-sync of many resources, any events that are processed before `Reconcile` is called are coalesced. If additional resource events arrive while the call to `Reconcile` is still running they will be coalesced in the next call to `Reconcile`.

The time in between calls to `Reconcile` is enforced by the rate limiters but its pretty quick by default.

arkodg (approved):
LGTM, thanks for introducing this design pattern to improve performance at scale !

arkodg:
> > this is a neat idea, thanks for implementing this @dboslee, trying to understand this better, by using `EnqueueRequestsFromMapFunc` we've managed to create a specific `Reconcile` request with the same name for all `Watch` cases, which get coalesced / rate limited by the controller (since the name is same ) so we end up getting 1 reconcile request within a specific duration ?
> 
> I think you've got it more or less. There isn't a specific duration here other than the rate limiting provided by the controller. Here are a few examples describing the behavior.
> 
> At startup a single call to reconcile is executed since the controller will wait for all caches to initialized and events are coalesced.
> 
> In general, during a re-sync of many resources, any events that are processed before `Reconcile` is called are coalesced. If additional resource events arrive while the call to `Reconcile` is still running they will be coalesced in the next call to `Reconcile`.
> 
> The time in between calls to `Reconcile` is enforced by the rate limiters but its pretty quick by default if not immediate when events occur.

ah makes sense, thanks for the explanation !

zirain (approved):
Thanks!

PR #1798, opened 2023-08-18
refactor: ir.TCPListener to support one to many listener to routes 

dboslee:
**What this PR does / why we need it**:
This change refactors the ir.TCPListener to be named directly after the corresponding gateway listener. The tcp/tls routes map to TCPRoutes attached to the TCPListener.

**Which issue(s) this PR fixes**:
Fixes #1635

PR #1814, opened 2023-08-22
use / instead of - in IR Route Name

arkodg:
* also use `/` instead of `-` in Authentication Filter naming 
* also replaced hostname dots to workaround https://github.com/envoyproxy/envoy/issues/4357


arkodg:
keeping in draft mode until https://github.com/envoyproxy/gateway/pull/1798 is merged

AliceProxy (approved):

zirain (approved):

PR #1826, opened 2023-08-25
feat: support statsmatch from proxy metric

tmsnan:
Description:
feat: support statsmatch from proxy metric.
Releate: #1669 #1761


internal/xds/bootstrap/bootstrap.go
```diff
@@ -113,8 +126,12 @@ func GetRenderedBootstrapConfig(proxyMetrics *egcfgv1a1.ProxyMetrics) (string, e
 	var (
 		enablePrometheus bool
 		metricSinks      []metricSink
+		StatsMatcher     StatsMatcherParameters
 	)
 
+	// Set default envoy proxy Stats
```
arkodg, line 132: why are we adding all existing ones ?
shouldnt be leave this field unset if `proxyMetrics.Matches` is nil and set it to `matches` if not nil ?

im guessing the need for this feature as a user is - "hey EG, you've enabled all these stats by default, I'd like to limit it to specific stats to reduce storage $$$"

tmsnan: You are right，when match is not empty, the default metric field should be added.
gateway enables all these stats by default.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -36,6 +36,9 @@ const (
 	envoyReadinessAddress = "0.0.0.0"
 	EnvoyReadinessPort    = 19001
 	EnvoyReadinessPath    = "/ready"
+
+	// DefaultEnvoyStatsMatcherPrefixes is the default subset of Envoy stats added when setting StatsMatcher.
```
arkodg, line 40: why is this needed ?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -138,6 +154,25 @@ func GetRenderedBootstrapConfig(proxyMetrics *egcfgv1a1.ProxyMetrics) (string, e
 				Port:    sink.OpenTelemetry.Port,
 			})
 		}
+
+		if proxyMetrics.Matches != nil {
+			enableStatConfig = true
+			// Set default envoy proxy Stats
+			StatsMatcher.Prefixs = append(StatsMatcher.Prefixs, strings.Split(DefaultEnvoyStatsMatcherPrefixes, ",")...)
```
arkodg, line 161: why are we also adding `DefaultEnvoyStatsMatcherPrefixes` here ?
imo we should ONLY we adding `proxyMetrics.Matches`

internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -8,6 +8,23 @@ admin:
     socket_address:
       address: {{ .AdminServer.Address }}
       port_value: {{ .AdminServer.Port }}
+{{- if .EnableStatConfig }}
```
arkodg, line 11: do we need an extra bool field, can we use ` {{- if not (kindIs "invalid" .StatsMatcher) }}`
relates to https://github.com/Masterminds/sprig/issues/53

internal/xds/bootstrap/bootstrap.go
tmsnan: There are two reasons：

1. [disable cluster manager stats will block the cds](https://github.com/envoyproxy/envoy/issues/28179), so cluster_manager stats is is required；
2. I think some basic stats are necessary，users may not be aware of these stats, but it is very necessary in the actual troubleshooting process.

internal/xds/bootstrap/bootstrap.yaml.tpl
tmsnan: get it

internal/xds/bootstrap/bootstrap.go
tmsnan: comment same as above

arkodg: How about enabling only the warming cluster stats for this case ? And adding a comment why in docs

tmsnan: Of course we can only expose warming cluster stats，but I think these stats(`DefaultEnvoyStatsMatcherPrefixes = "cluster_manager,listener_manager,server,cluster.xds_cluster"`) are very useful.

- [cluster_manager(https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#general):  `cluster_manager` can view the cluster size of the current proxy
- [listener_manager:](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/stats) `listener_manager` can view the cluster size of the current proxy
- cluster.xds_cluster: `cluster.xds_cluster` can view the interaction with the control plane gateway
- [server:](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/statistics)  Server related statistics are rooted at server

arkodg: the API you are implementing is an opt in API
why do we need to enable all other stats by default ? if the user wanted all stats they wouldn't use this API, since not setting a match enables all stats
the only reason to enable the warming cluster stats when a match has been defined is to make sure the critical functionality in system doesnt break

tmsnan: Maybe other default stats can be added later if more users need it later.

zirain: let's keep it empty for now.
if it's really needed, it can be done with a seperated PR.

arkodg: lets go with empty as @zirain suggested with a big disclaimer in the API that some stats are needed for basic functioning
relates to
https://github.com/envoyproxy/envoy/issues/9856
https://github.com/envoyproxy/envoy/pull/9883
https://github.com/envoyproxy/envoy/issues/14610

we can also add a match exclusion API field to opt out instead in the near future

arkodg: thoughts on updating the API to
```
metrics:
  matches:
  - include:
       type: "Prefix"
       value: ...
  - exclude:
       type: "Suffix"
       value: ....
 ```
and we can add disclaimer saying that `reducing stats helps reduce memory consumption but
eliminating some stats may after critical functionality, here are the stats that we strongly recommend not disabling .......`

Comment (tmsnan) [reply internal/xds/bootstrap/bootstrap.go:161]: I agree, adding a disclaimer is appropriate.
The idea of ​​updating the API is ok to me.

Comment (arkodg) [reply internal/xds/bootstrap/bootstrap.go:161]: Let's continue with this PR for now, and leave API as

Comment (tmsnan) [reply internal/xds/bootstrap/bootstrap.go]: done

File: internal/xds/bootstrap/bootstrap.yaml.tpl [hunk 5]
```diff
@@ -8,6 +8,23 @@ admin:
     socket_address:
       address: {{ .AdminServer.Address }}
       port_value: {{ .AdminServer.Port }}
+{{- if .StatsMatcher  }}
+stats_config:
+  stats_matcher:
+    inclusion_list:
+      patterns:
+      {{- range $a, $s := .StatsMatcher.Prefixs }}
```
zirain, line 16: can remove `$a` or use `_`?
use item instead of `s`

tmsnan: done

arkodg:
blocked on https://github.com/envoyproxy/gateway/pull/1877

api/config/v1alpha1/metric_types.go
```diff
@@ -11,7 +11,12 @@ type ProxyMetrics struct {
 	// Sinks defines the metric sinks where metrics are sent to.
 	Sinks []MetricSink `json:"sinks,omitempty"`
 	// Matches defines configuration for selecting specific metrics instead of generating all metrics stats
-	// that are enabled by default. This helps reduce CPU and memory overhead in Envoy.
+	// that are enabled by default. This helps reduce CPU and memory overhead in Envoy, but eliminating some stats
+	// may after critical functionality. Here are the stats that we strongly recommend not disabling:
```
arkodg, line 15: ❤️

arkodg (dismissed):
LGTM thanks !

AliceProxy (approved):
Needs a quick rebase but looks good

zirain (approved):

arkodg (approved):
thanks @tmsnan !

PR #1838, opened 2023-08-28
fix: prevent nil pointer panic because of nonexist attribute of Access Log

### LINKED ISSUE #1830: EnvoyGateway crashes (nil pointer) due to inappropriate AccessLog settings
*Description*:
Envoy Gateway crashes (nil pointer) when defining `EnvoyProxy` with `AccessLog` enabled using the `File` sink type. It happened when we didn't provide the `file` content under `spec.telemetry.accessLog.settings.sinks`.

*Repro steps*:
Apply below `EnvoyProxy` manifest:
```yaml
kind: EnvoyProxy
metadata:
  name: default
spec:
  telemetry:
    accessLog:
      settings:
      - format:
          json:
            host: '%REQ(:AUTHORITY)%'
          type: JSON
        sinks:
        - type: File
```

*Environment*:
Envoy Gateway v0.5.0 and latest

*Logs*:
```bash
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1d668af]

goroutine 72 [running]:
github.com/envoyproxy/gateway/internal/gatewayapi.processAccessLog(0xc001218780)
        /home/runner/work/gateway/gateway/internal/gatewayapi/listener.go:193 +0x28f
github.com/envoyproxy/gateway/internal/gatewayapi.(*Translator).ProcessListeners(0xc0008a1bf0?, {0xc0000128c8?, 0x1, 0x1}, 0x3c00900?, 0x55?, 0xc000bb1500)
        /home/runner/work/gateway/gateway/internal/gatewayapi/listener.go:52 +0x5fe
github.com/envoyproxy/gateway/internal/gatewayapi.(*Translator).Translate(0xc000083fc0?, 0xc000bb1500)
        /home/runner/work/gateway/gateway/internal/gatewayapi/translator.go:121 +0xbd
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate.func1({{0xc000080e40?, 0x16?}, 0x0?, 0xc000bb1500?})
        /home/runner/work/gateway/gateway/internal/gatewayapi/runner/runner.go:78 +0x16e
github.com/envoyproxy/gateway/internal/message.HandleSubscription[...](0x2769c40, 0xc0008a1f98?)
        /home/runner/work/gateway/gateway/internal/message/watchutil.go:36 +0x148
github.com/envoyproxy/gateway/internal/gatewayapi/runner.(*Runner).subscribeAndTranslate(0xc0000f5280, {0x2749548?, 0xc00013c5f0?})
        /home/runner/work/gateway/gateway/internal/gatewayapi/runner/runner.go:52 +0x75
        /home/runner/work/gateway/gateway/internal/gatewayapi/runner/runner.go:46 +0x272
```

ardikabs:
Fixes #1830 
Additionally, I also added a test for JSON format type Access Log.

### DIFF
--- a/api/config/v1alpha1/validation/validate_test.go
+++ b/api/config/v1alpha1/validation/validate_test.go
@@ -246,6 +246,58 @@ func TestValidateEnvoyProxy(t *testing.T) {
 			},
 			expected: false,
 		},
+		{
+			name: "should invalid when accesslog enabled using Text format, but `text` field being empty",
+			proxy: &egcfgv1a1.EnvoyProxy{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "test",
+					Name:      "test",
+				},
+				Spec: egcfgv1a1.EnvoyProxySpec{
+					Telemetry: egcfgv1a1.ProxyTelemetry{
+						AccessLog: &egcfgv1a1.ProxyAccessLog{
+							Settings: []egcfgv1a1.ProxyAccessLogSetting{
+								{
+									Format: egcfgv1a1.ProxyAccessLogFormat{
+										Type: egcfgv1a1.ProxyAccessLogFormatTypeText,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
+		{
+			name: "should invalid when accesslog enabled using File sink, but `file` field being empty",
+			proxy: &egcfgv1a1.EnvoyProxy{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "test",
+					Name:      "test",
+				},
+				Spec: egcfgv1a1.EnvoyProxySpec{
+					Telemetry: egcfgv1a1.ProxyTelemetry{
+						AccessLog: &egcfgv1a1.ProxyAccessLog{
+							Settings: []egcfgv1a1.ProxyAccessLogSetting{
+								{
+									Format: egcfgv1a1.ProxyAccessLogFormat{
+										Type: egcfgv1a1.ProxyAccessLogFormatTypeText,
+										Text: pointer.String("[%START_TIME%]"),
+									},
+									Sinks: []egcfgv1a1.ProxyAccessLogSink{
+										{
+											Type: egcfgv1a1.ProxyAccessLogSinkTypeFile,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
 	}
 
 	for i := range testCases {
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -175,6 +175,10 @@ func processAccessLog(envoyproxy *configv1a1.EnvoyProxy) *ir.AccessLog {
 		for _, sink := range accessLog.Sinks {
 			switch sink.Type {
 			case configv1a1.ProxyAccessLogSinkTypeFile:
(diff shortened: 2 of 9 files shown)


Xunzhuo:
Plz run `make testdata` and submit again, thanks.

internal/gatewayapi/listener.go
```diff
@@ -175,6 +175,10 @@ func processAccessLog(envoyproxy *configv1a1.EnvoyProxy) *ir.AccessLog {
 		for _, sink := range accessLog.Sinks {
 			switch sink.Type {
```
Xunzhuo, line 176: Do we need a default in switch to deal with the unsupported sink type? @zirain

zirain: that will block by kube-apiserver?

ardikabs: ya, I think that was already prevented from the schema validation by kube-apiserver, as `sink.Type` is actually an enum of `File` and `OpenTelemetry`.

arkodg:
imo this should be handled in https://github.com/envoyproxy/gateway/blob/3999cdd3604ad6f946a7cba795f2b034a1ae0e22/api/config/v1alpha1/validation/validate.go#L40 and we must error out early if `type: File` has a `nil` , `file` field, this will be caught in the provider layer which reconciles the resource and the error will surface in the `GatewayClass` status condition

internal/gatewayapi/listener.go
Xunzhuo: Right, thanks.

ardikabs:
> imo this should be handled in
> 
> https://github.com/envoyproxy/gateway/blob/3999cdd3604ad6f946a7cba795f2b034a1ae0e22/api/config/v1alpha1/validation/validate.go#L40
> 
> and we must error out early if `type: File` has a `nil` , `file` field, this will be caught in the provider layer which reconciles the resource and the error will surface in the `GatewayClass` status condition

but @arkodg , as per the [docs](https://gateway.envoyproxy.io/latest/api/config_types.html#fileenvoyproxyaccesslog), it said, it is expected to disable the access log for `type: File` if path was not defined or empty. So, it seems to me adding an error to the gatewayclass status would against with the docs, wdyt?

arkodg:
oh good catch @ardikabs , I think that is a docs error. we have a field called `disable` to disable access logs now, let me fix that https://gateway.envoyproxy.io/v0.5.0/api/config_types.html#proxyaccesslog

arkodg:
raised https://github.com/envoyproxy/gateway/pull/1850 to fix doc string discrepancy.

api/config/v1alpha1/validation/validate_test.go
```diff
@@ -246,12 +246,65 @@ func TestValidateEnvoyProxy(t *testing.T) {
...
+			},
+			expected: false,
+		},
 	}
 
 	for i := range testCases {
 		tc := testCases[i]
 		t.Run(tc.name, func(t *testing.T) {
 			err := ValidateEnvoyProxy(tc.proxy)
+			t.Log(err)
```
Xunzhuo, line 307: Is this required?

ardikabs: ah my bad @Xunzhuo, let me remove it.

Xunzhuo (approved):

PR #1840, opened 2023-08-28
feat: load balancer class support

### LINKED ISSUE #1793: Allow setting envoy proxy service spec.loadbalancerClass
*Description*:

For EKS clusters in AWS its recommended to use the aws-load-balancer-controller over the in-tree LoadBalancer provider.

However as of aws-load-blanacer-controller 2.5, it is no longer compatible with envoy-gateway.

The reason is an added mutating webhook that sets `spec.loadBalancerClass: service.k8s.aws/nlb` Once this has been set the envoy gateway controller can no longer converge the service object and is "stuck"

```
2023-08-16T17:18:54.170Z	ERROR	infrastructure	runner/runner.go:70	failed to create new infra	{"runner": "infrastructure", "error": "failed to create or update service ... : for Update: Service \"...\" is invalid: spec.loadBalancerClass: Invalid value: \"null\": may not change once set", "errorVerbose": "Service \"...\" is invalid: spec.loadBalancerClass: Invalid value: \"null\": may not change once set....
```

For now it seems like its possible to disable the mutating webhook in the aws-load-balancer-controller and rely on annotations to defer control to the aws-load-balancer-controller.

Though i think it would be a bit nicer to be able to configure this aspect of the generated `Service` object

*Relevant Links*:
 * [eks loadbalancer guide](https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html)
 * [eks aws loadbalancer controller guide](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html)
 * [aws-loadbalancer-controller 2.5.0 release notes](https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.5.0)
 * [#1772](https://github.com/envoyproxy/gateway/issues/1772)

wondersd:
**What type of PR is this?**

fix: fixes stuck controller when loadBalancerClass mutators exist and use of loadBalancerClass when mutators do not exist

updates EP CR to include ability to specify `Service.spec.loadBalancerClass`

Fixes #1793

Think this includes everything. My go is not well practiced so may be off. Still working on setting up for e2e test


arkodg:
hi @wondersd , thanks for raising a PR for this feature.
the PR looks good, but it looks like EG will have issues when a `service` has been created/deployed and user updates the `EnvoyProxy` spec with this new field, and EG will be unable to set this field since it cannot be mutated ?

wondersd:
@arkodg Yes that is correct. Assuming the user modifies it via a mutator on creation or otherwise deletes/recreates the service, the only presented option here would be to allow syncing what EG expects to reality.

It could be possible to update EG to destroy the `Service` to force sync this field to what EG expects. I figured this was out-of-scope and should have some more thought as it would guarantee cause downtime and would cascade into resource provisioning/de-provisioning. With just being able to alter what EG expects we can at least sync EGs exceptions with the reality.

arkodg:
sure, once this PR is merged, can you raise a follow up issue to handle the mutate case which will probably need a `delete` + `create`

arkodg (approved):
LGTM thanks !
ptal @qicz

zirain (approved):

PR #1843, opened 2023-08-29
fix a typo: disabled -> disable

rootsongjc:
**What type of PR is this?**

* docs: fix a typo: disabled -> disable, the API actually is `telemetry.accesslog.disable` not `telemetry.accesslog.disabled`. I fixed the typo in the docs and example YAML files. See the [API types](https://gateway.envoyproxy.io/latest/design/accesslog.html#proxyaccesslog-api-type).


zirain (approved):
Thanks

PR #1854, opened 2023-08-31
doc: fix incorrect item number in accesslog design doc

shawnh2:
**What type of PR is this?**

docs: typo fix

Fixes #


zirain (changes requested):
It's worked as expected.

The result is same, but it's more easy to insert a paragraph in the middle.

zirain:
actually, should change `2.` to `1.`

Xunzhuo (approved):

PR #1855, opened 2023-08-31
metrics(virtualHosts): add support for opt-in virtual host metrics stats

### LINKED ISSUE #1822: Opt in to per Hostname metrics
*Description*:

As an end user, I'd like to get more fine grained metrics per Hostname.

* EG creates a xds VirtualHost per GatewayAPI listener per Hostname
* Envoy provides [stats](https://www.envoyproxy.io/docs/envoy/v1.23.0/configuration/http/http_filters/router_filter#virtual-clusters)  per VirtualHost using the [VirtualCluster](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualcluster) field
* This can be populated if user opts in (since these stats bump memory usage) using the [EnvoyProxy](https://github.com/envoyproxy/gateway/blob/91c0888ad63631440286264f18f1bd8d2ccab4cd/api/config/v1alpha1/metric_types.go#L8) API

Another enhancement in the future could be opting in to generate per xRoute per Rule per Match metrics
* EG creates a xds Route per HTTPRoute per rule per match per hostname
* Envoy provides a [stats_prefix](https://www.envoyproxy.io/docs/envoy/v1.24.0/api-v3/config/route/v3/route_components.proto.html?highlight=we%20do%20recommend%20setting%20up%20stat%20prefix#envoy-v3-api-msg-config-route-v3-route) field in the xds Route to create per route stats

psinghal20:
**What type of PR is this?**

This PR adds support for enabling virtual host metrics stats by adding the virtual cluster configuration. 

There is an API change in the EnvoyProxy spec with the addition of `EnableVirtualHostStats` in the `Telemetry.Metrics` configuration.

I have added a unit test for translation and xdsIR generation. 

Fixes #1822


internal/xds/translator/translator.go
```diff
@@ -173,6 +176,25 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
...
 				}
+				if metrics != nil && metrics.EnableVirtualHostStats {
+					vHost.VirtualClusters = []*routev3.VirtualCluster{
+						{
+							Name: underscoredHostname,
+							Headers: []*routev3.HeaderMatcher{
+								{
+									Name: AuthorityHeaderKey,
+									HeaderMatchSpecifier: &routev3.HeaderMatcher_StringMatch{
+										StringMatch: &matcherv3.StringMatcher{
```
arkodg, line 187: probably want to make this a prefix match to handle the `example:com:8080` case

arkodg (approved):
Looks great, thanks @psinghal20 !

AliceProxy (approved):

Issue #1768, opened 2023-08-06
TLS termination breaks jwt authenticaton


dukes7gh:
*Description*:
When jwt authentication filter is applied to a HTTP route while TLS termination is used all the routes become unavailable returning 404. TLS termination seems to break GRPC routes also with the latest build when jwt filter is included.

*Repro steps*:
[tls-term-jwt.yaml.txt](https://github.com/envoyproxy/gateway/files/12270129/tls-term-jwt.yaml.txt)

* Relevant Links
https://gateway.envoyproxy.io/v0.5.0/user/secure-gateways.html
https://gateway.envoyproxy.io/v0.5.0/user/authn.html
https://github.com/envoyproxy/gateway/pull/1720

tanujd11:
@arkodg I can take a look at this.

PR #1856, opened 2023-08-31
fix tls with jwt authn usecase

tanujd11:
**What type of PR is this?**

Bug Fix

HTTPS and JWT authentication was not working in tandem because for some reason we were getting the information for JWTRequirement from HCM in DefaultFilterChain. DefaultFilterChain is not populated in case of HTTPS.

The question is irRoute already have all the information about JWT Requirement as the HCM of listener is also populated from route only which only checks if the listener has JWT Filter on the route and patch listener. Then why reverse check is done if httpfilter has the JWT requirements?  

Fixes https://github.com/envoyproxy/gateway/issues/1768

### DIFF
--- a/internal/xds/translator/authentication.go
+++ b/internal/xds/translator/authentication.go
@@ -195,7 +195,7 @@ func buildXdsUpstreamTLSSocket() (*corev3.TransportSocket, error) {
 
 // patchRouteWithJwtConfig patches the provided route with a JWT PerRouteConfig, if the
 // route doesn't contain it.
-func patchRouteWithJwtConfig(route *routev3.Route, irRoute *ir.HTTPRoute) error { //nolint:unparam
+func patchRouteWithJwtConfig(route *routev3.Route, irRoute *ir.HTTPRoute) error {
 	if route == nil {
 		return errors.New("xds route is nil")
 	}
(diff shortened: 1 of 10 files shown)


Xunzhuo (approved):

zirain:
can you add test case for this?

tanujd11:
@zirain somehow `make testdata` generated test file is failing in `go.test.coverage`. Default `make test` is working. This is due to some unicode sequence. I am not sure why is this happening with this PR only. Any ideas?

zirain:
seems related to https://github.com/envoyproxy/gateway/pull/1859

internal/xds/translator/authentication.go
```diff
@@ -195,29 +194,23 @@ func buildXdsUpstreamTLSSocket() (*corev3.TransportSocket, error) {
 }
 
 // patchRouteWithJwtConfig patches the provided route with a JWT PerRouteConfig, if the
-// route doesn't contain it. The listener is used to create the PerRouteConfig JWT
-// requirement.
-func patchRouteWithJwtConfig(route *routev3.Route, irRoute *ir.HTTPRoute, listener *listenerv3.Listener) error { //nolint:unparam
+// route doesn't contain it.
+func patchRouteWithJwtConfig(route *routev3.Route, irRoute *ir.HTTPRoute) error { //nolint:unparam
```
zirain, line 198: why `unparam`?

tanujd11: Have removed it. It was there earlier so somehow I missed that. Thanks for pointing out

arkodg (approved):
LGTM, thanks !

zirain (approved):

arkodg (approved):

Issue #1857, opened 2023-08-31
Allow to set imagePullPolicy for kube-rbac-proxy container


ElenaHenderson:
*Description*:
Allow to set imagePullPolicy for kube-rbac-proxy container in values.yaml for gateway-helm chart just like is already allowed for envoy-gateway container. I need this so I can run envoy gateway in air-gapped env.

*Repro steps*:
Go to https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
Note that there is no `imagePullPolicy` for `kube-rbac-proxy` container

PR #1858, opened 2023-08-31
fix: allow to set imagePullPolicy for kube-rbac-proxy container

ElenaHenderson:
**What type of PR is this?**
This PR adds ability to set imagePullPolicy for kube-rbac-proxy container for gateway-helm chart

I need this so I can control this setting for air-gapped env where images can not be pulled.

Fixes # https://github.com/envoyproxy/gateway/issues/1857


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -15,6 +15,7 @@ deployment:
     image:
       repository: gcr.io/kubebuilder/kube-rbac-proxy
       tag: v0.14.1
+    imagePullPolicy: Always
```
zirain, line 18: change this to `IfNotPresent`?

ElenaHenderson: Thank you! Updated.

zirain (approved):

arkodg (approved):

PR #1852, opened 2023-08-30
Added docs and fix for ref grant in service import case

tanujd11:
**What type of PR is this?**

**What this PR does / why we need it**:
Adds docs for the Multicluster services service import.
A fix is added for service import working with reference grant.
**Which issue(s) this PR fixes**:

Fixes #

PR #1869, opened 2023-09-06
:bug: Fix: follow-up of #1852

Xunzhuo:
**What type of PR is this?**

+ Add missing index
+ Set EnvoyProxy Service Type to ClusterIP, since LoadBalancer is not supported in default submariner env.
+ Upgrade RG to v1beta1.

### DIFF
--- a/examples/kubernetes/multicluster-service.yaml
+++ b/examples/kubernetes/multicluster-service.yaml
@@ -4,9 +4,6 @@ metadata:
   name: custom-proxy-config
   namespace: envoy-gateway-system
 spec:
-  logging:
-    level:
-      default: warn
   provider:
     kubernetes:
       envoyService:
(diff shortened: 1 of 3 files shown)


zirain (approved):

examples/kubernetes/multicluster-service.yaml
```diff
@@ -1,9 +1,29 @@
+apiVersion: config.gateway.envoyproxy.io/v1alpha1
+kind: EnvoyProxy
+metadata:
+  name: custom-proxy-config
+  namespace: envoy-gateway-system
+spec:
+  logging:
+    level:
+      default: warn
```
zhaohuabing, line 9: Is the logging level setting necessary?

Xunzhuo: Not necessary but does not impact anything, `default/warn` is the default value of logging.

zhaohuabing: Can we remove it since the default logging level is  already `warn`.

Xunzhuo: Done

zhaohuabing (approved):
LGTM. Thanks for cleaning up!

zirain (approved):

PR #1878, opened 2023-09-12
Bump golangci-lint to 1.54.2

zirain:
fixes: #1874


internal/gatewayapi/listener_test.go
```diff
@@ -45,7 +45,8 @@ func TestProcessTracing(t *testing.T) {
 		},
 	}
 
-	for _, c := range cases {
+	for i := range cases {
```
arkodg, line 48: prefer if kept the style same across entire code case and used the existing style of reassigning the var inside the loop i.e.
```
for _, c := range cases {
    c := c
```

zirain: updated

arkodg (approved):

Xunzhuo (commented):
Can you explain a bit about why these `x := x` lines?

zirain:
see https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

AliceProxy (approved):

PR #1886, opened 2023-09-13
chore: Add kubebuilder directive to setup default value

den3tsou:
This is a follow up PR for https://github.com/envoyproxy/gateway/pull/1661#discussion_r1323545758 to add kubebuilder directive to provide default value


zirain:
@arkodg EnvoyGateway is not a real CRD, do we really need this?

den3tsou:
This is not needed.

PR #1890, opened 2023-09-14
Delete gateway api admission server

arkodg:
No longer required, Gateway API has transitioned to
CEL based validation https://kubernetes.io/blog/2023/08/29/gateway-api-v0-8/#cel-validation


arkodg:
looks like we cannot manually omit the admission server resource since our tooling relies on the upstream install manifests
blocked on https://github.com/kubernetes-sigs/gateway-api/issues/2400

qicz:
gen-check needs fix

arkodg:
https://github.com/kubernetes-sigs/gateway-api/issues/2400 has been fixed upstream, nothing needs to be done in this PR since the omission of the webhook resources is handled in upstream itself, so we'll auto fix this the next time we bump to gateway-api v1.0

PR #1892, opened 2023-09-15
egctl: support config envoy-stats command

### LINKED ISSUE #1885: egctl support /stats
*Description*:
would be used.

[optional *Relevant Links*:]

https://www.envoyproxy.io/docs/envoy/latest/operations/admin.html#get--stats

tanujd11:
**What type of PR is this?**

Feature
Adds stats command from envoy proxies

Fixes #1885


tanujd11:
@zirain the command is working. However I can't think of any ways to unit test this whole feature. Any ideas?

internal/cmd/egctl/envoy_stats.go
```diff
@@ -0,0 +1,195 @@
...
+
+	kube "github.com/envoyproxy/gateway/internal/kubernetes"
+	"github.com/spf13/cobra"
+	"github.com/tetratelabs/multierror"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	summaryOutput          = "short"
```
zirain, line 23: istio `short` for consistency, let's use `statsd` or something like that. cc @arkodg , any suggestions?

internal/cmd/egctl/envoy_stats.go
```diff
@@ -0,0 +1,195 @@
...
+	prometheusOutput       = "prom"
+	prometheusMergedOutput = "prom-merged"
```
zirain, line 25: I don't think EG need this.

internal/cmd/egctl/config.go
```diff
@@ -180,8 +180,14 @@ func fetchRunningEnvoyPods(c kube.CLIClient, nn types.NamespacedName, labelSelec
...
 // portForwarder returns a port forwarder instance for a single Pod.
-func portForwarder(cli kube.CLIClient, nn types.NamespacedName) (kube.PortForwarder, error) {
-	fw, err := kube.NewLocalPortForwarder(cli, nn, 0, adminPort)
+func portForwarder(cli kube.CLIClient, nn types.NamespacedName, port ...int) (kube.PortForwarder, error) {
+	var err error
+	var fw kube.PortForwarder
+	if len(port) == 0 {
+		fw, err = kube.NewLocalPortForwarder(cli, nn, 0, adminPort)
+	} else {
+		fw, err = kube.NewLocalPortForwarder(cli, nn, 0, port[0])
```
zirain, line 189: if only port[0] is used, why do we need a slice here?

tanujd11: It was more or less a way to provide default port if no values is present. I can make it an `int` if it makes more sense

internal/cmd/egctl/envoy_stats.go
```diff
@@ -0,0 +1,200 @@
...
+	result, err := statsRequest(fw.Address(), path)
+	if err != nil {
+		return "", fmt.Errorf("failed to get stats on envoy for pod %s/%s: %v", podNamespace, podName, err)
+	}
+	return string(result), nil
+}
+
+func setupEnvoyClusterStatsConfig(kubeClient kubernetes.CLIClient, podName, podNamespace string, outputFormat string) (string, error) {
+	path := "clusters"
+	port := 19000
```
zirain, line 162: can you make this configurable? Technically this port should be configurable.

zirain: or else add TODO comment.

internal/cmd/egctl/envoy_stats.go
```diff
@@ -0,0 +1,200 @@
...
+
+var (
+	statsType, outputFormat string
+)
+
+func newEnvoyStatsConfigCmd() *cobra.Command {
+	var podName, podNamespace string
+
+	statsConfigCmd := &cobra.Command{
+		Use:   "envoy-stats [<type>/]<name>[.<namespace>]",
```
arkodg, line 36: can we follow the same UX as `egctl config` ?
https://github.com/envoyproxy/gateway/blob/b98c1d1cc88686765128a0228f9acbed14e7c26f/internal/cmd/egctl/config_cmd.go#L61
so this would look like
```
egctl stats envoy-proxy all <po> -n <ns>
```

arkodg: its a little more wordy, but it allows us to easier add stats for `envoy-ratelimit` and `envoy-gateway` ?

zirain: stats is not a kind of config, prefer to keep it separated.

arkodg: @zirain can you please rephrase ?
are you saying that
```egctl stats envoy-proxy``` | ```egctl stats envoy-ratelimit``` | ```egctl stats envoy-gateway```
dont make much sense ?

Comment (zirain) [reply internal/cmd/egctl/envoy_stats.go:36]: sorry, I miss understand what you said.

yes, `egctl stats` is better.

Comment (tanujd11) [reply internal/cmd/egctl/envoy_stats.go]: @arkodg @zirain the command now looks like `egctl stats envoy-proxy  <po> -n <ns>`. I don't think there is any need for all in this command. wdyt?

Comment (zirain) [reply internal/cmd/egctl/envoy_stats.go]: yes, it's OK to me.

Comment (tanujd11) [reply internal/cmd/egctl/envoy_stats.go]: It is exposed through `--`. It is same for istio. Not saying it should be the same way but the command is pretty much the same, so from a user perspective it's better in my opinion.

File: internal/cmd/egctl/envoy_stats.go [hunk 6]
```diff
@@ -0,0 +1,200 @@
...
+
+func newEnvoyStatsCmd() *cobra.Command {
+	var podName, podNamespace string
+
+	statsConfigCmd := &cobra.Command{
+		Use:   "envoy-proxy [<type>/]<name>[.<namespace>]",
+		Short: "Retrieves Envoy metrics in the specified pod",
+		Long:  `Retrieve Envoy emitted metrics for the specified pod.`,
+		Example: `  # Retrieve Envoy emitted metrics for the specified pod.
+  egctl experimental stats <pod-name[.namespace]>
```
arkodg, line 40:
```suggestion
  egctl experimental stats envoy-proxy <pod-name> -n <namespace>
```

arkodg (dismissed):
small nits in docs, else LGTM thanks !

zirain (approved):

Xunzhuo (approved):
Thanks !

PR #1893, opened 2023-09-17
feat: add field allocateLoadBalancerNodePorts to KubernetesServiceSpec

### LINKED ISSUE #1772: Allow custom (-options in the) Service object - allocateLoadBalancerNodePorts: false
*Description*:
would be used.

Allow for custom `Service` (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#servicespec-v1-core) objects to be provided in `KubernetesServiceSpec` (https://gateway.envoyproxy.io/latest/api/config_types.html#kubernetesservicespec) instead of only `ServiceType`.
The only aspect overridden by Envoy Gateway would need be the `selector` object in the `Service`.
Having the Service object exposed to configuration allows for all sorts of custom deployments .

Specifically, I am looking for a way to support `allocateLoadBalancerNodePorts: false`.
Likely others will also start to look into custom: `loadBalancerClass` or other options.

[optional *Relevant Links*:]

See: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation

den3tsou:
This PR adds field `allocateLoadBalancerNodePorts` to [KubernetesServiceSpec](https://gateway.envoyproxy.io/latest/api/config_types.html#kubernetesservicespec) for Envoy Proxy kind. The value will get propagated to the Envoy Proxy Sevice configuration, so users can choose to not use automatically assigned `NodePorts` by setting `false` for this field. Please see [ServiceSpec v1 core](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#servicespec-v1-core) for more info.

Fixes #1772


api/config/v1alpha1/validation/validate_test.go
```diff
@@ -602,3 +644,7 @@ func TestGetEnvoyProxyComponentLevelArgs(t *testing.T) {
 		})
 	}
 }
+
+func toPointer[T any](t T) *T {
```
den3tsou, line 648: Is there any good place for this kind of help function?

api/config/v1alpha1/validation/validate.go
```diff
@@ -93,6 +96,18 @@ func validateServiceType(spec *egcfgv1a1.EnvoyProxySpec) []error {
 	return errs
 }
 
+func validateServiceAllocateLoadBalancerNodePorts(spec *egcfgv1a1.EnvoyProxySpec) error {
```
arkodg, line 99: thoughts on flattening this and above `validateServiceType` into a single function called `validateService` ?

api/config/v1alpha1/validation/validate_test.go
arkodg: yah you can use https://github.com/envoyproxy/gateway/blob/main/internal/utils/ptr/ptr.go

internal/infrastructure/kubernetes/resource/resource_test.go
```diff
@@ -51,6 +52,19 @@ func TestExpectedServiceSpec(t *testing.T) {
 				ExternalTrafficPolicy: corev1.ServiceExternalTrafficPolicyTypeLocal,
 			},
 		},
+		{
+			name: "LoadBalancerWithAllocateLoadBalancerNodePorts",
+			args: args{service: &egcfgv1a1.KubernetesServiceSpec{
+				Type:                          egcfgv1a1.GetKubernetesServiceType(egcfgv1a1.ServiceTypeLoadBalancer),
+				AllocateLoadBalancerNodePorts: &allocateLoadBalancerNodePorts,
```
arkodg, line 59: can we use `ptrTo(true)` style in this file ?

api/config/v1alpha1/validation/validate.go
den3tsou: Sounds good to me. I will update it

internal/infrastructure/kubernetes/resource/resource_test.go
den3tsou: Sounds great. Updated.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #523, opened 2022-10-08
Whitenoise linter is not working well in GA


Xunzhuo:
*Description*:

See:

<img width="742" alt="image" src="https://user-images.githubusercontent.com/48784001/194692091-0d0b6e28-b39d-4ad9-8295-e1c593d54d2c.png">

https://github.com/envoyproxy/gateway/actions/runs/3208530844/jobs/5244475038

whitenoise linter is not working well in PR GA.

Xunzhuo:
BWT, do we really need to maintain this linter? How much value does it have?

arkodg:
can you repro this ? seems like its specific to the diff
we could add `set -x` to better understand the issue

Xunzhuo:
I think every PR lint GA will have this error. @arkodg

arkodg:
@Xunzhuo I dont see this issue on main, can this be closed ?

Xunzhuo:
It appears on build and test workflow in pull request, can you help check this too? @arkodg

arkodg:
> It appears on build and test workflow in pull request, can you help check this too? @arkodg

you're right, I see it, looks like its still relevant

danehans:
@shivanshu1333 are you still able to work on this issue?

Xunzhuo:
Hey @shivanshu1333, can you still work on this?

PR #1904, opened 2023-09-21
fix(ci): whitenoise lint failed when pull request

zaunist:
**What type of PR is this?**

Fix whitenoise lint error

Fixes [523](https://github.com/envoyproxy/gateway/issues/523)


Xunzhuo:
image

Thanks for working on it, I did not look deep into this issue, but it seems like still throw an error in Pull Request.

zaunist:
> image
> Thanks for working on it, I did not look deep into this issue, but it seems like still throw an error in Pull Request.

You're right, I'm not entirely sure about the exact problem either. This is my first test, and I will further study how to solve this issue carefully.

zaunist:
Please temporarily ignore the previous commit records above. I am currently investigating the cause of the issue on my own branch.

zaunist:
Hi, @Xunzhuo. I've identified the issue. In the github pull request CI environment, there is no branch information available, causing the git diff command to fail. Therefore, I added a git fetch command to retrieve the remote branch information and prefixed the branch names with "origin" to match them.

Here is my debug info for reference.

image

https://github.com/zaunist/gateway/actions/runs/6262614026/job/17005194717

arkodg:
@zaunist DCO is failing, can you squash, sign and push again ?

zaunist:
> @zaunist DCO is failing, can you squash, sign and push again ?

Done

arkodg:
DCO is still failing, you might need to squash commits and then resign and push

zirain (approved):

tools/src/whitenoise.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 TRAILING_WHITESPACE=0
 # only check changed docs in pr
-for file in $(git diff --cached --name-only --diff-filter=ACRMTU $GITHUB_BASE_REF | grep "\.md"); do
+git fetch
+for file in $(git diff --cached --name-only --diff-filter=ACRMTU origin/$GITHUB_BASE_REF| grep "\.md"); do
```
arkodg, line 5: thinking out loud, why isn't `git diff --name-only` good enough here ?

zaunist: The issue here is that the GitHub Actions environment does not have the branch information for Git, which causes an error when using `git diff` to compare the differences. To resolve this, I simply added a `git fetch` command to retrieve the branch information from the remote repository. For further debugging information, you can refer to this link: [https://github.com/zaunist/gateway/actions/runs/6262614026/job/17005194717](https://github.com/zaunist/gateway/actions/runs/6262614026/job/17005194717).

zaunist: @arkodg If we don't provide branch information, how can we compare the differences between the fork and main branches? I'm not quite understanding this point. Could you provide an example, please?

Xunzhuo (approved):
I am OK with current approach, thanks for working on it @zaunist.

Actually I think this one is not that high priority, if you would like to help, we can try to pick some important issues for you : )

zaunist:
> I am OK with current approach, thanks for working on it @zaunist.
> 
> Actually I think this one is not that high priority, if you would like to help, we can try to pick some important issues for you : )

Willing to make more contributions

PR #1905, opened 2023-09-25
fix: egctl stats invalid format

Xunzhuo:
**What type of PR is this?**

+ Fix json format output is invalid
+ Clean-up unused prom output type
+ Add alias and short for stats commands

### DIFF
--- a/internal/cmd/egctl/envoy_stats.go
+++ b/internal/cmd/egctl/envoy_stats.go
@@ -21,7 +21,6 @@ import (
 )
 
 const (
-	summaryOutput    = "statsd"
 	prometheusOutput = "prom"
 )
 
@@ -103,6 +102,12 @@ func newEnvoyStatsCmd() *cobra.Command {
 			}
 
 			switch outputFormat {
+			case jsonOutput:
+				statsBytes, err := json.Marshal(stats)
+				if err != nil {
+					return err
+				}
+				_, _ = fmt.Fprint(c.OutOrStdout(), string(statsBytes))
 			// convert the json output to yaml
 			case yamlOutput:
 				var out []byte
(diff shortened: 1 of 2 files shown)


internal/cmd/egctl/envoy_stats.go
```diff
@@ -124,7 +129,7 @@ func newEnvoyStatsCmd() *cobra.Command {
 			return nil
 		},
 	}
-	statsConfigCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", summaryOutput, "Output format: one of json|yaml|prom")
```
zirain, line 127: previous is intend, should change the description.

Xunzhuo: We support three types json/yaml/prom, the default value should be set with `prometheusOutput`, and remove the unused `summaryOutput`. What should we update here? I do not get it.

zirain: we don't need support json/yaml on stats.

internal/cmd/egctl/stats.go
```diff
@@ -11,8 +11,10 @@ import (
 
 func statsCommand() *cobra.Command {
 	c := &cobra.Command{
-		Use:  "stats",
-		Long: "Retrieve stats from envoy proxy.",
+		Use:     "stats",
+		Long:    "Retrieve Prometheus Stats from envoy proxy.",
+		Short:   "Retrieve stats from envoy proxy.",
+		Aliases: []string{"proxy"},
```
zirain, line 17: this line seems meaningless.

internal/cmd/egctl/envoy_stats.go
Xunzhuo: Not sure about that, so why adding json/yaml before cc @tanujd11 ?

internal/cmd/egctl/stats.go
Xunzhuo: Done

zirain (dismissed):

internal/cmd/egctl/stats.go
```diff
@@ -11,8 +11,9 @@ import (
 
 func statsCommand() *cobra.Command {
 	c := &cobra.Command{
-		Use:  "stats",
-		Long: "Retrieve stats from envoy proxy.",
+		Use:   "stats",
+		Long:  "Retrieve Prometheus Stats from envoy proxy.",
```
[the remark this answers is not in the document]
Xunzhuo: No matter the type is Prometheus/Json/Yaml, all of them are just Prometheus data format when dumping.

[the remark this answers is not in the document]
zhaohuabing: Sorry, I didn't get it. Not an expert of telemetry, so please correct me if I'm wrong. 

I mean that the stats data is generated by Envoy, and then the data can be consumed by Prometheus, Skywalking, or any Open Telemetry compatible backends. So why emphasize Prometheus here?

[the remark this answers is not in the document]
Xunzhuo: Thanks, done.

arkodg (approved):

Issue #1846, opened 2023-08-29
proposal: ClientTrafficPolicy


AliceProxy:
## Relates to
#1492
#1821
#1845 

## What is this?
`ClientTrafficPolicy` is a proposal for a new policy attachment resource that can be applied to `Gateway` resources. It is meant to supplement a lot of config that Gateway API currently lacks while also providing a way to specify global defaults and is the counterpart to the other proposed `BackendTrafficPolicy` resource.

## How would I use this resource?
You can attach a `ClientTrafficPolicy` to a `Gateway` to configure behaviour for how Envoy Proxy communicates with downstream clients.

## What if Gateway API implements a GEP that delivers the same functionality as some of the fields of this resource?
If Gateway improves any of their existing resources to deliver functionality that meets all of the needs of any of the below config,
then we will deprecate the field in this resource and use the Gateway API config instead. Ideally, a lot of this config can be
upstreamed into Gateway API in some form eventually since this resource only exists to solve areas where Gateway API is lacking for the needs of Envoy Gateway.

## Can you use multiple `ClientTrafficPolicy` resources at the same time?
You may use multiple `ClientTrafficPolicy` that target different `Gateways`, but you may not attach multiple `ClientTrafficPolicies` to the same `Gateway`. Merging config in this scenario would quickly become convoluted and confusing, and there are many edge-case scenarios that make supporting this undesirable.

## How will this be developed/implemented?

This issue serves as a proposal for the high-level design and fields of the API, but that does not mean that every field included below will be immediately available. My plan for the implementation is to add and implement one high-level field at a time until the whole resource is implemented. 

## API outline

```yaml
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: ClientTrafficPolicy
metadata:
  name: example-policy
spec:
  tls: # optional
    clientCerts: # optional
      forwarding: Enum(sanitize/forward-only/append-forward/sanitize-set/always-forward-only) # required
      setDetails: string # optional
      requireClientCert: bool  # optional
    version: # optional
      min: string # optional
      max: string # optional
    cipers: []Enum(list of supported cipher suites...) # optional
    ecdhCurves: []string # optional
    alpnProtocols: []Enum(h2/http-1.1) # optional
    sniOverride: string # optional
    certificateRevocationList: # optional
      secretRef: # required
        name: string # required
        namespace: # optional
  protocols: # optional
    protocolOrder: []Enum(http/https/proxy/tcp/tls/udp)
[...]

PR #1911, opened 2023-09-27
feat: ClientTrafficPolicy

arkodg:
* adds base API + CRD
* adding k8s provider reconciliation
* add skeleton in gateway-api translation layer

Relates to https://github.com/envoyproxy/gateway/issues/1846


AliceProxy (approved):

AliceProxy:
Thanks for getting the ball rolling on the API :rocket:

arkodg:
```
kubectl get clienttrafficpolicy -A -o yaml
apiVersion: v1
items:
  kind: ClientTrafficPolicy
  metadata:
    annotations:
        {"apiVersion":"gateway.envoyproxy.io/v1alpha1","kind":"ClientTrafficPolicy","metadata":{"annotations":{},"name":"my-policy","namespace":"default"},"spec":{"targetRef":{"group":"gateway.networking.k8s.io","kind":"Gateway","name":"eg","namespace":"default"}}}
    creationTimestamp: "2023-09-30T00:31:44Z"
    generation: 1
    name: my-policy
    namespace: default
    resourceVersion: "1226"
    uid: e7ee54ac-90b6-4517-a79b-952b23d5350d
  spec:
    targetRef:
      kind: Gateway
      name: eg
      namespace: default
  status:
    conditions:
    - lastTransitionTime: "2023-09-30T00:31:44Z"
      message: ClientTrafficPolicy has been accepted.
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
kind: List
metadata:
  resourceVersion: ""
```

internal/provider/kubernetes/controller.go
```diff
@@ -294,6 +294,21 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 		}
 	}
 
+	// Add all ClientTrafficPolicies
+	clientTrafficPolicies := egv1a1.ClientTrafficPolicyList{}
+	if err := r.client.List(ctx, &clientTrafficPolicies); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error listing clienttrafficpolicies: %v", err)
```
zirain, line 300: nit: lowercase is hard reading

arkodg: So you're suggesting to convert clienttrafficpolicies to ClientTrafficPolicies ?

zirain: yes

zirain (approved):

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -0,0 +1,261 @@
...
+
+			}
+
+			// Check if another policy targeting the same Gateway exists
+			if ok && (s.Len() > 0) {
+				// Maintain order here to ensure status/string does not change with same data
+				sections := s.UnsortedList()
+				sort.Strings(sections)
+				message := fmt.Sprintf("There are existing ClientTrafficPolicies that are overriding these sections %v", sections)
+
```
zhaohuabing, line 141: Should we set the `PolicyConditionAccepted` condition status to `False` if all the `Listeners` in the `Gateway` already have a `ClientTrafficPolicy` with a `SectionName` attached?

arkodg: thanks for highlighting this edge case,  I'd prefer if `PolicyConditionAccepted` is set to true, because imo the policy is accepted, it won't be able to `program` any section because its being overridden which is also highlighted

zhaohuabing: It's fine if 'programmed' condition status will be set to 'False'.

arkodg: yeah +1 to that, that will be part of the next PR when are actually translating an API field

zhaohuabing (approved):

PR #1913, opened 2023-09-28
Address 'helm lint' error <.watch.namespaces>: nil pointer

PetrMc:
**What type of PR is this?**

helm lint indentifies the following error:
```
templates/: template: gateway-helm/templates/envoy-gateway-rbac.yaml:3:27: executing "gateway-helm/templates/envoy-gateway-rbac.yaml" at <$kube.watch.namespaces>: nil pointer evaluating interface
```

two adjustments:
- values now hove the path `config.envoyGateway.provider.kubernetes.watch.namespaces`
- also adjusting logic to handle empty list in `templates.envoy-gateway-rbac.yaml`

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
@@ -1,6 +1,6 @@
 {{ $watchedNamespaces := list }}
 {{ if $kube := .Values.config.envoyGateway.provider.kubernetes }}
-{{ if and $kube.watch $kube.watch.namespaces }}
+{{ if and $kube.watch (gt (len $kube.watch.namespaces) 0) }}
 {{ $watchedNamespaces = $kube.watch.namespaces }}
 {{ end }}
 {{ end }}
--- a/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
@@ -1,9 +1,7 @@
 {{ $watchedNamespaces := list }}
 {{ if $kube := .Values.config.envoyGateway.provider.kubernetes }}
 {{ if $kube.watch }}
-{{ if or (not $kube.watch.namespaces) (eq (len $kube.watch.namespaces) 0) }}
-{{ $watchedNamespaces = list }}
-{{ else }}
+{{ if and $kube.watch.namespaces (gt (len $kube.watch.namespaces) 0) }}
 {{ $watchedNamespaces = $kube.watch.namespaces }}
 {{ end }}
 {{ end }}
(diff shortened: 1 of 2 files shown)


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -41,10 +41,14 @@ config:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
     provider:
       type: Kubernetes
+      kubernetes:
```
arkodg, line 44: This doesn't look right,  watch is nil by default

PetrMc: thanks @arkodg - tested with original values.yaml and helm lint fails if the empty "config.envoyGateway.provider.kubernetes.watch.namespaces" is not defined.

PetrMc: addressed nil in the template code. 

After checking $kube.watch, it uses an or function to evaluate two conditions:
a. Whether $kube.watch.namespaces is not set (not $kube.watch.namespaces) — this handles the nil case.
b. Whether the length of $kube.watch.namespaces is 0.

If either of those conditions is true, it sets $watchedNamespaces to an empty list; otherwise, it sets $watchedNamespaces to $kube.watch.namespaces.

This way, we handle both nil and zero-length lists appropriately.

charts/gateway-helm/templates/envoy-gateway-rbac.yaml
```diff
@@ -1,9 +1,14 @@
 {{ $watchedNamespaces := list }}
 {{ if $kube := .Values.config.envoyGateway.provider.kubernetes }}
-{{ if and $kube.watch $kube.watch.namespaces }}
+{{ if $kube.watch }}
+{{ if or (not $kube.watch.namespaces) (eq (len $kube.watch.namespaces) 0) }}
+{{ $watchedNamespaces = list }}
```
arkodg, line 5: this is also defined on line 1

arkodg:
@PetrMc can you also include the `helm` lint cmd as a make target in https://github.com/envoyproxy/gateway/blob/main/tools/make/lint.mk and add it as a dep to `lint` to ensure this doesnt regress again, thanks !

charts/gateway-helm/templates/envoy-gateway-rbac.yaml
PetrMc: is the new commit satisfying?

arkodg: looks good to me :) 
@Xunzhuo @zirain can you help review

arkodg:
`docker pull kindest/node:v1.28.0` is failing, I think docker hub is down

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM

Issue #809, opened 2022-12-14
How to Organize API Types


danehans:
Currently, EG API types reside in 2 locations:
- Envoy Gateway and Envoy Proxy configuration: `api/config/v1alpha1`.
- Extended functionality: `api/v1alpha1`.

This makes imports a bit confusing:
```
import (
	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
	egcfgv1a1 "github.com/envoyproxy/gateway/api/config/v1alpha1"
)
```

It's unclear what API types are being imported from `github.com/envoyproxy/gateway/api/v1alpha1`.

danehans:
From the 5/13/22 EG [community call](https://docs.google.com/document/d/1leqwsHX8N-XxNEyTflYjRur462ukFxd19Rnk3Uzy55I/edit?usp=sharing), @youngnick proposes:
> I recommend keeping all the types in the version directory (eg v1alpha1), with validation, helper, or utility code in subdirectories.

danehans:
My recommendation slightly differs from the above. I suggest having separate config and extension API groups. Helper functions, validation, etc. will live within each group for the associated types. This allows versioning the extension types separately from the config types. For example, the EnvoyProxy type can graduate faster than extension API types since they may be more widely used. Additionally, it simplifies transitioning the extension APIs into a separate repo.

Thoughts @youngnick

zirain:
+1 to `egv1 "github.com/envoyproxy/gateway/api/v1alpha1"` or `egv1alpha1 "github.com/envoyproxy/gateway/api/v1alpha1"`

youngnick:
@danehans, I think that's fine, although I suggest standardizing the import names to make reading the code easier.

danehans:
@arkodg @AliceProxy @Xunzhuo @skriss please comment if you have issue with the approach in https://github.com/envoyproxy/gateway/issues/809#issuecomment-1352277995.

arkodg:
@danehans i think theres a typo in the issue description -  `Extended functionality: api/v1alpha1`
as @zirain suggested I prefer all EG APIs to live under  `api/v1alpha1` under the same group for now so there are lesser groups to revision

danehans:
> i think theres a typo in the issue description

@arkodg thanks for catching this. I updated the description.

> I prefer all EG APIs to live under api/v1alpha1 under the same group for now so there are lesser groups to revision

[...]

PR #1921, opened 2023-10-03
move EnvoyProxy API to gateway.envoyproxy.io

arkodg:
* this keeps all CRDs/APIs under the gateway.envoyproxy.io group

* this is a breaking change, that needs to be highlighted in the release notes. Undesirable but acceptable since these APIs are experimental (alpha) and are not yet stable

Fixes: https://github.com/envoyproxy/gateway/issues/809


zirain:
will this be a break change for existing end users? especially upgrade?

arkodg:
> will this be a break change for existing end users? especially upgrade?

yeah this will be a breaking change, we had discussed this in the community meeting and looks like we reached some [consensus](https://github.com/envoyproxy/gateway/issues/809#issuecomment-1631570729) on this

qicz (approved):

zirain (approved):

PR #1926, opened 2023-10-03
Allow policy target namespace to be optional

arkodg:
* Defaults to the policy namespace as recommend by upstream
https://github.com/kubernetes-sigs/gateway-api/blob/d38e6c81ba3d3f137012ceeec8262ab40ad0cd81/apis/v1alpha2/policy_types.go#L47


zhaohuabing (approved):
LGTM. Thanks for cleaning up.

zirain (approved):

PR #1927, opened 2023-10-05
fix: check error first before accessing the result during translation

### LINKED ISSUE #1917: Panic because of invalid result during xDS translation
*Description*:
When integrating Envoy Gateway and Envoy Gateway Extension service, I observed that during the xDS translation result, when an error occurred, it would be panicking because there was some operation to the result variable before the error check.
https://github.com/envoyproxy/gateway/blob/7d9e16e4997e7118d70ea910a6ef9d0732426878/internal/xds/translator/runner/runner.go#L80-L98

*Repro steps*:

In my scenario, we can simply produce an error from one of the hooks registered through the Envoy Gateway Extension service. Then you will see Envoy Gateway no longer up and running because of panicking.

*Environment*:
Tested on Envoy Gateway v0.5.0, but the code also persists in the main branch.

*Logs*:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1792a64]

goroutine 97 [running]:
github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate.func1({{0x400086d170?, 0x1733654?}, 0x78?, 0x4000138f50?})
        /home/runner/work/gateway/gateway/internal/xds/translator/runner/runner.go:79 +0x1a4
github.com/envoyproxy/gateway/internal/message.HandleSubscription[...](0x2112300, 0x4000d4df88?)
        /home/runner/work/gateway/gateway/internal/message/watchutil.go:36 +0x104
github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate(0x40003ba180, {0x20f19b8?, 0x4000528410?})
        /home/runner/work/gateway/gateway/internal/xds/translator/runner/runner.go:52 +0x70
created by github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).Start
        /home/runner/work/gateway/gateway/internal/xds/translator/runner/runner.go:45 +0x1f8
```

*Expectation*:

The error during xDS translation should be returned and logged instead of panicking.

ardikabs:
Fixes #1917

### DIFF
--- a/internal/xds/translator/runner/runner.go
+++ b/internal/xds/translator/runner/runner.go
@@ -78,6 +78,10 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				}
 
 				result, err := t.Translate(val)
+				if err != nil {
+					r.Logger.Error(err, "failed to translate xds ir")
+					return
+				}
 
 				// Publish EnvoyPatchPolicyStatus
 				for _, e := range result.EnvoyPatchPolicyStatuses {
@@ -90,12 +94,8 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				// Discard the EnvoyPatchPolicyStatuses to reduce memory footprint
 				result.EnvoyPatchPolicyStatuses = nil
 
-				if err != nil {
-					r.Logger.Error(err, "failed to translate xds ir")
-				} else {
-					// Publish
-					r.Xds.Store(key, result)
-				}
+				// Publish
+				r.Xds.Store(key, result)
 			}
 		},
 	)
--- a/internal/xds/translator/runner/runner.go
+++ b/internal/xds/translator/runner/runner.go
@@ -80,22 +80,21 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				result, err := t.Translate(val)
 				if err != nil {
 					r.Logger.Error(err, "failed to translate xds ir")
-					return
-				}
-
-				// Publish EnvoyPatchPolicyStatus
-				for _, e := range result.EnvoyPatchPolicyStatuses {
-					key := ktypes.NamespacedName{
-						Name:      e.Name,
-						Namespace: e.Namespace,
+				} else {
+					// Publish EnvoyPatchPolicyStatus
+					for _, e := range result.EnvoyPatchPolicyStatuses {
+						key := ktypes.NamespacedName{
+							Name:      e.Name,
+							Namespace: e.Namespace,
+						}
+						r.EnvoyPatchPolicyStatuses.Store(key, e.Status)
 					}
-					r.EnvoyPatchPolicyStatuses.Store(key, e.Status)
-				}
-				// Discard the EnvoyPatchPolicyStatuses to reduce memory footprint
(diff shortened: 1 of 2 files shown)


ardikabs:
it seems that the e2e-test is a bit flaky. It succeeded on v1.28, but failed on v1.26.6, am I missing something here?
/cc @arkodg

internal/xds/translator/runner/runner.go
```diff
@@ -78,6 +78,10 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				}
 
 				result, err := t.Translate(val)
+				if err != nil {
```
arkodg, line 81: you can't `return` here, this is an endless loop, one bad config should not cause the entire translator to exit

ardikabs: sure, let me change it to move under the `else` directive.

ardikabs: out of curiosity, @arkodg if we are not allowed to `return` this, then why this is allowed here? 
https://github.com/envoyproxy/gateway/blob/2ba50ea89e7a7342f3c44962715201b2d382b862/internal/provider/kubernetes/controller.go#L1077-L1093
Because I thought that process was just a callable function, that would be executed on every incoming data.

arkodg: that is returning from the function, not from the go routine

internal/xds/translator/runner/runner.go
```diff
@@ -78,21 +78,20 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
-					}
-					r.EnvoyPatchPolicyStatuses.Store(key, e.Status)
-				}
-				// Discard the EnvoyPatchPolicyStatuses to reduce memory footprint
-				result.EnvoyPatchPolicyStatuses = nil
-
 				if err != nil {
 					r.Logger.Error(err, "failed to translate xds ir")
 				} else {
+					// Publish EnvoyPatchPolicyStatus
```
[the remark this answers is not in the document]
ardikabs: the issue is when an error occurs, the `result` variable is nil, then empty values wouldn't be a problem, but the access to the value is the problem.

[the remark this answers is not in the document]
ardikabs: as we can see here, 
https://github.com/envoyproxy/gateway/blob/2ba50ea89e7a7342f3c44962715201b2d382b862/internal/xds/translator/translator.go#L60-L97

internal/xds/translator/runner/runner.go
ardikabs: sure, just a side note, I forgot to mention as well, that the patch that I work on now is actually running in my org staging environment, and it works well.

arkodg (approved):
LGTM thanks !

internal/xds/translator/runner/runner.go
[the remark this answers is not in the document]
arkodg: thanks for root causing the panic

Xunzhuo (approved):
LGTM

PR #1935, opened 2023-10-08
opt: codes

Xunzhuo:
**What type of PR is this?**

`EnvoyPatchPolicyStatuses` is passing from component to component on top level, this is not followed other policy logics and make it look weird, and if more policies add in the future, the style is not synced and not easy to maintain.

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -124,16 +124,14 @@ func setupRunners(cfg *config.Server) error {
 	}
 
 	pResources := new(message.ProviderResources)
-	ePatchPolicyStatuses := new(message.EnvoyPatchPolicyStatuses)
 	// Start the Provider Service
 	// It fetches the resources from the configured provider type
 	// and publishes it
 	// It also subscribes to status resources and once it receives
 	// a status resource back, it writes it out.
 	providerRunner := providerrunner.New(&providerrunner.Config{
-		Server:                   *cfg,
-		ProviderResources:        pResources,
-		EnvoyPatchPolicyStatuses: ePatchPolicyStatuses,
+		Server:            *cfg,
+		ProviderResources: pResources,
 	})
 	if err := providerRunner.Start(ctx); err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


qicz (approved):

zhaohuabing (commented):
1. I'm not certain that we should put `EnvoyPatchPolicyStatuses` into the `ProviderResource`: `ProviderResource` is used in both the `providerrunner` config and the `gatewayapirunner` config, but  `EnvoyPatchPolicyStatuses` is only used in the `providerrunner`.  would like to hear @arkodg 's opinion on this.

2. It may be a personal preference - I don't like to add a middle structure to group all the statuses: these structures don't add additional functionalities, but make the file lengthy and make reading code a bit more difficult.

Xunzhuo:
> 1. I'm not certain that we should put `EnvoyPatchPolicyStatuses` into the `ProviderResource`: `ProviderResource` is used in both the `providerrunner` config and the `gatewayapirunner` config, but  `EnvoyPatchPolicyStatuses` is only used in the `providerrunner`.  would like to hear @arkodg 's opinion on this.

`EnvoyPatchPolicyStatuses` is used for both provider and xds-translator. Regarding `ClientTrafficPolicyStatuses` we introduced recently, it is also under `ProviderResources` instead of adding it directly on top of components.

We should be careful about adding new params/message on top of component runners, if not, there will be more and more new func params in component runners, and we need to individually new/close them on top of runners.

arkodg:
> * I'm not certain that we should put `EnvoyPatchPolicyStatuses` into the `ProviderResource`: `ProviderResource` is used in both the `providerrunner` config and the `gatewayapirunner` config, but  `EnvoyPatchPolicyStatuses` is only used in the `providerrunner`.  would like to hear @arkodg 's opinion on this.
> * It may be a personal preference - I don't like to add a middle structure to group all the statuses: these structures don't introduce additional functionalities, but add an unnecessary redirection and make reading code a bit more difficult.

the code does look cleaner now, but the reason `EnvoyPatchPolicyStatus` was made separate from `ProviderResources` is that it was published from the `xds-translator`, no other resource is published for there, rest are published from `gateway-api`. With this change, we are now passing the entire `ProviderResources` handle to the `xds-translator` which doesnt need it

Xunzhuo:
> > * I'm not certain that we should put `EnvoyPatchPolicyStatuses` into the `ProviderResource`: `ProviderResource` is used in both the `providerrunner` config and the `gatewayapirunner` config, but  `EnvoyPatchPolicyStatuses` is only used in the `providerrunner`.  would like to hear @arkodg 's opinion on this.
> > * It may be a personal preference - I don't like to add a middle structure to group all the statuses: these structures don't introduce additional functionalities, but add an unnecessary redirection and make reading code a bit more difficult.
> 
> the code does look cleaner now, but the reason `EnvoyPatchPolicyStatus` was made separate from `ProviderResources` is that it was published from the `xds-translator`, no other resource is published for there, rest are subscribed & published from `gateway-api`. With this change, we are now passing the entire `ProviderResources` handle to the `xds-translator` which doesnt need it, and we are assuming that `xds-translator` doesnt use that context rather than enforce it

There may be other `resource` published from xds translator, so it would be better to keep the resource under `resources`.

PR #1936, opened 2023-10-08
ci: ignore patch error 

Xunzhuo:
**What type of PR is this?**

Patch should not bring CI failed


qicz (approved):

arkodg:
by going ahead with this, we are making this the norm rather than the exception

PR #1938, opened 2023-10-09
infra: add support for merging gateways on to a single EnvoyProxy

cnvergence:
**What type of PR is this?**

- Setting the mergeGateways field in the EnvoyProxy resource linked to GatewayClass will result in merging
all Gateway Listeners
- Ensure that the tuple of port, protocol, and hostname is unique across all Listeners

Fixes:
- https://github.com/envoyproxy/gateway/issues/1230


internal/gatewayapi/helpers.go
```diff
@@ -400,6 +406,10 @@ func irStringKey(gatewayNs, gatewayName string) string {
 	return fmt.Sprintf("%s/%s", gatewayNs, gatewayName)
 }
 
+func irMergedInfraPortName(listener *ListenerContext) string {
+	return fmt.Sprintf("%s-%s-%s", listener.Name, listener.gateway.Namespace, listener.gateway.Name)
```
arkodg, line 410: maybe reuse `irHTTPListenerName` here ?

internal/gatewayapi/listener.go
```diff
@@ -27,31 +27,29 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
...
+		var foundPorts []*protocolPort
+
 		irKey := irStringKey(gateway.Gateway.Namespace, gateway.Gateway.Name)
-		gwXdsIR := &ir.Xds{}
-		gwInfraIR := ir.NewInfra()
-		gwInfraIR.Proxy.Name = irKey
-		gwInfraIR.Proxy.GetProxyMetadata().Labels = GatewayOwnerLabels(gateway.Namespace, gateway.Name)
-		if resources.EnvoyProxy != nil {
-			gwInfraIR.Proxy.Config = resources.EnvoyProxy
+		if t.MergeGateways {
```
arkodg, line 42: seeing this pattern used in many places, can we dedup it with `t.getIRKey(gwc, gateway)` ?

internal/gatewayapi/listener.go
```diff
@@ -27,31 +27,29 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
...
 	// to the Xds IR.
 	for _, gateway := range gateways {
-		// init IR per gateway
+		// Infra IR proxy ports must be unique.
+		var foundPorts []*protocolPort
+
 		irKey := irStringKey(gateway.Gateway.Namespace, gateway.Gateway.Name)
-		gwXdsIR := &ir.Xds{}
-		gwInfraIR := ir.NewInfra()
-		gwInfraIR.Proxy.Name = irKey
```
arkodg, line 39: these lines have been deleted but not readied below

internal/gatewayapi/testdata/multiple-gateways.in.yaml
```diff
@@ -0,0 +1,29 @@
+gateways:
```
arkodg, line 1: is this test case needed ?

internal/gatewayapi/testdata/merge-valid-multiple-gateways.in.yaml
```diff
@@ -0,0 +1,37 @@
...
+gateways:
+  - apiVersion: gateway.networking.k8s.io/v1beta1
+    kind: Gateway
+    metadata:
+      name: gateway-1
+      namespace: envoy-gateway
+    spec:
+      gatewayClassName: envoy-gateway-class
+      listeners:
+        - name: http
```
arkodg, line 18: can we add some more gateways or listeners in here to test where only one of the three tuple values (port, proto, hostname) is different, yet those are valid ?

internal/gatewayapi/helpers.go
```diff
@@ -400,6 +406,10 @@ func irStringKey(gatewayNs, gatewayName string) string {
...
+}
+
```
arkodg, line 412: will `irRouteName` and `irRouteDestinationName` collide since now there is nothing unique if the routes attach to multiple gateways resources which now will manifests into a single xds config
or can we reuse ?

internal/gatewayapi/validate.go
```diff
@@ -532,6 +532,29 @@ type portListeners struct {
 	hostnames map[string]int
 }
 
+func (t *Translator) validateConflictedMergedListeners(gateways []*GatewayContext) {
+	listenerSets := sets.Set[string]{}
+	for _, gateway := range gateways {
+		for _, listener := range gateway.listeners {
+			var hostname string
```
arkodg, line 539: safer to init to `*`

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -56,6 +57,17 @@ func ExpectedResourceHashedName(name string) string {
 	return fmt.Sprintf("%s-%s", config.EnvoyPrefix, hashedName)
 }
 
+// ExpectedContainerPortHashedName returns expected service name with max length of 15 characters.
+func ExpectedContainerPortHashedName(name string) string {
+	if len(name) > 15 {
```
arkodg, line 62: can we move this code into `providerutils` ?

internal/gatewayapi/helpers.go
```diff
@@ -400,6 +406,10 @@ func irStringKey(gatewayNs, gatewayName string) string {
 	return fmt.Sprintf("%s/%s", gatewayNs, gatewayName)
 }
 
+func irMergedInfraPortName(listener *ListenerContext) string {
```
arkodg, line 409: lets flip the order of concatenation, similar to `irHTTPListenerName` ? and we cannot use `-`, we need to use `/` , an invalid char as a delimiter

internal/gatewayapi/helpers.go
arkodg: and you then want to do something similar to https://github.com/envoyproxy/gateway/blob/2ecc54da663dc10c9257a61843cd14ba1bb77dcc/internal/provider/utils/utils.go#L33C1-L33C1 in the infra layer side

internal/provider/kubernetes/controller.go
```diff
@@ -1317,7 +1322,10 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	}
 
 	// Watch Gateway CRUDs and reconcile affected GatewayClass.
-	gPredicates := []predicate.Predicate{predicate.NewPredicateFuncs(r.validateGatewayForReconcile)}
+	gPredicates := []predicate.Predicate{
+		predicate.NewPredicateFuncs(r.validateGatewayForReconcile),
+		predicate.GenerationChangedPredicate{},
```
arkodg, line 1327: this seems like a very useful thing to do, should we add this predicate to all resources as a separate PR ?
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#GenerationChangedPredicate

internal/provider/kubernetes/helpers.go
```diff
@@ -197,12 +197,20 @@ func refsSecret(ref *gwapiv1b1.SecretObjectReference) bool {
 		(ref.Kind == nil || *ref.Kind == gatewayapi.KindSecret)
 }
 
-func infraServiceName(gateway *gwapiv1b1.Gateway) string {
+func infraServiceName(gateway *gwapiv1b1.Gateway, merged bool) string {
+	if merged {
```
arkodg, line 201: can we simplify this
```
var ns, name string
if  merged {
  ns, name = "", string(gateway.Spec.GatewayClassName)
} else {
  ns, name = gateway.Namespace , gateway.Name
}
 ......
```

internal/provider/kubernetes/predicates.go
```diff
@@ -168,7 +168,20 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
...
+	if ok {
+		gtw := r.findGateways(ctx, gclass)
+		if len(gtw.Items) != 0 {
+			for _, gw := range gtw.Items {
+				gw := gw
+				r.statusUpdateForGateway(ctx, &gw)
+			}
+		}
 
+		nsName := utils.NamespacedName(svc)
```
arkodg, line 182: curious why this (backend svc) is needed for this case (where the `svc` here is the generated envoy svc)?

arkodg: we should be returning `false` here

internal/provider/kubernetes/predicates.go
```diff
@@ -168,7 +168,20 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 		r.log.Info("unexpected object type, bypassing reconciliation", "object", obj)
 		return false
 	}
+	labels := svc.GetLabels()
+	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
+	if ok {
+		gtw := r.findGateways(ctx, gclass)
```
arkodg, line 174: should we instead check if this gclass is the accepted gclass in the providerResources watchable message ?
and if so  just fetch the gateways from the message ?

internal/provider/kubernetes/predicates.go
```diff
@@ -180,6 +193,19 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 	return r.isRouteReferencingBackend(&nsName)
 }
 
+// findGateways attempts finds a GatewayList using accepted GatewayClass name.
```
arkodg, line 196: based on above comment, this func may not be needed

internal/provider/kubernetes/predicates.go
```diff
@@ -286,7 +312,18 @@ func (r *gatewayAPIReconciler) validateDeploymentForReconcile(obj client.Object)
 		r.log.Info("unexpected object type, bypassing reconciliation", "object", obj)
 		return false
 	}
-
+	labels := deployment.GetLabels()
+	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
```
arkodg, line 316: same as above

arkodg:
this is looking pretty good @cnvergence
added some comments
would be great if you can add some more comments in the code as well so devs can understand why the gatewayclass label is being used instead of the gateway label or why the IR keys are being computed differently

cnvergence:
> hey @cnvergence guessing now with this change when `mergeGateway: true`, the envoy proxy data plane fleet is created and deleted based on when a new gatewayclass is accepted and removed ?

hey @arkodg, yes, the envoy proxy fleet is redeployed when gatewayclass is referencing mergeGateways: true

internal/gatewayapi/helpers.go
cnvergence: alright, I thought I could use `-` as this is mainly used for container ports, but I will change it to `/` and add this in the infrastructure helper function

internal/provider/kubernetes/helpers.go
cnvergence: true, also I have merged `infraServiceName` and `infraDeploymentName` as they were the same

internal/provider/kubernetes/predicates.go
cnvergence: good catch

internal/provider/kubernetes/predicates.go
cnvergence: yes, that is much simpler, thanks :)

internal/provider/kubernetes/controller.go
cnvergence: Sure, I will create a separate PR for it, during tests I have found many reconciliations in the logs on Gateway status updates.

cnvergence: xref: https://github.com/envoyproxy/gateway/pull/1988

internal/gatewayapi/helpers.go
```diff
@@ -400,6 +406,10 @@ func irStringKey(gatewayNs, gatewayName string) string {
...
+	return fmt.Sprintf("%s/%s/%s", listener.Name, listener.gateway.Namespace, listener.gateway.Name)
```
arkodg, line 410: can we replace this with `return irHTTPListenerName(listener)` to improve code reuse ?

internal/gatewayapi/testdata/merge-valid-multiple-gateways.in.yaml
```diff
@@ -0,0 +1,40 @@
...
+            namespaces:
+              from: Same
+  - apiVersion: gateway.networking.k8s.io/v1beta1
+    kind: Gateway
+    metadata:
+      name: gateway-2
+      namespace: envoy-gateway
+    spec:
+      gatewayClassName: envoy-gateway-class
+      listeners:
```
arkodg, line 31: there's a bug, I dont see any listeners in gateway-2 being reflected in the xdsIR

internal/gatewayapi/helpers.go
cnvergence: sure

Xunzhuo (approved):
LGTM, hope this can land in v0.6.0-rc1, tested locally and it works well:

### Config:

```yaml

kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
---
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  mergeGateways: true
---
kind: Gateway
metadata:
  name: merged-eg-1
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 8080
    protocol: HTTP
---
kind: Gateway
metadata:
  name: merged-eg-2
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 8081
    protocol: HTTP
---
kind: Gateway
metadata:
  name: merged-eg-3
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 8082
    protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: hostname1-route
spec:
  parentRefs:
    - name: merged-eg-1
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example
---
kind: HTTPRoute
metadata:
  name: hostname2-route
spec:
  parentRefs:
    - name: merged-eg-2
  hostnames:
    - "www.example2.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example2
---
kind: HTTPRoute
metadata:
  name: hostname3-route
spec:
  parentRefs:
    - name: merged-eg-3
  hostnames:
    - "www.example3.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example3

```
### Result:

```shell
❯ kg gateway -A
NAMESPACE   NAME          CLASS   ADDRESS          PROGRAMMED   AGE
default     merged-eg-1   eg      172.18.255.200   True         12m
default     merged-eg-2   eg      172.18.255.200   True         12m
default     merged-eg-3   eg      172.18.255.200   True         12m
❯ kg httproute -A
NAMESPACE   NAME              HOSTNAMES              AGE
default     hostname1-route   ["www.example.com"]    27s
default     hostname2-route   ["www.example2.com"]   27s
default     hostname3-route   ["www.example3.com"]   27s
❯ kg svc -n envoy-gateway-system
NAME                            TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)                                        AGE
envoy-eg-d8c59e83               LoadBalancer   10.96.161.114   172.18.255.200   8080:31539/TCP,8081:30783/TCP,8082:31013/TCP   12m
envoy-gateway                   ClusterIP      10.96.116.219   <none>           18000/TCP,18001/TCP                            14m
envoy-gateway-metrics-service   ClusterIP      10.96.69.160    <none>           8443/TCP                                       14m
❯ kg pod -n envoy-gateway-system
NAME                                 READY   STATUS    RESTARTS   AGE
envoy-eg-d8c59e83-68fdf67bc9-pgb2v   1/1     Running   0          95s
envoy-gateway-6f557cc687-h8nxh       2/2     Running   0          14m
```

Maybe we need an e2e test for this as a follow-up, and also a user-guide docs.

internal/provider/utils/utils.go
```diff
@@ -38,3 +34,24 @@ func GetHashedName(nsName string) string {
...
+// If mergedGateways is enabled or listener port name is larger than 15 characters it will return partially hashed name.
+// Listeners on merged gateways have a infraIR port name {GatewayNamespace}/{GatewayName}/{ListenerName}.
+func ExpectedContainerPortHashedName(name string) string {
+	if len(name) > 15 {
+		hashedName := HashString(name)
+		// replace `/` with `-` to create a valid K8s resource name
+		resourceName := strings.ReplaceAll(name, "/", "-")
+		listenerName := string(resourceName[2])
+
+		return fmt.Sprintf("%s-%s", listenerName, hashedName[0:14-len(listenerName)])
```
arkodg, line 54: this is going to be prone to collisions, id recommend hardcoding this to 7 chars of the listener and 8 chars of the hash.

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -72,8 +73,10 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 			if port.Protocol == ir.UDPProtocolType {
 				protocol = corev1.ProtocolUDP
 			}
+			// Listeners on merged gateways will have a port name {GatewayNamespace}/{GatewayName}/{ListenerName}.
```
arkodg, line 76: doesnt this need to be hashed to ensure length is under 15 ?

arkodg (approved):
LGTM @cnvergence, thanks for adding this feature

there are a couple of comments mainly around hashed names of container ports that can be addressed in a follow up PR

@tamalsaha can you also help test this feature out

tamalsaha:
Thanks! Does this support cross namespace gateway merge?

arkodg:
@tamalsaha yes it should

internal/provider/utils/utils.go
cnvergence: Yeah, probably it makes more sense, I don't like it either
Initially didn't want to make the listener name that short, I will change it like you proposed in the follow-up

internal/infrastructure/kubernetes/proxy/resource_provider.go
cnvergence: AFAIK, Service port names apply under standard DNS label names, then it should be fine

arkodg: that means the length is 63 ? but im guessing the individual max length of `GatewayNamespace` , `GatewayName` & `ListenerName` is 63 so this need to be trimmed

cnvergence: That should be trimmed as well, let's just reuse ExpectedResourceHashedName for it

Issue #1928, opened 2023-10-05
hostNetwork support


yangchuansheng:
*Description*:
>I'd like to ask about hostNetwork support in Custom Resource EnvoyProxy

[optional *Relevant Links*:]
>https://gateway.envoyproxy.io/v0.5.0/api/config_types.html#kubernetesdeploymentspec

arkodg:
@yangchuansheng can you share your use case why you need this feature ?

arkodg:
copying ChatGPT's answer
```
In Kubernetes, the hostNetwork is a Pod-level setting that allows you to share the network namespace with the host machine on which the Pod is scheduled. When you configure a Pod to use the hostNetwork, it essentially means that the Pod directly uses the host machine's network stack, and it doesn't have its own isolated network namespace. This can be useful in certain scenarios, but it comes with some trade-offs and should be used with caution.

Here are some reasons why users might configure hostNetwork in Kubernetes:

Performance: By using the host's network stack, you can potentially achieve better network performance for applications that require low-latency communication. This is particularly relevant for applications that demand high network throughput.
Access to Host Networking Resources: Some applications might need direct access to network resources that are only available on the host machine, such as specific network devices or services. Configuring the Pod to use the host network allows access to these resources.
Port Conflicts: If your Pod needs to bind to a port that's already in use on the host, you might configure hostNetwork to avoid port conflicts. This is because, by default, Pods have their own IP address and network namespace, which can lead to port collisions if the host is already using the desired port.
However, there are several important considerations and drawbacks to using hostNetwork:

Isolation: Using hostNetwork sacrifices network isolation, which is one of the key benefits of containerization and Kubernetes. It can lead to security and stability issues, as Pods can potentially interfere with the host's network configuration.
Network Conflicts: If multiple Pods are configured to use the host's network, there can still be conflicts, as they might contend for network resources or ports. Proper coordination and planning are necessary in such scenarios.
Compatibility: Not all Kubernetes clusters or cloud environments support the use of hostNetwork, so it may limit the portability of your workloads.
Security: Running containers in the host's network namespace can pose security risks, as containers have a level of isolation by design to prevent them from directly impacting the host system. Using hostNetwork reduces this isolation.
[...]

PR #1944, opened 2023-10-10
feature: add hostNetwork to the podSpec

tanujd11:
**What type of PR is this?**

Feature

added hostNetwork to the podSpec

Fixes https://github.com/envoyproxy/gateway/issues/1928


Xunzhuo (commented):
LGTM

Xunzhuo (approved):

zirain:
Before adding this, I want to clearify the use case. The use of HostNetwork is discouraged.

fanux:
@zirain Although the use of hostnetwork is not encouraged, it is an exception in the scenario of gateways. Firstly, gateways need to expose ports such as 80 and 443, and hostnetwork is very convenient for this. Moreover, a gateway may be shared by many applications, so there will be no port conflicts. Therefore, even in a production environment, it is very convenient to use it in this way. Otherwise, it would be necessary to rely on external load balancing to expose ports 80 and 443, which would greatly increase the complexity for users, especially in situations where there is no external load balancing.

Xunzhuo (approved):
Re-activate this issue. @fanux and his team come to me and provided me some contexts on this request, their team tries to put EG into production, but this is a blocker for them.

In their scenarios, I think this is reasonable and this would be also a common request when deploying single EG with mergeGateways enabled in a cluster, which means there will be only one single envoyproxy instance, and will be no listeners conflicts. AND in private cloud clusters, there offen will be no loadbalancer implementations, if the envoyproxy needs to expose http-80/https-443, this cannot be worked easily, which needs some changes to kubernetes.

So I am +1 for it.

shawnh2 (approved):

qicz (approved):
LGTM

arkodg:
Im still not convinced this is a good idea, the project supports multiple ways to setup a gateway in non LoadBalancer mode
* ClusterIP
* NodePort
* External IP (using `gateway.spec.addresses`)

Setting up `hostNetwork` is not going to be enough and not going to work out of the box if trying to access the deployment directly, because of the custom service to container port mapping for privileged ports, this is not a bug, but a design decision, since the container port is an internal implementation and all traffic must be routed to the deployment via the service

Issue #1845, opened 2023-08-29
proposal: SecurityPolicy


AliceProxy:
## Relates to
#1492  
#1821 
#1846 

## What is this?
SecurityPolicy is a proposal for a new policy attachment resource that can be applied to `Gateways` and `xRoute` resources. It is meant to supplement a lot of config that Gateway API currently lacks while also providing a way to specify global defaults. It is similar to the proposal for an `UpstreamTrafficPolicy` with the notable difference that this tackles authentication based features and will thus support `overrides` and `defaults`.

## How would I use this resource?
You can attach an `SecurityPolicy` to a `Gateway` to set global defaults that will apply to all children xRoute objects of that `Gateway`. You can also create an `SecurityPolicy` to attach to an xRoute object to configure things on a route-to-route basis and override global defaults if needed. 

## How will the `overrides` and `defaults` work?
If there is an `SecurityPolicy` X attached to a `Gateway`, and an `SecurityPolicy` Y attached to an `HTTPRoute`:
- `overrides` from policy X will win in a conflict with `overrides` from policy Y.
  - This is to support the use-case of a cluster admin/operator setting some global overrides that are not allowed to be changed on a route level
- `defaults` from policy Y will win in a conflict with `defaults` from policy X.
  - While cluster admins/operators might want to provide sane defaults, developers creating xRoute objects should still be able to override those defaults on a route-to-route basis (unless otherwise enforced by `Gateway` level overrides in the above scenario) 

## What if Gateway API implements a GEP that delivers the same functionality as some of the fields of this resource?
If Gateway improves any of their existing resources to deliver functionality that meets all of the needs of any of the below config,
then we will deprecate the field in this resource and use the Gateway API config instead. Ideally, a lot of this config can be
upstreamed into Gateway API in some form eventually since this resource only exists to solve areas where Gateway API is lacking for the needs of Envoy Gateway.

## Can you use multiple `SecurityPolicy` resources at the same time?
You may use multiple `AuthPolicies` that target different resources, but you may not attach multiple `AuthPolicies`
to the same resource. For example, it is invalid to attach two `AuthPolicies` to the same `Gateway`. Merging config in this scenario would quickly become convoluted and confusing, and there are many edge-case scenarios that make supporting this undesirable.

## How will this be developed/implemented?

[...]

PR #1950, opened 2023-10-11
design: SecurityPolicy

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1845


qicz (approved):

site/content/en/latest/design/security-policy.md
```diff
@@ -0,0 +1,114 @@
...
+* External Authorization
+* Basic Auth
+* API Key Auth
+* CORS
+
+## Design Decisions
+* This API will only support a single `targetRef` and can bind to a `Gateway` resource or a `HTTPRoute` or `GRPCRoute`.
+* This API resource MUST be part of same namespace as the targetRef resource
+* There can be only be ONE policy resource attached to a specific targetRef e.g. a `Listener` (section)  within a `Gateway`
+* If the policy targets a resource but cannot attach to it, this information should be reflected
```
zhaohuabing, line 97: Is this the solution for OIDC before per-route configuration is supported?  https://github.com/envoyproxy/envoy/issues/29641

site/content/en/latest/design/security-policy.md
```diff
@@ -0,0 +1,114 @@
...
+in the Policy Status field using the `Conflicted=True` condition.
+* If multiple polices target the same resource, the oldest resource (based on creation timestamp) will
+attach to the Gateway Listeners, the others will not.
+* If Policy A has a `targetRef` that includes a `sectionName` i.e. 
+it targets a specific Listener within a `Gateway` and Policy B has a `targetRef` that targets the same
+entire Gateway then
+  * Policy A will be applied/attached to the specific Listener defined in the `targetRef.SectionName`
+  * Policy B will be applied to the remaining Listeners within the Gateway. Policy B will have an additional
+  status condition `Overridden=True`.
+* A Policy targeting a xRoute (`HTTPRoute` or `GRPCRoute`) overrides a Policy targeting a Listener that is
```
zhaohuabing, line 107: A policy targeting an xRoute may conflict with a policy targeting a Gateway. For example, the following two policies contains ext auth configurations with different grcp_service configured, but only one grpc_service can be set at the HCM filter level. 

Policy 1
```yaml
kind: SecurityPolicy
metadata:
  name: ext-auth-policy-gateway
  namespace: default
spec:
  extAuth:
    grpcService:
      envoyGrpc:
          clusterName: ext-authz-service1
  targetRef:
    kind: Gateway
    name: eg
    namespace: default
```

```yaml
kind: SecurityPolicy
metadata:
  name: ext-auth-policy-gateway
  namespace: default
spec:
  extAuth:
    grpcService:
      envoyGrpc:
          clusterName: ext-authz-service2
  targetRef:
    kind: HTTPRoute
    name: foo
    namespace: default   
```

We need to define how to handle this kind of conflict in the design docs. I can think of two rules:

* The oldest one wins.
* Or the one with the largest scope wins: Gateway > Listener > xRoute

arkodg: in the example you shared Policy 2 overrides Policy 1 and its entire spec wins since a xRoute target has high precedence over Gateway or Gateway section target (listener), so only Policy2's spec will be applied and translated

site/content/en/latest/design/security-policy.md
arkodg: we cannot add OIDC into this API until per route is supported

site/content/en/latest/design/security-policy.md
zhaohuabing: There may be other HTTPRoutes. So Policy2 will be applied to HTTPRoute foo, and Policy1 should be applied to the owner listeners of foo as the default config for other HTTPRoutes. The conflict still exists.

A simpler example of a conflict:

HTTPRoute foo and bar target the same Gateway


```yaml
kind: HTTPRoute
metadata:
  name: foo
  namespace: default
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend-foo
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /foo
---

kind: HTTPRoute
metadata:
  name: bar
  namespace: default
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend-bar
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /bar

```


Policy foo targets HTTPRoute foo, Policy bar targets HTTPRoute bar, and they have different extAuth configuration
```yaml
kind: SecurityPolicy
metadata:
  name: ext-auth-policy-foo
  namespace: default
spec:
  extAuth:
    grpcService:
      envoyGrpc:
          clusterName: ext-authz-service2
      ......
  targetRef:
    kind: HTTPRoute
    name: foo
    namespace: default 
---

kind: SecurityPolicy
metadata:
  name: ext-auth-policy-bar
  namespace: default
spec:
  extAuth:
    grpcService:
      envoyGrpc:
          clusterName: ext-authz-service2
      ......
  targetRef:
    kind: HTTPRoute
    name: bar
    namespace: default   
```

In the Envoy configuration, only one grpc_service is allowed.

```yaml
http_filters:
  - name: envoy.filters.http.ext_authz
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
      grpc_service:  // only one grpc service is allowed at the HCM filter level
        envoy_grpc:
          cluster_name: ext-authz-service1
      ......
```

JWT filter doesn't have this problem because the JWT filter configuration allows multiple providers configured.

```yaml
providers:
  provider1:
    issuer: https://provider1.com
    local_jwks:
      inline_string: PUBLIC-KEY
  provider2:
    issuer: https://provider2.com
    local_jwks:
      inline_string: PUBLIC-KEY
```

site/content/en/latest/design/security-policy.md
zhaohuabing: Adding Per-route configuration for OIDC is a big change, and it won't be available soon. How about we start by providing OIDC at the Gateway and Listener levels and then add per-route support when it's ready?   Listener-level OIDC should cover most of what EG users need.

Per-route OIDC issue: https://github.com/envoyproxy/envoy/issues/29641

arkodg: ptal @envoyproxy/gateway-maintainers

site/content/en/latest/design/security-policy.md
arkodg: can we create multiple ext authz filters using different names ?
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_filters#route-specific-config

site/content/en/latest/design/security-policy.md
zhaohuabing: I verified that per-route configuration can be accomplished with "route level filter disabled support", so this is no longer blocking us. https://github.com/zhaohuabing/playground/blob/main/envoy/per-route-oauth2-oidc/envoy.yaml

zhaohuabing (approved):
LGTM. Thanks for putting up the design.

zirain (approved):

PR #1955, opened 2023-10-11
design/impl: control plane metrics monitoring

Xunzhuo:
**What type of PR is this?**

design: control plane metrics monitoring

Fixes https://github.com/envoyproxy/gateway/issues/700

https://gateway.xunzhuo.cafe/latest/design/eg-metrics


Xunzhuo:
I have set up metrics pkg basement which can be easily used in different component, and we support push/pull metrics.
But I want to know what/where metrics we should build in each components:

+ Provider ( already have controller-runtime metrics )
+ GWAPI Translator
+ Infra Manager
+ xDS Translator
+ Extension Manager

I would like to hear from @envoyproxy/gateway-maintainers .

arkodg:
thanks for kickstarting this @Xunzhuo, hoping this lands in v0.6.0

sharing my thoughts on what we should be measuring
* **Input Resources from K8s** 
  * `kubernetes_provider_reconcile_duration_seconds` - gauge that measures how much time was spent in the [reconciler ](https://github.com/envoyproxy/gateway/blob/76704d1124eccaf06b81634a96390c95722a0554/internal/provider/kubernetes/controller.go#L163) , note this may already be provided by controller-runtime (haven't looked into this yet)
* **Translate Resources** 
  * `<runner_duration_seconds` - gauge that measures how much time was spent in every runner during a subscribe event e.g. in [here](https://github.com/envoyproxy/gateway/blob/76704d1124eccaf06b81634a96390c95722a0554/internal/xds/server/runner/runner.go#L138), with a unique label for every runner
* **Output Resources to Envoy**
  * `xds_push_duration_seconds` - gauge that measures time taken to push xds from xds server to envoy
* **Write Status to API Server** 
   * `<resource_status_write_kubernetes_duration_seconds>` - gauge that measures how much time it took to write status back to k8s server (this [func](https://github.com/envoyproxy/gateway/blob/76704d1124eccaf06b81634a96390c95722a0554/internal/status/status.go#L75)) with a unique label for every resource type


request to add a design doc that highlights a few things, which are already answered by your PR
* why this is needed
* what are we exactly measuring
* how we are observing the metrics

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,634 @@
+---
+date: 2023-10-10
```
arkodg, line 2: ptal @envoyproxy/gateway-reviewers @envoyproxy/gateway-maintainers

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,634 @@
...
+
+These attributes can be expressed in logs (as keys of structured logs), traces (as attributes), and metrics (as labels).
+
+We aim to use attributes consistently where applicable. Where possible, these should adhere to codified Semantic Conventions; when not possible, they should maintain consistency across the project.
+
+### Extensibility
+
+Envoy Gateway supports both **PULL/PUSH** mode metrics, with Metrics exported via Prometheus by default.
+
+Additionally, Envoy Gateway can export metrics using the [OTEL HTTP metrics exporter](https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/#general), which pushes metrics to a remote OTEL collector.
```
arkodg, line 67: should be start off with GRPC exporter since that is what we use by default in data plane
https://github.com/envoyproxy/gateway/blob/57b00ed7e01d69b479307f3996fd7878cb985fbf/internal/xds/bootstrap/bootstrap.yaml.tpl#L112

site/content/en/latest/design/pprof.md
```diff
@@ -1,15 +1,19 @@
 ---
```
zirain, line 1: prefer to move this out of this PR

site/content/en/latest/design/eg-metrics.md
Xunzhuo: Sure, I have added that.

site/content/en/latest/design/pprof.md
Xunzhuo: The debug server changes are related to this PR, we need to separate the debug server (including pprof) out of admin server.

internal/metrics/definition.go
```diff
@@ -0,0 +1,146 @@
...
+// It's used to visualize the statistical distribution of these observations.
+
+const (
+	CounterType   MetricType = "Counter"
+	GaugeType     MetricType = "Gauge"
+	HistogramType MetricType = "Histogram"
+)
+
+// A Metric collects numerical observations.
+type Metric interface {
```
zirain, line 54: is these something from istio/istio/pkg/monitoring?

site/content/en/latest/design/pprof.md
zirain: why?

internal/metrics/definition.go
Xunzhuo: The interface abstraction refers the metric design in istio, did not figure out more, top level metric abstraction should be similar, but with different implementations and registers/exporters.

zirain: I think there's some lib like this, at least, you should not copy the code or add reference.

zirain: not 100%, but maybe https://github.com/tetratelabs/telemetry can help.

Xunzhuo: https://github.com/tetratelabs/telemetry/blob/master/metric.go the metric difinitions are pretty similar to istio/istio/pkg/monitoring too, as I said, the top level interface abstraction of metric should be similar.

Xunzhuo: Or are you proposing directly use `github.com/tetratelabs/telemetry` as the Interfaces, and we do implementations like https://github.com/tetratelabs/telemetry-opencensus? But this will make us not that easy to maintain the metrics library if we want to add/change more interfaces

site/content/en/latest/design/pprof.md
Xunzhuo: debug server for now is mostly used for pprof, may be exposed more debugging features in the future, it is optional and not recommended to be used in production env, but the metrics should always be enabled.

It is better to use a sparate port for debugging/pprof with admin port.

And we can remove the logics here: 

image

By using controller-runtime metrics server, the reuse the logics in controller-runtime to enable pprof, and it is also not allowed in controller-runtime to expose metrics and pprof in the same port.

zirain: One thing per PR, let focus on metrics in this PR.
You can do pprof in a seperated one.

internal/metrics/definition.go
zirain: yes, we should not copy everything from istio or do something similar.

site/content/en/latest/design/pprof.md
Xunzhuo: Okay, make sense

Xunzhuo:
required: https://github.com/envoyproxy/gateway/pull/1967

site/content/en/latest/design/pprof.md
Xunzhuo: https://github.com/envoyproxy/gateway/pull/1967

internal/metrics/definition.go
Xunzhuo: Current version is not a copy from istio, different in both interfaces and implementations. I have refactored a couple of times, and tried directly to use tetrate/telemtry or istio/pkg/monitoring, both ways are not ideal and not suitable for eg, which spent me a lot of time on it. I believe current version is the ideal design for eg comp metrics obs.

api/v1alpha1/envoygateway_types.go
```diff
@@ -69,6 +69,12 @@ type EnvoyGatewaySpec struct {
 	// +optional
 	Debug *EnvoyGatewayDebug `json:"debug,omitempty"`
 
+	// Telemetry defines the desired control plane telemetry related abilities.
+	// If unspecified, the telemetry is used with default configuration.
```
arkodg, line 73: what is the default configuration ? is it opt in ?

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -112,6 +116,51 @@ func DefaultEnvoyGatewayLogging() *EnvoyGatewayLogging {
...
+	if e.Telemetry != nil {
+		if e.Telemetry.Metrics.Prometheus == nil {
+			e.Telemetry.Metrics.Prometheus = DefaultEnvoyGatewayPrometheus()
+		}
+		if e.Telemetry.Metrics == nil {
+			e.Telemetry.Metrics = DefaultEnvoyGatewayMetrics()
+		}
+		return e.Telemetry
+	}
+	e.Telemetry = DefaultEnvoyGatewayTelemetry()
```
arkodg, line 130: why are we enabling prometheus metrics by default ?

api/v1alpha1/envoygateway_metric_types.go
```diff
@@ -0,0 +1,44 @@
...
+	Port int32 `json:"port,omitempty"`
+}
+
+// EnvoyGatewayPrometheusProvider will expose prometheus endpoint in pull mode.
+type EnvoyGatewayPrometheusProvider struct {
+	// Enable defines if enables the prometheus metrics in pull mode. Default is true.
+	//
+	// +optional
+	// +kubebuilder:default=true
+	Enable bool `json:"enable,omitempty"`
```
arkodg, line 43: why do we need an explicit `enable` flag, isn't the instantiation of `EnvoyGatewayPrometheusProvider` an implicit enable ?

api/v1alpha1/envoygateway_metric_types.go
```diff
@@ -0,0 +1,44 @@
...
+// EnvoyGatewayMetricSink defines control plane
+// metric sinks where metrics are sent to.
+type EnvoyGatewayMetricSink struct {
+	// Type defines the metric sink type.
+	// EG control plane currently supports OpenTelemetry.
+	// +kubebuilder:validation:Enum=OpenTelemetry
+	// +kubebuilder:default=OpenTelemetry
+	Type MetricSinkType `json:"type"`
+	// Host define the sink service hostname.
+	Host string `json:"host"`
```
arkodg, line 25: should we start using urls here instead of using 3 fields ?
cc @zirain

internal/message/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+	)
+
+	watchableHandleUpdateTimeSeconds = metrics.OpenTelemetrySink().NewHistogram(
+		"watchable_queue_handle_update_time_seconds",
+		"How long in seconds a update handled by watchable queue.",
+		[]float64{0.001, 0.01, 0.1, 1, 5, 10},
+	)
+
+	// metrics label definitions
+	// component is which component the update belong to.
```
arkodg, line 34: you cannot map a single component to watchable, because it involves a subscriber component and a publisher component.
Also we internally call components `runners`

internal/message/metrics.go
```diff
@@ -0,0 +1,38 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package message
+
+import "github.com/envoyproxy/gateway/internal/metrics"
+
+var (
```
arkodg, line 10: these metrics need to be split up in terms of subscribe and publish

api/v1alpha1/envoygateway_helpers.go
zirain: I think most people in Kubernetes are/will use prometheus, it should always enable by default.

internal/metrics/api.go
```diff
@@ -0,0 +1,73 @@
+// Copyright Envoy Gateway Authors
```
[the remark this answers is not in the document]
Xunzhuo: Sure, I would like to separate it with 3 PR:
+ design
+ library
+ metrics instrumentation

Xunzhuo:
Close in favor of separations

PR #1957, opened 2023-10-12
feat: init SecurityPolicy

zhaohuabing:
Relates to https://github.com/envoyproxy/gateway/issues/1845


internal/gatewayapi/securitypolicy.go
```diff
@@ -0,0 +1,250 @@
...
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
+	"github.com/envoyproxy/gateway/internal/status"
+	"github.com/envoyproxy/gateway/internal/utils/ptr"
+)
+
+func securityPolicyHasSectionName(policy *egv1a1.SecurityPolicy) bool {
+	return policy.Spec.TargetRef.SectionName != nil
+}
+
+func ProcessSecurityPolicies(securityPolicies []*egv1a1.SecurityPolicy,
```
arkodg, line 27: need to be able to support 3 levels of targetRefs
* Gateway
* Gateway with sectionName (Listener)
* HTTPRoute or GRPCRoute

api/v1alpha1/securitypolicy_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// Spec defines the desired state of SecurityPolicy.
+	Spec SecurityPolicySpec `json:"spec"`
+
+	// Status defines the current status of SecurityPolicy.
+	Status SecurityPolicyStatus `json:"status,omitempty"`
+}
+
+// SecurityPolicySpec defines the desired state of SecurityPolicy.
+type SecurityPolicySpec struct {
+	// TargetRef is the name of the Gateway resource this policy
```
arkodg, line 38: need to be able to support 3 levels of targetRefs

Gateway
Gateway with sectionName (Listener)
HTTPRoute or GRPCRoute

arkodg:
also needs a status eq check in https://github.com/envoyproxy/gateway/blob/57b00ed7e01d69b479307f3996fd7878cb985fbf/internal/status/status.go#L172

api/v1alpha1/securitypolicy_types.go
arkodg: as well compute `Override=True` as well as `Conflicted=True`

zhaohuabing:
Hold this since it has a similar structure to [BackendTrafficPolicy](https://github.com/envoyproxy/gateway/pull/1961).  These two should align with each other.

zhaohuabing:
If this is acceptable, I'd like to add design and user docs of SecurityPolicy later when we have a concrete feature coming in, like CORS or JWT.

arkodg (approved):

zirain (approved):

PR #1961, opened 2023-10-13
BackendTrafficPolicy init + design doc

AliceProxy:
Initializes the `BackendTrafficPolicy` custom resource and API.

I added some extra CRD validation to catch a couple things early instead of waiting to get to the runtime condition checks.
Here's some screenshots from when I was working on that so you can get an idea of what a user might see when trying to apply and fix an invalid resource. I'm happy to open a PR tweaking the existing ClientTrafficPolicy CRD to have similar validation if we're good with it.

Screenshot from 2023-10-12 09-55-17

Screenshot from 2023-10-12 09-55-53

Screenshot from 2023-10-12 09-56-07


zhaohuabing:
> Initializes the `BackendTrafficPolicy` custom resource and API.
> 
> I added some extra CRD validation to catch a couple things early instead of waiting to get to the runtime condition checks. 

I need similar validation for SecurityPolicy as well. Did you add a validation webhook?

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+					gwv1a2.PolicyConditionAccepted,
+					metav1.ConditionFalse,
+					gwv1a2.PolicyReasonConflicted,
+					message,
+				)
+				continue
+			}
+
+			// Check if any policies that target xRoute resources are overriding this policy so we can set the status
+			if _, ok := overriddenPolicyTargets[*gatewayKey]; ok {
```
zhaohuabing, line 152: The policy targeting a gateway section may not overlap with the policy targeting a xRoute. For exmaple, they may have different hostnames.

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+				message := fmt.Sprintf("There are existing BackendTrafficPolicies that are overriding these sections %v", sections)
+
+				status.SetBackendTrafficPolicyCondition(policy,
+					egv1a1.PolicyConditionOverridden,
+					metav1.ConditionTrue,
+					egv1a1.PolicyReasonOverridden,
+					message,
+				)
+			}
+
```
zhaohuabing, line 223: We should also check if there are existing policies at the route level.

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -0,0 +1,72 @@
...
+	gwapiv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+const (
+	// KindBackendTrafficPolicy is the name of the BackendTrafficPolicy kind.
+	KindBackendTrafficPolicy = "BackendTrafficPolicy"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=btpolicy
```
arkodg, line 19: btp ?

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -0,0 +1,72 @@
...
+	Spec BackendTrafficPolicySpec `json:"spec"`
+
+	// status defines the current status of BackendTrafficPolicy.
+	Status BackendTrafficPolicyStatus `json:"status,omitempty"`
+}
+
+// spec defines the desired state of BackendTrafficPolicy.
+type BackendTrafficPolicySpec struct {
+
+	// +kubebuilder:validation:XValidation:rule="has(self.sectionName) ? self.kind == 'Gateway': true", message="sectionName can only be set when kind is 'Gateway'."
```
arkodg, line 41: thanks for adding these !

raised https://github.com/envoyproxy/gateway/issues/1964 to add validation tests

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// +kubebuilder:validation:XValidation:rule="self.kind == 'Gateway' || self.kind == 'HTTPRoute' || self.kind == 'GRPCRoute' || self.kind == 'UDPRoute' || self.kind == 'TCPRoute' || self.kind == 'TLSRoute'", message="this policy can only have a targetRef.kind of Gateway/HTTPRoute/GRPCRoute/TCPRoute/UDPRoute/TLSRoute"
+	//
+	// targetRef is the name of the resource this policy
+	// is being attached to.
+	// This Policy and the TargetRef MUST be in the same namespace
+	// for this Policy to have effect and be applied to the Gateway.
+	TargetRef gwapiv1a2.PolicyTargetReferenceWithSectionName `json:"targetRef"`
```
arkodg, line 48: we should add a line here about policy hierarchy as well as policy override i.e.
* Only one policy can be successfully attached to a specific target
* Since multiple policies can be indirectly applied to the same target e.g. a policy targeting a Gateway as well as a policy targeting a xRoute attached to the same Gateway, the policy with the most specific target will always have higher precedence (in this case the policy targeting the xRoute) over the less specific one (the policy targeting the Gateway)

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+			targetedRoutes[routeKey] = true
+			route := routeMap[*target]
+			for _, parentRef := range GetParentReferences(route) {
+				namespace := ""
+				if parentRef.Namespace != nil {
+					namespace = string(*parentRef.Namespace)
+				}
+				overriddenPolicyTargets[types.NamespacedName{
```
arkodg, line 107: can you explain this logic ?

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+				metav1.ConditionTrue,
+				gwv1a2.PolicyReasonAccepted,
+				"BackendTrafficPolicy has been accepted.",
+			)
+		}
+	}
+
+	// Process the policies targeting Gateways with a section name
+	for _, policy := range backendTrafficPolicies {
+		if policy.Spec.TargetRef.SectionName != nil {
```
arkodg, line 126: in theory a route will never reach here because of the CEL validation in the CRD, but this maybe hard for the next dev looking at this to figure that out
can we add an extra checking in here
```
policy.Spec.TargetRef.Kind == KindGateway &&  .....
```

internal/gatewayapi/backendtrafficpolicy.go
arkodg: nvm, looks like we are proactively making sure the parent gateways are overridden, a comment here would be great

internal/gatewayapi/backendtrafficpolicy.go
arkodg: a xRoute will never enter this code section

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+					metav1.ConditionFalse,
+					gwv1a2.PolicyReasonConflicted,
+					message,
+				)
+				continue
+			}
+
+			// Check if any policies that target xRoute resources are overriding this policy so we can set the status
+			if _, ok := overriddenPolicyTargets[*gatewayKey]; ok {
+				message := "There are existing BackendTrafficPolicies that are overriding this config at the route level"
```
arkodg, line 153:
```suggestion
				message := "There are existing BackendTrafficPolicies that are overriding this config at the Gateway listener level"
```

internal/gatewayapi/backendtrafficpolicy.go
arkodg: I believe this is happening on L109

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+func resolveBTPolicyTargetRef(policy *egv1a1.BackendTrafficPolicy, gateways map[types.NamespacedName]*GatewayContext, routes map[types.NamespacedName]RouteContext) *types.NamespacedName {
+	targetNs := policy.Spec.TargetRef.Namespace
+	// If empty, default to namespace of policy
+	if targetNs == nil {
+		targetNs = ptr.To(gwv1b1.Namespace(policy.Namespace))
+	}
+
+	// Ensure Policy and target are in the same namespace
```
arkodg, line 252: curious if this can be moved to CEL Validation, not a blocker, we can raise a GH issue to track this

internal/gatewayapi/translator.go
```diff
@@ -184,3 +200,103 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
...
+		gw := gw
+		gateways[types.NamespacedName{Name: gw.Name, Namespace: gw.Namespace}] = true
+	}
+
+	// Check all the different types of routes to see if any belong to one of our relevant gateways
+	for _, httpRoute := range httpRoutes {
+		for _, parentGW := range httpRoute.Spec.ParentRefs {
+			key := types.NamespacedName{Name: string(parentGW.Name)}
+			if parentGW.Namespace != nil {
+				key.Namespace = string(*parentGW.Namespace)
```
arkodg, line 225: when `parentGW.Namespace == nil` ,  `key.Namespace = httpRoute.Namespace` and the same is true for all other route types

internal/gatewayapi/translator.go
```diff
@@ -129,6 +133,15 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
 	// Get Gateways belonging to our GatewayClass.
 	gateways := t.GetRelevantGateways(resources.Gateways)
 
+	// Get Routes belonging to our GatewayClass.
+	routes := t.GetRelevantRoutes(gateways,
```
arkodg, line 137: I dont think this is needed, L160-L170 already compute the relevant routes today e.g.
https://github.com/envoyproxy/gateway/blob/2ecc54da663dc10c9257a61843cd14ba1bb77dcc/internal/gatewayapi/route.go#L63
we just need to append it into a list and pass it over

internal/gatewayapi/translator.go
```diff
@@ -184,3 +200,103 @@ func (t *Translator) GetRelevantGateways(gateways []*v1beta1.Gateway) []*Gateway
 
 	return relevant
 }
+
+func (t *Translator) GetRelevantRoutes(relevantGateways []*GatewayContext,
```
arkodg, line 204: as shared above, dont think this func is needed

site/content/en/latest/design/backend-traffic-policy.md
```diff
@@ -0,0 +1,155 @@
...
+            type: PathPrefix
+            value: /
+---
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: BackendTrafficPolicy
+metadata:
+  name: default-ipv-policy
+  namespace: default
+spec:
+  protocols:
```
arkodg, line 101: can we use a different feature as an example, maybe load balancing
ipv6 routing is tied to L3 which EG cannot really control, it can only make sure that if ipv6 endpoints exist, it can successfully fetch it and program it

site/content/en/latest/design/backend-traffic-policy.md
```diff
@@ -0,0 +1,155 @@
...
+in the Policy Status field using the `Conflicted=True` condition.
+- If multiple polices target the same resource, the oldest resource (based on creation timestamp) will
+attach to the Gateway Listeners, the others will not.
+- If Policy A has a `targetRef` that includes a `sectionName` i.e.
+it targets a specific Listener within a `Gateway` and Policy B has a `targetRef` that targets the same
+entire Gateway then
+  - Policy A will be applied/attached to the specific Listener defined in the `targetRef.SectionName`
+  - Policy B will be applied to the remaining Listeners within the Gateway. Policy B will have an additional
+  status condition `Overridden=True`.
+
```
arkodg, line 150: needs another line about policy hierarchy/precendance ( Route > Listener > Gateway)

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -0,0 +1,338 @@
...
+					message,
+				)
+
+				continue
+			}
+
+			targetedRoutes[routeKey] = true
+			route := routeMap[*target]
+			for _, parentRef := range GetParentReferences(route) {
+				namespace := ""
```
arkodg, line 103: should default to namespace of the route

arkodg: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.ParentReference

internal/gatewayapi/backendtrafficpolicy.go
arkodg: the `sectionName` of the parentReference is not being saved in L107 so we dont know for sure if this listener section is a parent of the xRoute which has a policy targeted to it

arkodg:
this PR is close to the finish line, thanks for complementing the policy hierarchy complexity with test cases, added some comments, ptal

arkodg:
also 2 more things needed in this PR is
* a way to subscribe to the status
https://github.com/envoyproxy/gateway/blob/2ecc54da663dc10c9257a61843cd14ba1bb77dcc/internal/provider/kubernetes/controller.go#L1263
* a way to skip status write if the status in the server is the same
https://github.com/envoyproxy/gateway/blob/2ecc54da663dc10c9257a61843cd14ba1bb77dcc/internal/status/status.go#L171

internal/gatewayapi/backendtrafficpolicy.go
zhaohuabing: L109 `overriddenPolicyTargets` is only used to check conflicts with listeners at L152. Ideally, we should also check if there are existing policies at the route level that override the gateway-level policies.

arkodg:
closed in favor of https://github.com/envoyproxy/gateway/pull/2053

PR #1967, opened 2023-10-16
feat: separate debug server with admin

Xunzhuo:
**What type of PR is this?**

debug server for now is mostly used for pprof, may be exposed more debugging features in the future, it is optional and not recommended to be used in production env, but the metrics should always be enabled.

It is better to use a sparate port for debugging/pprof with admin port.


zirain (approved):

qicz (approved):
LGTM，Thanks

arkodg:
@Xunzhuo im not sure how this related to control plane metrics, if we want to separate `Admin/Debug` from `Metrics` wouldn't the config be different ?
```
telemetry:
  metrics:
    prometheus:
      listenAddress: .....
```

PR #1972, opened 2023-10-16
feat: introduce initContainers field on EnvoyProxy API

### LINKED ISSUE #1946: Support `init` containers for Envoy deployments managed by EG
*Description*:

EG uses the EnvoyProxy resource to configure Envoy deployments it creates and manages. Today the API allows users to configure [a lot of aspects of the Envoy deployment](https://github.com/envoyproxy/gateway/blob/2699416c9f086f8f8ee4bc788537f9ca1c791fd1/api/v1alpha1/shared_types.go#L49) in k8s, but does not allow init containers to be configured. This makes it challenging to do things like use Vault's init container to provide initial certs, or use [simple patterns for loading WASM binaries into an Envoy container](https://github.com/corazawaf/coraza-proxy-wasm/pull/154) without a custom Envoy image.

It would be helpful in an `InitContainer` field was added to the [`KubernetesDeploymentSpec`](https://github.com/envoyproxy/gateway/blob/2699416c9f086f8f8ee4bc788537f9ca1c791fd1/api/v1alpha1/shared_types.go#L49) to allow init containers to be provided by the user. Even better if this field was repeated to allow for multiple init containers. (The same could be said for making the `Container` field repeated, to allow multiple containers in the Envoy pod, though there's not a clear use case for that today unlike init containers.)

ardikabs:
Fixes #1946


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #1975, opened 2023-10-16
fix: update readme link according to new website

slayer321:
**What type of PR is this?**

`fix: update readme link according to new website`

As we have migrated to Hugo. I found that most of the link on our README is not pointing to correct link. So updated that.

Fixes #


arkodg (approved):
thanks !

zirain (approved):

PR #1981, opened 2023-10-17
design(docs/api): control plane metrics monitoring

Xunzhuo:
**What type of PR is this?**

design(docs/api): control plane metrics monitoring

Part of: https://github.com/envoyproxy/gateway/issues/700


api/v1alpha1/envoygateway_helpers.go
```diff
@@ -88,6 +92,70 @@ func DefaultEnvoyGatewayLogging() *EnvoyGatewayLogging {
...
+	return &EnvoyGatewayMetricsAddress{
+		Host: GatewayMetricsHost,
+		Port: GatewayMetricsPort,
+	}
+}
+
+// DefaultEnvoyGatewayPrometheus returns a new EnvoyGatewayMetrics with default configuration parameters.
+func DefaultEnvoyGatewayPrometheus() *EnvoyGatewayPrometheusProvider {
+	return &EnvoyGatewayPrometheusProvider{
+		// Enable prometheus pull by default.
```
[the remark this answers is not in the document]
Xunzhuo: As we discussed in Slack, we should enable prometheus in default, let us Use `disable` for it.

site/content/en/latest/install/install-helm.md
```diff
@@ -124,11 +124,13 @@ These are the ports used by Envoy Gateway and the managed Envoy Proxy.
...
-|:----------------------:|:---------:|:------:|
-| Xds EnvoyProxy Server  | 0.0.0.0   | 18000  |
-| Xds RateLimit Server   | 0.0.0.0   | 18001  |
-| Admin Server           | 127.0.0.1 | 19000  |
+| Envoy Gateway          | Address   |  Port  |  Configurable  |
+|:----------------------:|:---------:|:------:|    :------:    |
+| Xds EnvoyProxy Server  | 0.0.0.0   | 18000  |       Yes      |
+| Xds RateLimit Server   | 0.0.0.0   | 18001  |       Yes      |
+| Admin Server           | 127.0.0.1 | 19000  |       Yes      |
+| Metrics Server         |  0.0.0.0  | 19001  |       Yes      |
```
arkodg, line 132: none of these are configurable atm, can we add `No` and later move it to `Yes` ?

arkodg:
hey @Xunzhuo can you rm the `png` from the diff, thanks

arkodg:
@Xunzhuo this PR mostly LGTM from my end but it also has the design doc which includes the specific metric names as well as the design for the internal metrics lib which is being discussed in https://github.com/envoyproxy/gateway/pull/1982. Suggest either only keeping the API in the design doc and adding the metrics lib design later or moving the design doc out of this PR

site/content/en/latest/install/install-helm.md
Xunzhuo: Admin Server and Metrics Server can be configured by EnvoyGateway Config.

Xunzhuo:
> @Xunzhuo this PR mostly LGTM from my end but it also has the design doc which includes the specific metric names as well as the design for the internal metrics lib which is being discussed in #1982. Suggest either only keeping the API in the design doc and adding the metrics lib design later or moving the design doc out of this PR

Removed from the docs.

api/v1alpha1/envoygateway_metrics_types.go
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// EnvoyGatewayMetrics defines control plane push/pull metrics configurations.
+type EnvoyGatewayMetrics struct {
+	// Address defines the address of Envoy Gateway Metrics Server.
+	Address *EnvoyGatewayMetricsAddress
```
Xunzhuo, line 11: One last check point, I am not sure if we should serve the metrics server if we disabled prometheus.

If we do not serve the metrics server, we should be better to put `Address` into `EnvoyGatewayPrometheusProvider`.

If we still serve the metrics port if we disable the prometheus, the `Address` should be outside of the Prometheus.

cc @zirain @arkodg

arkodg: I would just remove Address for now, and hard code to `0.0.0.0:19001`, this is what is done in data plane

Xunzhuo: It is Okay to me, but if people disabled Prometheus, should we still serve 0.0.0.0:19001?

arkodg: We shouldn't, that port should be disabled

Xunzhuo: https://github.com/envoyproxy/gateway/pull/2021 is working on configuring metrics port, I think eg cp metrics port should be supported to be configurable.

arkodg: Yeah we should in the future, vote to add it in a future release than in this release. Getting the API right will require iterations and atm we don't have enough time to get it in

arkodg (dismissed):
LGTM thanks !

api/v1alpha1/envoygateway_metrics_types.go
Xunzhuo: Rmoved.

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
...
+
++ Supporting other stats sinks.
++ Only focusing on code design and does not provide specific code implementation.
+
+## Use-Cases
+
+The use-cases include:
+
++ Exposing Prometheus metrics in the Envoy Gateway Control Plane.
++ Pushing Envoy Gateway Control Plane metrics via the Open Telemetry Sink.
```
zirain, line 47: this more like the `Goals`, the user case should be more explicit

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
...
+	// Disable defines if disables the prometheus metrics in pull mode.
+	//
+	Disable bool `json:"disable,omitempty"`
+}
+
+```
+
+#### Example
+
++ The following is an example to enable prometheus metric.
```
zirain, line 176: prometheus is enable by default

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
...
+
+Metrics offer the greatest potential for providing guarantees. They often directly influence alerts and dashboards, making changes highly impactful. This contrasts with traces and logs, which are often used for ad-hoc analysis, where minor changes to information can be easily understood by a human.
+
+Moreover, there is precedent for this: [Kubernetes Metrics Lifecycle](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metric-lifecycle) has well-defined processes, and Envoy Gateway's dataplane (Envoy Proxy) metrics are de facto stable.
+
+Currently, all Envoy Gateway metrics lack defined stability. I suggest we categorize all existing metrics as either:
+
++ ***Deprecated***: a metric that is intended to be phased out.
++ ***Experimental***: a metric that is off by default.
++ ***Alpha***: a metric that is on by default.
```
zirain, line 106: how can you categorize these?

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
...
+At present, the Envoy Gateway control plane provides logs and controller-runtime metrics, without traces. Logs are managed through our proprietary library (`internal/logging`, a shim to `zap`) and are written to `/dev/stdout`.
+
+The absence of comprehensive and robust control plane metrics observability hinders the effective monitoring of Envoy Gateway in a production environment, a critical requirement before deploying Envoy Gateway into production.
+
+## Goals
+
+Our objectives include:
+
++ Supporting **PULL** mode for Prometheus metrics and exposing these metrics on the admin address.
++ Supporting **PUSH** mode for Prometheus metrics, thereby sending metrics to the Open Telemetry Stats sink.
```
zirain, line 33: You don't mention HTTP or gRPC here, it seems that you use HTTP by default, so what about gRPC?

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
...
++ Metrics in **PULL** mode: The OTEL collector can scrape Prometheus and export to X.
++ Metrics in **PUSH** mode: The OTEL collector can receive OTEL gRPC/HTTP exporter metrics and export to X.
+
+While the examples above involve OTEL collectors, there are numerous other systems available.
+
+#### Vendor extensions
+
+The OTEL libraries allow for the registration of Providers/Handlers. While we will offer the default ones (PULL via Prometheus, PUSH via OTEL HTTP metrics exporter) mentioned in Envoy Gateway's extensibility, we can easily allow custom builds of Envoy Gateway to plug in alternatives if the default options don't meet their needs.
+
+For instance, users may prefer to write metrics over the OTLP gRPC metrics exporter instead of the HTTP metrics exporter. This is perfectly acceptable -- and almost impossible to prevent. The OTEL has ways to register their providers/exporters, and Envoy Gateway can ensure its usage is such that it's not overly difficult to swap out a different provider/exporter.
```
zirain, line 86: does this doc spend too much paragraph on what's OTel, how OTel/OTel-collector work?

zirain (commented):
Great work, leave some comments.

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,238 @@
+---
+date: 2023-10-10
```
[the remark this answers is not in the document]
Xunzhuo: No, we should set date from now now, it sorted docs with date instead of naming.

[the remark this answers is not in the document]
zirain: The design doc needs to be kept updated, use alphabetical order is fine.

zirain:
please do not use force push, it's hard to track the changes for a large PR.

arkodg (approved):

site/content/en/latest/design/eg-metrics.md
```diff
@@ -0,0 +1,225 @@
...
+	// It's required if the sink type is OpenTelemetry.
+	OpenTelemetry *EnvoyGatewayOpenTelemetrySink `json:"openTelemetry,omitempty"`
+}
+
+type EnvoyGatewayOpenTelemetrySink struct {
+	// Host define the sink service hostname.
+	Host string `json:"host"`
+	// Protocol define the sink service protocol.
+	// +kubebuilder:validation:Enum=grpc;http
+	Protocol string `json:"protocol"`
```
zirain, line 143: from the default value of port, I suspect default value for this field is `grpc`.

arkodg: imo its fine to start w/o a default

zirain (approved):
Let's merge this first

PR #1988, opened 2023-10-17
add GenerationChanged predicate filter to all resources

cnvergence:
**What type of PR is this?**

Per review of https://github.com/envoyproxy/gateway/pull/1938#discussion_r1358956623, it would be valuable to add GenerationChanged predicate to all resources.
This skip events that have no change in the object Spec and only the metadata and/or status fields are changed.

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1295,16 +1295,16 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	if err := c.Watch(
 		source.Kind(mgr.GetCache(), &gwapiv1b1.GatewayClass{}),
 		handler.EnqueueRequestsFromMapFunc(r.enqueueClass),
-		predicate.NewPredicateFuncs(r.hasMatchingController),
 		predicate.GenerationChangedPredicate{},
+		predicate.NewPredicateFuncs(r.hasMatchingController),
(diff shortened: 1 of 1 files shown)


internal/provider/kubernetes/controller.go
```diff
@@ -1296,13 +1296,15 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
 		predicate.NewPredicateFuncs(r.hasMatchingController),
+		predicate.GenerationChangedPredicate{},
 	); err != nil {
 		return err
 	}
 
 	// Only enqueue EnvoyProxy objects that match this Envoy Gateway's GatewayClass.
 	epPredicates := []predicate.Predicate{
 		predicate.ResourceVersionChangedPredicate{},
+		predicate.GenerationChangedPredicate{},
```
arkodg, line 1307: is the predicate func as OR op or an AND op ?
if its an AND op then we should be adding `GenerationChangedPredicate` to the beginning of the list

arkodg (approved):
LGTM thanks !
should hopefully reduce reconciles

zhaohuabing (approved):
LGTM. Thanks for the optimization!

zirain (approved):
Thanks!

zirain:
@cnvergence sorry, I merge this by mistake, can you resubmit this and fix the CI?

PR #1999, opened 2023-10-18
fix: wrong gatewayclass status without paramsRef

cnvergence:

Fixing the bug: 
Providing GatewayClass without paramsRef
```bash
kind: GatewayClass
metadata:
  name: example-class
spec:
```
results in error in the provider, not accepting the GatewayClass.
```bash
ERROR provider kubernetes/controller.go:330 failed to process parametersRef for gatewayclass
{"runner": "provider", "name": "example-class", "error": "failed to find envoyproxy referenced by gatewayclass: example-class"}
```
```bash
  status:
    conditions:
    - message: 'Invalid parametersRef: failed to find envoyproxy referenced by gatewayclass:
        example-class'
      observedGeneration: 2
      reason: InvalidParameters
      status: "False"
      type: Accepted
 
NAME            CONTROLLER                                      ACCEPTED   AGE
example-class   gateway.envoyproxy.io/gatewayclass-controller   False      2m
```

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1629,11 +1629,6 @@ func (r *gatewayAPIReconciler) hasManagedClass(obj client.Object) bool {
 
 // processParamsRef processes the parametersRef of the provided GatewayClass.
 func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1b1.GatewayClass, resourceTree *gatewayapi.Resources) error {
-	if gc.Spec.ParametersRef != nil {
-		if !refsEnvoyProxy(gc) {
-			return fmt.Errorf("unsupported parametersRef for gatewayclass %s", gc.Name)
-		}
-	}
 	epList := new(egv1a1.EnvoyProxyList)
 	// The EnvoyProxy must be in the same namespace as EG.
 	if err := r.client.List(ctx, epList, &client.ListOptions{Namespace: r.namespace}); err != nil {
@@ -1688,12 +1683,16 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
 		}
 	}
 
-	if !found {
-		return fmt.Errorf("failed to find envoyproxy referenced by gatewayclass: %s", gc.Name)
-	}
-
-	if !valid {
-		return fmt.Errorf("invalid gatewayclass %s: %v", gc.Name, validationErr)
+	if gc.Spec.ParametersRef != nil {
+		if !refsEnvoyProxy(gc) {
+			return fmt.Errorf("unsupported parametersRef for gatewayclass %s", gc.Name)
+		}
+		if !found {
+			return fmt.Errorf("failed to find envoyproxy referenced by gatewayclass: %s", gc.Name)
+		}
+		if !valid {
+			return fmt.Errorf("invalid gatewayclass %s: %v", gc.Name, validationErr)
+		}
 	}
 
 	return nil


cnvergence:
PTAL @arkodg 
seems like we have missed something during this
https://github.com/envoyproxy/gateway/pull/1534#discussion_r1307999971

internal/provider/kubernetes/controller.go
```diff
@@ -1629,11 +1629,6 @@ func (r *gatewayAPIReconciler) hasManagedClass(obj client.Object) bool {
 
 // processParamsRef processes the parametersRef of the provided GatewayClass.
 func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1b1.GatewayClass, resourceTree *gatewayapi.Resources) error {
-	if gc.Spec.ParametersRef != nil {
-		if !refsEnvoyProxy(gc) {
-			return fmt.Errorf("unsupported parametersRef for gatewayclass %s", gc.Name)
-		}
-	}
```
arkodg, line 1636: here's the issue with this new code that was added
* its a `processParamsRef` function for a specific `gc` but it fetches all `EnvoyProxies` and then tries to remove the finalizer for unrelated `EnvoyProxies` 
I see another bug on L1675 where we are removing a finalizer from the EnvoyProxy without actually checking if the EnvoyProxy is being referenced by any other GatewayClass or not.


We are very close to our v0.6 release and I prefer reverting https://github.com/envoyproxy/gateway/pull/1534 and adding the feature later when all issues have been addressed

cnvergence: This L1675 was actually added to support this case https://github.com/envoyproxy/gateway/pull/1534#issuecomment-1691130263

arkodg: yes, but the way its written it, it can just rm another EnvoyProxy's finalizer which may actually be referenced by a GatewayClass

cnvergence: I assumed it was supporting only one gateway class and therefore only one envoy proxy config or am I understanding it wrong?
Finalizer will be only there if it was referenced by a GatewayClass that was accepted.

cnvergence: `its a processParamsRef function for a specific gc but it fetches all EnvoyProxies and then tries to remove the finalizer for unrelated EnvoyProxies`
I agree with that, it could definitely be refactored now and have a list taken out of this function

arkodg: the issue is - its a one level check for the EnvoyProxy without backtracking which GatewayClass it belongs to, it may belong to a accepted GatewayClass from another EG controller

cnvergence: Hmm, that could be the truth, so we would need to add the check for ControllerName as well

arkodg: ive raised https://github.com/envoyproxy/gateway/pull/2000 for the revert
@cnvergence we really appreciate the hard work on this feature, the provider codebase is already so complex, and there are many edge cases that needed to be handled before we introduce the `Envoyproxy` finalizer which is not a must have yet, so prefer reverting for now

cnvergence: I agree, there are a lot of corner cases, I would be happy to help again, after the 0.6.0.

arkodg: thanks for understanding, I actually prefer pausing on this one, until we've cleaned up some of the GC and EP handling in the code before re adding it

Issue #1998, opened 2023-10-18
Decision: Enable Prometheus Endpoint by default for EnvoyProxy


arkodg:
*Description*:
>Describe the issue.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
ptal @envoyproxy/gateway-maintainers 
this will make metrics consistent with control plane

arkodg:
im a +1 on this, and prefer if we add a `disable` knob in case someone wants to disable it when enabling otel

zirain:
I forget why it's like this, but that's what I thought, big +1.

PR #2015, opened 2023-10-19
Enable proxy prometheus by default

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/1998


api/v1alpha1/envoyproxy_types.go
```diff
@@ -84,7 +84,7 @@ type ProxyTelemetry struct {
 	Tracing *ProxyTracing `json:"tracing,omitempty"`
 
 	// Metrics defines metrics configuration for managed proxies.
-	Metrics *ProxyMetrics `json:"metrics,omitempty"`
+	Metrics ProxyMetrics `json:"metrics,omitempty"`
```
arkodg, line 87: should be a pointer, since this setting is optional

api/v1alpha1/metric_types.go
```diff
@@ -7,7 +7,7 @@ package v1alpha1
 
 type ProxyMetrics struct {
 	// Prometheus defines the configuration for Admin endpoint `/stats/prometheus`.
-	Prometheus *PrometheusProvider `json:"prometheus,omitempty"`
+	Prometheus PrometheusProvider `json:"prometheus,omitempty"`
```
arkodg, line 10: should be a pointer, since this setting is optional

api/v1alpha1/metric_types.go
```diff
@@ -72,4 +72,6 @@ type OpenTelemetrySink struct {
 }
 
 type PrometheusProvider struct {
+	// Disabled disables the Prometheus endpoint.
+	Disabled bool `json:"disabled,omitempty"`
```
arkodg, line 76: should be `disable` (in present tense)
e.g. https://github.com/envoyproxy/gateway/blob/f203fe51dbcc2b6da3d3d5ea8a1715aad0eb1a1e/api/v1alpha1/accesslogging_types.go#L10

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -110,12 +122,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egv1a1.Kube
 		}
 	}
 
-	var proxyMetrics *egv1a1.ProxyMetrics
-	if infra.Config != nil {
-		proxyMetrics = infra.Config.Spec.Telemetry.Metrics
-	}
-
-	if proxyMetrics != nil && proxyMetrics.Prometheus != nil {
+	if enablePrometheus(infra) {
```
arkodg, line 125: shouldnt this port always exist, since its the readiness port as well ?

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -125,6 +132,10 @@ func expectedProxyContainers(infra *ir.ProxyInfra, deploymentConfig *egv1a1.Kube
 
 	var bootstrapConfigurations string
 
+	var proxyMetrics *egv1a1.ProxyMetrics
+	if infra.Config != nil {
```
arkodg, line 136: probably want to check for `nil` at each level - ` && infra.Config.Spec.Telemetry != nil && infra.Config.Spec.Telemetry.Metrics != nil `

arkodg:
thanks @zirain ! added some minor comments,
can you also update https://gateway.envoyproxy.io/latest/user/proxy-observability/ & https://gateway.envoyproxy.io/latest/user/grafana-integration/ to reflect this change (guessing we just need to delete existing lines of text)

internal/infrastructure/kubernetes/proxy/resource.go
zirain: it's checked inside `GetRenderedBootstrapConfig`.

internal/infrastructure/kubernetes/proxy/resource.go
zirain: then we need to rename container name.
Keep it the same as before, will do it later.

internal/infrastructure/kubernetes/proxy/resource.go
arkodg: looks like `Telemetry` is not a pointer https://github.com/envoyproxy/gateway/blob/23c91a6f903dc101b6bc362ac6c3432aec14b316/api/v1alpha1/envoyproxy_types.go#L45
since its optional, shouldnt it be a pointer

internal/infrastructure/kubernetes/proxy/resource.go
zirain: and think about https://github.com/envoyproxy/gateway/pull/1787

internal/infrastructure/kubernetes/proxy/resource.go
zirain: updated, hope I don't miss anything.

arkodg (dismissed):
LGTM thanks !

AliceProxy (approved):
lgtm, I like prometheus enabled by default

zirain:
@arkodg @AliceProxy fix the conflict.

arkodg:
@zirain can you run make generate again

arkodg (approved):

Issue #567, opened 2022-10-14
Enabling CORS for a HTTPRoute 


Amila-Rukshan:
*Description*:

I see two ways to enable CORS in for a HTTPRoute either via a targetRef from a Policy resource or an extensionRef from a custom route filter. 

Will envoy-gateway support another CR like below for this purpose?

```
apiVersion: gateway.envoy.io/v1alpha1
kind: CORSPolicy
metadata:
 name: cors-policy-1
Spec:
  default:
    corsConfigurationEnabled: true
    accessControlAllowOrigins:
     - '*'
    accessControlAllowCredentials: true
    accessControlAllowHeaders:
     - authorization
     - Access-Control-Allow-Origin
     - Content-Type
     - accept
     - origin
    accessControlAllowMethods:
     - GET
     - PUT
     - POST
     - DELETE
     - PATCH
     - OPTIONS
  targetRef:
    kind: HTTPRoute / Gateway
    name: http-route-1 / my-gateway
``` 

with this setting, it should be able to configure [cors filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/cors_filter) in HTTPRoute or Gateway levels.

Any thoughts?

Thanks!

arkodg:
Hi @Amila-Rukshan my earlier comment was incorrect and the above GEP only drop extension support within `matches` disallowing custom matches, but still supports extensions under `filters` . This means there are still 2 options of either using PolicyAttachment or Filter Extensions. There is a detailed discussion here https://github.com/envoyproxy/gateway/issues/675, hoping the outcome will help drive this discussion

youngnick:
Sorry @arkodg, I missed this one earlier. I think that a Policy resource is going to be the best fit for now, that is enough implementation to be able to suggest that something like this be implemented upstream in the HTTPRoute object (which I think is not unreasonable, since it's a common request). Its support level will depend on the data plane support, but I can see it ending up as Extended at the very least.

arkodg:
thanks @youngnick
hey @Amila-Rukshan, it would be great if you would be able to contribute to the design and implementation for CORS, it could be a stretch goal for v0.3.0 / Jan 2023

danehans:
> we are still safe since we are going to use a policy for it

Policy will depend on the outcome of #675.

sunjayBhatia:
What Gateway API extension mechanism works best for this feature may come down to what UX is allowed

[...]

PR #2016, opened 2023-10-19
feat: CORS ir and xds translation

zhaohuabing:
This PR does the following:

* Adds ir and xds translation for CORS.
The CORS will be part of the SecurityPolicy API. The API and translation to ir will be included in a subsequent PR after the [SecurityPolicy](https://github.com/envoyproxy/gateway/pull/1957) PR has been completed.

* Sorts the HTTP filter in a rational order after patching all filters to the HCM filter chain, so each individual filter translator doesn't have to care about its own order anymore.

Relates to https://github.com/envoyproxy/gateway/issues/567


internal/ir/xds.go
```diff
@@ -309,6 +313,42 @@ type JwtRequestAuthentication struct {
...
+	// AllowHeaders defines the headers that are allowed to be sent with requests.
+	AllowHeaders []string `json:"allowHeaders,omitempty" yaml:"allowHeaders,omitempty"`
+	// ExposeHeaders defines the headers that can be exposed in the responses.
+	ExposeHeaders []string `json:"exposeHeaders,omitempty" yaml:"exposeHeaders,omitempty"`
+	// MaxAge defines how long the results of a preflight request can be cached.
+	MaxAge *metav1.Duration `json:"maxAge,omitempty" yaml:"maxAge,omitempty"`
+	// AllowCredentials defines whether the resource allows credentials.
+	// Defaults to false.
+	AllowCredentials bool `json:"allowCredentials,omitempty" yaml:"allowCredentials,omitempty"`
+	// AllowPrivateNetwork defines whether allow whose target server’s IP address
```
arkodg, line 333: dont think this is common use case, can we disable it for now ? and add when we get user requests for it ?

internal/ir/xds.go
```diff
@@ -309,6 +313,42 @@ type JwtRequestAuthentication struct {
...
+	// is more private than that from which the request initiator was fetched.
+	// Defaults to false.
+	AllowPrivateNetworkAccess bool `json:"allowPrivateNetwork,omitempty" yaml:"allowPrivateNetwork,omitempty"`
+}
+
+func (c *CorsPolicy) Validate() error {
+	var errs error
+
```
arkodg, line 341: imo all fields within `cors` are optional and these zero length checks are not required

internal/xds/translator/cors.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if !listenerContainsCorsPolicy(irListener) {
+		return nil
+	}
+
+	// Return early if filter already exists.
+	for _, httpFilter := range mgr.HttpFilters {
+		if httpFilter.Name == wellknown.CORS {
+			// This should not happen since this is the only place where the cors
```
arkodg, line 43: why is this an error ?
multiple IR Listeners could map to the same XDS listeners so if we find an existing CORS filters, we should just return early with `nil`

internal/xds/translator/cors.go
```diff
@@ -0,0 +1,183 @@
...
+			return fmt.Errorf("cors filter already exists in hcm: %+v", mgr)
+		}
+	}
+
+	corsFilter, err := buildHCMCorsFilter()
+	if err != nil {
+		return err
+	}
+
+	// Ensure the cors filter is the first one in the filter chain.
```
arkodg, line 54: if this is the case, we should add a comment in listener.go, saying that this patch func should be not be moved from the end, because it ensures the cors policy is added first

internal/xds/translator/cors.go
```diff
@@ -0,0 +1,183 @@
...
+		exposeHeaders             string
+		maxAge                    string
+		allowCredentials          *wrappers.BoolValue
+		allowPrivateNetworkAccess *wrappers.BoolValue
+	)
+
+	//nolint:gocritic
+	for _, origin := range irRoute.CorsPolicy.AllowOrigins {
+		if origin.Exact != nil {
+			allowOrigins = append(allowOrigins, &matcherv3.StringMatcher{
```
arkodg, line 126: can we reuse the existing stringMatcher function here ?https://github.com/envoyproxy/gateway/blob/f203fe51dbcc2b6da3d3d5ea8a1715aad0eb1a1e/internal/xds/translator/route.go#L152

arkodg:
thanks for picking this one up @zhaohuabing, added some minor comments

internal/ir/xds.go
zhaohuabing: Agree we can disable this for now till get explicit user requests.

internal/ir/xds.go
zhaohuabing: In theory, we can, but it's meaningless if you set a CorsPolicy but leave the domains or methods empty.

internal/xds/translator/cors.go
zhaohuabing: Refactored the code and sorted the HTTP filters after patching.

internal/ir/xds.go
zhaohuabing: Rethinking it, it can be useful when the backend service allows CORS and we want to explicitly disable it at the Gateway.  We should make them optional.

arkodg: I agree with you reg domain, we should really consider making it non option in the API

internal/ir/xds.go
```diff
@@ -309,6 +313,26 @@ type JwtRequestAuthentication struct {
...
+	AllowMethods []string `json:"allowMethods,omitempty" yaml:"allowMethods,omitempty"`
+	// AllowHeaders defines the headers that are allowed to be sent with requests.
+	AllowHeaders []string `json:"allowHeaders,omitempty" yaml:"allowHeaders,omitempty"`
+	// ExposeHeaders defines the headers that can be exposed in the responses.
+	ExposeHeaders []string `json:"exposeHeaders,omitempty" yaml:"exposeHeaders,omitempty"`
+	// MaxAge defines how long the results of a preflight request can be cached.
+	MaxAge *metav1.Duration `json:"maxAge,omitempty" yaml:"maxAge,omitempty"`
+	// AllowPrivateNetwork defines whether allow whose target server’s IP address
+	// is more private than that from which the request initiator was fetched.
+	// Defaults to false.
```
arkodg, line 332: can we rm this field, if its not being used ?

internal/ir/xds.go
```diff
@@ -45,6 +45,8 @@ var (
 	ErrAddHeaderDuplicate            = errors.New("header modifier filter attempts to add the same header more than once (case insensitive)")
 	ErrRemoveHeaderDuplicate         = errors.New("header modifier filter attempts to remove the same header more than once (case insensitive)")
 	ErrRequestAuthenRequiresJwt      = errors.New("jwt field is required when request authentication is set")
+	ErrCorsAllowOriginsEmpty         = errors.New("field AllowOrigins must be specified with at least a single origin entry")
```
arkodg, line 48: can we rm these since they are not being used ?

internal/xds/translator/httpfilters.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Add the cors filter, if needed
+	if err := patchHCMWithCorsFilter(mgr, irListener); err != nil {
+		return err
+	}
+
+	// Add the router filter
+	mgr.HttpFilters = append(mgr.HttpFilters, xdsfilters.HTTPRouter)
+
+	// Sort the filters in the correct order.
```
arkodg, line 101: I like this idea, because it may become hard for a dev to also update `newOrderedHTTPFilter`
just a comment, dont think we need to do much here, more comments in `newOrderedHTTPFilter` would be great

arkodg (approved):
Lgtm thanks !

zirain (approved):

Issue #1841, opened 2023-08-28
Allow to set Service LoadBalancer IP


mazzy89:
*Description*:
At the current state it is not possible to set to the created Service LoadBalancer an user static IP address. The property is not exposed. 

There are cases in various cloud providers like GCP where it is possibile to create a static IP address and reference it around.

There is a TODO in the EnvoyProxy ServiceSpec.

mazzy89:
I don't think it should be hard to set so I can wire a PR.

arkodg:
it looks like `loadBalancerIP` is deprecated https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer
looking at the GKE docs https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters can you achieve the same by using a `networking.gke.io/internal-load-balancer-subnet` annotation to specify a /32 subnet (`10.11.12.13/32`)?

arkodg:
yes you could always use GKE gateway but you'd be missing out on the extensive features in Envoy Gateway :) 

even Envoy Gateway supports `Gateway.Spec.Addresses` but it doesnt affect the Cloud Loabalancer

curious why `networking.gke.io/internal-load-balancer-subnet` won't work, my guess is, if you submit subnet to be 1 IP (/32 range for IPv4) , then the service must be assigned that IP

mazzy89:
As said above I have a public IP v4. That annotation is internal subnet. Doesn't not relate at all with public IP addresses.

arkodg:
ah yes, thanks for the clarification

mazzy89:
Then I think the way to go would be as the approach proposed done by Google https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-gateways#use_a_named_ip_address where the IP address would be referenced as a named IP address. However this would be a GCP only approach. what do you think?

arkodg:
I dont think above approach will work, the `Gateway` defined is tied to a GKE GatewayClass `gke-l7-rilb` which is tied to a GKE controller which is probably doing some magic under the hood to stitch together a path from the Google LB to the Loadbalancer Gateway service

mazzy89:
[...]

PR #2017, opened 2023-10-19
Support configuring loadBalancerIP in envoy svc

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/1841


api/v1alpha1/shared_types.go
```diff
@@ -191,6 +191,13 @@ type KubernetesServiceSpec struct {
 	// +optional
 	AllocateLoadBalancerNodePorts *bool `json:"allocateLoadBalancerNodePorts,omitempty"`
 
+	// LoadBalancerIP defines the IP Address of the underlying load balancer service. This field
+	// may be ignored if the load balancer provider does not support this feature.
+	// This field has been Deprecated, but it is still used for set the IP Address in some cloud
+	// providers such as GCP.
```
zirain, line 197: it's deprecated, maybe use annotations to support this?

arkodg: Not possible for GCP, the linked issue has all the details:)

zirain: thanks for clearify.

zirain: what will happen if someone config this by mistake on non-GCP?

arkodg: https://github.com/envoyproxy/gateway/issues/1841#issuecomment-1699472763
has the details, looks like all still support it except AWS and I'm unsure what happens when it is set, any idea @LanceEa ?

LanceEa: @zirain - see discussion here: https://github.com/envoyproxy/gateway/issues/1841#issuecomment-1699472763

tl;dr; using term "deprecated" is probably not correct but rather discouraged until k8s community/cloud providers have completely moved away from it.

zirain: Just want to make sure, it won't break anything.

arkodg: Thanks @LanceEa can you please review the API doc string in this PR

LanceEa: My guess is either it is ignored or like GCP the admission controller would reject the fields it doesn't support.

api/v1alpha1/shared_types.go
```diff
@@ -191,6 +191,13 @@ type KubernetesServiceSpec struct {
 	// +optional
 	AllocateLoadBalancerNodePorts *bool `json:"allocateLoadBalancerNodePorts,omitempty"`
 
+	// LoadBalancerIP defines the IP Address of the underlying load balancer service. This field
+	// may be ignored if the load balancer provider does not support this feature.
+	// This field has been Deprecated, but it is still used for set the IP Address in some cloud
```
LanceEa, line 196: nit: `set` --> `setting`

Might also be good to clarify that it is deprecated in k8s so maybe `This field has been Deprecated in K8s`. 

If you wanted to get fancy you could add a link but since this gets bundled into the CRD definition, this seems sufficient. Also, I think we have good history in the PR that users can search if have questions.

api/v1alpha1/shared_types.go
zirain: What I'm thinking is should EG auto-detect cloud provider, and reject/ignore this field if it's non-GCP?

api/v1alpha1/shared_types.go
arkodg: thanks, addressed

api/v1alpha1/shared_types.go
LanceEa: @arkodg - I agree. At that point I think we would be over engineering  it and would end up with more pain maintaining it. Instead I think leaning on documentation is a better approach.

zirain (dismissed):

api/v1alpha1/validation/validate.go
```diff
@@ -95,6 +96,15 @@ func validateService(spec *egv1a1.EnvoyProxySpec) []error {
 				errs = append(errs, fmt.Errorf("allocateLoadBalancerNodePorts can only be set for %v type", egv1a1.ServiceTypeLoadBalancer))
 			}
 		}
+		if serviceType, serviceLoadBalancerIP := spec.Provider.Kubernetes.EnvoyService.Type, spec.Provider.Kubernetes.EnvoyService.LoadBalancerIP; serviceType != nil && serviceLoadBalancerIP != nil {
+			if *serviceType != egv1a1.ServiceTypeLoadBalancer {
+				errs = append(errs, fmt.Errorf("loadBalancerIP can only be set for %v type", egv1a1.ServiceTypeLoadBalancer))
+			}
+
+			if net.ParseIP(*serviceLoadBalancerIP) == nil {
```
LanceEa, line 104: Not a blocker for me but ParseIP will also parse an IPV6 address successfully so it could be nice to give the user earlier feedback.

But realistically most users setting this field have read the Kubernetes docs or can at least search them easily.

LanceEa (approved):

api/v1alpha1/validation/validate.go
arkodg: thanks for flagging this, adding a check to only allow ipv4

AliceProxy (approved):

PR #2019, opened 2023-10-19
Add LoadBalancer IR to HTTPRoute

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1105


internal/ir/xds.go
```diff
@@ -952,3 +961,58 @@ type TCPKeepalive struct {
 	// Defaults to `75s`.
 	Interval *uint32 `json:"interval,omitempty" yaml:"interval,omitempty"`
 }
+
+// LoadBalancer defines the load balancer settings.
+// +k8s:deepcopy-gen=true
+type LoadBalancer struct {
+	// RoundRobin load balacning policy
```
AliceProxy, line 968: not a blocker right now since this is just IR, but do we want to add a `type` field here?

internal/xds/translator/cluster.go
```diff
@@ -61,10 +60,24 @@ func buildXdsCluster(clusterName string, tSocket *corev3.TransportSocket, protoc
...
+	if args.loadBalancer == nil {
+		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
+	} else if args.loadBalancer.LeastRequest != nil {
+		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
+	} else if args.loadBalancer.RoundRobin != nil {
+		cluster.LbPolicy = clusterv3.Cluster_ROUND_ROBIN
+	} else if args.loadBalancer.Random != nil {
+		cluster.LbPolicy = clusterv3.Cluster_RANDOM
+	} else if args.loadBalancer.ConsistentHash != nil {
+		cluster.LbPolicy = clusterv3.Cluster_MAGLEV
```
AliceProxy, line 78: Emissary supports maglev and ring hash to allow for session affinity based on request headers, auto-generated cookies, and source IP if we want to design around feature parity. The IR can always change so this just just a consideration for the public API for this.

AliceProxy (approved):

internal/ir/xds.go
arkodg: ah the `type` will def be in the user facing API, in this layer using the struct ptr instantiation to figure out the type

internal/xds/translator/cluster.go
arkodg: imo is there any use case to support ring hash over maglev ? afaik maglev is superior to ring hash when it comes down to consistent hashing

arkodg: ive only added only hash input - source ip, we should be able to add other such as cookie etc in the future

internal/ir/xds.go
AliceProxy: That totally works, just wasn't sure if we wanted to make it explicitly clear in the ir that only one of these can be provided for a route, etc. for newer contributors that might be unaware. I kinda like using types in the ir sometimes.

arkodg: I added a `Validate()` method to ensure only one can be set and added test cases in this PR :)

internal/xds/translator/cluster.go
AliceProxy: I think that's totally fine unless someone has a reason to request support for ring hash, then we can consider it at that point.

PR #2021, opened 2023-10-20
api: proposal proxy ports

zirain:
this's an alternative for https://github.com/envoyproxy/gateway/pull/1787
xref: https://github.com/envoyproxy/gateway/issues/1703
follow up of: https://github.com/envoyproxy/gateway/pull/2015#discussion_r1366079885


qicz (approved):

arkodg:
@zirain can we add this post v0.6 ?

zirain:
> @zirain can we add this post v0.6 ?

sure, it's not urgent.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -73,6 +79,20 @@ type EnvoyProxySpec struct {
...
+	// AdminPort defines the port for the Proxy admin interface.
+	// If unspecified, defaults to 19000.
+	// +optional
+	AdminPort *int32 `json:"adminPort,omitempty"`
+
+	// ExtraPort defines the extra port for the Proxy extra interface like
+	// HealthCheck, Prometheus metrics.
+	// If unspecified, defaults to 19001.
+	// +optional
+	ExtraPort *int32 `json:"extraPort,omitempty"`
```
arkodg, line 93: `extraPort` is not a great name imo
should we split this up into `prometheusPort` & `healthCheckPort` ?

zirain: nop, it will/should shared one port in most cases.

arkodg: can we then come up with a better name

zirain: do you have any good idea?

PR #2050, opened 2023-10-23
feat: add CEL validation for EnvoyProxy telemetry

shawnh2:
**What type of PR is this?**

add missing unit test cases

some unit test cases are missing according to https://app.codecov.io/gh/envoyproxy/gateway/blob/main/api%2Fv1alpha1%2Fvalidation%2Fenvoyproxy_validate.go

Fixes #


zirain (approved):

zirain (changes requested):
rethink about this, let's move to https://github.com/envoyproxy/gateway/pull/2059.

arkodg:
@shawnh2 we've moved to CEL based validations, the newer tests live in https://github.com/envoyproxy/gateway/blob/main/test/cel-validation/envoyproxy_test.go, can we move the new test cases there instead and also make sure a CEL validation exists for it
cc @zirain

shawnh2:
add missing CEL validation and test for `ProxyTelemetry`, PTAL @arkodg @zirain

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #2034, opened 2023-10-20
Improve infra port name collisions when Merge Gateways is enabled


arkodg:
*Description*:
>Describe the issue.

Relates to https://github.com/envoyproxy/gateway/pull/1938#discussion_r1367334125 & https://github.com/envoyproxy/gateway/pull/1938#discussion_r1367339721

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
cc @cnvergence

PR #2052, opened 2023-10-23
fix: Improve infra port name collisions

cnvergence:
**What type of PR is this?**

Fixes:
- https://github.com/envoyproxy/gateway/issues/2034


internal/provider/utils/utils.go
```diff
@@ -41,17 +41,24 @@ func HashString(str string) string {
 	return strings.ToLower(fmt.Sprintf("%x", h.Sum(nil)))
 }
 
-// ExpectedContainerPortHashedName returns expected container port name with max length of 15 characters.
+// ExpectedContainerPortName returns expected container port name with max length of 15 characters.
 // If mergedGateways is enabled or listener port name is larger than 15 characters it will return partially hashed name.
 // Listeners on merged gateways have a infraIR port name {GatewayNamespace}/{GatewayName}/{ListenerName}.
-func ExpectedContainerPortHashedName(name string) string {
+func ExpectedContainerPortName(name string) string {
```
arkodg, line 47: this logic is looking fairly complicated - the output is different based on length as well as special char `/` tokens
can this look more like
```
func GetHashedContainerPortName(nsName string) string {
	hashedName := HashString(nsName)
	// replace `/` with `-` to create a valid K8s resource name
	resourceName := strings.ReplaceAll(nsName, "/", "-")
	if len(resourceName) > 7 {
		return fmt.Sprintf("%s-%s", resourceName[0:7], hashedName[0:8])
	}
	return fmt.Sprintf("%s-%s", resourceName, hashedName[0:8])
}
```

arkodg (dismissed):
LGTM thanks !

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -50,12 +50,22 @@ var (
...
 func ExpectedResourceHashedName(name string) string {
-	hashedName := providerutils.GetHashedName(name)
+	hashedName := providerutils.GetHashedName(name, 48)
 	return fmt.Sprintf("%s-%s", config.EnvoyPrefix, hashedName)
 }
 
+// ExpectedContainerPortName returns expected container port name with max length of 15 characters.
+// If mergedGateways is enabled or listener port name is larger than 15 characters it will return partially hashed name up to the 7 characters.
+// Listeners on merged gateways have a infraIR port name {GatewayNamespace}/{GatewayName}/{ListenerName}.
+func ExpectedContainerPortName(infraPortName string) string {
```
arkodg, line 62: we are always hashing for `deployment` , `service` and `service Port Name`, why not always hash for `container port name` as well (instead of this if else)

cnvergence: I insisted on having this as a listener name, but on the other hand, this is still only 15 characters long.
Let's do a hashed name here as well

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #1964, opened 2023-10-13
Add CEL validation tests for xPolicy resources


arkodg:
*Description*:
We've recently started add CEL validations in our APIs e.g. https://github.com/envoyproxy/gateway/pull/1961
Would be great to add validation tests so its easier to understand what works and what doesnt
Here's an example of how upstream does it https://github.com/kubernetes-sigs/gateway-api/blob/2b9aeb0c7fa3d558e42074486c230406e7ac80ce/pkg/test/cel/grpcroute_test.go#L416

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
we should add valiation for other EG CRDs.

PR #2059, opened 2023-10-24
init CEL validation

zirain:
xref: https://github.com/envoyproxy/gateway/issues/1964

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -81,23 +81,29 @@ jobs:
     - uses: actions/checkout@v4
     - uses: ./tools/github-actions/setup-deps
 
-    - name: Download EG Binaries
-      uses: actions/download-artifact@v3
-      with:
-        name: envoy-gateway
-        path: bin/
-
-    - name: Give Privileges To EG Binaries
-      run: |
-        chmod +x bin/linux/amd64/envoy-gateway
-        chmod +x bin/linux/arm64/envoy-gateway
-
     # conformance
     - name: Run Standard Conformance Tests
       env:
         KIND_NODE_TAG: ${{ matrix.version }}
         IMAGE_PULL_POLICY: IfNotPresent
       run: make conformance
+
+  crd-validation:
+    runs-on: ubuntu-latest
+    needs: [ build ]
+    strategy:
+      matrix:
+        version: [ v1.26.6, v1.27.3, v1.28.0 ]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./tools/github-actions/setup-deps
+
+      # CRD Validation Tests
+      - name: Run Validation Tests
+        env:
+          KIND_NODE_TAG: ${{ matrix.version }}
+          IMAGE_PULL_POLICY: IfNotPresent
+        run: make validation
 
   e2e-test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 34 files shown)


api/v1alpha1/envoyproxy_types.go
```diff
@@ -97,6 +97,7 @@ type EnvoyProxyProvider struct {
 	// optional auxiliary control planes. Supported types are "Kubernetes".
 	//
 	// +unionDiscriminator
+	// +kubebuilder:validation:XValidation:message="type must be one of ['Kubernetes']",rule="self == 'Kubernetes'"
```
arkodg, line 100: is this needed ?
it already exists https://github.com/envoyproxy/gateway/blob/bdb0191dec04788a1b8e47b5a66f0cc99ef29afe/api/v1alpha1/shared_types.go#L36

tools/make/kube.mk
```diff
@@ -210,3 +210,12 @@ generate-egctl-releases: ## Generate egctl releases
 	curl -sSL https://github.com/envoyproxy/gateway/releases/download/latest/egctl_latest_darwin_arm64.tar.gz -o $(OUTPUT_DIR)/egctl_$(TAG)_darwin_arm64.tar.gz
 	curl -sSL https://github.com/envoyproxy/gateway/releases/download/latest/egctl_latest_linux_amd64.tar.gz -o $(OUTPUT_DIR)/egctl_$(TAG)_linux_amd64.tar.gz
 	curl -sSL https://github.com/envoyproxy/gateway/releases/download/latest/egctl_latest_linux_arm64.tar.gz -o $(OUTPUT_DIR)/egctl_$(TAG)_linux_arm64.tar.gz
+
+.PHONY: validation
```
arkodg, line 214: imo we should reuse `make test` or `make kube-test` for this and not create another target

.github/workflows/build_and_test.yaml
```diff
@@ -2,172 +2,167 @@ name: Build and Test
 on:
   push:
     branches:
-    - "main"
-    - "release/v*"
+      - "main"
```
arkodg, line 5: can we undo all changes from this file  ?
the existing indentation seems consistent with our code base

test/validation/main_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+//go:build validation
```
arkodg, line 6: can we call this `cel-validation` ?

.github/workflows/build_and_test.yaml
zirain: that won't pass the yaml lint.

test/validation/main_test.go
zirain: done

.github/workflows/build_and_test.yaml
arkodg: Not seeing any issues when Im running yamllint

zirain: you didn't change it?

zirain: updated

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):
LGTM, thanks.

Issue #2006, opened 2023-10-19
Remove RateLimitFilter & AuthenticaionFilter


arkodg:
*Description*:

Remove these filters and add these features as fields in `BackendTrafficPolicy` & `SecurityPolicy`

This will be a breaking change that needs to be highlighted in the release notes

zirain:
will `BackendTrafficPolicy` & `SecurityPolicy` ready in v0.6?

arkodg:
thats the plan @zirain, we are trying :)

Issue #1821, opened 2023-08-23
proposal: BackendTrafficPolicy


AliceProxy:
## Relates to
https://github.com/envoyproxy/gateway/issues/1492
https://github.com/envoyproxy/gateway/issues/1821
https://github.com/envoyproxy/gateway/issues/1845

## What is this?
`BackendTrafficPolicy` is a proposal for a new policy attachment resource that can be applied to `Gateways` and `xRoute` resources. It is meant to supplement a lot of config that Gateway API currently lacks while also providing a way to specify global defaults. It is focused on configuring the behaviour of traffic between EnvoyProxy and the backend service.

## What if Gateway API implements a GEP that delivers the same functionality as some of the fields of this resource?
If Gateway improves any of their existing resources to deliver functionality that meets all of the needs of any of the below config,
then we will deprecate the field in this resource and use the Gateway API config instead. Ideally, a lot of this config can be
upstreamed into Gateway API in some form eventually since this resource only exists to solve areas where Gateway API is lacking for the needs of Envoy Gateway.

## What about traffic between EnvoyProxy and the downstream clients?
I will be opening up a sister issue proposing an `InboundTrafficPolicy` (or similar) resource here shortly that will focus on this use case.

## What happens when I attach it to a `Gateway` resource?
If you apply this resource to a `Gateway` resource, any config that applies to the "listener level" will be configured there.
A lot of the config is more "route level," and so for those fields, they may not do anything if all you have is a Gateway, but they will be used as the default configuration for all children `xRoute` resources of the `Gateway` this resource is attached to. 

## What happens when I attach it to an `xRoute` resource (`HTTPRoute`, `GRPCRoute`, etc.)
If you apply this config to an `xRoute` resource, it will apply route-level config to this specific route. 
If you have an `BackendTrafficPolicy` attached to both a `Gateway` and an `xRoute` resource, then config from the one 
attached to the `xRoute` will win in a conflict. The "merging" in this case, will be primitive. If you have any sort of 
config for a top-level field such as `circuitBreakers`, then the whole field will be replaced rather than comparing all of the subfields
and merging/patching them.

## Can you use multiple `BackendTrafficPolicy` resources at the same time?
You may use multiple `BackendTrafficPolicies` that target different resources, but you may not attach multiple `BackendTrafficPolicies`
[...]

PR #2060, opened 2023-10-24
Add rateLimit to BackendTrafficPolicy

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1821 & https://github.com/envoyproxy/gateway/issues/2006


zhaohuabing (approved):

zhaohuabing:
What happens if both a "RateLimitFilter" and the "RateLimit" of a "BackendTrafficPolicy" attach the same route? Will we deprecate "RateLimitFilter"?

arkodg:
> What happens if both a "RateLimitFilter" and the "RateLimit" of a "BackendTrafficPolicy" attach the same route? Will we deprecate "RateLimitFilter"?

@zhaohuabing the plan is to remove RateLimitFilter (& AuthenticatinFilter) before we release v0.6

zirain (dismissed):

zirain (dismissed):

zirain (approved):

Issue #1966, opened 2023-10-14
Global downstream connection limits.


arkodg:
Add a overload manager configuration to limit the total number of downstream connections
https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager#limiting-active-connections

tmsnan:
It appears that the functionality is incomplete and unsuitable for production use. Should we enable it?
[](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/resource_monitors/downstream_connections/v3/downstream_connections.proto#extensions-resource-monitors-downstream-connections-v3-downstreamconnectionsconfig)

arkodg:
@tmsnan im fine waiting, but atm we dont have any way of enabling overload manager

tmsnan:
@arkodg  If possible, I could first add other overload manager features that are already fully supported.

arkodg:
@tmsnan sure, imo the others like heap size based will require an API addition, putting the burden of getting it right on the user .
Was trying to use this GH issue to enable sensible defaults

tmsnan:
@arkodg Maybe we can refer to Google VRP edge server configuration.
`overload_manager:
  refresh_interval: 0.25s
  resource_monitors:
  - name: "envoy.resource_monitors.fixed_heap"
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.resource_monitors.fixed_heap.v3.FixedHeapConfig
      # TODO: Tune for your system.
      max_heap_size_bytes: 2147483648  # 2 GiB
  actions:
  - name: "envoy.overload_actions.shrink_heap"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.95
  - name: "envoy.overload_actions.stop_accepting_requests"
    triggers:
    - name: "envoy.resource_monitors.fixed_heap"
      threshold:
        value: 0.98`
https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#best-practices-edge

shahar-h:
Please assign me

PR #2062, opened 2023-10-25
feat: support overload manager

tmsnan:
**What type of PR is this?**
support overload manager

Related #1966 #1048


tmsnan:
How to set the approximate fixed heap value of the system through a script needs to be considered further.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -73,6 +73,15 @@ type EnvoyProxySpec struct {
...
 	MergeGateways *bool `json:"mergeGateways,omitempty"`
+
+	// OverloadManager defines the configuration for the Overload Manager in envoy.
+	// Visit https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager#config-overload-manager
+	// to learn more about the syntax.
+	// Overload manager is disabled by default.
+	// After the feature is enabled, following two overload actions are configured to Envoy:
+	// 1.Shrink heap action is executed when 95% of the maximum heap size is reached.
+	// 2.Envoy will stop accepting requests when 98% of the maximum heap size is reached.
+	OverloadManager *OverloadManager `json:"overloadManager,omitempty"`
```
arkodg, line 84: imo we shouldnt expose envoyisms such as `overloadManager` directly, but do it in a way that is relevant to the EG user

api/v1alpha1/envoyproxy_types.go
```diff
@@ -213,6 +222,14 @@ type EnvoyProxyList struct {
 	Items           []EnvoyProxy `json:"items"`
 }
 
+// OverloadManager defines the configuration for the Overload Manager in envoy.
+type OverloadManager struct {
+	// The appropriate number of bytes can be different from system to system.
```
arkodg, line 227: we shouldnt add this API field if we cannot suggest a prescriptive way for a user to come up with a value to insert

api/v1alpha1/envoyproxy_types.go
tmsnan: You're right，the `MaxHeapSizeBytes` value should be exposed directly to enable the overload manager. I'm still thinking about how to set this value specifically.

api/v1alpha1/envoyproxy_types.go
guydc: The google VRP example uses a [2GiB max heap size](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge) when [executing a docker container](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/google_vrp#execution-environment) with 3GB of memory.

Proposal:
- By default, set MaxHeapSize to 70% of the envoy container's memory limit (which is known to EG, since it's generating the PodTemplate). If the limit is not set - don't configure anything. 
- Allow users to override this setting - configuring a custom percent. 

@arkodg , @tmsnan  - WDYT?

arkodg: hey @guydc agree with the approach ! since EG creates the envoy proxy, we have the the exact memory specs for the envoy proxy fleet
 imo we'll need to gather some more data before we decide on the the max heap and threshold values

arkodg: chatted with @KBaichoo who advised that we could consider a tighter limit like 80% to optimize resources, thanks !

PR #2063, opened 2023-10-25
LoadBalancer in BackendTrafficPolicy

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/1105


api/v1alpha1/loadbalancer_types.go
```diff
@@ -0,0 +1,55 @@
...
+	// "Random",
+	// "RoundRobin",
+	//
+	// +unionDiscriminator
+	Type LoadBalancerType `json:"type"`
+	// ConsistentHash defines the configuration when the load balancer type is
+	// set to ConsistentHash
+	//
+	// +optional
+	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty"`
```
zirain, line 24: haha, this's a good place to try XValidation

arkodg: Yes

arkodg: Can we do it as a follow up ?

zirain: sure, it's not a block

zhaohuabing, line 24: What happens if the type is `ConsistentHash` and the `ConsistentHash` is nil?

arkodg: this is what @zirain is referring to in https://github.com/envoyproxy/gateway/pull/2063#discussion_r1372418714
we can add CEL Validation to proactively check this

zirain (approved):

api/v1alpha1/loadbalancer_types.go
zirain: can you open a ticket for this?

arkodg: raised https://github.com/envoyproxy/gateway/issues/2066

PR #2080, opened 2023-10-26
feat: add subscribed message metrics instrumentations

Xunzhuo:
**What type of PR is this?**

image


zirain:
I think you can add a simple e2e test for prometheus

```golang
package main

import (
	"fmt"
	"io"
	"log"
	"net/http"
	"strings"
	"time"

	"github.com/prometheus/common/expfmt"
)

func main() {
	_, body, err := HTTPGet(fmt.Sprintf("http://127.0.0.1:%d/stats/prometheus", 9901))
	if err != nil {
		fmt.Println("get metric error: %v", err)
		return
	}
	reader := strings.NewReader(body)
	_, err = (&expfmt.TextParser{}).TextToMetricFamilies(reader)
	if err != nil {
		fmt.Println("parse metric error: %v body: %s", err, body)
	}
}

func HTTPGet(url string) (code int, respBody string, err error) {
	log.Println("HTTP GET", url)
	client := &http.Client{Timeout: time.Minute}
	resp, err := client.Get(url)
	if err != nil {
		log.Println(err)
		return 0, "", err
	}
	defer resp.Body.Close()
	body, err := io.ReadAll(resp.Body)
	if err != nil {
		log.Println(err)
		return 0, "", err
	}
	respBody = string(body)
	code = resp.StatusCode
	return code, respBody, nil
}

```

Xunzhuo:
@zirain Can you help with it in a sparate PR?

internal/message/watchutil.go
```diff
@@ -20,20 +45,36 @@ type Update[K comparable, V any] watchable.Update[K, V]
...
 		}
 	}
 	for snapshot := range subscription {
+		messageDepth.With(meta.LabelValues()...).Record(float64(len(subscription)))
 		for _, update := range snapshot.Updates {
-			handle(Update[K, V](update))
+			startHandleTime := time.Now()
+			handle(Update[K, V](update), errChans)
+			messageSubscribedTotal.With(meta.LabelValues()...).Increment()
+			messageSubscribedDurationSeconds.With(meta.LabelValues()...).Record(time.Since(startHandleTime).Seconds())
```
zirain, line 77: make these lines into a function will make sure you will not forget update anyone of those metrics.

internal/message/watchutil.go
```diff
@@ -20,20 +45,36 @@ type Update[K comparable, V any] watchable.Update[K, V]
 // it handles the case where the watchable.Map already contains
 // entries before .Subscribe is called.
 func HandleSubscription[K comparable, V any](
+	meta Metadata,
 	subscription <-chan watchable.Snapshot[K, V],
-	handle func(Update[K, V]),
+	handle func(updateFunc Update[K, V], errChans chan error),
 ) {
+	errChans := make(chan error, 10)
```
zirain, line 52: any reason for use 10 for buffer size?

internal/message/watchutil.go
```diff
@@ -6,11 +6,36 @@
...
 
+var logger = logging.DefaultLogger(v1alpha1.LogLevelInfo).WithName("watchable")
+
+type Metadata struct {
+	Runner   string
+	Resource string
+}
+
+func (m Metadata) LabelValues() []metrics.LabelValue {
+	labels := []metrics.LabelValue{}
```
zirain, line 28:
```suggestion
	labels := make([]metrics.LabelValue, 0, 2)
```

arkodg:
@Xunzhuo are we following metric naming conventions similar to EnvoyProxy ?

internal/gatewayapi/runner/runner.go
```diff
@@ -49,8 +49,8 @@ func (r *Runner) Start(ctx context.Context) (err error) {
 }
 
 func (r *Runner) subscribeAndTranslate(ctx context.Context) {
-	message.HandleSubscription(r.ProviderResources.GatewayAPIResources.Subscribe(ctx),
-		func(update message.Update[string, *gatewayapi.Resources]) {
+	message.HandleSubscription(message.Metadata{Runner: string(v1alpha1.LogComponentGatewayAPIRunner), Resource: "gatewayapi-resources"}, r.ProviderResources.GatewayAPIResources.Subscribe(ctx),
```
arkodg, line 52: can we replace `resource` with `message` and called this type `provider-resources` in line with 
https://github.com/envoyproxy/gateway/blob/main/internal/message/types.go

internal/message/metrics.go
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package message
+
+import "github.com/envoyproxy/gateway/internal/metrics"
+
+var (
+	messageDepth = metrics.NewGauge("message_depth", "Current depth of message queue.")
```
arkodg, line 11: can we replace `message` with `watchable` here ?

arkodg:
hey @Xunzhuo, ive made some suggestions for now mainly around naming, prefer- `watchable_subscribed_duration_seconds{resource="provider-resources", runner="gateway-api"}`

would also like to replace `subscribed` with something else, but can't think of anything rn

internal/message/metrics.go
```diff
@@ -0,0 +1,22 @@
...
+
+	watchableSubscribedDurationSeconds = metrics.NewHistogram("watchable_subscribed_duration_seconds", "How long in seconds a subscribed watchable is handled.", []float64{0.001, 0.01, 0.1, 1, 5, 10})
+
+	watchableSubscribedTotal = metrics.NewCounter("watchable_subscribed_total", "Total number of subscribed watchable.")
+
+	watchableSubscribedErrorsTotal = metrics.NewCounter("watchable_subscribed_errors_total", "Total number of subscribed watchable errors.")
+
+	runnerLabel = metrics.NewLabel("runner")
+
+	resourceLabel = metrics.NewLabel("resource")
```
arkodg, line 21: instead of `resource` can we call it `message` ?

Xunzhuo: How about `output` same as log value?

Xunzhuo:
```go
	yamlXdsIR, _ := yaml.Marshal(&result.XdsIR)
	r.Logger.WithValues("output", "xds-ir").Info(string(yamlXdsIR))
	yamlInfraIR, _ := yaml.Marshal(&result.InfraIR)
	r.Logger.WithValues("output", "infra-ir").Info(string(yamlInfraIR))
```

arkodg:
thanks @Xunzhuo, hey @zirain can you ptal

internal/message/watchutil.go
Xunzhuo: We need some buffers to make sure the errChan will not be stuck when observing errors from snapshot.Updates. As for the size, I am not sure about it, a suggested value from chatgpt.

zirain: can you add some comments for it, or a `TODO: find a suitable value` for this.

zirain:
can udpate an image of the latest output?

internal/message/watchutil.go
Xunzhuo: Done

Xunzhuo:
@zirain  updated the pic in PR desc.

zirain (approved):

qicz (approved):

Issue #2067, opened 2023-10-25
chore: refactory string matcher


zhaohuabing:
*Description*:

The string match types are a bit messy. They're defined in multiple APIs but look similar. For most EG APIs(Ratelimit and CORS for now, more in the future), the match types are just used to represent the [StringMatcher](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/string.proto) XDS API. 

What I'd like to change:

* Create a common StringMatcher that can be used by Ratelimit, CORS, and other APIs that need plain string matching
* Change the name of StringMatcher in Stats to avoid naming conflicting

I would like to raise a PR to refactor it if it makes sense.

PR #2102, opened 2023-10-28
chore: refactor string match

zhaohuabing:
fix:  https://github.com/envoyproxy/gateway/issues/2067

* use the shared string match for stats match condition
* rename string match type in shared types, if no special requirement like 'Distinct' in Rate limit, all the other EG APIs should consider using this common string match type
* use the shared string match for rate limit - removed after discussion


api/v1alpha1/ratelimit_types.go
```diff
@@ -108,49 +108,25 @@ type SourceMatch struct {
...
+
+	// Match specifies how to match against the value of the header.
 	// Do not set this field when Type="Distinct", implying matching on any/all unique
 	// values within the header.
-	//
-	// +optional
-	// +kubebuilder:validation:MaxLength=1024
-	Value *string `json:"value,omitempty"`
+	// Only one of Distinct or Match can be set.
+	Match *StringMatch `json:"match,omitempty"`
```
zhaohuabing, line 127: @arkodg @zirain @Xunzhuo  I plan to refactor rate limit HeaderMatch to use the shared string match. If you agree this is the right way, I'll go ahead.

arkodg: we should keep header matches aligned to upstream as much as possible.
for this case we defined in this api because we introduced a new match type called `distinct` that doesnt exist upstream
https://github.com/kubernetes-sigs/gateway-api/blob/5fd8bec3bedc7c325bf2227da48d041a735b9b98/apis/v1beta1/httproute_types.go#L98
https://github.com/kubernetes-sigs/gateway-api/blob/5fd8bec3bedc7c325bf2227da48d041a735b9b98/apis/v1/httproute_types.go#L439

Also although sprinkling different match types affects project mantainence, it benefits end users because for each different feature we are adding a different set of kubebuilder annotations for the fields, so we providing proactive feedback on the type of matches that we support.

tldr, if the match types are identical, it makes sense to consolidate, else imo we should wait until they reach parity

zhaohuabing: IMO, the Distinct Match type only matters when it comes to the Rate Limit. So, perhaps we could consider renaming HeaderMatch to RateLimitHeaderMatch. Let's leave it as is for now and see how we want to deal with the string match when we have new APIs in the mix.

api/v1alpha1/envoyproxy_metric_types.go
```diff
@@ -60,18 +60,18 @@ type ProxyPrometheusProvider struct {
...
 	//
 	// +kubebuilder:validation:Enum=RegularExpression;Prefix;Suffix
-	Type  MatcherType `json:"type"`
-	Value string      `json:"value"`
+	Type  StatsMatchType `json:"type"`
+	Value string         `json:"value"`
 }
 
-type MatcherType string
+type StatsMatchType string
```
arkodg, line 71: needs enum kubebuilder annotations

api/v1alpha1/envoyproxy_metric_types.go
```diff
@@ -23,7 +23,7 @@ type ProxyMetrics struct {
 	// `cluster.<cluster_name>.membership_degraded`，reference  https://github.com/envoyproxy/envoy/issues/9856,
 	// https://github.com/envoyproxy/envoy/issues/14610
 	//
-	Matches []Match `json:"matches,omitempty"`
+	Matches []StatsMatch `json:"matches,omitempty"`
```
arkodg, line 26: replacing this with `sringsMatch`, should be backwards compatible from a YAML perspective, right ?

zhaohuabing: Yes, we can. We should align the string match type across APIs. Currently supports exact, suffix, prefix, and Regex. Others can be included if needed in the future.

api/v1alpha1/envoyproxy_metric_types.go
zhaohuabing: Don't need it anymore :-)

zirain (approved):

arkodg (approved):
LGTM thanks !

arkodg:
should we merge this post v0.6 ?

zhaohuabing:
> should we merge this post v0.6 ?
> 

I'm OK with that.

zhaohuabing:
@arkodg can we merge it now? v0.6 has been released.

Issue #2103, opened 2023-10-28
Panic: `patchRouteWithRateLimit` should check if xdsRouteAction is nil


Xunzhuo:
*Description*:

```
2023-10-28T10:33:31.018Z	INFO	infrastructure	runner/runner.go:59	received an update	{"runner": "infrastructure"}
2023-10-28T10:33:31.018Z	INFO	provider	status/status.go:122	received a status update	{"runner": "provider", "namespace": "default", "name": "http-ratelimit"}
2023-10-28T10:33:31.019Z	INFO	xds-translator	runner/runner.go:54	received an update	{"runner": "xds-translator"}
2023-10-28T10:33:31.019Z	INFO	global-ratelimit	runner/runner.go:116	received a notification	{"runner": "global-ratelimit"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xd0 pc=0x185b2fc]

goroutine 211 [running]:
github.com/envoyproxy/gateway/internal/xds/translator.patchRouteWithRateLimit(0x0, 0x1b723a0?)
	/Users/bitliu/github/gateway/internal/xds/translator/ratelimit.go:136 +0x3c
github.com/envoyproxy/gateway/internal/xds/translator.patchRouteWithFilters(0x400026bb30, 0x0?)
	/Users/bitliu/github/gateway/internal/xds/translator/httpfilters.go:128 +0x5c
github.com/envoyproxy/gateway/internal/xds/translator.buildXdsRoute(0x4000187540)
	/Users/bitliu/github/gateway/internal/xds/translator/route.go:80 +0xdf8
github.com/envoyproxy/gateway/internal/xds/translator.(*Translator).processHTTPListenerXdsTranslation(0x4000b2d348, 0x40008b62e0, {0x400079f470, 0x1, 0x4000b2d268?}, 0x186bf40?, 0x20?, 0x0)
	/Users/bitliu/github/gateway/internal/xds/translator/translator.go:203 +0x8b0
github.com/envoyproxy/gateway/internal/xds/translator.(*Translator).Translate(0x4000b2d348, 0x400092c680)
	/Users/bitliu/github/gateway/internal/xds/translator/translator.go:67 +0x60
github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate.func1({{0x4000b006f0?, 0x0?}, 0x50?, 0x400092c680?}, 0x0?)
[...]

PR #2104, opened 2023-10-28
fix: panic when using nil xdsRouteAction

Xunzhuo:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/2103

cc @zhaohuabing


zirain (approved):

shawnh2 (approved):

zhaohuabing (approved):
LGTM! Thanks for fixing it!

arkodg:
thanks for catching this bit, weird how we got here, the diff looks good, curious how can the route action be empty when rate limit is set

PR #2108, opened 2023-10-29
helm: remove kube-rbac-proxy and update metrics service

Xunzhuo:
**What type of PR is this?**

Fixes #


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -49,10 +49,10 @@ config:
 
 envoyGatewayMetricsService:
   ports:
-    - name: https
-      port: 8443
+    - name: http
```
arkodg, line 52: it looks like this was pointing to the controller runtime metrics endpoint earlier
https://book.kubebuilder.io/reference/metrics
https://github.com/envoyproxy/gateway/blob/d65ab597196bba840c9ec87dca2676d6b0502a77/charts/gateway-helm/templates/envoy-gateway-deployment.yaml#L74

are we disabling that now ?

zirain (dismissed):

charts/gateway-helm/values.tmpl.yaml
Xunzhuo: We have collected them together, we have the controller runtime metrics and also our own metrics. They are exposing at 19001 together.

arkodg: do we then need to delete the above listen address in L75 ?

Xunzhuo: let me check.

Xunzhuo: Do not need it anymore.

shawnh2 (approved):

arkodg:
@Xunzhuo is this needed anymore https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/metrics-reader-rbac.yaml ?

arkodg (approved):
LGTM thanks !

Issue #422, opened 2022-09-26
HTTP3 support on gateway


kristiandrucker:
*Description*:
I'd like to ask about HTTP3 support in Envoy Gateway. Been looking in the official gateway-api sig as well as this repository but couldn't seem to find any mention of HTTP3.
Is it a feature that needs implementing, or is it just undocumented for now?

[optional *Relevant Links*:]
> https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http3
> https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ProtocolType

youngnick:
Gateway API hasn't defined HTTP3 standards yet, the tracking issue is kubernetes-sigs/gateway-api#687. This is probably blocked until there's more progress there.

tanujd11:
I think there is some progress on the gateway API issue for HTTP3 support stating no API changes are required. This to be supported by controller implementations.

tanujd11:
Hey @arkodg I would like to work on this.

zirain:
thanks @tanujd11

arkodg:
looks like the implementation in Envoy has been highlighted by 
https://github.com/kubernetes-sigs/gateway-api/issues/687#issuecomment-1223884691 (thanks  @alibo)

what is probably needed is a way for a user to opt in at the Gateway Listener (downstream) level to enable HTTP3 at the xds listener level (alt-svc header, udp listener, quic filter ... ) or should we enable this by default ?

@LanceEa ccing you since you've built this out in Emissary, any suggestion on how we can add this opt in knob using Gateway API ?

an interim solution could be to use EnvoyPatchPolicy post v0.5.0

tanujd11:
@arkodg Istio creates mirror listener to all HTTP ports. It is behind a feature flag however. The connection is created by client and they can decide based on alt-svc header received whether to upgrade the connection to HTTP3.

arkodg:
yeah @tanujd11 , I see 3 options here for native support
* add a feature flag in the [EnvoyProxy API (https://gateway.envoyproxy.io/latest/api/config_types.html#envoyproxyspec) that enables the mirror listener by default 
* we use some information from the Gateway API fields such as [Gateway protocol](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ProtocolType) to specifically set a HTTP3 Listener 
[...]

PR #2111, opened 2023-10-29
Downstream QUIC/HTTP3 Support

tanujd11:
**What type of PR is this?**

Feature

Adds HTTP3 support in envoy gateway

Fixes https://github.com/envoyproxy/gateway/issues/422


arkodg:
thanks for building this out @tanujd11 ! , will start reviewing this post the v0.6 release

arkodg:
hey @LanceEa can you help review this one ( since you have experience building this out before :) )

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -53,6 +53,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty"`
+	// EnableHTTP3 enables HTTP/3 support on the listener.
+	// Disabled by default.
+	// +optional
+	EnableHTTP3 bool `json:"enableHTTP3,omitempty"`
```
arkodg, line 59: lets make this `*bool` since its optional

internal/xds/translator/testdata/out/xds-ir/http3.routes.yaml
```diff
@@ -0,0 +1,17 @@
...
+    name: envoy-gateway/gateway-1/tls/*
+    routes:
+    - match:
+        prefix: /
+      name: httproute/default/httproute-1/rule/0/match/0/*
+      responseHeadersToAdd:
+      - append: true
+        header:
+          key: alt-svc
+          value: h3=":10443"; ma=86400
```
LanceEa, line 15: This port needs to be the port exposed externally outside the cluster to the client. So, in most cases involving a browser it is more than likely going to be `:443` and not the listener port. The LoadBalancer of type `Service` then needs to point traffic to 10443. That is what I was referring to as the tricky part because a single Service doesn't always allow for MixedProtocol. _note: that might have change since last time I looked into this but it is the more nuanced part to the setup._

LanceEa (commented):
I will defer to the maintainers on code coverage and general style but for the most part it looks good. I have left some feedback on the design and recommendations for future proofing it because support for mixedprotocols on the LoadBalancer Service and cloud vendors was limited last I looked.

internal/gatewayapi/listener.go
```diff
@@ -138,15 +138,19 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
...
 					Protocol:      proto,
 					ServicePort:   servicePort.port,
 					ContainerPort: containerPort,
 				}
-				// Only 1 listener is supported.
-				infraIR[irKey].Proxy.Listeners[0].Ports = append(infraIR[irKey].Proxy.Listeners[0].Ports, infraPort)
+
+				proxyListener := &ir.ProxyListener{
+					Name:  irHTTPListenerName(listener),
+					Ports: []ir.ListenerPort{infraPort},
```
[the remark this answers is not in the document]
LanceEa: This comment is referring to mixed protocols on a single Service of type `Loadbalancer`.

api/v1alpha1/clienttrafficpolicy_types.go
arkodg: I would love to just have a enable knob, but agree with @LanceEa, an empty struct maybe a good call for this API because some user may want to decide the listening port, and everyone loves tweaking port values to adhere with corp perimeter policies

tanujd11:
Hi @LanceEa @arkodg I have made changes to the API making it as following
```
http3Settings:
  enabled: true
```
At the moment only enabled is there. I can raise a follow up PR to expose separate port for QUIC and other alt-svc configs. Till then this feature can be termed as experimental. WDYT?

tanujd11:
Also it would be great if someone from @envoyproxy/gateway-reviewers / @envoyproxy/gateway-maintainers can try out the feature

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -59,6 +59,15 @@ type ClientTrafficPolicySpec struct {
...
 	EnableProxyProtocol *bool `json:"enableProxyProtocol,omitempty"`
+	// HTTP3Settings provides HTTP/3 configuration on the listener.
+	// +optional
+	HTTP3Settings *HTTP3Settings `json:"http3Settings,omitempty"`
+}
+
+type HTTP3Settings struct {
+	// Enabled enables HTTP/3 support on the listener.
+	// Disabled by default.
+	Enabled bool `json:"enabled,omitempty"`
```
arkodg, line 70: `Enabled` is not needed, instantiating the stunt should implicitly enable `http3`

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -59,6 +59,15 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	EnableProxyProtocol *bool `json:"enableProxyProtocol,omitempty"`
+	// HTTP3Settings provides HTTP/3 configuration on the listener.
+	// +optional
+	HTTP3Settings *HTTP3Settings `json:"http3Settings,omitempty"`
```
arkodg, line 64: prefer `http3`

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -59,15 +59,15 @@ type ClientTrafficPolicySpec struct {
...
-	HTTP3Settings *HTTP3Settings `json:"http3Settings,omitempty"`
+	HTTP3 *HTTP3Settings `json:"http3,omitempty"`
 }
 
+// HTTP3Settings provides HTTP/3 configuration on the listener.
 type HTTP3Settings struct {
-	// Enabled enables HTTP/3 support on the listener.
-	// Disabled by default.
-	Enabled bool `json:"enabled,omitempty"`
+	// TODO(tanujd11): placeholder for the http3 configuration at the moment
```
arkodg, line 70: Let's rm this TODO, this shows up on API docs

tanujd11: ohh shoot, completely slipped off my mind.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -291,6 +291,20 @@ func translateClientTrafficPolicyForListener(policySpec *egv1a1.ClientTrafficPol
 
 		// Translate Proxy Protocol
 		translateListenerProxyProtocol(policySpec.EnableProxyProtocol, httpIR)
+		// enable http3 if set and TLS is enabled
+		if httpIR.TLS != nil && policySpec.HTTP3 != nil {
+			httpIR.HTTP3 = &ir.HTTP3Settings{}
+			var proxyListenerIR *ir.ProxyListener
+			for _, proxyListener := range infraIR[irKey].Proxy.Listeners {
```
arkodg, line 298: nit: can we reuse `irListenerName` and also check if `irListenerName.Proxy != nil`

internal/ir/infra_test.go
```diff
@@ -20,7 +20,7 @@ func TestValidateInfra(t *testing.T) {
 		{
 			name:   "default",
 			infra:  NewInfra(),
-			expect: false,
+			expect: true,
```
arkodg, line 23: why is this `true` ?

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -80,6 +80,18 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 				TargetPort: target,
 			}
 			ports = append(ports, p)
+
+			if port.Protocol == ir.HTTPSProtocolType {
```
arkodg, line 84: should this live here on in the Gateway API layer  ?

internal/xds/translator/listener.go
```diff
@@ -137,6 +163,10 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 		}
 	}
 
+	if http3Listener {
+		mgr.CodecType = hcmv3.HttpConnectionManager_HTTP3
```
arkodg, line 167: will auto work ? 
is there a way for us to support http3 but also support something else in case the client doesnt support it ?

internal/gatewayapi/testdata/clienttrafficpolicy-status-conditions.out.yaml
```diff
@@ -275,11 +276,15 @@ infraIR:
     proxy:
       listeners:
       - address: ""
+        name: envoy-gateway/gateway-2/http
         ports:
         - containerPort: 10080
           name: http
           protocol: HTTP
           servicePort: 80
+      - address: ""
```
arkodg, line 285: should address be made a pointer so its optional and doesnt show up in yaml ?

internal/xds/translator/listener.go
tanujd11: Hi Arko, If HTTP3 is not supported then the other listener will be used. We have created a mirror UDP listener and the connection would be upgraded to HTTP3 according to the alt-svc header. So even if HTTP3 is not supported HTTP/HTTP2 will be used

internal/infrastructure/kubernetes/proxy/resource_provider.go
tanujd11: This is for opening UDP ports for only those ports which have HTTPS configured, That's a requirement for quic. So this condition is added here.

internal/ir/infra_test.go
tanujd11: As we removed NewProxyListener by default and only add it when needed(Since we replaced only 1 listener supported to multiple listener), there is no error generated in this test case, hence false. Before any empty listener was there which failed the validate function.

tanujd11:
Hi @arkodg , welcome back from the vacation. Your reviews on this will be greatly appreciated. Thanks.

arkodg (approved):

Xunzhuo (approved):
Great! Thanks for working on it : )

shawnh2 (approved):
lgtm

PR #2112, opened 2023-10-29
Enable HTTPRouteRewritePath test

arkodg:
* Addresses this conformance test https://github.com/kubernetes-sigs/gateway-api/blob/8f1b718543caae927bb4f57dfb538aff1efc8159/conformance/tests/httproute-rewrite-path.yaml#L32 by using a regex match and replace instead of a prefix replace, since a prefix replace adds a double slash
* tries to implement https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPPathModifier
* reverts #1125 which tried to solve the same problem
* trim the `/` suffix when it comes to prefix matches because Gateway API recommends ignoring the trailing `/` and envoy errors out when the trailing char is `/` for path separated prefix
* enhance IR sorting function so preference is given to exact match over path prefix match, this is needed when exact match of `/match` and path prefix match of `/match/` exists which need to reordered in envoy to make sure path prefix is not selected

Fixes: https://github.com/envoyproxy/gateway/issues/2004

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -107,9 +107,9 @@ func buildXdsRouteMatch(pathMatch *ir.StringMatch, headerMatches []*ir.StringMat
 				}
 			} else {
 				// Remove trailing /
-				*pathMatch.Prefix = strings.TrimSuffix(*pathMatch.Prefix, "/")
+				trimmedPrefix := strings.TrimSuffix(*pathMatch.Prefix, "/")
 				outMatch.PathSpecifier = &routev3.RouteMatch_PathSeparatedPrefix{
-					PathSeparatedPrefix: *pathMatch.Prefix,
+					PathSeparatedPrefix: trimmedPrefix,
 				}
 			}
 		} else if pathMatch.SafeRegex != nil {
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -43,7 +43,7 @@ func buildXdsRoute(httpRoute *ir.HTTPRoute) *routev3.Route {
 	case httpRoute.Redirect != nil:
 		router.Action = &routev3.Route_Redirect{Redirect: buildXdsRedirectAction(httpRoute.Redirect)}
 	case httpRoute.URLRewrite != nil:
-		routeAction := buildXdsURLRewriteAction(httpRoute.Destination.Name, httpRoute.URLRewrite)
+		routeAction := buildXdsURLRewriteAction(httpRoute.Destination.Name, httpRoute.URLRewrite, httpRoute.PathMatch)
 		if httpRoute.Mirrors != nil {
 			routeAction.RequestMirrorPolicies = buildXdsRequestMirrorPolicies(httpRoute.Mirrors)
 		}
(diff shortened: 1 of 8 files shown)


arkodg:
this is a workaround for https://github.com/envoyproxy/envoy/issues/26055, would be great to fix this upstream
cc @cpakulski @zhaohuabing

arkodg:
also relates to https://github.com/envoyproxy/gateway/pull/1125

zirain (commented):
LGTM, @qicz PTAL

Xunzhuo (approved):

internal/xds/translator/route.go
```diff
@@ -101,14 +101,15 @@ func buildXdsRouteMatch(pathMatch *ir.StringMatch, headerMatches []*ir.StringMat
 				Path: *pathMatch.Exact,
 			}
 		} else if pathMatch.Prefix != nil {
-			// when the prefix ends with "/", use RouteMatch_Prefix
-			if strings.HasSuffix(*pathMatch.Prefix, "/") {
+			if *pathMatch.Prefix == "/" {
```
qicz, line 104: IMO, `HasSuffix` can cover `Prefix = '/'` and suffix with `/`. if we remove the trailing `/` can not cover the use case that suffix with `/` like `/example/abc/`.

the `/example/abc` and `/example/abc/` are different.

internal/xds/translator/route.go
```diff
@@ -101,14 +101,15 @@ func buildXdsRouteMatch(pathMatch *ir.StringMatch, headerMatches []*ir.StringMat
...
 				outMatch.PathSpecifier = &routev3.RouteMatch_Prefix{
-					Prefix: *pathMatch.Prefix,
+					Prefix: "/",
 				}
 			} else {
+				// Remove trailing /
+				trimmedPrefix := strings.TrimSuffix(*pathMatch.Prefix, "/")
 				outMatch.PathSpecifier = &routev3.RouteMatch_PathSeparatedPrefix{
-					PathSeparatedPrefix: *pathMatch.Prefix,
```
qicz, line 111: if the path prefix is not suffix with `/`the `PathSpecifier` use the `PathSeparatedPrefix`.

internal/xds/translator/route.go
arkodg: according to the gateway API a trailing slash in prefix can be ignored, ptal at
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.PathMatchType
```
Matches based on a URL path prefix split by /. Matching is case sensitive and done on a path element by element basis. A path element refers to the list of labels in the path split by the / separator. When specified, a trailing / is ignored.

For example, the paths /abc, /abc/, and /abc/def would all match the prefix /abc, but the path /abcd would not.
```

internal/xds/translator/route.go
arkodg: is this related to your above comment, if so ptal at https://github.com/envoyproxy/gateway/pull/2112#discussion_r1377047715

qicz (approved):

zirain (approved):

Issue #881, opened 2023-01-10
Support OIDC based authentication


arkodg:
*Description*:
Support [OIDC](https://openid.net/specs/openid-connect-core-1_0.html) based authentication

Support in envoy 
* https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/oauth2_filter.html
* https://github.com/envoyproxy/envoy/issues/21982

Relates to https://github.com/envoyproxy/gateway/issues/336

lizan:
The Envoy oauth2 filter isn't really an OIDC authentication filter but issuing some custom custom format token in cookies/headers. We might want to look at some control plane for this like [vouch-proxy](https://github.com/vouch/vouch-proxy)

arkodg:
hey @lizan assigning this issue to you since you're looking into it, please feel to unassign if anything changes

arkodg:
@lizan do you think a new filter can be added to envoyproxy or the current oauth filter can be extended to support the OIDC flow (potentially including redirects, and an external storage backend [Redis] for server tokens) .
This would eliminate the need to achieve the same using an external service such as https://github.com/istio-ecosystem/authservice

cc @mattklein123 @ZackButcher

lizan:
Yes that is also doable. It is likely we need to do a new filter for OIDC. We just need to agree on a plan to support what flow and how to configure token storage etc. If that's more desired for EG community we can do that. Let's compare which path is better

mattklein123:
> Yes that is also doable. It is likely we need to do a new filter for OIDC. We just need to agree on a plan to support what flow and how to configure token storage etc. If that's more desired for EG community we can do that. Let's compare which path is better

Can we put together a short design doc on the options of sharing the filter vs. a new filter? I would imagine a lot of the logic will be shared. Perhaps there can be a common core shared between the 2 filters. I don't know enough details to know if that makes sense or not but we should see so we don't end up with substantial code duplication for a security critical feature.

lizan:
My bad I meant the opposite we shouldn't need a new filter for OIDC. The existing oauth filter can do a very basic flow of OIDC already.

For OIDC Discovery I don't think it should should be done in Envoy directly but should be a responsibility of control plane.

arkodg:
hey @lizan any update ?

lizan:
I've been testing several scenarios with oauth2 filter and looks like it can be used for OIDC authentication, so we don't need additional control plane for now. Will send a PR for API review soon.

[...]

PR #2122, opened 2023-10-30
feat: OIDC Gateway API

zhaohuabing:
What this PR does:
* add OIDC authentication to the SecurityPolicy API
* translate OIDC API to IR 

xds translation will be in a follow-up PR.

OIDC will be done per-route. The final xds output will look like this example yaml file: https://github.com/zhaohuabing/playground/blob/main/envoy/per-route-oauth2-oidc/envoy.yaml

Related: https://github.com/envoyproxy/gateway/issues/881


internal/provider/kubernetes/controller.go
```diff
@@ -388,6 +391,61 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+					r.log.Info("no matching ReferenceGrants found", "from", from.kind,
+						"from namespace", from.namespace, "target", to.kind, "target namespace", to.namespace)
+				default:
+					// RefGrant found
+					resourceMap.allAssociatedRefGrants[utils.NamespacedName(refGrant)] = refGrant
+					r.log.Info("added ReferenceGrant to resource map", "namespace", refGrant.Namespace,
+						"name", refGrant.Name)
+				}
+			}
+			resourceMap.allAssociatedNamespaces[secretNamespace] = struct{}{} // TODO Zhaohuabing do we need this line?
```
zhaohuabing, line 442: Do we need to check if this namespace exists?

api/v1alpha1/oidc_types.go
```diff
@@ -0,0 +1,64 @@
...
+	// no query or fragment components.
+	// +kubebuilder:validation:MinLength=1
+	Issuer string `json:"issuer"`
+
+	// TODO zhaohuabing validate the issuer
+
+	// The OIDC Provider's [authorization endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
+	// If not provided, EG will try to discover it from the provider's [Well-Known Configuration Endpoint](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
+	//
+	// +optional
```
arkodg, line 56: optional fields must be pointer types

api/v1alpha1/securitypolicy_types.go
```diff
@@ -56,6 +56,11 @@ type SecurityPolicySpec struct {
 	//
 	// +optional
 	JWT *JWT `json:"jwt,omitempty"`
+
+	// OIDC defines the configuration for the OpenID Connect (OIDC) authentication.
+	//
+	// +optional
+	OIDC *OIDC `json:"oidc,omitempty"`
```
arkodg, line 63: any use cases where this needs to be a list ?

api/v1alpha1/oidc_types.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+const OIDCClientSecretKey = "client_secret"
+
+// OIDC defines the configuration for the OpenID Connect (OIDC) authentication.
+type OIDC struct {
+	// The OIDC Provider configuration.
```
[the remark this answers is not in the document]
arkodg: can we add this as doc string comments  ?

api/v1alpha1/oidc_types.go
```diff
@@ -0,0 +1,64 @@
...
+	//
+	// +kubebuilder:validation:MinLength=1
+	ClientID string `json:"clientID"`
+
+	// The Kubernetes secret which contains the OIDC client secret assigned to
+	// the filter to be used in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	//
+	// This is an Opaque secret. The client secret should be stored in the key
+	// "client_secret".
```
arkodg, line 30: should this be snake case or camel case ?

arkodg:
hey @zhaohuabing this looks good, haven't looked at the implementation yet, added some minor comments reg API

api/v1alpha1/oidc_types.go
zhaohuabing: Probably uses camel case to keep the secret key aligned with the golang code.

internal/gatewayapi/securitypolicy.go
```diff
@@ -220,22 +250,39 @@ func resolveSecurityPolicyRouteTargetRef(policy *egv1a1.SecurityPolicy, routes m
...
 		cors *ir.CORS
 		jwt  *ir.JWT
+		oidc *ir.OIDC
+		err  error
 	)
 
 	if policy.Spec.CORS != nil {
-		cors = t.buildCORS(policy)
+		cors, err = t.buildCORS(policy.Spec.CORS)
+		if err != nil {
```
arkodg, line 266: I dont think we should return early, we should probably translate each field, and collect all `errs` and return them in the end

internal/gatewayapi/securitypolicy.go
```diff
@@ -316,23 +385,122 @@ func (t *Translator) buildCORS(policy *egv1a1.SecurityPolicy) *ir.CORS {
...
+		clientSecret *v1.Secret
+		err          error
+	)
+
+	from := crossNamespaceFrom{
+		group:     egv1a1.GroupName,
+		kind:      KindSecurityPolicy,
+		namespace: policy.Namespace,
 	}
+	if clientSecret, err = t.validateSecretRef(from, oidc.ClientSecret, resources); err != nil {
```
arkodg, line 426: are we validating anywhere that the `Secret` and `Policy` must be in the same namespace ?

internal/gatewayapi/securitypolicy.go
```diff
@@ -316,23 +385,122 @@ func (t *Translator) buildCORS(policy *egv1a1.SecurityPolicy) *ir.CORS {
...
+	}
+
+	val, ok := clientSecret.Data[egv1a1.OIDCClientSecretKey]
+	if !ok || len(val) == 0 {
+		return nil, fmt.Errorf("client secret not found in secret %s/%s", clientSecret.Namespace, clientSecret.Name)
+	}
+
+	// Discover the token and authorization endpoints from the issuer's
+	// well-known url if not explicitly specified
+	provider := oidc.Provider.DeepCopy()
```
arkodg, line 437: curious why deep copy is needed here

internal/ir/xds.go
```diff
@@ -319,6 +324,29 @@ type JWT struct {
...
+
+	// The OIDC client ID assigned to the filter to be used in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	ClientID string `json:"clientID"`
+
+	// The Kubernetes secret which contains the OIDC client secret assigned to the filter to be used in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	//
+	// This is an Opaque secret. The client secret should be stored in the key "clientSecret".
+	ClientSecret gwapiv1b1.SecretObjectReference `json:"clientSecret"`
```
arkodg, line 343: this should be []byte similar to https://github.com/envoyproxy/gateway/blob/c85f0e438ec3c41048739e794f27f3a33e853df0/internal/ir/xds.go#L219

https://github.com/envoyproxy/gateway/blob/c85f0e438ec3c41048739e794f27f3a33e853df0/internal/ir/xds.go#L147 also will need to get updated

internal/gatewayapi/validate.go
```diff
@@ -722,3 +725,61 @@ func (t *Translator) validateHostname(hostname string) error {
 
 	return nil
 }
+
+// validateSecretRef checks three things:
+//  1. Dose the secret reference has valid Group and kind
+//  2. If the secret reference is a cross namespace reference, is it permitted by
+//     any ReferenceGrant
+//  3. Does the secret exist
+func (t *Translator) validateSecretRef(
```
arkodg, line 734: shouldnt we have a similar function in the code for validating the Gateway listener TLS secret ?

internal/gatewayapi/securitypolicy.go
zhaohuabing: Agree 👍

internal/gatewayapi/securitypolicy.go
zhaohuabing: validateSecretRef checks three things:
  1. Does the secret reference have a valid Group and kind
  2. If the secret reference is a cross-namespace reference, is it permitted by any ReferenceGrant
  3. Does the secret exist

internal/gatewayapi/securitypolicy.go
zhaohuabing: Because the discovered token and authorization endpoints will be added to the copied provider.

internal/gatewayapi/validate.go
zhaohuabing: Plan to reuse this method for Gateway listener TLS secret as well later in a follow-up PR: 

internal/gatewayapi/validate.go line 288

// TODO zhaohuabing: reuse validateSecretRef

internal/ir/xds.go
zhaohuabing: The OIDC client secret will not be used as an inline string like TLS private key does, instead, it will be fetched from the EG xDS server through SDS, so we don't need []byte in the IR.

internal/gatewayapi/securitypolicy.go
arkodg: would prefer stricter approach first of making sure `secret` and `policy` are in same namespace
and we can relax this later

internal/gatewayapi/securitypolicy.go
```diff
@@ -273,16 +273,20 @@ func (t *Translator) translateSecurityPolicyForRoute(
...
+			errs = multierror.Append(errs, err)
 		}
 	}
 
-	// Apply IR to all relevant routes
 	// It can be difficult to reason about the state of the system if we apply
 	// part of the policy and not the rest. Therefore, we either apply all of it
 	// or none of it (when get errors when translating the policy)
+
+	if errs != nil {
```
arkodg, line 286: I dont think we should return early, maybe 1 feature is valid (e.g. `cors`) and one isn't (e.g. `jwt`), we should probably still
set those values but also return err for status feedback

internal/gatewayapi/securitypolicy.go
zhaohuabing: Gotcha

internal/gatewayapi/securitypolicy.go
zhaohuabing: Would it be confusing if the `Accepted` status is `false`, but part of the policy gets applied?

arkodg: I think thats fine, because the applied policy sub field is not tied to the error

api/v1alpha1/oidc_types.go
```diff
@@ -0,0 +1,74 @@
...
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	//
+	// +kubebuilder:validation:MinLength=1
+	ClientID string `json:"clientID"`
+
+	// The Kubernetes secret which contains the OIDC client secret to be used in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	//
+	// This is an Opaque secret. The client secret should be stored in the key
+	// "clientSecret".
```
arkodg, line 29: looking at https://kubernetes.io/docs/concepts/configuration/secret/, should it be kebab-case ?

zhaohuabing: How didn't I find that? Yes, we should follow existing examples.

arkodg (approved):
LGTM thanks !

zirain (approved):

zirain:
User documentation would be even better.

zhaohuabing:
User docs will be in a follow-up PR.

Issue #1953, opened 2023-10-11
Scale: Enabling leader election or Multiple EG writing status back for same resource.


Xunzhuo:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Envoy Gateway disabled leader election at default and didn`t expose it through envoy gateway config. When scaling cp replicas, Multiple EG will be writing status back for same resource.

We need to find out which is more expensive - enabling leader election and EG sending heartbeats to API server or multiple EG writing status back for same resource.

cc @envoyproxy/gateway-maintainers 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
> Envoy Gateway disabled leader election at default and didn`t expose it through envoy gateway config.

this suprise me, IMO, EG should enable leader election.

arkodg:
imo we should enable leader election (by default) to make sure only 1 EG controller can
  *  **write status** back
  *  can **create/write infra** (envoy proxy fleet & rate limit service)
  
but we need to make sure all EGs can **read/watch resources** and generate resources so any Envoy proxy can connect to it to get xds to support data plane scale out

cc @Xunzhuo can you help with this since you raise https://github.com/envoyproxy/gateway/pull/2123 :)

alexwo:
Hi, I'm interested in contributing. Could you please assign me to this issue?

arkodg:
thanks for picking this one up and detailing your plans !

approach LGTM, my suggestion would be to not spend any extra efforts on `Readiness Probes Adjustment` and disable the other replicas , and instead invest it directly on `Multi-Replica xDS Service` ( which is how its designed today where `kube-proxy` is helping us spread the load today to different xDS servers ). 
Only after 1 & 2, can we recommend users to confidently scale EG replicas

alexwo:
That sounds like a plan. We can move forward with implementing a comprehensive solution for xDS right from the start. 
Do you think it could still be beneficial to have an option for a deployment where only a single xDS instance is active? 
This setup might ensure better consistency, especially in scenarios where, for example, one replica might become outdated due to a synchronization issue, leading to potential discrepancies to some of it's connected envoys?

arkodg:
> @arkodg That sounds like a plan. We can move forward with implementing a comprehensive solution for xDS right from the start.
> 
[...]

PR #2123, opened 2023-10-31
feat: enable leader-election for eg controllers

Xunzhuo:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/1953

### DIFF
--- a/internal/provider/kubernetes/kubernetes.go
+++ b/internal/provider/kubernetes/kubernetes.go
@@ -37,7 +37,7 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 	mgrOpts := manager.Options{
 		Scheme:                 envoygateway.GetScheme(),
 		Logger:                 svr.Logger.Logger,
-		LeaderElection:         false,
+		LeaderElection:         true,
 		HealthProbeBindAddress: ":8081",
 		LeaderElectionID:       "5b9825d2.gateway.envoyproxy.io",
 	}


qicz (approved):

zhaohuabing:
Do we need it?  Can multiple EG pods run in parallel?

Xunzhuo:
> Do we need it? Can multiple EG pods run in parallel to share the workload?

@zhaohuabing, yes and like the question raised in https://github.com/envoyproxy/gateway/issues/1953:

We need to find out which is more expensive - enabling leader election and EG sending heartbeats to API server or multiple EG writing status back for same resource.

qicz:
> > Do we need it? Can multiple EG pods run in parallel to share the workload?
> 
> @zhaohuabing, yes and like the question raised in #1953:
> 
> We need to find out which is more expensive - enabling leader election and EG sending heartbeats to API server or multiple EG writing status back for same resource.

In fact, We just one replicas on prod env. but we can enable leader election and set the replicas `1`

zhaohuabing:
I am inclined to enable leader election by default. But we may need a knob in configuration to allow users to disable it in case that they have a fleet of Envoy instances and want to have multiple EG instances to share the workload of the XDS server.

Xunzhuo:
> I am inclined to enable leader election by default. But we may need a knob in configuration to allow users to disable it in case that they have a fleet of Envoy instances and want to have multiple EG instances to share the workload of the XDS server.

Yep, this should be configurable in Envoy Gateway Config.

Xunzhuo:
Let us make this into draft for now.

liorokman:
If leader election is enabled and the number of Envoy Gateway instances is >1, then will the XDS server on envoy instances that are not the leader return correct XDS information? 

As far as I can tell, the readiness probe that is [configured](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/kubernetes.go#L72) in the Envoy Gateway deployment always returns true.

In non-leader pods, does this not mean that the Kubernetes provider won't get notified or watch any resources and the XDS server provided by those non-leader pods will not contain valid information?

Because as long as the pods return "ready" in their readiness probe, traffic from Envoy Proxy to the gRPC XDS endpoint might get routed to one of these non-leader pods, and the Envoy EProxy configuration will be bad. 

This would mean that simply turning on the "leaderElection" flag is not a good idea.

arkodg:
hey @liorokman, you're right that this PR is not enough, had added some thoughts earlier in https://github.com/envoyproxy/gateway/issues/1953#issuecomment-1806263460

Xunzhuo:
close this and tracked in https://github.com/envoyproxy/gateway/issues/1953

Issue #2076, opened 2023-10-26
Enable HTTPRouteBackendProtocolH2C conformance test


arkodg:
*Description*:
>Describe the issue.

Conformance test -  https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-backend-protocol-h2c.go

This will require reading the `appProtocol` field on the `Service` resource and enabling `isHTTP2`

Skipped in https://github.com/envoyproxy/gateway/pull/2075

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #2136, opened 2023-10-31
conformance: Enable HTTPRouteBackendProtocolH2C conformance test

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/2076


internal/ir/xds.go
```diff
@@ -172,6 +172,7 @@ type HTTPListener struct {
 	// Routes associated with HTTP traffic to the service.
 	Routes []*HTTPRoute `json:"routes,omitempty" yaml:"routes,omitempty"`
 	// IsHTTP2 is set if the upstream client as well as the downstream server are configured to serve HTTP2 traffic.
+	// TODO: review the usage about this field
```
arkodg, line 175: we need some knob at the listener level to decide if we want to enable grpc-web & grpc-stats or not.
We should edit the description of this field to say `IsHTTP2 is set if the the listener is configured to serve HTTP2 traffic."

internal/ir/xds.go
```diff
@@ -466,15 +467,16 @@ func (r RouteDestination) Validate() error {
...
 
 // DestinationSetting holds the settings associated with the destination
 // +kubebuilder:object:generate=true
 type DestinationSetting struct {
 	// Weight associated with this destination.
 	// Note: Weight is not used in TCP/UDP route.
-	Weight    *uint32                `json:"weight,omitempty" yaml:"weight,omitempty"`
+	Weight *uint32 `json:"weight,omitempty" yaml:"weight,omitempty"`
+	// Protocol associated with this destination/port.
+	Protocol  model.AppProtocol      `json:"protocol" yaml:"protocol"`
```
arkodg, line 479: can we move `protocol` to the IR lib ?

internal/xds/translator/cluster.go
```diff
@@ -60,7 +61,15 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		cluster.RespectDnsTtl = true
 	}
 
-	if args.protocol == HTTP2 {
+	isHTTP2 := false
```
arkodg, line 64: just a nit, feel free to ignore
```
if args.settings != nil && args.settings[0].Protocol ..........
```

internal/ir/xds.go
arkodg: the same type works, into this file

arkodg: also should this be a ptr, since its optional ?

zirain: you means move `model.AppProtocol` to package ir? it's OK if you insist.
I put it under model, just because of it's common, shouldn't be too close to package IR.

zirain: > also should this be a ptr, since its optional?

it can be, but right now make TCP as default.

arkodg (approved):
LGTM thanks !

internal/ir/protocol.go
```diff
@@ -0,0 +1,27 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package ir
+
+type AppProtocol string
+
```
Xunzhuo, line 9: Thanks, here is a better place to keep it.

Xunzhuo (approved):
Tkx, LGTM

arkodg (approved):

PR #2138, opened 2023-10-31
feat: support FQDN address type for EndpointSlice

### LINKED ISSUE #1922: Support FQDN Address Type in EndpointSlice
*Description*:
We've recently added support for EndpointSlice, and current only support Static IPv4 Endpoint Addresses
We should also add support for FQDN address types 
https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#address-types

[optional *Relevant Links*:]

* looks like no changes are needed in the gateway api layer https://github.com/envoyproxy/gateway/blob/8f138f5ffdb050e4d61bee2328f27d1fe7303835/internal/gatewayapi/route.go#L1043
* in the xds translator layer, we'd need to detect whether the endpoint is a static address or FQDN and pass the right endpoint type (Static or DNS) https://github.com/envoyproxy/gateway/blob/8f138f5ffdb050e4d61bee2328f27d1fe7303835/internal/xds/translator/translator.go#L219
* Above step may require splitting up endpoints in multiple clusters if we need to handle mix endpoints (some static and some DNS)

shawnh2:
**What type of PR is this?**

Fixes #1922


internal/xds/translator/translator.go
```diff
@@ -461,5 +458,53 @@ const (
...
+		}
+		if len(fqdn) > 0 {
+			groups[DNS] = append(groups[DNS], &ir.DestinationSetting{
+				Weight:    ds.Weight,
+				Endpoints: fqdn,
+			})
+		}
+		if len(ips) > 0 {
+			groups[Static] = append(groups[Static], &ir.DestinationSetting{
+				Weight:    ds.Weight,
```
EltonzHu, line 488: From this implementation, you have already used weight to balance FQDN and IP. 
@arkodg Do we still want to go to this [simple solution](https://github.com/envoyproxy/gateway/issues/1922#issuecomment-1806292224)?

EltonzHu (changes requested):
Generally, code looks good to me. 
Just two minor comments:
1. Minor name change comment
2. From this implementation, IP and FQDN weight based balancing is achieved. Do we still want to go with this simple approach [here](https://github.com/envoyproxy/gateway/issues/1922#issuecomment-1806292224)?

shawnh2:
> Generally, code looks good to me. Just two minor comments:
> 
> 1. Minor name change comment
> 2. From this implementation, IP and FQDN weight based balancing is achieved. Do we still want to go with this simple approach [here](https://github.com/envoyproxy/gateway/issues/1922#issuecomment-1806292224)?

hi @EltonzHu , thanks for reviewing this. I haven't test this implementation yet, so it may not be right. If this implementation won't work, i will stick to the simple approach.

EltonzHu:
> > Generally, code looks good to me. Just two minor comments:
> > 
> > 1. Minor name change comment
> > 2. From this implementation, IP and FQDN weight based balancing is achieved. Do we still want to go with this simple approach [here](https://github.com/envoyproxy/gateway/issues/1922#issuecomment-1806292224)?
> 
> hi @EltonzHu , thanks for reviewing this. I haven't test this implementation yet, so it may not be right. If this implementation won't work, i will stick to the simple approach.

Sounds good. FYI, please be aware of this [issue](https://github.com/envoyproxy/gateway/issues/2170) opened and proposed solution with explicit locality setting on LB weights.

internal/xds/translator/translator.go
shawnh2: hi @EltonzHu i am sticking to the simple approach in this PR, PTAL

internal/ir/xds.go
```diff
@@ -520,6 +521,8 @@ type DestinationSetting struct {
 	// Protocol associated with this destination/port.
 	Protocol  AppProtocol            `json:"protocol" yaml:"protocol"`
 	Endpoints []*DestinationEndpoint `json:"endpoints,omitempty" yaml:"endpoints,omitempty"`
+	// AddressTypeState specifies the state of DestinationEndpoint address type.
+	AddressTypeState *DestinationAddressTypeState `json:"addressTypeState,omitempty" yaml:"addressTypeState,omitempty"`
```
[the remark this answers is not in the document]
shawnh2: updated

arkodg:
make gen-check is failing, you can run make testdata to rewrite test files

internal/gatewayapi/route.go
```diff
@@ -149,8 +149,14 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 		// a unique Xds IR HTTPRoute per match.
 		var ruleRoutes = t.processHTTPRouteRule(httpRoute, ruleIdx, httpFiltersContext, rule)
 
+		dstAddrTypeMap := make(map[ir.DestinationAddressType]int)
+
 		for _, backendRef := range rule.BackendRefs {
 			ds, backendWeight := t.processDestination(backendRef.BackendRef, parentRef, httpRoute, resources)
+			if !t.EndpointRoutingDisabled && ds != nil && len(ds.Endpoints) > 0 {
+				dstAddrTypeMap[*ds.AddressType]++
```
arkodg, line 157: probably want to check if `ds.AddressType` != nil

arkodg, line 157: also needs a `ds.AddressType != nil` check

internal/gatewayapi/route.go
```diff
@@ -163,14 +169,22 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
...
 			}
 		}
 
+		// TODO: support mixed endpointslice address type between backendRefs
+		if !t.EndpointRoutingDisabled && len(dstAddrTypeMap) > 1 {
+			parentRef.SetCondition(httpRoute,
+				gwapiv1.RouteConditionResolvedRefs,
+				metav1.ConditionFalse,
+				gwapiv1a1.RouteReasonResolvedRefs,
+				"Do not support mixed endpointslice address type between backendRefs")
```
arkodg, line 185:
```suggestion
				"mixed endpointslice address type between backendRefs is not supported")
```

internal/gatewayapi/route.go
```diff
@@ -1033,10 +1051,20 @@ func (t *Translator) processDestination(backendRef gwapiv1.BackendRef,
 		}
 	}
 
+	// TODO: support mixed endpointslice address type for the same backendRef
+	if !t.EndpointRoutingDisabled && addrType != nil && *addrType == ir.MIXED {
+		parentRef.SetCondition(route,
+			gwapiv1.RouteConditionResolvedRefs,
+			metav1.ConditionFalse,
+			gwapiv1a1.RouteReasonResolvedRefs,
+			"Do not support mixed endpointslice address type for the same backendRef")
```
arkodg, line 1060:
```suggestion
			"mixed endpointslice address type for the same backendRef is not supported")
```

arkodg:
few nits, else looks good, thanks !
cc @zirain

internal/gatewayapi/route.go
shawnh2: done

internal/gatewayapi/route.go
shawnh2: done

internal/gatewayapi/route.go
shawnh2: done

arkodg (dismissed):
LGTM thanks !

Xunzhuo (approved):
LGTM. Thanks

zirain (approved):

Issue #1907, opened 2023-09-26
Support retries


arkodg:
*Description*:

Add retry settings to be able to retry making requests to a backend

Relates to
* https://github.com/kubernetes-sigs/gateway-api/issues/1731
* https://github.com/envoyproxy/gateway/issues/1821#issuecomment-1729935509
* https://github.com/envoyproxy/gateway/pull/1899#issuecomment-1725910504

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
need something like TimeoutPolicy on upstream?

arkodg:
@zirain, based on @kflynn's comment [here](https://github.com/envoyproxy/gateway/issues/1821#issuecomment-1729935509) we should try and implement in this project as a top level field within `BackendTrafficPolicy`

zirain:
oh, may bad, there's only one client(the proxy) in ingress mode.
yes, we can put retries under `BackendTrafficPolicy `.

PR #2168, opened 2023-11-09
api: support retry on in BackendTrafficPolicy

tmsnan:
api: support retry on in BackendTrafficPolicy

Relates to #1821 & #1907


arkodg:
@kflynn and @AliceProxy, can you help review this one

kflynn:
@arkodg @tmsnan Sorry for the delay here! and many thanks for digging into this, I appreciate that. 🙂 

That said, I'd like to formalize something I've said on many other occasions:

**Flynn's First Law of Envoy:** Having end users interact directly with Envoy configuration concepts is _almost always_ a huge mistake. 🙂 

In Linkerd, the basic retry configuration is one line: `isRetryable: true`. You can go beyond that, but you don't have to. In Emissary, it's a little more complex than that, but it's still only... three lines, I think? The fact that Envoy has a ton of options is nearly always irrelevant to the user; let's not force them to use configuration structures that look anything like Envoy.

My strong preference here is that the base case should be a single line of configuration: to get sane retries with good defaults, the user should need _only_ to add `isRetryable: true`, or something very similar, at the top level of the `BackendTrafficPolicy`. Having a separate stanza for configuring policy is entirely reasonable but it should be optional.

It's certainly possible that I missed something here, and if so I apologize! 🙂 But in Envoy Gateway's name, the "Gateway" word is the really important bit, right? 🙂

tmsnan:
> @arkodg @tmsnan Sorry for the delay here! and many thanks for digging into this, I appreciate that. 🙂
> 
> That said, I'd like to formalize something I've said on many other occasions:
> 
> **Flynn's First Law of Envoy:** Having end users interact directly with Envoy configuration concepts is _almost always_ a huge mistake. 🙂
> 
> In Linkerd, the basic retry configuration is one line: `isRetryable: true`. You can go beyond that, but you don't have to. In Emissary, it's a little more complex than that, but it's still only... three lines, I think? The fact that Envoy has a ton of options is nearly always irrelevant to the user; let's not force them to use configuration structures that look anything like Envoy.
> 
> My strong preference here is that the base case should be a single line of configuration: to get sane retries with good defaults, the user should need _only_ to add `isRetryable: true`, or something very similar, at the top level of the `BackendTrafficPolicy`. Having a separate stanza for configuring policy is entirely reasonable but it should be optional.
> 
> It's certainly possible that I missed something here, and if so I apologize! 🙂 But in Envoy Gateway's name, the "Gateway" word is the really important bit, right? 🙂


IMO, the `RetryStrategy` in `BackendTrafficPolicy` offers a more comprehensive range of options. Users can customize the retry backoff strategy, retry budget strategy, and retry triggering conditions.  However, these advanced features may be too complex for typical Gateway users. We need to consider whether to fully retain these higher-order uses. 

It is indeed true that most users do not customize the configuration and prefer to use the recommended settings provided by the gateway to enable the retry feature. I agree with this. Perhaps we could keep the above-mentioned API and add a toggle switch to enable the commonly used retry function. @arkodg @kflynn

kflynn:
@tmsnan, so sorry for the delay (again)!

In practice, we haven't really found that _everything_ needs support, and many things just don't get used all that often. Here's an alternative that deliberately tries to minimize the amount of indentation and configuration needed for the things I think are very common:

## Suggestion

```yaml
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  ...
  isRetryable: true
  retryStrategy:
    maxBudget: 20    # percentage, default 20%
    maxRetries: 0    # count, if nonzero, maxBudget is ignored
    maxParallel: 3   # Optional
    minConcurrent: 3 # Optional

    retryOn:
      trigger: trigger-enum (see below)
      httpStatus: []int # valid when trigger is "retriable-status-code"

    perRetry:
      timeout: duration # optional
      idleTimeout: duration # optional
      backoff: # optional
        baseInterval: duration # required
        maxInterval: duration # optional
        # we can add rate limited based backoff config here if we want to

trigger-enum: one of
  5xx                       # HTTP events
  gateway-error             
  disconnect-reset
  connect-fail
  retriable-4xx
  refused-stream
  retriable-status-codes

  cancelled                 # GRPC events
  deadline-exceeded
  internal
  resource-exhausted
  unavailable
```

The idea here is that the proxy should know whether it's speaking HTTP or gRPC to a given backend. There's no point in configuring HTTP retries for a gRPC backend, and minimal config shouldn't require rewriting the retry config if you change types. (Obviously if you change protocols when you _have_ configured a trigger, you'd need to change that.)

## Examples

These are not meant to be exhaustive; they're meant to show specific cases that might come up.

### Minimal config

```yaml
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  isRetryable: true
```

Just give me retries and don't make me think about it. This will enable budgeted retries with a 20% retry budget. (Why budgeted and not counted? because the user experience is better for really simple configurations.)

### Minimal counted retries

```yaml
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  isRetryable: true
  retryStrategy:
    maxRetries: 1
```

Retry, but only once. (It might be worth considering whether the presence of `retryStrategy` implies `isRetryable: true`...)

### Retry only on 499 or 505 (I don't recommend this)

```yaml
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  isRetryable: true
  retryStrategy:
    retryOn:
      trigger: retriable-status-codes
      httpStatus: [ 499, 505 ]
```

Budgeted retries, but only retry on 499 or 505 status codes.

### Retry, controlling timing, on gRPC `cancelled`.

```yaml
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  isRetryable: true
  retryStrategy:
    retryOn:
      trigger: cancelled
    perRetry:
      timeout: 250ms
      idleTimeout: 5s
      backoff:
        baseInterval: 100ms
        maxInterval: 10s
```

## Summary

This is the kind of thing I mean when I say that we don't need to expose all the Envoy config. Envoy's API almost always considers every possible configuration as equal weight, as well as allowing a certain amount of configuration that's kind of nonsensical -- in practice people do not mix HTTP and gRPC to the same endpoint, for example, so configuring both at the same time just doesn't make much sense. We can - and should - be more opinionated, so that the common things are easy.

In, of course, my opinion. 🙂

tmsnan:
Thanks, the API looks much simpler, and there's really no need to distinguish between HTTP and gRPC retry trigger conditions.  It's a minor point that we might need to configure multiple conditions in the trigger simultaneously.
```
kind: BackendTrafficPolicy
metadata:
  name: example-policy
spec:
  isRetryable: true
  retryStrategy:
    retryOn:
      trigger:  gateway-error, retriable-4xx
```
I will update the API later.

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -65,6 +65,15 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty"`
+
+	// IsRetryAble indicates whether to enable budget retries with a 20% retry budget.
+	// +optional
+	IsRetryAble bool `json:"isRetryAble,omitempty"`
```
arkodg, line 71: I would prefer we didn't include that and instead relied only on `Retry`

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -65,6 +65,15 @@ type BackendTrafficPolicySpec struct {
...
+
+	// RetryStrategy provides more advanced retry usage, which allows users to customize the retry method (number of retries、retry budget and concurrency max retries),
+	// retry fallback strategy, and retry triggering conditions
+	// +optional
+	RetryStrategy *RetryStrategy `json:"retryStrategy,omitempty"`
```
arkodg, line 76: can we rename to `retry` ? all fields in here are singular

api/v1alpha1/retrystragegy_types.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RetryStrategy defines the retry strategy to be applied.
+type RetryStrategy struct {
+	// NumRetries is the number of retries to be attempted. Defaults to 0. If nonzero, maxBudget is ignored.
+	//
+	// +optional
+	MaxRetries *int `json:"maxRetries,omitempty"`
```
arkodg, line 17: is this `numRetries` or `maxRetries` ?

api/v1alpha1/retrystragegy_types.go
```diff
@@ -0,0 +1,130 @@
...
+	MinConcurrent *int `json:"minConcurrent,omitempty"`
+
+	// MaxParallel is the maximum number of parallel retries. If not specified, the default is 3. Priority lower than retry budget.
+	//
+	// +optional
+	MaxParallel *int `json:"maxParallel,omitempty"`
+
+	// RetryOn specifies the retry trigger condition.
+	//
+	// +optional
```
arkodg, line 39: lets add some defaults ?

tmsnan: like istio?
```
    retryOn:
      trigger: retriable-status-codes，connection-failure,  refused-stream, unavailable,cancelled
      httpStatus: [ 503 ]
```

arkodg: yeah as well as `numRetries`

arkodg:
hey @tmsnan, sorry for the delay in review
1. can we rm `MaxBudget` , `MaxParallel` & `MinConcurrent` in this first iteration
2. can we decide on sane defaults, so when a user sets
```
retry: {}
```

it results in
* NumRetries - X
* RetryOn.Triggers - [......]

tmsnan:
> hey @tmsnan, sorry for the delay in review
> 
> 1. can we rm `MaxBudget` , `MaxParallel` & `MinConcurrent` in this first iteration
> 2. can we decide on sane defaults, so when a user sets
> 
> ```
> retry: {}
> ```
> 
> it results in
> 
> * NumRetries - X
> * RetryOn.Triggers - [......]

done

api/v1alpha1/retry_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// Timeout is the timeout per retry attempt.
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+	// IdleTimeout is the upstream idle timeout per retry attempt.This parameter is optional and if absent there is no per try idle timeout.
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	IdleTimeout *metav1.Duration `json:"idleTimeout,omitempty"`
```
arkodg, line 93: can we also rm `IdleTimeout` in the first iteration ?

arkodg:
hey @tmsnan, this API is pretty close to the finish line, can you address the review comments ? tia

api/v1alpha1/retry_types.go
```diff
@@ -0,0 +1,114 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Retry defines the retry strategy to be applied.
+type Retry struct {
+	// NumRetries is the number of retries to be attempted. Defaults to 2.
+	//
+	// +optional
+	// +kubebuilder:default=2
+	NumRetries *int `json:"numRetries,omitempty"`
```
guydc, line 18: maybe int32 is more explicit. Also, we can validate that the value is >= 0.

api/v1alpha1/retry_types.go
```diff
@@ -0,0 +1,114 @@
...
+type RetryOn struct {
+	// Triggers specifies the retry trigger condition(Http/Grpc).
+	//
+	// +optional
+	Triggers []TriggerEnum `json:"triggers,omitempty"`
+
+	// HttpStatusCodes specifies the http status codes to be retried.
+	//
+	// +optional
+	HTTPStatusCodes []int `json:"httpStatusCodes,omitempty"`
```
guydc, line 41: maybe validate status in accepted range 1xx  - 5xx

tmsnan:
> hey @tmsnan, this API is pretty close to the finish line, can you address the review comments ? tia

Sorry for the late reply, there have been some family matters recently, and I will handle them soon.

arkodg:
> > hey @tmsnan, this API is pretty close to the finish line, can you address the review comments ? tia
> 
> Sorry for the late reply, there have been some family matters recently, and I will handle them soon.

Sorry to hear @tmsnan. 
No rush to pick up this PR, @guydc can help move this PR forward

guydc:
Closing as this work was merged in #2657

Issue #2057, opened 2023-10-24
Add helm lint to `make lint`


PetrMc:
*Description*:

Per @arkodg  [comment](https://github.com/envoyproxy/gateway/pull/1913#issuecomment-1739872090) 
we needed the way to validate the helm chart is capable with AWS `helm lint` 

After back and forth with AWS it's confirmed that the specific version `3.8.1' of Helm is required, ataching the screenshot here: 
![image](https://github.com/envoyproxy/gateway/assets/19846180/0e2be7b4-65b1-4d63-b50f-65c539c48d83)

*Relevant Links*:

the desired version - https://github.com/helm/helm/releases/tag/v3.8.1

Reference of the existing Helm tags - https://github.com/helm/helm/tags

arkodg:
thanks for raising this @PetrMc, would be good to add `helm lint` into our existing `lint` target

PR #2174, opened 2023-11-10
chore: add helm lint

misstick86:
**What type of PR is this?**

chore: add `helm lint` at `make lint` stage.

Fixes [2057](https://github.com/envoyproxy/gateway/issues/2057)

### DIFF
--- a/tools/make/helm.mk
+++ b/tools/make/helm.mk
@@ -27,3 +27,4 @@ helm-install: helm-generate ## Install envoy gateway helm chart from OCI registr
 
 helm-generate:
 	ImageRepository=${IMAGE} ImageTag=${TAG} envsubst < charts/gateway-helm/values.tmpl.yaml > ./charts/gateway-helm/values.yaml
+	helm lint charts/gateway-helm
--- a/charts/gateway-helm/Chart.yaml
+++ b/charts/gateway-helm/Chart.yaml
@@ -5,6 +5,7 @@ type: application
 
 version: v0.0.0-latest
 appVersion: "latest"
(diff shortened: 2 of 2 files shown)


tools/make/helm.mk
```diff
@@ -27,3 +27,7 @@ helm-install: helm-generate ## Install envoy gateway helm chart from OCI registr
 
 helm-generate:
 	ImageRepository=${IMAGE} ImageTag=${TAG} envsubst < charts/gateway-helm/values.tmpl.yaml > ./charts/gateway-helm/values.yaml
+
+lint: helm-lint
+helm-lint: helm-generate
```
arkodg, line 32: looking at our CI
https://github.com/envoyproxy/gateway/blob/main/.github/workflows/build_and_test.yaml
it looks like `helm-generate` is directly called
should we just move the `helm lint` command as a step inside `helm-generate` ?

misstick86: According to your recommendation, it is completely okay in ci. but I think `helm lint` should be a part of `make lint` Otherwise, if the user executes `make lint` separately, there will be no`helm lint` step.

arkodg: yeah instead it will get triggered in `helm generate` step and fail that step which imo is fine.
If we include `helm lint` in `lint` we will end up calling `helm generate` multiple times

misstick86: make sense. Already updated.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #1328, opened 2023-04-18
Support downstream and upstream Proxy Protocol


JuniorJPDJ:
Envoy Gateway should support Proxy Protocol on traffic coming from the world to support for example cases of Network Load Balancers from some cloud providers ([AWS example from Emissary Ingress docs](https://www.getambassador.io/docs/emissary/latest/topics/running/ambassador-with-aws#preserving-the-client-ip-address)) and other cases when we have some sort of L4 LB in front of the K8S cluster.

It should also support Proxy Protocol on egress traffic to support cases where eg. pointing to ExternalService outside of the cluster or we just need raw L4 socket (TCP or UDP) and source IP preservation.

I actually need the second case, but I know there are lots of people needing first one due to cloud deployments with L4 LBs in front of the k8s.

Istio for example has 2 issues for upstream proxy proto (https://github.com/istio/istio/issues/42257, https://github.com/istio/istio/issues/44342) and some forum posts from people implementing it with EnvoyFilter mechanism and it also has example of configuration in docs for downstream proxy proto due to cloud usage.

Upstream proxy protocol would for example allow placing nginx ingress controler behind Envoy Gateway without losing the source IPs when someone needs legacy ingress resource support: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-proxy-protocol

arkodg:
thanks for raising this issue @JuniorJPDJ, ive raised a issue in the upstream API repo

JuniorJPDJ:
Can you please link this issue here?

arkodg:
https://github.com/kubernetes-sigs/gateway-api/issues/1955

JuniorJPDJ:
Sure.
But IMO it should be extension, not part of main spec.
Lets see what upstream will say :D

arkodg:
@JuniorJPDJ yes an extension (policyAttachment or Filter) might be the way to go if the feature is not supported by the native upstream API

haq204:
Not sure if a policyAttachment or Filter would be necessary since you could also introduce it as extended protocol types in `Gateway.listeners[].protocol` (e.g. HTTPPROXY, HTTPSPROXY, etc.)

arkodg:
hey @haq204 thats a really good idea
+1 to it
now the protocol name is being overloaded with 3 nouns - `HTTP` + `S` + `PROXY` :)

JuniorJPDJ:
[...]

PR #2188, opened 2023-11-14
api: enableProxyProtocol in ClientTrafficPolicy

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1328


zirain:
Will there be other fields in the future? if so, prefer to something like: 
```
proxyProtocol:
  enable: true
  xxxx:
  yyy:
```

arkodg:
@zirain good question, the only thing I can think of is TLV, but unsure how common it is for end users to set it.
Also prefer implicit enable rather than explicit enable with a field

Xunzhuo:
Should ctp reuse the same proxyprotocol like btp https://github.com/envoyproxy/gateway/commit/99b1642b3c1ff20ed8f82ffb0ea8b88108dbdf96 ?

zhaohuabing:
> @zirain good question, the only thing I can think of is TLV, but unsure how common it is for end users to set it. Also prefer implicit enable rather than explicit enable with a field

We can make the `EnableProxyProtocol` as an empty structure for better future-proofing.

arkodg:
hey @Xunzhuo we can't use confirm from `btp` which picks a specific version (v1 or v2), this is server side, which can interpret both

zhaohuabing (approved):

arkodg:
emissary seems to have a `boolean` knob too
https://www.getambassador.io/docs/emissary/latest/topics/running/ambassador-with-aws#recommended-configuration
so does contour
https://projectcontour.io/guides/proxy-proto/

I would prefer to start with `bool` before we introduce a `struct` because there is no good use case yet

zirain (approved):
Thanks!

PR #2191, opened 2023-11-15
OIDC xds translation

zhaohuabing:
What this PR does:
* translate OIDC IR to XDS

docs and e2e test will be in a follow-up PRs.

OIDC will be done per-route. The final xds output will look like this example yaml file: https://github.com/zhaohuabing/playground/blob/main/envoy/per-route-oauth2-oidc/envoy.yaml

Related: 
* https://github.com/envoyproxy/gateway/issues/881
* https://github.com/envoyproxy/gateway/pull/2122


internal/xds/translator/httpfilters.go
```diff
@@ -37,14 +38,16 @@ func newOrderedHTTPFilter(filter *hcmv3.HttpFilter) *OrderedHTTPFilter {
 	order := 50
 
 	// Set a rational order for all the filters.
-	switch filter.Name {
-	case wellknown.CORS:
+	switch {
+	case filter.Name == wellknown.CORS:
 		order = 1
-	case jwtAuthnFilter:
+	case filter.Name == jwtAuthnFilter:
```
arkodg, line 44: shouldn't this be after Oauth2 ?

zhaohuabing: case isOAuth2Filter(filter):

zhaohuabing: Should we put `HTTPRateLimit` at the first?

internal/xds/translator/testdata/out/xds-ir/oidc.listeners.yaml
```diff
@@ -0,0 +1,135 @@
...
+      typedConfig:
+        '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+        commonHttpProtocolOptions:
+          headersWithUnderscoresAction: REJECT_REQUEST
+        http2ProtocolOptions:
+          initialConnectionWindowSize: 1048576
+          initialStreamWindowSize: 65536
+          maxConcurrentStreams: 100
+        httpFilters:
+        - name: envoy.filters.http.oauth2_first-route
```
arkodg, line 17: do we need to do anything here that disables processing these at the listener level ?

internal/xds/translator/translator.go
```diff
@@ -252,6 +260,16 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 			return err
 		}
 
+		// Create oauth2 token endpoint clusters, if needed.
+		if err := createOAuth2TokenEndpointClusters(tCtx, httpListener.Routes); err != nil {
+			return err
+		}
+
+		// Create oauth2 client and HMAC secrets, if needed.
```
arkodg, line 268: curious, does this order matter ?

internal/xds/translator/tracing.go
```diff
@@ -132,7 +132,7 @@ func processClusterForTracing(tCtx *types.ResourceVersionTable, tracing *ir.Trac
 		name:         clusterName,
 		settings:     []*ir.DestinationSetting{ds},
 		tSocket:      nil,
-		endpointType: DefaultEndpointType,
+		endpointType: EndpointTypeDNS,
```
arkodg, line 135: is this always DNS ?

internal/xds/translator/translator.go
```diff
@@ -195,7 +195,10 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 			}
 
 			// 1:1 between IR HTTPRoute and xDS config.route.v3.Route
-			xdsRoute := buildXdsRoute(httpRoute)
+			xdsRoute, err := buildXdsRoute(httpRoute)
```
arkodg, line 198: translation is best effort, we cannot bail out early and break other routes

internal/xds/translator/translator.go
```diff
@@ -240,6 +243,11 @@ func (t *Translator) processHTTPListenerXdsTranslation(tCtx *types.ResourceVersi
 		}
 		xdsRouteCfg.VirtualHosts = append(xdsRouteCfg.VirtualHosts, vHostsList...)
 
+		// Add per-route filter configs to the route config.
+		if err := patchRouteCfgWithPerRouteConfig(xdsRouteCfg, httpListener); err != nil {
```
arkodg, line 247: same as above, when cannot err out early and break other routes

internal/xds/translator/testdata/out/xds-ir/oidc.listeners.yaml
zhaohuabing: // patchRouteCfgWithPerRouteConfig appends per-route filter configurations to the
// route config.
// This is a generic way to add per-route filter configurations for all filters
// that has none-native per-route configuration support.
//   - For the filter type that without native per-route configuration support, EG
//     adds a filter for each route in the HCM filter chain.
//   - patchRouteCfgWithPerRouteConfig disables all the filters in the
//     typedFilterConfig of the route config.
//   - PatchRouteWithPerRouteConfig enables the corresponding oauth2 filter for each
//     route in the typedFilterConfig of the route.
//
// The filter types that have non-native per-route support: oauth2, basic authn
// Note: The filter types that have native per-route configuration support should
// use their own native per-route configuration.

internal/xds/translator/tracing.go
zhaohuabing: The previous default one is DNS, but the name is unclear, so I changed the name.

arkodg: cc @zirain who may have more context

internal/xds/translator/translator.go
zhaohuabing: I think the order here doesn't matter because DeltaXDS will eventually make the order correct when pushing all the resources to Envoy.

internal/xds/translator/translator.go
zhaohuabing: I think the order here doesn't matter because DeltaXDS will eventually make the order correct when pushing all the resources to Envoy

internal/xds/translator/testdata/out/xds-ir/oidc.listeners.yaml
arkodg: its not clear from the docs https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager
```
http_filters
(repeated [extensions.filters.network.http_connection_manager.v3.HttpFilter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-httpfilter)) A list of individual HTTP filters that make up the filter chain for requests made to the connection manager. [Order matters](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_filters#arch-overview-http-filters-ordering) as the filters are processed sequentially as request events happen.
```

so although its part of `http_filters`, it is disabled at per route level, which is interpreted by envoy ? should we also update envoy hcm http_filters docs to explain this ?

internal/xds/translator/translator.go
arkodg: we cannot err out due to syntactic issues or field validation issues, it must be due to a system issue

zhaohuabing: I had this doubt, but still did this way because all the other code in aslo returns eraly.

Should I change all of them? Or maybe we address this issue in a follow-up PR?

arkodg: sure, can you raise a GH issue for this

zhaohuabing: https://github.com/envoyproxy/gateway/issues/2200

internal/xds/translator/testdata/out/xds-ir/oidc.listeners.yaml
zhaohuabing: This generic per-route filter config feature is quite new. The Envoy docs actually don't mention it yet.

A workable example can be found here: https://github.com/zhaohuabing/playground/blob/main/envoy/per-route-oauth2-oidc/envoy.yaml

internal/xds/translator/tracing.go
zirain: thanks, it's more clear now.

zirain (approved):

PR #1945, opened 2023-10-10
fix typos in tls-cert-manager.md

lemonlinger:
Fixes #1931

PR #2197, opened 2023-11-16
doc: small patch cert-manager doc better experience

tanujd11:
Annotation for cluster-issuer was also changed it seems:

Checkout https://cert-manager.io/docs/usage/gateway/#supported-annotations


zirain (dismissed):

arkodg:
looks like we lost https://github.com/envoyproxy/gateway/pull/1945/files during the migration to Hugo

arkodg (approved):

Issue #1660, opened 2023-07-14
Support BasicAuth and KeyAuth based authentication


saltbo:
*Description*:

Support BasicAuth and KeyAuth based authentication

[optional *Relevant Links*:]

https://github.com/envoyproxy/envoy/issues/15365

zhaohuabing:
EG can support BasicAuth after the [credential injector http filter ](https://github.com/envoyproxy/envoy/pull/27769) is finished.

arkodg:
thinking out loud
* the user could represent the username password in the format used by `htpasswd`
https://httpd.apache.org/docs/2.4/programs/htpasswd.html
* which can be stored in a k8s secret
* we could use SDS to pass this info over to Envoy
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/secret.proto#secrets-configuration-proto
* last step in Envoy involves allowing the request if the `Authorization` header matches with one of the user:passwd entries, @zhaohuabing @cpakulski any idea how we can achieve this one ?

and something similar could be done for API Key as well

zhaohuabing:
@arkodg AFAIK, Envoy doesn't have built-in BasicAuth or API key support. But it may be accomplished by extensions such as lua, wasm, or  External Authorization.

zhaohuabing:
Basic Auth has been merged into Envoy. https://github.com/envoyproxy/envoy/pull/30079. I plan to add it to the SecurityPolicy in the next release.

arkodg:
@zhaohuabing can we reuse the same Envoy Filter for API key auth ?

zhaohuabing:
> @zhaohuabing can we reuse the same Envoy Filter for API key auth ? 

The problem is, unlike HTTP Basic Auth, there's no standard for API key auth. We may improve the http basic auth filter or and another filter to support API key auth following open API specs: https://swagger.io/docs/specification/authentication/api-keys/#:~:text=API%20keys%20are%20supposed%20to,mechanisms%20such%20as%20HTTPS%2FSSL.

arkodg:
yeah nice, it would look a lot like `from_headers` in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider

zhaohuabing:
Key auth is not supported yet. We can open another issue to track it.

[...]

PR #2199, opened 2023-11-16
feat: basic auth API

zhaohuabing:
Related to: https://github.com/envoyproxy/gateway/issues/1660

This PR adds API for HTTP Basic Auth:

```yaml
  kind: SecurityPolicy
  metadata:
    name: basic-auth
    namespace: default
  spec:
    basicAuth:
      users:
        name: "htpasswd-secret"
    targetRef:
      kind: HTTPRoute
      name: httproute-1
```


api/v1alpha1/basic_auth_types.go
```diff
@@ -0,0 +1,19 @@
...
+type BasicAuth struct {
+	// Username-hashed password pairs used to verify user credentials in the
+	// "Authorization" header.
+	//
+	// The value needs to be the htpasswd format, for example: "user1:{SHA}hashed_user1_password".
+	// Right now, ony SHA hash algorithm is supported.
+	// Reference to https://httpd.apache.org/docs/2.4/programs/htpasswd.html
+	//
+	// +kubebuilder:validation:MinItems=1
+	Users []string `json:"users"`
```
zhaohuabing, line 18: We could store the user list in a k8s secret, but I am inclined to just use a plain string array for easier UI. The passwords are hashed, so it should be ok to put them in the SecurityPolicy resources.

arkodg, line 18: must be a secret
we can either support multiple basic auth secrets https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret which might be too much to type out and manage or support a generic secret with `data` containing the key `htpasswd` with its contents in it

zhaohuabing: I vote to put `htpaswd` file in the `data` of a generic secret,  like the nginx k8s controller does: https://kubernetes.github.io/ingress-nginx/examples/auth/basic/

arkodg: yeah sgtm, reg key naming, should it be `.htpasswd` to reflect the name of the file ?

zhaohuabing: That's a good point.

api/v1alpha1/basic_auth_types.go
```diff
@@ -0,0 +1,22 @@
...
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+import gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+// BasicAuth defines the configuration for 	the HTTP Basic Authentication.
+type BasicAuth struct {
+	// The Kubernetes secret which contains the username-password pairs used to
```
arkodg, line 12:
```suggestion
	// The Kubernetes secret which contains the username-password pairs in htpasswd format, used to
```

zhaohuabing:
@saltbo

api/v1alpha1/basic_auth_types.go
```diff
@@ -0,0 +1,23 @@
...
+	// htpasswd format, used to verify user credentials in the "Authorization"
+	// header.
+	//
+	// This is an Opaque secret. The username-password pairs should be stored in
+	// the key ".htpasswd". As the key name indicates, the value needs to be the
+	// htpasswd format, for example: "user1:{SHA}hashed_user1_password".
+	// Right now, only SHA hash algorithm is supported.
+	// Reference to https://httpd.apache.org/docs/2.4/programs/htpasswd.html
+	// for more details.
+	Users gwapiv1b1.SecretObjectReference `json:"users"`
```
arkodg, line 22: prefer if this field is called `Htpasswd`

arkodg:

api/v1alpha1/basic_auth_types.go
zhaohuabing: I prefer `Users` because it's a user list, htpasswd is just the file format.

arkodg (approved):
LGTM thanks !

api/v1alpha1/basic_auth_types.go
```diff
@@ -0,0 +1,23 @@
...
+
+import gwapiv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+// BasicAuth defines the configuration for 	the HTTP Basic Authentication.
+type BasicAuth struct {
+	// The Kubernetes secret which contains the username-password pairs in
+	// htpasswd format, used to verify user credentials in the "Authorization"
+	// header.
+	//
+	// This is an Opaque secret. The username-password pairs should be stored in
```
arkodg, line 16: maybe also add a line mentioning that secret must live in the same namespace as the policy

arkodg (approved):

zirain (approved):

PR #2202, opened 2023-11-17
fix: xds translation return early, should be done in a best-effort manner

zhaohuabing:
fix: https://github.com/envoyproxy/gateway/issues/2200


zirain (approved):

Xunzhuo (approved):
looks good, thanks for improving it

PR #2206, opened 2023-11-18
feat: add support to extract from different identifier on `JWT`

### LINKED ISSUE #2186: Add support to extract from different identifier on `JWT` under Security Policy
*Description*:
Follow up from #1930.

Currently `JWT` only enables the default behavior to extract the JWT identifier from an HTTP request which uses `Authorization` header using Bearer format as well as query param using `access_token`. 
Thus, in my current use case, the requirement states that the JWT needs to be extracted from the `Cookie`, but because EG does not support this, I have to modify it using the Envoy Gateway Extension service. Hence, it would be much better if EG could support this behavior directly.

*Proposal*:

Add support to extract the JWT identifier from available options such as `from_cookies`, `from_headers`, and/or `from_params`, while keeping the default behavior if there are no extractors included in the API. 

We could introduce new fields for these additional extractors like below;

```yaml
kind: SecurityPolicy
metadata:
  name: jwt-example
spec:
  targetRef:
    kind: HTTPRoute
    name: foo
  jwt:
    providers:
      - name: example
        remoteJWKS:
          uri: https://raw.githubusercontent.com/envoyproxy/gateway/main/examples/kubernetes/jwt/jwks.json
        fromCookies:
          - session_access_token
        fromParams:
          - access_token
        fromHeaders:
          - name: x-access-token
            valuePrefix: ""
```

[optional *Relevant Links*:]

- [Envoy JWT Authentication](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider).

ardikabs:
**What type of PR is this?**

Introduce new fields on Security Policy's `JWT` API for adding support to enable JWT token extraction from different identifiers.

Fixes #2186


api/v1alpha1/jwt_types.go
```diff
@@ -81,3 +94,9 @@ type ClaimToHeader struct {
 	// to separate the JSON name path.
 	Claim string `json:"claim"`
 }
+
+// JWTFromHeader defines an HTTP header location to extract JWT token
+type JWTFromHeader struct {
+	Name        string `json:"name"`
+	ValuePrefix string `json:"valuePrefix"`
```
zhaohuabing, line 101: nit:
 * ValuePrefix is optional, should be pointer
 * Can we add some comments here to explain what ValuePrefix is?

zhaohuabing (commented):
Hey @ardikabs,  Are there any specific use cases where you require these fields? Envoy comes with a ton of configuration options, and our approach is to streamline the user experience by hiding the complexity of Envoy and giving user a simpler UI.

ardikabs:
> Hey @ardikabs, Are there any specific use cases where you require these fields? Envoy comes with a ton of configuration options, and EG won't want to expose them all to end users - that's the primary design goal of EG: make it easier for users to use Envoy as a gateway proxy. Our approach in EG is to streamline the user experience by hiding the complexity of Envoy and giving user a simpler UI.

Yes, currently in our use case, the running services are depends on JWT token extraction from Cookie. The reason why we stick with that because it is a running service and we are trying to migrate from Ingress to Gateway API. 
Changing the current implementation from cookie to header might be quite hassle, and might introduce unexpected behavior. 
Then, looking back at the Envoy implementation, this is doable.

api/v1alpha1/jwt_types.go
ardikabs: sure @zhaohuabing

zhaohuabing:
Sounds reasonable, defer to @arkodg

arkodg:
I think its reasonable, even this upstream GEP PR https://github.com/kubernetes-sigs/gateway-api/pull/2213/files  refers to fetching JWT from cookies

api/v1alpha1/jwt_types.go
```diff
@@ -54,7 +54,20 @@ type JWTProvider struct {
...
+
+	// FromHeaders is a list of HTTP request headers to extract the JWT token from.
+	// By default JWT Authentication extract JWT token from the Authorization HTTP Header using Bearer schema or
+	// extract it from access_token query parameters.
+	// See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto
+	FromHeaders []JWTFromHeader `json:"fromHeaders,omitempty"`
+
+	// FromCookies is a list of cookie names to extract the JWT token from.
+	// See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto
+	FromCookies []string `json:"fromCookies,omitempty"`
```
arkodg, line 66: can we only implement `cookies` for now ? since that is the only use case you are interested in ?

zhaohuabing (approved):
LGTM, thanks！

api/v1alpha1/jwt_types.go
```diff
@@ -54,7 +54,20 @@ type JWTProvider struct {
 	// The claim must be of type; string, int, double, bool. Array type claims are not supported
 	//
 	ClaimToHeaders []ClaimToHeader `json:"claimToHeaders,omitempty"`
-	// TODO: Add TBD JWT fields based on defined use cases.
+
```
arkodg, line 57: should this API look like
```
from:
- type: Cookie
  cookie: 
```

api/v1alpha1/jwt_types.go
```diff
@@ -54,7 +54,20 @@ type JWTProvider struct {
...
+	// FromHeaders is a list of HTTP request headers to extract the JWT token from.
+	// By default JWT Authentication extract JWT token from the Authorization HTTP Header using Bearer schema or
+	// extract it from access_token query parameters.
+	// See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto
+	FromHeaders []JWTFromHeader `json:"fromHeaders,omitempty"`
+
+	// FromCookies is a list of cookie names to extract the JWT token from.
+	// See: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto
```
arkodg, line 65: is this a `AND` case where all cookies specified MUST have a JWT in it ?

zhaohuabing:
You can specify multiple tokens and jwt filter will verify all of them.

api/v1alpha1/jwt_types.go
ardikabs: this works for me as well

api/v1alpha1/jwt_types.go
```diff
@@ -54,7 +54,11 @@ type JWTProvider struct {
 	// The claim must be of type; string, int, double, bool. Array type claims are not supported
 	//
 	ClaimToHeaders []ClaimToHeader `json:"claimToHeaders,omitempty"`
-	// TODO: Add TBD JWT fields based on defined use cases.
+
+	// ExtractFrom defines different ways to extract the JWT token from HTTP request.
+	// If empty, it defaults to extract JWT token from the Authorization HTTP request header using Bearer schema
+	// or access_token from query parameters.
+	ExtractFrom *JWTExtractor `json:"extractFrom,omitempty"`
```
arkodg, line 61: needs a `optional` annotation

api/v1alpha1/jwt_types.go
```diff
@@ -81,3 +85,28 @@ type ClaimToHeader struct {
...
+
+// JWTHeaderExtractor defines an HTTP header location to extract JWT token
+type JWTHeaderExtractor struct {
+	// Name is the HTTP header name to retrieve the token
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// ValuePrefix is the prefix that should be stripped before extracting the token.
+	// The format would be used by Envoy like "{ValuePrefix}<TOKEN>".
```
arkodg, line 97:
```suggestion
	// Here is the format that would be used by Envoy - "{ValuePrefix}<TOKEN>".
```

api/v1alpha1/jwt_types.go
```diff
@@ -81,3 +85,28 @@ type ClaimToHeader struct {
...
+// JWTExtractor defines a custom JWT token extraction from HTTP request.
+type JWTExtractor struct {
+	// Headers represents a list of HTTP request headers to extract the JWT token from.
+	Headers []JWTHeaderExtractor `json:"headers,omitempty"`
+
+	// Cookies represents a list of cookie names to extract the JWT token from.
+	Cookies []string `json:"cookies,omitempty"`
+
+	// Params represents a list of query parameters to extract the JWT token from.
+	Params []string `json:"params,omitempty"`
```
arkodg, line 111: is there a strong case for `Params` & `Headers` ? should be add it later and only add `Cookies` for now ?

api/v1alpha1/jwt_types.go
```diff
@@ -81,3 +85,28 @@ type ClaimToHeader struct {
...
+	Name string `json:"name"`
+
+	// ValuePrefix is the prefix that should be stripped before extracting the token.
+	// The format would be used by Envoy like "{ValuePrefix}<TOKEN>".
+	// For example, "Authorization: Bearer <TOKEN>", then the ValuePrefix="Bearer " with a space at the end.
+	ValuePrefix *string `json:"valuePrefix,omitempty"`
+}
+
+// JWTExtractor defines a custom JWT token extraction from HTTP request.
+type JWTExtractor struct {
```
arkodg, line 103: what happens when one of these fields is set ? does Envoy no longer peek into `Authorization: Bearer` ?

ardikabs: as per my observation, yes, it won't check to `Authorization` HTTP header, reference: https://github.com/envoyproxy/envoy/tree/main/api/envoy/config/filter/http/jwt_authn/v2alpha#the-locations-to-extract-jwt

api/v1alpha1/jwt_types.go
ardikabs: IMO, it is just yet another option to specify the desired location to extract JWT, omitting `Params` or `Headers` might not be necessary for my use case yet. But considering that those attributes are also ways to extract JWT tokens from different use cases, I included them now. 

But `add it later` is also works for me, thought @arkodg?

api/v1alpha1/jwt_types.go
ardikabs: add json tag `omitempty` should have the same behavior with `optional` annotation, should i still need to add this?
Reference: https://book.kubebuilder.io/reference/markers.html?highlight=omitempty#markers-for-configcode-generation

api/v1alpha1/jwt_types.go
arkodg: yeah lets add it later, prefer use case driven development

api/v1alpha1/jwt_types.go
arkodg: I believe`optional` is also added in other place to help with API docs

api/v1alpha1/jwt_types.go
ardikabs: noted @arkodg

api/v1alpha1/jwt_types.go
```diff
@@ -81,3 +87,9 @@ type ClaimToHeader struct {
 	// to separate the JSON name path.
 	Claim string `json:"claim"`
 }
+
+// JWTExtractor defines a custom JWT token extraction from HTTP request.
```
arkodg, line 91: we need to add a doc string here mentioning expected behavior when multiple values are set i.e. multiple cookies are set

arkodg:
hey @ardikabs, thanks for making the changes, PR looks good, added one  comment around API doc string, else LGTM

api/v1alpha1/jwt_types.go
ardikabs: added

arkodg (approved):
LGTM thanks !

Issue #882, opened 2023-01-10
code refactor: support features in xds layer using XdsPatch interface 


arkodg:
*Description*:
As we continue to enhance the [xds translator](https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/translator.go) to support more features, the code might become more complex, less readable and harder to maintain. 

One solution to this problem is to create a Patch interface so API Gateway feature libraries can patch in their translations to xds objects (listener, hcm, cluster) so the code is modular and more readable ,

zirain:
are you talking about something like [EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/)

arkodg:
no was referring to formalizing this interface call https://github.com/envoyproxy/gateway/blob/fb0c155f629ef72c8f7a7c7ccf9089bdad6680b2/internal/xds/translator/listener.go#L95

youngnick:
So, this is a _developer_ convenience, not a end user feature? If so, sounds good to me.

arkodg:
thanks for moving this in the Backlog, this is a good to have, and is not needed for the 0.3.0 release

arkodg:
completed with https://github.com/envoyproxy/gateway/pull/2210

PR #2210, opened 2023-11-19
refactor: simplify filter patching with an unified interface

zhaohuabing:
Why: as we continue to add more features such as CORS, JWT, OIDC, and other to XDS translator, the current HTTP filter patching code is growing increasingly complex and becoming harder to maintain.

This PR aims to unify the HTTP filter patching logics into an `httpFilter` interface to make them more modular and the code cleaner.

Related: https://github.com/envoyproxy/gateway/issues/882

### DIFF
--- a/internal/xds/translator/httpfilters.go
+++ b/internal/xds/translator/httpfilters.go
@@ -19,7 +19,12 @@ import (
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
-var httpFilters = [...]httpFilter{&cors{}, &jwt{}, &oidc{}}
+var httpFilters []httpFilter
+
+// registerHTTPFilter registers the provided HTTP filter.
+func registerHTTPFilter(filter httpFilter) {
+	httpFilters = append(httpFilters, filter)
+}
 
 // httpFilter is the interface for all the HTTP filters.
 //
--- a/internal/xds/translator/httpfilters.go
+++ b/internal/xds/translator/httpfilters.go
@@ -13,9 +13,54 @@ import (
 	hcmv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 
+	"github.com/envoyproxy/gateway/internal/xds/filters"
+	"github.com/envoyproxy/gateway/internal/xds/types"
+
 	"github.com/envoyproxy/gateway/internal/ir"
-	xdsfilters "github.com/envoyproxy/gateway/internal/xds/filters"
 )
+
+var httpFilters = [...]httpFilter{&cors{}, &jwt{}, &oidc{}}
+
+// httpFilter is the interface for all the HTTP filters.
+//
+// There are two ways to support per-route configuration for an HTTP filter:
+// - For the filters with native per-route configuration support:
+//   - patchHCM: EG adds the filter to the HCM filter chain only once.
+//   - patchRoute: EG adds the filter's native per-route configuration to each
+//     route's typedFilterConfig.
+//
+// - For the filters without native per-route configuration support:
+//   - patchHCM: EG adds a filter for each route in the HCM filter chain, the
+//     filter name is prefixed with the filter's type name, for example,
+//     "envoy.filters.http.oauth2", and suffixed with the route name. Each filter
+//     is configured with the route's per-route configuration.
+//   - patchRouteConfig: EG disables all the filters of this type in the
+//     typedFilterConfig of the route config.
+//   - PatchRouteWithPerRouteConfig: EG enables the corresponding filter for each
+//     route in the typedFilterConfig of that route.
+//
+// The filter types that haven't native per-route support: oauth2, basic authn
+// Note: The filter types that have native per-route configuration support should
+// always se their own native per-route configuration.
+type httpFilter interface {
+	// patchHCM patches the HttpConnectionManager with the filter.
+	patchHCM(mgr *hcmv3.HttpConnectionManager, irListener *ir.HTTPListener) error
+
+	// patchRouteConfig patches the provided RouteConfiguration with a filter's
+	// RouteConfiguration level configuration.
+	patchRouteConfig(rc *routev3.RouteConfiguration, irListener *ir.HTTPListener) error
+
+	// patchRoute patches the provide Route with a filter's Route level configuration.
+	patchRoute(route *routev3.Route, irRoute *ir.HTTPRoute) error
+
+	// patchResources adds all the other needed resources referenced by this
+	// filter to the resource version table.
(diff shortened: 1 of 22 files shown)


zhaohuabing:
Please hold this before https://github.com/envoyproxy/gateway/pull/2202 gets merged.

zhaohuabing:
Please hold this till https://github.com/envoyproxy/gateway/pull/2202 get merged.

zirain (approved):
I like OOO, lgtm

internal/xds/translator/testdata/out/xds-ir/jwt-ratelimit.clusters.yaml
```diff
@@ -40,6 +40,28 @@
   outlierDetection: {}
   perConnectionBufferLimitBytes: 32768
   type: EDS
+- commonLbConfig:
```
arkodg, line 43: why did this config change ?

internal/xds/translator/translator.go
```diff
@@ -469,7 +462,35 @@ func findXdsEndpoint(tCtx *types.ResourceVersionTable, name string) *endpointv3.
...
+		secret := r.(*tlsv3.Secret)
+		if secret.Name == name {
+			return secret
+		}
+	}
+
+	return nil
+}
+
+func addXdsSecretIfNotExist(tCtx *types.ResourceVersionTable, secret *tlsv3.Secret) error {
```
arkodg, line 481: prefer `addXdsSecret` , if we use `addXdsSecretIfNotExist` then we shouldnt be returning a `ErrXdsSecretExists` error

internal/xds/translator/translator.go
```diff
@@ -469,7 +462,35 @@ func findXdsEndpoint(tCtx *types.ResourceVersionTable, name string) *endpointv3.
...
+		return ErrXdsSecretExists
+	}
+
+	if err := tCtx.AddXdsResource(resourcev3.SecretType, secret); err != nil {
+		return err
+	}
+	return nil
+}
+
+func addXdsClusterIfNotExist(tCtx *types.ResourceVersionTable, args *xdsClusterArgs) error {
```
arkodg, line 493: same as https://github.com/envoyproxy/gateway/pull/2210/files#r1399882088

internal/xds/translator/oidc.go
```diff
@@ -262,15 +278,15 @@ func createOAuth2Secrets(tCtx *types.ResourceVersionTable, routes []*ir.HTTPRout
 		}
 
 		clientSecret := buildOAuth2ClientSecret(route)
-		if err := tCtx.AddXdsResource(resourcev3.SecretType, clientSecret); err != nil {
+		if err := addXdsSecretIfNotExist(tCtx, clientSecret); err != nil {
```
arkodg, line 281: what happens when 2 different policies (attached to 2 different routes) access the same secret ?

internal/xds/translator/httpfilters.go
```diff
@@ -13,10 +13,55 @@ import (
...
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 
+	"github.com/envoyproxy/gateway/internal/xds/filters"
+	"github.com/envoyproxy/gateway/internal/xds/types"
+
 	"github.com/envoyproxy/gateway/internal/ir"
-	xdsfilters "github.com/envoyproxy/gateway/internal/xds/filters"
 )
 
+var httpFilters = [...]httpFilter{&cors{}, &jwt{}, &oidc{}}
```
arkodg, line 22: is this needed ? someone will always forget to add this

internal/xds/translator/httpfilters.go
```diff
@@ -13,10 +13,55 @@ import (
...
+//     is configured with the route's per-route configuration.
+//   - patchRouteConfig: EG disables all the filters of this type in the
+//     typedFilterConfig of the route config.
+//   - PatchRouteWithPerRouteConfig: EG enables the corresponding filter for each
+//     route in the typedFilterConfig of that route.
+//
+// The filter types that haven't native per-route support: oauth2, basic authn
+// Note: The filter types that have native per-route configuration support should
+// always se their own native per-route configuration.
+type httpFilter interface {
```
arkodg, line 45: ❤️  nicely done, thanks for all the comments

internal/xds/translator/testdata/out/xds-ir/jwt-ratelimit.clusters.yaml
zhaohuabing: The same configuration, just the `commonLbConfig` location changed in its parent config. It's because of this line: 

```
		// Add all the other needed resources referenced by this filter to the
		// resource version table.
		if err := patchResources(tCtx, httpListener.Routes); err != nil {
			return err
		}

		// RateLimit filter is handled separately because it relies on the global
		// rate limit server configuration.
		// Check if a ratelimit cluster exists, if not, add it, if it's needed.
		if err := t.createRateLimitServiceCluster(tCtx, httpListener); err != nil {
			errs = multierror.Append(errs, err)
		}
```

zhaohuabing: The same configuration, just the `commonLbConfig` location changed in its parent config. It's because of this line: 

```
		// Add all the other needed resources referenced by this filter to the
		// resource version table.
		if err := patchResources(tCtx, httpListener.Routes); err != nil {
			return err
		}

		// RateLimit filter is handled separately because it relies on the global
		// rate limit server configuration.
		// Check if a ratelimit cluster exists, if not, add it, if it's needed.
		if err := t.createRateLimitServiceCluster(tCtx, httpListener); err != nil {
			errs = multierror.Append(errs, err)
		}
```

internal/xds/translator/translator.go
zhaohuabing: Yes, it does return a `ErrXdsSecretExists` if it's already there.

internal/xds/translator/oidc.go
zhaohuabing: Good catch! 

Normally this shouldn't happen because different clients don't share the same secret. However, we can not assume users won't apply the same policy to different routes. I think we shouldn't return error in this case.

zhaohuabing: That might happen, I choose to create a secret for each route and don't share secret across routes. I will add some comments so we (including me) won't get confused by the codes in the future.

internal/xds/translator/httpfilters.go
zhaohuabing: The tests will never pass if someone forget this, but it will be better to just modify one file. I replaced this a register method in each HTTPFilter.

arkodg (approved):
LGTM, thanks for adding this refactor

zirain (approved):

Issue #1902, opened 2023-09-20
Support slow start model


tmsnan:
*Description*:
>Specific configuration for the RoundRobin/LEAST_REQUEST load balancing policy.

[optional *Relevant Links*:]
>https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig

zirain:
should this be part of https://github.com/envoyproxy/gateway/issues/1821?

arkodg:
yah id prefer if we added as a field in under each LB type in https://github.com/envoyproxy/gateway/issues/1105

arkodg:
should be possible to add this API into the LoadBalancing config
https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/loadbalancer_types.go

PR #2219, opened 2023-11-20
feat:support slow start mode

tmsnan:
**What type of PR is this?**
Support slow start model

Releated: 
#1902
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig


api/v1alpha1/loadbalancer_types.go
```diff
@@ -24,6 +26,13 @@ type LoadBalancer struct {
 	//
 	// +optional
 	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty"`
+
+	// SlowStartWindow defines the duration of the warm up period for newly added host.
+	// If set, during slow start window, traffic sent to the newly added hosts will gradually increase in a linear manner.
+	// Currently this is only supported for RoundRobin and LeastRequest load balancers.
```
zirain, line 32: can you add CEL validation for this?

api/v1alpha1/loadbalancer_types.go
```diff
@@ -24,6 +26,13 @@ type LoadBalancer struct {
...
+	// For additional details,
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig
+	SlowStartWindow *metav1.Duration `json:"slowStartWindow,omitempty"`
```
shawnh2, line 35: add `// +optional` in comment for kubebuilder

internal/xds/translator/cluster.go
```diff
@@ -78,8 +78,27 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
 	} else if args.loadBalancer.LeastRequest != nil {
 		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
+		if args.loadBalancer.LeastRequest.SlowStartWindow > 0 {
+
```
shawnh2, line 82: extra line

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -425,6 +425,9 @@ func (t *Translator) buildLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.
 		lb = &ir.LoadBalancer{
 			LeastRequest: &ir.LeastRequest{},
 		}
+		if policy.Spec.LoadBalancer.SlowStartWindow != nil {
```
shawnh2, line 428: can you also add some test under https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata

zirain:
can you share the use case for this? add an e2e test for this if possible?
there're tons of feautes envoy support, EG won't expose all of them.
IMO, most of the edge case can be achieved by `EnvoyPath`?

shawnh2 (commented):
generally LGTM, just some nits and please fix ci

tmsnan:
> can you share the use case for this? add an e2e test for this if possible? there're tons of feautes envoy support, EG won't expose all of them. IMO, most of the edge case can be achieved by `EnvoyPath`?

1. I will try to implement this e2e test through indicator comparison later.
2. `Slow-start mode` is useful for certain backend applications that might otherwise be overwhelmed during warm-up. Such applications may respond slowly to requests or return errors immediately after a pod starts or a container restarts. Using a slow-start configuration to gradually increase traffic to the most recently started service endpoint can reduce the impact of this behavior on users. So I think it is necessary to provide this feature in load balancing. Secondly, the API does not provide complicated configuration (like envoy) to enable slow start. It only enables the function by configuring the slow response time window. If more fine-grained use is required, `EnvoyPath` is suitable.

api/v1alpha1/loadbalancer_types.go
tmsnan: done

internal/gatewayapi/backendtrafficpolicy.go
tmsnan: done

api/v1alpha1/loadbalancer_types.go
```diff
@@ -24,6 +27,15 @@ type LoadBalancer struct {
...
 	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty"`
+
+	// SlowStartWindow defines the duration of the warm up period for newly added host.
+	// If set, during slow start window, traffic sent to the newly added hosts will gradually increase in a linear manner.
+	// Currently this is only supported for RoundRobin and LeastRequest load balancers.
+	// For additional details,
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig
+	//
+	// +optional
+	SlowStartWindow *metav1.Duration `json:"slowStartWindow,omitempty"`
```
arkodg, line 38: should this live inside `roundRobin` and `leastRequest` since its not available for all lb types (like `consistentHash` ) ?

arkodg, line 38: another question regarding API - should it be `slowStart.window` to accommodate for future fields (like `aggression` ) or do we think that its not needed for the common use cases ?

arkodg: or should we just rely on the CEL validation here ?

tmsnan: IMO, currently, for most use cases, linear growth of traffic is enough. If there are many requirements for non-linear growth later, we will expand it. 😄

tmsnan: IMO, CEL may be better, maybe the envoy extension supports other loadbalancer type. 👀

arkodg: Cool, no strong opinion here, current API is fine

arkodg: I vote to make it a struct just in case we need to expand in the future

tmsnan: done

api/v1alpha1/loadbalancer_types.go
```diff
@@ -55,3 +65,13 @@ const (
...
 	SourceIPConsistentHashType ConsistentHashType = "SourceIP"
 )
+
+// SlowStart defines the configuration related to the slow start load balancer policy.
+type SlowStart struct {
+	// Window defines the duration of the warm up period for newly added host.
+	// During slow start window, traffic sent to the newly added hosts will gradually.
+	// Currently only supports linear growth of traffic. For additional details,
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig
+	Window *metav1.Duration `json:"window,omitempty"`
```
arkodg, line 75: is this a required field ? from the envoy API it looks like it

api/v1alpha1/loadbalancer_types.go
```diff
@@ -55,3 +65,13 @@ const (
 	// SourceIPConsistentHashType hashes based on the source IP address.
 	SourceIPConsistentHashType ConsistentHashType = "SourceIP"
 )
+
+// SlowStart defines the configuration related to the slow start load balancer policy.
+type SlowStart struct {
+	// Window defines the duration of the warm up period for newly added host.
+	// During slow start window, traffic sent to the newly added hosts will gradually.
```
arkodg, line 72:
```suggestion
	// During slow start window, traffic sent to the newly added hosts will gradually increase.
```

api/v1alpha1/loadbalancer_types.go
tmsnan: Yeap, added verification.

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -423,15 +423,29 @@ func (t *Translator) buildLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.
 		}
 	case egv1a1.LeastRequestLoadBalancerType:
 		lb = &ir.LoadBalancer{
-			LeastRequest: &ir.LeastRequest{},
+			LeastRequest: &ir.LeastRequest{
+				SlowStart: &ir.SlowStart{},
```
arkodg, line 427: instead of setting it here , lets set it under `policy.Spec.LoadBalancer.SlowStart != nil`

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -423,15 +423,29 @@ func (t *Translator) buildLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.
...
 		}
 	case egv1a1.RandomLoadBalancerType:
 		lb = &ir.LoadBalancer{
 			Random: &ir.Random{},
 		}
 	case egv1a1.RoundRobinLoadBalancerType:
 		lb = &ir.LoadBalancer{
-			RoundRobin: &ir.RoundRobin{},
+			RoundRobin: &ir.RoundRobin{
+				SlowStart: &ir.SlowStart{},
```
arkodg, line 442: same as above

internal/ir/xds.go
```diff
@@ -1062,3 +1071,10 @@ type ConsistentHash struct {
 	// Hash based on the Source IP Address
 	SourceIP *bool `json:"sourceIP,omitempty" yaml:"sourceIP,omitempty"`
 }
+
+// SlowStart defines the slow start configuration.
+// +k8s:deepcopy-gen=true
+type SlowStart struct {
+	// Window defines the duration of the warm up period for newly added host.
+	Window time.Duration `json:"window,omitempty" yaml:"window,omitempty"`
```
arkodg, line 1079: lets use `*metav1.Duration` here 
also `omitempty` is not needed since its a required field

internal/gatewayapi/testdata/backendtrafficpolicy-with-loadbalancer.in.yaml
```diff
@@ -91,3 +110,33 @@ backendTrafficPolicies:
       type: ConsistentHash
       consistentHash:
         type: SourceIP
+- apiVersion: gateway.envoyproxy.io/v1alpha1
```
arkodg, line 113: instead of creating a new route, can we just add this `slowStart` config to the existing config ?

internal/xds/translator/cluster.go
```diff
@@ -78,8 +78,30 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
...
+							SlowStartWindow: durationpb.New(args.loadBalancer.LeastRequest.SlowStart.Window),
+						},
+					},
+				}
+			}
+		}
 	} else if args.loadBalancer.RoundRobin != nil {
 		cluster.LbPolicy = clusterv3.Cluster_ROUND_ROBIN
+		if args.loadBalancer.RoundRobin.SlowStart != nil {
+			if args.loadBalancer.RoundRobin.SlowStart.Window > 0 {
```
arkodg, line 95: lets make this a `ptr` so we can check for `nil` ?

internal/gatewayapi/testdata/backendtrafficpolicy-with-loadbalancer.in.yaml
tmsnan: The existing route has been used to test `ConsistentHash`, so a new route needs to be added

internal/gatewayapi/backendtrafficpolicy.go
arkodg: this line should be deleted

internal/ir/xds.go
```diff
@@ -1080,3 +1088,10 @@ type ProxyProtocol struct {
 	// Version of proxy protocol to use
 	Version ProxyProtocolVersion `json:"version,omitempty" yaml:"version,omitempty"`
 }
+
+// SlowStart defines the slow start configuration.
+// +k8s:deepcopy-gen=true
+type SlowStart struct {
+	// Window defines the duration of the warm up period for newly added host.
+	Window *metav1.Duration `json:"window" yaml:"window"`
```
zhaohuabing, line 1096: Why window is a pointer? It should not be optional.

api/v1alpha1/loadbalancer_types.go
```diff
@@ -55,3 +65,14 @@ const (
...
 )
+
+// SlowStart defines the configuration related to the slow start load balancer policy.
+type SlowStart struct {
+	// Window defines the duration of the warm up period for newly added host.
+	// During slow start window, traffic sent to the newly added hosts will gradually increase.
+	// Currently only supports linear growth of traffic. For additional details,
+	// see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig
+	// +kubebuilder:validation:Required
+	Window *metav1.Duration `json:"window"`
```
zhaohuabing, line 76: Why `Window` is a pointer? It should not be optional.

zhaohuabing (commented):
LGTM. Just some small nits.

internal/ir/xds.go
arkodg: I suggested this, to make it easier for adding `if` checks within xds code

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -436,13 +436,33 @@ func (t *Translator) buildLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.
...
+		}
 	case egv1a1.RandomLoadBalancerType:
 		lb = &ir.LoadBalancer{
 			Random: &ir.Random{},
 		}
 	case egv1a1.RoundRobinLoadBalancerType:
 		lb = &ir.LoadBalancer{
-			RoundRobin: &ir.RoundRobin{},
+			RoundRobin: &ir.RoundRobin{
+				SlowStart: &ir.SlowStart{},
```
arkodg, line 455: this slow start field still exists on line 455, can you rm this ?

internal/ir/xds.go
zhaohuabing: Do you mean here? Can the second if be removed since `window` is a mandatory field?

```
		if args.loadBalancer.RoundRobin.SlowStart != nil {
			if args.loadBalancer.RoundRobin.SlowStart.Window != nil {
				cluster.LbConfig = &clusterv3.Cluster_RoundRobinLbConfig_{
					RoundRobinLbConfig: &clusterv3.Cluster_RoundRobinLbConfig{
						SlowStartConfig: &clusterv3.Cluster_SlowStartConfig{
							SlowStartWindow: durationpb.New(args.loadBalancer.RoundRobin.SlowStart.Window.Duration),
						},
					},
				}
			}
		}
```

internal/gatewayapi/backendtrafficpolicy.go
tmsnan: Done, my bad, I totally missed it many times.

arkodg (approved):

zirain (approved):

PR #2224, opened 2023-11-20
feat: basic auth impl

zhaohuabing:
Related to: https://github.com/envoyproxy/gateway/issues/1660

close https://github.com/envoyproxy/gateway/issues/2226

@saltbo


zhaohuabing:
Please hold this till https://github.com/envoyproxy/gateway/pull/2210 get merged.

internal/ir/xds.go
```diff
@@ -355,6 +357,14 @@ type OIDC struct {
 	Scopes []string `json:"scopes,omitempty" yaml:"scopes,omitempty"`
 }
 
+// BasicAuth defines the schema for the HTTP Basic Authentication.
+//
+// +k8s:deepcopy-gen=true
+type BasicAuth struct {
+	// The username-password pairs in htpasswd format.
+	Users string `json:"users,omitempty" yaml:"users,omitempty"`
```
arkodg, line 365: should this be []byte ?

arkodg: also needs to be unreadable, need to add logic in `Printable`

internal/xds/translator/basicauth.go
```diff
@@ -0,0 +1,189 @@
...
+			TypedConfig: basicAuthAny,
+		},
+	}, nil
+}
+
+func basicAuthFilterName(route *ir.HTTPRoute) string {
+	return perRouteFilterName(basicAuthFilter, route.Name)
+}
+
+func basicAuthConfig(route *ir.HTTPRoute) *basicauthv3.BasicAuth {
```
arkodg, line 99: shouldnt we be sending the data via SDS ?

zhaohuabing: The password is hashed so it's safe to include it in an inline string.  It can be improved by sending the user list through SDS, but we need to modify the Envoy Basic Auth filter first.  I can address this in a subsequent iteration when I have bandwidth.

internal/ir/xds.go
zhaohuabing: It can be changed to []byte.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2234, opened 2023-11-23
fix: add route descriptor prefix for ratelimit xds translator

### LINKED ISSUE #2225: Two or more BackendTrafficPolicy cause collision for ratelimit
*Description*:

When two TrafficBackendPolicy are applied for rate limiting a GRPCROUTE. rate limit service reports several error `Error loading configuration: duplicate descriptor composite key`

*Repro steps*:

```yaml

kind: BackendTrafficPolicy
metadata:
  name: serviceA-ratelimit-all-ips
spec:
  targetRef:
    kind: GRPCRoute
    name: serviceA
    namespace: default
  rateLimit:
    type: Global
    global:
      rules:
        - clientSelectors:
            - sourceCIDR:
                value: 0.0.0.0/0
                type: distinct
          limit:
            requests: 15
            unit: Hour
---
kind: BackendTrafficPolicy
metadata:
  name: serviceB-ratelimit-all-ips
spec:
  targetRef:
    kind: GRPCRoute
    name: serviceB
    namespace: default
  rateLimit:
    type: Global
    global:
      rules:
        - clientSelectors:
            - sourceCIDR:
                value: 0.0.0.0/0
                type: distinct
          limit:
            requests: 300
            unit: Hour
```

*Environment*:
v0.6.0 envoy gateway release + rate limiter deployed with redis 

*Logs*:

shawnh2:
**What type of PR is this?**

Fixes #2225


internal/xds/translator/ratelimit.go
```diff
@@ -138,16 +137,26 @@ func patchRouteWithRateLimit(xdsRouteAction *routev3.RouteAction, irRoute *ir.HT
...
 	// Rules are ORed
 	for rIdx, rule := range global.Rules {
-		rlActions := []*routev3.RateLimit_Action{}
+		var rlActions []*routev3.RateLimit_Action
 		// Matches are ANDed
 		for mIdx, match := range rule.HeaderMatches {
 			// Case for distinct match
 			if match.Distinct {
 				// Setup RequestHeader actions
-				descriptorKey := getRateLimitDescriptorKey(descriptorPrefix, rIdx, mIdx)
```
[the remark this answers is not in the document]
shawnh2: in the original code, `descriptorPrefix` sure is the correct route name. 

the question is: in the CIDR match logic, there's no `descriptorPrefix` for MaskedRemoteAddress descriptor. 

https://github.com/envoyproxy/gateway/blob/b816e34dd1b9b2c3caf70faf4a58114668475e23/internal/xds/translator/ratelimit.go#L357-L361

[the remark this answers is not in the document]
shawnh2: so if we apply [this config](https://github.com/envoyproxy/gateway/issues/2225#issuecomment-1820190562), we will get xds config:

```yaml
domain: first-listener
descriptors:
  - key: masked_remote_address
    value: 192.168.0.10/32
    rate_limit:
      requests_per_unit: 15
      unit: HOUR
      unlimited: false
      name: ""
      replaces: []
    descriptors: []
    shadow_mode: false
    detailed_metric: false

---

domain: first-listener
descriptors:
  - key: masked_remote_address
    value: 192.168.0.10/32
    rate_limit:
      requests_per_unit: 300
      unit: HOUR
      unlimited: false
      name: ""
      replaces: []
    descriptors: []
    shadow_mode: false
    detailed_metric: false
```

in ratelimit, these two config will generate exact same key, which is `first-listener.masked_remote_address.192.168.0.10/32`. lack of any route info, causing error.

arkodg:
hey @zhaohuabing can you help with this review ?

internal/xds/translator/ratelimit.go
```diff
@@ -138,16 +137,26 @@ func patchRouteWithRateLimit(xdsRouteAction *routev3.RouteAction, irRoute *ir.HT
 }
 
 func buildRouteRateLimits(descriptorPrefix string, global *ir.GlobalRateLimit) []*routev3.RateLimit {
-	rateLimits := []*routev3.RateLimit{}
+	var rateLimits []*routev3.RateLimit
+
+	routeDescriptor := &routev3.RateLimit_Action{
```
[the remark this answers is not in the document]
zhaohuabing: A route-level descriptor has been added to avoid name conflicts.  We could possibly only modify the descriptors for the source IP condition, but I like the unified approach in this PR.

zhaohuabing:
@shawnh2  Well done! Thanks for the fixing.
Could you please add a few lines of comments in the code to explain the structure of the descriptors? It will help those who come across it later.

internal/xds/translator/testdata/out/ratelimit-config/empty-header-matches.yaml
```diff
@@ -1,13 +1,19 @@
...
-  - key: first-route-key-rule-0-match--1
-    value: first-route-value-rule-0-match--1
-    rate_limit:
-      requests_per_unit: 5
-      unit: SECOND
-      unlimited: false
-      name: ""
-      replaces: []
-    descriptors: []
+  - key: key-first-route
```
zhaohuabing, line 3: slightly prefer just `first-route`, no `key` or `value` prefix.

internal/xds/translator/testdata/out/ratelimit-config/empty-header-matches.yaml
```diff
@@ -1,13 +1,19 @@
...
+    value: value-first-route
+    rate_limit: null
+    descriptors:
+      - key: key-rule-0-match--1
```
zhaohuabing, line 7: The same as above, slightly prefer just the rule and match index part, no key or value prefix.

internal/xds/translator/testdata/out/ratelimit-config/empty-header-matches.yaml
shawnh2: +1, it's better to remove redundant `key` and `value` prefix, since the descriptor key and value are already speaking for itself.

internal/xds/translator/testdata/out/ratelimit-config/empty-header-matches.yaml
shawnh2: removed

zhaohuabing (approved):
LGTM. Would appreciate it if some comments could be added to explain the layered descriptors.

zirain (commented):
LGTM, defer to @arkodg

internal/xds/translator/ratelimit.go
```diff
@@ -179,7 +190,7 @@ func buildRouteRateLimits(descriptorPrefix string, global *ir.GlobalRateLimit) [
 						},
 					},
 				}
-				rlActions = append(rlActions, action)
+				rlActions = append(rlActions, routeDescriptor, action)
```
arkodg, line 193: Are we appending the `routeDescriptor` multiple times in the for loop ?

shawnh2: yes. the `routeDescriptor` is also a `RateLimit_Action` type, should be append before each Header/CIDR match action:
https://github.com/envoyproxy/gateway/blob/6ec8afd85733b61daf0df8dc2aa10a9fb5da5239/internal/xds/translator/ratelimit.go#L151

arkodg (approved):
LGTM thanks for fixing this bug

PR #2241, opened 2023-11-27
feat: add option in chart to add hpa to gateway

gazal-k:
re #703

**What type of PR is this?**

Fixes #


charts/gateway-helm/templates/envoy-gateway-hpa.yaml
```diff
@@ -0,0 +1,44 @@
...
+  {{- with .Values.autoscaling.annotations }}
+  annotations: {{ toYaml . | nindent 4 }}
+  {{- end }}
+  labels:
+    control-plane: envoy-gateway
+  {{- include "eg.labels" . | nindent 4 }}
+  name: envoy-gateway
+  namespace: '{{ .Release.Namespace }}'
+spec:
+  scaleTargetRef:
```
shawnh2, line 14: i don't think this is the target for HPA.

ref: https://github.com/envoyproxy/gateway/issues/703#issuecomment-1827083833

Issue #2251, opened 2023-11-29
E2E: flaky BasicAuth/per_route_configuration_second_route test


shawnh2:
flaky e2e test case: https://github.com/envoyproxy/gateway/actions/runs/7031850497/job/19134515984?pr=2234#step:6:1714

zirain:
@zhaohuabing PTAL

PR #2253, opened 2023-11-29
fix flaky test

zhaohuabing:
fix https://github.com/envoyproxy/gateway/issues/2251#issuecomment-1831853013

### DIFF
--- a/test/e2e/tests/basic-auth.go
+++ b/test/e2e/tests/basic-auth.go
@@ -38,14 +38,14 @@ var BasicAuthTest = suite.ConformanceTest{
 	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
 		t.Run("valid username password", func(t *testing.T) {
 			ns := "gateway-conformance-infra"
-			routeNN := types.NamespacedName{Name: "http-with-basic-auth", Namespace: ns}
+			routeNN := types.NamespacedName{Name: "http-with-basic-auth-1", Namespace: ns}
 			gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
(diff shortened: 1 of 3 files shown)


test/e2e/tests/basic-auth.go
```diff
@@ -38,14 +38,14 @@ var BasicAuthTest = suite.ConformanceTest{
 	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
 		t.Run("valid username password", func(t *testing.T) {
 			ns := "gateway-conformance-infra"
-			routeNN := types.NamespacedName{Name: "http-with-basic-auth", Namespace: ns}
+			routeNN := types.NamespacedName{Name: "http-with-basic-auth-1", Namespace: ns}
 			gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
 			gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
-			SecurityPolicyMustBeAccepted(t, suite.Client, types.NamespacedName{Name: "basic-auth", Namespace: ns})
-			SecurityPolicyMustBeAccepted(t, suite.Client, types.NamespacedName{Name: "basic-auth-2", Namespace: ns})
```
shawnh2, line 45: curious about this fix, is [this](https://github.com/envoyproxy/gateway/actions/runs/7031850497/job/19134515984?pr=2234#step:6:1714) becasue the request has been sent before the SecurityPolicy got accepted?

zhaohuabing: Not sure this is the reason, but we should check that. Ideally, the test HTTP requests should be sent after the xDS has been pushed to Envoy, but we don't have a condition in SecurityPolicy to check that yet.

shawnh2 (approved):
lgtm

zirain (approved):

PR #2257, opened 2023-11-30
feat: add HorizontalPodAutoscaler support for EnvoyProxy API

ardikabs:
**What type of PR is this?**
Add Horizontal Pod Autoscaler support for the Envoy Proxy deployment.

Here are the summaries of these changes:
- Introduce the `envoyHpa` field under EnvoyProxy API to enable Horizontal Pod Autoscaler.
- Move from `reflect.DeepEqual` to `cmp.Equal` with the `IgnoreField` helper for `replicas` field once HorizontalPodAutoscaler was used while comparing the live object of the EnvoyProxy Deployment spec and the generated Envoy Proxy Deployment spec. 
- Add a new method on the `ResourceRender` interface, named `HorizontalPodAutoscaler()` to fetch the HPA object from renderer if any.

Address #703


api/v1alpha1/shared_types.go
```diff
@@ -275,3 +276,39 @@ const (
...
+	// +optional
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// maxReplicas is the upper limit for the number of replicas to which the autoscaler can scale up.
+	// It cannot be less that minReplicas. It defaults to 1 replica.
+	// See k8s.io.autoscaling.v2.HorizontalPodAutoScalerSpec
+	//
+	// +optional
+	// +kubebuilder:default=1
+	MaxReplicas int32 `json:"maxReplicas,omitempty"`
```
shawnh2, line 295:
```suggestion
	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```

ardikabs: changed

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -224,5 +226,46 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
...
+	}
+
+	hpa := &autoscalingv2.HorizontalPodAutoscaler{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "autoscaling/v2",
+			Kind:       "HorizontalPodAutoscaler",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: r.Namespace,
+			Name:      r.Name(),
```
shawnh2, line 255: i think it's necessary to make `labels` configurable, not so rush in this PR, we can do it in the future

internal/infrastructure/kubernetes/proxy/resource_provider_test.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
+
```
shawnh2, line 19:
```suggestion
```

internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -74,8 +78,52 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, r ResourceRender)
...
-		return !reflect.DeepEqual(deployment.Spec, current.Spec)
+		return !cmp.Equal(current.Spec, deployment.Spec, opts...)
+	})
+}
+
+// configureHPA configures HorizontalPodAutoscaler object in the kube api server based on the provided ResourceRender,
+// there are two operations on this method:
+// 1. if an HPA is defined and it doesn't exist then creates, or otherwise updates
+// 2. if an HPA is not defined then delete the object irrespective of its existence
+func (i *Infra) configureHPA(ctx context.Context, r ResourceRender) error {
```
shawnh2, line 100: let's still call it `createOrUpdateHPA` for consistency

api/v1alpha1/shared_types.go
```diff
@@ -275,3 +276,38 @@ const (
 	// https://github.com/google/re2/wiki/Syntax.
 	StringMatchRegularExpression StringMatchType = "RegularExpression"
 )
+
+// KubernetesHorizontalPodAutoscalerSpec defines Kubernetes Horizontal Pod Autoscaler settings of Envoy Proxy Deployment
+type KubernetesHorizontalPodAutoscalerSpec struct {
```
zirain, line 281: I don't think this's the right direction, at last, you need borrow everything from `k8s.io.autoscaling.v2.HorizontalPodAutoScalerSpec`.

internal/infrastructure/kubernetes/infra_resource.go
ardikabs: but the operation also include deletion when needed, like omitting `envoyHpa`, wdyt? @shawnh2

api/v1alpha1/shared_types.go
ardikabs: if we still provide scaleTargetRef, then technically user could specify unrelated resource than Envoy Proxy Deployment

api/v1alpha1/envoyproxy_types.go
```diff
@@ -127,6 +127,15 @@ type EnvoyProxyKubernetesProvider struct {
 	// +kubebuilder:validation:XValidation:message="loadBalancerIP can only be set for LoadBalancer type",rule="!has(self.loadBalancerIP) || self.type == 'LoadBalancer'"
 	// +kubebuilder:validation:XValidation:message="loadBalancerIP must be a valid IPv4 address",rule="!has(self.loadBalancerIP) || self.loadBalancerIP.matches(r\"^(((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\\.|$)){4})\")"
 	EnvoyService *KubernetesServiceSpec `json:"envoyService,omitempty"`
+
+	// EnvoyHpa defines the Horizontal Pod Autoscaler settings for Envoy Proxy Deployment.
+	// Once the HPA is being set, Replicas field from EnvoyDeployment will be ignored.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:message="minReplicas must be greater than 0",rule="!has(self.minReplicas) || self.minReplicas > 0"
```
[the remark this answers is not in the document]
ardikabs: added

api/v1alpha1/envoyproxy_types.go
```diff
@@ -127,6 +127,15 @@ type EnvoyProxyKubernetesProvider struct {
...
+	// +kubebuilder:validation:XValidation:message="maxReplicas must be greater than 0",rule="!has(self.maxReplicas) || self.maxReplicas > 0"
+	// +kubebuilder:validation:XValidation:message="maxReplicas cannot be less than minReplicas",rule="!has(self.minReplicas) || self.maxReplicas >= self.minReplicas"
```
zhaohuabing, line 137: Should maxReplicas > self.minReplicas ?

api/v1alpha1/shared_types.go
```diff
@@ -275,3 +276,35 @@ const (
...
+	// can scale down. It defaults to 1 replica.
+	//
+	// +optional
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// maxReplicas is the upper limit for the number of replicas to which the autoscaler can scale up.
+	// It cannot be less that minReplicas. It defaults to 1 replica.
+	//
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```
zhaohuabing, line 293: Should the MaxReplicas be mandatory?

api/v1alpha1/envoyproxy_types.go
ardikabs: this is up to EG implementation, as per the Kubernetes HPA, it allows `maxReplicas == minReplicas`, hence I used `maxReplicas >= minReplicas`, thought @zhaohuabing ?

api/v1alpha1/shared_types.go
ardikabs: what about both `minReplicas` and `maxReplicas` be mandatory?

internal/infrastructure/kubernetes/infra_resource.go
shawnh2: sure, lets keep it

api/v1alpha1/envoyproxy_types.go
zhaohuabing: Are there any specific cases where maxReplicas should be equal to mixReplicas? If no, this restriction should be enforced.

api/v1alpha1/shared_types.go
zhaohuabing: I have no strong opinion on `minReplicas`, but at least `maxReplicas` should be mandatory.

ardikabs: i have no concern, so let me update it

api/v1alpha1/envoyproxy_types.go
zhaohuabing: How about making maxReplicas > self.minReplicas?  I doubt there won't be any cases requiring that these two be the same value, but we can always relax this restriction later if any users report that they need it.

ardikabs: agreed

zhaohuabing (approved):
LGTM, Thanks.

internal/infrastructure/kubernetes/proxy/resource_provider.go
Xunzhuo: Can you tell the scenarios ?

internal/infrastructure/kubernetes/infra_resource.go
Xunzhuo: Why not move it into `deleteHPA`  lline 203 - 205 ?

ardikabs: Oh, hold on, I just noticed, what if I implement this in the deleteHPA function? Do you think it would be suitable, @Xunzhuo ?

```go
func (i *Infra) deleteHPA(ctx context.Context, r ResourceRender) error {
	hpa, err := r.HorizontalPodAutoscaler()
	if err != nil {
		return err
	}

	if hpa == nil {
		hpa = &autoscalingv2.HorizontalPodAutoscaler{
			ObjectMeta: metav1.ObjectMeta{
				Namespace: i.Namespace,
				Name:      r.Name(),
			},
		}
	}

	return i.Client.Delete(ctx, hpa)
}
```

Xunzhuo: Yes, I think this is more suitable, and configuteHPA can be focused on createOrUpdateHPA

ardikabs: make sense, let me update it @Xunzhuo

Xunzhuo (approved):
LGTM, I have tested locally and worked as expect.

internal/infrastructure/kubernetes/proxy/resource_provider.go
shawnh2: no specific scenarios on my mind right now. it's not a strong option, can be considered if someone runs into it in the future.

api/v1alpha1/shared_types.go
Xunzhuo: Current approach looks well, any better ideas, IMO HPA fields are hard to abstract.

shawnh2 (approved):
lgtm, thanks

Issue #1336, opened 2023-04-20
Support local ratelimit


zhaohuabing:
*Description*:
Support local ratelimt as described in envoy [Local rate limit]( https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter) doc.

Global ratelimit and local ratelimit are used in different scenarios:

* If you want to enforce a global access control policy to a particular resource, then global rate limiting should be used. A typical scenario is to set how often a user can access an API according to the user’s service level agreement. Docker Hub, for example, has a rate limiting policy for how often a user can pull images from its registy. Paid users have a higher quota than free users.

* Local ratelimit is very useful when proxy instances can be horizontally scaled out based on the client load. In this scenario, since each proxy instance get its own rate limit quota, the traffic that the fleet of envoy proxies can handle increase when more proxy instances are spun up.

arkodg:
+1 for this, the API could mimic/reuse existing data structures that exist from `GlobalRateLimit`
https://github.com/envoyproxy/gateway/blob/9d6d699e83b37f98f2eca554d68290234ae231c9/api/v1alpha1/ratelimitfilter_types.go#L53
The only issue I see is supporting the `Distinct` match case, which needs some upstream work
https://github.com/envoyproxy/envoy/issues/23351
cc @cpakulski @vikaschoudhary16

PR #706, opened 2022-11-07
Ratelimiting Design

arkodg:
Started a design doc highlighting the `WHAT` and `WHY`

Relates to https://github.com/envoyproxy/gateway/issues/670

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

PR #2258, opened 2023-11-30
feat: local rate limit

zhaohuabing:
Related issue: https://github.com/envoyproxy/gateway/issues/1336


api/v1alpha1/ratelimit_types.go
```diff
@@ -42,6 +52,26 @@ type GlobalRateLimit struct {
...
+type LocalRateLimit struct {
+
+	// Envoy requires a default rate limit to be set for each route.
+	// The other option is to set the default rate limit to unit32 max and set the
+	// fill interval to a short period, like 1 second. This will effectively make
+	// the default rate limit "unlimited".
+	// See https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting
+
+	// Default is the default rate limit if no rules match.
+	Default RateLimitValue `json:"default"`
```
zhaohuabing, line 65: Envoy requires a default rate limit to be set for each route.
The other option is to set the default rate limit to unit32 max and set the fill interval to a short period, like 1 second. This will effectively make the default rate limit "unlimited".

See https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting

api/v1alpha1/ratelimit_types.go
```diff
@@ -42,17 +54,45 @@ type GlobalRateLimit struct {
...
+	// See https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting
+
+	// Limit is the default rate limit for a route if no rules match.
+	// If a request does not match any of the rules, it is counted towards this limit.
+	//
+	// Note: Limit is applied per route. Even if a policy targets a gateway,
+	// each route in that gateway still has a separate rate limit bucket.
+	// For example, if a gateway has 2 routes, and the limit is 100r/s, then
+	// each route has its own 100r/s rate limit bucket.
+	Limit RateLimitValue `json:"default"`
```
zhaohuabing, line 73: **Why do we need this?** 
Envoy requires a rate limit bucket to be set for each route. If a request does not match any of the rules, it is counted towards the default limit. It's a two-level rate limit mechanism, a default bucket for the whole route, and separate buckets can be defined if more fine-grained rate limits are needed for specific headers or source IP ranges.

The other option is to set the default rate limit to unit32 max and set the fill interval to a very short period, like 1 second. This will effectively make the default rate limit "unlimited".

I'm inclined to add a `Default` bucket to both the Local and Global rate limits. The reasoning is that if the length of `ClientSelectors` of a `RateLimitRule` is 0, that `RateLimitRule` is already effectively a default rate limit for that route, but it's implicit and confusing to users, and what makes this even worse is that descriptors are processed sequentially in Envoy, so if this rule is not the first one in the rules, it won't take effect. Instead, adding a `Default` rate limit explicitly at the top level and requiring the length of `ClientSelectors` at least 1 results in a much more understandable and user-friendly UI.

If a user doesn't want a `Default` rate limit at the route level, we can make it optional and fall back to the unit32 max approach, but I prefer that the `Default` should be mandatory,  otherwise, the behavior of rate limit is undefined when the length of rules is 0.

See https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting

Need your input here: @arkodg @zirain @envoyproxy/gateway-maintainers

api/v1alpha1/ratelimit_types.go
```diff
@@ -42,17 +54,45 @@ type GlobalRateLimit struct {
...
 	// specific clients using attributes from the traffic flow.
 	// All individual select conditions must hold True for this rule
 	// and its limit to be applied.
-	// If this field is empty, it is equivalent to True, and
-	// the limit is applied.
 	//
-	// +optional
+	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=8
 	ClientSelectors []RateLimitSelectCondition `json:"clientSelectors,omitempty"`
```
zhaohuabing, line 97: I propose renaming `ClientSelectors` to `Matches` and `RateLimitSelectCondition` to `RateLimitMatch`

The rationale behind this suggestion is to align with the Gateway API's terminology and make it more intuitive for users. `Selector` is Envoy-specific, and using consistent terminology like `Matches` in both `RateLimit` and `HTTPRoute` enhances clarity for users interacting with the API.

internal/xds/translator/local_ratelimit.go
```diff
@@ -0,0 +1,299 @@
...
+	filterCfg := route.GetTypedPerFilterConfig()
+	if _, ok := filterCfg[localRateLimitFilter]; ok {
+		// This should not happen since this is the only place where the filter
+		// config is added in a route.
+		return fmt.Errorf(
+			"route already contains local rate limit filter config:  %s",
+			route.Name)
+	}
+
+	localRl := &localrlv3.LocalRateLimit{
```
zirain, line 149: `always_consume_default_token_bucket` is enable?

zhaohuabing: Yes, the default value is true.  
I believe it's reasonable because the route-level bucket is for all traffic going through that route. But we may need to add a line of comment in the API to explain this. WDYT @zirain @arkodg

zhaohuabing: The default value of Envoy is set to true.  I believe we may need to align the behaviors of both the local and global rate limit and set it to false. WDYT @zirain @arkodg

zirain: I don't know why envoy set default to ture, but I'm +1 on setting it to `false`

internal/xds/translator/local_ratelimit.go
```diff
@@ -168,6 +169,13 @@ func (*localRateLimit) patchRoute(route *routev3.Route, irRoute *ir.HTTPRoute) e
 			},
 		},
 		Descriptors: descriptors,
+		// By setting AlwaysConsumeDefaultTokenBucket to false, the descriptors
```
zirain, line 172: do you think we should highlight this in design/user doc?

api/v1alpha1/ratelimit_types.go
```diff
@@ -9,27 +9,39 @@ package v1alpha1
...
 	GlobalRateLimitType RateLimitType = "Global"
+
+	// LocalRateLimitType allows the rate limits to be applied on a per Envoy
+	// proxy instance basis.
+	LocalRateLimitType RateLimitType = "Local"
 )
 
 // GlobalRateLimit defines global rate limit configuration.
 type GlobalRateLimit struct {
+	// TODO zhaohuabing add default rate limit here.
```
arkodg, line 43: no TODOS in APIs, since this will show up in the API docs

api/v1alpha1/ratelimit_types.go
arkodg: this was a long discussion, the original API I proposed used `matches` for the exact reason you are suggesting but others wanted to highlight the distinction, more in https://github.com/envoyproxy/gateway/pull/706

api/v1alpha1/ratelimit_types.go
```diff
@@ -42,17 +54,45 @@ type GlobalRateLimit struct {
...
 type RateLimitRule struct {
 	// ClientSelectors holds the list of select conditions to select
 	// specific clients using attributes from the traffic flow.
 	// All individual select conditions must hold True for this rule
 	// and its limit to be applied.
-	// If this field is empty, it is equivalent to True, and
-	// the limit is applied.
 	//
-	// +optional
+	// +kubebuilder:validation:MinItems=1
```
arkodg, line 95: this needs to be reverted, empty is valid, and is used to rate limit all traffic

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +133,7 @@ const (
 	SourceMatchExact SourceMatchType = "Exact"
 	// SourceMatchDistinct Each IP Address within the specified Source IP CIDR is treated as a distinct client selector
 	// and uses a separate rate limit bucket/counter.
+	// Note: This is only supported for Global Rate Limits.
```
arkodg, line 136: I suggest creating separate fields for `Local` and `Global` or adding CEL validation we cannot support something like `Distinct` for local

api/v1alpha1/ratelimit_types.go
```diff
@@ -42,17 +54,45 @@ type GlobalRateLimit struct {
 	Rules []RateLimitRule `json:"rules"`
 }
 
+// LocalRateLimit defines local rate limit configuration.
+type LocalRateLimit struct {
+
+	// Envoy requires a default rate limit to be set for each route.
```
arkodg, line 60: this is not relevant to the end user, can this be rewritten with the app dev in mind ?

api/v1alpha1/ratelimit_types.go
arkodg: https://gateway.envoyproxy.io/v0.6.0/user/rate-limit/#rate-limit-all-requests highlights how you can have a default ratelimit for all requests, but also a specific rate limit for specific clients

arkodg: we may need something like `The other option is to set an implicit default rate limit as unit32 max and set the fill interval to a very short period, like 1 second. This will effectively make the default rate limit "unlimited".` if a default limit value is required, but we must ensure that it is truly `unlimited`

zhaohuabing: > https://gateway.envoyproxy.io/v0.6.0/user/rate-limit/#rate-limit-all-requests

Yes, the current approach also works, but I prefer an explicit "global" setting because

* Explict approach is confusing to the users.
* Descriptors are processed sequentially in Envoy, so if this "global" rule is not the first one in the rules, it won't take effect.

api/v1alpha1/ratelimit_types.go
zhaohuabing: This comment is just for EG developers. It won't show up in the generated CRD yaml and md file.

api/v1alpha1/ratelimit_types.go
zhaohuabing: This line won't show up in the API docs since there's a blank line between it and the rest of the code.

api/v1alpha1/ratelimit_types.go
zhaohuabing: I am inclined to add a CEL validation. It's simpler.

api/v1alpha1/ratelimit_types.go
zhaohuabing: If a "global" token bucket is added at the top level, then it would be invalid.

api/v1alpha1/ratelimit_types.go
zhaohuabing: I may missed it but I didn't find the discussion around `matches` vs `selectors`. Could you please point me to the comments? Thanks!

I'm not sure how it sounds to native speakers, but from a non-native speaker's perspective, I was really bothered by the term `Selector` when I first looked at the rate limit API.

api/v1alpha1/ratelimit_types.go
arkodg: it should then probably live in the design doc or translation lib, not in the API file

api/v1alpha1/ratelimit_types.go
zhaohuabing: I tried to use validate it with the below cel rule: 
```
// +kubebuilder:validation:XValidation:rule="self.rules.all(r, has(r.clientSelectors)? (r.clientSelectors.all(s, has(s.sourceCIDR) && has(s.sourceCIDR.type)? s.sourceCIDR.type!='Distinct' : true)) : true)", message="Distinct SourceMatchType is not supported for Local Rate Limit"
```
And verified it on  https://playcel.undistro.io/.

However,  k8s throws the below error, complaining it's invalid.

```
        	            	unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "backendtrafficpolicies.gateway.envoyproxy.io": CustomResourceDefinition.apiextensions.k8s.io "backendtrafficpolicies.gateway.envoyproxy.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[rateLimit].properties[local].properties[rules].x-kubernetes-validations[0].rule: Invalid value: apiextensions.ValidationRule{Rule:"self.rules.all(r, has(r.clientSelectors)? (r.clientSelectors.all(s, has(s.sourceCIDR) && has(s.sourceCIDR.type)? s.sourceCIDR.type!='Distinct' : true)) : true)", Message:"Distinct SourceMatchType is not supported for Local Rate Limit", MessageExpression:""}: compilation failed: ERROR: <input>:1:5: type 'list_type:{elem_type:{message_type:"selfType61480208.@idx"}}' does not support field selection
        	            	 | self.rules.all(r, has(r.clientSelectors)? (r.clientSelectors.all(s, has(s.sourceCIDR) && has(s.sourceCIDR.type)? s.sourceCIDR.type!='Distinct' : true)) : true)
```

api/v1alpha1/ratelimit_types.go
```diff
@@ -9,35 +9,54 @@ package v1alpha1
...
 type GlobalRateLimit struct {
+	// Rules are a list of RateLimit selectors and limits. Each rule and its
+	// associated limit is applied in a mutually exclusive way. If a request
+	// matches multiple rules, each of their associated limits get applied, so a
+	// single request might increase the rate limit counters for multiple rules
+	// if selected. The rate limit service will return a logical OR of the individual
+	// rate limit decisions of all matching rules. For example, if a request
+	// matches two rules, one rate limited and one not, the final decision will be
+	// to rate limit the request.
+	// +kubebuilder:validation:MaxItems=16
```
zhaohuabing, line 51: https://github.com/envoyproxy/ratelimit?tab=readme-ov-file#example-2

api/v1alpha1/ratelimit_types.go
```diff
@@ -9,35 +9,58 @@ package v1alpha1
...
+	Rules []RateLimitRule `json:"rules"`
+}
+
+// LocalRateLimit defines local rate limit configuration.
+type LocalRateLimit struct {
+	// Rules are a list of RateLimit selectors and limits. If a request matches
+	// multiple rules, the strictest limit is applied. For example, if a request
+	// matches two rules, one with 10rps and one with 20rps, the final limit will
+	// be based on the rule with 10rps.
```
zhaohuabing, line 61: Originally from https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting

api/v1alpha1/ratelimit_types.go
```diff
@@ -49,8 +72,14 @@ type RateLimitRule struct {
...
 	// and its limit to be applied.
-	// If this field is empty, it is equivalent to True, and
-	// the limit is applied.
+	//
+	// If no client selectors are specified, the rule applies to all traffic of
+	// the targeted Route.
+	//
+	// If the policy targets a Gateway, the rule applies to each Route of the Gateway.
+	// Please note that each Route has its own rate limit counters. For example,
+	// if a Gateway has two Routes, and the policy has a rule with limit 10rps,
```
arkodg, line 81: not true, limit will be 10rps for each and every route

arkodg (approved):
LGTM thanks !

zirain (approved):
LGTM, let's finish user-face doc with following PRs.

PR #2259, opened 2023-12-01
[Helm chart] Refactoring Kubernetes labels

nlamirault:
**What type of PR is this?**
 

- Support [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
- Support additional labels for all resources

Fixes #


charts/gateway-helm/templates/_helpers.tpl
```diff
@@ -39,7 +39,12 @@ helm.sh/chart: {{ include "eg.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
+app.kubernetes.io/component: gateway-api
```
shawnh2, line 42: i am not sure whether refer `gateway-api` as a component is accurate, any thoughts? 

zirain: IMO, default shouldn't

zirain: eg is built on top of gateway-api and envoy.

qicz: agree

nlamirault: do you have an idea for the component label ?

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -1,3 +1,7 @@
+## Additional labels to add to all resources
+additionalLabels: {}
```
arkodg, line 2: any idea if other projects also call this top level field `additionalLabels`?

charts/gateway-helm/templates/leader-election-rbac.yaml
```diff
@@ -44,7 +45,8 @@ metadata:
   name: {{ include "eg.fullname" . }}-leader-election-rolebinding
   namespace: '{{ .Release.Namespace }}'
   labels:
-  {{- include "eg.labels" . | nindent 4 }}
+    {{- include "eg.labels" . | nindent 4 }}
```
arkodg, line 48: looks like the idententation is changing for all these files, was it broken earlier ?

arkodg:

charts/gateway-helm/templates/leader-election-rbac.yaml
nlamirault: it is to add indentation to `app.kubernetes.io/component: rbac`

charts/gateway-helm/templates/_helpers.tpl
nlamirault: any idea ?

PR #2260, opened 2023-12-01
helm: add ServiceMonitor resource

nlamirault:
**What type of PR is this?**

Support `ServiceMonitor` resource to monitoring Envoy Gateway using the Prometheus Operator

Fixes #


nlamirault:
sure @zirain it was a mistake.

Xunzhuo (commented):
IMO, it is better to move this into addons or examples.

zirain:
> IMO, it is better to move this into addons or examples.

we can add a new helm chart for all these things, and it won't break the `core`.

nlamirault:
@zirain a new chart just for ServiceMonitor resource ?

charts/gateway-helm/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.serviceMonitor.enabled }}
```
guydc, line 1: Do we want an extra layer of protection with something like `{{ if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" -}}` ?

guydc:
> new chart just for ServiceMonitor resource ?

I think that other projects also take this approach: https://github.com/istio/istio/blob/4f54ef1ba6ff4608841e76c5233827322e683a86/samples/addons/extras/prometheus-operator.yaml#L48

+1 for an addons chart that will contain additional observability examples, like the ones here: https://github.com/envoyproxy/gateway/tree/main/examples/.

zirain:
I understand that why users want add this, but I do really don't want this be part of part `core` chart. 

Option 1: be part of examples.
Option 2: part of an new addon chart(maybe it's a little early)

arkodg:
adding some labels to this, so we prioritize making a decision here

arkodg:
should this be a add-on chart (along with prom and grafana with no EG resource) or a demo/extended chart (with EG as a subchart) ?

zirain:
> should this be a add-on chart (along with prom and grafana with no EG resource) or a demo/extended chart (with EG as a subchart) ?

either is better than current.

shawnh2:
Since #3470 landed, what is the status of this PR? Do we still need this?

arkodg:
@shawnh2 as part of https://github.com/envoyproxy/gateway/pull/3470, did we need to define a ServiceMonitor to ensure Prometheus can scrape properly ?

shawnh2:
> @shawnh2 as part of #3470, did we need to define a ServiceMonitor to ensure Prometheus can scrape properly ?

The `ServiceMonitor` is used for [Prometheus-operator](https://github.com/prometheus-operator/prometheus-operator), since we do not support this operator now, we can include this as a option in `gateway-addons-helm`.

arkodg:
closing this PR since its become inactive, the ServiceMonitor should be available in the `gateway-addons-helm` chart now
the `

charts/gateway-helm/templates/servicemonitor.yaml
nlamirault: done

Issue #2265, opened 2023-12-05
OOM killed


qicz:
*Description*:

watch some HTTPRoute that has some error. maybe the service does not exist. the EG has been killed due to  `Reconcile` them.

*Logs*:
![image](https://github.com/envoyproxy/gateway/assets/2174082/3caf06e7-a8a0-4574-9557-861733d4b90c)

qicz:
IMO, should set `RequeueAfter` to requeue

Xunzhuo:
I did not reproduce it, can you provide the steps to reproduce it ? @qicz

Xunzhuo:
Tried that and just HTTPRoute reported BackendNotFound, the eg works still well

qicz:
> Tried that and just HTTPRoute reported BackendNotFound, the eg works still well

 this report too often and there are more invalid HTTPRoute, the EG has been killed due to Reconcile them.

zzjin:
@qicz I'm facing same error here.
But My usage is setting about ~1300 `HTTPRoute` CR with about ~20 `Gateway` with `mergeGateway=true`. 
![image](https://github.com/envoyproxy/gateway/assets/679757/40ccbafc-9eea-4759-811d-5df50709635e)


May be it's not `non-exists` backends cause eg oom, but the count of gateway api crs cause eg oom, I'm facing that deployment `envoy-gateway` pod eats too many memory.
![image](https://github.com/envoyproxy/gateway/assets/679757/2e0c5d32-36bf-4e59-92ab-f934d667f4fd)
default eg memory limit is 1g, you can change this to unlimited, but the problem is still the problem.

arkodg:
@qicz in your logs, can you please paste the entire log showing the namespace and name of service, along with `kubectl` info on the service as well the httproute that is linking to it ?

qicz:
> @qicz in your logs, can you please paste the entire log showing the namespace and name of service, along with `kubectl` info on the service as well the httproute that is linking to it ?

@arkodg sorry reply slowly. the namespace and service are from my company app, so they have been cleared by me.

qicz:
[...]

PR #2275, opened 2023-12-06
fix: return error on failed requests in controller

cnvergence:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/2265


zhaohuabing:
Looks like the Gateway API conformance test allows non-existent backend:

        	            				/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.0/conformance/utils/suite/suite.go:213
        	            	context deadline exceeded
        	Test:       	TestGatewayAPIConformance/HTTPRouteInvalidNonExistentBackendRef

zhaohuabing (changes requested):
EG should continue reconciling if some backends are missing. This behavior is allowed according to Gateway API conformance tests.  And it's reasonable because EG should try its best to keep the Gateway functioning.

cnvergence:
Seems like I was wrong with my idea

qicz:
> EG should continue reconciling even if some backends are missing. This behavior is required by the Gateway API conformance tests. And it's reasonable because EG should try its best to keep the Gateway functioning.

IMO, delaying to requeue better for the EG memory consumption

PR #2280, opened 2023-12-07
ci: add PR comment and cc reviewers

Xunzhuo:
**What type of PR is this?**

We tried a couple of ways to auto request reviews from @envoyproxy/gateway-reviewers, but it failed, and I tried it too today, failed again.

So current way is to add a welcome comment to cc @envoyproxy/gateway-maintainers and @envoyproxy/gateway-reviewers, they should get the ping too.

### DIFF
--- a/.github/workflows/welcome.yaml
+++ b/.github/workflows/welcome.yaml
@@ -26,7 +26,7 @@ jobs:
             - Your PR needs to receive at least two approvals.
             - At least one approval must come from a member of the gateway-maintainers team.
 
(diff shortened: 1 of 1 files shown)


Xunzhuo:
demo https://github.com/uestcer/community/pull/191

zirain:
I have no objection to this, but not sure this's good to go, @envoyproxy/gateway-maintainers PTAL.

Xunzhuo:
@shawnh2 I like the idea.

> BTW, if you really want to add this, please mention that don't use force push like envoy does for better review.

sure @zirain

zirain (approved):

Issue #2282, opened 2023-12-07
MergeGateways: nil pointer dereference


guydc:
*Description*:
When using using the `MergeGateways` feature, the K8s provider's Deployment and Service reconciliations fail after restarting Envoy Gateway. The implementation assumes that the owning GatewayClass of these resources is already present in the resource tree: [service](https://github.com/envoyproxy/gateway/blob/52af65e2d35d6d8cbd02e3b8257d1784b0f31dd6/internal/provider/kubernetes/predicates.go#L183C26-L183C26), [deployment](https://github.com/envoyproxy/gateway/blob/52af65e2d35d6d8cbd02e3b8257d1784b0f31dd6/internal/provider/kubernetes/predicates.go#L316). It's possible that during the initial reconciliation of core resources, the Gateway API resources tree is not yet synced. 

*Repro steps*:
Apply the [quickstart config](https://github.com/envoyproxy/gateway/releases/download/latest/quickstart.yaml) with the following modifications to Gateway and EnvoyProxy resources: 

```
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: envoy-proxy
    namespace: envoy-gateway-system
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  namespace: envoy-gateway-system
  name: envoy-proxy
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        type: NodePort
  mergeGateways: true
```

Wait for envoy deployment and service to be ready. Restart Envoy Gateway and observe panic. 

*Environment*:
K8s: kind cluster, v1.27.3
Envoy Gateway: v0.0.0-latest

*Logs*:
```
[...]
runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 362 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1045d6420?, 0x105fca6c0})
        /workspace/go/src/github.com/envoyproxy/gateway/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x7c
[...]

Issue #2029, opened 2023-10-20
Add E2E for the MergeGateways feature


arkodg:
*Description*:
>Describe the issue.

Add an E2E in https://github.com/envoyproxy/gateway/tree/main/test/e2e for the MergeGateway features
Relates to https://github.com/envoyproxy/gateway/pull/1938

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

shawnh2:
it seems a little tricky to implement this e2e test case.

our e2e test env deploys eg with controller name:
https://github.com/envoyproxy/gateway/blob/c85f0e438ec3c41048739e794f27f3a33e853df0/test/config/gatewayclass.yaml#L6

if we have `mergeGateways` feature enabled, **it's better to create a new gatewayclass, not reuse the exist one**:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg-merge-gateways
spec:
  controllerName: gateway.envoyproxy.io/merge-gateways-controller  # needs another controller name
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
    namespace: envoy-gateway-system
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
  namespace: envoy-gateway-system
spec:
  mergeGateways: true  # enable
```

but that requires to deploy eg one more time, it's not worthy!

arkodg:
thanks for looking into this @shawnh2, maybe we cause pause on this until https://github.com/envoyproxy/gateway/issues/1231 is complete, which should make it easier to add and remove multiple gateway classes

PR #2283, opened 2023-12-08
fix: mergeGateways panics when restarting control plane

Xunzhuo:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/2282

### DIFF
--- a/internal/provider/kubernetes/predicates.go
+++ b/internal/provider/kubernetes/predicates.go
@@ -180,10 +180,13 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
 	if ok {
 		res, _ := r.resources.GatewayAPIResources.Load(gclass)
-		for _, gw := range res.Gateways {
-			gw := gw
-			r.statusUpdateForGateway(ctx, gw)
+		if res != nil && len(res.Gateways) > 0 {
+			for _, gw := range res.Gateways {
+				gw := gw
+				r.statusUpdateForGateway(ctx, gw)
+			}
 		}
+
 		return false
 	}
 
@@ -313,9 +316,11 @@ func (r *gatewayAPIReconciler) validateDeploymentForReconcile(obj client.Object)
 	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
 	if ok {
 		res, _ := r.resources.GatewayAPIResources.Load(gclass)
-		for _, gtw := range res.Gateways {
-			gtw := gtw
-			r.statusUpdateForGateway(ctx, gtw)
+		if res != nil && len(res.Gateways) > 0 {
+			for _, gw := range res.Gateways {
+				gw := gw
+				r.statusUpdateForGateway(ctx, gw)
+			}
 		}
 		return false
 	}


zirain:
can you add a (e2e)test for this to avoid regression in the future?

Xunzhuo:
Can this be a follow up ? @zirain

Xunzhuo:
Okay, let me update it later

Xunzhuo:
I still think we can start working on e2e after https://github.com/envoyproxy/gateway/issues/2029#issuecomment-1808623065

let us fix the bug first @zirain

zirain (approved):

internal/provider/kubernetes/predicates.go
```diff
@@ -179,7 +179,10 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 	// Only merged gateways will have this label, update status of all Gateways under found GatewayClass.
 	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
 	if ok {
-		res, _ := r.resources.GatewayAPIResources.Load(gclass)
+		res, ok := r.resources.GatewayAPIResources.Load(gclass)
+		if !ok {
+			return false
```
zzjin, line 184: If return false here, the service here(and deploy below) will always return false, witch means the `MergeGateways` do not work anymore.

Xunzhuo:
Let me double check it before merging it.

internal/provider/kubernetes/predicates.go
```diff
@@ -179,7 +179,10 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
...
+		}
 		for _, gw := range res.Gateways {
 			gw := gw
 			r.statusUpdateForGateway(ctx, gw)
```
cnvergence, line 188:
```suggestion
		}
	if res != nil && len(res.Gateways) > 0 {
		for _, gw := range res.Gateways {
			gw := gw
			r.statusUpdateForGateway(ctx, gw)
			}
		}
```

internal/provider/kubernetes/predicates.go
```diff
@@ -312,7 +315,10 @@ func (r *gatewayAPIReconciler) validateDeploymentForReconcile(obj client.Object)
...
 	gclass, ok := labels[gatewayapi.OwningGatewayClassLabel]
 	if ok {
-		res, _ := r.resources.GatewayAPIResources.Load(gclass)
+		res, ok := r.resources.GatewayAPIResources.Load(gclass)
+		if !ok {
+			return false
+		}
 		for _, gtw := range res.Gateways {
 			gtw := gtw
 			r.statusUpdateForGateway(ctx, gtw)
```
cnvergence, line 324:
```suggestion
		}
	if res != nil && len(res.Gateways) > 0 {
		for _, gw := range res.Gateways {
			gw := gw
			r.statusUpdateForGateway(ctx, gw)
			}
		}
```

internal/provider/kubernetes/predicates.go
cnvergence: we will need to add this check

cnvergence (changes requested):
@Xunzhuo ptal

internal/provider/kubernetes/predicates.go
```diff
@@ -179,10 +179,16 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
...
+		res, ok := r.resources.GatewayAPIResources.Load(gclass)
+		if !ok {
+			return false
+		}
+	if res != nil && len(res.Gateways) > 0 {
 		for _, gw := range res.Gateways {
 			gw := gw
 			r.statusUpdateForGateway(ctx, gw)
+			}
+		}
```
cnvergence, line 191: suggestion missed that
```suggestion

```

internal/provider/kubernetes/predicates.go
```diff
@@ -312,10 +318,16 @@ func (r *gatewayAPIReconciler) validateDeploymentForReconcile(obj client.Object)
...
+		if !ok {
+			return false
+		}
+	if res != nil && len(res.Gateways) > 0 {
+		for _, gw := range res.Gateways {
+			gw := gw
+			r.statusUpdateForGateway(ctx, gw)
+			}
+		}
 		}
```
cnvergence, line 331:
```suggestion
```

cnvergence: same

cnvergence (approved):

cnvergence (approved):
lgtm now

zirain (approved):

Issue #2170, opened 2023-11-09
Weighted BackendRefs not showing expected outcome


ardikabs:
*Description*:
I've got the intention to implement weighted backendRef for canary purposes. 
Following through this documentation https://gateway.envoyproxy.io/latest/user/http-traffic-splitting/#weighted-backendrefs, didn't show an expected outcome for the weighted, it means given the weight ratio between services, such as backend A set the weight to 8, then backend B set the weight to 2, the result of Envoy Proxy config (both RDS and LDS) doesn't show the expected implementation, which should be 8:2 ratio.

*Expectation*:
The expectation is similar to the doc mentioned which is
> The HTTPRoute below will configure the gateway to send 80% of the traffic to the backend service, and 20% to the backend-2 service.

*Repro steps*:
Assuming the backend services are ready as in `backend` and `backend-2` with 1 replica.

Then apply below `HTTPRoute`
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: http-headers
spec:
  parentRefs:
  - name: eg
  hostnames:
  - backends.example
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
      weight: 8
    - group: ""
      kind: Service
      name: backend-2
      port: 3000
      weight: 2
```

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
Latest / v0.6.0

*Logs*:

- RDS:
```yaml
dynamicRouteConfigs:
- lastUpdated: "2023-11-09T15:14:02.825Z"
  routeConfig:
    '@type': type.googleapis.com/envoy.config.route.v3.RouteConfiguration
    ignorePortInHostMatching: true
    name: envoy-gateway-system/eg/http
    virtualHosts:
    - domains:
      - backends.example
      name: envoy-gateway-system/eg/http/backends_example
      routes:
      - match:
          prefix: /
        name: httproute/envoy-gateway-system/http-headers/rule/0/match/0/backends_example
        route:
          cluster: httproute/envoy-gateway-system/http-headers/rule/0
  versionInfo: c0ca1f138a0b4e661716ad9b97eae569732c3b77b83d93ca93dbdc7c25daa9a9
```

- CDS:
```yaml
dynamicActiveClusters:
- cluster:
[...]

Issue #2125, opened 2023-10-31
Support Circuit Breaker


Xunzhuo:
*Description*:

Add Circuit Breaker support to BackendTrafficPolicy.

guydc:
@Xunzhuo - Can you assign this task to me?

zirain:
need user doc

Xunzhuo:
Track in https://github.com/envoyproxy/gateway/issues/2125

PR #2284, opened 2023-12-09
API: Support Circuit Breakers in BackendTrafficPolicy

guydc:
## API: Circuit Breakers

Relates to https://github.com/envoyproxy/gateway/issues/2125, and based on https://github.com/envoyproxy/gateway/issues/1821

## Overview

The [Backend Traffic Policy][] proposes support for Envoy's [Circuit Breaker][] configuration. Circuit breakers define distributed limits on the volume of requests and connections from Envoy to hosts in a [Cluster][], with the intention of applying back-pressure when the upstream fails or degrades. 

Envoy Proxy enables Circuit Breakers by default. Circuit Breakers cannot be completely disabled, but threshold values can be set to `MaxUint32`. The default threshold value in Envoy (1024) is not appropriate for high-throughput environments and can cause unexpected failures for users who are not aware of the default configuration. 

## Goals
* Propose API for Circuit Breakers. 
* Define the default Circuit Breaker settings.
* Allow users to customize commonly-used Circuit Breaker settings. 

## Design Decisions 
* Envoy Gateway enables Circuit Breakers by default.
  * If a `BackendTrafficPolicy` resource is not attached to `xRoute`/`Gateway` or does not specify threshold values, Envoy Gateway can use one of the following defaults:
    * Leave Circuit Breakers unset, effectively keeping the Envoy default threshold values.
    * Use opinionated values that are appropriate for typical Gateway scenarios.
    * "Disable" Circuit Breakers by setting the threshold values to `MaxUint32`.
  * Keeping the default Envoy values ensures greater resilience for most users. Advanced users with higher throughput requirements can adjust the settings.
  * Envoy metrics and access logs provide observability for Circuit Breaker overflow. Unexpected failures related to default circuit breaker overflow can be easily identified by Envoy Gateway users. Envoy Gateway troubleshooting documentation and Grafana dashboards can further assist in this area. 
* Advanced Circuit Breaking options are not supported, but the API definition will allow changes in the future. 
  * [Routing Priorities][] are currently not supported by Envoy Gateway. So, only a single Circuit Breaker `Thresholds` struct that translates to the `DEFAULT` priority is allowed.
  * [Retry Budgets][] are discussed in https://github.com/envoyproxy/gateway/pull/2168 and are not included. 
  * Per-Host Thresholds, Maximum Connection Pools and tracking of remaining resources are rarely supported by other projects. These settings are not included. 
* Per-Backend Circuit Breaker settings are not supported.
  * Some users may want and expect Circuit Breakers to apply per Backend, as mentioned [here](https://github.com/envoyproxy/gateway/issues/2170#issuecomment-1805221753). Per-Backend Circuit Breakers are better synchronized (within the scope of a single Envoy process) and react to failures quickly and efficiently.  
  * The Envoy Gateway [System Design][] maps `BackendRefs` to Envoy clusters. As a result, the same upstream backend can be represented by multiple Envoy clusters.
  * Backend Traffic Policies can only be attached to `Gateway` and `xRoute` resources at this time.
  * Support for Per-Backend settings requires further community discussion on Backend translation and Policy Attachment and considered out-of-scope for this PR. 
  
## API Example

```
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy-with-circuitBreakers
spec:
 [...]
  circuitBreakers: # optional
    thresholds: #optional
    - maxConnections: uint32 # optional, default=1024
      maxPendingRequests: uint32 # optional, default=1024
      maxParallelRequests: uint32 # optional, default=1024
      maxRetries: uint32 # optional, default=3
[...]    

```

[Backend Traffic Policy]: https://github.com/envoyproxy/gateway/issues/1821
[Circuit Breaker]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking
[System Design]: https://github.com/envoyproxy/gateway/blob/main/site/content/en/latest/design/system-design.md
[Cluster]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/upstream
[Routing Priorities]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-routingpriority
[Retry Budgets]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto.html#envoy-v3-api-msg-config-cluster-v3-circuitbreakers-thresholds-retrybudget


api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,43 @@
...
+	// +kubebuilder:validation:MaxItems:=1
+	// +optional
+	Thresholds []Thresholds `json:"thresholds,omitempty"`
+}
+
+type Thresholds struct {
+	// The maximum number of connections that Envoy will make to the referenced backend (per xRoute).
+	// Default: 1024
+	//
+	// +optional
```
shawnh2, line 22: better go with CEL for all these fields, for example:

```go
	// +kubebuilder:validation:Minimum=xxx
	// +kubebuilder:validation:Maximum=xxx
	// +kubebuilder:default=xxx
```

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -65,6 +65,12 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty"`
+
+	// Circuit Breaker settings for the upstream connections and requests.
+	// If not set, circuit breakers will be enabled with the highest supported thresholds
+	//
+	// +optional
+	CircuitBreakers *CircuitBreakers `json:"circuitBreakers,omitempty"`
```
thiagogcm, line 73: Why plural?

guydc: It's the term used by [Envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto), the [Backend Traffic Policy](https://github.com/envoyproxy/gateway/issues/1821) and other OSS projects like [Emissary Ingress](https://www.getambassador.io/docs/edge-stack/latest/topics/using/circuit-breakers#circuit-breakers). A single `thresholds` struct defines multiple breakers (requests, connection, etc. ), so plural form does sound appropriate.

Xunzhuo (commented):
Thanks for working on it, please fix the CI lint errors first

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -16,28 +16,35 @@ type CircuitBreakers struct {
...
+	// +kubebuilder:default=1024
 	// +optional
 	MaxConnections *uint32 `json:"maxConnections,omitempty"`
 
-	// The maximum number of pending requests that Envoy will allow to the referenced backend (per xRoute).
-	// Default: 1024
+	// The maximum number of pending requests that Envoy will queue to the referenced backend (per xRoute).
 	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
```
zirain, line 30: why is `4294967295` ?

shawnh2: this is the maximum value of `uint32`, but i think this `Maximum` validation can be optional

zirain: oh, my bad, think about `2147483647`.

zirain: can `-1` pass if the type is uint32?

shawnh2: no, it cannot

zirain: so these rules seem meaningless?

shawnh2: `Maximum` and `Minimum` sure seems meaningless, but can keep `default`, WDYT?

zirain: Yes, that's what it means.

guydc: Hi @zirain, @shawnh2.  Do note that the OpenAPI spec (used by K8s CRDs) doesn't really support unsigned ints: https://swagger.io/specification/. The controller-gen tools actually produce a schema that refers to these fields as `int32` in the generated CRD. The actual K8s API server behavior, from my limited check, is to treat these fields as `int64`. I think that the actual go type (`*uint32`) mostly impacts the unmarshalling done by client go. So, guaranteeing that the value stored is actually safe to cast to uint32 could be useful...

guydc: Another approach would be to use `int64` explicitly in the go types layer and have uint32 as a representation in the IR layer and downwards. The value range validation can occur either using the schema or during the IR translation. WDYT?

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,50 @@
...
+	// +optional
+	MaxRequests *uint32 `json:"maxParallelRequests,omitempty"`
+
+	// The maximum number of parallel retries that Envoy will allow to the referenced backend (per xRoute).
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:default=3
+	// +optional
+	MaxRetries *uint32 `json:"maxRetries,omitempty"`
```
tmsnan, line 49: `MaxRetries` is part of the retry and would be better included in the retry, the corresponding field is `maxParallel` #2168

tmsnan (changes requested):

api/v1alpha1/circuitbreaker_types.go
zirain: like Gateway API project, let's use `*int32` with valiation min and max?

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,50 @@
...
+package v1alpha1
+
+// CircuitBreakers defines the Circuit Breakers configuration.
+type CircuitBreakers struct {
+	// List of Circuit Breaker Thresholds
+	// At most one Thresholds resource is supported.
+	//
+	// +kubebuilder:validation:MaxItems:=1
+	// +optional
+	Thresholds []Thresholds `json:"thresholds,omitempty"`
```
zhaohuabing, line 15: As mentioned in the description of this PR, only one `Thresholds` is needed here because Envoy Gateway doesn't support [Routing Priorities](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-routingpriority)

api/v1alpha1/circuitbreaker_types.go
zhaohuabing: I believe EG should distinguish the `MaxConcurrentRetries` of a cluster and the `MaxRetries` of an individual request.

`MaxConcurrentRetries` belongs to the Circuit Breaker configuration, and `MaxRetries` belongs to the Retry configuration.

api/v1alpha1/backendtrafficpolicy_types.go
zhaohuabing: If we only need one `Thresholds` in the `CircuitBreakers` structure, the name should be singular.

api/v1alpha1/circuitbreaker_types.go
guydc: I kept this as list for two reasons:
- Future-proofing:  if/when EG does support routing priorities, there would be no need to add another list. We can just add another optional `priority` field to `Thresholds` and cancel the length validation. 
- Compatibility with the original proposal by @AliceProxy in BTP

We can decide to keep things simpler for the common use case of tweaking default circuit breakers, and, in the future, allow a list for non-default priorities. 

```
---
kind: BackendTrafficPolicy
metadata:
  name: example-policy-with-circuitbreakers
spec:
 [...]
  circuitBreakers: # optional
    maxConnections: uint32 # optional, default=1024
    maxPendingRequests: uint32 # optional, default=1024
    maxParallelRequests: uint32 # optional, default=1024
    maxRetries: uint32 # optional, default=3
    additionalThresholds: #optional [Future, not in this PR]
    - priority: [HIGH|...] 
      maxConnections: uint32 # optional, default=1024
      maxPendingRequests: uint32 # optional, default=1024
      maxParallelRequests: uint32 # optional, default=1024
      maxRetries: uint32 # optional, default=3
[...] 
``` 

If the maintainers support this I don't have any objection . WDYT?

api/v1alpha1/circuitbreaker_types.go
guydc: I agree. However, note that https://github.com/envoyproxy/gateway/pull/2168 deals with `RetryBudget` which is also a part of Envoy's Circuit Breaker settings. 

In Envoy, the separation of route and cluster settings is pretty clear. Multiple routes can point to the same cluster, and so `MaxRetries`, `RetryBudget` will apply to all the traffic coming from routes that share an upstream cluster. The motivation is to protect the upstream system from a retry storm. 

In Envoy Gateway, we have a cluster for each `xRoute`. So, it doesn't make much difference if these settings are managed under the `Retries` or the `CircuitBreakers` section. It is important that the users understand the implications of these settings - overflowing retries will be queued and later dropped.  

If in the future Envoy Gateway does support a notion of shared backends (e.g. by translating services to clusters in some situations) and Envoy Gateway will support a `SharedBackendTrafficPolicy`, I expect that this policy will include `CircuitBreakers` but not `Retries`. So, for future reusability, it could be better to have these settings under the circuit breaker types.  

I'm willing to drop `MaxRetries` from this PR for now. We can continue the discussion in https://github.com/envoyproxy/gateway/pull/2168 on the best location for these settings in the API and implement it as part of that PR. WDYT?

api/v1alpha1/circuitbreaker_types.go
guydc: Wouldn't it make more sense to use `*int64`? `MaxUInt32` > `MaxInt32`, so by using `*int32` users would not able to use the full value range provided by Envoy.

api/v1alpha1/circuitbreaker_types.go
zhaohuabing: Agree. A one-member list just looks wired to me.

api/v1alpha1/circuitbreaker_types.go
zhaohuabing: IMO, the concurrent max retries setting belongs to Circuit Breaker logic because it enforces back pressure on the clients. Therefore, EG probably should not mix it with the request retries configuration.

tmsnan: I think we should distinguish the difference between `circuit_breaker` and `retryStrategy` functions for users, and should not be limited to the design of envoy. Although  `retry budget` and `concurrent max retries` are implemented in `circuit_breaker` in envoy, for users, they are all retry functions( provides richer uses for the retry function).

Shared cluster is indeed a point that needs to be considered, but now I am not sure how to use it. Maybe it is an implementation similar to istio DestinationRule resource. If so, you can patch the BackendTrafficPolicy to dr.

zhaohuabing: > IMO, we should distinguish between the `circuit_breaker` and `retryStrategy` functions for users, as they offer distinct features, not limited to the design of Envoy. Even though `retry budget` and `concurrent max retries` are implemented in the `circuit_breaker` in Envoy, for users, these encompass retry functions that provide richer options for retry operations.

I vote -1 on this.

 Even though both have "retries" in their name, they serve two different purposes. The concurrent max retries setting is inherently associated with the Circuit Breaker, which fails requests quickly when a lot of retries happen and apply back pressure on downstream. On the other hand, request retries are specifically designed to mitigate transient network issues. Would love more insights from @kflynn and other @envoyproxy/gateway-maintainers

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -65,6 +65,12 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty"`
+
+	// Circuit Breaker settings for the upstream connections and requests.
+	// If not set, circuit breakers will be enabled with the highest supported thresholds
```
arkodg, line 70:
```suggestion
	// If not set, circuit breakers will be enabled with default thresholds
```

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type Thresholds struct {
+	// The maximum number of connections that Envoy will establish to the referenced backend (per xRoute).
```
arkodg, line 19:
```suggestion
	// The maximum number of connections that Envoy will establish to the referenced backend (per xRoute per rule).
```

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,50 @@
...
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:default=1024
+	// +optional
+	MaxConnections *uint32 `json:"maxConnections,omitempty"`
+
+	// The maximum number of pending requests that Envoy will queue to the referenced backend (per xRoute).
```
arkodg, line 27:
```suggestion
	// The maximum number of pending requests that Envoy will queue to the referenced backend (per xRoute per rule).
```

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// The maximum number of parallel requests that Envoy will make to the referenced backend (per xRoute).
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:default=1024
+	// +optional
+	MaxRequests *uint32 `json:"maxParallelRequests,omitempty"`
+
+	// The maximum number of parallel retries that Envoy will allow to the referenced backend (per xRoute).
```
arkodg, line 43: vote to rm this for now, raise a issue to track max parallel retries, and once the retry API is complete, we can revisit this field and decide on the right home for this

api/v1alpha1/circuitbreaker_types.go
guydc: Since @arkodg https://github.com/envoyproxy/gateway/pull/2284#discussion_r1428347907 and @tmsnan support removing `MaxRetries` from this API proposal, I'll go ahead and remove it. If we eventually decide that `CircuitBreakers` should contain these settings, we can add them later on.

api/v1alpha1/circuitbreaker_types.go
arkodg: I vote to keep it singular for now
```
circuitBreaker:
  maxConnections: 
  maxPendingRequests: 
  maxParallelRequests: 
  ```
  
  I dont see routing priority being added in the upstream Gateway API, and if it does, we can deprecate `circuitBreaker` in favor of `circuitBreakers`

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM

> vote to rm this for now, raise a issue to track max parallel retries, and once the retry API is complete, we can revisit this field and decide on the right home for this

Please raise an issue to make sure we can track this, thanks!

Issue #2286, opened 2023-12-10
Support setting common TLS properties on downstream listeners


liorokman:
*Description*:

Support the following attributes in `ClientTrafficPolicy`:

```yaml
 tls: # optional
    version: # optional
      min: string # optional
      max: string # optional
    ciphers: []Enum(list of supported cipher suites...) # optional
    ecdhCurves: []string # optional
    alpnProtocols: []Enum(h2/http-1.1) # optional
```

These are all part of the proposed content of the `ClientTrafficPolicy` specified in #1846 ( proposal: ClientTrafficPolicy ).

Since all of these attributes are optional, not specifying any of them would translate to the same Envoy configuration as before this feature.

*Relevant Links*:

The configuration provided in the `ClientTrafficPolicy` would be translated to Envoy's [TlsParameters](https://www.envoyproxy.io/docs/envoy/v1.28.0/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlsparameters) configuration.

zesiar0:
please assign issue #2422 and #2421 to me, I can finish them

PR #2287, opened 2023-12-10
api: Add TLS configuration attributes in ClientTrafficPolicy

liorokman:
This PR adds the relevant optional TLS configuration attributes to `ClientTrafficPolicy`

Related to #2286


api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+	// TLSv1.2 specifies TLS version 1.2
+	TLSv12 TLSVersion = "TLSv1_2"
+	// TLSv1.3 specifies TLS version 1.3
+	TLSv13 TLSVersion = "TLSv1_3"
+)
+
+type TLSVersions struct {
+	// Min specifies the minimal TLS verison to use
+	// +optional
+	Min TLSVersion `json:"min,omitempty"`
```
shawnh2, line 54: using `*TLSVersion`, optional fields are always ptr type

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Max specifies the maximal TLS verison to use
+	// +optional
+	Max TLSVersion `json:"max,omitempty"`
```
shawnh2, line 58: ditto

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+type TLSSettings struct {
+	// Version details the minimum/maximum TLS protocol verison that
+	// should be supported by this listener.
+	// +optional
+	Version *TLSVersions `json:"version,omitempty"`
```
shawnh2, line 12: do we need to setup a default value for this field in EG?

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// SignatureAlgorithms specifies which signature algorithms the listener should
+	// support.
+	// +optional
+	SignatureAlgorithms []string `json:"signatureAlgorithms,omitempty"`
+
+	// ALPNProtocols supplies the list of ALPN protocols that should be
+	// exposed by the listener. If left empty, ALPN will not be exposed.
+	// +optional
+	ALPNProtocols []string `json:"alpnProtocols,omitempty"`
```
shawnh2, line 31: suggest a enumerate type instead of `string`, since [envoyproxy TLS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto.html#extensions-transport-sockets-tls-v3-commontlscontext) only support two values: `h2,http/1.1` or `http/1.1`.

IMO, we can add http protocol version enumerate type under `api/v1alpha1/shared_types.go`, like:
```go
HTTPProtocolVersion1_1 = "http/1.1"
HTTPProtocolVersion2   = "http/2"
```

and we can translate their combination into envoyproxy supported values:

- `["http/1.1", "http/2"]` will be seen as `h2,http/1.1`
- `["http/1.1"]` will be seen as `http/1.1`
- invalid otherwise

WDYT?

liorokman: I think maybe it's better to keep this as strings. I tried to create a configuration in Envoy where the ALPN was something else and Envoy had no issue with that - I was able to use `openssl s_client` to negotiate that different protocol string. 

If somebody uses a custom version of Envoy that is aware of a different protocol, it should be possible to configure it from EG.

api/v1alpha1/tls_types.go
liorokman: Today EG simply doesn't specify any default, leaving the default up to Envoy itself. Maybe keep that behaviour, so that there will be backwards compatibility with previous EG versions?

api/v1alpha1/tls_types.go
liorokman: Fixed

api/v1alpha1/tls_types.go
liorokman: Fixed

api/v1alpha1/tls_types.go
shawnh2: fair enough, since `alpn_protocols` is also a string type in envoyproxy

shawnh2 (approved):
lgtm, thanks, defer to @envoyproxy/gateway-maintainers

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
```
zhaohuabing, line 31: I'm not sure that we need to expose `alpn_protocols` to API.  Are there any use cases other than the default `"h2", "http/1.1` ?

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+	CipherSuites []string `json:"ciphers,omitempty"`
+
+	// ECDHCurves specifies the set of supported ECDH curves.
+	// +optional
+	ECDHCurves []string `json:"ecdhCurves,omitempty"`
+
+	// SignatureAlgorithms specifies which signature algorithms the listener should
+	// support.
+	// +optional
+	SignatureAlgorithms []string `json:"signatureAlgorithms,omitempty"`
```
zhaohuabing, line 26: Also, are there any use cases that need to configure CipherSuites,ECDHCurves, and SignatureAlgorithms?

api/v1alpha1/tls_types.go
liorokman: In some cases we have customers that want to disable ALPN altogether, or disable negotiating HTTP/2 with ALPN.

Generally, there are cases where it makes sense to have a custom implemented protocol. For example, Istio uses a custom ALPN protocol to distinguish between types of traffic (e.g. https://github.com/istio/istio/issues/40680 ).

api/v1alpha1/tls_types.go
liorokman: If TLS protocol versions below 1.2 are used, the default cipher suite is not going to work. See the documentation [here](https://www.envoyproxy.io/docs/envoy/v1.28.0/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlsparameters).

The ability to control the supported curves and signature algorithms is needed for some scenarios where additional hardening is needed - the customer wants to specify exactly what should be used.

While most use-cases shouldn't need to touch the defaults, sometimes it's required.

api/v1alpha1/tls_types.go
zhaohuabing: > In some cases we have customers that want to disable ALPN altogether, or disable negotiating HTTP/2 with ALPN.

AFAIK, normally the server side wants to let the user side choose between HTTP 1.1 and HTTP2 for back compatibility. 

> Generally, there are cases where it makes sense to have a custom implemented protocol. For example, Istio uses a custom ALPN protocol to distinguish between types of traffic (e.g. https://github.com/istio/istio/issues/40680 ).

The Istio use case is legitimate, but it's not a Gateway scenario.

I suggest removing this from the API if we don't have a concrete use case right now. We can always add it later, but it will be really hard to remove an existing feature from API.

guydc: Envoy disables ALPN by default. Envoy Gateway enables ALPN and sets it to `[h2, http/1.1]` by default. It's a reasonable behavior, but perhaps users should have the option to control this parameter. 

In general, other protos are supported in the [IANA registry](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids). In rare cases, it could be necessary to support other protos, e.g. there was a period where `curl` was [spec-compliant](https://github.com/curl/curl/issues/12259) and used `http/1.0` as ALPN for HTTP/1.0. 

Other Gateways like Emissary-Ingress do support configuration of [alpn settings](https://www.getambassador.io/docs/emissary/latest/topics/running/tls#alpn-protocols). The Emissary docs imply that it could be necessary to explicitly set ALPN to `[h2]` for gRPC. I'm not sure if this generally mandatory or required for specific grpc clients. I assume that this was also the considered when ALPN was included in [CTP](https://github.com/envoyproxy/gateway/issues/1846). Maybe @AliceProxy can provide some context here as well.

zhaohuabing: defer to @envoyproxy/gateway-maintainers for suggestions.

api/v1alpha1/tls_types.go
zhaohuabing: defer to @envoyproxy/gateway-maintainers for more inputs.

arkodg:
should these be defined under `tls.options` within the Gateway ?
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewayTLSConfig

zhaohuabing:
> should these be defined under `tls.options` within the Gateway ? https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewayTLSConfig

Yes, if upstream them to Gateway API.

arkodg:
@zhaohuabing referring to the `options` field which is a map, which is a tls specific extension provided by upstream

zhaohuabing:
> @zhaohuabing referring to the `options` field which is a map, which is a tls specific extension provided by upstream

Ha, I haven't noticed that.

api/v1alpha1/tls_types.go
arkodg: The default min version is `TLSv1_2`
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto
`Using TLS protocol versions below TLSv1_2 has serious security considerations and risks.`
what is the reason to use  version below TLS 1.2 ?

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+package v1alpha1
+
+type TLSSettings struct {
+	// Version details the minimum/maximum TLS protocol version that
+	// should be supported by this listener.
+	// +optional
+	Version *TLSVersions `json:"version,omitempty"`
+
+	// CipherSuites specifies the set of cipher suites supported when
+	// negotiating TLS 1.0 - 1.2. This setting has no effect for TLS 1.3.
```
arkodg, line 15: needs to be enforced with CEL validation

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+	// +optional
+	CipherSuites []string `json:"ciphers,omitempty"`
+
+	// ECDHCurves specifies the set of supported ECDH curves.
```
arkodg, line 19: default needs to be specified

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,59 @@
...
+	// +optional
+	ECDHCurves []string `json:"ecdhCurves,omitempty"`
+
+	// SignatureAlgorithms specifies which signature algorithms the listener should
+	// support.
+	// +optional
+	SignatureAlgorithms []string `json:"signatureAlgorithms,omitempty"`
+
+	// ALPNProtocols supplies the list of ALPN protocols that should be
+	// exposed by the listener. If left empty, ALPN will not be exposed.
```
arkodg, line 29:
```suggestion
	// exposed by the listener. By default http/1.1 and http/2 are enabled .
```

api/v1alpha1/tls_types.go
arkodg: +1, need to use a custom type to enforce cube builder validations
check out https://github.com/envoyproxy/gateway/blob/52946bef5132a302b6f5edd6d154062f35906e22/api/v1alpha1/ratelimit_types.go#L136

api/v1alpha1/tls_types.go
liorokman: In some cases there are legacy clients that don't support TLS 1.2 but require an older version.

api/v1alpha1/tls_types.go
liorokman: But ALPN is more than just "h2" and "http/1.1". If I define this as an enum, it becomes impossible to use anything else. See @guydc's comment https://github.com/envoyproxy/gateway/pull/2287#discussion_r1425469498 below.

api/v1alpha1/tls_types.go
liorokman: Fixed.

api/v1alpha1/tls_types.go
liorokman: Added.

api/v1alpha1/tls_types.go
liorokman: Changed.

guydc:
> should these be defined under `tls.options` within the Gateway ? https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewayTLSConfig

@arkodg That's a good point. The last activity related to Gateway TLS `options` took place in late 2021: https://github.com/kubernetes-sigs/gateway-api/pull/886. I don't think that any common TLS options were introduced in Gateway-API and I'm not sure if any of the other Gateway-API implementations leverage this part of the spec. 

More recently, the Policy Attachment GEP mentioned `Listener` TLS configuration as an example for policy use cases: https://gateway-api.sigs.k8s.io/geps/gep-713/#direct-policy-attachment_1. Backend TLS settings are being implemented with a Policy resource: https://gateway-api.sigs.k8s.io/geps/gep-1897/. The situation here is a bit different, since it's not likely that the K8s Service resource will be extended to support backend TLS settings. So, a Policy resource is necessary. 

I feel that both alternatives (policy and options) are equally valid from the Gateway-API perspective. Policies are more flexible (e.g. supporting both `Gateway` and `Listener` attachment) and provide better API Lifecycle Management (kubebuilder validation, deprecation process, decoupling from restriction applied to k8s annotations, etc.). So, I prefer a Policy-based approach here. WDYT ?

api/v1alpha1/tls_types.go
arkodg: do you have a specific use case where you'd like to set a specific ALPN apart from h2 and http/1.1 ?
its safer to use enums and validate synchronously imo

arkodg:
> > should these be defined under `tls.options` within the Gateway ? https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewayTLSConfig
> 
> @arkodg That's a good point. The last activity related to Gateway TLS `options` took place in late 2021: [kubernetes-sigs/gateway-api#886](https://github.com/kubernetes-sigs/gateway-api/pull/886). I don't think that any common TLS options were introduced in Gateway-API and I'm not sure if any of the other Gateway-API implementations leverage this part of the spec.
> 
> More recently, the Policy Attachment GEP mentioned `Listener` TLS configuration as an example for policy use cases: https://gateway-api.sigs.k8s.io/geps/gep-713/#direct-policy-attachment_1. Backend TLS settings are being implemented with a Policy resource: https://gateway-api.sigs.k8s.io/geps/gep-1897/. The situation here is a bit different, since it's not likely that the K8s Service resource will be extended to support backend TLS settings. So, a Policy resource is necessary.
> 
> I feel that both alternatives (policy and options) are equally valid from the Gateway-API perspective. Policies are more flexible (e.g. supporting both `Gateway` and `Listener` attachment) and provide better API Lifecycle Management (kubebuilder validation, deprecation process, decoupling from restriction applied to k8s annotations, etc.). So, I prefer a Policy-based approach here. WDYT ?

jotting down some more pros and cons of each approach

**Using the `tls.options` within Gateway resource**
**Pros**
* config lives close to the `tls` listener config
* easier to debug since all `tls` config lives in one place
* dont need to create and maintain another resource (ClientTrafficPolicy) for just this case

**Cons**
* field and values are not strongly typed so can only be validated asynchronously and surfaced in the Gateway/Listener status
* cannot add CEL validation

**Creating a new field within ClientTrafficPolicy**
**Pros**
* Strongly typed config
* CEL validations can be added to provide instantaneous feedback

**Cons**
* tls config lives a little far away from the listener tls config, increasing cognitive load
* debugging involves examining multiple resources

liorokman:
> jotting down some more pros and cons of each approach
> 
> **Using the `tls.options` within Gateway resource** **Pros**
> 
> * config lives close to the `tls` listener config
> * easier to debug since all `tls` config lives in one place
> * dont need to create and maintain another resource (`ClientTrafficPolicy`) for just this case
> 
> **Cons**
> 
> * field and values are not strongly typed so can only be validated asynchronously and surfaced in the Gateway/Listener status
> * cannot add CEL validation
> 
> **Creating a new field within `ClientTrafficPolicy`** **Pros**
> 
> * Strongly typed config
> * CEL validations can be added to provide instantaneous feedback
> 
> **Cons**
> 
> * tls config lives a little far away from the listener tls config, increasing cognitive load
> * debugging involves examining multiple resources

I think that the strongly typed config is significantly more important than having the configuration close to the listener. 

Just the `ciphers` are complicated enough that having them in an untyped string is a bad idea. It's a [small mini-language](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Cipher-suite-configuration) that needs to be specified in exactly the right way, and having to also condense it into a string makes it harder to use.

api/v1alpha1/tls_types.go
liorokman: My point is that validation the values to just "h2" and "http/1.1" is limiting. If this is moved to enums, I think it would make sense to have all of the possible values in the [IANA registry](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids) available as accepted enums. 

EG is not there yet, but at some point it would make sense to have `ClientTrafficPolicy` documents attached also to TLS routes. When that happens, it would be possible to have other protocols supported. Examples would include SMTP, MQTT, or IMAP with EG terminating the TLS and routing to a service running in the cluster. If ALPN is limited to just the two possibilities - this won't be possible.

It honestly seems better to me to not set this as an enum. The default is set to `[ "h2", "http/1.1" ]`, meaning that most people wouldn't touch this. But when more control over this is needed, it makes more sense to me to make it possible to control this setting.

liorokman:
> jotting down some more pros and cons of each approach

More pros for the policy approach:

* A policy can be applied to all listeners in a `Gateway` resource in one place, so there's less copy-n-paste required when a gateway has multiple listeners that need to be hardened. 
* Allows for a separation of concerns - the person that configures a `Gateway` is not necessarily the same person that is responsible for fine-tuning TLS parameters (in some cases this is done by a security expert operating from a central perspective).

api/v1alpha1/tls_types.go
liorokman: Updated to be an enum, as agreed in the meeting.

api/v1alpha1/tls_types.go
kflynn: (This is a rare use case, but it is a real use case.)

api/v1alpha1/tls_types.go
kflynn: I would vote for being explicit here. Setting the default to "whatever Envoy wants" leaves behavior interesting undefined as the Envoy version changes.

As @arkodg notes, Envoy's current default minimum version is TLS 1.2, and there are some _very_ good reasons not to go below that. I'd vote to define Envoy Gateway's minimum as TLS 1.2, while still allowing users to select 1.0 or 1.1 if they need to.

(Allowing the _maximum_ to float with Envoy is OK with me, FTR.)

kflynn: (I'd also vote to call this stanza `versions` plural rather than `version` singular.)

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// TLSAuto allows Envoy to choose the optimal TLS Version
+	TLSAuto TLSVersion = "TLS_Auto"
+	// TLSv1_0 specifies TLS version 1.0
+	TLSv10 TLSVersion = "TLSv1_0"
+	// TLSv1_1 specifies TLS version 1.1
+	TLSv11 TLSVersion = "TLSv1_1"
+	// TLSv1.2 specifies TLS version 1.2
+	TLSv12 TLSVersion = "TLSv1_2"
+	// TLSv1.3 specifies TLS version 1.3
+	TLSv13 TLSVersion = "TLSv1_3"
```
kflynn, line 87: As noted on the wrong PR 🤦‍♂️, I'd rather these just be "v1.2" etc. Forcing the user to type "TLS_" all the type feels kind of pointless.

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type TLSVersions struct {
+	// Min specifies the minimal TLS version to use
+	// +optional
+	Min *TLSVersion `json:"min,omitempty"`
+
+	// Max specifies the maximal TLS version to use
+	// +optional
+	Max *TLSVersion `json:"max,omitempty"`
```
kflynn, line 97: Also as noted on the wrong PR 🤦‍♂️, I think it's important to allow setting either of these without being forced to set the other.

OTOH if - as I recommend - the default minimum is v1.2, then setting the maximum to e.g. v1.1 without also setting the minimum lower should be an error.

kflynn (changes requested):
Sorry for commenting on the wrong PR! 🤦‍♂️ I've added some here, and yes, I'd like to request some changes here. Thanks!

api/v1alpha1/tls_types.go
liorokman: > I would vote for being explicit here. Setting the default to "whatever Envoy wants" leaves behavior interesting undefined as the Envoy version changes.

"Whatever Envoy wants" is the current default, and is also what would happen if there is no `ClientTrafficPolicy` attached at all. 

The `TLS_Auto` constant maps all the way to the Envoy Proxy's enum, which is defined as `⁣Envoy will choose the optimal TLS version.` Setting the minimum in EG to something other than `TLS_Auto` means that if Envoy Proxy fixes its default, maybe for solving some future CVE, then Envoy Gateway doesn't automatically benefit from the fix in Envoy Proxy.

> (I'd also vote to call this stanza versions plural rather than version singular.)

I'm thinking to maybe get rid of the `versions` intermediate level, and align more closely with how Envoy Proxy models it. I think this would make it clearer.

WDYT?

api/v1alpha1/tls_types.go
liorokman: Fixed.

api/v1alpha1/tls_types.go
liorokman: This is currently the case. That's why both of these fields are optional. If you set one but not the other, then Envoy Proxy will use its default, (which is `TLS_AUTO`) and is documented as  `(DEFAULT) ⁣Envoy will choose the optimal TLS version.` This works correctly regardless of which of the fields was set or unset.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-enum-extensions-transport-sockets-tls-v3-tlsparameters-tlsprotocol

The implementation itself verifies that the minimum version is not greater than the maximum version. Since these are string constants, it's not easy to concisely validate correct ordering with CEL.

arkodg:
removing the `decision` label, this was discussed in the community meeting yesterday, and @zhaohuabing @zirain @guydc and I agreed on introducing this API in `ClientTrafficPolicy` instead of reusing the `tls.options` field within `Gateway` for the below reasons
* stronger validation of fields
* apply default settings to all listeners

This API can be deprecated once these fields are natively introduced upstream

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -59,6 +59,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	EnableProxyProtocol *bool `json:"enableProxyProtocol,omitempty"`
+	// TLS settings controlling how connections are established with the downstream host.
```
arkodg, line 62:
```suggestion
	// TLS settings configure TLS termination settings with the downstream client.
```

api/v1alpha1/tls_types.go
arkodg: yeah, we could rm `versions` in favor of `minVersion` & `maxVersion` , thoughts @kflynn ?

api/v1alpha1/clienttrafficpolicy_types.go
liorokman: Updated.

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// - ECDHE-RSA-AES256-GCM-SHA384
+	//
+	// In builds using BoringSSL FIPS the default cipher list is:
+	// - ECDHE-ECDSA-AES128-GCM-SHA256
+	// - ECDHE-RSA-AES128-GCM-SHA256
+	// - ECDHE-ECDSA-AES256-GCM-SHA384
+	// - ECDHE-RSA-AES256-GCM-SHA384
+	//
+	// +optional
+	CipherSuites []string `json:"ciphers,omitempty"`
```
arkodg, line 32: should this be made into an enum for better validation ?

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// ECDHCurves specifies the set of supported ECDH curves.
+	// In non-FIPS Envoy Proxy builds the default curves are:
+	// - X25519
+	// - P-256
+	//
+	// In builds using BoringSSL FIPS the default curve is:
+	// - P-256
+	// +optional
+	ECDHCurves []string `json:"ecdhCurves,omitempty"`
```
arkodg, line 42: should this be made into an enum for better validation ?

api/v1alpha1/tls_types.go
arkodg: does this need to be an enum ?

api/v1alpha1/tls_types.go
liorokman: It's not possible to make it into an enum, because the values in here are not always singular values. 

See the [cipher list documentation](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Cipher-suite-configuration) from Envoy Proxy's underlying library.

The values in this array are expressions, allowing things like `[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]`, which is an expression that says "either the first cipher or the second".

Additional possible values are to use `!` or '+' or '-' as the prefix for a cipher name. This cannot be expressed as an enum in Envoy Gateway.

api/v1alpha1/tls_types.go
liorokman: The set of valid values are not clear. 

Envoy Proxy doesn't define them - the underlying BoringSSL library does. If Envoy Proxy is compiled against a different version of BoringSSL, there may be different valid values than the ones the EnvoyGateway code base uses.

api/v1alpha1/tls_types.go
liorokman: This field is so much of a moving target that even the [Envoy Proxy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto) doesn't specify what the up-to-date values are. I'm not sure it makes sense to have an enum here.

api/v1alpha1/tls_types.go
kflynn: `minVersion` and `maxVersion` work for me.

I would definitely prefer having Envoy Gateway dictate a particular minimum version for predictability, but I don't feel all that strongly there.

api/v1alpha1/tls_types.go
kflynn: I'd go with @arkodg's take here:

```yaml
tls:
  minVersion: v1.2
  maxVersion: v1.3
```

(I don't care terribly if it's `v1.2` or `v1_2`, though I'd slightly prefer the `.`).

api/v1alpha1/tls_types.go
kflynn: In the other PR, it would actually throw an error if you set `min` and left `max` at `AUTO`, since `AUTO` is defined to have the value `-1`. 🙂 That's what prompted me to look at this.

I agree that expressing this in CEL is tricky. 🙁 I'm not actually sure it's possible...

api/v1alpha1/tls_types.go
kflynn: I agree with @liorokman that curves & ciphers should not be enums – there's not going to be a great way to validate, especially in BYOEnvoy cases. 🙁 This is a case where we should be thinking carefully about the behavior when there's an error, and documenting it. (Personally I'd like to see a status on the `Gateway` itself...)

api/v1alpha1/tls_types.go
liorokman: Modified the version specification to remove the intermediate level.

I also set the default minimal version default to be 1.2 and the default maximal version to be 1.3. This is actually exactly the same as it was before, since this is Envoy Proxy's default. I agree that it makes sense to document it more clearly.

api/v1alpha1/tls_types.go
liorokman: I removed the intermediate `version` structure.

api/v1alpha1/tls_types.go
liorokman: Once the API is agreed upon, I'll make sure to add tests (as part of the other PR) to verify that the verification works correctly.

api/v1alpha1/tls_types.go
arkodg: cool, lets keep it as is

api/v1alpha1/tls_types.go
arkodg: cool lets keep as is

api/v1alpha1/tls_types.go
arkodg: cool, lets keep it as is

api/v1alpha1/tls_types.go
```diff
@@ -0,0 +1,95 @@
...
+	// - ECDHE-RSA-AES256-GCM-SHA384
+	//
+	// In builds using BoringSSL FIPS the default cipher list is:
+	// - ECDHE-ECDSA-AES128-GCM-SHA256
+	// - ECDHE-RSA-AES128-GCM-SHA256
+	// - ECDHE-ECDSA-AES256-GCM-SHA384
+	// - ECDHE-RSA-AES256-GCM-SHA384
+	//
+	// +optional
+	CipherSuites []string `json:"ciphers,omitempty"`
```
arkodg, line 39: go field name and json field name dont match - `cipherSuites` and  `ciphers`

liorokman: Fixed.

api/v1alpha1/tls_types.go
liorokman: Added CEL validation to verify that  `tls.minVersion <= tls.maxVersion`

arkodg (dismissed):
LGTM, appreciate addressing all the review comments and enhancing the doc strings

arkodg:
@liorokman seeing some merge conflicts, can you rebase ?

arkodg (approved):

zirain (approved):

PR #2297, opened 2023-12-12
feat: Implement setting common TLS parameters in ClientTrafficPolicy

liorokman:

Implementation for setting Ciphers, ECDHCurves, ALPN protocols, Signature Algorithms, and TLS protocol versions in ClientTrafficPolicy.

Implements #2286
Based on #2287

Related to #1846


liorokman:
@envoyproxy/gateway-reviewers please review

zirain:
please don't try to implement before API accepted, maybe a waste of your time.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -335,3 +373,64 @@ func translateListenerProxyProtocol(enableProxyProtocol *bool, httpIR *ir.HTTPLi
...
+	// at least one certificate defined, which would cause httpIR to
+	// have a TLS structure.
+	if httpIR.TLS == nil {
+		return nil
+	}
+	var errs error
+	if tlsParams.Version != nil {
+		var minVersion, maxVersion int
+		var err error
+		if minVersion, err = tlsVersionToNumericVersion(tlsParams.Version.Min); err != nil {
```
kflynn, line 392: This seems to imply that settings like this:

```yaml
version:
  min: TLSv1_2
```

will not work, since it seems that the default for `max` is `TLS_Auto`.

I think that "auto" should mean "the earliest supported version" for minimum and "the latest supported version" for maximum. (While we're at it, I also don't see any value in forcing people to type "TLS_" everywhere: just let the versions be strings like "v1.3", and "auto".)

kflynn (changes requested):
This looks like a really good start. I'd like some more attention paid to the defaults (at minimum, we absolutely _must_ state what the defaults are!), and I kind of think that if the `version` stanza sets the minimum and maximum it should be named `versions` plural, but that's a really minor nit.

I'm marking this as "request changes" because the defaults thing is important, I think. Thanks for digging into this!

kflynn:
D'oh!! I should've read #2287 before commenting here. Mea culpa.

liorokman:
> D'oh!! I should've read #2287 before commenting here. Mea culpa.

Continuing the discussion on the API in #2287.

internal/gatewayapi/clienttrafficpolicy.go
liorokman: This was resolved by adopting defaults of minimum TLS v1.2 and maximum TLS v1.3.

The validation that `min <= max` was implemented as a CEL rule.

arkodg:
hey @liorokman can you rebase ?

internal/xds/translator/listener.go
```diff
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
+	gwv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
```
arkodg, line 27: this needs to be removed, we've decoupled the xds layer from the user api layer

internal/ir/xds.go
```diff
@@ -231,6 +229,25 @@ func (h HTTPListener) Validate() error {
 // TLSListenerConfig holds the configuration for downstream TLS context.
 // +k8s:deepcopy-gen=true
 type TLSListenerConfig struct {
+	// Certificates contains the set of certificates associated with this listener
+	Certificates []TLSCertificate `json:"certificates,omitempty" yaml:"certificates,omitempty"`
+	// MinVersion defines the minimal version of the TLS protocol supported by this listener.
+	MinVersion *egv1a1.TLSVersion `json:"minVersion,omitempty" yaml:"version,omitempty"`
```
arkodg, line 235: recommend using non `egv1a1` types in here

internal/ir/xds.go
```diff
@@ -179,8 +179,8 @@ type HTTPListener struct {
 	// Refer to https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost
 	// for more info.
 	Hostnames []string `json:"hostnames" yaml:"hostnames"`
-	// Tls certificate info. If omitted, the gateway will expose a plain text HTTP server.
-	TLS []*TLSListenerConfig `json:"tls,omitempty" yaml:"tls,omitempty"`
+	// Tls configuration. If omitted, the gateway will expose a plain text HTTP server.
+	TLS *TLSListenerConfig `json:"tls,omitempty" yaml:"tls,omitempty"`
```
arkodg, line 183: lets rename the struct to `TLSConfig` since its being reused here, and this isn't a listener, its a client

internal/xds/translator/listener.go
```diff
@@ -340,17 +341,18 @@ func addXdsTLSInspectorFilter(xdsListener *listenerv3.Listener) error {
...
-				AlpnProtocols: []string{"h3"},
+				TlsParams:     buildTLSParams(tlsConfig),
+				AlpnProtocols: buildALPNProtocols(tlsConfig.ALPNProtocols),
 			},
 			RequireClientCertificate: &wrappers.BoolValue{Value: false},
 		},
 	}
 
-	for _, tlsConfig := range tlsConfigs {
+	for _, tlsConfig := range tlsConfig.Certificates {
```
arkodg, line 355: won't these crash if `tlsConfig` is nil ?

internal/xds/translator/listener.go
```diff
@@ -371,15 +373,16 @@ func buildDownstreamQUICTransportSocket(tlsConfigs []*ir.TLSListenerConfig) (*co
...
 		CommonTlsContext: &tlsv3.CommonTlsContext{
-			AlpnProtocols:                  []string{"h2", "http/1.1"},
+			TlsParams:                      buildTLSParams(tlsConfig),
+			AlpnProtocols:                  buildALPNProtocols(tlsConfig.ALPNProtocols),
 			TlsCertificateSdsSecretConfigs: []*tlsv3.SdsSecretConfig{},
 		},
 	}
 
-	for _, tlsConfig := range tlsConfigs {
+	for _, tlsConfig := range tlsConfig.Certificates {
```
arkodg, line 385: same as https://github.com/envoyproxy/gateway/pull/2297/files#r1441086773

internal/xds/translator/listener.go
```diff
@@ -401,7 +404,57 @@ func buildXdsDownstreamTLSSocket(tlsConfigs []*ir.TLSListenerConfig) (*corev3.Tr
...
+	}
+	if isEmpty {
+		return nil
+	}
+	return p
+}
+
+func buildTLSVersion(version *gwv1a1.TLSVersion) tlsv3.TlsParameters_TlsProtocol {
+	lookup := map[gwv1a1.TLSVersion]tlsv3.TlsParameters_TlsProtocol{
+		gwv1a1.TLSv10: tlsv3.TlsParameters_TLSv1_0,
```
arkodg, line 438: thinking out loud, about the API, should the user strings be `1.2`, `1.3` instead of `v1_2` and `v1_3` ?

internal/xds/translator/listener.go
```diff
@@ -401,7 +404,57 @@ func buildXdsDownstreamTLSSocket(tlsConfigs []*ir.TLSListenerConfig) (*corev3.Tr
...
+		gwv1a1.TLSv12: tlsv3.TlsParameters_TLSv1_2,
+		gwv1a1.TLSv13: tlsv3.TlsParameters_TLSv1_3,
+	}
+	if r, found := lookup[*version]; found {
+		return r
+	}
+	return tlsv3.TlsParameters_TLS_AUTO
+}
+
+func buildALPNProtocols(in []gwv1a1.ALPNProtocol) []string {
```
arkodg, line 449: imo this is not needed if we use `[]string` in the IR

internal/gatewayapi/route.go
```diff
@@ -995,6 +996,9 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
 				},
 				TLS: &ir.TLS{Terminate: irTLSConfigs(listener.tlsSecrets)},
 			}
+			if irListener.TLS.Terminate != nil {
```
arkodg, line 999: nice !
a comment in here saying enable http/1.1 & http/2 alpn protocols by default would help

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -359,3 +362,44 @@ func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *i
...
+	// at least one certificate defined, which would cause httpIR to
+	// have a TLS structure.
+	if httpIR.TLS == nil {
+		return
+	}
+	// Make sure that the negotiated TLS protocol version is as expected if TLS is used,
+	// regardless of if TLS parameters were used in the ClientTrafficPolicy or not
+	httpIR.TLS.MinVersion = ptr.To(egv1a1.TLSv12)
+	httpIR.TLS.MaxVersion = ptr.To(egv1a1.TLSv13)
+	// If HTTP3 is enabled, the ALPN protocols array should be hardcoded
```
arkodg, line 377: cc @tanujd11 , this breaks upgrade or downgrades, right ?
imo this should be an append, not a set

arkodg:
@liorokman curious why the permissions was changed for so many files

liorokman:
> curious why the permissions was changed for so many files

When I started with the tests, it stood out like a sore thumb that so many yaml files were marked as executable.

internal/xds/translator/listener.go
liorokman: `tlsConfigs` can't be nil because this function is only called after checking that it isn't.

https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/listener.go#L190

internal/gatewayapi/clienttrafficpolicy.go
liorokman: I didn't append because if HTTP/3 was enabled, then only HTTP/3 is expected. 

The socket being configured is being explicitly setup as a Quic socket, so I'm not sure that accepting HTTP/1.1 or HTTP/2 would work in this case. It's also what the [original code](https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/listener.go#L347) did.

internal/xds/translator/listener.go
liorokman: Fixed.

internal/ir/xds.go
liorokman: Changed.

internal/ir/xds.go
liorokman: Renamed.

internal/xds/translator/listener.go
liorokman: Changed.

internal/gatewayapi/route.go
liorokman: Added a comment.

internal/xds/translator/listener.go
liorokman: I don't think it matters that much. I'll change to whatever is requested.

kflynn: FTR I think "1.2" and "1.3" are great. 🙂

internal/gatewayapi/clienttrafficpolicy.go
arkodg: ah thanks for the clarification

internal/xds/translator/listener.go
arkodg: thanks for clarifying this

internal/xds/translator/listener.go
arkodg: @liorokman sorry for churn, can you go ahead and update the strings to `1.2` & `1.3`

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -359,3 +362,45 @@ func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *i
...
+	// Make sure that the negotiated TLS protocol version is as expected if TLS is used,
+	// regardless of if TLS parameters were used in the ClientTrafficPolicy or not
+	httpIR.TLS.MinVersion = ptr.To(ir.TLSv12)
+	httpIR.TLS.MaxVersion = ptr.To(ir.TLSv13)
+	// If HTTP3 is enabled, the ALPN protocols array should be hardcoded
+	// for HTTP3
+	if httpIR.HTTP3 != nil {
+		httpIR.TLS.ALPNProtocols = []string{"h3"}
+	} else if tlsParams != nil {
+		httpIR.TLS.ALPNProtocols = make([]string, len(tlsParams.ALPNProtocols))
```
arkodg, line 382: shouldnt this line 
```
httpIR.TLS.ALPNProtocols = make([]string, len(tlsParams.ALPNProtocols))
```
be under `if lsParams.ALPNProtocols != nil`

else it will end up resetting the value

internal/gatewayapi/route.go
arkodg: just realized this is the logic for TCPRoute, what about HTTPRoute and GRPCRoute ?

arkodg: rethinking about this logic, here's my sugession
1. dont parse/set ALPN here
2. in the ClientTrafficPolicy layer, if ALPN is set, then set it in the IR
3. If the xds layer, if ALPN is nil, set to http/1.1 and http/2, if set, set to the value in ALPN field

internal/xds/translator/listener.go
liorokman: Done.

One small thing to notice: when setting the TLS protocol version to the string `1.0`, sometimes YAML/JSON parsers detect the value as a number and not as a string and remove the `.0` from the value. It's important to quote the value in order to keep the `.0` in it as part of a string.

This can happen when the YAML text is unmarshaled into untyped maps. I ran into this when updating the YAML files that comprise the tests. It also happens with the tool that generates the CRD, which is why the validation enum comment contains quotes.

Example:
```yaml
spec:
    tls:
       # This sometimes gets parsed wrong, since it becomes a number and 
       # loses the fractional part.
       minVersion: 1.0
       # This always gets parsed correctly as a string.
       minVersion: "1.0"
```

internal/gatewayapi/route.go
liorokman: Done.

internal/gatewayapi/clienttrafficpolicy.go
liorokman: Fixed.

arkodg:
hey @liorokman one more thing I noticed was the ALPN string for `http/2` - lets set it to `h2` in the API
its easier to define `h2c` in the future https://developer.mozilla.org/en-US/docs/Glossary/ALPN

apart from this, LGTM

this PR also needs a rebase

liorokman:
> hey @liorokman one more thing I noticed was the ALPN string for `http/2` - lets set it to `h2` in the API its easier to define `h2c` in the future https://developer.mozilla.org/en-US/docs/Glossary/ALPN

ALPN can never be used for `h2c`, by the very definition of `h2c` (HTTP/2 over cleartext). ALPN is a TLS extension, while `h2c` is only possible if TLS is not being used and HTTP/2 is not negotiated but rather used because of pre-knowledge. Without TLS there is no ALPN, but with TLS there is no `h2c`.

If EG supports `h2c`, it will not be a good fit to enable it by configuring TLS or ALPN, which means that these constants could never mean `h2c`. 

I suggest keeping the `http/2` string in the ALPN constants, since if `http/2` is being used it's never going to be anything other than `h2c`.

arkodg:
@liorokman I understand that `h2c` in the context of TLS doesnt make much sense, which is also why we shouldnt allow it in the API
but there is an existing RFC definition for ALPN protocols here https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids and we should reuse it imo

liorokman:
Is it OK to assume that valid Enum values for the ALPN field will always be from the IANA registry? Or should I keep the lookup function that currently allows the enum to be different than the technical values required on the wire?

arkodg:
> Is it OK to assume that valid Enum values for the ALPN field will always be from the IANA registry? Or should I keep the lookup function that currently allows the enum to be different than the technical values required on the wire?

yeah 1:1 works, and we can do w/o a lookup func for alpn for now imo

arkodg (approved):
LGTM thanks, appreciate you addressing all the comments !

api/v1alpha1/tls_types.go
```diff
@@ -5,8 +5,8 @@
 
 package v1alpha1
 
-// +kubebuilder:validation:XValidation:rule="has(self.minVersion) && self.minVersion == 'v1_3' ? !has(self.ciphers) : true", message="setting ciphers has no effect if the minimum possible TLS version is 1.3"
```
zirain, line 8: can you add some test for these in a separated PR.

zirain (approved):

api/v1alpha1/tls_types.go
liorokman: Added #2424

Issue #1231, opened 2023-03-27
Support multiple GatewayClass


arkodg:
*Description*:
Today, Envoy Gateway only supports a single GatewayClass. This decision was taken to simplify logic, limit permissioning and ensure sensitive material is isolated (across tenants) before dealing with multi tenancy (each GatewayClass is a tenant). 

Our proposed alternate solution is to request users to create another Envoy Gateway controller with a different name in a the same cluster in a different namespace and link the newer GatewayClass to it

Keeping this issue open to gather feedback from users.

wondersd:
Hi all,

I think it would be great if multiple `GatewayClasses` could be managed such that one controller is able to handle multiple "tenants" or even diverse `Gateway` instances within one logical "tenant". As it stands today, even with multiple EG controllers for each logical tenant, team X or environment Y, is still bound to only creating uniform `Gateway` instances within that tenant. If they wish to segregate internally for different purposes, e.g. internal vs external, or a critical service like an exposed database vs more forgiving lightweight api calls, they would have to install multiple EG controllers.

The resources allocated to it are dependent on what services are being proxied though it and are not 'automatic' as one may get via a cloud managed LB (like GKE's provided gateway api implementation mapping to GCE LB instances). If there is only ever one `Gateway` per `GatewayClass` then all is well. But trying to use multiple `Gateway` per `GatewayClass` one is not able to manage parameters of each envoy proxy cluster in accordance to what each separate `Gateway` is doing.

Having each `tenants` install a EG controller would require each tenant to have some higher level permissions to the cluster per each tenant, e.g. creating `ClusterRole` `ClusterRoleBindings` for the individualized EG controller or having to coordinate with a cluster administrator. In the case of many "tenants" or a high churn of "tenants" this is not easily manageable. This starts to lead down the road of installing a helm chart (or equivalent) per managed envoy proxy instance. Which, to me atleast, seems a bit underwhelming for the controller/operator model as its seen as a mechanism to provide "infrastructure as a service" within the k8s ecosystem. Here, providing a uniform service for managing envoy proxy clusters as defined by the gateway api spec.

I think it would be nice to either be able to specify some sort of glob pattern for watched `controllerName` like `gateway.envoyproxy.io/*`

and/or making

```
apiVersion: config.gateway.envoyproxy.io/v1alpha1
kind: EnvoyGateway
```

[...]

PR #2298, opened 2023-12-12
feat: support multiple GatewayClass per controller

cnvergence:
**What type of PR is this?**

- supporting multiple GatewayClass with different infrastructure deployment types
- extend Provider Resources watchable interface with GatewayClassResources map
- initial e2e test for multiple GatewayClass scenario

Fixes https://github.com/envoyproxy/gateway/issues/1231


arkodg:
@cnvergence kinda amazing to see how this PR doesnt require any changes to the gateway-api layer :)

internal/provider/kubernetes/controller.go
```diff
@@ -180,213 +181,206 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+			if err := r.addFinalizer(ctx, acceptedGC); err != nil {
+				r.log.Error(err, fmt.Sprintf("failed adding finalizer to gatewayclass %s",
+					acceptedGC.Name))
+				return reconcile.Result{}, err
+			}
 		}
+		// The Store is triggered even when there are no Gateways associated to the
+		// GatewayClass. This would happen in case the last Gateway is removed and the
+		// Store will be required to trigger a cleanup of envoy infra resources.
+		r.resources.GatewayAPIResources.Store(acceptedGC.Name, resourcesMap[acceptedGC.Name])
```
arkodg, line 381: seeing a bug 
* the gateway-api translator works at a few gateway class level 
https://github.com/envoyproxy/gateway/blob/e168ef961dbd40861af401bf70c781e2cba282fd/internal/gatewayapi/runner/runner.go#L63
* however in the gateway-api layer, the logic to clean up IRs is global, which will cause other gateways to be deleted
https://github.com/envoyproxy/gateway/blob/e168ef961dbd40861af401bf70c781e2cba282fd/internal/gatewayapi/runner/runner.go#L172

Possible solution
* filter on relevant keys (try and match GC) when cleaning
* or keep it flat, and pass 1 provider resource to the gateway-api layer with a list of GCs (the gateway-api translator can still work on 1 GC at a time)

cnvergence: right I checked it out, thanks @arkodg

internal/gatewayapi/runner/runner.go
```diff
@@ -76,10 +77,12 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 			}
 			// Translate to IR
 			result := t.Translate(val)
-
 			var curKeys, newKeys []string
 			// Get current IR keys
-			for key := range r.InfraIR.LoadAll() {
+			filteredInfra := r.InfraIR.LoadAllMatching(func(_ string, ir *ir.Infra) bool {
```
arkodg, line 82: im a little wary of relying on labels to decide translation.here is one suggestion to solve this, please lmk what you think
* change the pub-sub watchable interface to publish a `providerResources` which looks like `map[string]*gatewayi.Resources` where the key is the gateway class name
* loop through the `providerResources` and for each gatewayClass, call `gatewayapi.Translate`
* remember all the IR values
* once done looping, clean up any old IRs
* publish all IRs

cnvergence: Understood, I thought that it would be safe anyway to keep labels as they are used in IR only.
The problem was with dealing with the keys, runner already watches for changes in map[string]*gatewayi.Resources, where the key is a GatewayClass name and InfraIR has a key in the Gateway namespace/ Gateway name format or GatewayClass Name for merged gateways.

arkodg: not sure I understood the problem, can you rephrase that

arkodg: I understand your issue :) , the watchable message needs to be a map representing all the resources, so the key here needs to the controller name, but its value is also a map which is a list of tuples with key as gatewayclass name and val as child resources

cnvergence: sorry, it was easy to get confused, I should rephrased that it wasn't that big of an issue, just the label approach was much simpler than matching all gateways from Infra keys :)

cnvergence: ok, so you would like to change the interface completely and rework this update func, now I think I understand your idea
let me try this and I will get back with it

arkodg: thanks @cnvergence !

cnvergence: hey @arkodg, you would like to have this watchable message look like?
```code
type ControlledResources struct {
	ControlledResources watchable.Map[string, *ProviderResources]
}

// ProviderResources message
	// GatewayAPIResources is a map from a GatewayClass name to
	// a group of gateway API and other related resources.
	GatewayAPIResources watchable.Map[string, *gatewayapi.Resources]

	// GatewayAPIStatuses is a group of gateway api
	// resource statuses maps.
	GatewayAPIStatuses

	// PolicyStatuses is a group of policy statuses maps.
	PolicyStatuses
}
```
this seems to be a big refactor, so want to make sure

arkodg:
```
type ControllerResources map[string]*gatewayapi.Resources

// ProviderResources message
	// GatewayAPIResources is a map from a GatewayClass name to
	// a group of gateway API and other related resources.
	GatewayAPIResources watchable.Map[string, *ControllerResources]

	// GatewayAPIStatuses is a group of gateway api
	// resource statuses maps.
	GatewayAPIStatuses

	// PolicyStatuses is a group of policy statuses maps.
	PolicyStatuses
}
```

probably something more like this, wdyt ?

cnvergence: That should do the work for the translator and matching old IR, I will check on how to use the new ProviderResources in the kubernetes provider.

internal/gatewayapi/resource.go
```diff
@@ -20,6 +20,37 @@ import (
 
 type XdsIRMap map[string]*ir.Xds
 type InfraIRMap map[string]*ir.Infra
+type GatewayClassResources map[string]*Resources
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
arkodg, line 25: did you hand write this ?
you can use a annotation similar to https://github.com/envoyproxy/gateway/blob/988d4ed0be5a5f9dd39a784d52c3bb5c51ecc2aa/internal/gatewayapi/resource.go#L26 to autogenerate it

internal/gatewayapi/runner/runner.go
```diff
@@ -49,118 +49,117 @@ func (r *Runner) Start(ctx context.Context) (err error) {
...
-			// Also validate the ir before sending it.
-			for key, val := range result.InfraIR {
-				r.Logger.WithValues("infra-ir", key).Info(val.YAMLString())
-				if err := val.Validate(); err != nil {
-					r.Logger.Error(err, "unable to validate infra ir, skipped sending it")
-					errChan <- err
-				} else {
-					r.InfraIR.Store(key, val)
-					newKeys = append(newKeys, key)
+			for gc, resources := range *val {
```
arkodg, line 65: lets protect against `val == nil`

internal/gatewayapi/resource.go
```diff
@@ -20,6 +20,37 @@ import (
 
 type XdsIRMap map[string]*ir.Xds
 type InfraIRMap map[string]*ir.Infra
+type GatewayClassResources map[string]*Resources
```
arkodg, line 23: non blocking comment that can be tackled in a new PR - would a list `*Resources[]` be a better option here ?

internal/provider/kubernetes/controller.go
```diff
@@ -165,212 +166,206 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
-				acceptedGC.Name))
-			return reconcile.Result{}, err
-		}
-	} else {
-		// finalize the accepted GatewayClass.
-		if err := r.addFinalizer(ctx, acceptedGC); err != nil {
-			r.log.Error(err, fmt.Sprintf("failed adding finalizer to gatewayclass %s",
-				acceptedGC.Name))
-			return reconcile.Result{}, err
+			// If needed, remove the finalizer from the accepted GatewayClass.
```
arkodg, line 353: looks like the order is changed here a bit, shouldnt we update the finalizer and then run `Store`, so any labels show up in the gatewayapi layer and can be used/if required in the future ?

arkodg:
minor comments, but this is looking good @cnvergence ! massive feature, that im hoping lands in v1.
0.0

internal/gatewayapi/runner/runner.go
cnvergence: checking before dereferencing
```
val := update.Value
if update.Delete || val == nil {
	return
}
```

internal/gatewayapi/resource.go
cnvergence: I did generate it, I will add comment
deepcopy-gen returns GatewayClassResources, I needed to change it to return *GatewayClassResources

internal/provider/kubernetes/controller.go
cnvergence: I have changed the order, so the store was called when the last gateway was removed and then clean up finalizer. 
I could verify that running the store after the finalizers sometimes resulted in not cleaning up infra resources, e.g. deleting all resources at once.

internal/gatewayapi/resource.go
arkodg: because its a `map` ? , if it was a list would we hit this same issue ?

arkodg:
@cnvergence looks like e2e is failing

internal/provider/kubernetes/controller.go
cnvergence: I was wrong, now with the new structure, it should be called afterwards

cnvergence:
fixed those issues

internal/gatewayapi/resource.go
cnvergence: yes, I think it is because of the map

arkodg (approved):
LGTM thanks !

zirain (approved):

shawnh2 (approved):
LGTM, thanks

Issue #2295, opened 2023-12-11
JWT SecurityPolicy does not forward Authorization header by default


akhenakh:
The JWT SecurityPolicy if enabled does not forward the `Authorization` header upstream.

This is the default behaviour in Proxy if `forward` is not set to true:
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#jwtprovider

Yet it is a weird behaviour for a gateway to temper headers from the original request.

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#jwtprovider
relevant code https://github.com/envoyproxy/gateway/blob/f3e4e932d0749f38af339a92ec68cebbf29645ef/internal/xds/translator/jwt.go#L139-L145

Adding `Forward: true` will result of not removing the `Authorization` header.

I'm not sure myself what would be a sane default, but this could become a parameter or at least documented in Gateway.

akhenakh:
Slack discussion https://envoyproxy.slack.com/archives/C03E6NHLESV/p1701708137782169

zhaohuabing:
We have two options here:
1. Expose Envoy `forward` knob to EG `SecurityPolicy` API
2. Set `forward` to `true` when translating the xds

I am leaning towards option 2 because it doesn't harm to pass down the JWT authorization header. WDYT @arkodg @envoyproxy/gateway-maintainers

arkodg:
looks like we went ahead with 2. which was completed with #2300, can this be closed @zhaohuabing ?

zhaohuabing:
> looks like we went ahead with 2. which was completed with #2300, can this be closed @zhaohuabing ?

Correct. We don't have to modify API with option 2. And if we need to give users the choice to move the token out later, we can add it to the API.

PR #2300, opened 2023-12-13
forward jwt tokens after verification

zhaohuabing:
related: https://github.com/envoyproxy/gateway/issues/2295


zirain:
is there a test for this?

zhaohuabing:
> is there a test for this?

I think we don't need to add a test for this, it should work as long as jwt test passed.

zirain:
> > is there a test for this?
> 
> I think we don't need to add a test for this, it should work as long as jwt test passed.

we need make sure it's there, avoid regression in the future.

zhaohuabing:
> > > is there a test for this?
> > 
> > 
> > I think we don't need to add a test for this, it should work as long as jwt test passed.
> 
> we need make sure it's there, avoid regression in the future.

Totally agree if it's a bug. But if you look into the issue, it's actually a feature request other than a repeated bug. We don't want to add an individual e2e test for every single small configuration knob.

And the unit test should have already covered it.

zirain (approved):

Xunzhuo (approved):

Issue #2296, opened 2023-12-12
SecurityPolicy with JWT  extractFromCookies and CORSenabled fails with 401


apjoseph:
*Description*:
The following SecurityPolicy configuration results in 401 errors on OPTIONS requests; SPAs must use http-only cookies and include credentials in requests. However credentials won't be sent until  `Access-Control-Allow-Credentials"true` is returned in the response headers. This results in perpetual denial of requests.

```yaml
  jwt:
    providers:
      - name: auth0
        extractFrom:
          cookies:
            - BearerToken
        remoteJWKS:
          uri: https://web.example.com/.well-known/jwks.json
```

*Environment*:
0.0.0-latest
https://hub.docker.com/layers/envoyproxy/gateway-helm/v0.0.0-latest/images/sha256-246e6cd244ec3a6d6db37acf9a1aba80a4599a6630aa1be40e24875843c7a780?context=explore

zirain:
cc @zhaohuabing

apjoseph:
As an update I also tried manually appending this via an `EnvoyPatchPolicy` which failed with "virtualHosts already exists" 
From my testing, it appears that the patch policy is applied separately from the securitypolicy rather than being merged into it. 
I'd expect the PatchPolicies to be applied AFTER all *Route and SecurityPolicies have been applied and then merged into them.

I'm pretty new to Envoy/EnvoyGateway, but It looks like right now to make a simple fix to a security policy, like adding the missing `allowCredentials` option, I'd basically have to rewrite the routeConfig for the entire https listener and write out every aspect of my routes via xDS -meaning I couldn't apply any securitypolicies or *route CRDs without causing a conflict. 

If this is the case I find this a bit too inflexible for my needs. In generally I'm pretty pleased with this Gateway implementation, but if a simple tweak requires replacing the entire routeConfiguration of a listener, I feel like it defeats the whole separation of concerns idea that the Gateway API is supposed to address. It's basically like going back to the old idea of a statically configured httpd server. 

[...]

PR #2307, opened 2023-12-15
Add AllowCredentials knob to CORS setting

zhaohuabing:
**Why we need this?**
As an app developer, I want to send a cross-origin request to the server with credentials.

An example: https://github.com/envoyproxy/gateway/issues/2296#issuecomment-1857724471

Related: https://github.com/envoyproxy/gateway/issues/2296


zirain (approved):
LGTM

arkodg (approved):
LGTM thanks !

PR #2309, opened 2023-12-15
Add hostNetwork DNSPolicy support

zzjin:
And add support listen port under 1024 (like 443).

**What type of PR is this?**
fix: fix new introduced `hostNetwork` cannot speak to xds bug.

May fix: #2310

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -203,6 +203,33 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 		},
 	}
 
+	// Check if need Set Capabilities when listening on privileged ports
+	if deploymentConfig.Pod.HostNetwork && deploymentConfig.Container.SecurityContext == nil {
+		needCapabilities := false
+		for _, port := range ports {
+			if port.ContainerPort < 1024 {
+				needCapabilities = true
+				break
+			}
+		}
+
+		if needCapabilities {
+			containers[0].SecurityContext = &corev1.SecurityContext{
+				Capabilities: &corev1.Capabilities{
+					Add: []corev1.Capability{
+						"NET_BIND_SERVICE",
+					},
+					Drop: []corev1.Capability{
+						"ALL",
+					},
+				},
+				ReadOnlyRootFilesystem:   ptr.To(true),
+				AllowPrivilegeEscalation: ptr.To(true),
+				RunAsNonRoot:             ptr.To(false),
+			}
+		}
+	}
+
 	return containers, nil
 }
 


Xunzhuo (changes requested):
Plz fix the CI and make sure it is backport compatible.

internal/gatewayapi/listener.go
```diff
@@ -96,7 +96,15 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 
 			// Add the listener to the Xds IR
 			servicePort := &protocolPort{protocol: listener.Protocol, port: int32(listener.Port)}
-			containerPort := servicePortToContainerPort(servicePort.port)
+
+			containerPort := servicePort.port
```
shawnh2, line 100: just out of curiosity, is there any better way to improve this looong hideous nested nil check?

internal/gatewayapi/listener.go
```diff
@@ -96,7 +96,15 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
...
+			if !(resources.EnvoyProxy != nil && resources.EnvoyProxy.Spec.Provider != nil &&
```
shawnh2, line 101: just out of curiosity, is there any better way to improve this looong hideous nested nil check?

shawnh2:
hi @zzjin, you can try `make testdata` to automatically generate the correct test data.

internal/gatewayapi/listener.go
Xunzhuo: refer to https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/envoygateway_helpers.go#L80

make it as a common helper to use

arkodg:
we shouldnt be changing the container port logic, all clients should access the gateway via the service

Xunzhuo:
@arkodg Any better solutions for using EG in such scenarios:

1. a kubernetes cluster without lb implementation
2. expose envoyproxy with 80/443

In this scenario, EG is more like a real loadbalancer, it should be a common request.

arkodg:
@Xunzhuo you can route to the service port (80/443) via External IP

Xunzhuo:
@arkodg Can you elaborate it ?

arkodg:
> @arkodg Can you elaborate it ?

https://www.fadhil-blog.dev/blog/kubernetes-external-ip/

Xunzhuo:
Correct ... @arkodg I think I missed this, thanks for pointing out.

Xunzhuo:
close via https://github.com/envoyproxy/gateway/pull/2374

PR #2319, opened 2023-12-18
Upgrade use of generic ptr to cleaner codebase.

zzjin:
**What type of PR is this?**

chore: change pointer*ptr usage to cleaner codebase.

1. Upgrade usage from deprecated `k8s.io/utils/pointer` to generic version `k8s.io/utils/ptr`
2. Remove `internal/utils/ptr` cause as duplicated
3. Simpler some `ptr.To` call


Xunzhuo (dismissed):
lgtm

api/v1alpha1/kubernetes_helpers.go
```diff
@@ -9,7 +9,7 @@ import (
 	appv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	"k8s.io/utils/pointer"
```
zirain, line 12: can you also update golanglint-ci setting for this package?

zirain: https://github.com/envoyproxy/gateway/blob/f4252e6685a4378a9b2e75f802507711c464dfa6/tools/linter/golangci-lint/.golangci.yml#L27

zzjin: Updated.

zirain (approved):

Xunzhuo (approved):

Issue #2308, opened 2023-12-15
support EnvoyPatchPolicy when enabling mergeGateways


apjoseph:
*Description*:
when mergeGateways is true, EnvoyPatchPolicies will not be picked up.

*Repro steps*:
set mergeGateways to true, create an EnvoyPatchPolicy. 

*Environment*:
> 0.0.0-latest (12/12/2023)

shawnh2:
It seems EnvoyPatchPolicy feature is not supported when `mergeGateways` is enabled.

https://github.com/envoyproxy/gateway/blob/52946bef5132a302b6f5edd6d154062f35906e22/internal/gatewayapi/envoypatchpolicy.go#L40

Currently, EnvoyPatchPolicy can be only applied for per Gateway. I am not sure whether MergeGateways should support EnvoyPatchPolicy? cc @arkodg @cnvergence 

If we do, the `targetRef` of EnvoyPatchPolicy should be `GatewayClass` instead of `Gateway`, it makes more sense to me.

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: custom-response-patch-policy
  namespace: default
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: GatewayClass # Gateway
    name: eg
    namespace: default
```

zirain:
@kflynn PTAL

apjoseph:
As a side comment on patching, I think the generic EnvoyPatchPolicy would benefit from being split up into more targeted CRDs

I'd prefer something like

`HttpRoutePatchPolicy`
`HttpListenerPolicy`

with name/namespace/label-based selectors so that you can target specific HttpRoutes and listeners to patch. I think ~95% of users will mostly be patching filters. So you could have the following  

EnvoyHTTPRoutePatchPolicy

```yaml
filterPatches:
  match:
   - type: type.googleapis.com/envoy.extensions.filters.http.cors.v3.CorsPolicy
     replace:
       - op: add 
         path: /typed_per_filter_config/allow_credentials
         value:  true 
``` 

EnvoyHttpListenerPatchPolicy

```yaml
filterPatches:
  match:
   - type: type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
      - op: add
         path: /typed_filter_config/bypass_cors_preflight
         value:  true 
``` 

[...]

PR #2320, opened 2023-12-18
feat: support mergeGateways in EnvoyPatchPolicy

cnvergence:
**What type of PR is this?**

feat: support mergeGateways in EnvoyPatchPolicy

Support EnvoyPatchPolicy when MergeGateways is enabled and make sure that it targets GatewayClass instead Gateway. As when mergeGateways is enabled, there will be one Envoy deployment for all Gateways in the cluster.

Fixes https://github.com/envoyproxy/gateway/issues/2308


arkodg:
@cnvergence this does make sense, can you please expand on the WHY in the PR description ?

you will also need to update the validation logic https://github.com/envoyproxy/gateway/blob/bf8eaa45b268b2ebf73e44bc3300c73daf51ef67/internal/gatewayapi/envoypatchpolicy.go#L69 to make sure targetRef type is `KindGatewayClass` when `mergeGateways` is true

this feature will also require some additional docs to explain this to the end user

cnvergence:
@arkodg right, I will add this to the GatewayClass level
Also need to fix this https://github.com/envoyproxy/gateway/issues/2308#issuecomment-1864576614

cnvergence:
nvm, it works as expected

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2321, opened 2023-12-19
revert docs-preview action

zirain:
revert this first before found a better solution

cc @arkodg @Xunzhuo

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -6,10 +6,6 @@ on:
     - "release/v*"
     paths:
     - 'site/**'
-  pull_request_target:
-    types: [opened, synchronize, reopened]
-    paths:
-    - 'site/**'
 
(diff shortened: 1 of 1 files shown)


Xunzhuo (commented):
If we do so, you should change back to `pull_request` event, make sure docs lint pass.

Also if we need to revert the `pull_request_target`, we need to revert the `welcome msg` too.

zirain:
> If we do so, you should change back to `pull_request` event, make sure docs lint pass.
> 
> Also if we need to revert the `pull_request_target`, we need to revert the `welcome msg` too.

I forgot that action, can you raise a new PR, I can close this one.

Xunzhuo:
Can you merge them together in this PR ?

arkodg (approved):

AliceProxy (approved):

Issue #2327, opened 2023-12-20
Support insecure Remote JWKS endpoint


ardikabs:
*Description*:
> Follow up from https://github.com/envoyproxy/gateway/issues/1930, regarding support for insecure remote JWKS endpoint.

Currently, JWT policy only supports secure endpoints through HTTPS for the remote JWKS endpoint. For particular reasons such as when the JWKS endpoint is expected to be internally accessible within the Kubernetes cluster, in that case adding a secure layer might not needed.

The workaround that we have done so far was to patch the code to add support when the JWKS endpoint uses `http://...`, thus it will omit the `tls` socket config under the `cluster` config. The example patch can be seen [here](https://github.com/ardikabs/gateway/commit/0b0bc32d580c3cdcbf29beeb51a5f10489069d22). 

With the above patch, you can apply the below configuration.

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  name: sp-any-endpoint
  namespace: envoy-gateway-system
spec:
  jwt:
    providers:
      - claimToHeaders:
          - claim: userId
            header: x-user-id
          - claim: email
            header: x-username
          - claim: sessionId
            header: x-session-id
        name: internal
        remoteJWKS:
          uri: http://auth-service.auth-service-ns.svc.cluster.local/.well-known/jwks
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: hr-any-endpoint
    namespace: envoy-gateway-system
```

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zhaohuabing:
@ardikabs Thanks for the feedback. I think it's reasonable to support plain HTTP jwks and ext auth servers, but OIDC servers need to be HTTPS. WDYT @arkodg ?

PR #2328, opened 2023-12-20
chore: relax https for jwks

zhaohuabing:
This PR relaxes https scheme restriction for JWKS.

Related: https://github.com/envoyproxy/gateway/issues/2327


internal/gatewayapi/testdata/securitypolicy-with-jwt.in.yaml
```diff
@@ -86,11 +86,11 @@ securityPolicies:
         - header: one-route-example-key
           claim: claim1
       - name: example2
-        issuer: https://two.example.com
+        issuer: http://two.example.com
```
arkodg, line 89: we shouldn't keep the test cases to `https` since thats the default case imo

zhaohuabing: There're two jwks in this test file, the first one is `https`, the changed one is the second.

arkodg (dismissed):
LGTM thanks

zirain (dismissed):

internal/xds/translator/jwt.go
ardikabs, line 1: Hi @zhaohuabing , i think you also need to adjust this [line](https://github.com/envoyproxy/gateway/pull/2328/files#diff-8837c552f2ec37f96df5dcb1f648e6282d2935678e26d2b80e0453264048ce9fL276-L280) as well, otherwise when the remote jwks endpoint uses `http`, it will end up with creating a `tls` config for a non-https endpoint.

internal/xds/translator/utils.go
```diff
@@ -66,6 +67,7 @@ func url2Cluster(strURL string) (*urlCluster, error) {
 		hostname:     u.Hostname(),
 		port:         uint32(port),
```
ardikabs, line 68: we also need to have a default port for `http`, otherwise the user is required to explicitly mention the port when `http` is being used, in this case, `80`.

internal/xds/translator/testdata/in/xds-ir/jwt-multi-route-multi-provider.yaml
```diff
@@ -21,12 +21,12 @@ http:
...
           claim: claim.neteased.key
       - name: example2
-        issuer: https://www.two.example.com
+        issuer: http://www.two.example.com
         audiences:
         - one.foo.com
         - two.foo.com
         remoteJWKS:
-          uri: https://192.168.1.250:8080/jwt/public-key/jwks.json
+          uri: http://192.168.1.250:8080/jwt/public-key/jwks.json
```
arkodg, line 29: can we make the `localhost` endpoint the `http` endpoint and make this a `https` endpoint to mimic a real use

internal/xds/translator/testdata/in/xds-ir/jwt-multi-route-multi-provider.yaml
```diff
@@ -49,7 +49,7 @@ http:
         audiences:
         - foo.com
         remoteJWKS:
-          uri: https://localhost/jwt/public-key/jwks.json
+          uri: http://ejwks.json
```
ardikabs, line 52: seems typo here?

arkodg (approved):

PR #2329, opened 2023-12-20
fix: allow envoyHpa maxReplicas to be equal to minReplicas

shahar-h:
According to hpa [spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#horizontalpodautoscalerspec-v2-autoscaling) `maxReplicas` cannot be less that `minReplicas`. The current cel validation also forbids it to be equal to `minReplicas`.
This PR fixes the validation to be aligned with the spec and also adds a test.

### DIFF
--- a/api/v1alpha1/envoyproxy_types.go
+++ b/api/v1alpha1/envoyproxy_types.go
@@ -134,7 +134,7 @@ type EnvoyProxyKubernetesProvider struct {
 	// +optional
 	// +kubebuilder:validation:XValidation:message="minReplicas must be greater than 0",rule="!has(self.minReplicas) || self.minReplicas > 0"
 	// +kubebuilder:validation:XValidation:message="maxReplicas must be greater than 0",rule="!has(self.maxReplicas) || self.maxReplicas > 0"
-	// +kubebuilder:validation:XValidation:message="maxReplicas cannot be less than or equal to minReplicas",rule="!has(self.minReplicas) || self.maxReplicas > self.minReplicas"
+	// +kubebuilder:validation:XValidation:message="maxReplicas cannot be less than minReplicas",rule="!has(self.minReplicas) || self.maxReplicas >= self.minReplicas"
 	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/envoyproxy_types.go
```diff
@@ -134,7 +134,7 @@ type EnvoyProxyKubernetesProvider struct {
 	// +optional
 	// +kubebuilder:validation:XValidation:message="minReplicas must be greater than 0",rule="!has(self.minReplicas) || self.minReplicas > 0"
 	// +kubebuilder:validation:XValidation:message="maxReplicas must be greater than 0",rule="!has(self.maxReplicas) || self.maxReplicas > 0"
-	// +kubebuilder:validation:XValidation:message="maxReplicas cannot be less than or equal to minReplicas",rule="!has(self.minReplicas) || self.maxReplicas > self.minReplicas"
+	// +kubebuilder:validation:XValidation:message="maxReplicas cannot be less than minReplicas",rule="!has(self.minReplicas) || self.maxReplicas >= self.minReplicas"
```
shawnh2, line 137: do we have any user case here? 

i don't think setting `maxReplicas` equal to `minReplicas` is necessary. if we do, what's point of using HPA? suggest using replicas setting directly.
https://github.com/envoyproxy/gateway/blob/c2f88f0e9be766d3cd8eb75c686277ebf2da1b43/api/v1alpha1/shared_types.go#L58

any inputs? cc @envoyproxy/gateway-maintainers

shahar-h: No specific use case, I just think maxReplicas should be aligned with the spec.

zhaohuabing: I suggest leaving the current validation as it is unless we have a strong reason to change it.

zhaohuabing (commented):
I'm curious why K8s allows `maxReplicas` to be equal to `minReplicas`. 

imo, EG should leave the current validation as it is: `maxReplicas` must bigger than `minReplicas`, unless we have a strong reason to change it.

arkodg:
I would accept this PR to handle one use case - debugging something in EG while having hpa enabled but temporarily keeping the replicas constant, this can be achieved using replicas, but requires the user to delete and readd more config rather than simply editing fields

zhaohuabing:
> I would accept this PR to handle one use case - debugging something in EG while having hpa enabled but temporarily keeping the replicas constant, this can be achieved using replicas, but requires the user to delete and readd more config rather than simply editing fields

Ha, this is a legitimate use case. I guess that's also why k8s allows that.

zhaohuabing (approved):
LGTM. Thanks!

Xunzhuo (approved):

PR #2330, opened 2023-12-20
feat(translator): Implement BTP CircuitBreaker API

guydc:
**What type of PR is this?**
Feature, implementation of API defined in https://github.com/envoyproxy/gateway/pull/2284

Add support for configuring Envoy's [circuit breaking](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking) settings. 

Fixes https://github.com/envoyproxy/gateway/issues/2125


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -482,3 +496,57 @@ func (t *Translator) buildProxyProtocol(policy *egv1a1.BackendTrafficPolicy) *ir
...
+				return nil
+			}
+		}
+	}
+
+	return cb
+}
+
+func setCircuitBreakerPolicyErrorCondition(policy *egv1a1.BackendTrafficPolicy) {
+	message := "Unable to translate Circuit Breaker"
```
arkodg, line 539: can we pass an extra `errMsg` to the func to add some more error msg details such as `invalid maxPendingRequests value` etc

arkodg:
the PR is looking really good, thanks for adding an e2e as well !
added a minor comment around err message verbosity in the policy status

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -482,3 +496,57 @@ func (t *Translator) buildProxyProtocol(policy *egv1a1.BackendTrafficPolicy) *ir
...
+	status.SetBackendTrafficPolicyCondition(policy,
+		gwv1a2.PolicyConditionAccepted,
+		metav1.ConditionFalse,
+		gwv1a2.PolicyReasonInvalid,
+		message,
+	)
+}
+func int64ToUint32(in int64) (uint32, bool) {
```
shawnh2, line 547:
```suggestion

func int64ToUint32(in int64) (uint32, bool) {
```

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -482,3 +496,57 @@ func (t *Translator) buildProxyProtocol(policy *egv1a1.BackendTrafficPolicy) *ir
...
+			if ui32, ok := int64ToUint32(*pcb.MaxConnections); ok {
+				cb.MaxConnections = &ui32
+			} else {
+				setCircuitBreakerPolicyErrorCondition(policy)
+				return nil
+			}
+		}
+
+		if pcb.MaxParallelRequests != nil {
+			if ui32, ok := int64ToUint32(*pcb.MaxParallelRequests); ok {
```
shawnh2, line 517: can we add some test here? make sure `int64ToUint32` works as expected

shawnh2 (approved):
some nits, lgtm, thanks

Xunzhuo (approved):
Thanks, this PR overall looks good ! Pretty clear. Would love to add more e2e tests and a user guide docs in a follow-up.

cnvergence (approved):

internal/gatewayapi/backendtrafficpolicy.go
guydc: done

internal/gatewayapi/backendtrafficpolicy.go
guydc: done

PR #2332, opened 2023-12-21
Run CI on k8s v1.29.0

arkodg:
* rm'd v1.26 for now, since we are only running last 3 versions

* k8s support matrix should be revisited pre GA


zirain (approved):

zhaohuabing:
> * rm'd v1.26 for now, since we are only running last 3 versions
> * k8s support matrix should be revisited pre GA

Created an issue to follow this https://github.com/envoyproxy/gateway/issues/2333

zhaohuabing (approved):

Issue #2312, opened 2023-12-18
JWT SecurityPolicy blocks CORS preflight requests, breaking OIDC + JWT flow


apjoseph:
*Description*:
the JWT configuration in the Security filter does not include the `bypass_cors_preflight` [configuration option](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtauthentication). Without this setting, OPTIONS requests are blocked unless the JWT is provided. 

If the JWT is being passed through cookies with a standard OIDC frontend / JWT backend setup then the credentials won't be included since the preflight request needs to include the   `Access-Control-Allow-Credentials: true` response header` before credentials will be sent.

I think bypass_cors_preflight should be set to true by default. I can't think of a good reason why someone would want to require a JWT for OPTIONS requests since Envoy doesn't even send OPTIONS requests down to the backend service. As a result the backend application can't customize the response. So blocking by default will just confuse people while providing no real benefit.

zirain:
cc @zhaohuabing

zhaohuabing:
I agree that `bypass_cors_preflight` should be set by true by default. But I'm not 100% if there're any security concerns on this. Would love more inputs from@arkodg @envoyproxy/web-maintainers .

apjoseph:
@zhaohuabing Another potential solution is to allow different security policies to be applied when a different method is specified with the same path. This was actually my attempted workaround when I first encountered the issue. I created two http routes:

`api-route`

```
        matches:
          - path:
              type: PathPrefix
              value: /
```

api-route-cors
```
        matches:
          - path:
              type: PathPrefix
              value: /
           method: OPTIONS
```              

and then created corresponding SecurityPolicies for both HttpRoutes. I only included the JWT section in the SecurityPolicy for `api-route`.  I assumed this would prevent OPTIONS requests from being blocked; however when looking at the response from egctl I found that it was being applied to both HttpRoutes BUT only in some instances. Other times it would remove JWT auth from BOTH HttpRoutes -depending on which security policy was applied last. 

I am not sure whether this is intended behavior or not, but it was counter intuitive to me as I would have expected all attributes of a match to be taken into account when applying a security policy if they are in separate routes -especially since the security policy is explicitly bound to an HTTPRoute

[...]

PR #2336, opened 2023-12-21
enable BypassCorsPreflight in JWT by default

zhaohuabing:
Fix: https://github.com/envoyproxy/gateway/issues/2312


arkodg:
@zhaohuabing can you elaborate on why setting a `cors` setting is not the suggested approach ?

zhaohuabing:
hold

PR #2337, opened 2023-12-21
feat: add ImagePullSecrets to PodSpec

shahar-h:
Add ImagePullSecrets to PodSpec in order to support pulling `EnvoyProxy` and `RateLimit` images from private registries.
Use cases:
* Custom images
* Pulling images from a private mirror


zirain:
I think we can learn from istio project, introduce [K8SObjectOverlay](https://github.com/istio/api/blob/master/operator/v1alpha1/operator.pb.go#L1085), no need to add all the fields one by one.

arkodg:
@zirain the UX for https://github.com/envoyproxy/gateway/pull/2337#issuecomment-1866229984 is not great.
I do agree we need some generic way to patch k8s fields we dont have native support for

Pulling Envoy and RateLimit from air gapped & private registries is a common use case imo

zirain:
> @zirain the UX for [#2337 (comment)](https://github.com/envoyproxy/gateway/pull/2337#issuecomment-1866229984) is not great. I do agree we need some generic way to patch k8s fields we dont have native support for
> 
> Pulling Envoy and RateLimit from air gapped & private registries is a common use case imo

it's hard to say which one is common or not.

arkodg (approved):

zirain (approved):

PR #2338, opened 2023-12-21
feat(helm): support resources configuration for certgen job

shahar-h:
Add support for configuring resources requests and limits for `certgen` job.


zhaohuabing:
Is the default enough for certgen?

shahar-h:
> The default isn't enough for certgen?

Which default?

shahar-h:
`e2e-test` failed with:
```log
Error: Unable to download artifact(s): Unable to download and extract artifact: Request timeout: /actions-results/e317eb1c-05a8-4e46-b13a-4417a5c7835b/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/artifacts/envoy-gateway.zip?se=2023-12-22T16%3A56%3A31Z&sig=NyfXAH1AY5MjtpA3udLo7flsDsdt4hwmFPn5M%2B8i0B8%3D&sp=r&spr=https&sr=b&st=2023-12-22T16%3A46%3A31Z&sv=2021-12-02
```
How can I retrigger it?

shahar-h:
> > The default isn't enough for certgen?
> 
> Which default?

@zhaohuabing can you please reply?

zhaohuabing:
> > > The default isn't enough for certgen?
> > 
> > 
> > Which default?
> 
> @zhaohuabing can you please reply?

Sorry for the late response. I mean isn't the default resource enough for certgen?

zhaohuabing:
> > > > > The default isn't enough for certgen?
> > > > 
> > > > 
> > > > Which default?
> > > 
> > > 
> > > @zhaohuabing can you please reply?
> > 
> > 
> > Sorry for the late response. I mean isn't the default resource enough for certgen?
> 
> In case you meant reusing [deployment.envoyGateway.resources](https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/values.tmpl.yaml#L7-L13) also for certgrn job, I guess it's enough but since certgen probably requires much more lower requests than EG I think it's a good idea to add its own resources configuration.

Oops, My bad. I just noticed that Kubernetes doesn’t provide default resource limits out-of-the-box. I think we should define the limit for certgen in EG. But since it's a quick an lightweight task, we might not even expose it through the API.

shahar-h:
@zhaohuabing you can provide default values for certgen requests and limits, so they can still be overridden, same as for EG pod.

zhaohuabing:
> @zhaohuabing you can provide default values for certgen requests and limits, so they can still be overridden, same as for EG pod.

I think users probably won't need to override the requests and limits for certgen. Unlike EG or Envoy, whose resource consumption depends on the scale of the application,  certgen's resources are quite steady and predictable.

shahar-h:
@zhaohuabing maybe we can start with allowing users to configure `certgen` job resources via chart values and at later point provide internal defaults? Anyway it's optional so I don't see how it will cause any issue.
It's important to have resources for all workloads since there are policy engines that validate that, i.e. [Kyverno require-requests-limits policy](https://kyverno.io/policies/best-practices/require-pod-requests-limits/require-pod-requests-limits/).

zhaohuabing:
> @zhaohuabing maybe we can start with allowing users to configure `certgen` job resources via chart values and at later point provide internal defaults? Anyway it's optional so I don't see how it will cause any issue.
> It's important to have resources for all workloads since there are policy engines that validate that, i.e. [Kyverno require-requests-limits policy](https://kyverno.io/policies/best-practices/require-pod-requests-limits/require-pod-requests-limits/).
> 

I have no objection with settings a reasonable default, just don't want to expose it to the API because certgen won't scale up or down. However, I may be wrong, wait other @envoyproxy/gateway-maintainers to chime in.

zirain:
IMO, [advanced helm chart customization](https://istio.io/latest/docs/setup/additional-setup/customize-installation-helm/#advanced-helm-chart-customization) will help this.

shahar-h:
@zhaohuabing I think that resources should be configurable from chart values because some users would prefer for example to omit CPU limits in order to avoid CPU throttling. On the other hand some users would prefer to define CPU limits in order to make sure other workloads get more CPU time. See https://home.robusta.dev/blog/stop-using-cpu-limits.

arkodg (approved):

zirain (approved):

Issue #2343, opened 2023-12-22
Envoy rejects invalid RouteConfiguration


zhaohuabing:
*Description*:

Envoy rejects invalid routes.

HTTPRoute
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend
  namespace: default
  resourceVersion: "12821"
  uid: 0a2738a5-3bb6-491d-8fee-35203d92ba5f
spec:
  hostnames:
  - www.example.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
      weight: 1
    matches:
    - path:
        type: RegularExpression
        value: '*.foo.bar.com'
status:
  parents:
  - conditions:
    - lastTransitionTime: "2023-12-22T04:49:11Z"
      message: Route is accepted
      observedGeneration: 3
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2023-12-22T04:49:11Z"
      message: Resolved all the Object references for the Route
      observedGeneration: 3
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: gateway.envoyproxy.io/gatewayclass-controller
    parentRef:
      group: gateway.networking.k8s.io
      kind: Gateway
      name: eg
```

Envoy log

```
[2023-12-22 04:49:11.841][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: no argument for repetition operator: *
```

PR #2344, opened 2023-12-22
validate regex before sending to Envoy

zhaohuabing:
**What type of PR is this?**
This PR adds validation to `RegularExpression` to make sure it won't be rejected by Envoy.

Fixes https://github.com/envoyproxy/gateway/issues/2343


zirain (dismissed):

internal/xds/bootstrap/bootstrap.go
```diff
@@ -204,3 +208,10 @@ func GetRenderedBootstrapConfig(proxyMetrics *egv1a1.ProxyMetrics) (string, erro
 
 	return cfg.rendered, nil
 }
+
+func validateRegex(regex string) error {
```
arkodg, line 212: can we move this into https://github.com/envoyproxy/gateway/tree/main/internal/utils so it can be reused instead of defining it twice ?

arkodg:
thanks for adding more guardrails @zhaohuabing ! added a minor comment

zirain (approved):

arkodg (approved):

Issue #2339, opened 2023-12-21
retest not working any more?


zirain:
https://github.com/envoyproxy/gateway/pull/2320#issuecomment-1866892220

https://github.com/envoyproxy/gateway/actions/runs/7292906880/job/19874879024

zirain:
https://github.com/envoyproxy/gateway/actions/runs/7325748312/job/19950658377

zirain:
still need AZP?

zirain:
https://github.com/envoyproxy/gateway/actions/runs/7326349471/job/19951914308

arkodg:
@phlax has something recently changed with this gha ?

zirain:
after rollback, the action is normal now, https://github.com/envoyproxy/gateway/pull/2374#issuecomment-1877992501

zirain:
keep this open, we want reenable latest version in the future.

zirain:
xref: https://github.com/envoyproxy/toolshed/issues/1377

PR #2349, opened 2023-12-23
ci: fix retest command

zirain:
ref: https://github.com/envoyproxy/envoy/blob/main/.github/workflows/command.yml

fixes: https://github.com/envoyproxy/gateway/issues/2339


zhaohuabing (approved):

zirain:
I'd like merge this first to verify if everything works as expected.

PR #2358, opened 2023-12-25
api: Support suppressing the 'x-envoy-' headers added by the HTTP router

liorokman:
**What type of PR is this?**
This PR enhances the `ClientTrafficPolicy`  by adding the following option:

```yaml
   # Suppress the "x-envoy-" headers from being added by the HTTP router.
   suppressEnvoyHeaders: true
```

This PR adds the ability to suppress the "x-envoy-" headers added by the Envoy HTTP router.

The following headers are affected: 

* Upstream:
  *  [x-envoy-attempt-count](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-attempt-count)
  * [x-envoy-is-timeout-retry](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-is-timeout-retry)
  * [x-envoy-expected-rq-timeout-ms](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-expected-rq-timeout-ms)
  * [x-envoy-original-path](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-original-path)
  * [x-envoy-upstream-stream-duration-ms](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-upstream-stream-duration-ms)
* Downstream:
  * [x-envoy-upstream-service-time](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-upstream-stream-time)
  * [x-envoy-overloaded](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-overloaded)

Support for this flag is mentioned in #1846


zhaohuabing:
This will change both the downstream and upstream behavior. Would it be appropriate to put this in `ClientTraffciPolicy`?

zhaohuabing (commented):
This will change both the downstream and upstream behavior. Would it be appropriate to put this in ClientTraffciPolicy?

liorokman:
The relevant Envoy Proxy configuration is this: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto#extensions-filters-http-router-v3-router .

It affects both upstream and downstream. It's not possible to affect just the upstream or just the downstream - Envoy Proxy doesn't give that option.

I think that both `ClientTrafficPolicy` and `BackendTrafficPolicy` are equally unfit, but that `ClientTrafficPolicy` is still the better fit. I think that users that look for something like this are looking for a way to prevent their downstream clients from receiving these headers.

Do you think there's a better place in the Envoy Gateway configuration to specify this?

zhaohuabing:
> I think that both `ClientTrafficPolicy` and `BackendTrafficPolicy` are equally unfit, but that `ClientTrafficPolicy` is still the better fit. I think that users that look for something like this are looking for a way to prevent their downstream clients from receiving these headers.

Looks like this option only suppresses the headers generated by the router.

```
x-envoy-attempt-count
x-envoy-is-timeout-retry
x-envoy-expected-rq-timeout-ms
x-envoy-original-path
x-envoy-upstream-stream-duration-ms
```
And there are many more "x-envoy-xxx" headers generated by other filters, here is the full list: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers

So I wonder whether this will work for the use case you mentioned or not.

> Do you think there's a better place in the Envoy Gateway configuration to specify this?
Unfortunately, I don't have one either. EG doesn't have a home for such an option that affects both the downstream and upstream.  I agree that we can settle with one of them.

liorokman:
> So I wonder whether this will work for the use case you mentioned or not.

Looking at the list of headers mentioned in the [HTTP connection manager documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers):

* x-envoy-downstream-service-cluster is only added if the [add_user_agent](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-add-user-agent) option is set to true
* x-envoy-downstream-service-node is only added if the add_user_agent is set to true
* x-envoy-force-trace is relevant if it's already set in the incoming request
* x-envoy-internal was only sent upstream in my tests, and removed from the response, so no issues here.
* x-envoy-original-dst-host needs to be enabled explicitly with the [use_http_header](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-originaldstlbconfig-use-http-header) configuration.


This only leaves the [x-envoy-external-address](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#id15) header.

zhaohuabing (approved):
LGTM. Defer to other @envoyproxy/gateway-maintainers

zirain (approved):
LGTM

liorokman:
@zirain @zhaohuabing Can I submit a PR that implements this?

zirain:
This's a small changes, you can do the implement here, or you can send another one later.

PR #2359, opened 2023-12-25
feat(helm): support affinity configuration for EnvoyGateway pod

shahar-h:
**What type of PR is this?**
Feature - Add support for affinity configuration in `EnvoyGateway` pod.

A sample use case for affinity configuration is making `EnvoyGateway` HA on different nodes (By adding AntiAffinity rule).

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
@@ -33,6 +33,10 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
+      {{- with .Values.deployment.pod.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
       - args:
         - server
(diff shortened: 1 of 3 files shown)


zirain (changes requested):
please add a test for this

shahar-h:
> please add a test for this

How? There is currently no concept of helm chart testing in this repo unless I missed it.

zirain:
you can add an unit test or e2e

shahar-h:
@zirain Since I only made changes to helm chart, the only test that I could think of is a test that executes `helm template` with custom values and validates the output. However, there are no such tests currently. Do you want me to introduce a new concept of helm tests? something like [this](https://github.com/gruntwork-io/terratest/tree/master/examples/helm-basic-example)? If yes, should it be part of this PR or maybe it requires a prior discussion? Any other options that you can think of?

zirain:
I'm thinking of adding golden file test.

shahar-h:
So something like [this](https://medium.com/@zelldon91/advanced-test-practices-for-helm-charts-587caeeb4cb#2f01)? Should I start implement it in the current PR or it needs a further discussion on a separate issue?

shahar-h:
@zirain After reading a bit about golden files tests I feel that it should not be in the scope of this PR, and it requires a further discussion as part of a new issue.

zirain (approved):

zhaohuabing (approved):
LGTM. Thanks!

PR #2364, opened 2023-12-26
feat: Suppress envoy headers

liorokman:

This PR implements #2358 

Implements #2358


shawnh2 (approved):
lgtm

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -349,3 +353,14 @@ func translateListenerProxyProtocol(enableProxyProtocol *bool, httpIR *ir.HTTPLi
 		httpIR.EnableProxyProtocol = true
 	}
 }
+
+func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *ir.HTTPListener) {
+	// Return early if not set
+	if suppressEnvoyHeaders == nil {
```
Xunzhuo, line 359:
```go
**func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *ir.HTTPListener) {
	if suppressEnvoyHeaders != nil {
		httpIR.SuppressEnvoyHeaders = *suppressEnvoyHeaders
	}
}
```

How about simplifying it like this way?

liorokman: Done.

Xunzhuo (approved):

zirain (approved):

arkodg:
we may need to revisit where this field lives in the future and consider moving it into `BackendTrafficPolicy` if
* this feature is something that is tied to the developer persona instead of the system admin persona
* if there are valid use cases where a user would like to enable this feature at the `Gateway` level (all HTTPRoutes ) by default but disable it for some explicit routes (using a explicit `BackendTrafficPolicy` targeted at the route level)

PR #2367, opened 2023-12-27
refactor: update watch ns api & add tests

qicz:
**What type of PR is this?**

Fixes #


arkodg (approved):
good catch !

zirain (approved):

PR #2368, opened 2023-12-28
fix: fix backtrafficpolic is applied to routes with the same prefix

### LINKED ISSUE #2366: Using backendTrafficPolicies to patch  route resources may lead to unexpected results.
*Description*:

Using backendTrafficPolicies to patch  httproute resources may lead to unexpected results.

For example:
```
- apiVersion: gateway.networking.k8s.io/v1
  kind: Gateway
  metadata:
    namespace: envoy-gateway
    name: gateway-1
  spec:
    listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
httpRoutes:
- apiVersion: gateway.networking.k8s.io/v1
  kind: HTTPRoute
  metadata:
    namespace: default
    name: httproute-1
  spec:
    hostnames:
    - gateway.envoyproxy.io
    parentRefs:
    - namespace: envoy-gateway
      name: gateway-1
      sectionName: http
    rules:
    - matches:
      - path:
          value: "/"
      backendRefs:
      - name: service-1
        port: 8080
- apiVersion: gateway.networking.k8s.io/v1
  kind: HTTPRoute
  metadata:
    namespace: default
    name: httproute-1-mirror
  spec:
    hostnames:
    - gateway.envoyproxy.io
    parentRefs:
    - namespace: envoy-gateway
      name: gateway-1
      sectionName: http
    rules:
    - matches:
      - path:
          value: "/mirror"
      backendRefs:
      - name: service-1
        port: 8080
  kind: BackendTrafficPolicy
  metadata:
    namespace: default
    name: policy-for-route
  spec:
    targetRef:
      kind: HTTPRoute
      name: httproute-1
      namespace: default
    rateLimit:
      type: Global
      global:
        rules:
        - clientSelectors:
          - sourceCIDR:
              type: "Distinct"
              value: 192.168.0.0/16
          limit:
            requests: 20
            unit: Hour
```
backendTrafficPolicies will be patched to `httproute-1` and `httproute-1-mirror`

Reason:
https://github.com/envoyproxy/gateway/blob/bda0055a2cc53af10bbf7ca9b1834bb960e37a07/internal/gatewayapi/helpers.go#L381
https://github.com/envoyproxy/gateway/blob/bda0055a2cc53af10bbf7ca9b1834bb960e37a07/internal/gatewayapi/backendtrafficpolicy.go#L270

Maybe add `/`  to the prefix `%s/%s/%s/`

tmsnan:
**What type of PR is this?**

Using backendTrafficPolicies to patch route resources may lead to unexpected results.

Fixes #2366

### DIFF
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -378,15 +378,15 @@ func irUDPListenerName(listener *ListenerContext, udpRoute *UDPRouteContext) str
 }
 
 func irRoutePrefix(route RouteContext) string {
-	return fmt.Sprintf("%s/%s/%s", strings.ToLower(string(GetRouteType(route))), route.GetNamespace(), route.GetName())
+	return fmt.Sprintf("%s/%s/%s/", strings.ToLower(string(GetRouteType(route))), route.GetNamespace(), route.GetName())
 }
 
 func irRouteName(route RouteContext, ruleIdx, matchIdx int) string {
-	return fmt.Sprintf("%s/rule/%d/match/%d", irRoutePrefix(route), ruleIdx, matchIdx)
+	return fmt.Sprintf("%srule/%d/match/%d", irRoutePrefix(route), ruleIdx, matchIdx)
 }
 
 func irRouteDestinationName(route RouteContext, ruleIdx int) string {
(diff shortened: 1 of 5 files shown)


zhaohuabing (commented):
`SecurityPolicy` has the same issue, please also fix it.

PR #2369, opened 2023-12-28
fix: perfix match

zhaohuabing:
fix https://github.com/envoyproxy/gateway/issues/2366

### DIFF
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -263,7 +263,7 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 	}
 
 	// Apply IR to all relevant routes
-	prefix := irRoutePrefix(route)
+	prefix := irRoutePrefix(route) + "/" // Prevent mismatching routes with the same prefix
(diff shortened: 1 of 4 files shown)


zhaohuabing:
Close this one. I didn't notice that you already opened a PR @tmsnan

tmsnan:
@zhaohuabing, this is better, I have closed the other one 🤣

zirain (approved):

shawnh2 (approved):

arkodg:
@zhaohuabing @tmsnan thanks for finding and fixing this.
shouldnt the fix be to add `/` to the prefix helper function ?

zhaohuabing:
> @zhaohuabing @tmsnan thanks for finding and fixing this. shouldnt the fix be to add `/` to the prefix helper function ?

I thought about it. I'm a bit uncomfortable to change`%s/rule/%d/match/%d` to `%srule/%d/match/%d`. `/` is not part of the prefix but a delimiter.

arkodg:
@zhaohuabing one concern I have is, we may need to continue duplicating code (add the extra `/`) in the future instead of reusing it

zhaohuabing:
Agree. Fixed in https://github.com/envoyproxy/gateway/pull/2394

Issue #2346, opened 2023-12-22
Add ability to set k8s fields not defined in the EnvoyProxy provider config


arkodg:
*Description*:
Envoy Gateway allows users to customize the Envoy Proxy fleet resources such as - Deployment, Service and HPA. These customizations are a subset of the the entire resource spec

https://github.com/envoyproxy/gateway/blob/5d691f2b312fb1fadce7cd7220759334f796ee6d/api/v1alpha1/shared_types.go#L54

Would be great if we can add a generic way to patch/set any field in the above resources so
* users can make edits without waiting for the field to be added and exposed in a future release
* EG doesnt need to add all fields, especially ones such as `hostNetwork` that are not common case scenarios

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Xunzhuo:
@zirain Thoughts on this ?

zirain:
https://github.com/envoyproxy/gateway/pull/2337#issuecomment-1866229984

arkodg:
we could allow users to specify a merge patch (strategic or json merge patch) https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#use-a-strategic-merge-patch-to-update-a-deployment

arkodg:
thoughts on ?
```
merge:
  type: StrategicMerge
  spec:
    .......
```

zirain:
> thoughts on ?
> 
> ```
> merge:
>   type: StrategicMerge
>   spec:
>     .......
> ```

looks good to me

PR #2374, opened 2023-12-29
feat: support mergepatch to envoyproxy/ratelimit deployment

Xunzhuo:
**What type of PR is this?**

feat(infra): support mergepatch to envoyproxy/ratelimit deployment

Add unsupported fields in deployment, for private customizations like:

```yaml
kind: EnvoyProxy
metadata:
  name: mergepatch
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        patch:
          type: StrategicMerge
          value:
            spec:
              template:
                spec:
                  hostNetwork: true
---
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: mergepatch
```

Fixes https://github.com/envoyproxy/gateway/issues/2346


api/v1alpha1/kubernetes_helpers.go
```diff
@@ -112,3 +117,40 @@ func (hpa *KubernetesHorizontalPodAutoscalerSpec) setDefault() {
 		hpa.Metrics = DefaultEnvoyProxyHpaMetrics()
 	}
 }
+
+// ApplyMergePatch applies a merge patch to a deployment based on the merge type
+func (deployment *KubernetesDeploymentSpec) ApplyMergePatch(old *appv1.Deployment) (*appv1.Deployment, error) {
```
zirain, line 122: IMO, this more like a util function.

api/v1alpha1/shared_types.go
```diff
@@ -333,3 +339,22 @@ type KubernetesHorizontalPodAutoscalerSpec struct {
...
+	JSONMerge MergeType = "JSONMerge"
+)
+
+// KubernetesMergeSpec defines how to perform the merge operation
+type KubernetesMergeSpec struct {
+	// Type is the type of merge operation to perform
+	Type MergeType `json:"type"`
+
+	// Object contains the raw configuration for merged object
+	Object runtime.RawExtension `json:"object"`
```
zirain, line 359: what the different with `apiextensionsv1.JSON`?

api/v1alpha1/shared_types.go
```diff
@@ -52,6 +53,11 @@ const (
 
 // KubernetesDeploymentSpec defines the desired state of the Kubernetes deployment resource.
 type KubernetesDeploymentSpec struct {
+	// Merge defines how to perform the merge operation to deployment
+	//
+	// +optional
+	Merge *KubernetesMergeSpec `json:"merge,omitempty"`
```
arkodg, line 59: thoughts on calling this field `patch`, so it would look like
```
patch:
  type: StrategicMerge
  spec: 
```

api/v1alpha1/shared_types.go
```diff
@@ -333,3 +339,22 @@ type KubernetesHorizontalPodAutoscalerSpec struct {
...
+	// JSONMerge indicates a JSON merge patch type
+	JSONMerge MergeType = "JSONMerge"
+)
+
+// KubernetesMergeSpec defines how to perform the merge operation
+type KubernetesMergeSpec struct {
+	// Type is the type of merge operation to perform
+	Type MergeType `json:"type"`
+
+	// Object contains the raw configuration for merged object
```
arkodg, line 358: thoughts on creating a unique struct per type -
* KubernetesDeploymentPatchSpec
* KubernetesServicePatchSpec
which can define a more well defined `spec` field ?

Xunzhuo: I think the patch spec should be dynamic like envoypatchpolicy jsonpatch value

Xunzhuo: If we define more specific fields, aren`t we using the previous way, then what is the value of dynamic patch? I did not think of what common field is more needed in spec, can you elaborate ?

api/v1alpha1/shared_types.go
Xunzhuo: Mostly the same, but rawextension has a wrapper about object for object converting.

But for mergepatch, apiextensionsv1.JSON is enough.

api/v1alpha1/shared_types.go
Xunzhuo: Done

api/v1alpha1/kubernetes_helpers.go
Xunzhuo: Yes, since ratelimit and proxy are both using it, keeping it in kubernetes_helpers is a common util too.

api/v1alpha1/shared_types.go
arkodg: my recommendation is to use a stricter type such as https://pkg.go.dev/k8s.io/api/apps/v1#DeploymentSpec
this is a little different that our field https://github.com/envoyproxy/gateway/blob/316a0a215ca4c040697bc0231f2ded9dabf59909/api/v1alpha1/shared_types.go#L54 since we are explicit about fields being set, reg `patch`, we cannot garuntee same o/p across releases since internally we may change implementation i.e. append more labels etc

arkodg: the advantage of using a strict type here is that patch semantic errors are surfaced to the user before even patching the two configs together

arkodg: I took another peek into this, its not possible to use the actual Deployment or Service type here since those CRD validations will kick in, which are not valid when its a patch
resolving this one

api/v1alpha1/shared_types.go
```diff
@@ -333,3 +339,22 @@ type KubernetesHorizontalPodAutoscalerSpec struct {
...
+	JSONMerge MergeType = "JSONMerge"
+)
+
+// KubernetesPatchSpec defines how to perform the patch operation
+type KubernetesPatchSpec struct {
+	// Type is the type of merge operation to perform
+	Type MergeType `json:"type"`
+
+	// Object contains the raw configuration for merged object
+	Object apiextensionsv1.JSON `json:"object"`
```
arkodg, line 359: prefer if this field is called `value` which matches what is used in other APIs in the project
https://gateway.envoyproxy.io/latest/api/extension_types/#jsonpatchoperation
https://gateway.envoyproxy.io/latest/api/extension_types/#proxybootstrap

api/v1alpha1/shared_types.go
```diff
@@ -333,3 +339,22 @@ type KubernetesHorizontalPodAutoscalerSpec struct {
...
+	// StrategicMerge indicates a strategic merge patch type
+	StrategicMerge MergeType = "StrategicMerge"
+	// JSONMerge indicates a JSON merge patch type
+	JSONMerge MergeType = "JSONMerge"
+)
+
+// KubernetesPatchSpec defines how to perform the patch operation
+type KubernetesPatchSpec struct {
+	// Type is the type of merge operation to perform
+	Type MergeType `json:"type"`
```
arkodg, line 356: should this field be optional ? and should we use a default similar to what `kubectl` has ?

api/v1alpha1/kubernetes_helpers.go
arkodg: I would also prefer if we moved this out of the `api` directory into https://github.com/envoyproxy/gateway/tree/main/internal/kubernetes

arkodg:
as a follow up, we need to make sure that if a patch fails, we need to surface that to the `EnvoyProxy`'s GatewayClass status

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -265,6 +265,13 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		deployment.Spec.Replicas = nil
 	}
 
+	// apply merge patch to deployment
+	if merged, err := deploymentConfig.ApplyMergePatch(deployment); err == nil {
+		deployment = merged
+	} else {
+		return nil, err
```
arkodg, line 272: if there is a bad patch, we shouldnt fail the creation of deployment, imo we should not patch the deployment and populate the error in the GatewayClass status. Updating the status can be a followup

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -205,6 +205,13 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		}
 	}
 
+	// apply merge patch to deployment
+	if merged, err := r.rateLimitDeployment.ApplyMergePatch(deployment); err == nil {
```
arkodg, line 209: same comment as above

api/v1alpha1/validation/envoyproxy_validate.go
```diff
@@ -81,6 +85,21 @@ func validateProvider(spec *egv1a1.EnvoyProxySpec) []error {
 	return errs
 }
 
+func validateDeployment(spec *egv1a1.EnvoyProxySpec) []error {
+	var errs []error
+	if spec.Provider.Kubernetes != nil && spec.Provider.Kubernetes.EnvoyDeployment != nil {
```
arkodg, line 90: nit as follow up: can this be tackled via CEL ?

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
Xunzhuo: Good point, updated.

internal/infrastructure/kubernetes/proxy/resource_provider.go
Xunzhuo: Good point, updated.

api/v1alpha1/validation/envoyproxy_validate.go
Xunzhuo: Sure.

api/v1alpha1/shared_types.go
```diff
@@ -370,3 +376,25 @@ type KubernetesHorizontalPodAutoscalerSpec struct {
...
+	JSONMerge MergeType = "JSONMerge"
+)
+
+// KubernetesPatchSpec defines how to perform the patch operation
+type KubernetesPatchSpec struct {
+	// Type is the type of merge operation to perform
+	//
+	// By default, StrategicMerge is used as the patch type.
+	// +optional
+	Type MergeType `json:"type,omitempty"`
```
arkodg, line 396: should be a pointer since its optional

arkodg:
small nit, else LGTM !
lets raise a follow up to reuse the same structure for Service as well

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -265,6 +265,11 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		deployment.Spec.Replicas = nil
 	}
 
+	// apply merge patch to deployment
+	if merged, err := deploymentConfig.ApplyMergePatch(deployment); err == nil {
```
arkodg, line 269: lets log the error

arkodg (dismissed):
LGTM thanks !
lets raise some follow up GH issues that can hopefully be tackled pre GA
* to populate patch errors in GatewayClass status
* to support patching for envoy.service as well

arkodg (dismissed):

shawnh2 (approved):
LGTM, thanks for bringing this feature in

arkodg (approved):

PR #2379, opened 2023-12-29
chore: Enable OpenSSF Scorecard

### LINKED ISSUE #2347: Enable OpenSSF Scorecard
*Description*:
would be used.

Monitor security best practices using https://github.com/ossf/scorecard

[optional *Relevant Links*:]

mmorel-35:
**What type of PR is this?**
* chore: Enable OpenSSF Scorecard

* seup scorecard workflow ans pin actions versions

Fixes #2347

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,7 +9,7 @@ jobs:
   release:
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
 
       - name: Extract Release Tag and Commit SHA
         id: vars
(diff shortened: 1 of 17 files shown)


mmorel-35:
cc @phlax

.github/workflows/release.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
           echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
 
       - name: Login to DockerHub
-        uses: docker/login-action@v3
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d  # v3.0.0
```
zirain, line 22: why changes all these? will depbot update this later?

phlax: its more secure to reference them by has rather than by version - but we still need the version for dependabot so its added as a comment

arkodg (approved):
LGTM thanks !
looks like the same diff has been working well in Envoy Proxy
https://github.com/envoyproxy/envoy/blob/main/.github/workflows/scorecard.yml

zirain (approved):

zirain:
@mmorel-35 need more work?
image

mmorel-35:
Just needed to be executed first on the targeted repository. I see the score now .

PR #2384, opened 2023-12-31
api: Add disableMergeSlash and escapedSlashesAction to ClientTrafficPolicy

liorokman:

This PR adds settings into `ClientTrafficPolicy` to control slashes in the incoming request path URI.

The following settings are added:

```yaml
spec:
  path:
     escapedSlashesAction: UnescapeRedirect # or UnescapeForward, RejectRequest, KeepUnchanged
     disableMergeSlashes: false # set to true to prevent Envoy from merging slashes 
```

The proposed defaults are the current EnvoyGateway behavior - so not setting these parameters will not cause any change to the current behavior.

These settings map to the EnvoyProxy [HTTP Connection Manager](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager) configuration `path_with_escaped_slashes_action` and `merge_slashes` keys.

These parameters were proposed in issue #1846 .


arkodg:
-1 on this setting, since its against the recommendations in https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy
what is the use case here @liorokman

liorokman:
> -1 on this setting, since its against the recommendations in https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy what is the use case here @liorokman

The `mergeSlashes` setting recommended by the best practices is actually not part of the HTTP spec. See the comment on the setting's [documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-merge-slashes). There are instances of frameworks that require that proxies in the middle don't merge adjacent slashes - e.g. [Passenger](https://serverfault.com/questions/284475/keep-double-slashes-when-working-with-passenger).

With regards to the  `escapedSlashesAction` setting, the recommendations you cited actually recommend setting the value of this key based on your specific use-case:

> Additionally the [path_with_escaped_slashes_action setting](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-path-with-escaped-slashes-action) should be set according to following recommendations:
>
> REJECT_REQUEST if dowstream clients are expected to use [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) compliant normalized paths (i.e. gRPC clients).
>
> UNESCAPE_AND_REDIRECT if downstream client supports HTTP redirect (i.e. a browser). This option minimizes possibility of path confusion by forcing request to be re-issued with the same path across all parties: downstream client, Envoy and upstream server. Note that gRPC requests will still be rejected with the INTERNAL (13) error code, as gRPC clients do not support redirect.
> 
> KEEP_UNCHANGED for servers that are not [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) compliant and require encoded slashes.
>
> UNESCAPE_AND_FORWARD for servers that are known to treat escaped and unescaped slashes equivalently. Choosing this option may increase probablity of path confusion vulnerabilities if intermediaries perform path based access control.

In order to do so, it needs to be exposed in `ClientTrafficPolicy`.

In our use-case, we have some upstream servers that are not RFC 3986 compliant and we need to keep the encoded slashes as they are.


I think the current defaults are good, and I would definitely keep the current settings as defaults. But I think it does make sense to make these flags available for cases where the defaults don't work.

liorokman:
A few other OSS ingress controllers also allow controlling the `mergeSlashes` key:

- [Ambassador](https://www.getambassador.io/docs/edge-stack/latest/topics/running/ambassador#merge-slashes)
- [Contour](https://projectcontour.io/docs/1.23/config/api/#projectcontour.io/v1alpha1.EnvoyListenerConfig) (see "disableMergeSlashes" )
- [Gloo](https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/hcm/hcm.proto.sk/) (see "mergeSlashes")

arkodg:
thanks for added context @liorokman, some thoughts
* since the above path field are set at within the `hcm`, we are limited to defining it in the `ClientTrafficPolicy` which might not be the ideal home, `SecurityPolicy` could have been another place to keep it, but can't how we can toggle this setting on/off at the route level
* prefer if we have a top level field like `path` to hold all path specific setting
```
path:
  disableEergeSlashes:
```  
* prefer if `escapedSlashesAction` is represented as a `bool`instead of `enum` e.g. `allowEscapedSlashes`

liorokman:
> * prefer if we have a top level field like `path` to hold all path specific setting
> 
> ```
> path:
>   disableEergeSlashes:
> ```

These parameters are already grouped under a top level `path` field.
 
> * prefer if `escapedSlashesAction` is represented as a `bool`instead of `enum` e.g. `allowEscapedSlashes`

Changing this to a `bool` would mean that EG supports only two of the possible actions. It would only be possible to either `KEEP_UNCHANGED` or `UNESCAPE_AND_REDIRECT`. The documentation would have to explicitly say that if `allowEscapedSlashes` is set to `false`, the behavior is `UNESCAPE_AND_REDIRECT` and not any of the other options.

I think there's also a hidden issue here that when the route is a gRPC route the better value for this configuration key would probably be `REJECT_REQUEST`, because gRPC clients are expected to be RFC 3986 compliant to begin with.

api/v1alpha1/pathsettings_types.go
```diff
@@ -0,0 +1,54 @@
...
+	//
+	// Redirect occurs after path normalization and merge slashes transformations if
+	// they were configured. gRPC requests will be rejected with the INTERNAL (13)
+	// error code.
+	// This option minimizes possibility of path confusion exploits by forcing request
+	// with unescaped slashes to traverse all parties: downstream client, intermediate
+	// proxies, Envoy and upstream server.
+	// The “httpN.downstream_rq_redirected_with_normalized_path” counter is incremented
+	// for each redirected request.
+	UnescapeRedirect PathEscapedSlashAction = "UnescapeRedirect"
```
arkodg, line 33: prefer `UnescapeAndRedirect`

api/v1alpha1/pathsettings_types.go
```diff
@@ -0,0 +1,54 @@
...
+	// UnescapeForward unescapes %2F and %5C sequences and forwards the request.
+	// Note: this option should not be enabled if intermediaries perform path based access
+	// control as it may lead to path confusion vulnerabilities.
+	UnescapeForward PathEscapedSlashAction = "UnescapeForward"
```
arkodg, line 37: prefer `UnescapeAndForward`

arkodg:
thanks for the detailed comments @liorokman 
only have suggestions reg naming, else LGTM

reg default setting for `escapedSlashesAction` for gRPC, based on the doc string for `UNESCAPE_AND_REDIRECT`. - `Note that gRPC requests will still be rejected with the INTERNAL (13) error code, as gRPC clients do not support redirect.`
feels like the outcome is the same  as `REJECT_REQUEST` - error code `13` :)

api/v1alpha1/pathsettings_types.go
liorokman: Renamed

api/v1alpha1/pathsettings_types.go
liorokman: Renamed

arkodg (approved):
LGTM thanks for adding the extra details in the PR !

Xunzhuo (approved):

PR #2394, opened 2024-01-03
move "/" to route prefix

zhaohuabing:
This is a small change to move "/" to route prefix so we can reuse the `irRoutePrefix` function and avoid duplicating code.


arkodg (approved):
LGTM thanks !

tanujd11 (approved):

Issue #2377, opened 2023-12-29
Critical Bug: data plane will auto restart when restart HTTPRoute's pod


zzjin:
*Description*:
the envoy gateway data plane pod will restart when restart any pod that HTTPRoute spec-ed.

![output](https://github.com/envoyproxy/gateway/assets/679757/8b954a3f-27b3-4296-b33e-6657fb0f8d02)

*Repro steps*:
apply below crds:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
    namespace: envoy-gateway-system

---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
  namespace: envoy-gateway-system
spec:
  mergeGateways: true

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: envoy-gateway-config
  namespace: envoy-gateway-system
data:
  envoy-gateway.yaml: |
    apiVersion: gateway.envoyproxy.io/v1alpha1
    kind: EnvoyGateway
    provider:
      type: Kubernetes
    gateway:
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
    extensionApis:
      enableEnvoyPatchPolicy: true

---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
  namespace: sealos-system
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
    - name: https
      protocol: HTTPS
      port: 443
      allowedRoutes:
        namespaces:
          from: All
      tls:
        mode: Terminate
        certificateRefs:
          - kind: Secret
            group: ""
            name: wildcard-cert

---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: test-policy
  namespace: sealos-system
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: GatewayClass
    name: eg
    namespace: sealos-system
  type: JSONPatch
  jsonPatches:
  - type: "type.googleapis.com/envoy.config.listener.v3.Listener"
[...]

PR #2395, opened 2024-01-03
fix: implement comparable interface for ir.Infra to avoid multiple reconciling

cnvergence:
**What type of PR is this?**

With single envoy deployment for every listener, containers ports slice were causing it to restart every reconciliation.
This can be fixed by implementing similar interface used in ir.Xds for ir.Infra.
See also: https://github.com/envoyproxy/gateway/issues/1634

Fixes https://github.com/envoyproxy/gateway/issues/2377

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -12,6 +12,10 @@ on:
     - "release/v*"
     paths-ignore:
     - "**/*.png"
+
+permissions:
+  contents: read
+
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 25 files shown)


arkodg:
@cnvergence we shouldnt be sorting, since it leads to unstable outputs, instead we should we using a sorted output as a way to check equality, check out https://github.com/envoyproxy/gateway/pull/1795, we probably need something similar to infraIR

cnvergence:
great idea, let's do it instead

arkodg (approved):
LGTM thanks !

PR #2398, opened 2024-01-03
chore: Add top level permissions for GHA workflows

arkodg:


zirain (approved):

.github/workflows/cherrypick.yaml
```diff
@@ -5,6 +5,9 @@ on:
       - main
     types: ["closed"]
 
+permissions:
```
zirain, line 8: with `read`, it's able to create PR from action?

arkodg: tbh, im not sure if this action was ever working, merging this one to rm the codeql alerts, and will test this out

PR #2400, opened 2024-01-04
rollback retest action

zirain:
cc @arkodg 

@tmsnan PTAL, you're the original author.


arkodg (approved):
LGTM, can you raise a follow up so we can re enable in the future

cnvergence (approved):

zirain:
> LGTM, can you raise a follow up so we can re enable in the future

we can reuse https://github.com/envoyproxy/gateway/issues/2339

PR #2405, opened 2024-01-05
feat: support use listener port as containerPort

Xunzhuo:
**What type of PR is this?**

support use listener port as containerPort, allow users to difine container port via listener port, for direct access scenarios.


Xunzhuo:
close for duplicated with https://github.com/envoyproxy/gateway/pull/3333

Issue #2326, opened 2023-12-20
 SecurityPolicy -> HTTPRoute  Auth mismatch / broken Envoy config


apjoseph:
*Description*:
I've continued testing my very basic ouath2 app +jwt api SecurityFilter use case and have encountered what seems to be substantially missing/broken functionality  when dealing with different security filters on different routes. on the same or different backendRef services. 

I have two deployments 

- a react app using host `dev.example.com` where authentication should be handled by ouath2
- and a node app using host `api.dev.example.com` which should be secured by JWTs 

I discovered that it is not currently not to call the API from the frontend at `api.dev.example.com` with Envoy Gateway because Envoy itself currently does not allow setting the domain of a cookie. I would need domain=dev.example.com in order for CORS to allow cookies for `dev.example.com` to be passed to `api.dev.example.com`

To work around this limitation, I decided to try send requests to `dev.example.com/api ` to my API service and use an HTTPRoute to redirect anything at /api to the node.js backend.

From my understanding of the Gateway, I used the following route/ security configuration

### Routes
- /api

### Upstream Service
Backend Node API

### Configuration
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: {{.Values.web.name}}-web-api
  labels:
    {{- include " example.labels" . | nindent 4 }}
    {{- include " example.web.labels" . | nindent 4 }}
  annotations:
    external-dns.alpha.kubernetes.io/hostname: {{ include " example.baseDomainName" . }}
spec:
  parentRefs:
    - name: {{.Values.web.name}}-web
      sectionName: {{.Values.web.name}}-web
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: {{.Values.web.name}}-web-api
          namespace: {{.Release.Namespace}}
          port: 8080
      matches:
        - path:
            type: PathPrefix
            value: /api


---

apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  name: {{.Values.web.name}}-web-api
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: {{.Values.web.name}}-web-api

  jwt:
    providers:
      - name: auth0
        extractFrom:
          cookies:
            - BearerToken
        remoteJWKS:
          uri: https://{{include " example.auth.domain" .}}/.well-known/jwks.json
```

## OAuth2
###Routes
[...]

PR #2409, opened 2024-01-05
add redirectURL and signoutPath to OIDC

zhaohuabing:
This PR allows user to specify redirect URL, signout Path, and passthrough path in the  OIDC settings of the `SecurityPolicy` API.

For example, if there're two applications "bookstore" and "petstore", and we would like to define a `SecurityPolicy` for each one to configure its OIDC aunthentication. For each application, there're some public pages that don't need authentication.

HTTPRoutes for "bookstore" and "petstore":
```yaml
kind: HTTPRoute
metadata:
  name: bookstore
spec:
  hostnames:
  - www.example.com
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /bookstore
    backendRefs:
    - group: ""
      kind: Service
      name: bookstore
      port: 3000
  ```    
  
 
  ```yaml
  kind: HTTPRoute
metadata:
  name: petstore
spec:
  hostnames:
  - www.example.com
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /petstore
    backendRefs:
    - group: ""
      kind: Service
      name: petstore
      port: 3000
  ```
  
  SecurityPolicies for "bookstore" and "petstore"
  ```yaml
kind: SecurityPolicy
metadata:
  name: bookstore
spec:
  oidc:
    # make sure the redirect_uri can be caught by HTTPRoute bookstore
    redirect_uri: /bookstore/oauth2/callback 
     # exclude public pages
    pass_through: 
    - path:
        type: PathPrefix
        value: /petstore/pulic-page1
    - path:
        type: PathPrefix
        value: /petstore/pulic-page2    
    clientID: bookstore_client_id
    clientSecret:
      group: ""
      kind: Secret
      name: bookstore-client-secret
    provider:
      issuer: https://oauth2.myoidcprovider.com
  targetRef:
    kind: HTTPRoute
    name: bookstore
```

```yaml
kind: SecurityPolicy
metadata:
  name: petstore
spec:
  oidc:
    # make sure the redirect_uri can be caught by HTTPRoute petstore
    redirect_uri: /petstore/oauth2/callback 
    # exclude public pages
    pass_through:   
    - path:
        type: PathPrefix
        value: /petstore/pulic-page1
    - path:
        type: PathPrefix
        value: /petstore/pulic-page2    
    clientID: petstore_client_id
    clientSecret:
      group: ""
      kind: Secret
      name: petstore-client-secret
    provider:
      issuer: https://oauth2.myoidcprovider.com
  targetRef:
    kind: HTTPRoute
    name: petstore
```    

Related: https://github.com/envoyproxy/gateway/issues/2326


api/v1alpha1/oidc_types.go
```diff
@@ -36,20 +36,21 @@ type OIDC struct {
...
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	// If not specified, uses the default redirect URI "%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback"
+	RedirectURL *string `json:"redirectURL,omitempty"`
+
+	// The path to sign a user out, clearing their credential cookies.
+	// If not specified, uses a default signout path "/signout"
+	SignoutPath *string `json:"signoutPath,omitempty"`
+
+	// The paths that don't need authentication
+	PassThrough []string `json:"passThrough,omitempty"`
```
zhaohuabing, line 50: We need to agree on the API first before moving on.

@arkodg @apjoseph

api/v1alpha1/oidc_types.go
```diff
@@ -36,20 +36,21 @@ type OIDC struct {
 	// specified.
 	// +optional
 	Scopes []string `json:"scopes,omitempty"`
+
+	// The redirect URL to be used in the OIDC
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	// If not specified, uses the default redirect URI "%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback"
+	RedirectURL *string `json:"redirectURL,omitempty"`
```
arkodg, line 43: this should be `redirect_uri`
Also just noticed,  below fields should be `authorization_uri` and `token_uri`

api/v1alpha1/oidc_types.go
```diff
@@ -36,20 +36,21 @@ type OIDC struct {
...
+
+	// The path to sign a user out, clearing their credential cookies.
+	// If not specified, uses a default signout path "/signout"
+	SignoutPath *string `json:"signoutPath,omitempty"`
```
arkodg, line 47: this should be `logout`

should this be `logoutPath` or `logoutUri` ?

api/v1alpha1/oidc_types.go
arkodg: im a -1 on this, this API field is not needed imo, users can create 2 HTTPRoutes - 1 that is targeted by a `BackendTrafficPolicy` and one that is not to satisfy this use case

api/v1alpha1/oidc_types.go
zhaohuabing: > Also just noticed, below fields should be authorization_uri and token_uri

This might be a bit confusing with the `Endpoint` of a Cluster in Envoy, but I'm using the terminology from the OIDC specification, and Envoy oauth2 filter also uses the same terms.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2

https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint
```
3.1.2.  Authorization Endpoint
The Authorization Endpoint performs Authentication of the End-User. This is done by sending the User Agent to the Authorization Server's Authorization Endpoint for Authentication and Authorization, using request parameters defined by OAuth 2.0 and additional parameters and parameter values defined by OpenID Connect.

Communication with the Authorization Endpoint MUST utilize TLS. See [Section 16.17](https://openid.net/specs/openid-connect-core-1_0.html#TLSRequirements) for more information on using TLS.

https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint
```
3.1.3.  Token Endpoint
To obtain an Access Token, an ID Token, and optionally a Refresh Token, the RP (Client) sends a Token Request to the Token Endpoint to obtain a Token Response, as described in Section 3.2 of [OAuth 2.0](https://openid.net/specs/openid-connect-core-1_0.html#RFC6749) [RFC6749], when using the Authorization Code Flow.

Communication with the Token Endpoint MUST utilize TLS. See [Section 16.17](https://openid.net/specs/openid-connect-core-1_0.html#TLSRequirements) for more information on using TLS.
```

File: api/v1alpha1/oidc_types.go (hunk 3 above)
Comment (zhaohuabing) [reply api/v1alpha1/oidc_types.go:47]: It should be path because the schem and port are not included. For example `/petstore/logout`

File: api/v1alpha1/oidc_types.go (hunk 2 above)
Comment (zhaohuabing) [reply api/v1alpha1/oidc_types.go:43]: > this should be `redirect_uri`

Some of the APIs are using `URL`, for example, the rate limit server, and some are using `URI`, should they be unified?

Based on a quick search, it appears `URL` is more appropriate here.

```
Should I use URL or URI?
It depends on what you are trying to do. If you are trying to access a resource on the internet, such as a web page or an image, you should use a URL. A URL is a specific type of URI that specifies the location of a resource on the internet and includes a protocol at the beginning.

If you are trying to identify a resource by its name, rather than its location, you should use a URN (Uniform Resource Name). URNs are used to provide a persistent and location-independent way to identify resources.

If you are not sure which type of URI to use, you can use a generic URI. A URI is a generic term that refers to a string of characters that identify a resource. It can be used to identify any type of resource, not just those on the internet.
```

File: api/v1alpha1/oidc_types.go (hunk 1 above)
Comment (zhaohuabing) [reply api/v1alpha1/oidc_types.go]: `Passthrough` has been removed. We can revisit and add it later if needed.

File: api/v1alpha1/oidc_types.go (hunk 3 above)
Comment (arkodg) [reply api/v1alpha1/oidc_types.go:47]: envoy docs calls its signout but the oidc docs calls it `logout`, so prefer `logoutPath`

File: api/v1alpha1/oidc_types.go (hunk 2 above)
Comment (arkodg) [reply api/v1alpha1/oidc_types.go:43]: some more comparisons for naming https://github.com/istio-ecosystem/authservice/blob/master/docs/README.md#message-config-configconfigproto

Comment (arkodg) [reply api/v1alpha1/oidc_types.go:43]: using `url` sgtm !

Comment (arkodg) [reply api/v1alpha1/oidc_types.go:43]: liking rfc in case others would be interested in further reading https://datatracker.ietf.org/doc/html/rfc2396#section-1.2

Comment (zhaohuabing) [reply api/v1alpha1/oidc_types.go:43]: Regarding `authorization_uri/url` and `token_uri/url` vs `authorization_endpoint` and `token_endpoint`, I think the former is more common in the wild, but the latter is more consistent with the OIDC spec. I'm fine with either but slightly prefer the latter because of consistency with the OIDC spec and Envoy's OAuth2 filter.

Comment (arkodg) [reply api/v1alpha1/oidc_types.go:43]: ptal @envoyproxy/gateway-maintainers

Comment (zhaohuabing) [reply api/v1alpha1/oidc_types.go:43]: I think it's a good practice to use the same terms across all EG APIs if there're no existing well-known references to those terms. :-)

But my argument for using the same terms from the OIDC spec is that it's a well-known spec and it's easier for people to understand the terms if they are consistent with the spec. 

- If a user is familiar with OIDC, they might be confused if we use different terms in EG API.
- If a user is not familiar with OIDC, they may look up the OIDC spec to understand the terms, and they will also be confused if we use different terms in EG API.

File: internal/ir/xds.go [hunk 4]
```diff
@@ -433,6 +433,16 @@ type OIDC struct {
 	// The OIDC scopes to be used in the
 	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
 	Scopes []string `json:"scopes,omitempty" yaml:"scopes,omitempty"`
+
+	// The redirect URL to be used in the OIDC
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	RedirectURL string `json:"redirectURL,omitempty"`
+
+	// The path part of the redirect URL
```
arkodg, line 441: can this be removed, and obtained from `RedirectURL` ?

internal/gatewayapi/securitypolicy.go
```diff
@@ -447,19 +453,58 @@ func (t *Translator) buildOIDC(
...
 		ClientID:     oidc.ClientID,
 		ClientSecret: clientSecretBytes,
 		Scopes:       scopes,
+		RedirectURL:  redirectURL,
+		RedirectPath: redirectPath,
+		LogoutPath:   logoutPath,
 	}, nil
 }
 
+func extractRedirectPath(redirectURL string) (string, error) {
```
arkodg, line 488: can we use https://pkg.go.dev/net/url#URL instead ?

internal/ir/xds.go
zhaohuabing: It's obtained from `RedirectURL`, not exposed to API.

internal/gatewayapi/securitypolicy.go
zhaohuabing: We need to handle Envoy header values like `%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback`

arkodg (approved):
LGTM

zirain (approved):

PR #1068, opened 2023-02-21
support envoyproxy custom configurations.

qicz:
support #1038 

- adjust xds-cluster connection timeout to 10s
- support custom envoyproxy configurations, cluster: connectionTimeout & upstreamTLS

PR #2411, opened 2024-01-05
api: Support connection timeouts in BackendTrafficPolicy

### LINKED ISSUE #2401: Support upstream HTTP connection timeouts
*Description*:
Envoy supports a variety of upstream [connection timeouts](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts) that impact persistent HTTP connections establishment and lifecycle:
- [Cluster connect timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-connect-timeout): timeout for establishing a TCP connection. If TLS is used, the timeout also applies to TLS session setup. The Envoy default is 5s. The Envoy Gateway default is 10s. 
- [HTTP proxy idle timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-idle-timeout): Idle connections (without active requests) are closed after this time. Envoy default is 60m. 
- [HTTP proxy max connection duration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-max-connection-duration): When the max connection duration is reached, HTTP persistent connections are closed (irrespective of the connection's idleness). The default value is unlimited. 

Envoy Gateway should support configuration of these timeouts for upstream connections. Below are some examples and use cases: 
- Cluster connect timeout: 
  - Envoy Gateway default values prioritize security. However, other proxies like nginx use 60s as a [default](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout) for connect timeout, indicating that longer connection establishment times are possible.  
  - Users may need to increase this value. For example, when the latency to the backend is significant, the TCP/TLS handshake may not finish in time. 
- HTTP Idle timeout: 
  - Users may want to change idle timeout settings to better reflect their performance vs. resource consumption needs. 
  - It could be necessary to align the Envoy idle timeout with the backend idle timeout, if the backend does not send the `Connection: close` header before closing the connection. Otherwise, a race condition can occur where envoy will use a half-closed connection and fail with a reset. For example, see here: https://github.com/envoyproxy/envoy/issues/14981. 
- HTTP max connection duration: 
  - While the default value is not disruptive, Envoy docs [mention](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#connection-timeouts) that setting a maximum age is beneficial for cleanup of connections to targets that are no longer discoverable with DNS.

guydc:
Allow users to configure various upstream timeout options. 

Fixes #2401, relates to #1821


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -75,6 +75,11 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	CircuitBreaker *CircuitBreaker `json:"circuitBreaker,omitempty"`
+
+	// Timeout settings for the backend connections.
+	//
+	// +optional
+	ConnectionTimeouts *ConnectionTimeouts `json:"timeouts,omitempty"`
```
arkodg, line 82: prefer `timeout` here, we've used singular terms in all fields in this API. 
Also `connection` prefix is not needed imo

api/v1alpha1/timeout_types.go
```diff
@@ -0,0 +1,26 @@
...
+// the root of the repo.
+
+package v1alpha1
+
+import gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+// ConnectionTimeouts defines configuration for timeouts related to connections.
+type ConnectionTimeouts struct {
+	// The timeout for new network connection establishment, including TCP and TLS handshakes.
+	//
```
arkodg, line 13: should we create top level `tcp` and `http` fields here ?

api/v1alpha1/timeout_types.go
```diff
@@ -0,0 +1,26 @@
...
+	// +optional
```
arkodg, line 14: defaults should be mentioned here

api/v1alpha1/timeout_types.go
guydc: Seeing as there are several other HTTP (stream, ...) and TCP (proxy, protocol, ...) timeouts that could be relevant in the future, it makes sense to me.

api/v1alpha1/timeout_types.go
guydc: Are you referring to Envoy defaults, Envoy Gateway defaults or both? 

`connectTimeout` was customized by Envoy Gateway in #1068 to 10s. 
Do you think that we should:
* Make that more explicit and use a kubebuilder `default` marker for that here? 
* Remove the Envoy Gateway custom value and keep the Envoy default, now that this option is configurable for end users? 
* Just document the Envoy Gateway default and keep it as a translation layer constant? 

`idleTimeout` has an envoy default of 1hr. Should we just document that, or actually reflect that with a kubebuilder marker?

test/cel-validation/backendtrafficpolicy_test.go
```diff
@@ -11,6 +11,7 @@ package celvalidation
 import (
 	"context"
 	"fmt"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
```
shawnh2, line 14: sort imports please

zirain: why this pass ci? let me double check it.

api/v1alpha1/timeout_types.go
arkodg: comment should be enough here imo, & lets reflect EG defaults

api/v1alpha1/timeout_types.go
```diff
@@ -0,0 +1,43 @@
...
+	ConnectTimeout *gwapiv1.Duration `json:"connectTimeout,omitempty"`
+}
+
+type HTTPTimeout struct {
+	// The idle timeout for persistent HTTPTimeout connections. Idle time is defined as a period in which there are no active requests in the connection.
+	// Default: 1 hour.
+	//
+	// +optional
+	ConnectionIdleTimeout *gwapiv1.Duration `json:"connectionIdleTimeout,omitempty"`
```
arkodg, line 36: is the `connection` prefix needed here ?

guydc: Envoy also has `stream` timeouts, e.g. [stream idle timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-stream-idle-timeout). So, to distnguish between the two, I added the connection prefix.

arkodg: sgtm, thanks

api/v1alpha1/timeout_types.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// The maximum duration of an persistent HTTPTimeout connections.
+	// Default: unlimited.
+	//
+	// +optional
+	ConnectionDuration *gwapiv1.Duration `json:"connectionDuration,omitempty"`
```
arkodg, line 42: maxConnectionDuration instead ?

api/v1alpha1/timeout_types.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+type HTTPTimeout struct {
+	// The idle timeout for persistent HTTPTimeout connections. Idle time is defined as a period in which there are no active requests in the connection.
+	// Default: 1 hour.
+	//
+	// +optional
+	ConnectionIdleTimeout *gwapiv1.Duration `json:"connectionIdleTimeout,omitempty"`
+
+	// The maximum duration of an persistent HTTPTimeout connections.
```
arkodg, line 38: ```The maximum duration of a HTTP connection.```

arkodg (approved):
LGTM thanks !

tmsnan (approved):
LGTM

shawnh2 (approved):
lgtm, thanks

zirain (approved):

PR #2412, opened 2024-01-05
fix: tools management

mmorel-35:
**What type of PR is this?**
* fix OSSF Score 

Today the tools are managed by a dedicated go.mod file .
The problem is that scorecard consider tools as dependencies of the project and each vulnerability of those tools decrease the score of the project.

This store tool and its version in a txt file to be then `go install`ed in `tools/bin`.
It also adds an actions/cache so it is cached to not download it to much time.

Fixes most of the dependabot vulnerabilities


mmorel-35:
@phlax , what do you think ?

tools/src/buf.txt
```diff
@@ -0,0 +1 @@
+github.com/bufbuild/buf/cmd/buf@v1.28.1
```
zirain, line 1: maybe we can remove these file, define thing directly in makefile, wdyt @arkodg ?

arkodg:
@mmorel-35 thanks for working on it, and eliminating the false positives to bump up the scorecard score
can you try out the approach suggested by @zirain to collapse it all under one file - `tools.mk` ?

mmorel-35:
Hi ! I’ve merged install of the tools in tools.mk.
I don’t kown if there is a more elegant way to do this. I’m interested in any way to make this more simple.

Dependabot is not able to handle go modules outside of a go.mod file . I don’t know how it would be possible to use a bot to update those dependencies.
@arkodg , @zirain , @phlax ,

Any suggestion ?

arkodg:
hey @mmorel-35 if dependabot cannot handle updates outside `go.mod`, imo we should stick with the orig `go.mod` approach, there's a big benefit of auto updating these tools that are used in the CI workflow

maybe we need to manually bump those tool dependencies for now

zirain (approved):
this's fine to me, a stable version of tools always lead a stable result, we can upgrade these manually if needed.

zirain:
kindly ping @arkodg

arkodg:
ah yah, please give me a day or two to try it out manually, else we can merge this PR

arkodg:
looks like manually bumping the packages helped, this is no longer needed, thanks @mmorel-35 for trying this out

PR #2413, opened 2024-01-06
feat: Implement disableMergeSlashes and escapedSlashesAction

liorokman:
This PR implements the APIs added in #2384


internal/ir/xds.go
```diff
@@ -207,6 +207,8 @@ type HTTPListener struct {
 	// HTTP3 provides HTTP/3 configuration on the listener.
 	// +optional
 	HTTP3 *HTTP3Settings `json:"http3,omitempty"`
+	// Path contains settings for path URI manipulations
+	Path PathSettings `json:"path,omitempty"`
```
shawnh2, line 211: ptr type

internal/ir/xds.go
```diff
@@ -262,6 +264,22 @@ func (t TLSListenerConfig) Validate() error {
 	return errs
 }
 
+type PathEscapedSlashAction egv1a1.PathEscapedSlashAction
```
shawnh2, line 267: why not using `egv1a1.PathEscapedSlashAction` directly?

liorokman: Because if I use it directly, I would need to import `egv1a1` into the XDS layer in order to translate the API constants into XDS constants.

internal/ir/xds.go
liorokman: It doesn't make sense to have a pointer here. This is not an optional structure.

shawnh2: oops, my bad, not noticing

internal/gatewayapi/listener.go
```diff
@@ -104,6 +104,10 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 					Address: "0.0.0.0",
 					Port:    uint32(containerPort),
 					TLS:     irTLSConfigs(listener.tlsSecrets),
+					Path: ir.PathSettings{
```
arkodg, line 107: maintaining logic in two places in the gatewayapi layer may be harder to reason with for the next dev trying to figure out translation, thoughts ?
instead if `ir.path` is `nil`, the xds translator layer could set defaults

liorokman: > maintaining logic in two places in the gatewayapi layer may be harder to reason with for the next dev trying to figure out translation, thoughts ?

I think the IR should encapsulate what is expected of the XDS layer, and all defaults should be set in the IR layer. There should be no magic translations happening in the XDS layer - XDS translation should be a technical translation without filling in any controlled attribute.

> instead if ir.path is nil, the xds translator layer could set defaults

By that logic, should I also remove the `Address` default of `0.0.0.0` in this structure (multiple places in the `gatewayapi` folder) and the default path being set in [processAccessLog](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/listener.go#L168)?

arkodg (dismissed):

arkodg (approved):

shawnh2 (approved):

PR #2432, opened 2024-01-11
Feat: Support configuring externalTrafficPolicy on the envoy service

timricese:
**What type of PR is this?**
Adds possibility to configure externalTrafficPolicy on EnvoyProxy Service objects.

Currently there is no way to change the default "Local" externalTrafficPolicy, which is fine for the vast majority of use cases.

In our case, we run bare-metal without any cloud loadbalancer and rely on MetalLB and Cilium with "Cluster" externalTrafficPolicy to provide loadbalancing within the cluster. Without this, all requests arrive at one single envoy pod.


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2433, opened 2024-01-11
api: Enable controlling the HTTP version used to connect to the backend

liorokman:
This PR proposes adding an API to control the HTTP version that will be used when connecting to a backend configured via an `HTTPRoute`.

```yaml
kind: BackendTrafficPolicy
spec:
      useClientProtocol: true
```

The current behavior is that HTTP/2 is explicitly used if the route is a GRPC route, and it's possible to force HTTP/2-over-cleartext (`h2c`) by setting the `appProtocol` field in the backend Kubernetes Service to `kubernetes.io/h2c`. In all other cases EnvoyGateway generates configuration that forces HTTP/1.1 to the upstream.

The proposed API would allow configuring Envoy Proxy to use the same HTTP protocol version when connecting to the backend as was used by the downstream client.


arkodg:
hey @liorokman can you raise a GH issue to define the use case ? would like to get input from the community to be understand if this is needed or not

liorokman:
> hey @liorokman can you raise a GH issue to define the use case ? would like to get input from the community to be understand if this is needed or not

Opened #2437

guydc:
LGTM. Maybe we need to update the description with an up-to-date example of the API.

guydc (dismissed):

liorokman:
> Maybe we need to update the description with an up-to-date example of the API.

Done

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -98,6 +103,15 @@ type BackendTrafficPolicySpec struct {
 	Compression []*Compression `json:"compression,omitempty"`
 }
 
+type HTTPOptions struct {
+	// UseClientProtocol configures Envoy to prefer sending requests to TLS enabled backends using
```
arkodg, line 107: can you elaborate on why  preference this given to TLS enabled backend ?

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -98,6 +103,15 @@ type BackendTrafficPolicySpec struct {
 	Compression []*Compression `json:"compression,omitempty"`
 }
 
+type HTTPOptions struct {
```
arkodg, line 106: there are already some HTTP only fields at the top level like `FaultInjection`, so vote to keep this at a top level
will bring this up in the next envoy gateway community meeting

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -98,6 +103,15 @@ type BackendTrafficPolicySpec struct {
...
+	// UseClientProtocol configures Envoy to prefer sending requests to TLS enabled backends using
+	// the same HTTP protocol that the incoming request used. Defaults to false, which means
```
arkodg, line 108: prefer if we added a line in here saying something along the lines of - `this is especially useful when the client (connecting to the gateway) and backend support multiple HTTP versions (e.g. http/1.1 and http/2), and users would like to force the backend to use the same version the client is using`

api/v1alpha1/backendtrafficpolicy_types.go
liorokman: The goal is to use Envoy Proxy's [use_downstream_protocol_config](https://www.envoyproxy.io/docs/envoy/v1.29.3/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto#extension-envoy-upstreams-http-http-protocol-options) feature. 

If the underlying transport socket supports TLS, it also supports ALPN. This makes it possible for Envoy to try to use both HTTP/1.1 and HTTP/2 as needed when connecting to it. If TLS is not enabled, then Envoy Proxy will either use HTTP/1.1 or HTTP/2. This means that if the downstream client is using the "wrong" protocol, the connection will simply fail.

It's easier to restrict this feature to TLS enabled upstreams where selection can be done based on ALPN than to try to force the downstream to have the right protocol. Fixing the client protocol selection is hard because:

1. If the Gateway has multiple HTTPRoutes, each one to a different  upstream cluster with different supported protocols, then it's not possible to fix the Listener level ALPN array correctly. The route to be used is based on things that come later (e.g. the request path), so it's possible that one HTTPRoute has an upstream that supports only HTTP/2 and is matched on the `/first` path, while the other HTTPRoute attached to the same Gateway supports only HTTP/1.1 and is matched on the `/second` path. Envoy can't know which path is going to be used before the protocol is selected and the request actually sent to Envoy, and by that point it's too late to change the protocol.

1. If the downstream socket (the Gateway) is not TLS enabled, then there's no ALPN available to affect the client's selection anyways, and there would be a large chance of a misconfiguration that makes some HTTPRoutes unusable.

api/v1alpha1/backendtrafficpolicy_types.go
liorokman: OK. Will wait for the decision before changing this PR.

api/v1alpha1/backendtrafficpolicy_types.go
liorokman: OK. Effectively this means that the backend must be TLS enabled, or it can't support multiple HTTP versions.

arkodg: can you elaborate on how TLS / ALPN is related here ? 
as far as in understand, this knob is proactively deciding the upstream protocol version to use based on downstream connection

liorokman: A backend without TLS will either support HTTP/1.1 or HTTP/2, but never both.  That's the point of `h2c` - it requires pre-knowledge on the part of the client that the server requires HTTP/2.

Since HTTP/2 is a binary protocol, if a client like Envoy Proxy starts a connection by sending a textual HTTP/1.1 request, the server will disconnect it. 
Since HTTP/1.1 is a textual protocol, if a client like Envoy Proxy starts a connection by sending an HTTP/2 request, the server will disconnect it.
Without TLS + ALPN, there is no way for a client like Envoy Proxy to negotiate with the server which protocol it can use, and this means that effectively upstream can't support _multiple_ versions of HTTP without TLS+ALPN.

> (like h2c to upstream, if downstream is h2c or h2)

We can configure the Envoy Proxy client to try to connect to upstream with the downstream protocol even without TLS on the upstream protocol. What it would do is automatically fail all requests where the downstream protocol doesn't match the single upstream protocol that is possible without upstream TLS.

liorokman:
@guydc @arkodg PTAL

arkodg (approved):
LGTM thanks !

guydc (approved):

Issue #1059, opened 2023-02-17
`ext_authz`  Authentication


Xervon:
*Description*:
Are there any plans of supporting `ext_authz` request authentication?

arkodg:
@Xervon the project supports a native [Authn API](https://gateway.envoyproxy.io/v0.3.0/user/authn.html) and plans on allowing advanced users to configure envoy filters such as `ext_authz` as well, https://github.com/envoyproxy/gateway/issues/24 is the tracking issue that should address this use case, which should hopefully be part of the 0.4.0 release in April 2023

Xervon:
Thanks for the quick answer. I guess I should have clarified that I had already seen both of those. I was wondering if there were already plans to extend the `AuthenticationFilter` with support for `ext_authz` (similar to #881 as far as I understand that one). And if there are no plans regarding this, what would have to be done if I wanted to take a stab at this?

arkodg:
@Xervon imho [ext_authz](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto) is something an advanced user would be aware of, so would prefer taking the #24 route, understanding the shortcomings once implemented and used, and if it still required, expose the fields of the underlying envoy filter into the higher level authn API

arkodg:
thanks for sharing all the references for other projects @Xervon , the [original design doc](https://docs.google.com/document/d/1TlQjBy1utEwgrxE_HVT4-EHpVJ51hgnfMuAh0Q_uNoE/edit#heading=h.h6z8hgjmkv0g) did account for `ext_authz`, hoping other @envoyproxy/gateway-maintainers can chime in

arkodg:
speaking to many users at Kubecon, who would like to integrate with an external authz system like OPA 
this can now be added into https://gateway.envoyproxy.io/latest/api/extension_types/#securitypolicy

akhenakh:
Agree it's a very common need to point to an ext_authz_filter filter to talk to an authorization server, resorting to patching xDS for a basic need is not optimal

PR #2435, opened 2024-01-12
api: support for external authz

zhaohuabing:
External authorization API.

Related: https://github.com/envoyproxy/gateway/issues/1059


api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// Host ist the hostname of the gRPC External Authorization service
+	Host gwapiv1a2.Hostname `json:"host,omitempty" yaml:"host,omitempty"`
+
+	// Port is the network port of the gRPC External Authorization service
+	Port gwapiv1a2.PortNumber `json:"port"`
+
+	// TLS defines the TLS configuration for the gRPC External Authorization service.
+	// Note: If not specified, the proxy will talk to the gRPC External
+	// Authorization service in plaintext.
+	TLS *TLSConfig `json:"tlsSettings,omitempty" yaml:"tlsSettings"`
```
zhaohuabing, line 67: The existence of TLS indicates that the Envoy should talk to the GRPC service with tls. The content of TLSConfig can be empty if the system default certificate pool is used for verifying server certificates.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+type ExtAuth struct {
+	// Type decides the type of External Authorization.
+	// Valid ExtAuthServiceType values are "GRPC" or "HTTP".
+	//
+	// +unionDiscriminator
+	Type ExtAuthServiceType `json:"type"`
+
+	// GRPCService defines the gRPC External Authorization service
+	// Only one of GRPCService or HTTPService may be specified.
+	GRPCService *GRPCExtAuthService `json:"grpcService,omitempty" yaml:"grpcService"`
```
zirain, line 34: simply with `grpc`?

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// services is different. All headers will be included in the check request
+	// to a gRPC authorization server, whereas no headers will be included in the
+	// check request to an HTTP authorization server.
+	AllowedHeaders []string `json:"allowedHeaders,omitempty" yaml:"allowedHeaders,omitempty"`
+}
+
+// GRPCExtAuthService defines the gRPC External Authorization service
+type GRPCExtAuthService struct {
+	// Host is the hostname of the gRPC External Authorization service
+	Host gwapiv1a2.Hostname `json:"host,omitempty" yaml:"host,omitempty"`
```
zirain, line 59: I recall there's a discussion about things like this, looks like here you introduce an new way?

api/v1alpha1/ext_auth_types.go
zhaohuabing: No strong opinion, but GRPCService seems clearer to me.

api/v1alpha1/ext_auth_types.go
zhaohuabing: Yes, I think using a single URL to represent GRPC/HTTP auth service may not be feasible.

* No ideal way to represent tls for grpc service? grpcs:// is not a standard way.
* HTTP service may have an additional path, e.g. /authz
* Port number should not be optional for grpc service, but it is optional for HTTP service.(default to 80/443)
* Some other configuration options may be different for grpc and HTTP service. For example, HTTP service supports `AllowedUpstreamHeaders`, but grpc service does not.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// HTTPService defines the HTTP External Authorization service
+	// Only one of GRPCService or HTTPService may be specified.
+	HTTPService *HTTPExtAuthService `json:"httpService,omitempty" yaml:"httpService"`
+
+	// AllowWhenFailed indicates whether the request should be allowed when the
+	// authorization service has failed or has returned an HTTP 5xx error.
+	// When set to true, the request will be allowed.
+	// When set to false, the request will be denied with an HTTP 403 Forbidden error.
+	// Defaults to false.
+	AllowWhenFailed bool `json:"allowWhenFailed,omitempty" yaml:"allowWhenFailed"`
```
arkodg, line 45: should be use `FailOpen` here ? also is this field needed for the first iteration of the API

api/v1alpha1/ext_auth_types.go
arkodg: +1 to `grpc` and `http`

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+// HTTPExtAuthService defines the HTTP External Authorization service
+type HTTPExtAuthService struct {
+	// URL is the URL of the HTTP External Authorization service.
+	// The URL must be a fully qualified URL with a scheme, hostname,
+	// and optional port and path. Parameters are not allowed.
+	// The URL must use either the http or https scheme.
+	// If port is not specified, 80 for http and 443 for https are assumed.
+	// If path is specified, the authorization request will be sent to that path,
+	// or else the authorization request will be sent to the root path.
+	URL string `json:"url" yaml:"url"`
```
arkodg, line 79: shouldnt we use `Host` and `Port` here as well ?

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// establish a TLS handshake with the external authorization server.
+	//
+	// If this field is not specified, the proxy will not present a client certificate
+	// and will use the system default certificate pool to verify the server certificate.
+	// +optional
+	// +kubebuilder:validation:MaxItems=64
+	CertificateRefs []gwapiv1a2.SecretObjectReference `json:"certificateRefs,omitempty"`
+}
+
+// AuthorizationResponse defines how the authorization response impacts the client request and response.
```
arkodg, line 106: seems unused

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// Defaults to false.
+	AllowWhenFailed bool `json:"allowWhenFailed,omitempty" yaml:"allowWhenFailed"`
+
+	// AllowedHeaders defines the client request headers that will be included
+	// in the request to the external authorization service.
+	// Note: If not specified, the default behavior of different external authorization
+	// services is different. All headers will be included in the check request
+	// to a gRPC authorization server, whereas no headers will be included in the
+	// check request to an HTTP authorization server.
+	AllowedHeaders []string `json:"allowedHeaders,omitempty" yaml:"allowedHeaders,omitempty"`
```
arkodg, line 53: isn't this specific to HTTP ?

api/v1alpha1/ext_auth_types.go
zhaohuabing: We can remove it from the initial API.

api/v1alpha1/ext_auth_types.go
zhaohuabing: It can be changed to the following struct, more fields, but aligned with gRPC service.

```go
// HTTPExtAuthService defines the HTTP External Authorization service
type HTTPExtAuthService struct {
	// Host is the hostname of the HTTP External Authorization service
	Host gwapiv1a2.Hostname `json:"host" yaml:"host"`

	// Port is the network port of the HTTP External Authorization service
	Port gwapiv1a2.PortNumber `json:"port" yaml:"port"`

	// If path is specified, the authorization request will be sent to that path,
	// or else the authorization request will be sent to the root path.
	Path *string `json:"path,omitempty" yaml:"path,omitempty"`

	// TLS defines the TLS configuration for the gRPC External Authorization service.
	// Note: If not specified, the proxy will talk to the gRPC External
	// Authorization service in plaintext.
	TLS *TLSConfig `json:"tlsSettings,omitempty" yaml:"tlsSettings"`

	// Authorization response headers that will be added to the original client request
	// before sending it to the upstream server.
	// Note that coexistent headers will be overridden.
	AllowedUpstreamHeaders []string `json:"allowedUpstreamHeaders,omitempty"`
}
```

api/v1alpha1/ext_auth_types.go
zhaohuabing: > isn't this specific to HTTP ?

According to Envoy api, yes. https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-httpservice

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,111 @@
...
+	// TLS defines the TLS configuration for the HTTP External Authorization service.
+	// TLS is only valid when the URL scheme is https. If the URL scheme is
+	// https, and TLS is not specified, the proxy will use the system default
+	// certificate pool to verify the server certificate.
+	TLS *TLSConfig `json:"tlsSettings,omitempty" yaml:"tlsSettings"`
+
+	// Authorization response headers that will be added to the original client request
+	// before sending it to the upstream server.
+	// Note that coexistent headers will be overridden.
+	AllowedUpstreamHeaders []string `json:"allowedUpstreamHeaders,omitempty"`
```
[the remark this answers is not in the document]
zhaohuabing: Remove all yaml tags, yaml lib just reuses json tags.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,97 @@
...
+	// TLS defines the TLS configuration for the HTTP External Authorization service.
+	// TLS is only valid when the URL scheme is https. If the URL scheme is
+	// https, and TLS is not specified, the proxy will use the system default
+	// certificate pool to verify the server certificate.
+	TLS *TLSConfig `json:"tlsSettings,omitempty"`
+
+	// Authorization response headers that will be added to the original client request
+	// before sending it to the upstream server.
+	// Note that coexistent headers will be overridden.
+	AllowedUpstreamHeaders []string `json:"allowedUpstreamHeaders,omitempty"`
```
arkodg, line 83: optional tag

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,97 @@
...
+	// The URL must use either the http or https scheme.
+	// If port is not specified, 80 for http and 443 for https are assumed.
+	// If path is specified, the authorization request will be sent to that path,
+	// or else the authorization request will be sent to the root path.
+	URL string `json:"url" yaml:"url"`
+
+	// TLS defines the TLS configuration for the HTTP External Authorization service.
+	// TLS is only valid when the URL scheme is https. If the URL scheme is
+	// https, and TLS is not specified, the proxy will use the system default
+	// certificate pool to verify the server certificate.
```
arkodg, line 77: optional tag

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,97 @@
...
+	TLS *TLSConfig `json:"tlsSettings,omitempty"`
+
+	// Authorization response headers that will be added to the original client request
```
arkodg, line 80: what is the default case when this field is unset ? are all response headers sent to backend ?

api/v1alpha1/ext_auth_types.go
arkodg, line 83: prefer using `Backend` instead of `Upstream` since that term is used everywhere in the upstream and EG API
also thoughts on these field names `HeadersToBackend` and `HeadersToExtAuth` / `HeadersToAuthService`

api/v1alpha1/ext_auth_types.go
arkodg: sure, url sounds good here

api/v1alpha1/ext_auth_types.go
zhaohuabing: Added comment:
 If not specified, no authorization response headers will be added to the original client request.

api/v1alpha1/ext_auth_types.go
zhaohuabing: How about renaming `AllowedUpstreamHeaders` To `AuthServiceHeadersToBackend`  and `AllowedHeaders` To `ClientHeadersToAuthService`?

arkodg: its very long ......
@envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers any suggestions ?

zhaohuabing: Settled with  `AllowedBackendHeaders` until we figure out a better name.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +optional
+	TLS *TLSConfig `json:"tls,omitempty"`
+
+	// Authorization response headers that will be added to the original client
+	// request before sending it to the backend server.
+	// Note that coexisting headers will be overridden.
+	// If not specified, no authorization response headers will be added to the
+	// original client request.
+	// +optional
+	AllowedBackendHeaders []string `json:"allowedBackendHeaders,omitempty"`
```
arkodg, line 94: looks good, thanks !

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+	HTTP *HTTPExtAuthService `json:"http,omitempty"`
+
+	// AllowedHeaders defines the client request headers that will be included
+	// in the request to the external authorization service.
+	// Note: If not specified, the default behavior of different external authorization
+	// services is different. All headers will be included in the check request
+	// to a gRPC authorization server, whereas no headers will be included in the
+	// check request to an HTTP authorization server.
+	// +optional
+	AllowedHeaders []string `json:"allowedHeaders,omitempty"`
```
arkodg, line 51: thoughts on `AllowedExtAuthHeaders` ?

zhaohuabing: Not sure about this: `ExtAuth` here could have multiple meaning

* Headers send to ExtAuth
* Headers from ExtAuth Response

The second one is wrong.

The most unambiguous name should be `AllowedClientHeadersToExtAuth`, but I guess `AllowedClientHeaders` should be clear enough?

arkodg: `AllowedClientHeaders` (from client) conflicts with `AllowedBackendHeaders` (to backend) :)

zhaohuabing: `ClientHeadersToExtAuth` and `ExtAuthHeadersToBackend` ?

kflynn: Emissary's names for these ended up, in practice, often confusing people. For all that `HeadersToExtAuth` and `HeadersToBackend` may sound a bit excessive, I like that they will (hopefully) be very clear -- let's use them. 🙂

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+type HTTPExtAuthService struct {
+	// URL is the URL of the HTTP External Authorization service.
+	// The URL must be a fully qualified URL with a scheme, hostname,
+	// and optional port and path. Parameters are not allowed.
+	// The URL must use either the http or https scheme.
+	// If port is not specified, 80 for http and 443 for https are assumed.
+	// If path is specified, the authorization request will be sent to that path,
+	// or else the authorization request will be sent to the root path.
+	// +kubebuilder:validation:Pattern=`^(https?:)(\/\/[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)((:\d+)?)(\/[a-zA-Z0-9-._~%]+\/?)*$`
+	URL string `json:"url" yaml:"url"`
```
kflynn, line 79: This feels odd. Unless Envoy's extauth protocol has dramatically changed, the HTTP request made to an HTTP extauth implementation mirrors the original URL (optionally with an added prefix), so I think what we need here is the host & port rather than a full URL.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
+}
+
+// GRPCExtAuthService defines the gRPC External Authorization service
+type GRPCExtAuthService struct {
+	// Host is the hostname of the gRPC External Authorization service
+	Host gwapiv1a2.Hostname `json:"host"`
+
+	// Port is the network port of the gRPC External Authorization service
+	Port gwapiv1a2.PortNumber `json:"port"`
```
kflynn, line 60: Documentation note: we should document the specific gRPC message (and protobuf?) that will be used here.

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+	HTTP *HTTPExtAuthService `json:"http,omitempty"`
+
+	// HeadersToExtAuth defines the client request headers that will be
+	// included in the request to the external authorization service.
+	// Note: If not specified, the default behavior of different external
+	// authorization services is different. All headers will be included in the
+	// check request to a gRPC authorization server, whereas no headers will be
+	// included in the check request to an HTTP authorization server.
+	// +optional
+	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
```
kflynn, line 51: Documentation note: there are some headers that end up always needing to be included, whether they're in `HeadersToExtAuth` or not. We should document those.

kflynn (commented):
I actually meant for these review comments to just be inline. Sorry! 🙂

api/v1alpha1/ext_auth_types.go
zhaohuabing: @kflynn Do you mean that some headers should be sent to the ext auth service by default? 
Could you please elaborate on what headers should be added by default?

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +optional
+	TLS *TLSConfig `json:"tls,omitempty"`
+
+	// HeadersToBackend are the authorization response headers that will be added
+	// to the original client request before sending it to the backend server.
+	// Note that coexisting headers will be overridden.
+	// If not specified, no authorization response headers will be added to the
+	// original client request.
+	// +optional
+	HeadersToBackend []string `json:"headersToBackend,omitempty"`
```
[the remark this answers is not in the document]
zhaohuabing: Same as the above. @kflynn Could you please elaborate on what headers should be added by default?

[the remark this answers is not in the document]
zhaohuabing: @kflynn I haven't found these default headers in the Envoy Config: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-authorizationresponse

I guess you are referring `allowed_client_headers`?

zirain (dismissed):
LGTM

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// TLSConfig describes a TLS configuration.
+type TLSConfig struct {
```
arkodg, line 105: is there going to be another field for `ca.crt` or will `ca.crt` be part of this ?

zhaohuabing: Both the ca and private key are part of `TLSConfig`, similar to `GatewayTLSConfig` in the Gateway API.

arkodg:
@zhaohuabing can you link the envoy config that the tls config here will be translated into for `http` and `grpc` ?

zhaohuabing:
> @zhaohuabing can you link the envoy config that the tls config here will be translated into for `http` and `grpc` ?

The`TLSConfig` should be used as the [Cluster](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-msg-config-cluster-v3-cluster) [transport_socket](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-transport-socket) of the GRPC/HTTP ext auth service cluster.

For gRPC service, the cluster is specified in `EnvoyGrpc`'s `cluster_name` field. 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/grpc_service.proto#envoy-v3-api-msg-config-core-v3-grpcservice-envoygrpc
For HTTP service, the cluster is specified in `HttpUri`'s `cluster` field.
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/http_uri.proto#envoy-v3-api-msg-config-core-v3-httpuri

arkodg:
awesome thanks @zhaohuabing ! I brought this up to bring up a common question around proxy identity, do we want to support a unique key/cert for 
* ext auth
* backend TLS

or should we reuse listener/downstream certs ?

zhaohuabing:
> awesome thanks @zhaohuabing ! I brought this up to bring up a common question around proxy identity, do we want to support a unique key/cert for
> 
> * ext auth
> * backend TLS
> 
> or should we reuse listener/downstream certs ?

Envoy docs says that **Only a single TLS certificate is supported in client contexts. **

tls_certificates
(repeated [extensions.transport_sockets.tls.v3.TlsCertificate](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-tlscertificate)) Only a single TLS certificate is supported in client contexts. In server contexts, [Multiple TLS certificates](https://www.envoyproxy.io/docs/envoy/v1.29.0/intro/arch_overview/security/ssl#arch-overview-ssl-cert-select) can be associated with the same context to allow both RSA and ECDSA certificates and support SNI-based selection.

https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/transport_sockets/tls/v3/tls.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-commontlscontext

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// CACertRef is the reference to a Kubernetes ConfigMap that contains a
+	// PEM-encoded TLS CA certificate bundle, which is used to validate the
+	// certificate presented by the external authorization server.
+	// The referenced ConfigMap must contain a key named ca.crt.
+	//
+	// If not specified, the proxy will use the system default certificate pool to
+	// verify the server certificate.
+	//
+	// +optional
+	CACertRef *gwapiv1.LocalObjectReference `json:"caCertRef,omitempty"`
```
arkodg, line 128:
```suggestion
	CACertificateRefs *gwapiv1.LocalObjectReference `json:"caCertRef,omitempty"`
```

arkodg: lets keep this identical to existing API field https://github.com/envoyproxy/gateway/blob/236cba58b587e195728e779f06cc85e249ff43d2/api/v1alpha1/tls_types.go#L127C2-L127C81

api/v1alpha1/ext_auth_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// TLS certificate and private key. The certificate and private key will be
+	// used to establish a TLS handshake between the Envoy proxy and the external
+	// authorization server.
+	// The referenced Secret must contain two keys: tls.crt and tls.key.
+	//
+	// If this field is not specified, the Envoy proxy will not present a client
+	// certificate to the external authorization server.
+	//
+	// +optional
+	CertificateRef *gwapiv1a2.SecretObjectReference `json:"certificateRef,omitempty"`
```
arkodg, line 117: lets rm this for now, we need to decide if we want to reuse downstream listener certs for this or not, and we need to also answer the same question for backend TLS, lets move this to a seperate GH issue

api/v1alpha1/ext_auth_types.go
zhaohuabing: After discussing with @arkodg, we decided to remove all the fields in the TLSConfig for now and address them later in a separate issue since we need to confirm with Gateway API upstream how the client certificate should be specified for Backends.

arkodg (dismissed):
LGTM thanks !

zirain (approved):

arkodg (approved):

PR #2440, opened 2024-01-15
Fix handling the error when parsing the http request timeout for processing the http route rule

zhlsunshine:
Description:
  Handling the error when parsing the http request timeout for processing the http route rule. The timeout could be empty according to the definition of `HTTPRouteTimeouts` in `gateway-api/apis/v1/httproute_types.go`.


zirain:
please fix dco

zhlsunshine:
> please fix dco

Hi @zirain done! ^_^

internal/gatewayapi/route.go
```diff
@@ -221,16 +225,25 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 func processTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
 	if rule.Timeouts != nil {
 		if rule.Timeouts.Request != nil {
-			// TODO: handle parse errors
-			d, _ := time.ParseDuration(string(*rule.Timeouts.Request))
+			// Need to handle parsing errors because the timeout may be empty according to the
+			// definition of HTTPRouteTimeouts in https://github.com/kubernetes-sigs/gateway-api
+			// Set the default timeout to 10s
```
zhaohuabing, line 230: Could you please elaborate why the timeout may be empty when `rule.Timeouts.Request != nil`? Thanks!

zhlsunshine: Hi @zhaohuabing, I think `empty` is inaccuracy, it should be `unspecified`, I will change this comment. Following the definition of `HTTPRouteTimeouts`: `When this field is unspecified, request timeout behavior is implementation-specific`, I recommend that setting the default timeout would be better. Besides, this change also can handle some invalid request timeout setting.

zhaohuabing: Isn't the validation in gateway api strict enough?

```
// Duration is a string value representing a duration in time. The format is as specified
// in GEP-2257, a strict subset of the syntax parsed by Golang time.ParseDuration.
//
// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
type Duration string
```

zhlsunshine: Yes, it's strict enough if the validating web hook can always handle it. If so, shall I change the test case for this PR?

zhlsunshine: Change the value of `HTTPRequestTimeout` based on our discussion. Done!

internal/gatewayapi/route.go
```diff
@@ -221,16 +228,26 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 func processTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
 	if rule.Timeouts != nil {
 		if rule.Timeouts.Request != nil {
-			// TODO: handle parse errors
-			d, _ := time.ParseDuration(string(*rule.Timeouts.Request))
+			// Need to handle parsing errors because the timeout may be unspecified and depended on
```
zhaohuabing, line 231: There should not be errors because gateway api has added validation for the `Duration` type. EG can add another layer of validation here anyway. But the comment is not correct, I suggest removing it.

zhaohuabing (approved):
LGTM. Thanks!

zirain (approved):

internal/gatewayapi/route.go
zhlsunshine: okay, done!

arkodg:
why are we trying to set the timeout if timeout is invalid, won't the default value kick in anyway ? imo if there is a error parsing the timeout, we should flag it in the Status

PR #2454, opened 2024-01-16
feat(translator): Implement BTP Timeout API

guydc:
Implement the BackendTrafficPolicy Timeout API

Fixes https://github.com/envoyproxy/gateway/issues/2401


internal/ir/xds.go
```diff
@@ -391,6 +391,8 @@ type HTTPRoute struct {
 	ExtensionRefs []*UnstructuredRef `json:"extensionRefs,omitempty" yaml:"extensionRefs,omitempty"`
 	// Circuit Breaker Settings
 	CircuitBreaker *CircuitBreaker `json:"circuitBreaker,omitempty" yaml:"circuitBreaker,omitempty"`
+	// Connection timeout Settings
+	BackendTimeout *Timeout `json:"backendTimeout,omitempty" yaml:"backendTimeout,omitempty"`
```
arkodg, line 395: prefer if this was renamed to `Timeout` and we moved the existing `Timeout` field into it and renamed it to `RequestTimeout` under `HTTP`

internal/xds/translator/cluster.go
```diff
@@ -315,13 +312,38 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
+	if !hasCommonHTTPOptions && !isHTTP2 {
+		return nil
+	}
+
+	protocolOptions := httpv3.HttpProtocolOptions{}
+
+	if hasCommonHTTPOptions {
+		protocolOptions.CommonHttpProtocolOptions = &corev3.HttpProtocolOptions{
+			IdleTimeout:           durationpb.New(args.timeout.HTTP.ConnectionIdleTimeout.Duration),
+			MaxConnectionDuration: durationpb.New(args.timeout.HTTP.MaxConnectionDuration.Duration),
```
shawnh2, line 337: these two duration should be checked (whether is nil ptr) before assign values to`IdleTimeout` and `MaxConnectionDuration`, nor it will raise a nil pointer panic.

```yaml
http:
- name: "first-listener"
  #...
  routes:
  - name: "first-route"
    backendTimeout:
      tcp:
        connectTimeout: "31s"
      http:
        connectionIdleTimeout: "32s"
        #maxConnectionDuration: "33s" # paniced
    #...
```

shawnh2 (approved):
lgtm, thanks

zhaohuabing (commented):
This implementation LGTM, thanks!

A minor comment on the  name of `Timeout` API: we may want to change it to `BackendTimeout` because the TCP `ConnectTimeout` can't be reused for the future client-side Timeouts.

guydc:
> A minor comment on the name of `Timeout` structure in the API: we may want to change it to `BackendTimeout` because the TCP `ConnectTimeout` can't be reused for the future client-side Timeouts.

Can we translate this to [transport socket timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#envoy-v3-api-field-config-listener-v3-filterchain-transport-socket-connect-timeout) for downstream connections? If so, it can be reused...

internal/gatewayapi/route.go
```diff
@@ -227,12 +227,20 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 
 func processTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
 	if rule.Timeouts != nil {
+		var rto *ir.Timeout
+
+		if irRoute.Timeout != nil {
```
arkodg, line 232: nit: policy is computed/translated post route translation, so we can assume its empty here

guydc: I wanted to this to be safe in case there's a re-arrangement of the computation order. If you feel that we can make this assumption, i'll update the implementation accordingly.

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -803,8 +819,66 @@ func (t *Translator) buildCircuitBreaker(policy *egv1a1.BackendTrafficPolicy) *i
...
+	var tto *ir.TCPTimeout
+	var hto *ir.HTTPTimeout
+
+	pto := policy.Spec.Timeout
+
+	if pto.TCP != nil && pto.TCP.ConnectTimeout != nil {
+		d, err := time.ParseDuration(string(*pto.TCP.ConnectTimeout))
+		if err != nil {
+			setBackendTrafficPolicyTranslationErrorCondition(policy, "TCP Timeout", fmt.Sprintf("invalid ConnectTimeout value %s", *pto.TCP.ConnectTimeout))
+			return nil
```
arkodg, line 832: we should log error in status and continue processing instead of returning early here

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -803,8 +819,66 @@ func (t *Translator) buildCircuitBreaker(policy *egv1a1.BackendTrafficPolicy) *i
...
+				return nil
+			}
+			cit = ptr.To(metav1.Duration{Duration: d})
+		}
+
+		if pto.HTTP.MaxConnectionDuration != nil {
+			d, err := time.ParseDuration(string(*pto.HTTP.MaxConnectionDuration))
+			if err != nil {
+				setBackendTrafficPolicyTranslationErrorCondition(policy, "HTTP Timeout", fmt.Sprintf("invalid MaxConnectionDuration value %s", *pto.HTTP.MaxConnectionDuration))
+				return nil
```
arkodg, line 856: same as above

internal/gatewayapi/route.go
arkodg: sure, if thats the case, can you also add a comment here, so the next dev can understand why this done this way similar to the comment you have in policy around request timeout

internal/xds/translator/testdata/out/xds-ir/http-route-timeout.routes.yaml
```diff
@@ -14,4 +14,3 @@
       name: first-route
       route:
         cluster: first-route-dest
```
arkodg, line 16: what happened here, why did 5s get deleted ?

internal/xds/translator/testdata/in/xds-ir/http-route-timeout.yaml
```diff
@@ -10,7 +10,8 @@ http:
   routes:
   - name: "first-route"
     hostname: "*"
-    timeout: 5s
+    timeout:
+      requestTimeout: 5s
```
arkodg, line 14: test is broken, probably needs below config
```
timeout:
  http:
    requestTimeout: 5s
```

zhaohuabing:
> > A minor comment on the name of Timeout structure in the API: we may want to change it to BackendTimeout because the TCP ConnectTimeout can't be reused for the future client-side Timeouts.

> Can we translate this to [transport socket timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#envoy-v3-api-field-config-listener-v3-filterchain-transport-socket-connect-timeout) for downstream connections? If so, it can be reused...

Not sure about it, they look similar but not exactly the same. Are there any other possible different settings between client and backend timeouts?

@arkodg @zirain WDYT?

arkodg:
> > > A minor comment on the name of Timeout structure in the API: we may want to change it to BackendTimeout because the TCP ConnectTimeout can't be reused for the future client-side Timeouts.
> 
> > Can we translate this to [transport socket timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#envoy-v3-api-field-config-listener-v3-filterchain-transport-socket-connect-timeout) for downstream connections? If so, it can be reused...
> 
> Not sure about it, they look similar but not exactly the same. Are there any other possible different settings between client and backend timeouts?
> 
> @arkodg @zirain WDYT?

the go struct name can be changed anytime w/o breaking YAML API compatibility so no strong comments for now from my side

guydc:
Regarding E2E test for this feature: echo server does support a `/delay` [endpoint](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/echo-basic/echo-basic.go#L160). However, since these settings are per-connection and not per-request, timed-out connections will not terminate active delayed requests. In other words, it's difficult to demonstrate a connection timeout on a specific request. 

- idle timeout: "The idle timeout is defined as the period in which there are no active requests."
- max connection duration: "When max_connection_duration is reached and if there are no active streams, the connection will be closed" 

One way to verify timeout settings with an e2e test is to trigger some upstream connections, wait for timeouts to occur and then check the relevant envoy cluster metrics. 

I propose to handle the e2e test in a dedicated issue.

arkodg:
sure @guydc lets raise seperate GH issues for e2e and docs and tackle those in different PRs

arkodg (dismissed):
LGTM thanks !

arkodg (approved):
LGTM thanks !

PR #2464, opened 2024-01-20
Use go standard errors instead of multierror

mmorel-35:
**What type of PR is this?**
* chore: use go standard errors instead of multierror

standard errors now supports Joining list of errors so there is no need to rely on custom packages


zirain (approved):

arkodg (approved):
LGTM thanks !

PR #2474, opened 2024-01-22
api: add compressor/decompressor support

### LINKED ISSUE #2451: Support Compression
*Description*:
Support Payload Compression

[optional *Relevant Links*:]

Envoy - https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/compression/compression.html#compression

Upstream Issue - https://github.com/kubernetes-sigs/gateway-api/issues/1799

Relates to https://envoyproxy.slack.com/archives/C03E6NHLESV/p1705084783438099

soulxu:
**What type of PR is this?**
* "feat(translator): add new feature"

Add HTTP compressor/decompressor support

Fixes #2451


soulxu:
In the end, we need to add compressor/decompressor both for BackendTrafficPolicy and ClientTrafficPolicy. The BackendTrafficPolicy defines the compressor/decompressor for connection between envoy proxy and backend service. The ClientTrafficPolicy defines the compressor/decompressor for connection between downstream client and envoy proxy.

The envoy's compressor/decompressor provide very flexible API.  I also define the EnvoyGateway API most flexible. But if we think this is too complicited for the envoy gateway user, we can simplify it. Like we can assume that if the user want to compress the request of envoy proxy to backend, then the user also want to decompress the response of envoy proxy to backend, then we can make the API more easy. Anyway this is something we can consider.

cc @arkodg @akhenakh

soulxu:
> In the end, we need to add compressor/decompressor API both for `BackendTrafficPolicy` and `ClientTrafficPolicy`. The `BackendTrafficPolicy` defines the compressor/decompressor for connection between the envoy proxy and backend service. The `ClientTrafficPolicy` defines the compressor/decompressor for connection between the downstream client and envoy proxy.
> 
@akhenakh the ClientTrafficPolicy should match your usecase. But let me know if I missed something. thanks!

zhaohuabing (commented):
I suppose request compression and response decompression are rarely used? We probably doesn't need to support them at the first iteration.

So this should only be in `ClientTrafficPolicy`.

arkodg:
* imo the deciding factor between `BackendTrafficPolicy` and `ClientTrafficPolicy` is whether `compression` is more likely to be configured by the app team or platform team
* Id vote to expose minimal config in the first iteration e.g.
```
compression:
  type: Gzip
```
and either set defaults or rely on the envoy defaults
* prefer if this API PR only focuses on `compression` (can raise another PR later from `decompression`) since sending compressed response to the browser seems like the most common use case to solve

soulxu:
> * imo the deciding factor between `BackendTrafficPolicy` and `ClientTrafficPolicy` is whether `compression` is more likely to be configured by the app team or platform team
> * Id vote to expose minimal config in the first iteration e.g.
> 
> ```
> compression:
>   type: Gzip
> ```
> 
> and either set defaults or rely on the envoy defaults
> 
> * prefer if this API PR only focuses on `compression` (can raise another PR later from `decompression`) since sending compressed response to the browser seems like the most common use case to solve

@arkodg @zhaohuabing thanks for the feedback! I like the way just provide minimal config first, let us add more in the future. let me update the PR.

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -86,6 +86,16 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	Timeout *Timeout `json:"timeout,omitempty"`
+
+	// The compression config for the backend connections.
```
zirain, line 90: compression/decompression will reduce the body size, at the same it will increase the CPU usage, right?

soulxu: yes, it is all about the tradeoff I think.

zirain: can you point out this in the api doc?

soulxu: yea, will do. thanks for the review!

soulxu:
> * prefer if this API PR only focuses on `compression` (can raise another PR later from `decompression`) since sending compressed response to the browser seems like the most common use case to solve

@arkodg @zhaohuabing  I only keep the compression for the client traffic policy and backend traffic policy. That means the initial implementation only allow two cases:
* the compression of the response from the upstream to the envoy proxy
* the compression of the request from envoy proxy to backend

Let me know if this matches what you expect.

Thanks!

zirain:
@soulxu plz fix the conflict.

zhaohuabing:
> * the compression of the request from envoy proxy to backend

I thought only enable compression on the response and possibly decompression for the request?

api/v1alpha1/compression_types.go
```diff
@@ -0,0 +1,28 @@
...
+package v1alpha1
+
+type GzipCompressor struct {
+}
+
+type Compression struct {
+	// A compressor library to use for compression
+	//
+	// +required
+	CompressorLibrary *CompressorLibrary `json:"compressorLibrary,omitempty"`
```
arkodg, line 15: compressionLibrary adds an extra level, 
prefer
```
compression:
  type: "Gzip"
    gzip:
```

and in the init API maybe we can skip `gzip` struct and rely on defaults if possible

soulxu: got it, thanks!

soulxu:
> > * the compression of the request from envoy proxy to backend
> 
> I thought only compression on the response, possibly decompression for the request (this can be done later as @arkodg suggested)?

cool, let me make it more simple, thanks!

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -78,6 +78,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	Path *PathSettings `json:"path,omitempty"`
+	// The compression config for the downstream client connections.
```
arkodg, line 81: lets define this config in one place for now, I vote to rm this one and keep the one in `backendTrafficPolicy`

soulxu: Emm... then how we distinguish the upstream and downstream side compression in the future?

arkodg: afaik compression is defined at a request/response level per listener/route in envoy https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/compressor_filter

soulxu: Yes, for the only downstream side compression, we will enable the response direction compression by `response_direction_config` and disable the request direction compression by the `request_direction_config`

For the only upstream side compression, we will enable the requestion direction compression by `request_direction_config` and diable the response direction compression by `response_direction_config`

soulxu: > Yes, for the only downstream side compression, we will enable the response direction compression by `response_direction_config` and disable the request direction compression by the `request_direction_config`

This supposed to be defined by backend traffic policy.

> 
> For the only upstream side compression, we will enable the requestion direction compression by `request_direction_config` and diable the response direction compression by `response_direction_config`

This supposed to be defined by client traffic policy. That confuses me if define this in the backend traffic policy. What we do when want to enable the dowstream side compression?

arkodg: its defined in terms of `request` and `response` and `request` is disabled by default, in the future we can enhance config and enable request compression if needed
```
compression:
  type: Gzip
  gzip:
    request:
    response:
       ......
```

arkodg: a downstream request is also a upstream request, so it cannot really be differentiated b/w ClientTraffic and BackendTraffic imo. I'd put it in `BackendTraffic` for two reasons
1. feature is relevant to app teams
2. feature can be enabled/disabled at the route level

soulxu: ah, I begin to get your point now. thanks for explanation this for me!

> a downstream request is also a upstream request, so it cannot really be differentiated b/w ClientTraffic and BackendTraffic imo. I'd put it in `BackendTraffic` for two reasons
> 
> 1. feature is relevant to app teams

May I ask how to define this? I can understand that for the upstream side, the only the application knows that whether the application has the ability to compressor and decompressor.

For the downstream side, why this is defined by the app teams? since the app teams also know the ability of the applciation?

> 2. feature can be enabled/disabled at the route level

arkodg: > since the app teams also know the ability of the applciation

yeah since they have a better understanding of their L7 client, this is just my initial thought on where to put the config, would be good to hear counter arguments as well

soulxu: > > 2. feature can be enabled/disabled at the route level
> >    +1

Will we have something like PreRouteTrafficPolicy in the future? I didn't something like this today.

arkodg: no, the `BackendTrafficPolicy` can be attached either to the `Gateway` or to the `Route`

api/v1alpha1/shared_types.go
```diff
@@ -36,6 +36,11 @@ type GroupVersionKind struct {
 	Kind    string `json:"kind"`
 }
 
+// CompressorLibraryType defines the types of compressor library supported by Envoy Gateway.
```
arkodg, line 39: can we move this definition into compression_types.go

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -86,6 +86,11 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	Timeout *Timeout `json:"timeout,omitempty"`
+
+	// The compression config for the http streams.
+	//
+	// +optional
+	Compression []*Compression `json:"compression,omitempty"`
```
arkodg, line 93: why this is a list ?

api/v1alpha1/clienttrafficpolicy_types.go
soulxu: oop, I missed that, thanks!

api/v1alpha1/backendtrafficpolicy_types.go
soulxu: to support enable multiple compression algorithm

api/v1alpha1/compression_types.go
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+type GzipCompressor struct {
+}
+
+type Compression struct {
```
arkodg, line 11: can we add a doc string around defaults ? enabled for responses and disabled for requests as well as per compression type defaults

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: is that a valid use case ?

api/v1alpha1/compression_types.go
soulxu: yup ,will do, thanks

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: rethinking, feels like a valid use case to enable both gzip and brotli and decide based on client capability.
for that case, should the config look more like
```
compression:
  gzip: {}
  brotli: {}
```

the type is no longer needed since we aren't picking 1
a list also works, but extra validation needs to be added to ensure a specific type is not enabled multiple times

soulxu: yes, at least from the Envoy side https://github.com/envoyproxy/gateway/pull/2474#discussion_r1465707302

There is also option `choose_first` https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/compressor/v3/compressor.proto for the case of multiple compressors.

arkodg: cool, in the future, we could consider adding a `priority` field to orchestrate order

soulxu: > rethinking, feels like a valid use case to enable both gzip and brotli and decide based on client capability. for that case, should the config look more like
> 
> ```
> compression:
>   gzip: {}
>   brotli: {}
> ```
> 
> the type is no longer needed since we aren't picking 1 a list also works, but extra validation needs to be added to ensure a specific type is not enabled multiple times

@arkodg sorry, just found I missed this comment.

I use the list since I'm afraid there are configs may different for the different compressors. Like they may use the different `content_length`. If we put them into one `compression` message, then we will block this case.

arkodg: can you share a use case where we'd define `gzip` compression twice ?

api/v1alpha1/clienttrafficpolicy_types.go
zhaohuabing: > > since the app teams also know the ability of the applciation
> 
> yeah since they have a better understanding of their L7 client, this is just my initial thought on where to put the config, would be good to hear counter arguments as well

@arkodg I got your reason for putting this in the `BackendTrafficPolicy`, but I also feel it is against the intuition of the users.

The most two common use cases for `Compression` and `Decompression` are:
* Compression: compress the response from the backend to the client to save the network bandwidth.
* Decompression: decompress the request from the client to the backend to avoid the backend to do the decompression.

From the user's point of view, these two use cases are more related to client-proxy communication. I guess the real issue here is that the ClientTrafficPolicy is not able to do per-route configuration.

api/v1alpha1/backendtrafficpolicy_types.go
soulxu: no, I don't think we are going to define the same compression twice. I think it is all about multiple different compression.

zirain: if so, prefer 
```yaml
compression:
  gzip: {}
  brotli: {}
```

arkodg: in that case, imo we shouldnt use a list (which makes validation hard), instead use something like
```
compression:
  gzip: {}
  brotli: {}
 ```

File: api/v1alpha1/clienttrafficpolicy_types.go (hunk 3 above)
Comment (arkodg) [reply api/v1alpha1/clienttrafficpolicy_types.go]: imo the nouns we could have come up with were
* trafficPolicy - to deal with L7 request/response transformations
* clientConnectionPolicy - for client connection
* backendConnectionPolicy - for backend connection

`backendTrafficPolicy` imo is combining `backendConnectionPolicy` + `trafficPolicy`

File: api/v1alpha1/backendtrafficpolicy_types.go (hunk 5 above)
Comment (soulxu) [reply api/v1alpha1/backendtrafficpolicy_types.go:93]: emm...sorry, I found a case for multiple same compressor. 

```
    // There could be many compressors registered for the same content encoding, e.g. consider a
    // case when there are two gzip filters using different compression levels for different content
    // sizes. In such case we ignore duplicates (or different filters for the same encoding)
    // registered last.
```
https://github.com/envoyproxy/envoy/blob/45ab9cfa30c7631fe325847dd291e2dcc80c1894/source/extensions/filters/http/compressor/compressor_filter.cc#L389-L392

Comment (soulxu) [reply api/v1alpha1/backendtrafficpolicy_types.go:93]: And we probably need the order of multiple compressors also.

File: api/v1alpha1/clienttrafficpolicy_types.go (hunk 3 above)
Comment (soulxu) [reply api/v1alpha1/clienttrafficpolicy_types.go]: > imo the nouns we could have come up with were
> 
> * trafficPolicy - to deal with L7 request/response transformations

Thinking of the role for the compression API. I feel there are two rules may involve this.

One is the application developer, just as we discussed above, they know the capabilities of the application.
The second one is the cluster operator, like the case of @zhaohuabing said, the cluster operator wants to save the bandwidth.

But for the second case, it should be under the service mesh case. There is sidecar proxy in the front of the backend endpoint. Then cluster operator needn't care about the application capabilities.

To fulfill this case, in the future, we can have two APIs, `ApplicationTrafficPolicy` and `TrafficPolicy`.
I saw the similar case in the TLSBackendPolicy https://gateway-api.sigs.k8s.io/geps/gep-1897/#api

For today, we can just simplify and have one.

File: api/v1alpha1/compression_types.go [hunk 7]
```diff
@@ -0,0 +1,34 @@
...
+
+// CompressorType defines the types of compressor library supported by Envoy Gateway.
+//
+// +kubebuilder:validation:Enum=Gzip
+type CompressorType string
+
+// GzipCompressor defines the config for the Gzip compressor. There are some configs
+// available from the Envoy Proxy. Currently only use the default value for configs.
+// For the default value for those configs can be reference here:
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/compression/gzip/compressor/v3/gzip.proto#extension-envoy-compression-gzip-compressor
```
arkodg, line 16:
```suggestion
// GzipCompressor defines the config for the Gzip compressor.
// T he default values can be found here:
// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/compression/gzip/compressor/v3/gzip.proto#extension-envoy-compression-gzip-compressor
```

api/v1alpha1/compression_types.go
```diff
@@ -0,0 +1,34 @@
...
+type GzipCompressor struct {
+}
+
+// Compression defines the config of compression for the http streams. Currently
+// only the minial config was added. All configs from the Envoy Proxy are using
+// the default value. For the default value can be reference here:
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/compressor/v3/compressor.proto#extensions-filters-http-compressor-v3-compressor
+type Compression struct {
+	// CompressorType defines which type compressor wants to use for compression.
```
arkodg, line 25:
```suggestion
	// CompressorType defines the compressor type to use for compression.
```

api/v1alpha1/compression_types.go
```diff
@@ -0,0 +1,34 @@
...
+	//
+	// +required
+	Type CompressorType `json:"type,omitempty"`
```
arkodg, line 28:
```suggestion
	Type CompressorType `json:"type"`
```

api/v1alpha1/compression_types.go
```diff
@@ -0,0 +1,34 @@
...
+// available from the Envoy Proxy. Currently only use the default value for configs.
+// For the default value for those configs can be reference here:
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/compression/gzip/compressor/v3/gzip.proto#extension-envoy-compression-gzip-compressor
+type GzipCompressor struct {
+}
+
+// Compression defines the config of compression for the http streams. Currently
+// only the minial config was added. All configs from the Envoy Proxy are using
+// the default value. For the default value can be reference here:
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/compressor/v3/compressor.proto#extensions-filters-http-compressor-v3-compressor
```
arkodg, line 23:
```suggestion
// Compression defines the config of enabling compression.
// This can help reduce the bandwidth at the expense of higher CPU
```

arkodg (approved):
LGTM thanks

zirain (approved):

PR #2477, opened 2024-01-22
fix: fix checkObjectNamespaceLabels func param type

ShyunnY:
**What type of PR is this?**

Fix the checkObjectNamespaceLabels function param type so that it can support passing in common interface types. Avoid obtaining namespace in advance outside the function.

Fixes #

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -580,11 +580,11 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to 
 	if len(r.namespaceLabels) != 0 {
 		var rgs []gwapiv1b1.ReferenceGrant
 		for _, refGrant := range refGrants {
-			ns := refGrant.GetNamespace()
-			ok, err := r.checkObjectNamespaceLabels(ns)
+			refGrant := refGrant
+			ok, err := r.checkObjectNamespaceLabels(&refGrant)
 			if err != nil {
 				// TODO: should return? or just proceed?
-				return nil, fmt.Errorf("failed to check namespace labels for ReferenceGrant %s in namespace %s: %w", refGrant.GetName(), ns, err)
+				return nil, fmt.Errorf("failed to check namespace labels for ReferenceGrant %s in namespace %s: %w", refGrant.GetName(), refGrant.GetNamespace(), err)
 			}
 			if !ok {
 				// TODO: should log?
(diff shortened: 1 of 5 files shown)


zirain (approved):

ShyunnY:
@zirain The rest of the files cannot be covered by the test, do I need to deal with it?

shawnh2 (changes requested):
hi @ShyunnY please fix linter, thanks

ShyunnY:
> hi @ShyunnY please fix linter, thanks

Yes, I have fix it

internal/provider/kubernetes/controller.go
```diff
@@ -580,8 +580,10 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
 	if len(r.namespaceLabels) != 0 {
 		var rgs []gwapiv1b1.ReferenceGrant
 		for _, refGrant := range refGrants {
+			refGrant := refGrant
+
 			ns := refGrant.GetNamespace()
```
arkodg, line 585: are these `ns` needed ?

arkodg: same for other types

ShyunnY: In principle, we don’t need ns here, but I considered the future scalability and may still use this ns, so I kept it.

arkodg: if thats the case, do we need to change the signature of `checkObjectNamespaceLabels` ?

ShyunnY: First of all, I think that the step of obtaining the namespace in advance before calling the function may be redundant externally. We should move this step inside the function and let the function do this step for us.
Secondly, we may expand the judgment conditions in the future, using the interface form of metav1.object as a parameter, which will allow us to obtain more information.
Finally, subsequent modifications to this function will not affect external calls.

shawnh2: vote for removing these `ns`, since there's no other particular usage. if we encounter this in the future, we can simply change it back.

internal/provider/kubernetes/predicates.go
```diff
@@ -61,10 +62,14 @@ type NamespaceGetter interface {
...
 
 // checkObjectNamespaceLabels checks if labels of namespace of the object is a subset of namespaceLabels
-// TODO: check if param can be an interface, so the caller doesn't need to get the namespace before calling
-// this function.
-func (r *gatewayAPIReconciler) checkObjectNamespaceLabels(nsString string) (bool, error) {
-	// TODO: add validation here because some objects don't have namespace
+func (r *gatewayAPIReconciler) checkObjectNamespaceLabels(obj matav1.Object) (bool, error) {
+
+	var nsString string
+	// it requires extra condition?
```
shawnh2, line 68: is this comment a TODO?

ShyunnY: Strictly speaking, this may be a TODO. Because we need to judge cluster resources and some resources that do not have namespace.

internal/provider/kubernetes/controller.go
ShyunnY: Yes, of course we can do that. One reason why I didn't remove ns is because the subsequent error message used the ns variable. I didn't want to modify the original code too much. I thought that this ns variable might be used in the future, so I made it a retained variable. rather than removing it.
@shawnh2

internal/provider/kubernetes/predicates.go
shawnh2: got it! can you make this comment more clear? it will help those who come across it later.

internal/provider/kubernetes/controller.go
shawnh2: imo, removing these `ns` is kinda point of this PR. as for those used in error msg, directly using `xxx.GetNamespace()` works for me.

internal/provider/kubernetes/predicates.go
ShyunnY: yes, i have do that

shawnh2 (approved):
lgtm, thanks

arkodg (approved):

PR #2489, opened 2024-01-23
feat: support Envoy extra args

### LINKED ISSUE #2479: Support disablement of Envoy extensions
*Description*:
We would like to be able to disable envoy extensions as part of security hardening.
In order to achieve that `--disable-extensions <extension list>` cli flag needs to be provided to Envoy.

Today there are specific Envoy cli flags that can be configured explicitly as part of [EnvoyProxySpec](https://gateway.envoyproxy.io/latest/api/extension_types/#envoyproxyspec):
* [--log-level](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-l)
* [--component-log-level](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-component-log-level)
* [--concurrency](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-concurrency)
* [--config-yaml](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-config-yaml)

One approach to support `--disable-extensions` is to add an explicit configuration under EnvoyProxySpec.

A more generic approach would be to add support for `EnvoyExtraArgs` configuration, also under EnvoyProxySpec.
This will allow support of other Envoy cli args without changes in the code.
However, this approach raises some questions:
* Overriding existing args(i.e. `--log-level`) would cause Envoy to crash with `Argument already set!` error. Is this an acceptable behavior or do we want to somehow add a validation? Should args overriding be supported at all?
* Using non supported args will cause Envoy to crash with `Could not find match for argument` error.  Is this an acceptable behavior or do we want to somehow add a validation? I guess that we want to avoid holding a list of all supported cli args for validation.

[optional *Relevant Links*:]
https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-disable-extensions
https://github.com/envoyproxy/envoy/blob/main/source/extensions/extensions_metadata.yaml

shahar-h:
Add support for providing Envoy extra [command line options](https://www.envoyproxy.io/docs/envoy/latest/operations/cli) via `EnvoyProxy` resource.
Sample use case: disable some Envoy extension by providing [--disable-extensions](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-disable-extensions) arg.

Fixes #2479


arkodg (approved):
LGTM thanks !

api/v1alpha1/envoyproxy_types.go
```diff
@@ -67,6 +67,12 @@ type EnvoyProxySpec struct {
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
 
+	// ExtraArgs defines additional command line options that are provided to Envoy.
+	// More info: https://www.envoyproxy.io/docs/envoy/latest/operations/cli#command-line-options
+	//
```
zirain, line 72: can you add a warning message about duplicate arg?

shawnh2 (approved):
lgtm, thanks

api/v1alpha1/envoyproxy_types.go
zhaohuabing: We can tell users which args can/can't be configured through `ExtraArgs` in the comment here. So they won't have to try and guess.

zhaohuabing (changes requested):

api/v1alpha1/envoyproxy_types.go
shahar-h: Should I add a generic message or specify all internal args that can't be used(`--service-cluster`, `--service-node`, `--log-level`, `--conponent-log-level`, `--concurrency`, `--config-yaml`, `--cpuset-threads`)?

shahar-h:
Should I also add a sample usage in the [user guide](https://gateway.envoyproxy.io/v0.6.0/user/customize-envoyproxy/)?

api/v1alpha1/envoyproxy_types.go
zirain: I think it's fine to not list all of them.

zirain:
> Should I also add a sample usage in the [user guide](https://gateway.envoyproxy.io/v0.6.0/user/customize-envoyproxy/)?

yes, that would be better, and you can do that with a follow up PR.

api/v1alpha1/envoyproxy_types.go
shahar-h: That's what I thought, thanks

shahar-h: Added

zhaohuabing: It may be useful to tell users which args can/can't be configured through `ExtraArgs` in the comment here. So they won't need to try and guess. And we can add validation here as well.

zirain (commented):
LGTM, defer to @arkodg

arkodg:
hey @shahar-h can you run `make generate` and commit those changes and repush ?

shahar-h:
> hey @shahar-h can you run `make generate` and commit those changes and repush ?

done

arkodg (approved):

Issue #88, opened 2022-06-02
Use Case - Front Proxy with mTLS


jasonmccallister:
Last meeting we discussed providing some use cases on how people are deploying Envoy in production. We have somewhat of an interesting use case I thought I would share.

We deploy Envoy as a front proxy that sits behind a CDN. Our Envoy Proxy sits behind a network load balancer and has one TLS listener that requires all connections present a client certificate signed by the Cloudflare CA. Many CDN providers, such as Cloudflare and Fastly, provide a way to authenticate origin requests using a CA/self-signed certificate. See the [Cloudflare Authenticated Origin Pulls documentation](https://developers.cloudflare.com/ssl/origin-configuration/authenticated-origin-pull/explanation/) for additional details. This provides similar benefits as deploying Envoy as a sidecar with mTLS; only requests that originate from the CDN provider with a valid certificate will be accepted by the Envoy Proxy. 

This affords a few benefits as a front proxy:

1. Requires all requests come from the CDN provider (Cloudflare, etc.). If someone does manage to discovery a load balancer DNS record they cannot spoof headers/etc.
2. Forcing requests through the CDN provider also ensures requests are filtered by the WAF and DDOS protection is applied by the CDN.

Happy to discuss any specific challenges on getting this deployed and configured correctly, but this is one of our use cases.

### Example 

An example listener would look something like this:

```yaml
static_resources:
  listeners:
    - address:
        socket_address:
          address: 0.0.0.0
          port_value: 8443
      listener_filters:
        - name: "envoy.filters.listener.tls_inspector"
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector
      per_connection_buffer_limit_bytes: 32768 # 32 KiB
      access_log:
        - name: envoy.access_loggers.stdout
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog
      filter_chains:
        - filters:
            - name: envoy.filters.network.http_connection_manager
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
                codec_type: AUTO
                stat_prefix: ingress_tls
                rds:
                  route_config_name: ingress_tls
                  config_source:
                    api_config_source:
                      api_type: REST
                      transport_api_version: V3
                      cluster_names:
                        - xds_proxy
                      refresh_delay: 10s
[...]

PR #2490, opened 2024-01-23
feat: downstream mTLS

arkodg:
Relates to https://github.com/envoyproxy/gateway/pull/2483


internal/xds/translator/listener.go
```diff
@@ -400,6 +399,16 @@ func buildDownstreamQUICTransportSocket(tlsConfig *ir.TLSConfig) (*corev3.Transp
 			})
 	}
 
+	if tlsConfig.CACertificate != nil {
```
guydc, line 402: consider also sending the client certificate upstream, e.g.  by updating the HCM default for [forwarding](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-forwardclientcertdetails) to `SANITIZE_SET`. This is not strictly required for the use case in #88, but is a common requirement when a proxy performs client cert auth.

arkodg: I'll raise a separate issue for propagating `XFCC`

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -490,4 +500,55 @@ func translateListenerTLSParameters(tlsParams *egv1a1.TLSSettings, httpIR *ir.HT
...
+					return err
+				}
+
+				secretBytes, ok := secret.Data[caCertKey]
+				if !ok || len(secretBytes) == 0 {
+					return fmt.Errorf(
+						"caCertificateRef not found in secret %s", caCertRef.Name)
+				}
+
+				irCACert.Certificate = append(irCACert.Certificate, secretBytes...)
```
guydc, line 528: should we also verify that this is an x509 conforming certificate? What's the impact of a malformed certificate provided via SDS on translation of other resources/data plane behavior?

arkodg: raised https://github.com/envoyproxy/gateway/issues/2601 to track this

guydc (approved):

zirain (approved):

PR #2491, opened 2024-01-24
feat: add supported features to gateway class

levikobi:
**What type of PR is this?**

feat: add supported features to gateway class.

This PR implements [GEP-2162](https://github.com/kubernetes-sigs/gateway-api/blob/f80f447fbf6e759a196cab6532b70582d9a8bd70/geps/gep-2162/index.md), which is about adding a list of supported features to the status of the gateway class.


internal/status/gatewayclass.go
```diff
@@ -23,3 +25,25 @@ func SetGatewayClassAccepted(gc *gwapiv1.GatewayClass, accepted bool, reason, ms
 	gc.Status.Conditions = MergeConditions(gc.Status.Conditions, computeGatewayClassAcceptedCondition(gc, accepted, reason, msg))
 	return gc
 }
+
+// GetSupportedFeatures returns a list of supported Gateway-API features,
```
arkodg, line 29: recommend removing all the additions in the above files and inserting this inside `SetGatewayClassAccepted` in this file

internal/status/gatewayclass.go
```diff
@@ -23,3 +25,25 @@ func SetGatewayClassAccepted(gc *gwapiv1.GatewayClass, accepted bool, reason, ms
...
+// based on the running conformance tests suite.
+func GetSupportedFeatures() []gwapiv1.SupportedFeature {
+
+	// TODO(levikobi): This must be in sync with the cSuite supported features.
```
arkodg, line 33: would be ideal to define this, https://github.com/envoyproxy/gateway/blob/100d3103363c64a5be480ee0a2d66e585a9882f1/test/conformance/conformance_test.go#L49 & https://github.com/envoyproxy/gateway/blob/100d3103363c64a5be480ee0a2d66e585a9882f1/test/conformance/experimental_conformance_test.go#L98 once and use the same definition in all 3 places

arkodg:
hey @levikobi suggest running `make testdata`, this change probably ends up updates a bunch of golden yamldata

internal/status/gatewayclass.go
levikobi: Created a new file under `conformance` package with the unified declaration, let me know if that's ok

internal/status/gatewayclass.go
levikobi: The function you recommended removing is mainly required for the tests, as we need to get the list of supported features. How would you recommend getting the supported features in tests if we remove this function?

levikobi:
> hey @levikobi suggest running `make testdata`, this change probably ends up updates a bunch of golden yamldata

Thank you for your review @arkodg!
I addressed your comments, `make testdata` also seems to work now locally.

internal/cmd/egctl/translate.go
```diff
@@ -274,11 +274,13 @@ func translateGatewayAPIToGatewayAPI(resources *gatewayapi.Resources) (gatewayap
 			epInvalid = true
 			msg := fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err)
 			status.SetGatewayClassAccepted(resources.GatewayClass, false, string(gwapiv1.GatewayClassReasonInvalidParameters), msg)
+			status.SetGatewayClassSupportedFeatures(resources.GatewayClass)
```
arkodg, line 277: what I meant in my previous comment was if you called `SetGatewayClassSupportedFeatures` within `status.SetGatewayClassAccepted`, its easier to manage the code in case the code updates in the future

test/conformance/conformance_features.go
```diff
@@ -0,0 +1,17 @@
...
+// the root of the repo.
+
+package conformance
+
+import (
+	"sigs.k8s.io/gateway-api/conformance/tests"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+var CommonEnvoySuite = suite.Options{
```
arkodg, line 13:
```suggestion
var EnvoyGatewaySuite = suite.Options{
```

internal/status/gatewayclass.go
arkodg: I actually meant https://github.com/envoyproxy/gateway/pull/2491/files#r1468272780, hope this helps

internal/cmd/egctl/translate.go
levikobi: Thanks for the clarification. I agree entirely

arkodg:
@levikobi your changes might have affected e2e runs

levikobi:
> @levikobi your changes might have affected e2e runs

Apparently the `SupportedFeatures` list has a kubebuilder validation that checks for a hard-coded list of features (see [here](Gateway;GatewayPort8080;GatewayStaticAddresses;HTTPRoute;HTTPRouteDestinationPortMatching;HTTPRouteHostRewrite;HTTPRouteMethodMatching;HTTPRoutePathRedirect;HTTPRoutePathRewrite;HTTPRoutePortRedirect;HTTPRouteQueryParamMatching;HTTPRouteRequestMirror;HTTPRouteRequestMultipleMirrors;HTTPRouteResponseHeaderModification;HTTPRouteSchemeRedirect;Mesh;ReferenceGrant;TLSRoute)).
The supported features I took from the conformance suite contain features that aren't part of this list, which makes the update fail. That's why only e2e fails and not unit tests.

I saw that Cillium hard-coded their list of supported features. I'll dig around to see what others have done to develop a proper solution.
@arkodg If you have any preferences, please let me know.

internal/status/gatewayclass.go
```diff
@@ -14,12 +14,29 @@
...
+func GetSupportedFeatures() []gwapiv1.SupportedFeature {
+	supportedFeatures := conformance.EnvoyGatewaySuite.SupportedFeatures
+	supportedFeatures.Delete(conformance.EnvoyGatewaySuite.ExemptFeatures.UnsortedList()...)
+
+	ret := sets.New[gwapiv1.SupportedFeature]()
+	for _, feature := range supportedFeatures.UnsortedList() {
+		ret.Insert(gwapiv1.SupportedFeature(feature))
+	}
+	return sets.List(ret)
+}
```
LiorLieberman, line 34: consider computing this only once. otherwise you would do this computation every time the GWC is reconciled. 
Similar to the feedback I got in https://github.com/istio/istio/pull/48770

levikobi: Good idea, thanks!

levikobi:
The issue around the list of supported features is solved upstream via https://github.com/kubernetes-sigs/gateway-api/pull/2754 so I believe the tests should pass now

arkodg:
@levikobi dont we need to wait for a gateway-api release that has the change in CRDs ? 
currently pinned to v1.0.0 https://github.com/envoyproxy/gateway/blob/f21f9ff6590a8d52c1366ea0aa60228811711378/go.mod#L46

levikobi:
> @levikobi dont we need to wait for a gateway-api release that has the change in CRDs ? currently pinned to v1.0.0
> 
> https://github.com/envoyproxy/gateway/blob/f21f9ff6590a8d52c1366ea0aa60228811711378/go.mod#L46

Definitely, my bad

LiorLieberman:
I think a release should be out relatively soon (1-2 weeks from now). I suggest you keep an eye on it and merge after the release is out

levikobi:
> I think a release should be out relatively soon (1-2 weeks from now). I suggest you keep an eye on it and merge after the release is out

Waiting for this release. Once it's out, I'll update this PR (rebase + the change you suggested @LiorLieberman)

LiorLieberman:
@levikobi i think rc1 is out. 1.1 should be out soon after we get feedback on the rc

levikobi:
> @levikobi i think rc1 is out. 1.1 should be out soon after we get feedback on the rc

Thanks @LiorLieberman
I rebased and resolved the opened comments. Once the release is out we should be ready to finish things up here.

arkodg:
blocked on https://github.com/envoyproxy/gateway/issues/3265

arkodg:
hey @levikobi v1.1.0-rc2 is in, you should be unblocked now

levikobi:
Thanks @arkodg 
I just updated the PR

internal/cmd/egctl/testdata/translate/out/invalid-envoyproxy.all.yaml
```diff
@@ -38,6 +38,34 @@ gatewayClass:
       reason: InvalidParameters
       status: "False"
       type: Accepted
+    supportedFeatures:
+    - GRPCRoute
+    - Gateway
+    - GatewayHTTPListenerIsolation
```
arkodg, line 44: the skip test values still showed up here, they shouldnt

levikobi: Just to make sure I understand:
We have a list of `SkipTests.` Should we delete the `Features` of every skipped test from the final supported features list?

levikobi: I also noticed that the list of `MeshExtendedFeatures` appears in the supported features. This happens since we only specify `MeshCoreFeatures` in the `ExemptFeatures` list. It doesn't make sense (correct me if I'm wrong) to have the extended without the core.

Should I add the `MeshExtendedFeatures` to the `ExemptFeatures` list?

arkodg: Yes we should be deleting them from this list because we don't support it yet :) 

Sure let's also MeshExtendedFeatures to ExemptFeatures

mikemorris:
@robscott Have you thought about how GKE might handle this (given the whole lack of in-cluster GatewayClass resources approach we share on AKS) if we try to go in the direction of using this advertisement of supported functionality to configure the conformance suite?

/cc @snehachhabria

internal/cmd/egctl/testdata/translate/out/invalid-envoyproxy.all.yaml
levikobi: Thanks for the clarification @arkodg.  I've added the requested change.
Please notice though that features such as `Gateway` and `HTTPRoute` have been removed (as a dependency of skipped test named `GatewayHTTPListenerIsolation`).

arkodg: thanks for highlighting that, that is something worth discussing in upstream since `GatewayHTTPListenerIsolation` is an extended test

arkodg (dismissed):
LGTM thanks !

we probably need a follow up to decide whether this is opt in / opt out or always on once we gather more feedback from users

internal/gatewayapi/status/gatewayclass.go
```diff
@@ -17,7 +17,11 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+
+	"github.com/envoyproxy/gateway/test/conformance"
```
zirain, line 24: it's not a good idea to import a test package here.

test/conformance/conformance_features.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+	"sigs.k8s.io/gateway-api/pkg/features"
+)
+
+// SkipTests is a list of tests that are skipped in the conformance suite.
+var SkipTests = []suite.ConformanceTest{
+	tests.GatewayStaticAddresses,
+	tests.GatewayHTTPListenerIsolation,          // https://github.com/kubernetes-sigs/gateway-api/issues/3049
+	tests.HTTPRouteBackendRequestHeaderModifier, // https://github.com/envoyproxy/gateway/issues/3338
+}
```
zirain, line 20: we can move this under `internal` to avoid import test package.

levikobi: I tried to move `conformance_features.go` to `internal/gatewayapi` package, but that leads to an import cycle
```
package github.com/envoyproxy/gateway/internal/gatewayapi/status
	imports github.com/envoyproxy/gateway/internal/gatewayapi
	imports github.com/envoyproxy/gateway/internal/gatewayapi/status: import cycle not allowed
```

@zirain do you maybe have a suggestion regarding where you would like this file to go?
Do you want it straight under `internal` (`internal/conformance_features.go`)?

levikobi:
> LGTM thanks !
> 
> we probably need a follow up to decide whether this is opt in / opt out or always on once we gather more feedback from users

Thanks for the multiple reviews @arkodg!
I'm in for the follow up

test/conformance/conformance_features.go
zirain: what about `internal/gatewayapi/conformance/features.go`

levikobi: done

internal/gatewayapi/status/gatewayclass.go
levikobi: fixed

shawnh2 (approved):
LGTM! Thanks

zirain (approved):

PR #2487, opened 2024-01-23
api: Add support for enabling Trailers in HTTP/1.1

liorokman:
**What this PR does / why we need it**:

This PR adds support in `ClientTrafficPolicy` for enabling HTTP/1 trailers.

If enabled, it will translate to the `enable_trailers` attribute in Envoy's [HTTP1ProtocolOptions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto.html#config-core-v3-http1protocoloptions) configuration.

PR #2247, opened 2023-11-28
Add Upstream TLS Support

TasdidurRahman:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #https://github.com/envoyproxy/gateway/issues/2033

PR #2492, opened 2024-01-24
feat: Implement enabling HTTP/1.1 trailers

liorokman:

This PR implements #2487 .

I recommend that issue #2247 be updated so that if upstream TLS is enabled, then instead of forcing HTTP/1.1 if trailers are enabled by using `HttpProtocolOptions_ExplicitHttpConfig` in the cluster setup, ALPN should be allowed to choose the best upstream protocol by using `AutoConfig` on the cluster level and the trailers should be enabled in the correct place there.


zirain (commented):
I would like to see a e2e test for new feature.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2500, opened 2024-01-25
api: Support Server header transformations

liorokman:

This PR adds support for defining what should be done with the Server header. 

The relevant Envoy configuration is [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-serverheadertransformation).

This PR additionally moves the `SuppressEnvoyHeaders` attribute into a `headers` container in the client traffic policy document.


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -82,6 +76,45 @@ type ClientTrafficPolicySpec struct {
...
+	ServerHeaderOverwrite ServerHeaderTransformation = "Overwrite"
+	// ServerHeaderAppendIfAbsent causes the ServerName value to be used only if there is no
+	// Server header
+	ServerHeaderAppendIfAbsent ServerHeaderTransformation = "AppendIfAbsent"
+	// ServerHeaderPassThrough will not modify the value of the Server header.
+	ServerHeaderPassThrough ServerHeaderTransformation = "PassThrough"
+)
+
+// HeaderSettings providess configuration options for headers on the listener.
+type HeaderSettings struct {
```
arkodg, line 101: we discussed this in the community meeting today, here are some follow up questions
* can this use case we solved with the ResponseHeaderFilter ?
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPHeaderFilter
* can the `suppressEnvoyHeaders` be represented with another name closer to the goal of the field such as `passThrough` etc ?

liorokman: > we discussed this in the community meeting today, here are some follow up questions
> 
> * can this use case we solved with the ResponseHeaderFilter ?
>   https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPHeaderFilter

No. The `ResponseHeaderFilter` unconditionally adds/overwrites a header whose value is given in the filter. This feature's `passThrough` option will keep the value of the Server header provided by the upstream, and there's no way to express the `appendIfAbsent` option with `ResponseHeaderFilter`.

> * can the `suppressEnvoyHeaders` be represented with another name closer to the goal of the field such as `passThrough` etc ?

You meant `serverHeaderTransformation`, not `suppressEnvoyHeaders`, right? `suppressEnvoyHeaders` has nothing to do with passing anything through and isn't the topic of this PR.

I used the name of the underlying feature in Envoy Proxy for the `serverHeaderTransformation` field. I think `passThrough` isn't a good name - it's only one of three options, and is exactly the opposite of the `overwrite` option.

 I'm open to any other name you have that you think fits better.

arkodg: from a user perspective, (correct me if im wrong), what you're really trying to solve here is - hey envoy, dont add any new response headers like `x-envoy-*` or `server`
we should come up with a name for this knob to define this opt in behavior

arkodg: how about a knob called `addEnvoyResponseHeaders`, making adding envoy headers along with `server` as opt in, so we supressby default

zirain: maybe https://github.com/istio/api/pull/2240 will give a little help.

arkodg: thanks for sharing this @zirain, this is super useful
how about removing this API and disabling
* x-envoy-*
* server
by default

and we can introduce APIs in the future to opt into this

zirain: fine to me

liorokman: So, to summarise the above:

1. The `HeaderSettings` structure is removed, and the default behaviour becomes that the `X-Envoy-` headers are removed and the Server header is defined as `passthrough`. This is done to try to effectively hide the presence of an Envoy Proxy between the client and the upstream target.
2. There is no support at this time for allowing the `X-Envoy` headers. This makes it harder for developers to perform certain tasks, for example when trying to measure latency with the [x-envoy-upstream-service-time](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-upstream-service-time), or debugging various rewrite rules with the [x-envoy-original-path](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_filters/router_filter#x-envoy-original-path) header.

How to proceed? Should I close this PR and create a new PR removing the `suppressEnvoyHeaders` option and implementing the behavioural changes listed here?

arkodg: hey @liorokman, agree with 1. 
for 2. we can make a opt in API e.g.
```
headers:
  enableEnvoyHeaders: true
``

Comment (liorokman) [reply api/v1alpha1/clienttrafficpolicy_types.go:105]: Updated to reflect this thread.

This PR changes the API and keeps the project building, a different PR will actually implement the feature.

File: api/v1alpha1/clienttrafficpolicy_types.go [hunk 2]
```diff
@@ -86,6 +80,33 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	HTTP1 *HTTP1Settings `json:"http1,omitempty"`
+	// HeaderSettings provides configuration for header management.
+	//
+	// +optional
+	Headers *HeaderSettings `json:"headers,omitempty"`
+}
+
+// ServerHeaderTransformation specifies the transformation required for the Server header
```
arkodg, line 89: this is not needed

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -387,10 +387,11 @@ func translateClientIPDetection(clientIPDetection *egv1a1.ClientIPDetectionSetti
 	httpIR.ClientIPDetection = (*ir.ClientIPDetectionSettings)(clientIPDetection)
 }
 
-func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *ir.HTTPListener) {
-	if suppressEnvoyHeaders != nil {
-		httpIR.SuppressEnvoyHeaders = *suppressEnvoyHeaders
+func translateListenerHeaderSettings(headerSettings *egv1a1.HeaderSettings, httpIR *ir.HTTPListener) {
+	if headerSettings == nil {
```
arkodg, line 391: this logic looks incorrect, `SuppressEnvoyHeaders` is enabled by default unless `headerSettings != nil && headerSettings.EnableEnvoyDebugHeader == true`
this is the one you plan on handling in a follow up PR ?

liorokman: Yes, exactly.

The implementation PR will handle the logic change, this PR is just to get the API in place while still being able to build the project.

arkodg: Sg

api/v1alpha1/clienttrafficpolicy_types.go
liorokman: Missed that. Removed.

arkodg (approved):
LGTM thanks !

shawnh2 (approved):

PR #2501, opened 2024-01-25
api: Support preserving header letter-case on HTTP/1

liorokman:

This PR defines the API for configuring Envoy to preserve the letter case of HTTP headers on HTTP/1.1 requests and responses.


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -93,6 +93,10 @@ type HTTP1Settings struct {
 	// EnableTrailers defines if HTTP/1 trailers should be proxied by Envoy.
 	// +optional
 	EnableTrailers *bool `json:"enableTrailers,omitempty"`
+	// PreserveHeaderCase defines if Envoy should preserve the letter case of headers.
+	// By default, Envoy will lowercase all the headers.
+	// +optional
+	PreserveHeaderCase *bool `json:"preserveHeaderCase,omitempty"`
```
zirain, line 99: I really don't encourage this API, because headers should not be case sensitive.

zirain: fine, let us knee to real world.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2504, opened 2024-01-26
Set status when EnvoyPatchPolicy is disabled

### LINKED ISSUE #2502: Add a status to EnvoyPatchPolicy when the feature is disabled
*Description*:
would be used.

The EnvoyPatchPolicy API only works when enabled in the startup `EnvoyGateway` config as highlighted here 
https://gateway.envoyproxy.io/latest/user/envoy-patch-policy/#enable-envoypatchpolicy

We've seen many users forget to enable this field and are confused by the API is not working.

One solution is to add a status condition (`Accepted: False`) with a reason mentioning that the feature is disabled and must be enabled in order for this resource to work as expected.

soulxu:
**What type of PR is this?**
* "fix: there is no status update when EnvoyPatchPolicy is disabled.

Update the status of EnvoyPatchPolicy when the patch policy feature is disabled.

Fixes #2502


internal/gatewayapi/envoypatchpolicy.go
```diff
@@ -94,6 +94,16 @@ func (t *Translator) ProcessEnvoyPatchPolicies(envoyPatchPolicies []*egv1a1.Envo
 			continue
 		}
 
+		if !t.EnvoyPatchPolicyEnabled {
+			status.SetEnvoyPatchPolicyCondition(policy,
+				gwv1a2.PolicyConditionAccepted,
+				metav1.ConditionFalse,
+				gwv1a2.PolicyReasonInvalid,
```
arkodg, line 101: im leaning more towards a `Disabled` Reason instead of `Invalid`

internal/gatewayapi/envoypatchpolicy.go
```diff
@@ -94,6 +94,16 @@ func (t *Translator) ProcessEnvoyPatchPolicies(envoyPatchPolicies []*egv1a1.Envo
...
+				"EnvoyPatchPolicy feature is disabled",
```
arkodg, line 102:
```suggestion
				"EnvoyPatchPolicy is disabled in the EnvoyGateway configuration",
```

arkodg:
looks good ! minor comment around status condition naming

soulxu:
> looks good ! minor comment around status condition naming

thanks! updated

arkodg (approved):
LGTM thanks !

zirain (approved):

zirain:
wait https://github.com/envoyproxy/gateway/pull/2514

PR #2506, opened 2024-01-26
feat: Implement preserving header letter cases

liorokman:
This PR implements preserving HTTP/1.1 header letter-case.

Implements #2501


zirain:
can you add an e2e test for this?

liorokman:
> can you add an e2e test for this?

That would be very hard, since the echo server used for the e2e test normalizes the header cases.

internal/gatewayapi/testdata/clienttrafficpolicy-preserve-case.in.yaml
```diff
@@ -0,0 +1,36 @@
+clientTrafficPolicies:
+- apiVersion: gateway.envoyproxy.io/v1alpha1
+  kind: ClientTrafficPolicy
+  metadata:
+    namespace: envoy-gateway
+    name: target-gateway-1
+  spec:
+    http1:
+      enableTrailers: true
```
arkodg, line 9: nit: not needed

internal/xds/translator/cluster.go
```diff
@@ -314,44 +317,51 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
-		// TODO: If the cluster is TLS enabled, use AutoHTTPConfig instead of ExplicitHttpConfig
-		// so that when ALPN is supported enabling trailers doesn't force HTTP/1.1
-		protocolOptions := httpv3.HttpProtocolOptions{
-			UpstreamProtocolOptions: &httpv3.HttpProtocolOptions_ExplicitHttpConfig_{
-				ExplicitHttpConfig: &httpv3.HttpProtocolOptions_ExplicitHttpConfig{
-					ProtocolConfig: &httpv3.HttpProtocolOptions_ExplicitHttpConfig_HttpProtocolOptions{
-						HttpProtocolOptions: &corev3.Http1ProtocolOptions{
-							EnableTrailers: http1Trailers,
-						},
+	} else if http1Opts.EnableTrailers || http1Opts.PreserveHeaderCase {
```
arkodg, line 333: shouldnt this be `else if http1Opts.EnableTrailers` ?

internal/xds/translator/listener.go
```diff
@@ -44,12 +45,24 @@ func http1ProtocolOptions(opts *ir.HTTP1Settings) *corev3.Http1ProtocolOptions {
 	if opts == nil {
 		return nil
 	}
-	if opts.EnableTrailers {
-		return &corev3.Http1ProtocolOptions{
-			EnableTrailers: opts.EnableTrailers,
+	if !opts.EnableTrailers && !opts.PreserveHeaderCase {
+		return nil
+	}
+	r := &corev3.Http1ProtocolOptions{
```
arkodg, line 51: trailers are enabled by default ?

internal/xds/translator/cluster.go
liorokman: No, the check for `PreserveHeaderCase` is needed.

Both these features need to be set in the `Http1ProtocolOptions` structure, so it needs to be filled if either of these features is enabled.

internal/xds/translator/listener.go
liorokman: No, they are enabled only if `opts.EnableTrailers == true`.

This function returns `nil` only if both `EnableTrailers` and `PreserveHeaderCase` are false. If `PreserveHeaderCase` is true and `EnableTrailers` is false then setting the `EnableTrailers` field to false is simply keeping it at its default value of "disabled".

internal/gatewayapi/testdata/clienttrafficpolicy-preserve-case.in.yaml
liorokman: You're right, it's not technically needed. It doesn't hurt though, so I left it in since I noticed that I didn't remove it only after I had already pushed the PR.

internal/ir/infra.go
```diff
@@ -76,7 +76,8 @@ type HTTP3Settings struct {
 // HTTP1Settings provides HTTP/1 configuration on the listener.
 // +k8s:deepcopy-gen=true
 type HTTP1Settings struct {
```
arkodg, line 78: missed this earlier, this struct should be living in `xds.go`

internal/xds/translator/listener.go
arkodg: yah thanks for explaining that, that logic should work

arkodg (approved):
minor nits, LGTM overall, thanks !

internal/xds/translator/listener.go
zirain: can you address these as comment in the code?

zirain:
> > can you add an e2e test for this?
> 
> That would be very hard, since the echo server used for the e2e test normalizes the header cases.

a nginx/tomcat base pod in https://github.com/envoyproxy/gateway/blob/7f360211df6f8cd8b4b783618cc694734cf2930a/test/e2e/base/manifests.yaml would help? 

it won't be a blocker, we can do that in following pr.

zirain (approved):

internal/xds/translator/listener.go
liorokman: :-) 
The PR was merged before I could add this as a comment.

zirain: you can open another one

liorokman:
> > > can you add an e2e test for this?

> it won't be a blocker, we can do that in following pr.

Added in PR #2516 .

internal/xds/translator/listener.go
liorokman: Added the comment in #2517

internal/ir/infra.go
liorokman: Moved to `xds.go` in #2517

PR #2512, opened 2024-01-26
bug: fix printable for ir xds

arkodg:


zirain:
can you share more details about this? I think related to security?

arkodg:
> can you share more details about this? I think related to security?

yes was added in https://github.com/envoyproxy/gateway/pull/607, but recently broke

zirain (approved):

zhaohuabing (approved):

PR #2513, opened 2024-01-26
api: support upstream max requests per connection

### LINKED ISSUE #2508: Support Max Requests Per Connection in Backend Traffic Policy
*Description*:

Envoy's [Common HTTP Protocol Options](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions) can be used to set a limit on the total number of requests that can be handled by a connection. Once the limit is reached, the connection is closed. Several envoy-based projects expose this setting for upstream connections. 

This parameter can be used for two purposes:
- effectively disabling upstream connection reuse
- coordinating persistent connection lifespan with the upstream backend. For example, nginx resets persistent connection after 100/1k requests by [default](https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_requests). If the proxy does not preemptively close the connection, a TCP race condition may occur, leading to request failures. 

*Relevant Links*:
Contour: https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1alpha1.ClusterParameters
Istio: https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings 
Gloo Edge: https://docs.solo.io/gloo-edge/1.7.23/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/connection.proto.sk/#connectionconfig

guydc:
Adding `MaxRequestsPerConnection` to `CircuitBreaker`, based on discussion in #2508. The implementation will translate the new attribute to the appropriate cluster's common http protocol options. 

Fixes #2508


api/v1alpha1/circuitbreaker_types.go
```diff
@@ -7,14 +7,24 @@ package v1alpha1
 
 // CircuitBreaker defines the Circuit Breaker configuration.
 type CircuitBreaker struct {
+	// Settings related to both TCP and HTTP upstream connections
```
arkodg, line 10:
```suggestion
	// Settings related to both TCP upstream connections
```

guydc: TCP settings like `MaxConnections` will apply to both TCP (e.g. TLS Passthrough) and HTTP connections. The HTTP struct contains settings that are relevant strictly to HTTP connections.

arkodg: in that case, the top level `tcp` and `http` doesnt seem very useful here and can be flattened into one, wdyt ?

guydc: no problem. I'll update the API PR accordingly.

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -30,4 +30,11 @@ type CircuitBreaker struct {
 	// +kubebuilder:default=1024
 	// +optional
 	MaxParallelRequests *int64 `json:"maxParallelRequests,omitempty"`
+
+	// The maximum number of requests that Envoy will make over a single connection to the referenced backend defined within a xRoute rule.
+	//
```
arkodg, line 35: can you highlight the default in the doc string ?

arkodg (approved):
LGTM thanks !

Issue #2495, opened 2024-01-24
Weighted Backend couldn't set to 0


ardikabs:
*Description*:
I had filed [issue](https://github.com/envoyproxy/gateway/issues/2170) related to weighted backends (e.g., canary purposes). It has been fixed, but it doesn't appear to solve the problem.

Given the scenario for canary purposes:
1. Backend A currently serving production traffic. Set the weight to `100`.
2. Preparing backend B for canary deployment. Set the weight to `0`.
3. Increase gradually backend B weight according to backend A weight. For example 100-0, 95-5, 90-10, 80-20, and so forth.

But it turned on point 2, setting the weight to 0 rejected by Envoy Gateway, as it seems that Envoy Gateway uses `load_balancing_weight` field in `endpoints` under [ClusterLoadAssigment](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint.proto#config-endpoint-v3-clusterloadassignment) which explicitly mentioned at least 1 is provided.
> load_balancing_weight
([UInt32Value](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#uint32value)) Optional: Per priority/region/zone/sub_zone weight; at least 1. The load balancing weight for a locality is divided by the sum of the weights of all localities at the same priority level to produce the effective percentage of traffic for the locality. The sum of the weights of all localities at the same priority level must not exceed uint32_t maximal value (4294967295).

From Envoy Gateway logs, you can see errors like below:

> invalid LocalityLbEndpoints.LoadBalancingWeight: value must be greater than or equal to 1

**Proposal**:

FWIW, it looks like for the weighted backend requirement would be better to utilize the [WeightedCluster](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-weighted-clusters) component as [pointed out](https://github.com/envoyproxy/gateway/issues/2170#issuecomment-1805167600) by @lemonlinger.
AFAIK, Istio did this for weighted backends. 

*Repro steps*:

1. Having 2 different backends on HTTPRoute.
2. One backend set the weight to 100, the other to 0.
3. You will see error logged by Envoy Gateway like above.
4. And the endpoint won't be available in Envoy.

*Environment*:

Latest

*Logs*:

Full error log:
```
[...]

PR #2515, opened 2024-01-27
bug: skip backendRefs with weight set to 0

arkodg:
* Envoy's `LocalityLbEndpoints.LoadBalancingWeight` only accepts weights greater than 0. So skip such backends/endpoints in the gateway api layer
* From the API spec, no other processing is required for such cases https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendRef
```
If weight is set to 0, no traffic should be forwarded for this entry.
```

Fixes https://github.com/envoyproxy/gateway/issues/2495

### DIFF
--- a/internal/gatewayapi/translator_test.go
+++ b/internal/gatewayapi/translator_test.go
@@ -62,7 +62,7 @@ func TestTranslate(t *testing.T) {
 			}
 
 			// Add common test fixtures
-			for i := 1; i <= 3; i++ {
+			for i := 1; i <= 4; i++ {
 				svcName := "service-" + strconv.Itoa(i)
 				epSliceName := "endpointslice-" + strconv.Itoa(i)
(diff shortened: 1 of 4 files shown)


internal/gatewayapi/translator_test.go
```diff
@@ -62,7 +62,7 @@ func TestTranslate(t *testing.T) {
 			}
 
 			// Add common test fixtures
-			for i := 1; i <= 3; i++ {
```
zirain, line 65: why not user retry package?

zirain (approved):

internal/gatewayapi/translator_test.go
arkodg: this func create service resources on the fly so we dont need to manually add them in the YAML

shawnh2 (approved):

PR #2516, opened 2024-01-27
e2e: Add an e2e test for the header case-preserving feature

liorokman:
This PR adds an e2e test for the header case preservation feature implemented in #2506

### DIFF
--- a/test/e2e/base/manifests.yaml
+++ b/test/e2e/base/manifests.yaml
@@ -521,3 +521,111 @@ spec:
     allowedRoutes:
       kinds:
       - kind: UDPRoute
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gateway-preserve-case-backend
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: go-server
+  namespace: gateway-preserve-case-backend
+data:
+  go.mod: |
+         module srvr
+         go 1.21.3
+         require (
+         	github.com/andybalholm/brotli v1.0.5 // indirect
+         	github.com/klauspost/compress v1.17.0 // indirect
+         	github.com/valyala/bytebufferpool v1.0.0 // indirect
+         	github.com/valyala/fasthttp v1.51.0 // indirect
+         )
+  go.sum: |
+         github.com/andybalholm/brotli v1.0.5 h1:8uQZIdzKmjc/iuPu7O2ioW48L81FgatrcpfFmiq/cCs=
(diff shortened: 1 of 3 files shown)


test/e2e/tests/preservecase.go
```diff
@@ -0,0 +1,143 @@
...
+		fmt.Printf("Sending Request:\n%s\n\n", formatDump(dump, "< "))
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if true {
```
zirain, line 82: nit?

test/e2e/base/manifests.yaml
```diff
@@ -521,3 +521,110 @@ spec:
...
+  selector:
+    matchLabels:
+      app: golang-app
+  template:
+    metadata:
+      labels:
+        app: golang-app
+    spec:
+      containers:
+      - name: golang-app-container
```
zirain, line 603: this's a little trick, but it's clear to maintain the source code.

@arkodg WDYT?

test/e2e/tests/preservecase.go
liorokman: Sorry about that. Fixed.

test/e2e/base/manifests.yaml
arkodg: case preservation isn't possible with echo server today ?

liorokman: The echo server uses `net/http` which automatically and unconditionally normalizes headers with [CanonicalHeaderKey](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/http/request.go;l=165).

The minimal server I added uses [fasthttp](https://github.com/valyala/fasthttp) which supports not normalizing the headers.

The e2e test still isn't complete because while the server I added returns a header with non-standard casing, the `net/http` based client used in the test suite normalizes headers as well, so short of replacing the client side as well, it's not possible to add a check that returned headers' case is also preserved. (I did verify this manually using `curl` though.)

zhaohuabing: @liorokman  This is brilliant :-) I think Ext Auth e2e test can aslo use it.

arkodg: cool thanks for explaining the current limitations @liorokman 
lets go ahead with this approach since @zirain and @zhaohuabing are +1s on this

zirain (approved):

liorokman:
@envoyproxy/gateway-maintainers please merge

arkodg:
ptal @zhaohuabing

zhaohuabing (approved):
LGTM. Thanks!

PR #2522, opened 2024-01-29
api: making the value optional for JSONPatchOperation

### LINKED ISSUE #2456: Propose to add PrivateKeyProvider support in EnvoyGateway
The Envoy Private Key Provider (https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-privatekeyprovider) allows implementing different provider which can provide the private key operation acceleration through the hardware or private key protection through the hardware.
 
Today, there are two private key providers were implemented as Envoy contrib extensions:
* CryptoMB in Envoy 1.20 release (https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/private_key_providers/cryptomb/v3alpha/cryptomb.proto )
* QAT in Envoy 1.24 release ([ https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/private_key_providers/qat/v3alpha/qat.proto#extensions-private-key-providers-qat-v3alpha-qatprivatekeymethodconfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/private_key_providers/qat/v3alpha/qat.proto#extensions-private-key-providers-qat-v3alpha-qatprivatekeymethodconfig) )

This proposal aims to describe how to enable private key provider configuration in the EnvoyGateway through dynamic configuration and static configuration. Also, consider the aspect of envoy image and k8s resource scheduling.

Design doc: https://docs.google.com/document/d/1oDLsNK6olPOaeMQPxFfG04HTNaYO0qp1u7YV4OmyNBA/edit?usp=sharing

soulxu:
**What type of PR is this?**
* fix: making the move operation is usable.

The `move`, `remove` operation in the Json patch needn't the value parameter, so making that is optional in the Envoy Gateway API.

Part of fixes #2456


api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -118,7 +118,8 @@ type JSONPatchOperation struct {
 	// +optional
 	From *string `json:"from,omitempty"`
 	// Value is the new value of the path location.
-	Value apiextensionsv1.JSON `json:"value"`
+	// +optional
+	Value apiextensionsv1.JSON `json:"value,omitempty"`
```
zhaohuabing, line 122: 1. Optional field should be pointer
2. It would be helpful to explain when the `Value` should be nil and when not
3. Could we add a CEL validation for when this should be nil or not?

soulxu: > 1. Optional field should be pointer
got it, will do.

> 2. It would be helpful to explain when the `Value` should be nil and when not
`Value` only required by add and replace operation, will update the doc

> 3. Could we add a CEL validation for when this should be nil or not?
Yes, it should be do able, let me look at it.

Thanks for the review!

soulxu: > 3. Could we add a CEL validation for when this should be nil or not?

emm..., I got a hard time with this. I give try on this:

```
// +kubebuilder:validation:XValidation:rule=" !has(self.value) || (self.op != 'add' && self.op != 'replace') ",message="The Value field is required by add and replace operation"

```

It always shows that the value field is undefined.

In the end, I found this:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md#type-system-integration

'object' with x-kubernetes-preserve-unknown-fields: object / "message type", unknown fields are NOT accessible in CEL expression

arkodg: +1 to adding CEL as well as in the PR

soulxu: @arkodg @zhaohuabing it seems like CEL doesn't work here due to the `value` is an unstruct message, is it ok to do the validation in the go code?

arkodg: can we check if field is not `nil` if `op` is nor `move` ?

arkodg: if not possible, lets add a validation check https://github.com/envoyproxy/gateway/blob/241e83839acba4349c746acfd797869f40e37aca/internal/xds/translator/jsonpatch.go#L53 here ?

soulxu: > can we check if field is not `nil` if `op` is not `move` ?

we can't, the CEL doesn't know the field of `value`.

As you suggested, the go code validation is probably the right place

soulxu:
@zhaohuabing @arkodg sorry for disappearing a few days, I updated the code, would you like to review it again? thanks in advance!

api/v1alpha1/envoypatchpolicy_types.go
```diff
@@ -117,8 +117,10 @@ type JSONPatchOperation struct {
 	// Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
 	// +optional
 	From *string `json:"from,omitempty"`
-	// Value is the new value of the path location.
-	Value apiextensionsv1.JSON `json:"value"`
+	// Value is the new value of the path location. The value is only used by
+	// the operations of `add` and `replace`.
```
arkodg, line 121:
```suggestion
	// the `add` and `replace` operations.
```

internal/xds/translator/jsonpatch.go
```diff
@@ -51,6 +55,19 @@ func processJSONPatches(tCtx *types.ResourceVersionTable, envoyPatchPolicies []*
 				err          error
 			)
 
+			switch p.Operation.Op {
+			case AddOperation, ReplaceOperation:
+				if p.Operation.Value == nil {
+					status.SetEnvoyPatchPolicyInvalid(e.Status, "The add/replace operation required a value")
```
arkodg, line 61: lets add a Sprintf here to specify exact operation - `%s operation requires a value`

internal/xds/translator/jsonpatch.go
```diff
@@ -51,6 +55,19 @@ func processJSONPatches(tCtx *types.ResourceVersionTable, envoyPatchPolicies []*
...
+					continue
+				}
+			default:
+				if p.Operation.Value != nil {
+					status.SetEnvoyPatchPolicyInvalid(e.Status, "The remove/replace/copy/move operation needn't the value field specified")
```
arkodg, line 66: as as above ` value cannot be set for the %s operation`

internal/xds/translator/translator_test.go
```diff
@@ -177,6 +177,14 @@ func TestTranslateXds(t *testing.T) {
 			name:                      "jsonpatch-invalid-patch",
 			requireEnvoyPatchPolicies: true,
 		},
+		{
+			name:                      "jsonpatch-add-op-without-value",
```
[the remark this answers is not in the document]
soulxu: In the file of internal/xds/translator/testdata/out/xds-ir/jsonpatch-add-op-without-value.envoypatchpolicies.yaml

we can validate the status message, so we combine them together the error status probably will be override.

```
- name: first-policy
  namespace: default
  status:
    conditions:
    - lastTransitionTime: null
      message: The add operation required a value
      reason: Invalid
      status: "False"
      type: Programmed
```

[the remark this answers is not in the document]
arkodg: Oh good point

internal/xds/translator/jsonpatch.go
```diff
@@ -51,6 +55,21 @@ func processJSONPatches(tCtx *types.ResourceVersionTable, envoyPatchPolicies []*
 				err          error
 			)
 
+			switch p.Operation.Op {
+			case AddOperation, ReplaceOperation:
+				if p.Operation.Value == nil {
+					msg := fmt.Sprintf("The %s operation required a value", p.Operation.Op)
```
arkodg, line 61:
```suggestion
					msg := fmt.Sprintf("The %s operation requires a value", p.Operation.Op)
```

arkodg (approved):
LGTM thanks!

zhaohuabing (approved):

PR #2534, opened 2024-01-29
api: Add support for enabling HTTP/1.0 and HTTP/0.9

liorokman:

This PR adds the API for enabling support for HTTP/1.0 and HTTP/0.9 from downstream clients


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -97,6 +98,11 @@ type HTTP1Settings struct {
 	// By default, Envoy will lowercase all the headers.
 	// +optional
 	PreserveHeaderCase *bool `json:"preserveHeaderCase,omitempty"`
+	// EnableHTTP10 turns on support for HTTP/1.0 and HTTP/0.9 requests.
+	EnableHTTP10 *bool `json:"enableHttp10,omitempty"`
```
arkodg, line 102: I'd prefer if this was
```
htp10: {}
```

so in the future we could opt into injecting a `Host` header, but id prefer if this was a `bool` as well as the value was derived from the listener's hostname

liorokman: I'm not sure I follow your intent here. 

The Envoy Proxy documentation says that Envoy Proxy doesn't support HTTP/1.0 requests correctly iff these arrive without a Host header. That's why configuring a default host is recommended.

> accept_http_10
([bool](https://developers.google.com/protocol-buffers/docs/proto#scalar)) Handle incoming HTTP/1.0 and HTTP 0.9 requests. This is off by default, and not fully standards compliant. There is support for pre-HTTP/1.1 style connect logic, dechunking, and handling lack of client host iff default_host_for_http_10 is configured.
>
> default_host_for_http_10
([string](https://developers.google.com/protocol-buffers/docs/proto#scalar)) A default host for HTTP/1.0 requests. This is highly suggested if accept_http_10 is true as Envoy does not otherwise support HTTP/1.0 without a Host header. This is a no-op if accept_http_10 is not true.

According to the [Listener](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.Listener) definition in the `Gateway` resource, matching with a hostname is optional. Deriving the hostname from the listener is not guaranteed to be possible, because providing a hostname is not mandatory.

Would you prefer if the API was:

```yaml
spec:
   http1:
      enableHttp10: host-name-to-use-as-a-default
```

zhaohuabing: +1 It would be harder for debugging if placing hostnames in two different places.

liorokman: Updated the proposed API.

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -97,6 +97,20 @@ type HTTP1Settings struct {
...
+
+// HTTP10Settings provides HTTP/1.0 configuration on the listener.
+type HTTP10Settings struct {
+	// SetHostHeader defines if the HTTP/1.0 request is is missing the Host header,
+	// then the hostname associated with the listener should be injected into the
+	// request.
+	// If this is not set and an HTTP/1.0 request arrives without a host, then
+	// it will be rejected.
+	// +optional
+	SetHostHeder *bool `json:"setHostHeader,omitempty"`
```
arkodg, line 113: I know I proposed `set` earlier, would `addHostHeader` be a better name ?

liorokman: Since the host header is only added if it's missing, how about `defaultHost`?  That's also closer to what the Envoy Proxy configuration key is called.

arkodg: @envoyproxy/gateway-maintainers any preferences here ?

zirain: +1 for `useDefaultHost`

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -97,6 +97,20 @@ type HTTP1Settings struct {
...
 	// +optional
 	PreserveHeaderCase *bool `json:"preserveHeaderCase,omitempty"`
+	// HTTP10 turns on support for HTTP/1.0 and HTTP/0.9 requests.
+	// +optional
+	HTTP10 *HTTP10Settings `json:"http10,omitempty"`
+}
+
+// HTTP10Settings provides HTTP/1.0 configuration on the listener.
+type HTTP10Settings struct {
+	// SetHostHeader defines if the HTTP/1.0 request is is missing the Host header,
```
zhaohuabing, line 107:
```suggestion
	// SetHostHeader defines if the HTTP/1.0 request is is missing the Host header,
	// SetHostHeader defines if the HTTP/1.0 request is missing the Host header,
```

zhaohuabing: Nit :-)

api/v1alpha1/clienttrafficpolicy_types.go
arkodg: lets go with `useDefaultHost` unless @zhaohuabing disagrees :)

zhaohuabing: If `addHostHeaderWhenMissing` is too long, then `useDefaultHost`. :-)

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM. Thanks!

Issue #2252, opened 2023-11-29
preserving the client IP address


tmsnan:
*Description*:
>
Currently, the gateway supports two methods for obtaining the client IP address (proxy protocol  in the four-layer and x-forwarded-for in the seven-layer). But, I think some additional details need to be added.
1. proxy protocol
The Proxy protocol to enable Layer 4 listeners to retrieve client IP addresses. But currently gateway only supports http listener, we need to expand to support non-http/https scenarios.
#2203

2. x-forwarded-for
`x-forwarded-for` is a layer-7 protocol method for saving client IP addresses. The current gateway also utilizes it for current limiting based on the source IP. But,  source IP-based rate limiting in the gateway only supports direct connections to the gateway. To accommodate scenarios involving indirect connections to the gateway and obtain the correct client IP address, it may be necessary to introduce the [xff_num_trusted_hops](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for) field. In multi-hop scenarios, it is important to note that the XFF header may be susceptible to forgery.
![client_ip drawio](https://github.com/envoyproxy/gateway/assets/12018633/3dfa2e9f-d604-4fe7-9f6d-524665548ad3)

[optional *Relevant Links*:]
>
#946 
https://github.com/envoyproxy/gateway/blob/0add22b8c58111606243e688611c63b96784cb75/api/v1alpha1/clienttrafficpolicy_types.go#L55C1-L58C15
https://github.com/envoyproxy/gateway/issues/1328

robd003:
Is there a way to restrict X-Forwarded-For to only allow it from trusted CIDR blocks?

tmsnan:
@robd003, Do you want to restrict based on remote IP? rbac is being implemented.https://github.com/envoyproxy/gateway/issues/2250.

You can use the `source ip` field. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#config-rbac-v3-principal

robd003:
@tmsnan I'm not trying to block end-user IPs from accessing the proxy.

This is more about trusting only certain CIDR blocks to give valid X-Forwarded-For headers. It's very easy for someone to spoof this data, but if you limit to trusting the X-Forwarded-For to only certain CIDR blocks then you can be certain that end user IPs are correct and not forged.

zirain:
sounds like a WAF?

robd003:
[...]

Issue #2250, opened 2023-11-29
Support fined-grained access control in SecurityPolicy 


zhaohuabing:
## Relates to
https://github.com/envoyproxy/gateway/issues/1845

## What is this?
EG can leverage [RBAC](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/rbac) for implementing fine-grained access control, at both the `Gateway` and `xRoute` level.

The principal, obtained through the authentication process (such as OIDC, JWT, etc.), serves as the basis for defining access control policies.  Source IP-based access control can also be supported in this model.

The below is roughly how API will look like, but it's just an initial idea and definitely needs more input.
## API outline

```yaml
kind: SecurityPolicy
metadata:
  name: rbac-example
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: http-route
  jwt:   # a jwt or oidc configuration is needed to obtain the principal
  ......  
  authorization:
  - principals: ["john"]
    permissions:
      methods: ["GET", "POST"]
      paths: ["/foo", "/bar"]

```
Reference:https://github.com/envoyproxy/envoy/issues/7913

- [x] [Allow/Deny IP Subnets](https://github.com/envoyproxy/gateway/issues/2462)
- [x] [Authorization based on JWT claim ](https://github.com/envoyproxy/gateway/issues/3367)
- [ ] [Authorization based on Basic Auth Username](https://github.com/envoyproxy/gateway/issues/3624) https://github.com/envoyproxy/gateway/issues/2947
- [ ] Authorization based on client cert
- [ ] Authorization based on identity extracted from an arbitrary HTTP header

arkodg:
@zhaohuabing 
* can't the `paths` and `methods` be limited by the `HTTPRoute` match fields ?
* can't the `principal` be limited by the `issuer` field in JWT https://gateway.envoyproxy.io/v0.6.0/api/extension_types/#jwtprovider ?

zhaohuabing:
> @zhaohuabing
> 
> * can't the `paths` and `methods` be limited by the `HTTPRoute` match fields ?

I would suggest not to go this route because:
* It could get tricky, especially when the policy targets a gateway.  
* Separate of concerns:  `SecurityPolicy`  and `HTTPRoute` may be created by different roles, such as app developers and administrators

[...]

PR #2535, opened 2024-01-29
feat: Support Client IP Detection using XFF on ClientTrafficPolicy

davidalger:
**What type of PR is this?**

feat: Support Client IP Detection using XFF on ClientTrafficPolicy

Implements client IP detection support on `ClientTrafficPolicy` using `xff_num_trusted_hops` allowing listeners to be configured to trust XFF headers such as `x-forwarded-for` and `x-forwarded-proto`

Related to #2252


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -66,6 +66,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	EnableProxyProtocol *bool `json:"enableProxyProtocol,omitempty"`
+	// HTTPConnectionManager provides HTTP Connection Manager configuration on the listener.
```
arkodg, line 69: this relates to https://github.com/envoyproxy/gateway/pull/2500/files

what might make sense here is a config like
```
headers:
  xForwardedFor:
    numTrustedHops: ...
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -84,6 +88,25 @@ type ClientTrafficPolicySpec struct {
...
+// HTTPConnectionManagerSettings provides HTTP Connection Manager configuration on the listener.
+type HTTPConnectionManagerSettings struct {
+	// UseRemoteAddress controls whether the connection manager will use the real remote address of the
+	// client connection when determining internal versus external origin and manipulating various headers.
+	// Refer to https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for
+	// for more details. Enabled by default.
+	//
+	// +optional
+	// +kubebuilder:default=true
+	UseRemoteAddress *bool `json:"useRemoteAddress,omitempty"`
```
arkodg, line 100: lets introduce this field when there is a valid use case

site/content/en/latest/user/client-traffic-policy.md
```diff
@@ -285,5 +285,62 @@ You should now expect 200 response status and also see that source IP was preser
...
+http-xff-num-trusted-hops   Accepted   5s
+```
+
+Open port-forward to the admin interface port:
+
+```shell
+kubectl port-forward deploy/${ENVOY_DEPLOYMENT} -n envoy-gateway-system 19000:19000
+```
+
+Curl the admin interface port to fetch the configured value for `xff_num_trusted_hops`:
```
arkodg, line 329: would be great to see an example using `curl` where the request being sent already has a `x-forwarded-for` and the value of `num_trusted_hops` is impacting the `x-envoy-external-address` header

arkodg:
thanks for building this out @davidalger . Here's how I'm hoping we can make progress in this PR
* get some consensus that `ClientTrafficPolicy` is the right home for this field vs in `EnvoyProxy`
Youve already highlighted this - since `EnvoyProxy` attaches to a `GatewayClass`, it can affect multiple `Gateway`s and corresponding Envoy Proxy fleets, so introducing the field in `ClientTrafficPolicy` does make sense, im a +1 for this
* I'd prefer if we created a top level field called `headers` to hold all fields that affect proxy header transformations. Once we have consensus on this, we'd need to make that change first @liorokman 
* I'd prefer if we skipped introducing the `use_remote_address` field in this PR

liorokman:
> * I'd prefer if we created a top level field called `headers` to hold all fields that affect proxy header transformations. Once we have consensus on this, we'd need to make that change first @liorokman

I'm +1 for that. In fact, I've already proposed this in #2500 .

internal/xds/translator/listener.go
```diff
@@ -146,6 +146,16 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	} else {
 		statPrefix = "http"
 	}
+
+	// HTTP XFF configuration
+	var useRemoteAddress bool
+	if irListener.UseRemoteAddress != nil {
```
liorokman, line 152: Lines 151-157 can also be written as:

```golang
useRemoteAddres := ptr.Deref(irListener.UseRemoteAddress, true)
```

davidalger:
> I'd prefer if we created a top level field called headers to hold all fields that affect proxy header transformations. Once we have consensus on this, we'd need to make that change first @liorokman

I like the idea of organizing things better, but not 100% sure `headers` is the right place to put it since XFF isn't just about header transformation and falls more into general IP detection territory.

@arkodg Since HCM supports either a combination of `use_remote_address` and `xff_num_trusted_hops` or using `original_ip_detection_extensions` for this, what are your thoughts on the following structure?

```yaml
originalIpDetection:
  useRemoteAddress: true    # True by default, false if `extensions` is specified (not present in CRD, as suggested)
  xffNumTrustedHops: number # Mutually exclusive with `extensions` configures `xff_num_trusted_hops` built-in
  extensions:
    # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.custom_header.v3.CustomHeaderConfig
    customHeaders:
      headerName: string
      rejectWithStatus: enum
    # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.xff.v3.XffConfig
    xff:
      numTrustedHops: number
```

Supporting config for the custom headers IP detection extension will be of use to me in deployments of EG, so even if precluded from this MR for sake of keeping it single-purpose I would like to consider planning for it and can introduce it in a future change set.

site/content/en/latest/user/client-traffic-policy.md
davidalger: I will add this

internal/xds/translator/listener.go
davidalger: Good point 🤦🏻

arkodg:
so we have two options here, (please ignore the exact names of the fields)

* add this config under a newly created `headers` field
```
headers:
  xForwardedFor:
    numTrustedHops:
```

* add this config under a dedicated field around IP detection / network topology
```
ipDetection:
  numTrustedHops
```

will bring this up in the community meeting this week

arkodg:
discussed this in the community meeting today and folks prefer 2 over 1 (generic `headers` section)

* brainstorming some names - what should the top level field be - `remoteIP` or `originalIP` or `remoteIPDetection` or `originalIPDetection` or `originatingIP....`

* sidenote (not related to this PR) - should proxy protocol config be moved under this field ? cc @envoyproxy/gateway-maintainers

davidalger:
> brainstorming some names - what should the top level field be - remoteIP or originalIP or remoteIPDetection or originalIPDetection or originatingIP.... or clientIP..... Whichever term we decide will need to be used in other APIs (https://github.com/envoyproxy/gateway/issues/2250 and [RateLimit](https://gateway.envoyproxy.io/v0.6.0/api/extension_types/#ratelimitselectcondition))

`clientIPDetection` makes sense to me given the term 'client IP' lines up with what's used used to document `sourceCIDR` on RateLimit already and also has widespread use across the broader industry (Google LBs, CloudFlare, Nginx, etc) and it's meaning should be widely understood

arkodg:
> > brainstorming some names - what should the top level field be - remoteIP or originalIP or remoteIPDetection or originalIPDetection or originatingIP.... or clientIP..... Whichever term we decide will need to be used in other APIs (#2250 and [RateLimit](https://gateway.envoyproxy.io/v0.6.0/api/extension_types/#ratelimitselectcondition))
> 
> `clientIPDetection` makes sense to me given the term 'client IP' lines up with what's used used to document `sourceCIDR` on RateLimit already and also has widespread use across the broader industry (Google LBs, CloudFlare, Nginx, etc) and it's meaning should be widely understood

sg @davidalger, lets go with `clientIpDetection` !

davidalger:
@arkodg Implemented first pass here with the following structure:

```yaml
clientIPDetection:
  xffNumTrustedHops: number # Mutually exclusive with `extensions` configures `xff_num_trusted_hops` HCM built-in
  extensions:
    # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.custom_header.v3.CustomHeaderConfig
    customHeader:
      headerName: string
      rejectWithStatus: number
    # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.xff.v3.XffConfig
    xff:
      numTrustedHops: number
```

`use_remote_address` maintains it's default of `true` while being set to `false` if original IP detection extensions are used since the two are mutually exclusive

Once https://github.com/envoyproxy/envoy/pull/31831 is merged, I would see there being an additional field to control the trusted CIDRs:

```yaml
clientIPDetection:
  extensions:
    xff:
      trustedCIDRs:
      - "130.211.0.0/22"
      - "35.191.0.0/16"
```

Would it be better to keep things a little more flat given this won't support arbitrary extensions?

```yaml
clientIPDetection:
    xffNumTrustedHops: 2
    customHeaderExtension:
      // ...
    xffExtension:
      // ...
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -66,6 +66,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	EnableProxyProtocol *bool `json:"enableProxyProtocol,omitempty"`
+	// ClientIPDetectionSettings provides XFF and extension configuration for client IP detection on the listener.
+	//
+	// +optional
+	ClientIPDetection *ClientIPDetectionSettings `json:"clientIPDetection,omitempty"`
```
arkodg, line 72:
```suggestion
	ClientIPDetection *ClientIPDetectionSettings `json:"clientIpDetection,omitempty"`
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -84,6 +88,73 @@ type ClientTrafficPolicySpec struct {
 	HTTP1 *HTTP1Settings `json:"http1,omitempty"`
 }
 
+// ClientIPDetectionSettings provides XFF and extension configuration for client IP detection on the listener.
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.xffNumTrustedHops) && has(self.extensions))",message="extensions cannot be used in conjunction with xffNumTrustedHops"
+type ClientIPDetectionSettings struct {
```
arkodg, line 94: prefer if we make incremental changes, and this PR includes only the ability to set `numTrustedHops` to be able to retrieve the correct client IP from XFF.
```
clientIPDetection:
  xForwardedFor:
    numTrustedHops: <>
```

arkodg:
@davidalger curious if you know the difference between the numTrustedHops field in the top level vs extension setting in envoy ?

davidalger:
> curious if you know the difference between the numTrustedHops field in the top level vs extension setting in envoy?

At the moment the only difference I believe is the extensions were marked incompatible with `use_remote_address`, but otherwise the XFF extension is identical in function currently.

It's been asked by maintainers to implement additions to XFF handling in `original_ip_detection_extensions` and in the coming days it's looking like `sourceCIDRs` support will be added to the XFF extension (https://github.com/envoyproxy/envoy/pull/31831) and notably it will evaluate the `x-forwarded-for` header either by a fixed number of trusted hops, or by evaluating the client IP address against a list of trusted addresses.

So something that could be considered is supporting `numTrustedHops` at the top-level and `sourceCIDRs` (potential future feature) would use the extension but without any plans to support the seemingly duplicate `numTrustedHops` field. 

Something like this I imagine:

```yaml
clientIPDetection:
    xForwardedFor:
      # Top-level xff_num_trusted_hops field with no support for same field in XFF extension
      numTrustedHops: 2

      # Future ability to skip appending to xff header
      skipAppend: true

      # Future ability to evaluate xff header based on trusted CIDRs rather than number of hops (uses extension)
      sourceCIDRs:
      - "130.211.0.0/22"
      - "35.191.0.0/16"

    # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.custom_header.v3.CustomHeaderConfig
    customHeader:
      headerName: x-client-ip-address
      rejectWithStatus: 403
```

Anything other than `numTrustedHops` would result in `use_remote_address` being set to `false` due to incompatibility with extensions, otherwise the current default of `true` would apply.

I'll update this MR to only add the ability to set `numTrustedHops` and tackle `customHeader` separately after this is complete.

arkodg:
> > curious if you know the difference between the numTrustedHops field in the top level vs extension setting in envoy?
> 
> At the moment the only difference I believe is the extensions were marked incompatible with `use_remote_address`, but otherwise the XFF extension is identical in function currently.
> 
> It's been asked by maintainers to implement additions to XFF handling in `original_ip_detection_extensions` and in the coming days it's looking like `sourceCIDRs` support will be added to the XFF extension ([envoyproxy/envoy#31831](https://github.com/envoyproxy/envoy/pull/31831)) and notably it will evaluate the `x-forwarded-for` header either by a fixed number of trusted hops, or by evaluating the client IP address against a list of trusted addresses.
> 
> So something that could be considered is supporting `numTrustedHops` at the top-level and `sourceCIDRs` (potential future feature) would use the extension but without any plans to support the seemingly duplicate `numTrustedHops` field.
> 
> Something like this I imagine:
> 
> ```yaml
> clientIPDetection:
>     xForwardedFor:
>       # Top-level xff_num_trusted_hops field with no support for same field in XFF extension
>       numTrustedHops: 2
> 
>       # Future ability to skip appending to xff header
>       skipAppend: true
> 
>       # Future ability to evaluate xff header based on trusted CIDRs rather than number of hops (uses extension)
>       sourceCIDRs:
>       - "130.211.0.0/22"
>       - "35.191.0.0/16"
> 
>     # Configures type.googleapis.com/envoy.extensions.http.original_ip_detection.custom_header.v3.CustomHeaderConfig
>     customHeader:
>       headerName: x-client-ip-address
>       rejectWithStatus: 403
> ```
> 
> Anything other than `numTrustedHops` would result in `use_remote_address` being set to `false` due to incompatibility with extensions, otherwise the current default of `true` would apply.
> 
> I'll update this MR to only add the ability to set `numTrustedHops` and tackle `customHeader` separately after this is complete.

sounds like a good plan ! this keeps any envoyisms out of the API increasing its ability to be moved into the upstream gateway api in the future

davidalger:
@arkodg updated and should be ready for review

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -84,6 +88,21 @@ type ClientTrafficPolicySpec struct {
 	HTTP1 *HTTP1Settings `json:"http1,omitempty"`
 }
 
+// ClientIPDetectionSettings provides configuration for determining the original client IP address for requests.
+type ClientIPDetectionSettings struct {
+	// XForwardedForSettings provides configuration for using X-Forwarded-For headers for determining the client IP address.
+	XForwardedFor XForwardedForSettings `json:"xForwardedFor,omitempty"`
```
arkodg, line 94: should be a pointer, since this field can be optional, in case you choose to include something like `customHeaders` in the future

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -84,6 +88,21 @@ type ClientTrafficPolicySpec struct {
...
+}
+
+// XForwardedForSettings provides configuration for using X-Forwarded-For headers for determining the client IP address.
+type XForwardedForSettings struct {
+	// NumTrustedHops controls the number of additional ingress proxy hops from the right side of XFF HTTP
+	// headers to trust when determining the origin client's IP address.
+	// Refer to https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for
+	// for more details.
+	NumTrustedHops uint32 `json:"numTrustedHops"`
```
arkodg, line 103: should be a pointer, since its optional
needs a optional tag as well

arkodg, line 103:
```suggestion
	NumTrustedHops *uint32 `json:"numTrustedHops,omitempty"`
```

arkodg:
PR looks good @davidalger, added a minor comment around optional fields

api/v1alpha1/clienttrafficpolicy_types.go
davidalger: updated to be optional

api/v1alpha1/clienttrafficpolicy_types.go
davidalger: updated to be optional

arkodg (approved):
LGTM thanks !

zirain (approved):

zirain:
need a user doc for this?

arkodg:
> need a user doc for this?

was part of this PR https://gateway.envoyproxy.io/latest/user/client-traffic-policy/#configure-client-ip-detection

PR #2537, opened 2024-01-30
ext auth impl

zhaohuabing:
This PR does the following:
* External Authorization implementation
* Add Indexer and Watch for `Services` and `EndpointSlices` referenced by the External Authorization configuration of `SecurityPolicy` 
* Clean up the code
  * Move all the Indexers to a separate file to shrink the huge `controller.go`
  * Unify the request timeout for external services(JWT, oauth2, ext auth)
  * Add SNI for UpstreamTLSSocket

The BackendObjectReference validation and Cluster creation could be optimized and combined into those of the xRoutes.  
I prefer to improve it in a follow-up PR because this PR is already big enough.

Docs and E2E tests will aslo be in separate PRs.

Related: https://github.com/envoyproxy/gateway/issues/1059


internal/xds/translator/httpfilters.go
```diff
@@ -91,18 +91,20 @@ func newOrderedHTTPFilter(filter *hcmv3.HttpFilter) *OrderedHTTPFilter {
...
-	case filter.Name == basicAuthFilter:
+	case isFilterType(filter, basicAuthFilter):
 		order = 2
-	case isOAuth2Filter(filter):
+	case isFilterType(filter, oauth2Filter):
 		order = 3
 	case filter.Name == jwtAuthn:
 		order = 4
-	case filter.Name == wellknown.Fault:
+	case isFilterType(filter, extAuthFilter):
```
arkodg, line 100: should this be before `jwt` ?

internal/xds/translator/extauth.go
```diff
@@ -0,0 +1,339 @@
...
+				continue
+			}
+		}
+	}
+
+	return errs
+}
+
+// createExtAuthCluster creates the cluster for the external auth service.
+func createExtAuthCluster(tCtx *types.ResourceVersionTable,
```
guydc, line 238: I assume that multiple security policies may reference the same auth-server `host:port` combination. It seems that a cluster is identified by name.  Does this mean that `ErrXdsClusterExists` can be triggered very often? If multiple policies define different cluster settings (e.g. transport socket setting) for the same `host:port` auth server, only the first policy's settings will apply in practice?

zhaohuabing: @guydc Good catch!  I believe different policies shouldn't have different TLS settings for the same host+port combination?

zhaohuabing: @guydc Good catch!  I believe different policies shouldn't have different TLS settings for the same host+port combination?

internal/xds/translator/httpfilters.go
zhaohuabing: @arkodg you mean `jwtAuthn` should be `jwt`?

I just followed the Envoy filter name `envoy.filters.http.jwt_authn`

zhaohuabing: @arkodg Is that because ext auth may add jwt token in the original headers?

arkodg: Yah

internal/xds/translator/extauth.go
guydc: Yes, I totally agree that cluster deduplication is a good idea. And you're right that for TLS, I would expect that the same target port would be either tls-enabled or not. 

Still, consider the following situation: a TLS-enabled ext-auth server is referenced by policies A and B. Policy B is misconfigured (by mistake) by the end user to not use TLS. If Policy B is the first to be translated, it would also break auth-server integration for all routes that policy A is attached to (even though policy A is ok). 

In some cases, there could even be a legitimate reason to have different cluster settings for the same host:port. For example, if a specific host:port supports both HTTP and GRPC auth interfaces, policies may need to reference the same host:port with different cluster HTTP settings (e.g. enabling HTTP2 for grpc).

internal/xds/translator/extauth.go
```diff
@@ -0,0 +1,339 @@
...
+
+func httpService(http *ir.HTTPExtAuthService) *extauthv3.HttpService {
+	var (
+		enableTLS        = http.TLS != nil
+		uri              string
+		headersToBackend []*matcherv3.StringMatcher
+	)
+
+	if enableTLS {
+		uri = fmt.Sprintf("https://%s:%d", http.Host, http.Port)
```
liorokman, line 147: Instead of building `uri` up as a string, why not use `net/url`'s [URL](https://pkg.go.dev/net/url#URL).

Code would look something like this:

```golang
uri := net.URL{
    Host:  http.Host,
    Port:  http.Port,
    Path:  ptr.Deref(http.Path, ""),
}
if enableTLS {
    url.Scheme = "https"
} else {
    url.Scheme = "http"
}
// Convert to a string for usage latter on.
urlAsString := uri.String()
```

This has the following advantages:
1. Properly escaping elements in the path
2. Making sure that a valid URL is created in corner cases. Example - the current code fails to create a valid URL if the `*http.Path` doesn't start with a `/`.

zhaohuabing:
Hi guys. Apologies for the force push; I had to rebase the code due to changes in the API.

internal/xds/translator/extauth.go
```diff
@@ -0,0 +1,314 @@
...
+		return errors.New("xds resource table is nil")
+	}
+
+	var errs error
+	for _, route := range routes {
+		if !routeContainsExtAuth(route) {
+			continue
+		}
+		if route.ExtAuth.HTTP != nil {
+			if err := createExtServiceXDSCluster(
```
guydc, line 213: Can we do this more efficiently by treating security policies like do `accesslog`/`tracing` and translate their destinations to a new list of destinations of the XDS struct? Then, we can have a dedicated translation phase in the translator to these "reusable" clusters.

internal/xds/translator/jwt.go
```diff
@@ -181,8 +181,10 @@ func buildJWTAuthn(irListener *ir.HTTPListener) (*jwtauthnv3.JwtAuthentication,
 
 // buildXdsUpstreamTLSSocket returns an xDS TransportSocket that uses envoyTrustBundle
 // as the CA to authenticate server certificates.
-func buildXdsUpstreamTLSSocket() (*corev3.TransportSocket, error) {
+// TODO huabing: add support for custom CA and client certificate.
+func buildXdsUpstreamTLSSocket(sni string) (*corev3.TransportSocket, error) {
```
guydc, line 185: Should we use a `BackendRef` also for jwks and make the cluster generator in ext-auth a utility that's also used here? Or do we think that jwks is likely an out-of-cluster endpoint so prefer to not bother the user with FQDN services in this scenario? Either way, if we do that, better to do it in another PR.

internal/gatewayapi/securitypolicy.go
```diff
@@ -626,3 +645,221 @@ func (t *Translator) buildBasicAuth(
...
+			Authority:        authority,
+		}
+	}
+	return extAuth, nil
+}
+
+// TODO: zhaohuabing combine this function with the one in the route translator
+// validateExtServiceBackendReference validates the backend reference for an
+// external service referenced by an EG policy.
+// This can also be used for the other external services deployed in the cluster,
```
guydc, line 724: Yes, this can be a very useful utility for other translators, as well as other functions in this file. Maybe extract this to a more common location?

internal/xds/translator/extauth.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, err
+	}
+
+	extAuthAny, err := anypb.New(extAuthProto)
+	if err != nil {
+		return nil, err
+	}
+
+	return &hcmv3.HttpFilter{
+		Name: extAuthFilterName(route),
```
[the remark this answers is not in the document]
zhaohuabing: Yes, but would you mind if I address this in a follow-up PR? This is a common issue for a few other filters as well, and I would like to keep this PR focusing on Ext Auth.

internal/xds/translator/jwt.go
zhaohuabing: I believe jwks mostly resides outside of the cluster, so EG shouldn't use a `BackendRef` for it.

internal/xds/translator/extauth.go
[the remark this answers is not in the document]
guydc: Yes, sure.

internal/xds/translator/jwt.go
guydc: Sorry, my phrasing was confusing. I agree, no need for backendref here.

internal/gatewayapi/testdata/securitypolicy-with-extauth-invalid-no-matching-port.in.yaml
```diff
@@ -0,0 +1,64 @@
...
+      name: policy-for-gateway
+    spec:
+      targetRef:
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: gateway-1
+        namespace: default
+      extAuth:
+        http:
+          Name: http-backend
```
arkodg, line 59: shouldnt this be ?
```
http:
  backendRef:
    namespace: ...
    port: 80
```

internal/xds/translator/extauth.go
zhaohuabing: Looks like we can't? The clusters need to be handled inside the `processHTTPListenerXdsTranslation` because they're part of the HTTPListener.

internal/gatewayapi/testdata/securitypolicy-with-extauth-invalid-no-matching-port.in.yaml
zhaohuabing: BackendObjectReference is an incline struct in the API.

```
type GRPCExtAuthService struct {
}

type HTTPExtAuthService struct {

	Path *string `json:"path,omitempty"`

}
```

Or we can make it explicit like this:

```
type GRPCExtAuthService struct {
	BackendRef gwapiv1.BackendObjectReference 
}

type HTTPExtAuthService struct {
	BackendRef gwapiv1.BackendObjectReference

	Path *string `json:"path,omitempty"`

}
```

internal/xds/translator/basicauth.go
```diff
@@ -57,13 +57,6 @@ func (*basicAuth) patchHCM(mgr *hcmv3.HttpConnectionManager, irListener *ir.HTTP
 			continue
 		}
 
-		// skip if the filter already exists
```
arkodg, line 60: why is this deleted ?

internal/xds/translator/oidc.go
```diff
@@ -63,13 +62,6 @@ func (*oidc) patchHCM(mgr *hcmv3.HttpConnectionManager, irListener *ir.HTTPListe
 			continue
 		}
 
-		// skip if the filter already exists
```
arkodg, line 66: same q, why is this removed ?

arkodg:
thanks @zhaohuabing ! PR looks good, added some comments mainly to understand the code deletions

internal/xds/translator/basicauth.go
zhaohuabing: https://github.com/envoyproxy/gateway/pull/2537#discussion_r1477468505

arkodg: not sure I understand, can you please elaborate ?

zhaohuabing: If multiple policies target the same route, the latest wins.
If a gateway-level policy and route-level policy target the same route, the route-level policy wins.
And the per-route filter has the route name in it, so the same filter name will not appear more than once in the HCM filter chain.

arkodg (approved):
LGTM thanks !

guydc (approved):

PR #2539, opened 2024-01-30
feat(translator): implement max requests per connection

### LINKED ISSUE #2508: Support Max Requests Per Connection in Backend Traffic Policy

guydc:
Fixes #2508
Based on API in #2513


arkodg (approved):
LGTM thanks !

Xunzhuo (approved):
Looks good, thanks!

PR #2544, opened 2024-01-31
docs: improve api type members

zirain:
before:

image

after:

image


arkodg:
nice ! this LGTM, one thing I see missing is the `optional` / `required` tags, how should we surface that ?

arkodg (approved):

zirain:
> nice ! this LGTM, one thing I see missing is the `optional` / `required` tags, how should we surface that ?

let me find out and try to fix with another PR.

PR #2550, opened 2024-02-01
feat: add support for `egctl x status`

### LINKED ISSUE #2463: Surface resource status summary in egctl
*Description*:
would be used.

As a user, Id like to use egtcl to give me a summary of the status of all resources so I can quickly find out 
if resources have be translated and programmed correctly into the data plane

Suggestions for UX
* egctl x status
OR
* egctl x analyze

Relates to https://github.com/envoyproxy/gateway/issues/2028

[optional *Relevant Links*:]

shawnh2:
**What type of PR is this?**

feature: add `egctl x status` subcommand

some examples:

image

image

Fixes #2463


internal/cmd/egctl/status.go
```diff
@@ -0,0 +1,324 @@
...
+
+func runStatus(ctx context.Context, cli client.Client, table *tabwriter.Writer, resourceType, namespace string, quiet, verbose, allNamespaces bool) error {
+	var resourcesList client.ObjectList
+
+	if allNamespaces {
+		namespace = ""
+	}
+
+	switch strings.ToLower(resourceType) {
+	case "gc", "gcs", "gatewayclass", "gatewayclasses":
```
arkodg, line 93: lets only use singular terms similar to `kubectl get`

arkodg, line 93:
```suggestion
	case "gc", "gatewayclass":
```

arkodg: lets do this for all APIs

internal/cmd/egctl/status.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+func runStatus(ctx context.Context, cli client.Client, table *tabwriter.Writer, resourceType, namespace string, quiet, verbose, allNamespaces bool) error {
+	var resourcesList client.ObjectList
+
+	if allNamespaces {
+		namespace = ""
+	}
+
+	switch strings.ToLower(resourceType) {
```
arkodg, line 92: something for the future, dont want to block this PR, but `all` can be helpful here to fetch statuses of all CRDs at once

arkodg:
this PR looks good @shawnh ! added a minor comment

internal/cmd/egctl/status.go
shawnh2: done

internal/cmd/egctl/status.go
shawnh2: sure, i will add this in a follow-up PR

arkodg (approved):
LGTM thanks !

Issue #2551, opened 2024-02-01
Use BackendRef to define Ext Auth Service


arkodg:
*Description*:
>Describe the issue.

Consider using BackendRef definition https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendObjectReference In the Ext Auth API instead of the current host:port definition https://gateway.envoyproxy.io/latest/api/extension_types/#extauth

* safer, since backendRef access to another namespace can be controlled using a ReferenceGrant
* deduplicates user authoring the ext auth endpoint definition in multiple places
* can be handy in the future in case we want to apply a Policy to a BackendRef e.g. BackendTLSPolicy
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.BackendTLSPolicySpec

Relates to https://github.com/envoyproxy/gateway/issues/1059

Also relates to
*  ext-proc svc defintion https://github.com/envoyproxy/gateway/issues/2546
* upstream TLS https://github.com/envoyproxy/gateway/issues/2536

arkodg:
@zhaohuabing this was discussed in the community meeting today with @guydc , and I'm a +1 on this, wdyt
cc @envoyproxy/gateway-maintainers

zhaohuabing:
The implication of using 'BackendObjectRefrence' is that out of cluster ext auth services will not be supported. But I guess this is okay because normally the ext auth service should be deployed inside the cluster as a k8s service.

arkodg:
> The implication of using 'BackendObjectRefrence' is that out of cluster ext auth services will not be supported. But I guess this is okay because normally the ext auth service should be deployed inside the cluster as a k8s service.

Yah, you can still use FQDN EndpointSlices for the out of cluster case, but it isn't pretty

PR #2553, opened 2024-02-02
use BackendObjectReference to represent the ext auth service

zhaohuabing:
Related: https://github.com/envoyproxy/gateway/issues/2551

This PR does the following:

* Use BackendObjectReference to represent the ext auth service
* Remove TLS configuration from ext auth API, EG can leverage Gateway API [BackendPolicy](https://gateway-api.sigs.k8s.io/api-types/backendtlspolicy/) to specify the TLS configuration for ext auth services


api/v1alpha1/ext_auth_types.go
```diff
@@ -59,39 +59,24 @@ type ExtAuth struct {
...
-
-	// Port is the network port of the gRPC External Authorization service.
-	Port gwapiv1a2.PortNumber `json:"port"`
-
-	// TLS defines the TLS configuration for the gRPC External Authorization service.
-	// Note: If not specified, the proxy will talk to the gRPC External Authorization
-	// service in plaintext.
-	// +optional
-	TLS *TLSConfig `json:"tls,omitempty"`
+	// BackendObjectReference references a Kubernetes object that represents the
```
arkodg, line 62: Should this live in the top level ?

zhaohuabing: It can be changed to the below, I slightly prefer the current one.

```
type ExtAuth struct {
	// Type decides the type of External Authorization.
	// Valid ExtAuthServiceType values are "GRPC" or "HTTP".
	// +kubebuilder:validation:Enum=GRPC;HTTP
	// +unionDiscriminator
	Type ExtAuthServiceType `json:"type"`

	HTTP gwapiv1.BackendObjectReference

	GRPC gwapiv1.BackendObjectReference
	
	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
		
	// The below configurations only apply to HTTP external authorization service.
	// More will be added in the future.
	Path *string `json:"path,omitempty"`

}
```

arkodg: why do we need a unique field for HTTP and GRPC backends ?

arkodg (approved):
LGTM thanks !

arkodg (approved):

zirain (approved):

arkodg:
@zhaohuabing thoughts on  moving all fields ( even the common ones)  under either 'grpc'  or http  ?
That keeps the API consistent with other types like RateLimit

zhaohuabing:
> @zhaohuabing thoughts on moving all fields ( even the common ones) under either 'grpc' or http ? That keeps the API consistent with other types like RateLimit

I like this idea, but also a bit hesitate because there're a number of "common options" that could be added in the future. This could cause too many duplications.

```
{
  "grpc_service": {...},
  "http_service": {...},
  "transport_api_version": ...,
  "failure_mode_allow": ...,
  "failure_mode_allow_header_add": ...,
  "with_request_body": {...},
  "clear_route_cache": ...,
  "status_on_error": {...},
  "metadata_context_namespaces": [],
  "typed_metadata_context_namespaces": [],
  "route_metadata_context_namespaces": [],
  "route_typed_metadata_context_namespaces": [],
  "filter_enabled": {...},
  "filter_enabled_metadata": {...},
  "deny_at_disable": {...},
  "include_peer_certificate": ...,
  "stat_prefix": ...,
  "bootstrap_metadata_labels_key": ...,
  "allowed_headers": {...},
  "include_tls_session": ...,
  "charge_cluster_response_stats": {...}
}
```

arkodg:
Option 1
```
extAuth:
 
  type: GRPC # all GRPC knobs
  grpc:
    backendRef:
     ....
    headersToExtAuth 
```
vs

Option 2
```
extAuth:
  headersToExtAuth: 
  grpc: #grpc specific only
    backendRef:
     ........
     
```

arkodg:
lets go with 2, it maybe easier for the user to swap between http and grpc ext auth service with option 2

PR #2563, opened 2024-02-05
fix: LoadBalancerIP validation to prevent trailing period

davidalger:
**What type of PR is this?**

fix: LoadBalancerIP validation to prevent trailing period

Current validation allowed `loadBalancerIP` to have a trailing period. This corrects the match expression to reject values such as `1.2.3.4.` and moves the validation rules for shared types to reside with the objects they are validating.

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -210,6 +210,8 @@ const (
 )
 
 // KubernetesServiceSpec defines the desired state of the Kubernetes service resource.
+// +kubebuilder:validation:XValidation:message="allocateLoadBalancerNodePorts can only be set for LoadBalancer type",rule="!has(self.allocateLoadBalancerNodePorts) || self.type == 'LoadBalancer'"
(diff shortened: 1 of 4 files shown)


api/v1alpha1/shared_types.go
```diff
@@ -243,6 +245,8 @@ type KubernetesServiceSpec struct {
 	// may be ignored if the load balancer provider does not support this feature.
 	// This field has been deprecated in Kubernetes, but it is still used for setting the IP Address in some cloud
 	// providers such as GCP.
+	//
+	// +kubebuilder:validation:XValidation:message="loadBalancerIP must be a valid IPv4 address",rule="self.matches(r\"^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$\")"
```
zirain, line 249: Currently you can only use RE, nothing else.

zirain (approved):

api/v1alpha1/shared_types.go
davidalger: It's unfortunate, since lookahead would have easily worked here to avoid the duplication were it supported.

arkodg (approved):
LGTM thanks !

PR #2566, opened 2024-02-05
feat: Support Client IP Detection using Custom Header

davidalger:
**What type of PR is this?**

feat: Support Client IP Detection using Custom Header

Similar to #2535 this extends ClientTrafficPolicy to support using a custom header rather than XFF headers to perform  Client IP Detection. Where envoy proxies reside behind a CDN or a cloud providers global LBs, this is useful as often the client IP is provided to the backend by the downstream CDN/LB proxy when handling the request.

Two examples:

- GCP backend services used to route traffic from a global LB to envoy proxies [configured with](https://cloud.google.com/load-balancing/docs/https/custom-headers-global) `"customRequestHeaders":["x-client-ip-address:{client_ip_address}"]` in which case the `x-client-ip-address` may be used to determine the address in a more favorable manner than counting number of hops in the XFF header.
- Cloudflare CDN [configured with](https://developers.cloudflare.com/network/true-client-ip-header/) the `True-Client-IP` header enabled to include the end user’s IP address on all requests to envoy proxies.

Proposed configuration support on `ClientTrafficPolicy`:

```yaml
  clientIPDetection:
    customHeader:
        name: X-Client-IP-Address
        failClosed: true
```


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -107,6 +116,28 @@ type XForwardedForSettings struct {
...
+// Refer to https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/original_ip_detection/custom_header/v3/custom_header.proto
+// for more details.
+type CustomHeaderExtensionSettings struct {
+	// HeaderName of the of the header containing the original downstream remote address, if present.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:Pattern="^[A-Za-z0-9-]+$"
+	//
+	HeaderName string `json:"headerName"`
```
arkodg, line 130:
```suggestion
	Name string `json:"headerName"`
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -107,6 +116,28 @@ type XForwardedForSettings struct {
...
+	// RejectWithStatus is the HTTP response status to use when detection fails, if present. May be
+	// any valid HTTP response status code within the range 400-511 (inclusive).
+	//
+	// +kubebuilder:validation:Minimum=400
+	// +kubebuilder:validation:Maximum=511
+	//
+	// +optional
+	RejectWithStatus *uint32 `json:"rejectWithStatus,omitempty"`
```
arkodg, line 138: instead of this, should this be ?
```
failClosed:
  status
```

davidalger: I like `failClosed` and keeping the terms uniform here. Would it make sense to make it flatter so `failClosed` is similarly boolean?

```yaml
failClosed: bool       # default false
failStatus: uint32     # default 400
```

davidalger: Might be ok to just go with a single boolean rejecting with 400 bad request when `failClosed: true` instead of making the status code configurable here as well. Wdyt?

arkodg: `failClosed: bool` with a default status sgtm, from the docs, looks like the preferred default is `403` 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/original_ip_detection/custom_header/v3/custom_header.proto

davidalger: 403 sounds good, let's go with it

internal/xds/translator/listener.go
```diff
@@ -149,9 +189,10 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	}
 
 	// Client IP detection
-	var xffNumTrustedHops uint32
-	if irListener.ClientIPDetection != nil && irListener.ClientIPDetection.XForwardedFor != nil {
-		xffNumTrustedHops = ptr.Deref(irListener.ClientIPDetection.XForwardedFor.NumTrustedHops, 0)
+	var useRemoteAddress = true
+	var originalIPDetectionExtensions = originalIPDetectionExtensions(irListener.ClientIPDetection)
+	if originalIPDetectionExtensions != nil {
+		useRemoteAddress = false
```
arkodg, line 195: does this have to be false ?
this knobs helps the next upstream with middleware client ip info, unsure how its tied to downstream client ip info which is being used to determine true client ip for logging, acl, ratelimiting

davidalger: > does this have to be false? this knobs helps the next upstream with middleware client ip info

It has to be `false`. Both top-level `xff_num_trusted_hops` and `use_remote_address` are mutually exclusive with `original_ip_detection_extensions`. There's a note in the docs about it, and if you try setting it to `true` while also configuring `original_ip_detection_extensions` envoy will reject the config:

```
[2024-02-07 22:30:29.118][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) default/eg/http: Original IP detection extensions and use_remote_address may not be mixed   
```

And yes, this does mean when `customHeader` is configured the `X-Forwarded-For` header will not have the envoy proxy IP appended to it since `use_remote_address` is `false` and it's assumed you aren't relying on the XFF headers when using alternate forms of client IP detection.

> unsure how its tied to downstream client ip info which is being used to determine true client ip for logging, acl, ratelimiting

The value of the named header will be present in log entries as the `downstream_remote_address`, so I would assume that it would still be used for ACLs and rate limits.

`AllowExtensionToSetAddressAsTrusted: true` is set on the extension which at minimum allows the address to also be used to determine if the address is "internal" or not based on the CIDR range.

arkodg (approved):
LGTM thanks !

internal/xds/translator/listener.go
arkodg: thanks for explaining this

davidalger:
@arkodg anything required from me to get this one across the finish line?

arkodg:
> @arkodg anything required from me to get this one across the finish line?

waiting for another review from a maintainer before merging this one, cc @zhaohuabing @zirain

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -89,11 +89,20 @@ type ClientTrafficPolicySpec struct {
 }
 
 // ClientIPDetectionSettings provides configuration for determining the original client IP address for requests.
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.xForwardedFor) && has(self.customHeader))",message="customHeader cannot be used in conjunction with xForwardedFor"
```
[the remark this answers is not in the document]
davidalger: That is correct. The top-level `xff_num_trusted_hops` knob in Envoy is mutually exclusive with `original_ip_detection_extensions`. Even if Envoy supported both simultaneously though, I'm not sure it would make sense to have two methods of downstream client IP detection configured at the same time as it would muddy where the logged downstream IP address came from.

When `customHeader` is specified, any XFF headers are essentially ignored.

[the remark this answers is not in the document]
zhaohuabing: > One of 'xForwordFor' and 'customHeader' must be specified?

Can we also add a CEL validation for that?

zirain (approved):

api/v1alpha1/clienttrafficpolicy_types.go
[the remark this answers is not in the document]
davidalger: @zhaohuabing I think I misunderstood your question at first. They are mutually exclusive, but neither one is required. When both are omitted, they are simply not configured and the default behavior is to use the remote address which is correct when Envoy is directly exposed to clients rather than residing behind a CDN or cloud-provider LB. Not sure any additional CEL validation is needed here.

AliceProxy (approved):

PR #2567, opened 2024-02-06
oidc support specify hmac key

yowenter:
so the routes will be able to share the same oauth flow cookie

**What type of PR is this?**
 
* "feat(oidc): add oidc hmac secret configuration "

Serveral routes may use the same oidc configuration. eg, the micro services route use different path prefix.
However, their oauth flow cookie can not be shared since the hmac verification will fail if each route use random generated hmac secret. [hmac validate](https://github.com/envoyproxy/envoy/blob/3d67a3f940db59057a6f22db81fae8ecf2bcf5d3/source/extensions/filters/http/oauth2/filter.cc#L429)

```
bool OAuth2Filter::canSkipOAuth(Http::RequestHeaderMap& headers) const {
  // We can skip OAuth if the supplied HMAC cookie is valid. Apply the OAuth details as headers
  // if we successfully validate the cookie.
  validator_->setParams(headers, config_->tokenSecret());
  if (validator_->isValid()) {
    config_->stats().oauth_success_.inc();
    if (config_->forwardBearerToken() && !validator_->token().empty()) {
      setBearerToken(headers, validator_->token());
    }
    ENVOY_LOG(debug, "skipping oauth flow due to valid hmac cookie");
    return true;
  }
  ENVOY_LOG(debug, "can not skip oauth flow");
  return false;
}
```

The OIDC configuration should support use same hmac key for verification, the different route will share the same oauth flow cookie.


arkodg:
@yowenter have you tried creating one SecurityPolicy and defining all matches in the same `HTTPRoute` which the policy can target

zhaohuabing:
@yowenter @arkodg 

Ideally, I prefer not exposing HMAC to EG API since it's just an implementation detail of the Envoy OAuth2 filter.

This reminds me of another potential issue: Per-route OIDC probably doesn't work because the scope of token and hmac cookies set by the Envoy OAuth2 filter is at the root path. So if two `HTTPRoutes` are under the same host, and two `SecurityPolicies` with different OIDC configurations target these two `HTTPRotes`, their cookies will overwrite each other. This is because the cookies are scoped to the same domain and path "/".

https://github.com/envoyproxy/envoy/blob/3d67a3f940db59057a6f22db81fae8ecf2bcf5d3/source/extensions/filters/http/oauth2/filter.cc#L41

yowenter:
> @yowenter have you tried creating one SecurityPolicy and defining all matches in the same `HTTPRoute` which the policy can target

@arkodg  good point, I may try later.

yowenter:
> @yowenter have you tried creating one SecurityPolicy and defining all matches in the same `HTTPRoute` which the policy can target

Hi, there, I've tried using only one security policy.
 However, it raised err:
`* spec.rules: Too many: 25: must have at most 16 items`

zhaohuabing:
Hi @yowenter  EG now uses a shared HMAC secret generated by the CertGen job.

PR #2568, opened 2024-02-06
ci: codedev token

zirain:
fix: https://github.com/envoyproxy/gateway/pull/2558

doc: https://docs.codecov.com/docs/adding-the-codecov-token#github-actions


zirain:
@phlax @Xunzhuo need token for codedev, can you help to set the secret?

Xunzhuo:
Will update it tomorrow

zirain:
> Will update it tomorrow

not sure why it's passed now.

.github/workflows/build_and_test.yaml
```diff
@@ -52,8 +52,9 @@ jobs:
     - name: Run Coverage Tests
       run: make go.test.coverage
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@4fe8c5f003fae66aa5ebb77cfd3e7bfbbda0b6b0  # v3.1.5
+      uses: codecov/codecov-action@e0b68c6749509c5f83f984dd99a76a1c1a231044  # v4.0.1
       with:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
arkodg, line 57: how did this token get injected into secrets ?

zirain: i have no idea why it passed now, maybe  there is a org secret.

phlax: i would assume that no password is being passed - i think that is standard for opensource/public/prs - and dont see any secret defined anywhere

arkodg: we do have the codecov app installed for the repo, which maybe related

phlax: yeah - its installed as an app - so it gets notifications and probs ci results from that

Xunzhuo:
Should we close this ?

arkodg:
> Should we close this ?

yeah, I crosschecked and we are not adding a TOKEN to our repo or org secrets, so not sure what's happening here

Issue #2546, opened 2024-01-31
Support HTTP External Processing Filter


guydc:
*Description*:
Envoy’s HTTP [External Processing Filter]( https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter) can be used by vendors and extension developers to enhance Envoy Proxy's data-plane feature set. The `ext-proc` filter implements a callout to an external grpc service in request and response flows. The external processor service can inspect and mutate requests and responses, access context data such as attributes and dynamic metadata and emit dynamic metadata. 

The external processing filter is included in the envoy base image and its security posture is robust. The external processing protocol and/or filter are also used in the industry for extensibility and integration: [GCP Application LB](https://cloud.google.com/service-extensions/docs/callouts-overview), [Gloo Edge](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/extproc/about/), [Imperva WAF](https://docs.imperva.com/bundle/v14.7-web-application-firewall-user-guide/page/79024.htm). 

Envoy Gateway can support the external processing filter in a way that is similar to the [ext-authz implementation](https://github.com/envoyproxy/gateway/pull/2435). If supported, this filter will fulfill some of the GA requirements outlined in #2025 related to Envoy Proxy extensibility.

arkodg:
+1 to this
my preference would be to include this into #2025 as a new Policy and combine all dev extensions into one new Policy resource type 
cc @Xunzhuo

zirain:
we could do this with EnvoyPatchPolicy now, right? what the benifit for create an new policy?

arkodg:
a dedicated ExtensionAPI can do a better job
* dealing with config validation
* reducing surface area of extension
* doing extra work such as downloading wasm from registry

zirain:
> doing extra work such as downloading wasm from registry

you want download wasm in controlplane?

zirain:
I'm +1 for a dedicated ExtensionAPI, maybe named (Gateway|Proxy|Envoy)(Http/Networking)Filter, focus on insert/replace filters, and use ECDS to send XDS.

Xunzhuo:
+1 for adding support to this, thanks for raisin it @guydc.

guydc:
Please assign to me.

Issue #2025, opened 2023-10-20
State and Future of EnvoyGateway Extension Points


Xunzhuo:
## Current State

In v0.4.0, eg introduces [ExtensionManager](https://gateway.envoyproxy.io/latest/design/extending-envoy-gateway/), grpc based extension hooks, to dynamically modify xds resources generated by EG.

In v0.5.0, eg introduces [EnvoyPatchPolicy](https://gateway.envoyproxy.io/latest/design/envoy-patch-policy/), a CRD resource, to dynamically patch xds resource generated by EG.

In the two above approaches, the extension target is the whole xds resources. It unlocks the full power of xds for vendors/teams to extend EG abilities, which is powerful and flexible. 

## What`s More?

EnvoyProxy itself provides some extension http filters, which allows people dynamically extend envoyproxy without restarting it, like [Lua](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter), [WASM](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/wasm_filter) and new introduced [Go](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/golang_filter) Extension Filter.

I propose EG to provide high level APIs(CRDs) to abstract these extension points. From my experience of using EnvoyFilter in istio, compared with the full xds patch/modify, these ways have more specific target scopes and easier to be control, which are safer , easier and less dangerous to be used.

## ExtensionRef or PolicyAttachment ?

ExtensionRef can be used per-route-rule/per-backend, and PolicyAttachment can be used to whole Gateway/xRoute.

+ WASMFilter or WASMExtensionPolicy CRD
+ LuaFilter or LuaExtensionPolicy CRD
+ GoFilter  or GoExtensionPolicy CRD

arkodg:
Im a +1 on this idea, it can be a safer, less scoped way than [EnvoyPatchPolicy](https://gateway.envoyproxy.io/latest/design/envoy-patch-policy/)  of extending EG

denniskniep:
@zhaohuabing @arkodg  is there already some progress on integrating GoLang Filters (HTTP, Network)

Envoy Docs:
https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/golang_filter#config-network-filters-golang
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/golang_filter#config-http-filters-golang


Envoy Examples:
[...]

PR #2570, opened 2024-02-06
API: EnvoyExtensionPolicy

guydc:

This PR introduces a new policy type, `EnvoyExtensionPolicy`, that can be used to add multiple HTTP and Network extension filters (such as lua, wasm, golang, ext-proc), and define their order of execution. 

Relates to #2546, #2025, #2877


api/v1alpha1/trafficextensionypolicy_types.go
```diff
@@ -0,0 +1,90 @@
...
+	KindTrafficExtensionPolicy = "TrafficExtensionPolicy"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=sp
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Accepted")].reason`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// TrafficExtensionPolicy allows the user to configure various traffic extensibility options for the Gateway.
```
arkodg, line 24: prefer `EnvoyExtensionPolicy` since these extensions are specific to envoy or `ExtensionPolicy` if these extensions are common across proxy implementations

guydc:
In yesterday's community meeting, @arkodg raised the following requirements:
- Users may need to attach multiple policies in order to configure multiple extension filters (a common use case for WASM, LUA)
- Users may need to control the logical phase of filter execution (e.g. before auth, before rate limit, ... )  
- Users may need to control the order of execution of extension filters amongst themselves (within a specific phase)

Istio's implementation for [WASM plugins](https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/) uses `phase` and `priority` to achieve this. Envoy Gateway can follow similar principles. The definition of such phases can have an impact on #2571, for example limiting the level of flexibility that users will have when re-arranging the Envoy Gateway built-in extensions. 

@envoyproxy/gateway-maintainers - WDYT? Should we follow the istio pattern and implement phases and priorities for user-defined extensions?

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	Headers *ExtProcHeaderProcessingMode `json:"request,omitempty"`
+	// Defines body processing mode
+	Body *ExtProcBodyProcessingMode `json:"response,omitempty"`
+}
+
+// ExtProcProcessingMode defines if and how headers and bodies are sent to the service.
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/processing_mode.proto#envoy-v3-api-msg-extensions-filters-http-ext-proc-v3-processingmode
+type ExtProcProcessingMode struct {
+	// Defines header and body treatment for requests
+	// Default: SEND
```
liorokman, line 40: Set the default with 
```golang
// +kubebuilder:default:=Send
```

( https://book.kubebuilder.io/reference/markers/crd-validation )

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	Request *ProcessingModeOptions `json:"request,omitempty"`
+	// Defines header and body treatment for responses
+	// Default: NONE
+	Response *ProcessingModeOptions `json:"response,omitempty"`
```
liorokman, line 44: Set the default with

```golang
// +kubebuilder:default:=None
```

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	ReceivingNamespaces  []MetadataNamespaces `json:"receivingNamespaces,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.group) || self.service.backendRef.group == \"\") : true", message="group is invalid, only the core API group (specified by omitting the group field or setting it to an empty string) is supported"
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.kind) || self.service.backendRef.kind == 'Service') : true", message="kind is invalid, only Service (specified by omitting the kind field or setting it to 'Service') is supported"
+//
+// ExtProc defines the configuration for External Processing.
+type ExtProc struct {
+	// Service defines the configuration of the external processing service
+	Service *ExtProcService `json:"service"`
```
liorokman, line 75: If `Service` is an optional field in the api, add the `omitempty` attribute to the field tags.
If it isn't, don't use a pointer here.

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	// Default: request and response headers are sent, bodies are not sent
+	ProcessingMode *ExtProcProcessingMode `json:"processingMode,omitempty"`
+	// Attributes defines which envoy request and response attributes are provided as context to external processor
+	// Default: no attributes are sent
+	Attributes *ExtProcAttributes `json:"attributes,omitempty"`
+	// MetadataOptions defines options related to the sending and receiving of dynamic metadata
+	// Default: no metadata context is sent or received
+	MetadataOptions *ExtProcMetadataOptions `json:"metadataOptions,omitempty"`
+	// The timeout for a response to be returned from the external processor
+	// Default: 200ms
```
liorokman, line 86: Add a `+kubebuilder:default` comment for `MessageTimeout`.

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -0,0 +1,120 @@
...
+	gwapiv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+const (
+	// KindEnvoyExtensionPolicy is the name of the EnvoyExtensionPolicy kind.
+	KindEnvoyExtensionPolicy = "EnvoyExtensionPolicy"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=sp
```
liorokman, line 19: This `shortName` is the same as the `shortName` defined for the `SecurityPolicy` resource. 

Consider removing the short name, or changing it to `ep`.

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	// defines attributes to send for Request processing
+	Request []string `json:"request,omitempty"`
+	// defines attributes to send for Response processing
+	Response []string `json:"response,omitempty"`
+}
+
+// MetadataNamespaces defines metadata namespaces that can be used to forward or receive dynamic metadata
+type MetadataNamespaces struct {
+	// Specifies a list of metadata namespaces whose values, if present, will be passed to the ext_proc service as an opaque protobuf::Struct.
+	Untyped []string `json:"untyped,omitempty"`
```
liorokman, line 58: add `// +optional` ?

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+	// Specifies a list of metadata namespaces whose values, if present, will be passed to the ext_proc service as a protobuf::Any.
+	Typed []string `json:"typed,omitempty"`
```
liorokman, line 60: add `// +optional` ?

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,102 @@
...
+type ExtProcMetadataOptions struct {
+	ForwardingNamespaces []MetadataNamespaces `json:"forwardingNamespaces,omitempty"`
+	ReceivingNamespaces  []MetadataNamespaces `json:"receivingNamespaces,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.group) || self.service.backendRef.group == \"\") : true", message="group is invalid, only the core API group (specified by omitting the group field or setting it to an empty string) is supported"
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.kind) || self.service.backendRef.kind == 'Service') : true", message="kind is invalid, only Service (specified by omitting the kind field or setting it to 'Service') is supported"
+//
+// ExtProc defines the configuration for External Processing.
+type ExtProc struct {
```
liorokman, line 73: Add `// +optional` as relevant for the fields in the struct?

guydc:
Incorporated points from the discussion in today's community meeting:
- Filter Ordering in `EnvoyProxy`
- Disablement of Extensibility by default, enablement in `EnvoyGateway`
- Remove distinction between HTTP/Network in the interface 
- Allow route attachment

arkodg:
@zhaohuabing lets pause on https://github.com/envoyproxy/gateway/pull/2877 until this API is done, and the WASM API can be incorporated as a field within this API

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -0,0 +1,87 @@
...
+// EnvoyExtensionPolicy allows the user to configure various envoy extensibility options for the Gateway.
+type EnvoyExtensionPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the desired state of EnvoyExtensionPolicy.
+	Spec EnvoyExtensionPolicySpec `json:"spec"`
+
+	// Status defines the current status of EnvoyExtensionPolicy.
+	Status EnvoyExtensionPolicyStatus `json:"status,omitempty"`
```
arkodg, line 33: probably PolicyStatus here

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,6 +104,14 @@ type EnvoyProxySpec struct {
...
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// FilterOrdering defines the order of Envoy HTTP filter execution. If a partial filter list is provided, the default
+	// order will apply, and only the listed filters will be re-ordered amongst themselves.
+	// Default: CORS, External-Processing, WASM, External-Authorization, Basic-Authorization, OAuth2, JWT-Authentication,
+	// Fault-Injection, Local-Rate-Limit, Global-Rate-Limit
+	//
+	// +optional
+	FilterOrdering []EnvoyFilterName `json:"filterOrdering,omitempty"`
```
arkodg, line 114: suggest keep filter ordering out of scope of this API, and solving it separately in https://github.com/envoyproxy/gateway/issues/2571

zhaohuabing:
Should the spec support defining extensions as a list?

Unlike other filters(Basic Auth, OAuth2, JWT, etc), where typically only one filter per HTTPRoute is needed, users often require multiple custom Envoy extensions(wasm, lua, golang, ext auth?, etc.) of the same type for different purposes.  For example, they might use a wasm filter for WAF and another for auth on the same HTTPRoute.  

```
kind: EnvoyExtensionPolicy
metadata:
  name: my-extensions
spec:
  targetRef:
    kind: HTTPRoute
    name: backend
  wasm:
    - name: wasm_plugin_1
      vmid: my_vm_id # optional
      rootId: my_root_id  # optional
      code: 
        configMap: wasmConfigMap # wasm code is stored in a configMap
      # http: 
      #   uri: https://foo.bar.com/example.wasm
      #   sha: 
      # image: <oci-registry>.foo.bar.com/wasm-to-oci:v1  
      config:
        "@type": "type.googleapis.com/google.protobuf.StringValue"
        value: |
          {}  
    - name: my_plugin_2
      ...
  ExtProc:
    - name: exproc_plugin_1 
      ....
    - name: exproc_plugin_2     
  Lua:
    - name: lua_plugin_1 
      ....
    - name: lua_plugin_2 
.....        
```

And if we make it a list, another emerging problem is that the order of all these filters would be difficult to manage.

zhaohuabing (commented):
Should the spec support defining extensions as a list?

Unlike "normal" filters(Rate Limit, CORS, Basic Auth, OAuth2, JWT, etc), where typically only one filter per HTTPRoute is needed, users often require multiple custom Envoy extensions(wasm, lua, golang, ext auth?, etc.) of the same type for different purposes.  For example, they might use a wasm filter for WAF and another for auth on the same HTTPRoute.  

```
kind: EnvoyExtensionPolicy
metadata:
  name: my-extensions
spec:
  targetRef:
    kind: HTTPRoute
    name: backend
  wasm:
    - name: wasm_plugin_waf
     ...
    - name: my_plugin_auth
      ...
  ExtProc:
    - name: exproc_plugin_1 
      ....
    - name: exproc_plugin_2     
  Lua:
    - name: lua_plugin_1 
      ....
    - name: lua_plugin_2 
.....        
```

guydc:
> Should the spec support defining extensions as a list?

+1

I think that's a reasonable ask, when considering that we allow something similar with envoy patches and it is indeed a very common use case. Maybe we can even avoid dealing with priority and multi-attachment in the first phases for this policy, if we support this + route attachment in the initial iteration.

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -0,0 +1,76 @@
...
+	// int32.max has the lowest priority.
+	// Defaults to 0.
+	//
+	// +optional
+	Priority int32 `json:"priority,omitempty"`
+
+	// ExtProc defines the configuration for the external processor extension.
+	//
+	// +optional
+	ExtProc *ExtProc `json:"extProc,omitempty"`
```
arkodg, line 62: as per huabing's point, should we make this a list ?

guydc:
To make things a bit simpler, I removed ext-proc from this PR. After the top-level resource is added, @zhaohuabing and I can work on ext-proc/wasm in parallel.

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -0,0 +1,71 @@
...
+	Spec EnvoyExtensionPolicySpec `json:"spec"`
+
+	// Status defines the current status of EnvoyExtensionPolicy.
+	Status gwapiv1a2.PolicyStatus `json:"status,omitempty"`
+}
+
+// EnvoyExtensionPolicySpec defines the desired state of EnvoyExtensionPolicy.
+type EnvoyExtensionPolicySpec struct {
+	// +kubebuilder:validation:XValidation:rule="self.group == 'gateway.networking.k8s.io'", message="this policy can only have a targetRef.group of gateway.networking.k8s.io"
+	// +kubebuilder:validation:XValidation:rule="self.kind in ['Gateway']", message="this policy can only have a targetRef.kind of Gateway"
```
arkodg, line 39: can target `Gateway` and `Route`

api/v1alpha1/envoygateway_types.go
```diff
@@ -151,6 +151,10 @@ type ExtensionAPISettings struct {
 	// EnableEnvoyPatchPolicy enables Envoy Gateway to
 	// reconcile and implement the EnvoyPatchPolicy resources.
 	EnableEnvoyPatchPolicy bool `json:"enableEnvoyPatchPolicy"`
+
+	// EnableEnvoyExtensionPolicy enables Envoy Gateway to
```
arkodg, line 155: @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers thoughts on this ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -134,6 +134,20 @@ type ShutdownConfig struct {
 	MinDrainDuration *metav1.Duration `json:"minDrainDuration,omitempty"`
 }
 
+// FilterOrdering defines the ordering of well-known.
+type FilterOrdering struct {
```
arkodg, line 138: lets rm this struct from this API

guydc: my bad, that's there by mistake

site/content/en/latest/contributions/design/envoy-extension-policy.md
```diff
@@ -0,0 +1,148 @@
...
+      status condition `Overridden=True`.
+* A Policy targeting the most specific scope wins over a policy targeting a lesser specific scope.
+  i.e. A Policy targeting a `Listener` overrides a Policy targeting the `Gateway` the listener/section is a part of.
+
+
+## Alternatives
+* The project can indefinitely wait for these configuration parameters to be part of the [Gateway API][].
+* The project can implement support for HTTP traffic extensions using vendor-specific [Gateway API Route Filters][]
+  instead of policies. However, this option will is less convenient for definition of gateway-level extensions.
+
```
arkodg, line 143: * using the envoypatchpolicy
* or the extension manager

site/content/en/latest/contributions/design/envoy-extension-policy.md
```diff
@@ -0,0 +1,148 @@
+---
+title: "EnvoyExtensionPolicy "
+---
+
+## Overview
+
+This design document introduces the `EnvoyExtensionPolicy` API allowing system administrators to configure traffic
```
arkodg, line 7: can we incorporate some of these below points in this doc

* there are now 2 kinds of extensibility today
  * extending EG API by configuring xDS directly using EnvoyPatchPolicy & ExtensionManager
  * the other kind, highlighted in this doc that extends and adds newer functionality on top of what Envoy offers
  
Also a note around how users have been using  EnvoyPatchPolicy & ExtensionManager to directly configure this extensions, but this API provides a safer and more scoped API

arkodg (approved):
LGTM thanks !

api/v1alpha1/envoygateway_types.go
zirain: I'm fine to remove all these, it will not happen unless you create a CR.

zirain (approved):

PR #2574, opened 2024-02-07
fix: Deprecated field error when using RequestHeaderModifier filter

davidalger:
**What type of PR is this?**

fix: Deprecated field error when using RequestHeaderModifier filter

Applying the attached `HTTPRoute` currently results in envoy logging a warning:

```
4-02-07 15:09:47.684][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.config.core.v3.HeaderValueOption Using deprecated option 'envoy.config.core.v3.HeaderValueOption.append' from file base.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

<details>
  <summary><code>HTTPRoute</code> which triggers the warning</summary>

```yaml
kind: HTTPRoute
metadata:
  name: backend
  namespace: default
spec:
  hostnames:
  - www.example.com
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - kind: Service
      name: backend
      port: 3000
    matches:
    - path:
        type: PathPrefix
        value: /
    filters:
    - type: RequestHeaderModifier
      requestHeaderModifier:
        add:
        - name: X-Lorem
          value: dolor
```

</details>

This change updates the route translator to use the `AppendAction` enum which replaces the deprecated boolean field.

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -329,12 +329,20 @@ func buildXdsAddedHeaders(headersToAdd []ir.AddHeader) []*corev3.HeaderValueOpti
 	headerValueOptions := make([]*corev3.HeaderValueOption, len(headersToAdd))
 
 	for i, header := range headersToAdd {
+		var appendAction corev3.HeaderValueOption_HeaderAppendAction
+
+		if header.Append {
+			appendAction = corev3.HeaderValueOption_APPEND_IF_EXISTS_OR_ADD
+		} else {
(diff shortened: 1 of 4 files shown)


arkodg (approved):
LGTM thanks !

shawnh2 (approved):

Issue #2543, opened 2024-01-31
Simple regexp length error not caught by Gateway


akhenakh:
A simple regexp is rejected by Envoy Proxy:

Envoy Proxy is measuring a regexp complexity and compare it with  `max_program_size`.

According to this, increasing it per entry has been deprecated several years ago
https://github.com/envoyproxy/envoy/blob/main/api/envoy/type/matcher/regex.proto#L41

Using 0.6.0 Gateway, this value is still 100 which is quite low, for example this regexp is valued 118: 
`^/v4/advisory/[0-9a-z-]{36}/clone$`.

Envoy Proxy will reject it at runtime, we found it the hard way, since on a restart it will then refuse to load every route 404ing on everything.

Can we increase the default in Gateway and raise an error earlier in Gateway rather than in Proxy?

zirain:
will custom bootstrap help this? should EG make `runtime_flags` configurable?

akhenakh:
It seems other gateways are defaulting to 1000.

In the meantime yes I'm trying to change via replacing the bootstrap...
      static_layer:
              re2.max_program_size.error_level: 1000

arkodg:
thanks for raising this issue, looks like the outcome of this issue is
* figure out what the `re2.max_program_size.error_level` should be for the default case
* set it in the bootstrap
* add a validation in the gateway-api layer to make sure we reject any regexes, else envoy will NACK and reject all xds

as mentioned by zirain, workaround for now is to use a custom bootstrap
https://gateway.envoyproxy.io/latest/user/customize-envoyproxy/#customize-envoyproxy-bootstrap-config

davidalger:
> It seems other gateways are defaulting to 1000.

I did some looking into this. I've seen this in a few places as well as issues opened because 1000 is also not high enough. For example, here are a couple "high cost" expressions:

* `^/api/v1/object/[^\/]{124}/action$` has program size of 1003
* `^/static/[a-zA-Z0-9._-]{0,200}/[a-zA-Z0-9._-]{0,200}$` has program size of 2005

Size can increase linearly based on the length of a single match due to how cost is calculated in RE2. For the first example the cost increases by 8 as the length of the match is incremented, or: `11 + 8 * 124 == 1003`

Program size of regular expressions can also apparently change between RE2 versions, which happened when the cost of `.` was reduced in 2019 breaking a test in Envoy ([xref](https://github.com/envoyproxy/envoy/issues/)9174#issue-530487212)

[...]

PR #2576, opened 2024-02-07
fix: Envoy rejects XDS at runtime losing all routes on restart

davidalger:
**What type of PR is this?**

fix: Envoy rejects XDS at runtime losing all routes on restart

By default Envoy rejects regex routes with a max program size > 100, which is very easy to trigger with simple regex matches:

```
rejected: regex '^/api/v1/object/[\w]{32}/action$' RE2 program size of 107 > max program size of 100
```

When envoy restarts, it loses all it's config resulting in 404 responses to all requests. This change lifts the deprecated validation done by `max_program_size` leaving in place a warning threshold which can be used to monitor for high complexity regex routes without putting stability of the system at risk. See thread on the related issue for full details.

Relates #2543


shawnh2 (approved):
lgtm

PR #2577, opened 2024-02-08
feat: Add support for HTTP/1.0 and HTTP/0.9

liorokman:
This PR implements the API for enabling HTTP/1.0 and HTTP/0.9   (#2534)


internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -393,14 +417,36 @@ func translateListenerSuppressEnvoyHeaders(suppressEnvoyHeaders *bool, httpIR *i
...
+		return nil
 	}
 	httpIR.HTTP1 = &ir.HTTP1Settings{
 		EnableTrailers:     ptr.Deref(http1Settings.EnableTrailers, false),
 		PreserveHeaderCase: ptr.Deref(http1Settings.PreserveHeaderCase, false),
 	}
+	if http1Settings.HTTP10 != nil {
+		var defaultHost *string
+		if ptr.Deref(http1Settings.HTTP10.UseDefaultHost, false) {
+			for _, hostname := range httpIR.Hostnames {
```
arkodg, line 431: imo we should make this more strict - making sure a single hostname exists and its not a wildcard

liorokman: > making sure a single hostname exists and it doesnt include a wildcard (even *.example.com should be rejected) 

It's already like this.

I couldn't find a way to create a test that defines a `Gateway` resource which would translate to the `Hostnames` array containing more than one entry at any time. Notice that the `Gateway` [resource](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.Listener) doesn't allow providing more than one hostname per listener.  

The code initializes the array so that it always contains exactly one entry - either the single value supported by the `Listener` structure or a wildcard entry.

arkodg (approved):
LGTM thanks !

PR #2579, opened 2024-02-08
Change route sorting order to Exact > RegularExpression > PathPrefix

vixns:
**What type of PR is this?**

fix

Like <https://github.com/kubernetes-sigs/gateway-api/issues/1770>, I've encountered a case when converting the following ingress definition:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: sentry
  namespace: sentry
spec:
  tls:
    - hosts:
        - sentry.selfhosted.tld
  rules:
    - host: sentry.selfhosted.tld
      http:
        paths:
          - path: /api/store
            pathType: ImplementationSpecific
            backend:
              service:
                name: sentry-relay
                port:
                  number: 3000
          - path: /api/[1-9][0-9]*/(.*)
            pathType: ImplementationSpecific
            backend:
              service:
                name: sentry-relay
                port:
                  number: 3000
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: sentry-web
                port:
                  number: 9000
```

to

```yaml
kind: HTTPRoute
metadata:
  name: sentry
  namespace: sentry
spec:
  hostnames:
    - sentry.selfhosted.tld
  parentRefs:
      kind: Gateway
      name: eg
      sectionName: https
  rules:
    - backendRefs:
        - group: ''
          kind: Service
          name: sentry-relay
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /api/store
        - path:
            type: RegularExpression
            value: ^/api/[1-9][0-9]*/.*$
    - backendRefs:
        - group: ''
          kind: Service
          name: sentry-web
          port: 9000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /
```

The RegularExpression is never matched because the PathPrefixes takes precedence

Based on  `Note: The precedence of RegularExpression path matches are implementation-specific.` (<https://github.com/kubernetes-sigs/gateway-api/pull/1855>), I propose in this PR to changes the route precedence order from `Exact > PathPrefix > RegularExpression` to `Exact > RegularExpression > PathPrefix`

### DIFF
--- a/internal/gatewayapi/sort.go
+++ b/internal/gatewayapi/sort.go
@@ -47,6 +47,7 @@ func (x XdsIRRoutes) Less(i, j int) bool {
 			if x[j].PathMatch.SafeRegex != nil {
 				return true
 			}
+		}
 	}
 	// Equal case
 
--- a/internal/gatewayapi/sort.go
+++ b/internal/gatewayapi/sort.go
@@ -18,24 +18,18 @@ func (x XdsIRRoutes) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
 func (x XdsIRRoutes) Less(i, j int) bool {
 
 	// 1. Sort based on path match type
-	// Exact > PathPrefix > RegularExpression
+	// Exact > RegularExpression > PathPrefix
 	if x[i].PathMatch != nil && x[i].PathMatch.Exact != nil {
 		if x[j].PathMatch != nil {
-			if x[j].PathMatch.Prefix != nil {
-				return false
-			}
-			if x[j].PathMatch.SafeRegex != nil {
+			if x[j].PathMatch.Prefix != nil || x[j].PathMatch.SafeRegex != nil {
 				return false
 			}
 		}
 	}
 	if x[i].PathMatch != nil && x[i].PathMatch.Prefix != nil {
 		if x[j].PathMatch != nil {
-			if x[j].PathMatch.Exact != nil {
+			if x[j].PathMatch.Exact != nil || x[j].PathMatch.SafeRegex != nil {
 				return true
-			}
-			if x[j].PathMatch.SafeRegex != nil {
-				return false
 			}
 		}
 	}
@@ -45,7 +39,7 @@ func (x XdsIRRoutes) Less(i, j int) bool {
 				return true
 			}
 			if x[j].PathMatch.Prefix != nil {
-				return true
+				return false
 			}
 		}
 	}
@@ -96,11 +90,11 @@ func pathMatchCount(pathMatch *ir.StringMatch) int {
 		if pathMatch.Exact != nil {
 			return len(*pathMatch.Exact)
 		}
-		if pathMatch.Prefix != nil {
-			return len(*pathMatch.Prefix)
-		}
 		if pathMatch.SafeRegex != nil {
 			return len(*pathMatch.SafeRegex)
+		}
+		if pathMatch.Prefix != nil {
+			return len(*pathMatch.Prefix)
 		}
 	}
 	return 0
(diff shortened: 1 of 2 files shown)


zirain:
@arkodg I recall we discussed offline, should EG provide an API about matching order?

arkodg:
met @vixns at Kubecon EU who brought this up in person :) 
Lets gather some more info on why ingress-nginx and contour prioritize regex over prefix
@AliceProxy curious to know what Emissary does ?
@zirain do you know what Istio does ?

internal/gatewayapi/sort.go
```diff
@@ -18,25 +18,19 @@ func (x XdsIRRoutes) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
 func (x XdsIRRoutes) Less(i, j int) bool {
 
 	// 1. Sort based on path match type
-	// Exact > PathPrefix > RegularExpression
+	// Exact > RegularExpression > PathPrefix
 	if x[i].PathMatch != nil && x[i].PathMatch.Exact != nil {
 		if x[j].PathMatch != nil {
-			if x[j].PathMatch.Prefix != nil {
```
arkodg, line 24: prefer the existing way, it shows the order of precedence - easier to read

internal/gatewayapi/sort.go
```diff
@@ -18,25 +18,19 @@ func (x XdsIRRoutes) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
...
-			if x[j].PathMatch.Exact != nil {
+			if x[j].PathMatch.Exact != nil || x[j].PathMatch.SafeRegex != nil {
 				return true
 			}
-			if x[j].PathMatch.SafeRegex != nil {
-				return false
-			}
 		}
 	}
 	if x[i].PathMatch != nil && x[i].PathMatch.SafeRegex != nil {
```
arkodg, line 36: lets move this block above to show the precedence

arkodg:
finally took a look at the config in the PR where `PathPrefix`
```
            type: PathPrefix
            value: /
```
is used as a fallback

This implies that if EG wants to promote or allow this usage of `PathPrefix`, we need to change the precendance ordering to `Exact > RegularExpression > PathPrefix`  and ask users to build very strict and scoped regex matches

arkodg:
hey @vixns added some suggestions, mainly around code style for this function

arkodg:
should we also consider cherrypicking this into v1.0 ?

zirain:
xref: 

https://github.com/istio/istio/blob/454d155c00d51d2e11478f8e904266b3cbd9f81a/pilot/pkg/config/kube/gateway/conversion.go#L779

vixns:
Hi @arkodg, it was nice to met in Paris.
I've updated the code style as asked.

Yes this precedence change is mainly for catchalls, which are a very common and imo should be easy to read and write.
To illustrate this PR's motivation, how would you write the catchall from the HTTProute in this PR's description without changing the precedence order ?

arkodg (dismissed):
LGTM thanks !

arkodg:
@envoyproxy/gateway-maintainers should we back port this to v1.0 ?

arkodg (approved):

zirain (approved):

arkodg:
The client timeout e2e test seems to be failing

zirain:
> The client timeout e2e test seems to be failing

maybe we should disable it first, I can find out a way to make it stable now.

arkodg:
> > The client timeout e2e test seems to be failing
> 
> maybe we should disable it first, I can find out a way to make it stable now.

+1, thanks @zirain !

PR #2582, opened 2024-02-09
fix: when service type = clusterIP set the address to service.clusterIP

### LINKED ISSUE #2166: If Service Type is ClusterIP set spec.addresses.value into ClusterIP
When creating a GW using ClusterIP:

```yaml
kind: EnvoyProxy
metadata:
  name: proxy-config-internal
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        type: ClusterIP
```

There is no way to set the IP address to be used, if we use spec.adresses it will set the `externalIPs` not the `clusterIP`:
```yaml
kind: Gateway
metadata:
  name: eg-internal
spec:
  addresses:
    - type: IPAddress
      value: 10.43.146.30
```
```
NAME          CLASS         ADDRESS        PROGRAMMED   AGE
eg-internal   eg-internal   10.43.146.30   True         11m
```
But is in fact this will produce:
```
spec:
  clusterIP: 10.43.65.18
  clusterIPs:
  - 10.43.65.18
  externalIPs:
  - 10.43.146.30
  internalTrafficPolicy: Cluster
```
In a ClusterIP case, setting the clusterIP with spec.addresses.value would probably be more useful.

ShyunnY:
**What type of PR is this?**
fix:  when service type = clusterIP set the address to service.clusterIP 

This commit solves the problem that when setting Service Type=ClusterIP and customizing the address for Gateway.spec.addresses, the address is incorrectly set to externalIPs.

I fixed it so that it can be correctly set to the ClusterIP

Fixes #2166

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource_provider.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider.go
@@ -120,7 +120,18 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 	serviceSpec := resource.ExpectedServiceSpec(envoyServiceConfig)
 	serviceSpec.Ports = ports
 	serviceSpec.Selector = resource.GetSelector(labels).MatchLabels
-	serviceSpec.ExternalIPs = r.infra.Addresses
+
+	if (*envoyServiceConfig.Type) == egv1a1.ServiceTypeClusterIP {
+		if len(r.infra.Addresses) > 0 {
+			// Since K8s Service requires specify no more than one IP for each IP family
+			// So we only use the first address
+			// if address is not set, the automatically assigned clusterIP is used
+			serviceSpec.ClusterIP = r.infra.Addresses[0]
+			serviceSpec.ClusterIPs = r.infra.Addresses[0:1]
+		}
+	} else {
+		serviceSpec.ExternalIPs = r.infra.Addresses
+	}
 
 	svc := &corev1.Service{
 		TypeMeta: metav1.TypeMeta{
--- a/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
@@ -42,6 +42,13 @@ func newTestInfraWithAnnotations(annotations map[string]string) *ir.Infra {
 	return newTestInfraWithAnnotationsAndLabels(annotations, nil)
 }
 
+func newTestInfraWithAddresses(addresses []string) *ir.Infra {
+	infra := newTestInfraWithAnnotationsAndLabels(nil, nil)
+	infra.Proxy.Addresses = addresses
+
+	return infra
+}
+
 func newTestInfraWithAnnotationsAndLabels(annotations, labels map[string]string) *ir.Infra {
 	i := ir.NewInfra()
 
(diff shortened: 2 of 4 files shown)


internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -120,7 +120,17 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
...
 	serviceSpec.Ports = ports
 	serviceSpec.Selector = resource.GetSelector(labels).MatchLabels
-	serviceSpec.ExternalIPs = r.infra.Addresses
+
+	if (*envoyServiceConfig.Type) == egv1a1.ServiceTypeClusterIP {
+		if len(r.infra.Addresses) > 0 {
+			// we just need the first one address
+			// if address is not set, the automatically assigned clusterIP is used
+			serviceSpec.ClusterIP = r.infra.Addresses[0]
+			serviceSpec.ClusterIPs = r.infra.Addresses[0:1]
```
arkodg, line 129: should we assign all for `ClusterIPs` ?

arkodg:
we'll also need to update the logic in https://github.com/envoyproxy/gateway/blob/d44f4be1d60b778b8acefc2b67008a8947fa62a5/internal/status/gateway.go#L30 to make sure that the `ClusterIP` is updated in the status
the `if/else` will need to be refactored a bit

internal/infrastructure/kubernetes/proxy/resource_provider.go
ShyunnY: No, actually we can only set one clusterIPs, and clusterIPs[0] must be equal to clusterIP, and clusterIPs stipulates that each IP family can only set one IP.

**For example**: 
When we create the following Service resource and set multiple clusterIPs
```yaml
apiVersion: v1
kind: Service
metadata:
  name: envoy-custom
spec:
  clusterIP: 10.102.168.100
  clusterIPs:
    - 10.102.168.100
    - 10.102.168.101
  ports:
  - name: envoy-http
    port: 10009
    protocol: TCP
    targetPort: 10000
  selector:
    app.kubernetes.io/name: envoy
    gateway.envoyproxy.io/owning-gateway-name: eg-gw
  type: ClusterIP
```

Actually api-server throws `The Service "envoy-custom" is invalid: spec.clusterIPs: Invalid value: []string{"10.102.168.100", "10.102.168.101"}: may specify no more than one IP for each IP family` error

So I think that when configuring `envoyService.type=ClusterIP`, if the user specifies addresses when configuring the gateway, we only need to get the first address.

I will add more detailed comments to the code explaining this

ShyunnY:
> we'll also need to update the logic in
> 
> https://github.com/envoyproxy/gateway/blob/d44f4be1d60b778b8acefc2b67008a8947fa62a5/internal/status/gateway.go#L30
> 
> to make sure that the `ClusterIP` is updated in the status
> the `if/else` will need to be refactored a bit

Yes, I've made changes

shawnh2 (approved):
lgtm, thanks

one more thing, it's not recommended to merge code from your main branch, so next time please create a new branch, thanks

zirain (approved):

internal/infrastructure/kubernetes/proxy/resource_provider.go
arkodg: thanks for explaining this, TIL about why `ClusterIPs` exist :)

PR #2585, opened 2024-02-10
feat: Suppress 'X-Envoy' headers and pass-through the upstream 'Server' header by default

liorokman:
Implements the API introduced in #2500


arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM

Issue #2507, opened 2024-01-26
Security policy from different route interferes authentication


sadovnikov:
*Description*:

Reusing the same `Gateway`, we created two pairs of `HTTPRoute` with matching `SecurityPolicy`, which use identical IDP settings for OIDC. The HTTP Routes are on different hosts. Both hosts are CNAMEs to the same A-record, with the IP address of the Gateway service

| hostname | path | http route | security policy |
| --- | --- | --- | --- |
| reference-apps-twin.platform-staging.gcp.internal.xxx.org | /cbdp-k8s-sample-eg | cbdp-k8s-sample-web-eg | cbdp-k8s-sample-web-eg |
| reference-apps.platform-staging.gcp.internal.xxx.org | /cbdp-k8s-sample-pr-195 | cbdp-k8s-sample-web-pr-195 | cbdp-k8s-sample-web-pr-195 |

Opening of https://reference-apps.platform-staging.gcp.internal.xxx.org/cbdp-k8s-sample-pr-195 in an incognito tab goes through the login process and the application page gets opened.

However, the opening of https://reference-apps-twin.platform-staging.gcp.internal.xxx.org/cbdp-k8s-sample-eg/ in a new incognito tab gives "too many redirects" after the login page. This link starts working when the `cbdp-k8s-sample-web-pr-195` Security Policy, which belongs to a different HTTP Route, is deleted.

*Repro steps*:

* the `envoy-proxy-min-err.json` is the Envoy configuration with both Security Policies present
* the `envoy-proxy-min-ok.json` is the Envoy configuration without the `cbdp-k8s-sample-web-pr-195` Security Policy
* YAML of HTTP Routes and Security Policies will be added in the issue comment 

[envoy-proxy-min-err.json](https://github.com/envoyproxy/gateway/files/14065885/envoy-proxy-min-err.json)
[envoy-proxy-min-ok.json](https://github.com/envoyproxy/gateway/files/14065886/envoy-proxy-min-ok.json)

*Environment*:

Envoy Gateway is deployed using the `v0.0.0-latest` Helm chart. 

```
dependencies:
- name: gateway-helm
  version: v0.0.0-latest
  repository: oci://docker.io/envoyproxy
  alias: eg
```

Container 

```
Containers:
  envoy:
    Container ID:  containerd://b8648147d345bcf50d6f68aab09149ece2737c29d72fb0c6fb18115cb946699d
    Image:         envoyproxy/envoy-dev:latest
[...]

PR #2586, opened 2024-02-12
add a catch-all route if needed

zhaohuabing:
fix: https://github.com/envoyproxy/gateway/issues/2507


internal/gatewayapi/translator.go
```diff
@@ -201,12 +202,60 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
...
 	return newTranslateResult(gateways, httpRoutes, grpcRoutes, tlsRoutes,
 		tcpRoutes, udpRoutes, clientTrafficPolicies, backendTrafficPolicies,
 		securityPolicies, xdsIR, infraIR)
 
 }
 
+// For filters without native per-route support, we need to add a catch-all route
+// to ensure that these filters are disabled for non-matching requests.
+// https://github.com/envoyproxy/gateway/issues/2507
+func addCatchAllRoute(xdsIR map[string]*ir.Xds) {
```
arkodg, line 217: cc @mt-inside

internal/gatewayapi/translator.go
```diff
@@ -201,12 +202,60 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
...
+						len(r.QueryParamMatches) == 0 {
+						needCatchAllRoutePerHost[host] = false
+					}
+				}
+			}
+
+			for host, needCatchAllRoute := range needCatchAllRoutePerHost {
+				if needCatchAllRoute {
+					http.Routes = append(http.Routes, &ir.HTTPRoute{
+						Name: "catch-all",
```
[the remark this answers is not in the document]
zhaohuabing: Does the route name need to be unique across all hostnames?

internal/gatewayapi/translator.go
```diff
@@ -201,12 +202,60 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
...
+						PathMatch: &ir.StringMatch{
+							Prefix: ptr.To("/"),
+						},
```
arkodg, line 247: what happens when a prefix match of `/` already exists ? :)

arkodg: are we relying on first match / last match for this ?

zhaohuabing: > what happens when a prefix match of `/` already exists ? :)

We skip at line 242 

> Are we relying on first match / last match for this ?
Yes. The matches have been sorted, right?

internal/gatewayapi/translator.go
```diff
@@ -203,12 +207,61 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
...
+			// skip if there is already a catch-all route
+			for host := range needCatchAllRoutePerHost {
+				for _, r := range http.Routes {
+					if (r.Hostname == host &&
+						r.PathMatch != nil &&
+						r.PathMatch.Prefix != nil &&
+						*r.PathMatch.Prefix == "/") &&
+						len(r.HeaderMatches) == 0 &&
+						len(r.QueryParamMatches) == 0 {
+						needCatchAllRoutePerHost[host] = false
```
arkodg, line 241: `delete` key maybe better here

internal/gatewayapi/translator.go
```diff
@@ -203,12 +207,61 @@ func (t *Translator) Translate(resources *Resources) *TranslateResult {
...
+					}
+				}
+			}
+
+			for host, needCatchAllRoute := range needCatchAllRoutePerHost {
+				if needCatchAllRoute {
+					underscoredHost := strings.ReplaceAll(host, ".", "_")
+					http.Routes = append(http.Routes, &ir.HTTPRoute{
+						Name: fmt.Sprintf("%s/catch_all", underscoredHost),
```
arkodg, line 250:
```suggestion
						Name: fmt.Sprintf("%s/catch-all-return-404", underscoredHost),
```

internal/gatewayapi/translator.go
zhaohuabing: Hah, it's good to know that it's safe to delete a key within range in golang. :-)

arkodg (approved):
LGTM thanks @zhaohuabing !

PR #2605, opened 2024-02-14
api: Support Timeouts in ClientTrafficPolicy

### LINKED ISSUE #2598: Add timeouts support to clientTrafficPolicy
*Description*:
Add `requestTimeout` and `requestHeadersTimeout` to clientTrafficPolicy.

According to Envoy documentation, These fields *should be configured in the presence of untrusted downstreams*.

Additionally, according to the [Envoy best practices](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy), configuring the request_timeout is recommended. It is emphasized that this setting "must be disabled for long-lived and streaming requests". hence, instead of enabling it by default, I propose making this field configurable.

These fields weren't previously discussed in [this issue](https://github.com/envoyproxy/gateway/issues/1048) because `requestHeadersTimeout` was not explicitly mentioned in the best-practices document, and `requestTimeout` wasn't enforced by default to avoid potential disruptions to certain clients (as discussed above).

*Links*:
- `request_timeout` and `request_headers_timeout` in [Envoy Docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-request-headers-timeout) 
- [Envoy best practices](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy)

yaelSchechter:
This PR will allow users to configure timeouts for client connections in the `ClientTrafficPolicy`. 

This is implemented as a struct to allow future addition of related fields (such as `request_headers_timeout` and `idle_timeout`).

Naming rationale:  `requestReceivedTimeout` is the envoy's `request_timeout`. This name indicates that the timeout includes the time it took for the upstream to receive the request. 

**Example**:
```yaml
spec:
  timeout:
    http:
      requestReceivedTimeout: 300s
```
 
Fixes #2598


guydc (approved):
Aligned with previous discussion in the community meeting. 
LGTM.

liorokman (approved):
LGTM

davidalger (changes requested):
Perhaps this wasn't meant to be a finished work when pushed, but the added fields here aren't used anywhere and don't do anything. They need translation into XDS that's added to the listener as well as test coverage.

If an example is needed, I recently made a similar addition to the ClientTrafficPolicy type that might help point you to the right places: https://github.com/envoyproxy/gateway/pull/2535/files

Pay specific attention to `internal/gatewayapi/clienttrafficpolicy.go` and `internal/ir/xds.go` for translation as well as `internal/xds/translator/listener.go` where it's injected into `HttpConnectionManager` settings.

liorokman:
@davidalger features that change APIs are usually split into two PRs, the first one to agree upon the API itself and the second one to implement the feature. For example, #2534 and #2577 , or #2487 and #2492.

davidalger:
> features that change APIs are usually split into two PRs, the first one to agree upon the API itself and the second one to implement the feature

How did I miss this? 😮 Nvm me then 🙈

arkodg:
thanks for pointing out the ideal workflow @liorokman (GH issue -> API PR -> Implementation PR) !
@davidalger thanks for proactively helping out other contributors !

api/v1alpha1/timeout_types.go
```diff
@@ -41,3 +41,18 @@ type HTTPTimeout struct {
...
+	// +optional
+	HTTP *HTTPClientTimeout `json:"http,omitempty"`
+}
+
+type HTTPClientTimeout struct {
+	// The timeout for processing the client request. Starting upon request initiation and ending when the last byte is sent to backend.
+	// Default: 300 seconds.
+	//
+	// +optional
+	RequestProcessTimeout *gwapiv1.Duration `json:"requestProcessTimeout,omitempty"`
```
arkodg, line 57: `requestProceess` doesnt convey the exact meaning of the knob imo
thoughts on `processingRequestTimeout` or even `requestTimeout` ?

davidalger: +1 for `requestTimeout` with a descriptive comment indicating something closer to what the docs describe:

> The amount of time that Envoy will wait for the entire request to be received. The timer is activated when the request is initiated, and is disarmed when the last byte of the request is sent upstream OR when the response is initiated.

Alternately something like `requestRecievedTimeout` seems less ambiguous than any variation of `process` timeout and conveys the intent while differentiating this from the (upstream) request timeout which can be specified via timeouts.request on the HTTPRoute

arkodg: +1 for `requestRecievedTimeout`

guydc: `requestTimeout` is slightly overloaded as gateway-api already defines a structure like `timeout.request` for [httproutes](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteTimeouts). 

I'm  fine with `requestRecievedTimeout` or just `requestTimeout` if we're not concerned that this would be confusing (due to the other GW API field).

yaelSchechter: Thanks for the feedback, I changed the field name to `requestRecievedTimeout` and added a more descriptive comment as @davidalger suggested.

guydc (approved):

arkodg (dismissed):
LGTM thanks !

arkodg (approved):

zirain (approved):

Issue #2452, opened 2024-01-16
Support Routing based on JWT Claims


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Users would like to route to a specific backend by matching on JWT claims

Looks like a popular issue raised in upstream as well https://github.com/kubernetes-sigs/gateway-api/issues/920 but was closed due to lack of ownership

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
* Some limitations we have from an Upstream API perspective -  there is no first class field for `claims` within the `matches` field within `HTTPRoute`, this would be made it very easy for the user to author this intent
* Envoy Gateway has a feature `claimToHeaders` https://gateway.envoyproxy.io/v0.6.0/api/extension_types/#claimtoheader
 to convert JWT claims to HTTP Headers which can be used for header matching, but from an implementation perspective JWT filter is applied at the route level, post route match. To support above use case, we could
   * automagically move the jwt filter to the hcm level if `claimToHeaders` is set
   * add a new field under `jwt` called `useForRouting` which also moves the jwt filter to the listener level

arkodg:
there's a `clear_route_cache` option in the JWT filter config https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider which clears the route decision and recomputes it, but based on the info
```
Clears route cache in order to allow JWT token to correctly affect routing decisions. Filter clears all cached routes when:

The field is set to true.
At least one claim_to_headers header is added to the request OR if payload_in_metadata is set.
```
it doesnt look like it needs to be set

tmsnan:
@arkodg, Its comments may be ambiguous, requiring clear_route_cache to be set.
https://github.com/envoyproxy/envoy/blob/1f1c32083150dced241336a6e2734e05bcead4b8/source/extensions/filters/http/jwt_authn/authenticator.cc#L380C1-L382C4

arkodg:
nice find @tmsnan, does that logic need to be updated to
```
provider.clear_route_cache() || .......
``` 
?

tmsnan:
No changes needed @arkodg. Recalculating routes may cause performance issues. In some scenarios, only header or metadata information needs to be added without requiring rerouting, such as ratelimit-based JWT claims. 
Envoy comment needs to be updated.

[...]

PR #2612, opened 2024-02-14
api: recomputeRoute field in JWT

arkodg:
Add a field called `useForRouting` that signals to Envoy Gateway that the headers generated from the claims are used to make routing decisions

Internally this field will be used to
~~* insert a catch-all route with a 404 direct response identical to https://github.com/envoyproxy/gateway/pull/2586 which makes sure the jwt filter with `claimToHeader` is applied before recomputing routing decision~~
* enable `clear_route_cache` to recompute routing decision https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider

Relates to https://github.com/envoyproxy/gateway/issues/2452


arkodg:
outlining a few other alternatives
* add a similar field at top level within `jwt` instead of creating it within `jwt.claimToHeaders` , which will have the same effect internally, since these settings internally are applied at top level
* automatically enable these settings w/o an API field if `jwt.claimToHeaders` is set. One downside here is, enabling `clearRouteCache` is expensive and may impact performance. This may not be ideal for a user who wants to `jwt.claimToHeaders` for ratelimiting

api/v1alpha1/jwt_types.go
```diff
@@ -85,6 +85,10 @@ type ClaimToHeader struct {
 	// (eg. "claim.nested.key", "sub"). The nested claim name must use dot "."
 	// to separate the JSON name path.
 	Claim string `json:"claim"`
+
+	// UseForRouting must be enabled if this header generated from the claim should be used for
+	// route matching decisions
+	UseForRouting *bool `json:"useForRouting,omitempty"`
```
shawnh2, line 91: need a `// +optional` comment for this

arkodg: thanks for catching this !

shawnh2 (approved):
lgtm

yuluo-yx (approved):
LGTM!

arkodg:
ptal @guydc @zhaohuabing, I've limited the use of the field to only recalculate route, and not add a fallback/catch-all route, 
we can use docs to add that manually

arkodg:
also looking at advice on API field naming, this is what ChatGPT said

```
Here are some alternate API field names for recomputeRoute:

calculateRoute
updateRoute
computeNewRoute
refreshRoute
recalculateRoute
```

api/v1alpha1/jwt_types.go
```diff
@@ -52,8 +52,16 @@ type JWTProvider struct {
...
 	// The claim must be of type; string, int, double, bool. Array type claims are not supported
 	//
+	// +optional
 	ClaimToHeaders []ClaimToHeader `json:"claimToHeaders,omitempty"`
 
+	// RecomputeRoute clears the route cache and recalculates the routing decision.
+	// This field must be enabled if the headers generated from the claim are used for
+	// route matching decisions.
+	//
+	// +optional
```
guydc, line 62: Do we need to state something about the impact on other features? e.g. filters that fired prior to jwt (CORS, ExtAuth, Basic, ...) will only execute in the context of the initial match?

guydc (commented):
+1 to `updateRoute`

api/v1alpha1/jwt_types.go
```diff
@@ -52,8 +52,16 @@ type JWTProvider struct {
...
+	RecomputeRoute *bool `json:"recomputeRoute,omitempty"`
```
zhaohuabing, line 63: Should we verify that when recomputeRoute is enabled, the length of claimToHeaders is not zero?

zhaohuabing:
recomputeRoute sounds good to me, "updateRoute" is a bit ambiguous.

api/v1alpha1/jwt_types.go
arkodg: yes !

zirain (approved):

api/v1alpha1/jwt_types.go
zirain: can you update cel validation?

api/v1alpha1/jwt_types.go
arkodg: added a comment around this

api/v1alpha1/jwt_types.go
arkodg: done ! thanks @zhaohuabing for helping me with my first CEL expression :)

zhaohuabing (approved):
LGTM. Thanks!

Issue #2572, opened 2024-02-07
Support Filters in BackendRefs


MregXN:
I have encountered difficulties while using the Kubernetes Gateway API Filters under BackendRefs. Is the current envoy gateway already supported?

arkodg:
@MregXN this is not supported, highlighted in https://gateway.envoyproxy.io/latest/user/gatewayapi-support/#httproute
keeping this issue open to we can highlight this using a status condition

MregXN:
I think I can help in making this issue achievable. 
But it seems that the fields of `ir.HTTPRoute` are not adequate for Filters in BackendRefs support because all filters will be translatrd into `Redirect` filed and there is no way to differentiate based on backendref. Should we need to some how extend the new field to  `ir.HTTPRoute` ? @arkodg @cnvergence

arkodg:
@MregXN can you share your use case of why you'd prefer enabling the filter on the backendRef instead of the route

MregXN:
@arkodg 
I have two backend services pointing to the same hostname and path, with different weights set to distribute access traffic proportionally to the corresponding backend services. However, I now need to apply forwarding operations to one of the backend services only. Therefore, I need to use a filter that only affects one backend. Using the route will affect both backends simultaneously.

arkodg:
@MregXN by forwarding operations, do you mean appending headers to the request or response ?

MregXN:
@arkodg 
It has nothing to do with the header. I hope the user can receive the 302 status code when their traffic is routed to the specific backend.

baildagq:
@MregXN hi, is there any process?

Xunzhuo:
I think this is a reasonable request for filter implementation, thoughts on this @arkodg ?

MregXN:
> @MregXN hi, is there any process?

I haven't started as I think we haven't reached a consensus yet.

dprotaso:
BackendRef Filters are needed by Knative - https://github.com/knative-extensions/net-gateway-api/issues/566

We specifically append headers after a traffic split - this enables our scale to zero feature as well as our ability to handle 'burst capacity'

arkodg:
I miscalculated the effort to get this in :( 
The implementation seems straightforward
[...]

PR #2620, opened 2024-02-15
fix: add unsupported status condition for filters within BackendRef

cnvergence:
**What type of PR is this?**

- highlight unsupported BackendRef filters using a status condition
- introduce BackendRefContext in translator

Fixes 
- https://github.com/envoyproxy/gateway/issues/2572

### DIFF
--- a/internal/gatewayapi/contexts.go
+++ b/internal/gatewayapi/contexts.go
@@ -417,3 +417,41 @@ func (r *RouteParentContext) HasCondition(route RouteContext, condType gwapiv1.R
 	}
 	return false
 }
+
+// BackendRefContext represents a generic BackendRef object that can reference Gateway objects.
+type BackendRefContext struct {
+	HTTPBackendRef *gwapiv1.HTTPBackendRef
+	GRPCBackendRef *v1alpha2.GRPCBackendRef
+	BackendRef     *gwapiv1.BackendRef
+}
+
+// GetBackendRef returns the BackendRef object from BackendRefContext matching route Kind.
+func (b *BackendRefContext) GetBackendRef(routeKind gwapiv1.Kind) *gwapiv1.BackendRef {
+	switch routeKind {
+	case KindHTTPRoute:
+		if b.HTTPBackendRef != nil {
+			return &b.HTTPBackendRef.BackendRef
+		}
+	case KindGRPCRoute:
+		if b.GRPCBackendRef != nil {
+			return &b.GRPCBackendRef.BackendRef
+		}
+	default:
+		return b.BackendRef
+	}
+	return nil
+}
+
+func (b *BackendRefContext) GetHTTPFilters() []gwapiv1.HTTPRouteFilter {
+	if b.HTTPBackendRef != nil {
+		return b.HTTPBackendRef.Filters
+	}
+	return nil
+}
+
+func (b *BackendRefContext) GetGRPCFilters() []v1alpha2.GRPCRouteFilter {
+	if b.GRPCBackendRef != nil {
+		return b.GRPCBackendRef.Filters
+	}
+	return nil
+}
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/route.go
```diff
@@ -173,6 +173,15 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 		dstAddrTypeMap := make(map[ir.DestinationAddressType]int)
 
 		for _, backendRef := range rule.BackendRefs {
+			if len(backendRef.Filters) > 0 {
```
arkodg, line 176: should we `continue`, we have been failing fast and early in other places

internal/gatewayapi/route.go
```diff
@@ -173,6 +173,15 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
...
+				parentRef.SetCondition(httpRoute,
+					gwapiv1.RouteConditionResolvedRefs,
+					metav1.ConditionFalse,
+					"UnsupportedRefValue",
+					"The filters field within BackendRef is not supported",
+				)
+			}
```
arkodg, line 183: probably also need to add this for tcproute, tlsroute and udproute ?

cnvergence: thought so, but they don't have it https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.BackendRef

internal/gatewayapi/route.go
cnvergence: we could, let's fail fast then on this

internal/gatewayapi/route.go
arkodg: thanks for checking !

internal/gatewayapi/route.go
arkodg: rethinking again, this code should live inside [validateBackendRef](https://github.com/envoyproxy/gateway/blob/ac0243749b01d1d60b32bae88cb0effb71b61102/internal/provider/kubernetes/helpers.go#L215) which gets called here
https://github.com/envoyproxy/gateway/blob/ac0243749b01d1d60b32bae88cb0effb71b61102/internal/gatewayapi/route.go#L1080

which automatically converts the `weight` into a `invalidWeight` ensuring that a percentage of traffic gets 503d

cnvergence: looks good, but the field Filters is within gwapiv1.HTTPBackendRef or gwapiv1a1.GRPCBackendRef.
we process BackendRef based on gwapiv1.BackendRef

arkodg: we've had to do something similar for xRoutes, so we created `RouteContext`
https://github.com/envoyproxy/gateway/blob/cc88e9ccd486d14af48138f733ad70ee61a581c6/internal/gatewayapi/contexts.go#L155

may need something similar here

cnvergence: if we want to extend it, then sure
thanks for explaining

internal/gatewayapi/contexts.go
```diff
@@ -417,3 +417,27 @@ func (r *RouteParentContext) HasCondition(route RouteContext, condType gwapiv1.R
 	}
 	return false
 }
+
+// BackendRefContext represents a generic BackendRef object that can reference Gateway objects.
```
arkodg, line 421: id suggest this be an interface with 2 methods
* GetBackendRef()
* GetFilters()

and the implementation could look similar to https://github.com/envoyproxy/gateway/blob/c5cd964da2f2ec1951ba72d45b54eef8246e96ee/internal/gatewayapi/contexts.go#L216
```


e.g. reflect.ValueOf(route).Elem().FieldByName("BackendRef")

cnvergence: I think the struct would be still needed or am I wrong?

arkodg (approved):
LGTM, thanks for incorporating the feedback

cnvergence:
sure, found the context in https://github.com/envoyproxy/gateway/issues/1252

arkodg (approved):

shawnh2 (approved):

Issue #2610, opened 2024-02-14
E2E Zero DownTime Test when upgrading Envoy Proxy Versions


arkodg:
I executed a naive test:
- Environment: kind, metallb, EG quickstart.yaml
- envoy proxy replicas: 2
- upgrade: 0.6.0 => 0.0.0-latest using `helm upgrade`
- load simulation during upgrade: `hey -c 100 -q 10 -z 300s -host www.example.com http://172.18.255.200/ `  

The upgrade caused some client-facing failures during the test:

```
Error distribution:
  [8]	Get "http://172.18.255.200/": EOF
  [32]	Get "http://172.18.255.200/": dial tcp 172.18.255.200:80: connect: connection refused
  [1]	Get "http://172.18.255.200/": read tcp 172.18.0.1:55220->172.18.255.200:80: read: connection reset by peer
  [1]	Get "http://172.18.255.200/": read tcp 172.18.0.1:55260->172.18.255.200:80: read: connection reset by peer
```
It's probably possible to tune some of the parameters mentioned in my previous comment to achieve a hitless upgrade under certain test conditions (RPS, connection reuse, HTTP version, ...). But, I'm not sure that we can claim to have a hitless upgrade in general, based on such test. 

So, I propose that for the GA scope, we focus on an upgrade test that ensures request convergence to successful execution after the upgrade. A limited hitless upgrade test can be a stretch-goal. 

In the future, we can explore:
- Implementing a graceful envoy shutdown feature and providing guidance on configuring envoy for hitless in-place upgrades
- Supporting canary deployments 

WDYT?

_Originally posted by @guydc in https://github.com/envoyproxy/gateway/issues/1712#issuecomment-1944561296_

guydc:
@arkodg - I think that this test can really focus on validating the functionality implemented in #2633. Meaning, we will just test a restart of a multi-replica proxy deployment, without also upgrading EG/Envoy in the process. 

We have another task for validating the entire upgrade process holistically in #1710. Here, we may run into failures due to other reasons (e.g. control plane unavailability for new envoy instances). 

Are you ok with the proposed scoping?

Issue #1712, opened 2023-07-26
Handle EnvoyProxy Image version upgrades


arkodg:

arkodg:
thanks @cnvergence !
Thinking out loud, an outcome of this issue could be a E2E where all client requests are successful while 2 replicas of Envoy Proxy are undergoing a rolling restart.

cnvergence:
@arkodg coming back to this after a while, could you please point me to where should I start?
I did check the upgrade and it seems like it is handled, but I may be wrong.

As for the E2E, I assume I should add a new scenario to the e2e test suite :)

arkodg:
hey I know @chauhanshubham was looking into a similar test for control plane upgrades which would invariably also upgrade envoy proxy, should we just collapse those two e2e tests into one where we perform an upgrade with a last known EG minor version, and ensure that
1. there is no config churn in the data plane during an upgrade
2. there is no traffic drop while this upgrade this happens (assuming we always have 2 replicas of control plane and data plane running)

guydc:
I'm concerned that a hitless in-place upgrade of envoy is not trivial. 

A graceful termination of envoy and may require:
- Failing LB/Kubelet probes to stop new connection from being established to terminating pods
- Triggering envoy to drain listeners 
- Delaying pod termination until all connections are terminated
- Additional factors to consider:
  - IaaS LBs behave differently in response to targets becoming unhealthy (e.g. AWS will reset all existing connections, while GCP/Azure will stop establishing new connections but retain existing connections) and have varying levels of configurability for HCs. 
  - The number of envoy pods per node and LB's `ExternalTrafficPolicy` impact the correctness of LB HCs. 

It's also important to avoid race conditions where a new instance of envoy is receiving traffic before it was configured (e.g. due to order of component restart, failures in new control plane version, etc.). 

Some prior art: 
- Contour: https://github.com/projectcontour/contour/blob/main/design/envoy-shutdown.md
- Gloo Edge: https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/

guydc:
@arkodg 

I executed a naive test:
- Environment: kind, metallb, EG quickstart.yaml
- envoy proxy replicas: 2
- upgrade: 0.6.0 => 0.0.0-latest using `helm upgrade`
- load simulation during upgrade: `hey -c 100 -q 10 -z 300s -host www.example.com http://172.18.255.200/ `  

[...]

PR #2633, opened 2024-02-16
feat: Gracefully drain listeners before envoy shutdown on pod termination

davidalger:
**What this PR does**:

This PR is a step in the direction of achieving a hitless upgrade. Some challenges to achieving this have been outlined by @guydc in https://github.com/envoyproxy/gateway/issues/1712#issuecomment-1943985961 and https://github.com/envoyproxy/gateway/issues/2610

This proposed implementation of a PreStop lifecycle hook will allow Envoy proxies to gracefully shutdown when pods are terminated by Kubernetes. This is done by:

a) Initiating a graceful drain sequence via the [drain_listeners?graceful](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#operations-admin-interface-drain) admin endpoint.

b) Sleeping for a minimum drain duration of 15 seconds allowing ample time for endpoint deprogramming to occur.

c) Monitors `total.server_connections` statistic exiting once total connection count reaches `0`

d) Times out after 600 seconds after which Envoy will be sent SIGTERM by Kubernetes regardless of how many open connections remain. This timeout lines up with the default value for [`--drain-time-s`](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-drain-time-s)

e) Configures `terminationGracePeriodSeconds` to 900 seconds. This provides ample time for Envoy to cleanly exit after connections have been drained.

**Some prior art w.r.t graceful shutdown for Envoy**:

* Contour: https://github.com/projectcontour/contour/blob/main/design/envoy-shutdown.md
* Gloo Edge: https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/
* Contour: https://www.reddit.com/r/RedditEng/comments/1aqsxqf/proper_envoy_shutdown_in_a_kubernetes_world
* Istio: https://github.com/istio/istio/blob/master/pkg/envoy/agent.go#L151-L152

**Summary of why we need this:**

* Kubernetes uses SIGTERM to signal processes to gracefully shutdown, however Envoy [does not support](https://github.com/envoyproxy/envoy/issues/7841) gracefully draining connections during shutdowns triggered by SIGTERM.
* Endpoint deprogramming is eventually consistent and the time this takes varies somewhat based on the k8s implementation and size of the cluster.

**Caveats**:

When TCP listeners are used (rather than HTTP listeners), connections will not drain per the below note from draining architecture documentation. This means in the event of long-living TCP connections, Envoy proxies will simply wait in the PreStop hook for the 600 second timeout to occur after which these connections will be closed when Envoy receives the SIGTERM from Kubernetes.

> Note that although draining is a per-listener concept, it must be supported at the network filter level. Currently the only filters that support graceful draining are [Redis](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/redis_proxy_filter#config-network-filters-redis-proxy), [Mongo](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/mongo_proxy_filter#config-network-filters-mongo-proxy), [Thrift](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/thrift_proxy_filter#config-network-filters-thrift-proxy) (if the envoy.reloadable_features.thrift_connection_draining runtime feature is enabled), and [HTTP connection manager](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/http_conn_man#config-http-conn-man).

Reference: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/draining


davidalger:
**Testing / Load generator with unaltered EG `quickstart.yaml`:**

_run from another pod within the cluster_
```shell
hey -c 100 -q 10 -z 300s -host www.example.com http://172.18.255.200/
```

<details>
  <summary><code>PreStop logs</code> from rolling restart of single replica deployment</summary>

```
$ kubectl logs -n envoy-gateway-system deploy/envoy-default-eg-e41e7b31 -f | grep -v start_time
Found 2 pods, using pod/envoy-default-eg-e41e7b31-85c848c76b-w8p5t
[2024-02-16 21:23:58.508][1][warning][main] [source/server/server.cc:923] There is no configured limit to the number of allowed active downstream connections. Configure a limit in `envoy.resource_monitors.downstream_connections` resource monitor.
[2024-02-16 21:24:21.965][37][info][PreStop] initiating graceful drain with 15 second minimum drain period and 600 second timeout
[2024-02-16 21:24:21.969][37][info][PreStop] total connections: 100
[2024-02-16 21:24:22.974][37][info][PreStop] total connections: 100
[2024-02-16 21:24:23.978][37][info][PreStop] total connections: 97
[2024-02-16 21:24:24.983][37][info][PreStop] total connections: 97
[2024-02-16 21:24:25.988][37][info][PreStop] total connections: 97
[2024-02-16 21:24:26.992][37][info][PreStop] total connections: 97
[2024-02-16 21:24:27.997][37][info][PreStop] total connections: 97
[2024-02-16 21:24:29.001][37][info][PreStop] total connections: 64
[2024-02-16 21:24:30.005][37][info][PreStop] total connections: 64
[2024-02-16 21:24:31.010][37][info][PreStop] total connections: 64
[2024-02-16 21:24:32.014][37][info][PreStop] total connections: 64
[2024-02-16 21:24:33.018][37][info][PreStop] total connections: 64
[2024-02-16 21:24:34.023][37][info][PreStop] total connections: 29
[2024-02-16 21:24:35.027][37][info][PreStop] total connections: 29
[2024-02-16 21:24:36.032][37][info][PreStop] total connections: 29
[2024-02-16 21:24:37.036][37][info][PreStop] total connections: 29
[2024-02-16 21:24:37.037][37][info][PreStop] minimum drain period reached; allowed to exit when total connections reaches zero
[2024-02-16 21:24:38.042][37][info][PreStop] total connections: 29
[2024-02-16 21:24:39.046][37][info][PreStop] total connections: 7
[2024-02-16 21:24:40.050][37][info][PreStop] total connections: 7
[2024-02-16 21:24:41.055][37][info][PreStop] total connections: 7
[2024-02-16 21:24:42.059][37][info][PreStop] total connections: 7
[2024-02-16 21:24:43.064][37][info][PreStop] total connections: 7
[2024-02-16 21:24:44.068][37][info][PreStop] total connections: 1
[2024-02-16 21:24:45.073][37][info][PreStop] total connections: 1
[2024-02-16 21:24:46.077][37][info][PreStop] total connections: 1
[2024-02-16 21:24:47.082][37][info][PreStop] total connections: 1
[2024-02-16 21:24:48.086][37][info][PreStop] total connections: 1
[2024-02-16 21:24:49.091][37][info][PreStop] total connections: 0
[2024-02-16 21:24:49.096][1][warning][main] [source/server/server.cc:895] caught ENVOY_SIGTERM
[2024-02-16 21:24:49.100][1][warning][config] [./source/extensions/config_subscription/grpc/grpc_stream.h:155] DeltaAggregatedResources gRPC config stream to xds_cluster closed: 13, 
```

</details>

After restarting the single replica deployment, Envoy was scaled up to 5 replicas and restarted twice with the `hey` load generator continuing to run.


<details>
  <summary><code>load generator</code> output demonstrating 100% success responses for the test</summary>

```shell
$ hey -c 100 -q 10 -z 300s -host www.example.com http://172.18.255.200/

Summary:
  Total:	300.0754 secs
  Slowest:	0.4689 secs
  Fastest:	0.0012 secs
  Average:	0.1215 secs
  Requests/sec:	812.2893
  
  Total data:	121874000 bytes
  Size/request:	500 bytes

Response time histogram:
  0.001 [1]	|
  0.048 [3263]	|■
  0.095 [4661]	|■
  0.142 [222708]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.188 [4640]	|■
  0.235 [6894]	|■
  0.282 [990]	|
  0.329 [370]	|
  0.375 [141]	|
  0.422 [66]	|
  0.469 [14]	|


Latency distribution:
  10% in 0.1130 secs
  25% in 0.1168 secs
  50% in 0.1196 secs
  75% in 0.1224 secs
  90% in 0.1264 secs
  95% in 0.1520 secs
  99% in 0.2142 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0000 secs, 0.0012 secs, 0.4689 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0006 secs, 0.0000 secs, 0.1484 secs
  resp wait:	0.0015 secs, 0.0002 secs, 0.2476 secs
  resp read:	0.0594 secs, 0.0000 secs, 0.3291 secs

Status code distribution:
  [200]	243748 responses



```

</details>

davidalger:
**Testing / Load generator with EG `quickstart.yaml` plus 1s delay:**

_run from another pod within the cluster_
```
hey -c 20 -q 10 -z 300s -host www.example.com http://172.18.255.200/
```

<details>
  <summary><code>fault-injection-delay</code> simulating 1s response times</summary>

```yaml
kind: BackendTrafficPolicy
metadata:
  name: fault-injection-delay
spec:
  targetRef:
    kind: HTTPRoute
    name: backend
  faultInjection:
    delay:
      fixedDelay: 1s
```

</details>


<details>
  <summary><code>PreStop logs</code> from rolling restart of single replica deployment</summary>

```
$ kubectl logs -n envoy-gateway-system deploy/envoy-default-eg-e41e7b31 -f | grep -v start_time
Found 2 pods, using pod/envoy-default-eg-e41e7b31-7854bd848-qpd82
[2024-02-16 21:31:03.157][1][warning][main] [source/server/server.cc:923] There is no configured limit to the number of allowed active downstream connections. Configure a limit in `envoy.resource_monitors.downstream_connections` resource monitor.
[2024-02-16 21:33:14.274][37][info][PreStop] initiating graceful drain with 15 second minimum drain period and 600 second timeout
[2024-02-16 21:33:14.278][37][info][PreStop] total connections: 20
[2024-02-16 21:33:15.286][37][info][PreStop] total connections: 20
[2024-02-16 21:33:16.308][37][info][PreStop] total connections: 20
[2024-02-16 21:33:17.327][37][info][PreStop] total connections: 20
[2024-02-16 21:33:18.347][37][info][PreStop] total connections: 19
[2024-02-16 21:33:19.353][37][info][PreStop] total connections: 19
[2024-02-16 21:33:20.374][37][info][PreStop] total connections: 19
[2024-02-16 21:33:21.395][37][info][PreStop] total connections: 19
[2024-02-16 21:33:22.405][37][info][PreStop] total connections: 19
[2024-02-16 21:33:23.426][37][info][PreStop] total connections: 19
[2024-02-16 21:33:24.443][37][info][PreStop] total connections: 19
[2024-02-16 21:33:25.484][37][info][PreStop] total connections: 19
[2024-02-16 21:33:26.498][37][info][PreStop] total connections: 19
[2024-02-16 21:33:27.517][37][info][PreStop] total connections: 19
[2024-02-16 21:33:28.539][37][info][PreStop] total connections: 16
[2024-02-16 21:33:29.568][37][info][PreStop] total connections: 16
[2024-02-16 21:33:30.594][37][info][PreStop] total connections: 16
[2024-02-16 21:33:30.598][37][info][PreStop] minimum drain period reached; allowed to exit when total connections reaches zero
[2024-02-16 21:33:31.608][37][info][PreStop] total connections: 16
[2024-02-16 21:33:32.629][37][info][PreStop] total connections: 16
[2024-02-16 21:33:33.656][37][info][PreStop] total connections: 16
[2024-02-16 21:33:34.681][37][info][PreStop] total connections: 16
[2024-02-16 21:33:35.707][37][info][PreStop] total connections: 16
[2024-02-16 21:33:36.729][37][info][PreStop] total connections: 16
[2024-02-16 21:33:37.848][37][info][PreStop] total connections: 16
[2024-02-16 21:33:38.862][37][info][PreStop] total connections: 14
[2024-02-16 21:33:39.889][37][info][PreStop] total connections: 14
[2024-02-16 21:33:40.900][37][info][PreStop] total connections: 14
[2024-02-16 21:33:41.914][37][info][PreStop] total connections: 14
[2024-02-16 21:33:42.926][37][info][PreStop] total connections: 14
[2024-02-16 21:33:43.943][37][info][PreStop] total connections: 11
[2024-02-16 21:33:44.948][37][info][PreStop] total connections: 11
[2024-02-16 21:33:45.958][37][info][PreStop] total connections: 11
[2024-02-16 21:33:46.969][37][info][PreStop] total connections: 11
[2024-02-16 21:33:47.992][37][info][PreStop] total connections: 11
[2024-02-16 21:33:49.007][37][info][PreStop] total connections: 10
[2024-02-16 21:33:50.026][37][info][PreStop] total connections: 10
[2024-02-16 21:33:51.048][37][info][PreStop] total connections: 10
[2024-02-16 21:33:52.055][37][info][PreStop] total connections: 10
[2024-02-16 21:33:53.068][37][info][PreStop] total connections: 10
[2024-02-16 21:33:54.094][37][info][PreStop] total connections: 8
[2024-02-16 21:33:55.101][37][info][PreStop] total connections: 8
[2024-02-16 21:33:56.109][37][info][PreStop] total connections: 8
[2024-02-16 21:33:57.121][37][info][PreStop] total connections: 8
[2024-02-16 21:33:58.134][37][info][PreStop] total connections: 8
[2024-02-16 21:33:59.147][37][info][PreStop] total connections: 7
[2024-02-16 21:34:00.166][37][info][PreStop] total connections: 7
[2024-02-16 21:34:01.184][37][info][PreStop] total connections: 7
[2024-02-16 21:34:02.206][37][info][PreStop] total connections: 7
[2024-02-16 21:34:03.328][37][info][PreStop] total connections: 7
[2024-02-16 21:34:04.341][37][info][PreStop] total connections: 4
[2024-02-16 21:34:05.462][37][info][PreStop] total connections: 4
[2024-02-16 21:34:06.505][37][info][PreStop] total connections: 4
[2024-02-16 21:34:07.625][37][info][PreStop] total connections: 4
[2024-02-16 21:34:08.652][37][info][PreStop] total connections: 1
[2024-02-16 21:34:09.678][37][info][PreStop] total connections: 1
[2024-02-16 21:34:10.687][37][info][PreStop] total connections: 1
[2024-02-16 21:34:11.710][37][info][PreStop] total connections: 1
[2024-02-16 21:34:12.731][37][info][PreStop] total connections: 1
[2024-02-16 21:34:13.860][37][info][PreStop] total connections: 0
[2024-02-16 21:34:13.880][1][warning][main] [source/server/server.cc:895] caught ENVOY_SIGTERM
[2024-02-16 21:34:13.886][1][warning][config] [./source/extensions/config_subscription/grpc/grpc_stream.h:155] DeltaAggregatedResources gRPC config stream to xds_cluster closed: 13, 
```

</details>

After restarting the single replica deployment, Envoy was scaled up to 5 replicas and restarted twice with the `hey` load generator continuing to run.


<details>
  <summary><code>load generator</code> output demonstrating 100% success responses for the test</summary>

```shell
hey -c 20 -q 10 -z 300s -host www.example.com http://172.18.255.200/

Summary:
  Total:	300.7588 secs
  Slowest:	1.1698 secs
  Fastest:	0.9969 secs
  Average:	1.0085 secs
  Requests/sec:	19.8165
  
  Total data:	2980000 bytes
  Size/request:	500 bytes

Response time histogram:
  0.997 [1]	|
  1.014 [5073]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  1.031 [798]	|■■■■■■
  1.049 [65]	|■
  1.066 [3]	|
  1.083 [0]	|
  1.101 [0]	|
  1.118 [0]	|
  1.135 [1]	|
  1.152 [9]	|
  1.170 [10]	|


Latency distribution:
  10% in 1.0017 secs
  25% in 1.0033 secs
  50% in 1.0061 secs
  75% in 1.0110 secs
  90% in 1.0168 secs
  95% in 1.0216 secs
  99% in 1.0363 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0003 secs, 0.9969 secs, 1.1698 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0004 secs, 0.0000 secs, 0.0308 secs
  resp wait:	1.0035 secs, 0.9960 secs, 1.0380 secs
  resp read:	0.0020 secs, 0.0000 secs, 0.0292 secs

Status code distribution:
  [200]	5960 responses


```

</details>

davidalger:
@arkodg @guydc Would like thoughts from the both of you on the proposed implementation and approach demonstrated here and also whether these defaults are sufficient or need knobs to adjust:

* minimum drain duration (currently 15s)
* graceful drain timeout (currently 600s)
* terminationGracePeriodSeconds (currently 900s)
* total.server_connections exit threshold (currently 0)

guydc:
> This lines up with the default value for [--parent-shutdown-time-s](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-parent-shutdown-time-s)

Is there any significance to using the same value here, if we're not using the envoy hot reload feature?

davidalger:
>> This lines up with the default value for [--parent-shutdown-time-s](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-parent-shutdown-time-s)
>
> Is there any significance to using the same value here, if we're not using the envoy hot reload feature?

@guydc No particular significance to the value being the same in the context of k8s, but it is why I chose 5 minutes greater than the drain timeout since the docs state `the drain time should be less than the parent shutdown time set via the [--parent-shutdown-time-s](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-parent-shutdown-time-s) option`.

The important bit is that `terminationGracePeriodSeconds` must be higher than the graceful drain timeout so that k8s will send a SIGTERM instead of a SIGKILL if the drain runs up the clock against the full 600 second drain timeout.

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -204,12 +206,65 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
...
 }
 
+func expectedContainerLifecycle() *corev1.Lifecycle {
+	var envoyAdminAddr = fmt.Sprint(bootstrap.EnvoyAdminAddress)
+	var envoyAdminPort = fmt.Sprint(bootstrap.EnvoyAdminPort)
+
+	return &corev1.Lifecycle{
+		PreStop: &corev1.LifecycleHandler{
+			Exec: &corev1.ExecAction{
+				Command: []string{"/bin/bash", "-c", strings.ReplaceAll(strings.TrimSpace(dedent.Dedent(`
```
guydc, line 223: is this portable to distroless images?

davidalger: That's actually a really good call out, as I'm pretty sure it _won't work_ in any pre-built distroless image. Probably won't break anything, but won't do anything either since the exec will fail and k8s will proceed with an immediate SIGTERM to Envoy. So now I'm having a sudden realization that v0.6.0 started using the distroless image, but the dev builds (which I tested with) don't. 🤦🏻

The options are more limited with distroless images since they typically don't include extra binaries such as bash, cat, or awk very likely leaving us with the only real viable long-term option being to build our own Envoy proxy images to include something like a wrapper process or a static executable to manage the the termination process. There are other approaches that could work without building our own proxy image, but would tend to be fragile or require exposing the admin port.

arkodg: can we implement a) and b) using the HTTP and Sleep pre stop hooks ?
https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-implementations

davidalger: There is unfortunately no way to accomplish this only leveraging k8s sleep and http lifecycle hooks:

* HTTP requests cannot be made to a port bound to `127.0.0.1` (I've tried this) so requests to the Envoy admin port would require it be exposed which would pose a security risk.
* Envoy doesn't exit after a graceful drain since the process was designed for hot reloading. It would always run out the clock on the `terminationGracePeriodSeconds` even if there were no open connections.
* Sleep prestop actions are [too new](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3960-pod-lifecycle-sleep-action/README.md) and cannot be used even to just introduce a basic X second sleep for deprogramming as they won't reach beta until k8s v1.30.

Here is what I'm going to propose as a path forward:

* Implement a sub-command in Envoy Gateway called `envoy shutdown` which will execute the graceful drain and block until both the min-drain-period has elapsed and open connections have closed (similar to what the bash script does currently)
* Implement a sub-command in Envoy Gateway called `envoy shutdown-manager` which exposes an HTTP endpoint on port 19002 and responds to `/shutdown` which will block until the graceful drain initiated by the `envoy shutdown` sub-command has completed.
* Add a `shutdown-manager` container to the Envoy proxy pod running the `envoy shutdown-manager` http server.
* HTTP `preStop` hook on the `envoy` container requesting `/shutdown` on port 19002
* Exec `preStop` hook on the `shutdown-manager` container which executes the `envoy shutdown` sub-command

When the pod gets a request to shutdown, the `preStop` hook will be triggered on both the `envoy` and `shutdown-manager` containers simultaneously. The `preStop` on the `shutdown-manager` container will execute the graceful drain and the `preStop` on the `envoy` pod will block until the drain is complete. Each of the containers will receive the SIGTERM from k8s per-normal when the `preStop` hook on the container completes.

I've mocked this up as a proof-of-concept to test the setup. Unless there are objections I will plan on re-implementing the drain process per the above outline which is very similar to what Contour did previously and  appears it will work based on my testing.

arkodg: hey @davidalger , exposing `/shutdown` on `envoy` has the same security risk, if its exposed on `0.0.0.0`, it can be called by anyone .

arkodg: @davidalger does `preStop.httpGet.host: localhost` work ?
https://github.com/kubernetes/kubernetes/blob/b069ef6e13a2f44ce844acd9ff88a6957fc6d418/pkg/apis/core/types.go#L2201

davidalger: > does preStop.httpGet.host =  localhost work ?

Tried this. It fails with a `PreStopHook failed` warning added to the pod status and the container exits without allowing the drain to occur. I'm fairly certain this is because the request is coming from kubelet, which isn't local to the pod and thus can't actually make a request to the loopback.

> exposing /shutdown on envoy has the same security risk, if its exposed on 0.0.0.0, it can be called by anyone .

This is mitigated by having the `/shutdown/ready` endpoint largely be a dumb wait that does nothing more than block envoy container shutdown until the the drain logic in the exec prestop hook on the `shutdown-manager` container finishes.

There really are two reasons why the actual drain logic is in an exec to a sub-command on the eg binary:

a) The security aspect, since http prestop hooks must be exposed to the rest of the cluster

b) The sub-command will be useful for a local (any non-k8s) EG deployment when such may be implemented in the future

I've pushed up my work in progress for those that may like to take a look. There are still some things to do to finish it up, but it's functional. Couple example log sets from restarts under varying load, with no failed requests:

<details>
  <summary><code>example 1</code> logs from <code>shutdown-manager</code> container</summary>

```
2024-02-21T14:47:39.037Z    INFO    shutdown-manager    envoy/shutdown_manager.go:72    starting shutdown manager
2024-02-21T14:47:55.365Z    INFO    shutdown-manager    envoy/shutdown_manager.go:86    received shutdown ready request
2024-02-21T14:47:55.398Z    INFO    shutdown-manager    envoy/shutdown_manager.go:115    initiating graceful drain with 15 second minimum drain period and 600 second timeout
2024-02-21T14:47:55.399Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 10
2024-02-21T14:47:56.401Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 10
2024-02-21T14:47:57.402Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 10
2024-02-21T14:47:58.404Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 10
2024-02-21T14:47:59.408Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 9
2024-02-21T14:48:00.412Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 9
2024-02-21T14:48:01.414Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 9
2024-02-21T14:48:02.421Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 9
2024-02-21T14:48:03.427Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 9
2024-02-21T14:48:04.428Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 5
2024-02-21T14:48:05.438Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 5
2024-02-21T14:48:06.441Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 5
2024-02-21T14:48:07.443Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 5
2024-02-21T14:48:08.454Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 5
2024-02-21T14:48:09.461Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 0
2024-02-21T14:48:10.464Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 0
2024-02-21T14:48:10.464Z    INFO    shutdown-manager    envoy/shutdown_manager.go:140    minimum drain period reached; will exit when total connections is 0
2024-02-21T14:48:10.464Z    INFO    shutdown-manager    envoy/shutdown_manager.go:148    graceful drain sequence completed
2024-02-21T14:48:10.473Z    INFO    shutdown-manager    envoy/shutdown_manager.go:62    received SIGTERM
2024-02-21T14:48:11.386Z    INFO    shutdown-manager    envoy/shutdown_manager.go:97    shutdown readiness detected
Stream closed EOF for envoy-gateway-system/envoy-default-eg-e41e7b31-65bb79b45-mkgdt (shutdown-manager)
```

</details>


<details>
  <summary><code>example 2</code> logs from <code>shutdown-manager</code> container</summary>

```
2024-02-21T14:47:55.887Z    INFO    shutdown-manager    envoy/shutdown_manager.go:72    starting shutdown manager
2024-02-21T14:48:56.058Z    INFO    shutdown-manager    envoy/shutdown_manager.go:86    received shutdown ready request
2024-02-21T14:48:56.087Z    INFO    shutdown-manager    envoy/shutdown_manager.go:115    initiating graceful drain with 15 second minimum drain period and 600 second timeout
2024-02-21T14:48:56.089Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 100
2024-02-21T14:48:57.089Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 100
2024-02-21T14:48:58.090Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 100
2024-02-21T14:48:59.091Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 100
2024-02-21T14:49:00.091Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 100
2024-02-21T14:49:01.093Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 74
2024-02-21T14:49:02.094Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 74
2024-02-21T14:49:03.094Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 74
2024-02-21T14:49:04.096Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 74
2024-02-21T14:49:05.096Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 74
2024-02-21T14:49:06.097Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 33
2024-02-21T14:49:07.099Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 33
2024-02-21T14:49:08.100Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 33
2024-02-21T14:49:09.101Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 33
2024-02-21T14:49:10.102Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 33
2024-02-21T14:49:11.103Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 11
2024-02-21T14:49:11.103Z    INFO    shutdown-manager    envoy/shutdown_manager.go:140    minimum drain period reached; will exit when total connections is 0
2024-02-21T14:49:12.105Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 11
2024-02-21T14:49:13.106Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 11
2024-02-21T14:49:14.106Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 11
2024-02-21T14:49:15.107Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 11
2024-02-21T14:49:16.108Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:17.109Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:18.111Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:19.111Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:20.112Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:21.113Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:22.114Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:23.115Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:24.116Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:25.117Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:26.119Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:27.120Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:28.121Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:29.122Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:30.124Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 1
2024-02-21T14:49:31.133Z    INFO    shutdown-manager    envoy/shutdown_manager.go:196    total connections: 0
2024-02-21T14:49:31.133Z    INFO    shutdown-manager    envoy/shutdown_manager.go:148    graceful drain sequence completed
2024-02-21T14:49:31.148Z    INFO    shutdown-manager    envoy/shutdown_manager.go:62    received SIGTERM
2024-02-21T14:49:32.080Z    INFO    shutdown-manager    envoy/shutdown_manager.go:97    shutdown readiness detected
Stream closed EOF for envoy-gateway-system/envoy-default-eg-e41e7b31-65bb79b45-crxbp (shutdown-manager)
```

</details>

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -204,6 +205,34 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
...
+					HTTPGet: &corev1.HTTPGetAction{
+						Path:   "/shutdown/ready",
+						Port:   intstr.FromInt32(envoy.ShutdownReadyPort),
+						Scheme: corev1.URISchemeHTTP,
+					},
+				},
+			},
+		},
+		{
+			Name:            "shutdown-manager",
```
arkodg, line 219: should this be a sidecar container ?
* this will ensure that `preStop` for `shutdown-manager` runs before `envoy`
* building off this
  * preStop exec `shutdown` performs a `touch` on the tmp file and exits
  * preStop exec `/shutdown/ready` performs the envoy shutdown procedure and blocks until complete
  * `/shutdown/ready` can return a non 200 error code immediately if tmp file doesnt exist

davidalger: Not sure of the shutdown sequencing as I don't see much about it in the [announcement blog post](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/) or on the [docs page](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) other than indications that pod termination is not blocked by sidecar containers.

I could look into this if we are OK with dropping support for 1.28 where sidecars containers are feature gated only being enabled by default in 1.29. Not sure I'd recommend this though with 1.29 not yet being available as stable/GA on [AKS](https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar) or [GKE](https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule-for-release-channels)

arkodg: ah good point, we cant rely on order yet, since sidecars just GA'd

then I propose `/shutdown/ready` sleeps for 1s (ensuring in case of termination, the other preStop had the chance to create the tmp file) before checking whether the tmp file exists or not. 
If it doesnt, return `404`, else run the shutdown steps and return `200`.
This eliminates the need for the `for` within the http handler.

arkodg: at a high level, big +1 with this approach

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Shutdown is called from a preStop hook where it will block until envoy can
+// gracefully drain open connections prior to pod shutdown.
+func Shutdown(drainTimeout time.Duration, minDrainDuration time.Duration, exitAtConnections int) error {
+	var startTime = time.Now()
+	var allowedToExit = false
+
+	// Reconfigure logger to write to stdout of main process if running in Kubernetes
+	if _, k8s := os.LookupEnv("KUBERNETES_SERVICE_HOST"); k8s && os.Getpid() != 1 {
+		logger = logging.FileLogger("/proc/1/fd/1", "shutdown-manager", v1alpha1.LogLevelInfo)
```
guydc, line 112: There's a discussion [here](https://www.reddit.com/r/docker/comments/97src6/application_output_to_docker_logs_works_only_when/.) about this option maybe not working if running the process as non-root. 
Not sure if this concern is applicable in this case...

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -0,0 +1,209 @@
...
+	// ShutdownReadyFile is the file used to indicate shutdown readiness.
+	ShutdownReadyFile = "/tmp/shutdown-ready"
+)
+
+// ShutdownManager serves shutdown manager process for Envoy proxies.
+func ShutdownManager(readyTimeout time.Duration) error {
+	// Setup HTTP handler
+	handler := http.NewServeMux()
+	handler.HandleFunc("/shutdown/ready", func(w http.ResponseWriter, r *http.Request) {
+		shutdownReadyHandler(w, r, ShutdownReadyFile)
```
guydc, line 42: Can this timeout before the handler actually detects shutdown readdiness due to the server timeouts and/or kubelet timeouts? If so, what would be the behavior?

guydc (commented):
Great work!

internal/cmd/envoy/shutdown_manager.go
davidalger: > writing directly to /proc/1/fd/1 will be problematic if the process drops root privileges but PID 1 is still running as root

I don't think it will apply in this case, as neither PID 1 nor the exec'd process are running as root since the container sets the uid:gid in the build ([see here](https://github.com/envoyproxy/gateway/blob/main/tools/docker/envoy-gateway/Dockerfile#L7))

internal/infrastructure/kubernetes/proxy/resource.go
davidalger: > then I propose /shutdown/ready sleeps for 1s (ensuring in case of termination, the other preStop had the chance to create the tmp file) before checking whether the tmp file exists or not.
If it doesnt, return 404, else run the shutdown steps and return 200.

Not sure I follow. As soon as the `/shutdown/ready` request returns (no matter the response code) kubelet will immediately send sigterm to PID 1 (envoy).

Was the concern with the `for` loop that it could never return? I've modified this to ensure requests don't endlessly poll after the ready-timeout: 4a3b90f69d547cf36b1e8882a4d2f9ab5e6f9eb5 since `http.TimeoutHandler` of course wouldn't interrupt the processing just breaking the connection.

arkodg: the concerns I was trying to mitigate were
* using more CPU with the `for`
* allowing non kubelet clients to hoard the shutdown handler CPU

Since we cannot establish order between the `envoy` and `shutdown-manager` container, I was proposing we use a delay (`1s` / `2s`) to ensure that for the `preStop` case, we can be confident that the other `preStop` has signaled to us (using a file) that its time to shutdown.

For the non preStop case, it would just be a delay + return 404

internal/cmd/envoy/shutdown_manager.go
davidalger: It has a [configurable timeout](https://github.com/envoyproxy/gateway/pull/2633/files#diff-7a44a61d8a97c19e682d34666a5b0616719ef5a0ab3fa0b857a4863d8294e5f0R67-R68
) (defaulting to 10s longer than the drain timeout) as a safeguard, but it won't timeout prior to this per documented and observed behavior.

If this did return before the `exec` prestop completed the drain, envoy would be sent a SIGTERM and exit after which you may observe errors such as this one logged until the drain timeout elapses and the shutdown command exits:

```
shutdown-manager 2024-02-22T20:48:54.259Z    ERROR    shutdown-manager    envoy/shutdown_manager.go:145    error getting total connections    {"error": "Get \"http://127.0.0.1:19000//stats?filter=^server\\\\.total_connections$&format=json\": dial tcp 127.0.0.1:19000: connect: connection refused"}
```

This should never occur with properly configured (staggered) timeouts, so I would hesitate to code around it for an initial implementation preferring to let the errors be logged to get eyes on the potential problem.

internal/infrastructure/kubernetes/proxy/resource.go
davidalger: Updated in b3dcffd4b3b455cc391190a3c553b52295cd524b

.github/workflows/build_and_test.yaml
```diff
@@ -101,6 +101,7 @@ jobs:
       env:
         KIND_NODE_TAG: ${{ matrix.version }}
         IMAGE_PULL_POLICY: IfNotPresent
+        TAG: latest
```
arkodg, line 104: curious why this is needed ?
https://github.com/envoyproxy/gateway/blob/c5cd964da2f2ec1951ba72d45b54eef8246e96ee/tools/make/env.mk#L17 is the default

api/v1alpha1/envoyproxy_types.go
```diff
@@ -115,6 +120,25 @@ type EnvoyProxyProvider struct {
 	Kubernetes *EnvoyProxyKubernetesProvider `json:"kubernetes,omitempty"`
 }
 
+// ShutdownConfig defines configuration for graceful envoy shutdown process.
+type ShutdownConfig struct {
+	// DrainTimeout defines the graceful shutdown timeout. This should be less than the pod's terminationGracePeriodSeconds.
```
arkodg, line 125:
```suggestion
	// DrainTimeout defines the graceful drain timeout. This should be less than the pod's terminationGracePeriodSeconds.
```

arkodg, line 125: 10min seems like a long time for a default value
I do understand that this will be lesser in case active connections fall to 0, but 10min seems long

api/v1alpha1/envoyproxy_types.go
```diff
@@ -115,6 +120,25 @@ type EnvoyProxyProvider struct {
...
+	// If unspecified, defaults to 600 seconds.
+	//
+	// +optional
+	DrainTimeout *metav1.Duration `json:"drainTimeout,omitempty"`
+	// MinDrainDuration defines the minimum drain duration allowing time for endpoint deprogramming to complete.
+	// If unspecified, defaults to 15 seconds.
+	//
+	// +optional
```
arkodg, line 133: can we do without `MinDrainDuration` & `ExitAtConnections` in the first iteration ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -139,6 +163,16 @@ type EnvoyProxyKubernetesProvider struct {
 	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
 }
 
+// EnvoyProxyDeploymentSpec defines the desired state of the Kubernetes deployment resource.
```
arkodg, line 166: suggest not adding this in the first iteration, in case we need to take another approach in the future
the user will still be able to change settings using  the new `patch` field in https://gateway.envoyproxy.io/latest/api/extension_types/#kubernetesdeploymentspec

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -0,0 +1,234 @@
...
+	// Signal to shutdownReadyHandler that drain process is started
+	if _, err := os.Create(ShutdownActiveFile); err != nil {
+		logger.Error(err, "error creating shutdown active file")
+		return err
+	}
+
+	logger.Info(fmt.Sprintf("initiating graceful drain with %.0f second minimum drain period and %.0f second timeout",
+		minDrainDuration.Seconds(), drainTimeout.Seconds()))
+
+	// Send request to Envoy admin API to initiate the graceful drain sequence
```
arkodg, line 151: would it make more sense for shutdown logic to be moved into `shutdownReadyHandler`

so the shutdown command can
* create a file
* wait until file is deleted

api/v1alpha1/envoyproxy_types.go
davidalger: Good catch! Fixed in f06cc01

davidalger: I don't think 10 minutes is all that high for an edge proxy, and the docs for the drain process seem to agree:

> In edge scenarios, it might be desirable to have a very long drain time. In service to service scenarios, it might be possible to make the drain and shutdown time much shorter (e.g., 60s/90s).

ref: https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-drain-time-s

Case in point, some non-Envoy balancers in front on some of our k8s Ingresses have a 20 minute drain period for HTTP requests. Streaming and TCP connections (which can't be signaled to gracefully close) can also easily run up the clock.

api/v1alpha1/envoyproxy_types.go
davidalger: Removed `ExitAtConnections` in fb4e3ed but `MinDrainDuration` really should stay. It's an important knob that should be configurable. I will be using it when rolling this out to infra I manage, and it's being used in `test/config/gatewayclass.yaml` to prevent a conservative default of 15 seconds from slowing CI processes.

api/v1alpha1/envoyproxy_types.go
davidalger: I keep forgetting about that patch field :) updated in cf106fa, modifying params on this container shouldn't be a common case outside of the timeouts anyways

internal/cmd/envoy/shutdown_manager.go
davidalger: I strongly believe the logic here should be kept in an `exec` preStop. There is no reason really to push this into the HTTP server, and doing so opens the door to multiple drain processes being triggered once the file is in place. Unless there is a significant reason behind the nudge here, I'd like to keep it as-is.

arkodg: I see, you'd prefer if the `shutdown` cmd performs the shutdown logic, while the http handler performs the simpler block in case the file exists, that sgtm

api/v1alpha1/envoyproxy_types.go
arkodg: can you elaborate a little more on the endpoint deprogramming part ? this timeout is added to ensure the proxy's VIP is removed from kube-proxy or other external loadbalancers ?

api/v1alpha1/shared_types.go
```diff
@@ -21,6 +21,14 @@ const (
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
 	DefaultEnvoyProxyImage = "envoyproxy/envoy:distroless-dev"
+	// DefaultShutdownManagerCPUResourceRequests for shutdown manager cpu resource
+	DefaultShutdownManagerCPUResourceRequests = "10m"
+	// DefaultShutdownManagerMemoryResourceRequests for shutdown manager memory resource
+	DefaultShutdownManagerMemoryResourceRequests = "32Mi"
+	// DefaultShutdownManagerImageName is the default image used for the shutdown manager.
+	DefaultShutdownManagerImageName = "envoyproxy/gateway-dev"
```
arkodg, line 29: can we combine these two fields ? similar to `DefaultEnvoyProxyImage` ?

davidalger: @arkodg That's how it originally was, see here for why: https://github.com/envoyproxy/gateway/pull/2633#discussion_r1503008338

I'm ok with either approach tbh, we just have to pick one.

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -273,6 +275,14 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 	return deployment, nil
 }
 
+func expectedTerminationGracePeriodSeconds(cfg *egv1a1.ShutdownConfig) *int64 {
```
arkodg, line 278: shouldn't this be equal to `DrainTimeout` ?

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -0,0 +1,234 @@
...
+	if resp, err := http.Post(fmt.Sprintf("http://%s:%d/drain_listeners?graceful&skip_exit",
```
arkodg, line 152: will this cause the health check defined in https://github.com/envoyproxy/gateway/blob/338515a6f963212bf1e20a38574fa0cc4f6aee30/internal/xds/bootstrap/bootstrap.yaml.tpl#L93 to fail ?

also curious why is `skip_exit` needed ?

api/v1alpha1/shared_types.go
arkodg: ah I just realized the conundrum you're in, big +1 to the approach taken using the ldflags !

internal/infrastructure/kubernetes/proxy/resource_provider.go
davidalger: `terminationGracePeriodSeconds` needs to be higher than the drain timeout so the pod has time to handle the SIGTERM before being sent a SIGKILL. It doesn't have to be much, but it does need to be marginally higher, and assuming other timeouts are functioning it should never be reached under normal circumstances.

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -0,0 +1,234 @@
...
+	// been reached and total connections reaches threshold or timeout is exceeded
+	for {
+		elapsedTime := time.Since(startTime)
+
+		conn, err := getTotalConnections()
+		if err != nil {
+			logger.Error(err, "error getting total connections")
+		}
+
+		if elapsedTime > minDrainDuration && !allowedToExit {
```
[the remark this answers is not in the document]
davidalger: Probably a matter of personal preference, but I opted for this method so that drain logs make it obvious what is happening and one gets visibility on connection counts changing more immediately after starting the graceful drain vs waiting X seconds silently before starting to log for the drain process.

api/v1alpha1/envoyproxy_types.go
arkodg: if this value is present to deal with eventual consistency of the various distributed networking components, prefer if we dropped this to 5s

api/v1alpha1/envoyproxy_types.go
arkodg: thanks for sharing that link, default value of `10m` comes from envoy itself, so sgtm

arkodg (dismissed):
thanks for building this out @davidalger, all users will benefit from this work, and not have to deal with custom images or scripts !
have a few non blocking questions, overall LGTM !

internal/cmd/envoy/shutdown_manager.go
davidalger: The healthcheck endpoint will not fail. It is unimportant though as the the ready status of the endpoint will immediately become false when the pod enters the terminating status triggering anything that uses the endpoint to begin deprogramming.

`skip_exit` is useful to ensure new connections are not unexpectedly blocked if a drain takes than `drain-time-s` on the Envoy proxy, so it becomes important where long drain timeouts are configured. It's a knob recently added to the drain API to support Istio's graceful drain process as they transition to native sidecars: https://github.com/envoyproxy/envoy/pull/30003 and I thought it made sense to include it so we don't have to also configure `drain-time-s` on Envoy where longer than 10m drain periods are needed.

api/v1alpha1/envoyproxy_types.go
davidalger: > to deal with eventual consistency of the various distributed networking components

That's exactly what it's for. I've lowered the default to 5s

internal/cmd/envoy/shutdown_manager.go
arkodg: I think its important to fail the healthcheck endpoint, so downstream LBs can deprogram the endpoint faster

arkodg: referring to calling this endpoint explicitly as the first step of the shutdown process
https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--healthcheck-fail
cc @guydc

internal/xds/translator/listener.go
```diff
@@ -254,6 +254,10 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 		return err
 	}
 
+	// Remove /healthcheck/fail from endpoints that will trigger a drain of listeners for better control
```
arkodg, line 257: this should probably live in `buildTCPListener` & `buildQUICListener`
https://github.com/envoyproxy/gateway/blob/72fadb7c1dbe7d660c2900527c34f53cfcfce7ac/internal/xds/translator/listener.go#L130

davidalger: that makes sense, addressed in afcbf23

arkodg (approved):

internal/cmd/envoy/shutdown_manager.go
guydc: I think that the LB concern is related to a specific deployment type (1 proxy per Node with `ExternalTrafficPolicy: Local`). This is already handled in K8s 1.26 by: https://kubernetes.io/blog/2022/12/30/advancements-in-kubernetes-traffic-engineering/#traffic-loss-from-load-balancers-during-rolling-updates

guydc (approved):
LGTM

PR #2639, opened 2024-02-17
Implement IP deny/allow ACL rules

### LINKED ISSUE #2462: Support Allow/Deny IP Subnets
*Description*:
would be used.

As a user, I would like to limit the client IP address to a few subnets for some cases as well as also deny specific subnets

zetaab:
**What type of PR is this?**
feat: add Support for Allow/Deny IP Subnets

Usually there needs to be way to deny / allow subnets for single routes / entire gateway.

Fixes #2462

### Notes to reviewers

disclaimer: This is my first bigger PR to this repository

#### Questions

- Gateway level ACL definitions works. However, if route do have any ACL rules the route will override the Gateway ACL settings. Like mentioned in https://github.com/envoyproxy/gateway/issues/2462#issuecomment-1906591842 how this should work? Is this behavior ok?
- This applies rules only to http, there is RBAC policy for network level as well (but those cannot be applied to route level). It might be needed if envoy gateway is used for tcp stuff? That is perhaps task to another PR?
- Not sure does it work out of the box with ipv6, I do not have k8s with ipv6 to verify it.

I have verified this PR manually by following test cases:

```
route: allow ip - ok (allowed ip works and others not)
route: deny ip - ok (denied ip does not work and others will work)
route: allow big cidr, deny single ip(s) from allowed cidr - ok (big cidr works except denied ips)
route: deny big cidr, allow single ip(s) from denied cidr - not working, but it should not either? (whole big cidr does not work, no matter is there something in allow)

gateway: allow ip - ok (allowed ip works and others not)
gateway: deny ip - ok (denied ip does not work and others will work)
gateway: allow big cidr, deny single ip(s) from allowed cidr - ok (big cidr works except denied ips)
gateway: deny big cidr, allow single ip(s) from denied cidr - not working, but it should not either? (whole big cidr does not work, no matter is there something in allow)
```


api/v1alpha1/acl_types.go
```diff
@@ -0,0 +1,25 @@
...
+// +kubebuilder:validation:XValidation:rule="(has(self.allow) || has(self.deny))",message="one of allow or deny must be specified"
+//
+// ACL defines the IP deny/allow configuration.
+type ACL struct {
+	Allow []IPSpec `json:"allow,omitempty"`
+	Deny  []IPSpec `json:"deny,omitempty"`
+}
+
+// IPSpec defines the configuration for IP.
+type IPSpec struct {
```
zirain, line 17: is there a common struct from upstream(gateway api or kubernetes)? maybe use `a.b.c.d/24` for better UX?

zirain:
Please submit the API design first, otherwise there may be a lot of waste.

zetaab:
@zirain I cannot find any mentioning about such policy in Contributing or developer guide in this project. What you mean with API design? Do you mean that I should make separate PR about stuff that is going under api/? I tried to look old PRs and all of them have API changes in same PR. Could you give example what I should do here.

PR #2640, opened 2024-02-17
chore: Update Envoy proxy image to envoy:distroless-dev in main

davidalger:

This PR updates the Envoy proxy image used to distroless dev builds for the `main` branch.

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -19,7 +19,7 @@ const (
 	// DefaultDeploymentMemoryResourceRequests for deployment memory resource
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
-	DefaultEnvoyProxyImage = "envoyproxy/envoy-dev:latest"
+	DefaultEnvoyProxyImage = "envoyproxy/envoy:distroless-dev"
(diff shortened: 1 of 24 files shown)


zirain (approved):
LGTM

I recall istio want to use distroless by default, but there are a lot of historical problems.

zhaohuabing:
EG should pin a specific envoy version and use a distroless image when releasing v1.0.0. But I think it's ok to use non-distroless dev images at the moment?

davidalger:
@zhaohuabing The main reason for this is so we're testing and building with the same Envoy base image build EG is released with. Fully on board with pinning specific versions for EG releases and using the latest dev build for EG main branch; but I see it as a risk, albeit a minor one, to be building and testing with an ubuntu base when EG will pin and release with distroless builds of Envoy. The EG image used for dev is already distroless.

Is it often needed to jump into the Envoy containers with a shell for debug purposes? My experience could be different of than others working on the project of course but it isn't something I've needed to do.

arkodg (approved):

arkodg:
@zhaohuabing, imo if we feel debugging pain, we should surface it in a GH issue and try and and solve it with `egctl`.
Using the `distroless` flavor in `main` and releases ensures lesser surprises (like [this](https://github.com/envoyproxy/gateway/pull/2633#discussion_r1493036329) one )

PR #2652, opened 2024-02-19
api: Authorization API design

zetaab:
**What type of PR is this?**

api: Authorization API design

The below snippet is what an authorization config looks like, the first iteration only supports `ClientCIDR` as the `Principal`. JWT Claims, custom headers, and client certs will be supported in the future.
```
authorization:
  defaultAction: "Deny"
  rules:
    - action: "Allow"
      principal:
        clientCIDR: ["10.0.1.0/24"]
```

Related: https://github.com/envoyproxy/gateway/issues/2462 https://github.com/envoyproxy/gateway/issues/2250


api/v1alpha1/acl_types.go
```diff
@@ -0,0 +1,24 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// +kubebuilder:validation:XValidation:rule="(has(self.allow) || has(self.deny))",message="one of allow or deny must be specified"
```
zhaohuabing, line 8: What will happen if both allow and deny are specified? Should this be allowed?

zetaab: actually that was same question that I asked myself. However, there is usecase for that.

Lets say that I have following config:

```
  acl:
    allow:
    - cidr: "1.2.3.0/24"
    deny:
    - cidr: "1.2.3.120/32"
```

so this configuration means that it will accept all traffic from 1.2.3.0/24. However, it will block that single ip address 1.2.3.120. So it can be used for "allow big chunk and deny smaller pieces of it"

zhaohuabing (commented):
Should another type of client identity be considered in this PR?  
For example, EG likely needs to support fine-grained access control based on JWT claims, or peer identity extracted from client certificate attributes.
Under the hook, all types of client identities(IP, JWT, peer identity) will be backed by the envoy RBAC filter. There are a few other implementations that can be referenced:
* Ambassdar: similar to this PR, but I think Ambassdar doesn't support other type of credentials. https://www.getambassador.io/docs/edge-stack/latest/topics/running/ambassador#ip-allow-and-deny
* Istio: a fat authorization policy including all types of client identities https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source

@envoyproxy/gateway-maintainers

zetaab:
@zhaohuabing yep, I was thinking to take that issue as next. However, in http filter level ACL is handled before JWT filter. So there must be 2 different RBAC filters because of that? Otherwise it is not possible to make RBAC filter based on JWT results. At least I do not like the way that ACL is also handled after authentication etc. That will expose JWT auth to everyone and they can try to hack things.

api/v1alpha1/acl_types.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// ACL defines the IP deny/allow configuration.
+type ACL struct {
```
arkodg, line 11: this will need to be generic for multiple attributes
```
acl/authorization:
  type: ClientCIDR
  clientCIDR:
    allow:
    - 1.2.3.0/24    
```

or 

```
acl/authorization:
- type: ClientCIDR
  clientCIDR:
  - 1.2.3.0/24
  action: ALLOW

Comment (zetaab) [reply api/v1alpha1/acl_types.go:11]: @arkodg did you read https://github.com/envoyproxy/gateway/pull/2652#issuecomment-1952652492 any comments to that? I do understand if all others do want everything gathered to single api. But then that also means that the http filter is going to be after all auth modules all called. That will be potential weakness that comes to security.

Comment (arkodg) [reply api/v1alpha1/acl_types.go:11]: based on the snippet in https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rbac_filter#role-based-access-control-rbac-filter, I prefer replacing `attributes` with `clientSelectors`

Comment (zetaab) [reply api/v1alpha1/acl_types.go:11]: @arkodg but in that case I would like to have acl which do 2 things: first it will only allow internal ip addresses and secondly I want to give access to only sub x. In that case I do not have any other option than put all acl rules after jwt and other filters? Or can there be a way to split the configuration to two different filters which is executed in different places?

Comment (arkodg) [reply api/v1alpha1/acl_types.go:11]: This is a good point, don't have a good answer for this right now, but this becomes another goal of #2571.

Comment (zetaab) [reply api/v1alpha1/acl_types.go:11]: thinking something like 

```
authorization:
- order: 1
  rules:
  - clientSelectors:
      clientCIDR:
      - 1.2.3.0/24
    action: ALLOW
```

Comment (arkodg) [reply api/v1alpha1/acl_types.go:11]: thats a good start, we'll have to deal with this in a holistic way for all features, so prefer if we deferred the API for filter ordering here, else this PR will be not make much progress

Comment (zetaab) [reply api/v1alpha1/acl_types.go:11]: if this split is going to be possible later, I can skip that for now. But to get things forward, I am thinking how the configuration could look like in case of normal rbac policies.

Lets imagine 3 use-cases:
1) I want allow 1.2.3.0/24
2) I want that only JWT role admin is allowed to access all paths in service
3) I want that JWT role user is allowed to access all paths in service instead /admin/*

```
authorization:
  rules:
  - clientSelectors:
      clientCIDR:
      - 1.2.3.0/24
    permissions:
    - any: true
    action: ALLOW
  - clientSelectors:
      jwt:
        path:
        - payload
        - role
        value:
          stringMatch:
            exact: admin
    permissions:
    - any: true
    action: ALLOW
  - clientSelectors:
      any: true
    permissions:
    - not:
        header:
          name: ":path"
          regexMatch: "/admin(/.*)?"
    action: ALLOW
```

how does that look like?

Comment (zetaab) [reply api/v1alpha1/acl_types.go:11]: two selectors: the policy does not need that much rows and is less verbose

```
authorization:
  rules:
  - clientSelectors: (can we allow multiple selectors under single rule item?)
      clientCIDR:
      - 1.2.3.0/24
      jwt:
        path:
        - payload
        - role
        value:
          stringMatch:
            exact: admin
    permissions:
    - any: true
    action: ALLOW
  - clientSelectors:
      any: true
    permissions:
    - not:
        header:
          name: ":path"
          regexMatch: "/admin(/.*)?"
    action: ALLOW
```

**Comment by zetaab** [comment]
```
authorization:
  - action: ALLOW
    rules:
    - from:
        cidr:
        - 1.2.3.0/24
        jwt:
          path:
          - payload
          - role
          value:
            stringMatch:
              exact: admin
      to:
      - any: true
    - from:
      - any: true
      to:
      - not:
          header:
            name: ":path"
            regexMatch: "/admin(/.*)?"
  - action: DENY
    rules:
    - from:
        cidr:
        - 4.4.4.0/24
      to:
      - any: true
```

just trying to think some ideas..

File: api/v1alpha1/acl_types.go (hunk 2 above)
Comment (arkodg) [reply api/v1alpha1/acl_types.go:11]: yah, we should allow multiple selectors under a rule`, it mimics the policies field in https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-rbac, all selectors should be `AND`d together

**Comment by zetaab** [comment]
@envoyproxy/gateway-maintainers before implementing anything, could others give opinion also about the format in https://github.com/envoyproxy/gateway/pull/2652#discussion_r1495014346 (usecases explained in previous post)

**Comment by zetaab** [comment]
@arkodg @zhaohuabing updated the PR according the ideas.. but still not sure is everything like should.

File: api/v1alpha1/authorization_types.go [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+	Any bool `json:"any,omitempty"`
+}
+
+// RuleActionType specifies the types of authorization rule action.
+// +kubebuilder:validation:Enum=AllowRuleType;DenyRuleType;LogRuleType
+type RuleActionType string
+
+const (
+	// AllowRuleType is the action to allow the request.
+	AllowRuleType RuleActionType = "ALLOW"
```
arkodg, line 52: imo lets use `Allow`, `Deny`, `Log` to use the same casing as other fields in the API such as the RateLimit Unit field (`Hour, `Second` etc)

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -52,13 +55,40 @@ func validateSecurityPolicySpec(spec *egv1a1.SecurityPolicySpec) error {
...
+
+	if err := ValidateAuthorisation(spec.Authorization); err != nil {
 		errs = append(errs, err)
 	}
 
 	return utilerrors.NewAggregate(errs)
 }
 
+// ValidateAuthorisation validates the provided Authorisation configuration.
+func ValidateAuthorisation(as *egv1a1.Authorization) error {
```
arkodg, line 72:
```suggestion
func ValidateAuthorization(as *egv1a1.Authorization) error {
```

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,47 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// Authorization defines the authorization configuration.
```
arkodg, line 8: what would deny all requests look like ?
```
authorization: {}
```
or
```
authorization:
  rules:
  - action: Deny
```

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,47 @@
...
+type Authorization struct {
+	// Rules contains all the authorization rules.
+	//
+	// +kubebuilder:validation:MinItems=1
+	Rules []Rule `json:"rules,omitempty"`
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// ClientSelectors contains the client selector configuration.
```
arkodg, line 18: can we add a little more explanation when multiple clientSelectors are added
* they are ANDd together, only if all selector are valid, then Action is performed

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,47 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// Authorization defines the authorization configuration.
+type Authorization struct {
+	// Rules contains all the authorization rules.
```
arkodg, line 10: what happens when there are many rules, but none of them match, is the default action to `Deny` ?
if so we should add a comment on it

zetaab: if we have any allow rule and it does not match, then the default action is Deny

api/v1alpha1/authorization_types.go
zetaab: it could like
```
authorization:
  rules:
  - action: Deny
```

api/v1alpha1/authorization_types.go
arkodg: thanks, what about the case of multiple rules matching for the same request - Rule 1's action is `allow`, and Rule2's action is `deny`, how is the decision computed here ?

zetaab: lets assume that we have case:

- allow cidr 1.2.3.0/24
- deny cidr 1.2.3.123/32

if the client ip is 1.2.3.123 it will be denied, otherwise all other ips from that 1.2.3.0/24 block will be allowed.

arkodg: thanks for explaining this, and what about 
```
deny cidr 1.2.3.0/24
allow cidr 1.2.3.123/32
```

will `1.2.3.123` get allowed or denied ?

from a users perspective, they'd like to know which of these below are true
* first match or
* best match or
* if any match has a deny action , then deny

zetaab: that does not work (or depends on the filter order?). In that case 1.2.3.123 will be denied. I have not tested but it might be because of the order of allow and deny rbac filters. 

Earlier my rbac filters was in order: allow, deny. So if someone wants that working then it might be possible if filters are in order: deny, allow. 

I am thinking how this is actually going to be implemented, if we have following:

allow cidr 8.0.0.0/8
deny cidr 8.1.0.0/16
allow cidr 8.1.1.1/32

how this is translated to filters? Should we keep the order and make own rbac filter from each clientSelector item (and rules inside single clientSelector is just ANDed)? Then it might be possible to make these kind of things. (not sure I have not tested).

zetaab:
@arkodg I tried this api and implemented the authorization for it: 

1) other selectors is going to be handled here https://github.com/envoyproxy/gateway/commit/64bffe1e4ddf7ee7ca63a4b3b92ee5b56adef94b#diff-39171fec9cb0b38b175c0b1d846990f95dd907518d1d7757857aaa7bca5ea528R107

2) https://github.com/envoyproxy/gateway/commit/64bffe1e4ddf7ee7ca63a4b3b92ee5b56adef94b#diff-39171fec9cb0b38b175c0b1d846990f95dd907518d1d7757857aaa7bca5ea528R127 there needs to be possibility to configure the permissions somehow

```
  - clientSelectors:
      clientCIDR:
      - 1.2.3.0/24
    permissions:
    - any: true
    action: ALLOW
```

could solve the issue.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,53 @@
...
+	// matches the action for the request is allow.
+	//
+	// +kubebuilder:validation:MinItems=1
+	Rules []Rule `json:"rules,omitempty"`
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// ClientSelectors contains the client selector configuration.
+	// All selectors are and together and only if all selector are valid
```
arkodg, line 23:
```suggestion
	// All selectors are AND'd together and only if all selector are valid
```

arkodg:
thanks for running the tests !
* does Case 4 imply "most specific match wins" ? (you can try to reverse order and test to ensure its not last match or first match that wins)
* also weird Case 3 not working is a little weird, I wouldn't block the API on it, because it should be solvable using other methods

zetaab:
case 5

```
  authorization:
    rules:
    - clientSelector:
        - clientCIDR:
          - 44.94.107.109/32
      action: Deny
```

result: single ip address is denied

case 6:

```
  authorization:
    rules:
    - clientSelector:
        - clientCIDR:
          - 44.94.107.109/32
      action: Deny
    - clientSelector:
        - clientCIDR:
          - 44.94.0.0/16
          - 44.65.0.0/16
      action: Allow
```

result: single ip address is still denied

case 7:

```
  authorization:
    rules:
    - clientSelector:
        - clientCIDR:
          - 44.94.0.0/16
      action: Deny
    - clientSelector:
        - clientCIDR:
          - 44.94.107.109/32
          - 44.65.0.0/16
      action: Allow
```

result: 44.65.0.0/16 cidr works, but nothing works from 44.94.0.0/16 cidr. So deny is somehow overriding allow rules

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Authorization defines the authorization configuration.
+type Authorization struct {
+	// Rules contains all the authorization rules.
+	// If rules contains at least one Allow rule and none of them
+	// matches the action for the request is deny.
+	// If rules contains at least one Deny rule and none of them
+	// matches the action for the request is allow.
+	// If rules contains Allow and Deny,
+	// the Deny rules are overriding the Allow rules from the overlapping parts.
```
arkodg, line 16:
```suggestion
	// and the request matches both, `Deny` takes precedence.
```

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,55 @@
...
+	// +kubebuilder:validation:MinItems=1
+	ClientSelectors []ClientSelector `json:"clientSelector,omitempty"`
+
+	// Action defines the action to be taken if the rule matches.
+	Action RuleActionType `json:"action"`
+}
+
+// ClientSelector contains the client selector configuration.
+type ClientSelector struct {
+	// ClientCIDRs is a list of CIDRs.
```
arkodg, line 37: This condition is met if any of the values in the list of CIRDs matches.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// Authorization defines the authorization configuration.
+type Authorization struct {
+	// Rules contains all the authorization rules.
+	// If rules contains at least one Allow rule and none of them
```
arkodg, line 11: After chatting with @ZackButcher today, here are some suggestions to simplify this API
1. Default Deny -  if no rules are defined or no rules match, then `Deny`
```
authorization: {}
```

is same as 

```
authorization:
  rules:
  - action: Deny
```

is same as client IP being `3.3.3.3` and not matching any rules
```
authorization:
  rules:
  - clientSelectors:
       clientCIDRs:
        - 1.1.1.1/32  
     action: Deny
  - clientSelectors:
       clientCIDRs:
        - 2.2.2.2/32  
     action: Allow     
```

2. If Allow and Deny rules match, `Deny` takes precedence

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// ClientSelectors contains the client selector configuration.
+	// All selectors are AND'd together and only if all selector are valid
+	// the Action is performed.
+	//
+	// +kubebuilder:validation:MinItems=1
+	ClientSelectors []ClientSelector `json:"clientSelector,omitempty"`
```
arkodg, line 29: thinking out loud, does this need to be a `list` ? , I know I had suggested it earlier 🙈

api/v1alpha1/authorization_types.go
arkodg: trying to make sure we add a statement that highlights that if there are a list of CIDRs, and 1 of them match, then the `action` is executed

api/v1alpha1/authorization_types.go
arkodg: However 2. breaks this case - I'd like to deny the entire subnet `1.1.1.0/24` except `1.1.1.1`

arkodg: to solve above case, we use the `matcher_list` field https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/rbac/v3/rbac.proto#extensions-filters-http-rbac-v3-rbac and perform first match wins
cc @ZackButcher

arkodg: if we do first match, then for the below config
```
authorization:
  rules:
    - clientSelectors:
       methods:
        - POST  
     action: Deny
  - clientSelectors:
       clientCIDRs:
        - 1.1.1.1/32  
     action: Allow
  - clientSelectors:
       clientCIDRs:
        - 1.1.1.0/24  
     action: Deny
```

This would allow us to
* reject all `POST` requests
* allow client with IP `1.1.1.1` for other requests
* deny others

zetaab: in this case last deny is not needed. If the client ip does not match 1.1.1.1 its denied already before that

arkodg: agree, we are essentially using list order as an inherently priority order for rules

api/v1alpha1/authorization_types.go
zetaab: it depends is all items under ClientSelector going to have array or not. Like in case of cidrs (array), this can be just object. However, when like jwt is added it needs to have array as well.

zetaab:
@arkodg I would like to move to coding phase in case of this API. This PR now contains more than 40 reviews, kind of saying: how much more is still needed? Could for instance some of the other maintainers give feedback about this PR and then we could somehow merge this soonish?

api/v1alpha1/authorization_types.go
arkodg: all items under `clientSelector` will be an array to implement the OR/ANY case, so we should use an array for `jwt` settings as well

arkodg:
> @arkodg I would like to move to coding phase in case of this API. This PR now contains more than 40 reviews, kind of saying: how much more is still needed? Could for instance some of the other maintainers give feedback about this PR and then we could somehow merge this soonish?

if we moved the `clientSelector` to a singular field, and defined first `rule` wins semantics i.e. first rule has higher precedence over next one, I'd have no other concerns.

This API and implementation, will unfortunately not be able to make it into v1.0.0 since we will be releasing v1.0.0 RC early next week, so we'll only be able to merge this after v1.0.0 is released.

@envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers please review, and share any feedback

zetaab:
@arkodg modified the field now, as 1.0.0 is released I could continue working on this?

arkodg:
thanks for kick starting this PR again @zetaab !
the open question left for me is the `match` semantics -  I find the current definition is the API confusing, which is why I recommended for `first match wins` with a default `Deny`. I'll bring this up in the community meeting on Thursday, so others can weigh in as well

davidalger:
This came up in the community meeting today. I'm in support of `first match wins` with a default `Deny` design as it preserves an important level of flexibility as described here: https://github.com/envoyproxy/gateway/pull/2652#discussion_r1506954516

It also does still allow for the inverse where the auth is allow by default by placing a catch all Allow rule as the last rule with Deny rules preceding it. For example: 

```yaml
authorization:
  rules:
  - clientSelectors:
       clientCIDRs:
        - 123.123.0.0/24
     action: Deny
  - clientSelectors:
       clientCIDRs:
        - 0.0.0.0/0
     action: Allow
```

This would allow us to

* deny from IPs falling in 123.123.0.0/24
* allow others (the built-in default-deny would never come in to play since the last rule allows everything)

arkodg:
hey @zetaab still working on this one ?

zetaab:
@arkodg well not really, I have tried to get it done but always something is missing. So I have kind of lost energy to do anything. If someone want to continue - feel free.

arkodg:
Thanks for the update @zetaab, sorry you feel this way, I'll try and take your work forward in a follow PR

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+	// the Action is performed.
+	//
+	ClientSelector ClientSelector `json:"clientSelector,omitempty"`
+
+	// Action defines the action to be taken if the rule matches.
+	Action RuleActionType `json:"action"`
+}
+
+// ClientSelector contains the client selector configuration.
+type ClientSelector struct {
```
zhaohuabing, line 35: `ClientSelector` sounds a bit odd , what do you think of `principal`?  I'm not a native speaker, so my feeling on naming may not be correct anyway :-)

```
// Principle defines the configuration for a principle.
type Principle struct {
	ClientCIDRs []string `json:"clientCIDR,omitempty"`

	// Other type of principal in future: jwt claim, client cert, etc.
}
```

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+	Rules []Rule `json:"rules,omitempty"`
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// ClientSelector contains the client selector configuration.
+	// All selectors are AND'd together and only if all selector are valid
+	// the Action is performed.
+	//
+	ClientSelector ClientSelector `json:"clientSelector,omitempty"`
```
zhaohuabing, line 28: Should we add another layer here so we can enforce authz on methods?

```
type Rule struct {
	// Action defines the action to be taken if the rule matches.
	Action RuleActionType `json:"action"`

	// Policies define which principals are allowed/denied access to specified permissions based on the action.
	Policies []AuthorizationPolicy `json:"policy"`
}

// AuthorizationPolicy defines the configuration for an authorization policy.
type AuthorizationPolicy struct {
	// Principals define the list of principals that are allowed/denied access based on the action.
	// If no principals are specified, all requests are matched.
	Principles []Principle `json:"principles"`

	// Permissions define the list of permissions that are allowed/denied access based on the action.
	// If no permissions are specified, all requests are matched.
	Permissions []Permission `json:"permissions"`
}
// Permission defines actions that the authorized principals can take.
type Permission struct {
	// Methods define the list of HTTP methods that are allowed/denied access based on the action.
	// If no methods are specified, all methods are matched.
	Method []string `json:"method"`
}
```

api/v1alpha1/authorization_types.go
arkodg: `principal` doesnt seem to define the attribute well, I agree `clientSelector` is also odd, @ZackButcher had brought up the same thing, wanna come up with a 3rd name :) ?

api/v1alpha1/authorization_types.go
arkodg: imo lets create the base structure and add `methods` in a follow up PR, its another attribute like IP
```
rules:
 clientSelector:
   methods:
   - GET
   clientCIDRs:
   - 1.0.0.0/8
 action: "Allow" 
```

api/v1alpha1/authorization_types.go
davidalger: would it work to call it `matches` similar to what's done on `HTTPRoute`?

arkodg: good point, there was a really long debate about this while designing RateLimit, and `clientSelectors` was chosen there to better describe the `match` field (its way to match/select on specific client attributes) , more in https://github.com/envoyproxy/gateway/pull/706

zhaohuabing:
I'm considering having two separate APIs—one for ACL and one for authorization.

Rationale:

* Typically, we don't need fine-grained access control(method, header, etc.) based on IP addresses; simply allowing or denying access is sufficient.
* It seems awkward to combine IP addresses and client identities (JWT tokens, client certs) within the same API.

zetaab:
IPs could be useful together with jwts. For instance, if I have /admin path. I want allow it only for admins which are coming from cidr x

zufardhiyaulhaq:
@zhaohuabing I don't think it's awkward.
one use case is in our company where we use mTLS & IP restriction or key auth with IP restriction.

This is to archive high security implementation in case client cert / key auth / jwt is compromised, attacker need to access the client infrastructure istead of using their local network.

zetaab:
@arkodg @zhaohuabing is there something that we could now do to get progress to this issue? I would really like to see this authorization feature. IMO it should be started from the ip addresses, then add more features like jwt

api/v1alpha1/authorization_types.go
zhaohuabing: I still feel like we should use `Principal` here, [it also aligns with Envoy's term](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-principal).

And methods should be in another `permission` field.

```
rules:
   principal:
     clientCIDRs: [1.0.0.0/8]
   permission: [GET]
   action: "Allow"  
```

zetaab:
```
rules:
- principal:
     clientCIDRs: [1.0.0.0/8]
     # header: userid // extract user identity from an HTTP header
     # jwtClaim: sub // extract user identity from a jwtClaim
     # other means: client cert, basic auth ...
     # any: true // matches to all principals?
   permission:
     mergemethod: and // supported values and, not, needs better name
     rules:
     - method: GET
       path:
         prefix: /foobar // could support exact, prefix, suffix, contains?
   action: "Allow"
```

so 3 things:
1) permissions should contain way to define path
2) principals may need any: true option which matches all requests? Like if I want to allow all requests to /public path
3) in permissions we may need to have also NOT use case. Like if I want to allow all requests to application but deny /admin path. In that case I would set principals any: true and permissions: mergemethod: not, rules: path: prefix: /admin

@zhaohuabing @arkodg

zhaohuabing: > 1) permissions should contain way to define path

The targeted HTTPRoute would be enough for most cases? I suggest we start with method in the first run.

> 2) principals may need any: true option which matches all requests? Like if I want to allow all requests to /public path

We can make API simpler by allowing `principal` to be optional, which means any.

arkodg: `subject` over `clientSelector` sgtm, can we address these changes as part of a new PR ?

zhaohuabing: `subject` is good. Another new PR is also OK.

zetaab:
@arkodg @zhaohuabing please review. I have updated the PR according the spec

zhaohuabing (approved):
LGTM! Thank you so much for this PR and your patience during the review process.  Authorization is a great addition to the `SecurityPolicy`!

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+	// If empty, all subjects are included.
+	//
+	// +optional
+	Subjects []Subject `json:"subjects,omitempty"`
+
+	// Permissions contains allowed HTTP methods.
+	// If empty, all methods are matching.
+	//
+	// +optional
+	Permissions []string `json:"permissions,omitempty"`
```
arkodg, line 28: can we omit `Permissions` from the init API ?

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// Authorization defines the authorization configuration.
+type Authorization struct {
+	// Rules contains all the authorization rules.
+	//
```
arkodg, line 11: we need to add some explicit comments here, saying - "Rules are executed in order, if a rule matches, the `action` is executed and further rules are skipped"

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+	Action RuleActionType `json:"action"`
+}
+
+// Subject contains the subject configuration.
+type Subject struct {
+	// ClientCIDR contains client cidr configuration.
+	// Valid examples are "192.168.1.0/24" or "2001:db8::/64"
+	//
+	// +optional
+	ClientCIDR *string `json:"clientCIDR,omitempty"`
```
arkodg, line 40: this field needs to be plural, highlighting `OR` semantics

zetaab:
@arkodg your comments are not down the line what was talked earlier in comments. So I do not know what to do here. If I do changes that you asked, is @zhaohuabing going to accept those.

Anyway as I see it, the biggest issue with the spec is that we should try implementing cases and after that we can see does it really work and support all the needed use cases. So I would like to get started with something. Should we do this in same PR? Like implement cidr support?

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,54 @@
...
+	Subjects []Subject `json:"subjects,omitempty"`
+
+	// Permissions contains allowed HTTP methods.
+	// If empty, all methods are matching.
+	//
+	// +optional
+	Permissions []string `json:"permissions,omitempty"`
+
+	// Action defines the action to be taken if the rule matches.
+	Action RuleActionType `json:"action"`
```
zhaohuabing, line 31: We may want to move `Action` to the top level to match it with the [Envoy RBAC filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto) configuration. So this API can be implemented with RBAC filter.

```
  - name: envoy.filters.http.rbac
    config:
      rules:
        action: ALLOW
        policies:
          "service-admin":
            permissions:
              - any: true
            principals:
              - authenticated:
                  principal_name:
                    exact: "cluster.local/ns/default/sa/admin"
              - authenticated:
                  principal_name:
                    exact: "cluster.local/ns/default/sa/superuser"
          "product-viewer":
            permissions:
                - and_rules:
                    rules:
                      - header:
                          name: ":method"
                          string_match:
                            exact: "GET"
                      - url_path:
                          path: { prefix: "/products" }
                      - or_rules:
                          rules:
                            - destination_port: 80
                            - destination_port: 443
            principals:
              - any: true
```

arkodg:
hey @zetaab completely understand your point, I'll try and prioritize this issue in the community meeting this week to see if we can make more progress on it, and will report back here

arkodg:
> Should we align the Authorization API design with the [Envoy RBAC filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#config-rbac-v3-rbac) configuration(move `Action` out of `Rule` to a higher level in the structure) for ease of implementation?
> 
> For example, this EG Authorization config:
> 
> ```
>   authorization:
>     action: ALLOW
>     rules: 
>       - name: "service-admin"
>         subjects:
>           - clientCert: "cluster.local/ns/default/sa/admin"
>           - clientCert: "cluster.local/ns/default/sa/superuser"
>       - name: "product-viewer"
>         permissions:
>           - "GET"
> ```
> 
> can be translated to the below Envoy Gateway RBAC filter:
> 
> ```
>   - name: envoy.filters.http.rbac
>     config:
>       rules:
>         action: ALLOW
>         policies:
>           "service-admin":
>             permissions:
>               - any: true
>             principals:
>               - authenticated:
>                   principal_name:
>                     exact: "cluster.local/ns/default/sa/admin"
>               - authenticated:
>                   principal_name:
>                     exact: "cluster.local/ns/default/sa/superuser"
> 
>           "product-viewer":
>             permissions:
>                 - header:
>                     name: ":method"
>                     string_match:
>                       exact: "GET"
>             principals:
>               - any: true
> ```
> 
> We may also need to add another layer on top of the above API, so we can have a more flexible authz logic like "allowing authenticated users access but excluding all users from 10.75.1.0/24"
> 
> EG Authorization config:
> 
> ```
>   authorization:
>     - action: ALLOW
>       pocies: 
>         - name: "service-admin"
>           subjects:
>             - clientCert: "cluster.local/ns/default/sa/admin"
>             - clientCert: "cluster.local/ns/default/sa/superuser"
>         - name: "product-viewer"
>           permissions:
>             - "GET"
>     - action: DENY
>       policies: 
>         - name: "all-users-in-a-specific-cidr"
>           subjects:
>             - cidr: "10.75.1.0/24"
> ```
> 
> The above configuration can be implemented with multiple RBAC filters:
> 
> ```
>   - name: envoy.filters.http.rbac
>     config:
>       rules:
>         action: DENY
>         policies:
>           "deny-a-specific-cidr":
>             permissions:
>               - any: true
>             principals:
>               - remote_ip: "1.2.3.0/24"  
>             
>    - name: envoy.filters.http.rbac
>     config:
>       rules:
>         action: ALLOW
>         policies:
>           "service-admin":
>             permissions:
>               - any: true
>             principals:
>               - authenticated:
>                   principal_name:
>                     exact: "cluster.local/ns/default/sa/admin"
>               - authenticated:
>                   principal_name:
>                     exact: "cluster.local/ns/default/sa/superuser"
> 
>           "product-viewer":
>             permissions:
>                 - header:
>                     name: ":method"
>                     string_match:
>                       exact: "GET"
>             principals:
>               - any: true
> ```
> 
> To conclude, the `Authorization` API will look like:
> 
> ```
> type Authorization struct {
>   Rules []Rule `json:"rules,omitempty"`
> }
> 
> type Rule struct {
>   Action RuleActionType `json:"action"`
> 
>   Policies []Policy `json:"rules,omitempty"`
> }
> 
> type Policy struct {
>   Subjects []Subject `json:"subjects,omitempty"`
> 
>   Permissions []string `json:"permissions,omitempty"`
> }
> ```
> 
> Does this make sense? @arkodg @zetaab

prefer if we take a user/usability approach first before measuring feasibility (mapping to envoy) for this API

zetaab:
that will look like

```
authorization:
  rules:
  - action: Allow
    policies:
    - subjects:
      - clientCIDR: 1.1.1.1
      permissions: ["GET"]
```

I am fine with that. At least it should support many use cases

zhaohuabing:
Following our discussion in this week's EG meeting, I did some rearch on prior art of Gateway implementations. Here are some findings:

## Istio Authorization Policy

This is the most relevant one. Istio Authz policy offers fine-grained authorization control based on host, path, and method. It supports source(subject) from cidr, workload identity, and user identity.

```
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: httpbin
  namespace: foo
spec:
  action: ALLOW
  rules:
  - from:
    - source:
        principals: ["cluster.local/ns/default/sa/sleep"]
    - source:
        namespaces: ["test"]
    to:
    - operation:
        methods: ["GET"]
        paths: ["/info*"]
    - operation:
        methods: ["POST"]
        paths: ["/data"]
    when:
    - key: request.auth.claims[iss]
      values: ["https://accounts.google.com"]
```

Since Istio Authz handles both the ingress and mesh traffic, `from` can be a principals, ipBlocks, and namespaces. `to` is a list of operations(permission on what resources), and `when` is a list of extra condtions of requests needed to be matched.

And there is a important difference between Istio Authz Policy and EG Authz Policy: Istio can have multiple Authz policies on a specific service, but EG only supports one `SecurityPolicy` per xRoute, so EG Authz API structure needs to look like a list of Istio Authz policies.

## APISIX supports authz on an HTTPRoure

APISIX's Authz concept is smilar to what we want to achieve in EG: authz on a sepcif HTTPRoute. It's less flexible than Istio Authz policy though.

```
apiVersion: apisix.apache.org/v2
kind: ApisixRoute
metadata:
 name: httpserver-route
spec:
 http:
 - name: rule1
   match:
     hosts:
     - httpbin.org
     paths:
       - /*
   backends:
   - serviceName: httpbin
     servicePort: 80
   plugins:
   - name: consumer-restriction
     enable: false
     config:
       allowed_by_methods:
       - user: "default_jack1"
         methods:
         - "POST"
         - "GET"
       - user: "default_jack2"
         methods:
         - "GET"
```

## Ambassador supports IP deny/allow list at the global level, without full-fleged authz policy.

```
apiVersion: getambassador.io/v2
kind: Module
metadata:
  name: ambassador
spec:
  config:
    ip_deny:
    - peer: 99.99.99.0
    - remote: 10.16.0.0/16
```

## Contour supports IP deny/allow on route level, without full-fleged authz policy.

```
apiVersion: projectcontour.io/v1
kind: HTTPProxy
metadata:
  name: basic
spec:
  virtualhost:
    fqdn: foo-basic.bar.com
    ipAllowPolicy:
      # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
      - cidr: 127.0.0.1/32
        source: Peer
  routes:
    - conditions:
      - prefix: /
      services:
        - name: s1
          port: 80
      # route-level ip filters override the virtualhost-level filters
      ipAllowPolicy:
      # traffic is allowed if it came from localhost (i.e. co-located reverse proxy)
      - cidr: 127.0.0.1/32
        source: Peer
      # and the request originated from an IP in this range 
      - cidr: 99.99.0.0/16
        source: Remote
```

## Other Gateway Implementations

Other Gateway implementations normally just delegate the Authz to an external process, such as an OPA.

Nginx: https://github.com/summerwind/opa-nginx-rbac
Kong: https://docs.konghq.com/hub/kong-inc/opa
Gloo: https://docs.solo.io/gloo-edge/latest/guides/security/auth/extauth/opa

Of all the implementations studied, Istio Authz and APISIX Authz APIs align most closely with EG's requirements.

api/v1alpha1/authorization_types.go
zhaohuabing: We may want to handle rule priority DENY > ALLOW, like [Istio Authz does](https://istio.io/latest/docs/reference/config/security/authorization-policy/).

arkodg: do you think how Istio defines it is easy to understand ?

I find it very hard, which is why I am a +1 for "first match in rules list wins"

zetaab:
```
type Authorization struct {
	Deny  []Policy `json:"deny,omitempty"`
	Allow []Policy `json:"allow,omitempty"`
	Log   []Policy `json:"log,omitempty"`
}
```

in which order these are going to be applied? it might be issue? or is there really need to have combination of different action types

zhaohuabing:
> ```
> type Authorization struct {
> 	Deny  []Policy `json:"deny,omitempty"`
> 	Allow []Policy `json:"allow,omitempty"`
> 	Log   []Policy `json:"log,omitempty"`
> }
> ```
> 
> in which order these are going to be applied? it might be issue? or is there really need to have combination of different action types

First `Deny` rules in the order of the list, then `Allow` rules, then Audit(Log) rules.

zetaab:
hmm it could work in that case, if its possible to implement like that

zhaohuabing:
> hmm it could work in that case, if its possible to implement like that

@arkodg mentioned a use case: allowing a small hole in a large CIDR block. This won't work for that because `Deny` goes first, so we need to use the other option in the PR code.

api/v1alpha1/authorization_types.go
zhaohuabing: Think again and agree, `First match wins` is more intuitive.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,62 @@
...
+	Rules []Rule `json:"rules,omitempty"`
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// Action defines the action to be taken if the rule matches.
+	Action RuleActionType `json:"action"`
+
+	// Policies contains the list of authorization policies.
+	Policies []Policy `json:"policies"`
```
arkodg, line 24: @zhaohuabing with an example, can you explain why we need an extra hierarchy of `policies` and why can't be achieve the same by moving `Subject` and `Permission` under `Rule` ?

zhaohuabing: I was thinking of something like the below:

```
authorization:
  rules:
    - action: "Allow"
      policies:
        - subject: ["Admin"]      
          permission: ["GET", "POST", "DELETE"]
        - subject: ["Viewer"]
          permission: ["Get"]
```

Of course, we can also make it like:

```
authorization:
  rules:
    - action: "Allow"
      subject: ["Admin"]      
      permission: ["GET", "POST", "DELETE"]  
    - action: "Allow"
      subject: ["Viewer"]      
      permission: ["GET", "POST", "DELETE"]  
```

But these would need more RBAC filters to implement it since one RBAC filter can only has one `Action`, either `Allow` or `Deny`.

arkodg:
```
But these would need more RBAC filters to implement it since one RBAC filter can only has one Action, either Allow or Deny.
```

I vote to prioritise usability over envoy feasibility here

api/v1alpha1/authorization_types.go
zhaohuabing: @arkodg This may be able to be achieved with [matcher API](https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/matcher.proto#envoy-v3-api-msg-xds-type-matcher-v3-matcher). However, Envoy doc says it's WIP and not recommended to be used in production.

arkodg: id still suggest us using it, we have used the other alpha APIs in the project already

zhaohuabing: Yeah, Macher API is much more flexible and solves all the concerns raised in the discussion. Let's go with it.

api/v1alpha1/authorization_types.go
zhaohuabing: We don't have this problem anymore since we opted to the Matcher API.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,56 @@
...
+// +kubebuilder:validation:Enum=Allow;Deny;Log
+type RuleActionType string
+
+const (
+	// Allow is the action to allow the request.
+	Allow RuleActionType = "Allow"
+	// Deny is the action to deny the request.
+	Deny RuleActionType = "Deny"
+	// Log is the action to log the request.
+	Log RuleActionType = "Log"
```
zhaohuabing, line 55: I'd like to remove the "Log" action since it's not related to Authz at all. If we need a knob on log decision, it should be within the Log API. @arkodg @zetaab

arkodg: It's usually used as a shadow tool, but fine to leave it out for.now

zirain: named it `DryRun`?

zhaohuabing: This action is just for logging decisions. We can use `shadow_matcher` for the shadow purpose.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,61 @@
...
+	// If the first rule denies the request, and the second rule allows it,
+	// the request will be denied.
+	//
+	// +optional
+	Rules []Rule `json:"rules"`
+
+	// DefaultAction defines the default action to be taken if no rules match.
+	// If not specified, the default action is Deny.
+	// +optional
+	DefaultAction *RuleActionType `json:"defaultAction"`
```
zhaohuabing, line 25: Add a `DefaultAction` to make the default behavior explicit.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Rule defines the single authorization rule.
+type Rule struct {
+	// Action defines the action to be taken if the rule matches.
+	Action RuleActionType `json:"action"`
+
+	// Subject contains the subject of the rule.
+	Subject Subject `json:"subjects,omitempty"`
```
zhaohuabing, line 34: We will need an "exclueSubject" as well.

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,65 @@
...
+type Authorization struct {
+	// Rules contains all the authorization rules.
+	// Rules are evaluated in order, the first matching rule will be applied,
+	// and the rest will be skipped.
+	//
+	// For example, if there are two rules: the first rule allows the request
+	// and the second rule denies it, when a request matches both rules, it will be allowed.
+	//
+	// +optional
+	Rules []Rule `json:"rules"`
```
arkodg, line 19:
```suggestion
	Rules []Rule `json:"rules, omitempty"`
```

arkodg (approved):
LGTM thanks @zhaohuabing & @zetaab for building this out !

api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,65 @@
...
+	Action RuleActionType `json:"action"`
+
+	// Principal specifies the client identity of a request.
+	Principal Principal `json:"principal"`
+
+	// Permissions contains allowed HTTP methods.
+	// If empty, all methods are matching.
+	//
+	// +optional
+	// Permissions []string `json:"permissions,omitempty"`
```
zirain, line 40: I'm not sure why this pass limiter, prefer to remove instead of commnet.

zirain (approved):
LGTM, can you clean the comment code in following PR?

PR #2657, opened 2024-02-20
api: support retry on in BackendTrafficPolicy

guydc:
api: support retry on in BackendTrafficPolicy

Relates to #1821,  #1907

Continue work by @tmsnan in #2168


api/v1alpha1/retry_types.go
```diff
@@ -0,0 +1,111 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Retry defines the retry strategy to be applied.
+type Retry struct {
+	// NumRetries is the number of retries to be attempted. Defaults to 2.
```
zhaohuabing, line 14: Is there a particular reason why the default retry is different from the envoy default(1)?

guydc: We're using [istio defaults](https://github.com/istio/istio/blob/b652ff114d85c2ed15034e1c3c87447d96b9c769/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go#L36) as a real-world example for reasonable defaults here.

arkodg (approved):
LGTM thanks !

shawnh2 (approved):
lgtm, thanks

zhaohuabing (approved):
LGTM. Thanks！

PR #2658, opened 2024-02-20
chore: use list instead of map for Provider Resources

arkodg:
* list adds order stability


zirain (approved):

zhaohuabing (approved):

cnvergence:
tests need to be updated

internal/message/types.go
```diff
@@ -21,7 +21,7 @@ import (
 type ProviderResources struct {
 	// GatewayAPIResources is a map from a GatewayClass name to
 	// a group of gateway API and other related resources.
-	GatewayAPIResources watchable.Map[string, *gatewayapi.GatewayClassResources]
```
cnvergence, line 24: *gatewayapi.Resources will also need a generated DeepCopy functions

internal/gatewayapi/resource.go
```diff
@@ -20,41 +20,6 @@ import (
 
 type XdsIRMap map[string]*ir.Xds
 type InfraIRMap map[string]*ir.Infra
-type GatewayClassResources map[string]*Resources
```
cnvergence, line 23: list will hit a similar issue
```bash
panic: watchable.DeepCopy: type is not copiable: *[]*gatewayapi.Resources
```

Xunzhuo:
@arkodg conflicts need to be resolved.

internal/gatewayapi/resource.go
arkodg: yeah I had to write my own `DeepCopy` as well

internal/gatewayapi/resource.go
```diff
@@ -20,41 +24,6 @@ import (
 
 type XdsIRMap map[string]*ir.Xds
```
Xunzhuo, line 25: Should we also use list for xdsIR infraIR instead of map

arkodg: thats an internal structure, iirc each resource is unrelated and the map improves lookup

arkodg: https://github.com/envoyproxy/gateway/blob/3e70498b63803d8bd78dcad31883bd7211cda3f8/internal/ir/xds.go#L87 helps deal with out of order equality when comparing ir xds

zirain (approved):
LGTM

Issue #2606, opened 2024-02-14
OIDC security policy is leaking on other routes


akhenakh:
I have one OIDC security policy to protect one httproute:

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  annotations:
  generation: 1
  labels:
    app.kubernetes.io/instance: auth-mm-api
    env: dev
  name: oidc-auth-mm
  namespace: auth-mm-api
  resourceVersion: "184503780"
spec:
  oidc:
    clientID: xxxxxxxxxxxxxxxx
    clientSecret:
      group: ""
      kind: Secret
      name: oidc-client-secret
    provider:
      issuer: https://login.microsoftonline.com/xxxxxxxxxxxxxxxxxx/v2.0
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: auth-mm-api-public-route
status:
  conditions:
  - lastTransitionTime: "2024-02-13T21:18:09Z"
    message: SecurityPolicy has been accepted.
    reason: Accepted
    status: "True"
    type: Accepted
 ```

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  annotations:
  creationTimestamp: "2024-02-02T13:04:06Z"
  generation: 2
  labels:
    app: auth-mm-api
    app.kubernetes.io/instance: auth-mm-api
    env: dev
  name: auth-mm-api-public-route
  namespace: auth-mm-api
  resourceVersion: "182167763"
spec:
  hostnames:
  - auth-mm-api.bloh.blah.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
    namespace: envoy-gateway-system
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: auth-mm-api
      port: 8085
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
status:
  parents:
  - conditions:
    - lastTransitionTime: "2024-02-08T17:05:42Z"
      message: Route is accepted
      observedGeneration: 2
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2024-02-08T17:05:42Z"
      message: Resolved all the Object references for the Route
      observedGeneration: 2
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: gateway.envoyproxy.io/gatewayclass-controller
[...]

PR #2663, opened 2024-02-20
Fix: requests not matching defined routes trigger per-route filters

zhaohuabing:
This PR disables per-route filters in `http_filters`, and removes the filter configuration in the `route_config`.
This fixes https://github.com/envoyproxy/gateway/issues/2606 without needing to add a `catch-all-return-404` route.

Fix https://github.com/envoyproxy/gateway/issues/2606
Related: https://github.com/envoyproxy/envoy/issues/32484


arkodg (approved):
LGTM thanks !

PR #2664, opened 2024-02-20
Add suffix for oauth cookies

zhaohuabing:
This PR adds a suffix for oauth cookies to prevent multiple oauth filters from overwriting each other's cookies.

The suffix is the hashed UID of the `SecurityPolicy` that contains the OIDC configuration.

For example:
```
                  bearerToken: BearerToken-5F93C2E4
                  idToken: IdToken-5F93C2E4
                  oauthExpires: OauthExpires-5F93C2E4
                  oauthHmac: OauthHMAC-5F93C2E4
                  refreshToken: RefreshToken-5F93C2E4
```

### DIFF
--- a/internal/gatewayapi/securitypolicy.go
+++ b/internal/gatewayapi/securitypolicy.go
@@ -495,16 +495,11 @@ func (t *Translator) buildOIDC(
 		logoutPath = *oidc.LogoutPath
 	}
 
-	nsName := types.NamespacedName{
-		Namespace: policy.GetNamespace(),
-		Name:      policy.GetName(),
-	}
 	h := fnv.New32a()
-	_, err = h.Write([]byte(nsName.String()))
-	if err != nil {
+	if _, err = h.Write([]byte(policy.UID)); err != nil {
 		return nil, fmt.Errorf("error generating oauth cookie suffix: %w", err)
 	}
-	suffix := strconv.Itoa(int(h.Sum32()))
+	suffix := fmt.Sprintf("%X", h.Sum32())
 
 	return &ir.OIDC{
 		Provider:     *provider,
--- a/internal/gatewayapi/testdata/securitypolicy-with-oidc-invalid-issuer.in.yaml
+++ b/internal/gatewayapi/testdata/securitypolicy-with-oidc-invalid-issuer.in.yaml
@@ -27,6 +27,7 @@ securityPolicies:
   metadata:
     namespace: default
     name: policy-non-exist-secretRef
+    uid: b8284d0f-de82-4c65-b204-96a0d3f258a1
   spec:
     targetRef:
       group: gateway.networking.k8s.io
--- a/internal/gatewayapi/testdata/securitypolicy-with-oidc-invalid-issuer.out.yaml
+++ b/internal/gatewayapi/testdata/securitypolicy-with-oidc-invalid-issuer.out.yaml
@@ -62,6 +62,7 @@ securityPolicies:
     creationTimestamp: null
     name: policy-non-exist-secretRef
     namespace: default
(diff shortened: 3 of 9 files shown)


internal/gatewayapi/securitypolicy.go
```diff
@@ -502,6 +502,7 @@ func (t *Translator) buildOIDC(
 		RedirectURL:  redirectURL,
 		RedirectPath: redirectPath,
 		LogoutPath:   logoutPath,
+		CookieSuffix: fmt.Sprintf("%s-%s", policy.Namespace, policy.Name),
```
arkodg, line 505: feels like config leak

arkodg: what about something like a hash on the clientID or policyNsName (stable yet irreversible) ?

internal/gatewayapi/securitypolicy.go
```diff
@@ -494,6 +495,17 @@ func (t *Translator) buildOIDC(
...
+	nsName := types.NamespacedName{
+		Namespace: policy.GetNamespace(),
+		Name:      policy.GetName(),
+	}
+	h := fnv.New32a()
+	_, err = h.Write([]byte(nsName.String()))
+	if err != nil {
+		return nil, fmt.Errorf("error generating oauth cookie suffix: %w", err)
+	}
+	suffix := strconv.Itoa(int(h.Sum32()))
```
liorokman, line 507: Instead of converting to a string with a decimal number, how about converting to a string with a hexidecimal number?

Since the hash function used generates a 32 bit number, this would consistently generate a string with the same length.

```golang
suffix := fmt.Sprintf("%0X", int(h.Sum32))
```

internal/gatewayapi/securitypolicy.go
```diff
@@ -494,6 +495,12 @@ func (t *Translator) buildOIDC(
 		logoutPath = *oidc.LogoutPath
 	}
 
+	h := fnv.New32a()
```
arkodg, line 498: can we reuse https://github.com/envoyproxy/gateway/blob/4c79ef9aa1e06d3b61b83e95f21d1705fa6568ba/internal/provider/utils/utils.go#L39 ?

liorokman: `HashString` uses SHA256, which is theoretically slower than FNV-32, and generates a 256-bit result.

I don't think it matters that much for this purpose that the hash being used is slower, or crypto-secure, but using the full 256 bits in the cookie suffix would result in 64 characters instead of the current 8. I think that would be a little long for a cookie name. 

Using only 32 bit from the result would theoretically be OK for this purpose, but I would prefer to use the FNV-32 hash function here so that the entire hash result would still be used.

liorokman (approved):
LGTM

internal/gatewayapi/securitypolicy.go
zirain: move it into a more common package, e.g. `internal/util`

arkodg: non blocking comment - less code, lesser bugs, lesser vulnerabilities, if this hash function is better, lets replace it with the one in `provider/utils`, as @zirain suggests, if `provider/utils` is the wrong home, lets move it to `utils`

arkodg (approved):

internal/gatewayapi/securitypolicy.go
zhaohuabing: Will do in a follow-up PR because it will impact many places.

zhaohuabing: https://github.com/envoyproxy/gateway/pull/2681

PR #2667, opened 2024-02-21
feat(translator): implement timeout in ClientTrafficPolicy

### LINKED ISSUE #2598: Add timeouts support to clientTrafficPolicy

yaelSchechter:
**What type of PR is this?**
Implement Timeout in ClientTrafficPolicy

Fixes #2598


internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -395,6 +400,34 @@ func translatePathSettings(pathSettings *egv1a1.PathSettings, httpIR *ir.HTTPLis
...
+	}
+
+	if clientTimeout.HTTP != nil {
+		if clientTimeout.HTTP.RequestReceivedTimeout != nil {
+			d, err := time.ParseDuration(string(*clientTimeout.HTTP.RequestReceivedTimeout))
+			if err != nil {
+				return err
+			}
+
+			if httpIR.Timeout == nil {
```
liorokman, line 415: Small nit, but maybe this is slightly more readable as a switch statement:

```golang
    switch {
        case httpIR.Timeout == nil:
                      httpIR.Timeout = &ir.ClientTimeout{}
                      fallthrough
         case httpIR.Timeout.HTTP == nil:
                      httpIR.Timeout = &ir.HTTPClientTimeout{}
                      fallthrough
    }

    httpIR.Timeout.HTTP.RequestReceivedTimeout = &metav1.Duration{
                                 Duration: d,
    }
```

yaelSchechter: Thanks, Lior
I think that this suggestion offers a cleaner structure, but is it a convention for implementing switch cases? I'm not sure that all readers are familiar with the `fallthrough` keyword.

liorokman: The `fallthrough` keyword is part of the language - readers should be familiar with the language.

liorokman (approved):
LGTM

zirain (approved):
LGTM, looking for doc and e2e with following PRs.

PR #2671, opened 2024-02-22
fix: Policies should apply only to gateways they were attached to when mergeGateways is true

### LINKED ISSUE #2666: When there are multiple Gateways and `mergeGateway `is enabled , using CORS on one Gateway will affect the others
*Description*:
When there are multiple Gateways and `mergeGateway` is enabled , using CORS on one Gateway will affect the others 

*Repro steps*:
Install the Gateway API CRDs and Envoy Gateway 
```shell
helm install eg oci://docker.io/envoyproxy/gateway-helm --version v0.0.0-latest -n envoy-gateway-system --create-namespace
```

Use aill-in-one yaml described below:

```shell
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config

---
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  mergeGateways: true 
---
apiVersion: v1
kind: Namespace
metadata:
  name: sealos-system
---
kind: Gateway
metadata:
  name: eg
  namespace: sealos-system
spec:
  gatewayClassName: eg
  listeners:
    - hostname: "*.192.168.0.15.nip.io"
      name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns-oehe
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  replicas: 1
  selector:
    matchLabels:
      app: bdkzlmibsivuiqav
  template:
    metadata:
      labels:
        app: bdkzlmibsivuiqav
    spec:
      containers:
      - image: nginx:1.25
        name: nginx
        ports:
        - containerPort: 80
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 100m
            memory: 64Mi
          requests:
            cpu: 10m
            memory: 16Mi
---
apiVersion: v1
kind: Service
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: bdkzlmibsivuiqav
  type: ClusterIP
---
kind: HTTPRoute
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  hostnames:
  - ntjxuedx.192.168.0.15.nip.io
  parentRefs:
    kind: Gateway
    name: eg
    namespace: sealos-system
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: bdkzlmibsivuiqav
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns-vrpg
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mfqjpuycbgjrtdww
  template:
    metadata:
      labels:
        app: mfqjpuycbgjrtdww
    spec:
      containers:
      - image: nginx:1.25
        name: nginx
        ports:
        - containerPort: 80
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 100m
            memory: 64Mi
          requests:
            cpu: 10m
            memory: 16Mi
---
apiVersion: v1
kind: Service
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: mfqjpuycbgjrtdww
  type: ClusterIP
---
kind: Gateway
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  gatewayClassName: eg
  listeners:
  - hostname: qccbahgo.qccbahgo
    name: http
    port: 80
    protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  hostnames:
  - qccbahgo.qccbahgo
  parentRefs:
    kind: Gateway
    name: mfqjpuycbgjrtdww
    namespace: ns-vrpg
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: mfqjpuycbgjrtdww
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
---
kind: SecurityPolicy
metadata:
  name: cors-example
  namespace: sealos-system
spec:
  targetRef:
    kind: Gateway
    name: eg
  cors:
    allowOrigins:
    - "http://*.foo.com"
    allowMethods:
    - PUT
    - GET
    - POST
    - DELETE
    - PATCH
    - OPTIONS
    maxAge: 600s
    allowCredentials: true
```

Get the name of the Envoy service created the by the  Gateway, because mergeGateway is enabled, the way to obtain the Envoy service differs from the quickstart.

```shell
export ENVOY_SERVICE=$(kubectl get svc -n envoy-gateway-system --selector=gateway.envoyproxy.io/owning-gatewayclass=eg -o jsonpath='{.items[0].metadata.name}')
```
Port forward to the Envoy service
```shell
kubectl -n envoy-gateway-system port-forward service/${ENVOY_SERVICE} 8888:80 &
```
Verify that the CORS headers are present in the response of the OPTIONS request from `http://www.foo.com`, and the host is `ntjxuedx.192.168.0.15.nip.io`

```shell
curl -H "Origin: http://www.foo.com" \
  -H "Host: ntjxuedx.192.168.0.15.nip.io" \
  -X OPTIONS -v -s \
  http://localhost:8888/
```
The following is the response, which is reasonable because this host is bound to the Gateway in the targetRef.

```shell
< HTTP/1.1 200 OK
< access-control-allow-methods: PUT, GET, POST, DELETE, PATCH, OPTIONS
< access-control-max-age: 600
```

Verify that the CORS headers are present in the response of the OPTIONS request from `http://www.foo.com`, and the host is `qccbahgo.qccbahgo`

```shell
curl -H "Origin: http://www.foo.com" \
  -H "Host: qccbahgo.qccbahgo" \
  -X OPTIONS -v -s \
  http://localhost:8888/
```
The following is the response, the same as above which is unreasonable because this host is unrelated to the Gateway in the targetRef.

```shell
< HTTP/1.1 200 OK
< access-control-allow-methods: PUT, GET, POST, DELETE, PATCH, OPTIONS
< access-control-max-age: 600
```

I think it is unreasonable. Our use case is similar to the following structure: users can only operate on CRs in their own namespaces. However, when using CORS, it affects other Gateways in the cluster, which poses significant risks.

img_v3_0275_843b3e72-c935-451c-8576-b575e7c8069g

liorokman:
Fixes #2666

### DIFF
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -357,7 +357,8 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
 		string(policy.Spec.TargetRef.Name),
 	)
 	for _, http := range ir.HTTP {
-		if t.MergeGateways && http.GatewayName != policyTarget {
+		gatewayName := http.Name[0:strings.LastIndex(http.Name, "/")]
+		if t.MergeGateways && gatewayName != policyTarget {
 			continue
 		}
 		for _, r := range http.Routes {
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -352,7 +352,14 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
 	// Should exist since we've validated this
 	ir := xdsIR[irKey]
 
+	policyTarget := irStringKey(
+		string(ptr.Deref(policy.Spec.TargetRef.Namespace, gwv1a2.Namespace(policy.Namespace))),
+		string(policy.Spec.TargetRef.Name),
+	)
 	for _, http := range ir.HTTP {
+		if t.MergeGateways && http.GatewayName != policyTarget {
+			continue
+		}
 		for _, r := range http.Routes {
 			// Apply if not already set
 			if r.RateLimit == nil {
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -100,11 +100,10 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 			switch listener.Protocol {
 			case gwapiv1.HTTPProtocolType, gwapiv1.HTTPSProtocolType:
 				irListener := &ir.HTTPListener{
-					GatewayName: irStringKey(gateway.Namespace, gateway.Name),
-					Name:        irHTTPListenerName(listener),
-					Address:     "0.0.0.0",
-					Port:        uint32(containerPort),
-					TLS:         irTLSConfigs(listener.tlsSecrets),
+					Name:    irHTTPListenerName(listener),
+					Address: "0.0.0.0",
+					Port:    uint32(containerPort),
+					TLS:     irTLSConfigs(listener.tlsSecrets),
 					Path: ir.PathSettings{
 						MergeSlashes:         true,
 						EscapedSlashesAction: ir.UnescapeAndRedirect,
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -100,10 +100,11 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 			switch listener.Protocol {
 			case gwapiv1.HTTPProtocolType, gwapiv1.HTTPSProtocolType:
 				irListener := &ir.HTTPListener{
-					Name:    irHTTPListenerName(listener),
-					Address: "0.0.0.0",
-					Port:    uint32(containerPort),
-					TLS:     irTLSConfigs(listener.tlsSecrets),
+					GatewayName: irStringKey(gateway.Namespace, gateway.Name),
+					Name:        irHTTPListenerName(listener),
+					Address:     "0.0.0.0",
+					Port:        uint32(containerPort),
+					TLS:         irTLSConfigs(listener.tlsSecrets),
 					Path: ir.PathSettings{
 						MergeSlashes:         true,
 						EscapedSlashesAction: ir.UnescapeAndRedirect,
--- a/internal/gatewayapi/securitypolicy.go
+++ b/internal/gatewayapi/securitypolicy.go
@@ -360,7 +360,8 @@ func (t *Translator) translateSecurityPolicyForGateway(
 		}
 	}
 
-	// Apply IR to all the routes within the specific Gateway
+	// Apply IR to all the routes within the specific Gateway that originated
+	// from the gateway to which this security policy was attached.
 	// If the feature is already set, then skip it, since it must have be
 	// set by a policy attaching to the route
 	//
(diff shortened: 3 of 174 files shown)


cnvergence (approved):
LGTM

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -345,7 +345,14 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
 	// Should exist since we've validated this
 	ir := xdsIR[irKey]
 
+	policyTarget := irStringKey(
```
arkodg, line 348: can we unify the logic and style used for route matching as well ?
https://github.com/envoyproxy/gateway/blob/c30d037a0018ddb131cacb2f041ae5bc81dcae10/internal/gatewayapi/backendtrafficpolicy.go#L284

with comments on - how prefix is generated and a helper match to compute a match

internal/gatewayapi/listener.go
```diff
@@ -100,10 +100,11 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
 			switch listener.Protocol {
 			case gwapiv1.HTTPProtocolType, gwapiv1.HTTPSProtocolType:
 				irListener := &ir.HTTPListener{
-					Name:    irHTTPListenerName(listener),
-					Address: "0.0.0.0",
-					Port:    uint32(containerPort),
-					TLS:     irTLSConfigs(listener.tlsSecrets),
+					GatewayName: irStringKey(gateway.Namespace, gateway.Name),
```
arkodg, line 103: if we use the `prefix` method approach we can do w/o the extra field and same on memory

internal/gatewayapi/backendtrafficpolicy.go
arkodg: https://github.com/envoyproxy/gateway/blob/c30d037a0018ddb131cacb2f041ae5bc81dcae10/internal/gatewayapi/helpers.go#L361

liorokman:
> thanks for fixing this one @liorokman, isn't a similar fix needed in ClientTrafficPolicy as well ?

The way that `ClientTrafficPolicy` works, it filters the relevant gateways according to the `Gateway` resources correctly. It already isolated policies to the correct listeners.

I've updated the unit-test to include all the different types of policies to make sure this behaviour isn't lost.

internal/gatewayapi/backendtrafficpolicy.go
liorokman: Sorry, I don't follow your meaning.

internal/gatewayapi/listener.go
liorokman: Done.

guydc (approved):
LGTM

arkodg (approved):
LGTM

PR #2681, opened 2024-02-23
chore: move provider/utils outside

zhaohuabing:


internal/utils/misc.go
```diff
@@ -25,19 +25,21 @@ func NamespacedName(obj client.Object) types.NamespacedName {
...
 }
 
-func HashString(str string) string {
-	h := sha256.New() // Using sha256 instead of sha1 due to Blocklisted import crypto/sha1: weak cryptographic primitive (gosec)
-	h.Write([]byte(str))
-	return strings.ToLower(fmt.Sprintf("%x", h.Sum(nil)))
+// Digest returns a 32-bit hashh of the input string.
+// The hash is represented as a capitalized hexadecimal string.
+func Digest(str string) string {
+	h := fnv.New32a()
```
arkodg, line 42: How did this not cause a change in envoy service names ?
Won't this change cause a churn in services that were deployed with v0.6 and now there will be no automated way to clean up those resources?

zhaohuabing: Yes, that would be a problem :-(

Then we need to continue using the old way: generate a 256 hash and only use the first 8 characters of the hex string.

arkodg: Yeah, would be ideal to avoid creating this messy situation for current users w/o an easy way to clean up

zirain (approved):

arkodg (approved):

Issue #2055, opened 2023-10-23
Add Overridden condition in the BackendTrafficPolicy status


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Relates to https://github.com/envoyproxy/gateway/issues/1821 & https://github.com/envoyproxy/gateway/pull/2053

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Issue #1934, opened 2023-10-07
Add the ability to override only a portion of the spec using Policies


arkodg:
*Description*:

With policies such as the [ClientTrafficPolicy](https://github.com/envoyproxy/gateway/issues/1846), a user can target a `section` within a target e.g. a `Listener` within a `Gateway` which applies or overrides the target with the `spec` of the policy.

Another use case is where a user wants to only override some fields i.e. the top level policy might setup defaults for fields like keep alive but a specific policy for a specific target/section may want to only apply an extra setting such as retries

One way to achieve this is to add a `type` field within the policy resource which can be set to `Replace` or `Merge`, which defaults to `Replace` (apply entire spec)

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

luvk1412:
My use case where i need this: 
we have a global security policy which have defined cors, each dev can select various predefined security policies for auth, cors and other stuff based on label selectors for there routes. To achieve this, merging would be required in our use case.

arkodg:
@guicassolato and the Kuadrant team had to solve a similar problem and their solution was a `strategy:Merge` field
more in https://docs.kuadrant.io/0.8.0/architecture/rfcs/0009-defaults-and-overrides/#design-option-strategy-field

arkodg:
Keeping this issue open to track docs

PR #2684, opened 2024-02-23
add overriding condition to BackendTrafficPolicy and SecurityPolicy

zhaohuabing:
fix https://github.com/envoyproxy/gateway/issues/2055


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -87,6 +92,32 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 				continue
 			}
 
+			// Populate the gatewayRouteMap that will be used to check policy overrides
+			var parents []gwapiv1.ParentReference
+			switch r := route.(type) {
```
arkodg, line 97: can use https://github.com/envoyproxy/gateway/blob/72fadb7c1dbe7d660c2900527c34f53cfcfce7ac/internal/gatewayapi/contexts.go#L216 here

arkodg: or even better can use https://github.com/envoyproxy/gateway/blob/72fadb7c1dbe7d660c2900527c34f53cfcfce7ac/internal/gatewayapi/contexts.go#L242 directly

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -108,6 +139,23 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 
 			t.translateBackendTrafficPolicyForGateway(policy, gateway, xdsIR)
 
+			// Check if this policy is overridden by other policies
```
arkodg, line 142: shouldnt this be executed before `translateBackendTrafficPolicyForGateway`

arkodg:
thanks for picking this one up @zhaohuabing 
* can the PR also apply the same logic for SecurityPolicy ?
* as part of this PR can we also fix the logic here 
https://github.com/envoyproxy/gateway/blob/72fadb7c1dbe7d660c2900527c34f53cfcfce7ac/internal/gatewayapi/backendtrafficpolicy.go#L364
as pointed out by you earlier, its adoping `Merge` semantics, when this should be a `Replace` i.e. if that route has an existing policy attached (maybe computed using the `routeMap`, we should skip applying any setting, rather than checking which is `nil`)

zhaohuabing:
> can the PR also apply the same logic for SecurityPolicy ?

Yes, I originally planned to apply this to `SecurityPolicy` when this one is approved.  But since you already mentioned this, I'm going to do it in this PR.

> as pointed out by you earlier, its adoping Merge semantics, when this should be a Replace i.e. if that route has an existing policy attached (maybe computed using the routeMap, we should skip applying any setting, rather than checking which is nil)

I thought it was intentional. For example, a default global config can be defined at the Gateway level, and app developers can fine-tune the configurations at the route level. Having a global config on the top, the app developers don't have to redefine every setting at the roue level, but only these they really want to modify.

arkodg:
> I thought it was intentional. For example, a default global config can be defined at the Gateway level, and app developers can fine-tune the configurations at the route level.

The benefit of having a global config on the top is that the app developers don't have to redefine every setting at the roue level, but only these they really want to modify. The downside is that it's difficult to figure out wich setting comes from which policy.

@zhaohuabing `Replace` by the most specific resource is the default https://gateway.envoyproxy.io/v0.6.0/design/security-policy/#design-decisions , and support for `Merge` will hopefully be tackled in the near future with https://github.com/envoyproxy/gateway/issues/1934

zhaohuabing:
> @zhaohuabing Replace by the most specific resource is the default https://gateway.envoyproxy.io/v0.6.0/design/security-policy/#design-decisions , and support for Merge will hopefully be tackled in the near future with https://github.com/envoyproxy/gateway/issues/1934

@arkodg Got it! Do you mind if I fix this in a separate PR?

arkodg:
> > @zhaohuabing Replace by the most specific resource is the default https://gateway.envoyproxy.io/v0.6.0/design/security-policy/#design-decisions , and support for Merge will hopefully be tackled in the near future with #1934
> 
> @arkodg Got it! Do you mind if I fix this in a separate PR?

Sure sgtm !

arkodg (approved):
LGTM thanks !

PR #2694, opened 2024-02-25
feat(EG K8S Provider): Enable leader election for EG controller

alexwo:
### What type of PR is this?
feat(controller): implement leader election for EG controller

### What this PR does / why we need it:
1. This PR introduces leader election for the Envoy Gateway (EG) controller to ensure that only one instance of the controller can perform write operations at any given time.
2. Ensure consistent availability of xDS (Envoy's discovery service) across the system.

The implementation follows best practices for distributed systems, enhancing the reliability and stability of the EG controller in multi-instance deployments.

### Which issue(s) this PR fixes:
This enhancement facilitates the scaling of xDS services and enables leader election support. 

### High Level Changes
1. We don't block controllers startup as we set NeedLeaderElection flag to false by default, apply to GatewayAPIController. 

2. The status handler controller postpones it's startup until it wins the leader election. After being elected, The EG GatewayAPIController initiates a reconciliation loop to process all pending statuses, ensuring that the status handler controller can pick up on all the resources and become up-to-date after assuming leadership.

3. The infra k8s provider is started after a pod is elected as leader, it directly will process all infra resources. no reconcile is required in this case.

4. It's possible to configure leader elections settings, to gain more control over how the leader election process happen, how quick, how resilient. Incase not provided those will default to k8s base values.
```YAML
leaseDuration : 18s
renewDeadline: 10s
retryPeriod: 2s
```

### This use cases were manually validated:

Blocking the GRPC Port of a Leader (with IP Tables Rule)
Result: the Envoy proxy is able to discover the xDS rules from at least one of the remaining EGs.

Blocking the GRPC Port of the leader + all secondary pods except one
Result: Despite the widespread blockage, the Envoy proxy manages to obtain xDS rules from the remaining unblocked EG, maintaining operational resilience.

Apply k8s quickstart ( confirm only leader update status / infra)
Result: status updates and infrastructure are exclusively managed by the leader controller.

K8S control plane API is not available for the leader instance
The leader restarts, however, other instances continue to serve xDS and take leadership when possible
https://github.com/envoyproxy/gateway/issues/1953


guydc:
Should this PR also include some work in the infrastructure controller?

alexwo:
> > Should this PR also include some work in the infrastructure controller?
> > Hi @guydc , We can attempt to ensure that there is only a single controller that can create/update e.g., infra related resources at a time.
> 
> The side effect at this point is that infra updates will result in concurrent updates and extra revision conflict's / retrys.

Added changes to ensure that infra will be created only by the elected EG instance.

alexwo:
I'm keeping this PR in draft mode for now, as I'm looking forward to enhancing it further with some manual testing and broader test coverage.

alexwo:
Leader Election Enabled and Rolling Update for EG and Envoy Proxy
My GIF Description

internal/infrastructure/runner/runner.go
```diff
@@ -41,13 +42,30 @@ func (r *Runner) Start(ctx context.Context) (err error) {
...
 
+	// Wait for leader if leader election is enabled
+	if r.LeaderElection.Enabled {
+		go func() {
+			select {
+			case <-ctx.Done():
+				return
+			case <-r.Elected:
+				initInfra()
+				r.Logger.Info("started")
```
shawnh2, line 63: this log can also be moved to `initInfra` func

internal/envoygateway/config/config.go
```diff
@@ -36,19 +38,72 @@ type Server struct {
...
 		Namespace:    env.Lookup("ENVOY_GATEWAY_NAMESPACE", DefaultNamespace),
 		DNSDomain:    env.Lookup("KUBERNETES_CLUSTER_DOMAIN", DefaultDNSDomain),
 		// the default logger
-		Logger: logging.DefaultLogger(v1alpha1.LogLevelInfo),
+		Logger:         logging.DefaultLogger(v1alpha1.LogLevelInfo),
+		LeaderElection: settings,
 	}, nil
 }
 
+func parseEnvDuration(envVar string, duration **time.Duration) error {
```
shawnh2, line 73: can we also make `duration` as a return value instead of using secondary pointer? (more like a gopher this way.

internal/provider/kubernetes/kubernetes.go
```diff
@@ -73,6 +79,13 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 		return nil, fmt.Errorf("unable to set up ready check: %w", err)
 	}
 
+	// Emit elected & continue with deployment of infra resources
+	go func() {
+		if <-mgr.Elected(); true {
```
shawnh2, line 84:
```suggestion
		if <-mgr.Elected() {
```

alexwo: Switched to using <-mgr.Elected(), given that it's a channel of type struct{}.

tools/make/kube.mk
```diff
@@ -113,6 +113,7 @@ install-ratelimit:
 	kubectl rollout restart deployment envoy-gateway -n envoy-gateway-system
 	kubectl rollout status --watch --timeout=5m -n envoy-gateway-system deployment/envoy-gateway
 	kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
+	tools/hack/deployment-exists.sh "app.kubernetes.io/name=envoy-ratelimit" "envoy-gateway-system"
```
arkodg, line 116: isn't this doing the same thing as L117 ?

internal/provider/runner/runner.go
```diff
@@ -43,7 +44,7 @@ func (r *Runner) Start(ctx context.Context) (err error) {
 		if err != nil {
 			return fmt.Errorf("failed to get kubeconfig: %w", err)
 		}
-		p, err := kubernetes.New(cfg, &r.Config.Server, r.ProviderResources)
+		p, err := kubernetes.New(cfg, &r.Config.Server, r.ProviderResources, r.Elected)
```
arkodg, line 47: this won't be needed, and should be available in `r.Config.Server`

internal/provider/kubernetes/kubernetes.go
```diff
@@ -73,6 +79,12 @@ func New(cfg *rest.Config, svr *config.Server, resources *message.ProviderResour
 		return nil, fmt.Errorf("unable to set up ready check: %w", err)
 	}
 
+	// Emit elected & continue with deployment of infra resources
```
arkodg, line 82: are we blocking here ?
imo we should continue to process watching resources, translating them, and populating the xds cache
we should only block on updating status

arkodg:
finally got to this, thanks for building this out ! adding some comments

internal/provider/kubernetes/kubernetes.go
alexwo: This won't be blocking as the manager is configured with:
Controller:              config.Controller{NeedLeaderElection: ptr.To(false)}
So as expected, we process watching resources, translating them, and populating the xds cache
and only block on updating status and the k8s envoy proxy provider (that can actually deploy envoys) if not elected.

alexwo:
> finally got to this, thanks for building this out ! adding some comments

Thanks!

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -1,5 +1,10 @@
 deployment:
   envoyGateway:
+    leaderElection:
+      enabled: true
```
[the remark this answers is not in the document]
alexwo: updated

internal/infrastructure/runner/runner.go
```diff
@@ -41,14 +42,30 @@ func (r *Runner) Start(ctx context.Context) (err error) {
...
+
+		// Enable global ratelimit if it has been configured.
+		if r.EnvoyGateway.RateLimit != nil {
+			go r.enableRateLimitInfra(ctx)
+		}
+		r.Logger.Info("started")
 	}
 
-	r.Logger.Info("started")
+	// Wait for leader if leader election is enabled
```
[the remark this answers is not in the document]
alexwo: added

internal/provider/runner/runner.go
alexwo: fixed

api/v1alpha1/envoygateway_types.go
```diff
@@ -88,6 +90,19 @@ type EnvoyGatewaySpec struct {
...
 	// +optional
 	ExtensionAPIs *ExtensionAPISettings `json:"extensionApis,omitempty"`
+
+	// LeaderElection defines the settings related to leader election
+	// +optional
+	LeaderElection *LeaderElection
+}
+
+type LeaderElection struct {
+	// LeaseDuration defines the time non-leader contenders will wait before attempting to claim leadership. It's based on the timestamp of the last acknowledged signal. The default setting is 15 seconds.
```
arkodg, line 100: can you follow the kubebuilder way of definition defaults, and also add optional tags
this may not be a CRD, but it helps with doc generation

charts/gateway-helm/values.tmpl.yaml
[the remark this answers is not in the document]
arkodg: this config need to be removed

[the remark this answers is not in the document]
arkodg: and moved under config.envoyGateway

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
```diff
@@ -42,6 +42,14 @@ spec:
         - server
         - --config-path=/config/envoy-gateway.yaml
         env:
+        - name: ENVOY_GATEWAY_LEADER_ELECTION_ENABLED
```
arkodg, line 45: needs to be removed

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -85,6 +86,11 @@ func (e *EnvoyGateway) NamespaceMode() bool {
 		len(e.Provider.Kubernetes.Watch.Namespaces) > 0
 }
 
+// DefaultLeaderElection returns a new LeaderElection with default configuration parameters.
```
arkodg, line 89: decision to make here, if we are enabling by default
* do we do it via helm (instantiate the field) or in process
* how do we disable it ?

good to make a call after looking around to see what the current state of existing CNCF projects is
does the average admin find default leaderElection stanza to be overwhelming or irrelevant

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
alexwo: this allows to disable leader election?

api/v1alpha1/envoygateway_helpers.go
alexwo: We can configure this at the helm level, allowing users to determine how Envoy should initiate upon deployment. Alternatively, we could adhere to the Envoy resource spec, in this case we need to decide on how we startup, probably with leader election enabled by default and possibly alter its mode of operation once the resource is processed. 

i'm uncertain whether disabling Envoy's leader election during runtime is interesting? or we can assume this is an eg deployment setting.
 
Istio leader election is enabled by default as per:

ENABLE_LEADER_ELECTION	Boolean	true	If enabled (default), starts a leader election client and gains leadership before executing controllers. If false, it assumes that only one instance of istiod is running and skips leader election.

https://github.com/istio/istio/pull/41141/files

arkodg: * lets define `envoyGateway.leaderElection: {}` in helm for now
* lets disable leader election if `envoyGateway.leaderElection` is `nil`

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
arkodg: added a related note below

internal/cmd/server.go
```diff
@@ -32,7 +32,7 @@ var (
 func getServerCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "server",
-		Aliases: []string{"serve"},
+		Aliases: []string{"c"},
```
arkodg, line 35: what's happening here

api/v1alpha1/envoygateway_helpers.go
arkodg: or we add a explicit `disabled *bool` flag in the `leaderElection` struct, will bring this up in the community meeting tomorrow

internal/provider/kubernetes/controller.go
```diff
@@ -901,6 +901,13 @@ func (r *gatewayAPIReconciler) addFinalizer(ctx context.Context, gc *gwapiv1.Gat
 
 // watchResources watches gateway api resources.
 func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.Manager, c controller.Controller) error {
+	// trigger a reconcile after getting elected
```
arkodg, line 904: curious why this is needed

internal/provider/kubernetes/kubernetes.go
arkodg: cool can we update the comment here - "Wait until election is complete, and then start reconciling resources"

arkodg: so then the question I have is, I dont see any code that is disabling status writes, how is this magically working :) ?

alexwo: almost, actually the status controller is already good with:
https://github.com/envoyproxy/gateway/blob/a667f23602794a38195cf0318a17e735bfcd33a1/internal/status/status.go#L106

internal/cmd/server.go
alexwo: should not be changed, thanks for catching fixed

internal/provider/kubernetes/controller.go
alexwo: When envoy gateway starts for the first time, it won't be a leader yet, however it will already reconcile resources as the controller main reconcile loop does not need a leader to run.

When the controller get elected, it won't be called again for resources it already reconciled. This is needed to fix the resources status and process infra (create proxies).

arkodg: but it doesn't need to revisit or update status, unless some resource has changed, and in that case the `Reconcile` would be triggered

internal/provider/kubernetes/kubernetes.go
arkodg: ah so the updates via the status updater client is auto disabled  ?

internal/provider/kubernetes/controller.go
alexwo: This ensures that we reconcile resources that have changed before there is an active elected leader.
use cases:
- Initial startup, when resources already exists
- Leader goes down during an outage
- EG upgrades / re-deployments

probably more..

internal/provider/kubernetes/kubernetes.go
alexwo: yes its a runnable, the manager won't call its start unless its a leader.
the client update won't be triggered.

type Runnable interface {
	// Start starts running the component.  The component will stop running
	// when the context is closed. Start blocks until the context is closed or
	// an error occurs.
	Start(context.Context) error
}

internal/provider/kubernetes/controller.go
alexwo: @arkodg 
I think the main point is that k8s does not trigger a reconcile after a leader gets elected. So we have to handle this on our end incase we would like the main reconcile loop to sync xds for all controllers, incase they are not leaders.

We can decide not do this but than we won't be able to serve xds from the other eg instances as it won't be at sync.


1. The main reconcile loop won't wait for leader election, it's start immediately on all pods. 
As there is no option to know which of the pods will get elected, the reconcile loops starts in all instances simultaneously.  we can change that and decide that only after the pod gets elected the reconcile process starts. ( i'm not sure if its more interesting)
(https://github.com/envoyproxy/gateway/blob/1f91c08ccec4df0e886fda603c6ffde3634202a7/internal/provider/kubernetes/controller.go#L136). 

2. This is expected as a pod has to wait a lease duration timeout before it attempts to be a leader, k8s won't trigger a reconcile on leader elected. 
- https://github.com/kubernetes/client-go/blob/v0.29.3/tools/leaderelection/leaderelection.go#L133
```
// LeaseDuration is the duration that non-leader candidates will
// wait to force acquire leadership. This is measured against time of
// last observed ack.
//
// A client needs to wait a full LeaseDuration without observing a change to
// the record before it can attempt to take over. 
```

arkodg: cool, thanks for explaining, it sounds like, we kick the method again to replay the writes, specifically for resources that were updated/created the leaderless interval ? is it a common pattern in most CNCF projects ?

api/v1alpha1/envoygateway_types.go
```diff
@@ -88,6 +90,20 @@ type EnvoyGatewaySpec struct {
 	//
 	// +optional
 	ExtensionAPIs *ExtensionAPISettings `json:"extensionApis,omitempty"`
+
+	// LeaderElection defines the settings related to leader election
+	// +optional
+	LeaderElection *LeaderElection `json:"leaderElection,omitempty"`
```
arkodg, line 96: just realized this lives in the top level today, this needs to be moved into the kubernetes specific provider field

internal/provider/kubernetes/controller.go
alexwo: Yes that's right, our goal is to close the leaderless interval. 
The typical approach in Kubernetes and CNCF ecosystems can be using a specific controller for xDS sync, which focuses solely on xDS operations without altering status or infrastructure. Additionally, a separate controller can be used to perform status updates, and make infrastructure modifications.

Based on this, we can outline that the xDS controller does not need to assume a leadership role, whereas the controller handling status and infrastructure updates does. Consequently, Kubernetes will initiate the reconciliation process.

While it's feasible to develop a distinct controller for managing xDS reconciliations and another solely for infrastructure and status modifications, such an overhaul might be extensive and potentially less appealing within the existing EG architecture.

api/v1alpha1/envoygateway_types.go
```diff
@@ -198,6 +210,9 @@ type EnvoyGatewayKubernetesProvider struct {
 	// OverwriteControlPlaneCerts updates the secrets containing the control plane certs, when set.
 	// +optional
 	OverwriteControlPlaneCerts *bool `json:"overwriteControlPlaneCerts,omitempty"`
+	// LeaderElection defines the settings related to leader election
```
zhaohuabing, line 213: Could we add some comments here to clarify the default behavior if `LeaderElection` is not specified?

Does the absence of `LeaderElectiono` imply that there is no leader election?  Should we always enable leader election and set a sane default? IMO, leader election is an internal mechanism of EG and should not be the user's concern unless necessary.

api/v1alpha1/envoygateway_helpers.go
arkodg: we discussed this in the community meeting yesterday, and we agreed that lets enable it by default, and add an explicit `Disable bool` knob to disable it

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -85,6 +85,11 @@ func (e *EnvoyGateway) NamespaceMode() bool {
...
+func DefaultLeaderElection() *LeaderElection {
+	return &LeaderElection{}
```
arkodg, line 90: this is a good place to assign the default values of the sub fields

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -36,6 +36,8 @@ config:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
     provider:
       type: Kubernetes
+      kubernetes:
```
arkodg, line 39: this will no longer be needed, since `leaderElection` will be enabled by default

internal/provider/kubernetes/controller.go
arkodg: sounds good, lets stick with this, can you also add a comment, explaining the why e.g. we kick the method again "Retrigger a reconcile after getting elected to replay the translation ( which will trigger status writes and/or infra resource creation) , specifically for resources that were updated/created during the leaderless interval"

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -148,6 +153,9 @@ func DefaultEnvoyGatewayPrometheus() *EnvoyGatewayPrometheusProvider {
 func DefaultEnvoyGatewayProvider() *EnvoyGatewayProvider {
 	return &EnvoyGatewayProvider{
 		Type: ProviderTypeKubernetes,
+		Kubernetes: &EnvoyGatewayKubernetesProvider{
```
arkodg, line 156: suggest move this code into `GetEnvoyGatewayKubeProvider`

api/v1alpha1/validation/envoygateway_validate_test.go
```diff
@@ -646,7 +646,6 @@ func TestEnvoyGatewayProvider(t *testing.T) {
 	assert.NotNil(t, envoyGateway.Provider)
 
 	envoyGatewayProvider := envoyGateway.GetEnvoyGatewayProvider()
-	assert.Nil(t, envoyGatewayProvider.Kubernetes)
```
arkodg, line 649: why is this removed ?

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -4,7 +4,7 @@ deployment:
       repository: ${ImageRepository}
       tag: '${ImageTag}'
     imagePullPolicy: Always
-    imagePullSecrets: []
```
arkodg, line 7: can we undo changes in this file ?

internal/provider/kubernetes/kubernetes_test.go
```diff
@@ -154,15 +159,8 @@ func testGatewayClassAcceptedStatus(ctx context.Context, t *testing.T, provider
 func testGatewayClassWithParamRef(ctx context.Context, t *testing.T, provider *Provider, resources *message.ProviderResources) {
 	cli := provider.manager.GetClient()
 
-	// Create the namespace for the test case.
 	// Note: The namespace for the EnvoyProxy must match EG's configured namespace.
 	testNs := config.DefaultNamespace
-	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNs}}
-	require.NoError(t, cli.Create(ctx, ns))
```
arkodg, line 161: so we moved the namespace creation from one test to the other ?

charts/gateway-helm/values.tmpl.yaml
alexwo: yes fixing it

api/v1alpha1/envoygateway_types.go
```diff
@@ -90,6 +92,21 @@ type EnvoyGatewaySpec struct {
...
+	// Default is 15 seconds.
+	LeaseDuration *time.Duration `json:"leaseDuration,omitempty"`
+	// RenewDeadline represents the time frame within which the current leader will attempt to renew its leadership status before relinquishing its position. The default setting is 10 seconds.
+	// Default is 10 seconds.
+	RenewDeadline *time.Duration `json:"renewDeadline,omitempty"`
+	// RetryPeriod denotes the interval at which LeaderElector clients should perform action retries. The default setting is 2 seconds.
+	// Default is 15 seconds.
+	RetryPeriod *time.Duration `json:"retryPeriod,omitempty"`
+	// Disabled allows to disable leader election (enabled by default)
+	Disabled *bool `json:"disabled,omitempty"`
```
arkodg, line 107: we are using Disable in other APIs, lets use `Disable` here
e.g. https://github.com/envoyproxy/gateway/blob/0f87fcbad1aaa2913af3f768da87ea72a695a14d/api/v1alpha1/envoygateway_metrics_types.go#L47

alexwo: indeed i will change it

internal/provider/kubernetes/kubernetes_test.go
alexwo: actually I think both tests needs the namespace, let me check it

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -39,6 +41,11 @@ func (e *EnvoyGateway) SetEnvoyGatewayDefaults() {
 	if e.Provider == nil {
 		e.Provider = DefaultEnvoyGatewayProvider()
 	}
+	if e.Provider.Kubernetes == nil {
```
arkodg, line 44: is this needed ?
aren't you performing the same thing again inside `DefaultGateway` ?

internal/provider/kubernetes/kubernetes_test.go
alexwo: Since all tests now require a namespace for leader election(k8s uses that), we initialize it within TestProvider, which subsequently initiates the other tests. This is why the namespace init was removed from testGatewayClassWithParamRef, it can just reuse the namespace.

```go
	testcases := map[string]func(context.Context, *testing.T, *Provider, *message.ProviderResources){
		"gatewayclass controller name":         testGatewayClassController,
		"gatewayclass accepted status":         testGatewayClassAcceptedStatus,
		"gatewayclass with parameters ref":     testGatewayClassWithParamRef,
		"gateway scheduled status":             testGatewayScheduledStatus,
		"httproute":                            testHTTPRoute,
		"tlsroute":                             testTLSRoute,
		"stale service cleanup route deletion": testServiceCleanupForMultipleRoutes,
	}
```

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -195,9 +212,18 @@ func (r *EnvoyGatewayProvider) GetEnvoyGatewayKubeProvider() *EnvoyGatewayKubern
...
+			r.Kubernetes.LeaderElection = DefaultLeaderElection()
+		}
 		return r.Kubernetes
 	}
 
+	if r.Kubernetes.LeaderElection == nil {
+		r.Kubernetes.LeaderElection = DefaultLeaderElection()
+	}
+
+	r.Kubernetes.LeaderElection = DefaultLeaderElection()
```
arkodg, line 225: why is this needed ?

internal/provider/kubernetes/kubernetes_test.go
arkodg: sounds good thanks

api/v1alpha1/envoygateway_helpers.go
alexwo: The DeafultGateway is overridden by the envoy cfg, this helps to ensure the defaults remain.
Incase we don't do it we need to set a default value in the envoy confg (cfgPath), which we did not want to?

```
	eg, err := config.Decode(cfgPath)
		if err != nil {
			logger.Error(err, "failed to decode config file", "name", cfgPath)
			return nil, err
		}
		// Set defaults for unset fields
		eg.SetEnvoyGatewayDefaults()
		cfg.EnvoyGateway = eg
```

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -195,6 +212,9 @@ func (r *EnvoyGatewayProvider) GetEnvoyGatewayKubeProvider() *EnvoyGatewayKubern
 
 	if r.Kubernetes == nil {
 		r.Kubernetes = DefaultEnvoyGatewayKubeProvider()
+		if r.Kubernetes.LeaderElection == nil {
```
arkodg, line 215: what about the case where `r.Kubernetes` is not nil, but `r.Kubernetes.LeaderElection` is `nil`

alexwo: indeed adding

api/v1alpha1/envoygateway_types.go
alexwo: added

alexwo: we will have a disable leader election setting to disable leader election, by default leader election is always on.
Users won't need to do anything unless they decide to disable it.

api/v1alpha1/envoygateway_helpers.go
alexwo: fixed

arkodg (approved):
LGTM, great to see this feature making it in
 thanks for addressing all the feedback !

zhaohuabing (approved):
LGTM. Thank you for all the effort and patience!

PR #2704, opened 2024-02-26
fix(helm): refactor monitoring service

nipsufn:
**What type of PR is this?**

fix / refactor monitoring service for envoy-gateway deployment

- deployment needs to expose port for it to be accessible via service
- there's no need to expose whole `ports` field from metrics service
- the port on deployment is static as it cannot be adjusted (https://gateway.envoyproxy.io/latest/design/eg-metrics/#envoygatewaymetrics)

N/A

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-metrics-service.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-metrics-service.yaml
@@ -1,6 +1,9 @@
 apiVersion: v1
 kind: Service
 metadata:
+  annotations:
+    prometheus.io/scrape: 'true'
+    prometheus.io/port: '19001'
   name: envoy-gateway-metrics-service
   namespace: '{{ .Release.Namespace }}'
   labels:
@@ -11,4 +14,7 @@ spec:
     control-plane: envoy-gateway
   {{- include "eg.selectorLabels" . | nindent 4 }}
   ports:
-	{{- .Values.envoyGatewayMetricsService.ports | toYaml | nindent 2 -}}
+  - name: http
+    port: {{ .Values.envoyGatewayMetricsService.port }}
+    protocol: TCP
+    targetPort: http-metrics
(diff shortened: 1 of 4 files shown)


arkodg:
thanks for finding this bug @nipsufn 
should we instead be defining the deployment port here https://github.com/envoyproxy/gateway/blob/72fadb7c1dbe7d660c2900527c34f53cfcfce7ac/charts/gateway-helm/values.tmpl.yaml#L15 ? cc @Xunzhuo

nipsufn:
It looks to me that `.Values.deployment.ports` is used in `envoy-gateway-service.yaml` directly...
If you think about dropping `.Values.envoyGatewayMetricsService` field then I see couple options:
- add `monitoring` port to `.Values.deployment.ports`, drop `envoy-gateway-metrics-service.yaml` and make users scrape metrics against `envoy-gateway-service.yaml`
- add `.Values.deployment.monitoringPorts` and use that for `envoy-gateway-metrics-service.yaml`

I'll also template already present `.Values.deployment.ports` to container spec on the deployment as they are missing.

zirain (dismissed):

charts/gateway-helm/templates/envoy-gateway-metrics-service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
```
arkodg, line 1: is there any strong reason for having this capabilities condition ?
for the case when this chart is installed before promethus is installed, this service wont get installed, which is not ideal, we ideally dont want to govern the order of install of components, so I'd vote to rm it

Xunzhuo: +1

zirain:
please fix DCO

BTW, can you also set prometheus annotations?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2707, opened 2024-02-26
e2e: basic multiple gc test

cnvergence:
**What type of PR is this?**

Adopted from https://gateway-api.sigs.k8s.io/api-types/gatewayclass/#gatewayclass

Fixes https://github.com/envoyproxy/gateway/issues/3364

### DIFF
--- a/test/e2e/multiple_gc/multiple_gc_test.go
+++ b/test/e2e/multiple_gc/multiple_gc_test.go
@@ -69,9 +69,9 @@ func TestMultipleGC(t *testing.T) {
 		internetGatewaySuite.Applier.GatewayClass = internetGatewaySuiteGatewayClassName
 		internetGatewaySuite.ControllerName = kubernetes.GWCMustHaveAcceptedConditionTrue(t, internetGatewaySuite.Client, internetGatewaySuite.TimeoutConfig, internetGatewaySuite.GatewayClassName)
 
-		t.Logf("Running %d MultipleGC tests", len(tests.InternetGCTests))
+		t.Logf("Running %d MultipleGC tests", len(tests.MultipleGCTests[internetGatewaySuiteGatewayClassName]))
 
-		err = internetGatewaySuite.Run(t, tests.InternetGCTests)
+		err = internetGatewaySuite.Run(t, tests.MultipleGCTests[internetGatewaySuiteGatewayClassName])
 		if err != nil {
 			t.Fatalf("Failed to run InternetGC tests: %v", err)
 		}
@@ -79,16 +79,7 @@ func TestMultipleGC(t *testing.T) {
 
 	t.Run("Private GC Test", func(t *testing.T) {
 		t.Parallel()
-
 		privateGatewaySuiteGatewayClassName := "private"
-		if flags.RunTest != nil && *flags.RunTest != "" {
-			t.Logf("Running E2E test %s with %s GatewayClass\n cleanup: %t\n debug: %t",
-				*flags.RunTest, *flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
-		} else {
-			t.Logf("Running E2E tests with %s GatewayClass\n cleanup: %t\n debug: %t",
-				*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
-		}
-
 		privateGatewaySuite, err := suite.NewConformanceTestSuite(suite.ConformanceOptions{
 			Client:               c,
 			GatewayClassName:     privateGatewaySuiteGatewayClassName,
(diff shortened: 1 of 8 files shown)


arkodg:
@cnvergence we could circumvent this issue by following what @shawnh2 did in https://github.com/envoyproxy/gateway/commit/decefccae531692b2e9b99929bf868deeb71b0e5 where he manually creates and applies the resources via the `client` API 😆

cnvergence:
> @cnvergence we could circumvent this issue by following what @shawnh2 did in [decefcc](https://github.com/envoyproxy/gateway/commit/decefccae531692b2e9b99929bf868deeb71b0e5) where he manually creates and applies the resources via the `client` API 😆

I see 😄 and still ended up using suite client, I will then try this as well

test/e2e/multiple_gc/multiple_gc_test.go
```diff
@@ -0,0 +1,119 @@
...
+	e2e.CheckInstallScheme(t, c)
+
+	if flags.RunTest != nil && *flags.RunTest != "" {
+		t.Logf("Running E2E test %s with %s GatewayClass\n cleanup: %t\n debug: %t",
+			*flags.RunTest, *flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+	} else {
+		t.Logf("Running E2E tests with %s GatewayClass\n cleanup: %t\n debug: %t",
+			*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+	}
+	t.Run("GC1Test", func(t *testing.T) {
```
arkodg, line 48: internet-gc-test so its clearer during debugging ?

test/e2e/multiple_gc/multiple_gc_test.go
```diff
@@ -0,0 +1,119 @@
...
+		// since this test suite reuse the base resources of previous test suite.
+		internetGatewaySuite.Applier.ManifestFS = []fs.FS{e2e.Manifests}
+		internetGatewaySuite.Applier.GatewayClass = internetGatewaySuiteGatewayClassName
+		internetGatewaySuite.ControllerName = kubernetes.GWCMustHaveAcceptedConditionTrue(t, internetGatewaySuite.Client, internetGatewaySuite.TimeoutConfig, internetGatewaySuite.GatewayClassName)
+
+		t.Logf("Running %d MultipleGC tests", len(tests.InternetGCTests))
+
+		err = internetGatewaySuite.Run(t, tests.InternetGCTests)
+		if err != nil {
+			t.Fatalf("Failed to run MergeGateways tests: %v", err)
```
arkodg, line 76:
```suggestion
			t.Fatalf("Failed to run InternetGC tests: %v", err)
```

test/e2e/multiple_gc/multiple_gc_test.go
```diff
@@ -0,0 +1,119 @@
...
+		t.Parallel()
+
+		privateGatewaySuiteGatewayClassName := "private"
+		if flags.RunTest != nil && *flags.RunTest != "" {
+			t.Logf("Running E2E test %s with %s GatewayClass\n cleanup: %t\n debug: %t",
+				*flags.RunTest, *flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+		} else {
+			t.Logf("Running E2E tests with %s GatewayClass\n cleanup: %t\n debug: %t",
+				*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+		}
```
shawnh2, line 90: curious why we need this ? do we already have this check as in L41~L47

test/e2e/tests/tests.go
```diff
@@ -14,4 +14,6 @@ var (
 	ConformanceTests   []suite.ConformanceTest
 	UpgradeTests       []suite.ConformanceTest
 	MergeGatewaysTests []suite.ConformanceTest
+	InternetGCTests    []suite.ConformanceTest
```
shawnh2, line 17: not a fan of this naming. should we just go as `MultipleGCTests` with type `map[string][]suite.ConformanceTest`, the GC name can be used as the map key ?

test/e2e/multiple_gc/multiple_gc_test.go
cnvergence: good catch, I made mistake during refactoring this PR

test/e2e/tests/tests.go
cnvergence: I agree, let's change the name for the tests, but I will keep this scenario naming just to keep up with the gw-api example

arkodg (approved):
LGTM thanks !

PR #2708, opened 2024-02-26
fix: Configure idle timeout when timeout is set on HTTPRoute

davidalger:
**What type of PR is this?**

Fixes bug where default `stream_idle_timeout` in connection manager interferes with request timeouts set on HTTPRoute when set greater than 5 minutes preventing high request timeouts from being used without patching connection manager settings.

**What this PR does**:

If an HTTP request `timeout` is configured on the route and is non-zero, the `idle_timeout` will be set to the greater of 1hr or the configured request timeout. If the `timeout` on the HTTPRoute is set to `0` the `idle_timeout` is also set to `0` disabling both timeouts on the route.

**How to reproduce**:

1. Deploy demo into kind cluster and set a fault injection delay of 500s:

```yaml
kind: BackendTrafficPolicy
metadata:
  name: fault-injection-delay
spec:
  targetRef:
    kind: HTTPRoute
    name: backend
  faultInjection:
    delay:
      fixedDelay: 500s
```

2. Configure timeout on HTTPRoute:

```yaml
kind: HTTPRoute
metadata:
  name: backend
spec:
  hostnames:
  - www.example.com
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - kind: Service
      name: backend
      port: 3000
    matches:
    - path:
        type: PathPrefix
        value: /
    timeouts:
      request: 600s
```

3. Issue request via `curl` noting the 504 response code at the 5 minute mark

```bash
$ time curl -sv -H "Host: www.example.com" http://172.18.255.200/
*   Trying 172.18.255.200:80...
* Connected to 172.18.255.200 (172.18.255.200) port 80
> GET / HTTP/1.1
> Host: www.example.com
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 504 Gateway Timeout
< content-length: 14
< content-type: text/plain
< date: Mon, 26 Feb 2024 18:20:32 GMT
< 
* Connection #0 to host 172.18.255.200 left intact
stream timeout
real	5m0.072s
user	0m0.017s
sys	0m0.023s
```

4. Note the reason for the early timeout in the request log:

```
"response_code":"504","response_flags":"DI,SI","response_code_details":"stream_idle_timeout"
```

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -8,6 +8,7 @@ package translator
 import (
 	"errors"
 	"strings"
+	"time"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	routev3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
@@ -62,7 +63,7 @@ func buildXdsRoute(httpRoute *ir.HTTPRoute) (*routev3.Route, error) {
 			// If there are invalid backends then a weighted cluster is required for the route
 			routeAction = buildXdsWeightedRouteAction(httpRoute)
 		} else {
-			routeAction = buildXdsRouteAction(httpRoute.Destination.Name)
+			routeAction = buildXdsRouteAction(httpRoute)
 		}
 		if httpRoute.Mirrors != nil {
 			routeAction.RequestMirrorPolicies = buildXdsRequestMirrorPolicies(httpRoute.Mirrors)
@@ -207,11 +208,12 @@ func buildXdsStringMatcher(irMatch *ir.StringMatch) *matcherv3.StringMatcher {
 	return stringMatcher
 }
 
-func buildXdsRouteAction(destName string) *routev3.RouteAction {
+func buildXdsRouteAction(httpRoute *ir.HTTPRoute) *routev3.RouteAction {
 	return &routev3.RouteAction{
 		ClusterSpecifier: &routev3.RouteAction_Cluster{
-			Cluster: destName,
+			Cluster: httpRoute.Destination.Name,
 		},
+		IdleTimeout: idleTimeout(httpRoute),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/xds/translator/route.go
```diff
@@ -239,9 +241,27 @@ func buildXdsWeightedRouteAction(httpRoute *ir.HTTPRoute) *routev3.RouteAction {
 				Clusters: clusters,
 			},
 		},
+		IdleTimeout: idleTimeout(httpRoute),
 	}
 }
 
+func idleTimeout(httpRoute *ir.HTTPRoute) *durationpb.Duration {
```
arkodg, line 248: I'd vote for below logic until we expose this as an API field
```
idleTimeout = 1hr
if idleTimout < requestTimeout {
    idleTimout = requestTimeout
}

Comment (davidalger) [reply internal/xds/translator/route.go:248]: I'm ok with this, with the caveat that it should still be set to zero so the request timeout will function per GEP-1742:

> A zero-valued timeout ("0s") MUST be interpreted as disabling the timeout
ref: https://gateway-api.sigs.k8s.io/geps/gep-1742/#timeout-values

This should work per envoy docs:

> value of 0 will completely disable the route’s idle timeout, even if a connection manager stream idle timeout is configured

ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-idle-timeout

Comment (arkodg) [reply internal/xds/translator/route.go:248]: yah agree with that

File: internal/xds/translator/testdata/out/xds-ir/http-route-timeout.routes.yaml [hunk 2]
```diff
@@ -14,6 +14,25 @@
       name: first-route
       route:
         cluster: first-route-dest
+        idleTimeout: 3600s
```
arkodg, line 17: nit: would be ideal if this wasn't set

arkodg (approved):
LGTM thanks !

zirain (approved):

internal/xds/translator/testdata/out/xds-ir/http-route-timeout.routes.yaml
davidalger: Two choices:

a) 1hr default set on the route only when the request timeout is passed on the route (as currently implemented)

b) Set `stream_idle_timeout` in the connection manager to 1hr (vs implicit default of 300s), only configuring per-route idle timeout if >1hr or 0

thoughts?

arkodg: Prefer b

Issue #2685, opened 2024-02-23
Client Traffic Policy: Support Connection Limiting


guydc:
*Description*:
Limiting active connections for an edge proxy is a security and resilience best practice, as described in the Envoy [edge best practices](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#best-practices-edge) document. 

When a single Envoy deployment is proxying multiple backends, it could be beneficial to apply granular downstream connection limits, e.g. per `Gateway` or `Listener`.  This creates a degree of resource isolation between different backends exposed by the proxy and allows Envoy Gateway administrators to allocate resources based on the expected volume of traffic, SLAs, etc. 

Envoy can limit L4 connections in various levels:
- [Process Connection Limits in overload manager](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/overload_manager/overload_manager#limiting-active-connections) (across all listeners): #1966
- [Listener Connection Limits](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/runtime): #1048
- [Filter Chain Connection Limits](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/connection_limit_filter#config-network-filters-connection-limit) : applies connection limits on filter chains

EG's Client Traffic Policy can be extended to support connection limits. The best candidate for implementation is the Envoy [Network Connection Limit Filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/connection_limit_filter#config-network-filters-connection-limit).

Issue #2670, opened 2024-02-22
Socket options SO_SNDBUF and SO_RCVBUF support


modatwork:
*Description*:

Envoy will handle massive connections when used as L4 proxy. Setting the `SO_SNDBUF` and `SO_RCVBUF` socket options helps to limit the memory usage on each connection. So it would be nice to have these options in the APIs.

[optional *Relevant Links*:]

For downstream, Envoy has a `socket_options` in the listener config, https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/config/listener/v3/listener.proto#config-listener-v3-listener

For upstream, Envoy has a `socket_options ` in the `upstream_bind_config` field of the cluster config, https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster

arkodg:
relates to https://github.com/envoyproxy/gateway/issues/2600

aoledk:
How about adding socket options array to ClientTrafficPolicy and BackendTrafficPolicy, for setting socket options in xDS listener and xDS cluster, to limit kernel socket's memory consumption.

zirain:
@arkodg is this fixed?

aoledk:
> @arkodg is this fixed?

No, this feature's api is still not defined. If we **should** do, I can work on this.

shawnh2:
reopen this one since lack of implementation

Issue #2600, opened 2024-02-12
Add a knob to make `per_connection_buffer_limit_bytes` configurable


arkodg:
Interesting discussion [here](https://github.com/envoyproxy/envoy/issues/8448) on compression performance and buffer sizing. Should we make `per_connection_buffer_limit_bytes` configurable as part of this issue? 

I can also think of future use cases that include buffering filters (ext-auth with body, WAF, ext-proc, lua, buffer) that should be part of the Envoy Gateway scope and require buffer size tuning.

_Originally posted by @guydc in https://github.com/envoyproxy/gateway/issues/2451#issuecomment-1937790606_

Currently default to 32Kb
https://github.com/envoyproxy/gateway/pull/1306

yaelShechter:
Please assign to me

PR #2709, opened 2024-02-26
api: connection limits

guydc:
 Adds a container for connection-level settings to CTP called `Connection`. Currently, connection will only support the `Limit` container for setting the max concurrent limits and close delay. Limits will be enforced using the Envoy [Connection Limit Filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/connection_limit/v3/connection_limit.proto#envoy-v3-api-msg-extensions-filters-network-connection-limit-v3-connectionlimit). 

When a CTP with a connection limit attaches to a `Gateway`, each `Listener` will inherit the parent gateway's limit. This proposal does not deal with Envoy [listener runtime connection limits](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/runtime).

In the future, we can consider using this container for #2600, #2670, downstream TCP keepalive, connection local rate limits, ...

Relates to #2685


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -88,6 +88,10 @@ type ClientTrafficPolicySpec struct {
 	//
 	// +optional
 	Timeout *ClientTimeout `json:"timeout,omitempty"`
+	// Connection provides configuration of client connection settings.
```
arkodg, line 91: thoughts on calling this `circuitBreaker` ?

guydc: Yea, `Connection` is a very vague name... I feel like `CircuitBreaker` implies that some sort of failure or unexpected degradation in service triggers the blocking behavior, which isn't really the case here. WDYT about `Resources` or `Limits`?

guydc: Some examples of how this feature is named in other products:
- Nginx: `limit_conn` - https://nginx.org/en/docs/http/ngx_http_limit_conn_module.html
- Kemp: `client connection limiter` (this is actually more of a connection rate limiter than a total connection limiter) - https://support.kemptechnologies.com/hc/en-us/articles/115004547103-How-To-Limit-The-Number-Of-Connections-Per-Second-From-A-Specific-Host
- HAProxy: `maxconn` - https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#maxconn
- Traefik: `InFlightConn` - https://doc.traefik.io/traefik/v3.0/middlewares/tcp/inflightconn/
- F5: `Connection Limits` - https://techdocs.f5.com/kb/en-us/products/big-ip_ltm/manuals/product/ltm-implementations-13-1-0/25.html
- Gloo Edge: `ConnectionLimits`- https://docs.solo.io/gloo-edge/main/guides/security/limit-connection/

So, I feel that `Connection[Limits|Settings|Options]` is a pretty intuitive name for a container which can also include other settings from the timeout and circuit breaker domains:

```yaml
connectionSettings:
  maxParallelConnections: 10000 # Total *[CircuitBreaker;Symmetric]*
  maxNewConnectionsPerSecond: 100 # [NEW;Downstream] CPS rate limit
  maxConnectionBufferSize: 100kb # *[NEW;Symmetric]* per connection memory allocation
  maxConnectionIdleTime: 600s # *[Timeout; Symmetric]*
  http:
    maxRequestsPerConnection: 5000 # *[CircuitBreaker; Symmetric]* close connection after this amount of request
    maxConnectionDuration: 3600s # *[CircuitBreaker; Symmetric]* close connection after this duration
    maxParallelRequests: 1000 # *[CircuitBreaker; Upstream]*  ATM (no envoy downstream equivalent that I know of, but could maybe be implemented through a simplified concurrency controller for adaptive concurrency filter. https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/adaptive_concurrency_filter#concurrency-controllers). 
    maxPendingRequest: 500 # *[CircuitBreaker; Upstream]* only for upstream, don't think that it makes sense for envoy to queue downstream requests... 
```

arkodg: thanks for doing the research !
I'm on board with the top level `connection` field now :) 
my only follow up since `limit` is not a struct but a value, how do we expose the `delay` knob, which is useful to thwart ddos

api/v1alpha1/connection_types.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+type ConnectionLimit struct {
+	// MaxActiveConnections is the maximum concurrent connections limit.
+	// When the limit is reached, incoming connections will be closed immediately.
+	// Default: unlimited.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MaxActiveConnections *int64 `json:"maxActiveConnections,omitempty"`
```
arkodg, line 25: thoughts on `value` ? , else `connections.limi.maxActiveConnections` seems repetitive  ...

api/v1alpha1/connection_types.go
```diff
@@ -0,0 +1,33 @@
...
+// Connection allows users to configure connection-level settings
+type Connection struct {
+	// Limit defines limits related to connections
+	//
+	// +optional
+	Limit *ConnectionLimit `json:"limit,omitempty"`
+}
+
+type ConnectionLimit struct {
+	// Value is the maximum concurrent connections limit.
```
arkodg, line 19:
```suggestion
	// Value of the maximum concurrent connections limit.
```

api/v1alpha1/connection_types.go
```diff
@@ -0,0 +1,33 @@
...
+	// When the limit is reached, incoming connections will be closed after the CloseDelay duration.
+	// Default: unlimited.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	Value *int64 `json:"value,omitempty"`
+
+	// CloseDelay defines the delay to use before closing connections that are rejected
+	// because of of the Value limit.
```
arkodg, line 28:
```suggestion
	// once the limit value is reached.
```

zhaohuabing (commented):
Will [Local rate limit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/connection_limit/v3/connection_limit.proto#envoy-v3-api-msg-extensions-filters-network-connection-limit-v3-connectionlimit) be included in this container?

zhaohuabing (approved):
Just one small thing, “Connection" looks a bit vague, probably ConnectionSetting ? I'm not good at naming anyway.

Not blocking :-)

arkodg (approved):
LGTM thanks !

Issue #2690, opened 2024-02-24
e2e: JWT Claim based routing


arkodg:
*Description*:
>Describe the issue.

Relates to https://github.com/envoyproxy/gateway/issues/2452

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #2712, opened 2024-02-27
e2e: add test for jwt claim base routing

zirain:
Fixes: https://github.com/envoyproxy/gateway/issues/2690

### DIFF
--- a/test/e2e/testdata/jwt.yaml
+++ b/test/e2e/testdata/jwt.yaml
@@ -1,0 +1,61 @@
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: SecurityPolicy
+metadata:
+  name: jwt-example
+  namespace: gateway-conformance-infra
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: jwt-claim-routing
+  jwt:
+    providers:
+      - name: example
+        recomputeRoute: true
+        claimToHeaders:
+          - claim: sub
+            header: x-sub
+          - claim: admin
+            header: x-admin
+          - claim: name
+            header: x-name
+        remoteJWKS:
+          uri: https://raw.githubusercontent.com/envoyproxy/gateway/main/examples/kubernetes/jwt/jwks.json
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: jwt-claim-routing
+  namespace: gateway-conformance-infra
+spec:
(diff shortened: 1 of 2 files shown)


test/e2e/tests/jwt.go
```diff
@@ -0,0 +1,80 @@
...
+							"Authorization": "Bearer " + v1Token,
+						},
+					},
+					Backend: "infra-backend-v1",
+					Response: http.Response{
+						StatusCode: 200,
+					},
+					Namespace: ns,
+				},
+				{
```
arkodg, line 58: lets add another test case to ensure this cannot be spoofed
```
					Request: http.Request{
						Path: "/get",
						Headers: map[string]string{
							"x-name": "Tom",
						},
					},
					Backend: "infra-backend-v2",
					Response: http.Response{
						StatusCode: 401,
					},
					Namespace: ns,
				},
```

test/e2e/testdata/jwt.yaml
```diff
@@ -0,0 +1,61 @@
...
+          port: 8080
+          weight: 1
+      matches:
+        - headers:
+            - name: x-name
+              value: Tom
+    # catch all
+    - backendRefs:
+        - kind: Service
+          name: infra-backend-v3
```
arkodg, line 55: lets make this non existent (`infra-backend-invalid`) since that is what users will most likely do

test/e2e/tests/jwt.go
```diff
@@ -74,7 +74,7 @@ var JWTTest = suite.ConformanceTest{
 					Request: http.Request{
 						Path: "/get",
 						Headers: map[string]string{
-							"Authorization": "Bearer " + v3Token,
+							"Authorization": "Bearer " + invalidToken,
```
arkodg, line 77: we probably need a valid token with a different user here

test/e2e/testdata/jwt.yaml
zirain: TID, I used to think it will be ignored if name is not found.

arkodg: status will be set to ResolvedRef=False and requests will get `500`
we ideally need something like https://github.com/envoyproxy/gateway/issues/2714 but it doesnt exist today :(

zirain: it's added to istio recently.

arkodg (approved):
LGTM thanks for adding this ! 
cc @zetaab

zhaohuabing (approved):

PR #2713, opened 2024-02-27
chore: add crd categories

zirain:
with this patch, you can get all eg resources with following command:
```shell
kubctl get envoy-gateway
```


liorokman (commented):
Maybe also add the `gateway-api` category?

zirain:
> Maybe also add the `gateway-api` category?

I think we should different with upstream.

liorokman (approved):
LGTM

arkodg (approved):
🚀

PR #2721, opened 2024-02-28
chore: update TestDeployment

zirain:
update TestDeployment, so the `default` test output became same as the real world


internal/infrastructure/kubernetes/proxy/resource_provider_test.go
```diff
@@ -550,6 +544,15 @@ func TestDeployment(t *testing.T) {
 				})
 			}
 
+			if os.Getenv("UPDATE_TESTDATA") != "" {
```
arkodg, line 547: can we use this approach https://github.com/envoyproxy/gateway/blob/26d9d9794f0a01bc584bd0273fcfd902467115a0/internal/cmd/egctl/translate_test.go#L29

https://github.com/envoyproxy/gateway/blob/26d9d9794f0a01bc584bd0273fcfd902467115a0/tools/make/golang.mk#L49

Xunzhuo (commented):
Have you tested if these files will be updated, each time we ran `make testdata`

zirain:
> Have you tested if these files will be updated, each time we ran `make testdata`

yes

Xunzhuo:
Thanks for confirming, I met this issue when I firstly add `make testdata` to eg.

arkodg (approved):

zhaohuabing (approved):

PR #2723, opened 2024-02-28
feat(helm): Allow supplying custom configmap

Demonsthere:
**What type of PR is this?**

- feature

This PR allows 2 thing:
1. Change the name of the envoy-gateway cm using an override
2. Allow the control of creating this cm

Those 2 options allow us to supply our own configmap, as well as keep the current schema of deploying it by default. This way one can disable the default config and sideload it when using this as a subchart

Fixes #


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -26,6 +26,10 @@ deployment:
     labels: {}
 
 config:
```
arkodg, line 28: can this be solved by setting `config: null` via values and using `{{- if .Values.config }}` in the template ?

arkodg: we could expose the configMap `name` ( defaults to `envoy-gateway-config`) using `config.name`

arkodg: another way to arrange this is
```
config:
  envoyGateway:
    name: ........
    value: ........
```

thoughts @Xunzhuo ?

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -26,6 +26,10 @@ deployment:
...
+  # -- Switch to false to prevent creating the configmap
+  enabled: true
```
Xunzhuo, line 30: This looks weird to me, what problems are we trying to solve?

arkodg: some users may not want the EG Helm Chart to create the configmap, but would like to hand write it themselves, but would still want the EG Deployment to mount that user defined configmap

charts/gateway-helm/values.tmpl.yaml
Demonsthere: sure, we can set the default name to `envoy-gateway-config`, i have chosen to hardcore it in the templates as i have seen that this is the default name hardcoded in the controller logic. Maybe we want to split the fields and leave `config.envoyGateway` as is, and have separate
```
configmap:
  enabled: true
  nameOverride: ""
```

shahar-h:
Any progress with this? We are also interested in this feature?

arkodg:
@shahar-h couldn't you input the same intent directly via Values.yaml ?

shahar-h:
@arkodg in our case `extensionManager.service.host` represents a kubernetes service host  - `<service-name>.<namespace>.svc.cluster.local`, and since the namespace part is dynamic according to chart installation command we need to have an option to make it templated. In case we have an option to provide our own configmap we can achieve that.

arkodg:
> @arkodg in our case `extensionManager.service.host` represents a kubernetes service host - `<service-name>.<namespace>.svc.cluster.local`, and since the namespace part is dynamic according to chart installation command we need to have an option to make it templated. In case we have an option to provide our own configmap we can achieve that.

since you are supplying the namespace in the helm command, could you also supply the same value to the values field, i.e. your helm cmd invocation is templated which seems like the common case, but the namespace will change ?
im trying to see if we can avoid this, but if not, lets find the right UX and get this in

shahar-h:
> > @arkodg in our case `extensionManager.service.host` represents a kubernetes service host - `<service-name>.<namespace>.svc.cluster.local`, and since the namespace part is dynamic according to chart installation command we need to have an option to make it templated. In case we have an option to provide our own configmap we can achieve that.
> 
> since you are supplying the namespace in the helm command, could you also supply the same value to the values field, i.e. your helm cmd invocation is templated which seems like the common case, but the namespace will change ? im trying to see if we can avoid this, but if not, lets find the right UX and get this in

I could send it from external chart values but this means that I need to send the whole service host, which means that I need to know the internal service name, which I want to avoid.

arkodg:
> > > @arkodg in our case `extensionManager.service.host` represents a kubernetes service host - `<service-name>.<namespace>.svc.cluster.local`, and since the namespace part is dynamic according to chart installation command we need to have an option to make it templated. In case we have an option to provide our own configmap we can achieve that.
> > 
> > 
> > since you are supplying the namespace in the helm command, could you also supply the same value to the values field, i.e. your helm cmd invocation is templated which seems like the common case, but the namespace will change ? im trying to see if we can avoid this, but if not, lets find the right UX and get this in
> 
> I could send it from external chart values but this means that I need to send the whole service host, which means that I need to know the internal service name, which I want to avoid.

sounds like to want the `ConfigMap` to be a template in other Helm Chart ? If this is the case, you'll need to enforce order to make sure the ConfigMap exists before Envoy Gateway Helm Chart kicks in

shahar-h:
I want the configmap to be installed as a template of a EG wrapper helm chart. Regarding the installation order, configmap is installed Before deployment(see https://helm.sh/docs/intro/using_helm/) so there is no issue here.

arkodg:
> cc @arkodg - WDYT?

the use case of supplying an external configmap makes sense, can we brainstorm a little more on the API for e.g. thoughts on
```
config:
  .......
  customConfigMap: <blah>
```

arkodg:

PR #2725, opened 2024-02-29
e2e: backend upgrade test

### LINKED ISSUE #1743: Handle Backend Pod Upgrades
If 2 backend pods are undergoing a rolling restart, add outlier detection and retry settings in Envoy proxy to ensure no traffic intended for the backend is dropped

guydc:
Tests backend upgrades - ensuring zero-downtime restarts of backend pods when retry and circuit breaker are configured. 

Following discussion with @arkodg in the last community meeting, we will use [fortio](https://github.com/fortio/fortio) to generate a load. Rationale for using fortio over hey, vegeta, k6 and other golang load generators:
- Apache-2.0 license 
- Relatively active and maintained project
- Owned by the fortio org, rather than a specific contributor
- Affinity to the Envoy project
- Convenient reporting interface

Depends on: #2773 , #2754 

Fixes #1743


Xunzhuo (dismissed):
LGTM

test/e2e/tests/backend-upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+		dp.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	restartTime := time.Now().Format(time.RFC3339)
+	dp.Spec.Template.ObjectMeta.Annotations[kubeRestartAnnotation] = restartTime
+
+	if err = c.Update(ctx, dp); err != nil {
+		return err
+	}
+
+	return wait.PollUntilContextTimeout(ctx, 1*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
```
zirain, line 140: can you use configuration from suite.TimeoutConfig?

test/e2e/tests/backend-upgrade.go
```diff
@@ -12,19 +12,20 @@ import (
 	"context"
 	"io"
 	"net/url"
-	"sigs.k8s.io/gateway-api/conformance/utils/config"
 	"testing"
 	"time"
 
+	"fortio.org/fortio/fhttp"
 	"fortio.org/fortio/periodic"
+
```
zirain, line 20:
```suggestion

```

test/e2e/tests/backend-upgrade.go
```diff
@@ -12,19 +12,20 @@ import (
...
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
```
zirain, line 27:
```suggestion

```

zirain (approved):

Xunzhuo (approved):

test/e2e/testdata/backend-upgrade.yaml
```diff
@@ -0,0 +1,38 @@
...
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: http-backend-upgrade
+    namespace: gateway-conformance-infra
+  circuitBreaker:
+    maxParallelRequests: 10000
+    maxConnections: 10000
+    maxPendingRequests: 10000
+  retry:
```
arkodg, line 16: late to this party, should we recommend `retryOn` connection failure over outlier detection (passive health check) ?

guydc: >  late to this party, should we recommend retryOn connection failure over outlier detection (passive health check) ?

I think that outlier detection plays a role in more prolonged backend downtime scenarios, where backends are finally expected to recover and rejoin the loadbalancing pool. In this specific test:
- rolling the deployment takes 1-2s
- outliers will never rejoin the pool, as their ClusterIPs will be eliminated in the rolling process 

When writing the tests, I saw the following envoy error codes:
- UF: was mitigated with `connect-failure` trigger
- UC: was mitigated with `reset` trigger
- UO: was mitigated by increasing `max_retries` and using a strategy that redispatches retries to different hosts
 
I was hesitant to use the `reset` trigger in this example, as some of the reset situations are mid-response, e.g. `upstream_reset_after_response_started`. So, there's a concern that non-idempotent requests can be retried without knowing that their response code was retriable, which is unsafe. That's also why `reset` is not in the default trigger list, from my PoV. 

Either way, if we want a test that demonstrates 100% success rate, I think that outlier detection alone would not be enough, as terminating pods would be ejected only after some requests already failed. The retry host selection predicate play a role similar to outlier detection in this example. 

So, when using this sort of config, users should be aware of the risk, and they can decide wether they are willing to accept failure for some requests that were in-flight during the upgrade, or retry them at the risk of corrupting data at the backend. 

I think that a more robust approach to ZDM on the backed would also require some work on the backend. Quite similar to was we did for EG's graceful shutdown:
- Define a HC endpoint on the backend
- Configure EG to probe the HC endpoint and setup EG upstream connection idle timeouts
- Upon termination, backend starts failing the HC endpoint and gracefully draining connections 
- Envoy identifies that the terminating pods in unhealthy and stops creating new connections
- idle timeouts kick-in on both sides and connections scale to 0
- backend pod is terminated safely

arkodg: thanks for writing out this detailed analysis !
would be great to convert some of this work into a user guide called - Backend Upgrades and share the `retry` logic and also highlight a better approach using active health checks

PR #2729, opened 2024-02-29
ratelimit support prometheus

zirain:
Right now, envoy-gateway controller and proxy support metric in prometheus, this patch will make ratelimit as same as the other components.


internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -91,7 +91,9 @@ const (
 	// ReadinessPath is readiness path for readiness probe.
 	ReadinessPath = "/healthcheck"
 	// ReadinessPort is readiness port for readiness probe.
-	ReadinessPort = 8080
+	ReadinessPort  = 8080
+	StatsdPort     = 9125
```
arkodg, line 95: can we use `19001` for prom, similar to rest https://gateway.envoyproxy.io/latest/install/install-helm/#open-ports ?

zirain: make sense, will update later

arkodg (approved):
LGTM thanks for adding this enhancement !

Xunzhuo (approved):

api/v1alpha1/envoygateway_types.go
```diff
@@ -326,6 +326,25 @@ type RateLimit struct {
...
+}
+
+type RateLimitMetrics struct {
+	// Prometheus defines the configuration for prometheus endpoint.
+	Prometheus *RateLimitMetricsPrometheusProvider `json:"prometheus,omitempty"`
+}
+
+type RateLimitMetricsPrometheusProvider struct {
+	// Disable the Prometheus endpoint.
+	Disable bool `json:"disable,omitempty"`
```
zhaohuabing, line 347: Does this need to be this deep?

zirain: keep same as others

arkodg (approved):

Issue #2601, opened 2024-02-12
Validate the CA Cert


arkodg:
*Description*:
>Describe the issue.

https://github.com/envoyproxy/gateway/pull/2490 adds support for downstream mTLS by introducing a way to input the CA Cert

This cert should be validated to make sure its contents are valid similar to what is done for the server tls certs in https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/tls.go

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

guydc:
Maybe also relevant:
- Server cert/key not malformed
- Server cert/key matching each other

arkodg:
@guydc server cert/key validation is already done today https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/tls.go

PR #2732, opened 2024-02-29
Validate the CA Cert

zirain:
Fixes: https://github.com/envoyproxy/gateway/issues/2601


internal/gatewayapi/testdata/tls/gen-certs.sh
```diff
@@ -0,0 +1,31 @@
...
+openssl rsa -in rsa-pkcs8-wildcard.key -out rsa-pkcs1-wildcard.key
+
+# ECDSA-p256
+
+openssl ecparam -name prime256v1 -genkey -noout -out ecdsa-p256.key
+openssl req -new -x509 -days $CERT_VALIDITY_DAYS -key ecdsa-p256.key -out ecdsa-p256-cert.pem -subj "/CN=foo.bar.com"
+
+# ECDSA-p384
+
+	openssl ecparam -name secp384r1 -genkey -noout -out ecdsa-p384.key
```
cnvergence, line 30: nit: indent

internal/gatewayapi/tls.go
```diff
@@ -96,3 +97,24 @@ func verifyHostname(cert *x509.Certificate, host *v1.Hostname) ([]string, error)
 
 	return nil, x509.HostnameError{Certificate: cert, Host: string(*host)}
 }
+
+func validateCertificate(data []byte) error {
```
arkodg, line 101: can we also insert this in https://github.com/envoyproxy/gateway/blob/c075d81d5a77fd72a7bda3b65c62c7bad49a7d1f/internal/gatewayapi/tls.go#L26, so we can validate server cert as well ?

internal/gatewayapi/testdata/tls/gen-certs.sh
zirain: done

internal/gatewayapi/tls.go
zirain: done

internal/gatewayapi/testdata/gateway-with-listener-with-valid-multiple-tls-configuration.out.yaml
```diff
@@ -28,20 +28,16 @@ gateways:
...
-      - lastTransitionTime: null
-        message: Listener has been successfully translated
-        reason: Accepted
-        status: "True"
-        type: Accepted
-      - lastTransitionTime: null
-        message: Listener references have been resolved
-        reason: ResolvedRefs
-        status: "True"
+        message: 'Secret envoy-gateway/tls-secret-ecdsa-1 must contain valid tls.crt
```
arkodg, line 31: we should update the cert material here then, because the testdata is for a valid case

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

PR #2733, opened 2024-02-29
fix(translator): Panic when translating routes with empty backends

davidalger:
This PR fixes a panic which occurs when adding either an `HTTPRoute` or `GRPCRoute` with no filters or backendRefs. For example:

```yaml
kind: HTTPRoute
metadata:
  name: empty-backends
spec:
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
```

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x190dadc]

goroutine 59 [running]:
github.com/envoyproxy/gateway/internal/xds/translator.buildXdsRouteAction(...)
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/route.go:214
github.com/envoyproxy/gateway/internal/xds/translator.buildXdsRoute(0x40007c2680)
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/route.go:66 +0x46c
github.com/envoyproxy/gateway/internal/xds/translator.(*Translator).processHTTPListenerXdsTranslation(0x4000b15348, 0x4000708ee0, {0x4000691d40, 0x1, 0x1ffa23a?}, 0x12?, 0x1?, 0x0)
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/translator.go:244 +0xab4
github.com/envoyproxy/gateway/internal/xds/translator.(*Translator).Translate(0x4000b15348, 0x40003f9700)
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/translator.go:79 +0x68
github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate.func1({{0x4000b74940?, 0x0?}, 0x0?, 0x40003f9700?}, 0x0?)
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/runner/runner.go:80 +0x1d0
github.com/envoyproxy/gateway/internal/message.HandleSubscription[...]({{0x1ff221a, 0xe?}, {0x1fe523f?, 0x6?}}, 0x4000141740?, 0x4000b15f88)
    /Users/davidalger/work/envoyproxy/gateway/internal/message/watchutil.go:76 +0xd74
github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate(0x40002c6100, {0x235bdf0?, 0x400079bd60?})
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/runner/runner.go:52 +0x88
created by github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).Start in goroutine 1
    /Users/davidalger/work/envoyproxy/gateway/internal/xds/translator/runner/runner.go:45 +0x1f0
Stream closed EOF for envoy-gateway-system/envoy-gateway-c589d95bf-mw66t (envoy-gateway)
```

### DIFF
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -209,7 +209,7 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 
 		// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
 		for _, ruleRoute := range ruleRoutes {
-			if ruleRoute.BackendWeights.Invalid > 0 && ruleRoute.Destination == nil {
+			if ruleRoute.Destination == nil && ruleRoute.Redirect == nil {
 				ruleRoute.DirectResponse = &ir.DirectResponse{
 					StatusCode: 500,
 				}
@@ -493,7 +493,7 @@ func (t *Translator) processGRPCRouteRules(grpcRoute *GRPCRouteContext, parentRe
 
 		// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
 		for _, ruleRoute := range ruleRoutes {
-			if ruleRoute.BackendWeights.Invalid > 0 && ruleRoute.Destination == nil {
+			if ruleRoute.Destination == nil && ruleRoute.Redirect == nil {
 				ruleRoute.DirectResponse = &ir.DirectResponse{
 					StatusCode: 500,
 				}
--- a/internal/gatewayapi/testdata/grpcroute-with-empty-backends.in.yaml
+++ b/internal/gatewayapi/testdata/grpcroute-with-empty-backends.in.yaml
@@ -1,0 +1,31 @@
+gateways:
+- apiVersion: gateway.networking.k8s.io/v1
+  kind: Gateway
+  metadata:
+    namespace: envoy-gateway
+    name: gateway-1
+  spec:
+    gatewayClassName: envoy-gateway-class
+    listeners:
+    - name: http
+      protocol: HTTP
+      port: 80
+      allowedRoutes:
+        namespaces:
+          from: All
+grpcRoutes:
+- apiVersion: gateway.networking.k8s.io/v1alpha2
+  kind: GRPCRoute
+  metadata:
+    namespace: default
+    name: grpcroute-1
+  spec:
+    parentRefs:
+    - namespace: envoy-gateway
+      name: gateway-1
+      sectionName: http
+    rules:
+    - matches:
+      - method:
+          service: com.ExampleExact
+          type: Exact
--- a/internal/gatewayapi/testdata/grpcroute-with-empty-backends.out.yaml
+++ b/internal/gatewayapi/testdata/grpcroute-with-empty-backends.out.yaml
@@ -1,0 +1,120 @@
+gateways:
+- apiVersion: gateway.networking.k8s.io/v1
+  kind: Gateway
+  metadata:
+    creationTimestamp: null
+    name: gateway-1
+    namespace: envoy-gateway
+  spec:
+    gatewayClassName: envoy-gateway-class
+    listeners:
+    - allowedRoutes:
+        namespaces:
+          from: All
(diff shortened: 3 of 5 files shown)


internal/gatewayapi/route.go
```diff
@@ -209,7 +209,7 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 
 		// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
 		for _, ruleRoute := range ruleRoutes {
-			if ruleRoute.BackendWeights.Invalid > 0 && ruleRoute.Destination == nil {
+			if (ruleRoute.BackendWeights.Invalid > 0 || len(rule.BackendRefs) == 0) && ruleRoute.Destination == nil {
```
arkodg, line 212: why isnt it hitting `ruleRoute.Destination == nil` ? that feels like the bug

arkodg: should it be a `or` / `||` instead ?

davidalger: `ruleRoute.BackendWeights.Invalid` is only incremented when `BackendRefs` is non-empty as a result of being set during destination processing, so it would seem to be an intentional gate. In the case of an empty `BackendRefs` though, `ruleRoute.BackendWeights.Invalid` will always be zero which is the cause of failure when publishing a route with no backend refs nor filters.

If I remove `ruleRoute.BackendWeights.Invalid > 0 &&` or change it to `||` instead, a handful of existing tests have `directResponse` show up in `xdsIR` output. For example, this one: https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/testdata/httproute-with-redirect-filter-hostname.in.yaml although maybe that is a poorly written test since applying that object would normally result in a `RequestRedirect filter must not be used together with backendRefs` validation error 🤔

arkodg: ah, cool, lets stick with the logic in this PR then

arkodg (dismissed):

internal/gatewayapi/route.go
davidalger: There is actually a problem with my proposed logic, since `backendRefs` cannot be used with the `RequestRedirect` filter it would break this route:

```yaml
kind: HTTPRoute
metadata:
  name: redirected
spec:
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /redirected
    filters:
    - type: RequestRedirect
      requestRedirect:
        scheme: https
        statusCode: 301
        hostname: "redirected.com"
```

davidalger: yeah, all the HTTPRouteRedirect are failing in the conformance tests kinda proving this fix is incomplete: https://github.com/envoyproxy/gateway/actions/runs/8103755213/job/22149232447?pr=2733

I think changing the logic to `if ruleRoute.Destination == nil && ruleRoute.Redirect == nil {` will cover all the cases (and it passed conformance locally) so I'm going to push that up in a minute

internal/gatewayapi/testdata/httproute-with-redirect-filter-full-path-replace-https.in.yaml
```diff
@@ -31,9 +31,6 @@ httpRoutes:
     - matches:
       - path:
           value: "/"
-      backendRefs:
```
arkodg, line 34: were these interfering with the tests ?

davidalger: nope, sorry. i've repushed this without any extra changes to keep it to a pure single-purpose fix

arkodg (approved):

zirain (approved):
LGTM

suprise this is not blocked bu validatingwebhook/CEL.

PR #2735, opened 2024-03-01
chore: filename be more gopher

zirain:


zhaohuabing (approved):
It's nice to know how file names should look like as a true gopher :-)

Xunzhuo (approved):

Issue #2679, opened 2024-02-23
e2e: Retry in BackendTrafficPolicy


guydc:
*Description*:
>Describe the desired behavior, what scenario it enables and how it

Relates to #2678 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #2738, opened 2024-03-01
e2e: add test for BackendTrafficPolicy retry

zirain:
Fixes: https://github.com/envoyproxy/gateway/issues/2679


guydc:
Very nice addition! I think that a lot of features that are difficult to verify with assertions on an HTTP requests/responses will benefit from metrics-based validation.

arkodg (approved):
LGTM thanks ! neat idea we can reuse going forward

Xunzhuo (approved):

PR #2739, opened 2024-03-01
fix: remove no needed new pointer creation

Xunzhuo:
**What type of PR is this?**

remove no needed new pointer creation


zirain:
I recall there's an improment in upstream go, but I cannot recall which version and commit, can you link it?

cnvergence (approved):
lgtm

Xunzhuo:
@zirain I did not get, and it should be none-related to this PR.

Can we get this in ?

zhaohuabing:
I guess this one: https://go.dev/blog/loopvar-preview

For Go 1.22, we plan to change for loops to make these variables have per-iteration scope instead of per-loop scope. 

It's fixed in Go 1.22, but why lint removes this check in 1.21 ?

zirain:
> I guess it's related to the variable scope in loop: https://go.dev/blog/loopvar-preview
> 
> > For Go 1.22, we plan to change for loops to make these variables have per-iteration scope instead of per-loop scope.
> 
> It's fixed in Go 1.22, but why lint removes this check in 1.21 ?

yeah, that's what I mean.

should we hold this until we move to go1.22?

arkodg (approved):

zirain (approved):

shawnh2:
This loopvar feature of go 1.22 requires enable env `GOEXPERIMENT=loopvar` (default closed) explicitly. Maybe we could consider enable this feature for other places as well.

PR #2746, opened 2024-03-01
chore: Cleanup test data for routes with RequestRedirect filter

davidalger:
This PR cleans up test data for `HTTPRoutes` using the `RequestRedirect` filter. When using the `RequestRedirect` filter, `backendRefs` should not be specified and applying a manifest with both present to a real cluster would result in validation errors:

```
The HTTPRoute "httproute-1" is invalid: spec.rules[0]: Invalid value: "object": RequestRedirect filter must not be used together with backendRefs
```


arkodg (approved):
LGTM thanks !

arkodg:
@davidalger is that error coming from CEL validation for upstream gateway api ?

arkodg:
looks like it is https://github.com/kubernetes-sigs/gateway-api/blob/f6102784e48833220d538e5a78309b71476529c4/apis/v1/httproute_types.go#L128

Xunzhuo (approved):

PR #2756, opened 2024-03-03
fix: reset translate in retry

### LINKED ISSUE #2755: bug: translation mistake in retry triggers
*Description*:
The EG interface uses the term `disconnect-reset`, while the underlying xds enum value is `reset`. 
Envoy fails silently and accepts misconfiguration. 

I think that the EG term is better, as `reset` is an overloaded term, and the `disconnect-` prefix helps users understand the difference between the `connection-failure` and the `disconnect-reset` triggers. 

The cluster translator should be fixed and translate `disconnect-reset` to `reset`.

guydc:
Fixes #2755

### DIFF
--- a/api/v1alpha1/retry_types.go
+++ b/api/v1alpha1/retry_types.go
@@ -44,7 +44,7 @@ type RetryOn struct {
 }
 
 // TriggerEnum specifies the conditions that trigger retries.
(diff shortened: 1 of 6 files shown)


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2757, opened 2024-03-03
feat(translator): add default retry budget and retry host predicate

### LINKED ISSUE #2754: set a default retry budget and host retry predicate
*Description*:
Currently, EG does not expose retry limit settings: #2322, and it doesn't configure a default value, leading to the envoy default of `max_retries: 3` to apply. 

While working on #2725 I noticed that the default envoy `max_retry` value is leading to retry overflow when there is a [significant amount of requests](https://github.com/envoyproxy/gateway/actions/runs/8118438104/job/22194033455?pr=2725#step:6:1611) to a restarting backend. 

Envoy provides a [sample](https://www.envoyproxy.io/docs/envoy/latest/faq/load_balancing/transient_failures#outlier-detection) that increases this value to 10 to deal with restart scenarios. Additionally, a host retry predicate is used, to allow dispatching retries to a different host, increasing the chance of a successful retry. 

Istio sets concurrent request limit and retry limit to `max_uint32` by default: https://github.com/istio/istio/blob/7f25d965f21154a32ab5e29aef9ae501fd8ef7a8/pilot/pkg/networking/core/v1alpha3/cluster_traffic_policy.go#L348. The high retry limit is required to support rolling update and restart scenarios. Istio also configures a host retry predicate. 

I propose that EG will currently set the `retry_budget` to 100% by default, meaning the total upstream request volume would be capped at 200% of `max_parallel_requests`. Additionally, EG will configure a host retry predicate by default, aligned with the envoy example (rejecting previously-used hosts from being used in a retry). 

[optional *Relevant Links*:]

- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto#config-cluster-v3-circuitbreakers-thresholds-retrybudget
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#envoy-v3-api-msg-config-route-v3-retrypolicy-retryhostpredicate
- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#retry-plugin-configuration

guydc:
This PR sets default values for:
-  retry budgets on any cluster: set to 100% of the configured max parallel request 
-  retry host predicate on any route that enables retries: omits previously-host hosts for retries

These values are inspired by envoy and istio recommendations, and intended to improve the overall success of retries when the backend is going through a rolling restart. 

Fixes #2754

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -674,6 +674,11 @@ xds:
       dynamicActiveClusters:
       - cluster:
           '@type': type.googleapis.com/envoy.config.cluster.v3.Cluster
+          circuitBreakers:
+            thresholds:
+            - retryBudget:
+                budgetPercent:
+                  value: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
@@ -690,6 +695,11 @@ xds:
           type: EDS
       - cluster:
           '@type': type.googleapis.com/envoy.config.cluster.v3.Cluster
+          circuitBreakers:
+            thresholds:
+            - retryBudget:
+                budgetPercent:
+                  value: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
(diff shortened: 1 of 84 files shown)


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2730, opened 2024-02-29
Adding OIDC e2e tests and Relaxing HTTPS restriction for OIDC token endpoint

zhaohuabing:
This PR adds e2e tests for OIDC. 
 * A keycloak has been installed as the OIDC idP for testing.
 * HTTPS restriction has been relaxed for OIDC token endpoint, so that the in-cluster keycloak can be used for testing.

Fix https://github.com/envoyproxy/gateway/issues/2726

PR #2768, opened 2024-03-04
Fix hmac secret

zhaohuabing:
This PR does the following:
* Generates an HMAC secret and stores it in a K8s secret in the CertGen job.
* Watches this k8s secret in the Kuberetes provider.
* Uses the HMAC generated in the above step for envoy ouath2 filter configuration.


internal/crypto/certgen.go
```diff
@@ -153,6 +154,11 @@ func GenerateCerts(cfg *config.Server) (*Certificates, error) {
 			return nil, err
 		}
 
+		oidcHMACSecret, err := generateHMACSecret()
```
arkodg, line 157: what happens during upgrade if this secret already exists ? do we skip or overwrite ?

arkodg: relates to https://github.com/envoyproxy/gateway/blob/ce1eb544d5e1b473ccd2c852be595d80fcf75ba8/internal/provider/kubernetes/secrets.go#L84

arkodg:
hey @zhaohuabing has this been tested ? are you sure we are reconciling this secret ? I think we may need some more code in https://github.com/envoyproxy/gateway/blob/ce1eb544d5e1b473ccd2c852be595d80fcf75ba8/internal/provider/kubernetes/controller.go#L138 to make sure this secret is part of the resources

internal/crypto/certgen.go
zhaohuabing: The creation of HMAC secret will be skipped in `CreateOrUpdateSecrets` if it's already there, like the other certs. No changes are needed.

zhaohuabing:
> hey @zhaohuabing has this been tested ? are you sure we are reconciling this secret ?
I haven't tested it manually yet. It should be covered by the OIDC e2e test: https://github.com/envoyproxy/gateway/pull/2730/

> I think we may need some more code in
> 
> https://github.com/envoyproxy/gateway/blob/ce1eb544d5e1b473ccd2c852be595d80fcf75ba8/internal/provider/kubernetes/controller.go#L138
> 
> to make sure this secret is part of the resources

Yes, I missed that.  It's fixed.

Xunzhuo (approved):

arkodg (approved):
🚀

PR #2769, opened 2024-03-04
feat(translator): set min retry limit of 100

### LINKED ISSUE #2754: set a default retry budget and host retry predicate

guydc:
In addition to #2757, set a default minimum value of `100` for the limit on concurrent retries. Since the concurrent budget is calculated based on the current active requests, the dynamic retry limit may be too low when the volume of traffic is not significant and there are many upstream failures.

In the future, we should make these options configurable, so that users can tune them to their needs. 

Fixes #2754

### DIFF
--- a/internal/xds/translator/cluster.go
+++ b/internal/xds/translator/cluster.go
@@ -276,8 +276,12 @@ func buildXdsClusterCircuitBreaker(circuitBreaker *ir.CircuitBreaker) *clusterv3
 	// related to pod restarts
 	cbt := &clusterv3.CircuitBreakers_Thresholds{
 		Priority: corev3.RoutingPriority_DEFAULT,
+		// concurrent retries limit = max(100, current-active-requests)
 		RetryBudget: &clusterv3.CircuitBreakers_Thresholds_RetryBudget{
 			BudgetPercent: &xdstype.Percent{
+				Value: 100,
+			},
+			MinRetryConcurrency: &wrapperspb.UInt32Value{
 				Value: 100,
 			},
 		},
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -679,6 +679,7 @@ xds:
             - retryBudget:
                 budgetPercent:
                   value: 100
+                minRetryConcurrency: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
@@ -700,6 +701,7 @@ xds:
             - retryBudget:
                 budgetPercent:
                   value: 100
+                minRetryConcurrency: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
@@ -726,6 +728,7 @@ xds:
             - retryBudget:
                 budgetPercent:
                   value: 100
+                minRetryConcurrency: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
@@ -747,6 +750,7 @@ xds:
             - retryBudget:
                 budgetPercent:
                   value: 100
+                minRetryConcurrency: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
@@ -768,6 +772,7 @@ xds:
             - retryBudget:
                 budgetPercent:
                   value: 100
+                minRetryConcurrency: 100
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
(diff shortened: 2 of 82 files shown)


internal/xds/translator/cluster.go
```diff
@@ -276,10 +276,14 @@ func buildXdsClusterCircuitBreaker(circuitBreaker *ir.CircuitBreaker) *clusterv3
 	// related to pod restarts
 	cbt := &clusterv3.CircuitBreakers_Thresholds{
 		Priority: corev3.RoutingPriority_DEFAULT,
+		// concurrent retries limit = max(100, current-active-requests)
 		RetryBudget: &clusterv3.CircuitBreakers_Thresholds_RetryBudget{
 			BudgetPercent: &xdstype.Percent{
```
arkodg, line 281: if `min_rety_concurrency` is adequate, should be rm `retryBudget:100%` setting ?

arkodg: also can we disable `min_rety_concurrency` by setting it to `0` ?

arkodg: what im trying to get at is can we do without this `RetryBudget` knob, if the 20% default is too low, can we increase `maxParallelRequests` (guesing the retry budget percentage is off this value) in the circuit breaker settings ?

guydc: I understand them as being complementary, with min retries acting as a guardrail. The budget-based limit will dynamically scale with traffic. However, in smaller numbers, the budget could be a problem. Imagine that you had some traffic that generated transient failures, and then traffic drops to 0: this means that the budget-based limit is now also 0 and all retries are aborted.

arkodg: @guydc isnt that what the knob is designed for ? to not invest more cycles when failure is evident ?
since youve already added the host predicate setting for retry it should hopefully retry and achieve success

guydc: > what im trying to get at is can we do without this `RetryBudget` knob, if the 20% default is too low, can we increase `maxParallelRequests` (guesing the retry budget percentage is off this value) in the circuit breaker settings ?

The circuit breaker value only define the upper limit of the budget. But, the actual limit is calculated dynamically from the volume of traffic seen. So, increasing one of the circuit breakers wouldn't necessarily produce the expected outcome (actual retry limit being increased).

arkodg: based on issues like https://github.com/envoyproxy/envoy/issues/30205, id like to avoid this setting for now

guydc: @arkodg - Sure, that makes sense. During the API discussion for retries, we preferred the more recent budget-based approach. 

I have to admit that I'm also a bit surprised by the behavior of this feature, and this follow-up PR is trying to further tweak our defaults so that our upgrades are hitless. Maybe the best approach would be to wait for this feature to receive some additional improvements, like the ones outlined in the mentioned issue. 

In that case, I can remove the budget altogether and replace it with some default min_retry value that's derived from one of the circuit breakers (e.g. 20% of parallel requests) . WDYT? Is that ok?

guydc:
closing in favor of #2773

PR #2773, opened 2024-03-04
API: support max_retries in circuit breakers API

### LINKED ISSUE #2322: Find the right home for maximum number of parallel retries
*Description*:
EG needs to find the right home for "maximum number of parallel retries"

* Option 1:  par of the [Circuit Breaker setting in BackendTrafficPolicy](https://github.com/envoyproxy/gateway/pull/2284)
* Option 2:  par of the [Retry setting in BackendTrafficPolicy](https://github.com/envoyproxy/gateway/pull/2168)

I prefer option 1 because: 
Even though both have "retries" in their name, "maximum number of parallel retries" and "request retires" serve two different purposes. The concurrent max retries setting is inherently associated with the Circuit Breaker, which fails requests quickly when  lot of retries happen and apply back pressure on downstream. On the other hand, request retries are specifically designed to mitigate transient network issues. Would love more insights from @kflynn and other @envoyproxy/gateway-maintainers

[optional *Relevant Links*:]
* https://github.com/envoyproxy/gateway/pull/2284
* https://github.com/envoyproxy/gateway/pull/2168

guydc:
Allows configuration of `max_retries` in BTP's circuit breakers. After some internal discussion, it seems that circuit breakers are a better location for this feature:
- Retry settings are per route and refer to per-request behavior
- Max retries setting are per-backend-per-route, similar to other circuit breakers
- Max retries settings are limiting a certain type of upstream requests, similar to other circuit breakers

We should enhance documentation and let users know that these retry-related settings exist in the circuit breakers section. 

Removes retry_budget for now, following the discussion here: https://github.com/envoyproxy/gateway/pull/2769#discussion_r1511691984

Fixes #2322


internal/xds/translator/cluster.go
```diff
@@ -301,6 +298,12 @@ func buildXdsClusterCircuitBreaker(circuitBreaker *ir.CircuitBreaker) *clusterv3
 				Value: *circuitBreaker.MaxParallelRequests,
 			}
 		}
+
+		if circuitBreaker.MaxParallelRetries != nil {
+			cbt.MaxRetries = &wrapperspb.UInt32Value{
+				Value: *circuitBreaker.MaxParallelRetries,
+			}
```
arkodg, line 305: lets be explicit with `1024` in case `circuitBreaker.MaxParallelRetries == nil`, in case we decide to run in non k8s env that doesnt set that default

arkodg: maybe we even need to set it if `circuitBreaker == nil`

guydc: Sure, will do. For the other settings, I think that there's no reason to do it, as they anyway default to 1024 on envoy level.

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM. Thanks!

PR #2782, opened 2024-03-05
fix: Delete unused status keys from watchable

### LINKED ISSUE #2660: Delete unused status keys from watchable
*Description*:

The status keys are only stored and not deleted. Either all the status keys or at least the stale ones should be deleted to free up memory
https://github.com/envoyproxy/gateway/blob/cf46fbe776918ad19444e26d637ffcc79676ca23/internal/gatewayapi/runner/runner.go#L112
& 
https://github.com/envoyproxy/gateway/blob/cf46fbe776918ad19444e26d637ffcc79676ca23/internal/xds/translator/runner/runner.go#L99

uniglot:
**What type of PR is this?**

fix: Delete unused status keys from watchable

The status keys are only stored to watchable maps and not deleted, consequently the amount of freeable memory keeps decreasing over time.

The approach to address this problem is very similar to the way to handle XdsIR and InfraIR. 

But I thought if I use `getIRKeysToDelete` directly to track the status diff, too many arrays need to be created to call the function for each status type. So I didn't use `getIRKeysToDelete` or create a similar function returning the list of keys to delete. Instead, I introduced `DeletableStatus` struct to temporarily store which keys are still relevant.

Fixes #2660 

It works, but the code is very repetitive, so it will be tedious and time-consuming to modify the code for now when other types of status need to be introduced. It'd be better if the watchable keys are handled by each status type's methods (but it requires larger modification) or at least the handlers are in a separate file.

If you want the code to be cleaner I'll try to refactor, or we can make a separate issue for it. I just wanted to get things work first.

\+ Checked linting & tests.


internal/gatewayapi/runner/runner.go
```diff
@@ -182,6 +203,153 @@ func (r *Runner) deleteAllIRKeys() {
...
+	HTTPRouteStatusKeys map[types.NamespacedName]bool
+	GRPCRouteStatusKeys map[types.NamespacedName]bool
+	TLSRouteStatusKeys  map[types.NamespacedName]bool
+	TCPRouteStatusKeys  map[types.NamespacedName]bool
+	UDPRouteStatusKeys  map[types.NamespacedName]bool
+
+	ClientTrafficPolicyStatusKeys  map[types.NamespacedName]bool
+	BackendTrafficPolicyStatusKeys map[types.NamespacedName]bool
+	SecurityPolicyStatusKeys       map[types.NamespacedName]bool
+	BackendTLSPolicyStatusKeys     map[types.NamespacedName]bool
```
arkodg, line 217: nit: can we move `BackendTLSPolicyStatusKeys` near L212, since its an upstream resource like `UDPRouteStatus`

internal/gatewayapi/runner/runner.go
```diff
@@ -182,6 +203,153 @@ func (r *Runner) deleteAllIRKeys() {
 	}
 }
 
+type DeletableStatus struct {
```
arkodg, line 206: prefer `StatusesToDelete`

internal/gatewayapi/runner/runner.go
```diff
@@ -182,6 +203,153 @@ func (r *Runner) deleteAllIRKeys() {
...
+	TCPRouteStatusKeys  map[types.NamespacedName]bool
+	UDPRouteStatusKeys  map[types.NamespacedName]bool
+
+	ClientTrafficPolicyStatusKeys  map[types.NamespacedName]bool
+	BackendTrafficPolicyStatusKeys map[types.NamespacedName]bool
+	SecurityPolicyStatusKeys       map[types.NamespacedName]bool
+	BackendTLSPolicyStatusKeys     map[types.NamespacedName]bool
+}
+
+func (r *Runner) getDeletableStatus() *DeletableStatus {
```
arkodg, line 220: prefer if this API was called `getAllStatuses`

it took me 2 readings to figure out your approach of collecting all statuses and then deleting the current ones from this truck and then deleting the remaining ones from watchable :)

internal/gatewayapi/runner/runner.go
```diff
@@ -56,15 +57,21 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
+			// IR keys for watchable
+			var curIRKeys, newIRKeys []string
+
 			// Get current IR keys
 			for key := range r.InfraIR.LoadAll() {
-				curKeys = append(curKeys, key)
+				curIRKeys = append(curIRKeys, key)
 			}
 
+			// Get the current DeletableStatus which manages status keys to be deleted
```
arkodg, line 72: lets add a comment here explaining the logic
1. Lets get all status keys from watchable and save it in this structure
2. Lets delete any current keys that are valid from this structure
3. The remaining keys, are the keys that are no longer valid, and will be deleted before we exit this function

internal/gatewayapi/runner/runner.go
```diff
@@ -182,6 +203,153 @@ func (r *Runner) deleteAllIRKeys() {
...
+	for key := range r.ProviderResources.ClientTrafficPolicyStatuses.LoadAll() {
+		r.ProviderResources.ClientTrafficPolicyStatuses.Delete(key)
+	}
+	for key := range r.ProviderResources.BackendTrafficPolicyStatuses.LoadAll() {
+		r.ProviderResources.BackendTrafficPolicyStatuses.Delete(key)
+	}
+	for key := range r.ProviderResources.SecurityPolicyStatuses.LoadAll() {
+		r.ProviderResources.SecurityPolicyStatuses.Delete(key)
+	}
+	for key := range r.ProviderResources.BackendTLSPolicyStatuses.LoadAll() {
```
arkodg, line 348: lets move this to L334

internal/xds/translator/runner/runner.go
```diff
@@ -90,19 +90,31 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 					return
 				}
 
+				// Get current status keys and mark them as deletable temporarily
```
arkodg, line 93: lets add comments similar to the one above

arkodg:
hey @uniglot, added minor comments, this is looking good !

uniglot:
@arkodg Applied all of your suggestions! I feel bad for making you reading 2 times 🥲

arkodg (approved):
LGTM thanks !

internal/gatewayapi/runner/runner.go
```diff
@@ -182,6 +205,153 @@ func (r *Runner) deleteAllIRKeys() {
...
+	GRPCRouteStatusKeys        map[types.NamespacedName]bool
+	TLSRouteStatusKeys         map[types.NamespacedName]bool
+	TCPRouteStatusKeys         map[types.NamespacedName]bool
+	UDPRouteStatusKeys         map[types.NamespacedName]bool
+	BackendTLSPolicyStatusKeys map[types.NamespacedName]bool
+
+	ClientTrafficPolicyStatusKeys  map[types.NamespacedName]bool
+	BackendTrafficPolicyStatusKeys map[types.NamespacedName]bool
+	SecurityPolicyStatusKeys       map[types.NamespacedName]bool
+}
```
zirain, line 220: maybe we should use set[T] instead of map[T]bool, but we can do that in the future.

zirain (approved):

PR #2786, opened 2024-03-05
fix: prevent policies targeting non-TLS listeners on the same port from conflicting

### LINKED ISSUE #2742: When there are multiple Gateways having listeners with same port and `mergeGateway` enabled, using CORS for each `Gateway` separately, only the first Gateway works.
*Description*:
When there are multiple Gateways having listeners with same port and `mergeGateway` enabled, using CORS for each `Gateway` separately, only to the first Gateway will work.

*Repro steps*:
Install the Gateway API CRDs and Envoy Gateway
```bash
helm install eg oci://docker.io/envoyproxy/gateway-helm --version v0.0.0-latest -n envoy-gateway-system --create-namespace
```
Use aill-in-one yaml described below:
```bash
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config

---
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  mergeGateways: true 
---
apiVersion: v1
kind: Namespace
metadata:
  name: sealos-system
---
kind: Gateway
metadata:
  name: eg
  namespace: sealos-system
spec:
  gatewayClassName: eg
  listeners:
    - hostname: "*.192.168.0.15.nip.io"
      name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns-oehe
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  replicas: 1
  selector:
    matchLabels:
      app: bdkzlmibsivuiqav
  template:
    metadata:
      labels:
        app: bdkzlmibsivuiqav
    spec:
      containers:
      - image: nginx:1.25
        name: nginx
        ports:
        - containerPort: 80
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 100m
            memory: 64Mi
          requests:
            cpu: 10m
            memory: 16Mi
---
apiVersion: v1
kind: Service
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  clusterIP: "10.96.0.2"
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: bdkzlmibsivuiqav
  type: ClusterIP
---
kind: HTTPRoute
metadata:
  name: bdkzlmibsivuiqav
  namespace: ns-oehe
spec:
  hostnames:
  - ntjxuedx.192.168.0.15.nip.io
  parentRefs:
    kind: Gateway
    name: eg
    namespace: sealos-system
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: bdkzlmibsivuiqav
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns-vrpg
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mfqjpuycbgjrtdww
  template:
    metadata:
      labels:
        app: mfqjpuycbgjrtdww
    spec:
      containers:
      - image: nginx:1.25
        name: nginx
        ports:
        - containerPort: 80
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 100m
            memory: 64Mi
          requests:
            cpu: 10m
            memory: 16Mi
---
apiVersion: v1
kind: Service
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  clusterIP: "10.96.0.3"
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: mfqjpuycbgjrtdww
  type: ClusterIP
---
kind: Gateway
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  gatewayClassName: eg
  listeners:
  - hostname: qccbahgo.qccbahgo
    name: http
    port: 80
    protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: mfqjpuycbgjrtdww
  namespace: ns-vrpg
spec:
  hostnames:
  - qccbahgo.qccbahgo
  parentRefs:
    kind: Gateway
    name: mfqjpuycbgjrtdww
    namespace: ns-vrpg
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: mfqjpuycbgjrtdww
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
---
kind: SecurityPolicy
metadata:
  name: cors-example
  namespace: ns-vrpg
spec:
  targetRef:
    kind: Gateway
    name: mfqjpuycbgjrtdww
  cors:
    allowOrigins:
    - "http://*.foo.com"
    allowMethods:
    - PUT
    - GET
    - POST
    - DELETE
    - PATCH
    - OPTIONS
    maxAge: 600s
    allowCredentials: true
```

Get the name of the Envoy service created the by the Gateway, because mergeGateway is enabled, the way to obtain the Envoy service differs from the quickstart. 
```bash
export ENVOY_SERVICE=$(kubectl get svc -n envoy-gateway-system --selector=gateway.envoyproxy.io/owning-gatewayclass=eg -o jsonpath='{.items[0].metadata.name}')
```
Port forward to the Envoy service
```bash
kubectl -n envoy-gateway-system port-forward service/${ENVOY_SERVICE} 8888:80 &
```
Verify that the CORS headers are present in the response of the OPTIONS request from http://www.foo.com, and the host is qccbahgo.qccbahgo

```bash
curl -H "Origin: http://www.foo.com" \
  -H "Host: qccbahgo.qccbahgo" \
  -X OPTIONS -v -s \
  http://localhost:8888/
```
The following is the response, indicating that the request was not allowed, which is not reasonable.

```bash
...
< HTTP/1.1 405 Method Not Allowed
< server: nginx/1.25.4
...
```
In this case, I applied the SecurityPolicy to the second Gateway, but the SecurityPolicy did not work, although when I checked the status of the SecurityPolicy, it showed as Accepted.

Furthermore, when I applied the SecurityPolicy only to the first Gateway, it works. The steps are as follows: 

1. Delete the existing SecurityPolicy
```bash
kubectl delete SecurityPolicy cors-example -n ns-vrpg
```
2. Apply the new SecurityPolicy to the first Gateway
```bash
kind: SecurityPolicy
metadata:
  name: cors-example
  namespace: sealos-system
spec:
  targetRef:
    kind: Gateway
    name: eg
  cors:
    allowOrigins:
    - "http://*.foo.com"
    allowMethods:
    - PUT
    - GET
    - POST
    - DELETE
    - PATCH
    - OPTIONS
    maxAge: 600s
    allowCredentials: true
```
3. Verify that the CORS headers are present in the response of the OPTIONS request from http://www.foo.com, and the host is ntjxuedx.192.168.0.15.nip.io

```bash
curl -H "Origin: http://www.foo.com" \
  -H "Host: ntjxuedx.192.168.0.15.nip.io" \
  -X OPTIONS -v -s \
  http://localhost:8888/
```
The following is the response,  indicating that the SecurityPolicy works.

```bash
< HTTP/1.1 200 OK
< access-control-allow-methods: PUT, GET, POST, DELETE, PATCH, OPTIONS
< access-control-max-age: 600
```
---

So why does the SecurityPolicy behave differently when applied to different gateways? I tried to find out the reason. I used egctl to analyze the differences in applying the SecurityPolicy to the first Gateway and the second Gateway separately. 

- When applying the SecurityPolicy to the first one, compared to the xds without the SecurityPolicy, as shown in the diagram, it adds an HTTP filter to the listener and adds the corresponding filter to the RouteConfiguration.

  - to Listener
  <div align=right>
  image
  </div>

  - to RouteConfiguration
   <div align=right>
  image
  </div>
- When applying the SecurityPolicy to the second one, compared to the xds without the SecurityPolicy, as shown in the diagram, it only adds the corresponding filter to the RouteConfiguration.

  - to RouteConfiguration

  image

So, I think this is the reason why the SecurityPolicy doesn't work when applied to the second Gateway. It lacks the HTTPFilter that should be added to the Listener. 

---

So, why does it lack?  Is it because the listener to which the HTTP filter should be added cannot be found or something? I think this is the key to solving the problem.

liorokman:

* Validate that multiple policies that affect listener configuration don't map to the same listener filter chain.
* Change the XDS listener generation so that instead of defaultFilterChain for non-TLS routes, a filterChain with a destinationPort matcher is used. This allows multiple policies attached to non-TLS listeners that differ on the destination port to provide different policies without conflicting.

Fixes #2742


arkodg:
thanks for working on this @liorokman, since this is a pretty big change in terms of how xds is generated, this is unlikely to make it into v1.0.0, will take a closer look at the PR soon

internal/xds/translator/listener.go
```diff
@@ -293,18 +293,32 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
...
-		if xdsListener.DefaultFilterChain != nil {
-			return errors.New("default filter chain already exists")
+	} else if !havePortMatch(xdsListener, irListener.Port) {
+		// No TLS, so create a matcher based on the port if one doesn't yet exist.
+		// Conflicts between listeners should have been detected in the GWAPI->IR
+		// translation level already.
+		filterChain.FilterChainMatch = &listenerv3.FilterChainMatch{
+			DestinationPort: wrapperspb.UInt32(irListener.Port),
 		}
-		xdsListener.DefaultFilterChain = filterChain
```
guydc, line 302: raised in yesterday's community meeting: when matching on the destination port, wouldn't there be a dedicated envoy listener that's already bound to that distinct port? meaning, for port-based matching, the default filter chain can still be used and there's no need to additionally do a filter chain match?

arkodg:
hey @liorokman, @guydc, @zhaohuabing and I discussed this issue in the community meeting yesterday, and prefer if we just partially rejected/skipped the policy if we detected this condition w/o changing any xds translation. 
It may be easier to implement first policy wins, and partially reject the second policy for this listener (similar to the `overlap` condition)
I had also raised this issue in the envoy slack channel, but didn't find a solution 
https://envoyproxy.slack.com/archives/C78M4KW76/p1709687283123669

liorokman:
It's not possible to reject or partially accept the second policy in this case because the problem is in the XDS translation and not the IR translation. The IR translation doesn't have an issue - when examining the IR for the listener that is _actually_ targeted by a policy there is no conflict. The conflict occurs because the XDS translation mashes multiple IRs into the same `defaultFilterContext`. 

In other words - it's not that two policies are in conflict with each other. It's that one single policy is asked to target a specific Listener and ends up affecting more than that.

Instead, the PR right now rejects policies that would affect listeners that they weren't the target of.

I made the XDS translation change because that way I could accept _more_ policies. Instead of declining to accept all policies targeting non-TLS listeners and routes attached to non-TLS listeners, the PR right now makes it possible to accept policies that don't target the same port.

liorokman:
Based on @guydc 's insight - no need to change the `defaultFilterChain` translation since a new listener is created anyways per port. I reverted that part of the patch.

What remains is that if a policy (either SecurityPolicy or ClientTrafficPolicy) is applied that would affect more than the listener that it was attached to, then the policy will be rejected.

arkodg:
@liorokman there is existing logic there checks for conflicts using a policyMap https://github.com/envoyproxy/gateway/blob/e4aea1b6e51f1a3810345e34cba4e39a94bbb5bb/internal/gatewayapi/clienttrafficpolicy.go#L72

do you think we should reuse that style, and add a `httpOnlyMap` that can make sure no 2 policies can affect the same http only port ?

I dont think we need to include `mergeGateways` since that feature has already kicked in earlier in the routing level and flattened all gateway listeners into IR listeners associated with a single key (per gatewayclass)

liorokman:
> I dont think we need to include `mergeGateways` since that feature has already kicked in earlier in the routing level and flattened all gateway listeners into IR listeners associated with a single key (per gatewayclass)


Good point. Fixed.

liorokman:
> @liorokman there is existing logic there checks for conflicts using a policyMap
> do you think we should reuse that style, and add a httpOnlyMap that can make sure no 2 policies can affect the same http only port ?

I tried to do this, and it didn't end up being any simpler to read. Is this critical?

arkodg:
> > @liorokman there is existing logic there checks for conflicts using a policyMap
> > do you think we should reuse that style, and add a httpOnlyMap that can make sure no 2 policies can affect the same http only port ?
> 
> I tried to do this, and it didn't end up being any simpler to read. Is this critical?

it helps keep the code and corresponding status (`Accepted=False` with `Conflicted` Reason) in the same place.
The current code checks that 2 policies cannot target the same `section` i.e. listener, this case is a variant where 2 policies cannot target the same `http` port

liorokman:
> > > @liorokman there is existing logic there checks for conflicts using a policyMap
> > > do you think we should reuse that style, and add a httpOnlyMap that can make sure no 2 policies can affect the same http only port ?
> > 
> > 
> > I tried to do this, and it didn't end up being any simpler to read. Is this critical?
> 
> it helps keep the code and corresponding status (`Accepted=False` with `Conflicted` Reason) in the same place. The current code checks that 2 policies cannot target the same `section` i.e. listener, this case is a variant where 2 policies cannot target the same `http` port

The current code that checks that 2 policies cannot target the same `section` doesn't have to look at information from inside the listener(s) identified by that section. It can do its job by looking at the strings in the policy itself.

To determine the port(s) targeted by a single policy document, the code needs to find the actual listener, or multiple listeners in the case where the policy attaches to an entire `Gateway`. We actually already have code that does this - the translation functions where the current PR does the validation.

Moving the check from the place where all of the information to do the validation exists to where this information doesn't yet exist would more or less duplicate the code from the translation functions. Is this your preference?

arkodg:
> Moving the check from the place where all of the information to do the validation exists to where this information doesn't yet exist would more or less duplicate the code from the translation functions. Is this your preference?

hoping others can chime in here as well, my preference would be to separate translation from validation, so the method could be called after getting the IR context before calling translate

also would be great if this layer did not have any envoyisms in the gateway-api layer, although we are in this situation because of the limitation in Envoy proxy, would be great to highlight that as a code comment, but use a more generic name (instead of `ShareEnvoyFilterChain`) for the helper method

liorokman:
> > Moving the check from the place where all of the information to do the validation exists to where this information doesn't yet exist would more or less duplicate the code from the translation functions. Is this your preference?
> 
> hoping others can chime in here as well, my preference would be to separate translation from validation, so the method could be called after getting the IR context before calling translate
> 
> also would be great if this layer did not have any envoyisms in the gateway-api layer, although we are in this situation because of the limitation in Envoy proxy, would be great to highlight that as a code comment, but use a more generic name (instead of `ShareEnvoyFilterChain`) for the helper method

I split the validation away from the translation, and renamed the `ShareEnvoyFilterChain` function to something a little more generic.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -285,6 +302,29 @@ func resolveCTPolicyTargetRef(policy *egv1a1.ClientTrafficPolicy, gateways []*Ga
 	return gateway
 }
 
+func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds) error {
```
arkodg, line 305: can this be also shared with SecurityPolicy and moved into helpers.go ?

arkodg (dismissed):
LGTM thanks !

internal/gatewayapi/clienttrafficpolicy.go
liorokman: Not easily. CTP is listener oriented, while SP is route oriented. The way that the correct listener is found is different in each of these validation methods.

guydc (approved):

arkodg (approved):

PR #2789, opened 2024-03-05
[release/v1.0] Pin Envoy patch version

arkodg:
* Pin envoy proxy to a specific patch release `v1.29.1` In security releases in Envoy Proxy will warrant a patch release bump in Envoy Gateway as well
* ~~Pin shutdown manager to `envoyproxy:gateway:v1.0.0-rc.1` The repository name matters here, the tag can be ignored since, at runtime the tag is computed using the shutdown manager version derived from the ld flag
https://github.com/envoyproxy/gateway/blob/930dcc1a4861609d6506891ea339722563bba53d/internal/infrastructure/kubernetes/proxy/resource.go#L270~~

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -20,13 +20,13 @@ const (
 	// DefaultDeploymentMemoryResourceRequests for deployment memory resource
 	DefaultDeploymentMemoryResourceRequests = "512Mi"
 	// DefaultEnvoyProxyImage is the default image used by envoyproxy
-	DefaultEnvoyProxyImage = "envoyproxy/envoy:distroless-v1.29-latest"
+	DefaultEnvoyProxyImage = "envoyproxy/envoy:distroless-v1.29.1"
 	// DefaultShutdownManagerCPUResourceRequests for shutdown manager cpu resource
 	DefaultShutdownManagerCPUResourceRequests = "10m"
 	// DefaultShutdownManagerMemoryResourceRequests for shutdown manager memory resource
 	DefaultShutdownManagerMemoryResourceRequests = "32Mi"
 	// DefaultShutdownManagerImage is the default image used for the shutdown manager.
-	DefaultShutdownManagerImage = "envoyproxy/gateway-dev:latest"
+	DefaultShutdownManagerImage = "envoyproxy/gateway:v1.0.0-rc.1"
 	// DefaultRateLimitImage is the default image used by ratelimit.
 	DefaultRateLimitImage = "envoyproxy/ratelimit:19f2079f"
 	// HTTPProtocol is the common-used http protocol.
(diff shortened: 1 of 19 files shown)


arkodg:
cc @davidalger

arkodg:
this is a little tricky, the CI is building a `envoyproxy:gateway-dev` image with a version set to short sha of the commit, but this PR sets the image name to `envoyproxy:gateway` so a image called `envoyproxy:gateway:<sha>` doesnt exist

one option is to enhance the logic in code - 
* if tag starts with `v` use  `envoyproxy:gateway`
* else use `envoyproxy:gateway-dev`

davidalger:
> one option is to enhance the logic in code -
> 
> * if tag starts with v use envoyproxy:gateway
> * else use envoyproxy:gateway-dev

+1 for this. sounds reasonable to me

Xunzhuo:
> this is a little tricky, the CI is building a `envoyproxy:gateway-dev` image with a version set to short sha of the commit, but this PR sets the image name to `envoyproxy:gateway` so a image called `envoyproxy:gateway:<sha>` doesnt exist
> 
> one option is to enhance the logic in code -
> 
> * if tag starts with `v` use  `envoyproxy:gateway`
> * else use `envoyproxy:gateway-dev`
> 
> wdyt @envoyproxy/gateway-maintainers

A more accurate regex for tag is better.

arkodg:
oh fun 

> > A more accurate regex for tag is better. Sometimes the sha will start with v too
> 
> @Xunzhuo good point... agreed, detection would need to be more specific than a letter prefix.
> 
> I'm looking at how this could be implemented and not sure how we can detect the tag given how the release works. Currently it's just copying the docker image: https://github.com/envoyproxy/gateway/actions/runs/8154910922/job/22289300255#step:5:1
> 
> Since the image `envoyproxy/gateway-dev:<sha>` is copied to `envoyproxy/gateway:<tag>` the ld flag is going to be the `<sha>` from the build pipeline:
> 
> ```
> $ docker run --rm envoyproxy/gateway:v1.0.0-rc.1 version
> ENVOY_GATEWAY_VERSION: v1.0.0-rc.1
> ENVOY_PROXY_VERSION: distroless-v1.29-latest
> GATEWAYAPI_VERSION: v1.0.0
> SHUTDOWN_MANAGER_VERSION: 7d48480
> GIT_COMMIT_ID: 7d48480fed10100a699ac360b2e1ace6eb49deab
> ```
> 
> `ENVOY_GATEWAY_VERSION` is set to the release based on what's in the `VERSION` file, but this won't work for `SHUTDOWN_MANAGER_VERSION` which needs to match the current build so the right image will be used in e2e tests.
> 
> Nothing immediately stands out as a simple and clean way of solving this without changing how the image is released so the ld flags can be set accordingly.

oh fun, thanks for pointing this out 
the fact that we dont rebuild a image when releasing, but instead promote a image (retag it), makes this chicken egg situation worse (gateway image defining a gateway image)

arkodg:
ive split up the problem and created a separate GH issue for shutdown manager to unblock this PR

zirain (approved):

Xunzhuo (approved):

PR #2795, opened 2024-03-06
Add referenced BackendRefs for ExtAuth to Resource Tree

zhaohuabing:
ExtAuth functionality was broken due to missing BackendRefs in the Resource Tree. This PR resolves this issue.
I have manually verified the fix in my dev environment and currently working on the associated e2e test. The e2e test will be in a separate follow-up PR.

This PR also adds BackendTLSPolicy support to the BackendRefs of the Ext Auth services.


internal/xds/translator/extauth.go
```diff
@@ -93,7 +93,10 @@ func extAuthFilterName(route *ir.HTTPRoute) string {
 }
 
 func extAuthConfig(extAuth *ir.ExtAuth) *extauthv3.ExtAuthz {
-	config := &extauthv3.ExtAuthz{}
+	config := &extauthv3.ExtAuthz{
+		TransportApiVersion: corev3.ApiVersion_V3,
```
arkodg, line 97: why not use `AUTO` here ?

internal/provider/kubernetes/controller.go
```diff
@@ -434,6 +416,50 @@ func (r *gatewayAPIReconciler) processSecurityPolicySecretRefs(
...
+		extAuth := policy.Spec.ExtAuth
+		if extAuth != nil {
+			var backendRef gwapiv1.BackendObjectReference
+			if extAuth.GRPC != nil {
+				backendRef = extAuth.GRPC.BackendRef
+			} else {
+				backendRef = extAuth.HTTP.BackendRef
+			}
+
+			backendNamespace := gatewayapi.NamespaceDerefOr(backendRef.Namespace, policy.Namespace)
```
arkodg, line 430: can you make sure that if a backendTLSPolicy exists for this backendRef that is included into `resources` as well ?

sidenote (not directly related to this PR) I dont see any tls context in the translation https://github.com/envoyproxy/gateway/blob/ef1934f1461157afc890cddac863418c3e90b4f0/internal/gatewayapi/securitypolicy.go#L753
shouldnt we be reusing this function ?
https://github.com/envoyproxy/gateway/blob/ef1934f1461157afc890cddac863418c3e90b4f0/internal/gatewayapi/route.go#L1074

internal/xds/translator/extauth.go
zhaohuabing: It's deprecated. We should set it to v3 explicitly.
```
	// When not specified, we assume v2, to ease migration to Envoy's stable API
	// versioning. If a client does not support v2 (e.g. due to deprecation), this
	// is an invalid value.
	//
	// Deprecated: Marked as deprecated in envoy/config/core/v3/config_source.proto.
	ApiVersion_AUTO ApiVersion = 0
```

internal/provider/kubernetes/controller.go
zhaohuabing: > shouldnt we be reusing this function ?

I have a TODO in the code and plan to do that but maybe later after v1.0.0. It requires massive modification on the current xRoute translation because the current Destination processing is tied to RouteParentContext.

internal/provider/kubernetes/controller.go
```diff
@@ -368,37 +252,139 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+				gwcResource.EndpointSlices = append(gwcResource.EndpointSlices, &endpointSlice)
+			}
+		}
+	}
+}
+
+// processSecurityPolicyObjectRefs adds the referenced resources in SecurityPolicies
 // to the resourceTree
-func (r *gatewayAPIReconciler) processSecurityPolicySecretRefs(
+// - Secrets for OIDC and BasicAuth
```
arkodg, line 373: you probably need to also make sure that the controller reconciles if the ext auth backend ref changes :)
so the indexer needs to be added or updated similar to https://github.com/envoyproxy/gateway/blob/ef1934f1461157afc890cddac863418c3e90b4f0/internal/provider/kubernetes/indexers.go#L380

zhaohuabing: @arkodg  It's already there https://github.com/envoyproxy/gateway/blob/ef1934f1461157afc890cddac863418c3e90b4f0/internal/provider/kubernetes/indexers.go#L406

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -6,13 +6,166 @@
...
+	}
+	tlsBundle.CACertificate = &ir.TLSCACertificate{
+		Certificate: []byte(ca),
+		Name:        fmt.Sprintf("%s/%s-ca", backendTLSPolicy.Name, backendTLSPolicy.Namespace),
+	}
+
+	return tlsBundle, nil
+}
+
+func (t *Translator) ProcessBackendTLSPoliciesAncestorRef(backendTLSPolicies []*gwapiv1a2.BackendTLSPolicy, gateways []*GatewayContext) []*gwapiv1a2.BackendTLSPolicy {
```
zhaohuabing, line 166: Why do we need to check whether the parent Gateway exist?

arkodg (approved):
LGTM thanks !
its a big change :). but the ext auth e2e and backend tls e2e should help verify this

arkodg:
e2e is failing, kicked the CI, lets wait for the tests to pass before we merge

Xunzhuo (approved):

Issue #2631, opened 2024-02-16
Replace Status with PolicyAncestorStatus in CTP, BTP and SP


arkodg:
*Description*:
>Describe the issue.

Use the upstream `PolicyAncestorStatus` definition which creates a top level structure per controller
https://gateway-api.sigs.k8s.io/geps/gep-713/#standard-status-struct

instead of what we have today e.g. https://github.com/envoyproxy/gateway/blob/1775624114aedee731bee565a0a100eeee80fc13/api/v1alpha1/backendtrafficpolicy_types.go#L97

Should fix https://github.com/envoyproxy/gateway/issues/2520

shawnh2:
by updating `Status` in Policies to `PolicyAncestorStatus`, this will also requires some changes in `egctl x status`

shawnh2:
should we change it directly to `PolicyAncestorStatus` or instead consider [`PolicyStatus`](https://github.com/kubernetes-sigs/gateway-api/blob/72f679c7c67228f5b4340e0c2f50cc04b36945b1/apis/v1alpha2/policy_types.go#L187) which each `Ancestors` can refer to objects that are either the Target of a policy or above it in terms of object hierarchy, as described [in here](https://github.com/kubernetes-sigs/gateway-api/blob/72f679c7c67228f5b4340e0c2f50cc04b36945b1/apis/v1alpha2/policy_types.go#L129)?

arkodg:
ah I meant using the struct you referenced https://github.com/kubernetes-sigs/gateway-api/blob/72f679c7c67228f5b4340e0c2f50cc04b36945b1/apis/v1alpha2/policy_types.go#L187, thanks for clarifying

arkodg:
blocked on https://github.com/envoyproxy/gateway/issues/2724#issuecomment-1970130948

PR #2802, opened 2024-03-06
fix: skip the ReasonTargetNotFound for all policies

### LINKED ISSUE #2724: BTP Status getting modified by unrelated GatewayClass
facing the same issue with BTP as in e2e test https://github.com/envoyproxy/gateway/actions/runs/8073622448/job/22057769156?pr=2665#step:6:2130

#2665 introduce merge gateways e2e test, that requires multiple-gatewayclass per controller feature. but this test will affects all the other test cases that have policies attached. because the new gatewayclass in controller will also update the policy status, even the policy is not belong to new GC's.

 so by fixing #2631, I think this issue can be resolved. assigning myself to these two issues.

_Originally posted by @shawnh2 in https://github.com/envoyproxy/gateway/issues/2520#issuecomment-1968509285_

shawnh2:
**What type of PR is this?**

currently, skip the `PolicyReasonTargetNotFound` for all policies as a workaround, will revisit this issue via #2631 

Fixes #2724


arkodg:
thanks @shawnh2, code looks good, just to be sure, can we add a test case where all 4 policy types target an unknown gateway, and we can make sure there is no `status` set for those

shawnh2:
> thanks @shawnh2, code looks good, just to be sure, can we add a test case where all 4 policy types target an unknown gateway, and we can make sure there is no `status` set for those

sure, some tests has been added.

arkodg (approved):
LGTM thanks !

zirain:
//retest

zirain (approved):

PR #2803, opened 2024-03-06
refactor: set instead of map for mergeGateways

### LINKED ISSUE #2796: use set[T] instead of map[T]bool

yeedove:
**What type of PR is this?**

-  `k8s.io/apimachinery/pkg/util/sets`
- `set[T]` instead of `map[T]bool` for mergeGateways

Fixes #2796


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -827,12 +827,13 @@ func (t *Translator) buildHTTPActiveHealthChecker(h *egv1a1.HTTPActiveHealthChec
 
 	var irStatuses []ir.HTTPStatus
```
shawnh2, line 828: can you also create `irStatuses` using `make`, like you did to other methods?

shawnh2 (approved):
LGTM, thanks!

zirain (approved):

PR #2805, opened 2024-03-06
api: support buffer limit in clientTrafficPolicy

### LINKED ISSUE #2600: Add a knob to make `per_connection_buffer_limit_bytes` configurable

yaelSchechter:

This PR introduces configurability for the` per_connection_buffer_limit_bytes` parameter. It uses the struct proposed in this PR: https://github.com/envoyproxy/gateway/pull/2709, pending its merge.

Fixes #2600


api/v1alpha1/connection_types.go
```diff
@@ -5,14 +5,21 @@
...
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
 
 // Connection allows users to configure connection-level settings
 type Connection struct {
 	// Limit defines limits related to connections
 	//
 	// +optional
 	Limit *ConnectionLimit `json:"limit,omitempty"`
+	// ConnectionBufferLimit provides configuration for the maximum buffer size for incoming connections.
```
zhaohuabing, line 19: @yaelSchechter This PR looks good! Just a tiny nit on the comment to clarify that the buffer size limit applies on each connection individually.  

```suggestion
	// ConnectionBufferLimit provides configuration for the maximum buffer size for each incoming connection.
```

yaelSchechter: Thanks @zhaohuabing! Fixed the comment as suggested.

api/v1alpha1/connection_types.go
```diff
@@ -17,6 +17,7 @@ type Connection struct {
 	// +optional
 	Limit *ConnectionLimit `json:"limit,omitempty"`
```
zhaohuabing, line 18: Now we get two Limits in this struct. :-)  Should we change this to `ConnectionLimit`?

@guydc @arkodg

arkodg: sure yah, not ideal, but lets do it

api/v1alpha1/connection_types.go
```diff
@@ -5,14 +5,22 @@
...
 type Connection struct {
 	// Limit defines limits related to connections
 	//
 	// +optional
 	Limit *ConnectionLimit `json:"limit,omitempty"`
+	// ConnectionBufferLimit provides configuration for the maximum buffer size for each incoming connection.
+	// Default: 32768 bytes.
+	//
+	// +optional
+	BufferLimit *resource.Quantity `json:"bufferLimit,omitempty"`
```
arkodg, line 23: any specific reason to use `quantity` and not `uint32` ?

yaelSchechter: Thinking about UX, using quantity as input lets users specify sizes in units like megabytes, gigabytes, etc., just like how we use `*Duration` in some cases, could make things smoother

api/v1alpha1/connection_types.go
```diff
@@ -5,14 +5,23 @@
...
+)
 
 // Connection allows users to configure connection-level settings
 type Connection struct {
 	// ConnectionLimit defines limits related to connections
 	//
 	// +optional
 	ConnectionLimit *ConnectionLimit `json:"connectionLimit,omitempty"`
+	// BufferLimit provides configuration for the maximum buffer size for each incoming connection.
+	// For example, 128Mi, 500m, 2G, etc.
```
arkodg, line 20: looks like the type accepts a lot more such as `500m` is not relevant for this field.
can we limit it to specific types relevant for memory

yaelSchechter: Do you mean, limit the use for only binary suffixes (`Ki | Mi | Gi | Ti | Pi | Ei`) and not decimal suffixed? (`m | "" | k | M | G | T | P | E`) 
If so, I can change the example and limit it using the kubebuilder schema validation.

yaelSchechter: Maybe `*uint32` is a better option after all, validating the `*resource.Quantity` might be a bit complex (unnecessarily) :) 
Changed to `*uint32`.

arkodg: fwiw I did like the fact that the abstracted type was handling known quantities very well

yaelSchechter: what do you think about runtime validation instead? users will get an error on the CTP but it won't prevent them from applying a resource with cpu specific suffixes

arkodg: sure, so the outlining the goal
* as a user, i'd like to specify mem in terms on absolute value (bytes), or known quantities (K, M, G, Ki, Mi, Gi) 

Options
* CEL validation for resource.Quanitity
* Runtime Validation of resource.Quanitity
* Creating our own type similar to what upstream did for Duration
https://github.com/kubernetes-sigs/gateway-api/blob/9183329ef9cff3f9222f001d2e112f325108c761/apis/v1/shared_types.go#L702

yaelSchechter: Changed back to `Quantity` and added a cel validation. It will check that the suffix, if present, is compatible with the suffixes for memory: [units of memory in k8s](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory). 

One more thing, values without suffix, like `1500` (bytes) must be specified with `""` (otherwise there is an error). Is it something known to users as k8s behavior, or should I document it as well?

arkodg: can you elaborate on `""` , that feels like something that might annoy the user :) 
thoughts on the string approach with kubebuilder validations, and then typecasting into `resource.Quantity` so the user doesnt hit the gotcha ?

api/v1alpha1/connection_types.go
```diff
@@ -5,14 +5,25 @@
...
 	// ConnectionLimit defines limits related to connections
 	//
 	// +optional
 	ConnectionLimit *ConnectionLimit `json:"connectionLimit,omitempty"`
+	// BufferLimit provides configuration for the maximum buffer size in bytes for each incoming connection.
+	// For example, 20Mi, 1Gi, 256Ki etc.
+	// Note that when the suffix is not provided, the value is interpreted as bytes.
+	// Default: 32768 bytes.
+	//
+	// +kubebuilder:validation:XValidation:rule="self.matches(r\"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\")",message="bufferLimit must be of the format \"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\""
```
liorokman, line 24: I suggest changing the CEL rule to:

`type(self) == string ? self.matches(r"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$") : true`

This way it works for raw numbers as well as for string values.

[CEL playground](https://playcel.undistro.io/?content=H4sIAAAAAAAAAyWMTwuCQBTEv8rj1UH7I3rokCDSIYrKCuoSuwaSb3OpVtldoai%2BeytdZn7MDPPGsrAFxtiDJWkCacBWzlTTWugqh3CaZRuoNawOuy2IWj8KG3DFlaG7iCGahKEoRYkjpGejyRhZK%2FdoXw153cSHJAFjtVRXSKFLAvdwqch4muOZReNpPmSh04HH5vvjIlvn8vOnW%2B6nfY4%2BxGB1S%2Fj9AZEILbWxAAAA)

api/v1alpha1/connection_types.go
yaelSchechter: Lior's suggestion solved the problem 👍

api/v1alpha1/connection_types.go
yaelSchechter: Thanks, that did the trick

guydc (approved):
LGTM

zirain (approved):

arkodg (approved):

PR #2801, opened 2024-03-06
fix bootstrap merge

zirain:
Fixes #2800

PR #2811, opened 2024-03-07
Merge yaml

zirain:
alternative of https://github.com/envoyproxy/gateway/pull/2801

cc @arkodg


zirain:
close this in favor of https://github.com/envoyproxy/gateway/pull/2801

Issue #2807, opened 2024-03-06
Enable HPA for Envoy Proxy introduce unexpected behavior


ardikabs:
*Description*:
Relates to https://github.com/envoyproxy/gateway/pull/2257

Once enabled `envoyHpa` on EnvoyProxy API some unexpected behavior was observed. 

During reconciling, Envoy Gateway managed to create `HorizontalPodAutoscaler` resource successfully. However, when it comes to any operation related to Gateway API resource, Envoy Gateway unexpectedly reset the replicas of Envoy Proxy deployment to their original value specified in `EnvoyProxy.spec.provider.kubernetes.envoyDeployment.replicas`, if this field is undefined, then it will set to 1 replica. Though, after some time, when the HPA controller kicks in again, it will set back to its `minReplicas`, but I am a bit concerned in the event of high traffic, and any operation for Gateway API applied, it enforces the Envoy Proxy deployment to be scaled-down.

*Repro steps*:

Apply the below EnvoyProxy manifest, then assuming some of HTTPRoute resources exist, then do any operation such as modification into that resource. You will observe a scale-down operation into EnvoyProxy deployment, from 3 replicas (specified in HPA's `minReplicas` field) to 1 replica (default value for EnvoyProxy replica).

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: default
  namespace: envoy-gateway-system
spec:
  logging:
    level:
      default: warn
  provider:
    kubernetes:
      envoyDeployment:
        pod:
          labels:
            service_name: envoy-proxy
            service_group: envoy
        container:
          resources:
            limits:
              cpu: 2
              memory: 2Gi
            requests:
              cpu: 1
              memory: 1Gi
      envoyHpa:
        maxReplicas: 12
        minReplicas: 3
        metrics:
          - resource:
              name: cpu
              target:
                averageUtilization: 60
                type: Utilization
            type: Resource
    type: Kubernetes
```

*Proposal*:

Even though the code has already been set up to prevent reverting replicas to their original value when the HPA controller kicks in (to adjust the Envoy Proxy deployment replicas), the Envoy Gateway seems still not aware when it comes in `createOrUpdateDeployment` function during reconciliation. 

https://github.com/envoyproxy/gateway/blob/af72b32773d8c5d238c02e9d15a0a4fd67e3f523/internal/infrastructure/kubernetes/infra_resource.go#L66-L86
[...]

PR #2816, opened 2024-03-07
fix: omit default replicas on Kubernetes Deployment

ardikabs:
**What type of PR is this?**

Fixes: #2807


Xunzhuo:
What if I set replicas at EnvoyProxy CR and enabled HPA too, we should have a priority over them.

ardikabs:
> What if I set replicas at EnvoyProxy CR and enabled HPA too, we should have a priority over them.

it is described in https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#migrating-deployments-and-statefulsets-to-horizontal-autoscaling, that's why I add a note on the docs string @Xunzhuo

api/v1alpha1/shared_types.go
```diff
@@ -348,6 +348,8 @@ const (
 )
 
 // KubernetesHorizontalPodAutoscalerSpec defines Kubernetes Horizontal Pod Autoscaler settings of Envoy Proxy Deployment.
+// When HPA is enabled, it is recommended that the value in `KubernetesDeploymentSpec.replicas` be removed, otherwise
```
ardikabs, line 351: check here @Xunzhuo

Xunzhuo: Yes, I mean if eg should take care the priority in code logics. cc @arkodg @zirain

qicz: +1, HPA will override replicas

arkodg: Prefer if we kept it simple for now, so our API matches K8s behavior

arkodg (approved):
LGTM thanks !
glad we caught this one early

arkodg:
@ardikabs can we also add a CEL validation to disable `hpa` and `replicas` from being set at the same time ?
this can alway be overriden using the `patch` field in case someone really wants to

ardikabs:
> @ardikabs can we also add a CEL validation to disable `hpa` and `replicas` from being set at the same time ?

if we add CEL validation, are we not trying to match Kubernetes behavior?

arkodg (approved):

zirain (approved):

PR #2817, opened 2024-03-07
fix: ReplaceFullPath not working for root path (/)

eternalphane:
**What type of PR is this?**

`ReplaceFullPath` should support replacing root path during URL rewrite.

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -314,7 +314,7 @@ func buildXdsURLRewriteAction(destName string, urlRewrite *ir.URLRewrite, pathMa
 		if urlRewrite.Path.FullReplace != nil {
 			routeAction.RegexRewrite = &matcherv3.RegexMatchAndSubstitute{
 				Pattern: &matcherv3.RegexMatcher{
-					Regex: "/.+",
+					Regex: "/.*",
 				},
 				Substitution: *urlRewrite.Path.FullReplace,
 			}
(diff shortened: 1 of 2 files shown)


zirain (changes requested):
In case this happen, please kindly add an e2e test.

arkodg:
@eternalphane can you share the config that not working ?
would be good to compare it with expectations from the spec https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPPathModifier

eternalphane:
@arkodg [This simple example](https://github.com/kubernetes-sigs/gateway-api/blob/main/examples/standard/http-rewrite.yam) is not working when accessing the root path:
```yaml
kind: HTTPRoute
metadata:
  name: http-filter-1
  namespace: gateway-api-example-ns1
spec:
  parentRefs:
  - name: my-filter-gateway
    sectionName: http
  hostnames:
  - my-filter.example.com
  rules:
  - filters:
    - type: URLRewrite
      urlRewrite:
        path:
          type: ReplaceFullPath
          replaceFullPath: /foo
```

arkodg:
@eternalphane is the request to my-filter.example.com  or my-filter.example.com/

eternalphane:
> is the request to my-filter.example.com or my-filter.example.com/

Both seem to be the same 😂
image

internal/xds/translator/route.go
```diff
@@ -314,7 +314,7 @@ func buildXdsURLRewriteAction(destName string, urlRewrite *ir.URLRewrite, pathMa
 		if urlRewrite.Path.FullReplace != nil {
 			routeAction.RegexRewrite = &matcherv3.RegexMatchAndSubstitute{
 				Pattern: &matcherv3.RegexMatcher{
-					Regex: "/.+",
+					Regex: "/.*",
```
arkodg, line 317: hey sorry for the delay here, lets tighten this up a bit with `^/.*$`

so something random like `blah/` doesnt get accepted

arkodg:
closed in favor of https://github.com/envoyproxy/gateway/pull/3530

PR #2818, opened 2024-03-07
feat: Support Upstream TLS to multiple Backends

### LINKED ISSUE #2774: Support Upstream TLS to multiple Backends
*Description*:
would be used.

Today we only build 1 TLS Socket based on the first Destination Setting associated with the first BackendRef
https://github.com/envoyproxy/gateway/blob/9b48ff913e742a49fd6b8ebbec2ce6b156d20e5a/internal/xds/translator/translator.go#L518

This needs to be enhanced to support a unique TLS socket per backendref

Should be possible using the `tls_socket_matches` field https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto

liorokman:

This PR moves upstream TLS configuration from the default `cluster.transport_socket` into `cluster.transport_socket_matchers`, and sets up the endpoints to allow matching to succeed.

Fixes #2774


internal/xds/translator/cluster.go
```diff
@@ -77,6 +78,30 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		cluster.TransportSocket = args.tSocket
 	}
 
+	for i, ds := range args.settings {
+		if ds.TLS != nil {
+			// for upstreamTLS , a fixed sni can be used. use auto_sni otherwise
```
arkodg, line 83: this comment is no longer valid, and can be removed

internal/xds/translator/cluster.go
```diff
@@ -77,6 +78,30 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
...
+			// https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/sni#faq-how-to-setup-sni:~:text=For%20clusters%2C%20a,for%20trust%20anchor.
+			socket, err := buildXdsUpstreamTLSSocketWthCert(ds.TLS)
+			if err != nil {
+				// TODO: Log something here
+				return nil
+			}
+			if args.proxyProtocol != nil {
```
arkodg, line 90: this should probably live outside the loop ?

internal/xds/translator/cluster.go
```diff
@@ -77,6 +78,30 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
...
+				socket = buildProxyProtocolSocket(args.proxyProtocol, socket)
+			}
+			cluster.TransportSocketMatches = append(cluster.TransportSocketMatches, &clusterv3.Cluster_TransportSocketMatch{
+				Name: fmt.Sprintf("destination-tls-config-%d", i),
```
arkodg, line 94: nit: each setting has a 1:1 mapping with a k8s backendRef
we could enhance https://github.com/envoyproxy/gateway/blob/f595862795047ddc35a2d181e8419f68af21187e/internal/ir/xds.go#L819 to add a `name` field and use that here in the metadata as `name: <nsName>`
(can be a big diff, so non blocker)

arkodg:
thanks for picking this one up @liorokman !

internal/xds/translator/cluster.go
liorokman: It can't live outside the loop because the proxy protocol socket wraps the socket created inside the loop.

internal/xds/translator/cluster.go
liorokman: done

internal/xds/translator/cluster.go
liorokman: I changed the name so that it's based on the cluster-name and is a little easier to understand.

Adding the name to the destination settings is a bigger change, since these are created in all sorts of places. Maybe a future PR?

arkodg (approved):
LGTM thanks !

zirain:
@liorokman need rebase

internal/xds/translator/cluster.go
liorokman: Done.

guydc (approved):

arkodg (approved):

Issue #2727, opened 2024-02-29
E2E test for ext auth


zhaohuabing:
*Description*:
>Describe the issue.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #2830, opened 2024-03-08
Ext auth e2e tests

zhaohuabing:
This PR adds e2e tests for Ext Auth with HTTP auth service.

It also:
* Export util methods to avoid `unparam` link errors.
* Use underscore to replace hyphen in test file names to align with the go file name convention.

E2E tests for GRPC auth service will be raised in a follow-up PR.

Related https://github.com/envoyproxy/gateway/issues/2727


arkodg (approved):
LGTM thanks !

Issue #2832, opened 2024-03-08
Unable to route to FQDN EndpointSlice when using TLSRoute


arkodg:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

Unable to route, seeing errors in Envoy
```
[2024-03-08 02:38:43.755][1][warning][main] [source/server/server.cc:925] There is no configured limit to the number of allowed active downstream connections. Configure a limit in `envoy.resource_monitors.downstream_connections` resource monitor.
[2024-03-08 02:41:40.372][1][warning][config] [source/extensions/config_subscription/grpc/delta_subscription_state.cc:269] delta config for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment rejected: malformed IP address: httpbin.org. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
[2024-03-08 02:41:40.372][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment rejected: malformed IP address: httpbin.org. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
[2024-03-08 02:41:40.372][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment rejected: malformed IP address: httpbin.org. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
{"start_time":"2024-03-08T02:42:06.176Z","method":"-","x-envoy-origin-path":"-","protocol":"-","response_code":"0","response_flags":"UH","response_code_details":"-","connection_termination_details":"-","upstream_transport_failure_reason":"-","bytes_received":"0","bytes_sent":"0","duration":"0","x-envoy-upstream-service-time":"-","x-forwarded-for":"-","user-agent":"-","x-request-id":"-",":authority":"-","upstream_host":"-","upstream_cluster":"tlsroute/default/httpbin/rule/-1","upstream_local_address":"-","downstream_local_address":"10.1.10.196:10443","downstream_remote_address":"192.168.65.4:34028","requested_server_name":"-","route_name":"-"}
```
[...]

PR #2833, opened 2024-03-08
bug: compute endpointType for all protocol types

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/2832

### DIFF
--- a/internal/xds/translator/cluster.go
+++ b/internal/xds/translator/cluster.go
@@ -57,6 +57,22 @@ const (
 	EndpointTypeStatic
 )
 
+func buildEndpointType(settings []*ir.DestinationSetting) EndpointType {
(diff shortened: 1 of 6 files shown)


zirain:
able to add an e2e test?

arkodg:
> able to add an e2e test?

the issue is do we want to route to an external FQDN ? Ive seen bad statuses being returned in the past due to middleware thinking we are bots etc, may cause flakes

zirain:
> > able to add an e2e test?
> 
> the issue is do we want to route to an external FQDN ? Ive seen bad statuses being returned in the past due to middleware thinking we are bots etc, may cause flakes

flakes better than error?

zirain (approved):

PR #2665, opened 2024-02-21
e2e: add e2e test for MergeGateways feature

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #2029

PR #2839, opened 2024-03-08
e2e: graceful envoy shutdown

### LINKED ISSUE #2610: E2E Zero DownTime Test when upgrading Envoy Proxy Versions

guydc:
- Update the upgrade suite to use its own namespace, gatewayclass, envoyproxy, gateway and backend. 
   - The new gatewayclass will have 2 replicas and 15s drain shutdown, which are required for this test to succeed. Increasing the replica count and shutdown drain time for existing gateways causes certain tests like local rate limit to fail and prolongs the suite. 
   - The need for another suite is explained in #2665, as only a single gatewayclass is currently supported in conformance suites
- Adds a test that rolls out the proxy deployment while running load:
  -  Restart is achieved by patching EnvoyProxy resource with a custom restart annotation. 
  - The shutdown manager should ensure that connections are drained gracefully and not closed forcibly. New connections are gradually loadbalanced to the active proxy envoy instance. 
- Small fixes to existing backend upgrade test and moving common parts to utils
- Minor adjustments of the EG upgrade test to the new suite's resources

Future improvements:
- Leverage the HA EnvoyProxy with Shutdown manager as part of the EG upgrade test, and check for hitless upgrades. 
- Move the backend upgrade test into this suite, and restart the upgrade "infra-backend". The current test restarts a common backend, in a way that can affect other tests in the suite.  
- When all limitations are resolved, we can consider merging this suite back into the main suite. 

Fixes #2610


guydc:
blocked on: https://github.com/envoyproxy/gateway/pull/2665#issuecomment-1984987844

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #2840, opened 2024-03-09
[Release-1.0] pin envoy to v1.29.2

zirain:
Envoy released v1.29.2, contains a codec revert which caused production incidents.


arkodg (approved):

zirain:
cc @Xunzhuo

PR #2841, opened 2024-03-09
add e2e tests for ext auth with grpc auth service

zhaohuabing:
This PR adds e2e tests for ext auth with grpc auth service.

It also adds the missing tls socket config for ext auth service clusters in the xds translator.

fix https://github.com/envoyproxy/gateway/issues/2727


test/e2e/testdata/ext-auth-grpc-securitypolicy.yaml
```diff
@@ -0,0 +1,53 @@
...
+  namespace: gateway-conformance-infra
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: http-with-ext-auth
+  extAuth:
+    grpc:
+      backendRef:
+        name: grpc-ext-auth
```
arkodg, line 51: can we use tls for this example ?

test/e2e/testdata/ext-auth-http-service.yaml
```diff
@@ -66,6 +66,12 @@ spec:
         volumeMounts:
         - name: http-ext-auth
           mountPath: /usr/src/app
+        readinessProbe:
```
arkodg, line 69: curious why this is needed ?

arkodg (dismissed):
LGTM thanks !

test/e2e/testdata/ext-auth-http-service.yaml
zhaohuabing: Because I want to avoid the test calls the ext auth service before it's ready. In that case, the test would fail.

zirain (approved):

PR #2845, opened 2024-03-09
chore: remove ProcessBackendTLSPoliciesAncestorRef

zhaohuabing:
`ProcessBackendTLSPoliciesAncestorRef` checks whether the `Gateway` in `policy.Status.Ancestors` exists, if not, it removes that Ancestor.

This might be problematic if multiple controllers are reconciling the `BackendTLSPolicy`, as the `policy.Status.Ancestors` could contain valid `Gateway` references from another controller.


zirain (commented):
LGTM, defer to @arkodg

shawnh2:
I've been working on adding `PolicyStatus` for xPolicy recently, like in #2846. Maybe `BackendTLSPolicy` can also be handled via this PR.

zhaohuabing:
> I've been working on adding `PolicyStatus` for xPolicy recently, like in #2846. Maybe `BackendTLSPolicy` can also be handled via this PR.

`BackendTLSPolicy` already adapted `gwapiv1a2.PolicyStatus`. This PR is trying to fix another issue.

internal/provider/kubernetes/controller.go
```diff
@@ -873,9 +873,9 @@ func (r *gatewayAPIReconciler) processBackendTLSPolicies(
 
 	for _, policy := range backendTLSPolicies.Items {
 		policy := policy
-		// Discard Status to reduce memory consumption in watchable
-		// It will be recomputed by the gateway-api layer
-		policy.Status = gwapiv1a2.PolicyStatus{} // todo ?
+		// The status of BackendTLSPolicies can't be discarded because the status
```
arkodg, line 876: can or can't ?

zhaohuabing: IMO can't

arkodg (approved):

PR #2846, opened 2024-03-09
feat: add PolicyStatus for BackendTrafficPolicy

### LINKED ISSUE #2631: Replace Status with PolicyAncestorStatus in CTP, BTP and SP

shawnh2:
Like [GW-API points out](https://github.com/kubernetes-sigs/gateway-api/blob/72f679c7c67228f5b4340e0c2f50cc04b36945b1/apis/v1alpha2/policy_types.go#L137), Policy should be treated as a shared resources. Different resources targeting the same policy may have different effects, and the status of these resources should be populated as its `AncestorStatus`.

related to #2631


zhaohuabing (approved):
LGTM. Thanks!

internal/provider/kubernetes/controller.go
```diff
@@ -833,7 +833,7 @@ func (r *gatewayAPIReconciler) processBackendTrafficPolicies(ctx context.Context
 		policy := policy
 		// Discard Status to reduce memory consumption in watchable
 		// It will be recomputed by the gateway-api layer
-		policy.Status = egv1a1.BackendTrafficPolicyStatus{}
+		policy.Status = gwapiv1a2.PolicyStatus{}
```
shawnh2, line 836: should we modify this like #2845?

zhaohuabing: I believe so, the status can't be discarded because other controllers may also modified it.

shawnh2: got it!

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -125,15 +136,24 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
...
-			gateway := resolveBTPolicyGatewayTargetRef(policy, gatewayMap)
+			gateway := resolveBTPolicyGatewayTargetRef(policy, t.GatewayControllerName, gatewayMap)
 			if gateway == nil {
 				continue
 			}
 
-			t.translateBackendTrafficPolicyForGateway(policy, gateway, xdsIR)
+			// Use the gatewayclass of gateway directly as its ancestor reference,
+			// since one gateway only associate with one gatewayclass.
+			ancestorRefs := []gwv1a2.ParentReference{
```
zhaohuabing, line 146: IMO, we should keep consistency and use `Gateway` as the ancestor if we can, which is suggested by the Gateway API. When a `Status` doesn't directly related to a specific `Gateway`(e.g, in the case of invalid configuration), we can use `GatewayClass` as the ancestor.

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -125,15 +136,24 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 			res = append(res, policy)
 
 			// Negative statuses have already been assigned so its safe to skip
-			gateway := resolveBTPolicyGatewayTargetRef(policy, gatewayMap)
+			gateway := resolveBTPolicyGatewayTargetRef(policy, t.GatewayControllerName, gatewayMap)
```
zhaohuabing, line 139: To simplify the code structure, I would like return errors directly from sub-methods and consolidate status handling within the outermost layer. This approach would also improve readability. 

It's just my preferred coding flavor, I recognize others may have different approaches. 

This is not blocking, we can handle it later if it makes sense.

zhaohuabing:
Overall looks good. Just a few nits.

internal/gatewayapi/backendtrafficpolicy.go
shawnh2: Modified

internal/gatewayapi/backendtrafficpolicy.go
shawnh2: make sense, I have done some refactor.

zhaohuabing (approved):
LGTM. Thanks!

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -86,14 +87,17 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 			res = append(res, policy)
 
 			// Negative statuses have already been assigned so its safe to skip
-			route := resolveBTPolicyRouteTargetRef(policy, routeMap)
-			if route == nil {
+			route, resolveErr := resolveBTPolicyRouteTargetRef(policy, routeMap)
+			if route == nil && resolveErr == nil {
```
arkodg, line 91: is this so that you can update `status.SetResolveErrorForPolicyAncestors` by first computing ``ancestorRefs` ?

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -86,14 +87,17 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
...
 				continue
 			}
 
 			// Find the Gateway that the route belongs to and add it to the
-			// gatewayRouteMap, which will be used to check policy overrides
-			for _, p := range GetParentReferences(route) {
+			// gatewayRouteMap and ancestor list, which will be used to check
+			// policy overrides and populate its ancestor status.
+			parentRefs := GetParentReferences(route)
```
arkodg, line 98: what happens if route is nil ?

internal/gatewayapi/backendtrafficpolicy.go
shawnh2: yes. in the previous commit, the `ResolveError` is calculated in each resolve method and then got surfaced, which requires `ancestorRefs` to be computing again, since the computing logic are the same. so by returning this `resolveErr`, we just need to compute `ancestorRefs` once.

internal/gatewayapi/backendtrafficpolicy.go
shawnh2: as you can see the changes in L297 and L308, the route got returned even it has resolve err in order to compute its `ancestorRefs`. so the route here will not be nil.

internal/gatewayapi/backendtrafficpolicy.go
arkodg: thoughts on simplifying this logic ?
```
if route == nil {
  continue
}

ancestorRefs := getBTPAncestorReferences()

if resolveErr != nil {
  status.SetResolveErrorForPolicyAncestors
  continue
}
```

shawnh2: yes, we can simplify the first if statement

shawnh2: as for the get btpAncestorRefs part, since this logic can be reused by other policies which can target to xroute, seems better to wrap it up into a more generic method.

shawnh2: > as for the get btpAncestorRefs part, since this logic can be reused by other policies which can target to xroute, seems better to wrap it up into a more generic method.

actually this loop is also used to collect `gatewayRouteMap`, not worthy to introduce another loop. so just done some simplification

Xunzhuo (approved):

zhaohuabing:
LGTM. Defer to @arkodg for the final call :-)

arkodg (approved):
LGTM thanks !

PR #2853, opened 2024-03-09
e2e: backend TLS policy

### LINKED ISSUE #2772: e2e: BackendTLSPolicy
*Description*:
>Describe the issue.

Relates to https://github.com/envoyproxy/gateway/issues/2033

[optional *Relevant Links*:]

guydc:

Simple E2E test for `BackendTLSPolicy`: uses the gateway-api echo server with TLS as backend. 

Fixes #2772


test/e2e/e2e_test.go
```diff
@@ -54,6 +56,14 @@ func TestE2E(t *testing.T) {
 	})
 
 	cSuite.Setup(t)
+	egSetup(t, cSuite)
 	t.Logf("Running %d E2E tests", len(tests.ConformanceTests))
 	cSuite.Run(t, tests.ConformanceTests)
 }
+
+// set up additional resources that are created and cleaned up programmatically like certificates
+func egSetup(t *testing.T, testSuite *suite.ConformanceTestSuite) {
```
shawnh2, line 65: can this be a better name?

test/e2e/tests/backend_tls.go
```diff
@@ -0,0 +1,48 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, BackendTLSTest)
+}
+
+var BackendTLSTest = suite.ConformanceTest{
```
zhaohuabing, line 24: Should we add a test to HTTP get an httproute without TLSPolicy and expect an error response?

test/e2e/e2e_test.go
```diff
@@ -54,6 +56,14 @@ func TestE2E(t *testing.T) {
 	})
 
 	cSuite.Setup(t)
+	addAdditionalResourcesToSuite(t, cSuite)
 	t.Logf("Running %d E2E tests", len(tests.ConformanceTests))
 	cSuite.Run(t, tests.ConformanceTests)
 }
+
+// set up additional resources that are created and cleaned up programmatically like certificates
+func addAdditionalResourcesToSuite(t *testing.T, testSuite *suite.ConformanceTestSuite) {
```
zhaohuabing, line 65: It's not a strong opinion, could we just put the ConfigMap and Secrets in the yaml files for BackendTLSPolicy tests? They're not used elsewhere.

zirain: at least, add a TODO

guydc: I was trying to follow the patterns in upstream GW-API for these sorts of test (e.g. TLS Route). My only concern is that these certs will expire at some point and break tests, so I prefer to create them programatically...

test/e2e/tests/backend_tls.go
guydc: yes, done

test/e2e/e2e_test.go
zirain: address this tought `My only concern is that these certs will expire at some point and break tests, so I prefer to create them programatically..`

guydc: Created a static 100-year CA and leaf cert. I expect that these test will be redundant anyway once GW-API creates conformance for that. 
Do we have github secret scanner that needs to be update with an exclude?

zhaohuabing (approved):
LGTM. Thanks!

zirain (approved):

PR #2857, opened 2024-03-10
docs: move Design docs under "Get Involved"

arkodg:
* design docs are more relevant to internal contributors than external users


zirain (approved):

Xunzhuo (commented):
should check if there are relative path changes

zirain:
> should check if there are relative path changes

Let me check if there's any tool help with dead link, I recall istio had that thing, not sure it works well here.

zhaohuabing (approved):

PR #2859, opened 2024-03-10
feat: add install and uninstall sub cmd to egctl

### LINKED ISSUE #2711: feat: add `egctl install` and `egctl uninstall`
*Description*:

Have we considered adding more installation methods for **eg**?

Currently we support installation through `k8s resource file` and `helm`. 

In my opinion, we can add `egctl install` command line method for installation and `egctl uninstall` method for uninstallation.

This can reduce other issues that users need to pay attention to, and reduces the installation steps compared to the other two methods. This is more friendly to users who are less familiar with k8s/helm.

ShyunnY:
**What type of PR is this?**

feat: Add install and uninstall subcommands to egctl

Fixes #2711

### DIFF
--- a/internal/cmd/egctl/experimental.go
+++ b/internal/cmd/egctl/experimental.go
@@ -26,6 +26,8 @@ func newExperimentalCommand() *cobra.Command {
 	experimentalCommand.AddCommand(newStatsCommand())
 	experimentalCommand.AddCommand(newStatusCommand())
 	experimentalCommand.AddCommand(newDashboardCommand())
+	experimentalCommand.AddCommand(newInstallCommand())
+	experimentalCommand.AddCommand(newUnInstallCommand())
 
 	return experimentalCommand
 }
--- a/internal/utils/helm/package.go
+++ b/internal/utils/helm/package.go
@@ -1,0 +1,489 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package helm
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/getter"
+	"helm.sh/helm/v3/pkg/registry"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/kubectl/pkg/cmd/util"
+
+	"github.com/envoyproxy/gateway/internal/cmd/options"
+)
+
+const (
+	helmOperateTimeout = time.Second * 300
+	egChartVersion     = "v1.0.0"
+	egReleaseName      = "envoy-gateway"
+	egReleaseNamespace = "envoy-gateway-system"
+)
+
+type PackageOptions struct {
+	DryRun           bool
+	SkipCRD          bool
+	Wait             bool
+	Version          string
+	Timeout          time.Duration
+	ReleaseName      string
+	ReleaseNamespace string
+	OnlyCRD          bool
+	WithCRD          bool
+}
+
+type PackageTool struct {
+	chartName string
+
+	// Helm dependency objects
+	envSettings     *cli.EnvSettings
+	actionConfig    *action.Configuration
+	actionInstall   *action.Install
+	actionUninstall *action.Uninstall
+	valuesOpts      *values.Options
+
+	logger Printer
+}
+
+func NewPackageTool() *PackageTool {
+	return &PackageTool{
+		envSettings:  cli.New(),
+		actionConfig: &action.Configuration{},
+		chartName:    "oci://docker.io/envoyproxy/gateway-helm",
+		valuesOpts:   &values.Options{},
+	}
+}
+
+// Setup Configuration required to initialize helm action.
+func (pt *PackageTool) Setup() error {
+
+	// Since envoy-gateway uses docker's oci to store charts,
+	// we need to create a registry client to make sure we can retrieve envoy-gateway chart
+	registryCli, err := registry.NewClient()
+	if err != nil {
+		return err
+	}
+	pt.actionConfig = &action.Configuration{
+		RegistryClient: registryCli,
+	}
+
+	kubectlFactory := options.DefaultConfigFlags
+	ns, _, err := kubectlFactory.ToRawKubeConfigLoader().Namespace()
+	if err != nil {
+		return err
+	}
+
+	if err = pt.actionConfig.Init(
+		kubectlFactory,
+		ns,
+		os.Getenv("HELM_DRIVER"),
+		pt.logger.Printf,
+	); err != nil {
+		return err
+	}
+
+	// Build the relevant helm command action
+	pt.actionInstall = action.NewInstall(pt.actionConfig)
+	pt.actionUninstall = action.NewUninstall(pt.actionConfig)
+
+	return nil
+}
+
+// SetInstallEnvSettings set the installation flags we are interested in
+func (pt *PackageTool) SetInstallEnvSettings(installCmd *cobra.Command, opts *PackageOptions) {
+
+	// add helm flags
+	// we use a temporary flag to be set by helm env flags,
+	// from which we can retrieve the flags we are interested
+	var tmpFlags pflag.FlagSet
+	pt.envSettings.AddFlags(&tmpFlags)
+	tmpFlags.VisitAll(func(flag *pflag.Flag) {
+		// TODO: Add more flags as needed?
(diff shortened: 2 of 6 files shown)


arkodg:
@alexwo meet @ShyunnY , @ShyunnY meet @alexwo :)
looks like you both are trying to create an internal helm lib to install and uninstall EG https://github.com/envoyproxy/gateway/pull/2859
would be great if the lib can be reused between `egctl` and `e2e`

ShyunnY:
> @alexwo meet @ShyunnY , @ShyunnY meet @alexwo :) looks like you both are trying to create an internal helm lib to install and uninstall EG #2859 would be great if the lib can be reused between `egctl` and `e2e`

This looks good and I think a more general option could be made using the HelmTool I wrote. :)

ShyunnY:
@arkodg Hello! I migrated HelmTool to the `/utils/helm` package and renamed it to PackageTool. Consider this naming because it bears the package management responsibilities of `eg` (e.g.  installation, uninstallation, and upgrades that may be added in the future, etc.). The use of the Helm library may not be limited to `egctl`. As you said, I pulled it out as a tool that others can use.

arkodg:
cc @alexwo

zirain:
should we put these under `experimetal` first?

ShyunnY:
This is a good idea.

internal/utils/helm/package.go
```diff
@@ -0,0 +1,473 @@
...
+			return err
+		}
+
+		if err := installCRDs(crdInfo, ht.actionConfig); err != nil {
+			return err
+		}
+
+		if exist, err := detectExistCRDs(crdInfo); exist == nil || !*exist {
+			if err != nil {
+				return fmt.Errorf("faile to install CRDs of envoy gateway")
```
shawnh2, line 240: typo `failed`

internal/utils/helm/package.go
```diff
@@ -0,0 +1,473 @@
...
+	return nil
+}
+
+// setCommonValue Set the common values needed for the installation chart.
+// We are not currently considering allowing users to define the following values
+func (ht *PackageTool) setCommonValue() {
+	ht.actionInstall.CreateNamespace = true
+	ht.actionInstall.GenerateName = false
+	ht.actionInstall.Description = "envoy gateway was installed using the egctl"
+	ht.actionInstall.Namespace = "envoy-gateway-system"
```
shawnh2, line 315: this is a fixed value now, imo users should be able to install EG under any namespace, this can be beneficial to EG's [multi-tenancy deployment mode](https://gateway.envoyproxy.io/v1.0.0/user/operations/deployment-mode/#multi-tenancy).

zirain: should be configurable via `--namspace` (`-n` in short)

internal/cmd/egctl/experimental.go
```diff
@@ -26,6 +26,8 @@ func newExperimentalCommand() *cobra.Command {
 	experimentalCommand.AddCommand(newStatsCommand())
 	experimentalCommand.AddCommand(newStatusCommand())
 	experimentalCommand.AddCommand(newDashboardCommand())
+	experimentalCommand.AddCommand(newInstallCommand())
```
shawnh2, line 29: hi @ShyunnY, the code looks good! 

since we have decided to set these two subcmds as experimental, can we add a section `egctl experimental install & uninstall` in https://gateway.envoyproxy.io/latest/user/operations/egctl/ to describe how to use these cmds.

and also, since we can install EG using egctl, maybe this doc https://gateway.envoyproxy.io/latest/install/install-egctl/ can be renamed as `install-with-egctl`, and how to install egctl will be a section in it. WDYT?

ShyunnY: IMO, I think https://gateway.envoyproxy.io/latest/install/install-egctl/ does not need to be changed, this part of the doc is meant to tell us how to install egctl, we are not supposed to add other information here. Currently install/uninstall is only in experimental status, so it is enough for us to just add description information at https://gateway.envoyproxy.io/latest/user/operations/egctl/.

When we later consider changing install/uninstall to stable, we will consider adding additional docs to specifically describe how to use egctl to install or uninstall eg.

I think this is more appropriate, what do you think? :)

shawnh2: SGTM

shawnh2 (approved):
LGTM! Also test against the [multi-tenancy mode](https://gateway.envoyproxy.io/v1.0.0/user/operations/deployment-mode/#multi-tenancy) and it works as expected. Thanks @ShyunnY 

There're some conflicts, can you resolve?

ShyunnY:
I may not be able to resolve this conflict at the moment, I may need help

ShyunnY:
@arkodg
It seems that I can't solve the go.mod problem. If I modify it according to lint, conflicts will occur. If I modify the conflicts, it will cause lint problems.

Also, do we really need to add tests?

arkodg:
sorry about the go mod rebase issues, will make sure to not merge any more go.mod PRs

arkodg:
instead of a test, lets raise a follow up to leverage this tool lib in our e2e

ShyunnY:
> instead of a test, lets raise a follow up to leverage this tool lib in our e2e

Since I couldn't resolve the `go.mod` conflict issue, I chose to undo the changes to `go.mod`, which may require troublesome troubleshooting after the merge, thank you again! 

After waiting for the code to be merged, I will follow up on the writing of related e2e.

internal/utils/helm/package.go
```diff
@@ -0,0 +1,493 @@
...
+		default:
+		}
+	})
+
+	installCmd.Flags().DurationVar(&opts.Timeout, "timeout", helmOperateTimeout, "time to wait for any individual Kubernetes operation")
+	installCmd.Flags().StringVar(&opts.Version, "version", egChartVersion, "specify a version constraint for the envoy gateway version to use")
+	installCmd.Flags().StringVar(&opts.ReleaseName, "release-name", egReleaseName, "name of the envoy-gateway release to install")
+	installCmd.Flags().StringVarP(&opts.ReleaseNamespace, "namespace", "n", egReleaseNamespace, "if set, specify the namespace where envoy gateway is installed")
+	installCmd.Flags().BoolVar(&opts.DryRun, "dry-run", false, "console output only, make no changes")
+	installCmd.Flags().BoolVar(&opts.SkipCRD, "skip-crd", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
```
arkodg, line 134:
```suggestion
	installCmd.Flags().BoolVar(&opts.SkipCRD, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
```

internal/utils/helm/package.go
```diff
@@ -0,0 +1,493 @@
...
+	installCmd.Flags().BoolVar(&opts.OnlyCRD, "only-crd", false, "if set, only install the crd")
```
arkodg, line 135:
```suggestion
	installCmd.Flags().BoolVar(&opts.OnlyCRD, "only-crds", false, "if set, only install the crd")
```

internal/utils/helm/package.go
```diff
@@ -0,0 +1,493 @@
...
+
+// SetUninstallEnvSetting set the uninstallation flags we are interested in
+func (pt *PackageTool) SetUninstallEnvSetting(uninstallCmd *cobra.Command, opts *PackageOptions) {
+
+	uninstallCmd.Flags().DurationVar(&opts.Timeout, "timeout", helmOperateTimeout, "time to wait for any individual Kubernetes operation")
+	uninstallCmd.Flags().StringVar(&opts.ReleaseName, "release-name", egReleaseName, "name of the envoy-gateway release to uninstall")
+	uninstallCmd.Flags().StringVarP(&opts.ReleaseNamespace, "namespace", "n", "", "if set, specify the namespace where envoy gateway is uninstalled")
+	uninstallCmd.Flags().BoolVar(&opts.Wait, "wait", false, "if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout")
+	uninstallCmd.Flags().BoolVar(&opts.DryRun, "dry-run", false, "console output only, make no changes")
+	uninstallCmd.Flags().BoolVar(&opts.WithCRD, "with-crd", false, "if set, the CRDs will also be removed")
```
arkodg, line 151:
```suggestion
	uninstallCmd.Flags().BoolVar(&opts.WithCRD, "with-crds", false, "if set, the CRDs will also be removed")
```

internal/utils/helm/package.go
```diff
@@ -0,0 +1,493 @@
...
+		switch flag.Name {
+		case "registry-config", "repository-config", "repository-cache":
+			installCmd.Flags().AddFlag(flag)
+		default:
+		}
+	})
+
+	installCmd.Flags().DurationVar(&opts.Timeout, "timeout", helmOperateTimeout, "time to wait for any individual Kubernetes operation")
+	installCmd.Flags().StringVar(&opts.Version, "version", egChartVersion, "specify a version constraint for the envoy gateway version to use")
+	installCmd.Flags().StringVar(&opts.ReleaseName, "release-name", egReleaseName, "name of the envoy-gateway release to install")
```
arkodg, line 131: maybe just `name` ?

arkodg:
minor comments around naming, thanks for building this out !

ShyunnY:
> minor comments around naming, thanks for building this out !

Thanks for your review! I will fix the test issue as soon as possible, thank you again!

internal/utils/helm/package.go
ShyunnY: > maybe just `name` ?

I have thought about this problem, but I think it is more suitable to use `release-name`.

+ it's self-explanatory
+ release-name does not seem to be easily ambiguous, at least when compared with name.

arkodg: we ideally like typing less when its comes to CLI commands, and the flag description does mention that this is related to release.
Lets get this in, and get feedback, since its experimental, we can always improve this later

arkodg (approved):
LGTM thanks !

Issue #1713, opened 2023-07-26
Handle EnvoyGateway control plane version upgrades


arkodg:

guydc:
As discussed in last community meeting with @arkodg and @zhaohuabing  - achieving a a hitless upgrade in this case may require #2694  to be merged. If the control plane is down during an upgrade and a new envoy pod is started, the proxy will not be programmed and drop all the traffic that reaches it. 

So, we can approach this incrementally:
1. A basic test that EG works after the upgrade,
2. Upgrade is hitless when there is traffic, e.g. 5k RPS

alexwo:
I would like to contribute. Could you please assign this task to me?

guydc:
closed by #2862

PR #2862, opened 2024-03-10
[e2e] eg release upgrade test

alexwo:
Tests eg upgrades - ensuring zero-downtime during the upgrade.

We want to ensure that the complete helm deployment is upgradable (Including helm template changes).

Direction:

1. Begin with the removal of the existing EnvoyGateway version using Helm uninstall.
2. Proceed to install the most recent stable release, or an alternate specified version if determined by the `last_version_tag` environment variable.
3. Introduce a minimal traffic load to ensure uninterrupted routing to an already deployed backend.
4. Update the EnvoyGateway to the latest main version.
5. Validate that both the upgrade and service connectivity were seamlessly maintained without disruptions.

Which issue(s) this PR fixes:
https://github.com/envoyproxy/gateway/issues/1713


test/e2e/tests/eg_upgrade.go
```diff
@@ -0,0 +1,178 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, BackendUpgradeTest)
+}
+
+var EGUpgradeTest = suite.ConformanceTest{
+	ShortName:   "EGUpgrade",
+	Description: "Upgrading from the lat eg version should not lead to failures",
```
zirain, line 33:
```suggestion
	Description: "Upgrading from the last eg version should not lead to failures",
```

test/e2e/tests/eg_upgrade.go
```diff
@@ -0,0 +1,178 @@
...
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/registry"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, BackendUpgradeTest)
```
zirain, line 28:
```suggestion
	ConformanceTests = append(ConformanceTests, EGUpgradeTest)
```

zirain:
I'm thinking should use a seprated test suite?

alexwo:
> > I'm thinking should use a seprated test suite?
> 
> 

Please share your feedback on the overall approach :) 

 If possible lets execute the tests at least once via the Continuous Integration system, as I'm uncertain about the completeness of local testing. If everything functions as expected, we can then move forward with restructuring this into a dedicated test suite

alexwo:
> > I'm thinking should use a seprated test suite?
> 
> Probably a good idea, as this may in the future depend on features like graceful drain that we will be enabled with a dedicated EnvoyProxy, as discussed #2839. Also, the install/uninstall/upgrade may be have side effects to other tests if this test is not executed in isolation after the main suite.

Please share your feedback on the overall approach.

If possible lets execute the tests at least once via the Continuous Integration system, as I'm uncertain about the completeness of local testing. If everything functions as expected, we can then move forward with restructuring this into a dedicated test suite

guydc:
Sorry Alex, didn't notice that the CI was off for this. Sure, I enabled it now.

test/e2e/tests/eg_upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+	ShortName:   "EGUpgrade",
+	Description: "Upgrading from the last eg version should not lead to failures",
+	Manifests:   []string{"testdata/eg-upgrade.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		t.Run("Upgrade from an older eg release should succeed", func(t *testing.T) {
+			relName := "eg"
+			depNS := "envoy-gateway-system"
+			lastVersionTag := os.Getenv("last_version_tag")
+			if lastVersionTag == "" {
+				lastVersionTag = "v0.0.0-latest" // Default version tag if not specified
```
arkodg, line 37: lets make this `v0.6.0`

test/e2e/tests/eg_upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+
+	registryClient, err := registry.NewClient()
+	if err != nil {
+		return err
+	}
+	install.SetRegistryClient(registryClient)
+	chartPath, err := install.LocateChart("oci://docker.io/envoyproxy/gateway-helm", cli.New())
+	if err != nil {
+		return err
+	}
```
arkodg, line 141: lets add a TODO comment in here saying we need to explicitly reinstall CRDs, and raise a separate GH issue to work on that as a follow up

test/e2e/tests/eg_upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+
+			// wait for everything to startup
+			kubernetes.NamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, namespaces)
+			loadSuccess := make(chan bool)
+			// can be used to abort the test after deployment restart is complete or failed
+			aborter := periodic.NewAborter()
+			t.Log("Starting load generation", "reqURL:", reqURL.String())
+
+			// Run load async and continue to restart deployment
+			go runLoadAndWait(t, suite.TimeoutConfig, loadSuccess, aborter, reqURL.String())
```
arkodg, line 75: so we're only checking if the traffic + route works post install ? can we also check if it works before ? (probably obvious, but helps the next dev debug this code better in case they are debugging upgrade failures)

arkodg:
@alexwo  the code looks good added some minor comments, thanks for getting this in into v1.0 !

alexwo:
> @alexwo the code looks good added some minor comments, thanks for getting this in into v1.0 !

Absolutely! You're very welcome! :)

ShyunnY:
This looks good, there are many similarities between the PRs you and I wrote about the Helm library, maybe we need a way to make this part of the code reusable in `egctl`/`e2e`!  ❤️ 
@alexwo

tools/make/kube.mk
```diff
@@ -121,7 +121,8 @@ run-e2e: install-e2e-telemetry
 	kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
 	kubectl apply -f test/config/gatewayclass.yaml
 ifeq ($(E2E_RUN_TEST),)
-	go test -v -tags e2e ./test/e2e --gateway-class=envoy-gateway --debug=true --cleanup-base-resources=$(E2E_CLEANUP)
+	go test -v -tags e2e ./test/e2e --gateway-class=envoy-gateway --debug=true
+	go test -v -tags e2e ./test/e2e/upgrade --gateway-class=envoy-gateway --debug=true --cleanup-base-resources=$(E2E_CLEANUP)
```
zirain, line 125: should also add smiliar line on L128?

alexwo: yes indeed, so that we have an option to run just it.

alexwo:
> This looks good, there are many similarities between the PRs you and I wrote about the Helm library, maybe we need a way to make this part of the code reusable in `egctl`/`e2e`! ❤️ @alexwo

Sure, I can check if we can maybe make it reusable.

ShyunnY:
> > This looks good, there are many similarities between the PRs you and I wrote about the Helm library, maybe we need a way to make this part of the code reusable in `egctl`/`e2e`! ❤️ @alexwo
> 
> Sure, I can check if we can maybe make it reusable.



> > This looks good, there are many similarities between the PRs you and I wrote about the Helm library, maybe we need a way to make this part of the code reusable in `egctl`/`e2e`! ❤️ @alexwo
> 
> Sure, I can check if we can maybe make it reusable.

Grateful! It looks like the `egctl install`/`egctl uninstall` I wrote relies more heavily on the Helm library part, maybe I can extract it into the utils package so that it can share functionality!

test/e2e/tests/eg_upgrade.go
alexwo: We can't do a hitless upgrade from v0.6.0, it results in connectivity issue as per:
upgrade completed after 12.14846937s with 41572 requests, 40555 success, 1017 failures and 1017 errors
[7735](https://github.com/envoyproxy/gateway/actions/runs/8235529982/job/22520849948#step:6:7736)

We can do v1.0.0-rc.1 , WDYT?

arkodg: do we know why it failed, this sounds like a release blocker

arkodg: or we can track this as a follow up issue, to unblock this PR

arkodg: also rereading your description, it sounds like upgrade passed, but hitless upgrade failed :) 
I think thats fine for v0.6.0 to `latest`, we can make sure `v1.0.0` to `latest` is hitless
any way you can tweak this PR to make sure some requests succeeded eventually, rather than make sure they are no failures ?

alexwo: yes, maybe we just keep v0.6.0 and i will tweak it to do the requests before the upgrade and after the upgrade. This should always work.

arkodg: awesome thanks !

arkodg (approved):
LGTM thanks for adding this !

guydc (approved):

PR #2863, opened 2024-03-10
e2e: add weighted backend 

### LINKED ISSUE #2238: Add e2e test for weighted backend
*Description*:
would be used.

Set up e2e test for weighted backend, prevent https://github.com/envoyproxy/gateway/pull/2187

ShyunnY:
**What type of PR is this?**

e2e: add backend weighted

Fixes #2238


test/e2e/tests/weighted_backend.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// CalculateWeight Calculate whether the expected weight is equal to the weight of the expected traffic
+func CalculateWeight(weight1, weight2 int, actualMap map[string]int) bool {
+
+	var count int
+	for _, v := range actualMap {
+		count += v
+	}
+	expectedV1 := float64(weight1) / float64(weight1+weight2) * float64(count)
```
zirain, line 208: I think we should use `AlmostEquals` here, 10 requests on `1:1`, may result `4:6` in real world.

arkodg:
@ShyunnY can you fix the lint issue, which is stopping the e2e from running

test/e2e/tests/weighted_backend.go
ShyunnY: Yes, this problem does occur. The actual traffic can only be infinitely close to the expected weight, but is not always guaranteed to be equal. (This is related to the cardinality of our request). Do you have a better solution? grateful!

ShyunnY:
> @ShyunnY can you fix the lint issue, which is stopping the e2e from running

Yes, I have solved it

test/e2e/tests/weighted_backend.go
ShyunnY: One solution is to increase the request base to 100 and give a deviation tolerance when calling the weight calculation.

zirain: try istio's way?
```golang
func AlmostEquals(a, b, precision int) bool {
	upper := a + precision
	lower := a - precision
	if b < lower || b > upper {
		return false
	}
	return true
}
```

ShyunnY: It seems that istio's idea is very good, I will choose this method! 
Gratitude!

test/e2e/tests/weighted_backend.go
```diff
@@ -0,0 +1,228 @@
...
+				} else {
+					// all we need is the pod Name prefix
+					podNamePrefix := ExtractPodNamePrefix(podName)
+					weightMap[podNamePrefix]++
+				}
+			}
+
+			for prefix, actual := range weightMap {
+				expect := expected[prefix]
+				if !AlmostEquals(actual, expect, 3) {
```
zirain, line 83: can you leave some comment here for people came here to understand

zirain (approved):
LGTM, thanks!

Xunzhuo (approved):

PR #2519, opened 2024-01-28
add e2e test for http3 & fix ci error

zesiar0:
**What type of PR is this?**
* feat: add e2e test

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #2423

PR #2876, opened 2024-03-11
fix: Don't override the ALPN array if HTTP/3 is enabled.

### LINKED ISSUE #2875: Enabling HTTP3 removes `h2` from TLS ALPN
*Description*:
Enabling HTTP3 in a `ClientTrafficPolicy` disables the `spec.tls.alpn` field, and sets ALPN in such a way that it advertises `h3` but not `h2`

*Repro steps*:

Create a `ClientTrafficPolicy` with the `http3` field set

## Expected result
```shell_session
$ openssl s_client -alpn h2 -connect [gateway ip]:443
...
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 2048 bit
Compression: NONE
Expansion: NONE
ALPN protocol: h2
Early data was not sent
Verify return code: 0 (ok)
```

## Actual result
```shell_session
$ openssl s_client -alpn h2 -connect [gateway ip]:443
...
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 2048 bit
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
```

*Environment*:
Gateway: `docker.io/envoyproxy/gateway-dev@sha256:cb9ff748dc8ea9c54e0cd5541658bff1aa3b8d68a945d7486561e4c593579f8c`
Proxy: `docker.io/envoyproxy/envoy@sha256:0d9cd17db2674550f06625acd5f347dee8bcdf59a812ebfd741202ce179b253b`

liorokman:
According to this [documentation](https://http3-explained.haxx.se/en/h3/h3-https#initial-connection), even when HTTP/3 is available, the first connection to a server would still be made using TCP and not QUIC. 

Since the `https` schema is reused for HTTP/3 URLs, the client can't know in advance that it should use QUIC, or even the port to use. The first request would not use HTTP/3, it would use HTTP/1.1 or HTTP/2 - as negotiated with ALPN.

Envoy Gateway should not override the ALPN string returned over the TCP listener with `h3`, since this prevents HTTP/2 from being used when it is available.

```mermaid
sequenceDiagram
   actor User
   participant eg as Envoy Gateway

   User ->> eg: Send first request with HTTP/2 over TCP
   eg->>User: Respond, add 'alt-svc' header
   User->>eg: Send next request with HTTP/3 over QUIC

```

Fixes #2875


zirain:
we need another positive/negative test case?

byhemechi (changes requested):
This commit doesn't remove the config validations for the behaviour changes

the ones i've found:


https://github.com/liorokman/gateway/blob/43733e79096e4e723d18ebe1cea38c446d429473/api/v1alpha1/clienttrafficpolicy_types.go#L37


https://github.com/liorokman/gateway/blob/43733e79096e4e723d18ebe1cea38c446d429473/test/cel-validation/clienttrafficpolicy_test.go#L206-L228

byhemechi:
I've removed/adjusted these respectively in my fork: 

https://github.com/byhemechi/envoy-gateway/commit/e2e3ba138f347c47c0829e233f58009e1b2a1136

byhemechi:
ahh whoops didnt redo the generated files

guydc (approved):
LGTM based on attached specs. 

Seems like the HTTP3 feature didn't work in a pretty common scenario. @envoyproxy/gateway-maintainers  - Should we add https://github.com/envoyproxy/gateway/pull/2519/files to GA scope?

byhemechi:
Probably more relevant to the e2e issue but i'd just like to mention that h3 itself worked completely fine if you connected to it directly (i.e. `curl --http3`), the issue was with browsers not respecting an h3 Alt-Svc from h/1.1. I'd imagine properly testing would take something like a full and proper headless chrome

PR #2877, opened 2024-03-11
api: wasm extension

zhaohuabing:
This PR proposes an API for supporting Wasm plugins:

```yaml
kind: EnvoyExtensionPolicy
metadata:
  name: wasm-extensions
spec:
  targetRef:
    kind: HTTPRoute
    name: backend
  wasm:
    - name: my_plugin_1
      code:
        image:  # the oci image that contains wasm code
          url: <oci-registry>.foo.bar.com/wasm-to-oci:v1 
          pullSecret: wasmPullSecret
        # http: https://foo.bar.com/example.wasm  # the HTTP URL containing the wasm code.
      sha256: xxxxxxx # SHA256 checksum that will be used to verify the wasm code.
      failOpen: true
      config:
        foo:
           foo1: value1
           foo2: value2
        bar: value3
      priority: 450   
    - name: my_plugin_2
      ...
```


api/v1alpha1/wasmextensionpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// Extensions is a list of Wasm extensions to be loaded by the Gateway.
+	// Order matters, as the extensions will be loaded in the order they are
+	// defined in this list.
+	//
+	// +kubebuilder:validation:MinItems=1
+	Extensions []WasmExtension `json:"extensions"`
+}
+
+// WasmExtension defines an wasm extension.
+type WasmExtension struct {
```
Xunzhuo, line 58: Should we add `fail_open` and `capability_restriction_config` on WasmExtension ?

guydc:
What persona will be using this API? IMHO it's a bit dangerous to let application owners inject code into envoy, even if executed in isolation. If we see this as a policy for Gateway admins, should we support attachment to routes?

zhaohuabing:
My apologies,  I haven't noticed the existing discussion and the draft PR and didn't mean to add a competing PR.  :-)

Before proposing the API, I considered two approaches to support data plane extensions (WASM, Lua, Golang, External Process, etc.
* A fat EnvoyExtensionPolicy containing all the extensions.
* A set of EnvoyXExtensionPolicy, each supporting a specific extension.

The main problem of the first approach is that it's hard to support multiple extensions of the same type, e.g. users may need to define multiple WASM filters for different purposes for a single HTTPRoute.

api/v1alpha1/wasmextensionpolicy_types.go
zhaohuabing: Yes, they can be added later. I just want to make the first API as simple as possible.

guydc:
> My apologies @guydc , I haven't noticed the existing discussion and the draft PR and didn't mean to add a competing PR. :-)

All good, I'm very happy that this topic is getting more attention and I hope that we can make this happen for v1.1.0. 

> Before proposing the API, I considered two approaches to support data plane extensions (WASM, Lua, Golang, External Process, etc.)

I agree that different extensions do have different structures, interfaces, concerns and use patterns. Some support per-route config while other don't, some are more modular and may repeat multiple time in the FC, while others are external and more likely to only be called-out once in the filter chain. 

I'm very much open to different extension types being supported for different use cases if other maintainers also support this approach.

zhaohuabing:
> I agree that different extensions do have different structures, interfaces, concerns and use patterns. Some support per-route config while other don't, some are more modular and may repeat multiple time in the FC, while others are external and more likely to only be called-out once in the filter chain.

> I'm very much open to different extension types being supported for different use cases if other maintainers also support this approach.

Let's bring this to the EG meeting and discuss further.  Even we option to have multiple extension types, these types do need to address the common concerns raised in https://github.com/envoyproxy/gateway/pull/2570 to ensure cohensive approach.

* Users may need to control the logical phase of filter execution (e.g. before auth, before rate limit, ... )
* Users may need to control the order of execution of extension filters amongst themselves (within a specific phase)

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// Code is the wasm code for the extension.
+	Code WasmCodeSource `json:"code"`
+
+	// Configuration for the wasm code.
+	Config *WasmConfig `json:"config"`
+}
+
+// WasmConfig defines the configuration for the wasm code.
+type WasmConfig struct {
```
zhaohuabing, line 36: Is there a better way to represent protobuf any.Any in the API?

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,66 @@
...
+	// Code is the wasm code for the extension.
+	Code WasmCodeSource `json:"code"`
+
+	// Configuration for the wasm code.
+	Config *WasmConfig `json:"config"`
+
+	// InsertBefore is the name of the filter that this Wasm extension should be
+	// inserted before.
+	// If not specified, this Wasm extension will be inserted before the router filter.
+	InsertBefore *EnvoyFilter `json:"insertBeforeFilter"`
```
zhaohuabing, line 37: This depends om https://github.com/envoyproxy/gateway/pull/2993

arkodg: here is my suggestion for prioritization of dealing with filter ordering to so we move faster in the API reviews
cc @guydc 
1. lets find a default home (order) for `ext_proc` and `wasm` (maybe right after `ext_auth`)
2. lets create per proxy ordering (which you have already done with #2993 (love it)) once we get use cases in
3. lets then consider creating per wasm location override once we get use cases in

zhaohuabing: @arkodg We may need to address ordering first because different WASM extensions should be in different locations in the chain.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,66 @@
...
+	// RootContext and Contexts if applicable (e.g., an Wasm HttpFilter and an Wasm AccessLog).
+	// If left blank, all extensions with a blank root_id with the same vm_id will share Context(s).
+	RootID *string `json:"rootID,omitempty"`
+
+	// Code is the wasm code for the extension.
+	Code WasmCodeSource `json:"code"`
+
+	// Config is the configuration for the Wasm extension.
+	// This configuration will be passed as a JSON string to the Wasm extension.
+	Config *apiextensionsv1.JSON `json:"config"`
```
zhaohuabing, line 37: Envoy supports struct, string, and bytes. Follow Istio practice to restrict input as a JSON string for a better UI:

* We can avoid media disruption in the K8s resources by just specifying the config as a YAML object. JSON is a subset of YAML, so it is very easy to convert to it from YAML.
* JSON has parsers in almost every language (as opposed to e.g. YAML), and certainly the ones that already compile to Wasm. It is also compatible with protobuf, should we want to use that in some of our SDKs.
* Developers don’t have to deal with parsing configuration. The language SDKs can use optimized code to unmarshall the JSON configuration and make the values easily accessible to developers in a language-specific way (this could e.g. be a map[string]interface{} in golang).
* Using a structured format for configuration allows for configuration injection and templating


https://docs.google.com/document/d/15GQnhOpLfvgyOZ8VDF5KD-q2cjXEMLdTQMm0YPswIu0/edit#heading=h.6zqg8n4aice5

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// +optional
+	HTTP *string `json:"http,omitempty"`
+
+	// SHA256 checksum that will be used to verify the wasm code.
+	// This field is required if the HTTP field is set.
+	// +optional
+	SHA256 *string `json:"sha256,omitempty"`
+
+	// Image is the OCI image containing the wasm code.
+	// Image *string `json:"image,omitempty"` //TODO: Add support for OCI image in the future.
```
zhaohuabing, line 83: OIC image is not supported by Envoy. So EG needs to download the wasm module from the OIC registry and mount it to the Envoy container.

zirain: there's no agent, we need to do this on controller?

zhaohuabing: IMO, EG doesn't need an agent like Istio does, since the Control Plane and Data Plane are in the same cluster.

zirain: envoy support http, EG can host an http endpoint?

arkodg:
hey can the PR description be used to reflect the newer API ?

api/v1alpha1/wasm_types.go
arkodg: a lot of the users already have their existing WASM module in a OCI registry so we should support this use case

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+// the root of the repo.
+
+package v1alpha1
+
+import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Wasm defines a wasm extension.
+type Wasm struct {
```
arkodg, line 13: can we autogenerate `name`, `vm_id` and `root_id` ? and leave it out of the first iteration ?

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Wasm defines a wasm extension.
```
arkodg, line 12: maybe some comments here around defaults, such as only `V8` type and `HTTP` filter based WASM modules are supported today

api/v1alpha1/wasm_types.go
zhaohuabing: We can generate the `name`, but the auto-generated name might not be as meaningful as the user-provided name. 

For `vm_id` and `root_id`, we might want to expose them to the API to allows users to share VM or RootContext.

zhaohuabing:
> hey can the PR description be used to reflect the newer API ?

Glad to see you back! @arkodg 

The description was modified.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+// WasmCodeSource defines the source of the wasm code.
+type WasmCodeSource struct {
+	// ConfigMap is the name of the ConfigMap containing the wasm code.
+	//
+	// The key in the ConfigMap should be the name of the Wasm. For example,
+	// if the Wasm is named "my-wasm-extension", the ConfigMap should have a key
+	// named "my-wasm-extension" and the value should be the wasm code.
+	// +optional
+	ConfigMap *string `json:"ConfigMap,omitempty"`
```
zhaohuabing, line 69: Should we support ConfigMap? Considering the size limit of CM(1M), it might not be very useful.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// will be inserted before the next filter found in the chain, if any. If no
+	// any other filters are found in the chain, this Wasm extension will be
+	// inserted before the router filter.
+	//
+	// If not specified, this Wasm extension will be inserted before the router filter.
+	// InsertBefore *EnvoyFilter `json:"insertBeforeFilter"`
+}
+
+// WasmCodeSource defines the source of the wasm code.
+type WasmCodeSource struct {
```
arkodg, line 62: should we start with only supporting 2 types - http and image ?
something like
```
type: "Image"
image:
  url:
  pullPolicy:
  pullSecret:
```

zhaohuabing: Yah, I also think CM is not so useful in production.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,84 @@
...
+	// If the specified filter is not found in the filter chain, this Wasm extension
+	// will be inserted before the next filter found in the chain, if any. If no
+	// any other filters are found in the chain, this Wasm extension will be
+	// inserted before the router filter.
+	//
+	// If not specified, this Wasm extension will be inserted before the router filter.
+	// InsertBefore *EnvoyFilter `json:"insertBeforeFilter"`
+}
+
+// WasmCodeSource defines the source of the wasm code.
```
arkodg, line 61: sidenote /non blocker - can code be refreshed w/o restarting Envoy ?

api/v1alpha1/wasm_types.go
arkodg: yeah we can tackle before v1.1 , but it doesnt need to block this PR

api/v1alpha1/wasm_types.go
zhaohuabing: I believe so. WASM filter is an HTTP filter in the HCM filter chain and the code is loaded on the fly.

zhaohuabing: I think I need to come up with a simple design doc before implementation.

zhaohuabing: > sidenote /non blocker - can code be refreshed w/o restarting Envoy ?

Seems that I misunderstood your comment :-) 

I guess we could change the sha256 to update the wasm code. I haven't verified that yet.

api/v1alpha1/wasm_types.go
zhaohuabing: @arkodg Envoy requires `name` to be unique within a VM. Since we have decided to leave vm_id out, EG will use a separate VM for each wasm extension, the name doesn't have to be unique in EnvoyExtensionPolicy.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,128 @@
...
+	// +optional
+	HTTP *HTTPWasmCodeSource `json:"http,omitempty"`
+
+	// Image is the OCI image containing the wasm code.
+	//
+	// Note that the image must be accessible from the Envoy Gateway.
+	// +optional
+	Image *ImageWasmCodeSource `json:"image,omitempty"`
+
+	// SHA256 checksum that will be used to verify the wasm code.
```
arkodg, line 80: lets rm SHA for now, until we decide how content pull will work ?

api/v1alpha1/wasm_types.go
```diff
@@ -79,7 +79,7 @@ type WasmCodeSource struct {
 
 	// SHA256 checksum that will be used to verify the wasm code.
 	// +optional
-	SHA256 *string `json:"sha256,omitempty"`
+	// SHA256 *string `json:"sha256,omitempty"`
```
zirain: should remove it instead of comment.

zhaohuabing: I leave this because it's likely to be used when implementing wasm extension.  

Of course, I'll remove it if I find out it's not necessary when implementing it.

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,128 @@
...
+	// +optional
+	// SHA256 *string `json:"sha256,omitempty"`
+}
+
+// WasmCodeSourceType specifies the types of RateLimiting.
+// +kubebuilder:validation:Enum=Global;Local
+type WasmCodeSourceType string
+
+const (
+	// HTTPWasmCodeSourceType allows the rate limits to be applied across all Envoy
```
arkodg, line 90: ratelimit ? :)

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,128 @@
...
+	URL string `json:"url"`
+
+	// PullSecret is a reference to the secret containing the credentials to pull the image.
+	PullSecret gwapiv1b1.SecretObjectReference `json:"pullSecret"`
+
+	// PullPolicy is the policy to use when pulling the image.
+	// If not specified, the default policy is IfNotPresent for images whose tag is not latest,
+	// and Always for images whose tag is latest.
+	// +optional
+	PullPolicy *PullPolicy `json:"pullPolicy,omitempty"`
```
arkodg, line 116: lets rm `pullPolicy` for now as well since its a k8s specific knob thats living in a generic API

api/v1alpha1/wasm_types.go
```diff
@@ -0,0 +1,128 @@
...
+	URL string `json:"url"`
+}
+
+// ImageWasmCodeSource defines the OCI image containing the wasm code.
+type ImageWasmCodeSource struct {
+	// URL is the URL of the OCI image.
+	URL string `json:"url"`
+
+	// PullSecret is a reference to the secret containing the credentials to pull the image.
+	PullSecret gwapiv1b1.SecretObjectReference `json:"pullSecret"`
```
arkodg, line 110: should this be pullSecretRef ?

api/v1alpha1/wasm_types.go
zhaohuabing: We probably can't reuse the k8s PullPolicy ?

```
const (
	// PullAlways means that kubelet always attempts to pull the latest image. Container will fail If the pull fails.
	PullAlways PullPolicy = "Always"
	// PullNever means that kubelet never pulls an image, but only uses a local image. Container will fail if the image isn't present
	PullNever PullPolicy = "Never"
	// PullIfNotPresent means that kubelet pulls if the image isn't present on disk. Container will fail if the image isn't present and the pull fails.
	PullIfNotPresent PullPolicy = "IfNotPresent"
)
```

zhaohuabing: Remove it for now.

arkodg (approved):
thanks !

zirain (approved):

arkodg:
hey @zhaohuabing lets start off with `HTTP` support, and temporarily pause on the design for the `Image` feature based on the discussions from https://github.com/envoyproxy/envoy/issues/33212

Issue #2269, opened 2023-12-06
conformance: flaky test GatewayModifyListeners


shawnh2:
https://github.com/envoyproxy/gateway/actions/runs/7111705161/job/19360397127#step:6:1111

shawnh2:
there're lots of failed reasons, but most of them are: `timeout: context deadline exceeded`.

some tests only failed occasionally and very difficult to debug and replay, hence the code should be fine, what may need to do is add a retry mechanism for the failed test case.

and also it seems upstream conformance does not support `retry` yet. maybe we can try to implement `retry` for EG first, to see how it works.

cc @envoyproxy/gateway-maintainers

arkodg:
@shawnh2 imo we need to figure out why its taking longer than 120s to compute and publish status

shawnh2:
GatewayModifyListeners & HTTPRouteInvalidCrossNamespaceParentRef, these two conformance tests seem to be very unstable recently, let me investigate this.

arkodg:
o/p from another run 
```
2024-04-16T07:30:22.7515182Z === RUN   TestGatewayAPIConformance/GatewayModifyListeners/should_be_able_to_remove_listeners,_which_would_then_stop_routing_the_relevant_traffic
2024-04-16T07:30:22.7589888Z     helpers.go:248: Gateways and Pods in gateway-conformance-infra namespaces ready
2024-04-16T07:30:22.7784581Z     helpers.go:217: Gateway gateway-conformance-infra/gateway-remove-listener expected observedGeneration to be updated to 2 for all conditions, only 0/2 were updated. stale conditions are: Accepted (generation 1), Programmed (generation 1)
2024-04-16T07:30:23.7784810Z     helpers.go:248: Gateways and Pods in gateway-conformance-infra namespaces ready
2024-04-16T07:30:23.7803677Z     helpers.go:655: Expected 1 Gateway status listeners, got 2
2024-04-16T07:30:24.7816499Z     helpers.go:655: Expected 1 Gateway status listeners, got 2
2024-04-16T07:30:25.7816031Z     helpers.go:655: Expected 1 Gateway status listeners, got 2
2024-04-16T07:30:26.7817410Z     helpers.go:655: Expected 1 Gateway status listeners, got 2
2024-04-16T07:30:27.7811108Z     helpers.go:655: Expected 1 Gateway status listeners, got 2
[...]

PR #2878, opened 2024-03-11
Feat: Support for loadBalancerSourceRanges on the envoy service

jaynis:
This PR adds support for `loadBalancerSourceRanges` on the envoy service. `loadBalancerSourceRanges` are used to specify firewall rules on the underlying load balancer of the kubernetes platform provider, if the provider supports this.


jaynis:
Failing tests seem to be related to #2269. Any idea how to overcome this?

arkodg:
@jaynis this API may make sense since the knob is specific to LB
to immediately unblock yourself, suggest using the new `patch` field
```
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  logging:
    level:
      default: warn
  provider:
    kubernetes:
      envoyService:
        patch:
          value:
            spec:
              loadBalancerSourceRanges: 1.0.0.0/8
    type: Kubernetes
```

api/v1alpha1/validation/envoyproxy_validate.go
```diff
@@ -115,6 +115,12 @@ func validateService(spec *egv1a1.EnvoyProxySpec) []error {
 				errs = append(errs, fmt.Errorf("allocateLoadBalancerNodePorts can only be set for %v type", egv1a1.ServiceTypeLoadBalancer))
 			}
 		}
+		if serviceType, serviceLoadBalancerSourceRanges :=
```
arkodg, line 118: can we add a IP Address check in here

arkodg:
hey @jaynis still working on this one ?

jaynis:
Sorry totally forgot about this one. What kind of IP address check do you have in mind there?

arkodg:
> Sorry totally forgot about this one. What kind of IP address check do you have in mind there?

something like https://github.com/envoyproxy/gateway/blob/d383ba54020a1bd38e2787eff230b055713f35e3/api/v1alpha1/validation/envoyproxy_validate.go#L123

jaynis:
I have added a CEL validation rule to check whether all `loadBalancerSourceRanges` have a valid CIDR notation. But now I am hitting some kind of schema validation rule cost budget. Limiting the maximum number of items, as suggested in the error message, does not help either.

Any suggestions how to proceed here @arkodg? Should I simply remove the CEL validation again?

arkodg:
> schema validation rule cost budget

yah should be fine for now, you've already added logic inside the validate method

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #2734, opened 2024-03-01
Change the `Merge` behavior to `Replace` for BackendTrafficPolicy and SecurityPolicy


zhaohuabing:
Change the `Merge` behavior to `Replace` for BackendTrafficPolicy and SecurityPolicy.

Replace by the most specific resource is the default https://gateway.envoyproxy.io/v0.6.0/design/security-policy/#design-decisions , and support for Merge will hopefully be tackled in the near future with https://github.com/envoyproxy/gateway/issues/1934

PR #2885, opened 2024-03-12
Change the Merge behavior to Replace for SecurityPolicy

zhaohuabing:
According to the [design decisions](https://gateway.envoyproxy.io/v0.6.0/design/security-policy/#design-decisions),  replacing with the most specific resource is the default behavior.

Part of https://github.com/envoyproxy/gateway/issues/2734


zirain (approved):

arkodg (approved):
thanks, same for BTP :) ?

PR #2887, opened 2024-03-12
bug: add h3 alpn by default if http3 is enabled

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/2875

### DIFF
--- a/internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
+++ b/internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
@@ -42,6 +42,7 @@
             alpnProtocols:
             - h2
             - http/1.1
+            - h3
             tlsCertificateSdsSecretConfigs:
             - name: envoy-gateway-tls-secret-1
               sdsConfig:
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -277,12 +277,12 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	if irListener.TLS != nil {
 		var tSocket *corev3.TransportSocket
 		if http3Listener {
-			tSocket, err = buildDownstreamQUICTransportSocket(irListener.TLS)
+			tSocket, err = buildDownstreamQUICTransportSocket(irListener.TLS, http3Listener)
 			if err != nil {
 				return err
 			}
 		} else {
-			tSocket, err = buildXdsDownstreamTLSSocket(irListener.TLS)
+			tSocket, err = buildXdsDownstreamTLSSocket(irListener.TLS, http3Listener)
 			if err != nil {
 				return err
 			}
@@ -388,7 +388,7 @@ func addXdsTCPFilterChain(xdsListener *listenerv3.Listener, irListener *ir.TCPLi
 	}
 
 	if isTLSTerminate {
-		tSocket, err := buildXdsDownstreamTLSSocket(irListener.TLS.Terminate)
+		tSocket, err := buildXdsDownstreamTLSSocket(irListener.TLS.Terminate, false)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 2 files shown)


internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
```diff
@@ -42,6 +42,7 @@
             alpnProtocols:
             - h2
             - http/1.1
+            - h3
```
zirain, line 45: should we sort this?

arkodg: I thought of it, would change every file ..... I'm fine if you are

liorokman:
I don't think this is needed.

Duplicating my response from the issue:

I took a look at how https://www.cloudflare.com/ works (redacted for readability):

```
$ openssl s_client -connect www.cloudflare.com:443 -alpn h3,http/1.1
SSL handshake has read 4713 bytes and written 418 bytes
Verification: OK
---
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 2048 bit
Compression: NONE
Expansion: NONE

ALPN protocol: http/1.1

Early data was not sent
Verify return code: 0 (ok)
---
HEAD / HTTP/1.1
Host: www.cloudflare.com 
---
HTTP/1.1 200 OK
Date: Tue, 12 Mar 2024 06:04:10 GMT
Content-Type: text/html; charset=utf-8

alt-svc: h3=":443"; ma=86400
```

You can see that I offer `h3` from the client as well as `http/1.1` (using openssl from the CLI, so no HTTP/2 for me), and the server rejects `h3` and accepts `http/1.1`. Later on, the `alt-svc` header is used to offer the `h3` protocol.

arkodg:
> I don't think this is needed.
> 
> Duplicating my response from the issue:
> 
> I took a look at how https://www.cloudflare.com/ works (redacted for readability):
> 
> ```
> $ openssl s_client -connect www.cloudflare.com:443 -alpn h3,http/1.1
> SSL handshake has read 4713 bytes and written 418 bytes
> Verification: OK
> ---
> New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
> Server public key is 2048 bit
> This TLS version forbids renegotiation.
> Compression: NONE
> Expansion: NONE
> 
> ALPN protocol: http/1.1
> 
> Early data was not sent
> Verify return code: 0 (ok)
> ---
> HEAD / HTTP/1.1
> Host: www.cloudflare.com 
> ---
> HTTP/1.1 200 OK
> Date: Tue, 12 Mar 2024 06:04:10 GMT
> Content-Type: text/html; charset=utf-8
> 
> alt-svc: h3=":443"; ma=86400
> ```
> 
> You can see that I offer `h3` from the client as well as `http/1.1` (using openssl from the CLI, so no HTTP/2 for me), and the server rejects `h3` and accepts `http/1.1`. Later on, the `alt-svc` header is used to offer the `h3` protocol.

its for the case where client starts off with `--http3-only` similar to the knob in `curl` https://curl.se/docs/http3.html

internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
zirain: we can do that in separated, let this one clean.

zirain (approved):

Xunzhuo (approved):

liorokman:
> its for the case where client starts off with `--http3-only` similar to the knob in `curl` https://curl.se/docs/http3.html

I took a network trace, and I think this PR went a little too far.

The `h3` protocol is required on the TLS context in the QUIC listener, and not in the regular TCP listener. When using `curl --http3`, curl only tries to negotiate `h3` on the QUIC based traffic.

arkodg:
@liorokman what about the `http3-only` case ?

liorokman:
When using `--http3-only` with `curl`, it means that `curl` will not try to open a TCP connection at all, but rather only use  QUIC over UDP. If there is no response on UDP, `curl` with `--http3-only` will simply fail. 

As opposed to `--http3`, where `curl` will fallback to HTTP/2 over TCP if it doesn't get an HTTP/3 response over UDP after a short delay (e.g. 100ms). 

In both cases, the TCP listener doesn't need to advertise `h3` in its ALPN protocol list in the TLS context, since by definition `h3` can't be used over TCP, and the QUIC listener doesn't need to advertise anything other than `h3` in its ALPN protocol list, since by definition only `h3` is possible over QUIC.

PR #2888, opened 2024-03-12
Change the Merge behavior to Replace for BackendTrafficPolicy

zhaohuabing:
Fix https://github.com/envoyproxy/gateway/issues/2734

### DIFF
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -457,7 +457,8 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
 			if r.RateLimit != nil || r.LoadBalancer != nil ||
 				r.ProxyProtocol != nil || r.HealthCheck != nil ||
 				r.CircuitBreaker != nil || r.FaultInjection != nil ||
-				r.TCPKeepalive != nil || r.Retry != nil {
+				r.TCPKeepalive != nil || r.Retry != nil ||
+				r.Timeout != nil {
 				continue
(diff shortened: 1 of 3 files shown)


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -448,7 +448,19 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
 		if t.MergeGateways && gatewayName != policyTarget {
 			continue
 		}
+
+		// A Policy targeting the most specific scope(xRoute) wins over a policy
+		// targeting a lesser specific scope(Gateway).
 		for _, r := range http.Routes {
+			// If any of the features are already set, it means that a more specific
+			// policy(targeting xRoute) has already set it, so we skip it.
+			if r.RateLimit != nil || r.LoadBalancer != nil ||
```
arkodg, line 457: `timeout`  is missing
how do we ensure this always stays in sync :) (good followup)

zhaohuabing: ir.HTTPRote has too many fields in it. We probably could add an intermediate layer and group all fields of an xPolicy into a structure. WDYT?

arkodg (approved):

internal/gatewayapi/backendtrafficpolicy.go
arkodg: yah makes sense, something like `TrafficFeatures *TrafficFeatures` and check for `nil`

guydc (approved):

PR #2902, opened 2024-03-13
skip publishing empty status for policies

arkodg:
* https://github.com/envoyproxy/gateway/pull/2802 skips computing status if a target resource cannot be found, mainly because that target maybe irrelevant to this specific translation, its hard to proactively find that out in the provider layer

* This fix ensures that any empty status is not published and resets any existing status for a policy

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -7,6 +7,7 @@ package runner
 
 import (
 	"context"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


zhaohuabing (approved):
Great job!

zhaohuabing (commented):
https://github.com/envoyproxy/gateway/blob/bb0a9a74abd7ef118246b2962ad0cac96ca31417/internal/provider/kubernetes/controller.go#L868

shawnh2 (approved):

zhaohuabing (approved):

PR #2907, opened 2024-03-13
fix: QUIC listeners should only advertise HTTP/3 over ALPN, and not HTTP/2 and HTTP/1.1

### LINKED ISSUE #2896: advertise h3 only on the QUIC listener
When using `--http3-only` with `curl`, it means that `curl` will not try to open a TCP connection at all, but rather only use  QUIC over UDP. If there is no response on UDP, `curl` with `--http3-only` will simply fail. 

As opposed to `--http3`, where `curl` will fallback to HTTP/2 over TCP if it doesn't get an HTTP/3 response over UDP after a short delay (e.g. 100ms). 

In both cases, the TCP listener doesn't need to advertise `h3` in its ALPN protocol list in the TLS context, since by definition `h3` can't be used over TCP, and the QUIC listener doesn't need to advertise anything other than `h3` in its ALPN protocol list, since by definition only `h3` is possible over QUIC.

_Originally posted by @liorokman in https://github.com/envoyproxy/gateway/issues/2887#issuecomment-1992356310_

liorokman:
Fixes #2896

### DIFF
--- a/internal/xds/translator/listener.go
+++ b/internal/xds/translator/listener.go
@@ -277,12 +277,12 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 	if irListener.TLS != nil {
 		var tSocket *corev3.TransportSocket
 		if http3Listener {
-			tSocket, err = buildDownstreamQUICTransportSocket(irListener.TLS, http3Listener)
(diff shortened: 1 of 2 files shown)


zirain (approved):

cnvergence (approved):

arkodg (approved):
thanks !

Issue #2810, opened 2024-03-06
Improve EG Gateway xDS & startup Reliability  (custom k8s health prob)


alexwo:
The proposed enhancement involves modifying the controller's "ready" status to accurately reflect the completion and synchronization of xds discovery processes. 

Specifically, the "ready" status indicator can transition to "true" when the xDS discovery has fully completed to store it's initial snapshot or when there is no reconciliation required.  (empty or new deployment).

This can ensure that envoy proxies are always in sync with latest xDS service and that an EG that has started is able to reconcile. 

-> If there is nothing to reconcile -> ready = true
-> if there are changes to reconcile, wait for xDS to complete -> ready = true
-> other wise -> ready = false

Currently, there may be certain cases where xDS is not completely synchronized at startup, which could cause new Envoy proxies to work with an incomplete xDS.

This can provide better guarantees that an operational EG consistently maintains an updated xDS, potentially also can allow avoiding situations where instances startup but fail during the initial reconcile.

Leader Election and multiple instances use case:
Will improve consistency in environments where multiple instances of EG run simultaneously by ensuring they start only once xDS server has persisted the latest state snapshot.  https://github.com/envoyproxy/gateway/issues/1953

arkodg:
makes sense, a workaround for this until this is implemented is to wake up slowly i.e. set `initialDelaySeconds` to a higher value

aoledk:
> makes sense, a workaround for this until this is implemented is to wake up slowly i.e. set `initialDelaySeconds` to a higher value

@arkodg But currently EG hasn't exposed `initialDelaySeconds` via Chart Values in https://github.com/envoyproxy/gateway/blob/78fe57ab58deb32b985a7ca6d8390194c4a5ef3b/charts/gateway-helm/templates/envoy-gateway-deployment.yaml#L66-L71

arkodg:
I was referring to `initialDelaySeconds` for envoy proxy (data plane) so it gets enough time to receive xds before receiving traffic (from external LB) once signaling its ready

aoledk:
Since EG is xDS resources provider and envoy is consumer, setting longer `initialDelaySeconds` for envoy can't prevent envoy from receiving incomplete xDS resources when EG is starting. This will lead to frequent envoy listener draining.

Workaround here maybe set longer `initialDelaySeconds` for EG to make sure envoy always consume complete xDS resources from EG.

arkodg:
[...]

PR #2918, opened 2024-03-13
feat(EG K8S Provider): Improve EG Gateway xDS & startup reliability

alexwo:
###  What type of PR is this?
feat(EG K8S Provider): Improve EG Gateway xDS & startup reliability

### What this PR does / why we need it:
- Service Readiness: By waiting for the xDS service to start up, we ensure that the service discovery mechanism is operational, which is crucial for dynamically managing service configurations and routing in microservices architectures.
- Stability and Reliability: This cautious approach prioritizes the stability and reliability of the system by preventing any unconfigured or misconfigured instances of Envoy from intercepting traffic, which could lead to service disruptions or degraded user experiences.
- Observability: Completing a successful reconcile cycle before startup can make it easier to monitor and debug the system, as it sets a clear baseline for the operational state of the Envoy instances.

###  Which issue(s) this PR fixes:
This enhancement facilitates the safer upgrades of EG gateways and features improved xDS consistency on startup.

###  High Level Changes
A custom health check is implemented to ensure that EG controller is deemed ready only once the xDS snapshot is persisted, or after at least one reconciliation is completed for empty or new clusters.
A custom watcher is added to initiate an initial dummy reconciliation, this ensures empty or news deployments can start (as a successful reconcile is a trigger for startup)

- Configuration Validation: Ensuring a successful reconcile cycle before startup means that the initial configuration passed to Envoy proxies is validated and error-free. This minimizes the chances of runtime errors related to misconfigurations.

https://github.com/envoyproxy/gateway/issues/2810

! Draft !


arkodg:
@alexwo is this necessary ?
based on https://github.com/envoyproxy/gateway/blob/main/internal/xds/cache/snapshotcache.go I believe we only push a snapshot if the IR exists

alexwo:
> @alexwo is this necessary ? based on https://github.com/envoyproxy/gateway/blob/main/internal/xds/cache/snapshotcache.go I believe we only push a snapshot if the IR exists, so if a IR hasn't been translated yet, it shouldnt result in a xds response (this needs to be cross checked)

Sure, If there are no items requiring translation, we don't anticipate an XDS response. Our focus is solely on verifying that the XDS server has successfully initiated and that a reconciliation loop, albeit inactive, has been executed. (empty / new cluster)

If there are any IRs translated due to a reconciliation loop containing items needing processing, we verify that a snapshot has indeed been persisted before deeming the instance ready. 

There are essentially two conditions that indicate the instance's health. (cluster with resources for xDS)

aoledk:
@alexwo You closed this PR in favor of @arkodg 's suggestion (https://github.com/envoyproxy/gateway/issues/2810#issuecomment-1981979019) to use longer `initialDelaySeconds` ?

PR #2920, opened 2024-03-14
chore: improve docs-serve

zirain:
after:
```shell
Environment: "dev"
Serving pages from memory
Running in Fast Render Mode. For full rebuilds on change: hugo server --disableFastRender
Web Server is available at http://localhost:1313/ (bind address 0.0.0.0)
```


zirain:
cc @Xunzhuo , hope this will not break netlify.

Xunzhuo (approved):
Thanks

PR #2932, opened 2024-03-15
refactor: collapse eg-metrics-svc into eg-svc

### LINKED ISSUE #2919: Collapse envoy-gateway-metrics-service into envoy-gateway service with a metrics port
*Description*:
>Describe the issue.

Can we reuse the `envoy-gateway` service and add a new port to it ? imo its easier for operations 

[optional *Relevant Links*:]

ShyunnY:
**What type of PR is this?**

refactor: Collapse envoy-gateway-metrics-service into envoy-gateway service with a metrics port

Fixes #2919


charts/gateway-helm/templates/_helpers.tpl
```diff
@@ -60,3 +60,11 @@ Create the name of the service account to use
...
 {{- end }}
 {{- end }}
+
+{{/*
+eg metrics annotations
+*/}}
+{{- define "eg.metrics.annotation" -}}
+prometheus.io/scrape: 'true'
+prometheus.io/port: '19001'
+{{- end }}
```
zirain, line 70: nit: new line at the end

zirain:
please kindly wait for https://github.com/envoyproxy/gateway/pull/2935

zirain:
@ShyunnY can you run `make gen-check` locally?

ShyunnY:
@zirain  Actually I can't run it, I need to change the existing `helm/default.yaml` file

zirain:
> @zirain Actually I can't run it, I need to change the existing `helm/default.yaml` file

`make helm-template` will update it, and it's part of `make gen-check`

charts/gateway-helm/templates/envoy-gateway-service.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
   labels:
     control-plane: envoy-gateway
   {{- include "eg.labels" . | nindent 4 }}
+  annotations:
```
zirain, line 9: Sorry, I may not clear, this should be on deployment/pod instead of service?

ShyunnY: I set it wrong, it should be on the pod. forgive me

zirain: what I learn from other projects is on Deployment/Pod, what the offical document said?

ShyunnY: There is no official recommendation on which one to use. However, when only capturing a small number of instances such as eg-controller, it may be better to set the annotation to Deployment/Pod.
In addition: when crawling a group of stateless applications with a large base, setting annotations to Service is indeed a common behavior :)

ShyunnY: I will change

zirain (approved):

arkodg (approved):
LGTM thanks !

arkodg (approved):

Issue #2930, opened 2024-03-15
HMAC secret envoy-gateway/envoy-oidc-hmac not found


hollanbm:
I believe this to be similar to #2823 

I upgraded from v0.6.0 There were No issues with the upgrade; just bumped the helm chart, and everything appeared to upgrade cleanly. I do not have any values defined, just the stock chart defaults.

However, when applying an OIDC security policy to an HTTP route, I am getting an error on the security policy. `HMAC secret envoy-gateway/envoy-oidc-hmac not found` 

Based on #2823 I would have expected the secret to be created for me. 

To attempt to debug this. I deleted the existing EnvoyProxy, GatewayClass, and Gateway resources from my cluster, then reapplied them. No change in behavior.

---

`kubectl describe securitypolicy test`
```yaml
Status:
  Ancestors:
    Ancestor Ref:
      Group:      gateway.networking.k8s.io
      Kind:       Gateway
      Name:       eg
      Namespace:  envoy-gateway
    Conditions:
      Last Transition Time:  2024-03-15T05:15:23Z
      Message:               HMAC secret envoy-gateway/envoy-oidc-hmac not found
      Observed Generation:   1
      Reason:                Invalid
      Status:                False
      Type:                  Accepted
    Controller Name:         gateway.envoyproxy.io/gatewayclass-controller
Events:                      <none>
```

The only thing unique about my setup so far is that I'm using the cert-manager API gateway annotations to generate certificates for the gateway.

```yaml
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-prod
  name: eg
  namespace: envoy-gateway
spec:
  gatewayClassName: eg
  listeners:
    - allowedRoutes:
        namespaces:
          from: All
      hostname: '*.test.domain.org'
      name: test-domain-org
      port: 443
      protocol: HTTPS
      tls:
        certificateRefs:
          - kind: Secret
            name: test-domain-org-tls
        mode: Terminate
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
  labels:
    app: test
  name: test
  namespace: test
spec:
  hostnames:
  - oidc.test.domain.org
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
    namespace: envoy-gateway
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: test
      port: 8080
[...]

PR #2934, opened 2024-03-15
Run certgen when upgrading

zhaohuabing:
Run certgen when upgrading EG helm chart because a new secret(HMAC) has been added after v0.0.6, and more may be added later. 

Fix https://github.com/envoyproxy/gateway/issues/2930

### DIFF
--- a/charts/gateway-helm/templates/certgen.yaml
+++ b/charts/gateway-helm/templates/certgen.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
   {{- include "eg.labels" . | nindent 4 }}
   annotations:
-    "helm.sh/hook": pre-install
+    "helm.sh/hook": pre-install, pre-upgrade
   {{- if .Values.certgen.job.annotations }}
     {{- toYaml .Values.certgen.job.annotations | nindent 4 -}}
   {{- end }}


zirain:
is there anyway to make sure it worked as excepted?

zhaohuabing:
> is there anyway to make sure it worked as excepted?

Just manually tested it on my dev environment：-）

zirain:
> > is there anyway to make sure it worked as excepted?
> 
> Just manually tested it on my dev environment：-）

I recall there's a upgrade test?

zhaohuabing:
The test looks like uninstalling the old version and then installing the new version, not exactly helm upgrade.

https://github.com/envoyproxy/gateway/blob/36505f4e0773f072952383280a5ce4145be40d77/test/e2e/tests/eg_upgrade.go#L66-L72

cnvergence (approved):

zirain:
> The test looks like uninstalling the old version and then installing the new version, not exactly helm upgrade.
> 
> We should have a proper upgrade test that helm upgrade the old release and run all the conformance tests and e2e tests on the upgrade version, this could be solved later.
> 
> https://github.com/envoyproxy/gateway/blob/36505f4e0773f072952383280a5ce4145be40d77/test/e2e/tests/eg_upgrade.go#L66-L72

can you update the test?

guydc:
@zhaohuabing , @zirain 
The test is:
1. Uninstalling main (in use by the regular conformance/e2e suite)
2. Installing latest rel
3. Upgrading from latest rel
https://github.com/envoyproxy/gateway/blob/36505f4e0773f072952383280a5ce4145be40d77/test/e2e/tests/eg_upgrade.go#L81

guydc:
Are there any possible adverse effects? e.g. will cert-gen override some other existing secrets that may block installation or break the pre-upgraded instance? For example, if some components are not dynamically reloading these secrets?

zhaohuabing:
> Are there any possible adverse effects? e.g. will cert-gen override some other existing secrets that may block installation or break the pre-upgraded instance? For example, if some components are not dynamically reloading these secrets?

Just add the missing HMAC secret, won't overwrite existing ones.

guydc (approved):

zirain (approved):

PR #2938, opened 2024-03-15
chore: use v1.0.0 as default for upgrade test

guydc:
Now that v1.0.0 is released, we should use it a baseline for upgrade tests in CI.


zhaohuabing (commented):
Is it reasonable to cover one more old versions(v 0.0.6)?

zirain (approved):

cnvergence (approved):

arkodg:
@zhaohuabing the release/v1.0 branch already has the v0.6 to v1.0 test

arkodg (approved):

zhaohuabing:
Do we support upgrade from v0.6 to latest？

arkodg:
@zhaohuabing id suggest we only start with supporting migrating from one minor version to the next minor version

PR #2942, opened 2024-03-15
fix: match SNI when using TLS listeners with hostname

michaelbeaumont:
**What type of PR is this?**

A fix for TLS listener with `hostname` handling.

This PR sets `filterChainMatch.serverNames` when using TLS listeners with a `hostname`.


arkodg:
hey @michaelbeaumont, for my understanding, is this PR is trying to solve  ?
```
Listener Protocol: TLS
TLS Mode: Terminate
Route Type: TCPRoute
```
https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls

afaik this should be supported today 
https://github.com/envoyproxy/gateway/pull/1431
https://gateway.envoyproxy.io/v1.0.0/user/security/tls-termination/

michaelbeaumont:
Hey, yeah traffic is terminated properly, but if I set `hostname` on that listener, there's no filtering of traffic by SNI as specified by the spec. [The final commit shows the difference](https://github.com/envoyproxy/gateway/pull/2942/commits/992dd33d1cab1e2a54ab5fb162f0868aff1125cd#diff-447319055fa6f1ac4b57037985f9051495d793be44984a6675266facaaf5d2d3R26) for [this listener](https://github.com/envoyproxy/gateway/pull/2942/commits/c473dcfe3f923e075663210e2fc10f7b55288880#diff-8e07addfcff6076812e513e97c7238cb3b3f5eb5f68630a17145c8168e89d53fR23).

arkodg:
got it makes sense, thanks for clarifying !
is there a way to modify the TCP Listener IR w/o affecting the HTTP Listener IR structure and achieving the same goal ?
Currently the `Hostnames` field within the HTTP Listener is used for Filter Chain creating with a `server_name` constaint

zhaohuabing:
@michaelbeaumont This PR looks Great - thank you for addressing the issue!

Just one small thing, could we add multiple hostnames in the gateway API translator and XDS translator tests for tls termination?  This could cover the scenario where a single listener needs to handle multiple SNIs? Thanks!

zhaohuabing (commented):
@michaelbeaumont Thank you for addressing the issue!  A few comments.

internal/xds/translator/testdata/out/xds-ir/tcp-route-tls-terminate.listeners.yaml
```diff
@@ -23,5 +23,31 @@
             sdsConfig:
               ads: {}
               resourceApiVersion: V3
+  - filterChainMatch:
+      serverNames:
+      - '*.envoyproxy.io'
+    filters:
+    - name: envoy.filters.network.tcp_proxy
+      typedConfig:
+        '@type': type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
```
zhaohuabing, line 32: Do we also add FilterChainMatch with serverNames if the filter is an HCM?

zhaohuabing (commented):
Could we add multiple hostnames in the gateway API translator and XDS translator tests for TLS termination? This could cover the scenario where a single listener needs to handle multiple SNIs. Thanks!

internal/gatewayapi/helpers.go
```diff
@@ -373,7 +373,7 @@ func irRouteDestinationName(route RouteContext, ruleIdx int) string {
 	return fmt.Sprintf("%srule/%d", irRoutePrefix(route), ruleIdx)
 }
 
-func irTLSConfigs(tlsSecrets []*v1.Secret) *ir.TLSConfig {
+func irTLSConfigs(tlsSecrets []*v1.Secret, listenerHostname *gwapiv1.Hostname) *ir.TLSConfig {
```
zhaohuabing, line 376:
```suggestion
func irTLSConfigs(tlsSecrets []*v1.Secret, sni *gwapiv1.Hostname) *ir.TLSConfig {
```

shuji-2019:
Is there any updates for this feature ?

internal/ir/xds.go
```diff
@@ -1123,13 +1126,12 @@ func (h TCPListener) Validate() error {
...
 // TLSInspectorConfig holds the configuration required for inspecting TLS
-// passthrough connections.
+// connections.
 // +k8s:deepcopy-gen=true
 type TLSInspectorConfig struct {
 	// Server names that are compared against the server names of a new connection.
 	// Wildcard hosts are supported in the prefix form. Partial wildcards are not
 	// supported, and values like *w.example.com are invalid.
-	// SNIs are used only in case of TLS Passthrough.
 	SNIs []string `json:"snis,omitempty" yaml:"snis,omitempty"`
```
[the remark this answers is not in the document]
michaelbeaumont: I think it's already exposed as `TLS.Passthrough` isn't it? I can add `SNIs` directly to `TLSConfig` if you prefer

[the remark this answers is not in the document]
arkodg: I'd vote for making `TLSInspectorConfig` an embedded field within `TLS` https://github.com/envoyproxy/gateway/blob/4d6da72fe523e732531fdfc02d499b0da3aaa185/internal/ir/xds.go#L1088 , removing the need for a dedicated `Passthrough` field

* passthrough logic can be computed using - `isTLSPassthrough := irListener.TLS != nil && irListener.TLS.Terminate == nil`
* this also eliminates the need to add `Inspector` into the `TLSConfig` field, which is common to HTTPS, which uses another field (`listener.Hostname`) for SNI inspection

aoledk:
@michaelbeaumont Are you still working on this? If not I can help, based on your previous work.

internal/ir/xds.go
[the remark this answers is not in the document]
michaelbeaumont: I pushed an update here, it changes a different part of the IR though, `passthrough` -> `inspector`, since that's now common between `mode: Terminate` and `mode: Passthrough`

michaelbeaumont:
Updated and merged

internal/gatewayapi/route.go
```diff
@@ -1082,6 +1082,14 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
 			accepted = true
 			irKey := t.getIRKey(listener.gateway)
 
+			tls := ir.TLS{
```
arkodg, line 1085: shouldnt this be wrapped inside an `if` 
e.g. https://github.com/envoyproxy/gateway/blob/33443f8846f105adaa57699fdd90202b5670c3ce/internal/gatewayapi/listener.go#L53

michaelbeaumont: I don't think anything's changed here, previously we set unconditionally:

```
TLS: &ir.TLS{Terminate: irTLSConfigs(listener.tlsSecrets)},
```

`irTLSConfigs` returns `nil` if there aren't any secrets. The line you linked to is also gated by checking the listener protocol.

arkodg: Cool thanks for the checking!

arkodg:
@michaelbeaumont 'make lint's is failing because of trailing whitespaces, once that's fixed, LGTM from my end !

arkodg:
@michaelbeaumont can you also fix DCO and force push ?

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM thanks!

michaelbeaumont:
@zhaohuabing I added another SNI to the test https://github.com/envoyproxy/gateway/pull/2942/files#diff-3481ca7cb1d20b614db1882fa576de2d303d434e519a17d12c2792436d6a0e18R28 actually

Issue #2891, opened 2024-03-12
Remove duplicated HTTP filters


zhaohuabing:
The current per-route filter implementation creates a filter in the HCM filter chain for each ir.HTTPRoute. This can result in duplicated HTTP filters with the same configuration and cause a huge Envoy configuration file.  To optimize this, we should only create one filter for each unique ODIC/ExtAuth/BaiscAuth configuration and reuse it across all applicable ir.HTTPRoute.

Thanks @guydc for identifying this during the ExtAuth PR review.

guydc:
should we close this?

arkodg:
this may need to be done for all per route filters :)

PR #2944, opened 2024-03-16
Remove duplicated http filters

zhaohuabing:
This PR removes duplicated HTTP filters from the HCM HTTP filter chain. Instead of creating one HTTP filter per route, now EG only creates one HTTP filter for each SecurityPolicy. The same filter will be enabled for all the applicable routes.

It also changes the filter suffix to the policy name to make it easier to debug.

Before:
name: envoy.filters.http.basic_auth/first-route

After:
name: envoy.filters.http.basic_auth/securitypolicy/default/policy-for-http-route-1

Fix https://github.com/envoyproxy/gateway/issues/2891


guydc (approved):

zirain (approved):

internal/gatewayapi/securitypolicy.go
```diff
@@ -366,20 +366,24 @@ func (t *Translator) translateSecurityPolicyForRoute(
 	}
 
 	if policy.Spec.OIDC != nil {
-		if oidc, err = t.buildOIDC(policy, resources); err != nil {
+		if oidc, err = t.buildOIDC(
+			irConfigName(policy),
```
arkodg, line 370: nit: curious why this passed for `buildOIDC` but computed for `buildBasicAuth`

zhaohuabing: I missed that :-(  

It shouldn't be needed anymore.

https://github.com/envoyproxy/gateway/pull/2992 @arkodg

PR #2945, opened 2024-03-16
chore: make clean should clean tools/bin

zirain:


arkodg (commented):
lets add a `tools.clean` and add this `rm` under it, with the definition living under 
https://github.com/envoyproxy/gateway/blob/main/tools/make/tools.mk

tools/make/tools.mk
```diff
@@ -63,3 +63,11 @@ tools/bin/$(notdir $(SHELLCHECK_TXZ)):
 	mkdir -p $(@D)
 	tar -C $(@D) -Jxmf $< --strip-components=1 shellcheck-v$(SHELLCHECK_VERSION)/shellcheck
 endif
+
+tools.clean: # Remove all tools
+	@$(LOG_TARGET)
+	rm -rf $(tools.bindir)
+
+.PHONY: clean
+clean: ## Remove all files that are created during builds.
```
arkodg, line 72: umm there's already a `clean` target under `go.mk`

zirain: there's another one in `docs.mk`, just follow the old way.

arkodg: ah so `make clean` will call all of them ?

arkodg (approved):

shawnh2 (approved):

PR #2948, opened 2024-03-17
feat: support failOpen in ext auth

### LINKED ISSUE #2897: Support failOpen in ext auth
ext authz setting is `failClosed` by default 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto

Some users use ext auth as a way to enrich request headers, and prefer this external call to be `failOpen` (its okay to send the request to upstream/backend in case the API call fails/times out)

[optional *Relevant Links*:]

yeedove:
**What type of PR is this?**

- support failOpen in ext auth, follow up PR

Fixes #2897


zhaohuabing (approved):
LGTM Thanks!

PR #2951, opened 2024-03-17
refactor: separate conditions computing for listeners and routes into status pkg

### LINKED ISSUE #860: refactor status conditions logic for Gateway API resources
*Description*:
The GatewayClass and Gateway logic for computing status conditions lives here
https://github.com/envoyproxy/gateway/blob/main/internal/status/conditions.go

and the logic for Routes lives here
https://github.com/envoyproxy/gateway/blob/085e6fa5a56066472a52e8c07bc1aefa3ffca5a2/internal/gatewayapi/contexts.go#L89

These should be consolidated to one place for improving code reuse

shawnh2:
**What type of PR is this?**

- move gateway listeners and route conditions computing logic from `internal/gatewayapi/context.go` to `internal/gatewayapi/status/`
- separate gateway, gatewayclass conditions computing and tests in `internal/status/conditions.go` into each corresponding place in `internal/gatewayapi/status/xxx.go`

Fixes #860


zirain:
should this live under `internal/gatewayapi`?

shawnh2:
> should this live under `internal/gatewayapi/status`?

currently the conditions computing for Gateway and GatewayClass lives in `internal/status`, we should group them together, so I also move them into `internal/status`.

zirain (approved):

arkodg:
thanks for unifying all the logic here @shawnh2 !
I'm a +1 on @zirain's idea of moving the status pieces under `/internal/gatewayapi` since they are all tied to GWAPI resources
we can move out the k8s piece i.e. the updator into `internal/kubernetes/status` 
https://github.com/envoyproxy/gateway/blob/main/internal/status/status.go

shawnh2:
> thanks for unifying all the logic here @shawnh2 !
> 
> I'm a +1 on @zirain's idea of moving the status pieces under `/internal/gatewayapi/status` since they are all tied to GWAPI resources
> 
> we can move out the k8s piece i.e. the updator into `internal/kubernetes/status` 
> 
> https://github.com/envoyproxy/gateway/blob/main/internal/status/status.go

Sure, so in this case we don't need `internal/status` anymore.

arkodg (approved):
LGTM thanks for this refactor !

arkodg (approved):

zhaohuabing (approved):
LGTM.  The status computation looks much better!

PR #2952, opened 2024-03-17
feat(translator): implement connection limit

### LINKED ISSUE #2685: Client Traffic Policy: Support Connection Limiting

guydc:
- Implements connection limits in CTP based on API defined in #2709. 
- Simple e2e test that demonstrates connection limiting by opening connections beyond the configured limit and expecting following requests to fail
- Docs

Fixes #2685


internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -644,3 +649,39 @@ func (t *Translator) translateListenerTLSParameters(policy *egv1a1.ClientTraffic
...
+func translateListenerConnection(connection *egv1a1.Connection, httpIR *ir.HTTPListener) error {
+	// Return early if not set
+	if connection == nil {
+		return nil
+	}
+
+	irConnection := &ir.Connection{}
+
+	if connection.Limit != nil {
+		hasLimit := false
```
arkodg, line 662: so `hasLimit` is being used to make sure at least one of the sub fields is set ? 
isn't this check already happening in CEL ?

api/v1alpha1/connection_types.go
```diff
@@ -20,9 +20,8 @@ type ConnectionLimit struct {
 	// When the limit is reached, incoming connections will be closed after the CloseDelay duration.
 	// Default: unlimited.
 	//
-	// +optional
 	// +kubebuilder:validation:Minimum=0
-	Value *int64 `json:"value,omitempty"`
+	Value int64 `json:"value,omitempty"`
```
arkodg, line 24: is there a default envoy uses here ?
there maybe a valid use case here where the user wants to use the default here (not set it) and set CloseDelay instead (where the default isn't good enough)

guydc: No default is mentioned in the docs. It looks like the filter constructor requires a value to be provided: https://github.com/envoyproxy/envoy/blob/e4bd0e6f70749a50dd122d5d7006e8913bb9e84c/source/extensions/filters/network/connection_limit/connection_limit.cc#L17C24-L17C53 by the filter config. So, I think that it's legitimate to require a value if the user decides to opt-in for this feature.

arkodg: Do you think EG should set a default here ?

zhaohuabing: It might be difficult to choose a default value for max connections because it depends on things we don't know before deploying EG in production: the scale of clients, the spec of the machine the EG is running on, etc. :-)

guydc: +1 to huabing's position

guydc:
renamed the container to "connectionLimit" based on discussion here: https://github.com/envoyproxy/gateway/pull/2805#discussion_r1533437654

api/v1alpha1/connection_types.go
```diff
@@ -9,20 +9,19 @@ import gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 // Connection allows users to configure connection-level settings
 type Connection struct {
```
zirain, line 11: will this be a little odd?
```yaml
connection:
  connectionLimit:
```

arkodg: its to tackle the issue outlined in #2805

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
Well done! Thanks!

Issue #2954, opened 2024-03-18
Refactor ir HTTPRoute: group related features into structures


zhaohuabing:
yah makes sense, something like `TrafficFeatures *TrafficFeatures` and check for `nil`

_Originally posted by @arkodg in https://github.com/envoyproxy/gateway/pull/2888#discussion_r1522007537_

PR #2926, opened 2024-03-15
refactor errors and xds for BackendTrafficPolicy and SecurityPolicy

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

- reuse the translated logic, currently there're a lot of dup code in `translateXXXPolicyForRoute` and `translateXXXPolicyForGateway` for BTP and SP, this PR extract the common part of these two methods into one shared method
- group BTP and SP features in `xds.go` into a struct, so we can perform `Any()` and `Validate()`, make the code structure more clean
- wrap the build error in translate method with its feature's name via `perr.WrapWithMessage()` for SP and BTP, providing more context about errors
- for BTP, change errors directly return behavior to errors join

**Which issue(s) this PR fixes**:

Fixes #2894 
Fixes #2954

PR #2955, opened 2024-03-18
Refactor: group security configuration in ir

zhaohuabing:
This PR groups security-related features in ir.HTTPRoute into a structure. 

Part of https://github.com/envoyproxy/gateway/issues/2954


internal/ir/xds.go
```diff
@@ -475,6 +462,56 @@ type HTTPRoute struct {
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty" yaml:"tcpKeepalive,omitempty"`
 	// Retry settings
 	Retry *Retry `json:"retry,omitempty" yaml:"retry,omitempty"`
+	// Security configuration for the route.
+	Security *Security `json:"security,omitempty" yaml:"security,omitempty"`
+}
+
+// Security holds the Security configuration associated with  the HTTP Route
+// +k8s:deepcopy-gen=true
+type Security struct {
```
zhaohuabing, line 471: Should we call this struct `Security`, `SecurityFeature`, or `SecurityPolicy`? I'm slightly inclined to `Security`.

shawnh2: hi @zhaohuabing, this work is similar to my recent PR https://github.com/envoyproxy/gateway/pull/2926/files#diff-b0330312d3f47f14bcdc8e1b589d84001a1969564c86ff59347bd56795db7ccaR451

think it through again, `SecurityFeatures` makes much sense to me. and also we can make this field inline, that would aviod so many nil check.

similar to [here](https://github.com/envoyproxy/gateway/pull/2926/files#diff-b0330312d3f47f14bcdc8e1b589d84001a1969564c86ff59347bd56795db7ccaR479), we can perform any check directly on this field

PR #2967, opened 2024-03-18
fix: Allow Policy to attach to multiple http listeners 

### LINKED ISSUE #2963: bug: Allow Policy to attach to multiple http listeners
@arkodg You are right - we do have a regression here. The following _should_ be possible but is not possible right now:

Setting a single `ClientTrafficPolicy` or a `SecurityPolicy` that targets a single `Gateway`, when that `Gateway` has more than one listener. The policy in this case is intended to apply the same settings to all of the listeners, and is valid, but it is currently being rejected.

If, instead of targeting the entire `Gateway`, the `SecurityPolicy` were to target a specific `HTTPRoute` then the correct behavior would have been to reject the `SecurityPolicy` since it's not possible to have different configuration for two routes  connected to the same listener. 

I think that this should be tracked as a separate issue though - WDYT?

_Originally posted by @liorokman in https://github.com/envoyproxy/gateway/issues/2742#issuecomment-2002369812_

liorokman:
Fixes #2963

### DIFF
--- a/internal/gatewayapi/clienttrafficpolicy.go
+++ b/internal/gatewayapi/clienttrafficpolicy.go
@@ -328,8 +328,14 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, 
 
 	// IR must exist since we're past validation
 	if httpIR != nil {
+		// Get a list of all other non-TLS listeners on this Gateway that share a port with
+		// the listener in question.
 		if sameListeners := listenersWithSameHTTPPort(xds, httpIR); len(sameListeners) != 0 {
 			if attachedToGateway {
+				// If this policy is attached to an entire gateway and the mergeGateways feature
+				// is turned on, validate that all the listeners affected by this policy originated
+				// from the same Gateway resource. The name of the Gateway from which this listener
+				// originated is part of the listener's name by construction.
 				gatewayName := irListenerName[0:strings.LastIndex(irListenerName, "/")]
 				conflictingListeners := []string{}
 				for _, currName := range sameListeners {
@@ -341,6 +347,8 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, 
 					return fmt.Errorf("affects additional listeners: %s", strings.Join(conflictingListeners, ", "))
 				}
 			} else {
+				// If this policy is attached to a specific listener, any other listeners in the list
+				// would be affected by this policy but should not be, so this policy can't be accepted.
 				return fmt.Errorf("affects additional listeners: %s", strings.Join(sameListeners, ", "))
 			}
 		}
--- a/internal/gatewayapi/clienttrafficpolicy.go
+++ b/internal/gatewayapi/clienttrafficpolicy.go
@@ -344,12 +344,12 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, 
 					}
 				}
 				if len(conflictingListeners) != 0 {
-					return fmt.Errorf("affects additional listeners: %s", strings.Join(conflictingListeners, ", "))
+					return fmt.Errorf("ClientTrafficPolicy is being applied to multiple http (non https) listeners (%s) on the same port, which is not allowed", strings.Join(conflictingListeners, ", "))
 				}
 			} else {
 				// If this policy is attached to a specific listener, any other listeners in the list
 				// would be affected by this policy but should not be, so this policy can't be accepted.
-				return fmt.Errorf("affects additional listeners: %s", strings.Join(sameListeners, ", "))
+				return fmt.Errorf("ClientTrafficPolicy is being applied to multiple http (non https) listeners (%s) on the same port, which is not allowed", strings.Join(sameListeners, ", "))
 			}
 		}
 	}
--- a/internal/gatewayapi/clienttrafficpolicy.go
+++ b/internal/gatewayapi/clienttrafficpolicy.go
@@ -124,7 +124,7 @@ func (t *Translator) ProcessClientTrafficPolicies(resources *Resources,
 				// It must exist since we've already finished processing the gateways
 				gwXdsIR := xdsIR[irKey]
 				if string(l.Name) == section {
-					err = validatePortOverlapForClientTrafficPolicy(l, gwXdsIR)
+					err = validatePortOverlapForClientTrafficPolicy(l, gwXdsIR, false)
 					if err == nil {
 						err = t.translateClientTrafficPolicyForListener(policy, l, xdsIR, infraIR, resources)
 					}
@@ -234,7 +234,7 @@ func (t *Translator) ProcessClientTrafficPolicies(resources *Resources,
 				irKey := t.getIRKey(l.gateway)
 				// It must exist since we've already finished processing the gateways
 				gwXdsIR := xdsIR[irKey]
-				if err := validatePortOverlapForClientTrafficPolicy(l, gwXdsIR); err != nil {
+				if err := validatePortOverlapForClientTrafficPolicy(l, gwXdsIR, true); err != nil {
 					errs = errors.Join(errs, err)
 				} else if err := t.translateClientTrafficPolicyForListener(policy, l, xdsIR, infraIR, resources); err != nil {
 					errs = errors.Join(errs, err)
@@ -312,7 +312,7 @@ func resolveCTPolicyTargetRef(policy *egv1a1.ClientTrafficPolicy, gateways map[t
 	return gateway.GatewayContext, nil
 }
 
-func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds) error {
+func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, attachedToGateway bool) error {
 	// Find Listener IR
 	// TODO: Support TLSRoute and TCPRoute once
 	// https://github.com/envoyproxy/gateway/issues/1635 is completed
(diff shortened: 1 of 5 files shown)


internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -312,7 +312,7 @@ func resolveCTPolicyTargetRef(policy *egv1a1.ClientTrafficPolicy, gateways map[t
 	return gateway.GatewayContext, nil
 }
 
-func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds) error {
+func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, attachedToGateway bool) error {
```
arkodg, line 315: adding comments for the new logic would really help

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -329,7 +329,20 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds)
 	// IR must exist since we're past validation
 	if httpIR != nil {
 		if sameListeners := listenersWithSameHTTPPort(xds, httpIR); len(sameListeners) != 0 {
-			return fmt.Errorf("affects additional listeners: %s", strings.Join(sameListeners, ", "))
+			if attachedToGateway {
```
arkodg, line 332: this is now O(n^3)
can we instead hold a map containing listener name as the key and policy as the value ?
* if key is missing, add key, and return
* if key is present, policy is same, return
* if key is present, and a different policy, reject

this takes an approach of first policy wins, which imo is better than no policy winning

zhaohuabing (commented):
We probably won't need this https://github.com/envoyproxy/gateway/pull/2970

arkodg:
@zhaohuabing we still need it for CTP which adds fields like `MergeSlashes` on the `hcm` directly, so we have a conflict for http (non https/tls) case

internal/gatewayapi/clienttrafficpolicy.go
liorokman: > this takes an approach of first policy wins, which imo is better than no policy winning

"First policy wins" doesn't make sense. It's not a conflict between two policies where one can win and the other can lose, it's that translating a single policy (no other policies existing in the system) would result in it affecting multiple listeners including listeners that should not be affected by the policy.

liorokman: > this is now O(n^3)

Looking closely at the ` validatePortOverlapForClientTrafficPolicy` method, it seems to me that it's `O(n)` where `n` is the number of HTTP listeners attached to a Gateway XDS representation.


```golang

func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds, attachedToGateway bool) error {
	irListenerName := irHTTPListenerName(l)
	var httpIR *ir.HTTPListener

         // O(n) loop

	for _, http := range xds.HTTP { 
		if http.Name == irListenerName {
			httpIR = http
			break
		}
	}

	// IR must exist since we're past validation
	if httpIR != nil {
                //           listenersWithSameHTTPPort is O(n) and it is called once. 
                //  This is an 'if' statment, not a 'for' statement.
		if sameListeners := listenersWithSameHTTPPort(xds, httpIR); len(sameListeners) != 0 {


			if attachedToGateway {
				gatewayName := irListenerName[0:strings.LastIndex(irListenerName, "/")]
				conflictingListeners := []string{}

                                // sameListeners is an array with an upper bound of O(n)

				for _, currName := range sameListeners {
					if strings.Index(currName, gatewayName) != 0 {
						conflictingListeners = append(conflictingListeners, currName)
					}
				}
				if len(conflictingListeners) != 0 {
					return fmt.Errorf("affects additional listeners: %s", strings.Join(conflictingListeners, ", "))
				}
			} else {
				return fmt.Errorf("affects additional listeners: %s", strings.Join(sameListeners, ", "))
			}
		}
	}
	return nil
}

```

arkodg: @liorokman adding some reasons why we should consider first policy wins
* we already oldest config wins in most places https://github.com/envoyproxy/gateway/blob/6fa99be5a9a3e5d16294492162181ff64db67003/internal/gatewayapi/clienttrafficpolicy.go#L43
* this allows existing traffic to not get interrupted/dropped, ensures that is policy1 targets listener1 at t1 and policy2 targets listener2 at t2, listener1 is not affected

zhaohuabing:
We don't need this for SecurityPolicy anymore.

internal/gatewayapi/securitypolicy.go
```diff
@@ -194,7 +194,7 @@ func (t *Translator) ProcessSecurityPolicies(securityPolicies []*egv1a1.Security
 			irKey := t.getIRKey(targetedGateway.Gateway)
```
zhaohuabing, line 194: We don't need this for SecurityPolicy anymore. It's handled in https://github.com/envoyproxy/gateway/pull/2970

internal/gatewayapi/clienttrafficpolicy.go
liorokman: @arkodg The bug occurs when there is exactly one policy. The scenario where there are two policies is out of scope for this issue. There's no way to implement "first policy wins", because "first policy wins" means "reject the policy provided".

Here's a recreation of the bug that this PR is trying to solve.:

```yaml
    kind: Gateway
    metadata:
      name: gateway
    spec:
      listeners:
        - name: http
          port: 80
          protocol: HTTP
          hostname: foo.example.com
          allowedRoutes:
            namespaces:
              from: Same
        - name: http2
          port: 80
          protocol: HTTP
          hostname: bar.example.com
          allowedRoutes:
            namespaces:
              from: Same
---
  kind: ClientTrafficPolicy
  metadata:
    name: target-gateway
  spec:
    targetRef:
      kind: Gateway
      name: gateway
      namespace: default
      sectionName: http
    path:
        escapedSlashesAction: RejectRequest
    timeout:
      http:
        requestReceivedTimeout: "5s"
      
```

The above single `ClientTrafficPolicy` tries to target only one listener - the one called `http` - and because both `http` and `http2` are non-TLS and on the same port, the `requestReceivedTimeout`  configured for `http2` will also be `5s`. Even worse, while `http` should have the `RejectRequest` action for escaped slashes, `http2` _should_ use the default value (`UnescapeAndForward`) but will instead be configured to `RejectRequest`.


Without this PR the translated IR will be correct - the various settings from the `ClientTrafficPolicy` will be set in the correct location - only on `http` and not on `http2`. The bug occurs because the correct IR is not translatable to valid XDS.

internal/gatewayapi/securitypolicy.go
liorokman: I removed the code that validates `SecurityPolicy` for this issue.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -328,8 +328,29 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds)
...
+						conflictingListeners = append(conflictingListeners, currName)
+					}
+				}
+				if len(conflictingListeners) != 0 {
+					return fmt.Errorf("affects additional listeners: %s", strings.Join(conflictingListeners, ", "))
+				}
+			} else {
+				// If this policy is attached to a specific listener, any other listeners in the list
+				// would be affected by this policy but should not be, so this policy can't be accepted.
+				return fmt.Errorf("affects additional listeners: %s", strings.Join(sameListeners, ", "))
```
arkodg, line 352:
```suggestion
				return fmt.Errorf("policy attached to multiple http (non https) listeners %s on the same port is not allowed", strings.Join(sameListeners, ", "))
```

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -328,8 +328,29 @@ func validatePortOverlapForClientTrafficPolicy(l *ListenerContext, xds *ir.Xds)
...
+				// originated is part of the listener's name by construction.
+				gatewayName := irListenerName[0:strings.LastIndex(irListenerName, "/")]
+				conflictingListeners := []string{}
+				for _, currName := range sameListeners {
+					if strings.Index(currName, gatewayName) != 0 {
+						conflictingListeners = append(conflictingListeners, currName)
+					}
+				}
+				if len(conflictingListeners) != 0 {
+					return fmt.Errorf("affects additional listeners: %s", strings.Join(conflictingListeners, ", "))
```
arkodg, line 347: same as below

arkodg:
thanks for adding comments @liorokman, added a minor comment around the error string to make it clearer why the policy was rejected, else LGTM !

internal/gatewayapi/clienttrafficpolicy.go
liorokman: The suggested error message is a bit misleading, since the policy document as applied by the user is explicitly not attached to any other listener.

How about `policy would affect additional http (non https) listeners to which it was not attached: %s` ?

arkodg: ah instead of `attached`, we could use `applied` :) 

`ClientTrafficPolicy is being applied to multiple http (non https) listeners %s on the same port, which is not allowed`

arkodg (approved):
LGTM thanks !

zirain (approved):

arkodg:
all e2e tests consistently seem to be consistently failing @liorokman

liorokman:
> all e2e tests consistently seem to be consistently failing @liorokman

@arkodg all e2e tests have passed.

PR #2970, opened 2024-03-18
fix: add missing http filters to the http filter chain

zhaohuabing:
This PR patches HTTP filters on the existing `HCM` of an HTTP listener's `DefaultFilterChain`.

Fix https://github.com/envoyproxy/gateway/issues/2742


internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
```diff
@@ -2,21 +2,18 @@
     socketAddress:
       address: 0.0.0.0
       portValue: 10443
-      protocol: UDP
```
arkodg, line 5: why did this yaml change ?

arkodg:
hey @zhaohuabing thanks for working on the fix + the refactor
can we split up the refactor into another follow PR ? It allows us to back port this into v1.0.0

internal/xds/translator/testdata/out/xds-ir/http3.listeners.yaml
zhaohuabing: The order of the quick and tcp listener changes.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #2965, opened 2024-03-18
bug: envoy proxy pods restarted always when configuration change


zetaab:
*Description*:
```
 % kubectl get pods -n envoy-gateway-system
NAME                                          READY   STATUS        RESTARTS   AGE
envoy-eg-external-a514c411-5ffc5d664b-74g5j   2/2     Running       0          2d19h
envoy-eg-external-a514c411-5ffc5d664b-dhmmg   2/2     Running       0          2d19h
envoy-eg-internal-7f4ff7e4-7fb9c8d7df-8kjgk   2/2     Terminating   0          27s
envoy-eg-internal-7f4ff7e4-7fb9c8d7df-krcqc   2/2     Terminating   0          25s
envoy-eg-internal-7f4ff7e4-8685896d59-4z8n8   1/2     Terminating   0          4m31s
envoy-eg-internal-7f4ff7e4-8685896d59-gqtd7   2/2     Running       0          6s
envoy-eg-internal-7f4ff7e4-8685896d59-lrd4f   2/2     Running       0          4s
envoy-gateway-5987f4589-9h6ts                 1/1     Running       0          3d
```

When I modify like httproutes it will lead to envoy pod restarts. This is situation that is not really good when using external loadbalancers in front of envoy. I do understand that envoy will drain connections. However, when envoy uses `externalTrafficPolicy: Local` by default, it means that external loadbalancer will mark only nodes as healthy which contains the envoy pods. Now when these pods are moving between machines, it will always take 10-30 seconds (depends how loadbalancer healthchecks are installed) that services will start replying again.

*Repro steps*:
use type loadbalancer service in front of envoy, if needed modify the external loadbalancer healthcheck intervals to 60 seconds (to see how it really behaves). Then modify httproute configurations and see when pods start restarting and moving between kubernetes nodes -> it will make the services unavailable for some seconds.

Instead of restarting pods, envoy configurations should be reloaded. Avoid modifying kubernetes deployment configuration itself all the time, it will make downtime when using external loadbalancers and externaltrafficpolicy local - the health checks are not that fast.

*Environment*:
eg 1.0.0

arkodg:
@zetaab can you share config to repro this ? is this specific when setting `mergeGateways` to `true` ?

cnvergence:
possibly related, https://github.com/envoyproxy/gateway/issues/2637

arkodg:
does this issue still exist @zetaab ?

zetaab:
[...]

PR #2973, opened 2024-03-18
fix: use unique container ports in deployment

### LINKED ISSUE #2964: bug: duplicate port definition when using mergeGateway
*Description*:
When using mergeGateway and creating multiple gateways which uses 10443 (https), it will lead to error

*Repro steps*:
1) create new gatewayclass with merge gateways enabled
2) create two different gateways to that gatewayclass

*Environment*:
eg 1.0.0

*Logs*:
```
 % kubectl rollout restart deployment/envoy-eg-internal-7f4ff7e4
Warning: spec.template.spec.containers[0].ports[2]: duplicate port definition with spec.template.spec.containers[0].ports[1]
```

when taking a look for the configuration:
```
        ports:
        - containerPort: 10080
          name: envoy-cfc433c8
          protocol: TCP
        - containerPort: 10443
          name: envoy-938fd695
          protocol: TCP
        - containerPort: 10443
          name: echose-ebb62894
          protocol: TCP
        - containerPort: 19001
          name: metrics
          protocol: TCP
```

that is incorrect, the containerport is unique it seems.

zetaab:
**What type of PR is this?**
fix: use unique ports in deployment

Kubernetes requires that container port numbers are unique. So those can exist only once. Current implementation is adding duplicate ports if there are like two listeners listening port 443.

This PR also reduces the amount of restarts needed for deployments https://github.com/envoyproxy/gateway/issues/2965 because it will not modify deployment configuration anymore that often when listeners are modifierd. 

Fixes #2964

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -107,6 +107,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 	var ports []corev1.ContainerPort
 
 	// Iterate over listeners and ports to get container ports
+	uniquePorts := make(map[string]struct{})
 	for _, listener := range infra.Listeners {
 		for _, p := range listener.Ports {
 			var protocol corev1.Protocol
@@ -119,12 +120,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 				return nil, fmt.Errorf("invalid protocol %q", p.Protocol)
 			}
 			port := corev1.ContainerPort{
-				// hashed container port name including up to the 6 characters of the port name and the maximum of 15 characters.
-				Name:          utils.GetHashedName(p.Name, 6),
+				Name:          fmt.Sprintf("port-%d-%s", p.ContainerPort, protocol),
 				ContainerPort: p.ContainerPort,
 				Protocol:      protocol,
 			}
-			ports = append(ports, port)
+			uname := fmt.Sprintf("%s-%d", protocol, p.ContainerPort)
+			if _, ok := uniquePorts[uname]; !ok {
+				ports = append(ports, port)
+				uniquePorts[uname] = struct{}{}
+			}
 		}
 	}
 
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -120,7 +120,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 				return nil, fmt.Errorf("invalid protocol %q", p.Protocol)
 			}
 			port := corev1.ContainerPort{
-				Name:          fmt.Sprintf("%s-%d", protocol, p.ContainerPort),
+				Name:          strings.ToLower(fmt.Sprintf("%s-%d", protocol, p.ContainerPort)),
 				ContainerPort: p.ContainerPort,
 				Protocol:      protocol,
 			}
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -120,7 +120,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 				return nil, fmt.Errorf("invalid protocol %q", p.Protocol)
 			}
 			port := corev1.ContainerPort{
-				Name:          fmt.Sprintf("port-%d-%s", p.ContainerPort, protocol),
+				Name:          fmt.Sprintf("%s-%d", protocol, p.ContainerPort),
 				ContainerPort: p.ContainerPort,
 				Protocol:      protocol,
 			}
--- a/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
@@ -106,6 +106,20 @@ func TestDeployment(t *testing.T) {
 			deploy:   nil,
 		},
 		{
+			caseName: "duplicate-listeners",
+			infra: func() *ir.Infra {
+				infra := newTestInfra()
+				infra.Proxy.Listeners[0].Ports = append(infra.Proxy.Listeners[0].Ports,
+					ir.ListenerPort{
+						Name:          "Envoy-second-https-listener",
+						Protocol:      ir.TCPProtocolType,
+						ContainerPort: envoyHTTPSPort,
+					})
+				return infra
+			}(),
+			deploy: nil,
+		},
+		{
 			caseName: "custom",
 			infra:    newTestInfra(),
 			deploy: &egv1a1.KubernetesDeploymentSpec{
--- a/internal/infrastructure/kubernetes/proxy/testdata/deployments/bootstrap.yaml
+++ b/internal/infrastructure/kubernetes/proxy/testdata/deployments/bootstrap.yaml
@@ -68,10 +68,10 @@ spec:
         name: envoy
         ports:
         - containerPort: 8080
-          name: TCP-8080
+          name: tcp-8080
           protocol: TCP
         - containerPort: 8443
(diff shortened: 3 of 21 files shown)


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -119,12 +120,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 				return nil, fmt.Errorf("invalid protocol %q", p.Protocol)
 			}
 			port := corev1.ContainerPort{
-				// hashed container port name including up to the 6 characters of the port name and the maximum of 15 characters.
-				Name:          utils.GetHashedName(p.Name, 6),
+				Name:          fmt.Sprintf("port-%d-%s", p.ContainerPort, protocol),
```
zirain, line 123: `protocol-port` seems better?
BTW, can you add a testcase for this?

zetaab:
@zirain fixed

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -119,12 +120,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 				return nil, fmt.Errorf("invalid protocol %q", p.Protocol)
 			}
 			port := corev1.ContainerPort{
-				// hashed container port name including up to the 6 characters of the port name and the maximum of 15 characters.
```
arkodg, line 122: @cnvergence can you recall why this hash was added ? I think it was to limit the `name` length, so we want want to change the args we pass, but we still may need to use `GetHashedName`

zetaab: the containerport name should not matter unless there is a kubernetes service which will refer to that name. https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2127-L2130

However, in case of envoy the service (type: loadbalancer) are using `targetPort` as Number instead of name. So the name is not used at all anywhere by default. https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5410-L5418

I do not see why we should still use `GetHashedName`. The only reason for that would be that name is going to be too long, but I cannot see that case really. If the name is [protocol]-[portnumber] the maximum lenght is going to be `SCTP-66666` which equals to 10. That is not too long.

zirain (commented):
Lgtm, defer to @arkodg and @cnvergence

internal/infrastructure/kubernetes/proxy/resource.go
cnvergence: We added the hash because we used listener port names as the container port names which could be larger than 15 chars
and in merged gateways we used the format with namespace/gateway/listener

cnvergence:
test fails are valid, so we are still missing something

internal/infrastructure/kubernetes/proxy/resource.go
cnvergence: this would be a good change for any deployment mode

zetaab:
@cnvergence 

```
=== RUN   TestE2E/FileAccessLog/Stdout
    helpers.go:402: error fetching Gateway: client rate limiter Wait returned an error: context deadline exceeded
    accesslog.go:44: 
        	Error Trace:	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.0/conformance/utils/kubernetes/helpers.go:423
        	            				/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.0/conformance/utils/kubernetes/helpers.go:337
        	            				/home/runner/work/gateway/gateway/test/e2e/tests/accesslog.go:44
        	            	error fetching Gateway: client rate limiter Wait returned an error: context deadline exceeded
        	Test:       	TestE2E/FileAccessLog/Stdout
        	Messages:   	error waiting for Gateway to have at least one IP address in status
```

for me this looks like we need better logs from envoy gateway controller, these logs in github actions does not tell much.

zetaab:
@cnvergence @arkodg @zirain I found the issue, kubernetes deployment container port name can contain only [a-z0-9-] uppercase chars are not allowed. It should work now

cnvergence:
Thanks for the update @zetaab, I will take a look again

cnvergence:
looks good, we might need to add this also to service ports to fix the issue
Just thinking, could we maybe check for uniqueness in the infra IR translator?

cnvergence:
I am referring to the original issue https://github.com/envoyproxy/gateway/issues/2964 and how to fix it, which is more related to the port values, https://github.com/envoyproxy/gateway/issues/2964#issuecomment-2007636363.
However, I approve of your port name change and I think we should get it merged.

internal/infrastructure/kubernetes/proxy/resource.go
arkodg: * can we instead fix the portName here https://github.com/envoyproxy/gateway/blob/7f5cbe84a2332a9f6b194ff2c276a320354a55f7/internal/gatewayapi/listener.go#L146 in the gatewayapi layer and reuse `p.Name` here ?
we'll need to make sure its not being used anywhere else

* and can we define the infraIRListenerPortName() https://github.com/envoyproxy/gateway/blob/7f5cbe84a2332a9f6b194ff2c276a320354a55f7/internal/gatewayapi/helpers.go#L375

sorry for nits, but trying to keep the code simpler to maintain

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -119,12 +120,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
...
-				Name:          utils.GetHashedName(p.Name, 6),
+				Name:          strings.ToLower(fmt.Sprintf("%s-%d", protocol, p.ContainerPort)),
 				ContainerPort: p.ContainerPort,
 				Protocol:      protocol,
 			}
-			ports = append(ports, port)
+			uname := fmt.Sprintf("%s-%d", protocol, p.ContainerPort)
```
arkodg, line 127: `uname` is not needed, you could directly use `uniquePorts[port.Name]`

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -119,12 +120,15 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
...
+			if _, ok := uniquePorts[uname]; !ok {
```
arkodg, line 128: can we also add a comment here e.g. "Only append unique ports"

cnvergence:
Quickly checked today and the problem now resurfaces in the service
```bash
ERROR	watchable	message/watchutil.go:56	observed an error	{"runner": "infrastructure", "error": "failed to create or update service envoy-gateway-system/envoy-eg-internal-7f4ff7e4: for Create: Service \"envoy-eg-internal-7f4ff7e4\" is invalid: [spec.ports[2].nodePort: Duplicate value: 32079, spec.ports[2]: Duplicate value: core.ServicePort{Name:\"\", Protocol:\"TCP\", AppProtocol:(*string)(nil), Port:443, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:\"\"}, NodePort:0}]"}
```

arkodg:
shouldn't this existing code in the gateway api translator https://github.com/envoyproxy/gateway/blob/7f5cbe84a2332a9f6b194ff2c276a320354a55f7/internal/gatewayapi/listener.go#L125 already handle the uniquePort case ?

zetaab:
looks like so, but for some reason these duplicate ports are ending to kubernetes deployment. So the logic is not fully applied to everywhere

arkodg:
> looks like so, but for some reason these duplicate ports are ending to kubernetes deployment. So the logic is not fully applied to everywhere

Worth debugging that and fixing that since there's already existing code that's trying to do the same thing

cnvergence:
hi @zetaab, I have fixed that logic in the translator, would you like to add this port naming change?

PR #2981, opened 2024-03-20
Refactor and fix early return: HTTP Listener XDS translator

zhaohuabing:
This PR:
* Add a few comments to `processHTTPListenerXdsTranslation` and extract an `addRouteToRouteConfig` sub-method to make the HTTPListener construction a bit easier to understand and maintain.
* Instead of returning early, making sure all the listeners are processed, and collecting all the errors during translation, return them at the end. 
* Rename `addXdsHTTPFilterChain` to `addHCMToXDSListener` to better reflect what this method does.


zirain:
any test for this?

arkodg (approved):
LGTM thanks for this refactor ! much easier to read now
hope we didn't miss any lines while moving the function into other methods

Issue #2536, opened 2024-01-30
Envoy Client Certs for Ext Auth and Backend TLS


arkodg:
*Description*:
>Describe the issue.

Raising this PR to make a decision on which client certs to use when originating a TLS connection to ext Auth and Backend TLS (relates to https://github.com/kubernetes-sigs/gateway-api/discussions/2743)

Option 1
Reuse Listener (Downstream) certs

Option 2
Define a common proxy cert in the `EnvoyProxy` config

Option 3 (not possible today)
Define certs in each config
* within the SecurityPolicy.ExtAuth.TLS field
* not possible in BackendTLS, since its a upstream API

arkodg:
ptal @envoyproxy/gateway-maintainers

zhaohuabing:
> Option 1
> Reuse Listener (Downstream) certs

The problem of this approach is that existing ext auth and jwt services may not be able to verify the Listener certs.

arkodg:
> EG doesn't use a global root ca like Istio does. So the problem of the first two approaches is that existing ext auth and jwt services may not be able to verify the client certs.

@zhaohuabing imo its a matter of sharing the CA (which is meant to be shared for validating trust anchor) with those entities (ext auth svc and backend)

zhaohuabing:
> > EG doesn't use a global root ca like Istio does. So the problem of the first two approaches is that existing ext auth and jwt services may not be able to verify the client certs.
> 
> @zhaohuabing imo its a matter of sharing the CA (which is meant to be shared for validating trust anchor) with those entities (ext auth svc and backend)


One possible solution is allow the user to specify an external Root CA cert in the EnvoyProxy config. This Root CA cert is generated by a well-known public/shared CA trusted by all the external services and backends. 
Then Envoy can use this root CA to generate the server certs for the Listeners and the client certs for the Backends and the external services.

guydc:
Option 1: it's possible (not 100% sure) that a server cert cannot always be used as a client cert, e.g. if it doesn't have the appropriate [extended key usage](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12) `TLS WWW client authentication`. For example, see [here](https://docs.apigee.com/how-to-guides/validating-certificate-purpose#extended-key-usage). 

Option 2 limits flexibility and option 3 creates a bit of duplication. I think that they're ok as mid-term mitigations. 
 
I think that in one of the community meetings, we discussed the following option as well:
[...]

Issue #2901, opened 2024-03-12
Backend TLS Settings


guydc:
*Description*:

Envoy Gateway should support configuration of backend TLS parameters such as ciphers, TLS versions, etc.

Similar to backend client certificate (#2536), these parameters can be configured in the `EnvoyProxy` resource by a Gateway Admin. Greater flexibility may be supported in the future with the evolution of `BackendTLSPolicy` in Gateway-API.

guydc:
dupe: https://github.com/envoyproxy/gateway/issues/2536

guydc:
closed by #3218

PR #2984, opened 2024-03-20
API for client cert

zhaohuabing:
Related: https://github.com/envoyproxy/gateway/issues/2536


api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,6 +87,11 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// ClientCertRef defines the reference to a Kubernetes Secret that contains
+	// the client certificate and private key for Envoy to use when connecting to
+	// backend services and external services, such as ExtAuth, ALS, OpenTelemetry, etc.
+	ClientCertRef *gwapiv1.SecretObjectReference `json:"clientCertRef,omitempty"`
```
zhaohuabing, line 94: How can we specify which backend service enforces client validation and which doesn't require it?

zhaohuabing, line 94: Do we want a container here for more TLS-related configurations like these mentioned in https://github.com/envoyproxy/gateway/issues/2901 ?

guydc: AFAIK, if the the TLS `server_hello` doesn't include a `certificate_request`, then the client [does not send the certificate](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.6). 

I can still think of some scenarios where sending a "technical" client cert when asked would be a problem. Some servers support multiple means of auth (client-cert, token, ... ), e.g. the K8s API server: https://kubernetes.io/docs/reference/access-authn-authz/authentication/. When proxying such a server, EG may be requested to present a client cert, and will present it. The server may try to authenticate the end-user based on the EG client cert, when the end-user actually intended for auth to be based on some L7 token. This will break such an integration.

guydc: +1 for container with additional settings

zhaohuabing: This is not the ideal location for backend services' TLS parameters since they may be different per backend. But we may still need a container for other client-side TLS options.

guydc: In the long term, I expect that this will be part of BTLS Policy, with all the flexibility required per-backend. In the meanwhile, this could be reasonable mitigation...

guydc (approved):
LGTM! As we also have an assignee for #2901 , it would be great to get this API in.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,6 +87,20 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// TLS is the TLS configuration for the Envoy proxy to use when connecting to
+	// backend services and external services, such as ExtAuth, ALS, OpenTelemetry, etc.
+	// +optional
+	TLS *EnvoyTLSConfig `json:"tls,omitempty"`
```
zirain, line 94: not good at TLS, is it possible need a slice here?

arkodg, line 94: should we call this ClientTLS ?

zhaohuabing: It's the client cert for Envoy to use when connecting to backend services. One would be enough unless we have use cases where we want to differentiate envoy identities when connecting to different backends.🤔

zhaohuabing: @guydc mentioned that he wants to a[dd some server-side setting](https://github.com/envoyproxy/gateway/issues/2901)s later. If that's the case, then it would be better to use a more generalized name.

guydc:
Hi @zhaohuabing ! Did you see this #3218? It also covers the API/Impl for client cert based on previous discussion in the community meeting.

zhaohuabing:
> Hi @zhaohuabing ! Did you see this #3218? It also covers the API/Impl for client cert based on previous discussion in the community meeting.

Let's do it in #3218 since my focus is now on wasm and filter ordering.

Issue #2571, opened 2024-02-06
Filter Ordering


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

* Some users would like to ratelimit request before authenticating them
* Some would like to first authenticate requests, retrieve claims based on the JWT, and ratelimit based on that

To address such cases, we need to introduce the ability to specify a order of preference in case the default order needs to be overridden (lives in https://github.com/envoyproxy/gateway/blob/3f37499cef2c76589ccbcf83a77fef92257cf5da/internal/xds/translator/httpfilters.go#L87 )

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zhaohuabing:
This is blocking `EnvoyExtensionPolicy`.

arkodg:
closing as completed with https://github.com/envoyproxy/gateway/pull/3273

PR #2991, opened 2024-03-20
feat(translator): Envoy Extension Policy and ExtProc

guydc:
 Initial Implementation of Envoy Extension Policy, supporting:
* Multiple Policy Attachment
* Route-level override
* Producing top-level IR container (EnvoyExtensionFeatures)
* Some reuse of ext-auth utilities

Includes a minimal implementation of the first extension use case: ext-proc. 
- EEP spec only supports a reference to an external processing service
- IR and XDS
- E2E test
- used to demonstrate status management, override behavior and multiple attachment for non-trivial policies

Relates to #2546, #2025, #2877


api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -55,6 +55,10 @@ type EnvoyExtensionPolicySpec struct {
 	//
 	// +optional
 	Priority int32 `json:"priority,omitempty"`
```
zhaohuabing, line 57: We might not need this anymore since the extensions are on a list now.

Should EG doesn't allow multiple ExtensionPolicies on the same target?

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.group) || self.service.backendRef.group == \"\") : true", message="group is invalid, only the core API group (specified by omitting the group field or setting it to an empty string) is supported"
+// +kubebuilder:validation:XValidation:rule="has(self.service) ? (!has(self.service.backendRef.kind) || self.service.backendRef.kind == 'Service') : true", message="kind is invalid, only Service (specified by omitting the kind field or setting it to 'Service') is supported"
+//
+// ExtProc defines the configuration for External Processing filter.
+type ExtProc struct {
```
zhaohuabing, line 16: We may need an `insertBefor` property in this struct to specify the place where this ExtProc extension should be inserted in the filter chain.

https://github.com/envoyproxy/gateway/blob/dbaa9a0ad8c35ac78909de30814cc63b08e0e2c1/api/v1alpha1/wasm_types.go#L42

api/v1alpha1/envoyextensionypolicy_types.go
guydc: I agree that having a list already achieves the main goal of priorities, which is to support multiple extensions on a route. I think that priority gives a bit more "organizational" flexibility, meaning: multiple extensions can be attached by different teams that own different policy resources. 

Of course, it also has the downside of translation being more expensive and structurally less consistent with other policies (BTP, Security, etc.). 

@envoyproxy/gateway-maintainers  - WDYT? Should we:
* Keep priority
* Remove priority
* Treat it as unimplemented

api/v1alpha1/ext_proc_types.go
guydc: Right. I think that this can be tackled in several iterations:

1. Extensions have a built-in location (e.g. after auth)
2. Support "global" filter ordering through: #2571
3. Support GW/Route specific filter ordering through a `InsertBefore`, `stage` or local `filterOrder`. 

For the first iteration, I propose to keep this as-is, and go for (3) when requirements come up.

api/v1alpha1/envoyextensionypolicy_types.go
zhaohuabing: > multiple extensions can be attached by different teams that own different policy resources.

It's possible that multiple teams are handling different extensions. However, I have a few concerns:

Coordination is necessary as these extensions might overlap, leading to issues like modifying the same header or one extension's output affecting another's behavior.
The priority structure is complex, with levels of prioritization among multiple EEPs and within extensions of a single EP.
In my opinion, managing a single EEP would be simpler, promoting better coordination and avoiding potential conflicts.

api/v1alpha1/ext_proc_types.go
```diff
@@ -0,0 +1,29 @@
...
+	// Service defines the configuration of the external processing service
+	Service ExtProcService `json:"service"`
```
arkodg, line 18: can we directly use `BackendRef` here ?

guydc: There are some additional backend-related attributes that we may want to support here in the future. For example, passing initial metadata to authenticate with the external-processing backend, choosing the sort of grpc client (envoy vs. google), etc. In Ext-Auth, we decided to use the a "service" container for this purpose: https://github.com/envoyproxy/gateway/blob/d020ce742bf873052bce590fcb3fb0a73e71ee77/api/v1alpha1/ext_auth_types.go#L66

arkodg: if we want to leave room for expansion, thoughts on ?
```
type ExtProc struct {
......
BackendRef BackendRef 
```

type BackendRef struct {
gwapiv1.BackendObjectReference
.....
}

arkodg: it matches the style adopted in upstream https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendRef where additional fields such as `weight` are added

api/v1alpha1/ext_proc_types.go
zhaohuabing: Thought again, we just need a `Priority` inside the Extension struct to allow users to adjust the location of individual extensions.

arkodg: within the list, we can always use the list order as the inherent order (ascending or descending)

zhaohuabing: Yes, we can use the list order for the priority between extensions, but we would also need an implicit priority field to decide the location of an extension in other filters(jwt, cors, oauth2,etc.)

guydc:
Notes from previous community meeting:
- We RM priority for the time being
- We will support, in a future iteration, the ability to control order on a per-extension-instance basis, similar to the global ordering concept we have in #2571 for built-in extension filters.

api/v1alpha1/ext_proc_types.go
```diff
@@ -15,15 +15,15 @@ import (
...
 	// Service defines the configuration of the external processing service
-	Service ExtProcService `json:"service"`
+	BackendRef ExtProcBackendRef `json:"backendRef"`
 }
 
 // ExtProcService defines the gRPC External Processing service using the envoy grpc client
 // The processing request and response messages are defined in
 // https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ext_proc/v3/external_processor.proto
-type ExtProcService struct {
+type ExtProcBackendRef struct {
```
zhaohuabing, line 24: Do we need this extra layer of ExtProcBackendRef?

internal/gatewayapi/runner/runner.go
```diff
@@ -79,12 +79,13 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
-					GatewayClassName:        v1.ObjectName(resources.GatewayClass.Name),
-					GlobalRateLimitEnabled:  r.EnvoyGateway.RateLimit != nil,
-					EnvoyPatchPolicyEnabled: r.EnvoyGateway.ExtensionAPIs != nil && r.EnvoyGateway.ExtensionAPIs.EnableEnvoyPatchPolicy,
-					Namespace:               r.Namespace,
-					MergeGateways:           gatewayapi.IsMergeGatewaysEnabled(resources),
+					GatewayControllerName:       r.Server.EnvoyGateway.Gateway.ControllerName,
+					GatewayClassName:            v1.ObjectName(resources.GatewayClass.Name),
+					GlobalRateLimitEnabled:      r.EnvoyGateway.RateLimit != nil,
+					EnvoyPatchPolicyEnabled:     r.EnvoyGateway.ExtensionAPIs != nil && r.EnvoyGateway.ExtensionAPIs.EnableEnvoyPatchPolicy,
+					EnvoyExtensionPolicyEnabled: r.EnvoyGateway.ExtensionAPIs != nil && r.EnvoyGateway.ExtensionAPIs.EnableEnvoyExtensionPolicy,
```
zhaohuabing, line 86: We probably don't need this switch?  Envoy extensions like wasm, lua, and ext proc are common requirements.

zhaohuabing (dismissed):
LGTM. Thanks! 🚀

internal/gatewayapi/translator_test.go
```diff
@@ -44,8 +44,9 @@ func mustUnmarshal(t *testing.T, val []byte, out interface{}) {
 
 func TestTranslate(t *testing.T) {
 	testCasesConfig := []struct {
-		name                    string
-		EnvoyPatchPolicyEnabled bool
+		name                        string
+		EnvoyPatchPolicyEnabled     bool
+		EnvoyExtensionPolicyEnabled bool
```
zirain, line 49: nit?

test/e2e/testdata/ext-proc-service.yaml
zirain, line 1: should this be part of manifest?

guydc: in ext auth, we also create the external auth service as a per-test manifest. I guess that if we use this server for additional tests in the future, it would make sense to move it to manifests...

internal/ir/xds.go
```diff
@@ -490,6 +490,8 @@ type HTTPRoute struct {
 	TCPKeepalive *TCPKeepalive `json:"tcpKeepalive,omitempty" yaml:"tcpKeepalive,omitempty"`
 	// Retry settings
 	Retry *Retry `json:"retry,omitempty" yaml:"retry,omitempty"`
+	// EnvoyExtensionFeatures
```
arkodg, line 493: prefer eliminating 1 level of hierarchy in the IR
* adding `ExtProcs []ExtProc` here
* flattening Service and adding those fields into `ExtProc`

internal/provider/kubernetes/predicates_test.go
```diff
@@ -514,6 +514,35 @@ func TestValidateServiceForReconcile(t *testing.T) {
 			service: test.GetService(types.NamespacedName{Name: "ext-auth-http-service"}, nil, nil),
 			expect:  true,
 		},
+		{
+			name: "service referenced by EnvoyExtensionPolicy ExtPrc GRPC service",
```
arkodg, line 518: can you also add one where the service is unrelated

arkodg:
minor comments, overall LGTM, thanks for adding this !

zhaohuabing (approved):

zhaohuabing (approved):

arkodg (approved):
LGTM thanks for building this out !

PR #2993, opened 2024-03-21
API: HTTP Filter ordering

zhaohuabing:
API for HCM HTTP filter ordering. This is a global-level config for all the HTTP listeners in Envoy proxy. If the Listener/HTTPRoute level filter ordering is required in the future, we can define a fine-grained API targeting Listener/HTTPRoute.

This is a blocker for https://github.com/envoyproxy/gateway/pull/2877

Related issue: https://github.com/envoyproxy/gateway/issues/2571


api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,46 @@ type EnvoyProxySpec struct {
...
+type FilterOrder struct {
+	// EnvoyFilters defines the order of filters in the filter chain.
+	// The order of filters in the list is the order in which they will be applied.
+	// +kubebuilder:validation:MinItems=9
+	// +kubebuilder:validation:MaxItems=9
+	EnvoyFilters []EnvoyFilter `json:"envoyFilters"`
 }
 
+// EnvoyFilter defines the type of Envoy HTTP filter.
+type EnvoyFilter string
```
zhaohuabing, line 115: `EnvoyFilter` or `EnvoyHTTPFilter`?

guydc: I think `EnvoyFilter` is fine. It can also include network filters in the future, but we will need to validate that the order makese sense, e.g. network filters before http filters.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,47 @@ type EnvoyProxySpec struct {
...
+// Default order of filters:
+// - envoy.filters.http.cors
+// - envoy.filters.http.ext_authz
+// - envoy.filters.http.basic_authn
+// - envoy.filters.http.oauth2
+// - envoy.filters.http.jwt_authn
+// - envoy.filters.http.fault
+// - envoy.filters.http.local_ratelimit
+// - envoy.filters.http.rate_limit
+// - envoy.filters.http.router
```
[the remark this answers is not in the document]
zhaohuabing: Good point!

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,47 @@ type EnvoyProxySpec struct {
...
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// FilterOrder defines the order of filters in the Envoy proxy's HTTP filter chain.
+	FilterOrder *FilterOrder `json:"filterOrder,omitempty"`
+}
+
+// FilterOrder defines the order of filters in the Envoy proxy's HTTP filter chain.
+// If unspecified, the default order of filters is applied.
+// Default order of filters:
+// - envoy.filters.http.cors
```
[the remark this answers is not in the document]
zhaohuabing: Maybe just the last part, eg, cors, basicAuth, oauth2, etc. ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,45 @@ type EnvoyProxySpec struct {
...
+// - envoy.filters.http.basic_authn
+// - envoy.filters.http.oauth2
+// - envoy.filters.http.jwt_authn
+// - envoy.filters.http.fault
+// - envoy.filters.http.local_ratelimit
+// - envoy.filters.http.rate_limit
+type FilterOrder struct {
+	// EnvoyFilters defines the order of filters in the filter chain.
+	// The order of filters in the list is the order in which they will be applied.
+	// +kubebuilder:validation:MinItems=8
```
guydc, line 108: How would this behave when users upgrade to a new version that includes an additional filter? I assume that we will increase min/max accordingly, which will make the existing end-user configuration invalid, causing things like gitops that applies user-defined resources to fail. Users that choose to change the order of some filters now need to know about all filters. 

Maybe we can avoid min/max here? Some ideas:
- Partial list will only impact ordering of the mentioned filters between themselves
- Instead of filter names, use a tuple of <filterName, filterPriority>, allowing users to control order specifically for filters. Native order will use 100-unit increments (0, 100, 200, ... ) to allow users plenty of room to re-order filters in-between the natively ordered filters

zhaohuabing: > Instead of filter names, use a tuple of <filterName, filterPriority>, allowing users to control order specifically for desired filters. Native order will use 100-unit increments (0, 100, 200, ... ) to allow users plenty of room to re-order filters in-between the natively ordered filters

Vote for this option because users normally only need to adjust one filter or two. They don't have to care about the orders of all filters.

guydc (dismissed):

zhaohuabing:
defer to @arkodg

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,46 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// FilterOrder defines the order of filters in the Envoy proxy's HTTP filter chain.
+	//
+	// +optional
+	FilterOrder []FilterOrder `json:"filterOrder,omitempty"`
```
arkodg, line 93: we are technically assigning priority - `0` , `100` etc, should this be `priority` instead of `order` ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,46 @@ type EnvoyProxySpec struct {
...
+}
+
+// FilterOrder defines the order of filters in the Envoy proxy's HTTP filter chain.
+// The order of filters is determined by the order field, where the filter with
+// the lowest order value is applied first.
+// If unspecified, the default order of filters is applied.
+// Default order of filters:
```
arkodg, line 100: did you pick this up from https://github.com/envoyproxy/go-control-plane/blob/main/pkg/wellknown/wellknown.go ?

arkodg:
I really like how simple this API is, great stuff @zhaohuabing !
my only q is naming - `FIlterOrder` or `FilterPriority` :)

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,46 @@ type EnvoyProxySpec struct {
...
+// - envoy.filters.http.ext_authz          100
+// - envoy.filters.http.basic_authn        200
+// - envoy.filters.http.oauth2             300
+// - envoy.filters.http.jwt_authn          400
+// - envoy.filters.http.fault              500
+// - envoy.filters.http.local_ratelimit    600
+// - envoy.filters.http.rate_limit         700
+type FilterOrder struct {
+	Filter EnvoyFilter `json:"filter"`
+	Order  uint32      `json:"order"`
```
arkodg, line 111: lets make this `int32` (negative) so we can add filters ahead of `cors` w/o moving it

api/v1alpha1/envoyproxy_types.go
zhaohuabing: I'm open to both. My choice may not be the most accurate as a non-native speaker.

zhaohuabing: Thought again, `priority` may be more appropriate because the number defined here is not exactly the `order` of the filter in the chain.

api/v1alpha1/envoyproxy_types.go
zhaohuabing: Yes. :-)

arkodg (dismissed):
LGTM thanks !

site/content/en/latest/api/extension_types.md
```diff
@@ -1126,6 +1138,21 @@ _Appears in:_
 | `path` | _string_ |  true  | Path defines the file path used to expose envoy access log(e.g. /dev/stdout). |
 
 
+#### FilterPriority
+
+
+
+FilterPriority defines the order of filters in the Envoy proxy's HTTP filter chain. The filter with the lower value is put before those with higher values in the filter chain. If unspecified, the default priority of filters is applied. Default priority of filters: - envoy.filters.http.cors               0 - envoy.filters.http.ext_authz          100 - envoy.filters.http.basic_authn        200 - envoy.filters.http.oauth2             300 - envoy.filters.http.jwt_authn          400 - envoy.filters.http.fault              500 - envoy.filters.http.local_ratelimit    600 - envoy.filters.http.ratelimit          700
```
zirain, line 1145: can you check output locally, this seems not very good.

zhaohuabing: Looks like crd-ref-docs can't handle list in comments.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,57 @@ type EnvoyProxySpec struct {
...
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// FilterPriority defines the order of filters in the Envoy proxy's HTTP filter chain.
+	//
+	// +optional
+	FilterPriority []FilterPriority `json:"filterPriority,omitempty"`
+}
+
+// FilterPriority defines the order of filters in the Envoy proxy's HTTP filter chain.
```
arkodg, line 96: the API looks good, but lets take a little more time to make sure if we 
* agree on this default order
* should we leave a gap for any upcoming feature like `rbac` or `compression` filter ?

we cannot change these default numbers going forward

zhaohuabing: > should we leave a gap for any upcoming feature like rbac or compression filter or is the 100 gap enough to handle these future cases ?

We could list all the filters that EG is likely to use.

zhaohuabing:
After discussing with @arkodg and @guydc during this week's EG meeting, let's hold off on this for now to gather more feedback.

@envoyproxy/gateway-maintainers your input on the default orders of the HTTP filters and comments on this API would be greatly appreciated. Thanks!

zhaohuabing:
> > After discussing with @arkodg and @guydc during this week's EG meeting, let's hold off on this for now to gather more feedback.
> > @envoyproxy/gateway-maintainers your input on the default orders of the HTTP filters and comments on this API would be greatly appreciated. Thanks!
> 
> I do think this's a blocker of #2877, a wasm filter can be inserted anywhere in the filter chain.

Just need more input to make sure we do the API right. We will revisit it right after the wasm extension implementation.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,57 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
+
+	// FilterPriority defines the order of filters in the Envoy proxy's HTTP filter chain.
+	//
+	// +optional
+	FilterPriority []FilterPriority `json:"filterPriority,omitempty"`
```
ZackButcher, line 93: We need to clarify the API: as I read it, it's all-or-nothing (you either set a filter order, in which case you need to set _every filter you want executed_, or you get the default) -- is that correct?

I think that's a reasonable API, but it makes simple things ("add a WASM filter before authn, please") as hard as hard things ("totally re-order how policy is applied, including my WASM, ext_proc, ext_authz, and rate limiting"). It's also very copy+paste heavy: you look up the official order and copy+paste it to start editing, and next time you copy+paste your edited version for the new one, and so on. This winds up making it more expensive to iterate on APIs as a project.

Instead it might be easier to have a split API, along the lines of:

```suggestion
        type FilterOrder struct {
                Add    []FilterPriority `json:"add,omitempty"`
                Custom []FilterPriority `json:"custom,omitempty"`
        }
        FilterOrder FilterOrder `json:"filterOrder,omitempty"`
```

Then for small changes a user can `add`, or for total control they can use `replace`:

```yaml
filterOrder:
  add:
    - filter: envoy.filters.http.wasm
      priority: 10
```
```yaml
filterOrder:
  custom:
    - filter: envoy.filters.http.cors
      priority: 901
    - filter: envoy.filters.http.ext_authz
      priority: 801
    - filter: envoy.filters.http.basic_authn
      priority: 701
    - filter: envoy.filters.http.oauth2
      priority: 601
    - filter: envoy.filters.http.jwt_authn
      priority: 501
    - filter: envoy.filters.http.fault
      priority: 401
    - filter: envoy.filters.http.local_ratelimit
      priority: 301
    - filter: envoy.filters.http.ratelimit
      priority: 201
```

And of course we'd expect the two to work together as well:
```yaml
# these are the same
filterOrder:
  add:
    - filter: envoy.filters.http.wasm
      priority: 10
  custom:
    - filter: envoy.filters.http.cors
      priority: 901
    - filter: envoy.filters.http.ext_authz
      priority: 801
    - filter: envoy.filters.http.basic_authn
      priority: 701
    - filter: envoy.filters.http.oauth2
      priority: 601
    - filter: envoy.filters.http.jwt_authn
      priority: 501
    - filter: envoy.filters.http.fault
      priority: 401
    - filter: envoy.filters.http.local_ratelimit
      priority: 301
    - filter: envoy.filters.http.ratelimit
      priority: 201
---
filterOrder:
  custom:
    - filter: envoy.filters.http.cors
      priority: 901
    - filter: envoy.filters.http.ext_authz
      priority: 801
    - filter: envoy.filters.http.basic_authn
      priority: 701
    - filter: envoy.filters.http.oauth2
      priority: 601
    - filter: envoy.filters.http.jwt_authn
      priority: 501
    - filter: envoy.filters.http.fault
      priority: 401
    - filter: envoy.filters.http.local_ratelimit
      priority: 301
    - filter: envoy.filters.http.ratelimit
      priority: 201
    - filter: envoy.filters.http.wasm
      priority: 10
```

zhaohuabing: @ZackButcher It doesn't have to be a full list.

Each filters have a default priority, as the comment shows:
// - envoy.filters.http.cors               0
//
// - envoy.filters.http.ext_authz          100
//
// - envoy.filters.http.basic_authn        200
//
// - envoy.filters.http.oauth2             300
//
// - envoy.filters.http.jwt_authn          400
//
// - envoy.filters.http.fault              500
//
// - envoy.filters.http.local_ratelimit    600
//
// - envoy.filters.http.ratelimit          700


If we want to change the default priority of the oauth2 filter to place it between cors and ext_auth, it can be done with the following config:

```
filterPriority:
  envoy.filters.http.oauth2: 50
```

For custom extensions like wasm, lua, and extproc, we plan to add a priority field in their individual spec, for exampel:

```
  wasm:
    - name: my_plugin_1
      priority: 450
      code:
        image:  # the oci image that contains wasm code
          url: <oci-registry>.foo.bar.com/my_plugin_1:v1 
          pullSecret: wasmPullSecret
      ...
    - name: my_plugin_2
      priority: 250
      code:
        image:  # the oci image that contains wasm code
          url: <oci-registry>.foo.bar.com/my_plugin_2:v1 
          pullSecret: wasmPullSecret
      ...      

```  

my_plugin_1 will be placed between jwt_authn and fault.
my_plugin_2 will be placed between basic_authn and jwt_authn.

The ultimate order of these filters are: cors(0),oauth2(50),basic_authn(200), wasm_my_plugin_2(250),jwt_authn(400), wasm_my_plugin_1(450),fault(500),local_ratelimit(600),ratelimit(700)

arkodg:
blocked on https://github.com/envoyproxy/gateway/issues/3046

arkodg:
hey @zhaohuabing , we brought this up in the community meeting yesterday and @davidalger had a great alternate suggestion - to use `before` and `after` keywords e.g.
```
filterOrder:
- name: envoy.filters.http.local_ratelimit
   before: envoy.filters.http.ext_authz
- name: envoy.filters.http.cors
   after: envoy.filters.http.local_ratelimit
```

this does improve readability of intent. Flattening this intent out, may result in a loop, which needs to be avoided and flagged in the status (but thats an internal detail)

this API can look similar inside the WASM API
```
spec:
.......
  filterOrder:
    before: envoy.filters.http.ext_authz
```

this will require us to rely on priority field in EEP to create order across EEPs

zhaohuabing:
> hey @zhaohuabing , we brought this up in the community meeting yesterday and @davidalger had a great alternate suggestion - to use `before` and `after` keywords e.g.
> 
> ```
> filterOrder:
> - name: envoy.filters.http.local_ratelimit
>    before: envoy.filters.http.ext_authz
> - name: envoy.filters.http.cors
>    after: envoy.filters.http.local_ratelimit
> ```
> 
> this does improve readability of intent. Flattening this intent out, may result in a loop, which needs to be avoided and flagged in the status (but thats an internal detail)
> 
> this API can look similar inside the WASM API
> 
> ```
> spec:
> .......
>   filterOrder:
>     before: envoy.filters.http.ext_authz
> ```
> 
> this will require us to rely on priority field in EEP to create order across EEPs

LGreatTM!

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,69 @@ type EnvoyProxySpec struct {
...
+	// Only one of Before or After must be set.
+	Before *EnvoyFilter `json:"before,omitempty"`
+
+	// After defines the filter that should come after the filter.
+	// Only one of Before or After must be set.
+	After *EnvoyFilter `json:"after,omitempty"`
+}
+
+// EnvoyFilter defines the type of Envoy HTTP filter.
+// +kubebuilder:validation:Enum=envoy.filters.http.cors;envoy.filters.http.ext_authz;envoy.filters.http.basic_authn;envoy.filters.http.oauth2;envoy.filters.http.jwt_authn;envoy.filters.http.fault;envoy.filters.http.local_ratelimit;envoy.filters.http.ratelimit;envoy.filters.http.wasm;envoy.filters.http.ext_proc
```
arkodg, line 135: ❤️

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,69 @@ type EnvoyProxySpec struct {
...
 }
 
+// FilterPosition defines the position of an Envoy HTTP filter in the filter chain.
+type FilterPosition struct {
+	// Name of the filter.
+	Name EnvoyFilter `json:"filter"`
+
+	// Before defines the filter that should come before the filter.
+	// Only one of Before or After must be set.
+	Before *EnvoyFilter `json:"before,omitempty"`
```
arkodg, line 127: CEL to make sure at least `Before` or `After` is set ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -86,8 +86,69 @@ type EnvoyProxySpec struct {
...
+	// Only one of Before or After must be set.
+	After *EnvoyFilter `json:"after,omitempty"`
+}
+
+// EnvoyFilter defines the type of Envoy HTTP filter.
+// +kubebuilder:validation:Enum=envoy.filters.http.cors;envoy.filters.http.ext_authz;envoy.filters.http.basic_authn;envoy.filters.http.oauth2;envoy.filters.http.jwt_authn;envoy.filters.http.fault;envoy.filters.http.local_ratelimit;envoy.filters.http.ratelimit;envoy.filters.http.wasm;envoy.filters.http.ext_proc
+type EnvoyFilter string
+
+const (
+	EnvoyFilterFault          EnvoyFilter = "envoy.filters.http.fault"
```
arkodg, line 139: comments for each of these ? so they appear in docs

arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM

Issue #2976, opened 2024-03-19
ReplacePrefixMatch not working 


zhaohuabing:
```
  - filters:
    - requestRedirect:
        path:
          replacePrefixMatch: /
          type: ReplacePrefixMatch
        port: 8080
        scheme: https
        statusCode: 301
      type: RequestRedirect
    matches:
    - path:
        type: PathPrefix
        value: /api/foo/public/
```        

In the given example we want to simply delete the whole Prefix, replace it with /. However, what we see happening is a double slash:

GET https://REDACTED/api/foo/public/bar
      accept: application/json
      accept-encoding: gzip,deflate,br
    ← 301 Moved Permanently
      transfer-encoding: chunked
      connection: close
      location: https://REDACTED:8080//bar

zhaohuabing:
This issue is because EG uses `path_separated_prefix` instead of `prefix` for prefix match.

The route matches `/api/foo/public`, then replace it with `/`, so `139.178.91.113:8080/api/foo/public/bar` becomes `139.178.91.113:8080//bar`

If we use `prefix`, then this will be solved, but `/api/foo/public` will match both `/api/foo/public/pages` and `/api/foo/publicpages`.  I think this is OK because according to https://github.com/kubernetes-sigs/gateway-api/pull/2055 , the trailing splash is meaningful.

path_separated_prefix
(string) If specified, the route is a path-separated prefix rule meaning that the :path header (without the query string) must either exactly match the path_separated_prefix or have it as a prefix, followed by /
For example, /api/dev would match /api/dev, /api/dev/, /api/dev/v1, and /api/dev?param=true but would not match /api/developer

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routematch-path-separated-prefix

zhaohuabing:
`https://redacted:8080//bar` works anyway because `mergeSlashes` is set to true by EG by default. So the HTTPRoute and backends will see `https://redacted:8080/bar` .

rodrigosiqueiradev:
Hi there, isn't the mergeSlashes created for convenience and not really to spec? Wouldn't it be healthier long-term to handle this situation here at the source instead?

arkodg:
hey @zhaohuabing thoughts on creating 2 xds matches for every IR match ? so we handle the `/foo/bar` and `/foo/bar/` case separately ?

zhaohuabing:
[...]

PR #2998, opened 2024-03-23
Fix: double slashes in redirect URL

zhaohuabing:
This PR fix  [double splashes in the redirect URL
](https://github.com/envoyproxy/gateway/issues/2976)

When an HTTPRoute's match is a path prefix and has a redirect filter with a replacePrefixMatch "/",
A redirect routeaction with  regex_rewrite is generated to match both the prefix with and without trailing "/" to avoid double "/" in the redirect URL.

For example, given the following HTTPRoute:

```
 - matches:
   - path:
       type: PathPrefix
       value: /api/foo/
   filters:
   - requestRedirect:
       path:
         replacePrefixMatch: /
         type: ReplacePrefixMatch
     type: RequestRedirect
```

 The following route will be created:

```
              routes:
              - match:
                  path_separated_prefix: "/api/foo"
                redirect:
                  regex_rewrite:       
                    pattern:
                      regex: "^\/api\/foo\/*"
                    substitution: /
```

See https://github.com/envoyproxy/gateway/issues/2976 for more details.

Fix: https://github.com/envoyproxy/gateway/issues/2976


arkodg:
hey @zhaohuabing I found some code under URLRewrite, https://github.com/envoyproxy/gateway/blob/485edbe9ac3f049d9bd3bfe9c6a36c63351ba127/internal/xds/translator/route.go#L332 , can we reuse this ?

internal/xds/translator/route.go
```diff
@@ -303,6 +312,25 @@ func buildXdsRedirectAction(redirection *ir.Redirect) *routev3.RedirectAction {
 	return routeAction
 }
 
+// useRegexRewriteForPrefixMatchReplace checks if the regex rewrite should be used for prefix match replace
+// due to the issue with Envoy not handling the case of "//" when the replace string is "/".
+// See: https://github.com/envoyproxy/envoy/issues/26055
+func useRegexRewriteForPrefixMatchReplace(pathMatch *ir.StringMatch, prefixMatchReplace string) bool {
+	return pathMatch != nil &&
+		pathMatch.Prefix != nil &&
+		*pathMatch.Prefix != "/" &&
```
arkodg, line 321: this line is new ?

arkodg:
looks good, thanks for reusing the existing code !
minor comment, else LGTM

internal/xds/translator/route.go
zhaohuabing: It's an edge case. We don't need to use prefixMatchReplace if the patchMatch prefix is "/".

```
            virtual_hosts:
            - name: service  
              domains:
              - "*" 
              routes:
              - match:
                  prefix: "/"
                redirect:
                  prefixRewrite: "/"
```

arkodg (approved):
LGTM thanks for fixing this quirk !
and also adding e2es 🚀

zirain (approved):

Issue #2953, opened 2024-03-18
CORS: Enable strict CORS enforcement with direct preflight response option


zhaohuabing:
The current `SecurityPolicy` CORS implementation forwards preflight requests down to the backend services if a request doesn't match its configured condition(origins, methods, headers, etc.) .  This makes the preflight response unpredictable, potentially allowing cross-origin requests even when they should be disallowed. 

Adding a configuration option to the CORS configuration, enabling direct preflight responses, would allow enforcement of strict CORS rules regardless of the backend behavior.

```
    cors:
      directResponse: true             // Send a direct response and disallow the cross-origin request when the request doesn't match
      allowOrigins:
      - "http://*.example.com"
      - "http://foo.bar.com"
      - "https://*"
      allowMethods:
      - GET
      - POST
      allowHeaders:
      - "x-header-1"
      - "x-header-2"
      exposeHeaders:
      - "x-header-3"
      - "x-header-4"
    
```

This depends on the upstream issue: https://github.com/envoyproxy/envoy/issues/14233

PR #3002, opened 2024-03-24
fix: do nof forward not matching prefilights to backends

zhaohuabing:
fix: https://github.com/envoyproxy/gateway/issues/2953

This PR sets the `ForwardNotMatchingPreflights` to false, allowing the CORS filter to directly respond to a CORS preflight request instead of routing it to the backend.

This change modifies the default behavior, but it aligns more closely with users'  intuition. We can expose this option to CORS API if sending to backends is required in the future.


arkodg (approved):
🚀

zirain (approved):

Issue #2880, opened 2024-03-11
Adding BTP support for TCPRoute


jakelevirne:
*Description*:
I'd like to apply a BackendTrafficPolicy to to a TCPRoute. The use case is for load balancing traffic to an MQTT broker (VerneMQ) which is run as a Service in my cluster backed by a StatefulSet with 3 replicas. I'm running a bare metal K8s cluster (microk8s) on Raspberry Pi (arm64) nodes.

From the VerneMQ [docs](https://docs.vernemq.com/guides/vernemq-on-kubernetes), the recommended approach is to use an ingress controller or external load balancer that can route TCP traffic, load balance based on source ip hashing, and provide TLS termination.

I wanted to try envoy gateway to see if it could fit the bill. But from a [Slack convo](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1710170946381669) with @arkodg  it seems like BackendTrafficPolicy does not yet apply to TCPRoutes. I tried targeting the policy to the Gateway and then to the TCPRoute but neither seemed to work. 

[optional *Relevant Links*:]
Gateway:
```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: vernemq-gateway
  namespace: vernemq
spec:
  gatewayClassName: envoy-gateway
  listeners:
  - name: vernemq-listener
    protocol: TCP
    port: 1883
    allowedRoutes:
      kinds:
      - kind: TCPRoute
```

TCPRoute:
```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: TCPRoute
metadata:
  name: vernemq-route
  namespace: vernemq
spec:
  parentRefs:
  - name: vernemq-gateway
    namespace: vernemq
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: my-vernemq
      port: 1883

```

BackendTrafficPolicy:
```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy
metadata:
  name: vernemq-backend-traffic-policy
  namespace: vernemq
spec:
# targetRef:
#    group: gateway.networking.k8s.io
#    kind: Gateway
#    name: vernemq-gateway
#    namespace: vernemq
  targetRef:
    group: gateway.networking.k8s.io
    kind: TCPRoute
    name: vernemq-route
    namespace: vernemq
  loadBalancer:
    type: ConsistentHash
    consistentHash:
      type: SourceIP

```

[...]

PR #3004, opened 2024-03-24
feat(policy): Adding BTP support for UDP/TCPRoute

alexwo:
Enables the application of BTP rules on gateway and TCP/UDP routes. 
Modifications have been incorporated into the IR mapper and XDS translator.

Adds below BTP rules for tcp and tls routes:
- LoadBalancer
- Timeout
- TCPKeepalive
- CircuitBreaker
- HealthCheck
- ProxyProtcol

Adds below BTP rules for UDP routes:
- LoadBalancer
- Timeout

https://github.com/envoyproxy/gateway/issues/2880


zhaohuabing:
Should we enforce BTP setting validation on TCP/UDP route and update the status? 

For example, HTTP-related settings such as `FaultInjection`, `Retries`, etc., can't be set when the targeting xRoute is TCP/UDP route.

JuniorJPDJ:
Can I also please ask to add TLSRoute support too?

alexwo:
> Can I also please ask to add TLSRoute support too?

TLS route is translated to a tcp listner

> Should we enforce BTP setting validation on TCP/UDP route and update the status?
> 
> For example, HTTP-related settings such as `FaultInjection`, `Retries`, etc., can't be set when the targeting xRoute is TCP/UDP route or the targeting gateway doesn't contain HTTPRoute.



> Can I also please ask to add TLSRoute support too?

Sure, I think that's alread

> Can I also please ask to add TLSRoute support too?

This is already included in the current pull request. Once merged, you can apply BTP either to specific TLS routes or to the entire gateway.

alexwo:
> Should we enforce BTP setting validation on TCP/UDP route and update the status?
> 
> For example, HTTP-related settings such as `FaultInjection`, `Retries`, etc., can't be set when the targeting xRoute is TCP/UDP route or the targeting gateway doesn't contain HTTPRoute.

Sounds good to me. Alternatively, we could update the status of each route to indicate which policies from the BTP are in effect, providing clarity to the user on how their route is affected. 

Since this functionality isn't available currently, it might be best to introduce it through a separate change. 
WDYT?

zhaohuabing:
> > Should we enforce BTP setting validation on TCP/UDP route and update the status?
> > For example, HTTP-related settings such as `FaultInjection`, `Retries`, etc., can't be set when the targeting xRoute is TCP/UDP route or the targeting gateway doesn't contain HTTPRoute.
> 
> Sounds good to me. Alternatively, we could update the status of each route to indicate which policies from the BTP are in effect, providing clarity to the user on how their route is affected.
> 
> Since this functionality isn't available currently, it might be best to introduce it through a separate change. WDYT?

Sounds good. Could you please raise an issue to track this?

arkodg:
hey @zhaohuabing we can't enforce validation here, because if the policy is applied to the Gateway, they could apply to `HTTPRoute`s as well as `TCPRoute`s

internal/xds/translator/listener.go
```diff
@@ -344,6 +344,25 @@ func findXdsHTTPRouteConfigName(xdsListener *listenerv3.Listener) string {
 	return ""
 }
 
+func buildTCPProxyHashPolicy(lb *ir.LoadBalancer) []*typev3.HashPolicy {
```
arkodg, line 347: can we reuse existing code from http ?

internal/xds/translator/testdata/out/xds-ir/listener-tcp-keepalive.clusters.yaml
```diff
@@ -66,3 +68,6 @@
   outlierDetection: {}
   perConnectionBufferLimitBytes: 32768
   type: EDS
+  upstreamConnectionOptions:
```
arkodg, line 71: the test is for listener side keep alive, how did this config come here (cluster) ?

internal/xds/translator/cluster.go
```diff
@@ -599,7 +599,7 @@ func buildXdsClusterUpstreamOptions(tcpkeepalive *ir.TCPKeepalive) *clusterv3.Up
 		ka.TcpKeepalive.KeepaliveProbes = wrapperspb.UInt32(*tcpkeepalive.Probes)
 	}
 
-	if tcpkeepalive.Probes != nil {
+	if tcpkeepalive.IdleTime != nil {
```
arkodg, line 602: thanks for catching this

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -427,6 +452,43 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
...
+			return errors.Wrap(err, "Timeout")
+		}
+	}
+
+	for _, tcp := range ir.TCP {
+		gatewayName := tcp.Name[0:strings.LastIndex(tcp.Name, "/")]
+		if t.MergeGateways && gatewayName != policyTarget {
+			continue
+		}
+
```
arkodg, line 467: you probably also need some code similar to L500-510
```
			// If any of the features are already set, it means that a more specific
			// policy(targeting xRoute) has already set it, so we skip it.
			// TODO: zhaohuabing group the features into a struct and check if all of them are set
			if r.RateLimit != nil || r.LoadBalancer != nil ||
				r.ProxyProtocol != nil || r.HealthCheck != nil ||
				r.CircuitBreaker != nil || r.FaultInjection != nil ||
				r.TCPKeepalive != nil || r.Retry != nil ||
				r.Timeout != nil {
				continue
			}
```

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -427,6 +452,43 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(policy *egv1a1.Back
...
+		}
+	}
+
+	for _, udp := range ir.UDP {
+		gatewayName := udp.Name[0:strings.LastIndex(udp.Name, "/")]
+		if t.MergeGateways && gatewayName != policyTarget {
+			continue
+		}
+
+		udp.LoadBalancer = lb
```
arkodg, line 485: same as above comment

arkodg:
thanks for adding support for this @alexwo !

zhaohuabing:
> hey @zhaohuabing we can't enforce validation here, because if the policy is applied to the Gateway, they could apply to `HTTPRoute`s as well as `TCPRoute`s

Should we enforce validation when the target is a xRoute? It would be confusing if HTTP related setting is configured for TCP/UDP route.

And it would be nice if we could allow users to know which settings in the BTP can be applied to which type of route. Probably add some comments?

internal/xds/translator/listener.go
alexwo: There isn't a significant amount of code that can be shared between the two functions since they initialize distinct types of HashPolicy resources. We can consolidate them into a single buildHashPolicy function, however it will lead to varying return types and likely necessitate some type casting.

arkodg: cool, lets stick to this

internal/gatewayapi/backendtrafficpolicy.go
alexwo: fixed

internal/gatewayapi/backendtrafficpolicy.go
alexwo: fixed

arkodg (approved):
LGTM thanks !

internal/xds/translator/testdata/out/xds-ir/listener-tcp-keepalive.clusters.yaml
arkodg: can you raise a follow up issue for this ?

guydc (approved):

guydc (approved):
LGTM, thanks for fixing the coverage check!

arkodg (approved):

JuniorJPDJ:
I just migrated to the latest commit and it works with TLSRoute! Thanks!

Issue #3016, opened 2024-03-25
can't configure pathPrefix for extAuth


tgunsch:
*Description*:
Hi,
We try to integrate our external auth service with envoy gateway. The external auth service has different endpoints (e.g. `/auth_api` for API's), so we want to configure a `SecurityPolicy` which routes to a specific path and not the root path. 

Unfortunately we can't find how to configure envoy gateway to do a request on a non-root path to the external auth service. Maybe we simply miss something ?

We tried to configure `spec.extAuth.http.path: /auth_api` to the `SecurityPolicy`, but this gives the following envoy config:
```json
    "httpService": {
      "serverUri": {
        "cluster": "securitypolicy/default/ext-auth-example/default/http-ext-auth",
        "timeout": "10s",
        "uri": "http://http-ext-auth.default:9002/auth_api"
      }
    }
```
and the path then seems to be ignored, i.e. when doing a request to  ` http://localhost:8888/myapp`, envoy do an external auth request to `http://http-ext-auth.default:9002/myapp` which ignores the path `/auth_api`

envoy support the configuration of `pathPrefix` for `ExtAuthz`, so when the following is configured (on istio this is done by configure `envoyExtAuthzHttp.pathPrefix`):

```json
"httpService": {
      "pathPrefix": "/auth_ui",
      "serverUri": {
        "cluster": "securitypolicy/default/ext-auth-example/default/http-ext-auth",
        "timeout": "10s",
        "uri": "http://http-ext-auth.default:9002"
      }
    },
```
and a request to ` http://localhost:8888/myapp` is done, then envoy do an external auth request to `http://http-ext-auth.default:9002/auth_ui/myapp`.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.
* Envoy docs: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-httpservice
* Istio ExternalAuth config with pathPrefix: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationHttpProvider

PR #3018, opened 2024-03-26
fix: set path prefix for http ext auth service

zhaohuabing:
fix: [can't configure pathPrefix for extAuth.](https://github.com/envoyproxy/gateway/issues/3016)


arkodg (approved):

arkodg:
can we also add this into e2e ?

zhaohuabing:
> can we also add this into e2e ?

The mock HTTP ext auth service for e2e doesn't care about path :-(

https://github.com/envoyproxy/gateway/blob/2882b7c65dee0ec266c948fd4cba3493f15cf220/test/e2e/testdata/ext-auth-http-service.yaml#L18-L32

PR #3019, opened 2024-03-26
refactor: group xds security features for security policy

shawnh2:
**What type of PR is this?**

related: #2954, separated from #2926


internal/ir/xds.go
```diff
@@ -475,6 +461,59 @@ type HTTPRoute struct {
...
+	}
+
+	return s.CORS != nil ||
+		s.JWT != nil ||
+		s.OIDC != nil ||
+		s.BasicAuth != nil ||
+		s.ExtAuth != nil
+}
+
+func (s *SecurityFeatures) Printable() {
```
zhaohuabing, line 497: Instead of modifying the fields directly, could we return a copy of SecurityFeatures here?

This approach is clearer and a bit easier to read.

internal/ir/xds.go
```diff
@@ -475,6 +461,59 @@ type HTTPRoute struct {
...
+	// OIDC defines the schema for authenticating HTTP requests using OpenID Connect (OIDC).
+	OIDC *OIDC `json:"oidc,omitempty" yaml:"oidc,omitempty"`
+	// BasicAuth defines the schema for the HTTP Basic Authentication.
+	BasicAuth *BasicAuth `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
+	// ExtAuth defines the schema for the external authorization.
+	ExtAuth *ExtAuth `json:"extAuth,omitempty" yaml:"extAuth,omitempty"`
+}
+
+// Any returns true if any of the features are already set.
+func (s *SecurityFeatures) Any() bool {
```
zhaohuabing, line 485: Nit: Any is a bit ambiguous, slightly prefer `Empty`, and uses `! Empty()` to tell if any of the features have been set.

zhaohuabing (commented):
Thanks for the refactoring. This looks good, just a few nits.

zhaohuabing (dismissed):
LGTM. Thanks!

zhaohuabing:
@shawnh2  Apologies for the delay on this PR. Can you fix the conflicts so we can move on?

shawnh2:
> @shawnh2 Apologies for the delay on this PR. Can you fix the conflicts so we can move on?

Resolved, thanks for reviewing this.

shawnh2:
[TestE2E/GatewayInfraResourceTest/create_gateway](https://github.com/envoyproxy/gateway/actions/runs/8629539473/job/23654546301?pr=3019#step:6:2950) seems a liitle unstable.

zhaohuabing (approved):
LGTM, thanks!

internal/ir/xds.go
```diff
@@ -495,6 +481,61 @@ type HTTPRoute struct {
...
+	ExtAuth *ExtAuth `json:"extAuth,omitempty" yaml:"extAuth,omitempty"`
+}
+
+// Empty returns true if all the features are not set.
+func (s *SecurityFeatures) Empty() bool {
+	if s == nil {
+		return true
+	}
+
+	return s.CORS == nil &&
```
arkodg, line 510: will we ever hit this case ? where s != nil, but all sub fields are nil ?

arkodg:
this looks good @shawnh2 ! sorry for the delay in reviewing it
have one minor/curious comment reg `Empty()` , else LGTM !

internal/ir/xds.go
shawnh2: we may very unlikely to hit the case that all fields are nil, but we may hit the case that only some (if any) of the fields are set, indicates `s` is not Empty, return false. This is exactly what we used for this method, as you can see in `translateSecurityPolicyForGateway` method.

I took the @zhaohuabing's suggestion, make this method called `Empty`, instead of `Any` (or other name), so it will be very comprehensible for anyone who read it.

arkodg: non blocking comment - I personally feel this is not needed, and `s != nil` is enough w/o the need for a `Empty` func

arkodg (approved):

internal/ir/xds.go
shawnh2: I will try this out, if it's working w/o Empty method, I will refactor this in the follow-up PR along with the refactor of BTP.

PR #3030, opened 2024-03-26
feat: added option to specify resources on the OIDC security policy

### LINKED ISSUE #3036: Support configuration of resources in SecurityPolicys OIDC filter
In order to request a certain resource from the authorization server a [resource](https://www.rfc-editor.org/rfc/rfc8707.html#name-resource-parameter) can be specified in the OIDC authentication request. This is a feature which envoy itself already supports in its [OIDC filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/oauth2_filter) and it just has to be made configurable via envoy gateways `SecurityPolicy`.

### Concrete use case:

The [kubernets/dashboard](https://github.com/kubernetes/dashboard) is a web front-end which enables cluster administrators to explore the cluster contents in a web browser. To authenticate, a valid bearer token has to be specified in each request to the dashboard. Managed clusters from cloud providers are often integrated with some IdP solution (for example [EntraID formerly known as AAD](https://www.microsoft.com/en/security/business/identity-access/microsoft-entra-id)), which means the kubernets API accepts tokens issued from the IdP. There are various guides on the internet using the [oauth2-proxy/oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) to achive this for the dashboard. In those cases the oauth2-proxy requests tokens from the IdP and sends them along with upstream requests to the dashboard, seamlessly enabling authentication for it. But with envoy gateway there is no need to use this additional service as envoy already has everything on board to do this. However, in the particular case of [AKS](https://azure.microsoft.com/de-de/products/kubernetes-service), a `resource` has to be specified when requesting such a token from the IdP, which the `SecurityPolicy` currently does not support.

I have already implemented this and would be happy to see this merged.

jaynis:
This PR adds the option to specify resources on the OIDC security policy in order to request them in the OIDC authorization request.

resolves #3036


arkodg:
hey @jaynis suggest creating an issue first, sharing the use case

jaynis:
Sure I can do that but this PR just extends the functionality of the gateway to a feature which envoy itself already has and which therefore can be assumed to have valid use cases I would say :smiley: .

arkodg (approved):
LGTM thanks !

cnvergence (approved):

zhaohuabing (approved):
Thanks for adding this useful feature!

Issue #2969, opened 2024-03-18
Run Envoy Gateway like DaemonSet


zvlb:
*Description*:
For many use-cases need to start Kubernetes API Gateway like DaemonSet on special nodes.

We discuss it in [Slack](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1710519615260429), and I found request of this feature in [issue](https://github.com/envoyproxy/gateway/issues/1393#issuecomment-1979046742).

To start the discussion, I can suggest API-scheme for [EnvoyProxyKubernetesProvider](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/envoyproxy_types.go#L139) in [EnvoyProxy](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/envoyproxy_types.go) (with Kubernetes Provider).

### I propose the following architecture for implementation:

Now it's looks:

```go
type EnvoyProxyKubernetesProvider struct {
	EnvoyDeployment *KubernetesDeploymentSpec `json:"envoyDeployment,omitempty"`

	EnvoyService *KubernetesServiceSpec `json:"envoyService,omitempty"`

	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
}
```

I think we can add field `type` and field for EnvoyDaemonset:

```go
type KubernetesProviderType string

type EnvoyProxyKubernetesProvider struct {
        // By default - Deployment.
        Type KubernetesProviderType `json:"type,omitempty"`

	EnvoyDeployment *KubernetesDeploymentSpec `json:"envoyDeployment,omitempty"`

        EnvoyDaemonset *KubernetesDaemonsetSpec `json:"envoyDaemonset,omitempty"`

         // Need add logic to NOT create service
	EnvoyService *KubernetesServiceSpec `json:"envoyService,omitempty"`

        // Don't work, if type = Daemonset
	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
}
```

And:

```go
// KubernetesDaemonsetSpec defines the desired state of the Kubernetes daemonset resource.
type KubernetesDaemonsetSpec struct {
	Patch *KubernetesPatchSpec `json:"patch,omitempty"`

	Pod *KubernetesPodSpec `json:"pod,omitempty"`

	Container *KubernetesContainerSpec `json:"container,omitempty"`

[...]

PR #3031, opened 2024-03-26
Run Envoy Gateway like DaemonSet 

zvlb:
Issue - https://github.com/envoyproxy/gateway/issues/2969

I decided not to wait for architecture discussions and implemented the functionality.

The issue suggested using the [OpenTelemetry Operator](https://github.com/open-telemetry/opentelemetry-operator/blob/eaf998f2ae9ca93ecf319d6864690a2bd7cbe0c7/apis/v1beta1/opentelemetrycollector_types.go#L88) structure for implementing the functionality. However, if we had followed the OpenTelemetry approach, we would have lost backward compatibility.

In the implemented functionality, the "old" logic remains intact, and the new logic is added for configuring Envoy Proxy as a DaemonSet.

A couple of points:

1. I haven't implemented the DaemonSet logic for EnvoyGateway, only for Gateway. (Here you can see a [placeholder](https://github.com/envoyproxy/gateway/pull/3031/commits/b80c5172915732eb18af8c39c6d99df4608f733b#diff-88e7c92f3bb0e20b1cc9fcb0bd4170cbfccad25d9759b09f42413e62f30064a1)). I hope this is not urgent and not needed in the near future (we can create a new task to implement this logic).
2. I have only written tests for envoyService (if it doesn't need to be brought up). I'll write tests for the main functionality as soon as it's confirmed that everything is okay and this functionality is accepted.
3. I'll write documentation after confirmation of the functionality as well, too.


api/v1alpha1/shared_types.go
```diff
@@ -92,6 +105,36 @@ type KubernetesDeploymentSpec struct {
...
 }
 
+// KubernetesDaemonsetSpec defines the desired state of the Kubernetes daemonset resource.
+type KubernetesDaemonSetSpec struct {
+	// Patch defines how to perform the patch operation to daemonset
+	//
+	// +optional
+	Patch *KubernetesPatchSpec `json:"patch,omitempty"`
+
+	// The daemonset strategy to use to replace existing pods with new ones.
```
arkodg, line 115: for your use case, do you need all these fields - pod, container, initcontainers ?

api/v1alpha1/shared_types.go
```diff
@@ -221,6 +271,11 @@ const (
 // +kubebuilder:validation:XValidation:message="allocateLoadBalancerNodePorts can only be set for LoadBalancer type",rule="!has(self.allocateLoadBalancerNodePorts) || self.type == 'LoadBalancer'"
 // +kubebuilder:validation:XValidation:message="loadBalancerIP can only be set for LoadBalancer type",rule="!has(self.loadBalancerIP) || self.type == 'LoadBalancer'"
 type KubernetesServiceSpec struct {
+	// Disable disables service if set to true.
+	// By default, the service is enabled.
```
arkodg, line 275: curious, why this is needed

api/v1alpha1/envoyproxy_types.go
```diff
@@ -137,13 +137,25 @@ type ShutdownConfig struct {
 // EnvoyProxyKubernetesProvider defines configuration for the Kubernetes resource
 // provider.
 type EnvoyProxyKubernetesProvider struct {
+	// Mode is the type of the Kubernetes resource provider. If unspecified, it defaults to "Deployment".
```
arkodg, line 140: do we need an explicit mode field or can we use CEL validation to make sure 
* if `Deployment` is set, then `DaemonSet` cannot be set

api/v1alpha1/shared_types.go
zvlb: In my case I test many cases, but most important:
```
kind: EnvoyProxy
metadata:
  name: envoy-proxy-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      mode: DaemonSet
      envoyService:
        disable: true
      envoyDaemonSet:
        pod:
          hostNetwork: true
          tolerations:
          - key: node-role.kubernetes.io/control-plane
            operator: Exists
          nodeSelector:
            node-role.kubernetes.io/control-plane: ""
```

Start Envoy Proxy like daemonset on special nodes with hostnetwork=true

May be for another cases someone need this?

api/v1alpha1/shared_types.go
```diff
@@ -147,6 +190,13 @@ type KubernetesPodSpec struct {
 	//
 	// +optional
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// Host networking requested for this pod. Use the host's network namespace.
+	// If this option is set, the ports that will be used must be specified.
+	// Default to false.
+	//
+	// +optional
+	HostNetwork bool `json:"hostNetwork,omitempty"`
```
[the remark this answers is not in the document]
zvlb: Why? It's what I realy need for my case. 
`patch` - not very comfortable

api/v1alpha1/shared_types.go
zvlb: Bc if I run EnvoyProxy like daemonSet with hostNetwork - I don't need service

api/v1alpha1/envoyproxy_types.go
zvlb: If in future someone add tag `+kubebuilder:default:="****"` to field in `KubernetesDeploymentSpec` or `KubernetesDaemonSetSpec`, then field `envoyDaemonSet` or `envoyDaemonSet`will not be empty. How to implement this case?

+ the implicit behavior is a bit confusing

zvlb:
Here's a more detailed description of your case:

I have bare-metal installations of Kubernetes clusters. Since the ServiceType LoadBalancer is not supported in my setup, I need to deploy Envoy Proxy on specific nodes to route traffic to pods.

arkodg:
Im not sure its going to work the way you've intended to
* there's a servicePort/ gateway listener port - containerPort mapping
* you will also need to set dns policy `ClusterFirstWithHostNet` 
more in https://github.com/envoyproxy/gateway/pull/2309
I'll defer to @zzjin @Xunzhuo for more details

api/v1alpha1/envoyproxy_types.go
zirain: will that be rejected by validating webhook?

zirain:
please kindly proposal API first, otherwise there will be a lot of waste.

api/v1alpha1/envoyproxy_types.go
zvlb: I'm not sure I understand what you're referring to. Does Envoy Gateway implement an Admission (Validation) webhook?

zirain: EG use [CEL](https://kubernetes.io/docs/reference/using-api/cel/)

zvlb:
Let's decide how to ultimately implement the necessary functionality and resolve contentious issues:

1. @arkodg believes that using the hostNetwork field is undesirable and suggests implementing this functionality through patch. I'm not sure if this is a good solution because it seems like we're providing users with functionality to manage the hostNetwork field, but doing so inconveniently. Considering there's a real use case for when this is needed, I see no reason not to expose this field.

2. @arkodg also suggested that we should expose the dnsPolicy field as controlled fields to set the value ClusterFirstWithHostNet for DaemonSets.

3. @arkodg сlarified whether we need to expose the container and initContainers fields for DaemonSet control. For my use case, this isn't necessary, but during implementation, I tried to maintain all the flexibility of DaemonSet configuration so that it works the same as Deployment.

Perhaps you want a different implementation for DaemonSet. I'm ready to listen to all proposals and implement what is approved.

arkodg:
> I will create 3 new PR for this. It's okay?

sure, but reviewers will most likely review in above order since imo they range from must have to good to have

api/v1alpha1/envoyproxy_types.go
zvlb: @arkodg Can you help here? Which kubebuilder tag need you for this logic?

arkodg: 1. use a ptr type called `mode` to decide b/w Deployment and DaemonSet, defaulting to Deployment
2. use CEL to make sure when `mode` is set to `DaemonSet` , user cannot populate the deployment field e.g. https://github.com/envoyproxy/gateway/blob/0f87fcbad1aaa2913af3f768da87ea72a695a14d/api/v1alpha1/ext_auth_types.go#L13

zvlb: > do we need an explicit mode field or can we use CEL validation to make sure

May be we don't need `mode`? Only `EnvoyDeployment` and `EnvoyDaemonSet` field with validation. If not set - use Deployment (with default preset)

arkodg: > > do we need an explicit mode field or can we use CEL validation to make sure
> 
> May be we don't need `mode`? Only `EnvoyDeployment` and `EnvoyDaemonSet` fields with validation. If not set - use Deployment (with default preset)

I prefer this approach

arkodg:
@zvlb suggest running `make lint`, `make test`, `make generate`  to repro the errors locally

PR #3037, opened 2024-03-27
test: supplementary metrics test

ShyunnY:
**What type of PR is this?**

test: The original metrics test is too simple, I have enhanced it

### DIFF
--- a/internal/metrics/sample_counter_test.go
+++ b/internal/metrics/sample_counter_test.go
@@ -5,19 +5,89 @@
 
 package metrics_test
 
-import "github.com/envoyproxy/gateway/internal/metrics"
+import (
+	"context"
+	"io"
+	"os"
+	"testing"
 
-var (
-	irUpdates = metrics.NewCounter(
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/stdout/stdoutmetric"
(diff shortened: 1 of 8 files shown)


ShyunnY:
+ Regarding the lint issue of yaml, we seem to be able to ignore it.
+ Regarding the problem of gomod, I can't solve it at the moment

zirain (approved):

Xunzhuo (approved):
Thanks for improving it!

internal/metrics/sample_counter_test.go
```diff
@@ -5,19 +5,92 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+
+	"github.com/envoyproxy/gateway/internal/metrics"
+)
+
+func TestCounter(t *testing.T) {
+
+	writer := os.Stdout
+	if true {
```
zirain, line 29: @ShyunnY why always true?

ShyunnY: Since the metric carries a timestamp, it needs to be written before each test is executed. (So explicitly make it true)

zirain: if always true, you shuld just remove this line?

zirain: BTW, these file will be updated every time you run the test, so you'd better improve it.

ShyunnY: Thanks for the suggestions, I will improve this soon, thank you again!

PR #3042, opened 2024-03-27
api: support IdleTimeout in clientTrafficPolicy

### LINKED ISSUE #2861: feat: add support to configuring downstream idle timeout
*Description*:
Add `idleTimeout` to clientTrafficPolicy.

Currently, the downstream-connection idle timeout is fixed at one hour, unless it is less than the request timeout.
I propose adding an idleTimeout parameter to the clientTrafficPolicy configuration. This parameter would allow users to specify the duration of the downstream-connection idle timeout according to their specific requirements.

Rationale:
1. **Prevention of Load Balancer Issues:** When running a proxy in an IaaS, the load balancer of that IaaS might have an idle timeout configured (e.g., GCP with an idle timeout of 600 seconds: [documentation]( https://cloud.google.com/load-balancing/docs/https/#timeouts_and_retries:~:text=Application%20Load%20Balancer%3A-,600%20seconds,-2)). 
If the idle timeout of the LB is reached, the LB closes the connection silently without notifying the proxy, causing following requests to fail. Having a configurable downstream idle timeout can prevent this issue by ensuring that Envoy's idle timeout aligns with or is shorter than the LB’s idle timeout.

2. **Saving Resources:** Configuring the downstream idle timeout helps save resources by reducing the amount needed to maintain connections.

3. **Drain Process Optimization:** In cases where Envoy needs to be drained, having a lower idle timeout enables faster termination of idle connections and speeding up the release of resources.

yaelSchechter:
**What type of PR is this?**
Support idle timeout in ClientTrafficPolicy. 

Fixes #2861


api/v1alpha1/timeout_types.go
```diff
@@ -50,9 +50,15 @@ type ClientTimeout struct {
...
 type HTTPClientTimeout struct {
-	// The duration envoy waits for the complete request reception. This timer starts upon request
+	// RequestReceivedTimeout is the duration envoy waits for the complete request reception. This timer starts upon request
 	// initiation and stops when either the last byte of the request is sent upstream or when the response begins.
 	//
 	// +optional
 	RequestReceivedTimeout *gwapiv1.Duration `json:"requestReceivedTimeout,omitempty"`
+
+	// The idle timeout for an HTTP connection. Idle time is defined as a period in which there are no active requests in the connection.
+	// Default: 1 hour.
```
zhaohuabing, line 60: Is there any particular reason the default value deviates from Envoy upstream?

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto

yaelSchechter: According to envoy documentation, the default is an hour: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-msg-config-core-v3-httpprotocoloptions
I guess that they just put different examples in your link

api/v1alpha1/timeout_types.go
```diff
@@ -50,9 +50,15 @@ type ClientTimeout struct {
...
 
 type HTTPClientTimeout struct {
-	// The duration envoy waits for the complete request reception. This timer starts upon request
+	// RequestReceivedTimeout is the duration envoy waits for the complete request reception. This timer starts upon request
 	// initiation and stops when either the last byte of the request is sent upstream or when the response begins.
 	//
 	// +optional
 	RequestReceivedTimeout *gwapiv1.Duration `json:"requestReceivedTimeout,omitempty"`
+
+	// The idle timeout for an HTTP connection. Idle time is defined as a period in which there are no active requests in the connection.
```
arkodg, line 59:
```suggestion
	// IdleTimeout for an HTTP connection. Idle time is defined as a period in which there are no active requests in the connection.
```

arkodg:
one nit to fix godoc strings, else looks good !

api/v1alpha1/timeout_types.go
yaelSchechter: thanks, fixed

arkodg (dismissed):
🚀

api/v1alpha1/timeout_types.go
zhaohuabing: Never mind. I looked into the wrong place (stream timeout instead of connection timeout)

guydc (approved):

arkodg (approved):

PR #3052, opened 2024-03-28
feat: HeadersWithUnderscoreActions on ClientTrafficPolicy

davidalger:
**What type of PR is this?**

Adds configuration to `ClientTrafficPolicy` to support overriding the default for `HeadersWithUnderscoreActions` in common HTTP protocol options.

Envoy defaults have `headers_with_underscores_action` set to `ALLOW` while Envoy Gateway defaults to `REJECT_REQUEST` despite the underscore character being allowed in header names by RFC-7230. This may be an ok default for EG to be more strict out of the box, but it breaks compatibility with preexisting API client configurations which send underscores in header names.


Xunzhuo (approved):

internal/xds/translator/listener.go
```diff
@@ -236,7 +236,7 @@ func (t *Translator) addXdsHTTPFilterChain(xdsListener *listenerv3.Listener, irL
 		MergeSlashes:                 irListener.Path.MergeSlashes,
 		PathWithEscapedSlashesAction: translateEscapePath(irListener.Path.EscapedSlashesAction),
 		CommonHttpProtocolOptions: &corev3.HttpProtocolOptions{
-			HeadersWithUnderscoresAction: corev3.HttpProtocolOptions_REJECT_REQUEST,
+			HeadersWithUnderscoresAction: translateHeadersWithUnderscoresAction(irListener.Headers),
```
arkodg, line 239: nit: this package seems to use `build` prefix over `translate`

arkodg (dismissed):
LGTM thanks !

arkodg:
looks like a rebase is needed @davidalger

arkodg (approved):

Issue #3041, opened 2024-03-27
Model OpenTelelemetry Sinks as a BackendRef


arkodg:
*Description*:
>Describe the issue.

- [x]  Add BackendRef field to all proxy OpenTelemetry fields (in Proxy Tracing and Metrics And Access Logs) in v1.1
- [x]  Deprecate `Protocol`, `Host`, `Port` in v1.1
- [ ]  Remove above fields in v1.2

Background

Currently the sink has fields such as `Host`, `Port`

* In EnvoyProxy config - https://github.com/envoyproxy/gateway/blob/f699edf3663f2b7e5a5d37756403e1551a7d3976/api/v1alpha1/envoyproxy_metric_types.go#L50

We've started to use BackendRef in newer APIs such as ext auth https://github.com/envoyproxy/gateway/blob/f699edf3663f2b7e5a5d37756403e1551a7d3976/api/v1alpha1/ext_auth_types.go#L70

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
ptal @envoyproxy/gateway-maintainers

zirain:
+1 on this, what about protocol?

arkodg:
@zirain is it safe to assume otel collector will always be running in the same cluster ?

For protocol we can use a seperate field or leverage appProtocol field inside the Service

guydc:
+1

Xunzhuo:
+1

zirain:
we can add appProtocol, also we can try https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

arkodg:
thanks folks, updated the issue description with the TODOs needed to complete this

arkodg:
sidenote - imo we should continue to use `host` `port` for any config in `EnvoyGateway` config, because of the bootstrapping problem - we dont want to rely on a provider (which may not be initialized yet or be part of another runner) to fetch the endpoints, having access to it directly helps

zirain:
need keep `protocol`?

arkodg:
> need keep `protocol`?

I vote for getting it from `appProtocol` . if we think that UX isn't working, we can add `protocol` in this API

zirain:
if it's a `service`, we should respect `appProtocol` first, then try exact from `name`?
Is there no possibility that the reference is not a `servcie` or doesn't contains `appProtocol `?

arkodg:
[...]

PR #3062, opened 2024-03-29
feat: gRPC Access Log Service (ALS) sink

davidalger:

This PR adds support for configuring a gRPC Access Log Service (ALS) logging sink implementing API added in #3078.


api/v1alpha1/accesslogging_types.go
```diff
@@ -73,14 +77,18 @@ const (
 //
 // +kubebuilder:validation:XValidation:rule="self.type == 'File' ? has(self.file) : !has(self.file)",message="If AccessLogSink type is File, file field needs to be set."
 // +kubebuilder:validation:XValidation:rule="self.type == 'OpenTelemetry' ? has(self.openTelemetry) : !has(self.openTelemetry)",message="If AccessLogSink type is OpenTelemetry, openTelemetry field needs to be set."
+// +kubebuilder:validation:XValidation:rule="self.type == 'GRPC' ? has(self.grpc) : !has(self.grpc)",message="If AccessLogSink type is GRPC, grpc field needs to be set."
 type ProxyAccessLogSink struct {
 	// Type defines the type of accesslog sink.
-	// +kubebuilder:validation:Enum=File;OpenTelemetry
+	// +kubebuilder:validation:Enum=File;GRPC;OpenTelemetry
```
zirain, line 83: grpc is a protocol name, perfer to name it as ALS or something like that.

davidalger: good point, and technically speaking OTel is a gRPC logger just a different API. I like using just `ALS`. I'll make this change!

api/v1alpha1/accesslogging_types.go
```diff
@@ -92,6 +100,41 @@ type FileEnvoyProxyAccessLog struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default="accesslog"
+	LogName string `json:"logName,omitempty"`
+	// Host define the extension service hostname.
+	Host string `json:"host,omitempty"`
+	// Port defines the port the extension service is exposed on.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	Port int32 `json:"port,omitempty"`
```
zirain, line 126: xref https://github.com/envoyproxy/gateway/issues/3041, we should not use host and port in new API.

can you proposal API first, otherwise there will be a lot of waste?

davidalger: api proposal: #3078

arkodg:
hey @davidalger still working on this one ?

zirain:
@davidalger are you still working on this? if not, I can take over this.

arkodg:
hey np @davidalger, congratulations on the new gig :) 
looks like @zirain already took the PR forward, so should be okay to close this one, would be great if you can review #3626

shawnh2:
closed in favor of #3626

Issue #2997, opened 2024-03-22
Support Unix Domain Sockets for Envoy services


guydc:
*Description*:
Unix Domain Sockets are sometimes used instead of TCP sockets, since communication over UDS has lower latency. Envoy Proxy supports UDS [addresses](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#network-addresses-proto) for upstream clusters. Currently, Gateway-API and Envoy Gateway do not not support UDS, since UDS cannot be represented by a K8s Service resource. 

In Envoy Gateway, support for UDS can improve latency to callout data-plane and control-plane services like Global Network/HTTP Rate Limit, Network/HTTP Ext Auth, HTTP Ext Proc and Secret Discovery Service (e.g. SPIRE) that run as sidecars or node-daemons. UDS is less likely to be used for communicating with backend applications, due to the high degree of coupling between the backend and the gateway. 

Gateway-API supports a notion of [implementation-specific Backend Object References](https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/?h=extensibility#approach-to-extensibility). This requirement is driven by Envoy-specific architecture constraints, and it is not likely to be useful for the general Gateway-API community. For this reason, Envoy Gateway can introduce a new (implementation-specific) resource representing UDS backends.

arkodg:
this was discussed in the call today
* @guydc @AliceProxy @kflynn are in favor of a single new CRD  ( like `UnixDomainService ` ) that acts as a custom backendRef only for unix domain sockets
* there are some security implications to supporting uds, it maybe possible to access the envoy admin server directly if its listening on a unix domain socket (its not by default today) . There's no version of `ReferenceGrant` that exists for things like this today
* if we are finally going down the path of supporting non k8s backend types, I'd suggest for us to also consider a generic way to represent a address (e.g. config in envoy https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-address) which can handle `UnixDomainSocket` , `IPAddress` and `FQDN` , which I've been trying to champion ever since the project started https://github.com/envoyproxy/gateway/issues/36, but the solution is a suboptimal one requiring multiple configs (`Service` and `EndpointSlice` https://gateway.envoyproxy.io/latest/user/traffic/routing-outside-kubernetes/)
* any approach here will probably need to be opt in, and not enabled by default, because this is allowing access to the same host as the proxy

[...]

Issue #36, opened 2022-04-28
Support routing to non K8s upstream services


arkodg:
Envoy Gateway intends to support upstream services & endpoint running in all environments as highlighted [here](https://github.com/envoyproxy/gateway/blob/main/GOALS.md#all-environments). This issue tracks how this can be
achieved using the [Gateway API](https://gateway-api.sigs.k8s.io)

arkodg:
possible today by creating EndpointSlice and linking them to a Service
https://gateway.envoyproxy.io/latest/user/traffic/routing-outside-kubernetes/

PR #3063, opened 2024-03-30
API: Backend

guydc:

**This is a preliminary API proposal for discussion purposes only.** The new `Backend` API allows users to reference UDS, FQDN and IP backends with an alternative to the K8s `Service` resources. See design doc for more details. 

Relates to #2997, #36


zirain:
so these's something like `ServiceEntry` in `Istio`?

site/content/en/latest/contributions/design/backend.md
```diff
@@ -0,0 +1,151 @@
...
+    caCertRefs:
+      - name: backend-tls-checks-certificate
+        group: ''
+        kind: ConfigMap
+    hostname: example.com
+```
+
+## Design Decisions
+* All existing and future `BackendObjectReference` in Envoy Gateway MUST support the `Backend` kind. 
+* Gateway-API and Envoy Gateway policies that attach to Services ([BackendTLSPolicy][], [BackendLBPolicy][])
```
arkodg, line 111: -1 on attaching BTS to a Backend, we technically have 4 levels of implicit defaults and overrides (`gateway`, `;listener`, `route` and `route rule` soon), adding more will further increase complexity

site/content/en/latest/contributions/design/backend.md
```diff
@@ -0,0 +1,151 @@
...
+* To limit the overall maintenance effort related to supporting of non-Service backends, the `Backend` API SHOULD 
+  support multiple generic address types (UDS, FQDN, IPv4, IPv6), and MUST NOT support vendor-specific backend types.
+* Both `Backend` and `Service` resources may appear in the same `BackendRefs` list.
+* The Optional `Port` field SHOULD NOT be evaluated when referencing a `Backend`.  
+* Referenced `Backend` resources MUST be translated to envoy endpoints, similar to the current `Service` translation.
+* Certain combinations of `Backend` and `Service` are incompatible. For example, a Unix Domain Socket and a FQDN service
+  require different cluster service discovery types (Static/EDS and Strict-DNS accordingly).
+* If a Backend that is referenced by a route cannot be translated, the `Route` resource will have an `Accepted=False` 
+  condition with a `UnsupportedValue` reason. 
+
```
arkodg, line 131: is there a valid use to supports UDS addresses in HTTPRoute ? I can't think of any

guydc: If the gateway is deployed in a very coupled manner with the backend, it could make sense. For example, if both the backend(s) and gateway are daemonsets, they may communicate over a host-level UDS. 

Proxies do support this in general: [nginx](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#upstream), [haproxy](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#4.2-server)

site/content/en/latest/contributions/design/backend.md
guydc: How is that different from supporting this for services, though? In terms of defaults/overrides, doesn't this introduce the same amount of layers that exists today?

guydc:
@zirain, @Xunzhuo - There are some difference from `ServiceEntry`. 
- `ServiceEntry` doesn't support Unix Domain Sockets
- You can route from `VirtualService` to external services by just specifying their address. You don't need a `ServiceEntry` for that, but a `ServiceEntry` allows you to configure some setting related to DNS resolution, (m)TLS, etc. In EG, this resource (or `Service` + FQDN `Endpoint`) is mandatory for external routing.

site/content/en/latest/contributions/design/backend.md
guydc: As discussed in the community meeting, we can choose to reject UDS for xRoute BackendRefs. The External Service use case is still valuable here.

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,107 @@
...
+	AddressProtocolTypeUDP AddressProtocolType = "UDP"
+)
+
+type ApplicationProtocolType string
+
+const (
+	// ApplicationProtocolTypeH2C defines the HTTP/2 prior knowledge application protocol.
+	ApplicationProtocolTypeH2C ApplicationProtocolType = "H2C"
+	// ApplicationProtocolTypeWS defines the WebSocket over HTTP protocol.
+	ApplicationProtocolTypeWS ApplicationProtocolType = "WS"
```
zhaohuabing, line 32: Should "WS" be a standalone ApplicationProtocolType?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,107 @@
...
+	Name string `json:"name,omitempty"`
+
+	// SocketAddress is a [FQDN|IP]:[Port] address
+	SocketAddress *SocketAddress `json:"socketAddress,omitempty"`
+
+	// UnixDomainSocketAddress is a unix domain socket path
+	UnixDomainSocketAddress *UnixDomainSocketAddress `json:"unixDomainSocketAddress,omitempty"`
+
+	// ApplicationProtocol determines the application protocol to be used, e.g. HTTP2.
+	ApplicationProtocol *ApplicationProtocolType `json:"applicationProtocol,omitempty"`
```
zhaohuabing, line 70: Should follow [IANA standard service names](https://www.iana.org/assignments/service-names).

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,107 @@
...
+const (
+	// KindBackend is the name of the Backend kind.
+	KindBackend = "Backend"
+)
+
+type AddressProtocolType string
+
+const (
+	// AddressProtocolTypeTCP defines the TCP address protocol.
+	AddressProtocolTypeTCP AddressProtocolType = "TCP"
```
zhaohuabing, line 21: Address can be layer2/layer3/layer4/layer7, probably just name it as Protocol?

api/v1alpha1/backend_types.go
guydc: I tried to keep this in-sync with https://gateway-api.sigs.k8s.io/geps/gep-1911/#supporting-protocols

api/v1alpha1/backend_types.go
guydc: I think that eventually, we use this indication to create cluster settings that are protocol-specific. Envoy supports a closed list of protocols: HTTPx, TLS, TCP and some [other protos](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/other_protocols). So, I do think that it makes sense to limit this list to supported protocols only (both in envoy and envoy gateway).

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,108 @@
...
+
+// UnixDomainSocketAddress describes TCP/UDP unix domain socket address, corresponding to Envoy's Pipe
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-pipe
+type UnixDomainSocketAddress struct {
+	Path string `json:"path"`
+}
+
+// BackendSpec describes the desired state of BackendSpec.
+type BackendSpec struct {
+	// +kubebuilder:validation:MaxItems=1
```
arkodg, line 95:
```suggestion
	// +kubebuilder:validation:MinItems=1
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// BackendAddress describes are backend address, which is can be either a TCP/UDP socket or a Unix Domain Socket
+// corresponding to Envoy's Address: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-address
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.socketAddress) || has(self.unixDomainSocketAddress))",message="one of socketAddress or unixDomainSocketAddress must be specified"
+// +kubebuilder:validation:XValidation:rule="(has(self.socketAddress) && !has(self.unixDomainSocketAddress)) || (!has(self.socketAddress) && has(self.unixDomainSocketAddress))",message="only one of socketAddress or unixDomainSocketAddress can be specified"
+type BackendAddress struct {
+	// Name is the unique name of the backend address
+	Name string `json:"name,omitempty"`
```
arkodg, line 61: is name needed ?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,108 @@
...
+
```
arkodg, line 62: should we add a top level `Type` - FQDN, IPv4, IPv6, UDS ?

api/v1alpha1/backend_types.go
guydc: I originally intended to use this to support named-section attachment of policies. However, since named-section attachment is not generally supported for xRoutes (only service ports are supported as a section), I think that we can decide that this is out of scope for now. We can add names later on to support this if needed.

api/v1alpha1/backend_types.go
guydc: UDS is distinguished from other types with a dedicated struct, since it doesn't follow the `host:port` pattern anyway. We can add IPv4/IPv6/FQDN type field. The value that I see for such a type is simplification of validation (only need to check against the expected type). However, it creates a more verbose API. I'm OK with both. WDYT?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,156 @@
...
+const (
+	// AddressTypeFQDN defines the RFC-1123 compliant fully qualified domain name address type.
+	AddressTypeFQDN ProtocolType = "FQDN"
+	// AddressTypeUDS defines the unix domain socket address type.
+	AddressTypeUDS ProtocolType = "UDS"
+	// AddressTypeIPv4 defines the IPv4 address type.
+	AddressTypeIPv4 ProtocolType = "IPv4"
+	// AddressTypeIPv6 defines the IPv4 address type.
+	AddressTypeIPv6 ProtocolType = "IPv6"
```
zirain, line 29: can we support IPv4 and IPv6 at the same time?

guydc: I couldn't find anything in the envoy docs to suggest that it's not possible. Judging by this example: https://istio.io/latest/blog/2023/experimental-dual-stack/, istio uses both IPv4 and IPv6 endpoints in the same cluster for dualstack support, as seen here:

```
ENDPOINT                 STATUS      OUTLIER CHECK     CLUSTER
[...]
10.244.0.26:9000         HEALTHY     OK                outbound|9000||tcp-echo.dual-stack.svc.cluster.local
fd00:10:244::1a:9000     HEALTHY     OK                outbound|9000||tcp-echo.dual-stack.svc.cluster.local
[...]
```

Having said that, I think that we can avoid IPv6 support in this API for the time being, and maybe address this at a later stage, together with #184.

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,154 @@
...
+
+// SocketAddress describes TCP/UDP socket address, corresponding to Envoy's SocketAddress
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-socketaddress
+// +notImplementedHide
+type SocketAddress struct {
+	// Address defines to the FQDN or IP address of the backend service.
+	Address string `json:"address"`
+
+	// Port defines to the port of of the backend service.
+	Port int32 `json:"port"`
```
zhaohuabing, line 102: Nit: uint32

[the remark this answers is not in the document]
guydc: In most of our APIs, we use int32 and kubebuilder min/max validations for portys. I think that this is because the CRDs are actually generated as int32, even when the go type is uint32, which will delay the validation to the processing phase. So, I kept it int32 but added a kubebuilder validation.

zhaohuabing:
LGTM. Thanks for adding this valuable API. It unlocks the full potential of EG to many more use cases. 

Defer to other @envoyproxy/gateway-maintainers

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,157 @@
...
+	// Port defines to the port of of the backend service.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	Port int32 `json:"port"`
+
+	// Protocol defines to the the transport protocol to use for communication with the backend.
+	//
+	// +optional
+	Protocol *ProtocolType `json:"protocol,omitempty"`
```
arkodg, line 110: can we do w/o this field in the init PR and rely on info from the `xRoute` for the protocol ?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,157 @@
...
+type BackendSpec struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	// +kubebuilder:validation:XValidation:rule="self.all(f, f.type == 'FQDN') || !self.exists(f, f.type == 'FQDN')",message="FQDN addresses cannot be mixed with other address types"
+	BackendAddresses []BackendAddress `json:"addresses,omitempty"`
+
+	// ApplicationProtocol defines the application protocol to be used, e.g. HTTP2.
+	//
+	// +optional
+	ApplicationProtocol *ApplicationProtocolType `json:"applicationProtocol,omitempty"`
```
arkodg, line 131: can we do w/o this field in the init PR and rely on info from the xRoute for the protocol ?

api/v1alpha1/backend_types.go
zhaohuabing: Would this result in protocol mismatching if the protocol of `xRoute` is not the same as the `Backend`?

api/v1alpha1/backend_types.go
arkodg: thoughts on ?
```
	FQDN *SocketAddress `json:"fqdn,omitempty"`
	
	IP *SocketAddress `json:"ip,omitempty"`

	// Unix defines the unix domain socket path
	//
	// +optional
	Unix *UnixDomainSocketAddress `json:"unix,omitempty"`
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,157 @@
...
+type UnixDomainSocketAddress struct {
+	Path string `json:"path"`
+}
+
+// BackendSpec describes the desired state of BackendSpec.
+// +notImplementedHide
+type BackendSpec struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	// +kubebuilder:validation:XValidation:rule="self.all(f, f.type == 'FQDN') || !self.exists(f, f.type == 'FQDN')",message="FQDN addresses cannot be mixed with other address types"
```
arkodg, line 125: its very wordy rn, `address` shows up in 3 layers
```
backendAddresses:
- socketAddress:
     address: 1.2.3.4
     port: 80
```

api/v1alpha1/backend_types.go
arkodg: that depends on the backend server type, so app dev should be using the right Route type imo

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,159 @@
...
+      - name: backend-tls-checks-certificate
+        group: ''
+        kind: ConfigMap
+    hostname: example.com
+```
+
+## Design Decisions
+* All instances of `BackendObjectReference` in Envoy Gateway MAY support referencing the `Backend` kind.
+* In some cases, Envoy Gateway may reject references to a `Backend` resource. For example, a backend with a UDS address 
+  will be rejected on xRoute references.
```
arkodg, line 111: for xRoute lets also disable `localhost` and `127.0.0.1`

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,159 @@
...
+* To limit the overall maintenance effort related to supporting of non-Service backends, the `Backend` API SHOULD 
+  support multiple generic address types (UDS, FQDN, IPv4, IPv6), and MUST NOT support vendor-specific backend types.
+* Both `Backend` and `Service` resources may appear in the same `BackendRefs` list.
+* The Optional `Port` field SHOULD NOT be evaluated when referencing a `Backend`.  
+* Referenced `Backend` resources MUST be translated to envoy endpoints, similar to the current `Service` translation.
+* Certain combinations of `Backend` and `Service` are incompatible. For example, a Unix Domain Socket and a FQDN service
+  require different cluster service discovery types (Static/EDS and Strict-DNS accordingly).
+* If a Backend that is referenced by a route cannot be translated, the `Route` resource will have an `Accepted=False` 
+  condition with a `UnsupportedValue` reason. 
+
```
arkodg, line 139: I'd vote to make this opt in at the `EnvoyGateway` level by an admin to avoid
https://github.com/advisories/GHSA-vw47-mr44-3jf9

api/v1alpha1/backend_types.go
zhaohuabing: We can add validation to check if there's a protocol mismatching if the Backend API also has a Protocol field and reflect it to the xRoute status. Otherwise, the mismatching leaks to runtime, making it harder for debugging.

api/v1alpha1/backend_types.go
guydc: I think that it's not currently possible to infer if HTTP1 or HTTP2 should be used upstream in HTTP routes. So, we need something similar to K8s service port application protocols.

arkodg: suggest making it a list if we want to add it in the first iteration
relates to https://github.com/kubernetes/enhancements/issues/3726

reg appProtocol naming, can we reuse anything from upstream K8s https://github.com/kubernetes/enhancements/pull/3727 ?

site/content/en/contributions/design/backend.md
guydc: I have no objection to doing this. However, one can easily bypass this check by using a FQDN that resolves to 127.0.0.1. This is also true for services with FQDN endpoints. 

The real solution here is to use something like a network policy to block this sort of egress traffic from the proxy pod. However, network policies are not supported in every K8s distribution (e.g. kind). 

I think that that covering this in a security doc is probably the best approach. 

WDYT?

arkodg: for services with FQDN endpoints / endpointslices , the service is owned by the app dev (lower priviledge) and the endpointslice is owned by the admin (higher priviledge) , so if the admin wants to muck around, we can let them, but this is more around not allowing app devs to be able to access the proxy endpoint directly . I'd try to avoid as much as we can, and document what we can't in a separate doc

api/v1alpha1/backend_types.go
guydc: > suggest making it a list

Can you explain why we need a list based on the KEP?

api/v1alpha1/backend_types.go
arkodg: yeah my vote would be to capture this as a issue and tackle once the need arises

api/v1alpha1/backend_types.go
arkodg: +1 to naming suggestion

arkodg: also prefer `appProtocols` over `applicationProocols`

zirain:
will this help https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/routing-outside-kubernetes/ without using service/endpointslice?

arkodg:
> will this help https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/routing-outside-kubernetes/ without using service/endpointslice?

yes this API is a better approach to do non k8s routing

site/content/en/contributions/design/backend.md
guydc: Updated the design doc to reflect this. Will implement the validation in the translator, as it's specific to xRoutes and not all instances of `Backend`.

api/v1alpha1/backend_types.go
guydc: done

api/v1alpha1/backend_types.go
guydc: Agree. When ahead and created dedicated structs to make validation a bit easier. For a yaml perspective, FQDN/IP look the same.

api/v1alpha1/backend_types.go
arkodg: could be simplified to
```
spec:
  endpoints:
  - ipv4:
      address: 1.2.3.4
      port: 80   
```

```
spec:
  endpoints:
  - ipv6:
      address: fe80::42:acff:fe11:2
      port: 80   
```

```
spec:
  endpoints:
  - unix:
       path: /var/run/blah.sock
```

zirain:
> > will this help https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/routing-outside-kubernetes/ without using service/endpointslice?
> 
> yes this API is a better approach to do non k8s routing

@guydc is this scenario in your plan?

guydc:
> > > will this help https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/routing-outside-kubernetes/ without using service/endpointslice?
> > 
> > 
> > yes this API is a better approach to do non k8s routing
> 
> @guydc is this scenario in your plan?

I'm targeting two scenarios:
- Simplify routing to out-of-cluster endpoints (for xRoutes and other backendRefs), as you mention 
- Support unix domain socket communication for reduced latency to extension servers (Ext Auth, Ext Proc, RL, Telemetry sinks.. ) 

I don't intend to implement support for `Backend` everywhere immediately. I'll start out with HTTPRoutes and ExtProc support for `Backend` referencing. I hope that that other use cases will be implemented gradually based on requirement from users.

site/content/en/latest/contributions/design/backend.md
guydc: as discussed, this was a misunderstanding. BackendTrafficPolicy will not attach to backend resources.

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,153 @@
...
+type BackendAddress struct {
+	// FQDN defines a FQDN address
+	//
+	// +optional
+	FQDN *FQDNAddress `json:"fqdn,omitempty"`
+
+	// IP defines a IPv4 address
+	//
+	// +optional
+	IP *IPAddress `json:"ip,omitempty"`
```
arkodg, line 65: should we split this up into `IPv4` and `IPv6` or leave it like this and infer internally using a lib ?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,153 @@
...
+	Path string `json:"path"`
+}
+
+// BackendSpec describes the desired state of BackendSpec.
+// +notImplementedHide
+type BackendSpec struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	// +kubebuilder:validation:XValidation:rule="self.all(f, has(f.fqdn)) || !self.exists(f, has(f.fqdn))",message="fqdn addresses cannot be mixed with other address types"
+	BackendAddresses []BackendAddress `json:"addresses,omitempty"`
```
arkodg, line 122: imo `backend` prefix is not needed, the config looks like
```
backend:
  spec:
    backendAddresses
```

this could be simplified to
```
backend:
  spec:
    addresses:

Comment (guydc) [reply api/v1alpha1/backend_types.go:122]: Sorry, missed your comment. I'll rename to endpoints as you suggested.

Comment (arkodg) [reply api/v1alpha1/backend_types.go:122]: tbh I dont know which one is better - `endpoints` or `addresses`

File: api/v1alpha1/backend_types.go [hunk 18]
```diff
@@ -0,0 +1,153 @@
...
+// IPAddress describes TCP/UDP socket address, corresponding to Envoy's Socket Address
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-socketaddress
+// +notImplementedHide
+type IPAddress struct {
+	// Host defines to the IPv4 address of the backend service.
+	//
+	// +kubebuilder:validation:MinLength=7
+	// +kubebuilder:validation:MaxLength=15
+	// +kubebuilder:validation:Pattern=`^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`
+	Host string `json:"host"`
```
arkodg, line 82: `Host` or `Address` ?

guydc: With endpoints as a top-level, we can have address here, which I feel is more accurate. Updated it now,

api/v1alpha1/backend_types.go
guydc: ok, makes sense. Pluralized it.

api/v1alpha1/backend_types.go
guydc: I think that your suggestion is closer to both K8s Endpoint and Envoy Endpoint APIs, where endpoints have an address (or addresses). So, I think that it should be pretty intuitive to EG users.

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+type BackendEndpoint struct {
+	// FQDN defines a FQDN endpoint
+	//
+	// +optional
+	FQDN *FQDNEndpoint `json:"fqdn,omitempty"`
+
+	// IP defines an IPv4 endpoint
+	//
+	// +optional
+	IP *IPv4Endpoint `json:"ipv4,omitempty"`
```
arkodg, line 67:
```suggestion
	IPv4 *IPv4Endpoint `json:"ipv4,omitempty"`
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+// BackendSpec describes the desired state of BackendSpec.
+//
+// +notImplementedHide
+type BackendSpec struct {
+	// BackendEndpoints defines the endpoints to be used when connecting to the backend.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	// +kubebuilder:validation:XValidation:rule="self.all(f, has(f.fqdn)) || !self.exists(f, has(f.fqdn))",message="fqdn addresses cannot be mixed with other address types"
+	BackendEndpoints []BackendEndpoint `json:"endpoints,omitempty"`
```
arkodg, line 131:
```suggestion
	Endpoints []BackendEndpoint `json:"endpoints,omitempty"`
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-socketaddress
+//
+// +notImplementedHide
+type FQDNEndpoint struct {
+	// Address defines the FQDN address of the backend endpoint.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Address string `json:"address"`
```
arkodg, line 104: should this be `Host` or `Address`

guydc: Envoy Endpoints contain an address ([IP, FQDN](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-socketaddress) or [UDS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-pipe)). The term envoy uses for hostname/ip is "address". 

```sh
Endpoint:
{
  "address": {...},
[...]
}

Socket Address:
{
  "protocol": ...,
  "address": ...,
  "port_value": ...,
[...]
}
```

Kubernetes Endpoints contain an address (IP, FQDN), and refer to hostname/ip as "address". 
```sh
apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name: example-abc
addressType: IPv4
endpoints:
  - addresses:
      - "10.1.2.3"
```

So, I think that we can stick to that terminology.

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,162 @@
+---
+title: "Backend "
```
shawnh2, line 2:
```suggestion
title: "Backend"
```

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,162 @@
...
+## Alternatives
+* The project can indefinitely wait for these configuration parameters to be part of the [Gateway API][].
+* Users can leverage the existing [Envoy Patch Policy][] or [Envoy Extension Manager][] to inject custom envoy clusters
+  and route configuration. However, these features require a high level of envoy expertise, investment and maintenance. 
+
+[BackendObjectReference Extension]: https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/?h=extensi#approach-to-extensibility
+[internal listeners]: https://www.envoyproxy.io/docs/envoy/latest/configuration/other_features/internal_listener
+[unix domain sockets]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-pipe
+[Resource Backends]: https://kubernetes.io/docs/concepts/services-networking/ingress/#resource-backend
+[Services and FQDN endpoints]: https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/routing-outside-kubernetes/
```
shawnh2, line 151: can we use the relevant path for this?

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,162 @@
...
+    - "www.example.com"
+  rules:
+    - backendRefs:
+        - group: gateway.envoyproxy.io
+          kind: Backend
+          name: backend-mixed-ip-uds
+          weight: 1
+        - group: ""
+          kind: Service
+          name: backend
```
shawnh2, line 82:
```suggestion
          name: backend-svc
```

do you mean this?

site/content/en/contributions/design/backend.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Non Goals
+* Support specific backend types, such as S3 Bucket, Redis, AMQP, InfluxDB, etc.  
+
+## Implementation
+
+The `Backend` resource is an implementation-specific Gateway-API [BackendObjectReference Extension][]. 
+
+### Example
+Here is an example highlighting how a user can configure a route that forwards traffic to both a Service and a local
```
shawnh2, line 35: curious about here: do you mean we can configure a route that forwards traffic to a Service `backend-svc`, a locl UDS, and `10.244.0.28:3000`?

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+// +kubebuilder:resource:categories=envoy-gateway,shortName=be
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Accepted")].reason`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +notImplementedHide
+type Backend struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec defines the desired state of Backend.
```
shawnh2, line 45:
```suggestion
	// Spec defines the desired state of Backend.
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+	Spec BackendSpec `json:"spec"`
+
+	// status defines the current status of Backend.
```
shawnh2, line 48:
```suggestion
	// Status defines the current status of Backend.
```

api/v1alpha1/backend_types.go
```diff
@@ -0,0 +1,162 @@
...
+	Status BackendStatus `json:"status,omitempty"`
```
shawnh2, line 49: can you also include some basic status type and conditions in this PR as well?

api/v1alpha1/backend_types.go
guydc: Discussed in the community meeting and agreed on hostname.

api/v1alpha1/backend_types.go
guydc: done

site/content/en/contributions/design/backend.md
guydc: yes, thanks, fixed

arkodg (approved):
LGTM thanks ! looking forward to using this API

zhaohuabing (approved):
LGTM thanks!

PR #2793, opened 2024-03-06
chore: bump golang to 1.22

Xunzhuo:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Bump go version

PR #3068, opened 2024-03-31
chore: bump controller-tools to `0.14.0`

zirain:
fix `make manifests` panic with golang 1.22


internal/metrics/testdata/counter_metric.json
zirain, line 1: rename from *.yaml to *.json to avoid yamllint fail

zirain:
replaced by https://github.com/envoyproxy/gateway/pull/2793

PR #3071, opened 2024-04-01
chore: rename *.yaml to *.json

zirain:
cc @ShyunnY


internal/metrics/testdata/gauge_metric.json
```diff
@@ -40,7 +40,7 @@
                   }
                 ],
                 "StartTime": "0001-01-01T00:00:00Z",
-                "Time": "2024-03-27T21:54:18.8562259+08:00",
+                "Time": "2024-03-31T11:25:29.204058+08:00",
```
Xunzhuo, line 43: will this update each time we run make generate ?

ShyunnY:
> cc @ShyunnY

YAML is compatible with JSON. Do we need to modify it to JSON?

zirain:
> > cc @ShyunnY
> 
> YAML is compatible with JSON. Do we need to modify it to JSON?

this will case yamllint tool panic

on the otherside, content should match with ext

ShyunnY:
block

zirain:
replaced by https://github.com/envoyproxy/gateway/pull/3077

PR #3078, opened 2024-04-01
api: gRPC Access Log Service (ALS) sink

davidalger:
**What type of PR is this?**

API to support a gRPC Access Log Service (ALS) sink.

When implemented, a sink may be configured to send log messages to a gRPC stream with rich and strongly typed protobufs including all the details for the request.

We are using a lightweight gRPC logging service to post-process logs (ex: enrich with user session metadata) and then publish them to Stackdriver which in turn routes them to both BigQuery and Pub/Sub for further processing and long-term storage.

Currently we are relying on a patched bootstrap and two XDS patches on each listener to replace the default access logging config with an `HttpGrpcAccessLogConfig`, but would like to natively support this in the `EnvoyProxy` CRD in order to keep configuration simple and not rely on having a Terraform configured `EnvoyPatchPolicy` for each listener on our Gateways.

* xref: #3062
* xref: #3041 

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/grpc/v3/als.proto


api/v1alpha1/accesslogging_types.go
```diff
@@ -86,6 +96,60 @@ type ProxyAccessLogSink struct {
...
+	// to differentiate between different access logs coming from the same Envoy.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default="accesslog"
+	LogName string `json:"logName,omitempty"`
+	// Type defines the type of accesslog. Supported types are "HTTP" and "TCP". Defaults to "HTTP" when not specified.
+	// +kubebuilder:validation:Enum=HTTP;TCP
+	// +unionDiscriminator
+	// +kubebuilder:default="HTTP"
+	Type ALSEnvoyProxyAccessLogType `json:"type,omitempty"`
```
zirain, line 135: only suppot HTTP right now?

davidalger: since there's no config specific to TCP — basically just a different type — I figured we may as well include both in the API design. although we could easily pull the `type` field out entirely until a later date if it'd be more amenable to keep the API specific to HTTP for now.

api/v1alpha1/accesslogging_types.go
```diff
@@ -71,13 +77,17 @@ const (
 // ProxyAccessLogSink defines the sink of accesslog.
 // +union
 //
+// +kubebuilder:validation:XValidation:rule="self.type == 'ALS' ? has(self.als) : !has(self.als)",message="If AccessLogSink type is ALS, als field needs to be set."
```
shawnh2, line 80: can you add some test under https://github.com/envoyproxy/gateway/tree/main/test/cel-validation to make sure that this CEL works as expected?

api/v1alpha1/accesslogging_types.go
zhaohuabing, line 135: Consider making `Type` a mandatory field? Or we could simply remove the `Type` here.

api/v1alpha1/accesslogging_types.go
```diff
@@ -86,6 +96,60 @@ type ProxyAccessLogSink struct {
...
+	// BackendRef references a Kubernetes object that represents the gRPC service to which
+	// the access logs will be sent. Currently only Service is supported.
+	BackendRef gwapiv1.BackendObjectReference `json:"backendRef,omitempty"`
+	// LogName defines the friendly name of the access log to be returned in
+	// StreamAccessLogsMessage.Identifier. This allows the access log server
+	// to differentiate between different access logs coming from the same Envoy.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default="accesslog"
+	LogName string `json:"logName,omitempty"`
```
zhaohuabing, line 130: Can we generate this name?

davidalger: Since access log filters could be implemented in the future to send different logs to different places my preference would be to leave it a user-configurable 'friendly name' with a simple default of `accesslog` so it's not mandatory.

If it were generated, we'd have to determine a way to differentiate between different log sinks on the same listener so the name would be useful to differentiate the logs in the receiver.

With the quickstart demo deployed locally, here is a portion of an example log entry generated by our ALS server:

```json
"envoy_node": "envoy-default-eg-9fd49bdd-556b9fdf4864klx", // --service-node value from bootstrap
"envoy_version": "1.29.2",
"envoy_cluster": "default/eg",            // --service-cluster value from bootstrap
"envoy_log_name": "accesslog",       // logName value from CRD
"upstream_cluster": "httproute/default/backend/rule/0",
"route_name": "httproute/default/backend/rule/0/match/0/www_example_com"
```

Just to say that an EG generated name is already present, although missing the listener name.

Given this context and the possibility of arbitrary access log filters resulting in multiple sinks to the same logging service in the future, what are your thoughts?

api/v1alpha1/accesslogging_types.go
davidalger: I'll make it mandatory, keeping with conventions of existing type fields

api/v1alpha1/accesslogging_types.go
zhaohuabing: Thanks for the explanation. Sounds reasonable for me to allow the user to specify the log name.

For the default name, probably add the listener name and the protocol? Like: lisetener1/http/accesslog

api/v1alpha1/accesslogging_types.go
davidalger: added

zirain (dismissed):

zhaohuabing (commented):
LGTM. Defer to @arkodg

api/v1alpha1/accesslogging_types.go
```diff
@@ -88,6 +96,56 @@ type ProxyAccessLogSink struct {
...
+//
+// - `x-accesslog-text` - The access log format string when a Text format is used.
+// - `x-accesslog-attr` - JSON encoded key/value pairs when a JSON format is used.
+//
+// +kubebuilder:validation:XValidation:message="BackendRef only supports Service Kind.",rule="!has(self.backendRef.kind) || self.backendRef.kind == 'Service'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'HTTP' || !has(self.http)",message="The http field may only be set when type is HTTP."
+type ALSEnvoyProxyAccessLog struct {
+	// BackendRef references a Kubernetes object that represents the gRPC service to which
+	// the access logs will be sent. Currently only Service is supported.
+	BackendRef gwapiv1.BackendObjectReference `json:"backendRef"`
```
davidalger, line 122: @arkodg given the conversation surrounding #3091 today, would it be preferred to amend this to pluralize now or address with sweeping pass updating all affected areas when #3091 is done for everything else?

arkodg: prefer if we wait another day or two to make sure #3091 has majority votes, and a decision is made, and that can be incorporated in this PR

arkodg: looks like there were enough votes to support pluralizing `backendRefs` , suggest waiting on https://github.com/envoyproxy/gateway/pull/3080 which is trying to create a common `BackendRef` struct that all APIs can reuse

davidalger: sounds good, thanks for the followup. will wait on that one to polish this one up

zirain (commented):
LGTM

api/v1alpha1/accesslogging_types.go
```diff
@@ -86,6 +94,59 @@ type ProxyAccessLogSink struct {
...
+}
+
+type ALSEnvoyProxyHTTPAccessLogConfig struct {
+	// RequestHeaders defines request headers to include in log entries sent to the access log service.
+	// +optional
+	RequestHeaders []string `json:"requestHeaders,omitempty"`
+	// ResponseHeaders defines response headers to include in log entries sent to the access log service.
+	// +optional
+	ResponseHeaders []string `json:"responseHeaders,omitempty"`
+	// ResponseTrailers defines response trailers to include in log entries sent to the access log service.
```
arkodg, line 145: do you have a use case for trailers atm ?

api/v1alpha1/accesslogging_types.go
```diff
@@ -86,6 +94,59 @@ type ProxyAccessLogSink struct {
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:XValidation:message="BackendRefs only supports Service kind.",rule="self.all(f, f.kind == 'Service')"
+	BackendRefs []BackendRef `json:"backendRefs"`
+	// LogName defines the friendly name of the access log to be returned in
+	// StreamAccessLogsMessage.Identifier. This allows the access log server
+	// to differentiate between different access logs coming from the same Envoy.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	LogName *string `json:"logName,omitempty"`
```
arkodg, line 129: can we make this an internal detail and use system details for populating the `node` and `logName` field ?

arkodg (approved):
LGTM thanks !

PR #3082, opened 2024-04-02
feat: configure overload manager

### LINKED ISSUE #1966: Global downstream connection limits.

shahar-h:
**What type of PR is this?**
feature

Configure overload manager as recommended in [Configuring Envoy as an edge proxy](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy) guide with the following:
* Set global downstream connection limits to `50000`.
* Dynamically set `max_heap_size_bytes` to `80%` of `Envoy` container memory limits. in case there are no limits `max_heap_size_bytes` is not being set. This approach has been agreed [here](https://github.com/envoyproxy/gateway/pull/2062#discussion_r1491049011).

Fixes #1966 
Related #1048 

**Note regarding global downstream connection limits**:
latest stable Envoy(v1.29.2) [docs](https://www.envoyproxy.io/docs/envoy/v1.29.2/configuration/operations/overload_manager/overload_manager#limiting-active-connections) suggests to use new overload manager API and also states: "One could also set this limit via specifying an integer through the runtime key overload.global_downstream_max_connections, though this key is deprecated and will be removed in future".
However, when navigating to Downstream connections API [page](https://www.envoyproxy.io/docs/envoy/v1.29.2/api-v3/extensions/resource_monitors/downstream_connections/v3/downstream_connections.proto#downstream-connections-proto) it is stated that it's currently work-in-progress.

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -143,8 +143,14 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 		infra.Config.Spec.Telemetry != nil {
 		proxyMetrics = infra.Config.Spec.Telemetry.Metrics
 	}
+
+	maxHeapSizeBytes := caclulateMaxHeapSizeBytes(deploymentConfig.Container.Resources)
+
 	// Get the default Bootstrap
-	bootstrapConfigurations, err := bootstrap.GetRenderedBootstrapConfig(proxyMetrics)
+	bootstrapConfigurations, err := bootstrap.GetRenderedBootstrapConfig(&bootstrap.RenderBootsrapConfigOptions{
+		ProxyMetrics:     proxyMetrics,
+		MaxHeapSizeBytes: maxHeapSizeBytes,
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -388,4 +394,19 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec) []corev
 	} else {
 		return env
 	}
+}
+
+// caclulateMaxHeapSizeBytes calculates the maximum heap size in bytes as 80% of Envoy container memory limits.
+// In case no limits are defined '0' is returned, which means no heap size limit is set.
+func caclulateMaxHeapSizeBytes(envoyResourceRequirements *corev1.ResourceRequirements) uint64 {
+	if envoyResourceRequirements == nil || envoyResourceRequirements.Limits == nil {
+		return 0
+	}
+
+	if memLimit, ok := envoyResourceRequirements.Limits[corev1.ResourceMemory]; ok {
+		memLimitBytes := memLimit.Value()
+		return uint64(float64(memLimitBytes) * 0.8)
+	}
+
+	return 0
 }
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -130,6 +130,13 @@ envoyProxy:
                       path_config_source:
                         path: "/sds/xds-trusted-ca.json"
                       resource_api_version: V3
+        overload_manager:
+          refresh_interval: 0.25s
+          resource_monitors:
+          - name: "envoy.resource_monitors.global_downstream_max_connections"
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.resource_monitors.downstream_connections.v3.DownstreamConnectionsConfig
+              max_active_downstream_connections: 50000
     logging: {}
   status: {}
 gatewayClass:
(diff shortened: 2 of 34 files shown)


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -389,3 +395,18 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec) []corev
...
+	if envoyResourceRequirements == nil || envoyResourceRequirements.Limits == nil {
+		return 0
+	}
+
+	if memLimit, ok := envoyResourceRequirements.Limits[corev1.ResourceMemory]; ok {
+		memLimitBytes := memLimit.Value()
+		return uint64(float64(memLimitBytes) * 0.8)
+	}
+
+	return 0
```
guydc, line 411: can we add a unit test covering the `return 0` use case?

shahar-h: Added [here](https://github.com/envoyproxy/gateway/pull/3082/commits/3f34cf82330114d35e7c03fac35c36066d34d9d0)

guydc (commented):
LGTM, Thanks!

guydc (approved):

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -389,3 +395,18 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec) []corev
 		return env
 	}
 }
+
+// caclulateMaxHeapSizeBytes calculates the maximum heap size in bytes as 80% of Envoy container memory limits.
+// In case no limits are defined '0' is returned, which means no heap size limit is set.
+func caclulateMaxHeapSizeBytes(envoyResourceRequirements *corev1.ResourceRequirements) uint64 {
+	if envoyResourceRequirements == nil || envoyResourceRequirements.Limits == nil {
```
arkodg, line 402: also lets make this method best effort
so it returns a *unit64, when we cannot compute the size lets return `nil`
and not add a max any `FixedHeapConfig` for that case

arkodg: looks like you already handle `0`, suggest changing the API signature to return a ptr instead so the caller can use `nil` as a way to make a decision

arkodg:
looks good ! just added one comment around using a ptr to improve the distinction b/w unset/unknown and 0

internal/infrastructure/kubernetes/proxy/resource.go
shahar-h: `0` is an illegal value for `max_heap_size_bytes`, so it means it's `disabled`. It can be returned in 2 cases:
1. No memory limits are defined
2. Memory limits are `0`, which has the same affect as `1`

I don't see a value in changing the signature to ptr unless I missed something.

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -389,3 +395,18 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec) []corev
 		return env
 	}
 }
+
+// caclulateMaxHeapSizeBytes calculates the maximum heap size in bytes as 80% of Envoy container memory limits.
+// In case no limits are defined '0' is returned, which means no heap size limit is set.
+func caclulateMaxHeapSizeBytes(envoyResourceRequirements *corev1.ResourceRequirements) uint64 {
```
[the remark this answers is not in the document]
shahar-h: The function implementation is specific to k8s provider(memory limits).  Only the function name is related to envoy. How is it different from all container args declared [here](https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/proxy/resource.go#L163-L186) which are also envoy specific?

internal/infrastructure/kubernetes/proxy/resource.go
arkodg: cool if `0` is `disabled`, then sgtm, thats not clear from the API doc https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/resource_monitor/fixed_heap/v2alpha/fixed_heap.proto#config-resource-monitor-fixed-heap-v2alpha-fixedheapconfig

internal/infrastructure/kubernetes/proxy/resource.go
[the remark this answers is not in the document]
arkodg: fair point, lets keep it here

arkodg (approved):
LGTM thanks !

internal/infrastructure/kubernetes/proxy/resource.go
shahar-h: Right, I verified it by defining 0 and got validation error from Envoy at runtime.

PR #3085, opened 2024-04-03
fix: rate limit doesn't work with two(and more) listeners

zhaohuabing:
This PR fixes the issue: rate limit doesn't work with two(and more) listeners

This was caused by [the missing name in RateLimitConfig resource](https://github.com/envoyproxy/gateway/blob/0f87fcbad1aaa2913af3f768da87ea72a695a14d/internal/xds/translator/ratelimit.go#L309). 

Fix: https://github.com/envoyproxy/gateway/issues/3086


arkodg:
thanks for debugging and fixing this one !
can you also add a yaml test in https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/ratelimit-config, so its easier to spot this in the future

zhaohuabing:
> thanks for debugging and fixing this one ! can you also add a yaml test in https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/ratelimit-config, so its easier to spot this in the future

* Add `name` field to https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/ratelimit-config verify the name is not nil.  
*  An e2e test has been added to prevent this happens again in the future.

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):

arkodg:
hitting https://github.com/codecov/codecov-action/issues/1359

PR #3090, opened 2024-04-04
correct kubectl command to show deployment in envoy-gateway-system na…

eitansuez:
**What type of PR is this?**
* "docs: updates/fixes kubectl command in user guide 'customize envoyproxy' to reference envoy-gateway-system namespace (was missing)"

i believe it was an omission - the deployment in question resides in envoy-gateway-system


arkodg:
thanks @eitansuez ! can you also make the same change for the `latest` version as well

zirain (approved):

eitansuez:
ah, thanks, i'd overlooked that.  ok, change applied also to latest.

arkodg (approved):
LGTM thanks !

PR #3129, opened 2024-04-08
fix egctl does not support custom controller name.

### LINKED ISSUE #3140: egctl does not support custom controller name
using custom envoy-gateway-config with custom controller name , but the egctl just support default controller name

qicz:
**What type of PR is this?**

Fixes #3140


internal/cmd/egctl/translate.go
```diff
@@ -276,7 +276,7 @@ func translateGatewayAPIToIR(resources *gatewayapi.Resources) (*gatewayapi.Trans
 	}
 
 	t := &gatewayapi.Translator{
-		GatewayControllerName:   egv1a1.GatewayControllerName,
+		GatewayControllerName:   string(resources.GatewayClass.Spec.ControllerName),
```
arkodg, line 279: probably good to check if `resources.GatewayClass` is `nil` or not, before dereferncing it

qicz: https://github.com/envoyproxy/gateway/blob/67008f83dc7e10df2b5c94c578a9379c3b688be7/internal/cmd/egctl/translate.go#L691-L694

did it

arkodg: won't this crash if no `GatewayClass` is provided ?

qicz: @arkodg yes, gateway class checked before translate  resouce 
https://github.com/envoyproxy/gateway/blob/67008f83dc7e10df2b5c94c578a9379c3b688be7/internal/cmd/egctl/translate.go#L332-L334

arkodg (approved):

shawnh2 (approved):

PR #3130, opened 2024-04-08
Use <proto>-<port> for naming service and container ports

arkodg:
Takes inspiration from https://github.com/envoyproxy/gateway/pull/2973 to name port, not off the listener but off the port-proto ensuring that patch (during updates) also works

Fixes https://github.com/envoyproxy/gateway/pull/3130#issuecomment-2045093056 by adding a stable name

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/quickstart.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/quickstart.all.yaml
@@ -80,7 +80,7 @@ infraIR:
         name: envoy-gateway-system/eg/http
         ports:
         - containerPort: 10080
-          name: http
+          name: http-80
           protocol: HTTP
           servicePort: 80
       metadata:
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -7,7 +7,6 @@ package gatewayapi
 
 import (
 	"fmt"
-	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -147,7 +146,7 @@ func (t *Translator) processInfraIRListener(listener *ListenerContext, infraIR I
 	}
 
 	infraPort := ir.ListenerPort{
-		Name:          strings.ToLower(fmt.Sprintf("%s-%d", proto, servicePort.port)),
+		Name:          irListenerPortName(proto, servicePort.port),
 		Protocol:      proto,
 		ServicePort:   servicePort.port,
 		ContainerPort: servicePortToContainerPort(servicePort.port),
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -7,6 +7,7 @@ package gatewayapi
 
 import (
 	"fmt"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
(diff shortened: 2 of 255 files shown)


internal/cmd/egctl/testdata/translate/out/quickstart.all.yaml
```diff
@@ -80,7 +80,7 @@ infraIR:
         name: envoy-gateway-system/eg/http
         ports:
         - containerPort: 10080
-          name: http
+          name: http-80
```
zirain, line 83: like this naming rule

arkodg:
cc @cnvergence @ardikabs

cnvergence:
that was fast!

owenhaynes:
@arkodg 

In my example we have 4+ listeners attached to a merge host gateway using the listeners name "https". Every time the gateway checks to see if the deployment is in-sync the container port name is different as its randomly getting a different listener name, this causes the deployment to get updated as the specs do not match.

This is tested with image `envoyproxy/gateway-dev:5a15902` which is the tag which contains the backports.

owenhaynes:
Here is the EnvoyProxy patchs for people to use if they are having the same problem until this resolved.

```yaml
    envoyProxy:
      provider:
        kubernetes:
          envoyService:
            patch:
              value:
                spec:
                  ports:
                  - name: https
                    port: 443
                    protocol: TCP
                    targetPort: 10443
          envoyDeployment:
            patch:
              value:
                spec:
                  template:
                    spec:
                      containers:
                      - name: envoy
                        ports:    
                        - containerPort: 10443
                          name: https
                          protocol: TCP
```

internal/gatewayapi/listener.go
```diff
@@ -145,12 +146,8 @@ func (t *Translator) processInfraIRListener(listener *ListenerContext, infraIR I
 		proto = ir.UDPProtocolType
 	}
 
-	infraPortName := string(listener.Name)
-	if t.MergeGateways {
-		infraPortName = irHTTPListenerName(listener)
-	}
 	infraPort := ir.ListenerPort{
-		Name:          infraPortName,
+		Name:          strings.ToLower(fmt.Sprintf("%s-%d", proto, servicePort.port)),
```
zirain, line 150: what about move to a helper/util packge?

cnvergence (approved):
lgtm

internal/gatewayapi/listener.go
arkodg: moved into a helper pkg

zirain (approved):

shawnh2 (commented):
seems we should also change the e2e test case for this one https://github.com/envoyproxy/gateway/blob/e537033f7f8cbd718b9c03b8c7ef2b24d241c1dd/test/e2e/tests/gateway_infra_resource.go#L132

according to the failure of e2e test https://github.com/envoyproxy/gateway/actions/runs/8688528949/job/23858439302?pr=3130#step:6:2966

zirain:
lint fails
```
Error: : # github.com/envoyproxy/gateway/test/e2e/tests
Error: test/e2e/tests/gateway_infra_resource.go:25:2: "github.com/envoyproxy/gateway/internal/utils" imported and not used (typecheck)
Error: could not import github.com/envoyproxy/gateway/test/e2e/tests (-: # github.com/envoyproxy/gateway/test/e2e/tests
Error: test/e2e/tests/gateway_infra_resource.go:25:2: "github.com/envoyproxy/gateway/internal/utils" imported and not used) (typecheck)
Error: could not import github.com/envoyproxy/gateway/test/e2e/tests (-: # github.com/envoyproxy/gateway/test/e2e/tests
Error: test/e2e/tests/gateway_infra_resource.go:25:2: "github.com/envoyproxy/gateway/internal/utils" imported and not used) (typecheck)
```

zirain (approved):

Issue #3111, opened 2024-04-07
Failed to update listener port of a Gateway CR


modatwork:
*Description*:

After updating the Gateway listener’s port, Envoy should listen on the new port and stop listening on the old port.

*Reproduction Steps*:

1. Set up the Envoy gateway and Custom Resources (CRs) by following the quickstart guide. Initial setup works fine.
2. Update the listener port in the Gateway Custom Resource (CR) to change it from 80 to 8001. After applying the updated Gateway CR, Envoy is not ready on either port 80 or 8001.

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 8001
```

*Environment*:

gateway version: gateway-dev:latest (imageID: docker-pullable://envoyproxy/gateway-dev@sha256:4a84a34a7955e4cf7fc276725957c11283c515b2f1e93684d753fa0582e540ba)

*Logs*:

```
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:141    reconciling gateways    {"runner": "provider"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:719    processing Gateway      {"runner": "provider", "namespace": "default", "name": "eg"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/routes.go:235        processing HTTPRoute    {"runner": "provider", "namespace": "default", "name": "backend"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:493    processing OIDC HMAC Secret     {"runner": "provider", "namespace": "envoy-gateway-system", "name": "envoy-oidc-hmac"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:316    processing Backend      {"runner": "provider", "kind": "Service", "namespace": "default", "name": "backend"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:330    added Service to resource tree  {"runner": "provider", "namespace": "default", "name": "backend"}
2024-04-07T08:29:53.884Z        INFO    provider        kubernetes/controller.go:364    added EndpointSlice to resource tree    {"runner": "provider", "namespace": "default", "name": "backend-nphpz"}
[...]

PR #3134, opened 2024-04-08
refactor: infra client CreateOrUpdate to ServerSideApply

ardikabs:
**What type of PR is this?**
Full context #3111

This PR will refactor the infra client which is moving from an imperative to a declarative way to manage the resources, specifically with the server-side apply mechanism, including a full refactor on `CreateOrUpdate` to `ServerSideApply`.
The reason, quoted from https://github.com/envoyproxy/gateway/issues/3111#issuecomment-2041443070:
> Apply expects to manage the entire lifecycle of the object, so when you created independently, it had no record of the original applied content, so it didn't know your updated manifest should remove the old port.

In previous attempts, we created a resource using the Create operation but then attempted updates using the Apply operation. This led us to identify issues on certain fields, such as `deployment.spec.template.spec.containers[*].ports[*]`, that returned the following error message:
```
{"runner": "infrastructure", "error": "failed to create or update deployment envoy-gateway-system/envoy-envoy-gateway-system-e2e-test-infra-7104d1de: for Update: Deployment.apps \"envoy-envoy-gateway-system-e2e-test-infra-7104d1de\" is invalid: spec.template.spec.containers[0].ports[1].name: Duplicate value: \"http-e0603c49\""}
```

I've also included an e2e test to ensure at least for the Gateway listener changes, specifically for the port name and number.

### DIFF
--- a/internal/infrastructure/kubernetes/infra_client.go
+++ b/internal/infrastructure/kubernetes/infra_client.go
@@ -31,27 +31,6 @@ func (cli *InfraClient) ServerSideApply(ctx context.Context, obj client.Object) 
 	}
 
 	return nil
-
-	// return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-	// 	if err := cli.Client.Get(ctx, key, current); err != nil {
-	// 		if kerrors.IsNotFound(err) {
-	// 			// Create if it does not exist.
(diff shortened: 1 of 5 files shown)


ardikabs:
cc @zirain

PR #3138, opened 2024-04-09
API for authorization

zhaohuabing:
@arkodg @zetaab @rooque @guydc 

Related issue: https://github.com/envoyproxy/gateway/issues/2250

### DIFF
--- a/api/v1alpha1/authorization_types.go
+++ b/api/v1alpha1/authorization_types.go
@@ -1,0 +1,76 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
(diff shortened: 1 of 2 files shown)


api/v1alpha1/authorization_types.go
```diff
@@ -0,0 +1,76 @@
...
+	JWTClaim *JWTClaim
+}
+
+// JWTClaim defines a JWT claim that can be used to authorize requests.
+type JWTClaim struct {
+	// Name is the name of the claim.
+	Name string `json:"name"`
+
+	// Value is the value of the claim.
+	Value string `json:"value"`
```
zetaab, line 68: Just thinking like groups parameter. Use case could be "i want allow group x only" or "i want allow/deny groups x y z".

zhaohuabing:
> thanks for picking this up @zhaohuabing , Id suggest by first taking #2652 forward, and then adding more attributes like `methods` and `jwtClaims`

Sure, I didn't notice that one. Close this for now.

PR #3154, opened 2024-04-09
revises instructions to render yaml in yaml format, removing the kube…

### LINKED ISSUE #3137: docs: yaml artifacts in guides should code-highlight
*Description*:

most of the tasks combine [a] displaying yaml to be applied with [b] applying it to the cluster using the shorthand:

```shell
cat <<EOF | kubectl apply -f -
---
some: yaml
EOF
```

this is great and convenient copy-paste.
but if the resources were presented first as bona-fide yaml, you'd get better yaml highlighting/color-coding, which makes it easier for the reader to review the content.
it also gives the reader an opportunity to copy just the yaml if they want to save these CRs on their file system and study them before doing a kubectl apply.
(also if the reader is not using the bash shell, that command might not work)

suggestion:  revise all instructions in the user guides / tasks such that we display yaml instead, along with an instruction to kubectl apply the resource[s] to the cluster:

```yaml
---
some: yaml
```

eitansuez:
…ctl apply -f - <<EOF part (#3137)

**What type of PR is this?**

This is a docs-related PR

First draft implementation of #3137:  revises yaml snippets in lab instructions to render as yaml.

Fixes #3137

### DIFF
--- a/site/content/en/latest/install/custom-cert.md
+++ b/site/content/en/latest/install/custom-cert.md
@@ -17,8 +17,8 @@ We use Cert-Manager to manage the certificates. You can install it by following 
 
    *You should not use the self-signed issuer in production, you should use a real CA issuer.*
 
-   ```shell
-   cat <<EOF | kubectl apply -f -
+   ```yaml
+   ---
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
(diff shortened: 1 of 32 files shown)


eitansuez:
i'm open to feedback.  a case can be made both for and against this idea of removing the `cat <<EOF | kubectl apply -f` convenience.

arkodg:
hey @eitansuez can we support both styles using tabs ?

eitansuez:
> hey @eitansuez can we support both styles using tabs ?

exactly.  i'm just waiting until tabs capability is added to the hugo config.  i assume @yuluo-yx is working on that?

arkodg:
> > hey @eitansuez can we support both styles using tabs ?
> 
> exactly. i'm just waiting until tabs capability is added to the hugo config. i assume @yuluo-yx is working on that?

it was completed in another PR https://github.com/envoyproxy/gateway/pull/3208

eitansuez:
i'm going to close this PR and start a new one rooted at main's HEAD.

PR #3156, opened 2024-04-09
fix(chart): set ttlSecondsAfterFinished on certgen job to 30 by default

JuniorJPDJ:
This is painless workaround for ArgoCD bug: argoproj/argo-cd#6880

### DIFF
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -48,7 +48,7 @@ certgen:
   job:
     annotations: {}
     resources: {}
-    ttlSecondsAfterFinished: 0
+    ttlSecondsAfterFinished: 30
   rbac:
     annotations: {}
     labels: {}
--- a/site/content/en/latest/install/api.md
+++ b/site/content/en/latest/install/api.md
@@ -26,7 +26,7 @@ The Helm chart for Envoy Gateway
 |-----|------|---------|-------------|
 | certgen.job.annotations | object | `{}` |  |
 | certgen.job.resources | object | `{}` |  |
-| certgen.job.ttlSecondsAfterFinished | int | `0` |  |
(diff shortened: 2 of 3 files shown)


zhaohuabing (approved):
@JuniorJPDJ  Thanks for coming up the workaroud for using EG with ArgoCD. Is it because ArgoCD needs the job to exist for a while before it is finished?

arkodg:
thanks for raising this ! I think `30s` also works here. 
weighing the pros and cons of setting this as a default - I believe the con of this is slower clean up (pressure on the k8s API server) ?

JuniorJPDJ:
@zhaohuabing probably. It's not specified anywhere in the ArgoCD docs, but ArgoCD freezes forever on sync phase when job is deleted too quickly.

60s also allows to read logs without much timing pressure ;)

@arkodg you are probably right, but I'm not sure if it creates any pressure, and if it does, its so small and done only when upgrading/installing the chart, as it's just a hook

arkodg:
> @zhaohuabing probably. It's not specified anywhere in the ArgoCD docs, but ArgoCD freezes forever on sync phase when job is deleted too quickly.
> 
> 60s also allows to read logs without much timing pressure ;)
> 
> @arkodg you are probably right, but I'm not sure if it creates any pressure, and if it does, its so small and done only when upgrading/installing the chart, as it's just a hook

cool, lets get this in, can we reduce this to `30` ?

JuniorJPDJ:
I've not seen any problems, but let me try in my setup again first.
Give me a few minutes :D

arkodg:
@JuniorJPDJ can you run `make generate` and commit the changes

JuniorJPDJ:
rebased and `make generate`d

arkodg (approved):
LGTM thanks !

PR #3157, opened 2024-04-09
feat: add support for optional JWT authentication

### LINKED ISSUE #3081: Support optional JWT authentication for an endpoint
*Description*:
There is a use case when both authenticated and non-authenticated users can access an endpoint, it would help like when it is intended to apply a rate limit based on the user login, say authenticated users will have a dedicated rate limit threshold compared to non-authenticated users.

For example on integrating optional JWT authentication with rate limit:

1. `/api/v1/auth/generate-otp` (authenticated): 100 requests per hour per user login
2. `/api/v1/auth/generate-otp` (non-authenticated): 10 requests per hour per originating IP address

Since our implementation is built upon the specifications outlined in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtrequirement, accomplishing this can be done using the following configuration:

```
name: envoy.filters.http.jwt_authn
typed_config:
  '@type': type.googleapis.com/envoy.extensions.filters.http.jwt_authn.v3.JwtAuthentication
  providers:
    provider_example:
      forward: true
      remote_jwks:
        cache_duration: 300s
        http_uri:
          cluster: raw_githubusercontent_com_443
          timeout: 10s
          uri: https://raw.githubusercontent.com/envoyproxy/gateway/main/examples/kubernetes/jwt/jwks.json
  requirement_map:
    optional_auth:
      requires_any:
        requirements:
          - provider_name: provider_example
          - allow_missing: {}
```

**Proposal**
Therefore in the Envoy Gateway implementation, we can add a new field such as `allowMissing` to the `SecurityPolicy` to support this case, like below:

```
kind: SecurityPolicy
metadata:
  name: jwt-example
spec:
  targetRef:
    kind: HTTPRoute
    name: foo
  jwt:
    providers:
    - name: example
      remoteJWKS:
        uri: https://raw.githubusercontent.com/envoyproxy/gateway/main/examples/kubernetes/jwt/jwks.json
    allowMissing: true
```

Luckily, as we already have the `requiresAny` implementation in the code, integrating it above should be straightforward.
https://github.com/envoyproxy/gateway/blob/941679847a80caa4dda9783cf94aa153bb2837b8/internal/xds/translator/jwt.go#L168-L176

[optional *Relevant Links*:]

ardikabs:
**What type of PR is this?**

Fixes #3081


api/v1alpha1/jwt_types.go
```diff
@@ -8,6 +8,10 @@ package v1alpha1
 // JWT defines the configuration for JSON Web Token (JWT) authentication.
 type JWT struct {
 
+	// AllowMissing determines whether a missing JWT is acceptable, defaulting to false if not specified.
+	// Note: Even if allowMissing is set to true, JWT authentication will still fail if an invalid JWT is presented.
+	AllowMissing *bool `json:"allowMissing,omitempty"`
```
arkodg, line 13: hey @ardikabs, we discussed this in the community meeting, and most preferred `Optional` over `AllowMissing` , can we update this ?

ardikabs: @arkodg 
can we keep the naming convention `AllowMissing` in the IR while using `Optional` in the user interface (CRD)?

arkodg: sure sgtm

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #1393, opened 2023-05-04
Run Envoy Gateway locally on host


arkodg:
*Description*:

- [x]  Design document https://github.com/envoyproxy/gateway/pull/1382
- [x]  Define Provider API https://github.com/envoyproxy/gateway/pull/1402
- [x]  Implement File ResourceProvider https://github.com/envoyproxy/gateway/pull/3159
- [x]   Implement Host InfrastructureProvider https://github.com/envoyproxy/gateway/issues/4481
- [ ]   User Facing docs

[optional *Relevant Links*:]
Relates to Relates to https://github.com/envoyproxy/gateway/issues/1001 & https://github.com/envoyproxy/gateway/issues/1035

arkodg:
the API is in, implementation is remaining, unassigned myself since I won't be able to pick up the implementation anytime soon due to other higher priority issues in the project, if anyone in the community is interested in picking this one up, please go for it

vladimirfx:
Testing Envoy Gateway v1.0.0-rc.1 as a replacement for Cilium Gateway implementation. There are plenty of options (great!) but I can't understand how to use Envoy Gateway on the edge cluster where VIP is assigned to the distinct node. 

Can I configure the Envoy Gateway to run as DaemonSet?

shawnh2:
I would love to see this kicks in after GA, wanna give it a try, so assign myself.

arkodg:
moving this to v1.2.0 to track docs work

PR #3159, opened 2024-04-10
feat: add file resource provider for EG standalone mode

shawnh2:
**What type of PR is this?**

This PR implement two component for file resource provider: *Notifier* and *ResourceStore*.

- _Notifier_ monitors paths(files or dirs) on host via fsnotify, and send Write/Remove event to _ResourceStore_
- _ResourceStore_ load and store resources according to the received event

related #1393

---

There're some TODOs in this PR, tracked by #3213 #3207


zhaohuabing (commented):
As discussed in today's meeting, I prefer to remove the `EnvoyGatewayCustomProvider` and add a flat `EnvoyGatewayFileProvider`(or a better name).

arkodg:
> As discussed in today's meeting, I prefer to remove the `EnvoyGatewayCustomProvider` and add a flat `EnvoyGatewayFileProvider`(or a better name).

Custom is needed to allow for different resource and infra provider

zhaohuabing:
> > As discussed in today's meeting, I prefer to remove the `EnvoyGatewayCustomProvider` and add a flat `EnvoyGatewayFileProvider`(or a better name).
> 
> Custom is needed to allow for different resource and infra provider

Add it later when a use case comes out?

arkodg:
> > > As discussed in today's meeting, I prefer to remove the `EnvoyGatewayCustomProvider` and add a flat `EnvoyGatewayFileProvider`(or a better name).
> > 
> > 
> > Custom is needed to allow for different resource and infra provider
> 
> Add it later when a use case comes out? The File/Host Provider is not a custom provider, it's a known one.

use case exists today - file resource provider and host infra provider to implement standalone case, running EG in a docker container or bare metal directly w/o any API server

Xunzhuo (commented):
Do we plan on releasing this in v1.1.0 ?

cc @arkodg

arkodg:
> Do we plan on releasing this in v1.1.0 ?
> 
> cc @arkodg

yeah I'm a +1 for this, prioritizing other PRs right now, will get to this one soon

shawnh2:
Some tests will be added after the first round of reviews.

shawnh2:
will split changes of `api/v1alpha1/validation/envoygateway_validate.go` into a separate PR, since it's only a refactor and easy to conflict.

sky92zwq:
Hi, does it set `-c` or `--config-path` to switch to file-mode? how to implement the config file?

api/v1alpha1/validation/envoygateway_validate.go
```diff
@@ -79,6 +83,31 @@ func validateEnvoyGatewayKubernetesProvider(provider *egv1a1.EnvoyGatewayKuberne
 	return nil
 }
 
+func validateEnvoyGatewayFileProvider(provider *egv1a1.EnvoyGatewayCustomProvider) error {
+	if provider == nil {
+		return fmt.Errorf("empty custom provider settings for file provider")
+	}
+
+	rType, iType := provider.Resource.Type, provider.Infrastructure.Type
+	if rType != egv1a1.ResourceProviderTypeFile || iType != egv1a1.InfrastructureProviderTypeHost {
```
arkodg, line 92: prefer if we didnt provide this check,  we may have a case where a sidecar hosts a web server and downloads resources and writes them into the directory, and the infra could continue to be provisioned in k8s

api/v1alpha1/validation/envoygateway_validate.go
```diff
@@ -79,6 +83,31 @@ func validateEnvoyGatewayKubernetesProvider(provider *egv1a1.EnvoyGatewayKuberne
...
+
+	if provider.Resource.File == nil {
+		return fmt.Errorf("field 'file' should be specified when resource type is 'File'")
+	}
+
+	if len(provider.Resource.File.Paths) == 0 {
+		return fmt.Errorf("no paths were assigned for file resource provider to watch")
+	}
+
+	if provider.Infrastructure.Host == nil {
```
arkodg, line 104: same as above

internal/cmd/server.go
```diff
@@ -108,21 +110,24 @@ func getConfigByPath(cfgPath string) (*config.Server, error) {
...
 	// https://github.com/envoyproxy/gateway/issues/43
 	ctx := ctrl.SetupSignalHandler()
 
-	// Setup the Extension Manager
-	extMgr, err := extensionregistry.NewManager(cfg)
-	if err != nil {
-		return err
+	// Setup the Extension Manager for Kubernetes provider.
+	var extMgr types.Manager
+	if cfg.EnvoyGateway.Provider.Type == egv1a1.ProviderTypeKubernetes {
```
arkodg, line 120: curious why this is needed ? imo its not required

internal/infrastructure/manager.go
```diff
@@ -36,14 +36,19 @@ type Manager interface {
...
+	case egv1a1.ProviderTypeKubernetes:
 		cli, err := client.New(clicfg.GetConfigOrDie(), client.Options{Scheme: envoygateway.GetScheme()})
 		if err != nil {
 			return nil, err
 		}
 		mgr = kubernetes.NewInfra(cli, cfg)
-	} else {
-		// Kube is the only supported provider type for now.
+
+	case egv1a1.ProviderTypeFile:
```
arkodg, line 48: i dont think this is a valid case - `file` is a valid resource provider, not infra provider

arkodg:
hey @shawnh2 this PR looks good, added some minor comments !

api/v1alpha1/validation/envoygateway_validate.go
shawnh2: sounds reasonable, so is this case a `file provider + k8s infra` combination ?

internal/cmd/server.go
shawnh2: the extension manager will init a k8s client, so it only should be running with k8s provider.

internal/infrastructure/manager.go
shawnh2: this is a old api. updated

internal/provider/file/file.go
shawnh2, line 1: add tests for these new files

arkodg (dismissed):
LGTM thanks !
excited to use this feature

zhaohuabing (dismissed):
🚀

cnvergence (approved):

api/v1alpha1/envoygateway_types.go
```diff
@@ -271,7 +271,11 @@ type EnvoyGatewayCustomProvider struct {
 	// This provider is used to specify the provider to be used
 	// to provide an environment to deploy the out resources like
 	// the Envoy Proxy data plane.
-	Infrastructure EnvoyGatewayInfrastructureProvider `json:"infrastructure"`
+	//
+	// Infrastructure is optional, if no provider is set,
+	// The Kubernetes will be used as the default infrastructure provider.
```
arkodg, line 276: we should make this explicit, and not default to k8s here, because users have expressed a desire for EG to not provision any infra, more in https://envoyproxy.slack.com/archives/C03E6NHLESV/p1722535811941329?thread_ts=1722481296.411989&cid=C03E6NHLESV

internal/provider/file/resources.go
```diff
@@ -0,0 +1,335 @@
...
+
+// TODO(sh2): This function is copied and updated from internal/cmd/egctl/translate.go.
+// This function should be able to process arbitrary number of resources, so we
+// need to come up with a way to extend the GatewayClass and EnvoyProxy field to array
+// instead of single variable in gatewayapi.Resources structure.
+//
+// - This issue is tracked by https://github.com/envoyproxy/gateway/issues/3207
+//
+// convertKubernetesYAMLToResources converts a Kubernetes YAML string into GatewayAPI Resources.
+func convertKubernetesYAMLToResources(str string) (*gatewayapi.Resources, error) {
```
sky92zwq, line 102: should deal with `ReferenceGrant` type resource in this func?

api/v1alpha1/envoygateway_types.go
shawnh2: updated

internal/provider/file/resources.go
shawnh2: Yes, it should. This PR only leverage the basic functionality for file-provider, as described in the comments here, this function will be revisited along with the one in egctl x translate.

shawnh2:
> Hi, does it set `-c` or `--config-path` to switch to file-mode? how to implement the config file?

You can start file-provider mode like `envoy-gateway server -c config.yaml`, with `config.yaml`:

```yaml
kind: EnvoyGateway
gateway:
provider:
  type: Custom
  custom:
    resource:
      type: File
      file:
        paths: ["test/"]
logging:
  level:
    default: info
```

arkodg (approved):
thanks !

Xunzhuo (approved):
LGTM. Thanks

PR #3164, opened 2024-04-10
feat: Wasm extension HTTP code source

zhaohuabing:
This PR implements the HTTP code source for Wasm extension.


api/v1alpha1/wasm_types.go
```diff
@@ -78,8 +78,9 @@ type WasmCodeSource struct {
 	Image *ImageWasmCodeSource `json:"image,omitempty"`
 
 	// SHA256 checksum that will be used to verify the wasm code.
-	// +optional
-	// SHA256 *string `json:"sha256,omitempty"`
+	//
+	// kubebuilder:validation:Pattern=`^[a-f0-9]{64}$`
+	SHA256 string `json:"sha256"`
```
zhaohuabing, line 83: Make SHA256 mandatory for now to avoid downloading wasm and calculate SHA in the control plane.

We can make it optional in the future without breaking changes if we opt to build a cache in the control plane.

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -46,12 +46,12 @@ type EnvoyExtensionPolicySpec struct {
...
 	TargetRef gwapiv1a2.PolicyTargetReferenceWithSectionName `json:"targetRef"`
 
-	// WASM is a list of Wasm extensions to be loaded by the Gateway.
+	// Wasm is a list of Wasm extensions to be loaded by the Gateway.
 	// Order matters, as the extensions will be loaded in the order they are
 	// defined in this list.
 	//
 	// +optional
-	WASM []Wasm `json:"wasm,omitempty"`
+	Wasm []Wasm `json:"wasm,omitempty"`
```
zhaohuabing, line 54: Struggled between `WASM` and `Wasm`. Looks like `Wasm` is [the official abbreviation](https://webassembly.org/). :-)

Xunzhuo: It looks a bit confused, the name is Wasm and the type is the slice of Wasm ([]Wasm)

zhaohuabing: Yeah, it's not pretty. I'm following the existing ExtProc setting here.

Alternative:

```
Wasms []Wasm `json:"wasm,omitempty"`

ExtProcs []ExtProc `json:"extProc,omitempty"`
```

```
WasmList []Wasm `json:"wasm,omitempty"`

ExtProcList []ExtProc `json:"extProc,omitempty"`
```

@arkodg @guydc

zhaohuabing: Follow existing EG API types to use `Wasms` and `ExtProcs`.

arkodg: We are using acronyms here, let's use Wasm and ExtProc

guydc (approved):
LGTM 

Would be great to have a minimal e2e test for one of the sources to demonstrate functional correctness.

api/v1alpha1/wasm_types.go
```diff
@@ -31,14 +31,16 @@ type Wasm struct {
 	// RootID is a unique ID for a set of extensions in a VM which will share a
 	// RootContext and Contexts if applicable (e.g., an Wasm HttpFilter and an Wasm AccessLog).
 	// If left blank, all extensions with a blank root_id with the same vm_id will share Context(s).
-	// RootID *string `json:"rootID,omitempty"`
+	// RootID must match the root_id parameter used to register the Context in the Wasm code.
+	RootID *string `json:"rootID,omitempty"`
```
zhaohuabing, line 35: This is needed and can't be generated by EG. See: https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/921039ae983ce053bf5cba78a85a3c08ff9791e5/proxy_wasm_intrinsics.cc#L27-L28

test/e2e/testdata/wasm.yaml
```diff
@@ -0,0 +1,55 @@
...
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: http-with-http-wasm-source
+  wasm:
+  - name: wasm-filter
+    rootID: my_root_id
+    code:
+      type: HTTP
+      http:
+        url: https://raw.githubusercontent.com/envoyproxy/envoy/main/examples/wasm-cc/lib/envoy_filter_http_wasm_example.wasm
```
zhaohuabing, line 54: @arkodg The HTTP code source does support HTTPS schema.

api/v1alpha1/wasm_types.go
```diff
@@ -92,6 +95,9 @@ const (
 
 	// ImageWasmCodeSourceType allows the user to specify the wasm code in an OCI image.
 	ImageWasmCodeSourceType WasmCodeSourceType = "Image"
+
+	// ConfigMapCodeSourceType allows the user to specify the wasm code in a ConfigMap.
```
arkodg, line 99: can we move this into a separate PR ?

api/v1alpha1/wasm_types.go
arkodg, line 35: why can't we reuse `name` internally here ?

arkodg: If we can generate this field internally, using name, my vote is to do that and not expose another field the user needs to think about how to populate

zhaohuabing: We can't generate `RootID`, See: https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/921039ae983ce053bf5cba78a85a3c08ff9791e5/proxy_wasm_intrinsics.cc#L27-L28

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -46,12 +46,12 @@ type EnvoyExtensionPolicySpec struct {
 	// TargetRef
 	TargetRef gwapiv1a2.PolicyTargetReferenceWithSectionName `json:"targetRef"`
 
-	// WASM is a list of Wasm extensions to be loaded by the Gateway.
+	// Wasm is a list of Wasm extensions to be loaded by the Gateway.
```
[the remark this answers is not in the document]
zhaohuabing: Following the name convention here：https://webassembly.org/

```
WebAssembly (abbreviated Wasm) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable compilation target for programming languages, enabling deployment on the web for client and server applications.
```

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #3153, opened 2024-04-09
tab character in helm template


networkhermit:
*Description*:

I find that envoy-gateway-service.yaml contains a tab character, I believe it is added by accident.

https://github.com/envoyproxy/gateway/blob/c943042cb61f5248df6bf9f38d555d4222f65c0d/charts/gateway-helm/templates/envoy-gateway-service.yaml#L14

yigang01:
can you assign this to me @arkodg

shawnh2:
closed in favor of #3169

PR #3169, opened 2024-04-10
fix: delete tab character

yigang01:
fix：#3153


yigang01:
> /重新测试



> /retest

May I ask how to operate it，again commit code？

shahar-h:
Proposal: you can add add a script that validates there are no tabs in chart yaml files by executing the following for each file: 
```shell
grep -q $'\t' <yaml file>
```
This will prevent such cases in future.

liorokman:
This should have been caught by the `yamllint` tool. Maybe something is wrong with how it's being invoked as part of the build?

liorokman (approved):

shahar-h:
> This should have been caught by the `yamllint` tool. Maybe something is wrong with how it's being invoked as part of the build?

That's because chart is ignored: https://github.com/envoyproxy/gateway/blob/c943042cb61f5248df6bf9f38d555d4222f65c0d/tools/linter/yamllint/.yamllint#L3-L8

liorokman:
But the linter should still be run on the output of the template, after `make generate-manifests` is run.

zirain:
> > This should have been caught by the `yamllint` tool. Maybe something is wrong with how it's being invoked as part of the build?
> 
> That's because chart is ignored:
> 
> https://github.com/envoyproxy/gateway/blob/c943042cb61f5248df6bf9f38d555d4222f65c0d/tools/linter/yamllint/.yamllint#L3-L8

there's a render result located `test/helm/default.yaml`, so I think helm will ignore that.
anyway it's fine to remove it.

zirain (approved):

shahar-h:
Right, but the output is generated correctly with space indentation.

Issue #3170, opened 2024-04-10
Support Additional External Processing Options


guydc:
*Description*:
Envoy Gateway allows an External Processing filter to be configured. However, only a BackendRef can be provided, and all other External Processing attributes are set to Envoy defaults. 

End-users may need to configure additional options to create meaningful integrations with external processing services. The following options are important:
- Defining how bodies and headers (of both requests and responses) are being processed
- Which Dynamic Metadata namespaces are available and mutable for the external processor
- Which Envoy attributes are shared as context with the external processor
- Timeout settings
- FailOpen mode
- Supporting multiple BackendRefs

Below is an example of the API:
```sh
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyExtensionPolicy
metadata:
  name: ext-proc-policy
  namespace: default
spec:
  extProc:
  - backendRef:
      group: ""
      kind: Service
      name: myExtProc
      port: 3000
    processingMode:
      request:
        headers: Send
        body: Buffered
      response:
        headers: Skip
        body: Streamed
    attributes:
      request:
      - xds.route_metadata
      - connection.requested_server_name
      response:
      - request.path
    metadataOptions:
      forwardingNamespaces:
        untyped:
        - envoy.filters.http.ext_authz
      receivingNamespaces:
        untyped:
        - envoy.filters.http.my_custom
    messageTimeout: 5s
    failOpen: true
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
    namespace: default
```

*Relevant Links*:
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto

PR #3171, opened 2024-04-10
api: ext-proc processing mode

guydc:
Allows users to define how request and response header/body are processed by exteral processing service. 

By default, nothing is sent to the external processor. If an empty request/response struct is defined, headers of request/response are sent. If the body attribute is configured, the body is processed according to the defined processing mode (Stream, Buffer, BufferPartial).

xref: #3170


test/cel-validation/envoyextensionpolicy_test.go
```diff
@@ -11,6 +11,8 @@ package celvalidation
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/ptr"
```
shawnh2, line 14: sort the imports please

zirain: please wait https://github.com/envoyproxy/gateway/pull/3177

zirain:
@guydc sorry, linter works as expected now, can you fix it?

zirain:
@guydc sorry, linter works as expected now, can you fix it?

zirain (commented):
LGTM

shawnh2 (approved):
lgtm

api/v1alpha1/ext_proc_types.go
```diff
@@ -9,6 +9,39 @@ import (
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// +kubebuilder:validation:Enum=Streamed;Buffered;BufferedPartial
+type ExtProcBodyProcessingMode string
+
+const (
+	StreamedExtProcBodyProcessingMode          ExtProcBodyProcessingMode = "Streamed"
```
zhaohuabing, line 16: Can we add comments for each mode so they can be generated in API docs?

zhaohuabing:
Minor/non-blocking comment on naming:

`ProcessingMode` is not a very clear name, but I can't think of a better name either.

arkodg (approved):
LGTM thanks !

arkodg:
> Minor/non-blocking comment on naming:
> 
> `ProcessingMode` is not a very clear name, but I can't think of a better name either.

yeah this API field is doing two things
* what to send (body, header)
* when to send it (stream, buffered)

can't think of a better name, so lets go ahead with the envoy specific name, since this feature is completely tied to envoy

zhaohuabing (approved):
LGTM, thanks.

PR #3177, opened 2024-04-11
chore: update golint

zirain:


test/cel-validation/main_test.go
```diff
@@ -28,6 +28,10 @@ import (
 var c client.Client
 
 func TestMain(m *testing.M) {
```
zirain, line 30: Thanks golint! This fix a memory leak on my mac when cel test failed.

arkodg (approved):
LGTM thanks !

PR #3182, opened 2024-04-12
Use native per-route config for basic auth

zhaohuabing:
Switching to native per-route config for basic auth to shrink HCM filter chain size.

[Per-route config for basic auth](https://github.com/envoyproxy/envoy) has been supported in Envoy.


internal/xds/translator/basicauth.go
```diff
@@ -31,97 +32,81 @@ type basicAuth struct {
...
+		basicAuthProto *basicauthv3.BasicAuth
+		basicAuthAny   *anypb.Any
+		err            error
+	)
 
-	if err := basicAuthProto.ValidateAll(); err != nil {
+	basicAuthProto = &basicauthv3.BasicAuth{
+		Users: &corev3.DataSource{
+			Specifier: &corev3.DataSource_InlineBytes{
+				InlineBytes: basicAuth.Users,
```
guydc, line 84: Are we using the first route's security policy basic config to define the top-level HCM "global" config? It's overriden on each route, but makes XDS look a bit strange... Maybe just leave it blank?

guydc (commented):
Big +1 for reducing FC size, one small question on the top-level filter config

internal/xds/translator/basicauth.go
zhaohuabing: We can't leave it blank because the `Users` field is mandatory.

guydc (approved):

arkodg (approved):
thanks !

PR #3187, opened 2024-04-14
fix: use ptr.To[uint32] intead of func()*uint32

ShyunnY:
**What type of PR is this?**

fix: use ptr.To[uint32] intead of func()*uint32


liorokman (approved):
LGTM

zirain (approved):

PR #3190, opened 2024-04-15
feat: Reconcile BackendRefs on EnvoyProxy

davidalger:

This adds support for BackendRefs on EnvoyProxies to the reconciler so the same endpoint discovery used for HTTPRoute refs will be used rather than DNS based discovery using service FQDNs.


internal/gatewayapi/listener.go
```diff
@@ -288,3 +314,45 @@ func processMetrics(envoyproxy *egv1a1.EnvoyProxy) *ir.Metrics {
...
+		endpoints   []*ir.DestinationEndpoint
+		addrType    *ir.DestinationAddressType
+		servicePort v1.ServicePort
+		backendTLS  *ir.TLSUpstreamConfig
+	)
+
+	// TODO (davidalger) Handle case where Service referenced by backendRef doesn't exist
+	serviceNamespace := NamespaceDerefOr(backendRef.Namespace, envoyproxy.Namespace)
+	service := resources.GetService(serviceNamespace, string(backendRef.Name))
+	for _, port := range service.Spec.Ports {
```
davidalger, line 329: This will panic if the backendRef refers to a Service which cannot be found. Ideally this would surface as a message on the status somewhere, similar to how it does on XRoutes.

Status isn't currently being used on EnvoyProxy as far as I can tell. It seems it could make sense to put errors such as this on the Gateway rather than the EnvoyProxy since it's something which will arise when processing the Gateway.

@arkodg @zirain Do either of you have thoughts on where errors such as this should be surfaced?

arkodg: ah
1. any EnvoyProxy errors are surfaced in the `GatewayClass` , lets tackle this separately, and raise a issue to track this because moving it to `Accepted=False` will break everything in the next reconciliation :) 
2. For now lets stick a Direct Response that returns `500` similar to https://github.com/envoyproxy/gateway/blob/2c4d385fcfb02841483cc36e0292175f60249bbb/internal/gatewayapi/route.go#L492

internal/xds/bootstrap/bootstrap.go
```diff
@@ -141,6 +141,7 @@ func GetRenderedBootstrapConfig(opts *RenderBootsrapConfigOptions) (string, erro
 		StatsMatcher     StatsMatcherParameters
 	)
 
+	// TODO (davidalger) push these through XDS?
```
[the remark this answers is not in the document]
arkodg: @zirain is there a way to dynamically setup metrics cluster via xds ?

arkodg:
hey @davidalger @zirain, just realized that there's a massive overlap with this PR and https://github.com/envoyproxy/gateway/pull/3293   , is this anything from this PR that is still required ?

Issue #3198, opened 2024-04-15
optional client certificate for mutual TLS


zufardhiyaulhaq:
*Description*:
Hi, we have a use case of securing one of the APIs with mutual TLS while keeping other APIs with TLS only. currently, mutual TLS is applied on gateway objects where it will protect the listener level.

by making client certificate validation optional, we can add make 1 domain support both mutual TLS and TLS and we can filter APIs that require mTLS with https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert later.

[optional *Relevant Links*:]
most API gateways support this mechanism, for example, Kong or Tyk.
1. https://docs.konghq.com/hub/kong-inc/mtls-auth/
2. https://tyk.io/docs/basic-config-and-security/security/mutual-tls/client-mtls/

arkodg:
this use case is also highlighted in the upstream GEP for this feature https://gateway-api.sigs.k8s.io/geps/gep-91/#deferred
also makes sense to add this knob into CTP https://gateway.envoyproxy.io/v1.0.1/api/extension_types/#clientvalidationcontext

arkodg:
xfcc is being tracked with https://github.com/envoyproxy/gateway/issues/2599

zufardhiyaulhaq:
I am working on this

shawnh2:
closed via #3199

Issue #2614, opened 2024-02-15
e2e: downstream mTLS


arkodg:
*Description*:
>Describe the issue.

Add an e2e for validating clients using a CA Cert as a trust anchor

Relates to https://github.com/envoyproxy/gateway/issues/88

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

guydc:
Also consider testing additional variants of this feature, such as:
- Optional client certificate validation
- XFCC behavior and format

PR #3199, opened 2024-04-15
feat: add requiredClientCertificate on mutual TLS to make it optional

zufardhiyaulhaq:
**What type of PR is this?**
support https://github.com/envoyproxy/gateway/issues/3198
making client certificate validation optional

by making client certificate validation optional, we can add make 1 domain support both mutual TLS and TLS and we can filter APIs that require mTLS with https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert later.

most API gateways support this mechanism, for example, Kong or Tyk.

    https://docs.konghq.com/hub/kong-inc/mtls-auth/
    https://tyk.io/docs/basic-config-and-security/security/mutual-tls/client-mtls/

https://github.com/envoyproxy/gateway/issues/3198


zufardhiyaulhaq:
@arkodg

api/v1alpha1/tls_types.go
```diff
@@ -110,6 +110,12 @@ const (
 // to the Gateway.
 // By default, no client specific configuration is validated.
 type ClientValidationContext struct {
+	// RequiredClientCertificate set to true means that Envoy will reject connections without a valid client certificate.
```
arkodg, line 113: I'd flip the meaning of the noun here, so the default value is `false`
The two terms that come to mind are `optional` and `allowMissing`
https://github.com/envoyproxy/gateway/pull/3157 is trying to address a similar problem for JWT

arkodg:
thanks for the quick PR, it looks good ! the only comment I have is around naming

api/v1alpha1/tls_types.go
zufardhiyaulhaq: we cannot set the default value to false to support backward compatibility.
should we use `DisabledClientCertificateValidation`?

arkodg: thats too wordy imo

zufardhiyaulhaq: since it's under ClientValidation, should we set this as Enabled or Disabled? hmm

arkodg: the meaning of the noun picked, should convey the default meaning, e.g. `optional` or `allowMissing`  highlights that the default is `false`

zufardhiyaulhaq: if that is the case, I propose with `Required` and the default is true.
since it's under ClientValidationContext struct.

or do you have better name @arkodg?

arkodg: I prefer `optional` which defaults to `false`
will bring this up in the community meeting this week to speed up consensus on this

zufardhiyaulhaq: I am using `Required` and set default to `true`.
I am not sure why I cannot use `!` whn using optional. somehow translator_test.go failing

zhaohuabing: Both names sound good to me. However, I prefer to use `false` as default to keep consistent with other APIs.

zufardhiyaulhaq: Hi @arkodg @zhaohuabing change `Required` to `Optional` and use false by default.
please help recheck! thanks

api/v1alpha1/tls_types.go
```diff
@@ -110,6 +110,10 @@ const (
 // to the Gateway.
 // By default, no client specific configuration is validated.
 type ClientValidationContext struct {
+	// Optional set to false means that Envoy will reject connections without a valid client certificate.
```
arkodg, line 113:
```suggestion
	// Optional set to true accepts connections even when a client doesn't present a certificate.
	// Defaults to false, which rejects connections without a valid client certificate.
```

internal/ir/xds.go
```diff
@@ -281,6 +281,8 @@ type TLSConfig struct {
 	Certificates []TLSCertificate `json:"certificates,omitempty" yaml:"certificates,omitempty"`
 	// CACertificate to verify the client
 	CACertificate *TLSCACertificate `json:"caCertificate,omitempty" yaml:"caCertificate,omitempty"`
+	// RequiredClientCertificate to enforce client certificate
+	RequiredClientCertificate bool `json:"requiredClientCertificate,omitempty" yaml:"requiredClientCertificate,omitempty"`
```
arkodg, line 285:
```suggestion
	RequireClientCertificate bool `json:"requireClientCertificate,omitempty" yaml:"requiredClientCertificate,omitempty"`
```

arkodg:
minor comments around naming and doc string, else LGTM

guydc:
@zufardhiyaulhaq - can you please rerun code-generation?

zufardhiyaulhaq:
@guydc done, please check

guydc (approved):
LGTM. 

Would be great if we could also add an e2e test for this feature in a follow-up PR, but I guess that this is blocked on #2614 for now.

arkodg (approved):
LGTM thanks !

Issue #3200, opened 2024-04-15
gzip compression on /stats/prometheus


zufardhiyaulhaq:
*Description*:
envoy gateway expose prometheus format metrics on /stats/prometheus. prometheus query this endpoints and there are network cost between prometheus & envoy gateway.

implement compression to reduce network traffic cost.

[optional *Relevant Links*:]
N/A

zirain:
fixed by #3201

PR #3201, opened 2024-04-15
feat: support HTTP compression on stats/prometheus

zufardhiyaulhaq:
**What type of PR is this?**
 envoy gateway expose prometheus format metrics on /stats/prometheus. prometheus query this endpoints and there are network cost between prometheus & envoy gateway.

add HTTP compression support for gzip, brotli, zstd

add HTTP compression support on /stats/prometheus
reduce network cost for getting the gateway metrics

https://github.com/envoyproxy/gateway/issues/3200


arkodg:
we should reuse the struct from https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/compression_types.go
and create a list type similar to https://github.com/envoyproxy/gateway/blob/9f00be3c0bbf145dee7cc97c7d0f110e0b83232b/api/v1alpha1/backendtrafficpolicy_types.go#L99C14-L99C28

arkodg:
since this translation may get complex, probably makes sense to generate a go struct, marshal into YAML and append that into the bootstrap

api/v1alpha1/envoyproxy_metric_types.go
```diff
@@ -79,4 +79,10 @@ type ProxyOpenTelemetrySink struct {
 type ProxyPrometheusProvider struct {
 	// Disable the Prometheus endpoint.
 	Disable bool `json:"disable,omitempty"`
+	// Enable the compression on Prometheus endpoint.
+	EnableCompression bool `json:"enableCompression,omitempty"`
```
zirain, line 83: I think we can make `CompressionLibrary` to `*string`?

zufardhiyaulhaq: I am using compression type as suggested by @arkodg

zirain: BTW, can you point out something like: `compression is not a silver bullet, it just a trade off. It will reduce the cost of network, but will increase the usage of cpu and memory.`

zufardhiyaulhaq: sure @zirain

zufardhiyaulhaq: this is done @zirain, please help check

zirain (approved):

zufardhiyaulhaq:
is the test flaky?

zufardhiyaulhaq:
@zirain let me know if we can merge this.
Thanks!

zirain:
@arkodg PTAL when you're back.

api/v1alpha1/envoyproxy_metric_types.go
```diff
@@ -79,4 +79,7 @@ type ProxyOpenTelemetrySink struct {
 type ProxyPrometheusProvider struct {
 	// Disable the Prometheus endpoint.
 	Disable bool `json:"disable,omitempty"`
+	// Configure the compression on Prometheus endpoint. Compression is useful in situations when bandwidth is scarce and large payloads can be effectively compressed at the expense of higher CPU load.
+	// +optional
+	Compression *Compression `json:"compression,omitempty"`
```
arkodg, line 84: the BTP API instantiated this as a list https://github.com/envoyproxy/gateway/blob/d162184e550b003c4095b49297aa6d5bee206783/api/v1alpha1/backendtrafficpolicy_types.go#L99 to support multiple compression settings, to support heterogenous clients (and compression settings), do we also need to make it a list here ?

zufardhiyaulhaq: could 1 envoy route support multiple compression libraries? AFAIK we need to select only 1 library right?

zirain: make it to list but limit maxitem to 1?

zufardhiyaulhaq: that doesn't make sense IMO

arkodg: I think envoy can support multiple compressor libraries per route, which is the argument for making it a list
but realistically the persona who creates the EnvoyProxy resource with this setting, also knows beforehand the compression supported by the peer, so a single scalar value is fine here

arkodg (approved):

Issue #3174, opened 2024-04-11
Support multiple backends in TCPRoute


shuji-2019:
*Description*:
> Multiple backends is supported in HTTPRoute, GRPCRoute and TLSRoute, but not [TCPRoute](https://github.com/envoyproxy/gateway/blob/a24263c4983bbffd8acdf913ffc97a2a7ba114a6/internal/gatewayapi/route.go#L981) and [UDPRoute](https://github.com/envoyproxy/gateway/blob/a24263c4983bbffd8acdf913ffc97a2a7ba114a6/internal/gatewayapi/route.go#L849). It's a necessary feature when backend with new version is rollout.

arkodg:
thanks for surfacing this
* needs to be fixed in https://github.com/envoyproxy/gateway/blob/d383ba54020a1bd38e2787eff230b055713f35e3/internal/gatewayapi/route.go#L991
* and in https://github.com/envoyproxy/gateway/blob/d383ba54020a1bd38e2787eff230b055713f35e3/internal/gatewayapi/route.go#L859

PR #3212, opened 2024-04-17
tcproute/udproute support multiple backends

zirain:
fix: https://github.com/envoyproxy/gateway/issues/3174


internal/gatewayapi/route.go
```diff
@@ -846,24 +846,22 @@ func (t *Translator) processUDPRouteParentRefs(udpRoute *UDPRouteContext, resour
...
+			// Skip further processing if route destination is not valid
+			if ds == nil || len(ds.Endpoints) == 0 {
+				valid = false
+				continue
+			}
+
+			destSettings = append(destSettings, ds)
+		}
+		if !valid {
 			continue
```
arkodg, line 862: we should add a status for the `invalid` case

zirain: correct me if I'm wrong, looks like we didn't do that thing for all routes?

internal/gatewayapi/route.go
```diff
@@ -846,24 +846,22 @@ func (t *Translator) processUDPRouteParentRefs(udpRoute *UDPRouteContext, resour
...
-		}
 
-		backendRef := udpRoute.Spec.Rules[0].BackendRefs[0]
-		ds, _ := t.processDestination(backendRef, parentRef, udpRoute, resources)
-		// Skip further processing if route destination is not valid
-		if ds == nil || len(ds.Endpoints) == 0 {
+		valid := true
+		for _, backendRef := range udpRoute.Spec.Rules[0].BackendRefs {
+			ds, _ := t.processDestination(backendRef, parentRef, udpRoute, resources)
+			// Skip further processing if route destination is not valid
```
arkodg, line 853: imo we should follow the same approach we use in HTTP case, where for an invalid destination, we track the weight as invalid, and build a weighted cluster based on valid and invalid weights across the multiple backendRefs

internal/gatewayapi/route.go
```diff
@@ -978,23 +975,21 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
...
-				"One and only one backend is supported",
-			)
-			continue
-		}
 
-		backendRef := tcpRoute.Spec.Rules[0].BackendRefs[0]
-		ds, _ := t.processDestination(backendRef, parentRef, tcpRoute, resources)
-		// Skip further processing if route destination is not valid
-		if ds == nil || len(ds.Endpoints) == 0 {
+		valid := true
```
arkodg, line 979: same as above

internal/gatewayapi/route.go
zirain: May I know what's the benifit for doing that?

arkodg: its defined in the spec https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.TCPRouteRule

zirain: seems different to https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPBackendRef?

arkodg: also mentioned in https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule

internal/xds/translator/route.go
```diff
@@ -221,27 +218,30 @@ func buildXdsStringMatcher(irMatch *ir.StringMatch) *matcherv3.StringMatcher {
...
 }
 
-func buildXdsRouteAction(httpRoute *ir.HTTPRoute) *routev3.RouteAction {
-	return &routev3.RouteAction{
-		ClusterSpecifier: &routev3.RouteAction_Cluster{
-			Cluster: httpRoute.Destination.Name,
-		},
-		IdleTimeout: idleTimeout(httpRoute),
+func buildXdsRouteAction(backendWeights *ir.BackendWeights) *routev3.RouteAction {
+	if backendWeights.Invalid == 0 {
```
arkodg, line 222: can we add a comment here saying we only use a weighted cluster when we have invalid weights

internal/xds/translator/route.go
```diff
@@ -71,13 +71,10 @@ func buildXdsRoute(httpRoute *ir.HTTPRoute) (*routev3.Route, error) {
...
-		var routeAction *routev3.RouteAction
-		if httpRoute.BackendWeights.Invalid != 0 {
-			// If there are invalid backends then a weighted cluster is required for the route
-			routeAction = buildXdsWeightedRouteAction(httpRoute)
-		} else {
-			routeAction = buildXdsRouteAction(httpRoute)
-		}
+		backendWeights := httpRoute.Destination.ToWeightedBackend()
+		routeAction := buildXdsRouteAction(backendWeights)
+		routeAction.IdleTimeout = idleTimeout(httpRoute)
```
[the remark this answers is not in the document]
arkodg: thanks for adding this note :). the diff is massive, stared with this file and the IR and this approach makes sense

internal/ir/xds.go
```diff
@@ -926,14 +925,33 @@ func (r RouteDestination) Validate() error {
 	return errs
 }
 
+func (r *RouteDestination) ToWeightedBackend() *BackendWeights {
```
arkodg, line 928: why not `ToBackendWeights` ?

internal/gatewayapi/testdata/custom-filter-order.out.yaml
```diff
@@ -236,11 +236,10 @@ xdsIR:
         mergeSlashes: true
       port: 10080
       routes:
-      - backendWeights:
-          invalid: 1
-          valid: 0
-        directResponse:
```
arkodg, line 242: why is `directResponse` getting removed ?

internal/gatewayapi/testdata/backendtrafficpolicy-status-conditions.out.yaml
```diff
@@ -575,11 +574,10 @@ xdsIR:
         mergeSlashes: true
       port: 10080
       routes:
-      - backendWeights:
-          invalid: 1
-          valid: 0
-        directResponse:
```
arkodg, line 581: why is `directResponse` getting removed

internal/gatewayapi/testdata/gateway-with-listener-with-tcproute-with-mismatch-port-protocol.out.yaml
```diff
@@ -94,3 +94,10 @@ xdsIR:
     - address: 0.0.0.0
       name: envoy-gateway/gateway-1/tcp
       port: 10162
+      routes:
+      - destination:
+          name: tcproute/default/tcproute-1/rule/-1
+          settings:
+          - weight: 1
+        name: tcproute/default/tcproute-1
+        tls: {}
```
arkodg, line 103: curious why we have an empty `tls` struct here

internal/gatewayapi/testdata/gateway-with-listener-with-udproute-with-multiple-backends.in.yaml
```diff
@@ -26,8 +26,8 @@ udpRoutes:
       rules:
         - backendRefs:
             - name: service-1
-              port: 8080
+              port: 8162
```
arkodg, line 29: how did the port change here ?

internal/ir/xds.go
```diff
@@ -926,14 +925,33 @@ func (r RouteDestination) Validate() error {
...
+	}
+
+	return w
+}
+
 // DestinationSetting holds the settings associated with the destination
 // +kubebuilder:object:generate=true
 type DestinationSetting struct {
 	// Weight associated with this destination.
-	// Note: Weight is not used in TCP/UDP route.
```
arkodg, line 933: can we add a comment here saying that invalid endpoints are represents with a non zero weight with an empty endpoints list ?

internal/gatewayapi/testdata/backendtrafficpolicy-status-conditions.out.yaml
zirain: fixed, see https://github.com/envoyproxy/gateway/pull/3212/commits/115fdb1f3692ff97846f523c4706fba25eb59ab8

internal/gatewayapi/route.go
```diff
@@ -516,7 +516,8 @@ func (t *Translator) processGRPCRouteRules(grpcRoute *GRPCRouteContext, parentRe
 
 		// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
 		for _, ruleRoute := range ruleRoutes {
-			if ruleRoute.Destination == nil && ruleRoute.Redirect == nil {
+			noValidBackends := ruleRoute.Destination == nil || ruleRoute.Destination.ToBackendWeights().Valid == 0
```
arkodg, line 519: shouldnt this be `invalidBackendsExist` instead of `noValidBackends`  i.e. some instead of all ?

zirain: noValidBackends = invalidBackendsExist | emptyBackends

internal/gatewayapi/testdata/gateway-with-listener-with-tcproute-with-mismatch-port-protocol.out.yaml
zirain: see https://github.com/envoyproxy/gateway/pull/3212/commits/9084108b4c3ac8736a0a17a9764f534029e71261

internal/gatewayapi/route.go
arkodg: but `noValidBackends` also implies validBackends == 0 which is not true here

internal/gatewayapi/route.go
```diff
@@ -1158,12 +1136,12 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
 
 	backendNamespace := NamespaceDerefOr(backendRef.Namespace, route.GetNamespace())
 	if !t.validateBackendRef(backendRefContext, parentRef, route, resources, backendNamespace, routeType) {
-		return nil, weight
+		return &ir.DestinationSetting{Weight: &weight}
```
arkodg, line 1139: can we add a comment here highlighting that this is referring to an invalid weight

internal/gatewayapi/route.go
```diff
@@ -493,30 +493,31 @@ func (t *Translator) processGRPCRouteRules(grpcRoute *GRPCRouteContext, parentRe
...
 				}
+				route.Destination.Settings = append(route.Destination.Settings, ds)
 			}
 		}
 
 		// If the route has no valid backends then just use a direct response and don't fuss with weighted responses
 		for _, ruleRoute := range ruleRoutes {
-			if ruleRoute.Destination == nil && ruleRoute.Redirect == nil {
+			noValidBackends := ruleRoute.Destination == nil || ruleRoute.Destination.ToBackendWeights().Valid == 0
+			if noValidBackends && ruleRoute.Redirect == nil {
```
arkodg, line 520: isnt something like this needed in `processHTTPRouteRules` as well ?

internal/gatewayapi/route.go
zirain: in current code validBackends == 0 equals emptyBackends

internal/gatewayapi/route.go
zirain: good catch

arkodg (approved):
LGTM thanks !

shawnh2 (approved):
lgtm

Issue #3210, opened 2024-04-16
Merge Gateway Tracing Root Trace service name random


owenhaynes:
*Description*:
Traces that are emitted when using merge gateway  have a random gateway name as  the service name. They do not use the correct gateway name that the route is attached to.

*Repro steps*:

Run merge gateway with multiple gateways with named listeners with routes attached. You should notice pretty quickly that the service name does not match up with the gateway that the route and host belongs to

 
*Environment*:
envoy gateway 1.0.1

cnvergence:
thanks for reporting @owenhaynes

PR #3214, opened 2024-04-17
fix: match serviceName in tracing with mergedGateways topology

cnvergence:
**What type of PR is this?**

When mergeGateways is enabled, tracing should set its service name to GatewayClass name, as there will be only one service for every gateway listener.

Fixes https://github.com/envoyproxy/gateway/issues/3210


zirain (commented):
LGTM, can you add a translator test?

shawnh2 (approved):
lgtm

cnvergence:
just thinking if we should use `GatewayClassName . EnvoyNamespace` format instead

arkodg (approved):

zirain (approved):

zirain:
it would be better to see a xds translator test, but LGTM now.

PR #3215, opened 2024-04-17
chore: use golang 1.N.P version syntax

guydc:
CodeQL is reporting an invalid golang toolchain version syntax:

> As of Go 1.21, toolchain versions [must use the 1.N.P syntax](https://go.dev/doc/toolchain#version).
> 1.22 in tools/src/setup-envtest/go.mod does not match this syntax and there is no additional toolchain directive, which may cause some go commands to fail.
> CodeQL also found 9 other warnings like this. See the workflow log for details.


zirain (approved):

shawnh2 (approved):
Thanks for spotting this, maybe that is the reason why I always receive `toolchains not found for go1.22` error on my local dev env.

Issue #3069, opened 2024-03-31
Support cluster settings for non-xRoute BackendRefs


guydc:
*Description*:
Currently, Envoy Gateway supports a variety of envoy cluster settings through the `BackendTrafficPolicy` resource, such as:
- TCP Keep Alive
- Timeouts
- Health Checks
- Load Balancing algorithm
- Circuit Breakers

Envoy Gateway is adopting the use of `BackendRef` to represent external services in various resources, such as SecurityPolicy (ExtAuth), EnvoyExtensionPolicy (ExtProc), EnvoyProxy (OTEL sinks). With this change, reuse of existing backend-traffic policies that attach to Services, such as `BackendTLSPolicy`, is now possible for these backends. However, it's not possible to attach a `BackendTrafficPolicy` to Services. As a result, the above-mentioned cluster settings are not configurable for non-xRoute referenced backends. 

**Options:**
1. Split BackendTrafficPolicy to two policies. One policy will contain all fields that have an affinity to Envoy listeners (retries, rate limits, Fault... ) and will attach to `Gateway` and `xRoutes`, while the other will contain all fields that have affinity to Envoy Clusters and will attach to `Services`. 
2. Support attachment of BackendTrafficPolicy to Services.
3. Create an extended BackendRef resource that supports configuration of relevant cluster settings, similar to support for weights in Gateway-API: https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.BackendRef

arkodg:
+1 for 3.

davidalger:
+1 for 3.

arkodg:
rethinking this one, we should also discuss option 4 - adding cluster settings into the `Backend` API . This could eliminate duplication of cluster settings if multiple Policies are targeting the same backend.

liorokman:
Looking at all of the suggested implementation options, I think none of them will work.

The problem is that most (if not all) of the knobs available in `BackendTrafficPolicy` actually are attached to an Envoy Cluster definition, while the `Backend` resource and the `BackendTargetRef` arrays refer specifically to endpoints inside the Cluster definition.

> * Split BackendTrafficPolicy to two policies. One policy will contain all fields that have an affinity to Envoy listeners (retries, rate limits, Fault... ) and will attach to Gateway and xRoutes, while the other will contain all fields that have affinity to Envoy Clusters and will attach to Services.
> * Support attachment of BackendTrafficPolicy to Services.

[...]

PR #3218, opened 2024-04-18
feat: support backend tls settings with envoyproxy

alexwo:
This change allows to enhance the security and adaptability of connections to backend services. By allowing administrators to specify TLS settings, such as which cryptographic ciphers and TLS protocol versions to use, it allows to better align their security posture with industry best practices and regulatory requirements.

**Approach**:
- Enable specification of default values through a TLS settings section in the EnvoyProxy backend.
- Use these settings as the default TLS configurations whenever a BackendTLS policy is specified for a backend, applying the EnvoyProxy defaults to affected backends.

Once the BackendTLS policy or an alternative method for providing these settings is implemented/accepted in upstream, opt for more specific settings than those provided by envoyproxy backendtls settings.

https://github.com/envoyproxy/gateway/issues/2901


guydc:
cc @zhaohuabing who was also looking into backend MTLS in  #2984

internal/xds/translator/utils.go
```diff
@@ -161,3 +161,17 @@ func createExtServiceXDSCluster(rd *ir.RouteDestination, tCtx *types.ResourceVer
 	}
 	return nil
 }
+
+func ParseTLSProtocol(tlsVersion ir.TLSVersion) (tlsv3.TlsParameters_TlsProtocol, error) {
```
guydc, line 165: can we reuse the util that we have in CTP for this mapping?

internal/gatewayapi/testdata/envoyproxy-tls-settings.in.yaml
```diff
@@ -0,0 +1,133 @@
+gateways:
```
guydc, line 1: Can we also add MTLS + TLS Settings tests for ext-auth and ext-proc, demonstrating how this applied to these sorts of backends?

guydc (commented):
Thanks for this!
Some comments on reuse of existing types.

arkodg:
thanks for working on this @alexwo !
this PR is doing two things
1. added support for specifying `clientCert` which doesnt exist in upstream today
2. added support for specifying tls params such as `cipher_suites`

both 1. and 2. are added into the global `EnvoyProxy` API
should 2. live inside BTP, so it can be modified on a per route (per upstream) basis ?

api/v1alpha1/envoyproxy_types.go
```diff
@@ -116,6 +117,34 @@ type EnvoyProxySpec struct {
...
+// BackendTLSConfig describes the BackendTLS configuration for Envoy Proxy.
+type BackendTLSConfig struct {
+	// ClientCertificateRef defines the reference to a Kubernetes Secret that contains
+	// the client certificate and private key for Envoy to use when connecting to
+	// backend services and external services, such as ExtAuth, ALS, OpenTelemetry, etc.
+	// +optional
+	ClientCertificateRef *gwapiv1.SecretObjectReference `json:"clientCertificateRef,omitempty"`
+	// Ciphers defines only support the specified cipher list when negotiating BackendTLS 1.0-1.2 (this setting has no effect when negotiating BackendTLS 1.3).
+	// +optional
+	Ciphers []string `json:"ciphers,omitempty"`
```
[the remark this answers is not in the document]
alexwo: We can re-use it, however, it also hot ClientValidation and ALPNProtocol, this are less interesting in this context. 
Would it be better to re-structure TLSSettings and include the common only?

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -78,7 +78,26 @@ func (t *Translator) processBackendTLSPolicy(
 	}
 
 	status.SetAcceptedForPolicyAncestors(&policy.Status, ancestorRefs, t.GatewayControllerName)
-
+	// apply defaults as per envoyproxy
+	if resources.EnvoyProxy != nil {
```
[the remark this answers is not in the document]
alexwo: Right, if support for additional resources is implemented in upstream, we can move / extend this logic to other areas as well. Do you see an advantage in making this change now? If we don’t currently need to apply it o the gateway level for example, perhaps we can delay this refactor till gateway add support or we see a use case. wdyt?

alexwo:
> thanks for working on this @alexwo ! this PR is doing two things
> 
> 1. added support for specifying `clientCert` which doesnt exist in upstream today
> 2. added support for specifying tls params such as `cipher_suites`
> 
> both 1. and 2. are added into the global `EnvoyProxy` API should 2. live inside BTP, so it can be modified on a per route (per upstream) basis ?

Hi @arkodg ,

Probably this make sense, 

> thanks for working on this @alexwo ! this PR is doing two things
> 
> 1. added support for specifying `clientCert` which doesnt exist in upstream today
> 2. added support for specifying tls params such as `cipher_suites`
> 
> both 1. and 2. are added into the global `EnvoyProxy` API should 2. live inside BTP, so it can be modified on a per route (per upstream) basis ?
Hi @arkodg ,
This will feature better flexibility this way, perhaps it can make more sense for (2) like settings, @guydc WDYT?

internal/xds/translator/utils.go
alexwo: yes, i think we should be able i will check it.

guydc:
> It can allow for better flexibility, perhaps it can make more sense for (2) settings, @guydc WDYT?

The major downside I see in using BTP is that it can only attach to xRoute backends. So, if we need to apply TLS settings for other backends (extproc, extauth, OTEL, ALS ... ), we will need to add TLS setting in the scope of of #3069. 

I'm fine with both options, as long as we can support TLS settings for both xRoute and non-xRoute backends. @arkodg - WDYT?

arkodg:
> > It can allow for better flexibility, perhaps it can make more sense for (2) settings, @guydc WDYT?
> 
> The major downside I see in using BTP is that it can only attach to xRoute backends. So, if we need to apply TLS settings for other backends (extproc, extauth, OTEL, ALS ... ), we will need to add TLS setting in the scope of of #3069.
> 
> I'm fine with both options, as long as we can support TLS settings for both xRoute and non-xRoute backends. @arkodg - WDYT?

yeah for non xRoute we'll need to embed it into our `backendRef`

by moving 2. into BTP, we add the ability to fine tune settings, but now `tls` settings are spread out in 3 places, which can get confusing
* `clientCertificateRef` in `EnvoyProxy.Spec.BackendTLS`
* `tlsParams` in `BackendTrafficPolicy.Spec.TLS`
* `caCertificateRefs` in `BackendTLSPolicy.Spec.Validation`

alexwo:
> * BackendTLSPolicy

Does it make sense to move clientCertificateRef, to BTP as well?  this way we can have things less spread.

guydc:
> If we go with the BTP direction, does it make sense to move clientCertificateRef, to BTP as well? This way we can have things less spread.

I think so. It's possible that different backends will require a different client cert. It's not uncommon for organizations to use private PKI(s) for client certs, and so backends may have different trust stores and CAs. If we go with EnvoyProxy-level config, we basically assume that all backends can be configured to trust the EG client certificate.

api/v1alpha1/envoyproxy_types.go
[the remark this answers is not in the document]
arkodg: we can create a struct that represents the `TLSParams` and embed that struct in both place and handle the `ClientCertificateRef` and `ClientValidation` fields seperately

arkodg:
@guydc @alexwo as platform administrators, would you say that the feature is necessary, because you are trying to enforce specific tls parameters (ciphers, tls versions). If this is true, then the argument of putting it in `EnvoyProxy.Spec.BackendTLS` can be made for 2. since for this use case, we'd want the platform admin to enforce these settings w/o allowing the app dev (route level BTP owner) to override it.

guydc:
> as platform administrators, would you say that the feature is necessary, because you are trying to enforce specific tls parameters (ciphers, tls versions)

Yes, our use case is related to defining system-wide upstream TLS settings, and so EnvoyProxy makes a lot of sense to me.

arkodg:
> > as platform administrators, would you say that the feature is necessary, because you are trying to enforce specific tls parameters (ciphers, tls versions)
> 
> Yes, our use case is related to defining system-wide upstream TLS settings, and so EnvoyProxy makes a lot of sense to me.

cool, i'm a +1 for this as well

api/v1alpha1/envoyproxy_types.go
```diff
@@ -116,6 +117,20 @@ type EnvoyProxySpec struct {
...
+}
+
+// BackendTLSConfig describes the BackendTLS configuration for Envoy Proxy.
+type BackendTLSConfig struct {
+	// ClientCertificateRef defines the reference to a Kubernetes Secret that contains
+	// the client certificate and private key for Envoy to use when connecting to
+	// backend services and external services, such as ExtAuth, ALS, OpenTelemetry, etc.
+	// +optional
+	ClientCertificateRef *gwapiv1.SecretObjectReference `json:"clientCertificateRef,omitempty"`
+	TLSSettings          `json:",inline"`
```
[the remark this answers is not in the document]
alexwo: In this context, it make sense as everything is under BackendTLS, probably its less clear if we do something like: backendTLS.tls.

guydc (approved):
LGTM, Thanks!

internal/xds/translator/translator.go
```diff
@@ -768,10 +768,29 @@ func buildXdsUpstreamTLSCASecret(tlsConfig *ir.TLSUpstreamConfig) *tlsv3.Secret
 func buildXdsUpstreamTLSSocketWthCert(tlsConfig *ir.TLSUpstreamConfig) (*corev3.TransportSocket, error) {
 
 	var tlsCtx *tlsv3.UpstreamTlsContext
+	var tlsParams *tlsv3.TlsParameters
```
arkodg, line 771: can we reuse https://github.com/envoyproxy/gateway/blob/27a770f4db8ff275591934f4bc3080ba0d1aaddc/internal/xds/translator/listener.go#L572 ?

alexwo: /retest

guydc (approved):

arkodg (approved):
LGTM thanks !

sherwinwangs:
> clientCertificateRef

that would be a better design, how someone will be contribute to this feature

Issue #3048, opened 2024-03-28
Support modifying http2 listener settings


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it

As a user, I'd like to increase the current defaults of these below settings for my setup
Defaults are derived from https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge#configuring-envoy-as-an-edge-proxy

- [x] [initial_stream_window_size](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-stream-window-size)

- [x] [initial_connection_window_size](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-connection-window-size)

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

guydc:
+1

arkodg:
@guydc does this fall into `connection` or a new `http2` field within https://gateway.envoyproxy.io/latest/api/extension_types/#clienttrafficpolicyspec ?

guydc:
I think that a new `http2` field would be more appropriate.

arkodg:
discussed this in the meeting today, `http2` seems like a popular option, will go with this in the init PR

PR #3222, opened 2024-04-19
HTTP2 settings API

zhaohuabing:
Related: https://github.com/envoyproxy/gateway/issues/3048


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -200,6 +204,23 @@ type HTTP10Settings struct {
...
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	InitialStreamWindowSize *uint32 `json:"initialStreamWindowSize,omitempty"`
+
+	// InitialConnectionWindowSize sets the initial window size for HTTP/2 connections.
+	// If not set, the default value is 1 MiB(1024*1024).
+	// +kubebuilder:validation:Minimum=65535
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	InitialConnectionWindowSize *uint32 `json:"initialConnectionWindowSize,omitempty"`
```
zhaohuabing, line 221: Should EG enforce that connection window size >= stream window size?

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -200,6 +204,23 @@ type HTTP10Settings struct {
 	UseDefaultHost *bool `json:"useDefaultHost,omitempty"`
 }
 
+// HTTP2Settings provides HTTP/2 configuration on the listener.
+type HTTP2Settings struct {
+	// InitialStreamWindowSize sets the initial window size for HTTP/2 streams.
```
arkodg, line 209: can we create a custom type for `resource.Quantiy` defined in https://github.com/envoyproxy/gateway/blob/2c4d385fcfb02841483cc36e0292175f60249bbb/api/v1alpha1/connection_types.go#L24 and reuse it here ?

zhaohuabing: kubebuilder won't generate validation for alias type. If we define a new type, we need to redefine every `resource.Quantiy` function that EG used for the new type `Size`.


https://github.com/envoyproxy/gateway/actions/runs/8772985239/job/24072530543?pr=3222

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -707,6 +707,7 @@ func translateListenerConnection(connection *egv1a1.Connection, httpIR *ir.HTTPL
 
 	if connection.BufferLimit != nil {
 		bufferLimit, ok := connection.BufferLimit.AsInt64()
+		connection.BufferLimit.DeepCopy()
```
guydc, line 710: should this be assigned to a var?

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -200,6 +204,19 @@ type HTTP10Settings struct {
...
+	// If not set, the default value is 64 KiB(64*1024).
+	// +optional
+	InitialStreamWindowSize *Size `json:"initialStreamWindowSize,omitempty"`
+
+	// InitialConnectionWindowSize sets the initial window size for HTTP/2 connections.
+	// If not set, the default value is 1 MiB(1024*1024).
+	// +optional
+	InitialConnectionWindowSize *Size `json:"initialConnectionWindowSize,omitempty"`
```
guydc, line 217: Two additional knobs that can be important here are:
- `max_concurrent_streams` - the secure-by-default value of 100 is quite strict, and users may want to increase it.
- `override_stream_error_on_invalid_http_message` - see discussion on how this field should be configured based on wether envoy is a L1 or L2 proxy: https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two. In other words, there's no "correct" value here, and we should make it configurable for users based on their topology.

guydc (commented):
Thanks! Would love to see two more knobs supported.

api/v1alpha1/clienttrafficpolicy_types.go
zhaohuabing: `override_stream_error_on_invalid_http_message` should be added in http1,http2,and http3 settings. Could we address this in another PR?

guydc: created #3245

guydc (approved):
LGTM, Thanks!

arkodg (approved):
LGTM thanks !

PR #3223, opened 2024-04-19
docs: (#3137) show 'apply yaml' either from stdin, or from file using…

### LINKED ISSUE #3137: docs: yaml artifacts in guides should code-highlight

eitansuez:
… tabs

**What type of PR is this?**

* docs: retrofits tasks to provide yaml in two forms, using tabs:  1. a command can be copy-pasted to apply directly with kubectl, and 2. show the yaml, useful for reviewing the configuration and saving to a file.

this is a convenience in the docs.  oftentimes it's really nice to be able to show yaml with code highlighting for ease of legibility, but it's also convenient to provide a kubectl apply -f command that can copy-paste directly into the shell.

Fixes #3137


eitansuez:
change applied to just one task: "http request headers".  let me know what you think and i'll start working on retrofitting all other tasks.

arkodg (approved):
LGTM thanks !

eitansuez:
> LGTM thanks !

i had to push a lint fix.  it looks like it might have invalidated your review?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #3229, opened 2024-04-21
crds: remove status

zirain:
since EG move to use `gwapiv1a2.PolicyStatus`, this won't work any more.


shawnh2 (approved):
LGTM! Users can also scrape the policy status via `egctl x status xpolicy` now.

guydc (approved):

PR #3230, opened 2024-04-21
refactor: remove explicit test case definition for xds translator test

### LINKED ISSUE #2229: Remove the need to specify test name in xds translator lib
*Description*:
>Describe the issue.

The Xds Translator test defined here https://github.com/envoyproxy/gateway/blob/87bc96fededc5de24ef14776be91c1802453cde1/internal/xds/translator/translator_test.go#L44
relies on the test being explicitly defined in this function.

This can be improved by removing the need to do so, similar to what is done in the Gateway API Translate test https://github.com/envoyproxy/gateway/blob/87bc96fededc5de24ef14776be91c1802453cde1/internal/gatewayapi/translator_test.go#L45

this should reduce number of merge conflicts in the project

[optional *Relevant Links*:]

shawnh2:
**What type of PR is this?**

Fixes #2229


internal/xds/translator/translator_test.go
```diff
@@ -41,357 +41,166 @@ var (
...
-			name: "http-route-rewrite-url-host",
-		},
-		{
-			name: "http-route-timeout",
-		},
-
-		{
-			name: "ratelimit",
+		"multiple-listeners-same-port": {
+			requireSecrets: true,
```
arkodg, line 78: can this check also be removed, and secrets be asserted if a secrets file exists ?

internal/xds/translator/testdata/out/xds-ir/mixed-tls-jwt-authn.listeners.yaml
```diff
@@ -14,26 +15,6 @@
           initialStreamWindowSize: 65536
           maxConcurrentStreams: 100
         httpFilters:
-        - name: envoy.filters.http.jwt_authn
```
arkodg, line 17: why is this changing ?

shawnh2: because this test file has been ignored (not explicit defined in test cases) until now, its content is outdated.

internal/xds/translator/translator_test.go
shawnh2: good idea!

shawnh2: by removing `requireSecrets` and checking secrets from translator result, I found some `xxx.secret.yaml` were missing, so added in this PR as well

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #2999, opened 2024-03-23
Support attachment of policies to multiple Gateways


guydc:
*Description*:
Gateway operators may want to attach common policies to a group of gateways. Currently, duplicated instances of a policy need to be created and attached to each Gateway to achieve this. This can lead to significant configuration duplication, especially when many virtual gateways are merged to a single envoy deployment. 

`EnvoyPatchPolicy` introduced a concept of [GatewayClass-atttached](https://github.com/envoyproxy/gateway/blob/0be74abf7c0f863064a03001f27fdcfe4c7a2279/site/content/en/v1.0.0/user/extensibility/envoy-patch-policy.md?plain=1#L107) policies for the merge gateways scenario. 

This form of attachment can be extended to other policies like `ClientTrafficPolicy`, `BackendTrafficPolicy`, `SecurityPolicy`, `EnvoyExtensionPolicy` and `BackendTLSPolicy`.

arkodg:
Im a -1 on this

2 ways to apply policies (Gateway, Route) is sufficient to solve most cases, 3 ways makes the concepts of `defaults`, `overrides` and `merges` very complicated to comprehend for the user and compute for the translator

* For `EnvoyPatchPolicy` we only allow attaching to the `GatewayClass` if `mergeGateways` is set

guydc:
> im a +1 for pluralizing `targetRefs` so multiple `Gateway`s can be targeted , would ideally like to get guidance on this from upstream before implementing this, this is something we can time box

Yes, that can also work pretty well, as long as selectors are supported, and users don't have to maintain an explicit gateway list.

arkodg:
cc @AliceProxy who was a big +1 for ^

davidalger:
+1 on pluralizing `targetRefs`, this would also be useful for targeting multiple `HTTPRoutes` via the same `BackendTrafficPolicy` for example where groups of routes may require an elevated timeout.

arkodg:
related discussion in upstream https://github.com/kubernetes-sigs/gateway-api/pull/2966#discussion_r1565161217

sadovnikov:
Gateway API [v1.1.0-rc1](https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.1.0-rc1) was released yesterday. It includes "_the targetRef field is now a targetRefs list and these references no longer include a namespace field_", which is good. However, it seems to be done only for the `BackendTLSPolicy` and to be implemented as a list of `targetRef`. 

[...]

Issue #2913, opened 2024-03-13
Should be able to attach OIDC SecurityPolicy to Gateways


mt-inside:
*Description*:
Currently, a `SecurityPolicy` that configures OIDC will only work if attached to an HTTPRoute. This is because the `callbackURL` and `logoutPath` have to match the path prefix of the HTTPRoute. (Although SecurityPolicies attached to Gateways aren't rejected, which is confusing)

I have a use-case which I believe is common: one Gateway for all my "internal apps" (Grafana, etc). Rather than bind this to an internal network, I'd like to do "zero trust" - expose it to the internet, and enable OIDC SSO. So I have many HTTPRoutes attached to that Gateway, all of which want OIDC.

Currently I have to make a SecurityPolicy for each one, and attach each one to the relevant HTTPRoute. This is kinda messy from an admin point-of-view.
What's worse is that because the SecurityPolicy's callbackURL has to match not only the HTTPRoute's path but also the OAuth2 client's callback path, I have to make an IdP OAuth2 Client for each HTTPRoute. At my company we use gsuite, which has no API for this, so every time I deploy a new app I have to click-ops a new OAuth2 client, deal with its secret key, etc.

I would be much better if I could make one OAuth2 client, one SecurityPolicy, and attach it to an entire Gateway.

arkodg:
Id vote for the user creating the `fallback` route for this case i.e. creating a prefix match for `/` with a non existent backend.

zhaohuabing:
EG needs to report an error when a `SecurityPolicy` targeting a `Gateway`, and we need to add this workaround to the user docs.

arkodg:
@zhaohuabing can you highlight the current limitation here wrt routing, redirect and oidc filter ?

zhaohuabing:
The current per-route filter implementation adds a filter for each OIDC/Basic Auth/Ext Auth, etc. config in the HCM http filter chains, and disable them by default, these filter are enable at route-leve base on the targeted route/gateway of the SecurityPolicy.

This approach works well for Basic Auth/Ext Auth, but OIDC is a different:  the redirectURL and logoutURL must match the HTTPRoute in order to make the IDP redirection works. So we can't apply the same OIDC config for the HTTRoutes of a Gateway because their matching paths are different. 

As @arkodg suggested, a fallback HTTPRoute on `/` should solve this

```
 Gateway1
    Listener1 443
      HTTPRoute1 /myapp1
      HTTPRoute2 /myapp2
[...]

PR #3252, opened 2024-04-23
feat: allow the same OIDC configuration target multiple HTTPRoutes

zhaohuabing:
This PR changes OIDC cookie suffix to OIDC `clientID` so multiple `HTTPRoutes` can share the same OIDC configuration.

Related: https://github.com/envoyproxy/gateway/issues/3253

@sadovnikov


internal/gatewayapi/securitypolicy.go
```diff
@@ -589,7 +589,18 @@ func (t *Translator) buildOIDC(
...
+	// cookie suffix. This allows us to share the same cookies across multiple
+	// oauth filters at multiple routes. Once users log in at one route, they will
+	// be able to access other routes with the same OIDC redirect URL (the same
+	// OIDC clientID) without logging in again.
+	//
+	// zhaohuabing: This is a workaround to allow multiple routes to share the same OIDC configuration.
+	// We should change this back to use policy UID after Gateway API supports
+	// targeting a policy to multiple routes.
+	// See https://github.com/kubernetes-sigs/gateway-api/discussions/2927#discussioncomment-8991869
+	suffix := utils.Digest32(*oidc.RedirectURL)
```
sadovnikov, line 603: The `oidc` is not present in every security policy and the `redirectURL` might be missing

internal/gatewayapi/securitypolicy.go
```diff
@@ -589,7 +589,18 @@ func (t *Translator) buildOIDC(
 	}
 
 	// Generate a unique cookie suffix for oauth filters
-	suffix := utils.Digest32(string(policy.UID))
+	// We use the digest of the redirect URL to generate the cookie suffix so that
```
arkodg, line 592: i dont think this is is needed 
https://github.com/envoyproxy/gateway/pull/2944 adds support for a single oauth2 filter based on the policy name instead of 1 per route, so using a suffix off policy should work

zhaohuabing: This supports multiple routes with the same OIDC configuration.

internal/gatewayapi/securitypolicy.go
zhaohuabing: You mean `RedirectURL` may be null? We should use `ClientID` here.

internal/gatewayapi/securitypolicy.go
arkodg: the issue linked is #2913 which is referring to targeting a `SecurityPolicy` to a Gateway with oidc set

suggest opening a new issue to handle this case of how to author different `SecurityPolicies` targeting different routes and yet reuse the same cookie prefix, for that case, my suggestion would be to expose an optional `CookiePrefix` field in the API

zhaohuabing: I believe only the OIDC configurations with the same `ClientID` should/must use the same `CookiePrefix`. If this is true, exposing `CookiePrefix` to API would be unnecessary and result in abuse.

zhaohuabing:
@sadovnikov During today's meeting, @arkodg  mentioned that EG is going to support targeting Policy to multiple targets. I suggest we hold this and focus on https://github.com/envoyproxy/gateway/issues/2999.

sadovnikov:
@zhaohuabing sounds good. Do you have an idea when https://github.com/kubernetes-sigs/gateway-api will have a new release?

zhaohuabing:
> @zhaohuabing sounds good. Do you have an idea when https://github.com/kubernetes-sigs/gateway-api will have a new release?

[The decision has been made](https://gateway-api.sigs.k8s.io/geps/gep-2648/#multiple),  and It doesn't rely on gateway API since `SecurityPolicy` is EG's extension. I just need to work on this when I have some circles.

PR #3260, opened 2024-04-24
fix: Add SourceMatchType validation for acceptable values

owenhaynes:
Adds validation to SourceMatch to only allow the accepted values. This is to prevent people using a lowercase  version of "Distinct" which will then default to using "Exact" for the ratelimiting. 

I lost a few hours to this 😄


zirain:
you maye need to run `make -k gen-check` locally.

owenhaynes:
I see no errors when running `make -k gen-check` locally

arkodg (approved):

shawnh2 (approved):

PR #3261, opened 2024-04-24
ci: add osv-scanner workflow

### LINKED ISSUE #3167: Dependencies vulnerability scanning
*Description*:
Currently There is an underlying dependency vulnerability scan as part of [OpenSSF Scorecard workflow](https://github.com/envoyproxy/gateway/blob/main/.github/workflows/scorecard.yml) which is triggered on push to main branch.
OpenSSF is using osv-scanner behind the scenes:
https://github.com/ossf/scorecard/blob/49c0eed3a423f00c872b5c3c9f1bbca9e8aae799/docs/checks.md#vulnerabilities
https://google.github.io/osv-scanner/
We currently have score `7` since there are several vulnerabilities:
image

It is not clear what is the origin of the vulnerabilities unless we run `osv-scanner` locally.
Do you think we should add an additional scan as part of separate GH workflow or keep it like it is?
We can use either [trivy](https://github.com/marketplace/actions/aqua-security-trivy#using-trivy-to-scan-your-git-repo) or [osv-scanner](https://google.github.io/osv-scanner/github-action/). I think `osv-scanner` should be preferred so we can be aligned with OpenSSF scorecard.

In addition, should this be triggered also on PRs? It has the downside of blocking PRs in case of unfixed vulnerabilities.

Please share your thoughts.

shahar-h:
* Added [osv-scanner](https://google.github.io/osv-scanner/) workflow in order to detect vulnerabilities in project dependencies.
This tool was chosen in order to be aligned with OpenSSF scorecard, which also [uses](https://github.com/ossf/scorecard/blob/49c0eed3a423f00c872b5c3c9f1bbca9e8aae799/docs/checks.md#vulnerabilities) it behind the scenes to determine vulnerabilities score.
* The workflow is triggered on both PR and merge to `main` branch. When triggered from PR it only fails in case new vulnerabilities are introduced as part of the PR. See more details [here](https://google.github.io/osv-scanner/github-action/#scan-on-pull-request).
* I fixed some of the detected vulnerabilities as part of this PR(let me know if you think this should be in a separate PR):
  * Updated `sigs.k8s.io/controller-tools` in `tools/sec/controller-gen` to v0.15.0.
  * Updated go version from `1.12` to `1.22.2` in `site` folder.
  * Updated `github.com/bufbuild/buf` to `v1.31.0`.
  * Updated `google.golang.org/protobuf` to `v1.33.0`
  * Updated `sigs.k8s.io/controller-runtime/tools/setup-envtest` to latest.
* There is a single active [vulnerability](https://github.com/advisories/GHSA-8r3f-844c-mc37) which will be resolved with the next `google.golang.org/grpc/cmd/protoc-gen-go-grpc` release. See related [issue](https://github.com/grpc/grpc-go/issues/7092).
* I also noticed that `tools/src/protoc-gen-go` and `tools/src/protoc-gen-go-grpc`  are missing from [dependabot configuration](https://github.com/envoyproxy/gateway/blob/main/.github/dependabot.yml). Is there any particular reason for that? If not I can add them in a separate PR.

Fixes #3167


zirain (commented):
LGTM

guydc (approved):
LGTM, thanks!

arkodg (approved):
LGTM thanks for bringing this in !

Issue #3000, opened 2024-03-24
shutdownmanager image should be customizable


vixns:
Hi,

Shutdownmanager image is hardcoded at https://github.com/envoyproxy/gateway/blob/0be74abf7c0f863064a03001f27fdcfe4c7a2279/api/v1alpha1/shared_types.go#L29

When testing an image from a fork, or when publishing to a private repository, this hardcoded value is on the way.

arkodg:
Blocked on https://github.com/envoyproxy/gateway/issues/2808

zirain:
duplicate with #2808

Issue #2808, opened 2024-03-06
Figure out shutdownmanager's image repo and tag for dev and release


arkodg:
> > A more accurate regex for tag is better. Sometimes the sha will start with v too
> 
> @Xunzhuo good point... agreed, detection would need to be more specific than a letter prefix.
> 
> I'm looking at how this could be implemented and not sure how we can detect the tag given how the release works. Currently it's just copying the docker image: https://github.com/envoyproxy/gateway/actions/runs/8154910922/job/22289300255#step:5:1
> 
> Since the image `envoyproxy/gateway-dev:<sha>` is copied to `envoyproxy/gateway:<tag>` the ld flag is going to be the `<sha>` from the build pipeline:
> 
> ```
> $ docker run --rm envoyproxy/gateway:v1.0.0-rc.1 version
> ENVOY_GATEWAY_VERSION: v1.0.0-rc.1
> ENVOY_PROXY_VERSION: distroless-v1.29-latest
> GATEWAYAPI_VERSION: v1.0.0
> SHUTDOWN_MANAGER_VERSION: 7d48480
> GIT_COMMIT_ID: 7d48480fed10100a699ac360b2e1ace6eb49deab
> ```
> 
> `ENVOY_GATEWAY_VERSION` is set to the release based on what's in the `VERSION` file, but this won't work for `SHUTDOWN_MANAGER_VERSION` which needs to match the current build so the right image will be used in e2e tests.
> 
> Nothing immediately stands out as a simple and clean way of solving this without changing how the image is released so the ld flags can be set accordingly.

oh fun, thanks for pointing this out 
the fact that we dont rebuild a image when releasing, but instead promote a image (retag it), makes this chicken egg situation worse (gateway image defining a gateway image)

_Originally posted by @arkodg in https://github.com/envoyproxy/gateway/issues/2789#issuecomment-1981587689_

arkodg:
From a correctness perspective
[envoyproxy/gateway:v1.0.0-rc.1](https://hub.docker.com/layers/envoyproxy/gateway/v1.0.0-rc.1/images/sha256-ef041655bb9cb9327479f6c3aa46f08050531f729986fdb653583dbd8944a5de?context=explore) is the same as [envoyproxy/gateway-dev:7d48480](https://hub.docker.com/layers/envoyproxy/gateway-dev/7d48480/images/sha256-ef041655bb9cb9327479f6c3aa46f08050531f729986fdb653583dbd8944a5de?context=explore)

[...]

PR #3269, opened 2024-04-24
feat: expose shutdownmanager image to EnvoyGageway API

zhaohuabing:
This PR exposes shutdown manager image to EnvoyGageway API, so it can be populated via a helm value.

fix :https://github.com/envoyproxy/gateway/issues/3000 https://github.com/envoyproxy/gateway/issues/2808


internal/infrastructure/kubernetes/proxy/testdata/deployments/shutdown-manager.yaml
```diff
@@ -249,7 +249,7 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.name
-        image: envoyproxy/gateway-dev:v1.2.3
+        image: privatereop/envoyproxy/gateway-dev:v1.2.3
```
arkodg, line 252:
```suggestion
        image: privaterepo/envoyproxy/gateway-dev:v1.2.3
```

arkodg:
thanks for adding this !
minor comment around typo, but this looks good !
as part of this fix, we need to make two more changes
* remove existing logic around shutdown versions using build flags https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/internal/cmd/version/version.go#L24
cc @davidalger 
* by default, always make sure we set `EnvoyGateway.Shutdown.Image` in helm

is this something you want to tackle in this PR or follow up ?

zhaohuabing:
> remove existing logic around shutdown versions using build flags
[gateway/internal/cmd/version/version.go](https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/internal/cmd/version/version.go#L24)

The default image version is displayed in the `envoy-gateway version` command, as the `ENVOY_PROXY_VERSION` does. Or should we remove the version info for both the ShutdownManager and the Envoy Proxy?

tools/make/helm.mk
```diff
@@ -33,4 +33,4 @@ helm-generate:
 
 helm-template: ## Template envoy gateway helm chart.
 	@$(LOG_TARGET)
-	helm template eg charts/gateway-helm --set deployment.envoyGateway.image.tag=latest > ./test/helm/default.yaml --namespace=envoy-gateway-system
+	helm template eg charts/gateway-helm --set deployment.envoyGateway.image.tag=latest --set config.envoyGateway.provider.kubernetes.shutdownManager.image="docker.io/envoyproxy/gateway-dev:latest" > ./test/helm/default.yaml --namespace=envoy-gateway-system
```
arkodg, line 36: why do we need the extra `set` ?

arkodg (approved):
LGTM thanks !

shawnh2 (approved):
lgtm

cnvergence (approved):
🚀

tools/make/helm.mk
zirain: > | config.envoyGateway.provider.kubernetes.shutdownManager.image | string | `"docker.io/envoyproxy/gateway:latest"` | |

make this reasonable?

wimnat:
When is this likely to get released?  I've just had to roll back to 0.6 because our k8s clusters have a gatekeeper rule that blocks images that are not from our private repo.  Without this fix to override the shutdown-manager image, I'm stuck :(

PR #3273, opened 2024-04-25
feat: support custom HTTP filter ordering

zhaohuabing:
This PR allows users to adjust HTTP filter orders to suit their specific use cases.
It also:
* adds a constraint limiting each `EnvoyExtensionPolicy` to have a maximum of 16 wasm/extproc extensions.
* Fixes the JSON tag in `FilterPosition` API

Note: this PR doesn't handle the ordering of individual wasm/extproc extensions. This will be addressed in a follow-up PR, as it involves API change and may need more discussion.

Related: https://github.com/envoyproxy/gateway/issues/2571 https://github.com/envoyproxy/gateway/pull/2993


internal/ir/xds.go
```diff
@@ -84,6 +84,8 @@ type Xds struct {
 	UDP []*UDPListener `json:"udp,omitempty" yaml:"udp,omitempty"`
 	// EnvoyPatchPolicies is the intermediate representation of the EnvoyPatchPolicy resource
 	EnvoyPatchPolicies []*EnvoyPatchPolicy `json:"envoyPatchPolicies,omitempty" yaml:"envoyPatchPolicies,omitempty"`
+	// FilterOrder holds the custom order of the HTTP filters
```
arkodg, line 87: should this be part of the Translator context instead ?

arkodg: referring to here https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/internal/xds/translator/translator.go#L43

arkodg: oh nvm I see both

internal/xds/translator/httpfilters.go
```diff
@@ -140,16 +144,64 @@ func (o OrderedHTTPFilters) Swap(i, j int) {
...
 
-	for i := 0; i < len(filters); i++ {
-		filters[i] = orderedFilters[i].filter
+	l := list.New()
+	for i := 0; i < len(orderedFilters); i++ {
+		l.PushBack(orderedFilters[i])
 	}
+
+	// Sort the filters in the custom order.
+	for i := 0; i < len(filterOrder); i++ {
```
guydc, line 161: Is this a topological sorting problem? Can we use an established algorithm and library to create this sort? It would make it a bit easier to be convinced in the functional correctness.

guydc (commented):
Do we need to validate that we don't have cycles like:
```
  spec:
    filterOrder:
      - name: envoy.filters.http.jwt_authn
        before: envoy.filters.http.cors
      - name: envoy.filters.http.cors
        before: envoy.filters.http.jwt_authn
```

arkodg:
> Do we need to validate that we don't have cycles like:
> 
> ```
>   spec:
>     filterOrder:
>       - name: envoy.filters.http.jwt_authn
>         before: envoy.filters.http.cors
>       - name: envoy.filters.http.cors
>         before: envoy.filters.http.jwt_authn
> ```

imo if we apply in order, we wont hit this issue, right ?

internal/xds/translator/httpfilters.go
```diff
@@ -91,28 +95,28 @@ func newOrderedHTTPFilter(filter *hcmv3.HttpFilter) *OrderedHTTPFilter {
...
+	case isFilterType(filter, basicAuthFilter):
 		order = 4
 	case isFilterType(filter, oauth2Filter):
 		order = 5
-	case filter.Name == jwtAuthn:
+	case isFilterType(filter, jwtAuthn):
 		order = 6
 	case isFilterType(filter, extProcFilter):
-		order = 7
+		order = 100 + mustFilterIndex(filter.Name)
```
arkodg, line 111: curious around why the num changed here, did the logic change, are we trying to create a hole to accomodate more filters ?

internal/xds/translator/httpfilters.go
arkodg: this logic could be simplified
1. init - ordered list
2. loop across filterOrder (overrideList)
3. find `name`, and delete it from list
4. find `before` or `after` and insert it before/after in list

Its still m (orig list)*n (override list) with both sizes being fairly small, but easier to comprehend

internal/xds/translator/httpfilters.go
zhaohuabing: Because there're more than one wasm/extproc filters for an HTTPRoute.

zhaohuabing:
> > Do we need to validate that we don't have cycles like:
> > ```
> >   spec:
> >     filterOrder:
> >       - name: envoy.filters.http.jwt_authn
> >         before: envoy.filters.http.cors
> >       - name: envoy.filters.http.cors
> >         before: envoy.filters.http.jwt_authn
> > ```
> 
> imo if we apply in order, we wont hit this issue, right ?

Add a check for a more friendly UI.

internal/xds/translator/httpfilters.go
zhaohuabing: > Is this a [topological sorting](https://en.wikipedia.org/wiki/Topological_sorting) problem? Can we use/reference an established algorithm (and possibly library) to create this sort? It would make it a bit easier to be convinced in the functional correctness.

There's no good way to combine two DAGs: the origin one and the custom one.  Would the current implementation be good enough since we have checked the cycle dependencies in Gateway API reconciler?

internal/xds/translator/httpfilters.go
```diff
@@ -223,6 +280,16 @@ func isFilterType(filter *hcmv3.HttpFilter, filterType string) bool {
 	return strings.HasPrefix(filter.Name, filterType)
 }
 
+// mustFilterIndex returns the index of the filter in its filter type.
+func mustFilterIndex(filterName string) int {
```
arkodg, line 284:
```suggestion
func filterIndex(filterName string) int {
```

arkodg (approved):
LGTM thanks
thanks for adding those extra comments which will help the next dev looking at the code !

zirain (approved):

PR #3274, opened 2024-04-25
feat: Optionally override Oauth Cookie Names

sam-burrell:
**What type of PR is this?**
Added ability to optionally specify a cookieSuffix in the OIDC spec.

This is necessary to link up with the JWT element of the SecurityPolicy to ultimately make passing claimHeaders from the IdToken upstream for authentication in upstream apps clear and easy.

For example it would be used like this

```yaml
kind: SecurityPolicy
metadata:
  name: grafana-public-test
spec:
  targetRef:
    kind: HTTPRoute
    name: grafana-public-test
  oidc:
    provider:
      issuer: "FOOO"
    clientID: "BAR"
    clientSecret:
      kind: Secret
      name: internal-app-client-secret
    logoutPath: "/logout"
    redirectURL: https://grafana.com/oauth2/callback
    cookieSuffix: "grafana"
    scopes:
      - openid
      - email
  jwt:
    providers:
      - name: "FOO"
        extractFrom:
          cookies:
           - IdToken-grafana
        claimToHeaders:
          - claim: email
            header: X-Email
          - claim: group
            header: X-Group
        remoteJWKS:
          uri: "https://PROVIDER/.well-known/jwks.json"
```


arkodg:
this makes a lot of sense
cc @mt-inside @zhaohuabing 

2 open questions
* do we need to revisit the `prefix` naming here ?
* should this be a prefix, suffix or the ability to rewrite entire cookie name ?
e.g. this is the config from the istio auth service 
https://github.com/istio-ecosystem/authservice/blob/f2ef0e41de8a2d664b35a51bd43748e0ca4d10e8/config/v1/oidc/config.proto#L176

hoping we rely on work done in the past, to influence naming here

zhaohuabing:
I'd vote to keep the current naming, but allow users to specify a custom suffix, and explain how to avoid session name conflicts in the EG doc if they want to do that.

A better approach would be to send the ID token through an `x-envoy-gateway-id-token` header and avoid users messing with the cookies, but it needs some work in Envoy upstream.

sam-burrell:
I am happy to do anything else required to move this forward. 

Side note - great work with this we are using envoy gateway successfully as our main ingress in our 300 plus service ecs to eks migration.

zhaohuabing:
> should this be a prefix, suffix or the ability to rewrite entire cookie name ?

I'm inclined to add a suffix because the default one is a digest of the policy UID, it's kind of strange to put a random string as a prefix.

arkodg:
> > should this be a prefix, suffix or the ability to rewrite entire cookie name ?
> 
> I'm inclined to add a suffix because the default one is a digest of the policy UID, it's kind of strange to put a random string as a prefix.

should we just allow the user to set the entire name ?

zhaohuabing:
> > > should this be a prefix, suffix or the ability to rewrite entire cookie name ?
> > 
> > 
> > I'm inclined to add a suffix because the default one is a digest of the policy UID, it's kind of strange to put a random string as a prefix.
> 
> should we just allow the user to set the entire name ?

Agree, entire name is better.

api/v1alpha1/oidc_types.go
```diff
@@ -30,6 +30,11 @@ type OIDC struct {
 	// +kubebuilder:validation:Required
 	ClientSecret gwapiv1b1.SecretObjectReference `json:"clientSecret"`
 
+	// The optional cookie suffix to be added to Bearer and IdToken cookies in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	// If not specified, uses a randomly generated suffix
+	CookieSuffix *string `json:"cookieSuffix,omitempty"`
```
arkodg, line 36:
```suggestion
	IdTokenCookieName *string `json:"idTokenCookieName,omitempty"`
```

api/v1alpha1/oidc_types.go
```diff
@@ -30,6 +30,11 @@ type OIDC struct {
...
+
```
arkodg, line 37: or even `idTokenName`

zhaohuabing:
@sam-burrell Could you please address @arkodg 's comment? So we can move on this.

zetaab:
ping @sam-burrell could you check the review comments

denniskniep:
Hi,

this feature is very welcome 👍 

I propose additionally to @arkodg´s review comment to make all cookie names configurable (and not only idToken cookie):

https://github.com/envoyproxy/gateway/blob/6dcf78aee3aa88fd7feae6fb7449ba6f3b9c57ff/internal/xds/translator/oidc.go#L161-L167

* bearerTokenCookieName
* oauthHmacCookieName
* oauthExpiresCookieName
* idTokenCookieName
* refreshTokenCookieName

arkodg:
> Hi,
> 
> this feature is very welcome 👍
> 
> I propose additionally to @arkodg´s review comment to make all cookie names configurable (and not only idToken cookie):
> 
> https://github.com/envoyproxy/gateway/blob/6dcf78aee3aa88fd7feae6fb7449ba6f3b9c57ff/internal/xds/translator/oidc.go#L161-L167
> 
> * bearerTokenCookieName
> * oauthHmacCookieName
> * oauthExpiresCookieName
> * idTokenCookieName
> * refreshTokenCookieName

if we need have predictable names for all the cookies then I'd suggest making the struct a little more tiered
```
cookieName:
  idToken:
  ........
```

zhaohuabing:
> Hi,
> 
> this feature is very welcome 👍
> 
> I propose additionally to @arkodg´s review comment to make all cookie names configurable (and not only idToken cookie):
> 
> https://github.com/envoyproxy/gateway/blob/6dcf78aee3aa88fd7feae6fb7449ba6f3b9c57ff/internal/xds/translator/oidc.go#L161-L167
> 
> * bearerTokenCookieName
> * oauthHmacCookieName
> * oauthExpiresCookieName
> * idTokenCookieName
> * refreshTokenCookieName

Exposing BearerTokenCookieName and idTokenCookieName to API makes sense to me, but others should not be in the API as they are trivial details of the Envoy OAuth2 filter implementation.

sam-burrell:
@zhaohuabing @arkodg 

We have refactored this and now made this the ability to optionally overide oauth cookie names.
Is there anything else we can help with?

denniskniep:
> > Hi,
> > this feature is very welcome 👍
> > I propose additionally to @arkodg´s review comment to make all cookie names configurable (and not only idToken cookie):
> > https://github.com/envoyproxy/gateway/blob/6dcf78aee3aa88fd7feae6fb7449ba6f3b9c57ff/internal/xds/translator/oidc.go#L161-L167
> > 
> > 
> > * bearerTokenCookieName
> > * oauthHmacCookieName
> > * oauthExpiresCookieName
> > * idTokenCookieName
> > * refreshTokenCookieName
> 
> Exposing BearerTokenCookieName and idTokenCookieName to API makes sense to me, but others should not be in the API as they are trivial details of the Envoy OAuth2 filter implementation.
> 
> @denniskniep do you need these cookies somewhere?

@zhaohuabing BearerTokenCookieName and idTokenCookieName is sufficient for my current use case

arkodg:
hey this diff looks good !
can we add a test in the [gateway-api lib](https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata) and the [xds translator lib](https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/xds-ir) to avoid any future regressions ? the `make testdata` target should help generate the o/p data for you once you've written the input config

coro:
@arkodg Added, thanks!

internal/gatewayapi/testdata/securitypolicy-with-oidc-custom-cookies.out.yaml
```diff
@@ -0,0 +1,169 @@
...
+  status:
+    ancestors:
+    - ancestorRef:
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: gateway-1
+        namespace: envoy-gateway
+      conditions:
+      - lastTransitionTime: null
+        message: HMAC secret envoy-gateway-system/envoy-oidc-hmac not found
```
arkodg, line 132: something is off, `Accepted=False`, you may have forgotten to include the `Secret`, the `security` IR field is also empty

api/v1alpha1/oidc_types.go
```diff
@@ -75,3 +81,17 @@ type OIDCProvider struct {
...
 	TokenEndpoint *string `json:"tokenEndpoint,omitempty"`
 }
+
+// OIDCCookieNames defines the names of cookies to use in the Envoy OIDC filter.
+type OIDCCookieNames struct {
+	// The name of the cookie used to store the BearerToken in the
+	// [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
+	// If not specified, defaults to "BearerToken-(randomly generated uid)"
+	// +optional
+	BearerToken *string `json:"bearerToken,omitempty"`
```
zhaohuabing, line 91: The name `BearerToken` is confusing in this context.  Let's rename it `AccessToken` to keep consistent with https://github.com/envoyproxy/gateway/pull/3423

```suggestion
	AccessToken *string `json:"accessToken,omitempty"`
```

internal/xds/translator/oidc.go
```diff
@@ -172,6 +172,17 @@ func oauth2Config(oidc *ir.OIDC) (*oauth2v3.OAuth2, error) {
 			Resources:  oidc.Resources,
 		},
 	}
+
+	if oidc.CookieNameOverrides != nil &&
+		oidc.CookieNameOverrides.BearerToken != nil {
+		oauth2.Config.Credentials.CookieNames.BearerToken = *oidc.CookieNameOverrides.BearerToken
```
zhaohuabing, line 178: Let's also use `AccessToken` as the default name.

```
BearerToken:  fmt.Sprintf("AccessToken-%s", oidc.CookieSuffix)
```

coro:
@arkodg @zhaohuabing Added your changes, thanks!

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM thanks!

arkodg:
@coro can you run `make testdata` again ? looks like some test o/p changed

coro:
@arkodg I believe that was unrelated to my change - it seems the example certificate used in that particular test with a diff (`internal/gatewayapi/testdata/gateway-with-listener-with-valid-multiple-tls-configuration-with-same-algorithm-different-fqdn.out.yaml`) had expired May 24th. I've rebased onto the latest commit and it has resolved that diff.

arkodg (approved):

PR #3281, opened 2024-04-26
chore: Bump k8s.io

zirain:
fix: #3237


internal/provider/kubernetes/controller.go
```diff
@@ -909,147 +909,185 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
+			}),
+			&predicate.TypedGenerationChangedPredicate[*gwapiv1.GatewayClass]{},
+			predicate.NewTypedPredicateFuncs[*gwapiv1.GatewayClass](r.hasMatchingController))); err != nil {
 		return err
 	}
 
 	// Only enqueue EnvoyProxy objects that match this Envoy Gateway's GatewayClass.
-	epPredicates := []predicate.Predicate{
-		predicate.GenerationChangedPredicate{},
-		predicate.ResourceVersionChangedPredicate{},
```
zirain, line 930: @arkodg is this a must?

zhaohuabing (approved):

guydc (approved):

internal/provider/kubernetes/controller.go
arkodg: can you track the orig commit for `ResourceVersionChangedPredicate` ? did the type get removed ?

zirain: I can find a typed Predicate about that, actually I didn't understand why EG need it only for this.

PR #3283, opened 2024-04-26
fix e2e EnvoyShutdown

### LINKED ISSUE #3262: flaky: e2e TestEGUpgrade/EnvoyShutdown
https://github.com/envoyproxy/gateway/actions/runs/8811709532/job/24187972908?pr=2665#step:6:17406

shawnh2:
**What type of PR is this?**

- add `test` into codecov ignore path
- trying to fix #3262 by increasing the HTTPRoute Duration and move load generation after the deployment has been restarted
- fix the order of merge-gateways e2e test in makefile

Fixes #3262


guydc:
> move load generation after the deployment has been restarted

Hi @shawnh2! The purpose of the tests is to run load during the restart and assert that there's no loss of requests... Running it after restart means that we're not really testing for hitless shutdown anymore..

shawnh2:
> > move load generation after the deployment has been restarted
> 
> 
> 
> Hi @shawnh2! The purpose of the test is to run load during the restart and assert that there's no loss of requests... Running it after restart means that we're not really testing for hitless shutdown anymore.. 

Thanks for the clarification, by root causing this e2e test case, sometimes after the restart process, the url that one request send to may be a 404 path, meaning there are some loss requests.

Maybe by sending another request to detect the url path right after the restart process can avoid this situation.

alexwo:
> > > move load generation after the deployment has been restarted
> > 
> > 
> > Hi @shawnh2! The purpose of the test is to run load during the restart and assert that there's no loss of requests... Running it after restart means that we're not really testing for hitless shutdown anymore..
> 
> Thanks for the clarification, by root causing this e2e test case, sometimes after the restart process, the url that one request send to may be a 404 path, meaning there are some loss requests.
> 
> Maybe by sending another request to detect the url path right after the restart process can avoid this situation.

Maybe the upgrade process isn't hitless anymore, I hadn't noticed the 404 issue before, so perhaps that's a new. Maybe worth to workaround it for now ,however, we probably solve the real issue causing requests to 404.

guydc:
Hi @shawnh2 , @alexwo  - that's interesting input. Yes, I assume that if failures are 404s, then we have a problem with new  envoy proxies receiving traffic before they are programmed. 

In that case, I think we can split this activity to two parts:
- Run load and log load stats, but only assert on success of converging single-user requests. Basically, only make sure that restarted proxies are eventually functional and stop this test from being flaky.
- Investigate how to fix the underlying issue causing 404s.

WDYT?

alexwo:
> Hi @shawnh2 , @alexwo - that's interesting input. Yes, I assume that if failures are 404s, then we have a problem with new envoy proxies receiving traffic before they are programmed.
> 
> In that case, I think we can split this activity to two parts:
> 
> * Run load and log load stats, but only assert on success of converging single-user requests. Basically, only make sure that restarted proxies are eventually functional and stop this test from being flaky.
> * Investigate how to fix the underlying issue causing 404s.
> 
> WDYT?

It's also possible that existing proxies may refresh their cache when attempting to obtain a new cache snapshot version from newly started EGs. Maybe something as initialDelaySeconds suggested here can help:
https://github.com/envoyproxy/gateway/issues/2810#issuecomment-1981979019

Or potentially:
https://github.com/envoyproxy/gateway/pull/2918

zirain:
this test skiped in https://github.com/envoyproxy/gateway/pull/3306, can you remove this SkipTests?

arkodg:
hey @shawnh2 is this PR still needed ?

shawnh2:
close this for now, will raise a fix if we have a clue.

PR #3286, opened 2024-04-26
bump github.com/envoyproxy/go-control-plane

mmorel-35:
**What type of PR is this?**

bump github.com/envoyproxy/go-control-plane
Replace github.com/golang/protobuf/ptypes with google.golang.org/protobuf/types

Fixes #


guydc:
This version appears to be mirrored from [envoy main](https://github.com/envoyproxy/envoy/commit/c15b875d2f03f583d8bbaa3a43325196c69b73de) (pending 1.31). If we release EG 1.1.0 according to the current timeline (https://github.com/envoyproxy/gateway/milestone/20), It's possible that envoy 1.31 would not be released yet. 

@envoyproxy/gateway-maintainers  - should we pin envoy images and go-control-plane version to 1.30, or can we continue using latest artifacts here?

guydc:
Hi @mmorel-35  - TBH, I'm not sure what our policy should be here, we can keep this open an defer to other maintainers...

guydc:
Hi @mmorel-35 . We discussed this in the community meeting, and decided that since we use latest envoy images, we should also use latest go-control-plane in main. When we cut an RC, we'll pin both component versions and deal with any incompatible feature. So, you can resume this activity.

guydc (approved):

arkodg (approved):

PR #3287, opened 2024-04-27
ci: add trivy image vulnerabilities scan workflow

### LINKED ISSUE #3168: Add Docker image vulnerabilities scan
*Description*:
* Add Docker image vulnerabilities scan
* We can use [Trivy](https://github.com/marketplace/actions/aqua-security-trivy#scan-ci-pipeline) which is the most popular tool.
It's worth mentioning that Trivy currently doesn't detect go stdlib vulnerabilities, see related [issue](https://github.com/aquasecurity/trivy/issues/4133).
* I checked also other tools:
   * [docker-scout](https://github.com/docker/scout-action) and [Snyk](https://github.com/snyk/actions/tree/master/docker) require an account, Trivy doesn't.
   * [Grype](https://github.com/anchore/grype) outputs the same results as Trivy, but Trivy is more popular.
* Should we trigger scan also on PRs?

shahar-h:
Introduced Trivy image vulnerabilities scan workflow:
* This workflow will be triggered on merge to main and also once a week. I thought about triggering it also on PRs but I decided to avoid it in order to not block all PRs when a new vulnerability is detected.
* I didn't upload scan results to GitHub code scanning since when choosing `sarif` output format results don't appear in stdin. Since code scanning results are not public I kept the default `table` format so anyone can see it. See related [issue](https://github.com/aquasecurity/trivy-action/issues/116).
* Currently only `eg` image is scanned. Regarding `envoy-proxy` and `rate-limit` images, I think that they should be scanned as part of their ow repos CI process. We can discuss it later if needed.

Fixes #3168

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -13,19 +13,19 @@ permissions:
 jobs:
   image-scan:
     permissions:
-      contents: read # for actions/checkout to fetch code
+      contents: read  # for actions/checkout to fetch code
     name: Image Scan
     runs-on: ubuntu-22.04
     steps:
-      - name: Checkout code
-        uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f  # v4.1.3
+    - name: Checkout code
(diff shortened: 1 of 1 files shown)


guydc (approved):
LGTM, Thanks!

Issue #3183, opened 2024-04-12
enhancement: use IfNotPresent for versioned helm chart release


networkhermit:
*Description*:
> Most versioned helm chart in the wild default to `IfNotPresent` for `imagePullPolicy`, versioned helm chart uses tagged image so it rarely change, and change the default is good for air gapped environment.

[optional *Relevant Links*:]
https://github.com/envoyproxy/gateway/blob/231b3b476bb897b9679f7e303904e8a4e9cc3999/charts/gateway-helm/values.tmpl.yaml#L6

ShyunnY:
This looks good, what do you think? @zirain

arkodg:
2 options here
* change it to `IfNotPresent` only for pinned releases
* set the default to `IfNotPresent` and change it to `Always` for `v0.0.0-latest`

ptal @envoyproxy/gateway-maintainers

ShyunnY:
> 2 options here
> 
> * change it to `IfNotPresent` only for pinned releases
> * set the default to `IfNotPresent` and change it to `Always` for `v0.0.0-latest`
> 
> ptal @envoyproxy/gateway-maintainers

vote 2

guydc:
+1 for `set the default to IfNotPresent and change it to Always for v0.0.0-latest`.

zhaohuabing:
+1 for `set the default to IfNotPresent and change it to Always for v0.0.0-latest`. 

This is not related, but I'm curious, why `v0.0.0-latest` instead of `latest`?

PR #3289, opened 2024-04-28
helm: use IfNotPresent for versioned helm chart release

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3183


guydc (approved):

zhaohuabing (approved):

test/helm/default.yaml
```diff
@@ -369,7 +369,7 @@ spec:
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: cluster.local
         image: docker.io/envoyproxy/gateway-dev:latest
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
arkodg, line 372: curious how it defaults to `IfNotPresent` when the default is `Always` in the Makefile
https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/tools/make/kube.mk#L65

zirain: is this used for helm install/template?

PR #3292, opened 2024-04-28
e2e: fix connectionLimit e2e test

ShyunnY:
Some minor changes regarding error messages and `Dial` connections


ShyunnY:
PTAL @guydc

guydc:
Hi @ShyunnY. I'm not a 100% sure why the test was flaky to begin with. We can change the error handling, and maybe that'll help understand failures better when they occur next time. 

I think that a more robust approach to this test could actually be:
- Create many connections (e.g. 100).
- Assert on stats that indicate the limit is holding (both number of active and rejected connections): https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/connection_limit_filter#statistics

We have an example in the retry test for stat assertions: 
https://github.com/envoyproxy/gateway/blob/2322ef0a405f5aeb72ec9d5accb9a722d90775c3/test/e2e/tests/retry.go#L53

WDYT?

ShyunnY:
This is a good proposal. Our statistical metrics will have better test stability.

I will make some minor modifications to your proposal:
+ We set the number of connections equal to `connectionLimit + 3` (there is no need to set it too high, setting 3 is to prevent some connections from being lost)
+ We grab the metrics data of `connectionLimit` to see if there is a **restricted connection**

@guydc

guydc (approved):
LGTM, thanks!

zirain (approved):

Issue #3161, opened 2024-04-10
refactor BackendHostAndPort


zirain:
can we find a way to reuse https://github.com/envoyproxy/gateway/blob/c943042cb61f5248df6bf9f38d555d4222f65c0d/internal/gatewayapi/route.go#L1068 instead ?

_Originally posted by @arkodg in https://github.com/envoyproxy/gateway/pull/3080#discussion_r1558944670_

PR #3293, opened 2024-04-29
refactor telemetry backendRefs

zirain:
fix: #3161


internal/gatewayapi/listener.go
```diff
@@ -127,10 +129,31 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR XdsIRMap
...
+	}
+
+	xdsIR.Metrics, err = t.processMetrics(envoyProxy, resources)
+	if err != nil {
+		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+			fmt.Sprintf("Invalid metrics backendRefs: %v", err))
+		return
+	}
+
+	status.RemoveGatewayListenersNotValidCondition(gwCtx.Gateway)
```
zirain, line 156: related to https://github.com/kubernetes-sigs/gateway-api/pull/2924, seems EG should report status on `Gateway`?

internal/ir/xds.go
```diff
@@ -1427,11 +1427,12 @@ type JSONAccessLog struct {
...
-	Attributes map[string]string `json:"attributes,omitempty" yaml:"attributes,omitempty"`
-	Host       string            `json:"host" yaml:"host"`
-	Port       uint32            `json:"port" yaml:"port"`
-	Resources  map[string]string `json:"resources,omitempty" yaml:"resources,omitempty"`
+	Text         *string               `json:"text,omitempty" yaml:"text,omitempty"`
+	Attributes   map[string]string     `json:"attributes,omitempty" yaml:"attributes,omitempty"`
+	Host         string                `json:"host" yaml:"host"`
+	Port         uint32                `json:"port" yaml:"port"`
+	Resources    map[string]string     `json:"resources,omitempty" yaml:"resources,omitempty"`
+	Destinations []*DestinationSetting `json:"destinations,omitempty" yaml:"destinations"`
```
shawnh2, line 1435:
```suggestion
	Destinations []*DestinationSetting `json:"destinations,omitempty" yaml:"destinations,omitempty"`
```

internal/provider/kubernetes/controller.go
```diff
@@ -1465,6 +1466,60 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
...
+								Kind:      backendRef.BackendObjectReference.Kind,
+								Namespace: gatewayapi.NamespacePtrV1Alpha2(backendNamespace),
+								Name:      backendRef.Name,
+							}] = struct{}{}
+						}
+					}
+				}
+
+				if telemetry.Tracing != nil {
+					for _, backendRef := range telemetry.Tracing.Provider.BackendRefs {
```
shawnh2, line 1510: the logic of processing backendRefs for accesslog/metrics/tracing are quite same, is there anyway to abstract them?

internal/gatewayapi/listener.go
shawnh2: I dont understand how it is related to this issue, can you give me a hint?

internal/gatewayapi/route.go
```diff
@@ -1235,14 +1215,50 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
 			"Mixed endpointslice address type for the same backendRef is not supported")
 	}
 
-	ds = &ir.DestinationSetting{
-		Weight:      &weight,
+	ds.Weight = &weight
+	return ds, weight
+}
+
+func (t *Translator) serviceDestinationSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string,
```
shawnh2, line 1222:
```suggestion
func (t *Translator) processServiceDestinationSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string,
```

internal/gatewayapi/listener.go
zirain: think about service invalid/not exists in the first loop, and create later.

internal/provider/kubernetes/controller.go
zirain: need revist those struct later.

shawnh2 (approved):
LGTM! no questions from my side.

internal/ir/xds.go
```diff
@@ -1478,11 +1478,12 @@ type JSONAccessLog struct {
...
-	Attributes map[string]string `json:"attributes,omitempty" yaml:"attributes,omitempty"`
-	Host       string            `json:"host" yaml:"host"`
-	Port       uint32            `json:"port" yaml:"port"`
-	Resources  map[string]string `json:"resources,omitempty" yaml:"resources,omitempty"`
+	Text         *string               `json:"text,omitempty" yaml:"text,omitempty"`
+	Attributes   map[string]string     `json:"attributes,omitempty" yaml:"attributes,omitempty"`
+	Host         string                `json:"host" yaml:"host"`
+	Port         uint32                `json:"port" yaml:"port"`
+	Resources    map[string]string     `json:"resources,omitempty" yaml:"resources,omitempty"`
+	Destinations []*DestinationSetting `json:"destinations,omitempty" yaml:"destinations,omitempty"`
```
arkodg, line 1486: probably want to use `RouteDestination` here
https://github.com/envoyproxy/gateway/blob/c5ec046be14c4bb9049a7d3633900d2b8207d40f/internal/ir/xds.go#L911 to reuse clusters internally

zirain: let's do that in 1.2, in which EG remove support for host and port?

internal/ir/xds.go
```diff
@@ -1536,11 +1537,12 @@ type JSONPatchOperation struct {
...
-	Host         string                      `json:"host"`
-	Port         uint32                      `json:"port"`
-	SamplingRate float64                     `json:"samplingRate,omitempty"`
-	CustomTags   map[string]egv1a1.CustomTag `json:"customTags,omitempty"`
+	ServiceName         string                      `json:"serviceName"`
+	Host                string                      `json:"host"`
+	Port                uint32                      `json:"port"`
+	SamplingRate        float64                     `json:"samplingRate,omitempty"`
+	CustomTags          map[string]egv1a1.CustomTag `json:"customTags,omitempty"`
+	DestinationSettings []*DestinationSetting       `json:"destinationSettings,omitempty"`
```
arkodg, line 1545: the naming is inconsistent b/w Tracing and OpenTelemetryAccessLog

internal/provider/kubernetes/controller.go
```diff
@@ -226,18 +239,6 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 			gwcResource.Namespaces = append(gwcResource.Namespaces, namespace)
 		}
 
-		// Process the parametersRef of the accepted GatewayClass.
```
arkodg, line 229: curious why we moved this away from L241

internal/ir/xds.go
arkodg: imo its a small change we can handle here

arkodg: reusing clusters is already handled in xds, so we only need to populate the new struct and read from it

internal/provider/kubernetes/controller.go
```diff
@@ -1465,6 +1466,60 @@ func (r *gatewayAPIReconciler) processParamsRef(ctx context.Context, gc *gwapiv1
 				validationErr = fmt.Errorf("invalid envoyproxy: %w", err)
 				continue
 			}
+
```
arkodg, line 1469: we probably also need to enhance https://github.com/envoyproxy/gateway/blob/e5f86303d83d732f8c3890a6868c62e29a3e3029/internal/provider/kubernetes/predicates.go#L217 to make sure we trigger a reconcile when the `Service` changes

zirain: yeah, but want to do that in an seperated PR.

internal/ir/xds.go
zirain: will take a look later

internal/provider/kubernetes/controller.go
zirain: need update `allAssociatedBackendRefs` first.

internal/xds/translator/testdata/out/xds-ir/tracing.clusters.yaml
```diff
@@ -25,25 +25,20 @@
...
       locality:
-        region: tracing|otel-collector.monitoring.svc.cluster.local|4317/backend/0
-  name: tracing|otel-collector.monitoring.svc.cluster.local|4317
+        region: tracing-0/backend/0
+  name: tracing-0
   outlierDetection: {}
   perConnectionBufferLimitBytes: 32768
   respectDnsTtl: true
   type: STRICT_DNS
-  typedExtensionProtocolOptions:
```
arkodg, line 45: hmm why did this get removed ? guessing this is a grpc endpoint, so we need http2

internal/xds/translator/testdata/out/xds-ir/tracing-endpoint-stats.listeners.yaml
```diff
@@ -51,8 +51,7 @@
               '@type': type.googleapis.com/envoy.config.trace.v3.OpenTelemetryConfig
               grpcService:
                 envoyGrpc:
-                  authority: otel-collector.monitoring.svc.cluster.local
```
arkodg, line 54: why is authority removed ? is this intentional ?

zirain: it's not, mistake by hand write input.

internal/xds/translator/testdata/out/xds-ir/tracing-endpoint-stats.clusters.yaml
```diff
@@ -27,27 +27,22 @@
...
-  name: tracing|otel-collector.monitoring.svc.cluster.local|4317
+        region: tracing-0/backend/0
+  name: tracing-0
   outlierDetection: {}
   perConnectionBufferLimitBytes: 32768
   respectDnsTtl: true
   trackClusterStats:
     perEndpointStats: true
   type: STRICT_DNS
-  typedExtensionProtocolOptions:
```
arkodg, line 49: isnt this needed, since this is a grpc endpoint

internal/gatewayapi/listener.go
```diff
@@ -227,16 +250,28 @@ func processAccessLog(envoyproxy *egv1a1.EnvoyProxy) *ir.AccessLog {
...
 
-				if sink.OpenTelemetry.Host != nil {
-					al.Host = *sink.OpenTelemetry.Host
+				// TODO: how to get authority from the backendRefs?
+				ds, err := t.processBackendRefs(sink.OpenTelemetry.BackendRefs, envoyproxy.Namespace, resources)
+				if err != nil {
+					return nil, err
+				}
+				al.Destination = ir.RouteDestination{
+					Name:     fmt.Sprintf("accesslog-%d", idx), // TODO: rename this, so that we can share backend with tracing?
```
arkodg, line 262: +1 to this prefix, once we add cluster level features like `circuitBreaking` which will live on `backendRef` level , we won't be able to reuse `tracing-` here

internal/gatewayapi/listener.go
```diff
@@ -251,25 +286,35 @@ func processAccessLog(envoyproxy *egv1a1.EnvoyProxy) *ir.AccessLog {
...
+		return nil, nil
 	}
 	tracing := envoyproxy.Spec.Telemetry.Tracing
 
-	// TODO: remove support for Host/Port in v1.2
-	var host string
-	var port uint32
-	if tracing.Provider.Host != nil {
-		host, port = *tracing.Provider.Host, uint32(tracing.Provider.Port)
+	// TODO: how to get authority from the backendRefs?
```
arkodg, line 300: name.namespace ?

internal/gatewayapi/listener.go
arkodg: im not sure I understand this either, if the service gets crated later, we'll reconcile again, compute status again and update the old one, we cannot add `status.RemoveGatewayListenersNotValidCondition(gwCtx.Gateway)` here, because this value may have set by another logic in the translator

internal/gatewayapi/listener.go
zirain: but we should reuse with tracing provider, normally there's only one collector per cluster.

internal/gatewayapi/listener.go
zirain: what about a service outside of cluster.

internal/xds/translator/testdata/out/xds-ir/tracing-endpoint-stats.clusters.yaml
zirain: fixed

internal/gatewayapi/listener.go
arkodg: it may not be possible to reuse if we decide to put features inside `backendRef` field, if feature live in top level policy, it may be possible

internal/gatewayapi/listener.go
arkodg: yah that we'd need to add an explicit field in the future

internal/gatewayapi/listener.go
zirain: you're both right, I misunderstand somehting.

arkodg (approved):
LGTM, thanks !

Issue #2947, opened 2024-03-17
Extract username from basic auth and forward it to backends


zhaohuabing:
I would ask for an update of the basic auth in the `SecurityPolicy` to extract the username and add it to the headers sent to the backends.

The use case is the following, after user auth success I need to add the username in the request header that will be used by the application after. 

This depends on the Envoy upstream PR: https://github.com/envoyproxy/envoy/pull/32943

akorosi:
+1

s0uky:
+1 
Is there some progress in this issue, pls? 
Thank you

s0uky:
not stale

surenraju:
I noticed that `BasicAuthPerRoute` does not support the `forward_username_header` field—only `BasicAuth` does.

Reference: [basic_auth.proto#L56](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/basic_auth/v3/basic_auth.proto#L56)

```
envoy-proxy  | [2025-02-18 18:14:36.611][1][info][main] [source/server/server.cc:1060] exiting
envoy-proxy  | Unable to unpack as envoy.extensions.filters.http.basic_auth.v3.BasicAuthPerRoute: [type.googleapis.com/envoy.extensions.filters.http.basic_auth.v3.BasicAuth] {
envoy-proxy  |   users {
envoy-proxy  |     inline_string: "suren:{SHA}ABw9tVKwslVgEf5YCxffU9+LB+CM= "
envoy-proxy  |   }
envoy-proxy  |   forward_username_header: "x-custom-user"
envoy-proxy  | }
```

From my understanding, when a `SecurityPolicy` is attached to an `HTTPRoute`, it creates an Envoy route configuration using `BasicAuthPerRoute`. For example, the generated route configuration looks like this:

https://gateway.envoyproxy.io/docs/tasks/security/basic-auth/

```
"routes": [
  {
    "match": {
      "prefix": "/"
    },
    "route": {
      "cluster": "httproute/default/backend/rule/0",
      "upgrade_configs": [
        {
          "upgrade_type": "websocket"
        }
      ]
    },
    "metadata": {
      "filter_metadata": {
        "envoy-gateway": {
          "resources": [
            {
              "kind": "HTTPRoute",
              "namespace": "default",
              "name": "backend"
            }
          ]
        }
      }
    },
    "typed_per_filter_config": {
      "envoy.filters.http.basic_auth": {
[...]

PR #3312, opened 2024-04-30
feat: forward basic auth user

zhaohuabing:
This PR extracts username from the HTTP authorization header and add it to a custom header, providing a user identity that can be directly used in [the Authorization Policy](https://github.com/envoyproxy/gateway/pull/2652) and by other filters or backends.

```
      basicAuth:
        users:
          name: "users-secret1"
        forwardUsernameHeader: x-basic-auth-user
```

Fix: https://github.com/envoyproxy/gateway/issues/2947
Related: https://github.com/envoyproxy/gateway/issues/2250 https://github.com/envoyproxy/gateway/pull/2652


arkodg:
thoughts on `userNameToHeader` similar to the existing `claimToHeader` field in jwt ?

api/v1alpha1/basic_auth_types.go
```diff
@@ -24,4 +27,10 @@ type BasicAuth struct {
 	//
 	// Note: The secret must be in the same namespace as the SecurityPolicy.
 	Users gwapiv1b1.SecretObjectReference `json:"users"`
+
+	// The name of the HTTP header that will be used to forward the username to the upstream server.
+	// Please note that the header name is a global setting for all the routes in a Gateway.
```
zhaohuabing, line 32: `UserNameToHeader` can also be per route, but will need some upstream work.

arkodg: imo we should wait until per route support is added, else a per route setting will enable this at the global level

zhaohuabing: OK, let's wait for the upstream.

Issue #3304, opened 2024-04-29
support oci image as a code source for wasm


zhaohuabing:
Majority of the WASM Modules now live in OCI Registries, so EG should be able to set an OCI image as the WASM code source through its API. There are two ways to accomplish this:

* Envoy supports oci_uri as a remote code source for its wasm filter, which is not supported yet, tracked in upstream issue https://github.com/envoyproxy/envoy/issues/33212
* EG fetches wasm modules from OCI registries and serves the envoy proxies through a control plane side HTTP server.

PR #3313, opened 2024-05-01
design: add wasm extension supports OCI image code source

zhaohuabing:
Related: https://github.com/envoyproxy/gateway/issues/3304


zirain:
I'm not sure which is a better idea:
1. extend EG controller support this
2. introduce an new component.

guydc:
Is there potential for a new component to be reused by other envoy control planes (e.g. istio)? If so, it could be beneficial from long-term maintenance perspective... 

Can we implement this based on some existing OCI registry cache?

zhaohuabing:
> Is there potential for a new component to be reused by other envoy control planes (e.g. istio)? If so, it could be beneficial from long-term maintenance perspective...

To simplify the process, I'd like to create this component as part of EG rather than a standalone project. This will help us avoid unnecessary dependencies/management(install, upgrade,etc.) overhead.


> Can we implement this based on some existing OCI registry cache?

We can use https://github.com/google/go-containerregistry

site/content/en/contributions/design/wasm-extension.md
```diff
@@ -0,0 +1,54 @@
...
+
+![](../wasm-extension.png)
+
+### Control Plane Wasm File Cache
+
+Envoy lacks native OCI image support, therefore, EG needs to download Wasm modules from OIC registries, 
+cache them locally in the file system, and serve them to Envoy over HTTP.
+
+**HTTP Code Source:** For HTTP code source, we have two options: serve Wasm modules directly from their 
+original HTTP URLs, or cache them in EG (as with OIC images). Caching both the HTTP Wasm modules and OCI
```
arkodg, line 23:
```suggestion
original HTTP URLs, or cache them in EG (as with OCI images). Caching both the HTTP Wasm modules and OCI 
```

site/content/en/contributions/design/wasm-extension.md
```diff
@@ -0,0 +1,54 @@
...
+images inside EG can make UI consistent, for example, sha256sum can be calculated on the EG side and made 
+optional in the API. EG can also periodically pull Wasm modules and update the local file cache.
+
+**Memory Optimization:** Since we cache Wasm modules in the file system, we can optimize the memory usage 
+of the cache and the HTTP server to avoid introducing too much memory consumption by this feature. 
+For example, when receiving a Wasm pulling request form the Envoy, we can open the Wasm file and write 
+the file content directly to response, and then close the file. There won’t be significant caching involved. 
+However, we need to balance the memory usage and the efficiency, which could be addressed in implementation.
+
+**Caching Mechanism:** Cached files will be evicted based on LRU（Last recently used）algorithm.
```
arkodg, line 33: can you add more clarity here ? is there an existing registry cache we can use ?

site/content/en/contributions/design/wasm-extension.md
```diff
@@ -0,0 +1,54 @@
...
+### Control Plane Wasm File Cache
+
+Envoy lacks native OCI image support, therefore, EG needs to download Wasm modules from OIC registries, 
+cache them locally in the file system, and serve them to Envoy over HTTP.
+
+**HTTP Code Source:** For HTTP code source, we have two options: serve Wasm modules directly from their 
+original HTTP URLs, or cache them in EG (as with OCI images). Caching both the HTTP Wasm modules and OCI 
+images inside EG can make UI consistent, for example, sha256sum can be calculated on the EG side and made 
+optional in the API. EG can also periodically pull Wasm modules and update the local file cache.
+
```
arkodg, line 27: how would resource isolation/authorization work ? i.e. how will EG make sure its not serving a private OCI image with private creds to another proxy or route filter ?

site/content/en/contributions/design/wasm-extension.md
zhaohuabing: Istio wasm cache fetches wasm module from OCI registry and HTTP url, and saves them to local files. We can reuse most of them and add an HTTP server in front.

https://github.com/istio/istio/tree/master/pkg/wasm

site/content/en/contributions/design/wasm-extension.md
arkodg: im referring to authz here
1. `user1` added a image cred as a `Secret` in `ns1`, and linked this in the `EEP` config for a specific `oci1` private URL targeting `httproute1`
2.  how will EG ensure `user2` created EEP in `ns2` cannot access the same `oci1` from the EG cache

zhaohuabing: We make sure the credential within that EEP wasm setting is correct. In this case:

If `EEP1` in `ns1` created by `user1` and `EEP2` in `ns2` created by `user2` refer to the same OCI image, `user1` need to create `secret1` in `ns1` and `user2` need to `secret2` in `ns2`, both with the correct credential to access to the registry.

site/content/en/contributions/design/wasm-extension.png
shawnh2, line 1: this should live under https://github.com/envoyproxy/gateway/tree/main/site/static/img

zirain (approved):

shawnh2 (approved):

PR #3320, opened 2024-05-02
fix helm IfNotPresent

zirain:
1. `ImagePullPolicy` be one of params of `make helm-generate`
2. make sure set `ImagePullPolicy=Always/IfNotPresent` when run `make helm-package` and `generate-manifests`

cc @arkodg


guydc (approved):

arkodg (approved):
thanks for the cleanup !

PR #3332, opened 2024-05-06
fix: add proxy protocol always as first listenerFilter

### LINKED ISSUE #2968: bug: race condition? when using mergeGateway and multiple gateways
*Description*:

I have following installation
```
% kubectl get gateway -A             
NAMESPACE              NAME       CLASS         ADDRESS          PROGRAMMED   AGE
echoserver             foobar     eg-internal   10.222.156.49    True         34m
envoy-gateway-system   internal   eg-internal   10.222.156.49    True         50m
```

full yaml spec: https://gist.github.com/zetaab/8caa34f5072d5a8efc5c2425c331c561

and httproutes https://gist.github.com/zetaab/149545f3e0ae17c0b925bafd3512d1eb

When I am adding httproutes and envoy proxies are restarting, it will randomly all services unavailable. When envoy pods are starting I can see following in logs

```
[2024-03-18 06:46:10.571][1][info][upstream] [source/common/listener_manager/lds_api.cc:99] lds: add/update listener 'envoy-gateway-system/internal/https'
[2024-03-18 06:46:10.573][1][info][upstream] [source/common/listener_manager/lds_api.cc:99] lds: add/update listener 'envoy-gateway-system/internal/http'
```

This means that all services are running (also services that are coming from other gateways than https). However, when the log says

```
[2024-03-18 07:17:24.768][1][info][upstream] [source/common/listener_manager/lds_api.cc:99] lds: add/update listener 'echoserver/foobar/https-foo'
[2024-03-18 07:17:24.769][1][info][upstream] [source/common/listener_manager/lds_api.cc:99] lds: add/update listener 'envoy-gateway-system/internal/http'
```

Nothing will work. 

```
 % curl https://foo.bar -v -k     
*   Trying 10.222.156.49:443...
* Connected to foo.bar (10.222.156.49) port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
* LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to foo.bar:443 
* Closing connection
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to foo.bar:443
```

```
% curl https://eg-int.company.com -v
*   Trying 10.222.156.49:443...
* Connected to eg-int.company.com (10.222.156.49) port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to eg-int.company.com:443 
* Closing connection
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to eg-int.company.com:443 
```

Why listeners are loaded in different order sometimes?

*Repro steps*:
1) use mergeGateways
2) create two different gateways
3) add and delete httproutes
4) soonish you should see the situation that listener will fail for some reason (there is no error anywhere but the port is not just listening)

*Environment*:
eg 1.0.0

*Logs*:

took listener configurations using egctl `egctl config envoy-proxy listener -n envoy-gateway-system -l gateway.envoyproxy.io/owning-gatewayclass=eg-internal`

not working:
https://gist.github.com/zetaab/2e0f2f00174d4b6189290e095ebe5cf5

working:
https://gist.github.com/zetaab/08d2f7b3bfbd04a28be14bd990552214

like can be seen: sometimes its missing 400+ rows of configurations. When this happens, I need to delete all other gateways than primary one (located in envoy-gateway-system) and then add other gateways back. Then everything starts to work again.

zetaab:
**What type of PR is this?**

fix: add proxy protocol always as first listenerFilter

As you can see I did issue about 3 months ago and since that I have had random issues that envoy will not reply anything back. Now I decided that I will go through the egctl dumps and I found out that the order in tls and proxy protocol listenerFilters are incorrect somehow.

not working
```
                  "listenerFilters": [
                    {
                      "name": "envoy.filters.listener.tls_inspector",
                      "typedConfig": {
                        "@type": "type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector"
                      }
                    },
                    {
                      "name": "envoy.filters.listener.proxy_protocol",
                      "typedConfig": {
                        "@type": "type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol"
                      }
                    }
                  ],
```

working
```
                  "listenerFilters": [
                    {
                      "name": "envoy.filters.listener.proxy_protocol",
                      "typedConfig": {
                        "@type": "type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol"
                      }
                    },
                    {
                      "name": "envoy.filters.listener.tls_inspector",
                      "typedConfig": {
                        "@type": "type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector"
                      }
                    }
                  ],
```

Fixes #2968

### DIFF
--- a/internal/xds/translator/proxy_protocol.go
+++ b/internal/xds/translator/proxy_protocol.go
@@ -32,7 +32,8 @@ func patchProxyProtocolFilter(xdsListener *listenerv3.Listener, irListener *ir.H
 	proxyProtocolFilter := buildProxyProtocolFilter()
 
 	if proxyProtocolFilter != nil {
-		xdsListener.ListenerFilters = append(xdsListener.ListenerFilters, proxyProtocolFilter)
+		// Add the Proxy Protocol filter as first to listeners.
+		xdsListener.ListenerFilters = append([]*listenerv3.ListenerFilter{proxyProtocolFilter}, xdsListener.ListenerFilters...)
 	}
 }
 
--- a/internal/xds/translator/proxy_protocol_test.go
+++ b/internal/xds/translator/proxy_protocol_test.go
@@ -1,0 +1,58 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package translator
+
+import (
+	"testing"
+
+	listenerv3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+	"github.com/stretchr/testify/require"
+
+	"github.com/envoyproxy/gateway/internal/ir"
+)
+
+func TestPatchProxyProtocolFilter(t *testing.T) {
+	type testCase struct {
+		name     string
+		listener *listenerv3.Listener
+	}
+
+	irListener := &ir.HTTPListener{
+		EnableProxyProtocol: true,
+	}
+
+	testCases := []testCase{
+		{
+			name: "listener with proxy proto available already",
+			listener: &listenerv3.Listener{
+				ListenerFilters: []*listenerv3.ListenerFilter{
+					{
+						Name: wellknown.ProxyProtocol,
+					},
+				},
+			},
+		},
+		{
+			name: "listener with tls, append proxy proto",
+			listener: &listenerv3.Listener{
+				ListenerFilters: []*listenerv3.ListenerFilter{
+					{
+						Name: wellknown.TLSInspector,
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			patchProxyProtocolFilter(tc.listener, irListener)
+			// proxy proto filter should be added always as first
+			require.Equal(t, wellknown.ProxyProtocol, tc.listener.ListenerFilters[0].Name)
+		})
+	}
+}


zetaab:
cc @arkodg perhaps worth of cherry-picking this to 1.0.x series as well. It is currently broken if people do use proxy protocol and tls.

guydc (commented):
Hi @zetaab. Overall makes sense to me. Can you maybe create a unit test to protect this from breaking in the future?

arkodg:
hey @zetaab thanks for fixing this !
added a label so we'll be sure this include this in `v1.0.2` as well
as @guydc suggested, can we add a xds translator test for this ?

zetaab:
@arkodg @guydc I could not find way to reproduce that with normal yaml tests. Always proxy protocol ended as first without the change. So I decided to do just normal unit test for that

arkodg:
looks like the existing test was already checking this and was computing this correctly
https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/testdata/out/xds-ir/listener-proxy-protocol.listeners.yaml
https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/testdata/in/xds-ir/listener-proxy-protocol.yaml

how is it that this test case is passing ?

zetaab:
I have no idea how to reproduce that really. I had such configuration in envoy gateway, sometimes it was working sometimes not. I just restarted EG + envoy pods n times and it happened sometimes.

arkodg (approved):

guydc (approved):

Issue #3226, opened 2024-04-20
Don't switch container port


zvlb:
I continue to work on the possibility of using Envoy Gateway on bare metal installations of Kubernetes. In my case, services with the LoadBalancer type are not available in the cluster, so I need to launch containers on the required ports (which are specified in the CR Gateway) without executing logic in the [servicePortToContainerPort](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/helpers.go#L231) function.

At the moment, I see two solutions:

1. Add a flag to Envoy that disables the need to execute `servicePortToContainerPort`. I'm not very fond of this option because if I need multiple Gateways, not all of which should have `CAP_NET_BIND_SERVICE`, there will be problems.
2. Add a field to EnvoyProxy that will control this behavior.

Do you have any other suggestions? I can implement the agreed functionality.

arkodg:
relates to https://github.com/envoyproxy/gateway/pull/2405
@Xunzhuo is a knob to opt out of this translation the only option ?

zvlb:
If someone can create name for this options in field `EnvoyProxyKubernetesProvider` I can start to implement it)

arkodg:
@zvlb would be great if you can take #2405 forward (as a new PR)
some open questions are
* right home - should this field live inside [EnvoyGateway](https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/api/v1alpha1/envoygateway_types.go#L194) config or [EnvoyProxy](https://github.com/envoyproxy/gateway/blob/9a626f262317b90946219d7f85069221cce2c4b6/api/v1alpha1/envoyproxy_types.go#L241)

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/3333

Issue #2310, opened 2023-12-15
Cannot listen host 80/443 when set `hostNetwork=true`


zzjin:
*Description*:
When use eg at bare-metal env but without any other lb support, we can set `hostNetwork=true` to listen node machine's port.
But at previous commit, we cannot listen port below 1024.

*Repro steps*:
Setup eg with `mergeGateways=true` and `hostNetwork=true`, and apply demo GatewayClass&Gateway, 

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
    namespace: envoy-gateway-system
---

apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
  namespace: envoy-gateway-system
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        pod:
          hostNetwork: true
      envoyService:
        type: NodePort
  mergeGateways: true
---

apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
  namespace: sealos-system
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
    - name: https
      protocol: HTTPS
      port: 443
      allowedRoutes:
        namespaces:
          from: All
      tls:
        mode: Terminate
        certificateRefs:
          - kind: Secret
            group: ""
            name: wildcard-cert
```

*Environment*:
envoy gateway version: main(9aebfdc)

*logs*

Deployment gateway listen on host port 10080&10443

May be fixed by: #2309

arkodg:
@zzjin the deployment should be reached only via the service, and not directly. The service is still exposed on `80`. What IP are you using ? NodeIP or Svc Cluster IP ?

zzjin:
> @zzjin the deployment should be reached only via the service, and not directly. The service is still exposed on `80`. What IP are you using ? NodeIP or Svc Cluster IP ?

NodeIP.

Just want it works like `ingress-nginx`, which can listen host's port when there have no working load-balancers environment like k3s/kind.

ilnurkhal:
>we can set hostNetwork=true

[...]

PR #3333, opened 2024-05-06
Add feature for "don't switching container port"

zvlb:
issue - https://github.com/envoyproxy/gateway/issues/3226, https://github.com/envoyproxy/gateway/issues/2310


api/v1alpha1/envoyproxy_types.go
```diff
@@ -264,6 +264,13 @@ type EnvoyProxyKubernetesProvider struct {
 	//
 	// +optional
 	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
+
+	// DisablePortShift disables the port shifting feature in the Envoy Proxy.
+	// When set to false (default value), if the service port is a privileged port (1-1023), add a constant to the value converting it into an ephemeral port.
+	// This allows the container to bind to the port without needing a CAP_NET_BIND_SERVICE capability.
+	//
+	// +optional
+	DisablePortShift *bool `json:"disablePortShift,omitempty"`
```
arkodg, line 273: any other alternate name suggestions ?`DisablePortShift` is too vague imo and doesnt account for the fact that this issue is tied to `ContainerPort` 
cc @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

zvlb: To come up with variable names is harder than writing code.

Something like `DisableShiftContainerPort`?

cnvergence: Could it be an other way around, like BindPort or StaticPort bool value?

zvlb: `BindPort` and `StaticPort` looks like fields for port value. Not for bool

arkodg: I prefer `UseListenerPortAsContainerPort` from https://github.com/envoyproxy/gateway/pull/2405/files
@envoyproxy/gateway-maintainers / @envoyproxy/gateway-reviewers wdyt

zhaohuabing: +1 for `UseListenerPortAsContainerPort` .

Reaons: I can tell what `UseListenerPortAsContainerPort` does by its name, but can't infer what `DisablePortShift`  without looking into the comment.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -264,6 +264,13 @@ type EnvoyProxyKubernetesProvider struct {
 	//
 	// +optional
 	EnvoyHpa *KubernetesHorizontalPodAutoscalerSpec `json:"envoyHpa,omitempty"`
+
+	// DisablePortShift disables the port shifting feature in the Envoy Proxy.
+	// When set to false (default value), if the service port is a privileged port (1-1023), add a constant to the value converting it into an ephemeral port.
+	// This allows the container to bind to the port without needing a CAP_NET_BIND_SERVICE capability.
```
zhaohuabing, line 270: Does this require user involvement, or does EG automatically add the `CAP_NET_BIND_SERVICE` capability to the Envoy container?

zvlb: I'm not sure about "automatically".
In my case I set this CAP in envoyProxy, like:
```
kind: EnvoyProxy
metadata:
  name: default
  namespace: envoy-gateway
spec:
  provider:
    kubernetes:
      disablePortShift: true.  # tmp name
      envoyDaemonSet:
        pod:
          nodeSelector:
            node-role.kubernetes.io/gateway: "true"
        container:
          image: envoyproxy/envoy:v1.29.2
          securityContext:
            allowPrivilegeEscalation: true
            capabilities:
              add:
              - NET_BIND_SERVICE
              drop:
              - ALL
            seccompProfile:
              type: RuntimeDefault
        patch:
          type: StrategicMerge
          value:
            spec:
              template:
                spec:
                  hostNetwork: true
    type: Kubernetes
```

And need to use NOT `distroless` images, bc `distroless` can't bind on 1-1024 ports

api/v1alpha1/envoyproxy_types.go
zvlb: No problem. I can use this variable. However, I see a small logical issue. If I plan to use a port higher than 1024 in the Listener (for example, 3000), there will be no 'port offsets'. That means the `UseListenerPortAsContainerPort` variable will be set to `false` (the default value), and at the same time, the listener port and the container port will be equal.

zvlb:
I rebase my branch from main and change field name from `DisablePortShift` to `UseListenerPortAsContainerPort`.

But I can't run e2e test localy. If something broken - I will test all and fix in monday. (I'm not sure about [route.go](https://github.com/zvlb/gateway/blob/disable-switch-ports/internal/gatewayapi/route.go)... Many changes. Now this package don't collect container ports)

zvlb:
Just rebase

zvlb:
Test errors: `Tokenless has reached GitHub rate limit`

AliceProxy (approved):
branch needs rebase again but lgtm

arkodg (approved):
LGTM thanks !

Issue #3265, opened 2024-04-24
Update Gateway API to v1.1.0-rc1


arkodg:
*Description*:
>Describe the issue.

Gateway API just released a new version, lets bring in the rc version

More in https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.1.0-rc1

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
Gateway API 1.1 timelines
Rc1: a few hours
v1.1: 2 weeks

arkodg:
related https://github.com/envoyproxy/gateway/issues/2624#issuecomment-2047564216

PR #3336, opened 2024-05-07
bump gatewayapi to 1.1.0-rc2

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3265

Key features:
- Bump GRPCRoute to `v1`
- Rename `gwapiv1a2.PolicyTargetReferenceWithSectionName` to `gwapiv1a2.LocalPolicyTargetReferenceWithSectionName`


AliceProxy (approved):

zhaohuabing (approved):

internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
```diff
@@ -322,6 +322,7 @@ grpcRoutes:
       - method:
           method: DoThing
           service: com.example.Things
+      sessionPersistence: null
```
arkodg, line 325: should be fixed in v1.1
https://github.com/kubernetes-sigs/gateway-api/pull/3059

internal/envoygateway/scheme.go
```diff
@@ -6,11 +6,11 @@
 package envoygateway
 
 import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
-	gwapiv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
```
arkodg, line 12: I dont think we can remove these yet
we need to continue to support GRPCRoute in v1beta1

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -199,16 +198,11 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 }
 
 func resolveBTPolicyGatewayTargetRef(policy *egv1a1.BackendTrafficPolicy, gateways map[types.NamespacedName]*policyGatewayTargetContext) (*GatewayContext, *status.PolicyResolveError) {
-	targetNs := policy.Spec.TargetRef.Namespace
```
arkodg, line 202: can we rm this ? what happens when the target Ns is different f

arkodg (approved):
LGTM thanks ! 
this bump does not look easy 

once merged, we should probably alert our users on Slack that old BackendTLSPolicy resources will no longer work, and will need to be updated

internal/gatewayapi/backendtrafficpolicy.go
arkodg: nvm, this field no longer exists

PR #3342, opened 2024-05-08
feat(api): Add support for the header hash policy

### LINKED ISSUE #3341: Add support for the header hash policy in the consistent hash load balancer
*Description*:

[Slack discussion here.](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1715069442245479)

While Envoy itself supports a [range of hash policies](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy) for consistent hash load balancing, the Envoy Gateway can only be configured with the [ConnectionProperties](https://github.com/envoyproxy/envoy/blob/8937b3d0bd28ed9f60e44744f4ad24149dee5d85/api/envoy/config/route/v3/route_components.proto#L896-L902) policy via the [SourceIP](https://gateway.envoyproxy.io/v1.0.1/api/extension_types/#consistenthashtype) consistent hash type. This is insufficient for situations where session affinity must be based on a sharding key that is in a request header.

For example, a request might include an resource identifier in a request header. When the request is processed, that resource will be cached on the upstream, so subsequent requests should be routed to the same instance. This routing must be done based on the value of that header, rather than the source IP address, which can change for a variety of reasons[^1].

Specifically, we want this feature to implement session affinity for gRPC messages based on their [protocol headers.](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)

[^1]: Whenever the downstream scales up/down, or for an public deployment, or when the client device roams and undergoes a network handover (which causing the connection to be reestablished), NAT rebinding, etc

*Any extra documentation required to understand the issue.*

* [ConnectionProperties proto](https://github.com/envoyproxy/envoy/blob/8937b3d0bd28ed9f60e44744f4ad24149dee5d85/api/envoy/config/route/v3/route_components.proto#L896-L902)
* [ConsistentHashType documentation](https://gateway.envoyproxy.io/v1.0.1/api/extension_types/#consistenthashtype)

harrisonturton:
**What type of PR is this?**

A user-facing API change.

See #3341 for justification, discussed in [this Slack thread.](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1715069442245479)

This PR adds support for the [Header](https://github.com/envoyproxy/envoy/blob/8bc2e917519936ab4cb1e9fa52fd5456a4fce4f7/api/envoy/config/route/v3/route_components.proto#L834-L846) hash policy in the Envoy Gateway API, which will allow users the configure the Envoy Gateway to use session affinity based on the value of a header. I've only included API changes here – I'll follow up with the implementation in a separate PR.

When implemented, this will fix #3341.


test/cel-validation/backendtrafficpolicy_test.go
```diff
@@ -174,6 +174,53 @@ func TestBackendTrafficPolicyTarget(t *testing.T) {
 			},
 			wantErrors: []string{},
 		},
+		{
+			desc: "consistentHash header field not nil when consistentHashType is header",
```
harrisonturton, line 178: I think this is where the validation tests are supposed to go, but when I run `make test` this directory is not listed in the output. I only see:

```
> make test
...
?       github.com/envoyproxy/gateway/internal/xds/utils        [no test files]
?       github.com/envoyproxy/gateway/test/e2e/tests    [no test files]
?       github.com/envoyproxy/gateway/proto/extension   [no test files]
...
```

So I'm not convinced they're passing yet. Is there a CI pipeline or different command I should be running?

api/v1alpha1/loadbalancer_types.go
```diff
@@ -52,9 +52,27 @@ const (
 )
 
 // ConsistentHash defines the configuration related to the consistent hash
-// load balancer policy
+// load balancer policy.
+// +union
+//
+// +kubebuilder:validation:XValidation:rule="self.type == 'Header' ? has(self.header) : !has(self.header)",message="If consistent hash type is header, the header field must to be set."
 type ConsistentHash struct {
```
harrisonturton, line 59: I've turned this into a union to configure the different hash policies.

I envision future policies will added as additional union variants:

```go
type ConsistentHash struct {
  Type ConsistentHashType `json:"type"`
  Header *Header          `json:"header"`
  Cookie *Cookie          `json:"cookie"`
  // And so on
}
```

api/v1alpha1/loadbalancer_types.go
```diff
@@ -52,9 +52,27 @@ const (
...
+
+	// Header configures the header hash policy when the consistent hash type is set to Header.
+	// +optional
+	Header *Header `json:"header,omitempty"`
+}
+
+// Header defines the header hashing configuration for consistent hash based
+// load balancing.
+type Header struct {
+	HeaderName string `json:"header_name"`
```
arkodg, line 75: following convention used in other place, just `Name` makes more sense here
* https://github.com/envoyproxy/gateway/blob/ed4222b78124de25d0c7094080567c45047d33c3/api/v1alpha1/ratelimit_types.go#L144
* https://github.com/kubernetes-sigs/gateway-api/blob/a29ddd4ed9395c0c192c19be853deb5e0a3a854d/apis/v1/httproute_types.go#L469

test/cel-validation/backendtrafficpolicy_test.go
arkodg: `make help` can come in handy here to find the right target
* `make tube-test` 
* `make go.test.cel`

should run the tests with the `cel` tag

api/v1alpha1/loadbalancer_types.go
```diff
@@ -53,17 +53,32 @@ const (
 
 // ConsistentHash defines the configuration related to the consistent hash
 // load balancer policy
+// +kubebuilder:validation:XValidation:rule="self.type == 'Header' ? has(self.header) : !has(self.header)",message="If consistent hash type is header, the header field must to be set."
 type ConsistentHash struct {
-	Type ConsistentHashType `json:"type"`
+	Type   ConsistentHashType `json:"type"`
+	Header Header             `json:"header,omitempty"`
```
arkodg: Needs to be a ptr type since it's an optional field

harrisonturton: This should be fixed on the most recent push

charts/gateway-helm/crds/generated/gateway.envoyproxy.io_backendtrafficpolicies.yaml
```diff
@@ -412,15 +412,29 @@ spec:
                       ConsistentHash defines the configuration when the load balancer type is
                       set to ConsistentHash
                     properties:
+                      header:
+                        description: Header configures the header hash policy when
+                          the consistent hash type is set to Header.
+                        properties:
+                          header_name:
```
harrisonturton, line 419: I can't get this to regenerate –`header_name` should be `name` after updating `loadbalancer_types.go`. I looked through `make help` and ran a variety of `make helm-package` and `make generate-artifacts` but no dice. What's happening here?

harrisonturton: Never mind, figured it out. `make manifests`

api/v1alpha1/loadbalancer_types.go
harrisonturton: Renamed!

arkodg (dismissed):
LGTM thanks !

guydc:
Overall LGTM. @arkodg  - should we start using `+notImplementedHide` for a new APIs?

arkodg:
Sure good idea ! we can then request contributors to remove it in the implementation PR.
Hey @harrisonturton can you add that tag to the API field (Header)

harrisonturton:
@arkodg added `+notImplementedHide` to `Header` and also removed mention of the new `Header` variant in the `ConsistentHashType` docstring. If the header struct was hidden from the documentation, but the `ConsistentHashType` docs described a "Header" variant, it would be confusing for users 😄  I'll add it back in the implementation PR

site/content/en/latest/api/extension_types.md
```diff
@@ -577,6 +577,7 @@ _Appears in:_
 | Value | Description |
 | ----- | ----------- |
 | `SourceIP` | SourceIPConsistentHashType hashes based on the source IP address.<br /> | 
+| `Header` | HeaderConsistentHashType hashes based on a request header.<br /> |
```
harrisonturton, line 580: FYI this line still exists in the generated documentation. Seems like it might be generated from the union discriminator validation comments? The `+notImplementedHide` did not hide it, even when added on `HeaderConsistentHashType`.

arkodg: cc @zirain can we disable this as well ?
@harrisonturton thanks for highlighting this, lets not block your PR on this

arkodg (dismissed):

guydc (approved):

arkodg:
@harrisonturton can you run `make generate` and commit those changes ?

harrisonturton:
Done – I checked the pipeline and was able to resolve it with `make manifests`. I tried `make generate` but it didn't update the helm charts 🤔

arkodg (approved):

site/content/en/latest/api/extension_types.md
zirain: this's expected, you cannot find  `header` under `ConsistentHash`.

PR #3345, opened 2024-05-08
feat(api): add idleTimeout to ClientTrafficPolicy for TCP listener

### LINKED ISSUE #3343: feat: add support to configuring downstream idle timeout for TCP listener
*Description*:
> Add `idleTimeout` to clientTrafficPolicy for TCP listener, for the same reason as #2861.

aoledk:

Add idleTimeout to ClientTrafficPolicy for TCP listener.

Fixes #3343


guydc (commented):
Overall looks good me, just one comment on the doc

api/v1alpha1/timeout_types.go
```diff
@@ -43,12 +43,26 @@ type HTTPTimeout struct {
...
+	TCP *TCPClientTimeout `json:"tcp,omitempty"`
+
 	// Timeout settings for HTTP.
 	//
 	// +optional
 	HTTP *HTTPClientTimeout `json:"http,omitempty"`
 }
 
+type TCPClientTimeout struct {
+	// IdleTimeout for a TCP connection. Idle time is defined as a period in which there are no
```
guydc, line 58: I assume that this will be implemented for the envoy TCP Proxy, right? Maybe we should state that these settings only apply to `tcproute`, `tlsroute` resources, so that users don't expect this timeout to apply also in case of HTTP  (over TCP) connections?

guydc (commented):
Overall looks good me, just one comment on the doc

api/v1alpha1/timeout_types.go
aoledk: Yeah, the state will be necessary to clear confusion. TCPClientTimeout can only be applied on listener whose protocol is TCP or TLS, it'll be used to set Envoy TCP Proxy's idle timeout eventually.

shawnh2 (approved):
LGTM!

arkodg (approved):
LGTM thanks !

PR #3348, opened 2024-05-08
feat(api): add consistent hash table size api

### LINKED ISSUE #3272: Support knob to configure consistent hash table size
*Description*:
> Backend whose LoadBalancer type is ConsistentHash will be translated to xDS cluster whose type is Maglev, but we can't configure maglev table size which is very important in long lived TCP load balancing use case. Bigger size, smaller disruption.

[optional *Relevant Links*:]
> https://gateway.envoyproxy.io/v1.0.1/api/extension_types/#consistenthash
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig

aoledk:

Add knob in BackendTrafficPolicy to configure consistent hash table size.

Fixes #3272


api/v1alpha1/loadbalancer_types.go
```diff
@@ -55,6 +55,14 @@ const (
 // load balancer policy
 type ConsistentHash struct {
 	Type ConsistentHashType `json:"type"`
+
+	// The table size for consistent hashing, must be prime number limited to 5000011.
+	//
+	// +kubebuilder:validation:Minimum=1
```
shawnh2, line 61: should we make it `2`? because `1` is actually not a prime number.

aoledk: Sure, make the minimum to the smallest prime number `2` now.

api/v1alpha1/loadbalancer_types.go
```diff
@@ -58,7 +58,7 @@ type ConsistentHash struct {
 
 	// The table size for consistent hashing, must be prime number limited to 5000011.
```
zirain, line 59: can we add CEL to validate it must be prime number?

aoledk: Seems like CEL in Kubernetes [^1] doesn't support complex validation logic like `IsPrime` function. But we can validate whether table size is prime number when translating ClientTrafficPolicy to IR.

```go
func IsPrime(n int) bool {
    if n <= 1 {
        return false
    }
    for i := 2; i*i <= n; i++ {
        if n%i == 0 {
            return false
        }
    }
    return true
}
```

[^1]: https://kubernetes.io/docs/reference/using-api/cel/#cel-options-language-features-and-libraries

shawnh2: SGTM

api/v1alpha1/loadbalancer_types.go
```diff
@@ -67,6 +67,14 @@ type ConsistentHash struct {
...
 	// +notImplementedHide
 	Header *Header `json:"header,omitempty"`
+
+	// The table size for consistent hashing, must be prime number limited to 5000011.
+	//
+	// +kubebuilder:validation:Minimum=2
+	// +kubebuilder:validation:Maximum=5000011
+	// +kubebuilder:default=65537
+	// +optional
+	TableSize *uint64 `json:"tableSize,omitempty"`
```
shawnh2, line 77: can you also add a comment `// +notImplementedHide`? we can remove this comment after this api has been implemented.

shawnh2 (approved):
LGTM! Thanks for adding this! Leave some comments.

api/v1alpha1/loadbalancer_types.go
aoledk: Sure.

arkodg (approved):
LGTM thanks !

PR #3357, opened 2024-05-09
feat(translator): Implement header hash policy for consistent hash load balancers

### LINKED ISSUE #3341: Add support for the header hash policy in the consistent hash load balancer

harrisonturton:
**What type of PR is this?**

Adds a new feature to the translator.

This PR adds support for the [Header](https://github.com/envoyproxy/envoy/blob/8bc2e917519936ab4cb1e9fa52fd5456a4fce4f7/api/envoy/config/route/v3/route_components.proto#L834-L846) hash policy in the Envoy Gateway API, which will allow users the configure the Envoy Gateway to use session affinity based on the value of a header.

See #3341 for justification, discussed in [this Slack thread.](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1715069442245479) The user-facing API changes were introduced in #3342, this PR implements them.

Fixes #3341


harrisonturton:
I think these CI failures are flakes. They are `context deadline exceeded` errors and seem to be unrelated to my changes.

shawnh2 (approved):
LGTM! Thanks for adding this.

cnvergence (approved):

arkodg (approved):
LGTM thanks !

PR #3371, opened 2024-05-11
feat: Adding extension server policy handling.

### LINKED ISSUE #2975: Support ExtensionContextPolicy resources to provide state to Extension Server hooks
*Description*:
The Extension Server protocol supports four hooks today. Only one of these hooks supports receiving context as part of its invocation.

While it's possible to register a CRD and attach specific instances of it to an `HTTPRoute` during calls to the route modification hook, it's not possible to get similar context with the other hooks. In the example extension server (#2931) this is worked around by developing a full operator for context, and then when the `HTTPListener` modification hook is called then correlating the internal extension server state with the hook details. 

I propose adding a policy document that would allow attaching context to calls to the extension hook server. For example:

```yaml
kind: ExtensionContextPolicy
metadata:
   name:  my-context
spec:
  targetRef:
    kind: Gateway
    name: eg
    namespace: default
  # Support sending multiple documents to the extension server as contextual information
  extensionRef:
  - group: example.myextension.io
     kind: LuaScript
     name: my-lua-script
---
apiVersion: example.myextension.io/v1
kind: LuaScript
metadata:
   name: my-lua-script
spec:
   lua: # some lua script
```

The `ExtensionContextPolicy` can be attached to:
1. `Gateway` resources (entire resource or just a section), to be sent to the Listener modification hook
2. `HTTPRoute` resources, to be sent to the Route modification hook

By passing the contextual resource as a parameter, extension server development is made easier since the extension server doesn't need to implement an operator itself for the common use-cases.

liorokman:
This PR implements extension server policies which can be attached to HTTP listeners.

See the full description in the issue (#2975)

Fixes #2975


guydc:
cc @AliceProxy

liorokman:
@AliceProxy @arkodg @guydc  PTAL

internal/gatewayapi/resource.go
```diff
@@ -56,6 +56,7 @@ type Resources struct {
 	SecurityPolicies       []*egv1a1.SecurityPolicy       `json:"securityPolicies,omitempty" yaml:"securityPolicies,omitempty"`
 	BackendTLSPolicies     []*gwapiv1a3.BackendTLSPolicy  `json:"backendTLSPolicies,omitempty" yaml:"backendTLSPolicies,omitempty"`
 	EnvoyExtensionPolicies []*egv1a1.EnvoyExtensionPolicy `json:"envoyExtensionPolicies,omitempty" yaml:"envoyExtensionPolicies,omitempty"`
+	ExtServerPolicies      []unstructured.Unstructured    `json:"extServerPolicies,omitempty" yaml:"extServerPolicies,omitempty"`
```
arkodg, line 59:
```suggestion
	ExtensionServerPolicies      []unstructured.Unstructured    `json:"extensionServerPolicies,omitempty" yaml:"extServerPolicies,omitempty"`
```

internal/gatewayapi/extensionserverpolicy.go
```diff
@@ -0,0 +1,202 @@
...
+		res = append(res, *policy)
+		targetRefs, err := extractTargetRefs(policy)
+		if err != nil {
+			t.Logger.Sugar().Errorf("error finding targetRefs for policy %s: %w", policy.GetName(), err)
+			continue
+		}
+		for _, currTarget := range targetRefs {
+			if currTarget.Kind != KindGateway {
+				// TODO: mark this targetRef as an error
+				t.Logger.Sugar().Errorf("extension policy %s doesn't target a Gateway", policy.GetName())
```
arkodg, line 59: we've tried to keep the gateway-api lib as minimal as possible, so it can be reused again in other places in the future, can we instead collect `errs` here and return and let the runner log them ?

internal/gatewayapi/extensionserverpolicy.go
```diff
@@ -0,0 +1,202 @@
...
+	"k8s.io/utils/ptr"
+	gwv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gwv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/envoyproxy/gateway/internal/gatewayapi/status"
+	"github.com/envoyproxy/gateway/internal/ir"
+	"github.com/envoyproxy/gateway/internal/utils"
+)
+
+func (t *Translator) ProcessExtensionServerPolicies(policies []unstructured.Unstructured,
```
arkodg, line 25: I dont see this being used anywhere

internal/utils/field/field.go
```diff
@@ -12,6 +12,22 @@ import (
 	"unicode"
 )
 
+func SetMapValues(input map[string]any, fieldName string, fieldValue any) {
```
arkodg, line 15: can we reuse SetValue  ?

internal/provider/kubernetes/status.go
```diff
@@ -397,6 +398,41 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
...
+					Resource:       &obj,
+					Mutator: MutatorFunc(func(obj client.Object) client.Object {
+						t, ok := obj.(*unstructured.Unstructured)
+						if !ok {
+							err := fmt.Errorf("unsupported object type %T", obj)
+							errChan <- err
+							panic(err)
+						}
+						tCopy := t.DeepCopy()
+						tCopy.Object["status"] = *val
```
arkodg, line 428: lets make sure the field matches the type `gwapiv1a2.PolicyStatus`

internal/provider/kubernetes/controller.go
```diff
@@ -45,17 +45,18 @@ import (
...
+	statusUpdater     Updater
+	classController   gwapiv1.GatewayController
+	store             *kubernetesProviderStore
+	namespace         string
+	namespaceLabel    *metav1.LabelSelector
+	envoyGateway      *egv1a1.EnvoyGateway
+	mergeGateways     sets.Set[string]
+	resources         *message.ProviderResources
+	extGVKs           []schema.GroupVersionKind
+	extServerPolicies []schema.GroupVersionKind
```
arkodg, line 59: why not reuse `extGVKs` here ?

internal/gatewayapi/extensionserverpolicy.go
AliceProxy: I see it in `translator.go`:

```go
extServerPolicies := t.ProcessExtensionServerPolicies(
		resources.ExtServerPolicies, gateways, xdsIR)

```

internal/provider/kubernetes/controller.go
AliceProxy: We should keep them separate since they are for different things and the extGVKs go to the post route modify hook, and extServerPolicies go to the post listener hook. Makes it easier to check which is which.

AliceProxy (approved):
No complaints from me. I :+1:'d a few of Arko's suggestion, but I don't think this should really break any existing extensions other than needing to update the function signatures when they pull the latest package import with these changes.

internal/gatewayapi/extensionserverpolicy.go
liorokman: It is being used in `translator.go`.

internal/provider/kubernetes/controller.go
liorokman: Like @AliceProxy wrote, keeping it separate captures the fact that these GVKs are used for different purposes in different places.

Maybe I should rename `extGVKs` to `extRouteFilters` ?

internal/utils/field/field.go
liorokman: I don't think so. `SetValue` assumes that the input is a struct, this function assumes that the input is a `map[string]any`. 

It's possible to force `SetValue` to do this, but I don't think that this would make the code clearer.

internal/gatewayapi/resource.go
liorokman: Done

internal/gatewayapi/extensionserverpolicy.go
liorokman: Done.

internal/gatewayapi/translator.go
```diff
@@ -101,12 +102,15 @@ type Translator struct {
 
 	// Namespace is the namespace that Envoy Gateway runs in.
 	Namespace string
+
+	Logger logging.Logger
```
arkodg, line 106: still seeing the `Logger` and `LoggedErrors` being defined here
can we follow the approach taken in the `gatewayapi` lib where we have a top level `errs` field e.g. https://github.com/envoyproxy/gateway/blob/cffaf27297d80349f78f7b56933880f7dae0cea7/internal/gatewayapi/backendtrafficpolicy.go#L299 , in this case it would be in the `Translate` function and we'd append the errors returned  from individual functions e.g. https://github.com/envoyproxy/gateway/blob/cffaf27297d80349f78f7b56933880f7dae0cea7/internal/gatewayapi/backendtrafficpolicy.go#L321

liorokman: The difference is that in `BackendTrafficPolicy` the errors are aggregated into a single `error` and then stored in the policy status. 

In this policy, the decision was not to update the policy status section for errors, which means that errors would need to be propagated all the way to the runner in order to be logged. This is a pretty large structural change.

I will remove the logger, but I think keeping the `LoggedErrors` approach is better.

arkodg: I was assuming this would be a best effort approach in the extension manager, but it looks like we've enhanced the gateway api translator to make sure the extension policy is conformant. In that case, lets populate the `Status` of the policy instead of using `LoggedErrors`

internal/provider/kubernetes/status.go
```diff
@@ -397,6 +398,41 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context) {
 		)
 		r.log.Info("envoyExtensionPolicy status subscriber shutting down")
 	}()
+
+	// EnvoyExtensionPolicy object status updater
+	go func() {
```
arkodg, line 403: can we wrap a `if ExtensionManager enabled` around this ?

internal/gatewayapi/translator.go
liorokman: When an extension policy doesn't implement the required contract - it has no `targetRef` or `targetRefs` attribute in its spec - then it's not possible to correctly place an error message in the policy status section. This is because finding the right place for the condition requires having the `targetRef`.

I think that there's value in logging this, though, to help with debugging sessions where the user wants to understand why the policy is not being sent to the extension server.

internal/provider/kubernetes/status.go
liorokman: Done.

internal/gatewayapi/runner/runner.go
```diff
@@ -97,6 +99,9 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				}
 				// Translate to IR
 				result := t.Translate(resources)
+				if result.LoggedErrors != nil {
```
arkodg, line 102: changing the `Translate` signature to return `result, err` and logging that would be more idiomatic

liorokman: I disagree.

The idiomatic nature would require that the error returned from `Translate` be treated as an actual error, since the caller can't distinguish between aggregated errors that are really errors and aggregated errors that should just be logged.

If the `Translate` signature changes to include an error, the `Runner`  should then change to treat the returned error _as an error_. Just logging a returned error and then ignoring it by continuing would appear like a bug.

If, in the future, there are show-stopping errors that would be returned from the `Translate` function, it would be impossible to distinguish those errors that should be handled by the `Translate` callers in some concrete way, and those that we just want printed to the log right now.

arkodg: this is a long running go routine / runner, we only want log errors, not return or shut down, and track them using telemetry metrics. 
There is currently no err object we are returning, so there is no distinction of errors, there are just logged errors

internal/gatewayapi/runner/runner.go
```diff
@@ -96,7 +98,11 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 					t.ExtensionGroupKinds = extGKs
 				}
 				// Translate to IR
-				result := t.Translate(resources)
+				result, err := t.Translate(resources)
+				if err != nil {
+					// Currently all errors that Translate returns should just be logged
+					r.Logger.Error(err, "errors detected during translation")
```
arkodg, line 104: non blocking comment, something we can add in a follow up
is to send the err to the `errChan` similar to
https://github.com/envoyproxy/gateway/blob/3772293d83239bc568e62d3feac7f46e91c3cb6d/internal/infrastructure/runner/runner.go#L84C6-L84C20
to bump the metrics

arkodg (approved):

guydc (approved):

PR #3388, opened 2024-05-13
bug: Route extension filters with different types but the same name and namespace aren't correctly cached

liorokman:
The provider caches `extensionRef` instances based on their name and namespace. But the actual object stored is an Unstructured instance, which means that if two instances of two different objects have the same name and namespace, then the cache would override the first with the second.

This PR adds a unit test to include this situation and fixes the current cache to also consider the extensionRef's type when caching it.

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -131,7 +131,7 @@ type resourceMappings struct {
 	// extensionRefFilters is a map of filters managed by an extension.
 	// The key is the namespaced name of the filter and the value is the
 	// unstructured form of the resource.
-	extensionRefFilters map[types.NamespacedName]unstructured.Unstructured
+	extensionRefFilters map[message.NamespacedNameAndType]unstructured.Unstructured
 }
 
 func newResourceMapping() *resourceMappings {
@@ -143,7 +143,7 @@ func newResourceMapping() *resourceMappings {
 		allAssociatedTCPRoutes:   map[string]struct{}{},
 		allAssociatedUDPRoutes:   map[string]struct{}{},
 		allAssociatedBackendRefs: map[gwapiv1.BackendObjectReference]struct{}{},
-		extensionRefFilters:      map[types.NamespacedName]unstructured.Unstructured{},
+		extensionRefFilters:      map[message.NamespacedNameAndType]unstructured.Unstructured{},
 	}
 }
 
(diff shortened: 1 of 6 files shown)


guydc (dismissed):
LGTM!

internal/provider/kubernetes/controller.go
```diff
@@ -121,14 +121,14 @@ type resourceMappings struct {
 	// extensionRefFilters is a map of filters managed by an extension.
 	// The key is the namespaced name of the filter and the value is the
 	// unstructured form of the resource.
-	extensionRefFilters map[types.NamespacedName]unstructured.Unstructured
+	extensionRefFilters map[message.NamespacedNameAndType]unstructured.Unstructured
```
shawnh2, line 124: can we reuse this structure or create a new similar one?

https://github.com/envoyproxy/gateway/blob/02ff9eb749ba7eda1a53dfcbd80fb2b6f46b23fe/internal/provider/kubernetes/helpers.go#L28

shawnh2: all fields are string type should be okay

liorokman: The problem with https://github.com/envoyproxy/gateway/blob/02ff9eb749ba7eda1a53dfcbd80fb2b6f46b23fe/internal/provider/kubernetes/helpers.go#L28 is that it doesn't contain the object's group, only the type.

This means that if it is used as the key, then these two different objects will be considered the same:

```yaml
---
apiVersion: some.group/v1
type: Type1
metadata:
   name: some-name
   namespace: default
---

apiVersion: some.other.group/v1
type: Type1
metadata:
   name: some-name
   namespace: default
```

shawnh2: so we can create a new structure with `group` here, instead of creating at `internal/message`

liorokman: Done.

internal/utils/misc.go
```diff
@@ -11,10 +11,30 @@ import (
 	"hash/fnv"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type NamespacedNameAndGroupKind struct {
```
shawnh2, line 19:
```suggestion
type NamespacedNameWithGroupKind struct {
```

prefer this name

internal/utils/misc.go
```diff
@@ -11,10 +11,30 @@ import (
...
+	types.NamespacedName
+	schema.GroupKind
+}
+
+// NamespacedNameAndType creates and returns object's NamespacedNameAndType.
```
shawnh2, line 24: comments also need to update

shawnh2 (commented):
Some comments on the naming, else LGTM

shawnh2 (approved):

AliceProxy (approved):
Nice work, thanks :+1:

Issue #3049, opened 2024-03-28
Better helm support for airgapped clusters with no access to public image registry


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Envoy Gateway uses 3 images

* docker.io/envoyproxy/gateway
* docker.io/envoyproxy/envoy
* docker.io/envoyproxy/ratelimit

Would be great if theres a simpler way to change the registry as a helm value without doing a helm template and edit

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

- [x] Shutdown Manager(Envoy Gateway) https://github.com/envoyproxy/gateway/pull/3269  https://github.com/envoyproxy/gateway/pull/3389
- [x] RateLimit https://github.com/envoyproxy/gateway/pull/3389
- [x] Statsd Exporter https://github.com/envoyproxy/gateway/issues/3369

zirain:
aren't they share same registry?

arkodg:
Let's track adding top level images configuration support for statsd exporter using this GH issue

shawnh2:
Should we close this? Leave undo to be tracked via https://github.com/envoyproxy/gateway/issues/3369

arkodg:
this is the parent issue / `epic`, so we should ideally leave both open :)

s0uky:
What about envoy proxy image? That should be configurable too.

zirain:
you should be able do that with https://gateway.envoyproxy.io/docs/api/extension_types/#envoyproxykubernetesprovider right now.

PR #3389, opened 2024-05-13
feat: add global image settings for all images in the EG chart

zhaohuabing:
This PR exposes a top-level global image configuration in the helm value file for images used in the EG helm chart.

```
global:
  images:
    envoyGateway:
      image: "private-hub/envoyproxy/gateway:abcdef12"
      pullPolicy: Always
      imagePullSecrets:
      - name: "secret1"
      - name: "secret2"
    ratelimit:
      image: "private-hub/envoyproxy/ratelimit:test"

```

Future work to support air-gapped envirionment:
* Support `pullPolicy` and `imagePullSecrets` for ratelimit
* Support  custom image setting for statsd-exporter 
* Support custom image setting for envoy

Those will be addressed in follow-up PRs.

part of: https://github.com/envoyproxy/gateway/issues/3049


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -1,9 +1,27 @@
...
+      image: ratelimit
+      tag: 'master'
+
 deployment:
   envoyGateway:
     image:
-      repository: ${ImageRepository}
-      tag: '${ImageTag}'
-    imagePullPolicy: ${ImagePullPolicy}
+      repository: ""
```
zhaohuabing, line 22: Image settings are retained here for backward compatibility.

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -1,9 +1,27 @@
+# The global settings for the Envoy Gateway Helm chart.
+# These values will be used if the values are not overridden in the other sections.
+global:
```
zhaohuabing, line 3: Not sure a `global` umbrella makes sense here. 

If may be other `global` options in the value files, however, the EG value file is not that large.

arkodg:
@zhaohuabing what happens what the image is set in top level as well as deployment, which one wins ? I'd vote for most specific one (deployment) wins

zhaohuabing:
> @zhaohuabing what happens what the image is set in top level as well as deployment, which one wins ? I'd vote for most specific one (deployment) wins

Deployment level wins. Ideally we should remove the deployment level config. I keep it for back compatibility.

site/content/en/latest/install/api.md
```diff
@@ -31,12 +31,11 @@ The Helm chart for Envoy Gateway
 | certgen.rbac.labels | object | `{}` |  |
 | config.envoyGateway.gateway.controllerName | string | `"gateway.envoyproxy.io/gatewayclass-controller"` |  |
 | config.envoyGateway.logging.level.default | string | `"info"` |  |
-| config.envoyGateway.provider.kubernetes.shutdownManager.image | string | `"docker.io/envoyproxy/gateway:latest"` |  |
 | config.envoyGateway.provider.type | string | `"Kubernetes"` |  |
 | createNamespace | bool | `false` |  |
-| deployment.envoyGateway.image.repository | string | `"docker.io/envoyproxy/gateway"` |  |
-| deployment.envoyGateway.image.tag | string | `"latest"` |  |
-| deployment.envoyGateway.imagePullPolicy | string | `"IfNotPresent"` |  |
+| deployment.envoyGateway.image.repository | string | `""` |  |
```
arkodg, line 36: should we delete this fields then ?

tools/make/helm.mk
```diff
@@ -28,9 +28,15 @@ helm-install: helm-generate ## Install envoy gateway helm chart from OCI registr
...
 .PHONY: helm-generate
 helm-generate:
-	ImageRepository=${IMAGE} ImageTag=${TAG} ImagePullPolicy=${IMAGE_PULL_POLICY} envsubst < charts/gateway-helm/values.tmpl.yaml > ./charts/gateway-helm/values.yaml
+	GatewayImage=${IMAGE}:${TAG} GatewayImagePullPolicy=${IMAGE_PULL_POLICY} envsubst < charts/gateway-helm/values.tmpl.yaml > ./charts/gateway-helm/values.yaml
 	helm lint charts/gateway-helm
 
-helm-template: ## Template envoy gateway helm chart.
+HELM_VALUES := $(wildcard test/helm/*.in.yaml)
+
+helm-template: ## Template envoy gateway helm chart.z
```
arkodg, line 36: imo lets create a new `helm-test` target instead of reusing this one

zirain: it not a test, because it will alwasy pass. it just help code review.

arkodg (approved):
LGTM thanks, this will def simplify operations !

zirain (approved):

site/content/en/latest/install/api.md
zhaohuabing: Just keep this for back compatibility.  If removed, the existing helm value file containing these values won't work, which will surprise users.

Issue #3278, opened 2024-04-26
Add a knob to make per_connection_buffer_limit_bytes of cluster configurable


aoledk:
*Description*:
> EG can configure `per_connection_buffer_limit_bytes` on xDS Listener currently via #2600, not available to configure `per_connection_buffer_limit_bytes` on xDS Cluster, default to `32KB`. A new knob will be better for entire buffer sizing control of  Envoy provisioned by EG.

[optional *Relevant Links*:]
> https://gateway.envoyproxy.io/latest/api/extension_types/#connection
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster

ShyunnY:
hey, @aoledk 
Do you have any ideas or are you currently working on this PR?

aoledk:
> hey, @aoledk Do you have any ideas or are you currently working on this PR?

@ShyunnY Perhaps this feature should be implemented via BackendTrafficPolicy, similar way like https://gateway.envoyproxy.io/latest/api/extension_types/#connection works. And I'm not working on this Issue.

aoledk:
@guydc When EG is used as L4 proxy and serves massive long lived TCP connection, the buffer size of <Downstream, Envoy> and <Envoy, Upstream> connections should be controlled, to reduce memory consumption of Envoy.

PR #3390, opened 2024-05-14
api: add the buffer limit knob to the cluster

ShyunnY:
**What type of PR is this?**

api: add the buffer limit knob to the cluster

This PR aims to define the configuration structure of the `per_connection_buffer_limit_bytes` feature knob

Related issues #3278


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -103,6 +104,15 @@ type BackendTrafficPolicySpec struct {
...
 	Compression []*Compression `json:"compression,omitempty"`
+
+	// PerConnectionBufferLimitBytes Soft limit on size of the cluster’s connections read and write buffers.
+	// If unspecified, an implementation defined default is applied (32768 bytes).
+	// For example, 20Mi, 1Gi, 256Ki etc.
+	// Note: that when the suffix is not provided, the value is interpreted as bytes.
+	//
+	// +kubebuilder:validation:XValidation:rule="type(self) == string ? self.matches(r\"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\") : type(self) == int",message="PerConnectionBufferLimitBytes must be of the format \"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\""
+	// +optional
+	PerConnectionBufferLimitBytes *resource.Quantity `json:"perConnectionBufferLimitBytes,omitempty"`
```
shawnh2, line 115: should we add a top level structure `Connection` like what we did in https://github.com/envoyproxy/gateway/blob/bd72474a1ff02b1e8c96efa6ea16f9b5ccb9b2e9/api/v1alpha1/connection_types.go#L14?

ShyunnY: IMO, we don't need the complexity of adding a parent type structure for this functionality.
+ First of all, this function is only used as a knob to supplement the configuration of the cluster's buffer.
+ Secondly, too much redundancy does not seem like a good thing (**less is more**). At least currently we have no plans to provide more other functions in BTP. In other words: If we need to add a `Connection` parent structure, maybe we should reconstruct the multiple types of BTP and divide them into `Connection`. (This may be implemented in the future)

shawnh2: In this way, `ConnectionBufferLimit` maybe a simple naming.

ShyunnY: this looks good :)

ShyunnY: Done!

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -103,6 +104,15 @@ type BackendTrafficPolicySpec struct {
...
 	Compression []*Compression `json:"compression,omitempty"`
+
+	// ConnectionBufferLimitBytes Soft limit on size of the cluster’s connections read and write buffers.
+	// If unspecified, an implementation defined default is applied (32768 bytes).
+	// For example, 20Mi, 1Gi, 256Ki etc.
+	// Note: that when the suffix is not provided, the value is interpreted as bytes.
+	//
+	// +kubebuilder:validation:XValidation:rule="type(self) == string ? self.matches(r\"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\") : type(self) == int",message="ConnectionBufferLimitBytes must be of the format \"^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$\""
+	// +optional
+	ConnectionBufferLimitBytes *resource.Quantity `json:"connectionBufferLimitBytes,omitempty"`
```
arkodg, line 115: should we move this under a `connection` field ? do we see any other values being added in here ?
```
connection:
  bufferLimit
```

zirain:
```yaml
connection:
  bufferLimit
```
is better for first look.

ShyunnY: If we wanted to stay in a backwards compatible way we could set it under "connection", but at the moment I think it might be more appropriate to just use one field (_as it's not yet clear what additional fields will be added_). what do you think?

@arkodg  :)

ShyunnY:
By the way, when `Connection` exists in both CTP and BTP, we can abstract them both into `api/v1alpha1/connection_types.go` (_only a small refactoring code is required_).

The current `connection_types.go` only defines about CTP:
https://github.com/envoyproxy/gateway/blob/8206e11d7ed3a04722b847f936437af66552cdef/api/v1alpha1/connection_types.go#L14-L26

Can we structure it like this?
```go
type struct BackendTrafficPolicyConnection {
  ...
}

type struct ClientTrafficPolicyConnection {
  ...
}

```

api/v1alpha1/backendtrafficpolicy_types.go
ShyunnY: In the end, I chose to add a parent type `Connection` to it instead of just using a single field.
The reason can be seen: https://github.com/envoyproxy/gateway/pull/3390#issuecomment-2119790669

shawnh2:
prefer `BackendConnection` and `ClientConnection`

ShyunnY:
wow, this looks great, I vote!
In this PR we will focus on the design of the bufferLimit API, and I will open an additional PR to solve the _reconstruction problem_ in the future.
@shawnh2

arkodg (approved):
LGTM thanks !

lets track the `connection` refactor with a separate GH issue

zirain (approved):

Issue #3361, opened 2024-05-09
Mixed metrics when enable Open Telemetry sink for EG CP & DP at the same time


shawnh2:
*Description*:

If we enable the Open Telemetry sink for EG control-plane and data-plane at the same time, by retrieving metrics from OTel-Collector, we will get mixed metrics from both CP & DP.

Do we need to separate these metrics? 

- maybe by adding a label field for each metrcis to indicate which it comes from (DP or CP) could avoid this

/cc @envoyproxy/gateway-maintainers 

*Repro steps*:

- CP follows https://gateway.envoyproxy.io/contributions/design/eg-metrics/#example
- DP follows https://gateway.envoyproxy.io/v1.0.1/tasks/observability/proxy-observability/#metrics

*Environment*:
latest

zirain:
TBH, I'm  not a fan of OpenTelemetry Metrics, is that anything like pod name or identity could be found?

shawnh2:
the only difference is: metrics for DP have label prefix `envoy`

zirain:
EG should send something for  identity

shawnh2:
If is possible, we can add `namespace` label for all the metrics from CP.

xref: https://github.com/envoyproxy/gateway/pull/3393#issuecomment-2114028252

zirain:
@shawnh2 can we close this?

PR #3393, opened 2024-05-15
feat: add metrics instrumentation for xds server

### LINKED ISSUE #2090: Implement Metrics instrumentation for xDS Server
*Description*:
would be used.

Implement Metrics instrumentation for xDS Server

shawnh2:
**What type of PR is this?**

image

Fixes #2090

### DIFF
--- a/internal/xds/cache/metrics.go
+++ b/internal/xds/cache/metrics.go
@@ -1,0 +1,20 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package cache
+
+import "github.com/envoyproxy/gateway/internal/metrics"
+
+var (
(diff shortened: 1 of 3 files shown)


Xunzhuo (commented):
I think we should have more histogram metrics for measuring the xds duration.

shawnh2:
> I think we should have more histogram metrics for measuring the xds duration.



> I think we should have more histogram metrics for measuring the xds duration.

great! thanks, let me trying add more metrics for this

ShyunnY:
Are you considering adding additional `namespace labels` to metrics? This will facilitate our use of ns select in PromQL.

shawnh2:
> Are you considering adding additional `namespace labels` to metrics? This will facilitate our use of ns select in PromQL.

thanks for reporting, this will be investigate via https://github.com/envoyproxy/gateway/issues/3361

arkodg (approved):
this is going to be very useful, thanks for adding this !

zirain (approved):
It would be better if you can update the grafana dashboard.

PR #3399, opened 2024-05-15
feat: authorization implementation

zhaohuabing:
This PR adds gateway API translator, XDS translator, and e2e tests for the `Authorization` setting of the `SecurityPolicy`.

Only source IP is supported for now.  Other principals such as JWT claims, basic auth, etc. will be supported later.

Fix: https://github.com/envoyproxy/gateway/issues/2462 
Related: https://github.com/envoyproxy/gateway/issues/2250

Example config:
```yaml
kind: HTTPRoute
metadata:
  name: http-with-authorization-client-ip
spec:
  parentRefs:
  - name: same-namespace
  rules:
  - matches:
    - path:
        type: Exact
        value: /protected
    backendRefs:
    - name: infra-backend-v1
      port: 8080
---
      
kind: SecurityPolicy
metadata:
  name: authorization-client-ip
spec:
  targetRef:
    kind: HTTPRoute
    name: http-with-authorization-client-ip
  authorization:
    defaultAction: Deny
    rules:
    - action: Allow
      principal:
        clientCIDRs:
        - 10.0.1.0/24
        - 10.0.2.0/24
---

kind: ClientTrafficPolicy
metadata:
  name: enable-client-ip-detection
spec:
  clientIPDetection:
    xForwardedFor:
      numTrustedHops: 1
  targetRef:
    kind: Gateway
    name: same-namespace
```

Test results:
```bash
curl -I --header "Host: www.example.com" --header "X-Forwarded-For: 10.0.2.1" http://172.18.255.203:80/protected
HTTP/1.1 200 OK

curl -I --header "Host: www.example.com" --header "X-Forwarded-For: 10.0.3.1" http://172.18.255.203:80/protected
HTTP/1.1 403 Forbidden
```


test/e2e/testdata/authorization-client-ip.yaml
```diff
@@ -0,0 +1,98 @@
...
+# This is a client traffic policy that enables client IP detection using a custom header.
+# So, the client IP can be detected from the custom header `client-ip` and used for authorization.
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: ClientTrafficPolicy
+metadata:
+  name: enable-client-ip-detection
+  namespace: gateway-conformance-infra
+spec:
+  clientIPDetection:
+    customHeader:
```
zhaohuabing, line 93: The current implementation works for the `customHeader` ClientIPDetection, but not for XFF.

In order to also support XFF, we need to use the [XFF IP detection extension](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/original_ip_detection/xff/v3/xff.proto#extension-envoy-http-original-ip-detection-xff) instead of the [xff_num_trusted_hops](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for) HCM setting for client IP detection.

@davidalger @ardikabs  Are there specific reasons why we prefer `xff_num_trusted_hops` over XFF IP detection extension in the implementation of https://github.com/envoyproxy/gateway/pull/2535 ? Would it be possible to switch to the XFF IP detection extension?

zufardhiyaulhaq:
@zhaohuabing questions, what is the differentiation of Authorization API & SecurityPolicy API.
since currently JWT, basic auth is located on securitypolicy. do we plan to move it to Authorization API in the future?

zhaohuabing:
@zufardhiyaulhaq Authorization is part of the `SeucrityPolicy`, as the example in the PR description shows.

zufardhiyaulhaq:
ahh just realized that @zhaohuabing thanks!

internal/xds/translator/authorization.go
```diff
@@ -0,0 +1,266 @@
...
+			return nil
+		}
+	}
+
+	rbacFilter, err := buildHCMRBACFilter()
+	if err != nil {
+		return err
+	}
+
+	// Ensure the RBAC filter is the first one in the filter chain.
```
zetaab, line 67: hmm does this work if we later want to use like JWT user information for the authorization? We might need way to specify the order?

zhaohuabing: It's not what it means. It's a copy-paste typo. 

The original one is "Make sure the router filter is the terminal filter in the chain."

Anyway, we sort all the filters afterward, so the order doesn't mater here.

zetaab:
I am testing this and got error

`'Unable to translate authorization rule: invalid CIDR address: 1.2.3.4'`

I could apply the configuration fine, but I was wondering why it was not provisioned. I could see that error in securitypolicy status. Just thinking, should it be validated in creation that ip address contains mask as well?

api/v1alpha1/authorization_types.go
```diff
@@ -16,50 +17,49 @@ type Authorization struct {
...
 }
 
-// Rule defines the single authorization rule.
-// +notImplementedHide
-type Rule struct {
+// AuthorizationRule defines a single authorization rule.
+type AuthorizationRule struct {
+	// Name is a user-friendly name for the rule. It's just for display purposes.
+	// +optional
+	Name *string `json:"name"`
```
arkodg, line 32:
```suggestion
	Name *string `json:"name,omitempty"`
```

api/v1alpha1/authorization_types.go
```diff
@@ -16,50 +17,49 @@ type Authorization struct {
...
-	DefaultAction *RuleActionType `json:"defaultAction"`
+	DefaultAction *AuthorizationAction `json:"defaultAction"`
 }
 
-// Rule defines the single authorization rule.
-// +notImplementedHide
-type Rule struct {
+// AuthorizationRule defines a single authorization rule.
+type AuthorizationRule struct {
+	// Name is a user-friendly name for the rule. It's just for display purposes.
```
arkodg, line 30: used for logging ?

test/e2e/testdata/authorization-client-ip.yaml
```diff
@@ -0,0 +1,98 @@
...
+    defaultAction: Deny
+    rules:
+    - action: Allow
+      principal:
+        clientCIDRs:
+        - 10.0.1.0/24
+        - 10.0.2.0/24
+---
+# This is a client traffic policy that enables client IP detection using a custom header.
+# So, the client IP can be detected from the custom header `client-ip` and used for authorization.
```
arkodg, line 85: not seeing the custom header config here, and the e2e is using `X-Forwarded-For` :)

test/e2e/testdata/basic-auth.yaml
```diff
@@ -59,7 +59,6 @@ spec:
     group: gateway.networking.k8s.io
     kind: HTTPRoute
     name: http-with-basic-auth-1
-    namespace: gateway-conformance-infra
```
arkodg, line 62: what's happening here ?

api/v1alpha1/authorization_types.go
zhaohuabing: It could be used for logging, but more importanly, an explicit name helps to explain each rule in a `SecurityPolicy`.

Instead of 
```yaml
  authorization:
    defaultAction: Deny
    rules:
    - action: Allow
      principal:
        clientCIDRs:
        - 10.0.2.0/24
    - action: Allow
      principal:
        clientCIDRs:
        - 10.0.1.0/24    
```

We can add a name to each rule so people can know what's the purpose of each rule, like this:

```yaml
  authorization:
    defaultAction: Deny
    rules:
    - name: us-west-region
      action: Allow
      principal:
        clientCIDRs:
        - 10.0.2.0/24
    - name: us-east-region 
      action: Allow
      principal:
        clientCIDRs:
        - 10.0.1.0/24   
```

internal/xds/translator/authorization.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Build a list of matchers based on the rules.
+	// The matchers will be evaluated in order, and the first one that matches
+	// will be used to determine the action, the rest of the matchers will be
+	// skipped.
+	// If no matcher matches, the default action will be used.
+	for _, rule := range authorization.Rules {
+		// Build the IPMatcher based on the client CIDRs.
+		ipRangeMatcher := &ipmatcherv3.Ip{
+			StatPrefix: "source_ip",
```
arkodg, line 157: should this be called `client_ip` instead ?

internal/xds/translator/authorization.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// If there are no matchers, the default action will be used for all requests.
+	// Setting the matcher type to nil since Proto validation will fail if the list
+	// is empty.
+	if len(matcherList) == 0 {
+		routeCfgProto.Rbac.Matcher.MatcherType = nil
+	}
+
+	if err = routeCfgProto.ValidateAll(); err != nil {
```
arkodg, line 245: id say not needed here, we are running this in https://github.com/envoyproxy/gateway/blob/2880a55b014a09deacadaa121396ea23d0a85fd4/internal/xds/types/resourceversiontable.go#L78

internal/xds/translator/authorization.go
```diff
@@ -0,0 +1,265 @@
...
+		})
+	}
+
+	// Set the default action.
+	defaultAction := denyAction
+	if authorization.DefaultAction == egv1a1.AuthorizationActionAllow {
+		defaultAction = allowAction
+	}
+
+	routeCfgProto := &rbacv3.RBACPerRoute{
```
arkodg, line 218: can we represent this as `matcherList` where we append the default action as a matcher to the end of the `mactherList` ?

internal/xds/translator/authorization.go
```diff
@@ -0,0 +1,265 @@
...
+								TypedConfig: ipMatcher,
+							},
+						},
+					},
+				},
+			},
+			OnMatch: &matcherv3.Matcher_OnMatch{
+				OnMatch: &matcherv3.Matcher_OnMatch_Action{
+					Action: &cncfv3.TypedExtensionConfig{
+						Name:        "action",
```
arkodg, line 204: should this `Name` be from the `Name` field from the Authz API ?

test/e2e/testdata/basic-auth.yaml
zhaohuabing: Guess I touched the `delete` key accidentally :-)

internal/xds/translator/authorization.go
zhaohuabing: We need to validate the RBACPerRoute message before converting it to an Any since the ValidateAll can't validate `Any` message.

arkodg (dismissed):
LGTM thanks !

zhaohuabing:
> I am testing this and got error
> 
> `'Unable to translate authorization rule: invalid CIDR address: 1.2.3.4'`
> 
> I could apply the configuration fine, but I was wondering why it was not provisioned. I could see that error in securitypolicy status. Just thinking, should it be validated in creation of securitypolicy that ip address contains mask as well?

Good point. I added a regex validation for CIDR. It should prevent most invalidate input such as lacking mask, but still need a second check in gateway API translation for some corner cases.

> Otherwise looks good, I have verified and looks like working like should. However, I am just thinking that we might need to split the filter to two different rbac filters? I am thinking like ip policies as first in order, and then others after authentication things like jwt (that we can use jwt information for authorization)?

I actually proposed the same during the API review. :-)

Most gateway implementations have a standalone IP Allow/Deny list. However, some, like the Istio Authorization API, integrate this feature into their Authorization API.

Incorporating the client IP range in the Authorization enables use cases that require a combination of the client IP and other conditions. For instance, " allows admin users access an HTTPRoute from use-west-region, but not from other regions", which is either impossible or significantly challenging  to implement with two independent filters.

arkodg (dismissed):

zirain (dismissed):
LGTM, please fix the conflict.

api/v1alpha1/shared_types.go
```diff
@@ -446,3 +446,8 @@ type BackendRef struct {
 	// Only service Kind is supported for now.
 	gwapiv1.BackendObjectReference `json:",inline"`
 }
+
+// CIDR defines a CIDR Address range.
+// A CIDR can be an IPv4 address range such as "192.168.1.0/24" or an IPv6 address range such as "2001:0db8:11a3:09d7::/64".
+// +kubebuilder:validation:Pattern=`((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\/([0-9]+))|((([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))\/([0-9]+))`
```
zetaab, line 452: perhaps prettier would be use golang in https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/validation/securitypolicy_validate.go to check that its valid?

zetaab (commented):
one nit, otherwise lgtm

api/v1alpha1/shared_types.go
zhaohuabing: It's two-level validation, first level is k8s API server, the second level is in gateway API translation in `parseCIDR` method.

zetaab (approved):

PR #3420, opened 2024-05-17
fix(translator): set ignoreCase for header matchers in extAuth

### LINKED ISSUE #3372: extAuth - headersToExtAuth translation case-sensitivity
*Description*:
headersToExtAuth translation case-sensitivity mismatch

*Repro steps*:
A config like this:
```
---
kind: SecurityPolicy
metadata:
  name: authentik-ext-auth
  namespace: default
spec:
  targetRef:
    kind: HTTPRoute
    name: httpbin
    namespace: default
  extAuth:
    headersToExtAuth:
      - Cookie
    ...
```
is translated to exact string matches including case. Unfortunately request headers are folded to lower-case somewhere (maybe this is the actual bug), so "Cookie" will not get forwarded to the auth backend and is silently dropped.

With this change in extauth.go I was able to make logging in work:
```
--- a/internal/xds/translator/extauth.go
+++ b/internal/xds/translator/extauth.go
@@ -104,6 +104,7 @@ func extAuthConfig(extAuth *ir.ExtAuth) *extauthv3.ExtAuthz {
                        MatchPattern: &matcherv3.StringMatcher_Exact{
                                Exact: header,
                        },
+                       IgnoreCase: true,
                })
        }
```

*Environment*:
gateway v1.0.1

*Notes*:
Maybe this should be specifiable in https://gateway.envoyproxy.io/latest/api/extension_types/#extauth instead of a stringarray instead.
The same problem might apply to headersToBackend as well.

haoqixu:
**What type of PR is this?**

This PR set ignoreCase to true for header matchers in extAuth.

HTTP Header names are case-insensitive according to [RFC 2616](http://www.ietf.org/rfc/rfc2616.txt) and [envoy will normalize the header keys to be all lowercase by default](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_casing).

Fixes #3372

### DIFF
--- a/internal/xds/translator/extauth.go
+++ b/internal/xds/translator/extauth.go
@@ -113,6 +113,7 @@ func extAuthConfig(extAuth *ir.ExtAuth) *extauthv3.ExtAuthz {
 			MatchPattern: &matcherv3.StringMatcher_Exact{
 				Exact: header,
 			},
+			IgnoreCase: true,
 		})
 	}
 
@@ -178,6 +179,7 @@ func httpService(http *ir.HTTPExtAuthService) *extauthv3.HttpService {
 			MatchPattern: &matcherv3.StringMatcher_Exact{
 				Exact: header,
 			},
+			IgnoreCase: true,
 		})
 	}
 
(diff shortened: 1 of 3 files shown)


arkodg (approved):
LGTM thanks !

shawnh2 (approved):
lgtm thanks!

PR #3422, opened 2024-05-17
chore: support multi nodes cluster

zirain:


tools/hack/create-cluster.sh
```diff
@@ -6,12 +6,32 @@ set -euo pipefail
...
+
+
+KIND_CFG=$(cat <<-EOM
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+EOM
+)
+
```
arkodg, line 19: nit: can you add https://kind.sigs.k8s.io/docs/user/quick-start/#multi-node-clusters this in here ?

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

Issue #3405, opened 2024-05-16
Extend OIDC Configurability


coro:
*Description*:
Allow users to configure additional elements of the Envoy OIDC filter in the `SecurityPolicy`.

Specific examples that would be very valuable to configure:
- [use_refresh_token](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#extensions-filters-http-oauth2-v3-oauth2config:~:text=rfc6749%23section%2D2.3.1-,use_refresh_token,-(BoolValue)%20If)
- [default_expires_in](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#extensions-filters-http-oauth2-v3-oauth2config:~:text=value%20is%20false.-,default_expires_in,-(Duration)%20The)
- [default_refresh_token_expires_in](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#extensions-filters-http-oauth2-v3-oauth2config:~:text=for%20AJAX%20requests.-,default_refresh_token_expires_in,-(Duration)%20The)
- [forward_bearer_token](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#extensions-filters-http-oauth2-v3-oauth2config:~:text=their%20credential%20cookies.-,forward_bearer_token,-(bool)%20Forward) (as per #3395)

Many of these are either not configured by EG or hard coded to a default: https://github.com/envoyproxy/gateway/blob/a2fbbad5db9a6fe4a1afa0bc20d0217d022275ca/internal/xds/translator/oidc.go#L117-L174 

*Relevant Links*:
Original issue: #3178 
Related Issue: #3395 

cc @zhaohuabing @denniskniep

Issue #3395, opened 2024-05-15
OIDC SecurityPolicy: Ability to set forward_bearer_token to false


denniskniep:
*Description*:
Ability to set `forward_bearer_token` to `false` in [OIDC SecurityPolicy](https://gateway.envoyproxy.io/latest/tasks/security/oidc/)

(see also: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#extensions-filters-http-oauth2-v3-oauth2config)

Currently its statically set to true:
https://github.com/envoyproxy/gateway/blob/a2fbbad5db9a6fe4a1afa0bc20d0217d022275ca/internal/xds/translator/oidc.go#L148

Current Workaround possible via: [EnvoyPatchPolicy](https://gateway.envoyproxy.io/latest/tasks/extensibility/envoy-patch-policy/)

PR #3423, opened 2024-05-17
API: add more oidc configuration settings

zhaohuabing:
This RP adds more configuration settings to the OIDC API.

Related issue: https://github.com/envoyproxy/gateway/issues/3405 https://github.com/envoyproxy/gateway/issues/3395 https://github.com/envoyproxy/gateway/issues/3462

@coro @denniskniep


api/v1alpha1/oidc_types.go
```diff
@@ -50,6 +51,35 @@ type OIDC struct {
 	// The path to log a user out, clearing their credential cookies.
 	// If not specified, uses a default logout path "/logout"
 	LogoutPath *string `json:"logoutPath,omitempty"`
+
+	// ForwardBearerToken indicates whether the Envoy should forward the access
+	// token as a bearer token in the "Authorization" header to the backend.
+	// If not specified, defaults to false.
+	// +optional
+	ForwardBearerToken *bool `json:"forwardBearerToken,omitempty"`
```
zhaohuabing, line 59: This changes the current default behavior, which forwards the access token. 

We should mention this is the release doc.

arkodg: is there a batter name for `ForwardAccessTokenAsBearerToken` ?

api/v1alpha1/oidc_types.go
```diff
@@ -48,8 +49,50 @@ type OIDC struct {
...
+	// of the authorization server if it is provided. This field is only used when
+	// the expiry time is not provided by the authorization.
+	//
+	// If not specified, defaults to 0. In this case, the "expires_in" field in
+	// the authorization response must be set by the authorization server, or the
+	// OAuth flow will fail.
+	//
+	// +optional
+	// +notImplementedHide
+	DefaultTokenExpireTime *metav1.Duration `json:"defaultTokenExpireTime,omitempty"`
```
arkodg, line 75: should this be `DefaultIdTokenExpireTime` ?

api/v1alpha1/oidc_types.go
zhaohuabing: > is there a batter name for `ForwardAccessTokenAsBearerToken` ?

I think Bearer Token is normally an Access Token, so we probably don't need to emphasize that in the name?

api/v1alpha1/oidc_types.go
zhaohuabing: The Envoy doc is somewhat vague in[ its description](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2config) of `DefaultTokenExpireTime`.

However, after digging into the code, it's clear that this applies to both the ID token and Access Token.

https://github.com/envoyproxy/envoy/blob/a112766683a6b82d20843c70b8aa4d7eda29048e/source/extensions/filters/http/oauth2/oauth_client.cc#L183

api/v1alpha1/oidc_types.go
zhaohuabing: Thought about it again, it's better to make it clear even though the name becomes a bit longer.

api/v1alpha1/oidc_types.go
zhaohuabing: Or `DefaultIdAndAccessTokenExpireTime`, that's quite lengthy :-)

api/v1alpha1/oidc_types.go
zhaohuabing: Change it to ForwardAccessToken and add a comment to explain that the access token will be forwarded via the Authorization header Bearer scheme to the upstream.

zetaab (approved):
LGTM, going to need all of these :)

api/v1alpha1/oidc_types.go
```diff
@@ -48,8 +49,49 @@ type OIDC struct {
...
+	// of the authorization server if it is provided. This field is only used when
+	// the expiry time is not provided by the authorization.
+	//
+	// If not specified, defaults to 0. In this case, the "expires_in" field in
+	// the authorization response must be set by the authorization server, or the
+	// OAuth flow will fail.
+	//
+	// +optional
+	// +notImplementedHide
+	DefaultTokenExpireIn *metav1.Duration `json:"defaultTokenExpireIn,omitempty"`
```
arkodg, line 74:
```suggestion
	DefaultTokenExpiresIn *metav1.Duration `json:"defaultTokenExpireIn,omitempty"`
```

arkodg: we need a better name here `DefaultTokenExpiresIn` that encapsulates default expiry duration for the token

zhaohuabing: `DefaultTokenExpiryDuration`
`DefaultTokenTTL`

zirain: +1 on `DefaultTokenTTL`

arkodg (approved):

PR #3424, opened 2024-05-18
fix: filter BackendTLSPolicy by namespace

### LINKED ISSUE #3409: TLSBackendPolicy´s TargetRef Name must be unique across all namespaces
My TLSBackendPolicy has a Service as TargetRef. This service name must be unique across all namespaces. Otherwise it will not consistently select the certs for the appropriate routes.

Maybe the problem is that `policies` are not filtered by namespace with `backendNamespace `?
https://github.com/envoyproxy/gateway/blob/c5ec046be14c4bb9049a7d3633900d2b8207d40f/internal/gatewayapi/backendtlspolicy.go#L127-L134

Reproduce:
* Create 2 Services with same name but in different namespaces
* Create 2 TLSBackendPolicy with same name but in different namespaces
* Create 2 HTTPRoutes with same name but in different namespaces and different Hostnames

haoqixu:
**What type of PR is this?**

This PR filters BackendTLSPolicy by namespace.

From https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.1.0:
> the targetRef field is now a targetRefs list and these references no
longer include a namespace field.

From https://gateway-api.sigs.k8s.io/api-types/backendtlspolicy/:
> The BackendTLSPolicy and the Service must reside in the same namespace in order to prevent the complications involved with sharing trust across namespace boundaries. 

Fixes #3409


arkodg (approved):
LGTM thanks !

shawnh2 (approved):
LGTM! Thanks for the fix

PR #3437, opened 2024-05-20
chore: enable gci in golangci-lint

shawnh2:
we have gci settings in https://github.com/envoyproxy/gateway/blob/e847d68c2665d471912d66250990937b73787dc6/tools/linter/golangci-lint/.golangci.yml#L39,

but this setting is somehow not be enabled.


zirain (approved):
Use to think it's enabled.

arkodg (approved):
thanks for finding and fixing this gap !

PR #3302, opened 2024-04-29
refactor: error processing for xPolicy

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #2894, separated from #2926

PR #3440, opened 2024-05-20
feat(translator): Implement consistent hash table size in BackendTrafficPolicy

### LINKED ISSUE #3272: Support knob to configure consistent hash table size

aoledk:

 Implement consistent hash table size in BackendTrafficPolicy, follow up #3348

Fixes #3272


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -788,24 +801,32 @@ func (t *Translator) buildLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.
 		}
 	}
 
-	return lb
+	return lb, nil
 }
 
-func (t *Translator) buildConsistentHashLoadBalancer(policy *egv1a1.BackendTrafficPolicy) *ir.ConsistentHash {
+func (t *Translator) buildConsistentHashLoadBalancer(policy *egv1a1.BackendTrafficPolicy) (*ir.ConsistentHash, error) {
```
shawnh2, line 807: can we directly pass `policy.Spec.LoadBalancer.ConsistentHash`, in instead of the whole `BackendTrafficPolicy`?

internal/gatewayapi/testdata/backendtrafficpolicy-with-loadbalancer-invalid-consistent-hash-table-size.in.yaml
shawnh2, line 1: can we also add a yaml test about table size that violating the prime number rule?

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -1230,3 +1251,17 @@ func makeIrTriggerSet(in []egv1a1.TriggerEnum) []ir.TriggerEnum {
 	}
 	return irTriggers
 }
+
+// isPrimeNumber judges whether the number is a prime number
+func isPrimeNumber(n uint64) bool {
```
shawnh2, line 1256: can we use the prime number check method in stdlib https://pkg.go.dev/math/big#Int.ProbablyPrime?

this method is 100% accurate for inputs less than 2⁶⁴ since our maximum table size is 5000011.

internal/gatewayapi/backendtrafficpolicy.go
aoledk: The reason I pass the whole `BackendTrafficPolicy` is many existing internal building functions in function `buildAbc(policy)` used to building `Spec.Abc` is invoked as `buildAbcXyz(policy)`.

To keep the same function signature style.

- `buildHealthCheck(policy)`
  - `buildPassiveHealthCheck(policy)`
  - `buildActiveHealthCheck(policy)`
- `buildRateLimit(policy)`
  - `buildGlobalRateLimit(policy)`
  - `buildLocalRateLimit(policy)`

internal/gatewayapi/testdata/backendtrafficpolicy-with-loadbalancer-invalid-consistent-hash-table-size.in.yaml
aoledk: There are two test cases in this file, tableSize as `1024`(not a prime number) and `5000012`(exceed MAX tableSize).

shawnh2: got it thanks

internal/gatewayapi/backendtrafficpolicy.go
aoledk: Has replaced with function from `go/math`.

internal/gatewayapi/backendtrafficpolicy.go
shawnh2: got it! lgtm anyway

arkodg (dismissed):
LGTM thanks !

shawnh2:
LGTM! #3302 has introduced a new error processing way for BTP, can you update the code, then we are good to go!

shawnh2 (approved):
LGTM thanks

arkodg (approved):

Issue #2599, opened 2024-02-12
Propagate XFCC header to the backend


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Propagate the XFCC header to the client after validating the client via mutual TLS
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-forwardclientcertdetails

Decisions
1) Defaults / Opt in or opt out

2) Where should it live

a) https://gateway.envoyproxy.io/latest/api/extension_types/#headersettings
or
b) https://gateway.envoyproxy.io/latest/api/extension_types/#clientvalidationcontext

- [x]  https://github.com/envoyproxy/gateway/pull/3202
- [ ]  Docs: XFCC
- [ ]  E2E: XFCC

zufardhiyaulhaq:
how expensive is the computing process if we opt-in for the XFCC?

this can also being used later in the authorization API
https://github.com/envoyproxy/gateway/pull/2652

zufardhiyaulhaq:
no stale

PR #3202, opened 2024-04-15
feat: support forward client cert config XFCC header

zufardhiyaulhaq:
**What type of PR is this?**
configure envoy gateway to propagate XFCC header

**What this PR does / why we need it**:
this can be used to authorize client based on certificate from XFCC header

**Which issue(s) this PR fixes**:
https://github.com/envoyproxy/gateway/issues/2599

PR #3443, opened 2024-05-20
API: Follow up on XFCC header

zhaohuabing:
This a follow-up PR to address comments on https://github.com/envoyproxy/gateway/pull/3202.
* XFCC header setting naming
* Add validation for xfcc setting
* Add some comments

Close https://github.com/envoyproxy/gateway/issues/2599

### DIFF
--- a/api/v1alpha1/clienttrafficpolicy_types.go
+++ b/api/v1alpha1/clienttrafficpolicy_types.go
@@ -104,8 +104,15 @@ type HeaderSettings struct {
 	// +optional
 	EnableEnvoyHeaders *bool `json:"enableEnvoyHeaders,omitempty"`
 
-	// Configure Envoy proxy how to handle the x-forwarded-client-cert (XFCC) HTTP header.
-	// When enabled, Hash and By is always set
+	// XForwardedClientCert configures how Envoy Proxy handle the x-forwarded-client-cert (XFCC) HTTP header.
+	//
+	// x-forwarded-client-cert (XFCC) is an HTTP header used to forward the certificate
+	// information of part or all of the clients or proxies that a request has flowed through,
+	// on its way from the client to the server.
+	//
+	// Envoy proxy may choose to sanitize/append/forward the XFCC header before proxying the request.
+	//
+	// If not set, the default behavior is sanitizing the XFCC header.
 	// +optional
 	XForwardedClientCert *XForwardedClientCert `json:"xForwardedClientCert,omitempty"`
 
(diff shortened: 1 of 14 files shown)


cnvergence (approved):

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -139,57 +146,63 @@ const (
...
-	// Whether to forward the URI type Subject Alternative Name of the client cert.
-	ClientCertDataURI ClientCertData = "Uri"
+	// XFCCCertDataSubject is the Subject field of the current client certificate.
+	XFCCCertDataSubject XFCCCertData = "Subject"
+	// XFCCCertDataCert is the entire client certificate in URL encoded PEM format.
+	XFCCCertDataCert XFCCCertData = "Cert"
+	// XFCCCertDataChain is the entire client certificate chain (including the leaf certificate) in URL encoded PEM format.
+	XFCCCertDataChain XFCCCertData = "Chain"
+	// XFCCCertDataDNS is the DNS type Subject Alternative Name field of the current client certificate.
+	XFCCCertDataDNS XFCCCertData = "DNS"
```
arkodg, line 203: how does k8s deal with 3 letter acronyms for PascalCase ?

zhaohuabing: Gateway API and EG have many existing examples of `HTTP`, `URL`, `UDP`, `TCP`, `DNS` etc.

But I can't find "Dns" or "Uri". Envoy does use PascalCase for all acronyms.

arkodg: thanks, https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.ProtocolType is a good example that highlights this

internal/xds/translator/listener.go
```diff
@@ -30,6 +30,7 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 	"k8s.io/utils/ptr"
 
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
```
arkodg, line 33: we shouldnt be linking any Gateway API fields here, this lib should only be linked with the IR

arkodg: nvm this is internal to this project, but we should ideally avoid this too

arkodg (approved):

Issue #2624, opened 2024-02-16
Support Cookie based Consistent Hashing / Session Affinity


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

There is currently support for SourceIP based ConsistentHashing Loadbalancing
https://github.com/envoyproxy/gateway/blob/1775624114aedee731bee565a0a100eeee80fc13/api/v1alpha1/loadbalancer_types.go#L62

Would be useful to also add cookie based consistent hashing support which enable sticky session / session affinity use cases

arkodg:
Gateway API recently added support for Session Persistence (hard affinity), see details in https://gateway-api.sigs.k8s.io/geps/gep-1619/
It doesnt however add support to add more cookie attributes such as `SameSite`, its rationale being it should be set to sane defaults e.g. `Strict` 
https://gateway-api.sigs.k8s.io/geps/gep-1619/#secure-httponly-samesite

There are multiple instances where users may want to override this, see 
* https://github.com/projectcontour/contour/issues/2965
* https://github.com/istio/istio/issues/27006

@envoyproxy/gateway-maintainers please weigh in, I'm a +1 for
*  adding this feature, which falls into session affinity (weak affinity) into BTP for v1.1 with support for setting multiple attributes ([supported in envoy now](https://github.com/envoyproxy/envoy/issues/27234))
* also support the `sessionPersistence` field within HTTPRoute using Stateful session filter https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto#envoy-v3-api-msg-extensions-filters-http-stateful-session-v3-statefulsession
* dont add support for `BackendLBPolicy`, it was specifically designed to support implementations without Route level support for session persistence, Envoy can support it, so imo lets reduce the number of ways to configure `sessionPersistence`

PR #3444, opened 2024-05-21
api: cookie based consistent hashing

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/2624

Will use this underlying feature in Envoy https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-hashpolicy-cookie


arkodg:
have two opens questions here for @envoyproxy/gateway-maintainers and @envoyproxy/gateway-reviewers 
1. should the `Cookie` structure get split up into two more sub fields - `Existing` and `Generated` to make it easier to divide up the cookie fields, else it can get confusing ?
2. Should the `Attributes` be renamed to `ExtraAttributes` ?

zhaohuabing:
> should the Cookie structure get split up into two more sub fields - Existing and Generated to make it easier to divide up the cookie fields, else it can get confusing ?

One Cookie structure won't cause confusion to me.   Two separate structures seem extreme since there is only one optional `TTL` field. 

>Should the Attributes field be renamed to ExtraAttributes ?

Are there any "built-in" default Attributes? If not, the "Extra" is not needed.

arkodg:
> > should the Cookie structure get split up into two more sub fields - Existing and Generated to make it easier to divide up the cookie fields, else it can get confusing ?
> 
> One Cookie structure won't cause confusion to me. Two separate structures seem excessive since there is only one optional `TTL` field.
> 
> > Should the Attributes field be renamed to ExtraAttributes ?
> 
> Are there any "built-in" default Attributes? If not, the "Extra" is not needed.

TTL field translates into Max-Age attribute

zhaohuabing:
> TTL field translates into Max-Age attribute

Then "Extra" makes sense to me, though I don't have a strong preference for it.

shawnh2 (approved):

shawnh2:
> 
> 1. should the `Cookie` structure get split up into two more sub fields - `Existing` and `Generated` to make it easier to divide up the cookie fields, else it can get confusing ?
> 
> 2. Should the `Attributes` field be renamed to `ExtraAttributes` ?

1. Just `Cookie` sounds good to me, by the definition here, cookie is only be generated if it is not exist in current request. So it works more like a default option.

2. `Attributes` sounds good too.

zhaohuabing (approved):
LGTM thanks!

PR #3448, opened 2024-05-21
fix: Grafana dashboards conformance

BadLiveware:
Fixes some panels in grafana to:
* Use the variable datasource like the rest of the panels
* Sets request latency unit to ms
* Makes envoy-pod-resource.json use variable datasource like the other dashboards


shawnh2 (approved):
LGTM! Thanks for correcting this

arkodg (approved):
LGTM thanks !

Issue #3257, opened 2024-04-23
Installation fails on vanilla cluster using the latest release


dprotaso:
*Description*:
Create a kind cluster and then apply the latest envoy gateway - it will fail

*Repro steps*:
```
$ kind create cluster
$ kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/latest/install.yaml..

customresourcedefinition.apiextensions.k8s.io/backendtlspolicies.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/gatewayclasses.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/gateways.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/grpcroutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/httproutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/referencegrants.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/tcproutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/tlsroutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/udproutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/backendtrafficpolicies.gateway.envoyproxy.io created
customresourcedefinition.apiextensions.k8s.io/clienttrafficpolicies.gateway.envoyproxy.io created
customresourcedefinition.apiextensions.k8s.io/envoyextensionpolicies.gateway.envoyproxy.io created
customresourcedefinition.apiextensions.k8s.io/envoypatchpolicies.gateway.envoyproxy.io created
customresourcedefinition.apiextensions.k8s.io/securitypolicies.gateway.envoyproxy.io created
namespace/envoy-gateway-system created
serviceaccount/envoy-gateway created
configmap/envoy-gateway-config created
clusterrole.rbac.authorization.k8s.io/eg-gateway-helm-envoy-gateway-role created
clusterrolebinding.rbac.authorization.k8s.io/eg-gateway-helm-envoy-gateway-rolebinding created
role.rbac.authorization.k8s.io/eg-gateway-helm-infra-manager created
role.rbac.authorization.k8s.io/eg-gateway-helm-leader-election-role created
rolebinding.rbac.authorization.k8s.io/eg-gateway-helm-infra-manager created
[...]

PR #3451, opened 2024-05-21
docs:use server side apply when installing EG using kubectl

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/3257


shawnh2 (approved):
lgtm

PR #3092, opened 2024-04-04
Run Envoy Gateway like DaemonSet

zvlb:
Previos PR - https://github.com/envoyproxy/gateway/pull/3031

Issue - https://github.com/envoyproxy/gateway/issues/2969

If all ok - I will add tests)

PR #3452, opened 2024-05-22
feat: add knob for disable service

zvlb:
Continue work with run Envoy Gateway on baremetal K8s installations without LoadBalancer.
Previos PRs: 
- [Run Envoy Gateway like DaemonSet](https://github.com/envoyproxy/gateway/pull/3092)
- [feat: add knob for switching container port](https://github.com/envoyproxy/gateway/pull/3333)

### In this PR:

I add new field for `KubernetesServiceSpec` for 'don't deploy service'-logic.

### Real case:

If I deploy Envoy Gateway with `EnvoyProxy`:
```yaml
kind: EnvoyProxy
metadata:
  name: default
  namespace: envoy-gateway
spec:
  provider:
    kubernetes:
      UseListenerPortAsContainerPort: true
      envoyDaemonSet:
        pod:
          nodeSelector:
            node-role.kubernetes.io/gateway: "true"
        container:
          image: envoyproxy/envoy:v1.29.2
          securityContext:
            allowPrivilegeEscalation: true
            capabilities:
              add:
              - NET_BIND_SERVICE
              drop:
              - ALL
            seccompProfile:
              type: RuntimeDefault
        patch:
          type: StrategicMerge
          value:
            spec:
              template:
                spec:
                  hostNetwork: true
    type: Kubernetes
```

I don't need service for starting Envoy, bc I will connect to envoy with host-port (`hostNetwork: true`)

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -245,6 +245,11 @@ const (
 // +kubebuilder:validation:XValidation:message="loadBalancerSourceRanges can only be set for LoadBalancer type",rule="!has(self.loadBalancerSourceRanges) || self.type == 'LoadBalancer'"
 // +kubebuilder:validation:XValidation:message="loadBalancerIP can only be set for LoadBalancer type",rule="!has(self.loadBalancerIP) || self.type == 'LoadBalancer'"
 type KubernetesServiceSpec struct {
+	// Disable disables service if set to true.
+	// By default, the service is enabled.
+	// +optional
+	Disable bool `json:"disable,omitempty"`
+
 	// Annotations that should be appended to the service.
 	// By default, no annotations are appended.
 	//
--- a/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoyproxies.yaml
+++ b/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoyproxies.yaml
@@ -9901,6 +9901,11 @@ spec:
                               Annotations that should be appended to the service.
                               By default, no annotations are appended.
                             type: object
+                          disable:
+                            description: |-
+                              Disable disables service if set to true.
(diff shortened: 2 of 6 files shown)


zvlb:
I think this PR - last PR with logic.

After this I will create 1 PR with docs for describe "How use Envoy Gateway without LoadBalancer service"

arkodg:
disabling a service will mean the `Gateway` status condition will always be set to `Programmed=False`, because the address is being fetched from the `Service`, we need to address how to resolve this https://github.com/envoyproxy/gateway/blob/cad5c2348c33fcdb6d049d213865ffa34cfe8e0b/internal/provider/kubernetes/status.go#L402 first

zvlb:
It's hard)
I think need something like this:
- Check  KubernetesProvider. If used Daemonset and installed `hostNetwork` and `dnsPolicy` - don't check service.

But in code this will be the look hardcode...

arkodg:
@zvlb we'll need to figure out what is the valid address that can be added to `Gateway.Status.Addresses` for the DaemonSet case ? is node address good enough ?

ilnurkhal:
> disabling a service will mean the `Gateway` status condition will always be set to `Programmed=False`, because the address is being fetched from the `Service`, we need to address how to resolve this

It looks like there is another problem with [deployment check](https://github.com/envoyproxy/gateway/blob/435a4dc10f92349c7c6951dc8077ad911513ee2b/internal/gatewayapi/status/gateway.go#L163). I've tried daemonSet mode for envoyproxy and bumped into the `PROGRAMMED: False state`

Also in gateway's status:
```yaml
conditions:
    - lastTransitionTime: "2024-07-16T06:12:16Z"
      message: Deployment replicas unavailable
      observedGeneration: 1
      reason: NoResources
      status: "False"
      type: Programmed
```

PR #3459, opened 2024-05-23
feat(translator): Implement tcp idle timeout in ClientTrafficPolicy

### LINKED ISSUE #3343: feat: add support to configuring downstream idle timeout for TCP listener

aoledk:

Implement tcp idle timeout in ClientTrafficPolicy, follow up #3345

Fixes #3343


arkodg (dismissed):
LGTM thanks !

guydc (approved):
LGTM!

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -521,19 +525,34 @@ func translatePathSettings(pathSettings *egv1a1.PathSettings, httpIR *ir.HTTPLis
 	}
 }
 
-func translateClientTimeout(clientTimeout *egv1a1.ClientTimeout, httpIR *ir.HTTPListener) error {
+func buildClientTimeout(clientTimeout *egv1a1.ClientTimeout) (*ir.ClientTimeout, error) {
```
shawnh2, line 528: which func name prefix do we prefer here? I see we are using different prefix names for these func, like `buildXXX` and `translateXXX`?

Is a non blocking comment, but IMO we should unify them.

shawnh2: vote for `build`, it's consistent with all the build methods in BTP

arkodg (approved):

Issue #3355, opened 2024-05-09
Add-on / Extended Helm Chart


arkodg:
*Description*:
>Describe the issue.

Consider introducing a add-on / helm chart that may be common while Installing Envoy Gateway
e.g. Monitoring components highlighted in 
https://gateway.envoyproxy.io/v1.0.1/tasks/observability/proxy-observability/ & 
https://gateway.envoyproxy.io/v1.0.1/tasks/observability/grafana-integration/

Relates to https://github.com/envoyproxy/gateway/pull/2260

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
will this introduce a new image on registry?

arkodg:
yeah

zirain:
how to handle the image version, start from `v1.1.0`(same with others)?

arkodg:
Yeah keeping them in sync and using the same CI tooling seems like low maintenance

zirain:
> Yeah keeping them in sync and using the same CI tooling seems like low maintenance

sounds good

arkodg:
hey @shawnh2 hoping you can help with this one, imo the first 2 adds on in here could be `prometheus` and `grafana` and the dashboard assets to make data plane and control obs easier

shawnh2:
Will add the rest dependencies for the addons-helm as a follow-up PR, once this addons-helm is all set. Docs and Makefiles shall be updated.

PR #3470, opened 2024-05-24
helm: add envoy gateway addon helm chart support

shawnh2:
**What type of PR is this?**

ref https://github.com/envoyproxy/gateway/issues/3355


charts/dashboard-addon/dashboards/envoy-clusters.json
shawnh2, line 1: do we need to remove https://github.com/envoyproxy/gateway/tree/main/examples/grafana/dashboards ?

charts/dashboard-addon/Chart.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: v2
+name: dashboard-addon
```
zirain, line 2: gateway-addon would be better?

arkodg: +1 to `envoy gateway add-ons`

shawnh2: SGTM

charts/dashboard-addon/Chart.lock
```diff
@@ -0,0 +1,12 @@
+dependencies:
+- name: prometheus
+  repository: https://prometheus-community.github.io/helm-charts
+  version: 23.1.0
+- name: grafana
+  repository: https://grafana.github.io/helm-charts
+  version: 6.58.4
+- name: envoy-gateway-helm
```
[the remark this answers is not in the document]
shawnh2: of course

shawnh2:
This PR only include DP metrics and Grafana support, will send another PR to include CP metrics as well (https://github.com/envoyproxy/gateway/issues/3474#issuecomment-2132817429)

charts/dashboard-addon/Chart.yaml
shawnh2: changed

charts/dashboard-addon/dashboards/envoy-clusters.json
shawnh2: > do we need to remove https://github.com/envoyproxy/gateway/tree/main/examples/grafana and https://github.com/envoyproxy/gateway/tree/main/examples/prometheus from examples ?

shawnh2: and update the makefile tools?

arkodg: yeah having one source of truth makes a lot of sense, thanks !

charts/gateway-addon/Chart.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: v2
```
arkodg, line 1: the current dir structure is `gateway-helm` , can we rename the dir here to `gateway-add-ons-helm` which would also represent the chart name in the OCI registry.
reg naming, I'll bring this up in the community meeting today

arkodg:
hey @shawnh2 , @zhaohuabing @guydc and I brought this up in the community meeting today and aligned on
* calling this chart `gateway-demo-helm` 
* adding envoy gateway as a sub chart (sorry for the churn here)

shawnh2:
> This PR only include DP metrics and Grafana support, will send another PR to include CP metrics as well ([#3474 (comment)](https://github.com/envoyproxy/gateway/issues/3474#issuecomment-2132817429))

lets include Grafana dashboard for CP as well after https://github.com/envoyproxy/gateway/pull/3506 landed. 

and the doc under https://gateway.envoyproxy.io/latest/tasks/observability/ can be updated with this helm demo support.

shawnh2:
> hey @shawnh2 , @zhaohuabing @guydc and I brought this up in the community meeting today and aligned on
> 
> * calling this chart `gateway-demo-helm`
> * adding envoy gateway as a sub chart (sorry for the churn here)

If we are including `envoy-gateway` helm as well, we may not install this demo chart just using one `helm install`, cause eg and prom/grafana should be installed at different namespace. But helm does not support installing dependencies at different namespace. xref: https://github.com/helm/helm/issues/10905

So:

- Do we accpet installing all the dependencies at the same one namespace?
- There's another tool [helmfile](https://helmfile.readthedocs.io/en/latest/) could solve this problem, but will introduce more operations for users.
- I can understand why we are trying to include eg as a dependency, to provide users with a one-stop eg-with-addons experience. But exclude eg does not affect this, we still can provide one-stop abilities for all the addons. And can solve this problem more easily.

arkodg:
ah then lets go with the previous approach of `gateway-addons-helm` and rm `envoy-gateway` from this chart

zhaohuabing:
Nit: should we name this directory "gateway-addons"? Currently, there are two addons(Prometheus and Grafana), and we may have more in the future.

charts/dashboard-addon/dashboards/envoy-clusters.json
shawnh2: done

charts/gateway-addon/Chart.yaml
shawnh2: rename to `gateway-addons-helm` by final decision

tools/make/helm.mk
```diff
@@ -1,35 +1,44 @@
...
 #
-# All make targets related to helmß are defined in this file.
+# All make targets related to helms are defined in this file.
 
 include tools/make/env.mk
 
 IMAGE_PULL_POLICY ?= IfNotPresent
 OCI_REGISTRY ?= oci://docker.io/envoyproxy
 CHART_NAME ?= gateway-helm
+ADDONS_CHART_NAME ?= gateway-addons-helm
```
arkodg, line 10: can we follow similar pattern as we use for `image` ?
https://github.com/envoyproxy/gateway/blob/51a3a9dd6e71daa568db066b34cd62eae9c69185/tools/make/image.mk#L11
this will allow us to build and package individual package using commands like `make helm-package.gateway-helm` and also use top level commands like `make helm-package` to run for all charts

shawnh2: done

arkodg (approved):
LGTM thanks !

arkodg:
hey @phlax once this PR merges, we'll be pushing a new Helm Chart `envoyproxy/gateway-addons-helm` , will the CI be able to push using the existing PAT or do we need to create the repo before hand ?

zirain (approved):
LGTM, may need wait @phlax first.

guydc (approved):

Issue #3392, opened 2024-05-15
Helm reference site docs not updated when helm values are updated


arkodg:
@arkodg there is already a link to helm chart values:
<img width="620" alt="image" src="https://github.com/envoyproxy/gateway/assets/16555289/9a735fd1-327b-4d50-a1ed-748005192d29">
https://gateway.envoyproxy.io/v1.0.1/install/install-helm/#using-valuesyaml-file-for-complex-installation

However, I noticed that values section in chart [README.md](https://github.com/envoyproxy/gateway/tree/main/charts/gateway-helm#values) is not updated.
The reason is only `site/content/en/latest/install/api.md` is updated when running `make helm-readme-gen`: https://github.com/envoyproxy/gateway/blob/c30d09f284e1ecc0a18a59dbe015987c6a713b1f/tools/make/docs.mk#L38-L42

_Originally posted by @shahar-h in https://github.com/envoyproxy/gateway/issues/3366#issuecomment-2106128670_

yuluo-yx:
@arkodg Can you give me some guidance? My understanding is that I need to add a command to helm-readme-gen to move the readme.md file along with the API.md file when it is generated and moved, is this correct?

```
mv charts/gateway-helm/README.md site/content/en/latest/install/README.md 
```

arkodg:
hey @yuluo-yx thanks for picking this one up, I think there are 2 issues here
1. replacing `mv charts/gateway-helm/api.md site/content/en/latest/install/api.md` with `cp charts/gateway-helm/README.md site/content/en/latest/install/api.md` . I dont think why we have `api.md` @Xunzhuo @zirain do you know it exists
2. triggering docs GHA when `charts/` changes https://github.com/envoyproxy/gateway/blob/main/.github/workflows/docs.yaml

zirain:
https://gateway.envoyproxy.io/latest/install/api/

PR #3492, opened 2024-05-28
docs: use helm-docs template

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3392


shawnh2 (approved):

arkodg (approved):
LGTM thanks !

zirain:
> hey @zirain can we also copy these files for v1.0.1 ?

it will be meaningless if there's no 1.0.2.

PR #3494, opened 2024-05-28
feat: Support connecting to an Envoy Gateway Extension Server over Unix Domain Sockets

liorokman:
Support connecting to an extension server over UDS, allowing for lower latency in the communication between Envoy Gateway and an extension server.


api/v1alpha1/envoygateway_types.go
```diff
@@ -480,10 +480,18 @@ type XDSTranslatorHooks struct {
 	Post []XDSTranslatorHook `json:"post,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="(has(self.Path) && !has(self.Host)) || (has(self.Host) && !has(self.Path))",message="transport configuration must be either tcp or uds"
 // ExtensionService defines the configuration for connecting to a registered extension service.
 type ExtensionService struct {
+	// Path defines the path to the extension service Unix Domain Socket file.
```
arkodg, line 486: should we replace this with `backendRef` ? @guydc @AliceProxy

api/v1alpha1/envoygateway_types.go
```diff
@@ -480,10 +480,18 @@ type XDSTranslatorHooks struct {
...
+	//
+	// +optional
+	Path string `json:"path,omitempty"`
+
 	// Host define the extension service hostname.
-	Host string `json:"host"`
+	//
+	// +optional
```
arkodg, line 493: I'd prefer if we deprecated the current fields and added something that looked like
https://github.com/envoyproxy/gateway/blob/6291a370d5ba965607f1dd2121793ca2e88f9f0a/api/v1alpha1/backend_types.go#L58 e.g.
```
type: Unix
unix:
  path: /path/to/domain.sock
```

```
type: FQDN
fqdn:
  hostname: name.ns.svc.cluster.local
  port: 8080
```

liorokman: Done

api/v1alpha1/envoygateway_types.go
```diff
@@ -483,15 +483,22 @@ type XDSTranslatorHooks struct {
...
 	// Port defines the port the extension service is exposed on.
+	// Deprecated: use the backend attribute instead
 	//
 	// +optional
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:default=80
 	Port int32 `json:"port,omitempty"`
 
+	// Backend points to where the extension server can be found.
+	Backend BackendEndpoint `json:"backend"`
```
arkodg, line 500: should this be an embedded field ? since TLS is on this level

liorokman: Done

guydc (approved):
LGTM!

AliceProxy (approved):
lgtm

PR #3495, opened 2024-05-28
feat(translator): implement backend API

### LINKED ISSUE #2997: Support Unix Domain Sockets for Envoy services

guydc:
Implements backend API 
- Support translation of `fqdn`, `ip` and `unix` endpoints in IR and XDS
- Support H2C upstream application protocol hint
- Support `HTTPRoute` and `EnvoyExtensionPolicy` references to `Backend`
- Extend `BackendTLSPolicy` translation tests for `HTTPRoute` and `Backend` use cases
- Minor refactor of test manifests and translation to increase code reuse

Fixes #2997 
Relates to #36


internal/gatewayapi/route.go
```diff
@@ -1176,6 +1178,17 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
 			Endpoints:   endpoints,
 			AddressType: addrType,
 		}
+		// TODO: support mixed endpointslice address type for the same backendRef
```
arkodg, line 1181: can we move this out of `switch` to make it common amongst all backend types ?

internal/gatewayapi/status/backend.go
```diff
@@ -0,0 +1,42 @@
...
+	return be
+}
+
+// computeBackendAcceptedCondition computes the Backend Accepted status condition.
+func computeBackendAcceptedCondition(be *egv1a1.Backend, accepted bool) metav1.Condition {
+	switch accepted {
+	case true:
+		return newCondition(string(egv1a1.BackendReasonInvalid), metav1.ConditionTrue,
+			string(egv1a1.BackendConditionAccepted),
+			"The Backend was accepted Envoy Gateway", time.Now(), be.Generation)
```
arkodg, line 36:
```suggestion
			"The Backend was accepted by Envoy Gateway", time.Now(), be.Generation)
```
```suggestion
			"The Backend was accepted Envoy Gateway", time.Now(), be.Generation)
```

arkodg, line 36: do we need the  "Envoy Gateway" suffix?

internal/provider/kubernetes/controller.go
```diff
@@ -1181,6 +1219,28 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 		return err
 	}
 
+	// Watch Backend CRUDs and process affected *Route objects.
+	if r.envoyGateway.ExtensionAPIs != nil && r.envoyGateway.ExtensionAPIs.EnableBackend {
+		backendPredicates := []predicate.TypedPredicate[*egv1a1.Backend]{
+			predicate.NewTypedPredicateFuncs[*egv1a1.Backend](func(be *egv1a1.Backend) bool {
```
arkodg, line 1225: ObservedGeneration predicate ? we haven't added the `ObjectMeta` yet so it won't work
the issue I see here is constant reconciling - 
* reconcile backend
* translate backend and update status
* translate again (may be stopped by watchable layer which checks for equality)

internal/provider/kubernetes/controller.go
```diff
@@ -387,25 +392,40 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
...
-		if err := r.client.List(ctx, endpointSliceList, opts...); err != nil {
-			r.log.Error(err, "failed to get EndpointSlices", "namespace", string(*backendRef.Namespace),
-				backendRefKind, string(backendRef.Name))
-		} else {
-			for _, endpointSlice := range endpointSliceList.Items {
-				endpointSlice := endpointSlice
-				r.log.Info("added EndpointSlice to resource tree", "namespace", endpointSlice.Namespace,
-					"name", endpointSlice.Name)
-				gwcResource.EndpointSlices = append(gwcResource.EndpointSlices, &endpointSlice)
+		if endpointSliceLabelKey != "" {
```
arkodg, line 411: lets move this inside the Service case ?

internal/provider/kubernetes/controller.go
arkodg: https://github.com/envoyproxy/gateway/blob/44330ef0ea6b0d6725e2cb40ab7b7837dccd1fb7/internal/gatewayapi/resource.go#L172

internal/provider/kubernetes/controller.go
guydc: It also applies to `ServiceImports`. I can make it a bit more explicit by checking the `Kind` as well as the label's existence.

internal/provider/kubernetes/controller.go
guydc: Can you elaborate on why `TypedGenerationChangedPredicate` won't work here?

internal/provider/kubernetes/controller.go
arkodg: ah lets then leave it as is

internal/provider/kubernetes/controller.go
arkodg: nvm it will, took a look at the struct https://github.com/envoyproxy/gateway/blob/6f30397e23ed06218896b7e5a0a8dad3368775f3/api/v1alpha1/backend_types.go#L43
and you've added `ObjectMeta`,  so should work, lets add it

internal/gatewayapi/route.go
guydc: The conditions is slightly different for `Backend` vs. the native K8s resources (depending on EndpointRouting toggle) and so is the message. I'll split the validation and status update.

arkodg (approved):
LGTM thanks for implementing this and making it easier to route to non k8s endpoints !

can we document the remaining items (disallowing `localhost` , other xRoute backend) as subtasks ?

arkodg (approved):

internal/ir/xds.go
```diff
@@ -1075,21 +1077,33 @@ type DestinationEndpoint struct {
...
+	if d.Path != nil {
+		if d.Port != 0 {
+			errs = errors.Join(errs, ErrDestEndpointUDSPortInvalid)
+		}
+		if d.Host != "" {
+			errs = errors.Join(errs, ErrDestEndpointUDSHostInvalid)
+		}
+	} else { // IP or FQDN
+		err := validation.IsDNS1123Subdomain(d.Host)
+		_, pErr := netip.ParseAddr(d.Host)
```
liorokman, line 1098: Even though the field is called "ipv4", I think IPv6 addresses would still work here. `netip.ParseAddr` has no issue with IPv6 addresses.

internal/gatewayapi/validate.go
```diff
@@ -260,6 +268,80 @@ func (t *Translator) validateBackendServiceImport(backendRef *gwapiv1a2.BackendR
...
+
+	var unsupportedRef bool
+
+	for _, bep := range backend.Spec.Endpoints {
+		switch {
+		case bep.Unix != nil:
+			unsupportedRef = true
+		case bep.FQDN != nil && bep.FQDN.Hostname == "localhost":
+			unsupportedRef = true
+		case bep.IPv4 != nil && bep.IPv4.Address == "127.0.0.1":
```
liorokman, line 325: Even though the field is called `IPv4`, there's nothing that really prevents IPv6 addresses from being specified here as far as I can tell.

Consider also rejecting the IPv6 localhost address ? `::1` ?

internal/gatewayapi/validate.go
```diff
@@ -260,6 +268,80 @@ func (t *Translator) validateBackendServiceImport(backendRef *gwapiv1a2.BackendR
...
+		return false
+	}
+
+	var unsupportedRef bool
+
+	for _, bep := range backend.Spec.Endpoints {
+		switch {
+		case bep.Unix != nil:
+			unsupportedRef = true
+		case bep.FQDN != nil && bep.FQDN.Hostname == "localhost":
```
liorokman, line 323: If the intent is to prevent localhost from being the target, in case of an FQDN the better approach would be to try to resolve the hostname and see that it doesn't resolve to the localhost.

liorokman (commented):
Looking at this carefully again, I think maybe it would make sense to rename the `IPv4` field in `BackendEndpoint` to `IP` since using IPv6 addresses would probably already just work.

internal/gatewayapi/validate.go
liorokman: What would happen if somebody were to use an IP in the `FQDN` field of `BackendEndpoint`? I think it would just work, except that it would be possible to use `127.0.0.1` here.

internal/gatewayapi/validate.go
liorokman: Also, localhost on Linux systems is not just `127.0.0.1`. It's `127.0.0.1/8`, which means that this check is wrong.

The correct way to check this would be to check if the provided address is contained in `127.0.0.0/8`, not to string-compare to `127.0.0.1`.

guydc: yea, good point, we just discussed this in slack. I'll update the validation to comply with the K8s one for EP Slices that checks for the entire range.

internal/gatewayapi/validate.go
guydc: > If the intent is to prevent localhost from being the target, in case of an FQDN the better approach would be to try to resolve the hostname and see that it doesn't resolve to the localhost. 

Users can change the DNS record this after the resource is accepted. Then, it becomes a race with the next translation until the config is rejected. So, it's still possible that there's a time window where localhost routing is possible. 

We touched this point here: https://github.com/envoyproxy/gateway/pull/3063#discussion_r1597285050. I'm not sure that there really is a good solution for that. The compromise is to enforce validations equivalent to what K8s allows in EP slices.

internal/ir/xds.go
guydc: This is enforced with CEL at the moment: 
https://github.com/envoyproxy/gateway/blob/7f2038fdba662c2de9a3fb4139800c24ecc875b6/api/v1alpha1/backend_types.go#L84

internal/gatewayapi/validate.go
guydc: > What would happen if somebody were to use an IP in the FQDN field of BackendEndpoint? 

There's a CEL validation that will probably reject IPv6 Addresses (as ":" is nto supported). But, IPv4 addresses are actually accepted by this validation. I'll add additional programatic validations.  
https://github.com/envoyproxy/gateway/blob/7f2038fdba662c2de9a3fb4139800c24ecc875b6/api/v1alpha1/backend_types.go#L103

internal/provider/kubernetes/controller.go
shawnh2, line 1: a non-blocking comment: https://github.com/envoyproxy/gateway/blob/eaa685d5174918c6093d08728b1014e5c7be8b52/internal/provider/kubernetes/status_updater.go#L196 and https://github.com/envoyproxy/gateway/blob/eaa685d5174918c6093d08728b1014e5c7be8b52/internal/provider/kubernetes/status_updater.go#L311 can also need a update.

internal/gatewayapi/route.go
```diff
@@ -1491,3 +1524,65 @@ func getTargetBackendReference(backendRef gwapiv1a2.BackendObjectReference) gwap
...
+		}
+	}
+
+	if len(addrTypeMap) > 0 && dstAddrType == nil {
+		dstAddrType = ptr.To(ir.MIXED)
+	}
+
+	for _, ap := range backend.Spec.AppProtocols {
+		if ap == v1alpha1.AppProtocolTypeH2C {
+			dstProtocol = ir.HTTP2
```
shawnh2, line 1579: need a break here?

internal/gatewayapi/route.go
```diff
@@ -1191,24 +1193,55 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
...
 			gwapiv1a2.RouteReasonResolvedRefs,
-			"Mixed endpointslice address type for the same backendRef is not supported")
+			valMsg)
 	}
 
 	ds.Weight = &weight
 	return ds
 }
 
+func validateDestinationSettings(destinationSettings *ir.DestinationSetting, endpointRoutingDisabled bool, kind *gwapiv1.Kind) (bool, string) {
```
shawnh2, line 1229: instead of `string`, can be return a `error` here? can use `error.Error()` to retrieve the string of one error later.

guydc:
IP:
-	API Changed from IPv4/IPv6 =>  IP for future use, as suggested
-	IPv6 still not supported and rejected in validation (CEL, Controller). 
-	Validate that target is not in the entire loopback range, for both xRoute and non-xRoute

FQDN:
-	Fix CEL validation to not support wildcards
-	Add controller validation that FQDN is not an IP address and has at least two segments, for both xRoute and non-xRoute

zhaohuabing:
This one is huge :-). It would be easier to review and quicker to merge if it could be split into multiple PRs.

zhaohuabing:
> This one is huge :-)
> 
> It would be easier to review and quicker to merge if it were split into multiple PRs.

I don’t mean this one, as it’s already under review, but in the future, it might be helpful to split large PRs into multiple smaller ones for easier and quicker reviews.

internal/gatewayapi/status/backend.go
```diff
@@ -0,0 +1,42 @@
...
+
+// UpdateBackendStatusAcceptedCondition updates the status condition for the provided Backend based on the accepted state.
+func UpdateBackendStatusAcceptedCondition(be *egv1a1.Backend, accepted bool, msg string) *egv1a1.Backend {
+	be.Status.Conditions = MergeConditions(be.Status.Conditions, computeBackendAcceptedCondition(be, accepted, msg))
+	return be
+}
+
+// computeBackendAcceptedCondition computes the Backend Accepted status condition.
+func computeBackendAcceptedCondition(be *egv1a1.Backend, accepted bool, msg string) metav1.Condition {
+	switch accepted {
```
shawnh2, line 32: why not use the simple `if..else` for this boolean ?

liorokman (approved):

arkodg (approved):

PR #3469, opened 2024-05-24
ext_auth support backendRefs

zirain:
xref: https://github.com/envoyproxy/gateway/issues/3091

PR #3497, opened 2024-05-29
fix api nit

zirain:
xref: https://github.com/envoyproxy/gateway/pull/3469#discussion_r1618090041


shawnh2 (approved):
thanks for the fix

zhaohuabing (approved):

Issue #3496, opened 2024-05-29
Secret / configmap updates do not trigger a reconcile


alexwo:
*Description*:
Secret and configmap updates do not trigger a reconcile.
This can impact certificate changes and certain rotation procedures, config updates.

*Repro steps*:
Since secrets don’t have a spec or generation field, this predicate causes us to ignore any updates to secret keys. Completely replacing the secret will trigger reconciliation, but if someone only updates the secret keys, those changes will be skipped.

https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/controller.go#L1209

alexwo:
related -> https://github.com/envoyproxy/gateway/issues/3500

shawnh2:
Fixed in favor of #3499

PR #3499, opened 2024-05-29
Fix:  secrets/config map updates do not trigger reconciles

alexwo:
ensure secret/config map changes trigger a reconcile

**What type of PR is this?**
<!--
Remove TypedGenerationChangedPredicate from configmaps / secrets as those don't have spec nor generation, they will not trigger a reconcile even if they change.

Ensure that update to secret and configmas will tigger a reconcile, currently updates are ignored.

Fixes # https://github.com/envoyproxy/gateway/issues/3496

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1228,16 +1228,12 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	}
 	if err := c.Watch(
 		source.Kind(mgr.GetCache(), &corev1.Secret{},
-			handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, s *corev1.Secret) []reconcile.Request {
-				return r.enqueueClass(ctx, s)
-			}),
 			secretPredicates...)); err != nil {
 		return err
 	}
 
 	// Watch ConfigMap CRUDs and process affected ClienTraffiPolicies and BackendTLSPolicies.
 	configMapPredicates := []predicate.TypedPredicate[*corev1.ConfigMap]{
-		predicate.TypedGenerationChangedPredicate[*corev1.ConfigMap]{},
 		predicate.NewTypedPredicateFuncs[*corev1.ConfigMap](func(cm *corev1.ConfigMap) bool {
 			return r.validateConfigMapForReconcile(cm)
 		}),


zirain:
https://github.com/envoyproxy/gateway/blob/72c0cc757c7c4703dfc14345dbd995240af663d7/internal/provider/kubernetes/controller.go#L1095

same exists in v1.0.0?

liorokman (approved):
LGTM

alexwo:
> https://github.com/envoyproxy/gateway/blob/72c0cc757c7c4703dfc14345dbd995240af663d7/internal/provider/kubernetes/controller.go#L1095
> 
> same exists in v1.0.0?



> https://github.com/envoyproxy/gateway/blob/72c0cc757c7c4703dfc14345dbd995240af663d7/internal/provider/kubernetes/controller.go#L1095
> 
> same exists in v1.0.0?

Yes, seems to be about 7 month old.

zirain:
Can you add a test?

alexwo:
> Can you add a test?

yes

> Can you add a test?

yes! :)

alexwo:
> Can you add a test?

Do you think we should add a component test?
I have a larger change that will also ensure we can perform secret rotations for backend mTLS. 
If this makes sense, we can combine this fix with my larger change, which will also include an end-to-end (e2e) test to cover secret updates.

Let me know if this approach makes sense, or if we should include a separate e2e test specifically for this?

zirain:
I'm talking something like `TestValidateServiceForReconcile`

arkodg:
looking at https://pkg.go.dev/k8s.io/api/core/v1#Secret , it contains `ObjectMeta` which contains the `Generation` field
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta , does k8s not bump this val when the secret changes ?

alexwo:
> looking at https://pkg.go.dev/k8s.io/api/core/v1#Secret , it contains `ObjectMeta` which contains the `Generation` field https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta , does k8s not bump this val when the secret changes ?


The generation field is typically used in resources with a spec section to track changes to the spec. Each time the spec of a resource changes, the generation is incremented by the Kubernetes API server. 

For resources like Secrets and ConfigMaps, changes are made directly to the data, and these resources do not have a spec field or a generation field. Therefore, updates to Secrets and ConfigMaps do not automatically trigger reconciliations in controllers that rely on generation changes.

arkodg:
thanks for the explaination @alexwo !

arkodg (approved):

quyenhoang96:
Hi @arkodg. I'm happy that Envoy gateway supports extauth integration in a simple and fast way. I am using extauth feature for authentication. But it is not working with a service with a different namespace than the extauth service. When do you plan to release this cherrypick/release-v1.0.2 version?

guydc (approved):

PR #3503, opened 2024-05-30
chore: improve extension server unit tests

liorokman:
When running unit tests that touch the Extension Server, go through the gRPC code and the regular extension manager 
instead of creating a testing-only variant. 

This PR enables using an in-memory gRPC server implementation as the extension server, instead of bypassing the manager code during tests.

This improves coverage and removes testing code from the built binary.


guydc (approved):
LGTM

shawnh2 (approved):

PR #3507, opened 2024-05-30
feat: backend TLS SAN validation

guydc:
According to BTLS docs, the gateway should ensure that the presented certificate matches the `Hostname` value:

> Hostname defines the server name indication (SNI) the Gateway should use in order to connect to the backend, and **must match the certificate served by the backend pod**.

[optional *Relevant Links*:]
- https://gateway-api.sigs.k8s.io/api-types/backendtlspolicy/?h=backendtls#hostname


arkodg (approved):
LGTM thanks !

shawnh2 (approved):

PR #3418, opened 2024-05-17
feat: add generate API and Helm doc process with zh content update

wilsonwu:
**What type of PR is this?**
The API docs don't need to translate, so add generate API process with zh content update:
1. add a copy cmd to zh content after API generation both for api and helm doc.
2. update the `docs-api-headings` support zh content update

Not sure it is OK or not, please help to review.

PR #3513, opened 2024-06-01
fix: merge race between #3418 and #3444,#3423

shawnh2:
**What type of PR is this?**

Fixes #


arkodg (approved):
Thanks Shawn !

PR #3515, opened 2024-06-02
e2e: eg install and uninstall test

### LINKED ISSUE #3323: e2e: add `egctl install/uninstall`  test
*Description*:
> Write corresponding e2e tests for `egctl x install` and `egctl x uninstall` for verification

ShyunnY:
**What type of PR is this?**
e2e: eg install and uninstall test

This PR adds e2e tests for the parts of `Helm.PackageTool` used in `egctl`.

1. I set both install/uninstall to the same ConformanceTest. This avoids duplicating the install/uninstall process for separate tests
2. Minor refactoring of the upgrade part, which I think belongs to the `package manager`  feature.

Fixes #3323

### DIFF
--- a/test/config/gatewayclass.yaml
+++ b/test/config/gatewayclass.yaml
@@ -165,4 +165,38 @@ metadata:
 spec:
   controllerName: gateway.envoyproxy.io/gatewayclass-controller
 ---
+kind: GatewayClass
+apiVersion: gateway.networking.k8s.io/v1
+metadata:
+  name: lifecycle
+spec:
+  controllerName: gateway.envoyproxy.io/gatewayclass-controller
+  parametersRef:
+    group: gateway.envoyproxy.io
+    kind: EnvoyProxy
+    name: lifecycle-config
+    namespace: envoy-gateway-system
+---
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: EnvoyProxy
+metadata:
+  name: lifecycle-config
+  namespace: envoy-gateway-system
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyDeployment:
+        replicas: 2
+        patch:
+          type: StrategicMerge
+          value:
+            spec:
+              template:
+                spec:
+                  containers:
+                    - name: envoy
+                      readinessProbe:
+                        initialDelaySeconds: 5
 
+---
--- a/test/e2e/lifecycle/eg_lifecycle_test.go
+++ b/test/e2e/lifecycle/eg_lifecycle_test.go
@@ -1,0 +1,67 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+//go:build e2e
+// +build e2e
+
+package lifecycle
+
+import (
+	"flag"
+	"io/fs"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/gateway-api/conformance/utils/flags"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+	"sigs.k8s.io/gateway-api/pkg/features"
+
+	"github.com/envoyproxy/gateway/test/e2e"
+	"github.com/envoyproxy/gateway/test/e2e/tests"
+)
+
+func TestEGLifecycle(t *testing.T) {
+	flag.Parse()
+
+	cfg, err := config.GetConfig()
+	require.NoError(t, err)
+
+	c, err := client.New(cfg, client.Options{})
+	require.NoError(t, err)
+
+	if flags.RunTest != nil && *flags.RunTest != "" {
+		t.Logf("Running E2E test %s with %s GatewayClass\n cleanup: %t\n debug: %t",
+			*flags.RunTest, *flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+	} else {
+		t.Logf("Running E2E tests with %s GatewayClass\n cleanup: %t\n debug: %t",
+			*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug)
+	}
+
+	cSuite, err := suite.NewConformanceTestSuite(suite.ConformanceOptions{
+		Client:               c,
+		GatewayClassName:     *flags.GatewayClassName,
+		Debug:                *flags.ShowDebug,
+		CleanupBaseResources: *flags.CleanupBaseResources,
+		ManifestFS:           []fs.FS{e2e.Manifests},
+		RunTest:              *flags.RunTest,
+		// SupportedFeatures cannot be empty, so we set it to SupportGateway
+		// All e2e tests should leave Features empty.
+		SupportedFeatures: sets.New[features.SupportedFeature](features.SupportGateway),
+	})
+	if err != nil {
+		t.Fatalf("Failed to create test suite: %v", err)
+	}
+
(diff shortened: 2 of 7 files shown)


test/e2e/tests/eg_uninstall_install.go
```diff
@@ -0,0 +1,107 @@
...
+func init() {
+	PackageTests = append(PackageTests, EGUninstallAndInstallTest)
+}
+
+var EGUninstallAndInstallTest = suite.ConformanceTest{
+	ShortName:   "EGUninstallAndInstall",
+	Description: "Uninstall and Install Envoy Gateway using Helm Package Tool",
+	Manifests:   []string{"testdata/eg-uninstall-install.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		t.Run("Uninstall and Install Envoy Gateway should succeed", func(t *testing.T) {
```
shawnh2, line 35: can we split this into two sub test run, one is uninstall and the other is install ?

test/e2e/tests/tests.go
```diff
@@ -12,6 +12,6 @@ import "sigs.k8s.io/gateway-api/conformance/utils/suite"
 
 var (
 	ConformanceTests   []suite.ConformanceTest
-	UpgradeTests       []suite.ConformanceTest
+	PackageTests       []suite.ConformanceTest
```
shawnh2, line 15: let's stick to `PackageManageTests`

test/e2e/tests/eg_uninstall_install.go
ShyunnY: Let's imagine that if we use two sub-tests:
1. Install process: you need to uninstall the eg resource first, then install the eg resource, and finally test whether it is installed normally
2. Uninstall process: uninstall the eg resource directly. (How do we test the success of uninstallation?)

We can see that if we use two sub-tests, most of the processes are overlapped. I think we have more benefits by using a combined test

shawnh2: What I means here is to split the whole test case into multiple `test.Run`, like first `t.Run("check svc xxx")`, then `t.Run("uninstall xxx")`, `t.Run("install xxx")`, not create a new test case for install/uninstall.

In this way, we can have a good understanding of the procedures in this test. It's technical feasible I think.

ShyunnY: Done!

test/e2e/tests/tests.go
ShyunnY: Done!

test/config/gatewayclass.yaml
```diff
@@ -72,19 +72,19 @@ spec:
 kind: GatewayClass
 apiVersion: gateway.networking.k8s.io/v1
 metadata:
-  name: upgrade
+  name: package
```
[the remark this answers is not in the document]
ShyunnY: Imagine: install, uninstall, upgrade... These operations seem to be related to eg **package**!

So I think we can name all these operations as package. If you have a better name, we can migrate

guydc:
maybe call use the term `lifecycle` instead of package/upgrade? It's slightly broader.

ShyunnY:
> maybe use the term `lifecycle` instead of package/upgrade? It's slightly broader.

sounds good!

guydc:
Hi @ShyunnY ! should we also update the existing "upgrade" test to use the new helm util?

ShyunnY:
> Hi @ShyunnY ! should we also update the existing "upgrade" test to use the new helm util?

@guydc 
I'm not sure if what you mean by _"update the existing "upgrade" test to use the new helm util"_ is: we should modify the logic in e2e/upgrade.go to use the tool in `internal/utils/helm/package.go` for the upgrade.
Hopefully my understanding is correct  :)

ShyunnY:
@guydc 

I thought about it, let's solve the current PR first. In the future I will open an issue: about adding `egctl` upgrade feature, I will enhance helm util and let `upgrade e2e` use helm util.
What do you think?

guydc:
yes, sure, let's get this in first.

test/config/gatewayclass.yaml
```diff
@@ -86,19 +86,19 @@ spec:
...
```
arkodg, line 89: shouldnt this be `lifecyle` as well ?

ShyunnY: > shouldnt this be `lifecyle` as well ?

This worries me a bit: I think `lifecycle` doesn't seem to have good prominence.
On the contrary I would think that `package lifecycle` might be better?

test/e2e/tests/eg_lifecycle.go
```diff
@@ -19,13 +19,138 @@ import (
...
+			if lastVersionTag == "" {
+				lastVersionTag = "v0.0.0-latest" // Default version tag if not specified
+			}
+
+			ht := helm.NewPackageTool()
+			if err := ht.Setup(); err != nil {
+				t.Errorf("failed to setup of packageTool: %v", err)
+			}
+
+			t.Log("start install envoy gateway resource")
```
guydc, line 125: I think that there's also some value if the installation will install the local version in the branch (from source), instead of v0.0.0-latest. This way, we're validating the contents of the PR/CI branch, instead of the installability of a previously-released artifact. Plus, in its current state, the test may impact other tests in this suite, by changing the actual EG version deployed (from local branch to v0.0.0-latest). 

@arkodg, @ShyunnY  - WDYT? Should we install from sources or atleast cleanup at the end of the test and revert to the local branch version?

ShyunnY: @guydc  
This is a good suggestion, but I don't think it's necessary. We are focusing on `helmtool` conformance testing, and I don't think we should add complexity by verifying PR/CI here.

In addition, we specify the order of execution in the Makefile, will this still affect other test suites?

guydc:
> In addition, we specify the order of execution in the Makefile, will this still affect other test suites?

It will not impact other suites, but may impact other tests within this suite.

ShyunnY:
> > In addition, we specify the order of execution in the Makefile, will this still affect other test suites?
> 
> It will not impact other suites, but may impact other tests within this suite.

can you give an example?  Thanks!
Or can we specify the order in which the suites are executed?

guydc:
> > > In addition, we specify the order of execution in the Makefile, will this still affect other test suites?
> > 
> > 
> > It will not impact other suites, but may impact other tests within this suite.
> 
> can you give an example? Thanks! Or can we specify the order in which the suites are executed?

One example that comes to mind:
- Shutdown Test: if Uninstall/Install test executes before shutdown test, shutdown test will run on `v0.0.0-latest` instead of `main`. A regression in the shutdown manager component may go undiscovered in the PR, as `main` is not deployed during that test.  


I think that we can just seperate this test into another suite and ensure that it runs after everything else in the `run-e2e` make target.

test/e2e/lifecycle/eg_lifecycle_test.go
```diff
@@ -63,11 +63,30 @@ func TestEGUpgrade(t *testing.T) {
...
-	err = cSuite.Run(t, tests.UpgradeTests)
+	err = cSuite.Run(t, tests.LifecycleTests)
 	if err != nil {
 		t.Fatalf("Failed to run tests: %v", err)
 	}
 }
+
+// we reorder lifecycle test, and we want 'Envoy Shutdown' to be the first test to
+// run to avoid being affected.
+func reorderLifeCycleTest(t *[]suite.ConformanceTest) {
```
guydc, line 78: Maybe we can just register all the tests in the suite setup instead of allowing them to register themselves in arbitrary order?

ShyunnY:
It seems like we've been procrastinating for a long time, can we please get on with this work? :)

shawnh2 (dismissed):
LGTM! Probably need to wait https://github.com/envoyproxy/gateway/pull/3764 gets in first.

ShyunnY:
I'm sorry for delaying the resolution of this conflict.

In the latest commit, I thought to simplify the `lifecycle` process as much as possible. So I executed it as a separate E2E Test and scheduled it to be executed in the last task. Avoid affecting the rest of the E2E.

arkodg:

PR #3521, opened 2024-06-03
feat: allow to explicitly disable helm rendering certgen resources

### LINKED ISSUE #3501: helm: support disabling cert-gen job
Good day!

I see that inside of helm chart we are using `kind: Job` to generate certificates. I'd like to ask to support disabling of this job as I can issue these certificates with cert-manager like I proposed in another project https://github.com/wazuh/wazuh-kubernetes/pull/577

ShyunnY:
**What type of PR is this?**
feat: allow to explicitly disable helm rendering certgen resources

Fixes #3501

### DIFF
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -1,3 +1,4 @@
+{{- if ne (toString .Values.certgen) "nil" }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -62,3 +63,4 @@ subjects:
 - kind: ServiceAccount
   name: '{{ include "eg.fullname" . }}-certgen'
   namespace: '{{ .Release.Namespace }}'
+{{- end }}
--- a/charts/gateway-helm/templates/certgen.yaml
+++ b/charts/gateway-helm/templates/certgen.yaml
@@ -1,3 +1,4 @@
+{{- if ne (toString .Values.certgen) "nil" }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -48,3 +49,4 @@ spec:
   {{- if not ( kindIs "invalid" .Values.certgen.job.ttlSecondsAfterFinished) }}
   ttlSecondsAfterFinished: {{ .Values.certgen.job.ttlSecondsAfterFinished }}
   {{- end }}
+{{- end }}
--- a/site/content/en/latest/install/install-helm.md
+++ b/site/content/en/latest/install/install-helm.md
@@ -118,6 +118,8 @@ You can use the below command to install the envoy gateway using values.yaml fil
 helm install eg oci://docker.io/envoyproxy/gateway-helm --version v0.0.0-latest -n envoy-gateway-system --create-namespace -f values.yaml
 ```
 
+> _NOTE: If you want to use custom certificates instead of the default generated certificates, you can also set `certgen: nil` to turn off the certificate generation Job._
+
 ## Open Ports
 
 These are the ports used by Envoy Gateway and the managed Envoy Proxy.
--- a/test/helm/certgen-disable.in.yaml
+++ b/test/helm/certgen-disable.in.yaml
@@ -1,0 +1,2 @@
+# We explicitly set the certgen to nil to close certgen Job
+certgen: nil
--- a/test/helm/certgen-disable.out.yaml
+++ b/test/helm/certgen-disable.out.yaml
@@ -1,0 +1,436 @@
+---
+# Source: gateway-helm/templates/envoy-gateway-deployment.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: envoy-gateway
+  namespace: 'envoy-gateway-system'
+  labels:
+    helm.sh/chart: gateway-helm-v0.0.0-latest
+    app.kubernetes.io/name: gateway-helm
+    app.kubernetes.io/instance: eg
+    app.kubernetes.io/version: "latest"
+    app.kubernetes.io/managed-by: Helm
+---
+# Source: gateway-helm/templates/envoy-gateway-config.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: envoy-gateway-config
+  namespace: 'envoy-gateway-system'
+  labels:
+    helm.sh/chart: gateway-helm-v0.0.0-latest
+    app.kubernetes.io/name: gateway-helm
+    app.kubernetes.io/instance: eg
+    app.kubernetes.io/version: "latest"
+    app.kubernetes.io/managed-by: Helm
+data:
+  envoy-gateway.yaml: |
+    apiVersion: gateway.envoyproxy.io/v1alpha1
+    kind: EnvoyGateway
+    gateway:
+      controllerName: gateway.envoyproxy.io/gatewayclass-controller
+    logging:
+      level:
+        default: info
+    provider:
+      kubernetes:
+        rateLimitDeployment:
+          container:
+            image: docker.io/envoyproxy/ratelimit:master
+          patch:
+            type: StrategicMerge
+            value:
+              spec:
+                template:
+                  spec:
+                    containers:
+                    - imagePullPolicy: IfNotPresent
+                      name: envoy-ratelimit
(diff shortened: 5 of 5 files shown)


charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if ne (toString .Values.certgen) "nil" }}
```
arkodg, line 1: shouldnt this be a `nil` `{{ if .Values.certgen }}` check rather than a string eq check ?

arkodg: https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

zhaohuabing:
@ShyunnY we need to move `hmacSecretKey` out of the certgen if we want to disable it.

zhaohuabing (commented):
@ShyunnY we need to move hmacSecretKey out of the certgen if we want to disable it.

ShyunnY:
> @ShyunnY we need to move hmacSecretKey out of the certgen if we want to disable it in helm template.

@zhaohuabing 
Sorry, I don't quite understand what you mean by hmacSecretKey. Could you please mark it out? Thank you very much  :)

charts/gateway-helm/templates/certgen-rbac.yaml
ShyunnY: When we explicitly set `certgen: nil`, yaml does not consider this to be empty, so `{{ if .Values.certgen }}` will be executed incorrectly.

ShyunnY: Initially I also used the empty check, but it always failed. After many attempts, maybe helm treats yaml as a string when parsing it?

We can make a simple attempt:
```yaml
# values.yaml
certgen: nil

# template
{{- if eq .Values.certgen nil }}
msg: empty
{{- else }}
msg: no empty
{{- end }}
```
When we execute the `template` command, this will happen:
```shell
msg: no empty
```

Let's change another example
```yaml
# values.yaml
certgen: nil

# template
{{- if .Values.certgen }}
msg: no empty
{{- else }}
msg: empty
{{- end }}
```
After executing the `template` command:
```shell
msg: no empty
```

shawnh2: changing `certgen:` to `None` or `{}` shall work

shawnh2: for the context here, is `certgen` better to be a boolean ?

zhaohuabing:
> > @ShyunnY we need to move hmacSecretKey out of the certgen if we want to disable it in helm template.
> 
> @zhaohuabing Sorry, I don't quite understand what you mean by hmacSecretKey. Could you please mark it out? Thank you very much :)

@ShyunnY We can't disable certgen job since it also generates the hmac secret for OIDC.

https://github.com/envoyproxy/gateway/blob/371265969fe8b4246de3b78290a44d08ba8cb20a/internal/crypto/certgen.go#L157

ShyunnY:
> > > @ShyunnY we need to move hmacSecretKey out of the certgen if we want to disable it in helm template.
> > 
> > 
> > @zhaohuabing Sorry, I don't quite understand what you mean by hmacSecretKey. Could you please mark it out? Thank you very much :)
> 
> @ShyunnY We can't disable certgen job since it also generates the hmac secret for OIDC.
> 
> https://github.com/envoyproxy/gateway/blob/371265969fe8b4246de3b78290a44d08ba8cb20a/internal/crypto/certgen.go#L157

I think I understand.
Is there a better way to solve this problem? Do you have any good ideas?

charts/gateway-helm/templates/certgen-rbac.yaml
ShyunnY: In this context, I think it would be more appropriate to use this way:
```yaml
certgen:
# Add a knob to turn it on or off
enable: true/false
job:
annotations: {}
resources: {}
ttlSecondsAfterFinished: 30
rbac:
annotations: {}
labels: {}
```

zhaohuabing:
I think it's fine to keep this job here, but we should consider giving it a better name to clearly convey that it does more than generate certificates.

charts/gateway-helm/templates/certgen-rbac.yaml
arkodg: what is the most common usage in helm
* enable/disable
* enabled/disabled
* nil

ShyunnY:
> changing name will be a breaking change, and even if we do change name, whats an apt name ?
> 
> we could add helm doc strings highlighting that this job does 2 things - creates control plane cert and and the hmac secret

It seems that we need to highlight this in the `Install with Helm` and `Control Plane Authentication using custom certs` docs.
When users choose to set `certgen: disabled`, we must inform the corresponding warning message

charts/gateway-helm/templates/certgen-rbac.yaml
ShyunnY: Combined with your comments. I think it is better to design it as follows:
```yaml
# This means disabling the Job
certgen: disable
```

@arkodg

zhaohuabing:
> > changing name will be a breaking change, and even if we do change name, whats an apt name ?
> > we could add helm doc strings highlighting that this job does 2 things - creates control plane cert and and the hmac secret
> 
> It seems that we need to highlight this in the `Install with Helm` and `Control Plane Authentication using custom certs` docs. When users choose to set `certgen: disabled`, we must inform the corresponding warning message

We should leave the certgen job and just add helm docs to explain:
* By default, certgen won't override existing certs generated by cert manager.
* certgen also generates the HMAC for OIDC

ShyunnY:
> > > changing name will be a breaking change, and even if we do change name, whats an apt name ?
> > > we could add helm doc strings highlighting that this job does 2 things - creates control plane cert and and the hmac secret
> > 
> > 
> > It seems that we need to highlight this in the `Install with Helm` and `Control Plane Authentication using custom certs` docs. When users choose to set `certgen: disabled`, we must inform the corresponding warning message
> 
> We should keep the certgen job(won't add a `certgen: disable` option in the helm) and just add docs to explain:
> 
> * By default, certgen won't override existing certs generated by cert manager or other third-party tools.
> * certgen also generates the HMAC for OIDC.
> 
> If certgen is disabled, OIDC will fail due to lacking of HMAC secret.

I may not be very clear about the design of OIDC. Can we let OIDC generate a default random value when HMAC cannot be obtained to solve this problem?
Similar to `func generateHMACSecret() ([]byte, error)`?

zhaohuabing:
The HMAC generation is a one-time thing. The certgen job is the best place for it, or we need a separate job for it.

ShyunnY:
This is a difficult decision.
We would like to provide a `certgen: disable` Job to allow users to use custom certs. Using a new Job to generate HMAC to solve the problem of missing HMAC is a good idea, but it may increase complexity.

I would like to hear your opinions and see if we can find a balance to solve the problem :)

zhaohuabing:
Hi  @ShyunnY I may miss something here, but AFAIK, the certgen job won't prevent users from using their own certs because certgen won't override these certs if they already exist.

ShyunnY:
I may not understand it very well, so what should I do next?
@zhaohuabing

arkodg:
hey @ShyunnY thanks for working on this one, after bringing this up in the community meeting today, we decided to not support disabling this job, to avoid any issues with the OIDC feature.

ShyunnY:
> 嘿@ShyunnY感谢您对此所做的努力，在今天的社区会议上提出这个问题后，我们决定不支持禁用这项工作，以避免 OIDC 功能出现任何问题。

@arkodg  That's okay! It's great to hear that the community meeting discussed this and came to a conclusion.  

To sum up, I think we should add a little bit of documentation in [doc](https://gateway.envoyproxy.io/v1.0.1/install/custom-cert/): Highlight what certgen does, inform users to avoid manually disabling the `certgen` Job when using custom cert.

arkodg:
should that certgen explanation be moved to helm README instead https://gateway.envoyproxy.io/v1.0.1/install/api/ ?

ShyunnY:
Thanks everyone for the comments, I will close this PR

PR #3522, opened 2024-06-03
feat(translator): Support extension server hooks for TCP and UDP listeners in addition to HTTP listeners.

### LINKED ISSUE #2673: Add a post translation hook for non-HTTP listener
*Description*:

In the current extension design there is a `PostHTTPListenerModify` RPC for HTTP listener and a `PostTranslateModify` RPC for cluster and secret resources. We can not touch non-HTTP listeners until some policy is implemented in EG. Envoy has so many configuration that EG may not expose all the APIs. So adding non-HTTP listener to a post translation hook helps a lot. 

One usage is to support transparent mode TCP proxy. Users could add Original Src Filter[^1] in the hook and handle the route related work in their own. Another example is to add a custom TLS handshaker[^2] to the listener.

[optional *Relevant Links*:]

[^1]: Original Src Filter: https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/filters/listener/original_src/v3/original_src.proto#extensions-filters-listener-original-src-v3-originalsrc
[^2]: custom_handshaker: https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/transport_sockets/tls/v3/tls.proto.html

liorokman:

This PR calls the extension server "Listener" hook for TCP and UDP listeners. The hook is called only once for each XDS listener.

I kept the hook name as it was before (`PostHTTPListenerModify`) for backwards compatibility reasons.

Fixes #2673


internal/xds/translator/translator.go
```diff
@@ -86,28 +87,34 @@ func (t *Translator) Translate(ir *ir.Xds) (*types.ResourceVersionTable, error)
...
+		tCtx, xdsIR.HTTP, xdsIR.AccessLog, xdsIR.Tracing, xdsIR.Metrics, listenerExtPolicies); err != nil {
+		errs = errors.Join(errs, err)
+	}
+
+	if err := t.processTCPListenerXdsTranslation(tCtx, xdsIR.TCP, xdsIR.AccessLog, xdsIR.Metrics, listenerExtPolicies); err != nil {
 		errs = errors.Join(errs, err)
 	}
 
-	if err := processTCPListenerXdsTranslation(tCtx, ir.TCP, ir.AccessLog, ir.Metrics); err != nil {
+	if err := processUDPListenerXdsTranslation(tCtx, xdsIR.UDP, xdsIR.AccessLog, xdsIR.Metrics, listenerExtPolicies); err != nil {
```
arkodg, line 101: looks like we are passing `listenerExtPolicies` to `processUDPListenerXdsTranslation` so that we can append per listener type's `ExtensionRefs` to it. Can we instead move that out of the pure xds translations into an extension specific function like `notifyExtensionServerAboutListeners` ?

internal/xds/translator/translator.go
```diff
@@ -120,12 +127,60 @@ func (t *Translator) Translate(ir *ir.Xds) (*types.ResourceVersionTable, error)
...
+	}
+
+	var errs error
+	for _, l := range tCtx.XdsResources[resourcev3.ListenerType] {
+		listener := l.(*listenerv3.Listener)
+		pols, found := listenerExtPolicies[listener.Name]
+		if !found {
+			pols = nil
+		} else {
+			slices.SortFunc(pols, func(l, r *ir.UnstructuredRef) int {
```
arkodg, line 153: this looks new, we are trying to ensure we pass in fixed order ?

liorokman: It's not about ensuring that we pass the policies in a fixed order. It's so that we don't duplicate them. The code sorts the list of extension policies and then compacts the list with `slices.CompactFunc` to remove duplicates.

See the [extensionpolicy-tcp-udp-http](https://github.com/envoyproxy/gateway/pull/3522/files#diff-07191b26ca09321c0d10a8cff082064e48c0729b420826c25ea781bb82f00f07) unit test for an example of how this can happen. The TL;DR is that when there are multiple listeners defined in the same gateway resource where the port is the same but one is for HTTP and the other is TCP, then in the translation to xDS the same xDS listener is reused because it's the same port. 

This code section is about making sure that if the same extension server policy is attached to different IR level constructs that map to the same xDS level construct, then the extension server won't receive the same policy twice.

liorokman:
@envoyproxy/gateway-maintainers  PTAL

guydc (approved):
LGTM

internal/gatewayapi/extensionserverpolicy.go
```diff
@@ -201,5 +201,25 @@ func (t *Translator) translateExtServerPolicyForGateway(
 		})
 		found = true
 	}
+	for _, currListener := range gwIR.TCP {
```
arkodg, line 204: should this be flipped to ?
1. find the listeners from the `gateway` arg that are being targeted by this policy
2. find the IR for that listener
3. append ExtensionRefs to it

internal/xds/translator/translator.go
arkodg: ah go it, thoughts on adding a `findIRListenesrByHostPort` method similar to 
https://github.com/envoyproxy/gateway/blob/e866ec138ca1666e55073d980f2d1aaf477f36d9/internal/xds/translator/translator.go#L598 which returns the IR listeners, allowing you to append/send the extensionRefs directly in here

this way we wont need to embed any extension policy code into the listener translator functions, mantain any state in memory, or perform any of these SortFuncs here

internal/gatewayapi/extensionserverpolicy.go
liorokman: WDYM?

This is already how the function works.

`gwIR` is the IR representation for the gateway, and the function goes over all of the possible listeners searching for listeners that are being targeted by this policy. Gateways have listeners in three places (`HTTP`, `TCP`, and `UDP`) so there are three identical loops here.

In each loop the policy is appended if it targets the current listener. If the policy doesn't have a `SectionName` then all the listeners of the gateway are targeted - otherwise the `SectionName` must match the listener's name.

internal/xds/translator/translator.go
liorokman: Changed the logic to work like this. Not sure it's better, computation-wise.

arkodg: thanks, agree, its not better computation wise, but should improve readability and maintenance

internal/gatewayapi/extensionserverpolicy.go
arkodg: I was hoping we could avoid parsing o/p (IR) and instead find the IR, this would avoid steps such as `currListener.Name[strings.LastIndex(currListener.Name, "/")+1:]`, but it looks like we are already doing a similar parse in other places https://github.com/envoyproxy/gateway/blob/8a4b13dd19bd03adcf9f5c98ee249c5da7b17c17/internal/gatewayapi/backendtrafficpolicy.go#L369 but wrapping the `find` in a helper function . I'm going to resolve this comment because this code is following existing patterns

arkodg (dismissed):
LGTM thanks !

arkodg:
hey @liorokman can you rebase, the `chart` generated files have been commited

guydc (approved):
LGTM, thanks everyone!

arkodg (approved):

PR #3524, opened 2024-06-03
feat(translator): Add names to filterchains based on the listener name

### LINKED ISSUE #3461: Set unique name for FilterChain in xDS listener's FilterChains
*Description*:

After #3371 is done, we can attach policies to gateway or specific listener.

When multiple listeners differentiated by SNI are translated into single xDS listener, `PostHTTPListenerModify` will be invoked after every individual listener has been translated to single FilterChain, then extension server will receive the whole xDS listener and policies attaching to the specific listener.

But extension server can't know which FilterChain should be modified based on policies.

One solution is to set name for each FilterChain, name is `irListenerName(listener)` as mentioned in https://github.com/envoyproxy/gateway/issues/3307#issuecomment-2126474603. There will an exception where multiple non-TLS HTTP listeners will be translated to DefaultFilterChain, should be handle carefully.

[optional *Relevant Links*:]
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchain

liorokman:
This PR populates the filterChain name in xDS listeners based on the listener from which a filter chain originated.

Fixes #3461

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -974,6 +974,7 @@ xds:
                       path: /dev/stdout
                   cluster: tcproute/default/backend/rule/-1
                   statPrefix: tcp
+              name: tcproute/default/backend
             name: default/eg/tcp
             perConnectionBufferLimitBytes: 32768
       - activeState:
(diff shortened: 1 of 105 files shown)


arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM

Issue #3500, opened 2024-05-29
fix: update envoyproxy (global) backend mTLS to use the latest secret details upon secret changes


alexwo:
Currently, even if a secret is changed, globally configured mTLS backend settings won’t be refreshed for Envoy Proxy. This needs to be fixed so that if the client certificate referenced by Envoy Proxy is updated, the Envoy Proxy configuration remains in sync and is updated to use the new client certificate.

PR #3526, opened 2024-06-03
fix client backend mtls secrets updates 

alexwo:
**What type of PR is this?**
Ensure that if a secret is globally configured mtls secret is changed, envoy proxy refreshes repectivilty and uses the updated version when connecting to backends. 

Fixes https://github.com/envoyproxy/gateway/issues/3500


arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM

PR #3530, opened 2024-06-04
fix: ReplaceFullPath not working for root path (/)

arkodg:
Takes https://github.com/envoyproxy/gateway/pull/2817 forward


zhaohuabing (commented):
Do we need an e2e test to make sure it works?

arkodg:
> Do we need an e2e test to make sure it works?

hey @zhaohuabing added an e2e, ptal

zhaohuabing (approved):
LGTM thanks!

shawnh2 (approved):
lgtm

PR #3533, opened 2024-06-04
fix: correct gatewayclass deletion behaviour when merge gateways enabled

### LINKED ISSUE #2637: Unexpected Pod while deleting gateways with merge gateways enabled
### *Description*:

There's some unexpected Pod during this process for deleting gateways with `mergeGateways` features on.

### *Repro steps*:

1. Apply the manifest from [quickstart](https://gateway.envoyproxy.io/latest/user/quickstart/) 
2. Apply these config below, to create a GC named `mg` with MergeGateways feature on, and 3 GTWs with 3 HTTPRoutes attached to each GTW

<details><summary>kubectl apply -f mg.yaml</summary>

```yaml
kind: GatewayClass
metadata:
  name: mg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
---
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  mergeGateways: true
---
kind: Gateway
metadata:
  name: merged-eg-1
  namespace: default
spec:
  gatewayClassName: mg
  listeners:
    - allowedRoutes:
        namespaces:
          from: Same
      name: http
      port: 8080
      protocol: HTTP
---
kind: Gateway
metadata:
  name: merged-eg-2
  namespace: default
spec:
  gatewayClassName: mg
  listeners:
    - allowedRoutes:
        namespaces:
          from: Same
      name: http
      port: 8081
      protocol: HTTP
---
kind: Gateway
metadata:
  name: merged-eg-3
  namespace: default
spec:
  gatewayClassName: mg
  listeners:
    - allowedRoutes:
        namespaces:
          from: Same
      name: http
      port: 8082
      protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: hostname1-route
spec:
  parentRefs:
    - name: merged-eg-1
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example
---
kind: HTTPRoute
metadata:
  name: hostname2-route
spec:
  parentRefs:
    - name: merged-eg-2
  hostnames:
    - "www.example2.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example2
---
kind: HTTPRoute
metadata:
  name: hostname3-route
spec:
  parentRefs:
    - name: merged-eg-3
  hostnames:
    - "www.example3.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /example3

```

</details>

4. Everything is fine now

```bash
k get gc                                                             ✭ ✈
NAME            CONTROLLER                                      ACCEPTED   AGE
envoy-gateway   gateway.envoyproxy.io/gatewayclass-controller   True       65m
mg              gateway.envoyproxy.io/gatewayclass-controller   True       2s

k get gtw                                                            ✭ ✈
NAME          CLASS   ADDRESS          PROGRAMMED   AGE
merged-eg-1   mg      172.18.255.200   True         4s
merged-eg-2   mg      172.18.255.200   True         4s
merged-eg-3   mg      172.18.255.200   True         4s

k get po -n envoy-gateway-system                                      ✭ ✈
NAME                                 READY   STATUS    RESTARTS   AGE
envoy-gateway-6b8cdbfcdc-fbcb8       1/1     Running   0          66m  # <- focus on this
envoy-mg-e9949d90-6d8596c878-ff5rb   1/1     Running   0          14s
```

5. `k delete -f mg.yaml`

6. There're some unexpected Pod showing up

```bash
k get po -n envoy-gateway-system                                      ✭ ✈
NAME                                                  READY   STATUS        RESTARTS   AGE
envoy-default-merged-eg-2-c7655c02-dd7c5d5d7-wp62r    0/1     Terminating   0          2s  # <- unexpected
envoy-default-merged-eg-3-2a08f1cd-85dc9dbc6f-6wm7n   0/1     Terminating   0          2s  # <- unexpected
envoy-gateway-6b8cdbfcdc-fbcb8                        1/1     Running       0          66m
```

### *Environment*:

lattest

shawnh2:
**What type of PR is this?**

Fixes #2637


cnvergence (approved):
awesome

arkodg:
hey @shawnh2 tring to better understand the issue and fix, can you help outline all the steps.
Its unclear to me why this issue is specific to `mergeGateways` and skipping processing
of gateways means we'll never we able to clean them up (if a GatewayClass was deleted before a Gateway)

shawnh2:
Discussed this in today's community meeting, this fix will introduce another bug: the Gateways will not be deleted with this fix.

The expected behaviour should be if we remove GatewayClass before Gateways, the merge-gateways features should not be failed.

shawnh2:
Fow now, if we deleted GC first, merge-gateways feature will not fail.

But if we trying to delete all the resources (including GC, EnvoyProxy, GTWs, Routes), the finalizer of GC will not be removed. Because the EnvoyProxy got deleted before GC, this will casue controller to raise a reconcile error in here:

https://github.com/envoyproxy/gateway/blob/607d8bcd9a833841eee057b263c0bdc06de0fbea/internal/provider/kubernetes/controller.go#L238

and return, skip the finalizers check logic in here:

https://github.com/envoyproxy/gateway/blob/607d8bcd9a833841eee057b263c0bdc06de0fbea/internal/provider/kubernetes/controller.go#L288

---

So IMO the fix here will be: **introduce a finalizer for EnvoyProxy**

Since the EnvoyProxy is a `parameterRef` of GC, the EnvoyProxy should not be deleted if it got GC associated with it.

cnvergence:
I can help you out there, I was thinking about refactoring it and bringing this topic back

shawnh2:
> I can help you out there, I was thinking about refactoring it and bringing this topic back



> I can help you out there, I was thinking about refactoring it and bringing this topic back

Great, thanks, so this fix will only concentrating on the `merge-gateways should not fail if GC delete first` part.

cnvergence:
Awesome @shawnh2, I will make sure to test this one for you, as I have some examples lying around.
I did work on the finalizer before https://github.com/envoyproxy/gateway/issues/865 and I still think that it is worth bringing it back, now you have another point that supports this.

internal/provider/kubernetes/helpers.go
```diff
@@ -214,3 +215,13 @@ func classAccepted(gc *gwapiv1.GatewayClass) bool {
 
 	return false
 }
+
+// classMarkedForDeletion returns true if the provided GatewayClass is
+// marked for deletion and its finalizer has been removed.
+func classMarkedForDeletion(gc *gwapiv1.GatewayClass) bool {
+	if gc == nil {
```
arkodg, line 222: non blocker comment:do we hit this ? I understand why its added, to make sure we dont crash below, but should `classMarkedForDeletion` return `true` when `gc` is `nil`  ?

internal/provider/kubernetes/controller.go
```diff
@@ -233,8 +233,8 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 		}
 
 		// Process the parametersRef of the accepted GatewayClass.
-		// This should run before processBackendRefs
-		if managedGC.Spec.ParametersRef != nil && managedGC.DeletionTimestamp == nil {
+		// This should run before processBackendRefs.
+		if managedGC.Spec.ParametersRef != nil && !classMarkedForDeletion(managedGC) {
```
arkodg, line 237: can you help explain how this fixes the issue
1. gc is deleted
2. we skip this step
3. we are not passing any `EnvoyProxy` to the next layer, wont this cause infra churn ?

shawnh2: if we don't fix here: 
- if gc is deleted first, and it has a envoyproxy as parameterRef (which still exist)
- since we still have some gtws ref to this gc, so this gc won't be deleted due to finalizers
- but this line doesn't check whether gc still have finalizers, false positively think the gc is deleted, stop processing gc's parameterRef
- this will cause the gatewayapi layer won't get any info from envoyproxy, so the gatewayapi layer will not perform merge-gateways anymore

shawnh2: > can you help explain how this fixes the issue
> 
> 1. gc is deleted
> 
> 2. we skip this step
> 
> 3. we are not passing any `EnvoyProxy` to the next layer, wont this cause infra churn ?
> 
> 

as for 2, we didn't skip this step, we enhanced this step by adding extra finalizers check

shawnh2:
hold this, need to add finalizers for EnvoyProxy before resolving this.

cnvergence:
looking into finalizer, will send the PR soon

arkodg:

PR #3527, opened 2024-06-03
api: support user provided names for generated k8s resources

arkodg:
* Users prefer setting their own custom names for generated k8s resources for the Envoy Proxy fleet instead of the autogenerated resource names
  * https://envoyproxy.slack.com/archives/C03E6NHLESV/p1717068088779319
  * https://github.com/knative-extensions/net-gateway-api/pull/738
  * https://github.com/envoyproxy/gateway/issues/2141

PR #3537, opened 2024-06-04
feat: support custom names for generated k8s resources

arkodg:
Relates to https://github.com/envoyproxy/gateway/pull/3527

### DIFF
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -231,7 +231,7 @@ func GatewayClassOwnerLabel(name string) map[string]string {
 // OwnerLabels returns the owner labels based on the mergeGateways setting
 func OwnerLabels(gateway *gwapiv1.Gateway, mergeGateways bool) map[string]string {
 	if mergeGateways {
-		return GatewayClassOwnerLabel(gateway.Name)
+		return GatewayClassOwnerLabel(gateway.Spec.GatewayClassName)
 	}
 
 	return GatewayOwnerLabels(gateway.Namespace, gateway.Name)
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -228,6 +228,15 @@ func GatewayClassOwnerLabel(name string) map[string]string {
 	return map[string]string{OwningGatewayClassLabel: name}
 }
 
+// OwnerLabels returns the owner labels based on the mergeGateways setting
+func OwnerLabels(gateway *gwapiv1.Gateway, mergeGateways bool) map[string]string {
+	if mergeGateways {
+		return GatewayClassOwnerLabel(gateway.Name)
+	}
+
+	return GatewayOwnerLabels(gateway.Namespace, gateway.Name)
+}
+
 // servicePortToContainerPort translates a service port into an ephemeral
 // container port.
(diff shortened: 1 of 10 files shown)


cnvergence (changes requested):
probably we would need to also change it in kubernetes provider
https://github.com/envoyproxy/gateway/blob/607d8bcd9a833841eee057b263c0bdc06de0fbea/internal/provider/kubernetes/helpers.go#L151

arkodg:
hey @cnvergence can you elaborate why thats needed, the above diff is changing the value right before setting the resource name ?

arkodg:
oh @cnvergence I see the issue now, we are fetching based on name and not labels 😢 
https://github.com/envoyproxy/gateway/blob/cdd4e065d4a2a5e3657052664073972f06ac77dd/internal/provider/kubernetes/predicates.go#L467
&
https://github.com/envoyproxy/gateway/blob/cdd4e065d4a2a5e3657052664073972f06ac77dd/internal/provider/kubernetes/predicates.go#L483

cc @dprotaso this is the issue seen in https://github.com/knative-extensions/net-gateway-api/pull/738

internal/gatewayapi/helpers.go
```diff
@@ -228,6 +228,15 @@ func GatewayClassOwnerLabel(name string) map[string]string {
 	return map[string]string{OwningGatewayClassLabel: name}
 }
 
+// OwnerLabels returns the owner labels based on the mergeGateways setting
+func OwnerLabels(gateway *gwapiv1.Gateway, mergeGateways bool) map[string]string {
```
cnvergence, line 232: 🚀

cnvergence (approved):
LGTM

davem-git:
If this is addressed here, will it also fix when I rename the `envoyDeployment`:, the HPA `envoyHpa`? It no longer works it keeps the original generated name. There doesn't seem to be a way to patch that either

arkodg:
> If this is addressed here, will it also fix when I rename the `envoyDeployment`:, the HPA `envoyHpa`? It no longer works it keeps the original generated name. There doesn't seem to be a way to patch that either

thanks for catching this ! ive fixed the deployment targetRef naming in hpa.
I havent changed the name on the hpa resource itself, that needs an API change, which can be handled in a future PR

zhaohuabing (commented):
🚀

zhaohuabing (commented):
🚀

zhaohuabing (approved):

davem-git:
On this same issue. I've renamed the deployment by patching the name, this causes the gateway to break. While the deployment with 3 replicas I get this error 

```sh
  Last Transition Time:  2024-06-14T15:02:58Z
    Message:               The Gateway has been scheduled by Envoy Gateway
    Observed Generation:   1
    Reason:                Accepted
    Status:                True
    Type:                  Accepted
    Last Transition Time:  2024-06-14T15:03:07Z
    Message:               Deployment replicas unavailable
    Observed Generation:   1
    Reason:                NoResources
    Status:                False
    Type:                  Programmed
```

the gateway status is this 

```sh
NAME            CLASS           ADDRESS         PROGRAMMED   AGE
acme-envoy   acme-envoy   IPADDRESS   False        4m19s
```

arkodg:
@davem-git can you raise a separate GH issue to highlight the issue you are facing

PR #3540, opened 2024-06-05
feat(api): Add http health check api to ClientTrafficPolicy

### LINKED ISSUE #3258: HealthCheck config for HTTP Downstream
*Description*:

A HealthCheck config is required to setup a health check downstream HTTP filter, aiding the load balancer before Envoy in assessing Envoy’s health. Presently, a HealthCheck exists in the BackendTrafficPolicy, but it serves the upstream.

[optional *Relevant Links*:]

https://www.envoyproxy.io/docs/envoy/v1.29.2/configuration/http/http_filters/health_check_filter

aoledk:

When envoy HTTP listener is registered to LB, LB will issue health check requests to check whether the HTTP service is healthy, and we should configure LB to check traffic port instead of `19001` like https://github.com/envoyproxy/gateway/issues/3258#issuecomment-2073889689 mentioned.

So EG should let user to configure health check settings for specific HTTP listener, envoy will not pass through health check requests and respond with a 200 or a 503 depending on the current draining state of the server.

Fixes #3258


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -310,6 +316,19 @@ type HTTP2Settings struct {
...
+type HealthCheckSettings struct {
+	// Headers specifies a set of health check request headers to match on. The health check filter
+	// will check a request’s headers against all the specified headers. To specify the health check
+	// endpoint, set the `:path` header to match on.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Headers []gwapiv1.HTTPHeaderMatch `json:"headers,omitempty"`
```
arkodg, line 329: prefer we only exposed one field here `Path` and internally mapped the value from here to `:path` within `Headers` of https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/health_check_filter#health-check

arkodg (approved):
LGTM thanks !

shawnh2 (approved):

shawnh2 (approved):
LGTM! Thanks!

cnvergence (approved):

zhaohuabing (approved):
🚀

PR #3542, opened 2024-06-05
feat(translator): Set zone for endpoint and arrange endpoints by zone in EDS

### LINKED ISSUE #3538: Set zone from EndpointSlice for xDS Endpoint
*Description*:
> xDS Endpoint's zone will be needed as **first step** when user wants to apply `priority levels` or `zone aware routing` for cluster's load balancing to reduce cost.

[optional *Relevant Links*:]
> https://github.com/envoyproxy/gateway/issues/1909#issuecomment-2126782250

aoledk:

Set zone for endpoint and arrange endpoints by zone in EDS if zone is available in EndpointSlice. This is required before users can apply `priority levels` or `zone aware routing` load balancing strategy. 

Fixes #3538


internal/xds/translator/cluster.go
```diff
@@ -404,18 +405,18 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
-		// If we don't do this, Envoy will merge all LocalityLbEndpoints into one.
-		// We use the name of the backendRef as a pseudo region name.
-		locality := &endpointv3.LocalityLbEndpoints{
-			Locality: &corev3.Locality{
-				Region: fmt.Sprintf("%s/backend/%d", clusterName, i),
-			},
-			LbEndpoints: endpoints,
-			Priority:    0,
+			// If irEp has unset zone, leave it empty.
+			var zone string
```
arkodg, line 410: this breaks weight based routing, currently a setting has a weight, and there's 1:1 b/w xds locality and IR setting

arkodg: this can be fixed by copying the IR setting weight into `lbEndpoint.LoadBalancingWeight` (currently set to 1) and removing the locality weight (or setting it to 1)

aoledk: It seems like splitting 1 backend with specific weight (as 1 IR setting) to multiple localities with same specific weight doesn't break weighted based routing.

Compared to your suggestion in which we assign the specific weight to individual endpoint, the same endpoint has same chance to be chosen by envoy.

Since EG has enabled locality weighted load balancing, we must set locality weight, or else the locality will not be assigned any load [^1].

https://github.com/envoyproxy/gateway/blob/85cac8486dd8ba88841c36cd0cc65157c6bec7a2/internal/xds/translator/cluster.go#L85-L89

[^1]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#config-endpoint-v3-localitylbendpoints

arkodg:
closing this PR since its become inactive, Zone aware routing is now available on v1.4.0

PR #3532, opened 2024-06-04
feat: Support attaching EnvoyProxy resource to Gateways

haoqixu:
Fixes #3317

PR #3543, opened 2024-06-05
feat(translator): support switching between service/clusterIP routing

### LINKED ISSUE #1900: Support routing to Service Cluster IP

evacchi:
This PR surfaces the internal `EndpointRoutingDisabled` flag to the `EnvoyProxy` spec. This is still missing an e2e test, opening as draft to double-check. 

~The flag defaults to true as it is currently set internally. It might make sense to flip the name to `ClusterIPRoutingEnabled` (so that the default is, perhaps less surprisingly, "false")~

The new config option is called `routingType` and it can be set to `Service` or `Cluster`. 

Fixes #1900


api/v1alpha1/envoyproxy_types.go
```diff
@@ -68,6 +68,12 @@ type EnvoyProxySpec struct {
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
 
+	// EndpointRoutingDisabled can be set to true to use
+	// the Service Cluster IP for routing to the backend
+	// instead.
+	// +optional
+	EndpointRoutingDisabled *bool `json:"endpointRoutingDisabled,omitempty"`
```
arkodg, line 75: in the API, can we call this `RoutingType` , creating two enums `Service` and `Endpoint` and defaulting to `Endpoint` ?

arkodg:
thanks for picking this one up @evacchi ! enabling service routing should be make it easier to integrate with service meshes

internal/cmd/egctl/translate.go
```diff
@@ -280,7 +281,7 @@ func translateGatewayAPIToIR(resources *gatewayapi.Resources) (*gatewayapi.Trans
 		GatewayControllerName:   string(resources.GatewayClass.Spec.ControllerName),
 		GatewayClassName:        gwapiv1.ObjectName(resources.GatewayClass.Name),
 		GlobalRateLimitEnabled:  true,
-		EndpointRoutingDisabled: true,
+		EndpointRoutingDisabled: isEndpointRoutingDisabled(resources),
```
arkodg, line 284: lets keep this to `true` , if you look at the newly generated yaml we are loosing out on endpoint o/p because the input is not providing related endpointslices

internal/gatewayapi/testdata/envoyproxy-endpoint-routing-disabled.in.yaml
```diff
@@ -0,0 +1,71 @@
...
+            securityContext:
+              runAsUser: 1000
+              runAsGroup: 3000
+              fsGroup: 2000
+              fsGroupChangePolicy: "OnRootMismatch"
+            volumes:
+              - name: certs
+                secret:
+                  secretName: envoy-cert
+gateways:
```
arkodg, line 57: suggest updating the input to also include a httproute e.g. https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/testdata/httproute-attaching-to-gateway.in.yaml 
this will ensure we can validate the Service IP in the o/p xdsIR, currently its empty
and shaving off `envoyproxy.spec.provider` / `nil` , so this test case only focuses on `endpointType`

test/e2e/testdata/httproute-routingtype.yaml
```diff
@@ -0,0 +1,36 @@
...
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /backend-fqdn
+    backendRefs:
+    - group: gateway.envoyproxy.io
+      kind: Backend
```
arkodg, line 24: this feature won't work with `Backend` because it doesnt have a `Service` with a clusterIP field, can you use one of the existing values e.g. 
```
      backendRefs:
        - name: infra-backend-v1
          port: 8080
``

File: internal/cmd/egctl/translate.go (hunk 2 above)
Comment (evacchi) [reply internal/cmd/egctl/translate.go:284]: so I assume you mean that I have inadvertently flipped the default in `isEndpointRoutingDisabled`, i.e. the default should be `ServiceRoutingType`

Comment (arkodg) [reply internal/cmd/egctl/translate.go:288]: the default for the system, in the gateway-api layer is `EndpointType`, but for this test case in `cmd/egctl` we've flipped it to `Service` , yah its confusing, sorry :)

File: api/v1alpha1/envoyproxy_types.go [hunk 5]
```diff
@@ -68,6 +68,11 @@ type EnvoyProxySpec struct {
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
 
+	// RoutingType can be set to "service" to use Service Cluster IP for routing to the backend,
+	// or it can be set to "endpoint" to use Endpoint routing. The default is "endpoint".
```
arkodg, line 72:
```suggestion
	// or it can be set to "Endpoint" to use Endpoint routing. The default is "Endpoint".
```

api/v1alpha1/envoyproxy_types.go
```diff
@@ -68,6 +68,11 @@ type EnvoyProxySpec struct {
 	// +optional
 	Concurrency *int32 `json:"concurrency,omitempty"`
 
+	// RoutingType can be set to "service" to use Service Cluster IP for routing to the backend,
```
arkodg, line 71:
```suggestion
	// RoutingType can be set to "Service" to use the Service Cluster IP for routing to the backend,
```

test/e2e/testdata/httproute-routingtype.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: custom-proxy-config
+  namespace: gateway-conformance-infra
+spec:
+  routingType: service
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: httproute-to-backend-fqdn
```
arkodg, line 12:
```suggestion
  name: httproute-to-service
```

test/e2e/testdata/httproute-routingtype.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /backend-fqdn
```
arkodg, line 21:
```suggestion
        value: /to-cluster-ip
```

test/e2e/testdata/httproute-routingtype.yaml
```diff
@@ -0,0 +1,35 @@
...
+    backendRefs:
+      - name: infra-backend-v1
+        port: 8080
+---
+apiVersion: gateway.envoyproxy.io/v1alpha1
```
arkodg, line 26: lets rm this `Backend` resource

test/e2e/tests/routingtype.go
```diff
@@ -0,0 +1,48 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, EnvoyGatewayRoutingType)
+}
+
+var EnvoyGatewayRoutingType = suite.ConformanceTest{
+	ShortName:   "EnvoyGatewayRoutingType",
+	Description: "Routes with a backend ref to a backend",
```
arkodg, line 26:
```suggestion
	Description: "Tests routing to the backendRef's ClusterIP",
```

arkodg:
hey you also want to copy over the value into the translator here https://github.com/envoyproxy/gateway/blob/33fceb05ca611e30107ff841cac9d7762eea2048/internal/gatewayapi/runner/runner.go#L83

internal/cmd/egctl/translate.go
evacchi: I have restored the line to read `EndpointRoutingDisabled: true,`

internal/gatewayapi/testdata/envoyproxy-endpoint-routing.out.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - name: service-1
+        port: 8080
+      matches:
+      - path:
+          value: /
+  status:
+    parents:
+    - conditions:
+      - lastTransitionTime: null
+        message: No listeners included by this parent ref allowed this attachment.
```
arkodg, line 64: this is causing the route to fail, so the `routes` section is missing in the xds IR

arkodg: probably due to this section
```
    - allowedRoutes:
        namespaces:
          from: Same
```

so you can tweak route ns or change the `from` to `All`

internal/gatewayapi/runner/runner.go
```diff
@@ -231,6 +233,21 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 	r.Logger.Info("shutting down")
 }
 
+func isEndpointRoutingDisabled(resources *gatewayapi.Resources) bool {
+	endpointRoutingDisabled := true
+	if resources.EnvoyProxy != nil {
+		switch ptr.Deref(resources.EnvoyProxy.Spec.RoutingType, v1alpha1.ServiceRoutingType) {
```
arkodg, line 239: this should default to `v1alpha1.EndpointRoutingType`

internal/gatewayapi/runner/runner.go
```diff
@@ -231,6 +233,21 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 	r.Logger.Info("shutting down")
 }
 
+func isEndpointRoutingDisabled(resources *gatewayapi.Resources) bool {
+	endpointRoutingDisabled := true
```
arkodg, line 237: should be `false`

internal/gatewayapi/runner/runner.go
```diff
@@ -231,6 +233,21 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 	r.Logger.Info("shutting down")
 }
 
+func isEndpointRoutingDisabled(resources *gatewayapi.Resources) bool {
```
arkodg, line 236: in `internal/gateway/runner` we set `endpointRoutingDisabled: false` by default but in `/internal/cmd/egtl` we want to enable it by default to simplify the CLI operation

internal/gatewayapi/runner/runner.go
evacchi: ok I guess I got it all backwards

internal/gatewayapi/testdata/envoyproxy-endpoint-routing.out.yaml
evacchi: `from: All`

internal/gatewayapi/testdata/routingpolicy/envoyproxy-service-routing.out.yaml
```diff
@@ -0,0 +1,134 @@
...
+      port: 10080
+      routes:
+      - destination:
+          name: httproute/default/httproute-1/rule/0
+          settings:
+          - endpoints:
+            - host: 1.1.1.1
+              port: 8080
+            protocol: HTTP
+            weight: 1
```
evacchi, line 127: I think this is what we wanted right? 🤔

arkodg (approved):
LGTM thanks !

arkodg:
hey @evacchi one more hiccup, hopefully last one, there's a another PR https://github.com/envoyproxy/gateway/pull/3532 that is adding support for reconciling multiple `EnvoyProxies` that can attach to `Gateway`, and that impacts this PR.

I suggest deleting `IsEndpointRoutingDisabled` from the translator struct and instead recomputing it using the helper func that youve added in this PR, e.g. in here 
https://github.com/envoyproxy/gateway/blob/e866ec138ca1666e55073d980f2d1aaf477f36d9/internal/gatewayapi/route.go#L193

this will also eliminate the need to have a separate test function, and you'll be able to reuse the same one

Once this PR is in, #3532 can modify and pass the right `EnvoyProxy` into the helper func cc @haoqixu

evacchi:
apparently the e2e is failing 🤔

evacchi:
@arkodg wouldn't that flip back the default here to false?
 https://github.com/envoyproxy/gateway/blob/412b1b787dc30fbbff02056cc56e502ae3b0419c/internal/cmd/egctl/translate.go#L283

evacchi:
likely a flake

```
        --- FAIL: TestGatewayAPIConformance/GatewayModifyListeners/should_be_able_to_remove_listeners,_which_would_then_stop_routing_the_relevant_traffic (360.02s)
```

tools/hack/create-cluster-kind-cfg.yml
evacchi, line 1: committed by mistake

tools/hack/create-cluster.sh
evacchi, line 1: committed by mistake

arkodg (approved):
LGTM thanks !

cnvergence (approved):

evacchi:
failed tests are just timing out

arkodg:
> failed tests are just timing out

kicked the CI, you can also comment on the GH issue with `/retest` which retriggers the CI

PR #3556, opened 2024-06-06
chore: automatically import Grafana dashboards when installing gateway-addons-helm

shawnh2:
**What type of PR is this?**

When deploying `gateway-addons-helm` successfully, we will get dashboards automatically import:

image

Fixes #


charts/gateway-addons-helm/values.yaml
```diff
@@ -10,6 +11,38 @@ grafana:
...
+      apiVersion: 1
+      providers:
+        - name: 'envoy-gateway'
+          orgId: 1
+          folder: 'envoy-gateway'
+          type: file
+          disableDeletion: false
+          editable: true
+          options:
+            path: /var/lib/grafana/dashboards/envoy-gateway
```
zhaohuabing, line 25: I guess the dashboards in the followings configuration will be downloaded and placed in this directory by grafana?

charts/gateway-addons-helm/values.yaml
```diff
@@ -10,6 +11,38 @@ grafana:
...
+  dashboards:
+    envoy-gateway:
+      prometheus-stats:
+        gnetId: 1
+        revision: 1
+        datasource: Prometheus
+      # Scrape and load dashboards of Envoy Gateway
+      envoy-global-dashboard:
+        url: https://raw.githubusercontent.com/envoyproxy/gateway/main/charts/gateway-addons-helm/dashboards/envoy-global.json
```
arkodg, line 34: is this the best way to do it ? or should the dashboard we part of this chart ?

charts/gateway-addons-helm/values.yaml
zhaohuabing: Oh, it's not.

charts/gateway-addons-helm/values.yaml
arkodg: the drawback of URL is, its not versioned

shawnh2: got it, prefer loading dashboards from local disk

shawnh2: change it to load dashboards from configmap

arkodg (approved):
LGTM thanks !

charts/gateway-addons-helm/values.yaml
```diff
@@ -10,6 +11,21 @@ grafana:
   adminPassword: admin
   service:
     type: LoadBalancer
+  dashboardProviders:
```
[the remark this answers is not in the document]
shawnh2: yes, grafana will need to hold the imported dashboards from configmap according to the provider path.

zirain (approved):

Issue #3468, opened 2024-05-24
add CEL validation for BackendRef Group


zirain:
e.g. 

https://github.com/envoyproxy/gateway/blob/8f2e175de6ca0842f59be06ae37bbb6f196c69cd/api/v1alpha1/accesslogging_types.go#L117-L123

phantooom:
i search `BackendRefs []BackendRef` in eg codebase. i found all BackendRefs field had been added.
//+kubebuilder:validation:XValidation

What can I do for this PR? Could you give me some suggestions?

zirain:
Right now, only `Service` is support, should limit `group` is `empty`(use default value `core`) or `core`.

phantooom:
i try to add CEL validation for BackendRef Group. 
`
panic: Failed to start testenv: unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "envoyproxies.gateway.envoyproxy.io": CustomResourceDefinition.apiextensions.k8s.io "envoyproxies.gateway.envoyproxy.io" is invalid: [

spec.validation.openAPIV3Schema.properties[spec].properties[telemetry].properties[accessLog].properties[settings].items.properties[sinks].items.properties[als].properties[backendRefs].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.048576x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), 


spec.validation.openAPIV3Schema.properties[spec].properties[telemetry].properties[accessLog].properties[settings].items.properties[sinks].items.properties[openTelemetry].properties[backendRefs].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.048576x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]
`

should we restrict  the maxitem for  ProxyAccessLog.Sinks  ProxyAccessLog.Settings?

phantooom:
> i try to add CEL validation for BackendRef Group.
> 
> ```
> panic: Failed to start testenv: unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "envoyproxies.gateway.envoyproxy.io": CustomResourceDefinition.apiextensions.k8s.io "envoyproxies.gateway.envoyproxy.io" is invalid: [
> 
[...]

PR #3557, opened 2024-06-06
add CEL validation for BackendRef Group

phantooom:
**What type of PR is this?**
* "add CEL validation for BackendRef Group"

Fixes https://github.com/envoyproxy/gateway/issues/3468


api/v1alpha1/accesslogging_types.go
shawnh2, line 1: can you add some test under https://github.com/envoyproxy/gateway/tree/main/test/cel-validation ? just to check whether this validation is being enforced

api/v1alpha1/accesslogging_types.go
```diff
@@ -120,6 +120,7 @@ type ALSEnvoyProxyAccessLog struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=1
 	// +kubebuilder:validation:XValidation:message="BackendRefs only supports Service kind.",rule="self.all(f, f.kind == 'Service')"
+	// +kubebuilder:validation:XValidation:message="BackendRefs only supports Core group.",rule="self.all(f, f.group == '')"
```
shawnh2, line 123: `f.group == ''` is this correct ?

api/v1alpha1/accesslogging_types.go
phantooom: ok

api/v1alpha1/accesslogging_types.go
phantooom: https://github.com/kubernetes/api/blob/master/core/v1/register.go#L26
core api group is '' .   it may works.

api/v1alpha1/accesslogging_types.go
```diff
@@ -11,6 +11,8 @@ type ProxyAccessLog struct {
 	// Settings defines accesslog settings for managed proxies.
 	// If unspecified, will send default format to stdout.
 	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=50
```
phantooom, line 15: i add maxItems. if not do this cel estimation  will timeout.

api/v1alpha1/accesslogging_types.go
zirain: `core` is also valid?

phantooom: core is invalid. core api has no group. so only group=='' is valid. 

Correct me . if I'm wrong，thx~

zirain: I found some usage about `group: core` in mesh and gep-709, @arkodg anything recall?

arkodg: this looks good to me
`''` == `core` https://github.com/kubernetes-sigs/gateway-api/blob/45c7d4ff434f744b5bf55ed81dfd059d78cd33f3/apis/v1/object_reference_types.go#L97

arkodg (approved):

shawnh2 (approved):

PR #3564, opened 2024-06-07
feat: Wasm OCI image

zhaohuabing:
**This PR adds support for Wasm OCI image format to EG**. 
It allows EG to download Wasm images from remote registries and serve them to the Envoy fleet via a local HTTP server inside EG running on 18002.

* The EG Wasm cache fetches Wasm modules from their original URLs and stores them locally. Currently, the cache directory resides within the EG container, it can be made configurable to use a mounted volume if necessary.
* Gateway API translator updates the original Wasm downloading URLs in the `EnvoyExtensionPolicy` to point to the local EG HTTP server.
* To ensure a consistent UI experience, HTTP Wasm modules are also cached and served locally.
* To prevent unauthorized access to private images, the OIDC HMAC secret is used as a hash salt to generate unguessable downloading paths for Wasm modules.
* The cached Wasm modules are purged periodically. A cached Wasm module will be deleted if it hasn't been requested by Envoy proxies within a certain period. Currently, defaults to 24 hours, which can be configurable if needed.
* The total cache size is limited to 10G by default, which can be configurable if needed.

Implements: #3304
Design: https://github.com/envoyproxy/gateway/pull/3313


zhaohuabing:
Hi @envoyproxy/gateway-reviewers my apologies for the large PR. I would have broken it into multiple smaller ones if possible :-).  A lot of *.out.yaml test files have been updated, as a static cluster has been added to the Envoy bootstrap config. 

If you're going to review it, the most significant changes are in the internal/wasm package, which implements the Wasm file cache and an HTTP server to serve these files to the Envoy fleet.

Additionally, the Gateway API runner has been modified to initialize the Wasm cache server.

I am still working on setting up a local test OCI registry to avoid accessing Docker hub during e2e tests. However, this PR should be ready for review. I would appreciate it if you could review this at your convenience and share your comments.

api/v1alpha1/authorization_types.go
```diff
@@ -27,7 +27,8 @@ type Authorization struct {
 
 // AuthorizationRule defines a single authorization rule.
 type AuthorizationRule struct {
-	// Name is a user-friendly name for the rule. It's just for display purposes.
+	// Name is a user-friendly name for the rule.
+	// If not specified, EG will generate a unique name for the rule.
```
arkodg, line 31:
```suggestion
	// If not specified, Envoy Gateway will generate a unique name for the rule.n
```

api/v1alpha1/wasm_types.go
```diff
@@ -81,8 +81,21 @@ type WasmCodeSource struct {
 
 	// SHA256 checksum that will be used to verify the wasm code.
 	//
+	// If not specified, EG will not verify the downloaded wasm code.
```
arkodg, line 84:
```suggestion
	// If not specified, Envoy Gateway will not verify the downloaded wasm code.
```

api/v1alpha1/wasm_types.go
```diff
@@ -81,8 +81,21 @@ type WasmCodeSource struct {
...
+	// PullPolicy is the policy to use when pulling the Wasm module by either the HTTP or Image source.
+	// This field is only applicable when the SHA256 field is not set.
+	//
+	// If not specified, the default policy is IfNotPresent except for OCI images whose tag is latest.
+	//
+	// Note: EG does not update the Wasm module every time an Envoy proxy requests
+	// the Wasm module even if the pull policy is set to Always.
+	// It only updates the Wasm module when the EnvoyExtension resource version changes.
+	// +optional
+	PullPolicy *ImagePullPolicy `json:"pullPolicy,omitempty"`
```
arkodg, line 98: shouldnt this be inside `Image` ?

go.mod
```diff
@@ -163,35 +178,37 @@ require (
...
+	k8s.io/kube-openapi v0.0.0-20240521193020-835d969ad83a // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/kustomize/api v0.13.5-0.20230601165947-6ce0bf390ce3 // indirect
-	sigs.k8s.io/kustomize/kyaml v0.14.3-0.20230601165947-6ce0bf390ce3 // indirect
+	sigs.k8s.io/kustomize/api v0.17.2 // indirect
+	sigs.k8s.io/kustomize/kyaml v0.17.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
+
+replace dario.cat/mergo => github.com/imdario/mergo v0.3.12
```
arkodg, line 214: why is this needed ?

internal/utils/sets/set.go
```diff
@@ -0,0 +1,294 @@
...
+package sets
+
+import (
+	"fmt"
+	"slices" // nolint: depguard
+
+	"golang.org/x/exp/constraints"
+)
+
+type Set[T comparable] map[T]struct{}
```
arkodg, line 29: can we use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets instead

api/v1alpha1/wasm_types.go
zhaohuabing: It's also used for HTTP code source.

```
	// PullPolicy is the policy to use when pulling the Wasm module by either the HTTP or Image source.
	// This field is only applicable when the SHA256 field is not set.
	//
	// If not specified, the default policy is IfNotPresent except for OCI images whose tag is latest.
```

internal/wasm/httpserver.go
```diff
@@ -0,0 +1,228 @@
...
+	maxFailedAttempts = 10
+	// TODO: zhaohuabing make this configurable
+	attemptsResetInterval = 5 * time.Minute
+	attemptResetDelay     = 1 * time.Hour
+)
+
+var _ Cache = &HTTPServer{}
+
+// HTTPServer wraps a local file cache and serves the Wasm modules over HTTP.
+type HTTPServer struct {
```
arkodg, line 34: can we model this as a runner ?
gateway-api can publish the IR, and this runner could subscribe to this layer

api/v1alpha1/wasm_types.go
arkodg: ^ is a little confusing to understand

go.mod
zhaohuabing: It's because of incompatible changes between mergo@v1.0.0 and mergo@v0.3.x

go mod verify
go: github.com/imdario/mergo@v1.0.0: parsing go.mod:
        module declares its path as: dario.cat/mergo
                but was required as: github.com/imdario/mergo

go mod graph|grep " github.com/imdario/mergo"
github.com/envoyproxy/gateway github.com/imdario/mergo@v1.0.0
github.com/Masterminds/sprig/v3@v3.2.3 github.com/imdario/mergo@v0.3.11
github.com/rubenv/sql-migrate@v1.6.1 github.com/imdario/mergo@v0.3.13
helm.sh/helm/v3@v3.15.1 github.com/imdario/mergo@v0.3.13
k8s.io/apiextensions-apiserver@v0.30.1 github.com/imdario/mergo@v0.3.6
k8s.io/apiserver@v0.30.1 github.com/imdario/mergo@v0.3.6
k8s.io/cli-runtime@v0.30.1 github.com/imdario/mergo@v0.3.6
k8s.io/client-go@v0.30.1 github.com/imdario/mergo@v0.3.6
k8s.io/kubectl@v0.30.1 github.com/imdario/mergo@v0.3.6
sigs.k8s.io/controller-runtime@v0.18.4 github.com/imdario/mergo@v0.3.6
sigs.k8s.io/gateway-api@v1.1.0 github.com/imdario/mergo@v0.3.16
k8s.io/client-go@v0.18.4 github.com/imdario/mergo@v0.3.5
sigs.k8s.io/controller-runtime@v0.6.1 github.com/imdario/mergo@v0.3.9
k8s.io/client-go@v0.18.2 github.com/imdario/mergo@v0.3.5

api/v1alpha1/wasm_types.go
zhaohuabing: If SHA is not specified, we need to know whether the Wasm code need updating regardless its source is HTTP or OCI registry.

api/v1alpha1/wasm_types.go
```diff
@@ -81,8 +81,21 @@ type WasmCodeSource struct {
 
 	// SHA256 checksum that will be used to verify the wasm code.
 	//
+	// If not specified, Envoy Gateway will not verify the downloaded wasm code.
 	// kubebuilder:validation:Pattern=`^[a-f0-9]{64}$`
-	SHA256 string `json:"sha256"`
+	// +optional
+	SHA256 *string `json:"sha256"`
```
arkodg, line 87: this is tricky that it lives as top level, for image is this sha of the wasm module or the oci artifact

api/v1alpha1/wasm_types.go
```diff
@@ -100,31 +113,32 @@ const (
...
 type HTTPWasmCodeSource struct {
 	// URL is the URL containing the wasm code.
+	// +kubebuilder:validation:Pattern=`^((https?:)(\/\/\/?)([\w]*(?::[\w]*)?@)?([\d\w\.-]+)(?::(\d+))?)?([\/\\\w\.()-]*)?(?:([?][^#]*)?(#.*)?)*`
 	URL string `json:"url"`
 }
 
 // ImageWasmCodeSource defines the OCI image containing the wasm code.
 type ImageWasmCodeSource struct {
 	// URL is the URL of the OCI image.
+	// URL can be in the format of `registry/image:tag` or `registry/image@sha256:digest`.
```
arkodg, line 123: we should also mention the name and the expected file location for the module in the oci image

api/v1alpha1/wasm_types.go
zhaohuabing: Do you prefer having a SHA256 inside the HTTP and Wasm code source separately?

internal/wasm/options.go
```diff
@@ -0,0 +1,108 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	DefaultPurgeInterval         = 1 * time.Hour
+	DefaultModuleExpiry          = 24 * time.Hour
+	DefaultHTTPRequestTimeout    = 15 * time.Second
+	DefaultHTTPRequestMaxRetries = 5
+	DefaultPullTimeout           = 5 * time.Minute
+	DefaultMaxCacheSize          = 1024 * 1024 * 1024 * 10 // 10GB
```
arkodg, line 34: is this static or dynamic ? id suggest making this lesser, e.g. to 1GB

zhaohuabing: It's the default value for the cached wasm file size. We can expose this and other cache configuration options to API if needed in the future.

internal/wasm/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		"number of Wasm remote fetch cache entries.",
+	)
+
+	wasmCacheLookupCount = metrics.NewCounter(
+		"wasm_cache_lookup_count",
+		"number of Wasm remote fetch cache lookups.",
+	)
+
+	wasmRemoteFetchCount = metrics.NewCounter(
+		"wasm_remote_fetch_count",
```
arkodg, line 49: prob needs to be split up into 2 - error, total
cc @shawnh2 any prior art here for naming

zhaohuabing: There are two different naming styles in EG:

```
xds_snapshot_creation_total
xds_snapshot_creation_failed
xds_snapshot_creation_success
```

```
status_update_total
status_update_failed_total
status_update_success_total
status_update_conflict_total
```

Prefer the first one. We also need to align "failed, failure, error" to a single term.

internal/ir/xds.go
```diff
@@ -2226,17 +2226,26 @@ type Wasm struct {
...
+	Code *HTTPWasmCode `json:"httpWasmCode,omitempty"`
 }
 
 // HTTPWasmCode holds the information associated with the HTTP Wasm code source.
+// +k8s:deepcopy-gen=true
 type HTTPWasmCode struct {
-	// URL is the URL of the Wasm code.
-	URL string `json:"url"`
+	// EGServingURL is the URL of the Wasm code served by the local EG HTTP server.
+	EGServingURL string `json:"egServingURL"`
```
arkodg, line 2240: prefer `ServingURL` and `OriginalURL`

internal/wasm/httpfetcher.go
```diff
@@ -0,0 +1,224 @@
...
+// requestTimeout is a timeout for each HTTP/HTTPS request.
+// requestMaxRetry is # of maximum retries of HTTP/HTTPS requests.
+func NewHTTPFetcher(requestTimeout time.Duration, requestMaxRetry int, logger logging.Logger) *HTTPFetcher {
+	if requestTimeout == 0 {
+		requestTimeout = 5 * time.Second
+	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	// nolint: gosec
+	// This is only when a user explicitly sets a flag to enable insecure mode
+	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
arkodg, line 71: is this configurable in the UI ?

internal/wasm/metrics.go
arkodg: 1 looks more like envoy stats
2 looks like prom naming https://prometheus.io/docs/practices/naming/
@Xunzhuo @shawnh2 can you help drive/unify this naming

arkodg (approved):
overall LGTM !

there are some non blocking comments added in the PR, that can be tackled in follow ups

long term, my vote would be to move the wasm server into its own runner (like ratelimit)
this eliminates the shared fate problem of delaying xDS due to one slow URL download.

this also creates a eventual consistency problem, which can be resolved by implementing retries in the wasm cluster in envoy

internal/wasm/httpfetcher.go
zhaohuabing: Not yet, we can expose it to API when needed.

internal/wasm/metrics.go
zhaohuabing: Tracked with https://github.com/envoyproxy/gateway/issues/3684

go.mod
```diff
@@ -2,12 +2,15 @@ module github.com/envoyproxy/gateway
...
+replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.16
+
 require (
 	fortio.org/fortio v1.65.0
 	fortio.org/log v1.12.2
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/cncf/xds/go v0.0.0-20240423153145-555b57ec207b
-	github.com/davecgh/go-spew v1.1.1
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
+	github.com/docker/cli v26.1.3+incompatible
```
guydc, line 13: any way to avoid the docker dependency and use more generic oci libs here?

guydc (commented):
In general, can we have some sort of a feature toggle here? Some components like the runner and the servers will execute even when the feature is not in use. LAter we can expose this by default, when we're confident and/or extracted to a different component.

internal/gatewayapi/runner/runner.go
```diff
@@ -34,24 +52,67 @@ type Config struct {
...
+
 func (r *Runner) Name() string {
 	return string(egv1a1.LogComponentGatewayAPIRunner)
 }
 
 // Start starts the gateway-api translator runner
 func (r *Runner) Start(ctx context.Context) (err error) {
 	r.Logger = r.Logger.WithName(r.Name()).WithValues("runner", r.Name())
+
+	go r.startWasmCache(ctx)
```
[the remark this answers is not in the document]
zhaohuabing: Fail the wasm translation in EEP gateway API translation if cache fails to start.

go.mod
zhaohuabing: The docker lib is used to parse docker compatible config file to get auth info.

zhaohuabing:
> long term, my vote would be to move the wasm server into its own runner (like ratelimit)
this eliminates the shared fate problem of delaying xDS due to one slow URL download.

We probably could:
* move the Wasm cache server into a standalone runner
* fail the httproute containing Wasm extension if tat Wasm module has not been downloaded yet
* trigger the API translation again when a Wasm module is available in the cache

guydc (approved):
Seeing as most of the implementation that will execute regardless of EEP presence is taken from istio and already has sufficient production burn time, I take back my request for a feature gate.

arkodg (approved):

Issue #3396, opened 2024-05-15
OIDC SecurityPolicy: original authorization header removed


denniskniep:
*Description*:
I use the [OIDC SecurityPolicy](https://gateway.envoyproxy.io/latest/tasks/security/oidc/) to secure a route to an upstream application.
But the upstream application needs a different jwt token than issued by the envoy oauth2 filter. This jwt is added by the application frontend to the `authorization` header on every request.

Envoy oauth2 filter overwrites this `authorization` header. 

Therefore I created an EnvoyPatchPolicy to set `forward_bearer_token` to `false` (and also a Ticket to make it configurable in the SecurityPolicy CRD: https://github.com/envoyproxy/gateway/issues/3395)

The issue is that after setting `forward_bearer_token` to `false` no `authorization` header is forwarded to the upstream application.

When setting loglevel of envoy to debug then you can see the following logs:

Incoming Request with  `authorization` header from application frontend:

```
[2024-05-15 10:27:02.943][14][debug][http] [source/common/http/conn_manager_impl.cc:393] [Tags: "ConnectionId":"8"] new stream                                                                                                                                                
[2024-05-15 10:27:02.943][14][debug][http] [source/common/http/conn_manager_impl.cc:1192] [Tags: "ConnectionId":"8","StreamId":"12035930501644059014"] request headers complete (end_stream=true):
':method', 'GET'
':authority', 'example.com'
':scheme', 'https'
':path', '/my/path'
'accept', 'application/json, text/plain, */*'                                                                                                                                                                                                                                 
'content-type', 'application/json'                                                                                                                                                                                                          
'authorization', 'Bearer ey....'
'cookie', 'OauthHMAC-4105b4aa=Fr...; OauthExpires-4105b4aa=1715771641'
```


And after `skipping oauth flow due to valid hmac cookie` the  `authorization` header is gone!
```
[2024-05-15 10:27:02.943][14][debug][http] [source/common/http/conn_manager_impl.cc:1175] [Tags: "ConnectionId":"8","StreamId":"12035930501644059014"] request end stream
[2024-05-15 10:27:02.943][14][debug][connection] [./source/common/network/connection_impl.h:98] [Tags: "ConnectionId":"8"] current connecting state: false
[2024-05-15 10:27:02.943][14][debug][oauth2] [source/extensions/filters/http/oauth2/filter.cc:438] skipping oauth flow due to valid hmac cookie
[...]

PR #3567, opened 2024-06-08
oidc: preserve authorization header

zhaohuabing:
Envoy OAuth2 filter deletes the HTTP authorization header by default, which surprises users.

This PR sets the Envoy OAuth2 filter's `preserve_authorization_header` field to true if `ForwardAccessToken` is false. This flag can be flipped and added to the EG API if needed in the future.

fix: https://github.com/envoyproxy/gateway/issues/3396#issuecomment-2155699257


internal/xds/translator/oidc.go
```diff
@@ -169,9 +169,10 @@ func oauth2Config(oidc *ir.OIDC) (*oauth2v3.OAuth2, error) {
 				},
 			},
 			// every OIDC provider supports basic auth
-			AuthType:   oauth2v3.OAuth2Config_BASIC_AUTH,
-			AuthScopes: oidc.Scopes,
-			Resources:  oidc.Resources,
+			AuthType:                    oauth2v3.OAuth2Config_BASIC_AUTH,
+			AuthScopes:                  oidc.Scopes,
+			Resources:                   oidc.Resources,
+			PreserveAuthorizationHeader: true,
```
arkodg, line 175: if we land this now, we'll have to wait for Envoy's v1.31 which may be around July 15th, and EG is trying to release in the first week of July

zhaohuabing: I thought we alway release behind Envoy. Let's hold this.

arkodg: We mostly do :( ,  due to the v1.0 release, we moved off, hoping we get back on track for v1.2

zetaab: so are you kind of saying that this fix will not land to 1.1? So we need to wait until eg v1.2 is out that this will be supported? It will take some months to land this? really?

denniskniep: If we set `PreserveAuthorizationHeader` always to true, it will be no longer possible to set `ForwardBearerToken` to true. Because it will fail due to a invalid config exception.

Therefore I propose to make `PreserveAuthorizationHeader` also individually configurable.

zhaohuabing: > If we set PreserveAuthorizationHeader always to true, it will be no longer possible to set ForwardBearerToken to true. Because it will fail due to a invalid config exception.

Good point. We can disable  PreserveAuthorizationHeader if ForwardBearerToken is true.  

But do we need this in the API?

zhaohuabing: Good point. We can disable PreserveAuthorizationHeader if ForwardBearerToken is true.

There's two cases:

* ForwardBearerToken true: forward the oauth2 token, override the original one if it exists.
* ForwardBearerToken false: don't touch the original authorization header if it exists.

Do we need an explicit  PreserveAuthorizationHeader or RemoveAuthorizationHeader in  the API?  Do you have any use cases that need this configuration knob?

denniskniep: Regarding the versioning:

To merge this, we have to wait anyhow for the next envoy proxy release 1.31. This will include the added functionality. 

From my point of view it makes sense to release envoy gateway 1.1 with envoy proxy 1.30 as soon as its ready.
And directly after envoy proxy 1.31 is released, we can adopt it in envoy gateway and immediately release envoy gateway 1.2 (even without further envoy gateway functionality)

After this  envoy gateway 1.3 can be planned with additional functionality on eg side, but still with envoy proxy 1.31.

Benefit is, that we are supporting most recent envoy proxy version as soon as possible.

Does this make sense to you? What do you think?

zhaohuabing: I also believe that adopting version 1.31 at the earliest opportunity would be beneficial. Let's discuss this in the next EG meeting?

denniskniep: > Good point. We can disable PreserveAuthorizationHeader if ForwardBearerToken is true.
> 
> There's two cases:
> 
> * ForwardBearerToken true: forward the oauth2 token, override the original one if it exists.
> * ForwardBearerToken false: don't touch the original authorization header if it exists.
> 
> Do we need an explicit PreserveAuthorizationHeader or RemoveAuthorizationHeader in the API? Do you have any use cases that need this configuration knob?

 I don't have currently any use case for `forwardBearerToken: false` AND `preserveAuthorizationHeader: false`

So I am fine with your proposal.

But I am wondering, what was the original envoy intention of that feature.

This is the comment above the remove source code line :
>   // Sanitize the Authorization header, since we have no way to validate its content. Also,
  // if token forwarding is enabled, this header will be set based on what is on the HMAC cookie
  // before forwarding the request upstream.
  
And that is the corresponding commit: "oauth2: do not blindly accept requests with a token in the Authorization header"
https://github.com/envoyproxy/envoy/commit/e0c52894b0adaee575e3e981814bece8ece58b40

denniskniep: > Release EG 1.2 immediately after Envoy 1.31 is released: However, this will require additional effort as we would need to support two releases simultaneously during their lifetimes.

I favor option 2, but I don´t get why it is necessary to support multiple versions? Is there a specific lifetime guarantee for each release?

I would appreciate a general project policy, that a new eg version is released as soon as possible with the as is feature set of main branch, if a new envoy proxy version is released. This should be done independent of the eg milestones.

zhaohuabing: @denniskniep Yes, the problem is the additional work to support two releases, for example, security patches and bug fixes.

zhaohuabing: @arkodg @guydc and I discussed this during the last EG meeting. We have two options:

	1.	We can release 1.2.0 in October with Envoy 1.32.0 and then align all future EG release cadences with Envoy. So this needs to wait until October to be merged.
	2.	Alternatively, we can release EG v1.1.0 RCs based on the latest Envoy and pin the final release with Envoy 1.31.0. However, I’m not sure if the 1.31.0 release date would be too late for the EG release.

@arkodg, please correct me if I didn’t capture this accurately.

zetaab:
still another discussion is https://github.com/envoyproxy/gateway/pull/3567/files#r1632057185 :) so this pr cannot work as-is. It needs changes

internal/xds/translator/oidc.go
arkodg: yah lets go with 2. , if Envoy doesnt release by 07/20, we can back these out of the release

arkodg (approved):

arkodg (approved):

denniskniep (approved):

zirain (approved):

Issue #3178, opened 2024-04-11
Intermittent 'OAuth Flow Failure' potentially caused by reuse of old connections in pool


coro:
*Description*:

We are frequently seeing routes hosted by the Gateway fail with a `401`, returning an error `OAuth flow failed.`. Looking at the logs it seems this occurs where Envoy decides to re-use an existing long-lived connection to our OIDC issuer. Failures seem to correlate with Envoy picking connections that are in the order of 50 minutes old.

*Repro steps*:
Using the following SecurityPolicy:
```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  name: grafana-public
  namespace: default
spec:
  oidc:
    clientID: <ID>
    clientSecret:
      group: ""
      kind: Secret
      name: oauth-secret
    logoutPath: /logout
    provider:
      issuer: https://oauth.id.jumpcloud.com
    redirectURL: https://grafana.my.website/oauth2/callback
    scopes:
    - openid
    - email
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: grafana-public
```

Intermittently requests to this route that involve an OAuth flow via our IdP result in this error message. Refreshing the page solves this issue.

*Environment*:
Gateway: v1.0.0 installed via Helm
Envoy: `envoyproxy/envoy:distroless-v1.29.2`

*Logs*:
Filtering debug logs based off the `ConnectionId` shows it's trying to use a connection from about 50 mins earlier, and immediately getting a `remote close`:
```
2024-04-11T11:15:24.284912467Z stderr F [2024-04-11 11:15:24.284][13][debug][connection] [./source/common/network/connection_impl.h:98] [Tags: "ConnectionId":"631"] current connecting state: true
2024-04-11T11:15:24.284917407Z stderr F [2024-04-11 11:15:24.284][13][debug][client] [source/common/http/codec_client.cc:57] [Tags: "ConnectionId":"631"] connecting
2024-04-11T11:15:24.284924487Z stderr F [2024-04-11 11:15:24.284][13][debug][connection] [source/common/network/connection_impl.cc:1009] [Tags: "ConnectionId":"631"] connecting to 104.18.6.220:443
2024-04-11T11:15:24.284981748Z stderr F [2024-04-11 11:15:24.284][13][debug][connection] [source/common/network/connection_impl.cc:1028] [Tags: "ConnectionId":"631"] connection in progress
[...]

PR #3568, opened 2024-06-08
oidc: add tcp keep alive to idp

zhaohuabing:
This PR adds a `tcpkeepalive` to the IDP provider cluster of OIDC.

This is a workaround before retry is supported in Envoy upstream. https://github.com/envoyproxy/envoy/issues/33572

Fix: https://github.com/envoyproxy/gateway/issues/3178

### DIFF
--- a/internal/xds/translator/oidc.go
+++ b/internal/xds/translator/oidc.go
@@ -274,6 +274,8 @@ func createOAuth2TokenEndpointClusters(tCtx *types.ResourceVersionTable,
 			settings:     []*ir.DestinationSetting{ds},
 			tSocket:      tSocket,
 			endpointType: cluster.endpointType,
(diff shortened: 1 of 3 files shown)


arkodg:
id prefer to understand what is happening in the envoy side, and why its reusing the connection even after receiving an RST, so prefer if we collected cluster metrics and logs in debug/trace level before we proposed a fix

coro:
Screenshot 2024-06-18 at 17 08 53

Here is a look at the metrics when we encountered this issue multiple times over one day. It seems like there is a strong correlation between instances of `OAuth flow failed` errors and the `envoy_cluster_upstream_cx_destroy_remote_with_active_rq` metric increasing. I've attached logs in debug level in the linked issue.

PR #3555, opened 2024-06-06
chore: update make helm-readme-gen

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #3547

PR #3569, opened 2024-06-08
doc: add certgen description into chart readme

ShyunnY:
**What type of PR is this?**

doc: add certgen description into chart readme

ref: #3521

### DIFF
--- a/charts/gateway-helm/README.md
+++ b/charts/gateway-helm/README.md
@@ -59,11 +59,7 @@ To uninstall the chart:
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| certgen.job.annotations | object | `{}` |  |
-| certgen.job.resources | object | `{}` |  |
-| certgen.job.ttlSecondsAfterFinished | int | `30` |  |
-| certgen.rbac.annotations | object | `{}` |  |
-| certgen.rbac.labels | object | `{}` |  |
+| certgen | object | `{"job":{"annotations":{},"resources":{},"ttlSecondsAfterFinished":30},"rbac":{"annotations":{},"labels":{}}}` | Certgen is used to generate the certificates required by EnvoyGateway. If you want to construct a custom certificate, you can generate a custom certificate through Cert-Manager before installing EnvoyGateway. Certgen will not overwrite the custom certificate. Please do not manually modify `values.yaml` to disable certgen, it may cause EnvoyGateway OIDC,OAuth2,etc. to not work as expected. |
 | config.envoyGateway.gateway.controllerName | string | `"gateway.envoyproxy.io/gatewayclass-controller"` |  |
 | config.envoyGateway.logging.level.default | string | `"info"` |  |
 | config.envoyGateway.provider.type | string | `"Kubernetes"` |  |
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -66,6 +66,10 @@ createNamespace: false
 
 kubernetesClusterDomain: cluster.local
 
+# Certgen is used to generate the certificates required by EnvoyGateway.
+# If you want to construct a custom certificate, you can generate a custom certificate through Cert-Manager before installing EnvoyGateway.
+# Certgen will not overwrite the custom certificate.
+# WARNING: Please do not manually modify `values.yaml` to disable certgen, which may cause EnvoyGateway OIDC,OAuth2,etc. to not work as expected.
 certgen:
   job:
     annotations: {}
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -66,10 +66,7 @@ createNamespace: false
 
 kubernetesClusterDomain: cluster.local
 
-# Certgen is used to generate the certificates required by EnvoyGateway.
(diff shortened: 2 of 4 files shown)


ShyunnY:
PTAL
@zhaohuabing 
@arkodg

ShyunnY:
ci `gen-check`  block on #3555

charts/gateway-helm/README.md
```diff
@@ -55,6 +59,16 @@ To uninstall the chart:
...
 
+### Certgen
+
+Certgen is used to generate the certificates required by EnvoyGateway.
+
+Envoy Gateway establishes a secure TLS connection for control plane communication between Envoy Gateway pods and the Envoy Proxy fleet. The TLS Certificates used here are self signed and generated using a job that runs before envoy gateway is created, and these certs and mounted on to the envoy gateway and envoy proxy pods.
+
+If you want to construct a custom certificate, you can generate a custom certificate through Cert-Manager before installing EnvoyGateway. Certgen will not overwrite the custom certificate.
+
+> _WARNING:_ Please do not manually modify `values.yaml` to disable certgen, which may cause EnvoyGateway OIDC to not work as expected.
```
zhaohuabing, line 70: We don't need to emphasize this cause there's no configuration knob in values.yaml file that disables certgen.

charts/gateway-helm/README.md
```diff
@@ -55,6 +59,16 @@ To uninstall the chart:
     helm delete eg
 ```
 
+### Certgen
+
+Certgen is used to generate the certificates required by EnvoyGateway.
```
zhaohuabing, line 64: It also creates the HMAC secret for verifying OAuth2 cookies.

charts/gateway-helm/README.md
ShyunnY: I think we need to emphasize this: this is to prevent users from unknowingly disabling certgen functionality by manually removing the certgen Job.
We still need to keep it, but it is definitely not controlled by `values.yaml`, I will change this

charts/gateway-helm/README.md
ShyunnY: Thank you for your supplement!

tools/helm-docs/readme.gotmpl
```diff
@@ -51,6 +51,16 @@ To uninstall the chart:
     helm delete eg
 ```
 
+### Certgen
+
+Certgen is used to generate the certificates required by EnvoyGateway.
```
arkodg, line 56: this should live in https://github.com/envoyproxy/gateway/blob/33fceb05ca611e30107ff841cac9d7762eea2048/charts/gateway-helm/values.tmpl.yaml#L68 which then gets generated using `make helm-docs`

ShyunnY: OK, I added it in `values.tmpl.yaml`.

At the same time, I kept the description in `readme.gotmpl`: I think certgen Job determines whether some functions of eg can be used normally, so I hope to inform users of its importance through "**multiple emphasis**".

arkodg: Can we move all the doc strings into `values.tmpl.yaml` where the doc string is meant to live, the `.gotmpl` is only meant for editing presentation

ShyunnY: Sorry, I misunderstood. We should indeed move this description text to `values.tmpl.yaml`, I will change it!

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -66,6 +66,10 @@ createNamespace: false
 
 kubernetesClusterDomain: cluster.local
 
+# Certgen is used to generate the certificates required by EnvoyGateway.
+# If you want to construct a custom certificate, you can generate a custom certificate through Cert-Manager before installing EnvoyGateway.
+# Certgen will not overwrite the custom certificate.
+# WARNING: Please do not manually modify `values.yaml` to disable certgen, which may cause EnvoyGateway OIDC,OAuth2,etc. to not work as expected.
```
arkodg, line 72:
```suggestion
# WARNING: Please do not manually modify `values.yaml` to disable certgen, it may cause EnvoyGateway OIDC,OAuth2,etc. to not work as expected.
```

arkodg:
@ShyunnY you will also need to run `make helm-docs` to generate the o/p

arkodg:
weird how it didn't fail `geo-check`

ShyunnY:
> helm-docs

Does the `helm-docs` Target exist?
I don't seem to see it in my local main branch:
```shell
$ make helm-docs
make[1]: Entering directory '/home/shyunn/envoy-gateway'
make[1]: *** No rule to make target 'helm-docs'.  Stop.
make[1]: Leaving directory '/home/shyunn/envoy-gateway'
make: *** [Makefile:18: _run] Error 2
```

shawnh2:
it's `helm-readme-gen` which arko was referring to. but it seems cannot generate content from the comments in helm chart.

arkodg:
looks like `# -- <some text>` is the doc string format

ShyunnY:
> looks like `# -- <some text>` is the doc string format

OK, I have done that.
I checked the `helm-docs` tool. It seems that after we add the `# -- <comment>` comment, the object's fields will be folded into the Default column. I checked the corresponding documentation and it seems that there is no way to solve it. But this is not a problem, it just looks a bit strange.

shawnh2 (approved):
LGTM! Thanks!

arkodg (approved):
Thanks !

PR #3580, opened 2024-06-10
bug: ClientTrafficPolicy resources are prevented from targeting a SectionName by an incorrect CEL validation

liorokman:
Removed CEL validation that prevents specifying a `SectionName` for `ClientTrafficPolicy` resources.

The `ClientTrafficPolicy` translation does in-fact support targeting a specific `SectionName`, the CEL validation is wrong.


arkodg (approved):
LGTM thanks !

guydc (approved):

PR #3581, opened 2024-06-10
feat: support plural targetRefs on policies

### LINKED ISSUE #2999: Support attachment of policies to multiple Gateways

liorokman:
This PR adds support for plural `targetRefs` on the various policies.

Fixes #2999


api/v1alpha1/policy_helpers.go
```diff
@@ -0,0 +1,36 @@
...
+	// Deprecated: use targetRefs instead
+	TargetRef *gwapiv1a2.LocalPolicyTargetReferenceWithSectionName `json:"targetRef,omitempty"`
+
+	// TargetRefs are the names of the Gateway resources this policy
+	// is being attached to.
+	TargetRefs []gwapiv1a2.LocalPolicyTargetReferenceWithSectionName `json:"targetRefs,omitempty"`
+
+	// TargetSelector contains a selector for resources that this policy
+	// is attached to.
+	TargetSelector map[string]string `json:"targetSelector,omitempty"`
```
arkodg, line 28: wouldnt this need more fields so it can also incorporate `Kind` and `Group` ? 
```
      kind: HTTPRoute
      selector/matchLabels:
        app: foo
```

liorokman: While it's possible to add this, I don't think it's needed.

The selector is checked only against routes and gateways anyways right now, and if a developer/admin wants to limit the matched policies to a specific `kind` it's possible to add the `kind` as a label.

arkodg: as a project, we've tried to use strongly typed fields as much as possible, my suggestion would be to provide a `kind` and `group` field here along with the labelSelector field

liorokman: I think a `kind` and `group` field as part of the target selector defeats the purpose here. If I add a `kind` and `group` field, then how would that work?

Let's say I want to attach to all of my routes (`HTTPRoute`, `GRPCRoute`, `TLSRoute`, and `TCPRoute`) to add some default setting. What do I specify in the `kind` field?

And, considering we enforce that the `group` field is always a single value using CEL, what should I as a user do with the `group` field? Fill it, and make sure it always matches the exactly one valid value? Leave it empty and assume that the default single value is being filled in?

arkodg: reg attaching to all xRoutes, 4 policies could be created per xRoute or 1 policy could be attached to the Gateway

charts/gateway-addons-helm/README.md
shawnh2, line 1: can you rebase ? the changes here are unintentional

internal/gatewayapi/translator_test.go
```diff
@@ -507,7 +507,7 @@ func TestTranslateWithExtensionKinds(t *testing.T) {
 
 func overrideOutputConfig(t *testing.T, data string, filepath string) {
 	t.Helper()
-	file, err := os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o755)
+	file, err := os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
```
shawnh2, line 510: should we change this in a separate PR? It causes a lot of file changes.

charts/gateway-addons-helm/README.md
liorokman: done.

internal/gatewayapi/translator_test.go
liorokman: This change didn't cause the file changes. I explicitly `chmod`-ed the yaml files which were marked as executable so that they wouldn't be.

This change only makes sure that new files being added are created without the executable bit set.

liorokman:
Not sure why the docs-lint build step is not successful. It's got something to do with some other change?

```
[404] https://gateway.envoyproxy.io/latest/tasks/traffic/http-routing/
[404] https://gateway.envoyproxy.io/latest/tasks/security/secure-gateways/
[404] https://gateway.envoyproxy.io/latest/tasks/traffic/grpc-routing/
[404] https://gateway.envoyproxy.io/latest/tasks/traffic/http-traffic-splitting/
[404] https://gateway.envoyproxy.io/latest/tasks/traffic/global-rate-limit/
```

api/v1alpha1/policy_helpers.go
arkodg: sure raised https://github.com/envoyproxy/gateway/issues/3607 to track this work

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -64,149 +64,167 @@ func (t *Translator) ProcessBackendTrafficPolicies(backendTrafficPolicies []*egv
 	// Map of Gateway to the routes attached to it
 	gatewayRouteMap := make(map[string]sets.Set[string])
 
+	handledPolicies := make(map[types.NamespacedName]*egv1a1.BackendTrafficPolicy)
```
arkodg, line 67: curious why we need this map ? if its being used to handle duplicates, any idea how we ended up with duplicates ?

liorokman: For example:

There's a gateway with 3 different routes. You create a single policy resource that targets two of those three routes with two different `targetRef` elements.

This map is a way to prevent the policy resource from being added twice to the `res` array.

arkodg: re-reading, its because we have multiple targetRefs, can we execute `res = append(res, policy)` right after
```
for _, currPolicy := range backendTrafficPolicies {
    res = append(res, curPolicy)
    ....
```
and avoid extra memory

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -38,9 +38,7 @@ func (t *Translator) processBackendTLSPolicy(
 		return nil, nil
 	}
 
-	ancestorRefs := []gwapiv1a2.ParentReference{
-		parent,
-	}
+	ancestorRefs := fixCurrentAncestorRefs(policy, parent)
```
arkodg, line 41: nit: prefer
```
ancestorRefs := getAncestorRefs(policy)
ancestorRefs = append(ancestorRefs, parent)
```

internal/gatewayapi/backendtrafficpolicy.go
liorokman: It's not that simple. The code has two loops, the first handling the policies that target routes and the second handling the targets that target gateways.  Once we have multiple `targetRefs`, it's possible for the same policy to target both a route and a different gateway.

The second loop needs to work on the same deep copy of the policy as the first loop, otherwise the `status` section of the policy is not correctly updated. If we do as you suggest in the first loop, then the second loop would need to go over all of the content of `res` and find the correct policy instance if its there. The easy way to do this is like the code does right now.

arkodg: ah thanks for explaining
do we want to support a policy targeting a `Gateway` and `HTTPRoute` at the same time ? does our library gracefully handle this case when it comes to defaults and overrides ?

liorokman: It's no different than two different policies targeting both the gateway and the route. The plural `targetRefs` is just a form of syntactic sugar.

arkodg (approved):
LGTM thanks !

guydc (approved):

PR #3596, opened 2024-06-12
feat(translator): Implement http health check filter

### LINKED ISSUE #3258: HealthCheck config for HTTP Downstream

aoledk:

Follow #3540 

Fixes #3258


arkodg:
thanks @aoledk this is great, can you also add an e2e for this that pings the healthCheck endpoint and gets a `200` back ?

test/e2e/tests/health_check.go
```diff
@@ -0,0 +1,72 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+
+	"github.com/envoyproxy/gateway/internal/gatewayapi"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HealthCheckTest)
+}
+
+var HealthCheckTest = suite.ConformanceTest{
```
shawnh2, line 28: can we rename this into `DownstreamHealthCheck`? Since this health check is in CTP, as we also have health check in BTP, and the `UpstreamHealthCheck` e2e test case will be added soon.

shawnh2 (commented):
LGTM! Just one comment on the naming.

test/e2e/tests/health_check.go
arkodg: +1 prefer `ListenerHealthCheck`

arkodg (dismissed):
LGTM thanks !

shawnh2 (approved):
LGTM! the linter need a fix :)

arkodg (approved):

Issue #3318, opened 2024-05-02
Support Route and VHost Metadata


guydc:
*Description*:
In Envoy, static [metadata](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-metadata) can be configured on virtual hosts, routes, clusters, etc. 

Static metadata can be used for various purposes:
- Enrichment of access logs with [metadata-based formatters](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/formatter/metadata/v3/metadata.proto.html#formatter-extension-for-printing-various-types-of-metadata-proto)
- Enrichment of traces with [metadata-based tags](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/tracing/v3/custom_tag.proto.html#envoy-v3-api-msg-type-tracing-v3-customtag-metadata)
- Providing configuration context for filters 
- Rate-limiting requests based on [metadata actions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#envoy-v3-api-msg-config-route-v3-ratelimit-action)
- Loadbalancing to a subset of endpoints that match [some metadata](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/subsets#example-route-with-metadata-criteria)
- [Transport socket matching](https://envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-msg-config-cluster-v3-cluster-transportsocketmatch)

In the context of Envoy Gateway, route metadata can be beneficial for observability and configuration of custom data-plane extensions, as described above.  

Envoy gateway can support route metadata using:
- A vendor-specific Gateway API [HTTP Route Filter](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteFilter). This is possibly a misuse of the filter mechanism, since there is no additional processing of traffic.
- A new or existing Envoy Gateway Policy
- An extended Gateway API HTTPRoute resources that supports route metadata, e.g. using [annotations](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.AnnotationValue).

AliceProxy:
:100: This type of stuff is super useful to get metadata per route or per ext_authz filter, etc. Makes it a lot easier to provide context about things like:
[...]

PR #3602, opened 2024-06-12
feat(translator): xds route and vhost metadata

guydc:
- Design doc for metadata propagation 
- Implement propagation of `HTTPRoute`/`GRPCRoute` metadata to envoy route resources
- Implement propagation of `Gateway` and `Listener` to Virtual Host metadata
- e2e test based on access logs - verifying that metadata is printed to the log. If injection of METADATA-based header is supported (https://github.com/envoyproxy/envoy/issues/34671), this test can be simplified to make an assertion on generated headers. 

Relates to #3318


internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
```diff
@@ -1082,6 +1082,13 @@ xds:
             routes:
             - match:
                 prefix: /
+              metadata:
+                filterMetadata:
+                  io.envoyproxy.gateway.metadata:
```
zhaohuabing, line 1087: Would we need to add other metadata in the future? If so, we might need another level within the metadata structure to accommodate them.

zhaohuabing, line 1087: `metadata.filterMetadata.io.envoyproxy.gateway.metadata` seems a bit redundant, maybe just use`io.envoyproxy.gateway` as the key?

guydc: I changed the field to be a list of resources for future proofing.

guydc: Actually, I'm having second thoughts about a list here. It's not so easy to address dynamic metadata in a list. So, I think that I'll go with your approach, but maybe slightly generalize to `Gateway`, `Route` and `Backend` top-level resources, which can then be GW/GWC, HTTP/GRPCRoute, Service/ServiceImport/Backend.

site/content/en/contributions/design/metadata.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+Envoy Gateway translates [Gateway-API][gw_api] in the following manner:
+- [Gateway][gw] metadata is propagated to envoy [listener][] metadata. If merge-gateways is enabled, [Gateway Class][gc] is used instead. 
+- [Gateway][gw] metadata and Listener Section name are propagated to envoy [virtual host][] metadata
+- [HTTPRoute][httpr] and [GRPCRoute][grpcr] metadata is propagated to envoy [route][] metadata. When Gateway-API adds support [named route rules][], the route rule name
+- [TCP/UDPRoute][tcpr] and [TLSRoute][tlsr] resource attributes are not propagated. These resources are translated to envoy filter chains, which do not currently support static metadata. 
+- [Service][svc], [ServiceImport][svci] and [Backend][] metadata and port name are propagated to envoy [cluster] metadata.
+
+## Usage
```
shawnh2, line 54: about `usgae`, can we also highlight this in the https://gateway.envoyproxy.io/latest/tasks/observability/proxy-observability/#logs ?

internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
zhaohuabing: I just thought the metadata structure should be able to accommodate multiple resources, but I don't know how exactly the structure should look like. Let's brainstorm in the next meeting?

internal/ir/xds.go
```diff
@@ -2251,3 +2253,20 @@ type DestinationFilters struct {
 	// RemoveResponseHeaders defines a list of headers to be removed from response.
 	RemoveResponseHeaders []string `json:"removeResponseHeaders,omitempty" yaml:"removeResponseHeaders,omitempty"`
 }
+
+// ResourceMetadata is metadata from the provider resource that is translated to an envoy resource
+// +k8s:deepcopy-gen=true
+type ResourceMetadata struct {
```
arkodg, line 2259: thoughts on leaving out `SectionName` and `GroupVersion` in the first iteration ?

internal/xds/translator/route.go
```diff
@@ -15,22 +15,32 @@ import (
...
 )
 
 const (
-	retryDefaultRetryOn             = "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes"
-	retryDefaultRetriableStatusCode = 503
-	retryDefaultNumRetries          = 2
+	retryDefaultRetryOn                 = "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes"
+	retryDefaultRetriableStatusCode     = 503
+	retryDefaultNumRetries              = 2
+	envoyGatewayMetadataNamespace       = "io.envoyproxy.gateway"
```
arkodg, line 29: should this be `envoy-gateway` instead ?

internal/xds/translator/route.go
```diff
@@ -15,22 +15,32 @@ import (
...
+	envoyGatewayMetadataKeyGroupVersion = "groupVersion"
+	envoyGatewayMetadataKeyKind         = "kind"
+	envoyGatewayMetadataKeyName         = "name"
+	envoyGatewayMetadataKeyNamespace    = "namespace"
+	envoyGatewayMetadataKeyAnnotations  = "annotations"
+	envoyGatewayMetadataKeySectionName  = "sectionName"
+	envoyGatewayMetadataKeyRoute        = "route"
```
arkodg, line 36: to which resource are we building this key off ? input gateway api, IR or xds filter ?

arkodg:
@zirain @guydc thoughts on adding this metadata by default to our default access log definition ?

internal/ir/xds.go
guydc: will remove `GroupVersion`. I'll add VHost to this PR in a bit, and it'll include a useful `SectionName`, so prefer to keep it.

internal/xds/translator/route.go
guydc: will change to `envoyGatewayXds*` to make it clear that this is a const used in generated XDS.

internal/xds/translator/route.go
guydc: > Several parts of Envoy configuration (e.g. listeners, routes, clusters) contain a [metadata](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-metadata) where arbitrary key-value pairs can be encoded. The typical pattern is to use the filter names in reverse DNS format as the key and encode filter specific configuration metadata in the value.

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/data_sharing_between_filters#metadata

However, I agree that:
- There is no filter in this case, so, we're not bound by these conventions. Istio uses a namespace called `istio` (and not `io.istio`) for similar purposes. 
- the long name is not convenient for access log formatters accessing the relevant keys.

zirain:
> @zirain @guydc thoughts on adding this metadata by default to our default access log definition ?

I don't think we should add them by default, users who want can add the command.

zhaohuabing:
> @zirain @guydc thoughts on adding this metadata by default to our default access log definition ?

IMO, adding them to the default access log format would be a nice surprise to EG users.

zhaohuabing (approved):
LGTM. This can be super helpful for EG users for debugging. I strongly suggest adding the metadata to the default log format - hiding this by default is a huge waste.

internal/gatewayapi/route.go
```diff
@@ -31,6 +31,8 @@ const (
 	// Request timeout, which is defined as Duration, specifies the upstream timeout for the route
 	// If not specified, the default is 15s
 	HTTPRequestTimeout = "15s"
+	// egPrefix is a prefix of annotation keys that are processed by Envoy Gateway
```
arkodg, line 34: unused ?

internal/xds/translator/metadata.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const (
+	envoyGatewayXdsMetadataNamespace      = "envoy-gateway"
+	envoyGatewayXdsMetadataKeyKind        = "kind"
+	envoyGatewayXdsMetadataKeyName        = "name"
+	envoyGatewayXdsMetadataKeyNamespace   = "namespace"
+	envoyGatewayXdsMetadataKeyAnnotations = "annotations"
+	envoyGatewayXdsMetadataKeySectionName = "sectionName"
+	envoyGatewayXdsMetadataKeyRoute       = "route"
```
arkodg, line 22: why are these keys needed ?
as a consumer I may want to access the `envoy-gateway` metadata, but I may not know if the xds resource maps to a `gateway` or `route` or both

guydc:
Sorry for force-push, I messed up the recent commit. 

We will go with the list approach, since it's most future proof, considering situations where multiple resources of the same kind (e.g. policy and overriding policy) can contribute metadata.

internal/gatewayapi/route.go
guydc: used in `filterEGPrefix` to filter the annotations that should be picked up for metadata.

arkodg: how do we document this opt in behavior ? user guide ?

internal/gatewayapi/route.go
```diff
@@ -719,6 +723,28 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
...
+		Namespace:   route.GetNamespace(),
+		Annotations: filterEGPrefix(route.GetAnnotations()),
+	}
+}
+
+func filterEGPrefix(in map[string]string) map[string]string {
+	out := map[string]string{}
+	for k, v := range in {
+		if strings.HasPrefix(k, egPrefix) {
+			out[strings.TrimPrefix(k, egPrefix)] = v
```
arkodg, line 739: should we trim the prefix or let it be ?

guydc: It feels a bit redundant to me, since only annotations with this prefix will be included.

arkodg: cool, lets leave as is

arkodg (approved):
LGTM thanks !
super excited to use this feature

zhaohuabing (approved):
🚀

PR #3608, opened 2024-06-14
docs: ext-proc

### LINKED ISSUE #3456: docs: Ext Proc
*Description*:
>Describe the issue.

Add docs highlighting the Ext Proc feature

Relates to https://github.com/envoyproxy/gateway/issues/2546

[optional *Relevant Links*:]

guydc:
docs for the ext-proc feature
Fixes #3456


site/content/en/latest/tasks/extensibility/ext-proc.md
```diff
@@ -0,0 +1,283 @@
...
+
+```shell
+kubectl get httproute/myapp -o yaml
+```
+
+### Configuration
+
+Create a new EnvoyExtensionPolicy resource to configure the external processing service. This EnvoyExtensionPolicy targets the HTTPRoute
+"myApp" created in the previous step. It calls the GRPC external processing service "grpc-ext-proc" on port 9002 for
+processing. The `processingMode` struct defines that headers of requests and responses will be sent to the external processing server.
```
arkodg, line 99: may a line in here saying, `by default all request headers are sent, but in this example we are sending none` would be helpful

site/content/en/latest/tasks/extensibility/ext-proc.md
```diff
@@ -96,7 +96,10 @@ kubectl get httproute/myapp -o yaml
 
 Create a new EnvoyExtensionPolicy resource to configure the external processing service. This EnvoyExtensionPolicy targets the HTTPRoute
 "myApp" created in the previous step. It calls the GRPC external processing service "grpc-ext-proc" on port 9002 for
-processing. The `processingMode` struct defines that headers of requests and responses will be sent to the external processing server.
+processing. 
+
+By default, requests and responses are not sent to the external processor. The `processingMode` struct is used to define what should be sent to the external processor.
```
arkodg, line 101: aren't request headers sent by default ?
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/processing_mode.proto#envoy-v3-api-msg-extensions-filters-http-ext-proc-v3-processingmode

arkodg: maybe we should explicitly mention `body` so readers like me dont confuse it with `headers`

guydc: yes, but not in EG. We wanted to simplify the API (avoid explicitly mentioning header processing modes) while still supporting opt-out for request and/or response processing and explicit body processing mode configuration. 

To accomodate this, the API currently behaves like this:
1. if request/response is set to an empty struct - headers are sent. 
2. if request/response is configured with a body - headers and body are sent.
3. is request/response is not configured at all - opt-out of calling the processor. 

I can extend the example to also include body forwarding in one of the flows to help explain how this works.

arkodg: thanks for explaining this, in that case, can we update above text mentioning that request headers will be sent ?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #3628, opened 2024-06-19
fix: do not propagate well-known port number for xds portRedirect

### LINKED ISSUE #3589: Redirects add port to Location header
### Description

When setting up any redirect, e.g. following the initial example at https://gateway.envoyproxy.io/v1.0.1/tasks/traffic/http-redirect/ (though removing the `backendRefs` which aren't actually allowed together with a redirect), Envoy Gateway will add an explicit port (in this case 443) to the location header:

```
$ curl -L -v --header "Host: redirect.example" "http://localhost:8888/get"
...
< HTTP/1.1 301 Moved Permanently
< location: https://www.example.com:443/get
```

(The example curl output in the docs actually doesn't show the port but that's not the behavior I observe.)

This is counter to the recommendation given by the Gateway API which on the `ports` field of the redirect filter (https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRequestRedirectFilter) states:

> Implementations SHOULD NOT add the port number in the ‘Location’ header in the following cases:
> 
> - A Location header that will use HTTP (whether that is determined via the Listener protocol or the Scheme field) and use port 80.
> - A Location header that will use HTTPS (whether that is determined via the Listener protocol or the Scheme field) and use port 443.

This happens even if the `port` field isn't actually set in the redirect and even if it's only a path redirect, with no change to the hostname, scheme, or port.

Using `egctl`, it looks like the following Envoy configuration is generated for the route for the example from the docs:

```
              redirect:
                portRedirect: 443
                schemeRedirect: https
```

Presumably, the `portRedirect` shouldn't be set in this case.

This probably doesn't matter in most cases since browsers apparently silently remove the port when following the redirect but some http clients don't which for us then lead to an issue in combination with OIDC because the URL with the port wasn't registered as a valid callback URL.

### Environment

- Envoy Gateway v1.0.1
- Envoy v1.29.3

shawnh2:
**What type of PR is this?**

Fixes #3589

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -345,7 +345,9 @@ func buildXdsRedirectAction(httpRoute *ir.HTTPRoute) *routev3.RedirectAction {
 	if redirection.Hostname != nil {
 		routeAction.HostRedirect = *redirection.Hostname
 	}
-	if redirection.Port != nil {
+	// Ignore the redirect port if it is a well-known port number, in order to
+	// prevent the port be added in the response's location header.
+	if redirection.Port != nil && *redirection.Port != 80 && *redirection.Port != 443 {
 		routeAction.PortRedirect = *redirection.Port
 	}
 	if redirection.StatusCode != nil {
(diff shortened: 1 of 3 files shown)


arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #3592, opened 2024-06-11
Consider support samplers and resource_detectors for OpenTelemetry tracer


zirain:
xref: [Samplers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/opentelemetry/samplers) [Resource Detectors](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/opentelemetry/resource_detectors)

I cannot recall whether EG support custom env flags to support resource detector.

zirain:
If Otel-collector is injected with linkerd for mTLS which broken the k8sattributes processor.
In this case, we need resource detector to simply the injection of resource attributes.

A workaround is using transform processor in otel-collector.
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: otel-telemetry
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      port: 80
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: All
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: otel-telemetry
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: otel-telemetry
spec:
  telemetry:
    tracing:
      provider:
        backendRefs:
          - name: otel-collector
            namespace: monitoring
            port: 4317
      tags:
        "k8s.cluster.name": "envoy-gateway"
        "k8s.pod.ip": "%ENVIRONMENT(ENVOY_POD_IP)%"
        "k8s.pod.uid": "%ENVIRONMENT(ENVOY_POD_UID)%"
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        container:
          env:
            - name: ENVOY_POD_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
            - name: ENVOY_POD_UID
              valueFrom:
                fieldRef:
                  fieldPath: metadata.uid
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: otel-telemetry
spec:
  parentRefs:
    - name: otel-telemetry
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /otel
      filters:
        - type: URLRewrite
          urlRewrite:
            hostname: www.httpbin.org
            path:
              type: ReplaceFullPath
              replaceFullPath: /headers
[...]

PR #3631, opened 2024-06-19
api: support OpenTelemetry Resource Detectors

zirain:
xref: #3592


arkodg:
@zirain can you highlight with some example how and why this feature can be used ?

zirain:
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.24.0/specification/resource/sdk.md#detecting-resource-information-from-the-environment

zirain:
close this before there more resource detectors introduced into upstream.

PR #3647, opened 2024-06-21
fix: upgrade test

### LINKED ISSUE #3311: flaky: TestEGUpgrade/EGUpgrade
https://github.com/envoyproxy/gateway/actions/runs/8896241845/job/24429998115

guydc:
- Add retries to locating charts from repo
- update to latest tag
Fixes #3311


arkodg:
the wasm e2e is flaky

arkodg (approved):

arkodg:
@guydc this maybe feeling due to the bump from v1alpha2 to v1alpha3 for BackendTLSPolicy, we'll need to add the upgrade steps in our release notes similar to https://gateway-api.sigs.k8s.io/guides/?h=upgrade#backendtlspolicy that will also need to be codified In the test

guydc:
> @guydc this maybe feeling due to the bump from v1alpha2 to v1alpha3 for BackendTLSPolicy, we'll need to add the upgrade steps in our release notes similar to https://gateway-api.sigs.k8s.io/guides/?h=upgrade#backendtlspolicy that will also need to be codified In the test

Good Point. Actually, the GW-API CRD version is an important aspect here, and we maybe need to rethink upgrade test strategy. 

In practice, we shouldn't/can't run tests from main against a cluster that has latest-rel installed:
- The conformance suite utils will fail on a version compatibility check between the bundled and installed CRD versions
- Even if the check was not in place, in reality, we can run into all sorts of mismatch issues, even for a simple test that just verifies a basic scenario.

PR #3650, opened 2024-06-22
refactor: refactor client/backend connection

### LINKED ISSUE #3445: Discussion: Refactoring Connection in BTP and CTP
***Description***:

We can uniformly abstract the connection feat in BTP and CTP and set them in `api/v1alpha1/connection_types.go`.

The current `connection` design is as follows. This does not seem to be convenient for us to design corresponding `connection` for different policies:
https://github.com/envoyproxy/gateway/blob/7d743067b22119056a357c3927fb302089f95a72/api/v1alpha1/connection_types.go#L14-L27

It seems that it would be better if we refactor in the following way:
```go
type struct BackendConnection {
  ...
}

type struct ClientConnection {
  ...
}

```

[optional ***Relevant Links***:]
Related PR: #3390

ShyunnY:
Fixes #3445


ShyunnY:
`BufferLimit` exists in both client and backend. Do we need to abstract it into a struct to provide sharing?

shawnh2:
I think this is unnecessary, keep api straight-forward and simple is good.

shawnh2 (approved):
LGTM thanks!

yuluo-yx (approved):

arkodg (approved):

PR #3652, opened 2024-06-22
docs: introduce shortcode boilerplate

zirain:
Introduce a new shortcode `boilerplate` to simply the usage.

the shortcode took from [istio/istio.io](https://github.com/istio/istio.io/blob/master/layouts/shortcodes/boilerplate.html)


shawnh2 (approved):
thanks for introducing this

PR #3653, opened 2024-06-23
docs: remove docs-api-headings

zirain:
use crd-ref-docs templates


shawnh2 (approved):
thanks for cleaning this!

arkodg (approved):

PR #3664, opened 2024-06-24
FEAT: jwt support custom CA

phantooom:
**What type of PR is this?**

* "FEAT: jwt support custom CA"

Fixes  https://github.com/envoyproxy/gateway/issues/3536


guydc:
Should we maybe turn `RemoteJWKS` to a proper backend reference and then use `BackendTLSPolicy` here?

arkodg:
@guydc we've followed the approach of directly defining the `certificateRef` for the ratelimit redis backend https://gateway.envoyproxy.io/v1.0.2/api/extension_types/#redistlssettings because its unlikely a k8s service and authoring 2 extra resources may be cumbersome for the user

PR #3630, opened 2024-06-19
api: Adding Zipkin Tracing support

alexandermarston:
**What type of PR is this?**
API to support Zipkin provider for Tracing telemetry.

**What this PR does / why we need it**:
At present, Envoy Gateway only supports the OpenTelemetry tracing provider.

This PR updates the underlying APIs to also support the Zipkin provider.

In addition, it also exposes a new option to specify provider specific configuration such as TraceId128Bit and SharedSpanContext for Zipkin.

**Which issue(s) this PR fixes**:
Related to #1827

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/zipkin.proto

PR #3668, opened 2024-06-24
feat: implement zipkin tracing

### LINKED ISSUE #1827: Enable selecting zipkin tracing provider
*Description*:
Currently Envoy Gateway only supports the OpenTelemetry tracer, this makes EG unsuitable to plug into platforms where other solutions are being used (especially regarding trace context propagation). Some may want to transition to OpenTelemetry in the future but others have significant investments in other solutions (often using the B3 context propagation standard from the Zipkin ecosystem) or use tracing ecosystems that simply can't be supported by OpenTelemetry (e.g. Apache SkyWalking).

As an ingress typically is the location where also the sampling decision is made, it is very important to support at least the core instrumentation that Envoy supports. At the very least start with Zipkin (for B3 propagation) and SkyWalking (native instrumentation is not OTel compatible)

alexandermarston:
**What type of PR is this?**
feat: implement zipkin tracing

This PR adds support for the Zipkin tracer.

Fixes #1827 

API Change: #3630


internal/xds/translator/tracing.go
```diff
@@ -23,26 +24,61 @@ import (
...
+			return protocov.ToAnyWithError(config)
+		}
+	case egv1a1.TracingProviderTypeZipkin:
+		providerName = envoyZipkin
+
+		providerConfig = func() (*anypb.Any, error) {
+			config := &tracecfg.ZipkinConfig{
+				CollectorCluster:         tracing.Destination.Name,
+				CollectorEndpoint:        "/api/v2/spans",
+				TraceId_128Bit:           *tracing.Provider.Zipkin.Enable128BitTraceID,
```
arkodg, line 68: will crash if ptr is null

arkodg: you'll need to use something like `ptr.Deref(<>, false)` as well as on L69

alexandermarston: oversight on my part, should be handled now

arkodg (approved):
LGTM thanks !

zirain (approved):
LGTM, thanks for adding this,
Would be better to add e2e test in the following.

internal/xds/translator/testdata/out/xds-ir/tracing-zipkin.clusters.yaml
```diff
@@ -0,0 +1,49 @@
...
+        region: tracing-0/backend/0
+  name: tracing-0
+  outlierDetection: {}
+  perConnectionBufferLimitBytes: 32768
+  respectDnsTtl: true
+  type: STRICT_DNS
+  typedExtensionProtocolOptions:
+    envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+      '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+      explicitHttpConfig:
```
zirain, line 48: this's not right, zipkin is HTTP.

Issue #3667, opened 2024-06-24
Support CEL Expressions in Access Logs


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Add the ability to log access logs only when specific conditions are met, mainly to reduce logging costs

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/filters/cel/v3/cel.proto

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

guydc:
+1. 
Is this just about CEL, or do we want to support additional filters? I think that status code and response flag filters are also pretty useful.

arkodg:
@guydc the status code and the response flag can we represented using CEL `response.code > 500` with the entire supported list in https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#request-attributes

guydc:
Ah, yea, makes sense. Let's keep it CEL for only for now. Thanks!

zirain:
still need doc and e2e for this.

PR #3669, opened 2024-06-24
api: support AccessLog filter

zirain:
xref: https://github.com/envoyproxy/gateway/issues/3667

if we plan to only support CEL filter, the API could be more simply.


api/v1alpha1/accesslogging_types.go
```diff
@@ -19,12 +19,39 @@ type ProxyAccessLog struct {
 type ProxyAccessLogSetting struct {
 	// Format defines the format of accesslog.
 	Format ProxyAccessLogFormat `json:"format"`
+	// Filters defines the extensions filters of accesslog.
+	// +notImplementedHide
+	Filters []ProxyAccessLogFilter `json:"filters,omitempty"`
```
arkodg, line 24: thoughts on ?
```
matches:
- "response.code >= 400"
- "(connection.mtls && request.headers['x-log-mtls'] == 'true') || request.url_path.contains('v1beta3')"
```

zirain: if we plan to support CEL only, it's fine.

arkodg: looks like I was able to convince @guydc https://github.com/envoyproxy/gateway/issues/3667#issuecomment-2187363606 :) 

also should the noun be `filters` or `matches`

api/v1alpha1/accesslogging_types.go
```diff
@@ -19,6 +19,10 @@ type ProxyAccessLog struct {
 type ProxyAccessLogSetting struct {
 	// Format defines the format of accesslog.
 	Format ProxyAccessLogFormat `json:"format"`
+	// Matches defines the match conditions for accesslog in CEL expression.
+	// Invalid expressions will be ignored.
```
arkodg, line 23: lets add a link in here to highlight all the available options

api/v1alpha1/accesslogging_types.go
```diff
@@ -19,6 +19,10 @@ type ProxyAccessLog struct {
 type ProxyAccessLogSetting struct {
 	// Format defines the format of accesslog.
 	Format ProxyAccessLogFormat `json:"format"`
+	// Matches defines the match conditions for accesslog in CEL expression.
```
arkodg, line 22:
```suggestion
	// Matches defines the match conditions for accesslog in CEL expression.
	// An accesslog will be emitted only when one or more match conditions are evaluated to true.
```

api/v1alpha1/accesslogging_types.go
```diff
@@ -19,6 +19,10 @@ type ProxyAccessLog struct {
...
+	// Invalid [CEL](https://www.envoyproxy.io/docs/envoy/latest/xds/type/v3/cel.proto.html#common-expression-language-cel-proto) expressions will be ignored.
```
arkodg, line 23: can we link to a page that shows the CEL examples ? along with all available attributes

zirain: it's on that page
image

api/v1alpha1/accesslogging_types.go
```diff
@@ -20,6 +20,7 @@ type ProxyAccessLogSetting struct {
 	// Format defines the format of accesslog.
 	Format ProxyAccessLogFormat `json:"format"`
 	// Matches defines the match conditions for accesslog in CEL expression.
+	// An accesslog will be emitted only when one or more match conditions are evaluated to true.
```
arkodg, line 23: curious if we can use CEL to validate the CEL 🙃  ? Or will it be hard keeping track of upstream OSS attributes?

arkodg (approved):
LGTM thanks !

api/v1alpha1/accesslogging_types.go
zirain: TBH, I don't know, mostly can't.

zhaohuabing (approved):
🚀

Issue #2417, opened 2024-01-07
Add e2e for active health checking


lemonlinger:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

shawnh2:
closed in favor of #3677

PR #3677, opened 2024-06-26
e2e: Add backend health check e2e case via active http

aoledk:

Check whether backend health check via active http works in e2e, inspired by https://github.com/projectcontour/contour/blob/main/test/e2e/httpproxy/http_health_checks_test.go.

EG use the default `50%` from Envoy as cluster [panic threshold](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/panic_threshold). In case of health check fail, all endpoints in cluster become unhealthy and panic mode is activated, request is distributed to all unhealthy `echo-basic` endpoints and get response from `echo-basic`, with same behavior as case of health check pass.

To make those two cases distinguishable, in case of health check fail, Envoy should not distribute request to unhealthy endpoint, respond with **503** `no healthy upstream` directly. This requires cluster panic mode is disabled.

Since EG doesn't have knob to configure cluster panic threshold currently, I use EnvoyPatchPolicy to disable cluster panic mode in this PR.

Issues about panic threshold in _contour_:
https://github.com/projectcontour/contour/issues/579.
https://github.com/projectcontour/contour/issues/1846

xref #2417


arkodg:
thanks for picking this one up @aoledk !
would be great to also make sure we can check if the `membership_degraded` stat https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats is also bumped
similar to whats done here https://github.com/ShyunnY/gateway/blob/c74f4fe7e7d9cf258e6d013934bffce85bbc9e9d/test/e2e/tests/connection_limit.go#L70

aoledk:
We can use `membership_healthy` to tell whether active http health check works, instead of `membership_degraded` which requires upstream host to respond with special `x-envoy-degraded` header [^1], IMO it's out of this issue's scope.

- In case of health check pass, `membership_healthy` should > 0.
- In case of health check fail, `membership_healthy` should == 0.

```
# TYPE envoy_cluster_membership_degraded gauge
envoy_cluster_membership_degraded{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 0
envoy_cluster_membership_degraded{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 0

# TYPE envoy_cluster_membership_excluded gauge
envoy_cluster_membership_excluded{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 0
envoy_cluster_membership_excluded{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 0

# TYPE envoy_cluster_membership_healthy gauge
envoy_cluster_membership_healthy{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 0
envoy_cluster_membership_healthy{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 2

# TYPE envoy_cluster_membership_total gauge
envoy_cluster_membership_total{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 2
envoy_cluster_membership_total{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 2
```

[^1]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/health_checking#arch-overview-health-checking-degraded

arkodg:
`membership_health` works, thanks !

test/e2e/tests/backend_health_check.go
```diff
@@ -46,6 +50,66 @@ var BackendHealthCheckActiveHTTPTest = suite.ConformanceTest{
...
+				func(_ time.Duration) bool {
+					// check membership_healthy stats from Prometheus
+					v, err := prometheus.QuerySum(promAddr, passPromQL)
+					if err != nil {
+						// wait until Prometheus sync stats
+						return false
+					}
+					t.Logf("cluster pass health check: membership_healthy stats query count: %v", v)
+
+					if v == 0 {
```
arkodg, line 80: shouldn't this be non zero ?

arkodg: you prob need to wait until non zero is reached

aoledk: I've turned to use non-zero cluster health check stats to verify whether backend health check works, for both pass and fail cases. And undo changes in `e2e/utils/prometheus.sum` to restore the behavior of only return `nil` when non-zero value is queried.

```
# TYPE envoy_cluster_health_check_attempt counter
envoy_cluster_health_check_attempt{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 2
envoy_cluster_health_check_attempt{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 2

# TYPE envoy_cluster_health_check_success counter
envoy_cluster_health_check_success{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 0
envoy_cluster_health_check_success{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 2

# TYPE envoy_cluster_health_check_failure counter
envoy_cluster_health_check_failure{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-fail/rule/0"} 2
envoy_cluster_health_check_failure{envoy_cluster_name="httproute/gateway-conformance-infra/http-with-health-check-active-http-pass/rule/0"} 0
```

zirain (approved):

PR #3681, opened 2024-06-26
DRAFT: ClientIPDetectionSettings: Add InternalAddressConfig

### LINKED ISSUE #3598: envoy-gateway: Additional configuration for determining internal address
*Description*
The http request header `x-forwarded-for` (XFF) is not beeing set when the request is origination from ip-ranges that envoy detects as internal. By default any request from private ip ranges ([RFC1918](https://datatracker.ietf.org/doc/html/rfc1918) / [RFC4193](https://datatracker.ietf.org/doc/html/rfc4193)) are considered internal. Envoy's list of internal CIDR ranges can be configured with the [internal address configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-internaladdressconfig) in the http connection manager.

When running the envoy gateway in an internal network, where the client will connect directly to the envoy-gateway without network address translation from a private ip addresses, it would still be interesting to know the client's IP address. When the XFF header is not set, the backend will only have the ip address of the proxying gateway, and not the actual client.

Please provide guidance for how to configure the envoy-gateway with custom CIDR ranges to consider as internal. Ideally it should be possible to set this via helm values, or at least via relevant kubernetes configuration objects. In the meantime I would also be interested in ways to configure this via low level interfaces, e.g. [overriding the bootstrap configuration](https://gateway.envoyproxy.io/latest/tasks/operations/customize-envoyproxy/#customize-envoyproxy-bootstrap-config) or any other way, if that is possible?

*Relevant Links*
 - API spec [HttpConnectionManager.InternalAddressConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-internaladdressconfig)
 - Upstream issue [envoyproxy/envoy#19844](https://github.com/envoyproxy/envoy/issues/19844)

cskarby:
**What type of PR is this?**
* feat: configure internal address according to ClientTrafficPolicy
* api: add internalAddressConfig fields in ClientTrafficPolicy.clientIPDetection

The http request header x-forwarded-for (XFF) is not beeing set when the request is origination from ip-ranges that envoy detects as internal. By default any request from private ip ranges ([RFC1918](https://datatracker.ietf.org/doc/html/rfc1918) / [RFC4193](https://datatracker.ietf.org/doc/html/rfc4193)) are considered internal. This PR is needed to allow users of envoy-gateway to configure [HttpConnectionManager.InternalAddressConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-internaladdressconfig).

Fixes #3598


shawnh2 (commented):
Please remove everything under `/internal`, we will start reviewing the api changes first

cskarby:
> Please remove everything under `/internal`, we will start reviewing the api changes first

Thank you for clarifying the process. We have started a discussion about the best way forward in ticket  #3598
I will investigate and come back with a new merge request.

Issue #3245, opened 2024-04-22
Support override_stream_error_on_invalid_http_message


guydc:
*Description*:
Envoy Docs:

> if you run level two Envoy version 1.11.1 or greater which terminates HTTP/2 or above, we strongly advise you to change the HttpConnectionManager configuration of your level two Envoy, by setting its downstream [validation of HTTP messaging option](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-stream-error-on-invalid-http-message) to true

Envoy Gateway should support configuration of this setting, so that users can adjust it, based on their deployment mode (L1 or L2) and protocol versions in use. 

*Relevant Links*:
https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two

shawnh2:
closed in favor of #3682

PR #3682, opened 2024-06-26
feat(translator): http2 upstream settings

### LINKED ISSUE #3670: Support HTTP/2 settings in BackendTrafficPolicy for upstreams
*Description*:

Several HTTP2 Protocol options can have an impact on resilience, performance and security of upstream connections. For example `max_concurrent_stream` defines "For upstream connections, [...] how many streams Envoy will initiate concurrently on a single connection." 

#3222 introduced an API for HTTP2 settings in downstream connections. Envoy Gateway can support the same settings for upstream connections in `BackendTrafficPolicy`. 

*Relevant Links*:
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-http2protocoloptions

guydc:
- Exposes HTTP2 settings for upstream in Backend Traffic Policy
- Adds support for the envoy `override_stream_error_on_invalid_http_message` option which is specifically relevant for L2 envoy deployments running http2: https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two

Fixes #3670, #3245


api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,33 @@ type BackendRef struct {
 // A CIDR can be an IPv4 address range such as "192.168.1.0/24" or an IPv6 address range such as "2001:0db8:11a3:09d7::/64".
 // +kubebuilder:validation:Pattern=`((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\/([0-9]+))|((([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))\/([0-9]+))`
 type CIDR string
+
+// HTTP2Settings provides HTTP/2 configuration on the listener.
```
zhaohuabing, line 483: Nit:
```suggestion
// HTTP2Settings provides HTTP/2 configuration for listeners and backends.
```

api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,33 @@ type BackendRef struct {
...
+type HTTP2Settings struct {
```
zhaohuabing, line 484: Should we use the same default value for both the listeners and clusters?

guydc: - envoy recommendations for secure-by-default are the same (window sizes). 
- max concurrent streams do not necessarily require hardening in the cluster context. 
- stream resetting behavior is mostly a downstream concern.  

We can move some of the defaulting behavior to the API, but some will remain in the XDS translator (due to differences between clusters and listeners). So, I propose we keep it as-is for now.

zhaohuabing: @guydc Thanks for the explanation!  I'm just curious because the recommendation is for listeners. They can be the same for Clusters if there's no unintentional side effects.

zhaohuabing (approved):
LGTM Thanks!

zhaohuabing (approved):

api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,33 @@ type BackendRef struct {
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	MaxConcurrentStreams *uint32 `json:"maxConcurrentStreams,omitempty"`
+
+	// ResetStreamOnError determines if Envoy will terminate the stream or the connection in the event of HTTP messaging error
+	// It's recommended for L2 Envoy deployments to set this value to true.
+	// https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two
+	// +optional
+	ResetStreamOnError *bool `json:"resetStreamOnError,omitempty"`
```
arkodg, line 510: `false` here really means `terminateConnectionOnError` , and `true` means `terminateStreamOnError` 
is there any other API name and value that can encapsulate this ?

guydc: Yea, talking about connections instead of streams would be easier for users to understand and better comprehend the impact. The only downside is that this could be confusing to someone who reads the envoy docs.

api/v1alpha1/shared_types.go
```diff
@@ -503,9 +503,10 @@ type HTTP2Settings struct {
...
 
-	// ResetStreamOnError determines if Envoy will terminate the stream or the connection in the event of HTTP messaging error
-	// It's recommended for L2 Envoy deployments to set this value to true.
+	// ResetConnectionOnError determines if Envoy will terminate the connection or just the offending stream in the event of HTTP messaging error
+	// It's recommended for L2 Envoy deployments to set this value to false.
 	// https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two
+	// Default: true
 	// +optional
-	ResetStreamOnError *bool `json:"resetStreamOnError,omitempty"`
+	TerminateConnectionOnMessagingError *bool `json:"terminateConnectionOnMessagingError,omitempty"`
```
arkodg, line 511: would `TerminateConnOnError` be too crypted ?

api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,34 @@ type BackendRef struct {
...
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	MaxConcurrentStreams *uint32 `json:"maxConcurrentStreams,omitempty"`
+
+	// ResetConnectionOnError determines if Envoy will terminate the connection or just the offending stream in the event of HTTP messaging error
+	// It's recommended for L2 Envoy deployments to set this value to false.
+	// https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two
+	// Default: true
+	// +optional
+	TerminateConnOnError *bool `json:"terminateConnOnError,omitempty"`
```
arkodg, line 511: lets move this to `TerminateConnectionOnError` since we already have `InitialConnectionWindowSize` in the same struct 🙈

api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,34 @@ type BackendRef struct {
...
+	InitialConnectionWindowSize *resource.Quantity `json:"initialConnectionWindowSize,omitempty"`
+
+	// MaxConcurrentStreams sets the maximum number of concurrent streams allowed per connection.
+	// If not set, the default value is 100.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	MaxConcurrentStreams *uint32 `json:"maxConcurrentStreams,omitempty"`
+
+	// ResetConnectionOnError determines if Envoy will terminate the connection or just the offending stream in the event of HTTP messaging error
```
arkodg, line 506:
```suggestion
	// TerminateConnectionOnError determines if Envoy will terminate the connection or just the offending stream in the event of HTTP messaging error
```

api/v1alpha1/shared_types.go
```diff
@@ -478,3 +479,34 @@ type BackendRef struct {
...
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	MaxConcurrentStreams *uint32 `json:"maxConcurrentStreams,omitempty"`
+
+	// TerminateConnOnError determines if Envoy will terminate the connection or just the offending stream in the event of HTTP messaging error
+	// It's recommended for L2 Envoy deployments to set this value to false.
+	// https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/level_two
+	// Default: true
+	// +optional
+	TerminateConnOnError *bool `json:"terminateConnOnError,omitempty"`
```
zhaohuabing, line 511: Should defaulting to `True` for a bool go against users' intuition?

guydc: Agreed, that's the most intuitive representation.

zhaohuabing (approved):
LGTM. Defer to @arkodg 🙂

PR #3692, opened 2024-06-28
chore: move benchmark-test job in build_and_test.yaml

shawnh2:
**What type of PR is this?**

Runs after `build` is succeeded, save some time in CI.

Fixes #


.github/workflows/build_and_test.yaml
```diff
@@ -135,6 +135,29 @@ jobs:
...
+
+    # Benchmark
+    - name: Run Benchmark tests
+      env:
+        KIND_NODE_TAG: v1.28.9
+        IMAGE_PULL_POLICY: IfNotPresent
+        # Args for benchmark test
+        BENCHMARK_RPS: 10000
+        BENCHMARK_CONNECTIONS: 100
+        BENCHMARK_DURATION: 90
```
arkodg, line 153: can we reduce this to 60s ? or even 30s ?
should further reduce CI time

shawnh2: 30s sounds fine

zirain (approved):

arkodg (approved):

PR #3697, opened 2024-06-28
chore: add "extproc" to extproc ir name

zhaohuabing:
Befor: "envoyextensionpolicy/default/policy-for-gateway/0"
After:  "envoyextensionpolicy/default/policy-for-gateway/extproc/0"

Also make it aligned with the wasm ir name.

### DIFF
--- a/internal/gatewayapi/envoyextensionpolicy.go
+++ b/internal/gatewayapi/envoyextensionpolicy.go
@@ -430,7 +430,7 @@ func (t *Translator) buildExtProcs(policy *egv1a1.EnvoyExtensionPolicy, resource
 	}
 
 	for idx, ep := range policy.Spec.ExtProc {
-		name := irConfigNameForEEP(policy, idx)
+		name := irConfigNameForExtProc(policy, idx)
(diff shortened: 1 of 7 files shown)


internal/gatewayapi/envoyextensionpolicy.go
```diff
@@ -534,12 +534,11 @@ func (t *Translator) buildExtProc(
...
 }
 
-func irConfigNameForEEP(policy *egv1a1.EnvoyExtensionPolicy, idx int) string {
+func irConfigNameForExtProc(policy *egv1a1.EnvoyExtensionPolicy, index int) string {
 	return fmt.Sprintf(
-		"%s/%s/%d",
-		strings.ToLower(egv1a1.KindEnvoyExtensionPolicy),
-		utils.NamespacedName(policy).String(),
-		idx)
+		"%s/extproc/%s",
```
arkodg, line 539: ext-proc instead ? cc @guydc

zhaohuabing: EG is using flattened names like "envoyextensionpolicy", and envoy is using "snake_case".

guydc (approved):
LGTM

Issue #3698, opened 2024-06-28
Why the memory is not increase linearly when the HTTPRoutes scales up?


arkodg:

arkodg:
thanks @shawnh2 for  picking this one up !

shawnh2:
According to the heap profile, looks like we have two places that consume the most memory:

## 1. The `DeepCopyInto` method in XDS ResourceVersionTable

Specifically: https://github.com/envoyproxy/gateway/blob/eeb62c88f8949f8da8a1278ec5515ffa1a004444/internal/xds/types/resourceversiontable.go#L38

<img width="1510" alt="image" src="https://github.com/user-attachments/assets/d7537593-80ce-473f-ae45-a55e6bee2645">

```plaintext
      flat  flat%   sum%        cum   cum%
      196.52MB 24.68% 79.92%   196.52MB 24.68%  reflect.New
   ...
         0     0% 97.46%   253.12MB 31.79%  github.com/envoyproxy/gateway/internal/message.HandleSubscription[go.shape.string,go.shape.*uint8]
         0     0% 97.46%   204.05MB 25.63%  github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate
         0     0% 97.46%   204.05MB 25.63%  github.com/envoyproxy/gateway/internal/xds/translator/runner.(*Runner).subscribeAndTranslate.func1
         0     0% 97.46%   203.55MB 25.56%  github.com/envoyproxy/gateway/internal/xds/types.(*ResourceVersionTable).DeepCopy
         0     0% 97.46%   203.55MB 25.56%  github.com/envoyproxy/gateway/internal/xds/types.(*ResourceVersionTable).DeepCopyInto
         0     0% 97.46%   222.07MB 27.89%  github.com/telepresenceio/watchable.(*Map[go.shape.string,go.shape.*uint8]).Store
         0     0% 97.46%   222.07MB 27.89%  github.com/telepresenceio/watchable.(*Map[go.shape.string,go.shape.*uint8]).unlockedStore
         0     0% 97.46%   222.07MB 27.89%  github.com/telepresenceio/watchable.DeepCopy[go.shape.*uint8]
         0     0% 97.46%   134.51MB 16.89%  google.golang.org/protobuf/internal/impl.(*MessageInfo).initOneofFieldCoders.func5
[...]

PR #3704, opened 2024-06-29
feat: Implement target selectors for policies.

### LINKED ISSUE #3607: Support targetSelectors in Envoy Gateway policies
I think that since this API still needs some discussion, I will remove the `targetSelector` for now from the PR and keep only the plural `targetRefs` in this PR.

The target selectors can be done in a separate PR after the API has been designed and agreed upon.

_Originally posted by @liorokman in https://github.com/envoyproxy/gateway/pull/3581#discussion_r1639514571_

liorokman:
This PR implements target selectors for policies

Fixes #3607


api/v1alpha1/policy_helpers.go
```diff
@@ -10,18 +10,33 @@ import (
...
 	// TargetRefs are the names of the Gateway resources this policy
 	// is being attached to.
 	TargetRefs []gwapiv1a2.LocalPolicyTargetReferenceWithSectionName `json:"targetRefs,omitempty"`
+
+	// TargetSelectors allow targeting resources for this policy based on labels
+	TargetSelectors []TargetSelector `json:"targetSelectors,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="has(self.group) ? self.group == 'gateway.networking.k8s.io' : true ", message="group must be gateway.networking.k8s.io"
+type TargetSelector struct {
```
arkodg, line 29: can we add a note here saying cross ns will require a ReferenceGrant ?

internal/gatewayapi/helpers.go
```diff
@@ -495,3 +496,37 @@ func irConfigName(policy client.Object) string {
 		strings.ToLower(policy.GetObjectKind().GroupVersionKind().Kind),
 		utils.NamespacedName(policy).String())
 }
+
+func getPolicyTargetRefs[T client.Object](policy egv1a1.PolicyTargetReferences, potentialTargets []T) []gwapiv1a2.LocalPolicyTargetReferenceWithSectionName {
+	dedup := map[gwapiv1a2.LocalPolicyTargetReferenceWithSectionName]byte{}
```
arkodg, line 501: sets instead ?

internal/gatewayapi/helpers.go
```diff
@@ -495,3 +496,37 @@ func irConfigName(policy client.Object) string {
...
+	for _, ref := range policy.GetTargetRefs() {
+		dedup[ref] = byte(0)
+	}
+
+	for _, currSelector := range policy.TargetSelectors {
```
arkodg, line 506: need to account for ReferenceGrant here

internal/gatewayapi/helpers.go
```diff
@@ -495,3 +496,37 @@ func irConfigName(policy client.Object) string {
...
+						Kind:  gwapiv1a2.Kind(gvk.Kind),
+						Name:  gwapiv1a2.ObjectName(obj.GetName()),
+					},
+				}] = byte(0)
+			}
+		}
+	}
+
+	ret := []gwapiv1a2.LocalPolicyTargetReferenceWithSectionName{}
+	for key := range dedup {
```
arkodg, line 528: can we garantee stability of list order with maps or sets ?

arkodg: also is there a way to make this stateless ? we are actively trying to reduce mem usage https://github.com/envoyproxy/gateway/issues/3698

liorokman: > can we garantee stability of list order with maps or sets ?

No. Sets are the same as the builtin maps, since they are implemented in the same way, and they do not guarantee the list order. When transforming the map to a list, I can sort the list. What would be the correct key in which the list should be ordered?

> is there a way to make this stateless ? we are actively trying to reduce mem usage

Using a `map` to deduplicate the set of relevant targets means that the code never allocates more memory than it would need to keep track of the targets of the policy. I don't see a way to collect the targets of the selectors and not keep _some_ state. Note that the memory is only allocated for the duration of the time it takes to handle the policy - it's not some cache that is kept and never released.

internal/gatewayapi/helpers.go
liorokman: Why?

Policies are not allowed to access targets across namespaces. These are all `LocalPolicyTargetReference` attachments.

Adding support for cross-namespace access for policies is a rather large change.

api/v1alpha1/policy_helpers.go
liorokman: Why? None of these policies can be attached across namespaces.

internal/gatewayapi/helpers.go
arkodg: ah thanks for reminding me, we only support same namespace today

api/v1alpha1/policy_helpers.go
arkodg: ah thanks for reminding me about this

internal/gatewayapi/helpers.go
arkodg: lets sort based on creation timestamp
```
	// Sort based on timestamp
	sort.Slice(ret, func(i, j int) bool {
		return ret[i].CreationTimestamp.Before(&(ret[j].CreationTimestamp))
	})
	```

guydc (approved):
LGTM, Thanks!

arkodg (approved):
thanks !

sadovnikov:
@arkodg, @liorokman, I installed `v1.1.0-rc.1` and tried to create a `SecurityPolicy` with the target selectors. It failed on validation. #3901 is logged for this.

Issue #3705, opened 2024-06-29
update docs for brew install 


zirain:
update installation doc after https://github.com/Homebrew/homebrew-core/pull/175940 merged.

PR #3714, opened 2024-06-30
docs: install with brew

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3705


arkodg (approved):
🚀  ha thanks for integrating into `brew`

shawnh2 (approved):

PR #3748, opened 2024-07-04
e2e: zkipin tracing test

zirain:
fix: https://github.com/envoyproxy/gateway/issues/3686

PR #3749, opened 2024-07-04
xds: should use TCP as default protocol

zirain:
found this when debuging https://github.com/envoyproxy/gateway/pull/3748


shawnh2 (approved):
thanks for fixing this

zhaohuabing (approved):

Xunzhuo (approved):
/lgtm

Issue #3737, opened 2024-07-03
METADATA not support on ALS


zirain:
METADATA not support on ALS, will investigate later.

_Originally posted by @zirain in https://github.com/envoyproxy/gateway/pull/3736#discussion_r1663403644_

PR #3751, opened 2024-07-04
accesslog: remove ALS gRPC initialMetadata

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3737

### DIFF
--- a/internal/xds/translator/accesslog.go
+++ b/internal/xds/translator/accesslog.go
@@ -6,7 +6,6 @@
 package translator
 
 import (
-	"encoding/json"
 	"errors"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


zhaohuabing:
@zirain  Could you please explaining why this fix https://github.com/envoyproxy/gateway/issues/3737 in the comment?

zirain:
> @zirain Could you please explain in the comment why this fixed #3737 ?

```
        // Header value.
	//
	// The same :ref:`format specifier <config_access_log_format>` as used for
	// :ref:`HTTP access logging <config_access_log>` applies here, however
	// unknown header values are replaced with the empty string instead of “-“.
	// Header value is encoded as string. This does not work for non-utf8 characters.
	// Only one of “value“ or “raw_value“ can be set.
```

Header valus support access log format, but didn't support extension formatter(e.g. METADATA, CEL etc.)
InitialMetadata is designed to send something for auth(e.g. got token from request header)
ALS doesn't need it in current stage.

zhaohuabing (approved):

arkodg (approved):

PR #3757, opened 2024-07-04
feat(translator): JsonPath in PatchPolicy

### LINKED ISSUE #1686: Allowing matching a filterchain when applying EnvoyPatchPolicy
EnvoyPatchPolicy utilities [JSON Patch](https://jsonpatch.com/) for applying patches to XDS resources, but JSON Patch lacks the ability to select a specific member in an array based on conditions other than its index.

For example: assuming there're two filter chains in a listener, there's no way that we could target the second filter chain based on its serverName `www.bar.com` when applying a patch.  

We won't be able to know the index of the targeted filter chain without inspecting the generated Envoy configuration, which is impossible when we want to use a controller to generate the patch. We could possibly create that patch manually, but it's cumbersome and unreliable, and the index might change with updates to the EG releases.

```yaml
      - address:
          socketAddress:
            address: 0.0.0.0
            portValue: 10443
        filterChains:
        - filterChainMatch:
            serverNames:
            - www.foo.com
          filters:
            typedConfig:
              ......          
        - filterChainMatch:
            serverNames:
            - www.bar.com
          filters:
            typedConfig:
              ......

```

denniskniep:
JsonPath can be utilized to select elements for JSONPatch in EnvoyPatchPolicy

Fixes #1686

**`EnvoyPatchPolicy` API Changes:** 
* Added `jsonPath`
* `path` is optional 

These changes are backwards compatible

**Change Summary:**
`jsonPath` is evaluated based on the json and the patch is done for all found items (pointers)

We can only "find" existing items in the json with jsonPath. Therefore we need a way to extend the found items (pointers). This enables us to add not existing properties. This is implemented by adding the string of the `path` to the end of the found items (pointers)

**Example:**
`jsonPath`: `"filter_chains[*].filters[*].typed_config"`
`path`: `"/preserve_external_request_id"`

will find following items (pointers):
`/filter_chains/0/filters/0/typed_config`
`/filter_chains/0/filters/1/typed_config`

Final jsonPointers which are used withing the JSONPatch:
`/filter_chains/0/filters/0/typed_config/preserve_external_request_id`
`/filter_chains/0/filters/1/typed_config/preserve_external_request_id`


denniskniep:
@arkodg can you plz have a quick look if this is the right direction?
If yes, I will start working on the docs  or should that be handled by a different PR?

internal/xds/translator/jsonpatch.go
```diff
@@ -84,9 +84,9 @@ func processJSONPatches(tCtx *types.ResourceVersionTable, envoyPatchPolicies []*
 				}
 			}
 
-			// If Path is "" and op is "add", unmarshal and add the patch as a complete
+			// If Path and JSONPath is "" and op is "add", unmarshal and add the patch as a complete
 			// resource
-			if p.Operation.Op == AddOperation && p.Operation.Path == EmptyPath {
+			if p.Operation.Op == AddOperation && p.Operation.IsPathEmpty() && p.Operation.IsJSONPathEmpty() {
```
arkodg, line 89:
```suggestion
			if p.Operation.Op == AddOperation && (p.Operation.IsPathEmpty() || p.Operation.IsJSONPathEmpty()) {
```

arkodg, line 89: amd when testing for IsEmpty - ` o.Path != nil && *o.Path == EmptyPath`

internal/xds/translator/jsonpatch.go
```diff
@@ -240,125 +240,150 @@ func processJSONPatches(tCtx *types.ResourceVersionTable, envoyPatchPolicies []*
...
-			// into and validated before saving it into the xds output resource
-			switch p.Type {
-			case resourcev3.ListenerType:
-				temp := &listenerv3.Listener{}
-				if err = protojson.Unmarshal(modifiedJSON, temp); err != nil {
-					tErr := fmt.Errorf(unmarshalErrorMessage(err, string(modifiedJSON)))
-					tErrs = errors.Join(tErrs, tErr)
-					continue
+			var jsonPointers []string
+			if p.Operation.JSONPath != nil {
```
arkodg, line 244: are we expecting JSONPath and Path to be set at the same time ?

arkodg: looks like JSONPath supports nested values
https://datatracker.ietf.org/doc/rfc9535/

denniskniep: @arkodg yes, both properties can be set at the same time, because JSONPath is only able to select what already exists in the JsonDocument. 

If one want to create a new property at a certain element, then this can be done by:
1. Select the existing element with `JSONPath` Property
2. `Path` Property is added to the selected existing element
3. Element can be created at a not existing path

internal/xds/translator/jsonpatch.go
denniskniep: Agree, will change this, then its more explicit.

denniskniep: I think we need to keep this condition, because else it would always run into that branch if
1. a JsonPath is specified and path is empty
2. a path is specified and JsonPath is empty

Both scenarios will result in a path to a certain element and we don't want to add as a complete resource.

Maybe I didn´t get the intention behind this proposal...?

arkodg: ah got it, thanks for clarifying

internal/xds/translator/testdata/in/xds-ir/jsonpatch-with-jsonpath.yaml
```diff
@@ -0,0 +1,163 @@
+envoyPatchPolicies:
```
arkodg, line 1: can we add examples with filters, which is the real value prop for using jsonpatch over path since its hard to figure out index

internal/xds/translator/jsonpathtopointer_test.go
```diff
@@ -0,0 +1,374 @@
...
+import (
+	"sort"
+	"strconv"
+	"testing"
+
+	"github.com/ohler55/ojg/jp"
+	"github.com/stretchr/testify/require"
+)
+
+type pointerTest struct {
```
arkodg, line 17: to maintain uniformity, can these tests be represented in this https://github.com/envoyproxy/gateway/blob/eeb62c88f8949f8da8a1278ec5515ffa1a004444/internal/ir/xds_test.go#L533 format ? (moving the struct def inside the test itself)

arkodg:
hey  @denniskniep sorry for the delay in review, the diff looks good !
added a few minor comments

internal/xds/translator/testdata/in/xds-ir/jsonpatch-with-jsonpath.yaml
denniskniep: Will add this test. It makes clear, that you do not need to go through all parent nested properties and you can select explicitly the route by name, which is much more reliable then selecting via index:
```
  - type: "type.googleapis.com/envoy.config.route.v3.RouteConfiguration"
    name: "first-listener"
    operation:
      op: "replace"
      jsonPath: "..routes[?(@.name=='second-route')].route.upgrade_configs"
      value:
      - upgrade_type: CONNECT
        connect_config:
          {}
```

but there is already this scenario:
`jsonPath: "..endpoints[*].load_balancing_weight"`
which addresses ALL endpoints, which is not possible right now

denniskniep:
> hey @denniskniep sorry for the delay in review, the diff looks good ! added a few minor comments

cool, thx

@arkodg addressed your comments

arkodg (approved):
LGTM thanks for adding this feature !
hopefully this makes it easier to author EnvoyPatchPolicies for modifying elements within a list

zirain (approved):

PR #3764, opened 2024-07-05
fix: enable upgrade test

guydc:
- e2e make: 
   - Cleanup manifests from previous suites before upgrade tests to create a clean test environment 
   - Remove conditional execution of upgrade test from make file, which is not in use
- upgrade suite:
  - Execute tests in a specific order, to ensure that the upgrade test is the last test in the suite. 
  - Upgrade test manifests must remain backwards-compatible to latest-version. So, instead of using the general e2e base manifests, a dedicated minimal manifests file is used. 
- upgrade test: 
  - Use EG helm tool where possible (uninstall, install)
  - Explicitly cleanup resources like GW, GWC before uninstall to avoid finalizer deadlocks
  - Remove CRDs as part of the initial uninstallation, to ensure latest-version is installed in a clean test environment
  - Explicitly apply base and test manifests and after re-installation (as they were removed due to CRD uninstall) 
  - Upgrade process:
    - Migrate specific CRDs like BTLS that must be removed and re-installed due to breaking changes in v1alpha3
    - Update CRDs to current version
    - Install current chart from source

Relates #3311


zirain:
> Cleanup manifests from previous suites before upgrade tests to create a clean test environment

IMO, upgrade test is special, whether need to run it in a separate environment？

guydc:
> > Cleanup manifests from previous suites before upgrade tests to create a clean test environment
> 
> IMO, upgrade test is special, whether need to run it in a separate environment？

Agree, but it also has certain k8s dependencies, so it'll require setting up 4 more test clusters... I'm concerned about delaying PRs too much. If maintainers are onboard, I can separate it to a dedicated cluster. 

Also, note that as long as we use the GWAPI conformance suite, we anyway have to install CRDs and some base content. Otherwise, the suite won't start (it looks for CRDs, GWC, ... ).

arkodg (approved):
LGTM thanks for codifying the upgrade process and enabling the test case back !

tools/make/kube.mk
```diff
@@ -141,20 +141,8 @@ run-e2e: install-e2e-telemetry
 ifeq ($(E2E_RUN_TEST),)
 	go test -v -tags e2e ./test/e2e --gateway-class=envoy-gateway --debug=true --cleanup-base-resources=false
 	go test -v -tags e2e ./test/e2e/merge_gateways --gateway-class=merge-gateways --debug=true --cleanup-base-resources=false
-	go test -v -tags e2e ./test/e2e/multiple_gc --debug=true --cleanup-base-resources=false
+	go test -v -tags e2e ./test/e2e/multiple_gc --debug=true --cleanup-base-resources=true
 	go test -v -tags e2e ./test/e2e/upgrade --gateway-class=upgrade --debug=true --cleanup-base-resources=$(E2E_CLEANUP)
-else
-ifeq ($(E2E_RUN_EG_UPGRADE_TESTS),false)
```
shawnh2, line 147: does it mean we can remove `E2E_RUN_EG_UPDATE_TESTS` as well?

zirain (approved):

zirain (approved):

PR #3768, opened 2024-07-06
docs: gRPC Access Log Service (ALS) sink

zirain:
fix: https://github.com/envoyproxy/gateway/issues/3745

PR #3770, opened 2024-07-07
api: make accesslog format optional

zirain:
for https://github.com/envoyproxy/gateway/pull/3768, this should be optional to simply the example.


guydc (approved):
LGTM

arkodg (approved):

PR #3771, opened 2024-07-07
fix: prevent xdsIR updates from overwriting RateLimit configs from other xdsIR

### LINKED ISSUE #3678: Ratelimit does not work across multiple GatewayClasses
*Description*:
EG currently uses Listener as domain for Ratelimit service config. It should instead use Gateway class not to override RL config for a case when when multiple Envoy deployments with diff gateway classes use same RL instance. 

*Repro steps*:

sharing.

*Environment*:

*Logs*:

sanposhiho:
**What type of PR is this?**

This PR tries to prevent xdsIR updates from overwriting RateLimit configs from other xdsIR.

When there are multiple gateway classes, the gateway API runner updates `r.XdsIR.Store` several times with different keys.
https://github.com/envoyproxy/gateway/blob/d784c3237765aa258c3526b32765dd9c3abf6415/internal/gatewayapi/runner/runner.go#L81

But, the ratelimit config runner create a ratelimit config from zero every time it gets updates of XdsIR.
https://github.com/envoyproxy/gateway/blob/d784c3237765aa258c3526b32765dd9c3abf6415/internal/globalratelimit/runner/runner.go#L114-L136

As a result, an update to a XdsIR could overwrite rate limit configs generated from other XdsIRs in the ratelimit config runner.

Fixes #3678


internal/globalratelimit/runner/runner.go
```diff
@@ -120,6 +124,7 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 					r.Logger.Error(err, "failed to update the config snapshot")
```
arkodg, line 124: is `r.addNewSnapshot(ctx, nil)` still valid ? looks like it will need to be replaced by `r.updateSnapshot(ctx, xdsResourcesToUpdate)` ?

arkodg:
thanks for picking this one up @sanposhiho !
the approach looks good, added a comment
can you also add a test in runner_test.go where you instrument some creates and deletes for multiple keys similar to https://github.com/envoyproxy/gateway/blob/181979697fc2f1254d37bab98d1820b72582fb9b/internal/gatewayapi/runner/runner_test.go#L28

internal/globalratelimit/runner/runner.go
sanposhiho: 🤦‍♂️

sanposhiho:
@arkodg Addressed your points.

zirain (approved):

internal/globalratelimit/runner/runner.go
```diff
@@ -109,17 +110,27 @@ func (r *Runner) serveXdsConfigServer(ctx context.Context) {
...
+	for _, xdsR := range rateLimitConfigsCache {
+		xdsResourcesToUpdate[resourcev3.RateLimitConfigType] = append(xdsResourcesToUpdate[resourcev3.RateLimitConfigType], xdsR...)
+	}
+
+	return xdsResourcesToUpdate
+}
+
 func (r *Runner) subscribeAndTranslate(ctx context.Context) {
+	// rateLimitConfigsCache is a cache of the rate limit config, which is keyed by the xdsIR key.
+	rateLimitConfigsCache := map[string][]cachetype.Resource{}
```
arkodg, line 124: is there a way to not maintain this cache here and instead create an API like `rCache := r.getSnapshot(ctx)` instead ? this will allow us to reduce one copy in memory

sanposhiho: Let me try; depending on the complexity that would give though, that looks a better idea.

sanposhiho: The change would be like this https://github.com/sanposhiho/gateway/commit/a9f0a190ed7ecc939225598c0e5ce96f23307ce9

So, either way, we have to have a cache to track which ir.Key generates which rate limit configurations. (A cache size would be smaller than the current cache of entire configs, that's one good point though.)
The cache is required because otherwise we wouldn't know which part of snapshot we have to update. (A `message.Update` comes from `watchable.Snapshot` doesn't have a previous state of ir.)

So, I'm not very motivated to proceed that way; For me, it looks like only a little gain (less memory usage) would come with a certain downside (complexity).

arkodg: hey @sanposhiho thanks for writing up the quick diff, what I was suggesting was a little different, instead of holding the entire cache on the stack using `rateLimitConfigsCache` which incurs a constant memory hit, can we retrieve the cache snapshot using an API like 
https://github.com/envoyproxy/go-control-plane/blob/1da4500d00e270d803caefbe0c20e4d3d162e586/pkg/cache/v3/snapshot.go#L124
every time we get a new new message ?

sanposhiho: Sorry, what's the difference from my diff? It fetches the previous snapshot with GetSnapshot -> GetResources, and try to rebuild a new snapshot from there already.

arkodg: I missed that 🙈 
lets go ahead with diff to keep the code complexity low and we can revisit these if/when we hit a memory bottleneck

arkodg (approved):
LGTM thanks !

Issue #3774, opened 2024-07-07
Allow configuring TLS settings for OIDC upstream


benediktwerner:
Our OIDC provider doesn't support Envoy's default ECDH curves so we currently have to configure additional curves via a patch policy.

It looks like EG v1.1 will already add a default `BackendTLSConfig` for the `EnvoyProxy` which seems good enough but at first glance, it looks like it currently isn't applied to the OIDC provider cluster?

zhaohuabing:
Related to https://github.com/envoyproxy/gateway/issues/3622

As discussed in the EG meeting, the provider can be modeled as a `Backend` resource and be associated with a `BackendTLSConfig`.

zhaohuabing:
Closing this as now OIDC provider TLS can be configured with `Backend` and `BackendTLSPolicy`.

Fixed with https://github.com/envoyproxy/gateway/pull/4227

PR #3799, opened 2024-07-09
API: model oidc provider as Backend to support TLS setting

zhaohuabing:
Related issue: https://github.com/envoyproxy/gateway/issues/3774


arkodg:
thanks for picking this one up @zhaohuabing ! the API presented looks good, but im hoping we wait until https://github.com/envoyproxy/gateway/issues/3069 is resolved, so that we can discuss the solution is a holistic way for all non xRoute backendRefs

PR #3812, opened 2024-07-10
Use Wasm instead of WASM

mathetake:
This is a really common mistake but WebAssembly is officially Wasm,
not WASM. This fixes that in various places in code base.


zirain (approved):

zhaohuabing (approved):
Thanks @mathetake for pointing this out and the fix!

PR #3654, opened 2024-06-23
docs: custom aliases

zirain:
use custom alias.html to make to `/tasks/quickstart` redirect to `/<current_version>/tasks/quickstart`

PR #3817, opened 2024-07-11
docs: add fixed links to the current eg docs

zhaohuabing:
This PR adds fixed versionless links to the current EG docs. The current(latest released) version of docs will always be in the subpath "/docs/", and all the previous versions will be in their corresponding versioned paths. 

In this way, any outside links to EG pages will always point to the current release without needing to change the URLs after a new version is released. 

Before:  
v1.0.2  https://gateway.envoyproxy.io/v1.0.2/tasks/traffic/circuit-breaker/
v0.6.0 https://gateway.envoyproxy.io/v0.6.0/tasks/traffic/circuit-breaker/
v0.5.0 https://gateway.envoyproxy.io/v0.5.0/tasks/traffic/circuit-breaker/
...

After: 
v1.0.2 https://gateway.envoyproxy.io/docs/tasks/traffic/circuit-breaker/
v0.6.0 https://gateway.envoyproxy.io/v0.6.0/tasks/traffic/circuit-breaker/
v0.5.0 https://gateway.envoyproxy.io/v0.5.0/tasks/traffic/circuit-breaker/
...

After releasing v1.1.0, rename the current `docs` to v1.0.2, and copy the latest to a new `docs`.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -55,7 +55,6 @@ func TestE2E(t *testing.T) {
 		// All e2e tests should leave Features empty.
 		SupportedFeatures: sets.New[features.SupportedFeature](features.SupportGateway),
 		SkipTests: []string{
-			tests.ClientTimeoutTest.ShortName,        // https://github.com/envoyproxy/gateway/issues/2720
 			tests.GatewayInfraResourceTest.ShortName, // https://github.com/envoyproxy/gateway/issues/3191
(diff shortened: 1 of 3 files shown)


Xunzhuo:
Can be solved by https://github.com/envoyproxy/gateway/pull/3654

zhaohuabing:
This way is simple and it won't change the URL in the browser address bar. @Xunzhuo

zhaohuabing:
> It will increase the duplicated doc files, making it hard maintain, cc @zirain thoughts on this?

We can just rename v1.0.2 to docs if we don't want the duplication.

PR #3819, opened 2024-07-11
docs: add fixed links to the current version of eg docs 

zhaohuabing:
This PR adds fixed versionless links to the current EG docs. The current(latest released) version of docs will always be copied to the subpath "/docs/" with the `make docs` command, and all the previous versions will be in their corresponding versioned paths. 

In this way, any outside links to EG pages will always point to the current release without needing to change the URLs after a new version is released. 

For example, the links to the circuit breaker task docs before and after this PR:

Before:  
v1.0.2  https://gateway.envoyproxy.io/v1.0.2/tasks/traffic/circuit-breaker/
v0.6.0 https://gateway.envoyproxy.io/v0.6.0/tasks/traffic/circuit-breaker/
v0.5.0 https://gateway.envoyproxy.io/v0.5.0/tasks/traffic/circuit-breaker/
...

After: 
v1.0.2 will be available at both links:
   https://gateway.envoyproxy.io/docs/tasks/traffic/circuit-breaker/  and 
   https://gateway.envoyproxy.io/v1.0.2/tasks/traffic/circuit-breaker/
v0.6.0 https://gateway.envoyproxy.io/v0.6.0/tasks/traffic/circuit-breaker/
v0.5.0 https://gateway.envoyproxy.io/v0.5.0/tasks/traffic/circuit-breaker/
...

When releasing v1.1.0, the `make docs-release TAG=v1.1.0` command will replace the current docs directory with the v1.1.0 docs.


zhaohuabing:
@arkodg @zirain @Xunzhuo what do you think about this approach? If we agree on this, I'll go ahead.

zirain:
> @arkodg @zirain @Xunzhuo what do you think about this approach? If we agree on this, I'll go ahead.

yeah, I'm +1 on this.

Xunzhuo:
Looks good! Please update release process docs too.

zhaohuabing:
> Looks good! Please update release process docs too.

The release process doesn't need to change. The ` make docs-release TAG=v1.1.0` command will replace the current docs directory with the v1.1.0 docs.

arkodg (approved):
LGTM thanks !
this will ensure users can safely hyperlink the versionless url and get the current latest release content and should also improve google search for this one link

zirain (approved):

PR #3827, opened 2024-07-12
chore: report a translate error to errChan to make it observed correctly

sanposhiho:
**What type of PR is this?**

This PR make the rate limit runner report a translate error to `errChan`.
It makes those errors correctly observed in the metric.

https://github.com/envoyproxy/gateway/blob/main/internal/message/watchutil.go#L57

Fixes #


shawnh2 (approved):
Thanks for improving this

Xunzhuo (approved):

PR #3831, opened 2024-07-12
chore: remove cherrypicker action

zirain:
this action didn't work for a long time, let's remove it now to reduce some nosie.


Xunzhuo (commented):
why there are some go mod changes>

zirain:
> why there are some go mod changes>

goland cause

Xunzhuo:
if so, we need to add lint to example server too, make sure its always go mod tidy

zirain:
> if so, we need to add lint to example server too, make sure its always go mod tidy

removed, let me do that in seperated one.

Xunzhuo (approved):

arkodg (approved):

PR #3835, opened 2024-07-12
fix: backendtls minversion

### LINKED ISSUE #3834: bug: nil pointer if backedtls.minVersion is set but backedtls.maxVersion is not
*Description*:
nil pointer if backedtls.minVersion is set but backedtls.maxVersion is not:
https://github.com/envoyproxy/gateway/blob/b9d149353e2c1872a8003a72dcd516c538b2ab18/internal/gatewayapi/backendtlspolicy.go#L72-L73

guydc:
**What type of PR is this?**

Fixes #3834


zirain (approved):

zirain:
can you add a test for this?

arkodg (approved):

Issue #2837, opened 2024-03-08
Potential Security Risk when SecurityPolicy fails for a Route


sam-burrell:
*Description*:

Our use case is to use the `SecurityPolicy` to define OIDC auth for public routes and we are attaching a `SecurityPolicy` to any type of Route (`HTTPRoute`,  `GRPCRoute`)

- If the SecurityPolicy fails for any reason the default behaviour is for the route to have no SecurityPolicy and be public - for us this feels like a security risk.
- Is there a way for use to ensure that if a `SecurityPolicy` fails the HTTPRoute fails?

arkodg:
this makes sense, it should not be very hard to implement, we already add a direct response of `500` in case any of the filters attached to a route cannot be resolved
what should the UX here be
* knob on `SecurityPolicy` or
* global knob in the startup config

levsha:
TIL the SecurityPolicy may also fail on completely external issues like a (temporary) misconfiguration of the OIDC issuer, with an error like `'Error fetching endpoints from issuer: Get "https://auth.internal.XXXXXX": stopped after 10 redirects'. I guess temporary connectivity problems to the issuer may also cause the gateway to fail open.
It means that this behavior is clearly security vulnerability: the gateway can fail open on issues that can be completely out of control of the cluster administrator. This makes the current implementation of `SecurityPolicy` just useless as no one would want to have an authentication system that fails open ...

arkodg:
if a BackendTrafficPolicy or a SecurityPolicy has not been `Accepted` , my preference is to return a direct response of `5XX` (500/503) for all routes tied to the policies
wdyt @envoyproxy/gateway-maintainers

zhaohuabing:
> if a BackendTrafficPolicy or a SecurityPolicy has not been `Accepted` , my preference is to return a direct response of `5XX` (500/503) for all routes tied to the policies wdyt @envoyproxy/gateway-maintainers

+1

The current behavior of translation is in a best-effort manner, which means that the failed xPolicy will not be accepted but the targeted xRoute will be translated to xDS and sent to the Envoy fleet anyway.   This default behavior is incorrect for BackendTrafficPolicy and a security risk for SecurityPolicy.

PR #3869, opened 2024-07-16
return 500 error for failed SecurityPolicies to avoid unauthorized access to xRoutes

zhaohuabing:
The current implementation of SecurityPolicy translation is in a best-effort manner, which means even if the translation of SecurityPolicy fails, the targeted xRoute will still be translated to xDS. This would result in unauthorized access to the xRoute.

This PR adds a 500 error direct response to the targeted xRoute when the SecurityPolicy translation fails.

fix https://github.com/envoyproxy/gateway/issues/2837

A `FailOpen` knob can be added to the SecurityPolicy API later if needed.


internal/gatewayapi/securitypolicy.go
```diff
@@ -422,13 +420,21 @@ func (t *Translator) translateSecurityPolicyForRoute(
...
+								JWT:           jwt,
+								OIDC:          oidc,
+								BasicAuth:     basicAuth,
+								ExtAuth:       extAuth,
+								Authorization: authorization,
+							}
+						} else {
+							// Return a 500 direct response to avoid unauthorized access
+							r.DirectResponse = &ir.DirectResponse{
+								Body:       ptr.To("Failed to apply SecurityPolicies"),
```
shawnh2, line 435: does the error matter? should we present it?

shawnh2 (approved):
Code LGTM, just got one small question.

Xunzhuo (commented):
Is this approach reasonable? Since this is a SecurityPolicy, should we accpet the best-effort logic to Security related features?

internal/gatewayapi/securitypolicy.go
zhaohuabing: It could be a hint for debugging, but we can't give out details in the response as it may leak control plane configurations that users don't have permission to access.

zhaohuabing:
> Is this approach reasonable? Since this is a SecurityPolicy, should we accpet the best-effort logic to Security related features?

It's reasonable because the current best-effort logic allows unauthorized access to the protected resources, which is a security risk.  A FailOpen knob can be added to the SecurityPolicy API later if needed.

Xunzhuo:
@zhaohuabing make senses, thanks for clarifying

internal/gatewayapi/securitypolicy.go
```diff
@@ -422,13 +420,21 @@ func (t *Translator) translateSecurityPolicyForRoute(
...
+						if errs == nil {
+							r.Security = &ir.SecurityFeatures{
+								CORS:          cors,
+								JWT:           jwt,
+								OIDC:          oidc,
+								BasicAuth:     basicAuth,
+								ExtAuth:       extAuth,
+								Authorization: authorization,
+							}
+						} else {
```
arkodg, line 432: should we assign values for all cases ? may help with debugging in logs, it would look like
```
r.Security = ......
if err != nil {
  r.DirectResponse = .....
}
```

arkodg:
hey @zhaohuabing should we create a seperate issue for BackendTrafficPolicy ?

internal/gatewayapi/securitypolicy.go
zhaohuabing: I saw that the existing filter translation giving out the errMsg in the body, should they also be removed?

arkodg:
hey @zhaohuabing can we also add a e2e for this ?

examples/extension-server/go.sum
zirain, line 1: will be coverd by https://github.com/envoyproxy/gateway/pull/3877

arkodg (approved):
LGTM thanks for fixing this one !

Xunzhuo (approved):

Issue #3872, opened 2024-07-16
Wrong gateway is chosen as HTTPRoute parent


shahar-h:
*Description*:
When creating 2 Gateways with the same name in 2 different namespaces, and then create for each one HTTPRoute, only one of the routes will be attached to a gateway. The reason is that Gateway is selected only by parentRef name in case namespace is omitted. The expected behavior in case namespace is omitted is to select the gateway in the local namespace.

*Repro steps*:
Create a GatewayClass:
```shell
cat <<EOF | kubectl apply -f -
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
EOF
```

Create Gateway and HTTPRoute with `www.example.com` hostname  in `consumer1` namespace:
```shell
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: consumer1
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
  namespace: consumer1
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      hostname: www.example.com
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: backend
  namespace: consumer1
---
apiVersion: v1
kind: Service
metadata:
  name: backend
  namespace: consumer1
  labels:
    app: backend
    service: backend
spec:
  ports:
    - name: http
      port: 3000
      targetPort: 3000
  selector:
    app: backend
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend
  namespace: consumer1
spec:
  replicas: 1
  selector:
    matchLabels:
      app: backend
      version: v1
  template:
    metadata:
      labels:
        app: backend
        version: v1
    spec:
      serviceAccountName: backend
      containers:
        - image: gcr.io/k8s-staging-gateway-api/echo-basic:v20231214-v1.0.0-140-gf544a46e
          imagePullPolicy: IfNotPresent
          name: backend
          ports:
            - containerPort: 3000
          env:
            - name: POD_NAME
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
            - name: NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
[...]

PR #3876, opened 2024-07-17
GetParentReferences should use namespace from RouteContext

zirain:
fix: https://github.com/envoyproxy/gateway/issues/3872


arkodg:
is this a regression from https://github.com/envoyproxy/gateway/pull/3532 ?

zirain:
> is this a regression from #3532 ?

looks like not

```
// GetParentReferences returns the ParentReference of the Route object.
func GetParentReferences(route RouteContext) []gwapiv1.ParentReference {
	rv := reflect.ValueOf(route).Elem()
	kind := rv.FieldByName("Kind").String()
	pr := rv.FieldByName("Spec").FieldByName("ParentRefs")
	if kind == KindHTTPRoute || kind == KindGRPCRoute {
		return pr.Interface().([]gwapiv1.ParentReference)
	}

	parentReferences := make([]gwapiv1.ParentReference, pr.Len())
	for i := 0; i < len(parentReferences); i++ {
		p := pr.Index(i).Interface().(gwapiv1.ParentReference)
		parentReferences[i] = UpgradeParentReference(p)
	}
	return parentReferences
}
```

arkodg (approved):
LGTM thanks !

internal/gatewayapi/testdata/httproute-with-multi-gateways-with-same-name.in.yaml
```diff
@@ -0,0 +1,45 @@
...
+          protocol: HTTP
+          port: 80
+          allowedRoutes:
+            namespaces:
+              from: All
+httpRoutes:
+  - apiVersion: gateway.networking.k8s.io/v1
+    kind: HTTPRoute
+    metadata:
+      namespace: default
```
arkodg, line 34: shouldn't this be `envoy-gateway` ?

arkodg: lets track this bug with a new GH issue

arkodg (approved):

zirain:
hold for a while

zirain:
@arkodg I think now we have the right fix.

internal/gatewayapi/helpers.go
```diff
@@ -109,7 +109,7 @@ func KindDerefOr(kind *gwapiv1.Kind, defaultKind string) string {
 // to a Gateway with the given namespace/name, irrespective of whether a
 // section/listener name has been specified (i.e. a parent ref to a listener
 // on the specified gateway will return "true").
-func IsRefToGateway(parentRef gwapiv1.ParentReference, gateway types.NamespacedName) bool {
+func IsRefToGateway(parentRefNamespace gwapiv1.Namespace, parentRef gwapiv1.ParentReference, gateway types.NamespacedName) bool {
```
arkodg, line 112: should we rename `parentRefNamespace` to `routeNamespace` in this diff, so its clear the value is derived from the namespace of the route, not the parentRef itself (like a Gateway or Listener) ?

zirain: updated

arkodg (approved):
LGTM thanks !

shawnh2 (approved):

PR #3878, opened 2024-07-17
rm gateway-api translation error message from direct response

arkodg:
* Responding back with an error message around translation errors may leak info to internet facing external clients around ingress internals


zirain (approved):

zhaohuabing (approved):
LGTM. Thanks!

PR #3880, opened 2024-07-17
chore: add gnu.org to LINKINATOR_IGNORE

shawnh2:
**What type of PR is this?**

Fix the flaky link detetct:

```plaintext
[200] site/public/latest/tasks/quickstart
[0] https://www.gnu.org/software/make
[200] site/public/latest/tasks/traffic/grpc-routing
```

Fixes #


shawnh2:
dup of #3879

PR #3868, opened 2024-07-16
Add e2e test for load balancing

shawnh2:
**What type of PR is this?**

**What this PR does / why we need it**:

Add e2e test for:
- [x] round robin lb
- [x] consistent hash + source ip lb
- [x] consistent hash + header lb

**Which issue(s) this PR fixes**:

Fix #2153

PR #3886, opened 2024-07-17
fix: nil pointer err during hash load balancing build

shawnh2:
**What type of PR is this?**

Found this while debugging https://github.com/envoyproxy/gateway/pull/3868, will add e2e test for cookie based LB in a follow-up PR.

Fixes #


arkodg (approved):
thanks for catching this one !

guydc (approved):

PR #3888, opened 2024-07-18
disable writing into GatewayClass.Status.SupportedFeatures

arkodg:
disable until the field moves from experiemental to stable so status writes for a GatewayClass dont fail when the datatypes differ

Relates to a breaking change soon to be introduced in https://github.com/kubernetes-sigs/gateway-api/pull/3200.


zirain (dismissed):

Xunzhuo (commented):
@arkodg plz run `make gen` to make CI happy

AliceProxy (approved):

Issue #3511, opened 2024-05-31
gateway pod suddenly died.


gecube:
Good day!

I was playing around quickstart. I definitely did not do something weird. But suddenly I got the next:

```
2024-05-31T16:47:15.757Z	INFO	admin	admin/server.go:37	starting admin server	{"address": "127.0.0.1:19000", "enablePprof": false}
2024-05-31T16:47:15.757Z	INFO	metrics	metrics/register.go:165	initialized metrics pull endpoint	{"address": "0.0.0.0:19001", "endpoint": "/metrics"}
2024-05-31T16:47:15.758Z	INFO	metrics	metrics/register.go:54	starting metrics server	{"address": "0.0.0.0:19001"}
2024-05-31T16:47:15.758Z	INFO	provider	runner/runner.go:41	Using provider	{"runner": "provider", "type": "Kubernetes"}
2024-05-31T16:47:15.759Z	INFO	provider	kubernetes/controller.go:104	created gatewayapi controller	{"runner": "provider"}
I0531 16:47:16.827924       1 request.go:697] Waited for 1.037175908s due to client-side throttling, not priority and fairness, request: GET:https://10.84.0.1:443/apis/flowcontrol.apiserver.k8s.io/v1beta2?timeout=32s
2024-05-31T16:47:17.831Z	INFO	provider	kubernetes/controller.go:1154	ServiceImport CRD not found, skipping ServiceImport watch	{"runner": "provider"}
2024-05-31T16:47:17.844Z	INFO	provider	kubernetes/controller.go:1429	Watching gatewayAPI related objects	{"runner": "provider"}
2024-05-31T16:47:17.844Z	INFO	gateway-api	runner/runner.go:49	started	{"runner": "gateway-api"}
2024-05-31T16:47:17.844Z	INFO	xds-translator	runner/runner.go:47	started	{"runner": "xds-translator"}
2024-05-31T16:47:17.845Z	INFO	xds-server	runner/runner.go:193	loaded TLS certificate and key	{"runner": "xds-server"}
2024-05-31T16:47:17.845Z	INFO	xds-server	runner/runner.go:93	started	{"runner": "xds-server"}
[...]

PR #3893, opened 2024-07-18
add startupProbe to all provisioned containers

arkodg:
* This ensures the readinessProbe kicks in only after the container has started
* max startup time is 300s - 30 (failureThreshold) x 10 (periodSeconds). After this the container is killed and the `restartPolicy` kicks in https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#startup-probe

Fixes: https://github.com/envoyproxy/gateway/issues/3511


AliceProxy (commented):
seems fine to me, but can you please update the branch and pull in the changes for the disabled test and then make sure CI passes first?

Xunzhuo (approved):
/lgtm

shawnh2 (approved):
LGTM

Issue #3894, opened 2024-07-19
EDS didn't update when deployment is create after service


zirain:
1. Create an Gateway with EnvoyProxy refereneced a backendr

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: proxy-config
    namespace: envoy-gateway-system
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: proxy-config
  namespace: envoy-gateway-system
spec:
  logging:
    level:
      default: debug
  telemetry:
    accessLog:
      settings:
        - sinks:
            - type: File
              file:
                path: /dev/stdout
            - type: ALS
              als:
                backendRefs:
                  - name: envoy-als
                    namespace: monitoring
                    port: 8080
                type: HTTP
---
apiVersion: v1
kind: Service
metadata:
  name: envoy-als
  namespace: monitoring
spec:
  selector:
    app: envoy-als
  ports:
    - name: grpc-als
      protocol: TCP
      appProtocol: grpc
      port: 8080
      targetPort: 8080
    - name: http-monitoring
      protocol: TCP
      port: 19001
      targetPort: 19001
```

2: Create deployment in monitoring namespace
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: envoy-als
  namespace: monitoring
spec:
  replicas: 1
  selector:
    matchLabels:
      app: envoy-als
  template:
    metadata:
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "19001"
      labels:
        app: envoy-als
    spec:
      containers:
        - name: envoy-als
          image: ghcr.io/zirain/als:latest
          imagePullPolicy: Always
          ports:
            - containerPort: 8080
            - containerPort: 19001
---
apiVersion: v1
kind: Service
metadata:
  name: envoy-als
  namespace: monitoring
spec:
  selector:
    app: envoy-als
  ports:
    - name: grpc-als
      protocol: TCP
      appProtocol: grpc
      port: 8080
[...]

PR #3895, opened 2024-07-19
validate for reconcile should check reference from EnvoyProxy

zirain:
fix: #3894


arkodg (approved):
good catch !

shawnh2 (approved):

Xunzhuo (approved):
Nice catch

Issue #3889, opened 2024-07-18
Add Support for modifying Container SecurityContext for Certgen job in Helm


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zhaohuabing:
Close this one as the default SecurityContext has been set to minimum permissions. 
Customization on Cetgen job SecurityContext should not be required anymore.

Issue #3887, opened 2024-07-18
Add Support for modifying Pod and Container SecurityContext for Envoy Gateway deploment in Helm


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Manoramsharma:
Hi @zhaohuabing after following this discussion for adding support for modifying `SecurityContext` for EG components like `Gateaway` and `certGen`. I came to find that how it can be something to enhance security by initially setting a default(minimize) permissions to the EG components but can be later modified by changing securityContext using this support. 

I saw this [PR](https://github.com/envoyproxy/gateway/pull/3940)  that configures the SecurityContext for EG components to minimize the required permission for operation. 

- CertGen
- Gateway
- ShutdownManager
- Envoy

How can I move forward with this issue as I would like to contribute to add this support for modifying the securityContext?

cc: @arkodg

zhaohuabing:
@Manoramsharma  , thank you for being willing to take this on. Before we proceed with exposing these SecurityContexts to the Helm chart Values file, I’d like to understand a concrete use case. We've already made efforts to minimize permissions to only what the EG components require.

Here’s my suggestion:

If there are any unnecessary permissions in the default settings that EG doesn’t need, we should consider removing them.
If a user has a requirement where the default settings don't fulfill their use case—such as needing additional permissions due to a customized Envoy deployment—we can then discuss how best to expose these SecurityContexts to the Helm chart Values file.
Does this approach make sense? Do you have a specific use case where you need to customize these SecurityContexts?

akorosi:
Hi,
We are adding Opentelemetry Collector Sidecar to gateway pods using specific annotations, but we see pods are failing, because 
`kubelet            Error: container has runAsNonRoot and image will run as root (pod: "envoy-gateway-77f56f76d-4lf2x_gateway(9ceff3d6-7a4a-4378-83e3-eb33e0452541)", container: otc-container)`

If I manually edit the Deployment and change securityContext to false, then pod can start properly. It would be good that this attribute could be customizable

zirain:
will https://gateway.envoyproxy.io/docs/api/extension_types/#kubernetespatchspec help?

arkodg:
for envoy proxy you can set the pod security context using `EnvoyProxy` API https://gateway.envoyproxy.io/docs/api/extension_types/#kubernetespodspec

[...]

PR #3897, opened 2024-07-19
helm: support for modifying SecurityContext for EG and CertGen pod

zhaohuabing:
implement: 
* https://github.com/envoyproxy/gateway/issues/3889
* https://github.com/envoyproxy/gateway/issues/3887

Please note:
* Only the Pod level SecurityContext configuration has been exposed to the helm Values file, as there's only one container in the EG and CertGen Pod.
* By default, the Pod level SecurityContext is nil, the original default SecurityContext setting has been merged to the Container level for consistency.

If this is not the best practice for configuring`SecurityContext`, I would appreciate your guidance. @envoyproxy/gateway-maintainers 

Also, to enhance security, the EG and CertGen Containers' SecurityContexts have been restricted to minimize permissions.

It would be better to include these changes in v1.1.0 since they're security-related hardening.


charts/gateway-helm/templates/certgen.yaml
```diff
@@ -38,12 +38,21 @@ spec:
...
+          runAsUser: 65534
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+            - ALL
       {{- include "eg.image.pullSecrets" . | nindent 6 }}
       restartPolicy: Never
+      {{- with .Values.certgen.job.securityContext }}
```
arkodg, line 52: will there be two `securityContext` sections at the same level if a user sets this value ?
is this an `append` or `override` style input field ?

zhaohuabing: No, it won't be two `SecurityContext` on the Pod level, if unset in the Values file, the Pod level `SecurityContext` will be nil, only Container level `SecurityContext` exists.

Example:  Default: https://github.com/envoyproxy/gateway/blob/5e31f2901903992cbde6471fb512dea3ee721ff8/test/helm/gateway-helm/global-images-config.out.yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: envoy-gateway
  namespace: 'envoy-gateway-system'
spec:
  template:
    spec:
      containers:
      - args:
        ....
        securityContext:
          runAsNonRoot: true
          capabilities:
            drop:
            - ALL
         ...
      serviceAccountName: envoy-gateway
      terminationGracePeriodSeconds: 10
      volumes:
      - configMap:
          defaultMode: 420
          name: envoy-gateway-config
        name: envoy-gateway-config
      - name: certs
        secret:
          secretName: envoy-gateway
  ```
  With customized `SecurityContex`: https://github.com/envoyproxy/gateway/blob/5e31f2901903992cbde6471fb512dea3ee721ff8/test/helm/gateway-helm/security-context.out.yaml
  
  ```yaml
  apiVersion: apps/v1
kind: Deployment
metadata:
  name: envoy-gateway
  namespace: 'envoy-gateway-system'
spec:
  template:
    spec:
      containers:
      - args:
        ....
        securityContext:
          runAsNonRoot: true
          capabilities:
            drop:
            - ALL
         ...
      securityContext:
        fsGroup: 1234
        runAsGroup: 1234
        runAsUser: 1234
      serviceAccountName: envoy-gateway
      terminationGracePeriodSeconds: 10
      volumes:
      - configMap:
          defaultMode: 420
          name: envoy-gateway-config
        name: envoy-gateway-config
      - name: certs
        secret:
          secretName: envoy-gateway
  ```

zhaohuabing: Not sure this is the best way to configure `SecurityContext`, would appreciate your input here @envoyproxy/gateway-maintainers

guydc:
@zhaohuabing Is this change backwards compatible? From them tests, it looks like some defaults are changing to be more hardened and pod level config is removed by default while container-level config is added. 
- Do we need to offer some guidance to users in release notes on this change? 
- Should we add some default values to keep the behavior backwards compatible? If not, is there a way for users to replicate the configuration that existed before this change?

@liorokman  - do you think that there's any impact on running an extension server as a sidecar to EG communicating over UDS?

zhaohuabing:
@guydc This is back compatible, but I'd like to hear more feedback before merging it. We don't have to land it on v1.1.0.

mcharriere:
hi, I'm looking forward for this improvement. My 2 cents:

Because of the way that SecurityContext work, I'm not sure if it makes much sense this strategy for configuration.

the SecurityContext in the containers take precedence over the PodSecurityContext; so effectively this means that no matter what the user sets in the exposed SecurityContext, the hardcoded value will be applied to the container.
Although I see no reason to relax these security settings, it might be confusing.

What I've seen in other charts is to expose both the `securityContext` for the container and the `podSecurityContext` and set some reasonable defaults in `values.yaml`.

zhaohuabing:
@mcharriere Thanks for your insight! 

I have opened a new PR https://github.com/envoyproxy/gateway/pull/3940 to add a reasonable default for container-level SecurityContext for all EG components. I hope we can use the minimum permissions as the default, and then expose SecurityConext to Values file if needed.

arkodg:
hey @zhaohuabing my suggestion would be to start off by adding support for modifying pod securityContext and making it a override

zhaohuabing:
> hey @zhaohuabing my suggestion would be to start off by adding support for modifying pod securityContext and making it a override

The SecurityContext for Envoy pod can be overridden with `EnvoyProxy.Spec.provider.kubernetes.envoyDeployment.pod.securityContext`. 

For Envoy Gateway and Cert Job pod, we may not need to customize their SecurityContexts. Should we wait until users require them?

PR #3917, opened 2024-07-21
printing name of the target when a policy fails to attach

sadovnikov:
**What type of PR is this?**

Policies in the `v1.1.x` versions will have multiple targets. This PR makes error messages on failures to attach to a target more specific.

No issues were logged for this


zirain (dismissed):

zirain:
it would be better to have a test for this.

sadovnikov:
> it would be better to have a test for this.

@zirain, the `resolveSecurityPolicyRouteTargetRef()` and other functions have parameters with complex structures. The code will be more testable if the validation of the namespaces and the attachment is extracted from all four files into a helper function. However, I believe there is a problem with the validation code too. Could you take a look at #3918?

PR #3928, opened 2024-07-22
feat(translator): add regex origins for CORS

Demacr:

I've added new field to `CORS` struct to get ability set flexible regexes to CORSes, because current situation allow only prefix wildcard. So in case of:
```
      - "tld\\.domain(?:\\..*)?"
      - ".*:\\/\\/localhost(?:\\:\\d+)?"
      - "http:\\/\\/((25[0-5]|(2[0-4]|1\\d|[1-9]|)\\d)\\.?\\b){4}:30\\d\\d"
      - null
```
like mobile app origins or when there is no origing (`null`) it is impossible to set CORS.
Second point why I did new field is to have back compatibility with existent configuration.

### DIFF
--- a/api/v1alpha1/cors_types.go
+++ b/api/v1alpha1/cors_types.go
@@ -26,11 +26,22 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 // +kubebuilder:validation:Pattern=`^(\*|https?:\/\/(\*|(\*\.)?(([\w-]+\.?)+)?[\w-]+)(:\d{1,5})?)$`
 type Origin string
 
+// RegexOrigin is
+//
+// For example, the following are valid origins:
+// - https://foo.example.com
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type RegexOrigin string
+
 // CORS defines the configuration for Cross-Origin Resource Sharing (CORS).
 type CORS struct {
 	// AllowOrigins defines the origins that are allowed to make requests.
 	// +kubebuilder:validation:MinItems=1
 	AllowOrigins []Origin `json:"allowOrigins,omitempty" yaml:"allowOrigins"`
+	// AllowRegexOrigins defines the origin regexes that are allowed to make requests.
+	AllowRegexOrigins []RegexOrigin `json:"allowRegexOrigins,omitempty" yaml:"allowRegexOrigins"`
 	// AllowMethods defines the methods that are allowed to make requests.
 	// +kubebuilder:validation:MinItems=1
 	AllowMethods []string `json:"allowMethods,omitempty" yaml:"allowMethods"`
--- a/api/v1alpha1/cors_types.go
+++ b/api/v1alpha1/cors_types.go
@@ -26,10 +26,16 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 // +kubebuilder:validation:Pattern=`^(\*|https?:\/\/(\*|(\*\.)?(([\w-]+\.?)+)?[\w-]+)(:\d{1,5})?)$`
 type Origin string
 
-// RegexOrigin is
+// RegexOrigin is defined by full-fledged regex, so it is possible to use any
+// scheme, domain, port or any other value. The value should be valid regex
+// string.
+// Syntax should be according to https://github.com/google/re2/wiki/Syntax.
+// Also, backslashes must be escaped.
 //
 // For example, the following are valid origins:
 // - https://foo.example.com
+// - https?://(.*\\.)?example\\.com
+// - .*://localhost(?:\\:\\d+)?
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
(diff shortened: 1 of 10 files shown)


internal/gatewayapi/securitypolicy.go
```diff
@@ -563,6 +563,12 @@ func (t *Translator) buildCORS(cors *egv1a1.CORS) *ir.CORS {
 		}
 	}
 
+	for _, origin := range cors.AllowRegexOrigins {
+		allowOrigins = append(allowOrigins, &ir.StringMatch{
+			SafeRegex: (*string)(&origin),
```
zirain, line 568: is there any way to make sure the regex is valid?

zhaohuabing:
> like mobile app origins or when there is no origing (null) it is impossible to set CORS.

@Demacr Thanks for the PR. Could you please elaborate on this?  It would help if you can give a real example where wildcard match doesn't solve but regex does. Thanks.

internal/gatewayapi/securitypolicy.go
Demacr: Hm, I'll look for validator.

Demacr:
@zhaohuabing I've mentioned rules, which cannot be matched by prefixed wildcard, in the PR message:
- 1st is our origin from mobile applications, it should have suffixed wildcard
- 2nd & 3rd could pass `Origin` struct validator
- 4th `null` origin uses our custom Figma's plugin, which call our API with such origin.

Finally, looks like it's possible to add just suffixed wildcard and to add posibility make schemaless origin.

internal/gatewayapi/securitypolicy.go
zhaohuabing: We can call regex.Validate(&origin) to validate the regex.

zhaohuabing:
> @zhaohuabing I've mentioned rules, which cannot be matched by prefixed wildcard, in the PR message:
> 
> * 1st is our origin from mobile applications, it should have suffixed wildcard
> * 2nd could be replaced by `Origin` struct validated regex
> * 3rd one is really rare, but impossible to change by regex which can pass `Origin` struct validator
> * 4th `null` origin uses our custom Figma's plugin, which call our API with such origin.
> 
> Finally, looks like it's possible to add just suffixed wildcard and to add posibility make schemaless origin. Excepting IPs.

Thanks, I think this request is fair enough. LGTM, and defer to other @envoyproxy/gateway-maintainers for any suggestions on the API change.

api/v1alpha1/cors_types.go
```diff
@@ -26,11 +26,28 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
...
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type RegexOrigin string
+
 // CORS defines the configuration for Cross-Origin Resource Sharing (CORS).
 type CORS struct {
 	// AllowOrigins defines the origins that are allowed to make requests.
 	// +kubebuilder:validation:MinItems=1
 	AllowOrigins []Origin `json:"allowOrigins,omitempty" yaml:"allowOrigins"`
```
zirain, line 48: Can we just loose the validation on AllowOrigins instead of introuducing an new field?

Demacr: I thought about it, but IMO it could be too radical. As I discovered, Envoy itself uses a lot of [types](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/string.proto#envoy-v3-api-msg-type-matcher-v3-stringmatcher) of matching string. So, my point is, if Envoy's `type.matcher.v3.StringMatcher` `Exact` type is faster than `SafeRegex` so it's really good point to have two options, for fast `Exact` and slower `SafeRegex`

zirain: I would expected a style as following:
```yaml
AllowOrigins:
  - type: Exact
    exact: xxxxx
  - type: Prefix
    prefrix: yyy
  - type: Regex
    regex: zzzz

```

zhaohuabing: Wildcard is a special case of regex, and it's commonly used for hostname matching.

Demacr: @zirain hm, that's really good idea, but it'll brake back compatibility. Maybe, make this by new field or by issuing new api version

arkodg: I prefer the suggestions provided by @zirain and @zhaohuabing, lets reuse the existing API field
1. the CEL validation will need to be relaxed
2. the logic https://github.com/envoyproxy/gateway/blob/8d82ea9748b57a0dada0bea133fbac018f9ba972/internal/gatewayapi/securitypolicy.go#L523 will need to be revisited to account for `Exact` , `Wildcard Hostnames` and `Regex` based hostnames

Demacr: @arkodg could you clarify your point: do you want to keep the only one `allowOrigins` with zirain's `type` field? Or zhaohuabing's way of two options `allowOrigins` with (Exact and Wildcard) and another with Regex.

arkodg: reuse `AllowOrigins []Origin` for regex as well w/o any breaking change to the API
it looks like contour does the same thing https://projectcontour.io/docs/1.28/config/cors/

arkodg:
hey @Demacr still plan on working on this one ?

api/v1alpha1/cors_types.go
arkodg: sure sgtm

arkodg:

PR #3929, opened 2024-07-22
feat: support custom StatsdExporter image

zirain:
Before make ratelimit support native prometheus endpoint, this patch will close https://github.com/envoyproxy/gateway/issues/3049.


zhaohuabing:
@zirain Could we use `KubernetesPatchSpec` as a workaround for this?  

A `StatsdExporter` in `EnvoyGateway` API seems confusing for me as this is for Ratelimit, not for Envoy Gateway deployment. (Even the existing `ShutdownManager` should not be there, it's my bad).

zirain:
> @zirain Could we use `KubernetesPatchSpec` as a workaround for this?
> 
> A `StatsdExporter` in `EnvoyGateway` API seems confusing for me as this is for Ratelimit, not for Envoy Gateway deployment. (Even the existing `ShutdownManager` should not be there, it's my bad).

Of course you can do that way, but TBH I think this's not a too bad idea.

zhaohuabing:
I suggest we support it with `KubernetesPatchSpec` and wait for native prometheus endpoint support in ratelimit,  @envoyproxy/gateway-maintainers please weight in.

zirain:
close since `KubernetesPatchSpec` support this.

PR #3930, opened 2024-07-22
feat: finalize envoyproxy for gatewayclass 

cnvergence:
**What type of PR is this?**

Restore [work](https://github.com/envoyproxy/gateway/pull/1534) on EnvoyProxy finalizers, only for those referenced by the GatewayClass.
We have ran into issues with Merged Gateways deployments. https://github.com/envoyproxy/gateway/pull/3533#issuecomment-2149533649

Fixes https://github.com/envoyproxy/gateway/issues/865


internal/provider/kubernetes/controller.go
```diff
@@ -991,28 +1018,56 @@ func (r *gatewayAPIReconciler) processBackends(ctx context.Context, resourceTree
...
 
-// addFinalizer adds the gatewayclass finalizer to the provided gc, if it doesn't exist.
-func (r *gatewayAPIReconciler) addFinalizer(ctx context.Context, gc *gwapiv1.GatewayClass) error {
-	if !slice.ContainsString(gc.Finalizers, gatewayClassFinalizer) {
-		base := client.MergeFrom(gc.DeepCopy())
-		gc.Finalizers = append(gc.Finalizers, gatewayClassFinalizer)
-		if err := r.client.Patch(ctx, gc, base); err != nil {
-			return fmt.Errorf("failed to add finalizer to gatewayclass %s: %w", gc.Name, err)
+// addFinalizer adds the gatewayclass or envoyproxy finalizer to the provided object, if it doesn't exist.
+func (r *gatewayAPIReconciler) addFinalizer(ctx context.Context, obj client.Object) error {
```
zhaohuabing, line 1048: Nit, non-blocking:
I would split this method into two: `addGatewayClassFinalizer` and `addEnvoyProxyFinalizer`. IMO, this could make the code more readable.

internal/provider/kubernetes/controller.go
```diff
@@ -991,28 +1018,56 @@ func (r *gatewayAPIReconciler) processBackends(ctx context.Context, resourceTree
...
+			objType.Finalizers = slice.RemoveString(objType.Finalizers, gatewayClassFinalizer)
+			if err := r.client.Patch(ctx, objType, base); err != nil {
+				return fmt.Errorf("failed to add finalizer to Gateway Class %s: %w", objType.Name, err)
+			}
 		}
+		return nil
+	case *egv1a1.EnvoyProxy:
+		if slice.ContainsString(objType.Finalizers, gatewayClassFinalizer) {
+			base := client.MergeFrom(objType.DeepCopy())
+			objType.Finalizers = slice.RemoveString(objType.Finalizers, gatewayClassFinalizer)
```
zhaohuabing, line 1036: Should use a string like "gatewayclass-exists-finalizer.gateway.envoyproxy.github.com" instead of "gateway-exists-finalizer.gateway.networking.k8s.io"

zhaohuabing:
HI @cnvergence thanks for picking this up. This PR overall looks good; I just a few comments.

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,46 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
+		}
+	} else {
+		// Add finalizer to EnvoyProxy if it references the Envoy Proxy.
+		if classRefsEnvoyProxy(managedGC, resourceTree.EnvoyProxyForGatewayClass) {
+			if err := r.addFinalizer(ctx, resourceTree.EnvoyProxyForGatewayClass); err != nil {
+				return fmt.Errorf("failed adding finalizer to gatewayclass %s: %w", managedGC.Name, err)
+			}
+		}
+		// Add finalizer to GatewayClass if there are gateways.
+		if err := r.addFinalizer(ctx, managedGC); err != nil {
```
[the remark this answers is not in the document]
cnvergence: I wanted to keep the logic of not having finalizers, when there are no gateways attached to the gc

internal/provider/kubernetes/controller.go
cnvergence: I do not have a strong opinion on this as well, as we won't be adding any new finalizers, so there is no big need for worrying about redundancy

internal/provider/kubernetes/controller.go
```diff
@@ -198,7 +198,7 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 
 		// Process the parametersRef of the accepted GatewayClass.
 		// This should run before processGateways and processBackendRefs
-		if managedGC.Spec.ParametersRef != nil && managedGC.DeletionTimestamp == nil {
+		if managedGC.Spec.ParametersRef != nil {
```
shawnh2, line 201: curious why are removing the `DeletionTimestamp` check ?

cnvergence: Leftover, I was doing some tests and wanted to avoid the issue of disabling the mergeGateways feature when GatewayClass is being deleted.
Let's restore it and solve it in the next iteration.

internal/provider/kubernetes/controller.go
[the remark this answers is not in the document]
zhaohuabing: @cnvergence Fix this then it should be good to go :-)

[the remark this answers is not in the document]
cnvergence: @zhaohuabing you mean moving this out of if-else block? I have moved it out

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,56 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
 }
 
+// processFinalizers encapsulates logic for managing finalizers on GatewayClass and EnvoyProxy objects.
+func (r *gatewayAPIReconciler) processFinalizers(ctx context.Context, managedGC *gwapiv1.GatewayClass, resourceTree *gatewayapi.Resources) error {
+	// Add finalizer to EnvoyProxy if it is referenced by the GatewayClass.
+	if classRefsEnvoyProxy(managedGC, resourceTree.EnvoyProxyForGatewayClass) && resourceTree.EnvoyProxyForGatewayClass.DeletionTimestamp.IsZero() && managedGC.DeletionTimestamp.IsZero() {
+		if err := r.addFinalizer(ctx, resourceTree.EnvoyProxyForGatewayClass); err != nil {
+			return fmt.Errorf("failed adding finalizer to Envoy Proxy %s: %w", resourceTree.EnvoyProxyForGatewayClass.Name, err)
+		}
+		if err := r.addFinalizer(ctx, managedGC); err != nil {
```
zhaohuabing, line 779: Do we need to add a finalizer to the GatewayClass here?

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,56 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
 	return nil, nil
 }
 
+// processFinalizers encapsulates logic for managing finalizers on GatewayClass and EnvoyProxy objects.
+func (r *gatewayAPIReconciler) processFinalizers(ctx context.Context, managedGC *gwapiv1.GatewayClass, resourceTree *gatewayapi.Resources) error {
+	// Add finalizer to EnvoyProxy if it is referenced by the GatewayClass.
+	if classRefsEnvoyProxy(managedGC, resourceTree.EnvoyProxyForGatewayClass) && resourceTree.EnvoyProxyForGatewayClass.DeletionTimestamp.IsZero() && managedGC.DeletionTimestamp.IsZero() {
+		if err := r.addFinalizer(ctx, resourceTree.EnvoyProxyForGatewayClass); err != nil {
+			return fmt.Errorf("failed adding finalizer to Envoy Proxy %s: %w", resourceTree.EnvoyProxyForGatewayClass.Name, err)
```
shawnh2, line 777:
```suggestion
			return fmt.Errorf("failed to add finalizer to Envoy Proxy %s: %w", resourceTree.EnvoyProxyForGatewayClass.Name, err)
```

internal/provider/kubernetes/controller.go
cnvergence: I want to add it to GatewayClass, when we reference EnvoyProxy and no Gateways are attached.
This is needed for situation, when GatewayClass is deleted before EnvoyProxy

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,56 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
+		}
+		if err := r.addFinalizer(ctx, managedGC); err != nil {
+			return fmt.Errorf("failed adding finalizer to gatewayclass %s: %w", managedGC.Name, err)
+		}
+	}
+
+	if len(resourceTree.Gateways) > 0 && managedGC.DeletionTimestamp.IsZero() {
+		if err := r.addFinalizer(ctx, managedGC); err != nil {
+			return fmt.Errorf("failed adding finalizer to gatewayclass %s: %w", managedGC.Name, err)
```
shawnh2, line 786:
```suggestion
			return fmt.Errorf("failed to add finalizer to gatewayclass %s: %w", managedGC.Name, err)
```

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,56 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
 	return nil, nil
 }
 
+// processFinalizers encapsulates logic for managing finalizers on GatewayClass and EnvoyProxy objects.
+func (r *gatewayAPIReconciler) processFinalizers(ctx context.Context, managedGC *gwapiv1.GatewayClass, resourceTree *gatewayapi.Resources) error {
```
shawnh2, line 773: done some test, found a behaviour seems like a bug to me:

1. running `kubectl apply -f examples/kubernetes/merged-gateways.yaml`
2. `kubectl delete envoyproxy -n envoy-gateway-system custom-proxy-config` will hold
3. `kubectl delete gc merged-eg` will hold
4. `kubectl delete gtw merged-eg-1 merged-eg-2 merged-eg-3`, all gtws are removed successfully, and step 3 return successfully, but step 2 still hold. By inspecting the finializers of `custom-proxy-config`, the finializer `gatewayclass-exists-finalizer.gateway.envoyproxy.github.com` still exist.

cnvergence: It seems like that's correct, the issue is that when the `GatewayClass` gets `deletionTimestamp`, we don't process `EnvoyProxy`, hence processed `resourceTree.EnvoyGatewayForGatewayClass` will be `nil`. (it also cause mergeGateways to be set to nil).
I can also try to fix this in this PR, as this makes the finalizers more complicated.

internal/provider/kubernetes/controller.go
zhaohuabing: Sorry, I may miss something here. Would it be a problem if a GatewayClass is deleted when its referencing EnvoyProxy exists?

internal/provider/kubernetes/controller.go
```diff
@@ -782,6 +769,47 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
+			return fmt.Errorf("failed to add finalizer to gatewayclass %s: %w", managedGC.Name, err)
+		}
+	}
+
+	// Check previously stored referenced Envoy Proxy and remove finalizer if not referenced anymore by the current managed GatewayClass.
+	managedResources := r.resources.GetResourcesByGatewayClass(managedGC.Name)
+	if managedResources != nil {
+		refEnvoyProxy := managedResources.EnvoyProxyForGatewayClass
+		if refEnvoyProxy != nil && !classRefsEnvoyProxy(managedGC, refEnvoyProxy) {
+			if err := r.removeFinalizer(ctx, refEnvoyProxy); err != nil {
```
zhaohuabing, line 786: EnvoyProxy can be referenced by multiple GatewayClasses.

internal/provider/kubernetes/controller.go
cnvergence: No, it is only checking if `EnvoyProxy` exists for the given `GatewayClass`.
As for now, it is working correctly for both scenarios, so `GatewayClass` deletion will hold until Gateways are cleaned up and then finalizers are removed from `GatewayClass` and `EnvoyProxy`.

internal/provider/kubernetes/controller.go
cnvergence: that's true, I have also tested this scenario, and `EnvoyProxy` finalizer will be there if it is referenced by another `GatewayClass`

internal/provider/kubernetes/controller.go
```diff
@@ -1613,6 +1669,17 @@ func (r *gatewayAPIReconciler) processGatewayClassParamsRef(ctx context.Context,
 	if err := r.processEnvoyProxy(ep, resourceMap); err != nil {
 		return err
 	}
+
+	if classRefsEnvoyProxy(gc, ep) {
+		if err := r.addFinalizer(ctx, ep); err != nil {
+			return fmt.Errorf("failed to add finalizer to Envoy Proxy %s: %w", ep.Name, err)
+		}
+		if err := r.addFinalizer(ctx, gc); err != nil {
```
shawnh2, line 1677: we should not add finalizer for gc at this moment, nor it will add `gateway-exists-finalizer.gateway.networking.k8s.io` finalizer for those gc that have no gateways attach to it.

we should this finalizer for those gc that have gateways attach to it.

internal/provider/kubernetes/controller.go
shawnh2: same thing happens here:

1. running `kubectl apply -f examples/kubernetes/merged-gateways.yaml` and create another GC (`foobar`) have envoyproxy `custom-proxy-config` as its `parametersRef`
2. `kubectl delete envoyproxy -n envoy-gateway-system custom-proxy-config` will hold
3. `kubectl delete gc merged-eg` will hold
4. `kubectl delete gtw merged-eg-1 merged-eg-2 merged-eg-3`, all gtws are removed successfully, but step 3 still hold (unexpected), surprisingly step 2 return (unexpected)

cnvergence: getting back to this, let me take a look, it seems like we have another problem, now with getting errors within processParamsRef

cnvergence: Thanks @shawnh2 @zhaohuabing
I needed to change the approach, because of the multiple gateway classes

cnvergence:
@shawnh2 @zhaohuabing  had you chance to take a look again?

arkodg:

PR #3940, opened 2024-07-23
set default SecurityContext for EG components

zhaohuabing:
This PR configures the default container SecurityContext for EG components to minimize the required permission for operation.
* CertGen
* Gateway
* RateLimit
* Statsd-exporter
* ShutdownManager
* Envoy

The default value is intended to meet the best practice for SecurityContext for the default images provided by the EG chart. For customized images that need other permission beyond the default, I plan to submit a follow-up PR to expose theses SecurityContext to the helm Values file.

CertGen
```yaml
        securityContext:
          capabilities:
            drop:
            - ALL
          privileged: false
          readOnlyRootFilesystem: true
          runAsNonRoot: true
          runAsGroup: 65534
          runAsUser: 65534
         seccompProfile:
           type: RuntimeDefault    
```

Gateway 
```yaml
    securityContext:
      capabilities:
        drop:
        - ALL
      privileged: false
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
```

RateLimit
```yaml
    securityContext:
      capabilities:
        drop:
        - ALL
      privileged: false
      readOnlyRootFilesystem: true
      runAsGroup: 65534
      runAsNonRoot: true
      runAsUser: 65534
      seccompProfile:
        type: RuntimeDefault
```

Statsd-exporter
```yaml
    securityContext:
      capabilities:
        drop:
        - ALL
      privileged: false
      readOnlyRootFilesystem: true
      runAsGroup: 65534
      runAsNonRoot: true
      runAsUser: 65534
      seccompProfile:
        type: RuntimeDefault
```
    
ShutdownManager
```yaml
    securityContext:
      capabilities:
        drop:
        - ALL
      privileged: false
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
```

Envoy
```yaml
    securityContext:
      capabilities:
        drop:
        - ALL
      privileged: false
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
```
@Demonsthere @arkodg


guydc:
CC @liorokman - any issue with integrating an extension server over UDS due to these changes?

liorokman:
> CC @liorokman - any issue with integrating an extension server over UDS due to these changes?

No. UDS doesn't need any privileges, other than that the two peers should both have read/write filesystem permissions to the socket file.

zhaohuabing:
> > CC @liorokman - any issue with integrating an extension server over UDS due to these changes?
> 
> No. UDS doesn't need any privileges, other than that the two peers should both have read/write filesystem permissions to the socket file.

@guydc @liorokman should we add the read/write filesystem permission to the default SecurityContext, or exclude it but allow users to customize the SecurityContext in the API when they need it? WDYT?

zhaohuabing:
> > CC @liorokman - any issue with integrating an extension server over UDS due to these changes?
> 
> No. UDS doesn't need any privileges, other than that the two peers should both have read/write filesystem permissions to the socket file.

Remove `readOnlyRootFilesystem: true` from Envoy's default SecurityContext as Envoy needs write permission for file log and UDS socket.

guydc:
CC @shahar-h

AliceProxy (commented):
I think these are good defaults but for the helm charts, I would prefer adding the config under the values and inserting it into the templates rather than hard coding it into the templates. That way, if a user is going to make any changes to Envoy Gateway such as using a custom built image/etc. they don't have to fork the repo to make changes.

liorokman:
> `readOnlyRootFilesystem: true` has been removed from Envoy's default SecurityContext as Envoy needs write permission for file log and UDS socket.

If UDS is being used as a separate side-car in the pod, then the file-system containing the socket file must be shared between the pods. This is usually done with a volume, which would not be the root filesystem.

`readOnlyRootFilesystem` can be set to true without any issue for the extension server.

Don't know about the file log requirements, though.

zhaohuabing:
> I think these are good defaults but for the helm charts, I would prefer adding the config under the values and inserting it into the templates rather than hard coding it into the templates. That way, if a user is going to make any changes to Envoy Gateway such as using a custom built image/etc. they don't have to fork the repo to make changes.

@AliceProxy Thank you very much for your input!

This PR tries to introduce reasonable default values for container SecurityContexts to minimize permissions to only what is necessary. I plan to submit [a follow-up PR](https://github.com/envoyproxy/gateway/pull/3897)  to expose some of these SecurityContext settings in the Helm values files. Given their likelihood for customization, I believe we should focus on exposing the SecurityContext settings for EG and Envoy in the Helm values files.

zhaohuabing:
> > `readOnlyRootFilesystem: true` has been removed from Envoy's default SecurityContext as Envoy needs write permission for file log and UDS socket.
> 
> If UDS is being used as a separate side-car in the pod, then the file-system containing the socket file must be shared between the pods. This is usually done with a volume, which would not be the root filesystem.
> 
> `readOnlyRootFilesystem` can be set to true without any issue for the extension server.
> 
> Don't know about the file log requirements, though.

AccessLog can be output to a file, the default path is "/dev/stdout", but it can be a local file. @zirain do you think we should make the root filesystem readonly by default?

Demonsthere:
Hi there! I think the best way compromise here would be to adhere to the k8s standards when it comes to security levels. The default values for components should follow the baseline policy from https://kubernetes.io/docs/concepts/security/pod-security-standards/ while the helm chart could allow for customization of both pod and container level policy to allow tetrate to run in a Restricted level namespace. 
A foolproof method of checking that is simply setting the labels on the namespace before pods are installed into it: 
```
apiVersion: v1
kind: Namespace
metadata:
  name: my-baseline-namespace
  labels:
    pod-security.kubernetes.io/enforce: baseline
    pod-security.kubernetes.io/enforce-version: latest
    pod-security.kubernetes.io/warn: baseline
    pod-security.kubernetes.io/warn-version: latest
---
apiVersion: v1
kind: Namespace
metadata:
  name: my-restricted-namespace
  labels:
    pod-security.kubernetes.io/enforce: restricted
    pod-security.kubernetes.io/enforce-version: latest
    pod-security.kubernetes.io/warn: restricted
    pod-security.kubernetes.io/warn-version: latest
```

@zhaohuabing when it comes to the root filesystem, if we could make it read only by default, then we may be able to run tetrate in the restricted mode, which would be the best case scenario for a edge gateway sitting between external users and internal resources 😉

zhaohuabing:
> when it comes to the root filesystem, if we could make it read only by default, then we may be able to run EG in the restricted mode, which would be the best case scenario for a edge gateway sitting between external users and internal resources 😉

@Demonsthere Some of these containers need write permission to run: 
* Gateway  need write permission for Wasm cache
* ShutdownManager need write permission for the creation of the shutdown check file
* Envoy need write permission for some operations such as file log, UDS socket, etc.

guydc:
LGTM. For the filesystem use case: If there's an option to opt-out of all the features that require FS (e.g. shutdown manager) access, then I propose to make it configurable and allow users to harden that value. Otherwise, let's go with a less hardened approach by default and perhaps track this as a requirement?

guydc (approved):

AliceProxy (approved):

arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM

Issue #3918, opened 2024-07-21
Namespaces of policies and targets


sadovnikov:
*Description*:

I'm testing policies' support of multiple targets in `v1.1.0-rc.1` and created #3917 to make the failures to attach messages more informative.

Although it's not directly related to that, I also noticed a potential problem with the validation of the namespaces. In all changed files, the `targetNs` value is [assigned](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L251C14-L251C30) from `policy.Namespace` and then it's [checked for mismatch](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L269) with itself. 

The `targetNs := policy.Namespace` assignment seems wrong.

[optional *Relevant Links*:]

* https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L251C14-L251C30
* https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L269

sadovnikov:
The #3917 is replaced with #3943

PR #3943, opened 2024-07-23
printing name of the target when a policy fails to attach

sadovnikov:
**What type of PR is this?**

Policies in the `v1.1.x` versions will have multiple targets. This PR makes error messages on failures to attach to a target more specific.

No issues were logged for this

@zirain, this PR replaces #3917, which I messed up with the wrong commits. It also includes changes in the test files.

### DIFF
--- a/internal/gatewayapi/testdata/securitypolicy-status-conditions.out.yaml
+++ b/internal/gatewayapi/testdata/securitypolicy-status-conditions.out.yaml
@@ -286,7 +286,7 @@ securityPolicies:
         namespace: envoy-gateway
       conditions:
       - lastTransitionTime: null
-        message: Unable to target HTTPRoute, another SecurityPolicy has already attached
+        message: Unable to target HTTPRoute httproute-1, another SecurityPolicy has already attached
           to it
         reason: Conflicted
         status: "False"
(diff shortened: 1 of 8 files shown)


internal/gatewayapi/testdata/securitypolicy-status-conditions.out.yaml
```diff
@@ -286,7 +286,7 @@ securityPolicies:
         namespace: envoy-gateway
       conditions:
       - lastTransitionTime: null
-        message: Unable to target HTTPRoute, another SecurityPolicy has already attached
+        message: Unable to target HTTPRoute httproute-1, another SecurityPolicy has already attached
```
zirain, line 289: should we add namespace here?

sadovnikov: This makes sense. However, the info about the route's namespace is missing. See #3918

sadovnikov:
The trailing spaces in the test files are removed

zirain (approved):

zhaohuabing (approved):
LGTM. Thanks for improving the status message, making it easier fo figure out the overriden policy.

PR #3947, opened 2024-07-24
fix: remove namespace in policies

Xunzhuo:
**What type of PR is this?**

fix(docs): remove namespace in policies


shawnh2 (approved):
Thanks for cleaning this!

cnvergence (approved):

PR #3950, opened 2024-07-24
Test data to demonstrate wrong validation of namespaces

sadovnikov:
**What type of PR is this?**

This PR extends the test files with one `SecurityPolicy` and `HTTPRoute` in a different namespace, to demonstrate #3918

### DIFF
--- a/internal/gatewayapi/testdata/securitypolicy-status-conditions.in.yaml
+++ b/internal/gatewayapi/testdata/securitypolicy-status-conditions.in.yaml
@@ -69,6 +69,28 @@ securityPolicies:
       group: gateway.networking.k8s.io
       kind: HTTPRoute
       name: unknown-httproute
+- apiVersion: gateway.envoyproxy.io/v1alpha1
+  kind: SecurityPolicy
+  metadata:
+    namespace: envoy-gateway
+    name: target-route-in-other-ns
+  spec:
+    targetRef:
+      group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      name: httproute-other-ns
+- apiVersion: gateway.envoyproxy.io/v1alpha1
+  kind: SecurityPolicy
+  metadata:
+    namespace: envoy-gateway
+    name: target-selectors-cross-namespaces
+  spec:
+    targetSelectors:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      matchLabels:
+        target.selector: "true"
+        security.policy.name: target-selectors-cross-namespaces
 httpRoutes:
 - apiVersion: gateway.networking.k8s.io/v1beta1
   kind: HTTPRoute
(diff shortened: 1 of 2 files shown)


zhaohuabing:
@sadovnikov Looks like the output file is incorrect, could you please run `make gen-check` and submit the change?

sadovnikov:
@zhaohuabing, I updated the testdata files. However, the testing setup does not want to accept a new namespace. 
The new `gateway-3` should be accepting routes from all namespaces, but it does not

image

zhaohuabing:
> @zhaohuabing, I updated the testdata files. However, the testing setup does not want to accept a new namespace. The new `gateway-3` should be accepting routes from all namespaces, but it does not

Hi @sadovnikov you can just add the `other-ns` namespace in the .in.yaml

```yaml
namespaces:
- apiVersion: v1
  kind: Namespace
  metadata:
    name: other-ns
```

sadovnikov:
Hi @zhaohuabing, thanks a lot for the tip about the namespace creation!

Unfortunately, status of neither HTTPRoutes nor SecurityPolicy status reflects the link between the two. However, from the empty `status.ancestors` of the `target-route-in-other-ns` policy, I conclude that this [condition](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L269-L278) never gets matched and the code can be removed? Do you agree with this conclusion and with #3918?

```- apiVersion: gateway.envoyproxy.io/v1alpha1
  kind: SecurityPolicy
  metadata:
    creationTimestamp: null
    name: target-route-in-other-ns
    namespace: envoy-gateway
  spec:
    targetRef:
      kind: HTTPRoute
      name: httproute-other-ns
  status:
    ancestors: null
```

zhaohuabing:
> Hi @zhaohuabing, thanks a lot for the tip about the namespace creation!
> 
> Unfortunately, status of neither `HTTPRoute` nor `SecurityPolicy` reflects the link between the two. However, from the empty `status.ancestors` of the `target-route-in-other-ns` policy, I conclude that this [condition](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L269-L278) never gets matched and the code can be removed? Do you agree with this conclusion and with #3918?
> 
> ```
>   kind: SecurityPolicy
>   metadata:
>     creationTimestamp: null
>     name: target-route-in-other-ns
>     namespace: envoy-gateway
>   spec:
>     targetRef:
>       group: gateway.networking.k8s.io
>       kind: HTTPRoute
>       name: httproute-other-ns
>   status:
>     ancestors: null
> ```

@sadovnikov The ancestor is null because the gateway api translator can't find the referenced `HTTPRoute envoy-gateway/httproute-other-ns` of the `SecurityPolicy`. Unlike `Gateway` and `HTTPRoute`, the `SecurityPolicy` doesn't have an affinity to a `GatewayClass`. Therefor, the translator doesn't know whether the HTTPRoute doesn't exist or simply doesn't belong to the current translator.

sadovnikov:
Hi @zhaohuabing, the point I'm trying to make is that this [assignment](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L251) and the following  [condition check](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/securitypolicy.go#L269-L278) are incorrect.

zhaohuabing:
> Reference in new issue

I got it. This checking is not needed anymore as cross-namespace references are not supported for policies.

sadovnikov:
@zhaohuabing, thank you very much for #3961 👍 
Should this PR be merged if you find the new test cases useful? or closed?

zhaohuabing:
Let's close this PR for now. If cross-ns references need to be supported, we can add tests for that later.

Issue #1295, opened 2023-04-13
Add Gateway API webhook validation functions to egctl translate


arkodg:
*Description*:

Run the upstream Gateway API validation functions such as https://github.com/kubernetes-sigs/gateway-api/blob/a78e09220feec7947eb1a08b658b45f3b4182d7f/apis/v1beta1/validation/gateway.go#L54 on the inputted resources in `egctl translate` https://github.com/envoyproxy/gateway/blob/5019d8b598e54fddeece07bbf6de666ec0875780/internal/cmd/egctl/translate.go#L221 to catch client side validation errors and mimic the same experience the user would have had if they were applying these resources using `kubectl apply`

Xunzhuo:
Can we have a `hgctl validate` cmd for validating resources ?

arkodg:
> Can we have a `hgctl validate` cmd for validating resources ?

we currently solve above case using `egctl translate --from gateway-api --to gateway-api` but we could also create a `validate` alias

Xunzhuo:
Thanks @liangyuanpeng !

Xunzhuo:
@liangyuanpeng hi, what is the status?

shawnh2:
This would be very helpful if we bring validations into translate, which will also help the validations in file-resources provider. 

For now, some validations are implemented via CEL, maybe what we need to do here is to find a way convert all the CEL rules into validations code.

zirain:
> For now, some validations are implemented via CEL, maybe what we need to do here is to find a way convert all the CEL rules into validations code.

can we run CEL validation locally like what we did in cel validation tests?

shawnh2:
> > For now, some validations are implemented via CEL, maybe what we need to do here is to find a way convert all the CEL rules into validations code.
> 
> can we run CEL validation locally like what we did in cel validation tests?

Have a discussion about this today in the meeting, it seems we cannot run CEL validation tests locally without kube-api-server support, so may need more investigation.

shawnh2:
It seems we can take https://github.com/google/cel-go as a example, and input our CEL validation rules like:

```go
ast, issues := env.Compile(`OUR-CEL-RULE`)
```

and then validate the input object like:

```go
out, details, err := prg.Eval(XXX)
```

So the only thing left unresolved now is how we retrieve our current CEL rules from comments.

arkodg:
oh nice @shawnh2 !
we can look into kube-builder converts CEL into `x-kubernetes-validations` fields 
[...]

PR #3960, opened 2024-07-27
feat: added support to process multiple gatewayclasses in translate.go

### LINKED ISSUE #3207: egctl x translate should be able to process multiple gatewayclasses
*Description*:

For now, the `kubernetesYAMLToResources` function in `egctl x translate` only able to process one GatewayClass, for multiple GC case, the translation will fail.

https://github.com/envoyproxy/gateway/blob/c0068a056f225e136a740823d6d0955cf380374e/internal/cmd/egctl/translate.go#L628

IMO, above function should return something like:

```go
type LoadResources struct {
        Resources
        
        GatewayClasses  []*gwv1.GatewayClass
        EnvoyProxies    []*egv1a1.EnvoyProxy
}
```

- The GC and EP should be stored in list instead of variable
- Above function should be placed in gatewayapi layer (This will be super helpful for file-provider in standalone mode to reuse this logic to load all kinds of resources from file)

Any thoughts on this? cc @envoyproxy/gateway-maintainers 

[optional *Relevant Links*:]

Manoramsharma:
**What type of PR is this?**

feat: Added support to process multiple gatewayclasses in translate.go

This PR is an attempt to add support to egctl x  translate command to process multiple gatewayClasses by modifying the kubernetesYAMLToResources function. This is just a supporting in the effort to add API webhook validation functions to egctl translate to catch client side validation errors and mimic the same experience the user. 

Support this issue: #1295 

Fixes #3207

### DIFF
--- a/internal/cmd/egctl/translate.go
+++ b/internal/cmd/egctl/translate.go
@@ -241,20 +241,20 @@ func translate(w io.Writer, inFile, inType string, outTypes []string, output, re
 		for _, outType := range outTypes {
 			// Translate
 			if outType == gatewayAPIType {
-				result.Resources, err = translateGatewayAPIToGatewayAPI(resources)
+				result.Resources, err = translateGatewayAPIToGatewayAPI(resources.Resources)
 				if err != nil {
 					return err
 				}
 			}
 			if outType == xdsType {
-				res, err := translateGatewayAPIToXds(dnsDomain, resourceType, resources)
+				res, err := translateGatewayAPIToXds(dnsDomain, resourceType, resources.Resources)
 				if err != nil {
 					return err
 				}
 				result.Xds = res
 			}
 			if outType == irType {
-				res, err := translateGatewayAPIToIR(resources)
+				res, err := translateGatewayAPIToIR(resources.Resources)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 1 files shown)


internal/cmd/egctl/translate.go
```diff
@@ -241,20 +241,20 @@ func translate(w io.Writer, inFile, inType string, outTypes []string, output, re
 		for _, outType := range outTypes {
 			// Translate
 			if outType == gatewayAPIType {
-				result.Resources, err = translateGatewayAPIToGatewayAPI(resources)
+				result.Resources, err = translateGatewayAPIToGatewayAPI(resources.Resources)
```
shawnh2, line 244: We cannot only pass the `resources.Resources` here, the `GatewayClass` and `EnvoyProxy` field is empty. These translate functions all need dedicated resources structure, which have only one GatewayClass and other resources that it has relationship with.

internal/cmd/egctl/translate.go
```diff
@@ -632,14 +632,25 @@ func addMissingServices(requiredServices map[string]*corev1.Service, obj interfa
 	}
 }
 
+// LoadResources struct to handle multiple GCs and EPs
+type LoadResources struct {
```
shawnh2, line 636: The structure looks good! But we cannot directly use it, we may need to create N `gatewayapi.Resources` according to how many GCs you have collected, and pass each generated `gatewayapi.Resources` to each translate function.

arkodg:

Issue #3690, opened 2024-06-27
Support DNS resolution settings


guydc:
*Description*:
Currently, Envoy Gateway overrides the envoy default values for certain DNS resolution:
https://github.com/envoyproxy/gateway/blob/8abf1efef4176092fe4db5ad9239e6d433ca57f2/internal/xds/translator/cluster.go#L145-L146

These settings are typically configurable in similar projects (see links below):
- Istio: `dns_refresh_rate`
- Ambassador: `respect_dns_ttl `
- Gloo Edge: `dns_refresh_rate`, `respect_dns_ttl`

Envoy Gateway can support these settings in `BackendTrafficPolicy`. 

*Relevant Links*:
- Ambassador: https://www.getambassador.io/docs/edge-stack/latest/topics/using/mappings#dns-configuration-for-mappings
- Gloo Edge: https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/upstream.proto.sk/
- Istio: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/

arkodg:
should these live in `Backend` instead ?

guydc:
Minutes from today's community meeting:
- Where: tweaking these settings makes sense mostly for FQDN endpoints. So, implementing this in the `Backend` API makes sense. 
- What: `dns_refresh_rate`, `respect_dns_ttl` are the knobs that make most sense to expose. Other options like dns type (`LOGICAL`, `STRICT`) should be auto-computed by EG. 
- Defaults: current defaults don't have a significant impact, since EG mostly resolves IPs from slices and does not rely on DNS often. When relying on DNS, lookup is less frequent, to reduce load on DNS servers.

alexwo:
Hi @guydc  / @arkodg ,
Since dns_refresh_rate and respect_dns_ttl are properties that apply to all resolutions within a cluster, and a cluster can actually have multiple different backend destinations, does it still make sense to have these settings for each backend?  (as different or conflicting values won't really reflect)

Perhaps using BackendTrafficPolicy would be a better option, as the settings there can apply directly at the cluster level on a 1:1 basis.

WDYT?

guydc:
Hi @alexwo ! 

Yes, that's a very good point. I tend to agree that it makes more sense to use BTP here. Only envoy endpoint-level config makes sense in the backend API.

guydc:
[...]

PR #3974, opened 2024-07-30
feat: Support DNS resolution settings

alexwo:
### DNS resolution settings
The purpose is to allow control over DNS resolution settings for non static cluster endpoints. 
With this PR, it will be possible to define custom DNS resolution settings through a BTP policy.

Fixes # https://github.com/envoyproxy/gateway/issues/3690


api/v1alpha1/dns_types.go
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type DNSSettings struct {
+	// DNSRefreshRate specifies the rate at which DNS records should be refreshed.
```
arkodg, line 11: can we specify the default here ?

api/v1alpha1/dns_types.go
```diff
@@ -0,0 +1,15 @@
...
+	DNSRefreshRate *metav1.Duration `json:"dnsRefreshRate,omitempty"`
+	// RespectDNSTTL indicates whether the DNS Time-To-Live (TTL) should be respected.
```
arkodg, line 13: can we add an extra line in here
```
If the value is set to true, the DNS refresh rate will be set to resource record's TTL
  ```

Comment (arkodg) [reply api/v1alpha1/dns_types.go:13]: also looks like this field defaults to `true` , lets add that info here

File: internal/xds/translator/translator.go [hunk 3]
```diff
@@ -772,6 +772,10 @@ func processXdsCluster(tCtx *types.ResourceVersionTable, httpRoute *ir.HTTPRoute
 		clusterArgs.backendConnection = bt.BackendConnection
 	}
 
+	if httpRoute.DNS != nil {
```
arkodg, line 775: is something similar needed for grpc, tcp, udp, tls ?

alexwo: Its probably can be still applicable If the route specifies a DNS hostname, it should function for all protocols within the context of the Envoy gateway cluster.

internal/xds/translator/cluster.go
```diff
@@ -144,6 +145,16 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		cluster.ClusterDiscoveryType = &clusterv3.Cluster_Type{Type: clusterv3.Cluster_STRICT_DNS}
 		cluster.DnsRefreshRate = durationpb.New(30 * time.Second)
 		cluster.RespectDnsTtl = true
+		if args.dns != nil {
+			if args.dns.DNSRefreshRate != nil {
+				if args.dns.DNSRefreshRate.Duration > 0 {
+					cluster.DnsRefreshRate = durationpb.New(args.dns.DNSRefreshRate.Duration)
+				}
+				if args.dns.RespectDNSTTL != nil {
```
guydc, line 153: should this live outside the the `args.dns.DNSRefreshRate` nil check, or does envoy require `DNSRefreshRate` to be set for `RespectDNSTTL` to be set?

internal/xds/translator/translator.go
guydc: @arkodg  - do you want to create an issue following DNS setting implementation across routes and non-routes? I think that we can start out with limited support for HTTP routes.

internal/xds/translator/cluster.go
alexwo: we default as per above:
	cluster.DnsRefreshRate = durationpb.New(30 * time.Second)
	cluster.RespectDnsTtl = true

internal/xds/translator/translator.go
arkodg: yeah for xRoutes is fine for now, what I was trying to get at this, the code is not uniform, for e.g. for TCP Route translation we have another code flow https://github.com/envoyproxy/gateway/blob/cbabdc623acb1c60c9e3790e9e9d870f117eec95/internal/xds/translator/translator.go#L576 so we missed adding it there

alexwo: right, thanks for catching it. I have just added that for tcp and udp routes.

arkodg: thanks @alexwo can you help with the tech debt here , is it possible to reuse `processXdsCluster` for other route types, so we dont miss something like this again

alexwo: Sure, I have modified processXdsCluster to be compatible with any type that can return cluster arguments,  currently used for TCP, UDP, and HTTP routes.​⬤

arkodg (approved):
LGTM thanks !

guydc (approved):

PR #3977, opened 2024-07-31
feat: Direct Response

### LINKED ISSUE #2714: Feature Request: Add support for configuring gateway to direct_response on HTTP routes.
*Description*:

It would be nice to support use case where users can configure the gateway to direct respond 200 on a HTTP route without upstream backend.

Demacr:
I decided to add Direct Response due to making endpoint stubs after deleting services. I found that #2714 issue has information that GW API still in development to add direct response and there is a way to workaround by `EnvoyPatchPolicy`. So I've implemented new CRD which can be used as `backendRef` in `HTTPRoute`.

p.s.: this PR is a draft, I'll add documentation later. Will be glad to discuss about conception of this solution.

---
Fixed #2714


api/v1alpha1/virtualbackend_types.go
```diff
@@ -0,0 +1,32 @@
...
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+const (
+	// KindSecurityPolicy is the name of the SecurityPolicy kind.
+	KindVirtualBackend = "VirtualBackend"
+)
+
+type Body string
```
liorokman, line 15: Maybe the body should be a `[]byte` instead of a string? Just like in `Secret` resources. That would allow having arbitrarily complex direct responses, including images, and would also clearly separate the Envoy Gateway CRD from the content of the direct response.

api/v1alpha1/virtualbackend_types.go
```diff
@@ -0,0 +1,32 @@
...
+type ResponseHeader string
+
+// VirtualBackend defines the configuration for direct response.
+type VirtualBackend struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              VirtualBackendSpec `json:"spec"`
+}
+
+type VirtualBackendSpec struct {
```
liorokman, line 28: Add some CEL validations here, and also document what the defaults are if nothing is provided.

For consistency with other EG APIs, if the value is optional:
* It should be a pointer 
* The JSON tag should specify "omitempty"
* There should be some documentation around what is used if the value was not specified.

This is especially important for the `StatusCode` field - there might not be a body but the direct response should always include a `StatusCode` - it shouldn't be possible to leave this empty.

liorokman (changes requested):

Xunzhuo (commented):
Can this be part of the Backend API ?

Xunzhuo:
Should we add a new CRD to support one feature ? In a long run, will it cause the API maintaince issues ?

api/v1alpha1/virtualbackend_types.go
```diff
@@ -0,0 +1,32 @@
...
+	Body                 Body             `json:"body,omitempty" yaml:"body"`
+	StatusCode           StatusCode       `json:"statusCode,omitempty" yaml:"statusCode"`
+	ResponseHeadersToAdd []ResponseHeader `json:"responseHeadersToAdd,omitempty" yaml:"responseHeadersToAdd,omitempty"`
```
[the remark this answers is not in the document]
Demacr: >Maybe call the field ResponseHeaders?

Agree

> Instead of a string array, maybe make this a map[string]string?

Hm, maybe make similar to `ir.AddHeader`?
```
type AddHeader struct {
	Name   string `json:"name" yaml:"name"`
	Value  string `json:"value" yaml:"value"`
	Append bool   `json:"append" yaml:"append"`
}
```

> Yeah, I'll add validation when PR will be approved conceptually.

Demacr:
> Can this be part of the Backend API ?

@Xunzhuo Firstly, I thought about it but it will require so many changes it won't be look good by architect.

> Should we add a new CRD to support one feature ? In a long run, will it cause the API maintaince issues ?

@Xunzhuo due to direct response gonna be [part of Gateway API](https://github.com/kubernetes-sigs/gateway-api/issues/2826) it's temporary solution for such functionality. But realization in GW API is unpredictable of time, like one guy took responsibility to add and since April no updates.
After implementation in GW API's side it could be deleted or kept for backward compatibility.

arkodg:
we'll have to unfortunately have to wait until the API is available in Gateway API
i suggest using https://gateway.envoyproxy.io/docs/tasks/extensibility/envoy-patch-policy/ as a workaround until then

zirain:
> we'll have to unfortunately have to wait until the API is available in Gateway API i suggest using https://gateway.envoyproxy.io/docs/tasks/extensibility/envoy-patch-policy/ as a workaround until then

is there an issue on upstream?

Demacr:
@zirain do you mean [this one](https://github.com/kubernetes-sigs/gateway-api/issues/2826)?

api/v1alpha1/virtualbackend_types.go
Demacr: @liorokman done.

api/v1alpha1/virtualbackend_types.go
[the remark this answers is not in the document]
Demacr: @liorokman done

internal/xds/translator/route.go
```diff
@@ -421,7 +421,17 @@ func buildXdsURLRewriteAction(destName string, urlRewrite *ir.URLRewrite, pathMa
 }
 
 func buildXdsDirectResponseAction(res *ir.DirectResponse) *routev3.DirectResponseAction {
-	routeAction := &routev3.DirectResponseAction{Status: res.StatusCode}
+	var directResponseDataSource *corev3.DataSource
+
+	if res.Body != "" {
+		directResponseDataSource = &corev3.DataSource{
+			Specifier: &corev3.DataSource_InlineString{
```
liorokman, line 428: Use `DataSource_InlineBytes` instead of `DataSource_InlineString`.

liorokman (changes requested):

api/v1alpha1/virtualbackend_types.go
```diff
@@ -0,0 +1,52 @@
...
+type StatusCode uint32
+
+// +kubebuilder:validation:Pattern=`^([\w-]+)$`
+type ResponseHeader string
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=envoy-gateway,shortName=vb
+//
+// VirtualBackend defines the configuration for direct response.
+type VirtualBackend struct {
```
zirain, line 26: @guydc do you thinks it's a good idea to support this in `Backend`?

arkodg:
ive raised https://github.com/envoyproxy/gateway/issues/4120 to discuss how we can move forward for Direct Response, hoping to get some consensus on the issue, before we can move forward with the implementation

api/v1alpha1/virtualbackend_types.go
guydc: Sorry for the very late response. IMHO, the best direction here is to support this through a custom filter, as outlined here: https://github.com/envoyproxy/gateway/issues/4120

zirain:
@arkodg should we close this, and move to HTTPRouterFilter?

arkodg:
> @arkodg should we close this, and move to HTTPRouterFilter?

yeah @zirain that makes sense, closing this PR, @Demacr if you're still interesting on adding support for Direct Response, can you raise another PR to add the feature as a field to [HTTPRouterFilter](https://github.com/envoyproxy/gateway/blob/051d5a43a49daf1bf759b756048abcaa176a4dcc/api/v1alpha1/httproutefilter_types.go#L33)

PR #3996, opened 2024-08-03
docs: Clarify sha-1 usage for htpasswd

pingiun:
**What type of PR is this?**

Docs change to clarify wording

Clarify that sha-1 is used for htpasswd checks

N/A


zirain:
this's generated from `api/`, please change the source code then run `make -k gen-check` to update the doc.

pingiun:
Not very useful to have an "edit this page" button on the page then

zirain:
> Not very useful to have an "edit this page" button on the page then

Thanks for head up, will send a PR to fix it.

PR #4000, opened 2024-08-05
feat: gateway http listener isolation

### LINKED ISSUE #3352: Enable GatewayHTTPListenerIsolation conformance test
*Description*:
>Describe the issue.

Details of the test are in https://github.com/kubernetes-sigs/gateway-api/pull/3047

Remove from https://github.com/envoyproxy/gateway/blob/0457ee7417288159eaa4422ee243b808a1696604/test/conformance/experimental_conformance_test.go#L33C3-L33C47 once fixed 

[optional *Relevant Links*:]

levikobi:
**What type of PR is this?**

feat(gateway-api): add support for gateway HTTP listener isolation.

* Adds support for gateway HTTP listener isolation.
* Enables GatewayHTTPListenerIsolation conformance test

Fixes #3352

### DIFF
--- a/internal/gatewayapi/helpers.go
+++ b/internal/gatewayapi/helpers.go
@@ -262,12 +262,12 @@ func servicePortToContainerPort(servicePort int32, envoyProxy *egv1a1.EnvoyProxy
 	return servicePort
 }
 
-// computeHosts returns a list of the intersecting hostnames between the route
-// and the listener.
-func computeHosts(routeHostnames []string, listenerHostname *gwapiv1.Hostname) []string {
+// computeHosts returns a list of intersecting listener hostnames and route hostnames
+// that don't intersect with other listener hostnames.
+func computeHosts(routeHostnames []string, listenerContext *ListenerContext) []string {
 	var listenerHostnameVal string
-	if listenerHostname != nil {
-		listenerHostnameVal = string(*listenerHostname)
+	if listenerContext != nil && listenerContext.Hostname != nil {
+		listenerHostnameVal = string(*listenerContext.Hostname)
 	}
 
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/helpers.go
```diff
@@ -314,6 +314,37 @@ func computeHosts(routeHostnames []string, listenerHostname *gwapiv1.Hostname) [
 	return hostnames
 }
 
+// removeIsolatedListeners returns a list of the hostnames between the route
```
arkodg, line 317: curious why this is needed ? is the issue that `computeHosts(GetHostnames)` is returning duplicates ?
can't we prune that o/p instead ?

levikobi: `computeHosts` returns a list of hostnames based on the intersection between the HTTPRoute's hostnames and a single parentRef's listener each time. 
However, to make this intersection correct with hostname isolation, we need to take into account the entire list of parentRef's listeners.

We can make the change inside `computeHosts` by passing the entire listener each time (instead of passing only `listener.Hostname`. I just thought it would be clearer this way.

@arkodg Please let me know if you want the change to be in `computeHosts` and I'll update the PR.

arkodg: ah thanks for clarifying the current limitation
would prefer if we redefined `computeHosts` to return  - `intersecting listener hostnames and route hostnames that dont intersect with other listener hostnames`

levikobi: 100%
I'll make the change

arkodg (approved):
LGTM thanks !

guydc (approved):
LGTM!

PR #4008, opened 2024-08-06
fix: multiple reference grants in same namespace

### LINKED ISSUE #2149: Multiple Reference Grant for specific service in the same namespace didn't work as expected.
*Description*:
If we separate `HTTPRoute` into a central namespace and place all related backends into their respective namespaces, this will necessitate the use of a `ReferenceGrant`. In our scenario, we include a `ReferenceGrant` as part of each application deployment, that refers to its Kubernetes Service. 
Everything seems working fine, but as soon as we add another application deployment in the existing namespace (that is contains an existing application with its `ReferenceGrant`), HTTPRoute starts showing an inconsistent status that shows `Backend ref to service <namespace>/<service_name> not permitted by any ReferenceGrant`.

*Initial Finding*:

We think this is because of inconsistent steps between these line of code:
* For finding the reference of `ReferenceGrant` for the associated service.
https://github.com/envoyproxy/gateway/blob/defed574484e2708933d5f911b8bfc35e7ac8ee2/internal/provider/kubernetes/controller.go#L496-L504

* For validating when referring to `ReferenceGrant` for the HTTPRoute.
https://github.com/envoyproxy/gateway/blob/defed574484e2708933d5f911b8bfc35e7ac8ee2/internal/gatewayapi/validate.go#L678-L682

The filtering mechanism only checks after the From spec, meaning, the first match with its `From` will be used. While on validation, it will check thoroughly on the `To` spec.

*Repro steps*:
Assuming you have deployed 3 services of Echoserver in the `testing` namespace.

Apply these `ReferenceGrant` manifests:

```yaml
kind: ReferenceGrant
metadata:
  name: echoserver-1-rg
  namespace: testing
spec:
  from:
      kind: HTTPRoute
  to:
    - group: ""
      kind: Service
      name: echoserver-1
---
kind: ReferenceGrant
metadata:
  name: echoserver-2-rg
  namespace: testing
spec:
  from:
      kind: HTTPRoute
  to:
    - group: ""
      kind: Service
      name: echoserver-2
---
kind: ReferenceGrant
metadata:
  name: echoserver-3-rg
  namespace: testing
spec:
  from:
      kind: HTTPRoute
  to:
    - group: ""
      kind: Service
      name: echoserver-3
```

Apply these for HTTPRoute manifests,

```yaml
kind: HTTPRoute
metadata:
  name: sample-01
spec:
  hostnames:
  - playground.internal
  parentRefs:
    kind: Gateway
    name: default-gateway
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: echoserver-1
      namespace: testing
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /echo1a
  - backendRefs:
    - group: ""
      kind: Service
      name: echoserver-2
      namespace: testing
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /echo2
  - backendRefs:
    - group: ""
      kind: Service
      name: echoserver-3
      namespace: testing
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /echo3
```

*Environment*:
v0.5.0 Envoy Gateway, but a similar code still exists in the recent version.

ardikabs:
**What type of PR is this?**

The current behavior of Envoy Gateway when working with multiple `ReferenceGrants` in the same namespace is problematic. When searching for a ReferenceGrant for corresponding resources, such as the `HTTPRoute`, it only checks the `ReferenceGrant.spec.from` semantics. However, during translation, the translator validates both the `ReferenceGrant.spec.from` and `ReferenceGrant.spec.to` semantics thoroughly.
As a result, the `findReferenceGrant` method only captures the first `ReferenceGrant` with the correct `From`, meaning other `ReferenceGrants` are not honored. This leads to the corresponding resource that uses two different resource references encountering an error message such as `Backend ref to service <namespace>/<service_name> not permitted by any ReferenceGrant` during translation.

Code for `findReferenceGrant`:
https://github.com/envoyproxy/gateway/blob/349760029d5ea70c003e291e8d1c9d0ddedc3e96/internal/provider/kubernetes/controller.go#L771-L782

Code for translation:
https://github.com/envoyproxy/gateway/blob/349760029d5ea70c003e291e8d1c9d0ddedc3e96/internal/gatewayapi/validate.go#L832-L857

**Proposed Solution**

This PR modifies the `findReferenceGrant` method to also check the To semantics.

Fixes #2149

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -769,13 +769,35 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to 
 	}
 
 	for _, refGrant := range refGrants {
-		if refGrant.Namespace == to.namespace {
-			for _, src := range refGrant.Spec.From {
-				if src.Kind == gwapiv1a2.Kind(from.kind) && string(src.Namespace) == from.namespace {
-					return &refGrant, nil
-				}
+		if refGrant.Namespace != to.namespace {
+			continue
+		}
+
+		var fromAllowed bool
+		for _, refGrantFrom := range refGrant.Spec.From {
+			if string(refGrantFrom.Kind) == from.kind && string(refGrantFrom.Namespace) == from.namespace {
+				fromAllowed = true
+				break
 			}
 		}
+
+		if !fromAllowed {
+			continue
+		}
+
+		var toAllowed bool
+		for _, refGrantTo := range refGrant.Spec.To {
+			if string(refGrantTo.Kind) == to.kind && (refGrantTo.Name == nil || *refGrantTo.Name == "" || string(*refGrantTo.Name) == to.name) {
+				toAllowed = true
+				break
+			}
+		}
+
+		if !toAllowed {
+			continue
+		}
+
+		return &refGrant, nil
 	}
 
 	// No ReferenceGrant found.
--- a/test/e2e/testdata/multi-referencegrants-same-namespace-services.yaml
+++ b/test/e2e/testdata/multi-referencegrants-same-namespace-services.yaml
@@ -1,0 +1,148 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: multireferencegrants-ns
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: app-backend-v1
+  namespace: multireferencegrants-ns
+spec:
+  selector:
+    app: app-backend-v1
+  ports:
+  - protocol: TCP
+    port: 8080
+    targetPort: 3000
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: app-backend-v1
+  namespace: multireferencegrants-ns
+  labels:
+    app: app-backend-v1
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: app-backend-v1
+  template:
+    metadata:
+      labels:
+        app: app-backend-v1
+    spec:
+      containers:
+      - name: app-backend-v1
+        image: gcr.io/k8s-staging-gateway-api/echo-basic:v20231214-v1.0.0-140-gf544a46e
+        env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: SERVICE_NAME
+          value: app-backend-v1
+        resources:
+          requests:
+            cpu: 10m
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: app-backend-v2
+  namespace: multireferencegrants-ns
+spec:
+  selector:
+    app: app-backend-v2
+  ports:
+  - protocol: TCP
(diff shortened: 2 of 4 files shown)


ardikabs:
since there is no existing unit test for the corresponding line, could you please advise me on whether I should add a unit test elsewhere?

internal/provider/kubernetes/controller.go
```diff
@@ -769,13 +769,37 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
-				}
+		if refGrant.Namespace != to.namespace {
+			continue
+		}
+
+		var fromAllowed bool
+		for _, refGrantFrom := range refGrant.Spec.From {
+			if string(refGrantFrom.Kind) == from.kind &&
+				string(refGrantFrom.Namespace) == from.namespace {
+				fromAllowed = true
```
shawnh2, line 780: should `break` here ?

internal/provider/kubernetes/controller.go
```diff
@@ -769,13 +769,37 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
+
+		if !fromAllowed {
+			continue
+		}
+
+		var toAllowed bool
+		for _, refGrantTo := range refGrant.Spec.To {
+			if string(refGrantTo.Kind) == to.kind &&
+				(refGrantTo.Name == nil ||
+					*refGrantTo.Name == "" ||
```
shawnh2, line 792: let's keep this three conditions in one line

internal/provider/kubernetes/controller.go
```diff
@@ -769,13 +769,37 @@ func (r *gatewayAPIReconciler) findReferenceGrant(ctx context.Context, from, to
...
+					string(*refGrantTo.Name) == to.name) {
+				toAllowed = true
```
shawnh2, line 794: ditto, `break` here ?

shawnh2:
> since there is no existing unit test for the corresponding line, could anyone advise me on whether I should add a unit test elsewhere?

An e2e test would be nice!

internal/provider/kubernetes/controller.go
ardikabs: added, thanks @shawnh2

internal/provider/kubernetes/controller.go
ardikabs: changed!

ardikabs:
> An e2e test would be nice!

@shawnh2 do you mean the integration, like [here](https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/kubernetes_test.go)?

ardikabs:
hi @shawnh2, 
I've added an e2e test, but it seems that some of the other tests, like `TestE2E/Wasm_HTTP_Code_Source`, are a bit flaky.

shawnh2:
they sure are flaky, and we are trying to repair them. 😿

shawnh2 (approved):
LGTM! Thanks for fixing this

arkodg (approved):
LGTM !
thanks for adding an e2e

Issue #3367, opened 2024-05-10
Authorization based on JWT claims


zhaohuabing:
*Description*:
Supports JWT claims as `Principal` in the `Authorization` configuration of a `SecurityPolicy`.

Part of https://github.com/envoyproxy/gateway/issues/2250

Blocked by https://github.com/envoyproxy/envoy/issues/34092
[optional *Relevant Links*:]

@zetaab

zetaab:
still valid

zetaab:
fyi @zhaohuabing & @arkodg that envoy blocking PR just got merged.

PR #4009, opened 2024-08-06
API: authorization api for jwt claims

zhaohuabing:
API for https://github.com/envoyproxy/gateway/issues/3367

A SecurityPolicy enforcing authz on a Gateway resource based on jwtClaims (sub=jason or sub=alice) and iss=foo.bar.com:

```yaml
  kind: SecurityPolicy
  metadata:
    namespace: envoy-gateway
    name: policy-for-gateway 
  spec:
    targetRef:
      kind: Gateway
      name: gateway-1
    authorization:
      defaultAction: Deny
      rules:
      - action: Allow
        principal:
          jwt:
            claims:
            - name: sub
              values: [jason, alice]
            - name: iss
              values: [foo.bar.com]
```

A SecurityPolicy enforcing authz on a HTTPRoute resource based on scope contains "read::message"

```yaml
  kind: SecurityPolicy
  metadata:
    namespace: envoy-gateway
    name: policy-for-route 
  spec:
    targetRef:
      kind: HTTPRoute
      name: route-1
    authorization:
      defaultAction: Deny
      rules:
      - action: Allow
        principal:
          jwt:
            sopes: [read::message]
```

Scopes and claims can be used together, they're And-ed if both are specified.
A SecurityPolicy enforcing authz on a HTTPRoute resource based on scope contains "read::message" and a custom claim message_type:

```yaml
  kind: SecurityPolicy
  metadata:
    namespace: envoy-gateway
    name: policy-for-route 
  spec:
    targetRef:
      kind: HTTPRoute
      name: route-1
    authorization:
      defaultAction: Deny
      rules:
      - action: Allow
        principal:
          jwt:
            sopes: [read::message]            
            claims:
            - name: message_type
              values: [foo, bar]
```


api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,55 @@ type Principal struct {
...
 	// You can use the `ClientIPDetection` or the `EnableProxyProtocol` field in
 	// the `ClientTrafficPolicy` to configure how the client IP is detected.
 	// +kubebuilder:validation:MinItems=1
 	ClientCIDRs []CIDR `json:"clientCIDRs"`
 
-	// TODO: Zhaohuabing the MinItems=1 validation can be relaxed to allow empty list
-	// after other principal types are supported. However, at least one principal is required
+	// JWTClaims are the claims in a JWT token.
+	//
+	// If multiple claims are specified, all claims must match for the rule to match.
```
zetaab, line 65: just thinking use case for two different users from two different issuers:

sub: user1 iss: issuer1
sub: user2 iss: issuer2

is there any way to allow these users? I do not think so. Basically jwtClaims should be "OR"d together and now I think it uses and everywhere?

zetaab: I mean

```
      rules:
      - action: Allow
        principal:
          jwtClaims:
          - type: string
             name: sub
             values: [user1]
          - type: string
            name: iss
            values:  [issuer1]
      - action: Allow
        principal:
          jwtClaims:
          - type: string
             name: sub
             values: [user2]
          - type: string
            name: iss
            values:  [issuer2]
```

but I think this does not work? rules should be "OR"d in this case

zetaab: ah actually, I am little bit slow

```
      rules:
      - action: Allow
        principal:
          jwtClaims:
          - type: string
             name: sub
             values: [user1, user2]
          - type: string
            name: iss
            values:  [issuer1, issuer2]
```

it might work like this. However, this now opens possible security issue. It allow also user1 from issuer2 AND user2 from issuer1

zhaohuabing: @zetaab I think this will work because rules are evaluated in order and the first matching rule will be applied.

```
      rules:
      - action: Allow
        principal:
          jwtClaims:
          - type: string
             name: sub
             values: [user1]
          - type: string
            name: iss
            values:  [issuer1]
      - action: Allow
        principal:
          jwtClaims:
          - type: string
             name: sub
             values: [user2]
          - type: string
            name: iss
            values:  [issuer2]
```

zetaab: In that case the spec looks good

zetaab (approved):
Seems good from my perspective

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,55 @@ type Principal struct {
...
+	// For example, if there are two claims: one for the audience and one for the issuer,
+	// the rule will match only if both the audience and the issuer match.
+	// Note: in order to use JWT claims for authorization, you must configure the
+	// JWT authentication in the same `SecurityPolicy`.
+	// +optional
+	// +notImplementedHide
+	JWTClaims []JWTClaim `json:"jwtClaims"`
```
arkodg, line 72: * how do we know which provider this claim belongs to 
https://github.com/envoyproxy/gateway/blob/866a6cea1af29e0bfbe761fc25b9baddc409ef80/api/v1alpha1/jwt_types.go#L26, do we need an optional `providerName` field here ?

arkodg: or are we asking users to use a field within the jwt like `iss` for differentiation ? as highlighted in https://github.com/envoyproxy/gateway/pull/4009/files#r1705517100

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,55 @@ type Principal struct {
...
+	// Type is the type of the claim.
+	// Valid values are "String" and "StringArray".
+	// For example, sub is a string claim, and groups is a string array claim.
+	// +kubebuilder:validation:Enum=String;StringArray
+	// +kubebuilder:default=String
+	// +unionDiscriminator
+	Type JWTClaimType `json:"type"`
+
+	// Name is the name of the claim.
+	Name string `json:"name"`
```
arkodg, line 86: are nested claims supported ?

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,55 @@ type Principal struct {
...
+
+	// Values are the values that the claim must match.
+	// If the claim is a string type, the specified value must match exactly.
+	// If the claim is a string array type, the specified value must match one of the values in the array.
+	// If multiple values are specified, one of the values must match for the rule to match.
+	Values []string `json:"values"`
```
arkodg, line 92: how would this work if values are boolean or integer ?

zhaohuabing: All the well-known claims that can be used for authorization, such as iss, sub, aud, email, name, roles, scope, and groups,  are either string or string array. Are there any other claims that could be used as identities but are of a different type?

We could consider making this API more flexible to also accommodate boolean and integer types, but it might introduce unnecessary complexity, and it's unclear if such an extension would be essential.  

@envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers please weigh in as you may have use cases for the jwt claims as authorization identities.

api/v1alpha1/authorization_types.go
zhaohuabing: Yeah, the name should by a `.` separated string representing the path to the key. For example: "my.nested.claim".

api/v1alpha1/authorization_types.go
zetaab: `email_verified` is boolean claim. That might be the only one that people could use for authorization

arkodg: can we represent claim values as string, and when matching convert a bool or int to string and compare ?

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,59 @@ type Principal struct {
...
+	// the rule will match only if both the audience and the issuer match.
+	// Note: in order to use JWT claims for authorization, you must configure the
+	// JWT authentication in the same `SecurityPolicy`.
+	// +optional
+	// +notImplementedHide
+	JWTClaims []JWTClaim `json:"jwtClaims"`
 }
 
+// JWTClaim specifies a claim in a JWT token.
+type JWTClaim struct {
```
arkodg, line 76: thoughts on this ?
```
- type: Exact
  name: foo
  value: bar
- type: Contains
  name: foo
  value: baz
```

this API allows for exact, contains and can even be extended for regex if needed, and `exact` can be default

arkodg: `type` / `matchType`

zhaohuabing: Is`Contains` matchType for string, string array, or both? If it works for both, it can be confusing.

api/v1alpha1/authorization_types.go
zhaohuabing: > can we represent claim values as string, and when matching convert a bool or int to string and compare ?

Agree. I'll update the API comments to reflect this.

api/v1alpha1/authorization_types.go
zhaohuabing: Thought it again. If we convert bool and int to string, a `ClaimType` here is misleading and a `MatchType` makes more sense.

So my understanding is:

- `Exact` is used to match a string/bool/int claim type.
- `Contains` is used to match a string Array claim type.

`Regex` can be confusing, better not to support it.

AliceProxy: Bit sure that I see the value of `Contains` or regex for the name of the claim. Would be helpful for the values. If forced to choose between `Contains` and the array for the values though, I think they are both valid use cases but do not completely overlap. If I want to allow a specific claim name but with one of x number of known values, then the array specifier would work for that, but `Contains` might not (depending on the values). But for example, if there were one where the value just needed to contain something, that could definitely be useful over the array of exacts since you might not be capable of easily specifying all accepted values with that method.

arkodg: hey @AliceProxy this is the example we are trying to solve with `Contains` , were you thinking of the same case ?

```
- matchType: Contains
  name: foo
  value: baz
```  

```
   "foo": [
      "bar",
      "baz"
   ]
```

another example is the `amr` claim highlighted in https://developer.okta.com/docs/guides/customize-tokens-returned-from-okta/main/#request-a-token-with-the-custom-claim

AliceProxy: @arkodg That makes sense. Ilike I mentioned, I think it is useful for the value. Maybe it's just me, but I see the name directly under matchType and it is unclear to me that it would only apply to the value since they are all on the same indent/object level.

AliceProxy: maybe just changing it to `valueType` would add clarity? I still do think there are cases where I would still want the slice instead of `Contains` if they don't share common substrings, etc. and we don't plan on supporting `Regex` as a type.

AliceProxy: I personally don't have any issues with Regex. I am quite fond of it for certain use-cases.

arkodg: adding some more links to similar work in other libraries/projects
* golang `slices` has a `Contains` API https://pkg.go.dev/golang.org/x/exp/slices#Contains, but since its a method within `Slices`, it more explanatory than matchType
* CEL https://github.com/google/cel-spec/blob/v0.7.0/doc/langdef.md#macros has `has` and `exists`
* the sets package has `Has` https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.Has

zhaohuabing: The term “MatchType” seems a bit vague to me - it implicitly suggests the type of the claim value(string vs string array) without being clear. I would prefer a more explicit name, such as "ValueType" or "ClaimType".

Given that bool and int values are converted to string, maybe "ValueType" is the better naming.

arkodg: another reason why I proposed the alternate API is
* the `values` field is a list, which made it unclear to what how the match was happening
* the `String` and `StringArray` fields tell me the data type of the `claim` but doesnt tell me how I'm matching on them

zhaohuabing: Unlike routing or rate limiting, which may use various matching types, "Exact“ is typically the sole method used for matching the principals(users, roles, groups, etc). Other kinds of matching types such as "Prefix", "Regex", and "Contains" are rarely used and might seem unusual if supported. Thus, 'Exact` aligns users' expectations.

In the context of `StringArray`, the specified value must "exactly" match one of the elements in the claim value, so it also seems clear to me.

missBerg: The scope claim is "special" because it isn't an array, is a space delimited list. 

When you want to check if a scope is present, it is very much a "contains".

Imagine you have a route/resource that requires the scope books:read however the claim scope will include more scopes in the string and they're space delimited.

@zhaohuabing how are we enabling us to check the presence of that a specific scope like books:read is present in the space delimited list in the scopes claim?

I've had to use regex in the past in my life and I really dislike it, especially since you don't know if the scope is at the beginning middle or end of the space delimited list.

zhaohuabing: @missBerg The authorization is intended to be used against ID tokens, where the scopes should already be resolved to fine-grained claims (The second column in the following table).


Scope | Claims
-- | --
email | email, email_verified
address | address
profile | name, family_name, given_name, middle_name, nickname, preferred_username, profile, picture, website, gender, birthdate, zoneinfo, locale, updated_at
phone | phone_number, phone_number_verified
openid | sub, auth_time, acr

missBerg: Do we need a different issue to enable AuthZ for JWT access tokens? 

It's a super valuable and important feature.

Evaluating the authorization level of an access token is the most typical token to use when deciding if a request should be routed onto the underlying target resource.

ID tokens are not proofs of authorization, they're proof to identity. If a system wants to use identity information to extrapolate authorization levels, that isn't OAuth based authorization. It's a custom mode of authorization. Not wrong, it's just not OAuth based authorization using access tokens.

zhaohuabing: @missBerg Thanks for pointing this out! I think claims from both the Access Token and ID token should be able to be used for authorization. We need to figure out a way to express both the `contians` matching and the `exact` match against an element in an array.

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,77 @@ type Principal struct {
...
 	ClientCIDRs []CIDR `json:"clientCIDRs"`
 
-	// TODO: Zhaohuabing the MinItems=1 validation can be relaxed to allow empty list
-	// after other principal types are supported. However, at least one principal is required
+	// JWT authorize the request based on the JWT claims and scopes.
+	// Note: in order to use JWT claims for authorization, you must configure the
+	// JWT authentication in the same `SecurityPolicy`.
+	// +optional
+	// +notImplementedHide
+	JWT *JWTPrincipal `json:"jwt,omitempty"`
```
zhaohuabing, line 68: Any better naming ideas for the field `JWT` and type `JWTPrincipal`?

api/v1alpha1/authorization_types.go
```diff
@@ -47,17 +50,77 @@ type Principal struct {
...
+	// +kubebuilder:default=String
+	// +unionDiscriminator
+	// +optional
+	ValueType *JWTClaimValueType `json:"valueType,omitempty"`
+
+	// Values are the values that the claim must match.
+	// If the claim is a string type, the specified value must match exactly.
+	// If the claim is a string array type, the specified value must match one of the values in the array.
+	// If multiple values are specified, one of the values must match for the rule to match.
+	Values []string `json:"values"`
```
zhaohuabing, line 114: I prefer using string for bool as "true"/"false" and integer if we need these types in the future. Any is too flexible, as it can represent nested structures besides simple types, which we won't use.

api/v1alpha1/authorization_types.go
```diff
@@ -55,9 +61,69 @@ type Principal struct {
 	ClientCIDRs []CIDR `json:"clientCIDRs"`
 
 	// TODO: Zhaohuabing the MinItems=1 validation can be relaxed to allow empty list
-	// after other principal types are supported. However, at least one principal is required
+	// after JWT principal is supported. However, at least one principal type is required
+
+	// JWT authorize the request based on the JWT claims and scopes.
+	// Note: in order to use JWT claims for authorization, you must configure the
+	// JWT authentication in the same `SecurityPolicy`.
```
arkodg, line 68: can this be codified using a CEL validation ?

api/v1alpha1/authorization_types.go
```diff
@@ -55,9 +61,69 @@ type Principal struct {
...
 }
 
+// JWTClaim specifies a claim in a JWT token.
+type JWTClaim struct {
+	// Name is the name of the claim.
+	// If it is a nested claim, use a dot (.) separated string as the name to
+	// represent the full path to the claim.
+	// For example, if the claim is in the "department" field in the "organization" field,
+	// the name should be "organization.department".
+	// +optional
```
arkodg, line 102: name is mandatory here

api/v1alpha1/authorization_types.go
```diff
@@ -55,9 +61,69 @@ type Principal struct {
...
+	// If multiple claims are specified, all claims must match for the rule to match.
+	// For example, if there are two claims: one for the audience and one for the issuer,
+	// the rule will match only if both the audience and the issuer match.
+	// +optional
+	Claims []JWTClaim `json:"claims,omitempty"`
+
+	// Scopes are a special type of claim in a JWT token that represents the permissions of the client.
+	//
+	// EG interprets the value of the scope claim as a space-separated string during matching,
+	// as defined in RFC 6749: https://datatracker.ietf.org/doc/html/rfc6749.
```
arkodg, line 89:
```suggestion
	// as defined in RFC 6749: https://datatracker.ietf.org/doc/html/rfc6749#page-23.
```

api/v1alpha1/authorization_types.go
```diff
@@ -55,9 +61,69 @@ type Principal struct {
...
+	// Name is the name of the claim.
+	// If it is a nested claim, use a dot (.) separated string as the name to
+	// represent the full path to the claim.
+	// For example, if the claim is in the "department" field in the "organization" field,
+	// the name should be "organization.department".
+	// +optional
+	Name string `json:"name"`
+
+	// ValueType is the type of the claim value.
+	// Only string and string array types are supported for now.
```
arkodg, line 106:
```suggestion
	// Only String and StringArray types are supported for now.
```

api/v1alpha1/authorization_types.go
```diff
@@ -55,9 +61,69 @@ type Principal struct {
...
+	//
+	// If multiple claims are specified, all claims must match for the rule to match.
+	// For example, if there are two claims: one for the audience and one for the issuer,
+	// the rule will match only if both the audience and the issuer match.
+	// +optional
+	Claims []JWTClaim `json:"claims,omitempty"`
+
+	// Scopes are a special type of claim in a JWT token that represents the permissions of the client.
+	//
+	// EG interprets the value of the scope claim as a space-separated string during matching,
```
arkodg, line 88:
```suggestion
	// ,
```
```suggestion
	// The value of the scopes field should be a space delimited string that is expected in the scope parameter,
```

api/v1alpha1/authorization_types.go
```diff
@@ -36,28 +36,95 @@ type AuthorizationRule struct {
...
+	Claims []JWTClaim `json:"claims,omitempty"`
+
+	// Scopes are a special type of claim in a JWT token that represents the permissions of the client.
+	//
+	// The value of the scopes field should be a space delimited string that is expected in the scope parameter,
+	// as defined in RFC 6749: https://datatracker.ietf.org/doc/html/rfc6749#page-23.
+	//
+	// If multiple scopes are specified, all scopes must match for the rule to match.
+	// +optional
+	Scopes []string `json:"scopes,omitempty"`
```
denniskniep, line 94: Can´t we treat `scopes` as normal claim and assume/ensure that specific claims are preprocessed by the JWT Filter?
i.e. via `normalize_payload_in_metadata.space_delimited_claims` ([see here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider-normalizepayload))
> Alters the payload representation in the request dynamic metadata to facilitate its use in matching.

Don´t know how the API implementation will look like, but if its done through metadata this might be an option.

api/v1alpha1/authorization_types.go
```diff
@@ -36,28 +36,95 @@ type AuthorizationRule struct {
...
+	// the name should be "organization.department".
+	Name string `json:"name"`
+
+	// ValueType is the type of the claim value.
+	// Only String and StringArray types are supported for now.
+	// +kubebuilder:validation:Enum=String;StringArray
+	// +kubebuilder:default=String
+	// +unionDiscriminator
+	// +optional
+	ValueType *JWTClaimValueType `json:"valueType,omitempty"`
```
denniskniep, line 112: Does it make sense to specify the Matcher instead of the ClaimValueType? I think that would bring more flexibility.

i.e. 
```
type JWTClaim struct {	
	Name string `json:"name"`
	ValueMatcher JWTClaimValueMatcher `json:"valueMatcher "`
}

type JWTClaimValueMatcher struct {
     equalsAny *JWTClaimEqualsValueMatcher  `json:"valueType,omitempty"`     
     notEqualsAny *JWTClaimEqualsValueMatcher  `json:"valueType,omitempty"`
     includesAll *JWTClaimEqualsValueMatcher  `json:"valueType,omitempty"`
     includesAny *JWTClaimEqualsValueMatcher  `json:"valueType,omitempty"`
     notIncludesAny *JWTClaimEqualsValueMatcher  `json:"valueType,omitempty"`
}

type JWTClaimEqualsValueMatcher struct {
     Values []string `json:"values"`
}
```

* equalsAny (assume claim is single string)
* notEqualsAny (assume claim is single  string)
* includesAll (assume claim is string array)
* includesAny (assume claim is string array)
* notIncludesAny (assume claim is string array)
* ...


What do you think?

api/v1alpha1/authorization_types.go
zetaab: scopes are needed when requesting JWT, not after verifying the tokens? And if this really is valid case (for us its not, we do not have scopes in tokens), it could be done using `[]JWTClaim`?

zhaohuabing: @zetaab both scopes and claims are optional, but at least one of them must be specified.

arkodg (approved):
LGTM thanks Huabing !

zirain (dismissed):

zirain:
```
--- FAIL: TestSecurityPolicyTarget (0.13s)
    --- FAIL: TestSecurityPolicyTarget/authorization-missing_principal (0.00s)
        securitypolicy_test.go:1126: Unexpected response while creating SecurityPolicy; got err=
            SecurityPolicy.gateway.envoyproxy.io "sp-1725430984228589885" is invalid: [spec: Invalid value: "object": if authorization.rules.principal.jwt is used, jwt must be defined, spec.authorization.rules[0].principal.jwt: Invalid value: "object": at least one of claims or scopes must be specified]
            ;missing strings within error=["at least one of clientCIDRs or jwt must be specified"]
FAIL
FAIL	github.com/envoyproxy/gateway/test/cel-validation	7.719s
```

api/v1alpha1/authorization_types.go
zhaohuabing: Hi @denniskniep In theory, we could make this API as flexible as the Envoy RBAC API, but I would like to make this API simpler and easier to understand as it's a user-facing API.  

I believe the normal use cases for JWT claim based authorization is that we just compare wether the claims and scopes are equals (equalsAny). I can't think of any use cases for a substring matching.

For the `not` use case, it can be accomplished with a `Allow` default action and a `Deny` rule:

```
      rules:
      - default: Allow
      - action: Deny
        principal:
          jwt:
            claims:
            - name: user
              values: [jason, alice]
```

zirain (approved):

PR #4010, opened 2024-08-06
chore: remove BackendTrafficPolicyConnection

zhaohuabing:
It seems like `BackendTrafficPolicyConnection` is a redundant struct.


zirain (approved):

guydc:
cc @liorokman I think that this is already covered by `BackendConnection` struct. Do we need this struct ?

guydc (approved):

liorokman:
Nothing is using this structure, it can be safely removed.

arkodg (approved):

Issue #3981, opened 2024-08-01
how to upgrade from 1.0 to 1.1 ?


paulo-worry:
*Description*:
[I Install v1.0 with YAML](https://gateway.envoyproxy.io/v1.0/install/install-yaml/)

Currently, I see that version 1.1 is released. I want to upgrade to 1.1, but I can't find any upgrade documents to refer to?

arkodg:
there are steps mentioned under the `v1.1. Upgrade Notes` in https://gateway.envoyproxy.io/docs/tasks/quickstart/ which should probably be moved to the Install Section

octonawish-akcodes:
@arkodg I want to work on this, can you assign it to me?

octonawish-akcodes:
@zirain @arkodg I raised a PR to fix the issue #3988

paulo-worry:
kubectl version
Client Version: v1.28.7
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.9-eks-036c24b

I got the following error during the upgrade:
`The CustomResourceDefinition "envoyproxies.gateway.envoyproxy.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
`

`kubectl apply -f ./install.yaml 
customresourcedefinition.apiextensions.k8s.io/backendlbpolicies.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/backendtlspolicies.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/gatewayclasses.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/gateways.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/grpcroutes.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/httproutes.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/referencegrants.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/tcproutes.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/tlsroutes.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/udproutes.gateway.networking.k8s.io configured
customresourcedefinition.apiextensions.k8s.io/backends.gateway.envoyproxy.io created
[...]

PR #3988, opened 2024-08-01
doc: fix re-arrangement of docs

octonawish-akcodes:
Re-arranged the docs to find out how to upgrade from `v1.0` to `v1.1`
Fixes #3981

PR #4013, opened 2024-08-07
move upgrade notes to install section

arkodg:
* also add the `--server-side` and `--force-conflicts` flag
* also updated min supported k8s version to v1.27

Fixes: https://github.com/envoyproxy/gateway/issues/3981

Takes https://github.com/envoyproxy/gateway/pull/3988 forward


zirain:
too much flaky on wasm e2e

Xunzhuo (approved):

Issue #3665, opened 2024-06-24
Support the Backend resource for Ext-Auth


arkodg:

guydc:
+1

Curious about the use case: cluster-external service?

arkodg:
same as ext proc, adding the ability to have a sidecar container that can connect over UDS,  but using the ext auth API for legacy reasons

zirain:
@liorokman is this fixed by https://github.com/envoyproxy/gateway/pull/3954?

PR #4016, opened 2024-08-08
feat: Support the Backend resource for Ext-Auth

zirain:
xref: https://github.com/envoyproxy/gateway/issues/3665

### DIFF
--- a/internal/gatewayapi/securitypolicy.go
+++ b/internal/gatewayapi/securitypolicy.go
@@ -896,9 +896,7 @@ func backendRefAuthority(resources *Resources, backendRef *gwapiv1.BackendObject
 
 	backendNamespace := NamespaceDerefOr(backendRef.Namespace, policy.Namespace)
 	backendKind := KindDerefOr(backendRef.Kind, KindService)
-
-	switch backendKind {
-	case egv1a1.KindBackend:
+	if backendKind == egv1a1.KindBackend {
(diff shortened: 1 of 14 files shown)


internal/gatewayapi/securitypolicy.go
```diff
@@ -892,6 +889,31 @@ func (t *Translator) buildExtAuth(policy *egv1a1.SecurityPolicy, resources *Reso
...
+	if backendRef == nil {
+		return ""
+	}
+
+	backendNamespace := NamespaceDerefOr(backendRef.Namespace, policy.Namespace)
+	backendKind := KindDerefOr(backendRef.Kind, KindService)
+	if backendKind == egv1a1.KindBackend {
+		backend := resources.GetBackend(backendNamespace, string(backendRef.Name))
+		if backend != nil {
+			// TODO: exists multi FQDN endpoints?
```
guydc, line 902: I think that we need to support user-defined authority for cases where there are multiple endpoints. We can leave this to a separate PR.

guydc (approved):
LGTM, Thanks!

arkodg (approved):
thanks !

internal/gatewayapi/securitypolicy.go
zirain: yes, telemetry also need it.

PR #4017, opened 2024-08-08
add-nodeSelector

magorKathy:
**What type of PR is this?**

It's needed so we can configure a NodeSelector for Envoy gateway  pods within values file.
A common use case for nodeSelector configuration is to control the placement of pods on specific nodes within a cluster based on node labels. This allows to ensure that certain workloads run only on nodes that meet specific criteria, such as hardware specifications, geographic location, or other custom attributes.

N/A


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -56,6 +56,7 @@ deployment:
     labels: {}
     topologySpreadConstraints: []
     tolerations: []
+  nodeSelector: {}
```
arkodg, line 59: should live at the same level/identation as `topologySpreadConstraints`

arkodg:
thanks for raising this PR @magorKathy, added a comment

charts/gateway-helm/values.tmpl.yaml
magorKathy: fixed. thanks

zirain:
please fix DCO, and run `make -k gen-check`

zirain:
donot merge main, this one is simple.
Otherwise I need approve CI for first-time contributor again and again, thanks!

zirain (approved):

arkodg (approved):
thanks !

Issue #4002, opened 2024-08-05
Graceful shutdown is not working as expected with default setup.


davem-git:
*Description*:
I'm working on implementing envoy-gateway as a replacement for our nginx controller.  I have some basic tests, a pod that returns a json block when hit an endpoint.  Using K6 as a testing sweet. I set up the following test.

```js
import http from 'k6/http';
import { sleep } from 'k6';

export const options = {
  stages: [
    { duration: '2m', target: 50 }, // ramp-up to 50 users
    { duration: '6m', target: 50 }, // stay at 50 users
    { duration: '2m', target: 0 },  // ramp-down to 0 users
  ],
};

export default function () {
  http.get(<url>/);
  sleep(1);
}
````

When I run this test and start a rollout restart of the envoy pods. I get the following errors

```sh
WARN[0070] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49547-><valid public address>:443: read: connection reset by peer"
WARN[0070] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49570-><valid public address>:443: read: connection reset by peer"
WARN[0071] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49587-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49573-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49601-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49594-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49555-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49595-><valid public address>:443: read: connection reset by peer"
WARN[0080] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49589-><valid public address>:443: read: connection reset by peer"
WARN[0082] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49572-><valid public address>:443: read: connection reset by peer"
WARN[0082] Request Failed                                error="Get \"<url>": read tcp 192.168.1.99:49563-><valid public address>:443: read: connection reset by peer"
[...]

PR #4021, opened 2024-08-09
reduce readinessProbe failureThreshold and periodSeconds

arkodg:
* Reduces time for the endpoint to be removed from the endpointSlice from `30s` (3 * 10) to `5s` (1 * 5)

* Since kube-proxy and CNIs rely on this info and so do external LBs like GKE https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer

* The default drain time is `30s` which was also equal to the max time for readinessProbe to fail

Relates to https://github.com/envoyproxy/gateway/issues/4002


arkodg:
cc @guydc @davidalger

zirain (approved):

cnvergence (approved):

PR #4031, opened 2024-08-12
fix: add header values as described in the documentation

### LINKED ISSUE #4023: HTTPHeaderFilter processing doesn't correctly support multiple header values.
*Description*:

The documentation for [HTTPHeaderFilter](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1#HTTPHeaderFilter) contains this statement:

> Configuration to set or add multiple values for a header must use [RFC 7230](https://rfc-editor.org/rfc/rfc7230.html) header value formatting, separating each value with a comma.

My expectation that adding a header via Envoy Gateway's route filters with a comma-delimited value would add the header in the same way that specifying the header multiple times with curl would do.

*Repro steps*:
1. Deploy the quickstart example
2. Apply the following `HTTPRoute`:
    ```yaml
    kind: HTTPRoute
    metadata:
      name: backend
      namespace: default
    spec:
      hostnames:
      - www.example.com
      parentRefs:
        kind: Gateway
        name: eg
      rules:
      - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
        filters:
        - requestHeaderModifier:
            add:
            - name: add-header
              value: foo,bar
          type: RequestHeaderModifier
        matches:
        - path:
            type: PathPrefix
            value: /
    ```
1. Run the following `curl` command:
    ```bash
    curl http://${ENVOY_GATEWAY_IP}/headers -H 'Host: www.example.com' -H 'Test: one' -H 'Test: two'
    ```
   Compare the header `Add-Header` and the header `Test` in the output:

    ```json
    {
     "path": "/headers",
     "host": "www.example.com",
     "method": "GET",
     "proto": "HTTP/1.1",
     "headers": {
      "Accept": [
       "*/*"
      ],
      "Add-Header": [
       "foo,bar"
      ],
      "Test": [
       "one",
       "two"
      ],
     },
     "namespace": "default",
     "ingress": "",
     "service": "",
     "pod": "backend-96f75bbf-778l6"
    }
    ```

liorokman:
Fixes #4023


guydc (approved):
LGTM

arkodg (approved):
LGTM thanks !

should we backport this into v1.0 and v1.1 ?

Issue #3055, opened 2024-03-29
Support LB priority for endpoints


guydc:
*Description*:
Envoy supports [routing priorities](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority#arch-overview-load-balancing-priority-levels):
- When load balancing traffic, envoy uses healthy endpoints (as determined by configured health checks). 
- Users can assign priorities to endpoints, determining which healthy endpoints are preferred for load balancing. 

This allows users to configure a failover mechanism from primary to secondary endpoints.  

Envoy Gateway currently configures a default priority for all endpoints: 
https://github.com/envoyproxy/gateway/blob/f9409e477e866a012757bb9eb4cbc9ac35c7f3a7/internal/xds/translator/cluster.go#L408

LB Priority can be implemented by extending existing Gateway-API resource: 
- `BackendRefs` may support a priority field, similar to the [existing weight](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.BackendRef) field. 
- The `BackendLBPolicy` [resource](https://gateway-api.sigs.k8s.io/geps/gep-1619/?h=backendobjectrefere#backendlbpolicy-api) may be extended to support additional LB considerations, such as priorty. 

Other options:
- EG can support this as an additional attribute of the extended BackendRef proposed in #3069 
- EG can support a custom BackendRef target, which will include a priority field, such as the proposed Backend API: #3063 
- EG `BackendTrafficPolicy` can be used to assign priorities. For example, the policy will hold a map of BackendRefs to priorities. Or, the policy can attach directly to Service resources.

arkodg:
ideally prefer if this was in upstream 
relates to https://github.com/kubernetes-sigs/gateway-api/discussions/2304 & https://github.com/kubernetes-sigs/gateway-api/issues/1802

arkodg:
if we add into field into non xRoute backendRefs, we may need to pluraize existing instantiations to `backendRefs`
* ext auth
* otel 
cc @guydc @zirain

guydc:
> if we add into field into non xRoute backendRefs, we may need to pluralize existing instantiations to `backendRefs`

Yes, also ext-proc. I don't object, and this is in-line with gateway-api. Do you think that we should consider a different intermediate approach?

arkodg:
[...]

PR #4033, opened 2024-08-12
feat: support LB priority for non xRoute endpoints 

alexwo:
### What does this PR do?
It enables users to configure failover backends.

### How is this achieved?
The proposal involves adding a Failover boolean property to backend refs, what allows to indicate whether the backend is designated as a failover.

This approach is illustrated in this PR for external extProc.

As per issue #3055


api/v1alpha1/wasm_types.go
```diff
@@ -49,9 +49,9 @@ type Wasm struct {
 	// +kubebuilder:default=false
 	FailOpen *bool `json:"failOpen,omitempty"`
 
-	// Priority defines the location of the Wasm extension in the HTTP filter chain.
```
guydc, line 52: I think that this should not change in this PR. This priority is related to location in the filter chain, not backend failover.

api/v1alpha1/shared_types.go
```diff
@@ -473,6 +473,9 @@ type BackendRef struct {
 	// BackendObjectReference references a Kubernetes object that represents the backend.
 	// Only Service kind is supported for now.
 	gwapiv1.BackendObjectReference `json:",inline"`
+	// Failover This indicates whether the backend is designated as a failover.
+	// Multiple failover backends can be configured for a single BackendService.⬤
```
guydc, line 477: nit: the ⬤ character is redundant

guydc (commented):
Can we also add an XDS translation test?

alexwo:
> Can we also add an XDS translation test?

added

alexwo:
> Can we also add an XDS translation test?

added

guydc:
overall LGTM. 

In terms of naming, I think that failover is a common industry term used by Akamai, Cloudflare, GCP, AWS and Azure in this context. Nginx and HAProxy use "backup". 

I think that "backup" can be a bit vague in this context:
- It's associated with data backup/restore operations
- It doesn't express the fact that the "backup" backend will automatically handle traffic in case of primary backend failure. Users may anticipate that they need to "enable" a backup server as a primary in case of failure.

guydc (dismissed):

api/v1alpha1/shared_types.go
```diff
@@ -473,6 +473,9 @@ type BackendRef struct {
 	// BackendObjectReference references a Kubernetes object that represents the backend.
 	// Only Service kind is supported for now.
 	gwapiv1.BackendObjectReference `json:",inline"`
+	// Failover This indicates whether the backend is designated as a failover.
+	// Multiple failover backends can be configured for a single BackendService.
```
zirain, line 477:
```suggestion
	// Multiple failover backends can be configured for a single BackendService.
	// +optional
```

api/v1alpha1/shared_types.go
```diff
@@ -473,6 +473,10 @@ type BackendRef struct {
...
```
arkodg, line 477:
```suggestion
	// Multiple failover backends can be configured.
```

api/v1alpha1/shared_types.go
```diff
@@ -473,6 +473,10 @@ type BackendRef struct {
...
+	// +optional
```
arkodg, line 478: "Its strongly recommended to configure active or passive healthchecks to ensure failover can be detected when the active backends are unhealthy and can be readjusted once the primary backends are healthy again. The overprovisioning factor is set to 1.4 i.e. the failover backends will receive traffic only once the health of the active backends drops below 72%"

internal/gatewayapi/securitypolicy.go
```diff
@@ -851,7 +854,7 @@ func (t *Translator) buildExtAuth(policy *egv1a1.SecurityPolicy, resources *Reso
 	authority = backendRefAuthority(resources, backendRef, policy)
 	pnn := utils.NamespacedName(policy)
 	if ds, err = t.processExtServiceDestination(
-		backendRef,
+		&egv1a1.BackendRef{BackendObjectReference: *backendRef, Failover: priority},
```
arkodg, line 857: can we modify this to `Failover: ptr.Deref(grpc.BackendRefs[0].Failover , false)`

arkodg: oh wait, it looks like this line is typecasting from bool from int32 ?

alexwo: We don't have to type cast (its bool type), priority should be named failover.

api/v1alpha1/shared_types.go
alexwo: updated

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4046, opened 2024-08-14
feat(translator): add support for configuring the GRPC Health Checker

liorokman:
**What this PR does / why we need it**:
This PR adds support GRPC health checks.

**Which issue(s) this PR fixes**:
Fixes #4044

PR #4063, opened 2024-08-16
fix: rename ServiceName to Service in the gRPC active health check structure.

liorokman:
Renames the field `ServiceName` to `Service` in the active health check configuration structure for gRPC health checks.

As requested in the comments in #4046


arkodg (approved):
thanks !

zirain (approved):

PR #4074, opened 2024-08-17
use sets and return stable result

zirain:
xref: https://github.com/envoyproxy/gateway/pull/4000


shawnh2 (approved):

arkodg (approved):
thanks, TIL that the sets implementation is stable

PR #4086, opened 2024-08-20
fix: remove the limitation for only one backendRef from ext-auth and ext-proc

### LINKED ISSUE #4006: Support multiple backendRefs
*Description*:
We currently limit the size of []BackendRef in policies (Security, Extension) and EnvoyProxy to 1, due to CEL validation complexity.  
Naturally, a list is used in the API because we do intend to support more than one BackendRef.

liorokman:
Support multiple Backends elements in ext-auth and remove the CEL limitation in ext-proc.

Fixes #4006


arkodg (approved):
LGTM thanks !

guydc (approved):

PR #4090, opened 2024-08-21
fix: active http healthcheck documents a default for expected status, but doesn't use it

### LINKED ISSUE #4089: Active Health check requires expectedStatuses field to work
## Description:
According to [HTTPActiveHealthChecker](https://gateway.envoyproxy.io/docs/api/extension_types/#httpactivehealthchecker), `expectedStatuses` is not required, but if we create a BTP without this field, policy gets accepted but following error logs can be seen in envoy gateway and active health check also doesn't work.
```shell
2024-08-20T22:35:46.590Z	ERROR	gateway-api	runner/runner.go:186	unable to validate xds ir, skipped sending it	{"runner": "gateway-api", "error": "field HTTPHealthChecker.ExpectedStatuses should be specified\nfield HTTPHealthChecker.ExpectedStatuses should be specified"}
2024-08-20T22:35:46.590Z	ERROR	watchable	message/watchutil.go:56	observed an error	{"runner": "gateway-api", "error": "field HTTPHealthChecker.ExpectedStatuses should be specified\nfield HTTPHealthChecker.ExpectedStatuses should be specified"}
```
## Repro steps:
Create a BTP with following config and observe eg logs
```yaml
kind: BackendTrafficPolicy
metadata:
  name: btp
spec:
  targetRef:
    kind: HTTPRoute
    name: backend-routes
  healthCheck:
    active:
      type: HTTP
      http:
        path: /status
```

## Expectation
Either of the below 2:
- Default should be added as it is not being added, or
- Change the `expectedStatuses`  to be a required field and hence policy should not get accepted if not mentioned

liorokman:
If no expected status was explicitly set, use the default value as described in the documentation.

Fixes #4089

### DIFF
--- a/internal/gatewayapi/clustersettings.go
+++ b/internal/gatewayapi/clustersettings.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


arkodg (approved):
LGTM thanks !

cnvergence (approved):

PR #4095, opened 2024-08-22
fix: don't lose timeout settings that originate from the route when translating the backend traffic policy

### LINKED ISSUE #4091: Applied BackendTrafficPolicy breaks HTTPRoute request timeout
*Description*:

When any BackendTrafficPolicy is applied to HTTPRoute the route request timeout setting is ignored. And there is no way to set such timeout in BackendTrafficPolicy. Please suggest a workaround if any.

*Repro steps*:

Route:

```yaml
kind: HTTPRoute
metadata:
  name: some-api
  labels:
    app.kubernetes.io/part-of: some-service
spec:
  parentRefs:
    - name: common
      namespace: gateway-system
  hostnames:
    - "some-api.example.com"
  rules:
    - backendRefs:
        - name: some-service
          namespace: some
          port: 8000
      timeouts:
        request: "130s"
```

Policy:

```
kind: BackendTrafficPolicy
metadata:
  name: conn-limit
spec:
  targetSelectors:
      kind: HTTPRoute
      matchLabels:
        app.kubernetes.io/part-of: some-service
  circuitBreaker:
    maxConnections: 1

```

Request timeout (130s) is ignored and default (15s) is used instead.

*Environment*:

EG version 1.1.0

liorokman:
Fixes #4091

### DIFF
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -384,6 +384,11 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 						continue
 					}
 
+					// Some timeout setting originate from the route.
+					if localTo, err := buildTimeout(policy.Spec.ClusterSettings, r); err == nil {
+						to = localTo
+					}
+
 					r.Traffic = &ir.TrafficFeatures{
 						RateLimit:         rl,
 						LoadBalancer:      lb,
@@ -396,15 +401,11 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 						BackendConnection: bc,
 						HTTP2:             h2,
 						DNS:               ds,
+						Timeout:           to,
 					}
 
 					// Update the Host field in HealthCheck, now that we have access to the Route Hostname.
 					r.Traffic.HealthCheck.SetHTTPHostIfAbsent(r.Hostname)
-
-					// Some timeout setting originate from the route.
-					if to, err = buildTimeout(policy.Spec.ClusterSettings, r); err == nil {
-						r.Traffic.Timeout = to
-					}
 
 					if policy.Spec.UseClientProtocol != nil {
 						r.UseClientProtocol = policy.Spec.UseClientProtocol
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -325,7 +325,7 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 	if policy.Spec.Retry != nil {
 		rt = t.buildRetry(policy)
 	}
-	if to, err = buildTimeout(policy.Spec.ClusterSettings, nil); err != nil {
+	if to, err = buildClusterSettingsTimeout(policy.Spec.ClusterSettings, nil); err != nil {
 		err = perr.WithMessage(err, "Timeout")
 		errs = errors.Join(errs, err)
 	}
@@ -385,7 +385,7 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 					}
 
 					// Some timeout setting originate from the route.
-					if localTo, err := buildTimeout(policy.Spec.ClusterSettings, r); err == nil {
+					if localTo, err := buildClusterSettingsTimeout(policy.Spec.ClusterSettings, r.Traffic); err == nil {
 						to = localTo
 					}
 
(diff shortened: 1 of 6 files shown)


guydc (dismissed):
LGTM, thanks!!

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -395,16 +401,12 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(policy *egv1a1.Backen
 						BackendConnection: bc,
 						HTTP2:             h2,
 						DNS:               ds,
+						Timeout:           to,
 					}
 
 					// Update the Host field in HealthCheck, now that we have access to the Route Hostname.
 					r.Traffic.HealthCheck.SetHTTPHostIfAbsent(r.Hostname)
 
-					// Some timeout setting originate from the route.
```
arkodg, line 403: shouldnt this have worked as well ?

internal/gatewayapi/clustersettings.go
```diff
@@ -83,7 +83,13 @@ func translateTrafficFeatures(policy *egv1a1.ClusterSettings) (*ir.TrafficFeatur
...
 func buildTimeout(policy egv1a1.ClusterSettings, r *ir.HTTPRoute) (*ir.Timeout, error) {
 	if policy.Timeout == nil {
-		return nil, nil
+		var ret *ir.Timeout
+		// Some timeout parameters originate from the route. Don't lose them if
+		// there is no timeout section in the policy.
+		if r != nil && r.Traffic != nil {
+			ret = r.Traffic.Timeout.DeepCopy()
+		}
+		return ret, nil
```
arkodg, line 92: why are we returning early here, if a route is non empty shouldnt we also compute route timeout set then and then return ?

internal/gatewayapi/backendtrafficpolicy.go
liorokman: The problem was an ordering problem. The timeout value was lost in the original ordering because `r.Traffic` (which contained the value collected from the `HTTPRoute` was overridden in line 392.

internal/gatewayapi/clustersettings.go
liorokman: Returning early here because the _policy_ doesn't contain any new timeout information. There's nothing to compute. The only thing that should happen is to not ignore whatever values are already in `ir.Timeout` due to how it was configured from the `HTTPRoute` resource.

arkodg: I think I understand the code flow now, we have two methods
1. [processTimeout ](https://github.com/envoyproxy/gateway/blob/ada279e082485846709ae1250dd2c6b466325626/internal/gatewayapi/route.go#L257)- which builds out the `timeout` struct by processing the HTTPRoute timeout section
2. buildTimeout - which builds out the `timeout` struct by processing BTP

to improve readability, my suggestion would be to rename the functions to `buildRouteTimeout` and `buildBTPTimeout` and add a 3rd func called `mergeTimeouts` which copies over the route timeout fields if set, and gets called in the route section

liorokman: `buildBTPTimeout` is a little misleading. That function is used wherever there are `ClusterSettings` involved - also for things like ext-proc, ext-auth, tracing, and logging cluster definitions.

I'll make a few changes to improve readability though.

internal/gatewayapi/route.go
```diff
@@ -254,15 +254,13 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
...
 	if rule.Timeouts != nil {
-		var rto *ir.Timeout
+		rto := &ir.Timeout{}
 
 		// Timeout is translated from multiple resources and may already be partially set
-		if irRoute.Traffic != nil && irRoute.Traffic.Timeout != nil {
-			rto = irRoute.Traffic.Timeout.DeepCopy()
-		} else {
-			rto = &ir.Timeout{}
+		if irRoute.Traffic != nil {
```
arkodg, line 262: curious when does this case kick in ?
does the route translation always happen before CTP translation, so won't `irRoute.Traffic` always be `nil` ?

liorokman: I'm pretty sure you're right. I think that this is more a case of defensive coding than something real.

arkodg: can we rm, to reduce code complexity ?

arkodg (dismissed):
LGTM thanks for the refactor
going forward, we are probably going to hit such situations very often where a feature will exist in the extension as well as in the core gateway api API, so we should consider following this design pattern there as well to merge the values

guydc (approved):
LGTM!

arkodg:
cherry-picking this one back is non trivial due to the refactoring, so removing the cherry-pick label for this

Issue #3919, opened 2024-07-22
OIDC: allows multiple subdomains share the same OIDC configuration


zhaohuabing:
The EG OIDC (OAuth2 filter under the hood) stores ID and access tokens in cookies without specifying `Domain`. This configuration prevents sharing the same OIDC configuration across subdomains.

The API might be something like this:
```
oidc:
      provider:
        issuer: "https://accounts.google.com/"
      clientID: "client1.apps.googleusercontent.com"
      clientSecret:
        name: "client1-secret"
      redirectURL: "https://www.example.com/bar/oauth2/callback"
      logoutPath: "/bar/logout"
      cookieDomain: ".example.com"
```

PR #4102, opened 2024-08-26
impl for setting OIDC token cookie domain

zhaohuabing:
This PR introduces the ability to set a domain for OAuth2 filter token cookies. By configuring the domain, these token cookies can be shared across multiple subdomains, enabling users to remain logged in while navigating between mulitple subdomains without the need for repeated logins.

```yaml
kind: SecurityPolicy
metadata:
  name: oidc-example
  namespace: default
spec:
  oidc:
    clientID: 758469837151-2tchi5dn2sh7o5kcc3m974nsfdvjlfpr.apps.googleusercontent.com
    clientSecret:
      group: ""
      kind: Secret
      name: my-app-client-secret
    cookieDomain: example.com
    logoutPath: /myapp/logout
    provider:
      issuer: https://accounts.google.com
    redirectURL: https://www.example.com:8443/myapp/oauth2/callback
  targetRefs:
    kind: Gateway
    name: eg
```

Related issue: #3919
Envoy PRs: https://github.com/envoyproxy/envoy/pull/35491 https://github.com/envoyproxy/envoy/pull/35845


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4115, opened 2024-08-26
docs: removes unimplemented attributes in the extproc example

mathetake:
**What type of PR is this?**
the title says.

This document is misleading by using unimplemented field named "attributes" 
which presumably will be implemented in https://github.com/envoyproxy/gateway/issues/3170

N/A


arkodg (approved):
thanks !

zirain (approved):

PR #4116, opened 2024-08-27
feat: support JSONPatches for proxy bootstrap modifications

### LINKED ISSUE #3691: EnvoyProxy Bootstrap configuration can't merge a list
I would like to change the value of `envoy.restart_features.use_eds_cache_for_ads` to false using this `EnvoyProxy` CR:

```yaml
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  bootstrap:
    type: Merge
    value: |
      layered_runtime:
        layers:
        - name: global_config
          staticLayer:
            envoy.restart_features.use_eds_cache_for_ads: false
```

Got two `global_config` keys in the bootstrap configuration.

```yaml
layeredRuntime:
  layers:
  - name: global_config
    staticLayer:
      envoy.restart_features.use_eds_cache_for_ads: true
      re2.max_program_size.error_level: 4294967295
      re2.max_program_size.warn_level: 1000
  - name: global_config
    staticLayer:
      envoy.restart_features.use_eds_cache_for_ads: false
```

Instead, the expected result should be :

```yaml
layeredRuntime:
  layers:
  - name: global_config
     staticLayer:
      envoy.restart_features.use_eds_cache_for_ads: false
      re2.max_program_size.error_level: 4294967295
      re2.max_program_size.warn_level: 1000
```

liorokman:
Fixes #3691


liorokman:
Note that the patch coverage check is artificially low because the PR moved some code from the `xds/translator` package to a common util package ( `util/jsonpatch`) and code that didn't change, other than some indentation, is now flagged as "untested" as part of this PR.

internal/provider/kubernetes/controller.go
```diff
@@ -1651,6 +1652,9 @@ func (r *gatewayAPIReconciler) processEnvoyProxy(ep *egv1a1.EnvoyProxy, resource
 	if err := validation.ValidateEnvoyProxy(ep); err != nil {
 		return fmt.Errorf("invalid envoyproxy: %w", err)
 	}
+	if err := bootstrap.Validate(ep.Spec.Bootstrap); err != nil {
```
guydc, line 1655: Can we cover this validation with a positive/negative test?

internal/ir/xds.go
```diff
@@ -1784,6 +1815,26 @@ func (o *JSONPatchOperation) IsJSONPathNilOrEmpty() bool {
 	return o.JSONPath == nil || *o.JSONPath == EmptyPath
 }
 
+func (o *JSONPatchOperation) Validate() error {
```
guydc, line 1818: can we add a single IR translation test with all invalid cases?

guydc (commented):
overall LGTM, if we can just add a few more tests to the new validation flow and IR it would be great.

I think that the decoupling the API package from XDS libs is a positive thing. 

cc @zhaohuabing

internal/provider/kubernetes/controller.go
liorokman: That's already included here: https://github.com/envoyproxy/gateway/pull/4116/files/c790f2e378c00d169e271049d17c050fb488dd86#diff-6783c0822a78972a488213dd33bef82297eaac567e74fd5ae639020e18448e3c

internal/ir/xds.go
liorokman: I've added a unit-test and made the `validate` method more aligned with RFC 6902

liorokman: I also tried to add CEL validation to the JSONPatchOperation structure in the API folder. 

Turns out that because the `Value` field there is `apiextensionsv1.JSON`, it's not possible to reference the `value` field in CEL tests. Any CEL expression that tries doesn't compile correctly. So no CEL tests on the `JSONPatchOperation` are possible if they reference the `value` attribute.

zirain (commented):
overall LGTM

api/v1alpha1/envoyproxy_types.go
```diff
@@ -358,19 +358,27 @@ const (
 )
 
 // ProxyBootstrap defines Envoy Bootstrap configuration.
+// +union
```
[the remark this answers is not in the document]
liorokman: Not sure. It appears in other places (e.g. `RateLimitSpec`,  `ActiveHealthCheckPayload`, `ActiveHealthCheck`, and more), so I added it here and kept the same semantics as in the other places.

guydc (approved):
LGTM, thanks!

zirain (approved):

PR #4117, opened 2024-08-27
docs: Clarification for type which calls itself "JSON" but actually cannot be.

mt-inside:
Clicking through the docs from `EnvoyProxy`, looking to customize a gateway Deployment or Service, you land on these docs. The type of `value` is given as k8s type `JSON`, but for what I assume are complicated api-machinery deserialization quirks, cannot acutally be a JSON string, which is how I read it. Adding clarification to that.

Ie
This is fine, and is not "JSON" in the simple meaning
```
value:
  foo: bar
```

Note that the above is first-class, ie NOT
```
value: |
  foo: bar
```

And this is actually _rejected_ 
```
value: '{ "foo": "bar" }'
```

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -456,7 +456,9 @@ const (
 	JSONMerge MergeType = "JSONMerge"
 )
 
-// KubernetesPatchSpec defines how to perform the patch operation
+// KubernetesPatchSpec defines how to perform the patch operation.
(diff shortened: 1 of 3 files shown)


arkodg (approved):
thanks !

zirain:
@mt-inside the doc is generated from `/api`, need to change there and run `make gen-check`.

mt-inside:
@zirain Done!

shawnh2 (approved):
LGTM

zirain (approved):

PR #4121, opened 2024-08-29
rename `failover` to `fallback` backend

arkodg:
`fallback` has been historically used for more to descibe a backend / server, while `failover` has been used to describe the operation of handling the case when the primary has failed


arkodg:
cc @alexwo

zirain (approved):

alexwo (approved):
LGTM

PR #4124, opened 2024-08-29
chore: add ExtensionGroupKinds debug log

zirain:


internal/gatewayapi/runner/runner.go
```diff
@@ -159,6 +159,7 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 						extGKs = append(extGKs, schema.GroupKind{Group: gvk.Group, Kind: gvk.Kind})
 					}
 					t.ExtensionGroupKinds = extGKs
+					r.Logger.Info("extension resources", "GVKs", extGKs)
```
arkodg, line 162: the size of `extGVKs` can be pretty big here, how about recording the length here instead ?

arkodg (approved):

shawnh2 (approved):

PR #4126, opened 2024-08-30
Fix IsNotFound check for secret and configmap

TasdidurRahman:
small fix : 
`unable to find the Secret/Configmap` should be returned when IsNotFound error is true.


shawnh2 (approved):
Thanks for spotting this

zirain (approved):

PR #4128, opened 2024-08-31
API: use BackendCluster to represent OIDCProvider

zhaohuabing:
Using BackendCluster to represent OIDCProvider allows user to customize backend connection settings for OIDC providers, such as retry policy and  tls setting.

Related: https://github.com/envoyproxy/gateway/issues/3178


zirain (approved):

coro:
@zhaohuabing Thanks for this PR! I'm struggling to build an image of the core proxy with your [configuration PR](https://github.com/envoyproxy/envoy/pull/35262) in to test with this PR in our system - would it be possible to produce a new distroless-dev image of Envoy so we can try this?

zhaohuabing:
Hi @coro You can test it after this PR has been merged.

coro:
@zhaohuabing I don't think there's been a dev build for 6 days, so that won't have your PR from 4 days ago unless I'm misunderstanding how it's built: https://hub.docker.com/layers/envoyproxy/envoy/distroless-dev/images/sha256-239f7d78c7377bc668574c3be860397e640d7cff3e0b038ced64b1bb15c2b104?context=explore

zhaohuabing:
> https://hub.docker.com/layers/envoyproxy/envoy/distroless-dev/images/sha256-239f7d78c7377bc668574c3be860397e640d7cff3e0b038ced64b1bb15c2b104?context=explore

Oh, I see. If you want to test now, you could probably build an envoy binary from the source:

```
ENVOY_DOCKER_BUILD_DIR=~/build ./ci/run_envoy_docker.sh './ci/do_ci.sh dev' 
```

and then build an image with a Dockerfile like this:

```
From envoyproxy/envoy:dev
COPY envoy /usr/local/bin/envoy
```

```
cp ~/build/envoy/x64/source/exe/envoy/envoy ./ 
docker build ./ -t docker.io/envoyproxy/envoy:test 
```

zirain:
envoy post jobs was broken, seems got fixed by https://github.com/envoyproxy/envoy/pull/35949, maybe need wait for a while.

phlax:
> I don't think there's been a dev build for 6 days

just read this and realized - our docker image building is broken 8(

ill work on a fix presently

phlax:
> ... seems got fixed by ...

that PR was an unrelated docs fix

internal/xds/translator/oidc.go
```diff
@@ -180,6 +180,13 @@ func oauth2Config(oidc *ir.OIDC) (*oauth2v3.OAuth2, error) {
 			Resources:  oidc.Resources,
 
 			PreserveAuthorizationHeader: preserveAuthorizationHeader,
+
+			RetryPolicy: &corev3.RetryPolicy{
```
arkodg, line 184: should be opt in, imo, to maintain the same API behavior with xRoutes

zhaohuabing: Discussed this with @arkodg this morning, it's better to put RetryPolicy into the `ClusterSettings` and use `BackendRef` to model the Auth server. In this way, the RetryPolicy can also be reused by jwt and other features.

zhaohuabing:
Discussed this with @arkodg this morning, it's better to put RetryPolicy into the ClusterSettings and use BackendRef to model the Auth server. In this way, the RetryPolicy can also be reused by jwt and other features.

phlax:
FYI: Envoy Docker publishing is fixed

api/v1alpha1/oidc_types.go
```diff
@@ -106,7 +106,17 @@ type OIDC struct {
 }
 
 // OIDCProvider defines the OIDC Provider configuration.
+// +kubebuilder:validation:XValidation:message="BackendRefs must be used, backendRef is not supported.",rule="!has(self.backendRef)"
```
zhaohuabing, line 109: Even though `BackendRefs` is used, only one `backendRef` will be supported.

zhaohuabing, line 109: Even though BackendRefs is used, only one backendRef will be supported.

liorokman: Maybe add a CEL validation to make sure that `has(self.backendRefs) && size(self.backendRefs) == 1` ?

api/v1alpha1/oidc_types.go
```diff
@@ -106,7 +106,17 @@ type OIDC struct {
...
 type OIDCProvider struct {
+	// BackendCluster defines the backend cluster to be used for the OIDC Provider.
```
arkodg, line 111: `BackendCluster` is an internal term, imo we shouldnt expose it to the user, `cluster` is a new term the user will need to understand which may confuse them with a k8s cluster

api/v1alpha1/oidc_types.go
zhaohuabing: We may support multiple services in the future. But I agree that we should make it clear that multiple providers is not supported yet.

arkodg (approved):

zirain (approved):

PR #4146, opened 2024-09-03
fix: [controller-runtime] log.SetLogger(...) was never called; logs w…

### LINKED ISSUE #4145: [controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
*Description*:
returned, etc.

*Repro steps*:

sharing.

*Environment*:

*Logs*:

qicz:
…ill not be displayed.

**What type of PR is this?**

Fixes #4145

### DIFF
--- a/internal/provider/kubernetes/kubernetes.go
+++ b/internal/provider/kubernetes/kubernetes.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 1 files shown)


zirain (dismissed):

internal/provider/kubernetes/kubernetes.go
```diff
@@ -44,6 +45,9 @@ func New(cfg *rest.Config, svr *ec.Server, resources *message.ProviderResources)
 		LeaderElectionNamespace: svr.Namespace,
 	}
 
+	// fix [controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
```
arkodg, line 48: can we rm the comment ? imo its not useful

zirain (approved):

shawnh2 (approved):

PR #4157, opened 2024-09-04
ci: fix osv vulnerability and license scans and add license overrides

shahar-h:
* Disable call analysis in both osv vulnerability and license scans until https://github.com/google/osv-scanner/issues/1220 is resolved.
* Add license overrides for packages with unknown licenses and for packages which got [license exception](https://github.com/cncf/foundation/blob/main/license-exceptions/README.md) from CNCF.
* Move osv-scanner config file to default location (`osv-scanner.toml`) in order to make it reusable also by openssf-scorecard.
* There are 2 remaining packages with an unapproved license (both have `MPL-2.0` license):
  * `github.com/hashicorp/go-getter`
  * `github.com/hashicorp/go-safetemp` - imported by  `github.com/hashicorp/go-getter`
  
  See related CNCF issue regarding license exception request for these packages: https://github.com/cncf/foundation/issues/624.
  We can either wait for this issue to be resolved (it's quite old) or consider to remove `github.com/hashicorp/go-getter` usage.
  import path:
  github.com/envoyproxy/gateway/internal/cmd/egctl
  github.com/replicatedhq/troubleshoot/pkg/convert
  github.com/replicatedhq/troubleshoot/pkg/analyze
  github.com/hashicorp/go-getter

### DIFF
--- a/osv-scanner.toml
+++ b/osv-scanner.toml
@@ -1,0 +1,122 @@
+[[IgnoredVulns]]
+id = "GO-2022-0646"
+reason = "No a real issue, just a warning about third party package."
+
+[[PackageOverrides]]
+name = "github.com/AdaLogics/go-fuzz-headers"
+version = "0.0.0-20230811130428-ced1acdcaa24"
+ecosystem = "Go"
+license.override = ["Apache-2.0"]
+reason = "Unknown license since package version is missing in pkg.go.dev"
+
+[[PackageOverrides]]
+name = "github.com/asaskevich/govalidator"
+version = "0.0.0-20230301143203-a9d515a09cc2"
+ecosystem = "Go"
+license.override = ["MIT"]
+reason = "Unknown license, remove once https://github.com/google/deps.dev/issues/87 is resolved"
+
+[[PackageOverrides]]
+name = "github.com/containers/storage"
+version = "1.55.0"
+ecosystem = "Go"
+license.override = ["Apache-2.0"]
+reason = "Unknown license, remove once https://github.com/google/deps.dev/issues/104 is resolved"
+
+[[PackageOverrides]]
+name = "github.com/distribution/distribution/v3"
+version = "3.0.0-beta.1"
+ecosystem = "Go"
+license.override = ["Apache-2.0"]
+reason = "Unknown license, remove once https://github.com/google/deps.dev/issues/105 is resolved"
+
+[[PackageOverrides]]
+name = "github.com/docker/go-metrics"
+version = "0.0.1"
+ecosystem = "Go"
+license.override = ["Apache-2.0"]
+reason = "This package has dual license - the code is licensed under the Apache 2.0 license and the docs under CC-BY-SA-4.0 license"
+
+[[PackageOverrides]]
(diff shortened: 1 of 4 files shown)


guydc (approved):
LGTM, Thanks!

guydc:
> We can get rid of [github.com/hashicorp/go-getter](https://github.com/hashicorp/go-getter) package by replacing the usage of [convert.ValidateOutputPath](https://github.com/envoyproxy/gateway/blob/301eeddfa4c2f7e4e612874a9c5b13a976156c3e/internal/cmd/egctl/collect.go#L77) function with a local copy. This is the function: https://github.com/replicatedhq/troubleshoot/blob/main/pkg/convert/output.go#L10-L19 

@zirain - WDYT? Do you see other uses for this lib? if not, by replacing it, we can finally start enforcing OSV scans.

zirain:
> > We can get rid of [github.com/hashicorp/go-getter](https://github.com/hashicorp/go-getter) package by replacing the usage of [convert.ValidateOutputPath](https://github.com/envoyproxy/gateway/blob/301eeddfa4c2f7e4e612874a9c5b13a976156c3e/internal/cmd/egctl/collect.go#L77) function with a local copy. This is the function: https://github.com/replicatedhq/troubleshoot/blob/main/pkg/convert/output.go#L10-L19
> 
> @zirain - WDYT? Do you see other uses for this lib? if not, by replacing it, we can finally start enforcing OSV scans.

it's fine to replace.

osv-scanner.toml
```diff
@@ -0,0 +1,122 @@
...
+license.override = ["BSD-3-Clause"]
+reason = "Unidentified license, remove once https://github.com/google/deps.dev/issues/86 is resolved"
+
+[[PackageOverrides]]
+name = "github.com/grafana/tempo"
+version = "1.5.0"
+ecosystem = "Go"
+# Override the license to an allowed one until https://github.com/google/osv-scanner/issues/1124 is resolved and we can skip it from licnese scanning instead
+license.override = ["Apache-2.0"]
+reason = "This package is only used in e2e tests so we can ignore its license"
```
zirain, line 122: we change change it to zipkin or other backend.
does loki has same problem?

shahar-h: Loki has the same `AGPL-3.0` license.

shahar-h:
> > > We can get rid of [github.com/hashicorp/go-getter](https://github.com/hashicorp/go-getter) package by replacing the usage of [convert.ValidateOutputPath](https://github.com/envoyproxy/gateway/blob/301eeddfa4c2f7e4e612874a9c5b13a976156c3e/internal/cmd/egctl/collect.go#L77) function with a local copy. This is the function: https://github.com/replicatedhq/troubleshoot/blob/main/pkg/convert/output.go#L10-L19
> > 
> > 
> > @zirain - WDYT? Do you see other uses for this lib? if not, by replacing it, we can finally start enforcing OSV scans.
> 
> it's fine to replace.

I'll open a separate PR for this.

zirain (approved):

PR #4158, opened 2024-09-04
feat(helm): Proper namespace handling and chart name

s0uky:
**What type of PR is this?**

This PR changes handling of namespace if helm template is used and also changes name of the chart to more logical.

Is more flexible for end users. Especially for the users who use Umbrella Helm chart in their envs.

### DIFF
--- a/charts/gateway-helm/Chart.yaml
+++ b/charts/gateway-helm/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: gateway-helm
+name: envoy-gateway
 description: The Helm chart for Envoy Gateway
 type: application
 
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "eg.fullname" . }}-certgen
-  namespace: '{{ .Release.Namespace }}'
+  namespace:  '{{ .Values.namespace | default .Release.Namespace }}'
   labels:
   {{- include "eg.labels" . | nindent 4 }}
   {{- if .Values.certgen.rbac.labels }}
@@ -18,7 +18,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ include "eg.fullname" . }}-certgen
-  namespace: '{{ .Release.Namespace }}'
+  namespace: '{{ .Values.namespace | default .Release.Namespace }}'
   labels:
   {{- include "eg.labels" . | nindent 4 }}
   {{- if .Values.certgen.rbac.labels }}
@@ -43,7 +43,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: {{ include "eg.fullname" . }}-certgen
-  namespace: '{{ .Release.Namespace }}'
+  namespace: '{{ .Values.namespace | default .Release.Namespace }}'
   labels:
   {{- include "eg.labels" . | nindent 4 }}
   {{- if .Values.certgen.rbac.labels }}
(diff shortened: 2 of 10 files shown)


charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "eg.fullname" . }}-certgen
-  namespace: '{{ .Release.Namespace }}'
+  namespace:  '{{ .Values.namespace | default .Release.Namespace }}'
```
zirain, line 5: is this recommand way for helm by introuducing a namespace field in `values.yaml`?

charts/gateway-helm/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: gateway-helm
+name: envoy-gateway
```
zirain, line 2: we cannot change this as we store chart in OIC registry, this's the image name.

charts/gateway-helm/templates/certgen-rbac.yaml
s0uky: Yes, in values.yaml will be only "namespace: name-of-ns" and it's easy to configure in whole chart. 
Because if you use Umbrella templating .Release.Namespace is always default and it's not possible to change it.

charts/gateway-helm/Chart.yaml
s0uky: Hmm, OK, understand. But it's misleading.

charts/gateway-helm/templates/certgen-rbac.yaml
zirain: `helm template -n ` didn't work?

s0uky: `helm template -n` you can't use if using Umbrella chart, you need to define namespace for every child Helm chart.

zirain: I didn't think it's a good practice to add things like this.

s0uky: Why? 
For example `ingress-nginx` has option `namespaceOverride` (this is more complex, than my PR):

https://github.com/kubernetes/ingress-nginx/pull/10539/files

arkodg: curious why a  `namespaceOverride` field is needed when it can be achieved in helm natively ?
```
helm install eg-marketing oci://docker.io/envoyproxy/gateway-helm --version v1.1.0 -n marketing --create-namespace
```

s0uky: because if you managing plenty of helm charts over [umbrella approach](https://helm.sh/docs/howto/charts_tips_and_tricks/#complex-charts-with-many-dependencies), you have to define namespace for every subchart.

zirain: IMO, we shouldn't support this(the benefits of introduction are far smaller than the complications), you can do this in your own fork.

s0uky: Which are the complications you see? Nothing changes in chart logic, only one option is added. 
Many CNCF projects adopted this logic.

s0uky: Another example in huge project which ArgoCD is:

https://github.com/argoproj/argo-helm/blob/fd70c609bd6e98e8b067c7c4915957c51acd786c/charts/argo-cd/templates/_helpers.tpl#L242

https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L11

arkodg:

honarkhah:
Noticed this was somehow also handling this issue https://github.com/envoyproxy/gateway/issues/8280

Issue #4152, opened 2024-09-03
EnvoyExtensionPolicy  support metadata_options in ExtProc


zirain:
this's useful when we're trying to use `metadata` to share info across filters(e.g. `envoy.filters.http.jwt_authn`)

cc @envoyproxy/gateway-maintainers WDYT?

guydc:
+1 

I would actually recommend exposing some additional options:  https://github.com/envoyproxy/gateway/pull/3247

arkodg:
-1 on this
to use metadata_options, the user would also need to add custom modifications to other parts of envoy as well that are not exposed by the EG API and would need EnvoyPatchPolicy or Extension Manager

zirain:
> -1 on this to use metadata_options, the user would also need to add custom modifications to other parts of envoy as well that are not exposed by the EG API and would need EnvoyPatchPolicy or Extension Manager

why need that? it's a common case to use metadata(e.g. `envoy.filters.http.jwt_authn`) in `ExtProc`.

arkodg:
> > -1 on this to use metadata_options, the user would also need to add custom modifications to other parts of envoy as well that are not exposed by the EG API and would need EnvoyPatchPolicy or Extension Manager
> 
> why need that? it's a common case to use metadata(e.g. `envoy.filters.http.jwt_authn`) in `ExtProc`.

How would the user configure the metadata in the JWT filter ? Wouldn't they need some extension mechanism to do it ?

zirain:
> > > -1 on this to use metadata_options, the user would also need to add custom modifications to other parts of envoy as well that are not exposed by the EG API and would need EnvoyPatchPolicy or Extension Manager
> > 
> > 
> > why need that? it's a common case to use metadata(e.g. `envoy.filters.http.jwt_authn`) in `ExtProc`.
> 
> How would the user configure the metadata in the JWT filter ? Wouldn't they need some extension mechanism to do it ?

no, no

zirain:
a demo configuration looks like following:
```
#                  - name: envoy.filters.http.jwt_authn
#                    typedConfig:
#                      "@type": type.googleapis.com/envoy.extensions.filters.http.jwt_authn.v3.JwtAuthentication
#                      providers:
#                        "jwt1":
#                          forward: true
#                          local_jwks:
#                            filename: /etc/envoy/jwks.json
#                          payloadInMetadata: "playload"
#                      requirementMap:
#                        "route":
#                          providerName: jwt1
[...]

PR #4160, opened 2024-09-05
api: add `Untyped` metadata support in ExtProc

zirain:
xref: #4152


api/v1alpha1/ext_proc_types.go
```diff
@@ -70,4 +70,16 @@ type ExtProc struct {
...
+
+	// MetadataOptions defines the metadata that should be forwarded to or received from the external processor.
+	//
+	// +optional
+	Metadata *MetadataOptions `json:"metadata,omitempty"`
+}
+
+type MetadataOptions struct {
+	// ForwardingNamespaces is a list of namespaces from which metadata should be forwarded to the external processor.
+	ForwardingNamespaces []string `json:"forwardingNamespaces,omitempty"`
```
guydc, line 82: Will this be translated to typed or untyped namespaces? if we want to support both options in the future, we maybe want to make the distinction here already, like here: https://github.com/envoyproxy/gateway/pull/3247/files#diff-2d475e566eff6abe27d7ed2892b3dd252980f2cfccc04bb859a14878c8ca71b1R63

zirain: can you resubmit that one instead of this?

guydc: it's stale, let's just go ahead with your PR.

guydc (approved):
LGTM

Xunzhuo (approved):
/lgtm

PR #4161, opened 2024-09-05
remove MPL packages

zirain:
follow up: https://github.com/envoyproxy/gateway/pull/4157

cc @shahar-h


.github/workflows/osv-scanner.yml
```diff
@@ -26,8 +26,6 @@ jobs:
       scan-args: |-
         --skip-git
         --recursive
-        --config
```
shahar-h, line 29: Thanks, I missed this in my PR. can you also add `--no-call-analysis=go` along with the TODO comment in line 39?

osv-scanner.toml
```diff
@@ -112,11 +112,3 @@ name = "stdlib"
 ecosystem = "Go"
 license.override = ["BSD-3-Clause"]
 reason = "Unidentified license, remove once https://github.com/google/deps.dev/issues/86 is resolved"
-
```
shahar-h, line 115: You can also remove:
* github.com/hashicorp/go-cleanhttp
* github.com/hashicorp/go-version

zirain: I want to keep it because it's in CNCF allowlist, it maybe included by accident later, WDYT?

shahar-h: You can take it from git history if needed but it's your call.

shahar-h (approved):
LGTM

guydc (dismissed):

test/e2e/tempopb/tempopb.go
```diff
@@ -0,0 +1,31 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package tempopb
+
+// this is a package copied from tempo project,
+// EG only use pb struct for testing purpose.
```
arkodg, line 9: to avoid any ambiguity wrt referencing , i'd suggest unmarshalling to `map[string]interface{}` so we rm this

internal/cmd/egctl/collect.go
```diff
@@ -74,7 +73,7 @@ func runCollect(collectOpts collectOptions) error {
 	basename := ""
 	if collectOpts.outPath != "" {
 		// use override output path
-		overridePath, err := convert.ValidateOutputPath(collectOpts.outPath)
+		overridePath, err := tb.ValidateOutputPath(collectOpts.outPath)
```
arkodg, line 76: i suggest adding an internal function to call `filepath.Abs` + `os.Stat` to validate, this feels like a common util func within nothing proprietary about it, and we can avoid the entire reference altogether

test/e2e/tempopb/tempopb.go
zirain: `map[string]interface{}` is untyped and not firendly to use, another option is switch to use `Zipkin`.

BTW, I don't think pb struct should be under MPL license.

arkodg (approved):
thanks !

arkodg (approved):

zirain:
```
+------------------------------+------+-----------+---------+---------+--------+
| OSV URL                      | CVSS | ECOSYSTEM | PACKAGE | VERSION | SOURCE |
+------------------------------+------+-----------+---------+---------+--------+
| https://osv.dev/GO-2024-3105 |      | Go        | stdlib  | 1.22.5  | go.mod |
| https://osv.dev/GO-2024-[31](https://github.com/envoyproxy/gateway/actions/runs/10747767939/job/29810620873?pr=4161#step:4:32)06 |      | Go        | stdlib  | 1.22.5  | go.mod |
| https://osv.dev/GO-2024-3107 |      | Go        | stdlib  | 1.22.5  | go.mod |
+------------------------------+------+-----------+---------+---------+--------+
```

will fix it in separated one.

Issue #4163, opened 2024-09-05
Add support for dynamic metadata in http ext authz


nothinux:
*Description*:
We have a use case that requires emitting specific information from the external authz service into the logs, and this can be achieved using dynamic metadata. In the grpc ext_authz, dynamic metadata can be included directly as part of `CheckResponse`. However, in the http ext_authz, the only method to pass this information is by sending it as response headers, which are then emitted via the `dynamic_metadata_from_headers`.

It would be great if the [dynamic_metadata_from_headers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-authorizationresponse) field can be exposed

arkodg:
can this be achieved today by `headersToBackend` and then removing the header using `RequstHeaderModifier` filter https://gateway-api.sigs.k8s.io/guides/http-header-modifier/#http-request-header-modifier?

nothinux:
@arkodg  yes, it can be archieved with that approach. However, we won't be able to retrieve the [ext_authz_duration](ext_authz_duration) data, since it's only available when `dynamic_metadata_from_headers` is set. We need that data to monitor the time taken to complete the authorization process for each request

arkodg:
this feels like an advanced use case, I suggest using https://gateway.envoyproxy.io/docs/tasks/extensibility/envoy-patch-policy/ to achieve this
lets keep this issue open to see if others in the community are also interested in this

arkodg:
we are beginning to add first class support for dynamic metadata
https://github.com/envoyproxy/gateway/pull/5023 added the ability to read/write metadata for ext proc, we can follow something similar here for ext authz

arkodg:
cc @nareddyt

PR #4164, opened 2024-09-05
feat: add dynamic metadata support in http ext authz

nothinux:
**What type of PR is this?**

add `headersToMetadata` field to support emitting dynamic metadata from authorization response header

Fixes https://github.com/envoyproxy/gateway/issues/4163


api/v1alpha1/ext_auth_types.go
```diff
@@ -78,4 +78,11 @@ type HTTPExtAuthService struct {
 	// original client request.
 	// +optional
 	HeadersToBackend []string `json:"headersToBackend,omitempty"`
+
+	// HeadersToMetadata are the authorization response headers that will be emitted
+	// as dynamic metadata to be consumed by next filter.
+	// Note: this metadata lives in envoy.filters.http.ext_authz and can only be consumed
+	// by the next filter if explicitly specified.
+	// +optional
+	HeadersToMetadata []string `json:"headersToMetadata,omitempty"`
```
zirain, line 87: looks like only support exact with this PR, what will happen if we want to support prefix/suffix/regex in the future?

BTW, please proposal the API first.

nothinux: yes, this will only support exact matching for now, similar to `HeadersToBackend`.

Apologies, I wasn't aware a proposal was needed. I have now created it on https://github.com/envoyproxy/gateway/issues/4163.

nothinux: we will need to change the API for `HeadersToBackend` also if we want to support prefix/suffix/regex in the future, do you think we need it @zirain? if so I'll try to change the implementation for both

zirain: let's focus on metadata first, WDYT?

zhaohuabing: I believe it may be sufficient to just support the exact match for the `HeadersToMetadata`, considering that
* HTTP header names are typically fixed and don't change dynamically, and the number of HTTP headers is usually quite limited
* A downstream filter that needs to extract the value from a Metada will also need to know the Metadata name in advance, so having a fixed name seems reasonable

Supporting prefix/suffix/regex may unnecessarily complicate the API.

That said, I could be mistaken on this.  @envoyproxy/gateway-maintainers, please feel free to chime in.

zhaohuabing:
@nothinux  Would you still be interested in working on this PR? If so, could you please raise an API PR as @zirain suggested? (Or you can just use this PR but just keep API change :-))

arkodg:
@zhaohuabing @zirain @nothinux I'm a -1 on introducing this change. There's isnt any notion of reading and writing metadata yet for end users, its still an advanced case.

zhaohuabing:
@arkodg Do you mean that Metadata is an internal mechanism of Envoy, and it should not be directly exposed to end users?

The real users of this API are the ext auth developers, who should already be familiar with Envoy and the concept of Metadata. 

Thinking it through, all custom extensions - Ext Auth, WASM, and Ext Proc, face similar problems: their APIs are designed for developers, not end users.

api/v1alpha1/ext_auth_types.go
nothinux: yes, I think adding prefix, suffix, or regex might complicate the API. For now, could we consider only supporting exact matches @zirain?

arkodg:

PR #4167, opened 2024-09-06
Impl: JWT claim authorization

zhaohuabing:
Implementation for:  https://github.com/envoyproxy/gateway/pull/4009
Related issue: https://github.com/envoyproxy/gateway/issues/3367

This PR also changes the default value of`PayloadInMetada` of the JWT filter from `JwtProvider.Issuer` to `JwtProvider.Name` as `JwtProvider.Issuer` is optional and we need the metadata in the authorization.


examples/kubernetes/http-routing.yaml
```diff
@@ -70,7 +70,7 @@ spec:
             requests:
               cpu: 10m
 ---
-apiVersion: gateway.networking.k8s.io/v1
+apiVersion: gate  way.networking.k8s.io/v1
```
zetaab, line 73: not needed change?

zetaab (commented):
one nit, but I am also thinking that which JWT this is now going to use? There is use-case for both `access_token` and `id_token`. Is there way to specify which one it should use?

zhaohuabing:
> one nit, but I am also thinking that which JWT this is now going to use? There is use-case for both `access_token` and `id_token`. Is there way to specify which one it should use?

It's up to you. A JWT provider must be defined in the JWT configuration within the same `SecurityPolicy`.

If we want to use the ID Token retreived from the OIDC, we need to modify the OAuth2 filter to allow it to add the ID Token in the headers.

zetaab:
I can get this working with idtoken by following configuration (it verifies sub AND that I have some single group that is needed... just for testing)

```
kind: SecurityPolicy
metadata:
  name: demo-api-jwt
  namespace: demo
spec:
  targetRef:
    kind: HTTPRoute
    name: demo-api
  jwt:
    providers:
    - name: cognito
      extractFrom:
        cookies:
        - IdToken
      claimToHeaders:
      - claim: sub
        header: x-sub
      - claim: custom:jobtitle
        header: x-jobtitle
      - claim: custom:department
        header: x-department
      - claim: email
        header: x-email
      - claim: cognito:groups
        header: x-groups
      - claim: name
        header: x-name
      remoteJWKS:
        uri: https://cognito-idp.eu-central-1.amazonaws.com/xxx/.well-known/jwks.json
  authorization:
    defaultAction: Deny
    rules:
    - action: Allow
      principal:
        jwt:
          provider: cognito
          claims:
          - name: sub
            values: [ca28333f-177f-4b7a-90c3-snip]
          - name: cognito:groups
            valueType: StringArray
            values: [AAD_FOOBAR]
```

great job @zhaohuabing

zhaohuabing:
> I can get this working with idtoken by following configuration (it verifies sub AND that I have some single group that is needed... just for testing)
@zetaab Thanks for verifying it!

internal/xds/translator/authorization.go
```diff
@@ -262,6 +278,243 @@ func (*rbac) patchRoute(route *routev3.Route, irRoute *ir.HTTPRoute) error {
...
+				{
+					Segment: &networkinput.DynamicMetadataInput_PathSegment_Key{
+						Key: "scope",
+					},
+				},
+			},
+		}
+
+		// The scope is a space-delimited string, so we use a regex matcher to match the scope.
+		pattern := fmt.Sprintf(`(^|.*\s)%s(\s.*|$)`, scope)
```
arkodg, line 355: @denniskniep had presented a nice suggestion earlier of using https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider-normalizepayload, that may make this faster

api/v1alpha1/authorization_types.go
```diff
@@ -76,12 +81,22 @@ type Principal struct {
 //
 // +kubebuilder:validation:XValidation:rule="(has(self.claims) || has(self.scopes))",message="at least one of claims or scopes must be specified"
 type JWTPrincipal struct {
+	// Provider is the name of the JWT provider that used to verify the JWT token.
+	// The provider must be configured in the JWT authentication.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Provider string `json:"provider"`
```
arkodg, line 89: can we do without this or is this needed ?

test/e2e/testdata/authorization-jwt.yaml
```diff
@@ -0,0 +1,137 @@
...
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: http-with-authorization-jwt-combined
+  jwt:
+    providers:
+    - name: example
+      issuer: https://foo.bar.com
+      remoteJWKS:
+        uri: https://www.zhaohuabing.com/download/jwks.json
```
arkodg, line 123: can we use GH here instead

api/v1alpha1/authorization_types.go
zhaohuabing: It's a must as the JWT auth filter stores the extracted JWT claims under its namespace, and we need this to correlate the JWT principal with the JWT claims in the metadata.

```yaml
   envoy.filters.http.jwt_authn:
      jwt_provider_name:
         scopes: "read write"
         name: "John"
         sub: "John"
```

test/e2e/testdata/authorization-jwt.yaml
zhaohuabing: Sure, I planned to change it to the EG Github one after this PR https://github.com/envoyproxy/gateway/pull/4200, but forgot it. :-)

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4168, opened 2024-09-06
CORS: support wildcard matching for AllowMethods and AllowHeaders

zhaohuabing:
Support wildcard matching for AllowMethods and AllowHeaders
Related: https://github.com/envoyproxy/gateway/issues/4196


api/v1alpha1/cors_types.go
```diff
@@ -29,18 +30,33 @@ type Origin string
 // CORS defines the configuration for Cross-Origin Resource Sharing (CORS).
 type CORS struct {
 	// AllowOrigins defines the origins that are allowed to make requests.
-	// +kubebuilder:validation:MinItems=1
-	AllowOrigins []Origin `json:"allowOrigins,omitempty" yaml:"allowOrigins"`
+	//
+	// +optional
+	AllowOrigins []Origin `json:"allowOrigins,omitempty"`
```
zhaohuabing, line 35: A empty `AllowOrigins` allows users to explicitly disable CORS for an HTTPRoute.

zirain (dismissed):

guydc (approved):
LGTM. Do we need to label this somehow so that we include a breaking change notice in release notes?

api/v1alpha1/cors_types.go
```diff
@@ -29,18 +32,33 @@ type Origin string
...
+
 	// AllowMethods defines the methods that are allowed to make requests.
-	// +kubebuilder:validation:MinItems=1
-	AllowMethods []string `json:"allowMethods,omitempty" yaml:"allowMethods"`
+	//
+	// +optional
```
arkodg, line 41: does unset field mean all methods are allowed ?
relates to https://github.com/envoyproxy/gateway/issues/4196

zhaohuabing: @arkodg unset field means all methods are **not** allowed.

```yaml
  cors:
    allowOrigins:
    - http://*.foo.com
    - http://*.foo.com:80
```

```bash
➜  ~ curl -H "Origin: http://www.foo.com" \
  -H "Host: www.example.com" \
  -X OPTIONS -v -s \
http://172.18.255.200  \
  1> /dev/null
*   Trying 172.18.255.200:80...
* Connected to 172.18.255.200 (172.18.255.200) port 80
> OPTIONS / HTTP/1.1
> Host: www.example.com
> User-Agent: curl/8.8.0-DEV
> Accept: */*
> Origin: http://www.foo.com
> Access-Control-Request-Method: GET
> 
* Request completely sent off
< HTTP/1.1 200 OK
< date: Tue, 10 Sep 2024 07:51:15 GMT
< content-length: 0
```

zhaohuabing:
Although the Envoy documentation doesn’t explicitly mention it, the implementation currently allows the `*` symbol to match all HTTP methods and headers. Should we consider making this behavior more strict by using the `HTTPMethod` and `HTTPHeaderName` types from the Gateway API, or should we continue allowing * to match everything? 

@arkodg @guydc 

I tend to keep the current behavior for back compatiblity.

luvk1412:
Adding my use case here where we require some way to mention Allow all.

- We want to have a global CORS `SecurityPolicy` where cluster Operators can restrict the origins only. If Cluster Operator controls the `allowHeaders` or `exposeHeaders`, then devs have to come to operators every time they add a new header to their req/resp. Devs can ideally create a separate `SecurityPolicy` and override, but we would ideally wan't devs to use only predefined ones via label selectors and not allow them to create new `SecurityPolicy`.
- For `allowMethods`, we can add all enums to policy, as enum set is limited, but adding `*` or some other cleaner approach to allow all would be preferred. I have one more use case in mind but i am not sure if that would be faced practically: Currently allowed enum set contains all the standard Methods but technically methods don't need to be only those, for eg `WebDAV` protocol(we happen to run a `WebDAV` server) has methods like 'PROPFIND` etc., but I don't think any browser/js allows hitting non standard methods. So this might not be an issue for CORS, just mentioning in case i am missing something here and someone else can add.

So for us it would be preferable if there is a way to define allow all for `allowOrigins` , `allowMethods`, `allowHeaders` and `exposeHeaders`
@zhaohuabing

zhaohuabing:
Sounds reasonable. I'll bring this to this week's meeting for further discussion. However, to support * for exposeHeaders, we'll need to change the Envoy side.

luvk1412:
From [Access-Control-Expose-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers)
> For requests without credentials, a server can also respond with a wildcard value

So for requests without creds/cookies, below
```yaml
  exposeHeaders:
  - '*'
 ```
would make it behave like allow all @zhaohuabing

File: api/v1alpha1/cors_types.go [hunk 3]
```diff
@@ -29,18 +32,33 @@ type Origin string
 // CORS defines the configuration for Cross-Origin Resource Sharing (CORS).
 type CORS struct {
 	// AllowOrigins defines the origins that are allowed to make requests.
-	// +kubebuilder:validation:MinItems=1
```
arkodg, line 32: we should also enhance the comments to explain how each field maps to each header
for e.g. `AllowOrigions` specifies the allowed domains for the `Access-Control-Allow-Origin` header. Use `*` to allow any origin.

zhaohuabing:
I misunderstood the `access-control-expose-headers` header. It's enforced by the browser, not the server. :-)

api/v1alpha1/cors_types.go
```diff
@@ -29,18 +29,47 @@ type Origin string
...
-	MaxAge *metav1.Duration `json:"maxAge,omitempty" yaml:"maxAge,omitempty"`
+	// It specifies the value in the Access-Control-Max-Age header.
+	//
+	// +optional
+	MaxAge *metav1.Duration `json:"maxAge,omitempty"`
+
 	// AllowCredentials indicates whether a request can include user credentials
 	// like cookies, authentication headers, or TLS client certificates.
-	AllowCredentials *bool `json:"allowCredentials,omitempty" yaml:"allowCredentials,omitempty"`
+	//
```
arkodg, line 68: can a similar line around which header this corresponds to, be added in here, `Access-Control-Allow-Credentials`

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4170, opened 2024-09-06
api: access log types

### LINKED ISSUE #3600: Support Listener Access Logging
*Description*:
Envoy supports several types of Access Loggers. The [access log](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage) that is most commonly in use is configured in HCM or TCP/UDP/Thrift Proxy. Envoy additionally supports a [listener](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto)-level access log. The listener access log is particularly useful for:
- Providing information related to downstream connection establishment issues, e.g. using the `%DOWNSTREAM_TRANSPORT_FAILURE_REASON%` operator, and following the [TLS troubleshooting guide](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#arch-overview-ssl-trouble-shooting).
- Providing information when other network-filter level access loggers are not invoked (e.g. when there is no matched filter chain, `NR` response flag).
- Reducing log volume: users may want to emit connection-scoped attributes (Source IP/Port, TLS Ciphers, SNI, Client Certificate Attributes, ...) just once per connection, instead of repeatedly emitting the same values with each HTTP log. 

Currently, Envoy Gateway uses the user-defined access log format for the listener access log, but the listener access log is only emitted when there is no matching filter chain.

https://github.com/envoyproxy/gateway/blob/0abdda7a033f0e37647177a588a904c90b783149/internal/xds/translator/accesslog.go#L216

Envoy Gateway should also support listener access logs for TLS troubleshooting. Common issues include:
- Downstream TLS parameter mismatch: supported versions, ciphers, ...
- Downstream Client Certificate validation issues: expiration, trust store compatibility, ...

To support the TLS troubleshooting use case, Envoy Gateway may:
1. Make the listener access log fully configurable for end-users (including format, filters, etc.)
2. Conditionally emit a listener access log containing the `%DOWNSTREAM_TRANSPORT_FAILURE_REASON%` when a TLS connection error occurs. This can be done by appending the TLS details to the user-defined/default access log format, or by emitting a different log that is fully controlled by EG and contains relevant information such as Client IP, Available TLS Context (SNI, Cert details, ...)

To filter listener logs based on a TLS failure,  a CEL access log filter can be used:

```
filter:
extension_filter:
  name: envoy.access_loggers.extension_filters.cel
  typed_config:
    '@type': type.googleapis.com/envoy.extensions.access_loggers.filters.cel.v3.ExpressionFilter
    expression: connection.transport_failure_reason != "-"
```

To support general-purpose listener access logs, the Telemetry Proxy Access Log Setting API may be extended with a new `source` attributes with values `network` (default, TCP/HTTP Proxy Log) and `listener`.  For example:

```
  telemetry:
    accessLog:
      settings:
      - format:
          source: [network|listener] 
          text: |
            [%START_TIME%] "%DOWNSTREAM_REMOTE_ADDRESS%" "%DOWNSTREAM_TRANSPORT_FAILURE_REASON%" "%CONNECTION_TERMINATION_DETAILS%" "%CONNECTION_ID%" "%REQUESTED_SERVER_NAME%"
          type: Text
        matches:
          - 'connection.transport_failure_reason != "-"'
        sinks:
          - file:
              path: /dev/stdout
            type: File
```

Alternatively, a new telemetry attribute called listenerAccessLog can be supported:

```
  telemetry:
    listenerAccessLog:
      settings:
      - format:
          text: |
            [%START_TIME%] "%DOWNSTREAM_REMOTE_ADDRESS%" "%DOWNSTREAM_TRANSPORT_FAILURE_REASON%" "%CONNECTION_TERMINATION_DETAILS%" "%CONNECTION_ID%" "%REQUESTED_SERVER_NAME%"
          type: Text
        matches:
          - 'connection.transport_failure_reason != "-"'
        sinks:
          - file:
              path: /dev/stdout
            type: File
```

guydc:
**What type of PR is this?**
Add support for scoping access log settings to a specific component such as envoy listener or http connection manager. Only support the listener scope at this time. 

If user does not define a type, then configuration would apply to both listener and HCM/TCP/UDPProxy, same as today. If the user defines type: listener, the custom user access log would replace the default one provided by envoy. 

Users may need to customize access logs of specific envoy modules. 
- Certain access log operators are only valid for specific scopes.   
- Access logs may need to be filtered by different conditions when applied to different scopes.

Fixes #3600


api/v1alpha1/accesslogging_types.go
```diff
@@ -30,8 +30,34 @@ type ProxyAccessLogSetting struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=50
 	Sinks []ProxyAccessLogSink `json:"sinks"`
+	// Scope defines the scope of component emitting the accesslog, such as envoy listener, TCP proxy, HTTP connection manager.
+	// If not defined, the setting would apply to all scopes supported by envoy gateway
+	// +kubebuilder:validation:Enum=Listener
+	// +optional
+	// +notImplementedHide
+	Scope ProxyAccessLogScopeType `json:"scope,omitempty"`
```
arkodg, line 38: * currently this only applies to route right ?
* is `Type`a more consistent term ?
* is there a way to reduce the envoyisms here by limiting the terms to `Route` and `Listener` ?

guydc: > is there a way to reduce the envoyisms here by limiting the terms to Route and Listener ?

The best GW-API fit that I can think of is:
- `Frontend` (~ envoy listener)
- `xRoute` (HTTP/GRPCRoute ~ HCM access logs, TCP/UDPRoute ~ TCP/UDP Proxy access logs)
- `xBackend` (GRPC/HTTPBackend ~ Router access logs). Envoy currentl does
WDYT?

api/v1alpha1/accesslogging_types.go
```diff
@@ -30,8 +30,30 @@ type ProxyAccessLogSetting struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=50
 	Sinks []ProxyAccessLogSink `json:"sinks"`
+	// Type defines the component emitting the accesslog, such as Listener and Route.
+	// If not defined, the setting would apply to all accesslog types supported by envoy gateway
+	// +kubebuilder:validation:Enum=Listener
+	// +optional
+	// +notImplementedHide
+	Type ProxyAccessLogType `json:"type,omitempty"`
```
zirain, line 38: if it's optional, it should be a pointer.

api/v1alpha1/accesslogging_types.go
```diff
@@ -30,8 +30,30 @@ type ProxyAccessLogSetting struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=50
 	Sinks []ProxyAccessLogSink `json:"sinks"`
+	// Type defines the component emitting the accesslog, such as Listener and Route.
+	// If not defined, the setting would apply to all accesslog types supported by envoy gateway
+	// +kubebuilder:validation:Enum=Listener
+	// +optional
+	// +notImplementedHide
+	Type *ProxyAccessLogType `json:"type,omitempty"`
```
arkodg, line 38: also if we take the `type` route and both route and listener types exist, we'll have to define the `sinks` twice ?

zirain (commented):
I'm not sure if it's necessary, current konbs used for a long time in istio, and didn't saw too much complaining about it.

If really needed, this design LGTM.

api/v1alpha1/accesslogging_types.go
```diff
@@ -30,8 +30,30 @@ type ProxyAccessLogSetting struct {
...
+	ProxyAccessLogTypeListener ProxyAccessLogType = "Listener"
+	// ProxyAccessLogTypeRoute defines the accesslog for HTTP, GRPC, UDP and TCP Routes.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/udp/udp_proxy/v3/udp_proxy.proto#envoy-v3-api-field-extensions-filters-udp-udp-proxy-v3-udpproxyconfig-access-log
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto#envoy-v3-api-field-extensions-filters-network-tcp-proxy-v3-tcpproxy-access-log
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-access-log
+	ProxyAccessLogTypeRoute ProxyAccessLogType = "Route"
+	// ProxyAccessLogTypeBackend defines the accesslog for HTTP Backends.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto#envoy-v3-api-field-extensions-filters-http-router-v3-router-upstream-log
+	ProxyAccessLogTypeBackend ProxyAccessLogType = "Backend"
```
zirain, line 54: Per our discussion, let us remove this now.

guydc: done

api/v1alpha1/accesslogging_types.go
guydc: clarified the behavior in the doc string and PR desc. When users select a type, the definition will replace any EG defaults for that component and apply only to that component. If type is not specified - behavior would be the same as today (custom format applied to both routes and listeners, default listener format will be configured as well).

api/v1alpha1/accesslogging_types.go
```diff
@@ -30,8 +30,29 @@ type ProxyAccessLogSetting struct {
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=50
 	Sinks []ProxyAccessLogSink `json:"sinks"`
+	// Type defines the component emitting the accesslog, such as Listener and Route.
+	// If type not defined, the setting would apply to all accesslog types supported by envoy gateway.
+	// If type is defined, the default envoy gateway accesslog configuration for the mentioned component would be
```
arkodg, line 35: for my understanding this implies that a `nil` ProxyAccessLogSetting is different that a ProxyAccessLogSetting with a `nil` type set ?

guydc: Right. Let's simplify this and make sure that we don't have surprising behavior for end users where unscoped logs suddenly disappear. 
- if type is not set - behavior remains as today (applied to routes as-is, to listener with additional NR filter)
- if type is set - applies to relevant component.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4175, opened 2024-09-07
bump to go1.22.7

zirain:


zirain:
@arkodg do you think we should bump to `1.23.1`?

shawnh2 (approved):

zhaohuabing (approved):

arkodg:
> @arkodg do you think we should bump to `1.23.1`?

sure +1

Issue #3609, opened 2024-06-14
Support RecomputeRoute for Ext Auth


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

The JWT Authn feature has a knob `RecomputeRoute` to support recomputing the routing decision after going though the jwt filter which may add headers into the request .
https://github.com/envoyproxy/gateway/blob/4e5a2c636c469c2f6faeca759eec64d5b0e55d9d/api/v1alpha1/jwt_types.go#L68

Such a field is also needed in the ext auth filter which may enrich the request with headers added by the ext service that may be used for routing

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #4178, opened 2024-09-07
feat: support `RecomputeRoute` for ExtAuth 

zirain:
xref: https://github.com/envoyproxy/gateway/issues/3609


internal/ir/xds.go
```diff
@@ -883,6 +883,13 @@ type ExtAuth struct {
 	// This setting determines whether to prioritize accessibility over strict security in case of authorization service failure.
 	// +optional
 	FailOpen *bool `json:"failOpen,omitempty"`
+
+	// RecomputeRoute clears the route cache and recalculates the routing decision.
+	// This field must be enabled if the headers generated from the claim are used for
```
zhaohuabing, line 888:
```suggestion
	// This field must be enabled if the headers added or modified by the ExtAuth are used for
```

arkodg (approved):
🚀

PR #4183, opened 2024-09-08
feat: choose a default HTTP1.0 route based on HTTPRoutes if possible.

### LINKED ISSUE #4182: Improve the HTTP/1.0 default host heuristic to also account for routes when possible
*Description*:
HTTP/1.0 allows requests that don't have the `Host` header set in them. When HTTP/1.0 is enabled in ClientTrafficPolicy, it's possible to enable a heuristic that sets the Host header to the listener hostname if the request doesn't have a `Host` header.

If the listener doesn't have a hostname set, but it has a single route that _does_ have a hostname set then it makes sense to use that route's hostname as the default Host header.

liorokman:
When HTTP/1.0 is enabled for a listener and the `UseDefaultHost` flag is set to true, modify the heuristic for calculating the default hostname to add an additional step (number 2 below) as follows:

1. If the listener has a non-wildcard hostname - use it.
2. If there is exactly one `HTTPRoute` attached to the listener that has a single non-wildcard hostname - use it.
3. Fail with and error.

Fixes #4182


guydc (approved):
LGTM, thanks

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -669,8 +670,31 @@ func translateHTTP1Settings(http1Settings *egv1a1.HTTP1Settings, httpIR *ir.HTTP
...
+						// make the linter happy
+						theHost := route.Hostname
+						defaultHost = ptr.To(theHost)
+					}
+					if numMatchingRoutes > 1 {
+						defaultHost = nil
+						break
+					}
+				}
+			}
```
arkodg, line 690: the 2 if checks of `if defaultHost == nil {` did confuse me at first

can we delete L690-692 ?

so once we drop out of the `for` loop we can check for  `if numMatchingRoutes == 0` and `else if numMatchingRoutes != 1`

liorokman: Done.

arkodg (approved):
thanks !

guydc (approved):

Issue #4193, opened 2024-09-09
certgen job is missing parameters for configuring tolerations & affinities in helm chart 


ncsham:
*Description*:
>In our k8s infra , we taint all the nodes with certain label pair, and as a result for scheduling any pod, it requires a toleration to be specified. The issue is with the certgen job of the helm chart. 

Seems like it doesn't have option for giving the toleration for the certgen job in values.yaml . https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/certgen.yaml -> Can see that tolerations & affinity block is not mentioned anywhere.

Can we add the support for mentioning the affinities and tolerations for the certjen job in values.yaml.

PR #4195, opened 2024-09-09
affinity and toleration patch for certgen

ncsham:
**What type of PR is this?**
helm:  certgen : adds support for configuring tolerations and affinity of in the helm chart

By adding this fix, we should be able to configure on which k8s node the job should run on.

Fixes (https://github.com/envoyproxy/gateway/issues/4193)


arkodg:
thanks @ncsham !

* can we also add empty values into the Charts values
```
    affinity: {}
    tolerations: []
    nodeSelector: {}
    ```
https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/values.tmpl.yaml

* can we also add a test in https://github.com/envoyproxy/gateway/tree/main/test/helm/gateway-helm ?

**Comment by ncsham** [comment]
Have added the empty values in values template & also have added helm tests.

File: charts/gateway-helm/values.tmpl.yaml [hunk 1]
```diff
@@ -77,6 +77,10 @@ certgen:
   job:
     annotations: {}
     resources: {}
+    pod:
```
arkodg, line 80: do we need the top level `pod` field or can we do w/o it ? @envoyproxy/gateway-maintainers

arkodg: hey @ncsham lets rm `pod` from here, since its not an explicit field in the job api and not needed here in this API

nitheeshc: sure , will do

ncsham:
@arkodg  I have updated the manifests by removing pod: and kept tols & affis inside job: itself.

I forgot to sign the commit off while pushing and when rebased & merged and pushed back , others commits also came into the same pr.

Let me know if thats okay , or i can delete the pr & recreate again.

charts/gateway-helm/values.tmpl.yaml
ncsham: this is done

zirain:
> @arkodg I have updated the manifests by removing pod: and kept tols & affis inside job: itself.
> 
> I forgot to sign the commit off while pushing and when rebased & merged and pushed back , others commits also came into the same pr. sorry this is my first oss pr
> 
> Let me know if thats okay , or i can delete the pr & recreate again.

please rebase `main` branch, then `git push -f`

ncsham:
hey @zirain thanks , the issue was my main branch not in sync with actual repo ,now i synced the fork and rebased and force pushed.

zirain (approved):

PR #4201, opened 2024-09-10
bump ENVTEST_K8S_VERSION

zirain:
follow up: https://github.com/envoyproxy/gateway/pull/4169


tools/make/kube.mk
```diff
@@ -1,7 +1,7 @@
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION ?= 1.28.0
 # Need run cel validation across multiple versions of k8s
-ENVTEST_K8S_VERSIONS ?= 1.27.1 1.28.0 1.29.0
+ENVTEST_K8S_VERSIONS ?= 1.28.0 1.29.0 1.30.0 1.31.0
```
arkodg, line 4: can we pick the latest patch versions here ?
```
1.28.13 1.29.8 1.30.4 1.31.0
```

zirain: I tried(but failed), it's actual `kubebuilder` versions.

arkodg: ah thanks for sharing that link, here's what I see in the envtest releases
```
1.28.3 1.29.5 1.30.3 1.31.0
```

arkodg (approved):
thanks !

PR #4209, opened 2024-09-10
fix: reject invalid backends in route validation

### LINKED ISSUE #4203: Backends which should be rejected are still used as an HTTPRoute's destination
*Description*:
A `Backend` that doesn't pass validation should not be usable by `HTTPRoute` resources.

*Repro steps*:

Starting with the configuration created by the quickstart sample, I enabled the Backends API extension and created the following configuration resources:

```yaml
kind: Backend
metadata:
  name: admin
  namespace: default
spec:
  endpoints:
  - ip:
      address: 127.0.0.1
      port: 19000
---
kind: HTTPRoute
metadata:
  name: admin
  namespace: default
spec:
  hostnames:
  - localhost
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: gateway.envoyproxy.io
      kind: Backend
      name: admin
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
```

The expected behavior was that the `HTTPRoute` should be rejected because the `Backend` is invalid. The `Backend` is [invalid](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/backend.go#L65) because it uses the `127.0.0.1` IP as a target. 

Instead, Envoy Gateway accepted the route:

```yaml
status:
  parents:
  - conditions:
    - lastTransitionTime: "2024-09-09T14:22:03Z"
      message: Route is accepted
      observedGeneration: 2
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2024-09-09T14:22:03Z"
      message: Resolved all the Object references for the Route
      observedGeneration: 2
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    parentRef:
      kind: Gateway
      name: eg
```

When I setup kubernetes port-forwarding to the Envoy Proxy service, it's possible to access the Envoy Proxy's administration console from a browser.

Additionally, no status was set on the `Backend` resource.

*Environment*:
The latest head of the main branch, though it probably reproduces in released versions as well. Using a `kind` cluster with the quickstart setup.

*Logs*:

From **Envoy Proxy** - note the `upstream_host`:
```
{"start_time":"2024-09-09T14:23:19.454Z","method":"GET","x-envoy-origin-path":"/server_info","protocol":"HTTP/1.1","response_code":"200","response_flags":"-","response_code_details":"via_upstream","connection_termination_details":"-","upstream_transport_failure_reason":"-","bytes_received":"0","bytes_sent":"96866","duration":"3","x-envoy-upstream-service-time":"-","x-forwarded-for":"10.244.0.10","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36","x-request-id":"3776baa7-69bd-48b8-8bee-adf5372ad9c6",":authority":"localhost:8080","upstream_host":"127.0.0.1:19000","upstream_cluster":"httproute/default/admin/rule/0","upstream_local_address":"127.0.0.1:43822","downstream_local_address":"127.0.0.1:10080","downstream_remote_address":"127.0.0.1:58574","requested_server_name":"-","route_name":"httproute/default/admin/rule/0/match/0/localhost"}
```

No errors appear in the Envoy Gateway log.

guydc:
**What type of PR is this?**

Fixes #4203


arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4219, opened 2024-09-11
fix: write backend status

### LINKED ISSUE #4214: Backend resources don't get status updates
*Description*:
After a `Backend` resource is added, Envoy Gateway doesn't update it's status section.

There is [code](https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/backend.go#L19) that creates correct status updates for the `Backend` resource, but this isn't stored in the provider. 

*Repro steps*:

1. Enable the Backend API extension
2. Create a backend.

```sh
kubectl apply -f - << EOF                                                                                                                                                          
kind: Backend
metadata:
  name: httpbin
  namespace: default
spec:
  endpoints:
    - fqdn:
        hostname: httpbin.org
        port: 80
EOF
```

The resource _should_ have a status section indicating that the `Backend` was accepted. After the resource is applied to the cluster, no status section is created for the resource.

*Environment*:

The latest head of the `main` branch, using a local `kind` cluster.

*Logs*:

```
2024-09-11T07:36:45.804Z	INFO	provider	kubernetes/controller.go:381	processing Backend	{"runner": "provider", "kind": "Service", "namespace": "default", "name": "backend"}
```

guydc:
**What type of PR is this?**

Fixes #4214


arkodg (dismissed):
LGTM thanks !

arkodg (approved):

zirain (approved):

PR #4224, opened 2024-09-12
api: unhide `Matches`  in ProxyAccessLog

zirain:


arkodg (approved):
thanks !

arkodg:
guessing this also should be backported so it shows up in the v1.1 docs section ?

zirain:
> guessing this also should be backported so it shows up in the v1.1 docs section ?

done

arkodg (approved):

shawnh2 (approved):

Issue #4226, opened 2024-09-12
EnvoyGateway rateLimitDeployment ignores pod labels


oscarboher:
*Description*:
When configuring the rate limit deployment through the `envoy-gateway-config` configmap, the `provider.kubernetes.rateLimitDeployment.pod.labels` field is ignored.
Other fields at the same level like annotations are respected.

*Repro steps*:
* gateway v1.1 (experimental channel)
* envoy-gateway installed through helm chart v0.0.0-latest
* create GatewayClass with envoy controller.
* Update `envoy-gateway-config` configmap in `envoy-gateway-system` namespace with the following values:
  
  ````
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: envoy-gateway-config
    namespace: envoy-gateway-system
  data:
    envoy-gateway.yaml: |
      apiVersion: gateway.envoyproxy.io/v1alpha1
      kind: EnvoyGateway
      provider:
        type: Kubernetes
        kubernetes:
          rateLimitDeployment:
            replicas: 4
            pod:
              labels:
                my-label: test
              annotations:
                test-anno: test
      gateway:
        controllerName: gateway.envoyproxy.io/gatewayclass-controller
      rateLimit:
        backend:
          type: Redis
          redis:
            url: redis.redis-system.svc.cluster.local:6379
        telemetry:
          metrics:
            prometheus:
              disable: false
          
        
  ```
* Rollout restart envoy-gateway deployment for it to pick up the changes.
* Deployment envoy-ratelimit has the annotation `test-anno: test` in the pod spec, but not the label `my-label: test`

*Environment*:
* gateway v1.1 (experimental channel)
* envoy-gateway installed through helm chart v0.0.0-latest
* image gw controller: docker.io/envoyproxy/gateway-dev:latest

zirain:
this indeed a bug, @oscarboher would you like to send a fix?

BTW, I think we may also have issue on merging annotations.

zirain:
that's the point.

arkodg:
hey @oscarboher thanks for giving it a try, assigning this you for now, let us know how we can help

arkodg:
a workaround could be to use the `patch` field  https://gateway.envoyproxy.io/docs/api/extension_types/#kubernetesdeploymentspec

I haven't tested this config, but something similar should help here

```
rateLimitDeployment:
  patch:
    template:
      metadata:
        labels:
          my-label: test
```

oscarboher:
[...]

PR #4228, opened 2024-09-12
fix: rateLimitDeployment ignoring pod labels and annotation merge

oscarboher:
**What type of PR is this?**
fix
`rateLimitDeployment.pod.labels` where not applying to rate limit pods.
prometheus annotations where ignored if `rateLimitDeployment.pod.annotations` was defined

Fixes https://github.com/envoyproxy/gateway/issues/4226


internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -208,16 +217,16 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
...
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: r.rateLimitDeployment.Replicas,
 			Strategy: *r.rateLimitDeployment.Strategy,
 			Selector: selector,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels:      selector.MatchLabels,
-					Annotations: annotations,
+					Labels:      podLabels,
```
arkodg, line 228: these are immutable so updating the labels will result in a failed deployment
cc @sanposhiho

sanposhiho: To be accurate, .spec.selector is immutable, while .spec.template.labels isn't. But, we have to make sure the .spec.template.labels match .spec.selector, when updating it.

@oscarboher 
You can refer to the idea -
https://github.com/sanposhiho/gateway/blob/69bf8825422db0cb7b34097b8db5dfca19b8d477/internal/infrastructure/kubernetes/proxy/resource_provider.go#L195-L206
how we can make the selector stable while allowing patching labels.

So, basically, we should build the selector only with the labels that will never get changed. In other words, we shouldn't build the selector with the custom labels that users can change.

Also, you have to pay attention not to break the existing users who already have this deployment on their cluster.

oscarboher: Currently the .spec.selector is a selector with the labels:
```
"app.kubernetes.io/name": InfraName,
"app.kubernetes.io/component":  "ratelimit",
"app.kubernetes.io/managed-by": "envoy-gateway",
```
So i understand this must remain the same to not break things because of the selector immutability.

Isn't it enough to ensure that the podLabels include at least those labels for the selector to work, as my PR does? It does look a bit funny using the copy function but it works.
What am I missing? Or the idea is just to have a method that can be used elsewhere to maintain some selectors?

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -186,19 +187,28 @@ func (r *ResourceRender) ServiceAccount() (*corev1.ServiceAccount, error) {
...
 func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 	containers := expectedRateLimitContainers(r.rateLimit, r.rateLimitDeployment, r.Namespace)
-	labels := rateLimitLabels()
-	selector := resource.GetSelector(labels)
+	selector := resource.GetSelector(rateLimitLabels())
+
+	podLabels := rateLimitLabels()
+	if r.rateLimitDeployment.Pod.Labels != nil {
+		maps.Copy(podLabels, r.rateLimitDeployment.Pod.Labels)
+		maps.Copy(podLabels, rateLimitLabels())
```
arkodg, line 195: can L195 be removed ? since L192 is already setting it

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
arkodg: thanks @sanposhiho and @oscarboher, this change is safe since the `selector` isnt changing

sanposhiho: Oops, sorry, I thought `rateLimitLabels()` included `rateLimitDeployment.pod.labels` too. Yes, then nothing to worry here around the immutability.

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
oscarboher: So `maps.Copy(dest, src)` [overwrites keys existing ](https://pkg.go.dev/maps#Copy) in both maps with the `src` value. I want to make sure noone can override one of the selector labels from the `rateLimitLabels`, so I'm doing the last maps.Copy to overwrite with the selector labels in case someone added one of those labels in their values.

Also the destination map must not be nil so I cannot do something like:
```
podLabels := r.rateLimitDeployment.Pod.Labels
maps.Copy(podLabels, rateLimitLabels())
```
as it will fail if no pod labels are provided.
It looks a bit funny, I just didn't want to implement a function that merges two maps, but maybe it's worth it for clarity's sake. Let me know if you think so.

arkodg: ah thats a good way to ensure the labels can't be overridden
nit: this as a code comment would be helpful so the next dev gets the "why"

arkodg (dismissed):
LGTM thanks !

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4239, opened 2024-09-14
feat: Add maxUnavailable PDB support to helm chart

### LINKED ISSUE #4238: Helm: PDB maxUnavailable support
*Description*:
would be used.
I'd like to be able to configure maxUnavailable in the PodDisruptionbBudget

[optional *Relevant Links*:]

jukie:
**What type of PR is this?**

Adds ability to set `maxUnavailable` in the PodDisruptionBudget

Fixes #4238


charts/gateway-helm/templates/envoy-gateway-deployment.yaml
```diff
@@ -1,25 +1,3 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: envoy-gateway
-  namespace: '{{ .Release.Namespace }}'
-  labels:
-  {{- include "eg.labels" . | nindent 4 }}
----
-{{- if and .Values.podDisruptionBudget.minAvailable (ge (int .Values.podDisruptionBudget.minAvailable) 1) }}
```
arkodg, line 9: prefer keeping this as this to make this backwards compatible and only inserting
```
maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
```

along with the `if` that will needed to ensure this template kicks in
cc @alexwo @guydc

jukie: Do you mean you'd prefer having it always being applied, and if not input provided would default to minAvailable=0?

arkodg: the top level `if` could we tweaked to `{{- if and .Values.podDisruptionBudget }}` i.e. create the resource, if any field is set

arkodg:
can you also add a test or modify this one https://github.com/envoyproxy/gateway/blob/main/test/helm/gateway-helm/control-plane-with-pdb.in.yaml

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
jukie: Makes sense! I can adjust it a bit more if preferred

jukie:
@arkodg anything else I should add?

zirain (approved):

shawnh2 (approved):
LGTM

PR #4256, opened 2024-09-16
EnvoyPatchPolicy JsonPath docs & fixes

### LINKED ISSUE #4043: docs: JSONPath usage in EnvoyPatchPolicy
*Description*:
>Describe the issue.

Support for JSONPath was recently added to EnvoyPatchPolicy to be able to better select elements in a list
More in https://github.com/envoyproxy/gateway/pull/3757

Lets add an example in https://gateway.envoyproxy.io/docs/tasks/extensibility/envoy-patch-policy/ highlighting this use case

[optional *Relevant Links*:]

### LINKED ISSUE #4162: JSONPath not correctly translated to JSONPatch paths
*Description*:
Consider the following fragment of the Envoy Proxy bootstrap yaml:
```yaml
layered_runtime:
  layers:
  - name: global_config
    static_layer:
      re2.max_program_size.error_level: 4294967295
      re2.max_program_size.warn_level: 1000
  - name: some_other_layer
     static_layer:
        key.entry: value
```

I would like to add the key `envoy.some.key.or.other.made.up` into the `static_layer` of the layer called `global_config`.

A JSONPatch that works looks like this:

```yaml
- op: add
  path: /layered_runtime/layers/0/static_layer/envoy.some.key.or.other.made.up
  value: a value
```

But this patch assumes that the layer I want to add the key to is always the first. 

I tried to use a `JSONPath` instead, with the following example:

```yaml
- op: add
  jsonPath: $.layered_runtime.layers[?@.name=='global_config'].static_layer['envoy.some.key.or.other.made.up']
  value: a value
```

I have to use the bracket syntax (`['envoy.some.key.or.other.made.up']`) because the dots must not be treated as one of the JSONPath separators. This is valid JSONPath according to the [RFC](https://www.rfc-editor.org/rfc/rfc9535.html):

> JSONPath expressions move further down the hierarchy using dot notation ($.store.book[0].title) or the bracket notation ($['store']['book'][0]['title']); both replace XPath's / within query expressions, where dot notation serves as a lightweight but limited syntax while bracket notation is a heavyweight but more general syntax.

This results in the following key added to the correct hash:
```yaml
layers:
   - name: global_config
     static_layer:
      '[''envoy.some.key.or.other.made.up'']': a value
       envoy.restart_features.use_eds_cache_for_ads: false
       envoy.something.completely.made.up: arbitrary string
       re2.max_program_size.error_level: 4294967295
```

I get the same result for this JSONPath string as well:

```yaml
- op: add
  jsonPath: $['layered_runtime']['layers'][?@.name=='global_config']['static_layer']['envoy.some.key.or.other.made.up']
  value: a value
```

The translation from JSONPath to a Path used by the JSONPatch should not have the brackets as part of the result.

`$.layered_runtime.layers[?@.name=='global_config'].static_layer['envoy.some.key.or.other.made.up']` should translate to `/layered_runtime/layers/0/static_layer/envoy.some.key.or.other.made.up` and not to `/layered_runtime/layers/0/static_layer/'[''envoy.some.key.or.other.made.up'']'`

denniskniep:
**What type of PR is this?**
* "fix: throw error if jsonPath returns no jsonPointers"
* "fix: allow selecting properties in jsonPath with special chars"
* "fix: generate jsonPointer and encode special chars"
* "docs: improved api (property descriptions)"
* "docs: added example"

Fixes #4162 
Fixes #4043

### DIFF
--- a/api/v1alpha1/envoypatchpolicy_types.go
+++ b/api/v1alpha1/envoypatchpolicy_types.go
@@ -109,12 +109,16 @@ type JSONPatchOperationType string
 type JSONPatchOperation struct {
 	// Op is the type of operation to perform
 	Op JSONPatchOperationType `json:"op"`
-	// Path is the location of the target document/field where the operation will be performed
-	// Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
+	// Path is a JSONPointer expression. Refer to https://datatracker.ietf.org/doc/html/rfc6901 for more details.
+	// It specifies the location of the target document/field where the operation will be performed
 	// +optional
 	Path *string `json:"path,omitempty"`
-	// JSONPath specifies the locations of the target document/field where the operation will be performed
-	// Refer to https://datatracker.ietf.org/doc/rfc9535/ for more details.
+	// JSONPath is a JSONPath expression. Refer to https://datatracker.ietf.org/doc/rfc9535/ for more details.
+	// It produces one or more JSONPointer expressions based on the given JSON document.
+	// If no JSONPointer is found, it will result in an error.
+	// If the 'Path' property is also set, it will be appended to the resulting JSONPointer expressions from the JSONPath evaluation.
+	// This is useful when creating a property that does not yet exist in the JSON document.
+	// The final JSONPointer expressions specifies the locations in the target document/field where the operation will be applied.
 	// +optional
 	JSONPath *string `json:"jsonPath,omitempty"`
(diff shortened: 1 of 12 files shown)


liorokman (approved):
LGTM, thanks!

denniskniep:
cc: @arkodg

arkodg (approved):
thanks !

PR #4257, opened 2024-09-16
feat: enable local validations for egctl translate and file provider

### LINKED ISSUE #1295: Add Gateway API webhook validation functions to egctl translate

shawnh2:
**What type of PR is this?**

Fixes #1295


internal/gatewayapi/resource/strategy.go
```diff
@@ -0,0 +1,197 @@
...
+	// Since it's a lazy loading map, calling getCreateStrategyMapForCRDs
+	// before using this map, directly using this map is not recommended.
+	crdCreateStrategyMap versionedCreateStrategyMap
+	once                 sync.Once
+)
+
+func getCreateStrategyMapForCRDs() versionedCreateStrategyMap {
+	once.Do(func() {
+		// This error should never happen. If it does, stop loading strategy map.
+		crds, _ := loadCRDs()
```
arkodg, line 56: is this an `embed` to use the CRD def for validation ? curious what's the binary size bump ?

arkodg: you dont need to use `embed.FS` you can just use embed to create 2 strings off 2 files (upstream gateway API CRDs and EG CRDs) https://pkg.go.dev/embed#pkg-overview

internal/gatewayapi/resource/strategy.go
```diff
@@ -0,0 +1,197 @@
...
+		crdCreateStrategyMap = make(versionedCreateStrategyMap, len(crds))
+		for _, crd := range crds {
+			vcs, _ := getVersionedCreateStrategyForCRD(crd)
+			crdCreateStrategyMap[crd.Spec.Names.Kind] = vcs
+		}
+	})
+	return crdCreateStrategyMap
+}
+
+// getVersionedCreateStrategyForCRD will only get latest version of the input CRD and
```
arkodg, line 69: is this what `kubectl validate` uses under the hood ?

internal/gatewayapi/resource/strategy.go
shawnh2: Wow thanks! It really helps!

internal/gatewayapi/resource/strategy.go
shawnh2: sort of, we both use Strategy to validate resources before creating.

arkodg: yeah would make it more maintainable imo

internal/gatewayapi/resource/embed.go
```diff
@@ -0,0 +1,73 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
```
arkodg, line 5: can you `git rm` this file `internal/gatewayapi/resource/crd/gateway-crds.yaml`

tools/make/kube.mk
```diff
@@ -43,6 +43,8 @@ CONTROLLERGEN_OBJECT_FLAGS :=  object:headerFile="$(ROOT_DIR)/tools/boilerplate/
 manifests: $(tools/controller-gen) generate-gwapi-manifests ## Generate WebhookConfiguration and CustomResourceDefinition objects.
 	@$(LOG_TARGET)
 	$(tools/controller-gen) crd:allowDangerousTypes=true paths="./api/..." output:crd:artifacts:config=charts/gateway-helm/crds/generated
+	# save a copy of all crds for local resource validator.
+	find charts/gateway-helm/crds -name "*.yaml" -exec cat {} + > internal/gatewayapi/resource/crd/gateway-crds.yaml
```
arkodg, line 47: can we do w/o this extra copy ?

shawnh2: Cannot think of better way to avoid copy. 

`embed` enforce the path must not contain `.` or `..` path elements nor begin with a leading slash, so it means the `gateway-crds.yaml` must under the `resource` pkg.

arkodg: I would instead prefer if we created a top level `embed.go` in the root dir, so we eliminated this copy

shawnh2: brilliant idea!

arkodg: im not very proud of it 🙈 , it make the root dir a little impure, but its reduces the amount of diffs generated every time a CRD changes

arkodg (approved):
LGTM thanks ! 
this is a really neat feature

Xunzhuo (approved):
Nice!

Issue #1794, opened 2023-08-17
Support regex based path rewrites


borederman:
https://github.com/envoyproxy/gateway/blob/677cf033f74babcab70670d87a1cf1cd105e4b48/internal/xds/translator/route.go#L258

arkodg:
@borederman can you share the gateway API config / filter you are trying to apply ?

borederman:
For example with the following config:  

would transform `http://www.example.com/gw/api/get`  into  `http://www.example.com/get/api`

``` yaml
    filters:
    - type: URLRewrite
      urlRewrite:
        path:
          replaceFullPath: /\2/\1 
          type: ReplaceFullPath
    matches:
    - path:
        type: RegularExpression
        value: ^/gw/(api)/(.+)$  # as `RegexMatchAndSubstitute.pattern`
```

borederman:
> @borederman can you share the gateway API config / filter you are trying to apply ?

For example with the following config:  

would transform `http://www.example.com/gw/api/get`  into  `http://www.example.com/get/api`

``` yaml
    filters:
    - type: URLRewrite
      urlRewrite:
        path:
          replaceFullPath: /\2/\1 
          type: ReplaceFullPath
    matches:
    - path:
        type: RegularExpression
        value: ^/gw/(api)/(.+)$  # as `RegexMatchAndSubstitute.pattern`
```

borederman:
@arkodg  I do not know if the problem was described clearly or not.

arkodg:
is this same as https://github.com/kubernetes-sigs/gateway-api/issues/2177 ?

borederman:
oh, no.....

arkodg:
fixed with https://github.com/envoyproxy/gateway/pull/4258

PR #4258, opened 2024-09-16
feat(translator): implement httproutefilter and path regex rewrite

### LINKED ISSUE #4120: Consider adding a HTTPRouteFilterExtensionRef for Envoy Gateway
*Description*:
would be used.

The project has been able to add functionality not yet defined in the Gateway API using [Extensions](https://gateway.envoyproxy.io/docs/concepts/concepts_overview/#envoy-gateway-eg-api-resources)

However some features that naturally live at the `rule` level within a `xRoute` that are API specific, dont fit well within Policies such as 

* Direct Response https://github.com/envoyproxy/gateway/issues/2714
* Regex Based Rewrites https://github.com/envoyproxy/gateway/issues/1794
* Body Transformations https://github.com/envoyproxy/envoy/issues/35783

Adding support for these features in the Gateway API will take time, since building a common API across multiple implementations is not a trivial task, however this is impacting users migrating to Envoy Gateway from other projects who are unable to do so because these common features dont exist yet

Raising this issue to propose adding a new CRD that can be attached to a HTTPRoute rule using the `extensionRef` field within HTTPRoute https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteFilter

guydc:
**What type of PR is this?**

- Adds an Envoy Gateway Extension filter, `HTTPRouteFilter`
- Implements support for regex rewrite of URL paths in HTTPRouteFilter

Fixes #4120, #1794


api/v1alpha1/httproutefilter_types.go
```diff
@@ -56,14 +56,17 @@ const (
...
 	Pattern string `json:"pattern"`
 	// Substitution is an expression that replaces the matched portion.The expression may include numbered
 	// capture groups that adhere to syntax documented in https://github.com/google/re2/wiki/Syntax.
 	Substitution string `json:"substitution"`
 }
 
+// +kubebuilder:validation:XValidation:rule="self.type == 'ReplaceRegexMatch' ? has(self.replaceRegexMatch) : !has(self.replaceRegexMatch)",message="If HTTPPathModifier type is ReplaceRegexMatch, replaceRegexMatch field needs to be set."
 type HTTPPathModifier struct {
-	// +kubebuilder:validation:Enum=RegexHTTPPathModifier
+	// +kubebuilder:validation:Enum=ReplaceRegexMatch
```
zhaohuabing, line 68: Should EG `HTTPURLRewriteFilter` be a superset of https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPURLRewriteFilter?

For example, if I want to rewrite both the host and path by regex matching, it would not be feasible with the current EG implementation as EG HTTPRouteFilter doesn't support host rewrite and Gateway API takes precedence if both filters are sepecified.

internal/gatewayapi/filters.go
```diff
@@ -150,16 +152,20 @@ func (t *Translator) processURLRewriteFilter(
...
-		status.SetRouteStatusCondition(routeStatus,
-			filterContext.ParentRef.routeParentStatusIdx,
-			filterContext.Route.GetGeneration(),
-			gwapiv1.RouteConditionAccepted,
-			metav1.ConditionFalse,
-			gwapiv1.RouteReasonUnsupportedValue,
-			"Cannot configure multiple urlRewrite filters for a single HTTPRouteRule",
-		)
-		return
+		if filterContext.URLRewrite.Hostname != nil ||
```
zhaohuabing, line 155: It might be a corner case, but would it be possible to rewrite both the Hostname and the Path?

internal/provider/kubernetes/indexers.go
```diff
@@ -114,6 +119,26 @@ func backendHTTPRouteIndexFunc(rawObj client.Object) []string {
 	return backendRefs
 }
 
+func httpRouteFilterHTTPRouteIndexFunc(rawObj client.Object) []string {
+	httproute := rawObj.(*gwapiv1.HTTPRoute)
+	var httpRouteFilterRefs []string
+	for _, rule := range httproute.Spec.Rules {
+		for _, filter := range rule.Filters {
+			if filter.ExtensionRef != nil && string(filter.ExtensionRef.Kind) == resource.KindHTTPRouteFilter {
+				// If an explicit Backend namespace is not provided, use the HTTPRoute namespace to
```
zhaohuabing, line 128: Nit: remove this comment.

zhaohuabing (dismissed):
This PR looks great! I just have a nit on the comment, which is non-blocking.

internal/ir/xds.go
```diff
@@ -1375,6 +1388,42 @@ func (r HTTPPathModifier) Validate() error {
 	return errs
 }
 
+// ExtendedHTTPPathModifier holds instructions for how to modify the path of a request on a redirect response
+// with both core gateway-api and extended envoy gateway capabilities
+// +k8s:deepcopy-gen=true
+type ExtendedHTTPPathModifier struct {
+	HTTPPathModifier `json:",inline" yaml:",inline"`
+	// RegexMatchReplace provides a regex to match an a replacement to perform on the path.
+	RegexMatchReplace *RegexMatchReplace `json:"regexMatchReplace" yaml:"regexMatchReplace"`
```
arkodg, line 1397: `omitempty`

arkodg:
one nit around tag, else LGTM, thanks for adding this !

zirain (approved):

arkodg (approved):

Issue #1400, opened 2023-05-05
Support Custom Error Responses


qicz:
xref: https://www.envoyproxy.io/docs/envoy/v1.25.0/configuration/http/http_conn_man/local_reply.html?highlight=local%20reply

```yaml
apiVersion: config.gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: with-local-reply-config
spec:
  localReplyConfig:
    mappers:
    - filter:
        status_code_filter:
          comparison:
            op: EQ
            value:
              default_value: 400
              runtime_key: key_b
      status_code: 401
      body_format_override:
        text_format: "<h1>%LOCAL_REPLY_BODY% %REQ(:path)%</h1>"
        content_type: "text/html; charset=UTF-8"
    - filter:
        status_code_filter:
          comparison:
            op: EQ
            value:
              default_value: 500
              runtime_key: key_b
      status_code: 501
    bodyFormat:
      text_format: "%LOCAL_REPLY_BODY% %RESPONSE_CODE%"
```

 **using the go-control-panel API, so some file naming styles followed** 

API definition

```golang
// EnvoyProxySpec defines the desired state of EnvoyProxy.
type EnvoyProxySpec struct {
	// Provider defines the desired resource provider and provider-specific configuration.
	// If unspecified, the "Kubernetes" resource provider is used with default configuration
	// parameters.
	//
	// +optional
	Provider *EnvoyProxyProvider `json:"provider,omitempty"`

	// Logging defines logging parameters for managed proxies. If unspecified,
	// default settings apply. This type is not implemented until
	// https://github.com/envoyproxy/gateway/issues/280 is fixed.
	//
	// +kubebuilder:default={level: {system: info}}
	Logging ProxyLogging `json:"logging,omitempty"`

	// Bootstrap defines the Envoy Bootstrap as a YAML string.
	// Visit https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-bootstrap
	// to learn more about the syntax.
	// If set, this is the Bootstrap configuration used for the managed Envoy Proxy fleet instead of the default Bootstrap configuration
	// set by Envoy Gateway.
	// Some fields within the Bootstrap that are required to communicate with the xDS Server (Envoy Gateway) and receive xDS resources
	// from it are not configurable and will result in the `EnvoyProxy` resource being rejected.
[...]

PR #4259, opened 2024-09-16
api: custom error response

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/1400


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -69,6 +69,12 @@ type BackendTrafficPolicySpec struct {
 	// +optional
 	// +notImplementedHide
 	Compression []*Compression `json:"compression,omitempty"`
+
+	// CustomErrorResponse defines the configuration to override specific responses with a custom one.
+	//
+	// +optional
+	// +notImplementedHide
+	CustomErrorResponse []*CustomErrorResponse `json:"customErrorResponse,omitempty"`
```
arkodg, line 77: other options for names
* errorPage
* customErrorPage
* customError
* customResponse
* errorResponseOverride

api/v1alpha1/shared_types.go
```diff
@@ -609,3 +609,35 @@ type HTTP2Settings struct {
...
+	// Match configuration.
+	Match CustomErrorResponseMatch `json:"match"`
+	// Response configuration.
+	Response CustomResponse `json:"response"`
+}
+
+// CustomErrorResponseMatch defines the configuration for matching a user response to return a custom one.
+type CustomErrorResponseMatch struct {
+	// Status codes to match on. A match is successful if any of the status codes match.
+	StatusCodes []string `json:"statusCodes"`
```
arkodg, line 624: * should we support values like `4xx` ?
* should we support ranges like `500-520` ?

api/v1alpha1/backendtrafficpolicy_types.go
AliceProxy: I slightly prefer something like `responseOverride` since while the most common use-case is overriding error responses, users might find other use cases for it. No major complaints with `customErrorResponse` though.

api/v1alpha1/shared_types.go
```diff
@@ -609,3 +609,35 @@ type HTTP2Settings struct {
 	// +optional
 	OnInvalidMessage *InvalidMessageAction `json:"onInvalidMessage,omitempty"`
 }
+
+// CustomErrorResponse defines the configuration to override specific responses with a custom one.
+type CustomErrorResponse struct {
+	// Match configuration.
+	Match CustomErrorResponseMatch `json:"match"`
+	// Response configuration.
+	Response CustomResponse `json:"response"`
```
AliceProxy, line 618: If we lean more strongly into the "override" terminology for this feature, we could change `response` to `overrides`. I would have to double check the xDS, but I believe you can do a "partial override" to do things like change/add response headers, but only when an error happens. I was asked the other day if Emissary/Edge Stack supported that for a large org that needed it and while Emissary/Edge Stack currently don't it would be cool to support it here if possible.

api/v1alpha1/shared_types.go
AliceProxy: Yeah, some way to provide a range is definitely important since 4xx/5xx will be a very common use-case

api/v1alpha1/shared_types.go
```diff
@@ -609,3 +609,35 @@ type HTTP2Settings struct {
...
+
+	// Body of the Custom Response
+	Body CustomResponseBody `json:"body"`
+}
+
+// CustomResponseBody
+type CustomResponseBody struct {
+	// ValueRef contains the contents of the body.
+	// Only a reference to ConfigMap is supported.
+	ValueRef *gwapiv1.LocalObjectReference `json:"valueRef,omitempty"`
```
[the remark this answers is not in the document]
AliceProxy: Could support an in-line string, keep the configmap, local file, etc. I think confimap is the most important though.

AliceProxy (commented):
I like where you're going with this API. Thoughts on the following? Not all the examples in my example below need to be added to the first iteration, just a suggestion around what we might eventually want to support.


```yaml
responseOverrides:
  - match:
      status:
        type: value
        value: 404
    body:
    ...
  - match:
      status:
        type: range
        range:
          min: 400
          max: 499
      headers:
      - name: "x-foo"
        value: "bar" # no value means any value, same as providing "*"
    headers: # no headers means don't change the headers
    - type: "set"
      name: "x-foo"
      value: "baz"
    - type: "append"
      name: "x-bar"
      value: "baz"
    - type: "remove"
      name: "x-bar"
    body: # no balue means don't change the body
    ...
```

I think we want to give people an easy way to change responses (and maybe headers) under certain responses from the upstream. One open question with the above is how we would want to handle the case where you supply a range then an individual specific status such as in my example. I think intuitively I would expect the exact `404` to take precedence over the `400-499` range, but curious what others think.

I think that pattern of having a generic fallback page for general 4xx/5xx errors will be very common and also wanting specific pages for more well-known cases like `404`/`500` with a little extra personality/info like many other large sites do.

api/v1alpha1/backendtrafficpolicy_types.go
zirain: could it be used with any response codes? if so, prefer to `responseOverride`.

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
zirain: I think we should support in-line string for mostly simple use cases.

zirain:
instead of using an range match:
```
match:
      status:
        type: range
        range:
          min: 400
          max: 499
```

what about use CEL, and it will looks like `http.statusCode >= 400 || http.statusCode <= 499`?

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
arkodg: any suggestions on how to represent both  ?
```
InlineValue *string
ValueRef *gwapiv1.LocalObjectReference
```
and add a CEL to make sure only 1 can be selected ?

arkodg:
thanks for the review and suggestions @AliceProxy & @zirain ! will try and incorporate these changes in

arkodg:
> instead of using an range match:
> 
> ```
> match:
>       status:
>         type: range
>         range:
>           min: 400
>           max: 499
> ```
> 
> what about use CEL, and it will looks like `http.statusCode >= 400 || http.statusCode <= 499`?

CEL is flexible and powerful as you've shown in the example, I fear we'll end up losing on simplicity

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
AliceProxy: Yeah, for things like this, I'm partial to using a kind/type identifier like in my examples below. That way, let's say a company is running Envoy Gateway and they have a bunch of config, they hire a new engineer who isn't familiar with Envoy Gateway. From glancing at their config it is clear that these kinds of fields is an either/or situation and not something that can be both configured at the same time. 

A lot of times, that is fairly self-evident and users could just find out when they go to apply and the apply-time validation catches it, or they could reference the docs, but I think small additions like this help a lot with readability.

Maybe something kind of like:

```yaml
body:
  type: inline
  inline: | 
  This is a plain text body
```

```yaml
body:
  type: local-object-ref
  localObjectRef: ....
```

api/v1alpha1/backendtrafficpolicy_types.go
AliceProxy: Should be able to do it on any statuses AFAIK

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
arkodg: yeah cool, I can add a `type` field too, we've followed the `type` approach in most places of the API

api/v1alpha1/backendtrafficpolicy_types.go
zhaohuabing: I slightly prefer removing `error` from the name as it feels more future-proof. Either `responseOverride` or `customRespone` looks good.

api/v1alpha1/shared_types.go
zhaohuabing: I think status classes like `4xx` and `5xx` are good enough for now. I can't think of any use cases like `500-520`, but if they do exist, we can add that later.

api/v1alpha1/backendtrafficpolicy_types.go
qicz: `responseOverride` +1

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
qicz: IMO, should support `ContentType` for in-line string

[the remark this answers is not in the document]
zirain: > IMO, should support `ContentType` options for in-line string

good catch.

zirain:
> > instead of using an range match:
> > ```
> > match:
> >       status:
> >         type: range
> >         range:
> >           min: 400
> >           max: 499
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what about use CEL, and it will looks like `http.statusCode >= 400 || http.statusCode <= 499`?
> 
> CEL is flexible and powerful as you've shown in the example, I fear we'll end up losing on simplicity

let's use range right now.

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: thanks for suggestion everyone, renaming to `responseOverride`

api/v1alpha1/shared_types.go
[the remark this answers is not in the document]
arkodg: * there's already a field for `ContentType` defined
* I'll go ahead and add a `inline` support for the body

arkodg:
> I like where you're going with this API. Thoughts on the following? Not all the examples in my example below need to be added to the first iteration, just a suggestion around what we might eventually want to support.
> 
> ```yaml
> responseOverrides:
>   - match:
>       status:
>         type: value
>         value: 404
>     body:
>     ...
>   - match:
>       status:
>         type: range
>         range:
>           min: 400
>           max: 499
>       headers:
>       - name: "x-foo"
>         value: "bar" # no value means any value, same as providing "*"
>     headers: # no headers means don't change the headers
>     - type: "set"
>       name: "x-foo"
>       value: "baz"
>     - type: "append"
>       name: "x-bar"
>       value: "baz"
>     - type: "remove"
>       name: "x-bar"
>     body: # no balue means don't change the body
>     ...
> ```
> 
> I think we want to give people an easy way to change responses (and maybe headers) under certain responses from the upstream. One open question with the above is how we would want to handle the case where you supply a range then an individual specific status such as in my example. I think intuitively I would expect the exact `404` to take precedence over the `400-499` range, but curious what others think.
> 
> I think that pattern of having a generic fallback page for general 4xx/5xx errors will be very common and also wanting specific pages for more well-known cases like `404`/`500` with a little extra personality/info like many other large sites do.

@AliceProxy id say 1st match wins

api/v1alpha1/shared_types.go
arkodg: ive gone ahead and added `range`

AliceProxy:
> @AliceProxy I prefer 1st match wins, thats what we've done in the JWT Authz API

Sounds good. I havent looked too much at the current APIs around auth, but keeping consistent behaviour across similar patterns is more important.

AliceProxy (approved):
API looks great :+1:

zirain (approved):

PR #4271, opened 2024-09-18
OIDC: add nonce support to the oauth2 filter

zhaohuabing:
This PR introduce nonce support to the oauth2 authentication flow, enhancing EG's security posture by preventing malicious  users from injecting their own access token into a victim's session through CSRF attacks.

Additional background information can be found at:
* https://github.com/envoyproxy/envoy/issues/35232
* https://github.com/envoyproxy/envoy/pull/35919


arkodg (approved):
🎉  thanks for adding this in upstream @zhaohuabing !

Xunzhuo (approved):
Nice works! Thanks @zhaohuabing

PR #4274, opened 2024-09-18
ci: remove license override for github.com/moby/patternmatcher

shahar-h:

Remove license override for github.com/moby/patternmatcher in osv-scanner configuration since https://github.com/google/deps.dev/issues/106 is resolved.


zirain (approved):

arkodg (approved):
thanks for tracking the work in upstream!

PR #4279, opened 2024-09-18
fix: Reconcile on HTTPRoute labels change

### LINKED ISSUE #4278: Delay in SecurityPolicy change propagation for HTTPRoute when using targetSelectors
I am facing a delay in propagation of security policy when I change from one security policy to another security policy in a situation when security policy is applied to http-routes using labels.

I have tried few things to narrow down the cases in which it is happening:
Suppose i have following resources : 
- `sp-1` having targetSelectors for HTTPRoute as `sp: sp-1`
- `sp-2` having targetSelectors for HTTPRoute as `sp: sp-2`
- HTTPRoute : `route-1`

Then for above:
- If `route-1` has  `sp-1` applied via labels: `sp: sp-1` and now if I change the labels to `sp: sp-2` (to apply sp-2 to route-1), then it takes considerable amount of time for this change to propagate. I am verifying if changes are propagated or not via `egctl c envoy-proxy route`
- if `route-1` has `sp-1` applied via labels: `sp: sp-1`, and if i make some change in `sp-1` itself and apply, the policy change is immediately propagated in route as well.
- if `route-1` has `sp-1` applied via targetRefs for `route-1` in `sp-1`, then making change in policy or shifting targetRefs to `sp-2`, both changes immediately propagate.

So basically only in the first case, where policy is applied to a route via targetSelectors and we change from one predefined security policy to another predefined one, i see a delay in propagation and delay can be of minutes also. I want to know if this delay is expected and if yes is there a way to reduce it. 

This is reproducible on my local using latest dev eg version.

luvk1412:

Fixes #4278

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1105,6 +1105,7 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	// Watch HTTPRoute CRUDs and process affected Gateways.
 	httprPredicates := []predicate.TypedPredicate[*gwapiv1.HTTPRoute]{
 		predicate.TypedGenerationChangedPredicate[*gwapiv1.HTTPRoute]{},
+		predicate.TypedLabelChangedPredicate[*gwapiv1.HTTPRoute]{},
 	}
 	if r.namespaceLabel != nil {
 		httprPredicates = append(httprPredicates, predicate.NewTypedPredicateFuncs(func(hr *gwapiv1.HTTPRoute) bool {
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1080,8 +1080,8 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 
 	// Watch Gateway CRUDs and reconcile affected GatewayClass.
 	gPredicates := []predicate.TypedPredicate[*gwapiv1.Gateway]{
-		predicate.TypedGenerationChangedPredicate[*gwapiv1.Gateway]{},
-		predicate.TypedLabelChangedPredicate[*gwapiv1.Gateway]{},
+		predicate.Or(predicate.TypedGenerationChangedPredicate[*gwapiv1.Gateway]{},
+			predicate.TypedLabelChangedPredicate[*gwapiv1.Gateway]{}),
 		predicate.NewTypedPredicateFuncs(func(gtw *gwapiv1.Gateway) bool {
 			return r.validateGatewayForReconcile(gtw)
 		}),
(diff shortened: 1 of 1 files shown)


zirain (dismissed):

zirain:
```
    --- FAIL: TestGatewayAPIConformance/HTTPRouteObservedGenerationBump (60.06s)
        --- FAIL: TestGatewayAPIConformance/HTTPRouteObservedGenerationBump/observedGeneration_should_increment (60.04s)
```

internal/provider/kubernetes/controller.go
```diff
@@ -1105,6 +1105,7 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 	// Watch HTTPRoute CRUDs and process affected Gateways.
 	httprPredicates := []predicate.TypedPredicate[*gwapiv1.HTTPRoute]{
 		predicate.TypedGenerationChangedPredicate[*gwapiv1.HTTPRoute]{},
+		predicate.TypedLabelChangedPredicate[*gwapiv1.HTTPRoute]{},
```
arkodg, line 1108: can we add this for all xRoutes and Gateway as well ?

luvk1412:
> ```
>     --- FAIL: TestGatewayAPIConformance/HTTPRouteObservedGenerationBump (60.06s)
>         --- FAIL: TestGatewayAPIConformance/HTTPRouteObservedGenerationBump/observedGeneration_should_increment (60.04s)
> ```

@zirain i tried running conformance tests on local using `make conformance`, but they didn't run properly on my local it seems i saw a lot more errors. Also looked into similar issues in istio and cillium, but couldn't get what could have went wrong here.

internal/provider/kubernetes/controller.go
luvk1412: added @arkodg

zirain:
yeah, I believe you fix is good, but want to know why test failed.

arkodg:
@luvk1412 here's how I run conformance tests
1. Setup a k8s cluster (I use docker desktop k8s)
2. `make kube-deploy`
3. `make run-conformance`

we probably need to peek into the envoy gateway deployment logs to see if the issue is the gateway api runner is not getting  the message or is the status update ignoring the status update

internal/provider/kubernetes/controller.go
```diff
@@ -1105,6 +1106,7 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
```
zirain, line 1109:
```suggestion
		predicate.Or(predicate.TypedGenerationChangedPredicate[*gwapiv1.HTTPRoute]{},
			predicate.TypedLabelChangedPredicate[*gwapiv1.HTTPRoute]{}),
```

I think this's the right way to use this, see

https://github.com/kubernetes-sigs/controller-runtime/blob/a2e9eb74bc3d7f180fdda837408b4ee5841cd807/pkg/predicate/predicate.go#L255-L267

luvk1412: i have made these changes but not sure why the earlier implementation would fail conformance, i am still not able to run conformance on my local, so don't know if current one is passing. My understanding was the multiple entries instead of using `.predicateOr` would just result in multiple enqueueClass(though could be redundant) and with using would result in enqueueClass, but not sure why that should fail api conformance.

luvk1412:
conformance looks to be pasing now but not sure why multiple enQueue(which would happen on mentioning multiple predicates instead of using or) would fail `TestGatewayAPIConformance/HTTPRouteObservedGenerationBump/observedGeneration_should_increment` 🤔

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #2422, opened 2024-01-09
E2E for Downstream TLS Settings


arkodg:
*Description*:
>Describe the issue.

Relates to https://github.com/envoyproxy/gateway/pull/2297

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
@zesiar0 can you comment here, so we can assign this to you ?

zesiar0:
@arkodg please assign to me

arkodg:
@zesiar0 still planning on working on this ? (this issue is a release blocker for v1.0.0 releasing next week), if not, I can pick this one up, thanks !

zesiar0:
> @zesiar0 still planning on working on this ? (this issue is a release blocker for v1.0.0 releasing next week), if not, I can pick this one up, thanks !

Sorry, I will finish it today

Issue #2421, opened 2024-01-09
docs: Downstream TLS Settings


arkodg:
*Description*:
>Describe the issue.

Relates to https://github.com/envoyproxy/gateway/pull/2297 & https://github.com/envoyproxy/gateway/issues/2286

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

shawnh2:
thanks @zesiar0

arkodg:
@zesiar0 can you comment here so we can assign this to you ?

zesiar0:
> @zesiar0 can you comment here so we can assign this to you ?

sorry, I mistake the issue. Please this issue to me

PR #4293, opened 2024-09-20
feat(translator): client tls session resumption

### LINKED ISSUE #4268: Support Session Resumption settings
*Description*:
TLS Session in envoy can be resumed using stateful (session id) and/or stateless (session ticket) approaches. 

Envoy Proxy docs on TLS mention the importance of session resumption for performance and security in front proxy scenarios:

> Session resumption: Server connections support resuming previous sessions via TLS session tickets (see [RFC 5077](https://www.ietf.org/rfc/rfc5077.txt)). Resumption can be performed across hot restarts and between parallel Envoy instances (typically useful in a front proxy configuration).

The default Envoy Gateway settings for session resumption are sub-optimal from a performance perspective and can have some security implications. 

Since Envoy Gateway typically deploys multiple Envoy Proxy instances, in-memory session id caches and session ticket encryption keys are not synchronized. As a result, a TLS session would only be resumed if a new connection is established against the same Envoy Proxy Instance. 

From a security perspective, envoy users previously requested the option to disable both stateful and stateless session resumption, due to various risks associated with resumption.  

Envoy Gateway could: 
- Allow users to opt-out of session resumption for security purposes
- Allow users to synchronize session ticket encryption keys across instances to improve the chance of successful resumption
- Allow users to control session ticket encryption key LCM to meet security requirements (e.g. rotation frequency). 

This can be achieve by extending the TLS parameter support in Client Traffic Policy.

[optional *Relevant Links*:]
- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#tls
- https://github.com/envoyproxy/envoy/issues/9951
- https://github.com/envoyproxy/envoy/pull/28758

guydc:
**What type of PR is this?**

Introduces an API for TLS session management:
- Disable TLS resumption by default. Allow users to opt-in.
- Adds E2E Tests and Docs for ClientTrafficPolicy TLS settings

Fixes #4268, #2422, #2421


api/v1alpha1/tls_types.go
```diff
@@ -15,6 +15,20 @@ type ClientTLSSettings struct {
 	// +optional
 	ClientValidation *ClientValidationContext `json:"clientValidation,omitempty"`
 	TLSSettings      `json:",inline"`
+
```
zhaohuabing, line 18: Could we explain here what's the default strategy used by Envoy/EG? No TLS resumption, Stateless TLS resumption, or Stateful TLS resumption?

zhaohuabing: Would we add an intermediate layer `SessionResumption` here to group all the session related parameters?

api/v1alpha1/tls_types.go
```diff
@@ -133,3 +147,36 @@ type ClientValidationContext struct {
...
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// SessionTicketKeysRef is a reference to the secret containing the keys to use for encryption and decryption
+	// of session tickets. The first keys is used for encryption and all keys are candidate for decryption.
+	// Users are responsible for securely generating, storing and rotating keys.
+	// If empty, each Envoy Proxy instance will use local in-memory keys.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlssessionticketkeys
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	SessionTicketKeysRef []gwapiv1.SecretObjectReference `json:"sessionTicketKeysRef,omitempty"`
```
zhaohuabing, line 170: Should we mention the risks of the shared ticket secret and suggestions to mitigate the risks? Like Envoy does: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-tlssessionticketkeys

api/v1alpha1/tls_types.go
```diff
@@ -133,3 +147,36 @@ type ClientValidationContext struct {
...
+// +kubebuilder:validation:XValidation:rule="(self.enabled == false && self.sessionTicketKeysRef.size() > 0)",message="sessionTicketKeysRef can only be set when enabled is true"
+type StatelessSessionResumptionSettings struct {
+	// Enabled determines if stateless session resumption is supported. When Envoy Proxy is deployed with more than
+	// one replica, session ticket encryption keys are not synchronized between instances, possibly leading to resumption
+	// failures. Users can synchronize session ticket encryption keys be storing them in a secret and using the
+	// sessionTicketKeysRef option.
+	// Note that improper handling of session ticket encryption keys may result in loss of secrecy.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-tlssessionticketkeys
+	// Default: true
+	Enabled *bool `json:"enabled,omitempty"`
```
zhaohuabing, line 161: If we add an intermediate layer, `Enabled` can be removed.

```
SessionResumption:
  Type: Stateless
  keys:
  - ...
```

guydc:
Following yesterday's community meeting, some additional industry context. 

Currently, Mozilla's SSL config generator recommends disabling stateless resumption (session tickets) for nginx, apache and haproxy for security reasons. 
- https://ssl-config.mozilla.org/#server=apache&version=2.4.41&config=modern&openssl=1.1.1k&guideline=5.7
- https://ssl-config.mozilla.org/#server=haproxy&version=2.1&config=modern&openssl=1.1.1k&guideline=5.7
- https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=modern&openssl=1.1.1k&guideline=5.7

This is mostly due to insufficient rotation of session ticket encryption keys in these projects. There is an ongoing discussion on changing the recommendation for newer versions of nginx where keys are rotated: https://github.com/mozilla/server-side-tls/issues/135. 

nginx ingress by default disables stateful and stateless session resumption, but makes it possible to opt-in:
- https://github.com/kubernetes/ingress-nginx/blob/a9c9a9d51e48f50edd5e8e3084a5c485366ba6c4/rootfs/etc/nginx/template/nginx.tmpl#L426
- https://github.com/kubernetes/ingress-nginx/blob/a9c9a9d51e48f50edd5e8e3084a5c485366ba6c4/rootfs/etc/nginx/template/nginx.tmpl#L432

BoringSSL rotates encryption keys by default every 48 hours: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Session-tickets. Envoy doesn't change this behavior or make the rotation schedule configurable. Industry leaders like cloudflare rotate session ticket encryption keys every hour: https://blog.cloudflare.com/keyless-ssl-the-nitty-gritty-technical-details/. 

There is no security recommendation to disable stateful session resumption. However:
- Stateful resumption is no longer supported in TLSv1.3: https://www.wolfssl.com/tls-1-3-performance-resumption/ 
- Envoy has a well-known issue that can only be mitigated by disabling stateful session resumption: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routematch-tlscontextmatchoptions
- There is no way to synchronize session caches across proxy instances. The effort to support synchronization is significant: https://github.com/envoyproxy/envoy/issues/14553 

I propose that we change the current behavior and disable both resumption methods by default, but allow users to opt-in similar to nginx ingress.

api/v1alpha1/tls_types.go
guydc: Good input, I adopted this format.

api/v1alpha1/tls_types.go
guydc: after discussion yesterday, I reduced the scope of this PR to only deal with enable/disable of resumption.

api/v1alpha1/tls_types.go
guydc: Added a list of "typed" resumption settings and included docs explaining the default EG behavior.

api/v1alpha1/tls_types.go
zhaohuabing: Rethink on this,  a fixed list of 2 elements seems a bit strange, what about the following structure? cc @arkodg always has great idea on how to shape the API.

 ```yaml
 SessionResumptionSettings:
    stateful: {}
    stateless: {}
```

File: test/e2e/testdata/client-mtls.yaml [hunk 4]
```diff
@@ -69,3 +79,56 @@ spec:
         - kind: "Secret"
           group: ""
           name: "client-mtls-certificate"
+---
```
zirain, line 82: need to record how this is generated for future maintenance.

api/v1alpha1/tls_types.go
```diff
@@ -15,6 +15,20 @@ type ClientTLSSettings struct {
...
+	// +optional
+	SessionTimeout *gwapiv1.Duration `json:"sessionTimeout,omitempty"`
+
+	// SessionResumptionSettings determine the proxy's supported TLS session resumption option.
+	// By default, Envoy Gateway does not enable session resumption. Use sessionResumption to
+	// enable stateful and stateless session resumption. Users should consider security impacts
+	// of different resumption methods. Performance gains from resumption are diminished when
+	// Envoy proxy is deployed with more than one replica.
+	// +optional
+	SessionResumptionSettings *SessionResumptionSettings `json:"sessionResumption,omitempty"`
```
arkodg, line 31: prefer `SessionResumption`

api/v1alpha1/tls_types.go
```diff
@@ -15,6 +15,20 @@ type ClientTLSSettings struct {
 	// +optional
 	ClientValidation *ClientValidationContext `json:"clientValidation,omitempty"`
 	TLSSettings      `json:",inline"`
+
+	// SessionTimeout determines the maximum lifetime of a TLS session.
+	// https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_DEFAULT_SESSION_TIMEOUT
+	// Default: 7200s
+	// +optional
+	SessionTimeout *gwapiv1.Duration `json:"sessionTimeout,omitempty"`
```
arkodg, line 23: should this live inside `SessionResumption` ?

api/v1alpha1/tls_types.go
```diff
@@ -15,6 +15,20 @@ type ClientTLSSettings struct {
...
+
+	// SessionResumptionSettings determine the proxy's supported TLS session resumption option.
+	// By default, Envoy Gateway does not enable session resumption. Use sessionResumption to
+	// enable stateful and stateless session resumption. Users should consider security impacts
+	// of different resumption methods. Performance gains from resumption are diminished when
+	// Envoy proxy is deployed with more than one replica.
+	// +optional
+	SessionResumptionSettings *SessionResumption `json:"sessionResumption,omitempty"`
```
zirain, line 31: can we use following structure?
```
session:
  timeout:
  resumption:
```

if there're more than 1 property with session prefix, prefer to move them into a struct.

zhaohuabing: Should the `timeout` be part of the SessionResumption setting? 

* https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-downstreamtlscontext
* https://datatracker.ietf.org/doc/html/rfc5077#section-5.6

guydc: Well, it seems that different implementations have different meanings for session timeout. 

[BoringSSL](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_DEFAULT_SESSION_TIMEOUT) seems to treat this as a hard timeout after which a new handshake is required:
>  This is how long we are willing to use the secret to encrypt traffic without fresh key material.

In envoy, this is also used as a hint on session tickets. 

[OpenSSL](https://docs.openssl.org/master/man3/SSL_get_default_timeout) seems to treat this as a soft timeout after which resumption cannot occur, but an active session may still continue (?):
> Whenever a new session is negotiated, it is assigned a timeout value, after which it will not be accepted for session reuse. 

@arkodg requested that we leave `sessionTimeout` out of this PR. I will change the API to @zirain's proposal, to support for future addition of session-related settings that are not specific to resumption.

api/v1alpha1/tls_types.go
```diff
@@ -15,6 +15,10 @@ type ClientTLSSettings struct {
 	// +optional
 	ClientValidation *ClientValidationContext `json:"clientValidation,omitempty"`
 	TLSSettings      `json:",inline"`
+
+	// Session defines setting related to TLS session management.
```
arkodg, line 19:
```suggestion
	// Session defines settings related to TLS session management.
```

api/v1alpha1/tls_types.go
```diff
@@ -133,3 +137,40 @@ type ClientValidationContext struct {
 	// +optional
 	CACertificateRefs []gwapiv1.SecretObjectReference `json:"caCertificateRefs,omitempty"`
 }
+
+// Session defines setting related to TLS session management.
+type Session struct {
+	// Resumption determine the proxy's supported TLS session resumption option.
```
arkodg, line 143:
```suggestion
	// Resumption determines the proxy's supported TLS session resumption option.
```

zirain (approved):
require user-faced doc.

zhaohuabing (approved):
LGTM, Thanks.

PR #4298, opened 2024-09-21
Datadog tracing support

Hartigan:
Hi,

This pull request allows to configure Datadog tracing for envoy proxy via CRDs.

### DIFF
--- a/internal/xds/translator/tracing.go
+++ b/internal/xds/translator/tracing.go
@@ -28,6 +28,7 @@ import (
 const (
 	envoyOpenTelemetry = "envoy.tracers.opentelemetry"
 	envoyZipkin        = "envoy.traces.zipkin"
+	envoyDatadog       = "envoy.tracers.datadog"
 )
 
 type typConfigGen func() (*anypb.Any, error)
@@ -41,6 +42,16 @@ func buildHCMTracing(tracing *ir.Tracing) (*hcm.HttpConnectionManager_Tracing, e
 	var providerConfig typConfigGen
 
 	switch tracing.Provider.Type {
+	case egv1a1.TracingProviderTypeDatadog:
+		providerName = envoyDatadog
+
+		providerConfig = func() (*anypb.Any, error) {
+			config := &tracecfg.DatadogConfig{
+				ServiceName:      tracing.ServiceName,
+				CollectorCluster: tracing.Destination.Name,
+			}
+			return protocov.ToAnyWithError(config)
+		}
 	case egv1a1.TracingProviderTypeOpenTelemetry:
 		providerName = envoyOpenTelemetry
 
--- a/api/v1alpha1/tracing_types.go
+++ b/api/v1alpha1/tracing_types.go
@@ -26,6 +26,7 @@ type TracingProviderType string
 const (
 	TracingProviderTypeOpenTelemetry TracingProviderType = "OpenTelemetry"
 	TracingProviderTypeZipkin        TracingProviderType = "Zipkin"
+	TracingProviderTypeDatadog       TracingProviderType = "Datadog"
 )
 
 // TracingProvider defines the tracing provider configuration.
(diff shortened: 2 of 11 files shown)


arkodg:
@Hartigan would be ideal if we can avoid a vendor specific field, can datadog tracing be supported via Otel ?

Hartigan:
Hi @arkodg 

I didn't add new specific field, I just add value to enum:

```
telemetry:
  tracing:
    samplingRate: 100
    provider:
      type: Datadog
      backendRefs:
      - name: datadog
        namespace: datadog
        port: 8126
```

For configuration it uses the same fields as opentelemetry. Not smth like https://gateway.envoyproxy.io/docs/api/extension_types/#zipkintracingprovider

I guess, to enable datadog tracing, we need to enable a specific plugin for envoy proxy. Therefore, there should be a way to describe the plugin name via yaml.

I made this request because envoy proxy has integration with datadog: https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/datadog-tracing

Ofc, it can be configured in another way with otel collector and specific exporter: https://docs.datadoghq.com/opentelemetry/collector_exporter/

Thanks

arkodg:
this is probably the first feature that is vendor specific, and we dont have any precedence for this yet
@zhaohuabing @guydc and I discussed this yesterday and agreed that its okay to add a vendor specific feature if its already part of EnvoyProxy, looking for  👍 / 👎  from the rest of @envoyproxy/gateway-maintainers before we proceed with this PR

AliceProxy:
Datadog is massive in the industry, we should make supporting it natively as easy as possible, especially if we already have people contributing the changes for us. If it weren't supported by Envoy Proxy and we had to write a custom service to add that native support, then I could understand being a little more on-the-fence about vendor specific inclusions in the API.

I think in general I do prefer APIs that are vendor-agnostic and support all of the options you need to configure whatever specific vendor you are using, but in instances like this where more specific handling is available then we should support it rather than trying to force people to use otel as a middle layer if that experience is not just as good and easy to use as native Datadog support.

internal/xds/translator/tracing.go
```diff
@@ -41,6 +42,16 @@ func buildHCMTracing(tracing *ir.Tracing) (*hcm.HttpConnectionManager_Tracing, e
...
 
 	switch tracing.Provider.Type {
+	case egv1a1.TracingProviderTypeDatadog:
+		providerName = envoyDatadog
+
+		providerConfig = func() (*anypb.Any, error) {
+			config := &tracecfg.DatadogConfig{
+				ServiceName:      tracing.ServiceName,
+				CollectorCluster: tracing.Destination.Name,
+			}
```
arkodg, line 52: is `CollectorHostname` also needed ?

arkodg:
thanks @AliceProxy, looks like we have a majority now, removed the decision label

hey @Hartigan thanks for being patient, this PR should be ready for a review for, can you run `make manifests` && `make generate` and commit those changes ?
can you also add a test case in https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata ?

Hartigan:
@arkodg I will try to find time to add tests today/tomorrow

internal/xds/translator/tracing.go
Hartigan: As I understand, `CollectorHostname` has default value which relates to `CollectorCluster`: https://github.com/envoyproxy/envoy/blob/b203e3e826af06e3639229f816f2a11245476728/api/envoy/config/trace/v3/datadog.proto#L40-L42.  Also I didn't find similar fields for `OpenTelemetry` and `Zipkin`

arkodg: ok, have you been able to manually test this, and does it work ?

Hartigan: I see traces in Datadog APM from Envoy Proxy deployed with this gateway, and I found code which relates to this logic: https://github.com/envoyproxy/envoy/blob/39851cddac60be79f58ab33eafa35a1c9dda81a3/source/extensions/tracers/datadog/config.cc#L51-L58

Or I misunderstood the question and need somehow write a unit test for this

arkodg (approved):
LGTM thanks !

zirain (approved):

zirain:
it would be nice if you can add e2e and doc.

zirain:
@Hartigan can you add e2e and doc for this?

Hartigan:
Hi @zirain 

Yes, I will try to find time on this week to create pull request with documentation and e2e tests

Thanks

Issue #3209, opened 2024-04-16
enhancement: apply `EnvoyGateway`  without manually restarting the EG deployment 


zhaohuabing:
*Description*:
Currently, the EG deployment must be manually restarted after updating the ConfigMap containing `EnvoyGageway` configuration. 

To streamline the process, we could implement a mechanism where EG automatically detects the CM changes and restart itself.  

To achieve this, EG can either:

* Use a timestamp annotation in EG deployment template to trigger a restart.
* Reload the config and restart all runners without restarting itself.

ShyunnY:
We can implement a watcher through informer to help us hot refresh the controller configuration.

This looks like a great feature and I'm looking forward to it.

shawnh2:
+1 for this one

xref: #864

zirain:
Will filewatcher help this? I recall there's same requirement in istio for `istio.istio-system` configmap.

zirain:
On the otherhand, a filewatch require a long time to catch a event, informer is a better choice?

arkodg:
+1 to a non k8s specific way to solve this e.g. using https://github.com/fsnotify/fsnotify

zirain:
I recall there's some issue when using fsnotiry in kubernetes, so istio use a wrapper https://github.com/istio/pkg/blob/master/filewatcher/filewatcher.go

ShyunnY:
I think that in most scenarios where the gateway configuration needs to be refreshed, we may pay more attention to real-time performance. Can we accept the delay caused by fsnotiry?

shawnh2:
IMO, for Kubernetes provider, the Informer would be much accurate and fast than watching fsnotify event. According to the [ConfigMaps update strategy](https://kubernetes.io/docs/concepts/configuration/configmap/#mounted-configmaps-are-updated-automatically), the potential delay of automatically update mounted ConfigMaps (which basically imply using the fsnotify approach) lies in kubelet sync period and cache propagation.

The fsnotify could also work, but this approach may be very suitable for the File provider.

- If we seeking code generality, the fsnotify approach works for both.
- If we seeking the accuracy and realtime, the Informer for k8s and fsnotify for file provider are suggested.

arkodg:
if the `fsnotify` approach enables the changes under 5s, its good enough, imo

zirain:
> if the `fsnotify` approach enables the changes under 5s, its good enough, imo

I recall it may take 30s in the worst.

PR #4299, opened 2024-09-22
infra: support reloadable EnvoyGateway configuration

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3209


internal/envoygateway/config/runner/runner.go
```diff
@@ -0,0 +1,87 @@
...
+	return runnerName
+}
+
+func (r *Runner) Start(ctx context.Context) error {
+	if r.cfgPath == "" {
+		r.logger.Info("no config file provided, skipping config watcher")
+		return nil
+	}
+
+	watcher, err := fsnotify.NewWatcher()
```
shawnh2, line 43: Is possible to reuse the notifier under `internal/provider/file/notifier.go` ?

zirain: still WIP, will try it later.

zirain: IMO `internal/provider/file/notifier.go` had not user-firendly API, I will try to use lib from hugo in this PR, cause it's not the key point.

zirain:
failed because [ExtensionAPISettings](https://github.com/envoyproxy/gateway/blob/f5693f6bc1bba4d770e0fea9d9e2b64617f2140b/api/v1alpha1/envoygateway_types.go#L165) didn't updated. It's able to add new watcher after `EnvoyGateway` updated, but cannot remove watcher after `EnvoyGateway` updated.

internal/common/runner.go
```diff
@@ -0,0 +1,21 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package common
```
arkodg, line 6: prefer `runner` package under `internal/runner`

internal/envoygateway/config/loader/Loader.go
```diff
@@ -0,0 +1,109 @@
...
+		defer w.Close()
+		for {
+			select {
+			case events, ok := <-w.Events:
+				if !ok {
+					return
+				}
+
+				for _, e := range events {
+					r.logger.Info("receive fsnotify events", "name", e.Name, "op", e.Op.String())
```
arkodg, line 71:
```suggestion
					r.logger.Info("received fsnotify events", "name", e.Name, "op", e.Op.String())
```

internal/envoygateway/config/loader/Loader.go
```diff
@@ -0,0 +1,109 @@
...
+		r.logger.Info("no config file provided, skipping config watcher")
+		return nil
+	}
+
+	w, err := watcher.New(500*time.Millisecond, 500*time.Millisecond, false)
+	if err != nil {
+		r.logger.Error(err, "failed to create fsnotify watcher")
+		return err
+	}
+	r.logger.Info("watch file changes", "path", r.cfgPath)
```
arkodg, line 54:
```suggestion
	r.logger.Info("watching for changes to the EnvoyGateway configuration", "path", r.cfgPath)
```

internal/envoygateway/config/loader/Loader.go
```diff
@@ -0,0 +1,109 @@
...
+				}
+
+				for _, e := range events {
+					r.logger.Info("receive fsnotify events", "name", e.Name, "op", e.Op.String())
+				}
+
+				// Load the config file.
+				eg, err := config.Decode(r.cfgPath)
+				if err != nil {
+					r.logger.Error(err, "failed to decode config file", "name", r.cfgPath)
```
arkodg, line 77: should we silently fail this ? imo we should follow the same flow we do during startup

internal/gatewayapi/runner/runner.go
```diff
@@ -43,17 +46,21 @@ const (
 	serveTLSCaFilename   = "/certs/ca.crt"
 )
 
+var _ common.Runner = &Runner{}
+
 type Config struct {
-	config.Server
+	ServerCfg         *config.Server
```
arkodg, line 52: curious why :) ?

internal/gatewayapi/runner/runner.go
```diff
@@ -120,7 +138,7 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 			r.Logger.Info("received an update")
 			val := update.Value
 			// There is only 1 key which is the controller name
-			// so when a delete is triggered, delete all IR keys
+			// so when delete is triggered, delete all IR keys
```
arkodg, line 141: the prev grammar was correct, lets undo this

internal/gatewayapi/runner/runner.go
zirain: Using pointers instead of value copies?

internal/provider/kubernetes/kubernetes.go
```diff
@@ -110,7 +111,7 @@ func New(cfg *rest.Config, svr *ec.Server, resources *message.ProviderResources)
 	// Emit elected & continue with deployment of infra resources
 	go func() {
 		<-mgr.Elected()
-		close(svr.Elected)
```
arkodg, line 113: logic is changing here ...

internal/infrastructure/kubernetes/ratelimit_infra.go
```diff
@@ -52,10 +52,6 @@ func (i *Infra) CreateOrUpdateRateLimitInfra(ctx context.Context) error {
 
 // DeleteRateLimitInfra removes the managed kube infra, if it doesn't exist.
 func (i *Infra) DeleteRateLimitInfra(ctx context.Context) error {
-	if err := ratelimit.Validate(ctx, i.Client.Client, i.EnvoyGateway, i.Namespace); err != nil {
```
arkodg, line 55: curious why this was needed

internal/provider/kubernetes/kubernetes.go
zirain: a channel can only be closed once

arkodg: sg, do we have a leader election test ?
cc @alexwo

zirain: tracked in https://github.com/envoyproxy/gateway/issues/4420

internal/infrastructure/kubernetes/ratelimit_infra.go
zirain: will be fixed in https://github.com/envoyproxy/gateway/pull/4430

internal/xds/server/runner/runner.go
```diff
@@ -82,29 +84,37 @@ func (r *Runner) Start(ctx context.Context) (err error) {
...
-	r.Logger.Info("started")
+	r.logger.Info("started")
 	return
 }
 
+func (r *Runner) Reload(serverCfg *config.Server) error {
+	r.logger = serverCfg.Logger.WithName("xds-server-runner")
+	r.logger.Info("reloaded")
+
+	// TODO: Implement reload logic
```
zhaohuabing, line 103: Hi @zirain Is this PR still a draft?

zirain: nit, should be removed.

zhaohuabing:
Dealing with configuration reload for each runner can be cumbersome, especially as the `EnvoyGateway` may introduce additional settings in the future. Would it be more straightforward to restart all runners instead of reloading the configuration individually?

zirain:
close this first in favor of https://github.com/envoyproxy/gateway/pull/4451, maybe came back later.

PR #4313, opened 2024-09-23
set user group and user id for the default SecurityContext

zhaohuabing:
This PR sets the user group and user id for the default SecurityContext to harden EG deployment. This is recommended by some security frameworks and usually checked by security scanning tools.

Reference: https://hub.armosec.io/docs/c-0013
Related: https://github.com/envoyproxy/gateway/pull/3940


arkodg:
isnt this already baked into the image / distroless ?

zhaohuabing:
> isnt this already baked into the image / distroless ?

Yes, this PR just explicitly set the userID and GroupID in the container SecurityContext. This is recommended by some security frameworks and is checked by security scanning tools, eg,  https://hub.armosec.io/docs/c-0013

arkodg (approved):

Xunzhuo (commented):
lgtm

PR #4314, opened 2024-09-23
feat: support request response sizes stats

### LINKED ISSUE #4234: Support for enabling Request Response Size statistics
I want to have plots of [Request Response Size statistics](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html#request-response-size-statistics) in my eg setup which requires me to enable `request_response_sizes` in [config.cluster.v3.TrackClusterStats](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-trackclusterstats). 

In my knowledge there is currently no api in eg to do so(please correct me if am wrong) nor it is practically feasible to enable this using `EnvoyPatchPolicy` as we have lots of clusters and there is no way to select them all based on regex or something else, and adding for each cluster one by one is not feasible. 

Request to add support for this. `BackendTrafficPolicy` seems a good place to do so to me maybe ?

luvk1412:

Fixes #4234


api/v1alpha1/envoyproxy_metric_types.go
```diff
@@ -32,6 +32,11 @@ type ProxyMetrics struct {
 	// EnablePerEndpointStats enables per endpoint envoy stats metrics.
 	// Please use with caution.
 	EnablePerEndpointStats bool `json:"enablePerEndpointStats,omitempty"`
+
+	// EnableRequestResponseSizesStats enables publishing of histograms tracking header and body sizes of requests and responses.
+	//
+	// +optional
+	EnableRequestResponseSizesStats bool `json:"enableRequestResponseSizesStats,omitempty"`
```
zhaohuabing, line 39: How about `EnableBodySizesStats` ?

zhaohuabing, line 39: Nit: suggest using *bool for optional field.  EnablePerEndpointStats can also be changed to pointer. It's back-compatible.

Additionally (not in this PR), We should consider standardizing the optional bool in the APIs. While most of them are  already using pointer,  there are several that are not.

Xunzhuo: We can make `EnableRequestResponseSizesStats` as a bool pointer type, and raise another issue to discuss if we standarlize the other bool vars in APIs

luvk1412: I also thought `EnablePerEndpointStats` should be `* bool`, wrote it as `bool` to maintain consistency.

Now i have changed `EnableVirtualHostStats`, `EnablePerEndpointStats` and `EnableRequestResponseSizesStats` to `*bool` in `ProxyMetrics`.  `EnableVirtualHostStats` from usages looked like optional only to me, let me know if this needs to be changed back.

Ideally `ir.Metrics` could also have `*bool` but not changing here as i felt could result in unnecessary refactor for this issue, let me know if that needs to be changed.

Also wanted to add a function like
```
func boolValue(b *bool) bool {
    if b == nil {
        return false
    }
    return *b
}
```
ideally to extract bool, but wasn't sure where should i add it, (in same file or is there common utils used across project), so for now haven't added this.

luvk1412:
`    --- FAIL: TestGatewayAPIConformance/HTTPRouteSimpleSameNamespace (60.01s)`
Not sure if this is flaky or not, retest didn't re run conformance test. Any of my change is not related to gw api, so doubt its due to that.

zhaohuabing (dismissed):
LGTM Thanks!

zirain:
can you also add a positive use case?

luvk1412:
@zirain just for clarification you want me to add a new file in `internal/gatewayapi/testdata` right ? I ideally wanted to add a test where generated cluster is validated from input yaml, but i think that's not the case ig for tests in `internal/gatewayapi/testdata` ? and i can't find in project where such tests exist if they exist.

zirain:
> @zirain just for clarification you want me to add a new file in `internal/gatewayapi/testdata` right ? I ideally wanted to add a test where generated cluster is validated from input yaml, but i think that's not the case ig for tests in `internal/gatewayapi/testdata` ? and i can't find in project where such tests exist if they exist.
> 
> Edit: added a metric-enabled input file in `internal/gatewayapi/testdata` for now if that is what is required.

can you add xds translator test case

zirain (approved):
Thanks for adding this.

Would be better if you can add e2e and doc in the next.

arkodg:
would be good to also enhance the grafana dashboard in `gateway-addons-helm` to make it easier to consume these metrics

luvk1412:
> would be good to also enhance the grafana dashboard in `gateway-addons-helm` to make it easier to consume these metrics

you wan it to be done in this pr or i can open a separate issue for that @arkodg  ?

arkodg:
follow up is fine @luvk1412 , thanks

zhaohuabing (approved):

Issue #4022, opened 2024-08-09
docs: use kubectl patch with `--type=merge` in docs 


kranurag7:
*Description*:

I was trying out the following docs here https://gateway.envoyproxy.io/docs/tasks/security/tls-cert-manager/#creating-a-tls-gateway-listener
the docs as of now suggests the following command for patch using --patch-file flag. This is not working out for me somehow. (I'm using bash shell)
```bash
$ kubectl patch gateway eg --patch-file=- <<EOF
metadata:
  annotations:
    cert-manager.io/cluster-issuer: selfsigned
    cert-manager.io/common-name: "Hello World!"
spec:
  listeners:
  - name: https
    protocol: HTTPS
    hostname: www.example.com
    port: 443
    tls:
      mode: Terminate
      certificateRefs:
      - kind: Secret
        name: eg-https
EOF
error: unable to read patch file: open -: no such file or directory
$ kubectl version
Client Version: v1.30.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.2
```

I updated the command and used the following:
```bash
$ kubectl patch gateway eg --patch '
metadata:
  annotations:
    cert-manager.io/cluster-issuer: selfsigned
    cert-manager.io/common-name: "Hello World!"
spec:
  listeners:
  - name: https
    protocol: HTTPS
    hostname: www.example.com
    port: 443
    tls:
      mode: Terminate
      certificateRefs:
      - kind: Secret
        name: eg-https
' --type=merge
gateway.gateway.networking.k8s.io/eg patched
```
This worked good for me. Is there something I'm missing in the first case which is documented as of now. 

[optional *Relevant Links*:]

slack discussion: https://envoyproxy.slack.com/archives/C03E6NHLESV/p1723037137777729

arkodg:
he @kranurag7 thanks for surfacing this issue, would be great if you can fix it in the docs

PR #4316, opened 2024-09-23
docs: fix patch command for adding TLS gateway listener

Dean-Coakley:

Fixes https://github.com/envoyproxy/gateway/issues/4022


arkodg (approved):
thanks !

zirain (approved):

Issue #3980, opened 2024-07-31
Consider combining the xds-translator and the xds-server runners into 1 runner


arkodg:
*Description*:
>Describe the issue.

* The xds-translator runners subscribes to Xds IR and translates it to Xds https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/runner/runner.go 
* The xds-server runner subscribes to Xds and services Xds requests from Envoy clients https://github.com/envoyproxy/gateway/blob/main/internal/xds/server/runner/runner.go 

these two runners can be combined into 1 to reduce the extra layer of Xds memory saved in watchable https://github.com/envoyproxy/gateway/blob/847d554a7f60018d6a608fe1469c6016e6292a47/internal/message/types.go#L126

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
wdyt @envoyproxy/gateway-maintainers ?

zhaohuabing:
@arkodg the intention is to save the memory for the xds as we don't have to publish and subscript to the generated xds? Do you have an estimated memory we can save by merging these two runners?

qicz:
IMO needs to split the resources too.  some resources for subscription like gateway API resources and EG APIs, but others are just kept in memory. like follows:

for subscription:

```golang
type WatchableResources struct {
	// This field is only used for marshalling/unmarshalling purposes and is not used by
	// the translator

	// EnvoyProxyForGatewayClass holds EnvoyProxy attached to GatewayClass
	EnvoyProxyForGatewayClass *egv1a1.EnvoyProxy `json:"envoyProxyForGatewayClass,omitempty" yaml:"envoyProxyForGatewayClass,omitempty"`
	// EnvoyProxiesForGateways holds EnvoyProxiesForGateways attached to Gateways
	EnvoyProxiesForGateways []*egv1a1.EnvoyProxy `json:"envoyProxiesForGateways,omitempty" yaml:"envoyProxiesForGateways,omitempty"`

	GatewayClass            *gwapiv1.GatewayClass          `json:"gatewayClass,omitempty" yaml:"gatewayClass,omitempty"`
	Gateways                []*gwapiv1.Gateway             `json:"gateways,omitempty" yaml:"gateways,omitempty"`
	HTTPRoutes              []*gwapiv1.HTTPRoute           `json:"httpRoutes,omitempty" yaml:"httpRoutes,omitempty"`
[...]

PR #4326, opened 2024-09-24
feat: combine xds-translator and xds-server runners

arkodg:
* The combined runner is called `xds`
* This should eliminate one layer of caching resulting in mem savings

Fixes: https://github.com/envoyproxy/gateway/issues/3980


arkodg:
this is a breaking change
* the log components - `xdc-translator` and `xds-server` have been removed and a new log component `xds` has been added
* this also affects change in control plane metrics labels

arkodg:
looks like the mem usage increased, that looks odd
cc @shawnh2

Issue #4330, opened 2024-09-25
[Gateway API v1.2 ] GEP-1762: In Cluster Gateway Deployments


zhaohuabing:
To comply with [GEP-1762: In Cluster Gateway Deployments](https://gateway-api.sigs.k8s.io/geps/gep-1762/)

EG must:
* MUST mark the Gateway as Programmed and provide an address in Status.Addresses where the Gateway can be reached on each configured port.
* MUST label all generated resources (Service, Deployment, etc) with gateway.networking.k8s.io/gateway-name: my-gateway (where my-gateway is the name of the Gateway resource).
* MUST provision generated resources in the same namespace as the Gateway if they are namespace scoped resources.
Cluster scoped resources are not recommended.
* SHOULD name all generated resources my-gateway-example (<NAME>-<GATEWAY CLASS>). This is not simply NAME to reduce the chance of conflicts with existing resources. Where required, this can also serve as the prefix for the object.

cnvergence:
crosslinking: we have added this for GatewayNamespaceMode in https://github.com/envoyproxy/gateway/pull/5996

PR #4333, opened 2024-09-25
[Gateway API 1.2.0] Add gateway name label

zhaohuabing:
Implement the Gateway name label pfart of [GEP-1762: In Cluster Gateway Deployments](https://gateway-api.sigs.k8s.io/geps/gep-1762/)

> MUST label all generated resources (Service, Deployment, etc) with gateway.networking.k8s.io/gateway-name: my-gateway (where my-gateway is the name of the Gateway resource).

This PR only covers the gateway name label requirement in the GEP-1762.

Please note that this label has not been added to the generated resources of the merged gateway. 

Implement: https://github.com/envoyproxy/gateway/issues/4330


internal/cmd/egctl/testdata/translate/out/quickstart.all.yaml
```diff
@@ -87,6 +87,7 @@ infraIR:
         labels:
           gateway.envoyproxy.io/owning-gateway-name: eg
           gateway.envoyproxy.io/owning-gateway-namespace: envoy-gateway-system
+          gateway.networking.k8s.io/gateway-name: eg
```
arkodg, line 90: EG differs from the GEP in two ways
* label name `gateway.envoyproxy.io/owning-gateway-name`
* generated resource is in the same ns as the Gateway

this warrants a bigger discussion, lets bring this up in the community meeting next week

zhaohuabing: This PR only adds the required label `gateway.envoyproxy.io/owning-gateway-name` as it's compatible with the current EG implementation.

The existing `gateway.envoyproxy.io/owning-gateway-name` is reserved as it's used as a selector to filter out services and deployments related to a Gateway.

internal/gatewayapi/testdata/merge-valid-multiple-gateways-multiple-routes.out.yaml
```diff
@@ -216,6 +216,7 @@ infraIR:
       metadata:
         labels:
           gateway.envoyproxy.io/owning-gatewayclass: envoy-gateway-class
+          gateway.networking.k8s.io/gateway-name: gateway-1,gateway-2
```
zhaohuabing, line 219: All gateway names are concatenated and stored in a single label if mergeGateways is enabled.
 Probably not the best way to do this, should revisit this and come up with a better solution upstream.

arkodg:
Apart from passing the conformance test, how does the end user benefit from this change ?

zhaohuabing:
> Apart from passing the conformance test, how does the end user benefit from this change ?

No, but we could replace the existing `gateway.envoyproxy.io/owning-gateway-name` with this one in the next release.

arkodg:
I suggest waiting until https://github.com/kubernetes-sigs/gateway-api/issues/3366 is resolved which in indirectly related to the name of the label

adrianmace:
Hello folks 👋🏻 It looks like kubernetes-sigs/gateway-api#3366 is closed now, so I'd like this to be re-visited.

> how does the end user benefit from this change ?

I currently run multiple Gateway API implementations (Cilium & Envoy Gateway). I would like to select `Service` resources that are associated with _any_ gateway. Per the Gateway API specification, I should be able to get resources using the following:

```shell
kubectl get --all-namespaces service --selector 'gateway.networking.k8s.io/gateway-name'
```

But currently, this only returns `Service` resources associated with Cilium.

The real-world impact of this is I cannot configure a `serviceMutatorWebhookConfig.objectSelector.matchExpressions` value [in the AWS Load Balancer Controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/be5ec5b12ccfce5ae4e542d3f563fb672024714f/helm/aws-load-balancer-controller/values.yaml#L429), where my goal is for this controller to only handle Gateway-related Services.

**Q: Can this PR as currently defined be revisited?**

cc: @arkodg @zhaohuabing

arkodg:
@adrianmace this label will be available in v1.5 (end July) only for gateway namespace mode https://gateway.envoyproxy.io/docs/tasks/operations/gateway-namespace-mode/

for the default mode you'd have to use `--selector=gateway.envoyproxy.io/owning-gateway-namespace=<>,gateway.envoyproxy.io/owning-gateway-name=<>`

adrianmace:
@arkodg thank you for the fast response.

Unfortunately we do not want to use Gateway Namespace Mode (due to us wanting to provide a centralised gateway capability for a multi-tenant cluster, and since that feature is in Alpha).

Do you have plans to bring this label to the default mode in order to fully conform to GEP-1762?

arkodg:
@adrianmace EG conforms to GEP-1762 only in gateway namespace mode, in the default mode all proxy fleets run in the controller namespace `envoy-gateway-system` , so `name` is not enough, we also need the `namespace` label

adrianmace:
With this in mind, I would love for at least this label to be merged in.

PR #4334, opened 2024-09-26
api: direct response

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/2714


api/v1alpha1/shared_types.go
```diff
@@ -663,7 +663,14 @@ type CustomResponse struct {
...
 	// Body of the Custom Response
-	Body CustomResponseBody `json:"body"`
+	//
+	// +optional
+	Body *CustomResponseBody `json:"body,omitempty"`
+
+	// Status Code of the HTTP response
+	//
+	// +optional
+	StatusCode *string `json:"statusCode,omitempty"`
```
zirain, line 673: this is optional? what will happen if both `body` and `statusCode` is nil?

arkodg: Empty status code in this context will be a 200

Can add a cel validation to make sure at least one field must be set

zirain: > Empty status code for direct response will be a 200.
For response override it is the existing status code

Should address this? maybe not share the struct for better doc(I'm not sure about this)?

arkodg: Yup battling that same thought rn :)

Xunzhuo (dismissed):
let us get this in

api/v1alpha1/shared_types.go
zhaohuabing: Prefer two different types for direct response and override response for below reasons:
* For better doc as @zirain pointed out
* They may have different fields, for example, override response could add/remove headers

zirain (changes requested):
wait https://github.com/envoyproxy/gateway/pull/4334#discussion_r1779996658

zirain (approved):

api/v1alpha1/shared_types.go
```diff
@@ -669,7 +669,14 @@ type CustomResponse struct {
...
```
zhaohuabing, line 679: Nit: The `StatusCodeMatch` uses int for StatusCode, should we also use int here?

zhaohuabing (approved):

api/v1alpha1/shared_types.go
arkodg: oops, I'll just delete this field for now

zirain (approved):

zhaohuabing (approved):

Issue #4336, opened 2024-09-26
Some status updates are lost


zhaohuabing:
When upgrading Gateway API to v1.2.0, I discovered that the status updates were being discarded by the status updater, which caused [the upgrade test to fail waiting for the HTTPRoute to be accepted](https://github.com/envoyproxy/gateway/actions/runs/11048844503/job/30693628918?pr=4270).

Root cause: The status updater has been added as a Runnable in the controller manager, and started after the controller started watching resources.  If a resource change comes before the status updater is started, the status will be discarded. This only happens when EG is just started.

Test log:
```
helpers.go:609: 2024-09-26T07:11:55.597278214Z: Route gateway-upgrade-infra/http-backend-eg-upgrade expected 1 Parents got 0
    helpers.go:609: 2024-09-26T07:11:56.597265108Z: Route gateway-upgrade-infra/http-backend-eg-upgrade expected 1 Parents got 0
    helpers.go:609: 2024-09-26T07:11:57.597740078Z: Route gateway-upgrade-infra/http-backend-eg-upgrade expected 1 Parents got 0
    helpers.go:609: 2024-09-26T07:11:58.596946119Z: Route gateway-upgrade-infra/http-backend-eg-upgrade expected 1 Parents got 0
    helpers.go:402: 
        	Error Trace:	/home/ubuntu/go/pkg/mod/sigs.k8s.io/gateway-api@v1.2.0-rc1/conformance/utils/kubernetes/helpers.go:611
        	            				/home/ubuntu/go/pkg/mod/sigs.k8s.io/gateway-api@v1.2.0-rc1/conformance/utils/kubernetes/helpers.go:367
        	            				/home/ubuntu/go/pkg/mod/sigs.k8s.io/gateway-api@v1.2.0-rc1/conformance/utils/kubernetes/helpers.go:402
        	            				/home/ubuntu/gateway/test/e2e/tests/eg_upgrade.go:120
        	Error:      	Received unexpected error:
        	            	error fetching HTTPRoute: client rate limiter Wait returned an error: context deadline exceeded
        	Test:       	TestEGUpgrade/EGUpgrade/Upgrade_from_an_older_eg_release_should_succeed
        	Messages:   	error waiting for HTTPRoute to have parents matching expectations
```

HTTPRoute, status was not updated.
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  creationTimestamp: "2024-09-26T07:23:16Z"
[...]

PR #4337, opened 2024-09-26
fix: some status updates are discarded by the status updater

zhaohuabing:
Fix https://github.com/envoyproxy/gateway/issues/4336

### DIFF
--- a/internal/provider/kubernetes/status_updater.go
+++ b/internal/provider/kubernetes/status_updater.go
@@ -56,7 +56,6 @@ func (m MutatorFunc) Mutate(old client.Object) client.Object {
 type UpdateHandler struct {
 	log           logr.Logger
 	client        client.Client
-	sendUpdates   chan struct{}
 	updateChannel chan Update
 }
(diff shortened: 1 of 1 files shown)


arkodg:
thanks !
do we also need to change the order of status updater and k8s controller ?

zhaohuabing:
> thanks ! do we also need to change the order of status updater and k8s controller ?

I don't know if it's possible: status updater is added as a runnable in the k8s controller and start within it because it also needs leader election.

The current solution is removing the check for start and receive all the updates even it's not started yet.

arkodg (approved):
thanks for fixing this !

zirain (approved):

zhaohuabing:
Should we cherry-pick this to v1.1 branch @guydc @arkodg ?

arkodg:
> Should we cherry-pick this to v1.1 branch @guydc @arkodg ?

Sure let's add a label and make sure to add it to v1.1
4

PR #4340, opened 2024-09-26
bugfix: EG loglevel error for admin and metrics module

### LINKED ISSUE #4339: EG loglevel error for admin and metrics module
*Description*:
returned, etc.

*Repro steps*:
> set the EG logleve to `warn` with envoy-gateway-config but the admin and metrics can not assign the level.

```yaml
apiVersion: v1
data:
  envoy-gateway.yaml: |-
    kind: EnvoyGateway
    logging:
      level:
        default: warn
    gateway:
    provider:
      type: Kubernetes
      kubernetes:
        overwriteControlPlaneCerts: true
kind: ConfigMap
```

sharing.

*Environment*:

*Logs*:
```
2024-09-26T14:43:44.712Z        INFO    admin   admin/server.go:37      starting admin server   {"address": "127.0.0.1:19000", "enablePprof": false}
2024-09-26T14:43:44.712Z        INFO    metrics metrics/register.go:165 initialized metrics pull endpoint       {"address": "0.0.0.0:19001", "endpoint": "/metrics"}
2024-09-26T14:43:44.713Z        INFO    metrics metrics/register.go:54  starting metrics server {"address": "0.0.0.0:19001"}
```

qicz:
**What type of PR is this?**

Fixes #4339

### DIFF
--- a/internal/metrics/register.go
+++ b/internal/metrics/register.go
@@ -23,11 +23,14 @@ import (
 
 	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
+	log "github.com/envoyproxy/gateway/internal/logging"
 )
 
 const (
 	defaultEndpoint = "/metrics"
 )
+
+var metricsLogger log.Logger
 
 // Init initializes and registers the global metrics server.
 func Init(cfg *config.Server) error {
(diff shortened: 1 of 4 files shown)


zirain (approved):

internal/metrics/register.go
```diff
@@ -42,6 +45,8 @@ func Init(cfg *config.Server) error {
 	}
 
 	if !options.pullOptions.disable {
+		metricsLogger = cfg.Logger.WithName("metrics")
```
arkodg, line 48: why not move this to L37 ?

qicz: done

zirain (approved):

arkodg (approved):
LGTM thanks !

PR #4341, opened 2024-09-26
feat(translator): implement access log types

### LINKED ISSUE #3600: Support Listener Access Logging

guydc:
Fixes #3600


internal/xds/translator/testdata/in/xds-ir/accesslog-als-tcp.yaml
```diff
@@ -7,6 +7,7 @@ accesslog:
       path: "%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%"
       protocol: "%PROTOCOL%"
       response_code: "%RESPONSE_CODE%"
+    logType: Global
```
zirain, line 10: can we avoid this? that will make this PR a little smaller?

guydc: done

zirain (dismissed):

site/content/en/latest/tasks/observability/proxy-accesslog.md
```diff
@@ -249,3 +249,49 @@ Envoy Gateway provides additional metadata about the K8s resources that were tra
...
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: EnvoyProxy
+metadata:
+  name: otel-access-logging
+  namespace: envoy-gateway-system
+spec:
+  telemetry:
+    accessLog:
+      settings:
+        - type: Listener
```
arkodg, line 284: so this is a override style setting ? i.e. since a `Route` type is not defined, the define Route AccessLog setting is applied ?

internal/xds/translator/testdata/out/xds-ir/accesslog-types.routes.yaml
```diff
@@ -0,0 +1,2 @@
+- ignorePortInHostMatching: true
```
arkodg, line 1: lets add a valid xds ir route value so this gets populated properly

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4363, opened 2024-09-29
refactor: return 500 when BackendTLSPolicy translation fails

alexwo:
**What type of PR is this?**

What this PR does / why we need it:
- 500 for HTTP
- connection reset for TCP/UDP

Which issue(s) this PR fixes:
Fixes https://github.com/envoyproxy/gateway/issues/4087

### DIFF
--- a/internal/gatewayapi/backendtlspolicy.go
+++ b/internal/gatewayapi/backendtlspolicy.go
@@ -8,7 +8,7 @@ package gatewayapi
 import (
 	"fmt"
 	"reflect"
-
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 	gwapiv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
@@ -20,8 +20,11 @@ import (
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
-func (t *Translator) applyBackendTLSSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string, parent gwapiv1a2.ParentReference, resources *resource.Resources, envoyProxy *egv1a1.EnvoyProxy) *ir.TLSUpstreamConfig {
-	upstreamConfig, policy := t.processBackendTLSPolicy(backendRef, backendNamespace, parent, resources, envoyProxy)
+func (t *Translator) applyBackendTLSSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string, parent gwapiv1a2.ParentReference, resources *resource.Resources, envoyProxy *egv1a1.EnvoyProxy) (*ir.TLSUpstreamConfig, error) {
+	upstreamConfig, policy, err := t.processBackendTLSPolicy(backendRef, backendNamespace, parent, resources)
+	if err != nil {
+		return nil, err
+	}
 	return t.applyEnvoyProxyBackendTLSSetting(policy, upstreamConfig, resources, parent, envoyProxy)
 }
 
@@ -30,18 +33,13 @@ func (t *Translator) processBackendTLSPolicy(
 	backendNamespace string,
 	parent gwapiv1a2.ParentReference,
 	resources *resource.Resources,
-	envoyProxy *egv1a1.EnvoyProxy,
-) (*ir.TLSUpstreamConfig, *gwapiv1a3.BackendTLSPolicy) {
-	policy := getBackendTLSPolicy(resources.BackendTLSPolicies, backendRef, backendNamespace, resources)
+) (*ir.TLSUpstreamConfig, *gwapiv1a3.BackendTLSPolicy, error) {
+	policy := getBackendTLSPolicy(resources.BackendTLSPolicies, backendRef, backendNamespace,resources)
 	if policy == nil {
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	tlsBundle, err := getBackendTLSBundle(policy, resources)
-	if err == nil && tlsBundle == nil {
-		return nil, nil
-	}
-
 	ancestorRefs := getAncestorRefs(policy)
 	ancestorRefs = append(ancestorRefs, parent)
 
@@ -52,42 +50,16 @@ func (t *Translator) processBackendTLSPolicy(
 			policy.Generation,
 			status.Error2ConditionMsg(err),
 		)
-		return nil, nil
+		return nil, nil, err
 	}
 
 	status.SetAcceptedForPolicyAncestors(&policy.Status, ancestorRefs, t.GatewayControllerName)
-	// apply defaults as per envoyproxy
-	if envoyProxy != nil {
-		if envoyProxy.Spec.BackendTLS != nil {
-			if len(envoyProxy.Spec.BackendTLS.Ciphers) > 0 {
-				tlsBundle.Ciphers = envoyProxy.Spec.BackendTLS.Ciphers
-			}
-			if len(envoyProxy.Spec.BackendTLS.ECDHCurves) > 0 {
-				tlsBundle.ECDHCurves = envoyProxy.Spec.BackendTLS.ECDHCurves
-			}
-			if len(envoyProxy.Spec.BackendTLS.SignatureAlgorithms) > 0 {
-				tlsBundle.SignatureAlgorithms = envoyProxy.Spec.BackendTLS.SignatureAlgorithms
-			}
-			if envoyProxy.Spec.BackendTLS.MinVersion != nil {
-				tlsBundle.MinVersion = ptr.To(ir.TLSVersion(*envoyProxy.Spec.BackendTLS.MinVersion))
-			}
-			if envoyProxy.Spec.BackendTLS.MaxVersion != nil {
-				tlsBundle.MaxVersion = ptr.To(ir.TLSVersion(*envoyProxy.Spec.BackendTLS.MaxVersion))
-			}
-			if len(envoyProxy.Spec.BackendTLS.ALPNProtocols) > 0 {
-				tlsBundle.ALPNProtocols = make([]string, len(envoyProxy.Spec.BackendTLS.ALPNProtocols))
-				for i := range envoyProxy.Spec.BackendTLS.ALPNProtocols {
-					tlsBundle.ALPNProtocols[i] = string(envoyProxy.Spec.BackendTLS.ALPNProtocols[i])
-				}
-			}
-		}
-	}
-	return tlsBundle, policy
+	return tlsBundle, policy, nil
 }
 
-func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendTLSPolicy, tlsConfig *ir.TLSUpstreamConfig, resources *resource.Resources, parent gwapiv1a2.ParentReference, ep *egv1a1.EnvoyProxy) *ir.TLSUpstreamConfig {
+func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendTLSPolicy, tlsConfig *ir.TLSUpstreamConfig, resources *resource.Resources, parent gwapiv1a2.ParentReference, ep *egv1a1.EnvoyProxy) (*ir.TLSUpstreamConfig, error) {
 	if ep == nil || ep.Spec.BackendTLS == nil || tlsConfig == nil {
-		return tlsConfig
+		return tlsConfig, nil
 	}
 
 	if len(ep.Spec.BackendTLS.Ciphers) > 0 {
@@ -116,28 +88,41 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
 		ancestorRefs := []gwapiv1a2.ParentReference{
 			parent,
 		}
+		var err error
 		if ns != ep.Namespace {
+			err = fmt.Errorf("ClientCertificateRef Secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
 			status.SetTranslationErrorForPolicyAncestors(&policy.Status,
 				ancestorRefs,
 				t.GatewayControllerName,
 				policy.Generation,
-				status.Error2ConditionMsg(fmt.Errorf("client authentication TLS secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)))
-			return tlsConfig
+				status.Error2ConditionMsg(err))
+			return tlsConfig, err
 		}
 		secret := resources.GetSecret(ns, string(ep.Spec.BackendTLS.ClientCertificateRef.Name))
 		if secret == nil {
+			err = fmt.Errorf(
+				"failed to locate TLS secret for client auth: %s specified in EnvoyProxy %s",
+				types.NamespacedName{
+					Namespace: ep.Namespace,
+					Name:      string(ep.Spec.BackendTLS.ClientCertificateRef.Name),
+				}.String(),
+				types.NamespacedName{
+					Namespace: ep.Namespace,
+					Name:      ep.Name,
+				}.String(),
+			)
 			status.SetTranslationErrorForPolicyAncestors(&policy.Status,
 				ancestorRefs,
 				t.GatewayControllerName,
 				policy.Generation,
-				status.Error2ConditionMsg(fmt.Errorf("failed to locate TLS secret for client auth: %s in namespace: %s", ep.Spec.BackendTLS.ClientCertificateRef.Name, ns)),
+				status.Error2ConditionMsg(err),
 			)
-			return tlsConfig
+			return tlsConfig, err
 		}
 		tlsConf := irTLSConfigs(secret)
 		tlsConfig.ClientCertificates = tlsConf.Certificates
 	}
-	return tlsConfig
+	return tlsConfig, nil
 }
 
 func backendTLSTargetMatched(policy gwapiv1a3.BackendTLSPolicy, target gwapiv1a2.LocalPolicyTargetReferenceWithSectionName, backendNamespace string) bool {
--- a/internal/gatewayapi/backendtlspolicy.go
+++ b/internal/gatewayapi/backendtlspolicy.go
@@ -8,6 +8,7 @@ package gatewayapi
 import (
 	"fmt"
 	"reflect"
+
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -34,7 +35,7 @@ func (t *Translator) processBackendTLSPolicy(
 	parent gwapiv1a2.ParentReference,
 	resources *resource.Resources,
 ) (*ir.TLSUpstreamConfig, *gwapiv1a3.BackendTLSPolicy, error) {
-	policy := getBackendTLSPolicy(resources.BackendTLSPolicies, backendRef, backendNamespace,resources)
+	policy := getBackendTLSPolicy(resources.BackendTLSPolicies, backendRef, backendNamespace, resources)
 	if policy == nil {
 		return nil, nil, nil
 	}
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -184,8 +184,10 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 
 	// compute matches, filters, backends
 	for ruleIdx, rule := range httpRoute.Spec.Rules {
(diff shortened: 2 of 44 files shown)


zhaohuabing (commented):
@alexwo Thanks for picking up this! This PR looks great overall! I just have a few nits.

internal/gatewayapi/route.go
```diff
@@ -1196,7 +1216,7 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
 // the same proportion as the backend would have otherwise received
 func (t *Translator) processDestination(backendRefContext BackendRefContext,
 	parentRef *RouteParentContext, route RouteContext, resources *resource.Resources,
-) (ds *ir.DestinationSetting) {
+) (ds *ir.DestinationSetting, err error) {
```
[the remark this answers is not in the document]
alexwo: I think we should not ignore, this results in a bigger change. will push an updated version

internal/gatewayapi/testdata/envoyproxy-tls-settings-invalid-ns.out.yaml
```diff
@@ -310,7 +286,6 @@ xdsIR:
               - ECDSA-SECP256R1-SHA256
               sni: example.com
               useSystemTrustStore: true
-            weight: 1
```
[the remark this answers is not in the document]
alexwo: The route will not be added if it is a TCP/UDP route and contains invalid configurations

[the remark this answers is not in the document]
guydc: @alexwo - let's maybe update the PR description and title so that the release notes will reflect the two behaviors:
- 500 for HTTP
- connection reset for TCP/UDP

In terms of UX, users experience a reset in this case right? Just want to make sure that the connections on the client side don't hang and wait for some timeout, so that it's clear to users that the request is explicitly rejected.

internal/gatewayapi/validate.go
```diff
@@ -26,28 +26,30 @@ import (
 
 func (t *Translator) validateBackendRef(backendRefContext BackendRefContext, parentRef *RouteParentContext, route RouteContext,
```
guydc, line 27: Are the changes made here extending the scope of the PR to reject additional cases like invalid backend config, or is this already handled with a 500 response?

internal/gatewayapi/testdata/tlsroute-with-backend.out.yaml
```diff
@@ -121,8 +121,6 @@ xdsIR:
       routes:
       - destination:
           name: tlsroute/default/tlsroute-1/rule/-1
```
guydc, line 123: should the destination exist in this case?

internal/gatewayapi/testdata/tcproute-with-backend.out.yaml
```diff
@@ -92,9 +92,10 @@ tcpRoutes:
     parents:
     - conditions:
       - lastTransitionTime: null
-        message: Route is accepted
-        reason: Accepted
-        status: "True"
+        message: 'backend reference validation failed: backend is not supported for
```
guydc, line 95: nice!

guydc (approved):
LGTM. Few questions on behavior and scope.

internal/gatewayapi/testdata/tlsroute-with-backend.out.yaml
alexwo: nope, updated.

internal/gatewayapi/testdata/envoyproxy-tls-settings-invalid-ns.out.yaml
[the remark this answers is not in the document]
alexwo: If the route associated with the TLS connection is removed, ongoing connections that were relying on this route will be dropped. This will most likely result in a TCP reset (RST) or a termination of the TLS session, effectively severing the connection.

internal/gatewayapi/validate.go
alexwo: Yes, these changes capture cases where the backend configuration is invalid.

guydc (approved):

internal/gatewayapi/testdata/tlsroute-with-backend.out.yaml
```diff
@@ -121,8 +121,6 @@ xdsIR:
...
-          settings:
```
arkodg, line 124: curious what happened here, and why this got deleted ?

internal/gatewayapi/testdata/httproute-with-mirror-filter-service-not-found.out.yaml
```diff
@@ -72,9 +72,10 @@ httpRoutes:
     parents:
     - conditions:
       - lastTransitionTime: null
-        message: Route is accepted
-        reason: Accepted
-        status: "True"
+        message: 'Backend service validation failed: Service default/service-unknown
```
arkodg, line 75: this 2 similar conditions here
```
Backend service validation failed: Service default/service-unknown
```

&&


```
 Service default/service-unknown not found
```

internal/gatewayapi/testdata/httproute-with-mirror-filter-service-no-port.out.yaml
```diff
@@ -71,9 +71,10 @@ httpRoutes:
     parents:
     - conditions:
       - lastTransitionTime: null
-        message: Route is accepted
-        reason: Accepted
-        status: "True"
+        message: 'Error validating backend port: port number not specified for backend
+          reference.'
+        reason: UnsupportedValue
```
arkodg, line 76: similar comment as https://github.com/envoyproxy/gateway/pull/4363/files#r1792663080

internal/gatewayapi/route.go
```diff
@@ -1191,12 +1229,12 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
 	}
 }
 
-// processDestination takes a backendRef and translates it into destination setting or sets error statuses and
-// returns the weight for the backend so that 500 error responses can be returned for invalid backends in
-// the same proportion as the backend would have otherwise received
+// processDestination translate a backendRef into a destination settings.
```
arkodg, line 1232:
```suggestion
// processDestination translates a backendRef into a destination settings.
```

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -115,28 +118,31 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
 		ancestorRefs := []gwapiv1a2.ParentReference{
 			parent,
 		}
+		var err error
 		if ns != ep.Namespace {
+			err = fmt.Errorf("client authentication TLS secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
```
arkodg, line 123:
```suggestion
			err = fmt.Errorf("ClientCertificteRef Secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
```

internal/gatewayapi/route.go
```diff
@@ -112,7 +112,7 @@ func (t *Translator) processHTTPRouteParentRefs(httpRoute *HTTPRouteContext, res
 		// any conditions that come out of it have to go on each RouteParentStatus,
 		// not on the Route as a whole.
 		routeRoutes, err := t.processHTTPRouteRules(httpRoute, parentRef, resources)
-		if err != nil {
+		if err != nil && !parentRef.HasCondition(httpRoute, gwapiv1.RouteConditionAccepted, metav1.ConditionFalse) {
```
arkodg, line 115: curious why we need to change the status generation logic

internal/gatewayapi/route.go
```diff
@@ -470,7 +479,7 @@ func (t *Translator) processGRPCRouteParentRefs(grpcRoute *GRPCRouteContext, res
 		// any conditions that come out of it have to go on each RouteParentStatus,
 		// not on the Route as a whole.
 		routeRoutes, err := t.processGRPCRouteRules(grpcRoute, parentRef, resources)
-		if err != nil {
+		if err != nil && !parentRef.HasCondition(grpcRoute, gwapiv1.RouteConditionAccepted, metav1.ConditionFalse) {
```
arkodg, line 482: same as https://github.com/envoyproxy/gateway/pull/4363/files#r1792677269

internal/gatewayapi/route.go
```diff
@@ -965,11 +984,21 @@ func (t *Translator) processUDPRouteParentRefs(udpRoute *UDPRouteContext, resour
 		}
 
 		for _, backendRef := range udpRoute.Spec.Rules[0].BackendRefs {
-			ds := t.processDestination(backendRef, parentRef, udpRoute, resources)
-			if ds == nil {
+			ds, err := t.processDestination(backendRef, parentRef, udpRoute, resources)
+			// skip adding the route and provide the reason via route status.
+
+			if err != nil && !parentRef.HasCondition(udpRoute, gwapiv1.RouteConditionAccepted, metav1.ConditionFalse) {
```
arkodg, line 990: this isnt added for HTTPRoute and GRPCRoute but is added for UDP and TCP ?

internal/gatewayapi/route.go
```diff
@@ -1098,11 +1127,20 @@ func (t *Translator) processTCPRouteParentRefs(tcpRoute *TCPRouteContext, resour
...
+			// skip adding the route and provide the reason via route status.
+			if err != nil && !parentRef.HasCondition(tcpRoute, gwapiv1.RouteConditionAccepted, metav1.ConditionFalse) {
+				routeStatus := GetRouteStatus(tcpRoute)
+				status.SetRouteStatusCondition(routeStatus,
+					parentRef.routeParentStatusIdx,
+					tcpRoute.GetGeneration(),
+					gwapiv1.RouteConditionAccepted,
+					metav1.ConditionFalse,
+					"Failed to process the settings associated with the TCP route.",
+					err.Error(),
```
arkodg, line 1140: same as above

internal/gatewayapi/route.go
alexwo: For HTTP and TCP routes, we currently accept the route but return a 500 response. However, for UDP and TCP routes, the suggestion is to reject and not translate it at all incase of errors, and indicate this via the route status as traffic won't be routable at all.

internal/gatewayapi/route.go
alexwo: To clarify why the traffic cannot be routed, previously, the route was accepted, and the traffic was considered routable before this change.

internal/gatewayapi/testdata/httproute-with-mirror-filter-service-not-found.out.yaml
alexwo: Would it be better to provide a static message explaining why the status is not accepted, such as “invalid configuration detected”? This approach would avoid repeating the message in the accepted status reason. The types, such as “Accepted” and “ResolvedRefs,” are already distinct.

internal/gatewayapi/testdata/tlsroute-with-backend.out.yaml
alexwo: In this scenario, the route generally shouldn’t be operational. I’ll look into why the settings are being removed to better understand the problem. For now, the route is left as-is because fully invalidating it causes the conformance test to fail. Perhaps we can ignore the TLS route translation error and create a separate task to address it.

internal/gatewayapi/testdata/httproute-with-mirror-filter-service-not-found.out.yaml
arkodg: I think its fine for this PR

internal/gatewayapi/route.go
arkodg: trying to better understand why do we need an additional check here, isnt `if err != nil` enough ?

alexwo: At this stage, it is already possible that a route is not applicable due to other reasons. This check attempts to retain the original reason, as it may be more specific. However, if that is the case, no traffic would be routed regardless to the settings validations errors.

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -30,15 +33,15 @@ func (t *Translator) processBackendTLSPolicy(
...
 	policy := getBackendTLSPolicy(resources.BackendTLSPolicies, backendRef, backendNamespace)
 	if policy == nil {
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	tlsBundle, err := getBackendTLSBundle(policy, resources)
 	if err == nil && tlsBundle == nil {
-		return nil, nil
+		return nil, nil, nil
```
zhaohuabing, line 44: Nit: When will this condition be true? `err == nil && tlsBundle == nil`? Should we just remove this?

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -19,8 +19,11 @@ import (
 	"github.com/envoyproxy/gateway/internal/ir"
 )
 
-func (t *Translator) applyBackendTLSSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string, parent gwapiv1a2.ParentReference, resources *resource.Resources, envoyProxy *egv1a1.EnvoyProxy) *ir.TLSUpstreamConfig {
-	upstreamConfig, policy := t.processBackendTLSPolicy(backendRef, backendNamespace, parent, resources, envoyProxy)
+func (t *Translator) applyBackendTLSSetting(backendRef gwapiv1.BackendObjectReference, backendNamespace string, parent gwapiv1a2.ParentReference, resources *resource.Resources, envoyProxy *egv1a1.EnvoyProxy) (*ir.TLSUpstreamConfig, error) {
+	upstreamConfig, policy, err := t.processBackendTLSPolicy(backendRef, backendNamespace, parent, resources, envoyProxy)
```
zhaohuabing, line 23: The code from line 61 to line 86  is redundant because the EnvoyProxy TLS setting has been applied again within `applyEnvoyProxyBackendTLSSetting`.

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -115,28 +118,31 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
...
+		var err error
 		if ns != ep.Namespace {
+			err = fmt.Errorf("ClientCertificteRef Secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
 			status.SetTranslationErrorForPolicyAncestors(&policy.Status,
 				ancestorRefs,
 				t.GatewayControllerName,
 				policy.Generation,
-				status.Error2ConditionMsg(fmt.Errorf("client authentication TLS secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)))
-			return tlsConfig
+				status.Error2ConditionMsg(err))
```
zhaohuabing, line 128: Non-blocking: we probably should set the status of the EnvoyProxy instead of BackendTLSPolicy as the `ClientCertificateRef` is not part of the `BackendTLSPolicy`.  WDYT @arkodg ?

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -115,28 +118,31 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
...
 		}
 		secret := resources.GetSecret(ns, string(ep.Spec.BackendTLS.ClientCertificateRef.Name))
 		if secret == nil {
+			err = fmt.Errorf("failed to locate TLS secret for client auth: %s in namespace: %s", ep.Spec.BackendTLS.ClientCertificateRef.Name, ns)
 			status.SetTranslationErrorForPolicyAncestors(&policy.Status,
 				ancestorRefs,
 				t.GatewayControllerName,
 				policy.Generation,
-				status.Error2ConditionMsg(fmt.Errorf("failed to locate TLS secret for client auth: %s in namespace: %s", ep.Spec.BackendTLS.ClientCertificateRef.Name, ns)),
+				status.Error2ConditionMsg(err),
```
zhaohuabing, line 138: Non-blocking: same here.

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -115,28 +118,31 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
 		ancestorRefs := []gwapiv1a2.ParentReference{
 			parent,
 		}
+		var err error
 		if ns != ep.Namespace {
+			err = fmt.Errorf("ClientCertificteRef Secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
```
zhaohuabing, line 123:
```suggestion
			err = fmt.Errorf("ClientCertificateRef Secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)
```

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -115,28 +118,31 @@ func (t *Translator) applyEnvoyProxyBackendTLSSetting(policy *gwapiv1a3.BackendT
...
 				t.GatewayControllerName,
 				policy.Generation,
-				status.Error2ConditionMsg(fmt.Errorf("client authentication TLS secret is not located in the same namespace as Envoyproxy. Secret namespace: %s does not match Envoyproxy namespace: %s", ns, ep.Namespace)))
-			return tlsConfig
+				status.Error2ConditionMsg(err))
+			return tlsConfig, err
 		}
 		secret := resources.GetSecret(ns, string(ep.Spec.BackendTLS.ClientCertificateRef.Name))
 		if secret == nil {
+			err = fmt.Errorf("failed to locate TLS secret for client auth: %s in namespace: %s", ep.Spec.BackendTLS.ClientCertificateRef.Name, ns)
```
zhaohuabing, line 133: Suggest mentioning the EnvoyProxy in the Status to avoid confusion.
```suggestion
			err = fmt.Errorf(
				"failed to locate TLS secret for client auth: %s specified in EnvoyProxy %s",
				types.NamespacedName{
					Namespace: ep.Namespace,
					Name:      string(ep.Spec.BackendTLS.ClientCertificateRef.Name),
				}.String(),
				types.NamespacedName{
					Namespace: ep.Namespace,
					Name:      ep.Name,
				}.String(),
			)
```

internal/gatewayapi/backendtlspolicy.go
arkodg: we haven't added a status field yet in `EnvoyProxy` and continue to use the linked resource as the place to update status

internal/gatewayapi/route.go
arkodg: can you share an example of what the status looks like with and w/o this change ?

alexwo: hmm actually you are right this does not seem to happen.

arkodg:
@alexwo I dont see the conformance tests passing. I'm worried about the size of this change that not only applies to BTP but every error while translating route rules and backendRefs. For example if one out of the many backendRefs is invalid, only a portion of the requests should receive a 503
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPBackendRef

alexwo:
> @alexwo I dont see the conformance tests passing. I'm worried about the size of this change that not only applies to BTP but every error while translating route rules and backendRefs. For example if one out of the many backendRefs is invalid, only a portion of the requests should receive a 503 https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPBackendRef

@arkodg  indeed, the impact of the change is large, I can relate to it but hopefully all the tests in place get us covered.
The conformance tests pass, there was a lint issue and they did not start at all before.

If a portion of a backend ref configuration is invalid, say we are unable to process one the associated configurations, the question if routing to this backend ref should still happen?

arkodg:
> Now, if a part of the backend reference configuration is invalid — for example, if one of the associated configurations cannot be processed — the key question is: should routing to this backend reference still occur in such cases?

According to the spec, the answer is a portion of the requests (based on weights) should continue to be sent to the valid backendRef

alexwo:
> > Now, if a part of the backend reference configuration is invalid — for example, if one of the associated configurations cannot be processed — the key question is: should routing to this backend reference still occur in such cases?
> 
> According to the spec, the answer is a portion of the requests (based on weights) should continue to be sent to the valid backendRef

We’re processing all backend routes and addressing only falty configured ones. This should ensures alignment with the spec, as valid backend routes will continue to function, while only those with invalid configurations will fail to render. 
( we process all routes for each backend and do not stop at error. )

arkodg (approved):
LGTM thanks !

guydc (dismissed):
LGTM!
@alexwo  - can you update the PR?

zirain (approved):

PR #4343, opened 2024-09-26
chore: upgrade test should use v1.1.2 by default

guydc:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

PR #4384, opened 2024-10-03
bump gwapi to 1.2.0

zirain:

PR #4393, opened 2024-10-06
chore: upgrade test should use VERSION by default

zirain:
fixes: https://github.com/envoyproxy/gateway/pull/4343

xref: https://github.com/envoyproxy/gateway/pull/4384


test/e2e/tests/eg_upgrade.go
```diff
@@ -267,46 +277,60 @@ func updateChartCRDs(actionConfig *action.Configuration, gatewayChart *chart.Cha
...
-					continue
-				}
 
-				// previous version exists
-				existingVersion, err := getGWAPIVersion(existingCRD)
-				if err != nil {
-					return err
-				}
+			if existingVersion == "v1.0.0" {
+				existingCRD.Status.StoredVersions = []string{storedVersion}
```
guydc, line 330: Right now, gw-api recommends to delete CRDs to completely: https://gateway-api.sigs.k8s.io/guides/?h=upgrade#v11-upgrade-notes

Should we change our instructions to users to reflect the upgrade path that is tested by EG?  https://github.com/envoyproxy/gateway/blob/0f4cb27f0110051f7811122a2dd73652c93a15b2/site/content/en/v1.1/install/install-yaml.md?plain=1#L41

zirain: here should be `1.1` -> `1.2`, let me try to remove it.

shawnh2 (approved):

Issue #3173, opened 2024-04-10
Support additional host rewrite options


guydc:
*Description*:

Envoy supports multiple host header rewriting options:
- Literal rewriting: the host header is rewritten to a statically defined value
- Automatic rewriting: the host header is rewritten with the [DNS name](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-socketaddress) or endpoint [hostname](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-field-config-endpoint-v3-endpoint-hostname) of the selected envoy cluster. 
- Header-based rewriting: envoy rewrites the host header with a value consumed from a different header. The different header name is statically defined in
- Path regex rewriting: the host header is rewritten with a value derived by applying a regex to the request path. 

Unlike other headers, it's safer to rewrite the host header in the forwarding phase, to avoid possible issues related to failed route recalculation after the header is rewritten. Envoy's host rewrite feature are implemented in this way. 

Gateway-API and Envoy gateway support literal [host rewriting](https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/?h=rewrite#rewrites). 

Scenarios:
- Automatic rewriting: When multiple backends are configured (e.g. for weighted traffic splitting), it could be necessary to rewrite the hostname automatically, based on the dynamically selected backend. 
- Header-based rewriting: allows envoy to rewrite the host header based on information provided as a hint by the end user or extracted from the request (e.g. a jwt claim with the desired target host header transformed to a header)
- Path-Regex rewriting: rewriting to a subdomain that is encoded as a path parameter in a request to the top-level domain.  

Envoy Gateway can support additional host rewriting options by adding new options to a policy or implementing EG-specific HTTP Route Filters. 

Application developers are the likely users of this feature, as they know the desired host header value and structure in the backend. 

Prior art: these features are supported by some gateways in the K8s space: 
- Nginx Ingress support the annotation [upstream-vhost](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#custom-nginx-upstream-vhost), which can be used to inject dynamic values such as value from other headers, using nginx variables, e.g. `nginx.ingress.kubernetes.io/upstream-vhost: $http_x_custom_header`
[...]

PR #4410, opened 2024-10-08
api: host header rewrite

guydc:
Support additional host rewrite options in HTTPRouteFilter

Relates #3173


api/v1alpha1/httproutefilter_types.go
```diff
@@ -91,6 +106,26 @@ type HTTPPathModifier struct {
...
+
+	// SetFromHeader indicates that the Host header value would be replaced with the value of the header specified in setFromHeader.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-host-rewrite-header
+	// +optional
+	SetFromHeader *string `json:"setFromHeader,omitempty"`
+
+	// SetFromBackend indicates that the Host header value would be replaced by the DNS name of the backend if it exists.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-auto-host-rewrite
+	// +optional
+	SetFromBackend *bool `json:"setFromBackend,omitempty"`
```
zhaohuabing, line 126: What will happen when type=SetFromBackend && SetFromBacken=false ?

zhaohuabing, line 126: What will happen when type=SetFromBackend && SetFromBacken=false ?

api/v1alpha1/httproutefilter_types.go
```diff
@@ -37,6 +37,13 @@ type HTTPRouteFilterSpec struct {
 
 // HTTPURLRewriteFilter define rewrites of HTTP URL components such as path and host
 type HTTPURLRewriteFilter struct {
+	// Hostname is the value to be used to replace the Host header value during
+	// forwarding.
+	//
+	// Support: Extended
```
zhaohuabing, line 43: Nit: remove this line?

api/v1alpha1/httproutefilter_types.go
guydc: right, it's redundant, i'll remove it =]

api/v1alpha1/httproutefilter_types.go
```diff
@@ -37,6 +37,12 @@ type HTTPRouteFilterSpec struct {
 
 // HTTPURLRewriteFilter define rewrites of HTTP URL components such as path and host
 type HTTPURLRewriteFilter struct {
+	// Hostname is the value to be used to replace the Host header value during
+	// forwarding.
+	//
+	// +optional
+	// +notImplementedHide
+	Hostname *HTTPHostnameModifier `json:"hostname,omitempty"`
```
zhaohuabing, line 45: Nit: How about `Host` instead of `Hostname`? 
```suggestion
	Host *HTTPHostModifier `json:"host,omitempty"`
```

guydc: The GW-API terminology is hostname: https://gateway.envoyproxy.io/docs/tasks/traffic/http-urlrewrite/#rewrite-host-name, so I think this would be more consistent for users.

guydc:
@arkodg  - follow up to yesterday's community meetings: it seems that it's not possible to mutate the host header using the standard [header mutation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/header_mutation_filter) feature. Instead, the custom specific envoy host rewrite features must be used, as documented [here](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers):

> Neither :-prefixed pseudo-headers nor the Host: header may be modified via this mechanism. The :path and :authority headers may instead be modified via mechanisms such as [prefix_rewrite](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-prefix-rewrite), [regex_rewrite](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-regex-rewrite), and [host_rewrite](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-host-rewrite-literal).

zhaohuabing (approved):
LGTM, Thanks!

zirain (approved):

api/v1alpha1/httproutefilter_types.go
```diff
@@ -53,6 +59,18 @@ const (
...
 )
 
+// HTTPPathModifierType defines the type of Hostname rewrite.
+type HTTPHostnameModifierType string
+
+const (
+	// HeaderHTTPHostnameModifier indicates that the Host header value would be replaced with the value of the header specified in setFromHeader.
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-host-rewrite-header
+	HeaderHTTPHostnameModifier HTTPHostnameModifierType = "SetFromHeader"
+	// BackendHTTPHostnameModifier indicates that the Host header value would be replaced by the DNS name of the backend if it exists.
```
arkodg, line 69: too late, but sharing my preference here
```
type: Header
header: "foo"
```

api/v1alpha1/httproutefilter_types.go
```diff
@@ -53,6 +59,18 @@ const (
...
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-auto-host-rewrite
+	BackendHTTPHostnameModifier HTTPHostnameModifierType = "SetFromBackend"
```
arkodg, line 71: do we need Backend in the first iteration ? can we only start with header ?

api/v1alpha1/httproutefilter_types.go
guydc: I'll adjust it in the implementation PR.

api/v1alpha1/httproutefilter_types.go
guydc: this should be equivalent to the nginx ingress `nginx.ingress.kubernetes.io/upstream-vhost: $proxy_host` that allows the value of the host header to by dynamically rewritten to that of the target host. Prefer to support this if possible.

PR #4414, opened 2024-10-09
chore: fix osv scanner

zirain:


zhaohuabing (approved):
I also ran into this before. It looks like OSC scanner has issues identifying licences for unofficial releases.

arkodg (approved):

PR #4415, opened 2024-10-09
impl: custom error response

zhaohuabing:
Implement: https://github.com/envoyproxy/gateway/pull/4259 

Issue: https://github.com/envoyproxy/gateway/issues/1400


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,7 +74,6 @@ type BackendTrafficPolicySpec struct {
 	// If multiple configurations are specified, the first one to match wins.
 	//
 	// +optional
-	// +notImplementedHide
 	ResponseOverride []*ResponseOverride `json:"responseOverride,omitempty"`
```
zhaohuabing, line 77: I'm considering encapsulating the respond overriding rules within a top-level structure to make it more future-proof. This would allow us to add some global settings in the future.

```yaml
  respondOverride:
    rules:
      - match: ....
        response: ... 
      - match: ....
        response: ... 
```  

WDYT @arkodg

api/v1alpha1/shared_types.go
```diff
@@ -644,16 +658,16 @@ type StatusCodeMatch struct {
 	// Value contains the value of the status code.
 	//
 	// +optional
-	Value *string `json:"value,omitempty"`
-	// ValueRef contains the contents of the body
-	// specified as a local object reference.
-	// Only a reference to ConfigMap is supported.
+	Value *int `json:"value,omitempty"`
```
zhaohuabing, line 661: Change the type to `int` to align with the types used in `Range`.

internal/provider/kubernetes/predicates.go
```diff
@@ -605,11 +605,23 @@ func (r *gatewayAPIReconciler) validateConfigMapForReconcile(obj client.Object)
...
+
+	btpList := &egv1a1.BackendTrafficPolicyList{}
+	if err := r.client.List(context.Background(), btpList, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(configMapBtpIndex, utils.NamespacedName(configMap).String()),
+	}); err != nil {
+		r.log.Error(err, "unable to find associated BackendTrafficPolicy")
 		return false
 	}
 
-	return true
```
zhaohuabing, line 612: This was a bug in the `validateConfigMapForReconcile` method.

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: can you think of any global settings ?
`rules` adds one more level of config authorship

api/v1alpha1/shared_types.go
```diff
@@ -676,21 +691,39 @@ type CustomResponse struct {
...
 	//
 	// +optional
 	Inline *string `json:"inline,omitempty"`
+
 	// ValueRef contains the contents of the body
 	// specified as a local object reference.
 	// Only a reference to ConfigMap is supported.
 	//
+	// The contents of the ConfigMap must have a key-value pair where
+	// the key is `response.body` and the value is the body content.
```
arkodg, line 725: do we need to enforce the key name here ? Any key name (like file name) should be fine imo, EG will look into `value[0]` for the data

zhaohuabing: Explicitly asking for a fixed name creates a better UI, but we can fall back to value[0] if this key doesn't exit. WDYT?

api/v1alpha1/backendtrafficpolicy_types.go
zhaohuabing: I don't have a concrete example in my mind yet. A possible use case could be adding a global header in the response for all the rules?

This is non-blocking though.

api/v1alpha1/shared_types.go
```diff
@@ -627,33 +627,48 @@ type ResponseOverride struct {
 // CustomResponseMatch defines the configuration for matching a user response to return a custom one.
 type CustomResponseMatch struct {
 	// Status code to match on. The match evaluates to true if any of the matches are successful.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=50
 	StatusCode []StatusCodeMatch `json:"statusCode"`
```
arkodg, line 632: can we rename this to `statusCodes` ?

api/v1alpha1/shared_types.go
zhaohuabing: @envoyproxy/gateway-maintainers Could you please vote for the name of the response body key in the ConfigMap?
* response.body
* body 
* Or other names that you suggest?

api/v1alpha1/shared_types.go
zhaohuabing: Done. We may also want to rename the `ResponseOverride` in the BTP spec.

```
	// ResponseOverride defines the configuration to override specific responses with a custom one.
	// If multiple configurations are specified, the first one to match wins.
	//
	// +optional
	ResponseOverrides []*ResponseOverride `json:"responseOverrides,omitempty"`
```

arkodg: We've historically named all the top level fields  as singular names

api/v1alpha1/shared_types.go
arkodg: +1 for `request.body`

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -836,3 +863,81 @@ func makeIrTriggerSet(in []egv1a1.TriggerEnum) []ir.TriggerEnum {
...
+		}
+
+		response := ir.CustomResponse{
+			ContentType: ro.Response.ContentType,
+		}
+
+		if ro.Response.Body.Type != nil && *ro.Response.Body.Type == egv1a1.ResponseValueTypeValueRef {
+			foundCM := false
+			for _, cm := range configMaps {
+				if cm.Namespace == policy.Namespace && cm.Name == string(ro.Response.Body.ValueRef.Name) {
```
arkodg, line 900: is this a ptr, will it work for `default` ?

internal/gatewayapi/testdata/backendtrafficpolicy-with-response-override-invalid-valueref.in.yaml
```diff
@@ -0,0 +1,141 @@
...
+          backendRefs:
+            - name: service-1
+              port: 8080
+configMaps:
+  - apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: response-override-config
+      namespace: default
+    data: {}
```
arkodg, line 71: can we add some dummy data in here

internal/xds/translator/testdata/out/xds-ir/custom-response.listeners.yaml
```diff
@@ -0,0 +1,130 @@
...
+                            typedConfig:
+                              '@type': type.googleapis.com/envoy.type.matcher.v3.HttpResponseStatusCodeMatchInput
+                          valueMatch:
+                            exact: "500"
+                      - singlePredicate:
+                          customMatch:
+                            name: cel-matcher
+                            typedConfig:
+                              '@type': type.googleapis.com/xds.type.matcher.v3.CelMatcher
+                              exprMatch:
```
arkodg, line 74: 👀

arkodg: trusting the e2e, this API is complex, thanks for implementing this !

arkodg (approved):
LGTM ! implementing the feature using the custom response filter doesnt look easy, thanks for building this out

internal/gatewayapi/backendtrafficpolicy.go
zhaohuabing: if *ro.Response.Body.Type == egv1a1.ResponseValueTypeValueRef, then ro.Response.Body.ValueRef is not nil, otherwise, it would be a bug in Gateway API translation.

internal/gatewayapi/testdata/backendtrafficpolicy-with-response-override-invalid-valueref.in.yaml
zhaohuabing: This is a test for invalid configMap where the data is nil.



Output:
```
message: 'ResponseOverride: can''t find the key response.body in the referenced
          configmap response-override-config.'
```

internal/xds/translator/testdata/out/xds-ir/custom-response.listeners.yaml
zhaohuabing: The Envoy CEL matcher API is terrible :-) It's the AST for CEL:  `response.code >= 501 && response.code <= 511 `.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -124,6 +124,8 @@ type EnvoyProxySpec struct {
 	//
 	// - envoy.filters.http.ratelimit
 	//
+	// - envoy.filters.http.custom_response
```
zirain, line 127: is it possible to override the 429 response if we place custom response filter behind ratelimit filterm should this filter be the first one of http filters?

zhaohuabing: Good point. Should we allow users to override the ratelimit response if they added a matching conditing for 429 in the `ResponseOrverride`? @envoyproxy/gateway-maintainers

zirain: it's not just ratelimit, does authz/authn has same problem?

zhaohuabing: Yeah, I've added this to the last because users may want to override the responses of other filters, if they explicitly adding these status codes in the matching conditions.

zirain: IMO, it's better to put it at the first but no strong option, @arkodg please chime in as API designer.

arkodg: as a user, I would want the custom response to kick in after all filters have finished processing, to unify the way I deal with status code and error pages, so +1 to adding it to the end

zirain: we can change the order in the future, maybe we can open a ticket to continue our discussion.

zirain:
@zhaohuabing can you make CI happy?

zirain (approved):

arkodg (approved):

arkodg:
e2es are failing ...

api/v1alpha1/envoyproxy_types.go
zhaohuabing: > we can change the order in the future, maybe we can open a ticket to continue our discussion.

Let's keep this at the last since there're use cases to override the responses of other filters. Users can change the default order using `EnvoyProxy` if they want to reorder the filters.

arkodg (approved):

zirain (approved):

PR #4422, opened 2024-10-10
fix dashboard typos

### LINKED ISSUE #4421: bug: Some typos in grafana dashboard
*Description*:
returned, etc.

*Repro steps*:

Upstream Bytes Tx/Second uses `RX` metric by mistake.
image

Allocated memroy uses `SUM` by mistake (should be lower `sum`), causes dashboard crash when being edited.
image

sharing.

*Environment*:

*Logs*:

haorenfsa:
**What type of PR is this?**

fix: some typos in Grafana dashboard.

Fixes #4421

### DIFF
--- a/charts/gateway-addons-helm/dashboards/envoy-proxy-global.json
+++ b/charts/gateway-addons-helm/dashboards/envoy-proxy-global.json
@@ -1849,7 +1849,7 @@
             "uid": "${datasource}"
           },
           "editorMode": "builder",
-          "expr": "sum by(namespace) (rate(envoy_cluster_upstream_cx_rx_bytes_total{namespace=~\"$Namespace\"}[5m]))",
+          "expr": "sum by(namespace) (rate(envoy_cluster_upstream_cx_tx_bytes_total{namespace=~\"$Namespace\"}[5m]))",
           "instant": false,
           "legendFormat": "{{namespace}}",
           "range": true,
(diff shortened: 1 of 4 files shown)


zirain:
IIRC, that json generated from jsonnet, can you update it and then run `make -k gen-check`?

cc @shawnh2

charts/gateway-addons-helm/dashboards/envoy-proxy-global.json
```diff
@@ -1849,7 +1849,7 @@
             "uid": "${datasource}"
           },
           "editorMode": "builder",
-          "expr": "sum by(namespace) (rate(envoy_cluster_upstream_cx_rx_bytes_total{namespace=~\"$Namespace\"}[5m]))",
+          "expr": "sum by(namespace) (rate(envoy_cluster_upstream_cx_tx_bytes_total{namespace=~\"$Namespace\"}[5m]))",
```
shawnh2, line 1852: good catch!

shawnh2 (dismissed):

shawnh2:
> IIRC, that json generated from jsonnet, can you update it and then run `make -k gen-check`?
> 
> 
> 
> cc @shawnh2 

not really, these json are legacy dashboards, not generated from jsonnet.

haorenfsa:
> please run make -k gen-check to make the CI pass.
@shawnh2 done

shawnh2 (approved):
LGTM! Thanks

zirain (approved):

PR #4427, opened 2024-10-12
feat: add labels to envoyService config

### LINKED ISSUE #4417: Support labels configuration in EnvoyProxy spec
*Description*:
would be used.
I'd like the ability to configure labels on the `envoyService` spec within the EnvoyProxy resource so that I can do things like target a specific cilium IP pool.

Would it be accepted if I made this change in https://github.com/envoyproxy/gateway/blob/d1cc0ba4372fd7d3f8ede1a1278930e21c22b9f2/api/v1alpha1/shared_types.go#L258?

[optional *Relevant Links*:]

jukie:
**What type of PR is this?**

feat: add labels to envoyService config

This adds the ability to configure labels that should specifically be applied to the service within an EnvoyProxy spec. This is useful for doing things like selecting a cilium IP pool or other labels that a user _doesn't_ want to be applied to all related resources via the existing [infra labels](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/gateway_types.go#L722) option.

Fixes #4417


jukie:
@arkodg @zetaab whenever you're able to take a look please

zetaab (approved):

zetaab:
ah actually you need to generate crds again. could you run `make manifests`

zetaab (approved):

zirain (approved):

shawnh2 (approved):
LGTM

PR #4429, opened 2024-10-12
feat: allow running EnvoyProxy as DaemonSet

### LINKED ISSUE #2969: Run Envoy Gateway like DaemonSet

### LINKED ISSUE #4416: Gateways never become ready/programmed when running daemonset EnvoyProxy
*Description*:
returned, etc.

I'm trying to run envoy proxy as a daemonset which was implemented with https://github.com/envoyproxy/gateway/pull/3092.
I think the cause of this bug is https://github.com/envoyproxy/gateway/blob/d1cc0ba4372fd7d3f8ede1a1278930e21c22b9f2/internal/gatewayapi/status/gateway.go#L163-L168 and doing a lookup on deployment vs daemonset.

*Repro steps*:

Use the below configuration with an EnvoyProxy to trigger using a daemonset:
```
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
spec:
  mergeGateways: true
  provider:
    type: Kubernetes
    kubernetes:
      envoyDaemonSet:
        strategy:
          type: RollingUpdate
---
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: custom-proxy-config
---
kind: Gateway
metadata:
  name: eg
  namespace: gateway-demo
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
```

Then the programmed status will always show as false even though the gateway is actually functional:
```
Status:
  Conditions:
    Last Transition Time:  2024-10-09T16:12:58Z
    Message:               Deployment replicas unavailable
    Observed Generation:   1
    Reason:                NoResources
    Status:                False
    Type:                  Programmed
```
sharing.

*Environment*:

*Logs*:

jukie:
**What type of PR is this?**
feat: allow running EnvoyProxy as DaemonSet

Continuation of  #3092 to finish all daemonset implementations and fix a bug where the gateway status never gets updated because the code only checks for deployments.

Fixes #4416
Fixes #2969


internal/gatewayapi/status/gateway.go
```diff
@@ -159,15 +160,22 @@ func updateGatewayProgrammedCondition(gw *gwapiv1.Gateway, deployment *appsv1.De
 
 	// If there are no available replicas for the Envoy Deployment, don't
 	// mark the Gateway as ready yet.
-
-	if deployment == nil || deployment.Status.AvailableReplicas == 0 {
+	dep, okDep := envoyObj.(*appsv1.Deployment)
```
zirain, line 163: can we simply this with switch?

jukie: Thanks! Fixed

jukie:
Should actually work this time 🤞

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -294,12 +297,105 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
...
+					Kind:       ResourceKindDaemonset,
+					APIVersion: appsAPIVersion,
+					Name:       "envoy-gateway",
+					UID:        uid,
+				},
+			}
+		}
+	}
+
+	// apply merge patch to deployment
```
shawnh2, line 392: apply merge patch to daemonset

jukie: I'll do another check on the rest of the comments referencing `deployment` now

jukie: Fixed this one and cleaned up some otherse

shawnh2 (commented):
Hi @jukie, thanks for implementing the rest part of daemonset deploy mode, code generally LGTM!

Got a question about EG and RateLimit both in daemonset settings: the EG and RL pod will be running on each node, does it mean the global rate limiting becomes per-node rate limiting?

api/v1alpha1/envoygateway_types.go
```diff
@@ -202,6 +202,13 @@ type EnvoyGatewayKubernetesProvider struct {
 	// +optional
 	RateLimitDeployment *KubernetesDeploymentSpec `json:"rateLimitDeployment,omitempty"`
 
+	// RateLimitDaemonset defines the desired state of the Envoy ratelimit daemonset resource.
+	// If unspecified, default settings for the managed Envoy ratelimit daemonset resource
+	// are applied.
+	//
+	// +optional
+	RateLimitDaemonset *KubernetesDaemonSetSpec `json:"rateLimitDaemonset,omitempty"`
```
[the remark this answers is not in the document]
jukie: I might be mistaken but they should be able to exist in either mode alongside each other. That said, I can't think of a use-case for running mixed (maybe EG daemonset and ratelimit deployment?) so adding a check to prevent that makes sense. I'll look into that more tomorrow if you'd like.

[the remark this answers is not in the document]
jukie: > the EG and RL pod will be running on each node, does it mean the global rate limiting becomes per-node rate limiting?

I'm not currently using the feature but reading through the docs and code leads me to think that nothing would change. RL pods all share the same Redis backend so it shouldn't become a per-node rate limit or adjust any logic.

Ref:
- https://gateway.envoyproxy.io/latest/tasks/traffic/local-rate-limit/
- https://gateway.envoyproxy.io/latest/tasks/traffic/global-rate-limit/

jukie:
I'm having a bit of trouble trying to dig into the e2e failures for 1.31.
It consistently fails, but on a different test each time due to context deadline exceeded (`TestE2E/OIDC/http_route_without_oidc_authentication` or `TestMergeGateways/BasicMergeGateways/apply_a_gateway_with_conflicted_listener_and_a_httproute`). I'll try bumping up the defaults to see if anything changes but I don't see what part of my code would cause a significant increase in processing time.

charts/gateway-helm/templates/_rbac.tpl
```diff
@@ -43,6 +43,7 @@ apiGroups:
 - apps
 resources:
 - deployments
+- daemonsets
```
zirain, line 46: nice catch

zirain (approved):

arkodg (approved):
LGTM thanks !

PR #4433, opened 2024-10-13
feat: support setting custom scheduler profile in podSpec

### LINKED ISSUE #4432: Support for using custom kube-scheduler
*Description*:
would be used.

I'd like to use a custom scheduler in the podSpec of envoy-gateway resources

[optional *Relevant Links*:]

jukie:
**What type of PR is this?**

feat: support setting custom scheduler profile in podSpec

Adds support for configuring a custom kube-scheduler profile in podSpec

Ref: https://kubernetes.io/docs/reference/scheduling/config/
 
    Pods that want to be scheduled according to a specific profile can include the corresponding scheduler name in its .spec.schedulerName.
    By default, one profile with the scheduler name default-scheduler is created. This profile includes the default plugins described above. When declaring more than one profile, a unique scheduler name for each of them is required.
    If a Pod doesn't specify a scheduler name, kube-apiserver will set it to default-scheduler. Therefore, a profile with this scheduler name should exist to get those pods scheduled.

Fixes #4432


api/v1alpha1/shared_types.go
```diff
@@ -181,6 +181,13 @@ type KubernetesPodSpec struct {
 	//
 	// +optional
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// SchedulerName is the kubernetes scheduler that should be used for this pod.
+	// If specified, the pod will be dispatched by specified scheduler.
+	// If not specified, the pod will be dispatched by default scheduler.
+	//
+	// +optional
+	SchedulerName string `json:"schedulerName,omitempty"`
```
zirain, line 190: if it's optional, make it a pointer.

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -248,7 +248,7 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 					TerminationGracePeriodSeconds: ptr.To[int64](300),
 					DNSPolicy:                     corev1.DNSClusterFirst,
 					RestartPolicy:                 corev1.RestartPolicyAlways,
-					SchedulerName:                 "default-scheduler",
+					SchedulerName:                 r.rateLimitDeployment.Pod.SchedulerName,
```
zirain, line 251: if it's empty, use `default-scheduler` as default value?

jukie: That's already added by kubernetes when left empty. Wouldn't it be preferable to simply leave this field empty if not set by the user?

zirain: If they mean the same, let's be consistent with before for compatibility, WDYT?

zirain: will this change trigger a rollout of existing deployment/daemonset?

jukie: If a user has a mutating hook to inject a custom scheduler, it can become problematic when gateway tries to reconcile this resource and inject something else. By leaving it empty that allows for a clean merge.

This change will not trigger a rollout as kubernetes is already adding if this field is empty. I've confirmed this locally as well by removing the scheduler field from a deployment and daemonset resource.

api/v1alpha1/shared_types.go
jukie: This matches the core [PodSpec API](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec) and an empty string would already be processed as such. I can still change it if you'd like but aligning with core seemed preferable.

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
zirain: if it will not trigger a rollout, lgtm.

api/v1alpha1/shared_types.go
zirain: let's keep same as other optional fileds in this repo.

jukie: One problem with using a pointer is then we need a nil check or to set a default. For fields like SecurityContext, the K8s PodSpec type already passes through as a pointer and is handled there but for SchedulerName it wants a string.

jukie: Updated as a pointer

jukie: Should be resolved now @zirain

zirain (approved):

arkodg (changes requested):
this field doesnt seem like a common use case
you can still enable this using the `patch` field https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/#customize-envoyproxy-with-patches

jukie:
Sounds good

Issue #4181, opened 2024-09-08
Envoy gateway runs oom


hemenghong:
*Description*:
> The following is an envoy gateway flame diagram. Whether DeepCopy has problems?
<img width="2515" alt="0880c4c80710a29fdf24" src="https://github.com/user-attachments/assets/cfae0431-da04-41d0-8925-dcb073435e92">

*Repro steps*:
> After the EG runs for a period of time, the oom will appear

*Environment*:
>the latest version

shawnh2:
may be releated to https://github.com/envoyproxy/gateway/issues/3698

arkodg:
hey @hemenghong how many relevant resources (xRoutes, Gateway and related Secrets, Services etc) have been  applied ?
the default limits value is `1G`, please consider increasing it in the helm chart

hemenghong:
> hey @hemenghong how many relevant resources (xRoutes, Gateway and related Secrets, Services etc) have been applied ? the default limits value is `1G`, please consider increasing it in the helm chart

eg limit mem has been set to 2G.  
httproute has 90 and node has 100 and secret has 900 and service has 270 and endpointslice has 270.

arkodg:
closing this issue in favor of https://github.com/envoyproxy/gateway/issues/3860

PR #4454, opened 2024-10-16
bugfix: filter repeated resources and optimize memory usage.

### LINKED ISSUE #4395: `endpoints` repeated
Why `endpoints` repeated ? 
image

EnvoyGateway: v1.1.1

Here is my backend-traffic-policy:
image

qicz:
**What type of PR is this?**

Fixes #4395 and optimize for #4181 

before:
image
image
image

after
image
image
image

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -131,47 +131,6 @@ func newGatewayAPIController(mgr manager.Manager, cfg *config.Server, su Updater
 	return nil
 }
 
-type resourceMappings struct {
-	// Map for storing namespaces for Route, Service and Gateway objects.
-	allAssociatedNamespaces sets.Set[string]
-	// Map for storing EnvoyProxies' NamespacedNames attaching to Gateway or GatewayClass.
-	allAssociatedEnvoyProxies sets.Set[string]
-	// Map for storing TLSRoutes' NamespacedNames attaching to various Gateway objects.
-	allAssociatedTLSRoutes sets.Set[string]
-	// Map for storing HTTPRoutes' NamespacedNames attaching to various Gateway objects.
-	allAssociatedHTTPRoutes sets.Set[string]
-	// Map for storing GRPCRoutes' NamespacedNames attaching to various Gateway objects.
-	allAssociatedGRPCRoutes sets.Set[string]
-	// Map for storing TCPRoutes' NamespacedNames attaching to various Gateway objects.
-	allAssociatedTCPRoutes sets.Set[string]
-	// Map for storing UDPRoutes' NamespacedNames attaching to various Gateway objects.
-	allAssociatedUDPRoutes sets.Set[string]
-	// Map for storing backendRefs' NamespaceNames referred by various Route objects.
-	allAssociatedBackendRefs sets.Set[gwapiv1.BackendObjectReference]
-	// extensionRefFilters is a map of filters managed by an extension.
-	// The key is the namespaced name, group and kind of the filter and the value is the
-	// unstructured form of the resource.
-	extensionRefFilters map[utils.NamespacedNameWithGroupKind]unstructured.Unstructured
-	// httpRouteFilters is a map of HTTPRouteFilters, where the key is the namespaced name,
-	// group and kind of the HTTPFilter.
-	httpRouteFilters map[utils.NamespacedNameWithGroupKind]*egv1a1.HTTPRouteFilter
-}
-
-func newResourceMapping() *resourceMappings {
-	return &resourceMappings{
-		allAssociatedNamespaces:   sets.New[string](),
-		allAssociatedEnvoyProxies: sets.New[string](),
-		allAssociatedTLSRoutes:    sets.New[string](),
-		allAssociatedHTTPRoutes:   sets.New[string](),
-		allAssociatedGRPCRoutes:   sets.New[string](),
-		allAssociatedTCPRoutes:    sets.New[string](),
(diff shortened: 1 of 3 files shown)


zirain:
we need some test to guratee this

qicz:
> @qicz thanks for picking this one up, its worth investigating _why_ this happening before working around it by using a set

for my analysis, the `endpointslice` with more repeated value for the service

https://github.com/envoyproxy/gateway/blob/aab669dd99f0337afaaa7c22f7826d9ebd37cfeb/internal/provider/kubernetes/controller.go#L434-L455

logging:

image

arkodg:
ah thanks for pointing to the controller/provider code, it looks like since the service was referenced twice, we added the corresponding endpoint slice twice. The fix in this PR is not going to work since it has visibility into a single EndpointSlice

We can either fix this in the provider or in the gateway api layer

qicz:
> ah thanks for pointing to the controller/provider code, it looks like since the service was referenced twice, we added the corresponding endpoint slice twice. The fix in this PR is not going to work since it has visibility into a single EndpointSlice
> 
> We can either fix this in the provider or in the gateway api layer

I 'll fix this in gateway API layer

qicz:
I tried many times but only followed resources to support duplicate resource check logic under the current implementation.

```go
var cacheKinds = []string{
	"Gateway", "ReferenceGrant", "Namespace",
	"ServiceImport", "EndpointSlice",
	"Secret", "ConfigMap", "EnvoyPatchPolicy",
	"ClientTrafficPolicy", "BackendTrafficPolicy", "SecurityPolicy",
	"BackendTLSPolicy", "EnvoyExtensionPolicy", "HTTPRouteFilter",
}
```

internal/gatewayapi/resource/resource_cache.go
```diff
@@ -0,0 +1,85 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package resource
+
+// supportedCacheResourceKinds that are supported cache manager.
+// if some resources are supported remove the code slash comment.
+var supportedCacheResourceKinds = []string{
+	"EnvoyProxy",
```
shawnh2, line 11: can we reuse the const from `internal/gatewayapi/resource/supported_kind.go` ?

shawnh2 (commented):
thanks for the fix @qicz, now the mem usage decrease dramatically !

image

qicz:
> thanks for the fix @qicz, now the mem usage decrease dramatically !
> 
> image

yep, we are filtering the duplicate resources that will optimize the memory for many big objects used.
cc @arkodg

arkodg:
looks like we are already doing something similar in the provider layer using [this structure](https://github.com/envoyproxy/gateway/blob/deddeb472fa7307bf6e83f782d02ec2eba828bf1/internal/provider/kubernetes/controller.go#L160) to ensure deduplication, can we implement this there instead ?

qicz:
> looks like we are already doing something similar in the provider layer using [this structure](https://github.com/envoyproxy/gateway/blob/deddeb472fa7307bf6e83f782d02ec2eba828bf1/internal/provider/kubernetes/controller.go#L160) to ensure deduplication, can we implement this there instead ?

They are similar,but cache used in controller and translate, IMO should in gateway layer resource pkg. resourcemap just for controller @arkodg

arkodg:
> > looks like we are already doing something similar in the provider layer using [this structure](https://github.com/envoyproxy/gateway/blob/deddeb472fa7307bf6e83f782d02ec2eba828bf1/internal/provider/kubernetes/controller.go#L160) to ensure deduplication, can we implement this there instead ?
> 
> They are similar,but cache used in controller and translate, IMO should in gateway layer resource pkg. resourcemap just for controller @arkodg

since we already have an existing way to do this, prefer the provider approach to reduce code complexity

qicz:
> > > looks like we are already doing something similar in the provider layer using [this structure](https://github.com/envoyproxy/gateway/blob/deddeb472fa7307bf6e83f782d02ec2eba828bf1/internal/provider/kubernetes/controller.go#L160) to ensure deduplication, can we implement this there instead ?
> > 
> > 
> > They are similar,but cache used in controller and translate, IMO should in gateway layer resource pkg. resourcemap just for controller @arkodg
> 
> since we already have an existing way to do this, prefer the provider approach to reduce code complexity

you mean reuse resourcemap struct? or reuse  resourcemap set properties? @arkodg

qicz:
> > > > > looks like we are already doing something similar in the provider layer using [this structure](https://github.com/envoyproxy/gateway/blob/deddeb472fa7307bf6e83f782d02ec2eba828bf1/internal/provider/kubernetes/controller.go#L160) to ensure deduplication, can we implement this there instead ?
> > > > 
> > > > 
> > > > They are similar,but cache used in controller and translate, IMO should in gateway layer resource pkg. resourcemap just for controller @arkodg
> > > 
> > > 
> > > since we already have an existing way to do this, prefer the provider approach to reduce code complexity
> > 
> > 
> > you mean reuse resourcemap struct? or reuse resourcemap set properties? @arkodg
> 
> create an addition field within `resourceMappings` for service to ensure we add one unique item into the list

ok ,I'll try it but  maybe  will  occur cyle imports , translate pkg and controller cycle using.

arkodg:
@qicz my suggestion is to solve the problem in the provider package similar to what we are doing for other resources

arkodg (approved):
LGTM thanks !
great to see the massive drop in mem usage

zirain (approved):

PR #4460, opened 2024-10-16
fix: disable ALPN for non-HTTP routes

### LINKED ISSUE #4456: default ALPN settings break TLS-terminated TCP route
*Description*:

Envoy Gateway sets ALPN to `h2, http/1.1` by default. When Envoy is not handling the application layer protocol (e.g. TLS termination for TCP), Envoy should not advertise application layer protocols that may not be supported by the backend, and keep the downstream TLS socket ALPN settings undefined. 

In general, it seems that Envoy Gateway should perhaps support more flexibility around ALPN:
- Supporting non-HTTP ALPN protos for TLS routes that proxy non-HTTP protos
- Supporting ALPN disablement
- Changing the defaults for non-HTTP routes

*Repro steps*:
Follow the the official guide: https://gateway.envoyproxy.io/docs/tasks/security/tls-termination/

```
curl -v -HHost:www.example.com --resolve "www.example.com:8443:127.0.0.1"  --cacert example.com.crt https://www.example.com:8443/get
* Added www.example.com:8443:127.0.0.1 to DNS cache
* Hostname www.example.com was found in DNS cache
*   Trying 127.0.0.1:8443...
* Connected to www.example.com (127.0.0.1) port 8443
* ALPN: server accepted h2
[...]
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://www.example.com:8443/get
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: www.example.com]
* [HTTP/2] [1] [:path: /get]
* [HTTP/2] [1] [user-agent: curl/8.7.1]
* [HTTP/2] [1] [accept: */*]
> GET /get HTTP/2
> Host:www.example.com
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
* Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly.
```

Envoy encourages the client the use HTTP/2, which is not supported by the upstream server, leading to a failure. If the client is explicitly using HTTP/1, the request succeeds:

```
curl -v --http1.1 -HHost:www.example.com --resolve "www.example.com:8443:127.0.0.1" --cacert example.com.crt https://www.example.com:8443/get
[...]
* using HTTP/1.x
> GET /get HTTP/1.1
> Host:www.example.com
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK

```

guydc:
**What type of PR is this?**

- By default, all non-HTTPS routes opt-out of ALPN 
- If a client traffic policy defines ALPN protocols, non-HTTPS routes are opted-in 
- No changes to upstream ALPN settings 

Fixes #4456


internal/ir/xds.go
```diff
@@ -355,6 +355,8 @@ type TLSConfig struct {
 	StatelessSessionResumption bool `json:"statelessSessionResumption,omitempty" yaml:"statelessSessionResumption,omitempty"`
 	// StatefulSessionResumption determines if stateful (session-id based) session resumption is enabled
 	StatefulSessionResumption bool `json:"statefulSessionResumption,omitempty" yaml:"statefulSessionResumption,omitempty"`
+	// ALPNDisabled determined if ALPN extension is disabled
+	ALPNDisabled bool `json:"alpnDisabled,omitempty" yaml:"alpnDisabled,omitempty"`
```
arkodg, line 359: can we reuse `ALPNProtocols` ? and treat `nil` as a default and empty as disabled ?

liorokman: +1 on reusing `ALPNProtocols` instead of adding a new `ALPNDisabled` attribute.

internal/gatewayapi/helpers.go
```diff
@@ -436,6 +436,12 @@ func irTLSConfigs(tlsSecrets ...*corev1.Secret) *ir.TLSConfig {
 			PrivateKey:  tlsSecret.Data[corev1.TLSPrivateKeyKey],
 		}
 	}
+
+	// explicitly disable ALPN by setting an empty list for non-HTTPS use cases
+	if protocol != gwapiv1.HTTPSProtocolType {
+		tlsListenerConfigs.ALPNProtocols = ptr.To([]string{})
```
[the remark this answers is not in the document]
zirain: can you addrees it with comment.

internal/xds/translator/translator.go
```diff
@@ -896,7 +896,7 @@ func buildXdsUpstreamTLSSocketWthCert(tlsConfig *ir.TLSUpstreamConfig) (*corev3.
 		tlsCtx.CommonTlsContext.TlsParams = tlsParams
 	}
 
-	if len(tlsConfig.ALPNProtocols) > 0 {
+	if tlsConfig.ALPNProtocols != nil {
```
liorokman, line 899: I think it would be clearer to change `buildALPNProtocols` to return an empty `[]string` if `tlsConfig.ALPNProtocols` is nil. There's no need to check the length of the `tlsConfig.ALPNProtocols` array here - it's already checked in the `buildALPNProtocols` function.

internal/ir/xds.go
```diff
@@ -350,7 +350,7 @@ type TLSConfig struct {
 	// SignatureAlgorithms supported by this listener
 	SignatureAlgorithms []string `json:"signatureAlgorithms,omitempty" yaml:"signatureAlgorithms,omitempty"`
 	// ALPNProtocols exposed by this listener
-	ALPNProtocols []string `json:"alpnProtocols,omitempty" yaml:"alpnProtocols,omitempty"`
+	ALPNProtocols *[]string `json:"alpnProtocols,omitempty" yaml:"alpnProtocols,omitempty"`
```
arkodg, line 353: is the ptr needed ? the zero value of a slice is `nil`

guydc: Sure. If we keep omitEmpty, the .out files would omit alpnProtocols both for nil and empty slices, so that can make IR/XDS test reading a bit confusing. If we remove it, tests are a bit inflated with `alpnProtocols: null`. Any preference?

arkodg: ah thanks for explaining why you chose this route :) even if .out files omit alpnProtocols, the test title (-empty-alpn) should hopefully solve the confusion imo

guydc: ok, my bad, it's not just a readability issue. When .out files are parsed, all the omitted slices are assumed to be nil, while the IR structs that they are are compared to might be an empty slice (e.g. for TLSRoutes). The .out testdata would always be wrongly generated in this case. 
So, options are:
- Don't omitempty
- Customize JSON marshaling with additional tags, methods
- Pointer to slice

I added the first option in a new commit.

internal/gatewayapi/helpers.go
```diff
@@ -436,6 +436,16 @@ func irTLSConfigs(tlsSecrets ...*corev1.Secret) *ir.TLSConfig {
...
+
+	// Envoy Gateway defaults for ALPN:
+	// 1. HTTPS Routes: set IR to nil slice, which leads to xds-level
+	//    default of: h2, http/1.1. Envoy Proxy will support both HTTP2 and HTTP1.
+	// 2. Other xRoutes: set to empty list, disabling ALPN extension for Envoy Proxy.
+	//    ALPN cannot be used to negotiate application layer protocols like HTTP2.
+	if protocol != gwapiv1.HTTPSProtocolType {
+		tlsListenerConfigs.ALPNProtocols = []string{}
+	}
+
```
zhaohuabing, line 448: Nit: we probably can make this easier to understand by:
 * Adding a line to the comments to explain the default behavior of the ALPNProtocols in the connection to the backend
 
 Or 
   * Adding another method 
   
   ```
   irBackendConnectionTLSConfigs(tlsSecrets ...*corev1.Secret) *ir.TLSConfig {
     return irTLSConfigs("", tlsSecrets)
   ```

zhaohuabing (approved):
LGTM thanks!

liorokman (approved):
LGTM

PR #4461, opened 2024-10-16
feat: implement fallback for the Backend API

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/3055

### DIFF
--- a/api/v1alpha1/backend_types.go
+++ b/api/v1alpha1/backend_types.go
@@ -131,7 +131,6 @@ type BackendSpec struct {
 	// the health of the active backends falls below 72%.
 	//
 	// +optional
-	// +notImplementedHide
 	Fallback *bool `json:"fallback,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


zhaohuabing (approved):

api/v1alpha1/backend_types.go
```diff
@@ -131,7 +131,6 @@ type BackendSpec struct {
 	// the health of the active backends falls below 72%.
 	//
 	// +optional
-	// +notImplementedHide
```
zhaohuabing, line 134: Nit: a link to https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority#arch-overview-load-balancing-priority-levels will help users to understand how fallback works.

arkodg: we dont want to expose the internal details, only want to represent it as a boolean field

zirain (approved):

Issue #2193, opened 2023-11-15
AND NOT style matches for RateLimiting


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Consider this config
```
cat <<EOF | kubectl apply -f -
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  name: jwt-backend
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: backend
  jwt:
    providers:
    - name: example
      remoteJWKS:
        uri: https://test.eu.auth0.com/.well-known/jwks.json
      claimToHeaders:
      - claim: email
        header: x-user-email
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy 
metadata:
  name: policy-httproute
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: backend
    namespace: envoy
  rateLimit:
    type: Global
    global:
      rules:
      - clientSelectors:
        - headers:
          - name: x-user-email
            value: myemail@gmail.com
        limit:
          requests: 10
          unit: Minute
    - clientSelectors:
      - headers:
        - type: Distinct
          name: x-claim-email
      limit:
        requests: 5
        unit: Minute
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
      namespace: envoy
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: ""
          kind: Service
          name: backend
          port: 3000
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /
EOF
```

I want to allow [myemail@gmail.com](mailto:myemail@gmail.com) user to 10 req/min but all other user 5 req/min.
But with this config [myemail@gmail.com](mailto:myemail@gmail.com) user is getting 429 after 5 requests.

Relates to https://envoyproxy.slack.com/archives/C03E6NHLESV/p1699784524023299

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
thinking out loud, this use case can be achieved in two ways
1.  by adding a `dontMatch` field

```
    - clientSelectors:
      - headers:
        - type: Distinct
          name: x-claim-email
        - type: Exact
           name: x-claim-email
           dontMatch: true
[...]

PR #4464, opened 2024-10-17
api: fix HeaderMatch list type to allow invert matches on same header

rudrakhp:
**What type of PR is this?**

api: fix HeaderMatch list type to allow invert matches on same header

To allow header match rate limit filters discussed [here](https://github.com/envoyproxy/gateway/issues/2193#issuecomment-2417372126).

Ref #2193


arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #4424, opened 2024-10-11
Allow WASM extensions configuration from secret


daniel-white:
*Description*:
>I'm currently using the `config` option for a Wasm extension. I have a sensitive value that I don't want explicitly defined in my chart or as a value. It would be great if I could configure the Wasm extension with a secret.

[optional *Relevant Links*:]
>Wasm object definition: https://gateway.envoyproxy.io/docs/api/extension_types/#wasm

zhaohuabing:
Ideally the sensitive value should be a `Secret` reference in the `config`, and the only the Wasm module should be able to access it.

PR #4470, opened 2024-10-17
feat(translator): allow configuration of hostEnvKeys on WASM extensions

sgargan:
exposes the hostEnvKeys configuration for WASM extensons through envoy extension policies. This enables access to env vars that are set on the host envoy processes and is a convenient way to share secret meterial with WASM extensions.

Release Note: Yes


daniel-white:
this may solve my #4424

internal/ir/xds.go
```diff
@@ -2519,6 +2519,10 @@ type Wasm struct {
 	// original URL(either an HTTP URL or an OCI image) and serves it through the
 	// local HTTP server.
 	Code *HTTPWasmCode `json:"httpWasmCode,omitempty"`
+
+	// HostEnvKeys is a list of keys for environment variables from the host envoy process
+	// that should be passed into the Wasm VM.
+	HostEnvKeys []string `json:"hostEnvKeys,omitempty"`
```
arkodg, line 2525: `hostEnvKeys` is pretty self explanatory, also brainstorming some more options
* `vmConfig.envKeys`
* `env.Keys`
* `envKeys`

arkodg:
hey @sgargan, the PR looks great ! thanks for adding docs and test cases

added one comment around naming, if we get consensus from @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers on naming, we could try and get this into the v1.2.0-rc.1 release (late next week)

internal/ir/xds.go
arkodg: also linking some past work here from istio https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/#VmConfig

arkodg:
`make geo-check` is failing, you can run `make generate && make manifests` locally and commit those changes

site/content/en/docs/tasks/extensibility/wasm.md
```diff
@@ -151,9 +151,146 @@ spec:
...
+kubectl get envoyextensionpolicy/wasm-test-o yaml
+```
+
+### Wasm Extension Configuration through Environment variables
+
+It is also possible to configure a wasm extension using environment variables from the host envoy process. Keys for the env vars to be shared are defined in a `hostEnvKeys` block.
+
+This is especially useful for sharing secure data from environment vars on the envoy process set using [valueFrom](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables) a Kubernetes secret.
+
+Note that setting an env var on the envoy process requires a custom [EnvoyProxy](../../api/extension_types#envoyproxy) configuration.
```
daniel-white, line 234: it would be great to show an example here!

internal/ir/xds.go
zhaohuabing: +1 adding a `vmConfig` wrapper for more future proof as other VM settings may be required in the future.
Should we also consider allowing [explicitly specified key-value pairs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#envoy-v3-api-msg-extensions-wasm-v3-environmentvariables).

```
vmConfig:
   env:
      hostEnvKey:
        - host_key_1
        - host_key_2
      cutomEnv:
        - key:  custom_key_1
           value: custom_value_1 
         - key:  custom_key_2
           value: custom_value_2    
```

arkodg: hey @zhaohuabing what value does customEnvs add that cannot be represented via `config` ?

zhaohuabing: Yeah, customEnvs just gives users more options for passing configurations into the Wasm module.

sgargan: `hostEnvKeys` is the name that envoy used for this and I figured it would make it easier for someone to join the dots from the envoy config if it matched here.

For a vmConfig wrapper, would this just contain the `hostEnvKeys` or the `config` block too? happy to add it, but is there value to the extra nesting?

site/content/en/docs/tasks/extensibility/wasm.md
sgargan: 👍 I've added an example to the envoyproxy customization docs and referenced it here.

internal/ir/xds.go
arkodg: yeah I'm a +1 for `hostEnvKeys`, it leaves room for `customEnvs` . The other way to frame this is `env.HostKeys`, which leaves room for `env.Custom`.

zhaohuabing:
After discussing with @arkodg and @guydc in today's EG meeting, it was suggested to consider using Kubernets Secrets for the `CustomEnvs` to securely pass sensitive data to the Wasm modules. Additionally, ConfigMap should be supported as a storage option for non-sensitive configurations.

As the next step, I suggest raising a PR for the API design, followed by a follow-up PR for the implementation, once the community reaches a consensus on the API.

arkodg:
> @sgargan After discussing with @arkodg and @guydc in today's EG meeting, it was suggested to consider using Kubernets Secrets for the `CustomEnvs` to securely pass sensitive data to the Wasm modules. Additionally, ConfigMap should be supported as a storage option for non-sensitive configurations.
> 
> As the next step, I suggest raising a PR for the API design, followed by a follow-up PR for the implementation, once the community reaches a consensus on the API.
> 
> A similar example can be found in the `CustomResponseBody` API:
> 
> https://github.com/envoyproxy/gateway/blob/7188dad953db69490907431c290dcaf489124e2a/api/v1alpha1/shared_types.go#L706-L729

rethinking this one @guydc @zhaohuabing we can't allow the app dev who has access to the EEP to inject any env vars into the proxy, so we'll need to split up the task of injecting env vars (using `EnvoyProxy` resource by the platform admin) and allowing the app dev to use it by specifying this field

sgargan:
> rethinking this one @guydc @zhaohuabing we can't allow the app dev who has access to the EEP to inject any env vars into the proxy, so we'll need to split up the task of injecting env vars (using `EnvoyProxy` resource by the platform admin) and allowing the app dev to use it by specifying this field

there is definitely the danger that a dev with access to the EEP could exfil secrets that they shouldn't have access to if they could reason about what keys to use. This is mostly a problem if the EP is shared 

Perhaps the EEP should only be able to use a secret from the same namespace that it explicitly calls out by name along with the key names to expose it to the ? If the target Proxy is in a different namespace it might require duplicating the secret into that namespace with a reconciliation process to remove it when the referencing EEP is removed.

sgargan:
> As the next step, I suggest raising a PR for the API design, followed by a follow-up PR for the implementation, once the community reaches a consensus on the API.

Sure thing. Just so I'm clear on the process, are you suggesting a PR with just the wasm_types.go in it to capture the api conversation?

arkodg:
hey @sgargan , @guydc and I discussed this issue again today in the community meeting and prefer if the API was `env.HostKeys` instead of `hostEnvKeys` .

It should be fine to make all the changes in this PR, we ideally prefer if the API and Implementation PR are separate to keep the PR sizes small and easier to review, but also the API discussion ( and churn) may cause a lot of work/change in the PR for the commiter

daniel-white:
@arkodg i'm currently implementing a wasm plugin and i need to pass some sensitive values. is this likely going to land in 1.2?

zhaohuabing:
> rethinking this one @guydc @zhaohuabing we can't allow the app dev who has access to the EEP to inject any env vars into the proxy, so we'll need to split up the task of injecting env vars (using EnvoyProxy resource by the platform admin) and allowing the app dev to use it by specifying this field

@arkodg Sorry, I didn't quite get the security concern behind this. Does this mean that we can still use a ValueRef to a Secret, just like what EG does with the SecuritPolicy, but not an env from the host?

arkodg:
> > rethinking this one @guydc @zhaohuabing we can't allow the app dev who has access to the EEP to inject any env vars into the proxy, so we'll need to split up the task of injecting env vars (using EnvoyProxy resource by the platform admin) and allowing the app dev to use it by specifying this field
> 
> @arkodg Sorry, I didn't quite get the security concern behind this. Does this mean that we can still use a ValueRef to a local Secret, just like what EG does with the SecuritPolicy, but not an env from the host because the host envs are global scope and the app dev may not be allowed to access them?

yes, retrieving a value here requires inserting it to the host first (for `hostEnvKeys`, not `customKeys`), which means you could overwrite any existing val, that maybe critical for operation of the proxy or for another route

zhaohuabing:
> yes, retrieving a value here requires inserting it to the host first (for hostEnvKeys, not customKeys), which means you could overwrite any existing val, that maybe critical for operation of the proxy or for another route

Got it, since envs are used host wide, so it's should be manged by the admin who has a global view to all variables. Thanks for the explaination!

sgargan:
> this won't make it into v1.2, we just cut the RC release and can't add anymore API changes to this release

Apologies for the delay, I was ooo last week and couldn't get to it. I've made the suggested changes to introduce an `env` grouping for the keys.

api/v1alpha1/wasm_types.go
```diff
@@ -52,6 +60,10 @@ type Wasm struct {
 	// Priority defines the location of the Wasm extension in the HTTP filter chain.
 	// If not specified, the Wasm extension will be inserted before the router filter.
 	// Priority *uint32 `json:"priority,omitempty"`
+
+	// Env configures the environment for the Wasm extension
+	// +optional
+	Env WasmEnv `json:"env,omitempty"`
```
arkodg, line 66: optional - this needs to be a ptr type

api/v1alpha1/wasm_types.go
```diff
@@ -10,6 +10,14 @@ import (
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// WasmEnv defines the environment for the VM of a Wasm extension
```
zhaohuabing, line 13: Nit: 
```suggestion
// WasmEnv defines the environment variables for the VM of a Wasm extension
```

sgargan: updated

arkodg (approved):
LGTM thanks !

thanks for addressing all the comments

arkodg:
hey @sgargan there's a lint error
```
tools/bin/yamllint --config-file=tools/linter/yamllint/.yamllint $(git ls-files :*.yml :*.yaml | xargs -L1 dirname | sort -u) 
./internal/xds/translator/testdata/in/xds-ir/wasm.yaml
  Error: 94:22 [new-line-at-end-of-file] no new line character at the end of file
 ```

sgargan:
@arkodg I've updated the lint error, can this be merged?

arkodg (approved):

zhaohuabing (approved):

Issue #4472, opened 2024-10-18
Collect release note in each PR


zhaohuabing:
As a release manager, I need to go through every commit from the previous release to the latest one to write the release note.  This is time-consuming and slowing down the release process. 

We'll need to spread the work and collect the release note in each PR.

PR #4473, opened 2024-10-18
Release note

zhaohuabing:
Fix https://github.com/envoyproxy/gateway/issues/4472

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 **What type of PR is this?**
 <!--
-Your PR title should be descriptive, and generally start with type that contains a subsystem name with `()` if necessary 
+Your PR title should be descriptive, and generally start with type that contains a subsystem name with `()` if necessary
 and summary followed by a colon. format `chore/docs/api/feat/fix/refactor/style/test: summary`.
 Examples:
 * "docs: fix grammar error"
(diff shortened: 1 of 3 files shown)


zirain:
it would be better to have linter for this.

zhaohuabing:
> it would be better to have linter for this.

Yeah, we can add that later. Checking when reviewing the PR can help solving the current issue.

arkodg:
thanks @zhaohuabing the workflow looks great.

I was hoping we can revisit the Release Notes format, right now its focused on changes for a specific area, my preference would be to describe the change in terms of the feature being added or bug being fixed, wdyt @envoyproxy/gateway-maintainers , there's enough past work in this space we can get inspiration from

zhaohuabing:
@arkodg Good point! 

The current format follows Envoy, which makes sense for envoy as most peoploe working on envoy are developers, and they need to know the technical details.

However, I believe we should reorganize EG's changelog to better algin with the end users' perspective. As an end user, I would expect to see the following sections in the changelog:
* new features
* security update
* breaking changes
* bug fixes

Minor tasks, such as improvement of CI or fixes for flaky tests, don't need to be included in the changelog.

arkodg:
hey @zhaohuabing +1 to the proposed format

arkodg:
also additional sections called `performance` and `other` will be useful as well

guydc (approved):

zirain (approved):

Issue #4466, opened 2024-10-17
Allow to pull wasm image from insecure registry while using envoy patch policy


elvenlegolas:
*Description*:
I got an x509: certificate signed by unknown authority error because the wasm image registry was insecure while using envoy patch policy,I've read about the api docs but couldn't find where I can configure trusted registry in,so I wonder if it can be configured or is a feature to be provided?
Thanks!
[optional *Relevant Links*:]
https://github.com/istio/istio/issues/36571 is a similar situation.

zirain:
cc @zhaohuabing

zhaohuabing:
Hi @elvenlegolas As a temporary workaround, Can you use a HTTP code source for testing?

elvenlegolas:
> Hi @elvenlegolas As a temporary workaround, Can you use a HTTP code source for testing?

Ok I'll try it,btw the example wasm http url given in envoy gateway doc https://raw.githubusercontent.com/envoyproxy/envoy/main/examples/wasm-cc/lib/envoy_filter_http_wasm_example.wasm seems invalid now ,would you fix it as a reference?

arkodg:
i believe this is already solved, right @zhaohuabing  ?

zhaohuabing:
@arkodg The API is in. We still need implementation.

https://github.com/envoyproxy/gateway/blob/309750a6c7157463a3dd86ff8500cb5c58e05bf0/api/v1alpha1/wasm_types.go#L131-L135

kamxnet:
thanks! Let me take a look.

PhantomInTheWire:
it looks like @kamxnet is no longer working on this, can I pick this up instead @zhaohuabing?

zhaohuabing:
Thanks @PhantomInTheWire for picking this up! Feel free to proceed.

muralisankar32:
Are there any updates on this issue? Could see WasmCodeSourceTLSConfig in the documentation, but it still does not work.. Also it does not specify which key should be used for caCertificateRef..

PR #4492, opened 2024-10-21
api: Wasm code source TLS configuration

zhaohuabing:
API for: https://github.com/envoyproxy/gateway/issues/4466


zirain:
I think we still need allow insecure registry in some cause, e.g. dev, localhost?

zhaohuabing:
> I think we still need allow insecure registry in some cause, e.g. dev, localhost?

We should avoid insecure as it poses a potential security risk, a self-signed CA can be used for testing and development purposes.

api/v1alpha1/wasm_types.go
```diff
@@ -115,6 +115,11 @@ type HTTPWasmCodeSource struct {
 	// kubebuilder:validation:Pattern=`^[a-f0-9]{64}$`
 	// +optional
 	SHA256 *string `json:"sha256"`
+
```
arkodg, line 118: should this also use BackendCluster ?

zhaohuabing: It's possible, but using BackendCluster may not be the appropriate approach - the TLS setting are used by the control plane rather than the data plane.

arkodg: ah makes sense

api/v1alpha1/wasm_types.go
```diff
@@ -163,3 +173,14 @@ const (
...
+// WasmCodeSourceTLSConfig defines the TLS configuration when connecting to the Wasm code source.
+type WasmCodeSourceTLSConfig struct {
+	// CACertificateRef contains a references to
+	// Kubernetes objects that contain TLS certificates of
+	// the Certificate Authorities that can be used
+	// as a trust anchor to validate the certificates presented by the Wasm code source.
+	//
+	// Kubernetes ConfigMap and Kubernetes Secret are supported.
+	CACertificateRef gwapiv1.SecretObjectReference `json:"caCertificateRef"`
```
arkodg, line 185: which ns does this need to live in ?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4493, opened 2024-10-22
feat: implement Backend API for TLSRoute

zhaohuabing:
impl: https://github.com/envoyproxy/gateway/issues/4035


guydc (approved):
LGTM

internal/gatewayapi/validate.go
```diff
@@ -288,7 +288,7 @@ func (t *Translator) validateBackendRefBackend(backendRef *gwapiv1a2.BackendRef,
 		return false
 	}
 
-	if kind != resource.KindHTTPRoute {
+	if kind != resource.KindHTTPRoute && kind != resource.KindTLSRoute {
```
arkodg, line 291: sweet, was this the only fix needed ?
can we rm this check so we can support this for all routes ? #4036 #4037 #4038

PR #4505, opened 2024-10-23
Reduce the amount of configuration logging, and make it line-delimeted friendly

evankanderson:
**What type of PR is this?**

fix: Gateway API currently logs multi-line JSON objects in the otherwise line-delimited logs.

I noticed (looking through our logs system, which collects line-delimited logs from containers using fluent-bit) a number of logs of the form:

```
			"metadata": {
```

Needless to say, stitching these logs back into a coherent stream is a bit of a chore.  Additionally, it seems like envoy-gateway is logging most of the desired configuration every 1-2 minutes by default, which seems a bit excessive, particularly as configuration gets larger.

I've moved the configuration dumps behind a `-v 1` log verbosity flag, and collapsed the JSON onto a single (very long) line.  Defaulting to structured logs might make more sense, but I'm wondering if you actually need to log all the configuration, including both the k8s object (with annotations!) and the full envoy config.

I did not open an issue for this, as the fix seemed feasible to simply do.

Release Notes: Yes


evankanderson:
#4506 adds coverage for the Xds `JSONString()` method, but I'm not sure that testing the others is worthwhile.

internal/gatewayapi/runner/runner.go
```diff
@@ -173,7 +173,9 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				// Publish the IRs.
 				// Also validate the ir before sending it.
 				for key, val := range result.InfraIR {
-					r.Logger.WithValues("infra-ir", key).Info(val.JSONString())
```
arkodg, line 176: shouldnt this be changed to `.Debug` and they won't show up because the default logging is `Info` 
https://github.com/envoyproxy/gateway/blob/6f5ae8ebc8ca3c30257a74d5dedc868c96e15038/api/v1alpha1/envoygateway_helpers.go#L115
cc @zirain

evankanderson: `r.Logger` embeds a [`logr.Logger`](https://pkg.go.dev/github.com/go-logr/logr#Logger).  That interface doesn't believe in named levels beyond `Info()` and `Error()`, but it supports arbitrary levels of verbosity with `V()`.

zirain: yeah, that's right. there's no `Debug` in `logr.Logger`

zirain: |level|description|
|----|----|
|v=0 | Generally useful for this to always be visible to a cluster operator.|
|v=1 | A reasonable default log level if you don’t want verbosity.|
|v=2 | Useful steady state information about the service and important log messages that may correlate to significant changes in the system.This is the recommended default log level for most systems.|
|v=3 | Extended information about changes.|
|v=4 | Debug level verbosity.|
|v=5 | Trace level verbosity.|
|v=6 | Display requested resources.|
|v=7 | Display HTTP request headers.|
|v=8 | Display HTTP request contents.|
|v=9 | Display HTTP request contents without truncation of contents.|


if we want something like debug, should change the level to `4`.

evankanderson: I'm happy to switch this to 4; it was previously always-on, so I wasn't sure how much it should be hidden, but 4 makes sense to me.

arkodg: +1 to 4

evankanderson: Updated.

internal/gatewayapi/runner/runner.go
```diff
@@ -184,7 +186,9 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
 				}
 
 				for key, val := range result.XdsIR {
-					r.Logger.WithValues("xds-ir", key).Info(val.JSONString())
+					if vlog := r.Logger.V(1); vlog.Enabled() {
```
zirain, line 189: same here?

release-notes/current.yaml
```diff
@@ -14,7 +14,7 @@ new features: |
 
 # Fixes for bugs identified in previous versions.
 bug fixes: |
-  Add a bug fix here
+  Only log endpoint configuration in verbose logging mode (`-v 1` or higher)
```
zirain, line 17: and the release notes

internal/gatewayapi/runner/runner.go
evankanderson: Oops, thanks!  Done.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4506, opened 2024-10-23
Use custom marshaller to clarify redactions

evankanderson:
**What type of PR is this?**

fix: Make redacted fields show as `"[redacted]"` rather than `"W3JlZGFjdGVkXQ=="`

I started preparing to report a security issue with Envoy Gateway because I noticed that our logs contained a `"privateKey"` with a base64-encoded value.  Eventually, while trying to get the key to line up with the certificate, I base-64 decoded the contents, and saw that the value was `[redacted]`.  I'm hoping this avoids future users getting freaked out like I did.

As a maybe-beneficial side-effect, it shouldn't be necessary to call `Printable()` to avoid leaking XDS secret information when marshalling to/from JSON or YAML.

I didn't end up filing a bug yet.

Release Notes: No


evankanderson:
Note: this incorporates #4505 , as that changed the format of `JSONString()`.

evankanderson:
Hold off on reviewing this; it looks like I missed `make -k gen-check`, which seems to be getting a bunch of double-base-64 encodings from somewhere even when I account for empty-bytes --> `nil` rather than `[redacted]`.

evankanderson:
I think I've fixed the errors from `make gen-check`, though it did add a bit of extra (mostly test) complexity.

arkodg (approved):
sorry for the late reply
LGTM thanks !

shawnh2 (approved):
LGTM

PR #4508, opened 2024-10-24
feat: direct response

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/2714


internal/xds/translator/testdata/in/xds-ir/accesslog-cel.yaml
```diff
@@ -51,6 +51,3 @@ http:
       - endpoints:
         - host: "1.2.3.4"
           port: 50000
-    directResponse:
```
zirain, line 54: why this is removed?

arkodg: it wasn't adding any value to the access log tests

zirain: there's no new test?

zirain: maybe worth to do in an seperate PR, and add some test for DirectReponse in this one.

arkodg: if I keep this, need there will be new out files generated, because the `body` is now translated to xds

internal/xds/translator/route.go
```diff
@@ -440,8 +440,20 @@ func buildXdsURLRewriteAction(destName string, urlRewrite *ir.URLRewrite, pathMa
 	return routeAction
 }
 
-func buildXdsDirectResponseAction(res *ir.DirectResponse) *routev3.DirectResponseAction {
-	routeAction := &routev3.DirectResponseAction{Status: res.StatusCode}
+func buildXdsDirectResponseAction(res *ir.CustomResponse) *routev3.DirectResponseAction {
```
zhaohuabing, line 443: What if `res.StatusCode == nil && (res.Body == nil || res.Body == "")`?

arkodg: in the gateway api layer we are always setting the status code, so we should not be hitting this case

zhaohuabing: Oh I see, the default has been set to 200 in gateway translation.

It would be more clearer if a default 200 is also set in the xds translation.

internal/xds/translator/testdata/in/xds-ir/accesslog-cel.yaml
zirain: ok

zhaohuabing (approved):
LGTM thanks!

zirain (approved):

PR #4518, opened 2024-10-24
doc: add standalone deployment doc

### LINKED ISSUE #1393: Run Envoy Gateway locally on host

shawnh2:
**What type of PR is this?**

Fixes #1393

Release Notes: No


site/content/en/latest/tasks/operations/standalone-deployment-mode.md
```diff
@@ -0,0 +1,99 @@
...
+
+Update gateway-api resources:
+
+```shell
+cp examples/standalone/quickstart.yaml /tmp/envoy-gateway-test/quickstart.yaml
+```
+
+From Envoy Gateway log, you should be able to observe a Envoy Proxy has been started.
+
+### Test Connection
```
shawnh2, line 93: There's a connection problem that may require another cycle to look into. By applying resources from `examples/standalone/quickstart.yaml`, `curl --verbose --header "Host: www.example.com" http://127.0.0.1:8888` returns 500, seems we have some problems in envoyproxy xds.

site/content/en/latest/tasks/operations/standalone-deployment-mode.md
```diff
@@ -0,0 +1,137 @@
...
+Create a local directory just for testing:
+
+```shell
+mkdir -p /tmp/envoy-gateway-test
+```
+
+Compile Envoy Gateway binary from project by using command:
+
+```shell
+make build
```
shawnh2, line 41: currently, we let users to build envoy-gateway binary from scratch. should we include this binary in release ?

shawnh2:
wait https://github.com/envoyproxy/gateway/pull/4535

site/content/en/latest/tasks/operations/standalone-deployment-mode.md
zirain: Event in standalone mode, we should use docker instead of pure binary?

zhaohuabing: For now, I think it's fine to use docker to spin up the EG in the documentation.

However, we should include the binaries to the release assets and update the doc to add the binary option. This can be addressed is a follow-up PR.

site/content/en/latest/tasks/operations/standalone-deployment-mode.md
```diff
@@ -0,0 +1,137 @@
...
+---
+
+{{% alert title="Notice" color="warning" %}}
+
+Standalone mode is an experimental feature, please **DO NOT** use it in production.
+
+{{% /alert %}}
+
+Envoy Gateway also supports running in standalone mode. In this mode, you can configure the
+Envoy Gateway to use different file provider and infrastructure provider combinations.
```
zhaohuabing, line 12: Not a strong opinion - I suggest we remove this and rephrase the standalone mode as the only support combination is file provider and host infrastructure. It would be confusing for readers.

site/content/en/latest/tasks/operations/standalone-deployment-mode.md
arkodg: we shouldnt be use `make` command here, instead either use GH releases assets https://github.com/envoyproxy/gateway/releases/tag/v1.2.0-rc.1 (if `envoy-gateway` is not present, we should consider pushing it), or Docker Image, my vote is for GH releases

zirain: if you build binary, you will need to support all the platforms(e.g. macos, windows), TBH, I think there's a kind of waste?

arkodg: there are many production environments/server thats are running a server with a manager like systemd that have no docker installed

arkodg: similar to envoyproxy https://github.com/envoyproxy/envoy/releases/tag/v1.32.0 , imo `linux-x86_64` is good enough for now

shawnh2: like @zhaohuabing 's idea. we should include eg binary in release, and maybe just for linux only, this can be done as a follow-up.

but for now, using docker to start eg sgtm.

arkodg: raised https://github.com/envoyproxy/gateway/issues/4566 to track the GH release work

arkodg: the binary is now part of the artifacts, hope this unblocks you @shawnh2

examples/standalone/quickstart.yaml
```diff
@@ -0,0 +1,46 @@
...
+            value: /
+---
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Backend
+metadata:
+  name: backend
+spec:
+  endpoints:
+    - ip:
+        address: 0.0.0.0
```
arkodg, line 45: can we add a comment in the YAML saying `0.0.0.0` is only used for demo purposes, do not use this in a production environment

site/content/en/docs/tasks/operations/standalone-deployment-mode.md
```diff
@@ -0,0 +1,128 @@
...
+{{% alert title="Notice" color="warning" %}}
+
+Standalone mode is an experimental feature, please **DO NOT** use it in production.
+
+{{% /alert %}}
+
+Envoy Gateway also supports running in standalone mode. In this mode, Envoy Gateway
+does not need to rely on Kubernetes and can be deployed directly on bare metal or virtual machines.
+In the standalone deployment mode, the functions and features supported by Envoy Gateway are 
+basically the same as those in the Kubernetes deployment mode.
```
arkodg, line 14: my suggestion would be to rm L13-14 or specifically specify which features will not work in standalone mode

examples/standalone/quickstart.yaml
shawnh2: added

site/content/en/docs/tasks/operations/standalone-deployment-mode.md
shawnh2: removed

examples/standalone/quickstart.yaml
```diff
@@ -42,5 +42,5 @@ metadata:
 spec:
   endpoints:
     - ip:
-        address: 0.0.0.0
+        address: 0.0.0.0 # this address is for demo purpose only, do not use it in any production!
```
arkodg, line 45: in production

site/content/en/docs/tasks/operations/standalone-deployment-mode.md
```diff
@@ -0,0 +1,126 @@
...
+logging:
+  level:
+    default: info
+extensionApis:
+  enableBackend: true
+```
+
+As you can see, we have enabled the [Backend][] API, this API will be used to represent our local endpoints.
+
+Since Envoy Gateway is running in the standalone mode, we highly recommend using the [Backend][] API instead of
```
zhaohuabing, line 77: Probably just remove this line - it makes no sense to use K8s services and EndpointSlice in the host mode.

shawnh2: sgtm

Issue #3387, opened 2024-05-13
make experimental gateway api crds optional so that gateway-api upgrade don't break envoy gateway


networkhermit:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

Recently I upgraded the gateway-api to v1.1.0 from the standard channel, but I found that envoy-gateway is in error due to missing some experimental gateway api crds, even though I have not directly used GRPCRoute gateway.networking.k8s.io/v1alpha2 at all.

Make these experimental crds optional could help users to upgrade newer gateway-api releases.

envoy-gateway log:

```text
Error: failed to create provider Kubernetes: failted to create gatewayapi controller: no matches for kind "GRPCRoute" in version "gateway.networking.k
8s.io/v1alpha2"
Usage:
  envoy-gateway server [flags]

Aliases:
  server, serve

Flags:
  -c, --config-path string   The path to the configuration file.
  -h, --help                 help for server

failed to create provider Kubernetes: failted to create gatewayapi controller: no matches for kind "GRPCRoute" in version "gateway.networking.k8s.io/v
1alpha2"
```

More context: https://github.com/kubernetes-sigs/gateway-api/issues/3075 and [the way istio handle it](https://github.com/kubernetes-sigs/gateway-api/issues/3075#issuecomment-2107687646).

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
>Include the access logs and the Envoy logs.

arkodg:
we've implemented something similar for `ServiceImport` where we check if it exists before watching it. For this case we are not the CRD owners, so this decision was easy
https://github.com/envoyproxy/gateway/blob/3bc7cf1d00635f602bc4731a904675d982d34317/internal/provider/kubernetes/controller.go#L1099

Do we need to the same for
1. **Gateway API CRDs** - that are not installed from the EG Helm Chart, but pre-applied and managed by the cluster admin (e.g. GKE)  (helm not updating CRDs just highlights this case even more )
2. **EG CRDs** (BackendTrafficPolicy etc) - that are not installed in the cluster, where the user wants to run EG in minimal mode (Gateway API featureset) and pre install the Gateway API CRDs from the standard channel themselves.

[...]

PR #4519, opened 2024-10-24
make watching alpha CRDs optional

arkodg:
* The ownership of CRD installation is not tied to a single entity https://gateway-api.sigs.k8s.io/guides/crd-management/#who-should-manage-crds This results in multiple entities taking ownership of CRD installation
  * infra users
  * implementations
  * cloud providers

This complicates things for implementations who may not know which version and release of CRDs are installed, so this PR makes watching alpha versioned CRDs optional

* Even Envoy Gateway specific CRDs have been made optional to solve the use case where users want to only configure Gateway API resources

* GRPCRoute is the only exception, which is v1, but has been made optional because it just graduated to v1 in v1.2 but a lot of cloud providers or service mesh implementations have not moved to v1.2

Fixes: https://github.com/envoyproxy/gateway/issues/3387

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1172,92 +1172,112 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
 		return err
 	}
 
-	// Watch GRPCRoute CRUDs and process affected Gateways.
-	grpcrPredicates := []predicate.TypedPredicate[*gwapiv1.GRPCRoute]{
-		predicate.Or(predicate.TypedGenerationChangedPredicate[*gwapiv1.GRPCRoute]{},
-			predicate.TypedLabelChangedPredicate[*gwapiv1.GRPCRoute]{}),
-	}
-	if r.namespaceLabel != nil {
-		grpcrPredicates = append(grpcrPredicates, predicate.NewTypedPredicateFuncs[*gwapiv1.GRPCRoute](func(grpc *gwapiv1.GRPCRoute) bool {
-			return r.hasMatchingNamespaceLabels(grpc)
-		}))
-	}
-	if err := c.Watch(
-		source.Kind(mgr.GetCache(), &gwapiv1.GRPCRoute{},
-			handler.TypedEnqueueRequestsFromMapFunc[*gwapiv1.GRPCRoute](func(ctx context.Context, route *gwapiv1.GRPCRoute) []reconcile.Request {
-				return r.enqueueClass(ctx, route)
-			}),
-			grpcrPredicates...)); err != nil {
-		return err
-	}
-	if err := addGRPCRouteIndexers(ctx, mgr); err != nil {
-		return err
-	}
-
-	// Watch TLSRoute CRUDs and process affected Gateways.
-	tlsrPredicates := []predicate.TypedPredicate[*gwapiv1a2.TLSRoute]{
-		predicate.Or(predicate.TypedGenerationChangedPredicate[*gwapiv1a2.TLSRoute]{},
-			predicate.TypedLabelChangedPredicate[*gwapiv1a2.TLSRoute]{}),
-	}
-	if r.namespaceLabel != nil {
-		tlsrPredicates = append(tlsrPredicates, predicate.NewTypedPredicateFuncs[*gwapiv1a2.TLSRoute](func(route *gwapiv1a2.TLSRoute) bool {
-			return r.hasMatchingNamespaceLabels(route)
-		}))
-	}
-	if err := c.Watch(
-		source.Kind(mgr.GetCache(), &gwapiv1a2.TLSRoute{},
-			handler.TypedEnqueueRequestsFromMapFunc[*gwapiv1a2.TLSRoute](func(ctx context.Context, route *gwapiv1a2.TLSRoute) []reconcile.Request {
-				return r.enqueueClass(ctx, route)
-			}),
-			tlsrPredicates...)); err != nil {
-		return err
-	}
-	if err := addTLSRouteIndexers(ctx, mgr); err != nil {
-		return err
+	grpcRouteCRDExists := r.crdExists(mgr, resource.KindGRPCRoute, gwapiv1.GroupVersion.String())
+	if !grpcRouteCRDExists {
+		r.log.Info("GRPCRoute CRD not found, skipping GRPCRoute watch")
+	} else {
+		// Watch GRPCRoute CRUDs and process affected Gateways.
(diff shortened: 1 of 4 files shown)


arkodg:
manually tested this by

1. Installing v1.0.0 Gateway API CRDs from the Stable Channel
```
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml
```

2. Installed Envoy Gateway using Helm and the `--skip-crds` option
```
helm install eg oci://docker.io/arkodg/gateway-helm --version v0.0.0-test3 -n envoy-gateway-system --create-namespace --skip-crds
```

3. Installed the QuickStart
```
kubectl apply -f https://github.com/envoyproxy/gateway/releases/download/v1.1.2/quickstart.yaml -n default
```

4. Tested it
```
export GATEWAY_HOST=$(kubectl get gateway/eg -o jsonpath='{.status.addresses[0].value}')
curl --verbose --header "Host: www.example.com" http://$GATEWAY_HOST/get
*   Trying 127.0.0.1:80...
* Connected to 127.0.0.1 (127.0.0.1) port 80
> GET /get HTTP/1.1
> Host: www.example.com
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< x-content-type-options: nosniff
< date: Sat, 26 Oct 2024 00:38:16 GMT
< content-length: 456
< 
{
 "path": "/get",
 "host": "www.example.com",
 "method": "GET",
 "proto": "HTTP/1.1",
 "headers": {
  "Accept": [
   "*/*"
  ],
  "User-Agent": [
   "curl/8.4.0"
  ],
  "X-Envoy-Internal": [
   "true"
  ],
  "X-Forwarded-For": [
   "192.168.65.4"
  ],
  "X-Forwarded-Proto": [
   "http"
  ],
  "X-Request-Id": [
   "8f3833ab-c84c-4bb9-8f23-a3b165715ba0"
  ]
 },
 "namespace": "default",
 "ingress": "",
 "service": "",
 "pod": "backend-69fcff487f-d4fpj"
* Connection #0 to host 127.0.0.1 left intact
```

internal/provider/kubernetes/predicates_test.go
```diff
@@ -356,6 +356,9 @@ func TestValidateSecretForReconcile(t *testing.T) {
 	r := gatewayAPIReconciler{
 		classController: egv1a1.GatewayControllerName,
 		log:             logger,
+		spCRDExists:     true,
```
zirain, line 359: I'm think about somehting like `sets[GVK]` instead of introduce two much properties.

arkodg: yeah, the sets approach saves a few lines (~10) of code by not instantiating the `bool`, but will add the same amount when inserting values into the set

arkodg: I dont see this changes much in the future, imo, the APIs have kinda stabilized

zirain (approved):

internal/provider/kubernetes/controller.go
```diff
@@ -1172,92 +1214,112 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
-			handler.TypedEnqueueRequestsFromMapFunc[*gwapiv1.GRPCRoute](func(ctx context.Context, route *gwapiv1.GRPCRoute) []reconcile.Request {
-				return r.enqueueClass(ctx, route)
-			}),
-			grpcrPredicates...)); err != nil {
-		return err
-	}
-	if err := addGRPCRouteIndexers(ctx, mgr); err != nil {
-		return err
+	r.grpcRouteCRDExists = r.crdExists(mgr, resource.KindGRPCRoute, gwapiv1.GroupVersion.String())
+	if !r.grpcRouteCRDExists {
```
zhaohuabing, line 1218: Should'nt EG report an error instead of an info message here? Since GRPCRoute is now in the standard channel, its absence suggests an issue with the current deployment.

internal/provider/kubernetes/controller.go
```diff
@@ -1489,118 +1556,143 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
-		predicate.TypedGenerationChangedPredicate[*egv1a1.SecurityPolicy]{},
-	}
-	if r.namespaceLabel != nil {
-		spPredicates = append(spPredicates, predicate.NewTypedPredicateFuncs[*egv1a1.SecurityPolicy](func(sp *egv1a1.SecurityPolicy) bool {
-			return r.hasMatchingNamespaceLabels(sp)
-		}))
-	}
+	r.spCRDExists = r.crdExists(mgr, resource.KindSecurityPolicy, egv1a1.GroupVersion.String())
+	if !r.spCRDExists {
+		r.log.Info("SecurityPolicy CRD not found, skipping SecurityPolicy watch")
```
zhaohuabing, line 1617: Same here

internal/provider/kubernetes/controller.go
zhaohuabing: Makes sense. Maybe add a TODO for making it an error or removing in in the next release?

internal/provider/kubernetes/controller.go
```diff
@@ -1489,118 +1556,143 @@ func (r *gatewayAPIReconciler) watchResources(ctx context.Context, mgr manager.M
...
-		predicate.TypedGenerationChangedPredicate[*egv1a1.BackendTrafficPolicy]{},
-	}
-	if r.namespaceLabel != nil {
-		btpPredicates = append(btpPredicates, predicate.NewTypedPredicateFuncs[*egv1a1.BackendTrafficPolicy](func(btp *egv1a1.BackendTrafficPolicy) bool {
-			return r.hasMatchingNamespaceLabels(btp)
-		}))
-	}
+	r.btpCRDExists = r.crdExists(mgr, resource.KindBackendTrafficPolicy, egv1a1.GroupVersion.String())
+	if !r.btpCRDExists {
+		r.log.Info("BackendTrafficPolicy CRD not found, skipping BackendTrafficPolicy watch")
```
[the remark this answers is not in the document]
zhaohuabing: I think there're two approaches:
* Adds an opt-out configuration in `EnvoyGateway` API for EG CRs
* Reports a warning in the log

This doesn't have to be handled in this PR as we're approaching v1.2.0. It can be tracked in an issue and addressed later.

zirain (approved):

internal/provider/kubernetes/controller.go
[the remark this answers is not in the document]
arkodg: raised https://github.com/envoyproxy/gateway/issues/4553

zhaohuabing (approved):
LGTM thanks!

PR #4476, opened 2024-10-18
release: v1.2.0-rc.1

zhaohuabing:
release: v1.2.0-rc.1

PR #4523, opened 2024-10-25
Revert "chore: upgrade test should use VERSION by default (#4393)"

zhaohuabing:
Revert this for now to release v1.2.0-rc.1
https://github.com/envoyproxy/gateway/pull/4476#issuecomment-2436776156

### DIFF
--- a/test/e2e/tests/eg_upgrade.go
+++ b/test/e2e/tests/eg_upgrade.go
@@ -12,17 +12,14 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/cli"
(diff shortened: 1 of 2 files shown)


zirain:
why revert?

zirain:
#4393 fix other problem, you'd better send another one:
- revert the change in kube.mk
- chagne lastVersionTag to "v1.1.2"

arkodg:
we are going to always hit this issue when releasing, we can't revert each time, I would instead hardcode this in the test, and add it as a step post releasing the bits

zirain:
> we are going to always hit this issue when releasing, we can't revert each time, I would instead hardcode this in the test, and add it as a step post releasing the bits

we should change the logic `if version == "" || strings.Contains("rc")`

zirain:
oh, same when releasing `1.2.0`.

zhaohuabing:
Try to fix it with https://github.com/envoyproxy/gateway/pull/4524

zhaohuabing:
> Try to fix it with #4524

This only works for the main branch. If we try to release a minor release like v1.1.3 after v1.2.0, it will use v1.2.0 as for upgrade test.

I'll update the version in upgrade test after releasing v1.2.0 and use that PR as an example to add a step in the release note.

PR #4525, opened 2024-10-25
e2e: use grafana alloy instead of fluent-bit

zirain:
- this will make configuration simpler and test run faster
- easier to support IPv6 and Dual


shawnh2 (approved):

zhaohuabing:
Hi @zirain curios why grafana alloy is faster than fluent-bit?

zhaohuabing:
There are two copies of logs are in the loki?

zirain:
> Hi @zirain curios why grafana alloy is faster than fluent-bit?

it's faster on my local cluster.

zirain:
> I would prefer to keep the more popular one (for users who are using EG or likely to use it) rather than the faster one

I want to bring this back cause I cannot figure out how to make fluent-bit support IPv6 and Dual.

arkodg (approved):

zhaohuabing (approved):

Issue #4498, opened 2024-10-23
Scaling down routes not scaling down memory usage


arkodg:
### Metrics

|Test Name                |Envoy Gateway Memory (MiB)|Envoy Gateway CPU (s)|Envoy Proxy Memory (Avg) (MiB)|Envoy Proxy CPU (Avg) (s)|
|-                        |-                         |-                    |-                             |-                        |
|scale-up-httproutes-10   |107.11                    |0.74                 |25.85                         |30.44                    |
|scale-up-httproutes-50   |136.46                    |1.42                 |31.97                         |60.84                    |
|scale-up-httproutes-100  |136.98                    |2.75                 |38.13                         |91.64                    |
|scale-up-httproutes-300  |141.98                    |14.64                |60.31                         |124.75                   |
|scale-up-httproutes-500  |157.41                    |26.76                |80.52                         |158.63                   |
|scale-up-httproutes-1000 |179.77                    |57.89                |130.77                        |197.64                   |
|scale-down-httproutes-500|164.67                    |85.83                |130.91                        |233.84                   |
|scale-down-httproutes-300|152.14                    |96.66                |131.07                        |266.26                   |
|scale-down-httproutes-100|146.73                    |106.85               |127.11                        |298.33                   |
|scale-down-httproutes-50 |161.01                    |107.92               |127.04                        |329.02                   |
|scale-down-httproutes-10 |202.45                    |108.52               |126.36                        |359.48                   |

_Originally posted by @arkodg in https://github.com/envoyproxy/gateway/issues/4287#issuecomment-2430569347_

arkodg:
is this an issue with the measurement or with EG ?
cc @shawnh2

qicz:
imo. we should  provide routes  with different services or  backend for benchmark. 

nowly, just  for one service, the  EG metrics report show the minimal scene

qicz:
The `resource` cached many resources. Any resource that changes will send a signal, and runners will subscribe to the change message. There are many `DeepCopy`.

I think that we should provide a global cache resource mapping, and the runners subscribe to the simple change signal. in the `HandleSubscription` function or `Translator` gets the resource in the global cache.

![image](https://github.com/user-attachments/assets/2ffdc9ed-7e9b-4086-8b75-a69219b39144)

PR #4527, opened 2024-10-25
bugfix: optimize routes status unnecessary deepcopy.

### LINKED ISSUE #4516: huge memory usage
The `resource` cached many resources. Any resource that changes will send a signal, and runners will subscribe to the change message. There are many `DeepCopy`.

I think that we should provide a global cache resource mapping, and the runners subscribe to the simple change signal. in the `HandleSubscription` function or `Translator` gets the resource in the global cache.

image

for the route status, the `DeepCopy` uses a huge memory. Can we remove this DeepCopy?for my analysis to compare the status change we just cmp the `val.Parents` and `obj.Status.Parents`.  or move the data to the global cache?

like follows:

- remove `DeepCopy` and CMP the status directly

```golang
// HTTPRoute object status updater
go func() {
	message.HandleSubscription(r.resources.HTTPRouteStatuses.Subscribe(ctx),
		func(update message.Update[types.NamespacedName, *gwapiv1b1.HTTPRouteStatus]) {
			// skip delete updates.
			if update.Delete {
				return
			}
			key := update.Key
			val := update.Value
			r.statusUpdater.Send(status.Update{
				NamespacedName: key,
				Resource:       new(gwapiv1b1.HTTPRoute),
				Mutator: status.MutatorFunc(func(obj client.Object) bool {
					h, ok := obj.(*gwapiv1b1.HTTPRoute)
					if !ok {
						panic(fmt.Sprintf("unsupported object type %T", obj))
					}
					// remove DeepCopy
					// hCopy := h.DeepCopy() 
					// hCopy.Status.Parents = val.Parents
					// cmp the Parents
					t := gwapiv1b1.HTTPRouteStatus{}
					t.Parents = val.Parents
					if isStatusEqualHTTPRoute(obj.Status, t) {
						u.log.WithName(h.NamespacedName.Name).
						WithName(h.NamespacedName.Namespace).
						Info("status unchanged, bypassing update")
						return true
					}
					h.Status.Parents = val.Parents
					return false
				}),
			})
		},
	)
	r.log.Info("httpRoute status subscriber shutting down")
}()

func isStatusEqualHTTPRoute(a gwapiv1b1.HTTPRouteStatus, b gwapiv1b1.HTTPRouteStatus) bool {
	opts := cmp.Options{
		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
		cmpopts.IgnoreMapEntries(func(k string, _ any) bool {
			return k == "lastTransitionTime"
		}),
	}
	if cmp.Equal(a, b, opts) {
		return true
	}

	return false
}
```

```golang
func (u *UpdateHandler) apply(update Update) {
	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
		obj := update.Resource

		// Get the resource.
		if err := u.client.Get(context.Background(), update.NamespacedName, obj); err != nil {
			if kerrors.IsNotFound(err) {
				return nil
			}
			return err
		}

		if update.Mutator.Mutate(obj) {
			return nil
		}

		return u.client.Status().Update(context.Background(), obj)
	}); err != nil {
		u.log.Error(err, "unable to update status", "name", update.NamespacedName.Name,
			"namespace", update.NamespacedName.Namespace)
	}
}
```

image

qicz:
**What type of PR is this?**

Part Fixes #4516

Release Notes: Yes/No

### DIFF
--- a/internal/provider/kubernetes/status.go
+++ b/internal/provider/kubernetes/status.go
@@ -66,16 +66,20 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context, ext
 				r.statusUpdater.Send(Update{
 					NamespacedName: key,
 					Resource:       new(gwapiv1.HTTPRoute),
-					Mutator: MutatorFunc(func(obj client.Object) client.Object {
+					Mutator: MutatorFunc(func(obj client.Object) bool {
 						h, ok := obj.(*gwapiv1.HTTPRoute)
 						if !ok {
 							err := fmt.Errorf("unsupported object type %T", obj)
 							errChan <- err
 							panic(err)
 						}
-						hCopy := h.DeepCopy()
-						hCopy.Status.Parents = val.Parents
-						return hCopy
+
+						if isStatusEqual(h.Status.Parents, val.Parents) {
+							return true
+						}
+
+						h.Status.Parents = val.Parents
+						return false
 					}),
 				})
 			},
@@ -96,16 +100,20 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context, ext
 				r.statusUpdater.Send(Update{
 					NamespacedName: key,
 					Resource:       new(gwapiv1.GRPCRoute),
-					Mutator: MutatorFunc(func(obj client.Object) client.Object {
+					Mutator: MutatorFunc(func(obj client.Object) bool {
 						h, ok := obj.(*gwapiv1.GRPCRoute)
 						if !ok {
 							err := fmt.Errorf("unsupported object type %T", obj)
 							errChan <- err
 							panic(err)
 						}
-						hCopy := h.DeepCopy()
-						hCopy.Status.Parents = val.Parents
-						return hCopy
+
+						if isStatusEqual(h.Status.Parents, val.Parents) {
+							return true
+						}
+
+						h.Status.Parents = val.Parents
+						return false
 					}),
 				})
 			},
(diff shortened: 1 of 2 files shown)


internal/provider/kubernetes/status.go
```diff
@@ -66,16 +66,20 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context, ext
 				r.statusUpdater.Send(Update{
 					NamespacedName: key,
 					Resource:       new(gwapiv1.HTTPRoute),
-					Mutator: MutatorFunc(func(obj client.Object) client.Object {
+					Mutator: MutatorFunc(func(obj client.Object) bool {
```
arkodg, line 69: the API is for mutating obects but is now returning a `bool` ? this doesnt seem right

qicz: changed this api to boolean

arkodg: What is the new definition for the mutator function

zirain:
can you the memory diff before-after?

internal/provider/kubernetes/status.go
qicz: https://github.com/envoyproxy/gateway/blob/5b9553243da45d95c398e4ffde3551efd5ce1f07/internal/provider/kubernetes/status_updater.go#L38-L53

qicz:
> can you the memory diff before-after?

I think we should deal this after https://github.com/envoyproxy/gateway/issues/4498#issuecomment-2434213677

internal/provider/kubernetes/status_updater.go
```diff
@@ -37,16 +37,16 @@ type Update struct {
 
 // Mutator is an interface to hold mutator functions for status updates.
 type Mutator interface {
```
qicz, line 39: @arkodg here

internal/provider/kubernetes/status_updater.go
```diff
@@ -37,16 +37,16 @@ type Update struct {
...
-	Mutate(obj client.Object) client.Object
+	Mutate(obj client.Object) bool
 }
 
 // MutatorFunc is a function adaptor for Mutators.
-type MutatorFunc func(client.Object) client.Object
+type MutatorFunc func(client.Object) bool
```
qicz, line 44: @arkodg here

Windfarer:
any update?

arkodg:

PR #4528, opened 2024-10-25
refactor: safeset for controller cache resource.

qicz:
**What type of PR is this?**

Following #4454, optimize the resources cached with the safety set.

Release Notes: Yes/No

### DIFF
--- a/internal/provider/kubernetes/resource.go
+++ b/internal/provider/kubernetes/resource.go
@@ -6,6 +6,8 @@
 package kubernetes
 
 import (
+	"sync"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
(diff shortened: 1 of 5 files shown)


arkodg:
@qicz can you elaborate the issue in the description

internal/provider/kubernetes/resource.go
```diff
@@ -64,28 +66,60 @@ type resourceMappings struct {
 	httpRouteFilters map[utils.NamespacedNameWithGroupKind]*egv1a1.HTTPRouteFilter
 }
 
+type safeSet[T comparable] struct {
```
zirain, line 69: this should live in utils package with some  tests

qicz:
> @qicz can you elaborate the issue in the description

done

internal/utils/safeset/safeset.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type SafeSet[T comparable] struct {
+	lock   sync.RWMutex
+	Values sets.Set[T]
+}
+
+func NewSafeSet[T comparable](items ...T) *SafeSet[T] {
```
zirain, line 19:
```suggestion
func New[T comparable](items ...T) *SafeSet[T] {
```

internal/utils/safeset/safeset.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type SafeSet[T comparable] struct {
+	lock   sync.RWMutex
+	Values sets.Set[T]
```
zirain, line 16: why expose?

zirain:
@qicz can you share more details about why we need a lock here?

PR #4533, opened 2024-10-25
feat: set full URI for the envoy-gateway service using name and namespace

rajatvig:
Signed-off-by: Rajat Vig <rvig@etsy.com>

**What type of PR is this?**

Sets the `xdsServerHost` in Kubernetes to be fully qualified Service Name using the namespace.

Instead of defaulting to `envoy-gateway` sets the `xdsServerHost` to be `envoy-gateway.envoy-gateway-system.svc.cluster.local`.

Release Notes: No

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -83,6 +83,8 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 	containerSpec *egv1a1.KubernetesContainerSpec,
 	shutdownConfig *egv1a1.ShutdownConfig,
 	shutdownManager *egv1a1.ShutdownManager,
+	namespace string,
+	dnsDomain string,
 ) ([]corev1.Container, error) {
 	// Define slice to hold container ports
 	var ports []corev1.ContainerPort
@@ -132,7 +134,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 			TrustedCA:   filepath.Join("/sds", common.SdsCAFilename),
 		},
 		MaxHeapSizeBytes: maxHeapSizeBytes,
-		XdsServerHost:    ptr.To(fmt.Sprintf("%s.%s.svc.cluster.local", infra.Config.Name, infra.Config.Namespace)),
+		XdsServerHost:    ptr.To(fmt.Sprintf("envoy-gateway.%s.svc.%s", namespace, dnsDomain)),
 	}
 
 	args, err := common.BuildProxyArgs(infra, shutdownConfig, bootstrapConfigOptions, fmt.Sprintf("$(%s)", envoyPodEnvVar))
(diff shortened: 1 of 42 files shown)


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -132,6 +132,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 			TrustedCA:   filepath.Join("/sds", common.SdsCAFilename),
 		},
 		MaxHeapSizeBytes: maxHeapSizeBytes,
+		XdsServerHost:    ptr.To(fmt.Sprintf("%s.%s.svc.cluster.local", infra.Config.Name, infra.Config.Namespace)),
```
arkodg, line 135: the infra handle is for the proxy/data plane, and we want to specify the control plane namespace
it may be easier to populate the namespace and dns domain from the runner config and set it in here
https://github.com/envoyproxy/gateway/blob/6f5ae8ebc8ca3c30257a74d5dedc868c96e15038/internal/infrastructure/host/proxy_infra.go#L48

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -132,6 +134,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 			TrustedCA:   filepath.Join("/sds", common.SdsCAFilename),
 		},
 		MaxHeapSizeBytes: maxHeapSizeBytes,
+		XdsServerHost:    ptr.To(fmt.Sprintf("envoy-gateway.%s.svc.%s", namespace, dnsDomain)),
```
arkodg, line 137: can we use `config.EnvoyGatewayServiceName` here ?

zirain:
can you share more details about the benifit? reduce DNS pressure?

rajatvig:
> can you share more details about the benifit? reduce DNS pressure?

Yes. It would reduce the number of DNS resolve requests as just `envoy-gateway` traverses the full search path everytime. 

The typical `/etc/resolv.conf` has `ndots 5` and quite a few search paths varying on the Kubernetes setup.

You can test it out by running and comparing against a CoreDNS pod with logging turned on to see the diference in number of queries.

```
dig +search +short envoy-gateway AAAA
dig +search +short envoy-gateway.envoy-gateway-system.svc.cluster.local AAAA
```

We see 
Screenshot 2024-10-26 at 12 21 15

Should add, this is not true when `autopath @kubernetes` and `pods verified` is on with CoreDNS. In our setup, however, we get arbitrary IPv6 address failures when we turn on autopath despite not using a dual stack cluster. The reason from what I have gathered in Go http/client defaults to requesting both IPv4 and IPv6 records all the time.

internal/infrastructure/kubernetes/proxy/resource.go
rajatvig: done

arkodg (approved):
LGTM thanks !

zirain (approved):

zirain:
@rajatvig  looks like this broken xds_cluster dns reslove on IPv6 cluster?

PR #4535, opened 2024-10-26
feat: enable load backend resources

shawnh2:
**What type of PR is this?**

block #4518

Release Notes: Yes/No


zirain:
can we add test?

arkodg (approved):
🚀

shawnh2:
> can we add test?

we can, but the test will be duplicated as the egctl translate test

zirain (approved):

Issue #3640, opened 2024-06-20
e2e: Attaching EnvoyProxy to Gateways


arkodg:
*Description*:
>Describe the issue.

Add e2e for this feature introduced in https://github.com/envoyproxy/gateway/pull/3532

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #4548, opened 2024-10-28
e2e test for Gateway with EnvoyProxy

zhaohuabing:
Fix: https://github.com/envoyproxy/gateway/issues/3640
Release Notes: No

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -631,7 +631,7 @@ func (r *gatewayAPIReconciler) processSecretRef(
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


zirain:
This scenario has been tested with zipkin tracing test case, but it's fine to have an seperated one.

zhaohuabing:
> This scenario has been tested with zipkin tracing test case, but it's fine to have an seperated one.

Yeah, a standalone test would make it easier to identify whether the issue lies with the Zipkin or an Envoy parameter if the test fails.

arkodg (approved):
LGTM thanks !

PR #4550, opened 2024-10-28
ci: enable ipv6/dual test

zirain:

PR #4559, opened 2024-10-29
helm: make eg-addons support IPv6 cluster

zirain:
xref: https://github.com/envoyproxy/gateway/pull/4550#issuecomment-2442961025


zirain:
test on my linux(IPv6 kind cluster) and macos m1(IPv4 kind cluster), full test will be https://github.com/envoyproxy/gateway/pull/4550

internal/gatewayapi/testdata/securitypolicy-with-jwt-and-invalid-oidc.out.yaml
```diff
@@ -173,7 +173,8 @@ securityPolicies:
         sectionName: http
       conditions:
       - lastTransitionTime: null
-        message: 'OIDC: secret default/client2-secret does not exist.'
+        message: 'OIDC: error fetching endpoints from issuer: Get "https://accounts.google.com/.well-known/openid-configuration":
```
zhaohuabing, line 176: :-)

zirain: nit

.gitignore
```diff
@@ -34,6 +34,7 @@ charts/gateway-helm/values.yaml
 
 # dependency charts generated by addons helm.
 charts/gateway-addons-helm/charts/
+charts/gateway-addons-helm/tmpcharts/
```
zhaohuabing, line 37: Do we need this?

charts/gateway-addons-helm/values.yaml
```diff
@@ -186,7 +186,7 @@ opentelemetry-collector:
   config:
     exporters:
       prometheus:
-        endpoint: 0.0.0.0:19001
+        endpoint: "[${env:MY_POD_IP}]:19001"
```
zhaohuabing, line 189: Nit: does ::19001 work?

.gitignore
zirain: better to keep

charts/gateway-addons-helm/values.yaml
zirain: no, it would pass the internal check.

zirain:
```
_, _, err := net.SplitHostPort("::19001")
	if err != nil {
		fmt.Printf("err: %v", err)
	}
```
```
err: address ::19001: too many colons in address
```

.gitignore
zhaohuabing: Curious why this is needed?

zirain: there's some tmp files, make sure noone will commit it by accident.

charts/gateway-addons-helm/values.yaml
```diff
@@ -209,17 +207,36 @@ opentelemetry-collector:
...
+          scrape_configs:
+          - job_name: opentelemetry-collector
+            scrape_interval: 10s
+            static_configs:
+            - targets:
+              - "[${env:MY_POD_IP}]:8888"
     service:
+      telemetry:
+        metrics:
+          address: "[${env:MY_POD_IP}]:8888"
```
arkodg, line 239: who is populating this env ?

zirain: otel-collector pod

arkodg (approved):
LGTM thanks !
cc @juwon8891

charts/gateway-addons-helm/values.yaml
zirain: we'd better change upstream charts, I do believe they should change the default value.

zhaohuabing (approved):

Issue #4565, opened 2024-10-29
IPv6 dual-stack not working on IPv6 first clusters due to IPv4 fixed listeners


alessiodionisi:
*Description*:

I'm testing the latest release (v1.2.0-rc.1) that adds IPv6 and dual-stack support to Envoy Gateway. On my IPv6-first cluster the pods created for the gateway resource fail the startup probe because, by default, the listener is listening only on IPv4 instead of IPv6 or all interfaces.

https://github.com/envoyproxy/gateway/blob/19eb5f58c7d10228e5225ce8752e3c570c23e593/internal/xds/bootstrap/bootstrap.go#L29

https://github.com/envoyproxy/gateway/blob/19eb5f58c7d10228e5225ce8752e3c570c23e593/internal/xds/bootstrap/bootstrap.go#L42

These IPs are not customizable and are used directly in the `bootstrap.yaml.tpl` template, example:
https://github.com/envoyproxy/gateway/blob/19eb5f58c7d10228e5225ce8752e3c570c23e593/internal/xds/bootstrap/bootstrap.yaml.tpl#L68

*Repro steps*:

1. Deploy Envoy Gateway on the cluster
2. Configure a resource of type Gateway

*Environment*:

- Kubernetes cluster configured in dual-stack with IPv6 as primary:
  ```yaml
  apiVersion: kubeadm.k8s.io/v1beta4
  kind: ClusterConfiguration
  ...
  networking:
    podSubnet: fd00:10:244::/48,10.244.0.0/16
    serviceSubnet: fd00:10:96::/112,10.96.0.0/12
  ```
- Envoy Gateway version: v1.2.0-rc.1

*Logs*:
```
Startup probe failed: Get "http://[fd00:10:244:2::5b]:19001/ready": dial tcp [fd00:10:244:2::5b]:19001: connect: connection refused
```

arkodg:
thanks for testing this out @alessiodionisi ! we'll get this resolved by v1.2.0

arkodg:
ptal @juwon8891

alessiodionisi:
Quick update:

I've switched to `v0.0.0-latest` and patched the bootstrap configuration to listen on IPv6 instead of IPv4, and everything works correctly.

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
...
spec:
  ipFamily: DualStack
  bootstrap:
    value: |
      yaml content below
```

[...]

Issue #4572, opened 2024-10-30
IPv6 known issues


zirain:
- [x] FluentBit cannot start https://github.com/envoyproxy/gateway/pull/4525
- [x] Listen on `0.0.0.0` stead of `::1` in IPv6 first cluster https://github.com/envoyproxy/gateway/pull/4573
- [x] xds_cluster dns faield to reslove https://github.com/envoyproxy/gateway/pull/4691
- [x] e2e pod like `grpc-ext-auth` failed to start, may https://github.com/envoyproxy/gateway/issues/3866 help. https://github.com/envoyproxy/gateway/pull/4576
- [x] RateLimit cannot start on IPv6 https://github.com/envoyproxy/ratelimit/pull/774
- [x] `net.SplitHostPort` return `Failed to parse host "[fc00:f853:ccd:e793::ce:80]": address [fc00:f853:ccd:e793::ce:80]: missing port in address`, because `// A literal IPv6 address in hostport must be enclosed in square
// brackets, as in "[::1]:80", "[::1%lo0]:80"`.  https://github.com/envoyproxy/gateway/issues/4597
- [x] https://github.com/envoyproxy/gateway/issues/4761

zirain:
cc @arkodg @juwon8891

zirain:
Another issue:
```console
kubectl logs -n gateway-conformance-infra setup-keycloak-z2rrm
Defaulted container "setup-keycloak" out of: setup-keycloak, wait-for-keycloak (init)
+ /opt/keycloak/bin/kcadm.sh create users -s username=oidcuser -s enabled=true --server http://keycloak --realm master --user admin --password admin
Logging into http://keycloak as user admin of realm master
HTTPS required [invalid_request]
```

cc @zhaohuabing

zhaohuabing:
> Another issue:
> 
> ```
> kubectl logs -n gateway-conformance-infra setup-keycloak-z2rrm
> Defaulted container "setup-keycloak" out of: setup-keycloak, wait-for-keycloak (init)
> + /opt/keycloak/bin/kcadm.sh create users -s username=oidcuser -s enabled=true --server http://keycloak --realm master --user admin --password admin
> Logging into http://keycloak as user admin of realm master
> HTTPS required [invalid_request]
> ```
> 
> cc @zhaohuabing

Fixed in https://github.com/envoyproxy/gateway/pull/4601

PR #4573, opened 2024-10-30
fix: listener on IPv6 first cluster

zirain:
fix: https://github.com/envoyproxy/gateway/issues/4565
xref: https://github.com/envoyproxy/gateway/issues/4572
Separate from : https://github.com/envoyproxy/gateway/pull/4550

- EG controller will detect the cluster is IPv6 or IPv4 first base on pod IP.
- controller will change the default listener socket address base on that, use `::` instead of `0.0.0.0` when IPv6 first.


internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -277,6 +277,22 @@ func expectedRateLimitContainerEnv(rateLimit *egv1a1.RateLimit, rateLimitDeploym
 	namespace string,
 ) []corev1.EnvVar {
 	env := []corev1.EnvVar{
+		{
```
arkodg, line 280: unrelated to this PR

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
```diff
@@ -46,6 +46,10 @@ spec:
         - server
         - --config-path=/config/envoy-gateway.yaml
         env:
+        - name: POD_IP
```
arkodg, line 49: its a little fragile to use Envoy Gateway IP type to determine IP type for Envoy Proxy.
instead can we add the env to Envoy Proxy https://github.com/envoyproxy/gateway/blob/b0c6f8ca3c2b55ab1007a800264efe6226100db1/internal/infrastructure/kubernetes/proxy/resource.go#L26 and use that value directly in bootstrap using env substitution ?

zirain: does envoy support something like `${POD_IP}` and check the `POD_IP` and change the listener address to `::1`?

internal/infrastructure/kubernetes/ratelimit/resource.go
zirain: yes, I forgot revert this.

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
arkodg: yeah, chatgpt says yes

zirain: I can give a try, and we need POD_IP for dynamic listener.

zirain: maybe I'm not smart as gpt, but envoy report error with `malformed IP address: ${POD_IP}`.

internal/cmd/envoy/shutdown_manager.go
```diff
@@ -170,8 +170,9 @@ func Shutdown(drainTimeout time.Duration, minDrainDuration time.Duration, exitAt
 
 // postEnvoyAdminAPI sends a POST request to the Envoy admin API
 func postEnvoyAdminAPI(path string) error {
+	// TODO: change bootstrap.AdminAddress() to localhost because there're in the same pod?
 	if resp, err := http.Post(fmt.Sprintf("http://%s:%d/%s",
-		bootstrap.EnvoyAdminAddress, bootstrap.EnvoyAdminPort, path), "application/json", nil); err != nil {
+		bootstrap.AdminAddress(egv1a1.IPv4), bootstrap.EnvoyAdminPort, path), "application/json", nil); err != nil {
```
zirain, line 175: this may not right, what about change it to `localhost`? cc @arkodg

arkodg: +1 to `localhost`, good idea

internal/utils/net/ip.go
```diff
@@ -0,0 +1,50 @@
...
+	return ip.To4() == nil
+}
+
+// IsIPv6FirstPod returns true if the POD_IP environment variable is an IPv6 address.
+// WARNING: This function is only intended to be used in the context of Kubernetes.
+func IsIPv6FirstPod() bool {
+	return IsIPv6(os.Getenv("POD_IP"))
+}
+
+func PreferIPFamily(ipv6First bool, envoyProxy *egv1a1.EnvoyProxy) egv1a1.IPFamily {
```
zhaohuabing, line 34: We should default to IPv4 to align with the API semantics.

```
	// IPFamily specifies the IP family for the EnvoyProxy fleet.
	// This setting only affects the Gateway listener port and does not impact
	// other aspects of the Envoy proxy configuration.
	// If not specified, the system will operate as follows:
	// - It defaults to IPv4 only.
	// - IPv6 and dual-stack environments are not supported in this default configuration.
	// Note: To enable IPv6 or dual-stack functionality, explicit configuration is required.
	// +kubebuilder:validation:Enum=IPv4;IPv6;DualStack
	// +optional
	IPFamily *IPFamily `json:"ipFamily,omitempty"`
```

Or we want to change to:

Default to the IP family of th POD IP of the Envoy Proxy?

I prefer defaulting to IPv4 and leting users to explicitly specify IPv6 if they wenat, as it makes the configuration more explicit.

zhaohuabing, line 34: We should default to IPv4 to align with the API semantics.

```
	// IPFamily specifies the IP family for the EnvoyProxy fleet.
	// This setting only affects the Gateway listener port and does not impact
	// other aspects of the Envoy proxy configuration.
	// If not specified, the system will operate as follows:
	// - It defaults to IPv4 only.
	// - IPv6 and dual-stack environments are not supported in this default configuration.
	// Note: To enable IPv6 or dual-stack functionality, explicit configuration is required.
	// +kubebuilder:validation:Enum=IPv4;IPv6;DualStack
	// +optional
	IPFamily *IPFamily `json:"ipFamily,omitempty"`
```

Or we want to change to:

Default to the IP family of th POD IP of the Envoy Proxy?

Should we always default to IPv4 and leting users to explicitly specify IPv6 if they want?  It will make the configuration more explicit.

internal/cmd/envoy/shutdown_manager.go
zhaohuabing: +1 `localhost` make things easier

internal/utils/net/ip.go
```diff
@@ -0,0 +1,50 @@
...
+	if ipv6First {
```
zhaohuabing, line 35: To align with the `IPFamily` API docs, EG should default to IPv4.

```
	// IPFamily specifies the IP family for the EnvoyProxy fleet.
	// This setting only affects the Gateway listener port and does not impact
	// other aspects of the Envoy proxy configuration.
	// If not specified, the system will operate as follows:
	// - It defaults to IPv4 only.
	// - IPv6 and dual-stack environments are not supported in this default configuration.
	// Note: To enable IPv6 or dual-stack functionality, explicit configuration is required.
	// +kubebuilder:validation:Enum=IPv4;IPv6;DualStack
	// +optional
	IPFamily *IPFamily `json:"ipFamily,omitempty"`
```

Or do we want to change  to:

Default to the IP family of the POD IP of the Envoy Proxy/Envoy Gateway?  

By this way, EG tries to explicitly infer user's intention on the IP Family of the listening port, which may not be correct.

Should we always default to IPv4 and leting users to explicitly specify IPv6 if they want?  It will make the configuration more explicit.

zirain: > Should we always default to IPv4 and require users to explicitly specify IPv6 if desired? It will make the IPFamily choice more explicit.

this will make it's hard to use on IPv6 first cluster.

zhaohuabing: Yes, this approach requires one more `IPFamily` setting for users who want to use IPv6. 

But if we make it implicit, "Default to the IP family of the POD IP of the Envoy Proxy/Envoy Gateway" may be challenging to clearly explain in the EG doc.

zirain: yes, we should do that.

internal/xds/translator/listener.go
```diff
@@ -146,36 +147,6 @@ func originalIPDetectionExtensions(clientIPDetection *ir.ClientIPDetectionSettin
 	return extensionConfig
 }
 
-func setAddressByIPFamily(socketAddress *corev3.SocketAddress, ipFamily *ir.IPFamily, port uint32) []*listenerv3.AdditionalAddress {
```
arkodg, line 149: why does this function need to change ?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -39,14 +40,29 @@ const (
...
+	envoyReadinessAddressIPv6 = "::"
+	EnvoyReadinessPort        = 19001
+	EnvoyReadinessPath        = "/ready"
 
 	defaultSdsTrustedCAPath   = "/sds/xds-trusted-ca.json"
 	defaultSdsCertificatePath = "/sds/xds-certificate.json"
 )
 
+func AdminAddress(family egv1a1.IPFamily) string {
+	if family == egv1a1.IPv6 {
```
arkodg, line 53: will `localhost` here work ?

internal/xds/bootstrap/bootstrap.go
```diff
@@ -39,14 +40,29 @@ const (
...
+		return envoyAdminAddressIPv6
+	}
+	return envoyAdminAddress
+}
+
+func readinessAddress(family egv1a1.IPFamily) string {
```
arkodg, line 59: imo just the changes in this file should be enough to support pure IPv6 case as well as dual stack case, wdyt

zirain: yes

internal/xds/bootstrap/bootstrap.go
zirain: `malformed IP address: localhost`

internal/xds/translator/listener.go
zirain: Before this function change `socketAddress` and return `additionalAddress` in the same time.

internal/xds/bootstrap/bootstrap.go
zirain: need changes around dynamic istener for IPv6 first/pure cluster.

Issue #4264, opened 2024-09-17
HTTPRoute w/ two parent gateways - Both get configured, only one is shown in status, and which switches


BadLiveware:
*Description*:
I have 3 different gateways up and running, only 2 of which are used in this issue.
Only one gateway is shown in status, while both are configured. This results in things like external-dns only setting up DNS records for one gateway address.

*Repro steps*:
Overview:
```shell
─❯ kubectl get envoyproxy -A
NAMESPACE       NAME       AGE
envoy-gateway   cluster    2d22h
envoy-gateway   external   2d22h
envoy-gateway   internal   2d22h

─❯ kubectl get gatewayclass -A
NAME       CONTROLLER                                      ACCEPTED   AGE
cluster    gateway.envoyproxy.io/gatewayclass-controller   True       2d22h
external   gateway.envoyproxy.io/gatewayclass-controller   True       2d22h
internal   gateway.envoyproxy.io/gatewayclass-controller   True       13d

─❯ kubectl get gateway -A
NAMESPACE       NAME       CLASS      ADDRESS          PROGRAMMED   AGE
envoy-gateway   cluster    cluster    10.182.165.187   True         2d22h
envoy-gateway   external   external   10.182.83.109    True         2d22h
envoy-gateway   internal   internal   10.180.0.60      True         13d

─❯ kubectl get httproute -A
NAMESPACE     NAME          HOSTNAMES                                                         AGE
dev-ops       webapi        ["webapi.dev-ops.c.example.org","webapi.dev-ops.i.example.org"]   2d22h # This is attached to internal and cluster
echo-server   echo-server   ["echo-server-gw.dev.example.org"]                                14d # This is attached to external
```

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: internal
  namespace: envoy-gateway
spec:
  gatewayClassName: internal
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    hostname: '*.i.example.org'
    name: http
    port: 80
    protocol: HTTP
status:
  addresses:
  - type: IPAddress
    value: 10.180.0.60
  conditions:
  - lastTransitionTime: "2024-09-14T11:07:57Z"
    message: The Gateway has been scheduled by Envoy Gateway
    observedGeneration: 8
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-09-14T11:09:29Z"
[...]

PR #4587, opened 2024-10-31
fix: HTTPRoute status only shows one parent when targeting multiple Gateways from different GatewayClasses

zhaohuabing:
Fixes: https://github.com/envoyproxy/gateway/issues/4264

Release Notes: Yes


internal/provider/kubernetes/status.go
```diff
@@ -469,6 +469,29 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context, ext
...
 
+// mergeRouteParentStatus merges the old and new RouteParentStatus.
+// This is needed because the RouteParentStatus doesn't support strategic merge patch yet.
+func mergeRouteParentStatus(ns string, old, new []gwapiv1.RouteParentStatus) []gwapiv1.RouteParentStatus {
+	merged := make([]gwapiv1.RouteParentStatus, len(old))
+	_ = copy(merged, old)
+	for _, parent := range new {
+		found := -1
+		for i, existing := range old {
+			if parent.ParentRef.Name == existing.ParentRef.Name &&
```
arkodg, line 480: also && for `parent.ControllerName` && `parent.ParentRef.Kind`

internal/provider/kubernetes/status.go
```diff
@@ -469,6 +469,29 @@ func (r *gatewayAPIReconciler) subscribeAndUpdateStatus(ctx context.Context, ext
...
+				utils.NamespaceDerefOr(parent.ParentRef.Namespace, ns) == utils.NamespaceDerefOr(existing.ParentRef.Namespace, ns) {
```
arkodg, line 481: "k8s.io/utils/ptr"
we've updated a lot of the code in the repo to use `ptr.Deref(parent.ParentRef.Namespace, ns)` instead

internal/provider/kubernetes/status.go
arkodg: do we also need to check for sectionName / listener name if non nil ?

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #4127, opened 2024-08-30
Route with multiple parents has incorrect namespace in parentref status


adam-cattermole:
*Description*:
Creating a route with multiple parents with inferred namespaces shows incorrect namespace in parentRef status

*Repro steps*:
* Create namespace:
```sh
kubectl create namespace test
```
* Create two gateways and a route targeting them:
```sh
kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: gateway-a
  namespace: test
spec:
  gatewayClassName: envoygateway
  listeners:
  - name: default
    port: 80
    protocol: HTTP
    hostname: '*.a.example.com'
    allowedRoutes:
      namespaces:
        from: All
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: gateway-b
  namespace: test
spec:
  gatewayClassName: envoygateway
  listeners:
    - name: default
      port: 80
      protocol: HTTP
      hostname: '*.b.example.com'
      allowedRoutes:
        namespaces:
          from: All
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: targeted-route
  namespace: test
spec:
  hostnames:
    - targeted.a.example.com
    - targeted.b.example.com
  parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: gateway-a
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: gateway-b
  rules:
    - matches:
        - method: GET
          path:
            type: PathPrefix
            value: /toy
EOF
```

* Get the route from the cluster `kubectl get -n test httproute targeted-route -o yaml`:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #4592, opened 2024-10-31
fix: Route with multiple parents has incorrect namespace in parentRef status

zhaohuabing:
Fixes: https://github.com/envoyproxy/gateway/issues/4127

Release Notes: Yes

### DIFF
--- a/internal/gatewayapi/contexts.go
+++ b/internal/gatewayapi/contexts.go
@@ -238,21 +238,26 @@ func GetRouteStatus(route RouteContext) *gwapiv1.RouteStatus {
 	return &rs
 }
 
-// GetRouteParentContext returns RouteParentContext by using the Route
-// objects' ParentReference.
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/contexts.go
```diff
@@ -266,25 +271,19 @@ func GetRouteParentContext(route RouteContext, forParentRef gwapiv1.ParentRefere
 		panic("parentRef not found")
 	}
 
+	// Find the parent in the Route's Status.
 	routeParentStatusIdx := -1
-	defaultNamespace := gwapiv1.Namespace(metav1.NamespaceDefault)
```
zhaohuabing, line 270: This is the root cause of the bug.

zhaohuabing:
@guydc it would be good to also cherry-pick this to v1.1.3.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #4564, opened 2024-10-29
Add ability to ignore health status when active HC is used on endpoint removal


nezdolik:
We use active health checking and observe an issue where endpoints that have been removed from service discovery are stuck in "pending dynamic removal state" and some requests are routed to them (they still pass HC). It would be useful to be able to customise the setting `ignore_health_on_host_removal` on cluster level: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

nezdolik:
@arkodg this is high prio for us, if no one is available to pick this up, i could give it a try?

arkodg:
+1 on setting this by default, if the control plane is removing endpoints, they should be removed immediately
ptal @envoyproxy/gateway-maintainers , would be good to get this into v1.2 if possible

arkodg:
hey @nezdolik as an Envoy Proxy maintainer, would you recommend Envoy Gateway set this to `true` by default :) , is there any case where a user way want to opt out for when running Envoy Gateway in K8s specifically

nezdolik:
@arkodg it looked like to me that Envoy Gateway enables passive healthchecking by default (is this correct?). When passive HC mode is enabled, endpoints are immediately removed from priority when they don't appear in EDS update. Setting this option to `true` by default will align the same behaviour when active HC is used. Having this option set to `false` adds some redundancy layer in case of service discovery outage etc. I think setting it to true is reasonable default setting and if users have other use cases they will let us know.

nezdolik:
@guydc right this is the case, this is why this feature provides extra redundancy layer in case of service discovery issues

arkodg:
I'll go ahead and enable this by default on for static endpoint then

PR #4612, opened 2024-11-01
Set ignore_health_on_host_removal to true for static clusters

arkodg:
Removes the endpoint from the pool faster instead of waiting for the result of the active health check.
Since the control plane already has definitive endpoint health info from the EndpointSlice API, it's safe to set this.

Fixes: https://github.com/envoyproxy/gateway/issues/4564

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -778,6 +778,7 @@ xds:
               ads: {}
               resourceApiVersion: V3
             serviceName: httproute/default/backend/rule/0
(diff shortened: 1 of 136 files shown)


zhaohuabing (approved):
LGTM.
@arkodg the title should be "Set ignore_health_on_host_removal to true for dynamic clusters" ?

arkodg:
> LGTM. @arkodg the title should be "Set ignore_health_on_host_removal to true for dynamic clusters" ?

Static here is static endpoints delivered by EDS

zhaohuabing:
Oh I see. Thanks for the explaination.

guydc (approved):
LGTM

nezdolik:
lgtm

Issue #2940, opened 2024-03-15
Helm chart fails for Flux HelmRelease


webwurst:
*Description*:

Flux HelmChart can't get tags for `oci://docker.io/envoyproxy/gateway-helm`, probably because those don't strictly  follow the [Helm specification](https://helm.sh/docs/topics/charts/#charts-and-versioning):
> More complex SemVer 2 names are also supported, such as version: 1.2.3-alpha.1+ef365. But non-SemVer names are explicitly disallowed by the system.

This means the `v` prefix is not allowed. Happened to others before: https://github.com/fluxcd/helm-controller/issues/670, https://github.com/fluxcd/flux2/issues/3766.

*Repro steps*:
```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: HelmRepository
metadata:
  name: envoyproxy
spec:
  type: oci
  url: oci://docker.io/envoyproxy
---

apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: envoy-gateway
spec:
  chart:
    spec:
      chart: gateway-helm
      version: "~1.0.0"
      sourceRef:
        kind: HelmRepository
        name: envoyproxy
status:
  conditions:
  - lastTransitionTime: "2024-03-15T14:01:03Z"
    message: 'HelmChart ''envoy-gateway/envoy-gateway-envoy-gateway'' is not ready:
      chart pull error: failed to get chart version for remote reference: could not
      get tags for "gateway-helm": unable to locate any tags in provided repository:
      oci://docker.io/envoyproxy/gateway-helm'
````

```bash
$ skopeo list-tags docker://docker.io/envoyproxy/gateway-helm
{
    "Repository": "docker.io/envoyproxy/gateway-helm",
    "Tags": [
        "v0.0.0-latest",
        "v0.4.0",
        "v0.4.0-rc.1",
        "v0.5.0",
        "v0.5.0-rc.1",
        "v0.6.0",
        "v0.6.0-rc.1",
        "v1.0.0",
        "v1.0.0-rc.1"
    ]
}
```

Would be nice to push charts for version 1.0.0 and ongoing without `v`.

zirain:
cc @Xunzhuo

Xunzhuo:
We followed this way to release for our release git tag, image tag, oci helm chart image tag, all of them shared the same value: https://semver.org/spec/v1.0.0.html#tagging-specification-semvertag

[...]

PR #4636, opened 2024-11-06
fix: push a helm chart without v in ther version

zhaohuabing:
Fixes https://github.com/envoyproxy/gateway/issues/2940

Release Notes: Yes

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -57,7 +57,7 @@ jobs:
         shell: bash
         run: |
           echo "release_tag=$(echo ${GITHUB_REF##*/})" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -57,6 +57,7 @@ jobs:
         shell: bash
         run: |
           echo "release_tag=$(echo ${GITHUB_REF##*/})" >> $GITHUB_ENV
+          echo "cleaned_version_tag=${release_tag:1}" >> $GITHUB_ENV
```
arkodg, line 60: can we call this `without_v_release_tag` ?

arkodg (approved):
thanks !

zirain (approved):

PR #4642, opened 2024-11-06
docs for release v1.2.0

zhaohuabing:


site/layouts/shortcodes/yaml-version.html
```diff
@@ -2,9 +2,9 @@
 {{- with (eq $pagePrefix "latest") -}}
 {{- "latest" -}}
 {{- end -}}
-{{- with (strings.HasPrefix $pagePrefix "v1.1") -}}
-{{- "v1.1.3" -}}
```
zirain, line 6: I think you should add lines for v1.2, as it's a major version.

site/layouts/shortcodes/helm-version.html
```diff
@@ -6,5 +6,5 @@
 {{- "v1.1.3" -}}
 {{- end -}}
 {{- with (strings.HasPrefix $pagePrefix "doc") -}}
```
zirain, line 8: same here

site/hugo.toml
```diff
@@ -301,3 +301,7 @@ enable = true
 [[params.versions]]
   version = "v0.2"
   url = "/v0.2"
+
+[[params.versions]]
+  version = "v1.2"
+  url = "/v1.2"
```
zirain, line 307: move to L276? and should be alias to `/doc`?

zirain (approved):

PR #4643, opened 2024-11-06
fix: recover from panics that occur during envoy gateway's reconciliation

### LINKED ISSUE #4332: Recover from reconciler panics
*Description*:
Currently, a panic in the reconciliation flow of Envoy Gateway will lead to EG crashing: #4291, #2661, #1830, #2882. 

Controller frameworks like controller runtime and api-machinery provide the means to recover from panics:
- https://github.com/kubernetes/apimachinery/blob/7f7bf1108918f8accb385727c07e9e66b8b94b10/pkg/util/runtime/runtime.go#L49
- https://github.com/kubernetes-sigs/controller-runtime/pull/1627 

In the context of Envoy Gateway, a reconciliation crash would have several undesired side affects:
- last-known-good XDS caches would be deleted and not recovered after a restart 
- infra manager disrupted during infra reconciliation, possibly creating an inconsistent infra state where only some changes are applied

If a crash occurs during an upgrade, there is a risk that envoy proxies would be replaced (e.g. due to a new proxy version being used), but no configuration is provided by the control plane, leading to a complete outage for users. 

Envoy Gateway should consider recovering from panics by default or allowing users to opt-in for panic recovery. If implemented, metrics should be provided to users, so that operators are made aware of the fact that XDS translation is broken.

liorokman:
This PR catches panics that occur during calls to the watchable infrastructure. All the runners are affected, and panics that occur in them will be logged and reported instead of crashing Envoy Gateway.

Note that the Kubernetes client already recovered from panics caused during calls to the provider specific reconcile - this is the default behavior and the default was never changed.

Fixes #4332

Release Notes: No

### DIFF
--- a/internal/message/watchutil.go
+++ b/internal/message/watchutil.go
@@ -6,6 +6,8 @@
 package message
 
 import (
+	"fmt"
+	"runtime/debug"
 	"time"
 
 	"github.com/telepresenceio/watchable"
@@ -54,6 +56,14 @@ func HandleSubscription[K comparable, V any](
 	go func() {
 		for err := range errChans {
 			logger.WithValues("runner", meta.Runner).Error(err, "observed an error")
+			watchableSubscribeTotal.WithFailure(metrics.ReasonError, meta.LabelValues()...).Increment()
+		}
+	}()
+	defer close(errChans)
+	defer func() {
+		if r := recover(); r != nil {
+			logger.WithValues("runner", meta.Runner).Error(fmt.Errorf("%+v", r), "observed an panic",
+				"stackTrace", string(debug.Stack()))
 			watchableSubscribeTotal.WithFailure(metrics.ReasonError, meta.LabelValues()...).Increment()
 		}
 	}()
--- a/internal/message/watchutil.go
+++ b/internal/message/watchutil.go
@@ -38,6 +38,27 @@ func (m Metadata) LabelValues() []metrics.LabelValue {
 	return labels
 }
 
+// handleWithCrashRecovery calls the provided handle function and gracefully recovers from any panics
+// that might occur when the handle function is called.
+func handleWithCrashRecovery[K comparable, V any](
+	handle func(updateFunc Update[K, V], errChans chan error),
+	update Update[K, V],
+	meta Metadata,
+	errChans chan error,
+) {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.WithValues("runner", meta.Runner).Error(fmt.Errorf("%+v", r), "observed an panic",
+				"stackTrace", string(debug.Stack()))
+			watchableSubscribeTotal.WithFailure(metrics.ReasonError, meta.LabelValues()...).Increment()
+		}
+	}()
+	startHandleTime := time.Now()
+	handle(update, errChans)
+	watchableSubscribeTotal.WithSuccess(meta.LabelValues()...).Increment()
+	watchableSubscribeDurationSeconds.With(meta.LabelValues()...).Record(time.Since(startHandleTime).Seconds())
+}
+
 // HandleSubscription takes a channel returned by
 // watchable.Map.Subscribe() (or .SubscribeSubset()), and calls the
 // given function for each initial value in the map, and for any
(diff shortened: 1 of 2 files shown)


guydc:
Can we also emit some sort of metric, so that users can easily setup alerts on these issues?

liorokman:
> Can we also emit some sort of metric, so that users can easily setup alerts on these issues?

Done.

arkodg:
hey @liorokman thanks for adding this

would you be able to manually simulate a scenario where this code kicks in (when there is an exception in - `provider`, `gateway-api` or `xds-translator`) and avoid the process from panicking thereby allowing the xds server to continue running, which should allow the envoy proxy fleet to continue to scale out and fetch xds. 
a. does scaling out envoy proxies work (can be done using kubectl or hpa) ?
b. after recovery do the watchable runners stabilize, until an update ?
c. if the offending config is fixed, and more config is applied, does new xds get generated ?
d. does any of the above change with multiple replicas of Envoy Gateway

liorokman:
> would you be able to manually simulate a scenario where this code kicks in (when there is an exception in - `provider`, `gateway-api` or `xds-translator`) and avoid the process from panicking thereby allowing the xds server to continue running, which should allow the envoy proxy fleet to continue to scale out and fetch xds.

Manually simulating means writing code that panics on purpose. This isn't something that I would want to commit.

> a. does scaling out envoy proxies work (can be done using kubectl or hpa) ?

As long as Envoy Gateway doesn't crash, in theory there should be no problem scaling out Envoy Proxy.

> b. after recovery do the watchable runners stabilise, until an update ? 

As the unit test shows, if code that is called inside the handler provided to `HandleSubscription` panics, then that panic is contained and recovered. The handler will be called again for the next item in the queue.

If the runners stabilise would depend on the nature of the panic. A bad configuration would probably mean that the panic would occur again in that runner as long as that runner's queue contained something that triggers the panic. But an error in the `gateway-api` runner wouldn't crash any of the other runners, meaning the `xds-translator` would continue working.

> c. if the offending config is fixed, and more config is applied, does new xds get generated ? 

Probably? At some point the queue would trigger a reconcile cycle that doesn't panic. 

But I guess it would depend on the specific bug - if the panic causes some other unexpected effect before it is recovered, then all bets are off. 

> d. does any of the above change with multiple replicas of Envoy Gateway

If all the copies of Envoy Gateway that are running have the same bug causing a panic, then all of them should panic in the same place given the same configuration. Shouldn't matter how many copies are running if the bug is deterministic.

arkodg:
thanks @liorokman, would be great if we simulate this in a fork to make sure its happening

arkodg:
> > thanks @liorokman, would be great if we simulate this in a fork to make sure its happening
> 
> @arkodg
> 
> I created a local branch where I could trigger panics on demand in both the gatewayAPI and xds-translator runners.
> 
> I verified that causing a panic in either of these doesn't crash the XDS server by scaling the envoy proxy deployment while these runners were panic-ing and making sure that all instances of Envoy Proxy were configured correctly.
> 
> I verified that if the configuration is changed so that the panic doesn't occur, then everything resumes work as expected.
> 
> However:
> 
> If there are panics in either of these runners, and during that time the Envoy Gateway deployment is scaled, then new instances of Envoy Gateway are unable to generate a locally cached working XDS configuration. If, during this time, the Envoy Proxy deployment is scaled, and the new Envoy Proxy instances happen to reach one of the new Envoy Gateway deployments for their XDS configuration, then these instances will not have a valid XDS configuration and traffic routed through them will not work.
> 
> We can workaround this by making sure that the Envoy Gateway pods are not considered "healthy" until after the XDS translation has occurred at least once. @alexwo suggested a PR (#2918) to this effect at one point, but it wasn't merged.

thanks for testing this !
lets discuss the best way to solve the issue of this new case that arises with change - running xds servers that have an empty cache, in a new GH issue

internal/message/metrics.go
```diff
@@ -13,6 +13,11 @@ var (
 		"Current depth of watchable queue.",
 	)
 
+	panicCounter = metrics.NewCounter(
+		"panics_recovered_total",
```
arkodg, line 17: should this have the `watchable` prefix ?
cc @shawnh2

internal/message/watchutil.go
```diff
@@ -36,6 +38,28 @@ func (m Metadata) LabelValues() []metrics.LabelValue {
...
+// that might occur when the handle function is called.
+func handleWithCrashRecovery[K comparable, V any](
+	handle func(updateFunc Update[K, V], errChans chan error),
+	update Update[K, V],
+	meta Metadata,
+	errChans chan error,
+) {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.WithValues("runner", meta.Runner).Error(fmt.Errorf("%+v", r), "observed an panic",
```
arkodg, line 51:
```suggestion
			logger.WithValues("runner", meta.Runner).Error(fmt.Errorf("%+v", r), "observed a panic",
```

guydc (approved):

PR #4654, opened 2024-11-07
[release/v1.2] fix panic in provider when the direct response body is nil (#4647)

arkodg:
body is an optional field, it may be nil

(cherry picked from commit ad518329bd6025fb56493ff75b39029263de2d02)


zhaohuabing (approved):
Thanks Arko!

zirain (approved):

PR #4662, opened 2024-11-07
fix: loosen JWT issuer validation

### LINKED ISSUE #4534: Could we loose jwtProvider validation for issuer ?
*Description*:
>Describe the issue.

We use jwt authentication
```
---
kind: SecurityPolicy
metadata:
  name: security-policy
  namespace: eg-test
spec:
  targetRefs:
      kind: HTTPRoute
      name: my-http-route
  jwt:
    providers:
      - name: foo
        issuer: foo.<company-domain-name>.com
        remoteJWKS:
          uri: https://<mysts>/jwks/<foo>
```
But we see error in gateway controller log
```
2024-10-26T03:32:32.610Z	ERROR	gateway-api	runner/runner.go:186	unable to validate xds ir, skipped sending it	{"runner": "gateway-api", "error": "invalid issuer; must be a URL or email address: mail: missing '@' or angle-addr\ninvalid issuer; must be a URL or email address: mail: missing '@' or angle-addr"}
2024-10-26T03:32:32.610Z	ERROR	watchable	message/watchutil.go:56	observed an error	{"runner": "gateway-api", "error": "invalid issuer; must be a URL or email address: mail: missing '@' or angle-addr\ninvalid issuer; must be a URL or email address: mail: missing '@' or angle-addr"}
```
It is caused by this [issuer validation check](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/validation/securitypolicy_validate.go#L71-L78). However, the `iss` in our use case is `foo.<company-domain-name>.com`. Do you think if we can loose the validation check ?

[optional *Relevant Links*:]

ardikabs:
**What type of PR is this?**

This PR fixes JWT issuer validation, to loosen the validation constraint from 'either URI or Email' to

- Per [RFC-7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1), when the issuer contains ':', it MUST be a valid URI.
- Adding additional constraint, when the issuer contains '@', it MUST be a valid email address.
- Otherwise, accept arbitrary string.

Fixes #4534 

Release Notes: Yes/No


zhaohuabing:
@ardikabs Thanks for working on this. This change deserves a release note. Can you add a description to `release-notes/current.yaml` and include it in this PR?

zhaohuabing:
Dumb question: why do we need to enforce the email validation? According to [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1), it's not required.

> StringOrURI
      A JSON string value, with the additional requirement that while
      arbitrary string values MAY be used, any value containing a ":"
      character MUST be a URI [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)].  StringOrURI values are
      compared as case-sensitive strings with no transformations or
      canonicalizations applied.

ardikabs:
^ also defer to @arkodg about the email constraint

arkodg (approved):
LGTM thanks !

zirain (approved):

zhaohuabing (approved):
It's reasonable as the email validation alreay exits.
Thanks!

PR #4663, opened 2024-11-07
listen on ipv4 and ipv6 localhost addrs for admin server

arkodg:
Relates to https://github.com/envoyproxy/gateway/issues/4565

Release Notes: Yes

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -15,8 +15,9 @@ envoyProxyForGatewayClass:
               path: /dev/null
           address:
             socket_address:
-              address: 127.0.0.1
+              address: ::1
               port_value: 19000
+              ipv4_compat: true
         layered_runtime:
           layers:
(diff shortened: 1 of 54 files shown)


internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
```diff
@@ -15,8 +15,9 @@ envoyProxyForGatewayClass:
               path: /dev/null
           address:
             socket_address:
-              address: 127.0.0.1
```
zirain, line 18: plan to use this as the default value even for IPv4 cluster?

internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
```diff
@@ -15,8 +15,9 @@ envoyProxyForGatewayClass:
...
+              address: ::1
               port_value: 19000
+              ipv4_compat: true
```
zirain, line 20: won't work on `127.0.0.1`

zirain:
@arkodg I think we can close this.

PR #4671, opened 2024-11-07
feat: add body to ext auth

### LINKED ISSUE #4431: Add support for with_request_body in SecurityPolicy.spec.extAuth
Envoy supports sending the full request body to the external authorization server via the with_request_body filter configuration. Do you think that it is possible to expose such feature on the SecurityPolicy.spec.extAuth api?

My use case specifically relates to LLM model serving with an OpenAI compatible server. There you receive a request JSON body with a model field. I would like to have an opa-envoy authz server that is able to authorize requests based on the user identity and the model requested.

*Relevant Links*:
- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter#configuration-examples

AurelienPillevesse:
**What type of PR is this?**
Add gesture for body to ext auth

Fixes #4431


AurelienPillevesse:
The testing part (https://gateway.envoyproxy.io/contributions/develop/#testing) seems not to work on my computer so I tried to modify `internal/gatewayapi/testdata/securitypolicy-with-extauth-backend.in.yaml` and `internal/gatewayapi/testdata/securitypolicy-with-extauth-backend.out.yaml` manually and hope that it will pass.

Don't really know how to do it better.

zirain:
it would be good to send API first.

AurelienPillevesse:
From `api/` or `internal/gatewayapi/` or both ?

api/v1alpha1/ext_auth_types.go
```diff
@@ -33,6 +33,8 @@ type ExtAuth struct {
 	// +optional
 	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
 
+	BodyToExtAuth *string `json:"bodyToExtAuth,omitempty"`
```
arkodg, line 36: why this a string type and not a bool

AurelienPillevesse: I don't know how it works so I don't know if it will receive a string which is the body of the request or a bool to activate it because `HeadersToExtAuth` is an array of string

arkodg: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#envoy-v3-api-msg-extensions-filters-http-ext-authz-v3-buffersettings is the underlying API in envoy

do we need to expose `max_request_bytes` to the user ?

AurelienPillevesse: I don't think that's necessary for the moment

jonbk: In my opinion, all configuration options for with_request_body should be exposed. Otherwise, we risk facing the same issues that this PR is intended to solve, namely the lack of control over Envoy Proxy from the gateway. For instance, in the case of max_request_bytes, the Envoy documentation specifies that if the body exceeds this limit, Envoy will return HTTP 413 and will not initiate the authorization process once the buffer reaches the set value. This behavior can be a blocker, especially if this parameter isn’t configurable via the gateway.

Thank you very much for the great work!

api/v1alpha1/ext_auth_types.go
```diff
@@ -33,6 +33,8 @@ type ExtAuth struct {
 	// +optional
 	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
 
+	BodyToExtAuth *bool `json:"bodyToExtAuth,omitempty"`
```
zhaohuabing, line 36: Should we make `BodyToExtAuth` an empty structure for better future-proofing?

internal/xds/translator/extauth.go
```diff
@@ -117,6 +117,14 @@ func extAuthConfig(extAuth *ir.ExtAuth) *extauthv3.ExtAuthz {
 		})
 	}
 
+	if extAuth.BodyToExtAuth != nil {
+		config.WithRequestBody = &extauthv3.BufferSettings{
+			MaxRequestBytes:     1024,
+			AllowPartialMessage: false,
+			PackAsBytes:         false,
+		}
```
AurelienPillevesse, line 125: Arbitrary value (but must be greater than 0) for the `MaxRequestBytes` field, given that the `AllowPartialMessage` field is always equal to false.

zhaohuabing:
cc @guydc

api/v1alpha1/ext_auth_types.go
```diff
@@ -33,6 +33,8 @@ type ExtAuth struct {
 	// +optional
 	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
 
+	BodyToExtAuth *BodyToExtAuth `json:"bodyToExtAuth,omitempty"`
```
zhaohuabing, line 36: Nit: can you add a line of comment to explain this field?
Also mark it as `+optional`

api/v1alpha1/ext_auth_types.go
```diff
@@ -33,6 +33,10 @@ type ExtAuth struct {
 	// +optional
 	HeadersToExtAuth []string `json:"headersToExtAuth,omitempty"`
 
+	// BodyToExtAuth defines the Body to Ext Auth configuration.
+	// +optional
+	BodyToExtAuth *BodyToExtAuth `json:"bodyToExtAuth,omitempty"`
```
guydc, line 38: Maybe add a comment about the maximum body size supported for this option? If I understand the docs correctly, when the body size is greater than 1024, the request would be blocked with status code 413 instead of being sent to ext-auth.

guydc (commented):
overall looks good to me, just one comment on the docs and please rebase + fix test code generation

api/v1alpha1/ext_auth_types.go
AurelienPillevesse: In my understanding, the maximum body size will not be take into account but in the `BufferSettings` object, `AllowPartialMessage` is hardcoded to false for the moment (so I put `1024` as random value).

```go
extauthv3.BufferSettings{
	MaxRequestBytes:     1024,
	AllowPartialMessage: false,
	PackAsBytes:         false,
}
```

In my opinion, it's better the release this feature like that (so basic feature first) and in next iteration for this feature, allow the possibility for partial message.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-buffersettings

dromadaire54: I'm agree with that if the `AllowPartialMessage` values is false by default I think we can open another PR about that. This feature is very important because now we cannot add a body.

guydc: Tested the following config:

```yaml
static_resources:
  listeners:
    - address:
        socket_address:
          address: 0.0.0.0
          port_value: 8000
      filter_chains:
        - filters:
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
                codec_type: AUTO
                stat_prefix: ingress_http
                route_config:
                  name: local_route
                  virtual_hosts:
                    - name: upstream
                      domains:
                        - "*"
                      typed_per_filter_config:
                        envoy.filters.http.ext_authz:
                          "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
                          check_settings:
                            with_request_body:
                              max_request_bytes: 1
                              allow_partial_message: false
                      routes:
                        - match:
                            prefix: "/"
                          route:
                            cluster: upstream-service
                http_filters:
                  - name: envoy.filters.http.ext_authz
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
                      http_service:
                        server_uri:
                          uri: ext_authz
                          cluster: ext_authz-http-service
                          timeout: 0.250s
                  - name: envoy.filters.http.router
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router

  clusters:
    - name: upstream-service
      type: STRICT_DNS
      lb_policy: ROUND_ROBIN
      load_assignment:
        cluster_name: upstream-service
        endpoints:
          - lb_endpoints:
              - endpoint:
                  address:
                    socket_address:
                      address: upstream-service
                      port_value: 8080

    - name: ext_authz-http-service
      type: STRICT_DNS
      lb_policy: ROUND_ROBIN
      load_assignment:
        cluster_name: ext_authz-http-service
        endpoints:
          - lb_endpoints:
              - endpoint:
                  address:
                    socket_address:
                      address: ext_authz-http-service
                      port_value: 9002
```

Specifying a body large enough will lead to request being rejected with 413 without being inspected by the ext-auth server. 

```sh
curl -I http://localhost:8000
HTTP/1.1 403 Forbidden
date: Fri, 22 Nov 2024 15:30:44 GMT
server: envoy
content-length: 0
```

No body: rejected due to fail closed (ext auth doesn't really exist in this example). 

```sh
curl -X POST http://localhost:8000 -d "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
[...]
< HTTP/1.1 413 Payload Too Large
< content-length: 17
< content-type: text/plain
< date: Fri, 22 Nov 2024 15:33:56 GMT
< server: envoy
< connection: close
Payload Too Large
```

Large body: rejected before being processed by ext auth.

guydc (approved):
LGTM, thanks.

api/v1alpha1/ext_auth_types.go
arkodg: can we only support configuring `max_request_bytes` for now ? and set `AllowPartialMessage` and `PackAsBytes` as `false` . imo grpcService authors will most likely using ext proc over this so we dont need to expose `PackAsBytes`

api/v1alpha1/ext_auth_types.go
```diff
@@ -85,3 +89,23 @@ type HTTPExtAuthService struct {
 	// +optional
 	HeadersToBackend []string `json:"headersToBackend,omitempty"`
 }
+
+// BodyToExtAuth defines the Body to Ext Auth configuration
+type BodyToExtAuth struct {
+	// MaxRequestBytes is the maximum size of a message body that the filter will hold in memory.
+	// Envoy will return HTTP 413 and will not initiate the authorization process when buffer
+	// reaches the number set in this field.
+	// Note that this setting will have precedence over failureModeAllow.
```
arkodg, line 98:
```suggestion
	// Note that this setting will have precedence over failOpen mode.
```

api/v1alpha1/ext_auth_types.go
```diff
@@ -85,3 +89,23 @@ type HTTPExtAuthService struct {
...
+	MaxRequestBytes uint32 `json:"maxRequestBytes"`
```
arkodg, line 99:
```suggestion
	MaxRequestBytes *uint32 `json:"maxRequestBytes"`
```

arkodg (approved):
LGTM thanks !

guydc (approved):

Issue #4001, opened 2024-08-05
CRDs upgrade support


shahar-h:
*Description*:
Currently Gateway API and EG CRDs are located in EG `gateway-helm` chart crds folder. This means that CRDs are not upgraded on chart upgrade, and need to be manually upgraded beforehand.

Possible solutions:
1. Move crds inside templates.
Real world example: cert-manager chart has CRDs as part of templates folder and exposes `crds.enabled` flag.
It also allows to decide if we want to keep the CRDs on chart uninstallation with `crds.keep` flag, by leveraging `"helm.sh/resource-policy": keep` helm annotation.
See: https://cert-manager.io/v1.13-docs/installation/helm/#crd-considerations.
However, this is a breaking change for consumers that use EG chart as a sub chart and have custom resources as part of the main chart.
2. Split the current chart into crds chart and application chart, and instruct to install the crds chart first.
Real world example: Istio provides a base chart that installs CRDs: https://istio.io/latest/docs/setup/install/helm/#installation-steps.

Both 1 and 2 can also solve https://github.com/envoyproxy/gateway/issues/3094 by providing a separate flags(1)/charts(2) for Gateway API and EG CRDs.

Also Related:
* https://github.com/envoyproxy/gateway/issues/2031
* https://github.com/envoyproxy/gateway/issues/1721

WDYT?

shahar-h:
While trying to split the chart into 2 charts(app chart and CRDs chart) internally CRDs chart installation failed because helm release secret exceeded max 1MB size limit. In order to resolve that issue I split the CRDs chart into 2 more charts - eg CRDS and gateway api CRDs.

arkodg:
thinking out loud, we could add two additional templates (w/o deleting the `/crds` dir) for the CRDs - one for `gateway-api` and one for `envoy-gateway` and use knobs such as `applyCrds.all` `applyCrds.gateway-api` and `applyCrds.envoy-gateway`

This solves 2 use cases

* For cases where the cloud provider like GKE installs the gateway-api resources, a user could install EG with `helm install --skip-crds --set applyCrds.envoy-gateway=true`
* For upgrade cases , users could run `helm upgrade --set applyCrds.all=true`

the naming for the helm knobs can be improved, of course

shahar-h:
[...]

PR #4675, opened 2024-11-07
docs: add a note of helm not updating CRDs in the upgrade section

arkodg:


zirain:
should consider add a new helm chart called `gateway-crds-helm`?

zirain (approved):

arkodg:
> should consider add a new helm chart called `gateway-crds-helm`?

this being tracked with https://github.com/envoyproxy/gateway/issues/4001

zhaohuabing (approved):

PR #4681, opened 2024-11-08
fix: tcp listener is rejected when no route attached

### LINKED ISSUE #4680: TCP listener creation failed when there is no TCP route defined
error:
[2024-11-08 04:34:43.214][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) gateway-conformance-infra/same-namespace/tls: error adding listener '0.0.0.0:10443': no filter chains specified

configuration:
```yaml
kind: Gateway
metadata:
  name: same-namespace
spec:
  gatewayClassName: "{GATEWAY_CLASS_NAME}"
  listeners:
    - name: tls
      protocol: TLS
      port: 443
      tls:
        mode: Passthrough
      allowedRoutes:
        namespaces:
          from: Same
```

zhaohuabing:
Create a default route pointing to a static cluster without endpoints when there is no TCPRoutes attached to a TCP listener. This avoids Envoy rejecting a listener without a filter chain and reporting warning.

Fixes #4680

Release Notes: Yes

Alternative consideration:  
An alterantive approach could be to not create the xDS listener, but this behavior doesn't align with the Gateway semantics(which does defines a listener) and the behavior of the empty HTTP listener.

Other type of listeners:
* A HTTP listener without HTTPRoute will create the xDS HTTP listener and an empty RouteConfiguration.
* An UDP listener without UDPRoute won't create the xDS UDP listerner at all.


arkodg:
thanks for looking into this @zhaohuabing 
can you share what the flow looks like from a client perspective (e.g. using curl) when we either dont create a tcp listener vs tcp listener with dummy filter chain

zhaohuabing:
> thanks for looking into this @zhaohuabing can you share what the flow looks like from a client perspective (e.g. using curl) when we either dont create a tcp listener vs tcp listener with dummy filter chain

can't connect vs connection reset by peer

no tcp listener:

```
curl -v http://${GATEWAY_HOST}:8080
*   Trying 172.18.0.200:8080...
* connect to 172.18.0.200 port 8080 from 172.18.0.1 port 54918 failed: Connection refused
* Failed to connect to 172.18.0.200 port 8080 after 0 ms: Couldn't connect to server
* Closing connection
curl: (7) Failed to connect to 172.18.0.200 port 8080 after 0 ms: Couldn't connect to server

```

dummy filter chain

```
curl -v http://${GATEWAY_HOST}:8080                    
*   Trying 172.18.0.200:8080...
* Connected to 172.18.0.200 (172.18.0.200) port 8080
> GET / HTTP/1.1
> Host: 172.18.0.200:8080
> User-Agent: curl/8.5.0
> Accept: */*
> 
* Recv failure: Connection reset by peer
* Closing connection
curl: (56) Recv failure: Connection reset by peer
```

arkodg:
hey @zhaohuabing I prefer if we dont generate a xds listener for this case

zirain:
> hey @zhaohuabing I prefer if we dont generate an xds listener for this case wdyt @envoyproxy/gateway-maintainers

it would be better if possiable.

arkodg:
from @haq204's comment https://github.com/envoyproxy/gateway/issues/4680#issuecomment-2483856313 , we may want to have a listener up for responding to TCP health checks

internal/xds/translator/translator.go
```diff
@@ -627,6 +631,31 @@ func (t *Translator) processTCPListenerXdsTranslation(
 				errs = errors.Join(errs, err)
 			}
 		}
+
+		// If there are no routes, add a route without a destination to the listener to create a filter chain
+		// This is needed because Envoy requires a filter chain to be present in the listener, otherwise it will reject the listener and report a warning
+		if len(tcpListener.Routes) == 0 {
+			nullRouteCluster := &clusterv3.Cluster{
```
arkodg, line 638: hey curious, what happens if we set `xdsListener.FilterChains = []` for this case, does xDS still complain ?

zhaohuabing: Envoy will complain `no filter chains specified`. It's a validation inside the Envoy listener impl instead of a proto type built-in validation.

https://github.com/envoyproxy/envoy/blob/79958991ffe0cf8d59a4a351646c76d672dada83/source/common/listener_manager/listener_impl.cc#L763-L770

```
type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) gateway-conformance-infra/same-namespace/tls: error adding listener '0.0.0.0:10443': no filter chains specified
```

zhaohuabing: Envoy will complain `no filter chains specified`. It's a validation inside the Envoy listener impl instead of a proto type built-in validation.

https://github.com/envoyproxy/envoy/blob/79958991ffe0cf8d59a4a351646c76d672dada83/source/common/listener_manager/listener_impl.cc#L763-L770

```
type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) gateway-conformance-infra/same-namespace/tls: error adding listener '0.0.0.0:10443': no filter chains specified
```

internal/xds/translator/testdata/in/xds-ir/listener-connection-limit.yaml
```diff
@@ -44,26 +44,28 @@ tcp:
...
-    passthrough:
-      snis:
-      - bar.com
-  destination:
-    name: "tls-route-dest"
-    settings:
-    - endpoints:
-      - host: "1.2.3.4"
-        port: 50000
+  routes:
```
zhaohuabing, line 47: This is a bug in the listener-connection-limit test. The TCPRoute structure is incorrect.

internal/xds/translator/testdata/in/xds-ir/listener-connection-limit.yaml
```diff
@@ -44,26 +44,28 @@ tcp:
...
       value: 10
       closeDelay: 3s
   port: 10083
-  destination:
-    name: "tcp-route-dest"
-    settings:
-    - endpoints:
-      - host: "1.2.3.4"
-        port: 50000
+  routes:
```
zhaohuabing, line 65: This is a bug in the listener-connection-limit test. The TCPRoute in the TCP Listener ir is incorrect.

internal/xds/translator/testdata/in/xds-ir/tcp-endpoint-stats.yaml
```diff
@@ -5,11 +5,12 @@ tcp:
...
   port: 10080
-  destination:
-    name: "tcp-route-simple-dest"
-    settings:
-    - endpoints:
-      - host: "1.2.3.4"
-        port: 50000
-      - host: "5.6.7.8"
-        port: 50001
+  routes:
```
zhaohuabing, line 8: This is a bug in the tcp-endpoint-stats test. The TCPRoute in the TCP Listener ir is incorrect.

internal/xds/translator/testdata/in/xds-ir/tcp-req-resp-sizes-stats.yaml
```diff
@@ -5,11 +5,12 @@ tcp:
...
   port: 10080
-  destination:
-    name: "tcp-route-simple-dest"
-    settings:
-    - endpoints:
-      - host: "1.2.3.4"
-        port: 50000
-      - host: "5.6.7.8"
-        port: 50001
+  routes:
```
zhaohuabing, line 8: This is a bug in the tcp-req-resp-sizes-stats test. The TCPRoute in the TCP Listener ir is incorrect.

internal/xds/translator/testdata/in/extension-xds-ir/extensionpolicy-tcp-udp-http.yaml
```diff
@@ -66,6 +75,13 @@ tcp:
           controllerName: gateway.envoyproxy.io/gatewayclass-controller
   name: envoy-gateway/gateway-1/tcp1
   port: 10080
+  routes:
+  - destination:
```
zhaohuabing, line 79: Same here.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4500, opened 2024-10-23
api: support setting trusted CIDRs

rudrakhp:
**What type of PR is this?**

api: support setting trusted CIDRs

**What this PR does / why we need it**:

To configure Envoy API's [xff_trusted_cidrs](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/http/original_ip_detection/xff/v3/xff.proto#L48) param.

**Which issue(s) this PR fixes**:

Ref #4489 

Release Notes: Yes

Issue #4489, opened 2024-10-21
Allow setting of trusted CIDR blocks for X-Forwarded-For


robd003:
*Description*:
Last week Envoy just released support for trusted CIDR blocks when processing X-Forwarded-For headers.

It would be really useful to have a default configuration mode called "cloudflare" which would automatically load the IP ranges from these two dynamically updated lists:

Cloudflare IPv4 CIDR blocks: https://www.cloudflare.com/ips-v4/
Cloudflare IPv6 CIDR blocks: https://www.cloudflare.com/ips-v6/

[optional *Relevant Links*:]
https://github.com/envoyproxy/envoy/pull/31831
https://github.com/envoyproxy/envoy/releases/tag/v1.32.0

zhaohuabing:
A knob for configuring CIDR can be added to the `ClientTrafficPolicySpec.ClientIPDetection.XForwardedForSettings`,


```
NumTrustedHops: 2
TrustedCIDRs:
    - 173.245.48.0/20
    - 103.21.244.0/22
```

Given that [the IP lists are updated only once a year](https://www.cloudflare.com/ips/), I believe it’s reasonable to just manually update the CTP whenever Cloudflare makes changes.

arkodg:
+1 to a generic `trustedCIDRs`

rudrakhp:
@arkodg I would like to contribute to this, please assign if help wanted. Thanks!

robd003:
Can we keep this open until the code is actually released (1.3.0?) with documentation?

rudrakhp:
@arkodg do we have anything pending in this issue?

arkodg:
@rudrakhp are we tracking the regression associated with losing out on `X-Envoy-Internal-Address` anywhere ?

PR #4702, opened 2024-11-11
feat:support configuring xff trusted cidrs

rudrakhp:
**What type of PR is this?**

feat:support configuring xff trusted cidrs

Implement API introduced in envoyproxy/gateway#4500 

Refer #4489 

Release Notes: No


arkodg:
hey @rudrakhp can we add an e2e for this in this PR, or can you confirm this works by testing manually (and e2e can be added in a follow up)

rudrakhp:
@arkodg I tried to use original IP detection extension's `xff_num_trusted_hops` instead of the same param in the parent HCM filter (which is on deprecation path according to [envoy community](https://envoyproxy.slack.com/archives/C78M4KW76/p1731427567505609?thread_ts=1731239979.305009&cid=C78M4KW76)). But looks like it's breaking the way we build our RBAC filters for client CIDRs today ([this](https://github.com/envoyproxy/gateway/blob/639a441e05a1aca94b6064d9d838a09be375bd8a/test/e2e/tests/authorization_client_ip.go#L28) e2e was failing). 
So I decided to fall back to using only `xff_trusted_cidrs` in the new extension to ensure no change in `NumTrustedHops` behaviour. Just an FYI in case we might decide to use the extension for`NumTrustedHops` as well in the future which will require us to fix the RBAC.
I will add an E2E for trusted CIDRs in this PR meanwhile.

zhaohuabing:
> @arkodg I tried to use original IP detection extension's `xff_num_trusted_hops` instead of the same param in the parent HCM filter (which is on deprecation path according to [envoy community](https://envoyproxy.slack.com/archives/C78M4KW76/p1731427567505609?thread_ts=1731239979.305009&cid=C78M4KW76)). But looks like it's breaking RBAC somehow ([this](https://github.com/envoyproxy/gateway/blob/639a441e05a1aca94b6064d9d838a09be375bd8a/test/e2e/tests/authorization_client_ip.go#L28) e2e was failing). Started a thread [envoyproxy/envoy#37140](https://github.com/envoyproxy/envoy/issues/37140) to understand this further. Meanwhile only using `xff_trusted_cidrs` in the new extension to ensure no change in `NumTrustedHops` behaviour. Just an FYI in case we might decide to use the extension for `NumTrustedHops` as well in the future which will require us to fix this. I will add E2E for trusted CIDRs in this PR to ensure functionality.

@rudrakhp Probably related to https://github.com/envoyproxy/envoy/issues/34241

rudrakhp:
According to [HCM docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto):
>Extensions cannot be used in conjunction with [use_remote_address](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-use-remote-address) nor [xff_num_trusted_hops](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-xff-num-trusted-hops).

But due to known issues mentioned earlier, not able to move legacy num trusted hops to the new extension. Moving PR to draft till the envoy issue is resolved.

zhaohuabing:
@rudrakhp 


> But due to known issues mentioned earlier, not able to move legacy num trusted hops to the new extension. Moving PR to draft till the envoy issue is resolved.

To move this PR ahead, I think we can modify the existing RBAC e2e.
```
numTrustedHops: 2
```
https://github.com/envoyproxy/gateway/blob/2a10d474fdfe2f3a9689266355fa6d89a042daac/test/e2e/testdata/authorization-client-ip.yaml#L94

But I'm not sure which behavior is correct, the HCM [xff_num_trusted_hops](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-xff-num-trusted-hops) or the xff extension.

zhaohuabing:
Raised a PR https://github.com/envoyproxy/envoy/pull/37780 to seek help from the Enovy maintainers.

rudrakhp:
>To move this PR ahead, We probably could modify the existing RBAC e2e.

Not sure if that's the right thing to do, isn't it essentially changing behaviour for EGW users, probably breaking?

internal/xds/translator/listener.go
```diff
@@ -141,6 +147,35 @@ func originalIPDetectionExtensions(clientIPDetection *ir.ClientIPDetectionSettin
 			Name:        "envoy.extensions.http.original_ip_detection.custom_header",
 			TypedConfig: customHeaderConfigAny,
 		})
+	} else if clientIPDetection.XForwardedFor != nil {
+		var xffHeaderConfigAny *anypb.Any
+		if clientIPDetection.XForwardedFor.TrustedCIDRs != nil {
+			trustedCidrs := make([]*corev3.CidrRange, 0)
+			for _, cidr := range clientIPDetection.XForwardedFor.TrustedCIDRs {
+				parsedCidr := strings.Split(string(cidr), "/")
```
arkodg, line 155: can we use `net.ParseCIDR` here to derive prefix and mask ?

internal/xds/translator/testdata/out/xds-ir/client-ip-detection.listeners.yaml
```diff
@@ -19,15 +19,20 @@
...
         rds:
           configSource:
             ads: {}
             resourceApiVersion: V3
           routeConfigName: first-listener
         serverHeaderTransformation: PASS_THROUGH
         statPrefix: http-8081
-        useRemoteAddress: true
-        xffNumTrustedHops: 2
+        useRemoteAddress: false
```
arkodg, line 35: since `skipXffAppend: false` exists, we will maintain backwards compatibility where the gateway ip will be appended to the off ?

rudrakhp: Yes, when `use_remote_address: true` was used earlier, HCM's `skip_xff_append` is used, which defaults to `false` (see [this](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-skip-xff-append)). 
When moving to to the extension and setting `use_remote_address: false`, we will have to set the extension's `skip_xff_append`  to `false` explicitly since it defaults to `true` (see [this](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/original_ip_detection/xff/v3/xff.proto#envoy-v3-api-field-extensions-http-original-ip-detection-xff-v3-xffconfig-skip-xff-append)).

arkodg (approved):
LGTM thanks !

mathetake:
the e2e test failure on ipv6 seems not a flake but a real failure - making the CI broken on main

PR #4707, opened 2024-11-12
fix: translator reports errors for existing clusters and secretes

### LINKED ISSUE #4706: OIDC authentication and JWT authorization is unstable
*Description*:

When attempting to use OIDC authentication and pass that down to the new JWT authorization feature in v1.2.0, Envoy Gateway seems to fail to program the Envoy Proxy in a stable way.  If I apply my Security Policy for an HTTP Route as a whole unit, most of the time I see errors around fetching the JWT public key with the message `OAuth flow failed` in the browser.  Another time I only saw the OAuth flow failed without a corresponding reason in the Envoy Proxy log.  If I apply the Security Policy with the oidc section and then add the jwt section and finally add the authorization section, I am able to create a working configuration that appropriately authenticates and authorizes the returned JWT token.

If more logs are required, I can attempt to deploy a minikube and grab logs there.

*Repro steps*:

Sample resources:

```
---
kind: HTTPRoute
metadata:
  name: prometheus
  namespace: monitoring
spec:
  hostnames:
  - prometheus.example.com
  parentRefs:
    kind: Gateway
    name: example
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: prometheus
      port: 9090
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
---
kind: SecurityPolicy
metadata:
  name: prometheus
  namespace: monitoring
spec:
  authorization:
    rules:
    - action: Allow
      name: allow
      principal:
        jwt:
          claims:
          - name: groups
            valueType: StringArray
            values:
            - foobar
          provider: exjwt
  jwt:
    providers:
    - claimToHeaders:
      - claim: email
        header: x-user-email
      extractFrom:
        cookies:
        - IdToken
      issuer: https://oidc.example.com/auth/realms/example
      name: exjwt
      remoteJWKS:
        uri: https://oidc.example.com/auth/realms/example/protocol/openid-connect/certs
  oidc:
    clientID: prometheus
    clientSecret:
      group: ""
      kind: Secret
      name: prometheus-oidc
    cookieNames:
      idToken: IdToken
    provider:
      issuer: https://oidc.example.com/auth/realms/example
    scopes:
    - openid
    - email
    - profile
  targetRefs:
    kind: HTTPRoute
    name: prometheus
```

*Environment*:

Envoy Gateway v1.2.1
Envoy Proxy v1.32.1

*Relevant Logs*:

In Envoy Proxy when deploying as a whole unit:

```
[2024-11-11 21:42:19.638][1][error][filter] [source/extensions/filters/http/common/jwks_fetcher.cc:51] fetch: fetch pubkey [uri = https://oidc.example.com/auth/realms/example/protocol/openid-connect/certs] failed: [cluster = oidc_example_com_443] is not configured
[2024-11-11 21:42:19.638][1][warning][jwt] [source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc:115] Jwks async fetching url=https://oidc.example.com/auth/realms/example/protocol/openid-connect/certs: failed
```

In Envoy Gateway when deploying as a whole unit:

```
2024-11-11T21:42:19.625Z	ERROR	xds-translator	runner/runner.go:85	failed to translate xds ir	{"runner": "xds-translator", "error": "xds cluster exists\nxds cluster exists"}
2024-11-11T21:42:19.625Z	ERROR	watchable	message/watchutil.go:56	observed an error	{"runner": "xds-translator", "error": "xds cluster exists\nxds cluster exists"}
```

zhaohuabing:
Fixes #4706 xDS translation failed when oidc tokenEndpoint and jwt remoteJWKS are specified within the same security policy and using the same hostname

Refactor:  skips adding the cluster/secrets and returns nil to make the code cleaner and easier to maintain. It's safe to remove `ErrXdsClusterExists` and `ErrXdsSecretsExists` as they don't need to be handled in any places.

Release Notes: Yes

Test before the fix:

```
--- FAIL: TestTranslateXds (0.43s)
    --- FAIL: TestTranslateXds/securitypolicy-with-oidc-jwt-authz (0.00s)
        translator_test.go:142: securitypolicy-with-oidc-jwt-authz
        translator_test.go:143: 
                Error Trace:    /home/ubuntu/gateway/internal/xds/translator/translator_test.go:143
                Error:          Received unexpected error:
                                xds cluster exists
                Test:           TestTranslateXds/securitypolicy-with-oidc-jwt-authz
FAIL
FAIL    github.com/envoyproxy/gateway/internal/xds/translator   0.673s
FAIL
```

After:
```
ok      github.com/envoyproxy/gateway/internal/xds/translator   0.251s
```


arkodg:
hey @zhaohuabing is the issue that we are generating the same name for the jwks and oidc clusters if both are set in the policy ? shouldnt we be using an additional string value to differentiate them ?

zhaohuabing:
> hey @zhaohuabing is the issue that we are generating the same name for the jwks and oidc clusters if both are set in the policy ? shouldnt we be using an additional string value to differentiate them ?

For the cluster generated by a url, EG uses the host and port for the cluster name to avoid creating duplicated clusters for the same host+port combination.

```
jwt: 
 providers:
   - remoteJWKS:
       uri: https://oidc.example.com/auth/realms/example/protocol/openid-connect/cert

oidc:
 provider:
   tokenEndpoint: https://oidc.example.com/oauth/token
```

The name of the generated cluster: oidc_example_com_443.

We could change this logic to generate an unique name for each single oidc or jwt configuration. However, we should also ensure that the translator shouldn't throw error if the cluster already exists.

arkodg:
> > hey @zhaohuabing is the issue that we are generating the same name for the jwks and oidc clusters if both are set in the policy ? shouldnt we be using an additional string value to differentiate them ?
> 
> For the cluster generated by a url, EG uses the host and port for the cluster name to avoid creating duplicated clusters for the same host+port combination.
> 
> ```
> jwt: 
>  providers:
>    - remoteJWKS:
>        uri: https://oidc.example.com/auth/realms/example/protocol/openid-connect/cert
> 
> oidc:
>  provider:
>    tokenEndpoint: https://oidc.example.com/oauth/token
> ```
> 
> The name of the generated cluster: oidc_example_com_443.
> 
> We could change this logic to generate an unique name for each single oidc or jwt configuration. However, we should also ensure that the translator shouldn't throw error if the cluster already exists.

is it safe to reuse the same cluster configuration ? is the naming different when use the backendRefs field ?

internal/xds/translator/translator.go
```diff
@@ -788,10 +782,12 @@ func addXdsSecret(tCtx *types.ResourceVersionTable, secret *tlsv3.Secret) error
 	return nil
 }
 
+// addXdsCluster adds a xds cluster with args.
+// If the cluster already exists, it skips adding the cluster and returns nil
 func addXdsCluster(tCtx *types.ResourceVersionTable, args *xdsClusterArgs) error {
 	// Return early if cluster with the same name exists
 	if c := findXdsCluster(tCtx, args.name); c != nil {
-		return ErrXdsClusterExists
+		return nil
```
arkodg, line 790: this is not great from an API perspective for `addXdsCluster`, because the the old cluster with the same name may not have the same `xdsClusterArgs` as the current request, so the caller should decide how to handle this case, not this method imo

zhaohuabing:
> For OIDC provider with backendRefs, EG generate an unique cluster name like `securitypolicy/envoy-gateway/policy-for-gateway/0`.  

Ha, there is also a bug here, the index is always 0, which could generate duplicated name for different clusters if there're both ext auth and oidc whithin a SecurityPolicy.  Will fix it in this PR as well.

arkodg:
> > For OIDC provider with backendRefs, EG generate an unique cluster name like `securitypolicy/envoy-gateway/policy-for-gateway/0`.
> 
> Ha, there is also a bug here, the index is always 0, which generates duplicated name for different clusters if there're both ext auth and oidc whithin a SecurityPolicy. Will fix it in this PR as well.

nice catch, prob needs another prefix like `jwt`, `oidc` after policy name

internal/xds/translator/translator.go
zhaohuabing: Not a strong opinion, but should the caller call `findXdsCluster` first and decide whether they should handle this situation? 

The current approach has an implict assumption: every callers should check whether the return error isErrXdsClusterExists or not, and ignore ErrXdsClusterExists  - this cause duplications for every caller and can cause issues if the caller forget to handle this special case, like this one.

arkodg: You're absolutely right, the method is doing two many things, fine to split it up, but developers will need to make sure they call both functions when writing new logic

zhaohuabing: For all the current callers, they just need to call `addXdsCluster` as they just simply ignore `ErrXdsClusterExists` and does nothing.

I guess this pattern comes from the Kub Client API - it totally makes sense for Kubernetes API server as the API server has concurrent writes. EG doesn't need to use the same pattern here as it's a local cache and has no concurrent writes.

internal/gatewayapi/securitypolicy.go
```diff
@@ -374,21 +378,26 @@ func (t *Translator) translateSecurityPolicyForRoute(
...
+			extAutClusterIndex++
 		}
 
 		var oidc *ir.OIDC
 		if policy.Spec.OIDC != nil {
 			if oidc, err = t.buildOIDC(
 				policy,
 				resources,
-				gtwCtx.envoyProxy); err != nil {
+				gtwCtx.envoyProxy,  // TODO zhaohuabing: Only the last cluster is used as the OIDC name doesn't include the cluster index
```
zhaohuabing, line 394: This is a minor issue, will address it in a follow-up PR.

internal/gatewayapi/securitypolicy.go
```diff
@@ -374,21 +378,26 @@ func (t *Translator) translateSecurityPolicyForRoute(
...
+			extAutClusterIndex++
 		}
 
 		var oidc *ir.OIDC
 		if policy.Spec.OIDC != nil {
 			if oidc, err = t.buildOIDC(
 				policy,
 				resources,
-				gtwCtx.envoyProxy); err != nil {
+				gtwCtx.envoyProxy,  // TODO zhaohuabing: Only the last EnvoyProxy will be used as the OIDC name doesn't include the cluster index
```
zhaohuabing, line 394: This is a minor issue, will address it in a follow-up PR.

internal/gatewayapi/securitypolicy.go
```diff
@@ -361,6 +361,10 @@ func (t *Translator) translateSecurityPolicyForRoute(
 	// Apply IR to all relevant routes
 	prefix := irRoutePrefix(route)
 	parentRefs := GetParentReferences(route)
+	var (
+		extAutClusterIndex = 0
```
arkodg, line 365: why is the index needed ?

zhaohuabing: A cluster is created for each parent, so an index is added to the cluster name to avoid name conflict.

var (
		extAutClusterIndex = 0
		oidcClusterIndex   = 0
	)
	for _, p := range parentRefs {
		parentRefCtx := GetRouteParentContext(route, p)
		gtwCtx := parentRefCtx.GetGateway()
		if gtwCtx == nil {
			continue
		}

		var extAuth *ir.ExtAuth
		if policy.Spec.ExtAuth != nil {
			if extAuth, err = t.buildExtAuth(
				policy,
				resources,
				gtwCtx.envoyProxy,
				extAutClusterIndex,
			); err != nil {
				err = perr.WithMessage(err, "ExtAuth")
				errs = errors.Join(errs, err)
			}
			extAutClusterIndex++
		}

		var oidc *ir.OIDC
		if policy.Spec.OIDC != nil {
			if oidc, err = t.buildOIDC(
				policy,
				resources,
				gtwCtx.envoyProxy, // TODO zhaohuabing: Only the last EnvoyProxy will be used as the OIDC name doesn't include the cluster index
				oidcClusterIndex,
			); err != nil {
				err = perr.WithMessage(err, "OIDC")
				errs = errors.Join(errs, err)
			}
			oidcClusterIndex++
		}
               ...
	}
	return errs
}

arkodg: It's the same policy though, can't the clusters be reused ?

zhaohuabing: It's because the `EnvoyProxy` configuration is different for each parent.

There's still an issue here, only the last one will be applied as the OIDC name is identical, I've added a TODO and will address this later in a follow-up PR.

// TODO zhaohuabing: Only the last EnvoyProxy will be used as the OIDC name doesn't include the cluster index

arkodg: if its due to envoy proxy resource attributes that make the cluster unique, then lets append `/<ns>-<name>/` of EP instead to maximize reuse ?

arkodg: or listener name, the index here is not used to debug and is not useful for reuse

zhaohuabing: @arkodg can we address this in https://github.com/envoyproxy/gateway/issues/4720? This is a minor issue, but the scope of the change is relatively large as the name of the OIDC will aslo to be changed to make it unique for each EnvoyProxy/Cluster Name.

guydc (dismissed):

arkodg (dismissed):
LGTM thanks !

arkodg (approved):

zirain (approved):

PR #4722, opened 2024-11-15
fix(ratelimiter): prevent env vars being overwritten if set

ryanhristovski:
**What type of PR is this?**
fix(ratelimiter): prevent REDIS_URL from being overwritten if already set

**What this PR does / why we need it:**
This PR updates the logic to ensure that the REDIS_URL environment variable is only set if it isn’t already defined in the environment. This is especially important for when this [issue](https://github.com/envoyproxy/gateway/issues/1217) is solved and proper Redis authentication is setup.

Example: I want to be able to set the REDIS_URL to a secret, which isn't possible in a configmap.
```
gateway-helm:
  config:
    envoyGateway:
      gateway:
      provider:
        type: Kubernetes
        kubernetes:
          rateLimitDeployment:
            replicas: 1
            container:
              image: envoyproxy/ratelimit:master
            patch:
              type: StrategicMerge
              value:
                spec:
                  template:
                    spec:
                      containers:
                      - name: envoy-ratelimit
                        env:
                          - name: LOG_LEVEL
                            value: debug
                          - name: REDIS_URL
                            valueFrom:
                              secretKeyRef:
                                name: ratelimit-secret
                                key: redis_url
      rateLimit:
        backend:
          type: Redis
```

As of currently, not setting the `url` while having `rateLimit` set is putting Gateway in a crashloop.

**Which issue(s) this PR fixes:**
Fixes #

**Release Notes:** No


internal/infrastructure/kubernetes/ratelimit/resource.go
```diff
@@ -352,16 +353,18 @@ func expectedRateLimitContainerEnv(rateLimit *egv1a1.RateLimit, rateLimitDeploym
...
-				Name:  RedisSocketTypeEnvVar,
-				Value: "tcp",
-			},
-			{
+		env = append(env, corev1.EnvVar{
+			Name:  RedisSocketTypeEnvVar,
+			Value: "tcp",
+		})
+
+		// Only set REDIS_URL if it's not already set
```
arkodg, line 361: instead of this logic (which is fragile because it will need to be done for many env vars) , can we instead apply the user defined env after setting these ?

ryanhristovski: yeah that makes a lot of sense, will have an update soon

ryanhristovski: @arkodg after a second look, the env var is being written fine - my bad. The actual issue is in the envoy gateway validation code [here](https://github.com/envoyproxy/gateway/blob/7699578e222a0c7ce7cafbf8cad96d634e464dfb/api/v1alpha1/validation/envoygateway_validate.go#L167-L172)

As of now I see 2 solutions, both not aren't great and I'm open to any suggestions:
1. Remove the validation for `rateLimit.Backend.Redis` and  `rateLimit.Backend.Redis.URL == ""`. 
	-  Ofc this isn't ideal and could cause hard to track rate limit errors which would arise if no `rateLimit.Backend.Redis.URL` or `REDIS_URL` is set.
2. Extract the EnvoyGatewayKubernetesProvider patch value for RateLimitDeployment, unmarshaling its configuration, and traversing the resulting structure to determine if the REDIS_URL environment variable is defined.
	- This code would be hard to maintain due to deep nesting, brittle type assertions, and reliance on rigid structure traversal.

I wasn't able to get the REDIS_URL any other way that I tried 😓

arkodg: hey yah for now suggest adding a dummy value in `rateLimit.Backend.Redis.URL` to solve your use case
good to hear you're looking to contribute, a good place to start would be to describe your use case in a new GH issue or in https://github.com/envoyproxy/gateway/issues/1217 and hoping you can drive it to completion, we'll def be around to help out !

Issue #4744, opened 2024-11-20
dnslookupfamily returns ipv6 addresses for external clusters (oidc)


zetaab:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

I compiled new version from latest master and our OIDC is now broken.

```
[2024-11-20 07:57:48.870][1][debug][dns] [source/extensions/network/dns_resolver/cares/dns_impl.cc:391] dns resolution for cognito-idp.eu-central-1.amazonaws.com started
[2024-11-20 07:57:48.876][1][debug][dns] [source/extensions/network/dns_resolver/cares/dns_impl.cc:308] dns resolution for cognito-idp.eu-central-1.amazonaws.com completed with status 0
[2024-11-20 07:57:48.876][1][debug][upstream] [source/common/upstream/upstream_impl.cc:484] transport socket match, socket default selected for host with address [2a05:d014:32e:701:9334:4719:42de:9263]:443
[2024-11-20 07:57:48.876][1][debug][upstream] [source/common/upstream/upstream_impl.cc:484] transport socket match, socket default selected for host with address [2a05:d014:32e:700:f4dc:9de:938f:1329]:443
[2024-11-20 07:57:48.876][1][debug][upstream] [source/common/upstream/upstream_impl.cc:484] transport socket match, socket default selected for host with address [2a05:d014:32e:702:b316:2916:8253:ddff]:443
[2024-11-20 07:57:48.876][1][debug][upstream] [source/extensions/clusters/strict_dns/strict_dns_cluster.cc:201] DNS refresh rate reset for cognito-idp.eu-central-1.amazonaws.com, refresh rate 30000 ms
```

Like can be seen our oidc now tries to use ipv6. However, we do not have ipv6 connectivity in our cluster at all

example interfaces

```
/home/curl_user $ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
[...]

PR #4745, opened 2024-11-20
xds: use V4_PREFERRED dnsLookupFamily by default

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/4744

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -772,6 +772,7 @@ xds:
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
+          dnsLookupFamily: V4_PREFERRED
           edsClusterConfig:
             edsConfig:
               ads: {}
@@ -791,6 +792,7 @@ xds:
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
+          dnsLookupFamily: V4_PREFERRED
           edsClusterConfig:
             edsConfig:
               ads: {}
@@ -817,6 +819,7 @@ xds:
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
+          dnsLookupFamily: V4_PREFERRED
           edsClusterConfig:
             edsConfig:
               ads: {}
@@ -836,6 +839,7 @@ xds:
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
+          dnsLookupFamily: V4_PREFERRED
           edsClusterConfig:
             edsConfig:
               ads: {}
@@ -855,6 +859,7 @@ xds:
           commonLbConfig:
             localityWeightedLbConfig: {}
           connectTimeout: 10s
+          dnsLookupFamily: V4_PREFERRED
           edsClusterConfig:
             edsConfig:
               ads: {}
--- a/internal/cmd/egctl/testdata/translate/out/echo-gateway-api.cluster.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/echo-gateway-api.cluster.yaml
@@ -100,6 +100,7 @@ xds:
         commonLbConfig:
           localityWeightedLbConfig: {}
         connectTimeout: 10s
+        dnsLookupFamily: V4_PREFERRED
         edsClusterConfig:
           edsConfig:
             ads: {}
--- a/internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
+++ b/internal/cmd/egctl/testdata/translate/out/from-gateway-api-to-xds.all.json
@@ -458,6 +458,7 @@
                   "localityWeightedLbConfig": {}
                 },
                 "connectTimeout": "10s",
+                "dnsLookupFamily": "V4_PREFERRED",
                 "edsClusterConfig": {
                   "edsConfig": {
                     "ads": {},
@@ -487,6 +488,7 @@
                   "localityWeightedLbConfig": {}
                 },
                 "connectTimeout": "10s",
+                "dnsLookupFamily": "V4_PREFERRED",
                 "edsClusterConfig": {
                   "edsConfig": {
                     "ads": {},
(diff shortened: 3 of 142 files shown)


zirain:
cc @zetaab can you verify this on your env? hope it will work.

arkodg (approved):

zhaohuabing (changes requested):
@zirain @arkodg I think V4_PREFERRED won't work for IPv6 env where the envoy pod only has an IPv6 address.

> If V4_PREFERRED is specified, the DNS resolver will first perform a lookup for addresses in the IPv4 family and fallback to a lookup for addresses in the IPv6 family.

zirain:
> @zirain @arkodg I think V4_PREFERRED won't work for IPv6 env where the envoy pod only has an IPv6 address.
> 
> > If V4_PREFERRED is specified, the DNS resolver will first perform a lookup for addresses in the IPv4 family and fallback to a lookup for addresses in the IPv6 family.

tested passed on an IPv6 only cluster.

zhaohuabing:
> 3b26516

I mean "real" IPv6 only where the pod just has an IPv6 address and has no IPv4 address.

Does the pod in this test has an IPv4 address? If it only has an IPv6 address, it shouldn't be able to establish an connection to an IPv4 server.

zirain:
> > [3b26516](https://github.com/envoyproxy/gateway/commit/3b265169ee3579c77a3dc9ab196e64fcefebc76e)
> 
> I mean "real" IPv6 only where the pod just has an IPv6 address and has no IPv4 address.
> 
> Does the pod in this test has an IPv4 address? If it only has an IPv6 address, it shouldn't be able to establish an connection to an IPv4 server.

it's IPv6 only cluster, not IPv6 first.

zhaohuabing:
SP with a JWK configuration:

```yaml
kind: SecurityPolicy
metadata:
  name: demo-api-jwt
spec:
  jwt:
    providers:
    - name: test
      remoteJWKS:
        uri: https://www.zhaohuabing.com/misc/jwks.json
  targetRef:
    kind: HTTPRoute
    name: backend
```

DNS records for IPv4 and IPv6

```
➜  gateway git:(use/V4_PREFERRED) dig www.zhaohuabing.com     

;; ANSWER SECTION:
www.zhaohuabing.com.	247	IN	A	104.21.60.99
www.zhaohuabing.com.	247	IN	A	172.67.195.133

➜  gateway git:(use/V4_PREFERRED) dig AAAA www.zhaohuabing.com

;; ANSWER SECTION:
www.zhaohuabing.com.	244	IN	AAAA	2606:4700:3037::ac43:c385
www.zhaohuabing.com.	244	IN	AAAA	2606:4700:3034::6815:3c63
```

Test with IPv6 only, failed to fetch jwks

```
export IP_FAMILY=ipv6; export IMAGE_PULL_POLICY=IfNotPresent; make create-cluster kube-install-image kube-deploy
```

Envoy log
```
[2024-11-21 06:37:38.446][1][warning][jwt] [source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc:115] Jwks async fetching url=https://www.zhaohuabing.com/misc/jwks.json: failed
```

Test with IPv4 only, succeeded

```
export IP_FAMILY=ipv4; export IMAGE_PULL_POLICY=IfNotPresent; make create-cluster kube-install-image kube-deploy
```

```
[2024-11-21 06:32:39.285][1][debug][filter] [source/extensions/filters/http/common/jwks_fetcher.cc:92] onSuccess: fetch pubkey [uri = https://www.zhaohuabing.com/misc/jwks.json]: succeeded
```

zirain:
can you passed the test with `IPv4_only` on a IPv6 only cluster?

zhaohuabing (approved):
LGTM Thanks.

It would be useful to include a few lines explaining how IPFamily affects the Envoy DNS resolver.

zhaohuabing (approved):

zhaohuabing (dismissed):

zhaohuabing (approved):

api/v1alpha1/envoyproxy_types.go
```diff
@@ -136,14 +136,15 @@ type EnvoyProxySpec struct {
...
 	BackendTLS *BackendTLSConfig `json:"backendTLS,omitempty"`
-
 	// IPFamily specifies the IP family for the EnvoyProxy fleet.
-	// This setting only affects the Gateway listener port and does not impact
-	// other aspects of the Envoy proxy configuration.
-	// If not specified, the system will operate as follows:
-	// - It defaults to IPv4 only.
-	// - IPv6 and dual-stack environments are not supported in this default configuration.
-	// Note: To enable IPv6 or dual-stack functionality, explicit configuration is required.
+	// This setting affects the DNS resolver for the EnvoyProxy fleet.
```
arkodg, line 140: this is incorrect, this field is exclusive for the listener / envoy proxy and the cluster IPFamily is derived from the Service / EndpointSlice of the backenRef
cc @juwon8891

zirain: But right now, from logic the main affect is dnsLookupFamily, or I may missed something.

arkodg: yeah its using the listener IPFamily instead of the addressType field or ipFamily field of the destination, which looks like a bug

zirain: default < envoyproxy < destination?

arkodg: envoyProxy resource IPFamily is only meant for listener and Envoy Proxy fleet Service spec

BackendRef IPFamily determines cluster setting

zirain: as we use `::` and `IPv4Compact=true` this knob seems meaningless?

zirain: I reverted the API doc change, seems this PR didn't related to it.

arkodg: Yeah now it will only be used in the infra manager to create the Service spec

zhaohuabing (approved):

zhaohuabing:
> envoyProxy resource IPFamily is only meant for listener and Envoy Proxy fleet Service spec

> BackendRef IPFamily determines cluster setting

Are we planning to introduce another "BackendRef IPFamily" for the dns lookup IP family? Even though they're different, typically the listener IP famaily and the Backend IP Family should be the same.

arkodg (approved):

PR #4754, opened 2024-11-21
Fix: frequent 503 errors when connecting to a Service experiencing high Pod churn

zhaohuabing:
Reverts envoyproxy/gateway#4337
Fixes: https://github.com/envoyproxy/gateway/issues/4685

Release Note: Yes

To avoid regression of the bug fixed by #4337, a temporary buffer has been added to store events received before the Updater is enabled.
These events will be sent to the update channel once the Updater is enabled.

Tested the fix in my local env:

```bash
➜  k -n envoy-gateway-system logs envoy-gateway-7f7d68bd-q74m6 |grep "status update"                        
2024-11-26T05:56:03.461Z        INFO    provider        kubernetes/status_updater.go:185        received a status update while disabled, storing for later      {"runner": "provider", "event": {
"name":"ha-gateway","namespace":"gateway-upgrade-infra"}}
2024-11-26T05:56:03.764Z        INFO    provider        kubernetes/status_updater.go:185        received a status update while disabled, storing for later      {"runner": "provider", "event": {
"name":"upgrade"}}                                                                              
2024-11-26T05:56:03.767Z        INFO    provider        kubernetes/status_updater.go:185        received a status update while disabled, storing for later      {"runner": "provider", "event": {
"name":"http-backend-eg-upgrade","namespace":"gateway-upgrade-infra"}}
2024-11-26T05:56:03.767Z        INFO    provider        kubernetes/status_updater.go:185        received a status update while disabled, storing for later      {"runner": "provider", "event": {
"name":"ha-gateway","namespace":"gateway-upgrade-infra"}}
2024-11-26T05:56:28.146Z        INFO    provider        kubernetes/status_updater.go:139        started status update handler   {"runner": "provider"}
2024-11-26T05:56:28.146Z        INFO    provider        kubernetes/status_updater.go:197        sending stored status update    {"runner": "provider", "event": {"name":"ha-gateway","namespace":
"gateway-upgrade-infra"}}       
2024-11-26T05:56:28.146Z        INFO    provider        kubernetes/status_updater.go:197        sending stored status update    {"runner": "provider", "event": {"name":"upgrade"}}
2024-11-26T05:56:28.146Z        INFO    provider        kubernetes/status_updater.go:197        sending stored status update    {"runner": "provider", "event": {"name":"http-backend-eg-upgrade"
,"namespace":"gateway-upgrade-infra"}}  
[...]

Issue #4685, opened 2024-11-08
503s and `_No_route_to_host` errors due to routing to non-existent Endpoints


coro:
*Description*:
We have been seeing many 503 errors when connecting to a Service with a lot of Pod churn.

We also saw in the logs that in these cases, the `upstream_host` that Envoy was attempting to connect to was for Pods that no longer existed in the cluster. These Pods could have been terminated over 50 mins earlier.

*Repro steps*:
Simple setup of Gateway (AWS NLB) -> HTTPRoute -> Service pointing to a Deployment with a lot of Pod churn.

*Environment*:
Gateway: v1.2.1 (also seen on v1.1.3) (**not** seen on v1.1.1)
Envoy: v1.32.1 (also seen on v1.31.1) (**not** seen on v1.31.0)
EKS cluster v1.29

*Logs*:
```json
{
  "start_time": "2024-11-08T13:54:35.382Z",
  "method": "POST",
  "x-envoy-origin-path": "/v1/models/custom-model:predict",
  "protocol": "HTTP/1.1",
  "response_code": "503",
  "response_flags": "UF",
  "response_code_details": "upstream_reset_before_response_started{remote_connection_failure|delayed_connect_error:_No_route_to_host}",
  "connection_termination_details": "-",
  "upstream_transport_failure_reason": "delayed_connect_error:_No_route_to_host",
  "bytes_received": "494",
  "bytes_sent": "165",
  "duration": "3060",
  "x-envoy-upstream-service-time": "-",
  "x-forwarded-for": "10.0.99.64",
  "user-agent": "python-requests/2.32.3",
  "x-request-id": "3dbc6bcc-79b1-4ec1-855c-1569b69f5416",
  ":authority": "kserve-detection.prod.signal",
  "upstream_host": "10.0.101.243:8080",
  "upstream_cluster": "httproute/default/detection/rule/0",
  "upstream_local_address": "-",
  "downstream_local_address": "10.0.101.92:10080",
  "downstream_remote_address": "10.0.99.64:41562",
  "requested_server_name": "-",
  "route_name": "httproute/default/detection/rule/0/match/0/kserve-detection_prod_signal"
}
```

The generated cluster config for this route was:

```yaml
   "dynamic_active_clusters": [
    {
     "version_info": "a998a3b0100c28a29ff3ee662cb01dc45463f0260c9a6a50107dcc12a45938fc",
     "cluster": {
[...]

PR #4767, opened 2024-11-22
fix: decouple gateway status updates from the reconciler

zhaohuabing:
Fixes: https://github.com/envoyproxy/gateway/issues/4845
Release Note: Yes


internal/provider/kubernetes/status.go
```diff
@@ -564,34 +592,3 @@ func (r *gatewayAPIReconciler) updateStatusForGateway(ctx context.Context, gtw *
 		}),
 	})
 }
-
```
arkodg, line 567: what about https://github.com/envoyproxy/gateway/blob/71c0b5408f7f80ef8b9d1e0bcf511891839edecd/internal/provider/kubernetes/status.go#L524

arkodg:
were we deadlocking ?

internal/provider/kubernetes/predicates.go
```diff
@@ -294,7 +294,7 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 	// Check if the Service belongs to a Gateway, if so, update the Gateway status.
 	gtw := r.findOwningGateway(ctx, labels)
 	if gtw != nil {
-		r.updateStatusForGateway(ctx, gtw)
+		r.resources.GatewayStatuses.Store(utils.NamespacedName(gtw), &gtw.Status)
```
arkodg, line 297: this is not documented well, but the reason this existed, was so that this flow could update the `Status.Address` field and `Status.Conditions` while the other flow (published from Gateway API) would update `Status.Conditions`

wanted to highlight this

zhaohuabing: Yep, better to add a line in the comment to explain why need to send an update here and how the gateway address is updated.

zhaohuabing:
> were we deadlocking ?

It's my suspicion - I can't reproduce myself, but based on the description of the issue https://github.com/envoyproxy/gateway/issues/4685, it seems that the reconciler has been blocked because the pods changes haven't been propogated to the Envoy through xDS.

internal/provider/kubernetes/status.go
zhaohuabing: I suspect the call to `updateStatusForGateway` from inside the predicate blocked the reconciler.

arkodg: Aren't we deleting update status for gateway ?

zhaohuabing: No, the update is stored into the r.resources.GatewayStatuses instead.  The update event will be handled in the status updater which subscripts the GatewayStatuses.

internal/provider/kubernetes/predicates.go
```diff
@@ -294,7 +294,7 @@ func (r *gatewayAPIReconciler) validateServiceForReconcile(obj client.Object) bo
 	// Check if the Service belongs to a Gateway, if so, update the Gateway status.
 	gtw := r.findOwningGateway(ctx, labels)
 	if gtw != nil {
-		r.updateStatusForGateway(ctx, gtw)
+		r.updateGatewayStatus(gtw)
```
zhaohuabing, line 297: This is a bit difficult for the later comer to understand and mantain the Gateway status update logic here, instead of firing up an empty GatewayStatus update event, the predicate should just return true and handle the GatewayStatus properly in the Reconcile method. This can be addressed in a follow-up PR.

internal/provider/kubernetes/predicates.go
```diff
@@ -636,12 +636,26 @@ func (r *gatewayAPIReconciler) updateStatusForGatewaysUnderGatewayClass(ctx cont
...
+func (r *gatewayAPIReconciler) updateGatewayStatus(gateway *gwapiv1.Gateway) {
+	// The status added to GatewayStatuses is solely used to trigger the status updater
+	// and does not reflect the real changed status.
+	//
+	// The status updater will check the Envoy Proxy service to get the addresses of the Gateway,
+	// and check the Envoy Proxy Deployment/DaemonSet to get the status of the Gateway workload.
+	//
+	// Since the status does not reflect the actual changed status, we need to delete it first
+	// to prevent it from being considered unchanged. This ensures that subscribers receive the update event.
+	r.resources.GatewayStatuses.Delete(utils.NamespacedName(gateway))
```
arkodg, line 655: its unclear why the `Delete` is needed here
If the `Status` value is same as the previous value, it should be skipped, else updated

internal/gatewayapi/runner/runner.go
```diff
@@ -428,10 +428,11 @@ func (r *Runner) getAllStatuses() *StatusesToDelete {
 }
 
 func (r *Runner) deleteStatusKeys(ds *StatusesToDelete) {
-	for key := range ds.GatewayStatusKeys {
-		r.ProviderResources.GatewayStatuses.Delete(key)
-		delete(ds.GatewayStatusKeys, key)
-	}
+	// Do not delete the status keys for the Gateway because the Gateway status has also been stored into the ProviderResources
```
arkodg, line 431: an assumption we are making here is that the watchable will first kick and update status and then a service spec will be updated which will result in updating the address. These steps although are ordered in theory, the ordering cannot be guaranteed in the reconciler, which may cause issues related to stale values being set

arkodg: this logic is going to cause issues

I suggest reverting some of this and instead setting the value only if set i.e. if `Status.Address` is non zero value, then set it, or if `Status.Conditions` is non zero value, then set it. This ensures that the gateway api path doesnt update `Addresses`

internal/provider/kubernetes/predicates.go
zhaohuabing: The status added to GatewayStatuses here is solely used to trigger the status updaterand does not reflect the real changed status.

The status updater `updateStatusForGateway` will check the Envoy Proxy service to get the addresses of the Gateway, and check the Envoy Proxy Deployment/DaemonSet to get the status of the Gateway workload.

I plan to refactory this logic in a follow-up PR.

internal/gatewayapi/runner/runner.go
zhaohuabing: the gateway api path won't update `Address` or programming status, it's handled in the `updateStatusForGateway` of the status updater. https://github.com/envoyproxy/gateway/blob/dd1c66fb024e1ce76f59fe1480900307e9ae480d/internal/provider/kubernetes/status.go#L524-L548

internal/provider/kubernetes/predicates.go
zhaohuabing, line 655: There is a race condition between here and the gateway API runner. We'd better to put all the Gateway status calculation (the accept, programming, and address) in one go routine(Gateway API translation) to avoid race condition.

arkodg (approved):
LGTM thanks !

zhaohuabing:
@arkodg I'm going to close this PR as it has race condition problem.

Fixed this issue in #4754

arkodg (dismissed):
LGTM thanks !

internal/provider/kubernetes/predicates.go
zhaohuabing: Solved this by not overriding the status caclulated by the Gateway API translator.

internal/provider/kubernetes/controller.go
```diff
@@ -129,13 +129,25 @@ func newGatewayAPIController(mgr manager.Manager, cfg *config.Server, su Updater
...
 	}
+
+	// When leader election is enabled, only subscribe to status updates upon acquiring leadership.
+	if cfg.EnvoyGateway.Provider.Type == egv1a1.ProviderTypeKubernetes &&
+		!ptr.Deref(cfg.EnvoyGateway.Provider.Kubernetes.LeaderElection.Disable, false) {
+		go func() {
+			select {
+			case <-ctx.Done():
+				return
+			case <-mgr.Elected():
```
arkodg, line 144: afaik this won't work, you can only receive a value from a channel once
you can use `mgr.Elected().Done(func() {
  r.subscribeAndUpdateStatus(ctx, cfg.EnvoyGateway.EnvoyGatewaySpec.ExtensionManager != nil)
})`

or we'll need to move this func to https://github.com/envoyproxy/gateway/blob/23856723c58df0ad28e128481f558bce41078700/internal/provider/kubernetes/kubernetes.go#L110 where the handle exists

arkodg: my preference would be to replace the existing svr.Elected with a waitGroup which is a more easier way to signal once based on `mgr.Elected()` and have every one else (infra mgr, watchable status) wait until then
cc @alexwo

internal/envoygateway/config/config.go
```diff
@@ -37,19 +38,22 @@ type Server struct {
 	// Logger is the logr implementation used by Envoy Gateway.
 	Logger logging.Logger
 	// Elected chan is used to signal what a leader is elected
-	Elected chan struct{}
+	Elected *sync.WaitGroup
```
zirain, line 41: why we need a pointer here?

zhaohuabing:
Verified with https://github.com/envoyproxy/gateway/pull/4904.

internal/envoygateway/config/config.go
zhaohuabing: internal/infrastructure/runner/runner.go:35:25: copylocks: literal copies lock value from *cfg: github.com/envoyproxy/gateway/internal/infrastructure/runner.Config contains github.com/envoyproxy/gateway/internal/envoygateway/config.Server contains sync.WaitGroup contains sync.noCopy (govet)
        return &Runner{Config: *cfg}

arkodg (approved):
LGTM thanks !

guydc (approved):

guydc (approved):

PR #4782, opened 2024-11-26
xds: fix ipFamily always nil

### LINKED ISSUE #4761: xdsClusterArgs.ipFamily always nil
https://github.com/envoyproxy/gateway/blob/a924cec6cc3a1d8cb78be06b61b6d238472c199b/internal/xds/translator/cluster.go#L60

juwon8891:
chore: fix ipFamily always nil

This PR improves IP family configuration handling across different listener types and updates the DNS lookup strategy for better dual-stack support.
- Add consistent IP family handling across all listener types

Fixes #4761


internal/xds/translator/translator.go
```diff
@@ -463,6 +463,7 @@ func (t *Translator) addRouteToRouteConfig(
 			ea := &ExtraArgs{
 				metrics:       metrics,
 				http1Settings: httpListener.HTTP1,
+				ipFamily:      httpListener.IPFamily,
```
arkodg, line 466: this is incorrect, you are using the listener level `IPFamily` value in the cluster
you should be deriving this value using a new `IPFamily` field in 
https://github.com/envoyproxy/gateway/blame/48a0310082ceeaf4d3af17e6b6844c7d0fb5db0b/internal/ir/xds.go#L1297 instead and this should be set in the gateway api layer using the Backend Service / EndpointSlice info

juwon8891: Thanks for the feedback! I’ve updated the code to use a new IPFamily field

internal/gatewayapi/route.go
```diff
@@ -1261,9 +1261,11 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
 			Protocol:    protocol,
 			Endpoints:   endpoints,
 			AddressType: addrType,
+			IPFamily:    getIPFamily(envoyProxy),
```
arkodg, line 1264: `getIPFamily` gets the value from the `envoyProxy` which is the listener's IP Family, not the backend's. This value should be coming from https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-service-configuration-scenarios

juwon8891: Thanks for the feedback! I’ve updated the code to use a new IPFamily field
cc @zirain

internal/gatewayapi/helpers.go
```diff
@@ -610,20 +610,56 @@ func setIfNil[T any](target **T, value *T) {
...
+		case corev1.IPv6Protocol:
+			return ptr.To(ir.IPv6)
+		}
+	}
+
+	return nil
+}
+
+// getEnvoyIPFamily returns the IPFamily configuration from EnvoyProxy
+func getEnvoyIPFamily(envoyProxy *egv1a1.EnvoyProxy) *ir.IPFamily {
```
zirain, line 650: QQ: why we need another enum? cannot we reuse egv1a1.IPFamily?

internal/xds/translator/translator.go
```diff
@@ -465,6 +465,10 @@ func (t *Translator) addRouteToRouteConfig(
 				http1Settings: httpListener.HTTP1,
 			}
 
+			if len(httpRoute.Destination.Settings) > 0 {
```
zirain, line 468: is it possible to have different IPFamily in destination settings?

internal/gatewayapi/route.go
```diff
@@ -1313,6 +1314,7 @@ func (t *Translator) processDestination(backendRefContext BackendRefContext,
 	}
 
 	ds.Weight = &weight
+	ds.IPFamily = getServiceIPFamily(resources.GetService(backendNamespace, string(backendRef.Name)))
```
zirain, line 1317: better to move to L1282, only service support IPFamily now.

internal/gatewayapi/helpers.go
juwon8891: Thanks for the feedback! I’ve updated the code.

internal/gatewayapi/route.go
juwon8891: Thanks for the feedback! I’ve updated the code.

internal/xds/translator/translator.go
juwon8891: Yes, it's possible to have different IPFamily settings in destination configurations. I've also added test cases to confirm this behavior.

zirain: should can complete `IPFamily` from all `destinations`:
1. if all v4, use v4
2. if all v6, use v6
3. if has both, use Dual?

cc @arkodg , WDYT?

arkodg: sure SGTM @zirain

zirain: @juwon8891 can you rebase and fix this?

juwon8891: Sure I rebased.

internal/gatewayapi/helpers.go
```diff
@@ -610,20 +610,56 @@ func setIfNil[T any](target **T, value *T) {
...
+// getServiceIPFamily returns the IP family configuration from a Kubernetes Service
+// following the dual-stack service configuration scenarios:
+// https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-service-configuration-scenarios
+//
+// The IP family is determined in the following order:
+// 1. Service.Spec.IPFamilyPolicy == RequireDualStack -> DualStack
+// 2. Service.Spec.IPFamilies length > 1 -> DualStack
+// 3. Service.Spec.IPFamilies[0] -> IPv4 or IPv6
+// 4. nil if not specified
+func getServiceIPFamily(service *corev1.Service) *egv1a1.IPFamily {
```
zirain, line 622: it would be better if you can an ut for this.

internal/xds/translator/translator.go
```diff
@@ -464,6 +464,11 @@ func (t *Translator) addRouteToRouteConfig(
 			ea := &ExtraArgs{
 				metrics:       metrics,
 				http1Settings: httpListener.HTTP1,
+				ipFamily:      httpRoute.Destination.Settings[0].IPFamily,
+			}
+
+			if len(httpRoute.Destination.Settings) > 1 {
+				ea.ipFamily = determineIPFamily(httpRoute.Destination.Settings)
```
zirain, line 471: merge this into `determineIPFamily`

internal/gatewayapi/helpers.go
juwon8891: Thank you for your feedback! I've incorporated the suggested changes.

internal/xds/translator/translator.go
juwon8891: Thank you for your feedback! I've incorporated the suggested changes.

internal/gatewayapi/listener.go
```diff
@@ -118,17 +118,14 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
 						Address:  address,
 						Port:     uint32(containerPort),
 						Metadata: buildListenerMetadata(listener, gateway),
-						IPFamily: getIPFamily(gateway.envoyProxy),
+						IPFamily: getEnvoyIPFamily(gateway.envoyProxy),
```
zirain, line 121:
```suggestion
						IPFamily: ipFamily,
```

internal/gatewayapi/listener.go
```diff
@@ -144,7 +141,7 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
 						Name:     irListenerName(listener),
 						Address:  address,
 						Port:     uint32(containerPort),
-						IPFamily: getIPFamily(gateway.envoyProxy),
+						IPFamily: getEnvoyIPFamily(gateway.envoyProxy),
```
zirain, line 144:
```suggestion
						IPFamily: ipFamily,
```

zirain (commented):
small nit, otherwise LGTM.

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #4496, opened 2024-10-22
api: adds request/response attributes ExtProc options

mathetake:
**What type of PR is this?**

This PR adds obvious config options for the extproc: `requestAttributes` and `responseAttributes`.

**What this PR does / why we need it**:

While I am developing a custom WAF external processing, I want to be able to
access attributes information while processing the requests listed below:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#arch-overview-attributes

**Which issue(s) this PR fixes**:
ref: https://github.com/envoyproxy/gateway/issues/3170

PR #4794, opened 2024-11-27
api: ext-proc attributes

guydc:
**What type of PR is this?**

Following the discussion at KubeCon NA : https://docs.google.com/document/d/1PS5xLA0IDbj6McHIXXhShn51Zq37WvuistaidBPeHoE/edit?tab=t.0 

## Scope and Motivation
This API will allow users to define which attributes are sent to the external processor as context for requests/responses.

Attributes provide HTTP extensions with additional context (e.g. TCP, TLS and XDS attributes) that can be relevant inputs for the extension logic. 

## Comparison to other extension options
Many Envoy extensions are inherently capable of interaction with context [attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes): 
- lua: stream and connection attributes are available through [StreamInfo](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#streaminfo)
- wasm: get attributes through [get_property]()
  
For out-of-process extensions like ext-proc, Envoy must be explicitly configured to allow access to attributes and and define the scope of access. With the increase in ext-proc use cases, such as the [llm-instance-gateway](https://github.com/kubernetes-sigs/llm-instance-gateway), [envoy-ai-gateway](https://github.com/envoyproxy/ai-gateway) and externally-deployed WAFs, there is a greater need to provide extension with context.

## Security Considerations

Most [attributes are scoped to the current connection or stream](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes) by their prefix (`connection.*`, `request.*`, `response.*`), with the exception of `xds.*` attributes such as `xds.listener_metadata`, `xds.upstream_host_metadata`, `xds.node` and generic metadata and filter state access attributes. 

## Related Work
- Issue: https://github.com/envoyproxy/gateway/issues/3170
- Previous Attempts by various contributors: 
  - #4496 
  - #4160 
  - #3247

Release Notes: Yes


arkodg (approved):
LGTM thanks !

arkodg:
curious what happens when something like `connection.unknown` is used ? is it a no-op ?

guydc:
> curious what happens when something like `connection.unknown` is used ? is it a no-op ?

Yes, it's just ignored and not passed. Config is still accepted by the proxy.

zirain (approved):

Issue #4793, opened 2024-11-27
Hostname doesn't support subdomain with single character in Backend Object


nothinux:
*Description*:
Hi, we have an issue with backend object when trying to set the hostname with subdomain that has single character in it. For example if we set the hostname to this subdomain `backend.s.collector.example.com` it will trigger this error

`The Backend "backend" is invalid: spec.endpoints[0].fqdn.hostname: Invalid value: "backend.s.collector.example.com": spec.endpoints[0].fqdn.hostname in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9]))*$'`

However when we set it to `backend.staging.collector.example.com` there's no issue

Is the regex is expected to not allow subdomain with single character?
https://github.com/envoyproxy/gateway/blob/a383dd97e4c7fe726f064c03f48a06e8b049001f/api/v1alpha1/backend_types.go#L97-L98

Example manifest

```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: Backend
metadata:
  name: backend
  namespace: envoy-gateway-system
spec:
  endpoints:
  - fqdn:
      hostname: backend.s.collector.example.com
      port: 80
```

nothinux:
@zhaohuabing the fix actually on this PR https://github.com/envoyproxy/gateway/pull/4803, can you help to check?

guydc:
SGTM. We weren't compatible with the K8s validation by mistake: https://github.com/kubernetes/apimachinery/blob/96b97de8d6ba49bc192968551f2120ef3881f42d/pkg/util/validation/validation.go#L206.

PR #4803, opened 2024-11-28
fix: allow hostname to use subdomain with single label/character

nothinux:
Fixes https://github.com/envoyproxy/gateway/issues/4793

Release Notes: No


zhaohuabing:
@nothinux Thanks for fixing this. Could you please add a test to test/cel-validation/backend_test.go ?

nothinux:
added

nothinux:
@zhaohuabing the coverage test failed because only 4 items are allowed. can we double this limit? 

https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/backend_types.go#L119-L121

arkodg:
> > @zhaohuabing the coverage test failed because only 4 items are allowed. can we double this limit?
> > https://github.com/envoyproxy/gateway/blob/a383dd97e4c7fe726f064c03f48a06e8b049001f/api/v1alpha1/backend_types.go#L119-L121
> 
> I think it's fine to make it 8 cc @guydc

64, so it account for cases where users are auto generating Backend resources to route to custom backends ? Can we do that in a separate PR @nothinux ?

nothinux:
@arkodg  I've created separate PR to Increase it https://github.com/envoyproxy/gateway/pull/4822

zirain:
can you run `make -k gen-check` to make CI happy?

nothinux:
`gen-check` was fine, not sure why it's failing on `lint.shellcheck`

zirain:
> `gen-check` was fine, not sure why it's failing on `lint.shellcheck`

I believed crd files should be updated?

zhaohuabing:
@nothinux 

test/cel-validation/backend_test.go:90: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/envoyproxy/gateway) (gci)

nothinux:
> test/cel-validation/backend_test.go:90: File is not gci-ed with --skip-generated -s standard -s default -s prefix(github.com/envoyproxy/gateway) (gci)

Thanks for pointing that out, updated

zhaohuabing (approved):
LGTM thanks!

arkodg (approved):

PR #4475, opened 2024-10-18
reuse filewatcher

zirain:
1. reuse `filewatcher` package
2. move code under `util`

PR #4807, opened 2024-11-28
refactor: reuse the filewatcher for file-provider

shawnh2:
**What type of PR is this?**

replace #4475

Release Notes: No


zirain (approved):
LGTM, it would be good if we can have some tests for file provider.

shawnh2:
> LGTM, it would be good if we can have some tests for file provider.

Have some tests on my local env, will add them as a follow-up

arkodg (approved):

PR #4809, opened 2024-11-28
use a waitGroup instead of an enabled channel in the status updater

arkodg:
* use a waitGroup to synchronize to the `Send` method that the status updater is enabled and ready for updates

### DIFF
--- a/internal/provider/kubernetes/status_updater.go
+++ b/internal/provider/kubernetes/status_updater.go
@@ -7,7 +7,7 @@ package kubernetes
 
 import (
 	"context"
-	"errors"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


internal/provider/kubernetes/status_updater.go
```diff
@@ -168,40 +165,15 @@ type Updater interface {
...
-func (u *UpdateWriter) handleEventsReceivedBeforeEnabled() {
-	go func() {
-		for e := range u.eventsBeforeEnabled {
-			u.log.Info("sending stored status update", "event", e.NamespacedName)
-			u.updateChannel <- e
-		}
-		close(u.eventsBeforeEnabled)
-	}()
+	// Wait until updater is ready
+	u.wg.Wait()
```
zhaohuabing, line 175: Would the wait cause dead block like this issue?   https://github.com/envoyproxy/gateway/issues/4685

arkodg: This will only wait once during startup

zhaohuabing (approved):

internal/provider/kubernetes/status_updater.go
zhaohuabing: I suspect it may block because #4685 happened even when there is no wait.

zirain (approved):

PR #4817, opened 2024-11-29
listen on ipv4 addresses by default

arkodg:
* listen on `::` if the listener IPFamily is `IPv6` or `DualStack`
* Set `ipv4_compat` to `true` if `IPFamily` is `DualStack`


zirain (approved):

zirain:
will this make ipv6 test failed without additional IPFamily in EnvoyProxy?

arkodg:
> will this make ipv6 test failed without additional IPFamily in EnvoyProxy?
Yes we'll need to explicitly set IPFamily in EnvoyProxy for the ipv6 test

zhaohuabing (approved):

PR #4820, opened 2024-11-30
[release/v1.2] release note for v1.2.3

zhaohuabing:


site/layouts/shortcodes/helm-version.html
```diff
@@ -6,8 +6,8 @@
 {{- "v1.1.4" -}}
 {{- end -}}
 {{- with (strings.HasPrefix $pagePrefix "v1.2") -}}
-{{- "v1.2.2" -}}
+{{- "v1.2.3" -}}
```
arkodg, line 9: Should this be part of another PR, else this will cause the.docs.site to be broken for some duration

zhaohuabing: A better approach could be releasing first, then modify the site on the main branch.

zhaohuabing: I think a better approach would be publishing the release first, then modify the doc?

arkodg: Yeah +1

arkodg (approved):

zirain (approved):

PR #4837, opened 2024-12-03
api: remove Compression from btp

zirain:
this's introduced for a long time, but never implement.


arkodg:
hey @zirain this is part of the v1.3 milestone https://github.com/envoyproxy/gateway/issues/2451, hoping someone takes it forward, also this shouldnt confuse users since the field is hidden from docs

PR #4852, opened 2024-12-05
ci: enable IPv6 test

zirain:


.github/workflows/build_and_test.yaml
```diff
@@ -122,8 +122,8 @@ jobs:
         - version: v1.30.4
           ipFamily: ipv4
         # Enable these after https://github.com/envoyproxy/gateway/issues/4572 fixed
-        # - version: v1.31.0
-        #   ipFamily: ipv6   # only run ipv6 test on latest version to save time
+        - version: v1.31.0
```
arkodg, line 125: can we do the for v1.30 so we save some time ?

tools/make/kube.mk
```diff
@@ -40,6 +40,17 @@ endif
 YEAR := $(shell date +%Y)
 CONTROLLERGEN_OBJECT_FLAGS :=  object:headerFile="$(ROOT_DIR)/tools/boilerplate/boilerplate.generatego.txt",year=$(YEAR)
 
+.PHONY: prepare-ip-family
+prepare-ip-family: ## Prepare the environment for running e2e tests with ENVOY_PROXY_IP_FAMILY
+	@$(LOG_TARGET)
+ifeq ($(IP_FAMILY),ipv4)
```
arkodg, line 46: can we just reuse the EG enums in make file and scripts to reduce cognitive complexity ?

test/benchmark/config/gatewayclass.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
   name: proxy-config
   namespace: envoy-gateway-system
 spec:
+  ipFamily: $ENVOY_PROXY_IP_FAMILY
```
arkodg, line 19: neat idea :)

.github/workflows/build_and_test.yaml
zirain: sure

tools/make/kube.mk
zirain: `kind` use `ipv4` `ipv6` and `dual`, which is different with EG.

arkodg: ah cool

test/benchmark/config/gatewayclass.yaml
zirain: the bad news is that we need move all EnvoyProxy resources to this file.

zirain:
https://github.com/envoyproxy/gateway/pull/4853 is better

PR #4853, opened 2024-12-05
ci: enable ipv6 test

zirain:


test/config/gatewayclass.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
   name: proxy-config
   namespace: envoy-gateway-system
 spec:
+  ipFamily: IPv4
```
zirain, line 19: IpFamily needs to be set explicitly, `sed` will change it to right value.

zhaohuabing: Nit: use `{IP_FAMILY}` instead of `IPv4` to make it clearer?

zirain: use Ipv4 directly allow you apply this file without run `make prepare-ip-family` for debugging

arkodg (approved):
LGTM thanks for driving this !

zhaohuabing (approved):

PR #4856, opened 2024-12-06
fix: outlier detection disabled by default

### LINKED ISSUE #4854: Outlier detection (passive health checking) is enabled by default without a BTP
*Description*:
Passive health checking seems to be enabled by default without specifying a BTP. This might not always be preferred (even though passive health checks is super useful).

*Repro steps*:
Create a GW and HTTPRoute with no BTP and dump config, you will find `"outlier_detection": {}` present which enabled passive health with default values. 

*Environment*:
EG 1.2.3

*Logs*:
look at the `envoy_cluster_outlier_detection_ejections_active{}` metric to see that outlier detection is active.

lsjostro:
**What type of PR is this?**
* "fix(translator): outlier detection need to be explicit enabled by a BTP"

disable passive health check if not explicitly enabled by a BTP. It's misleading for users that passive health check is enabled with default values without user knowing.

Use case where you don't need outlier detection:
If you users have tiers of proxies (envoy have clusters backends which is also proxies), outlier detection shouldn't be enabled. 

Fixes #4854

Release Notes: No

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -781,7 +781,6 @@ xds:
           ignoreHealthOnHostRemoval: true
           lbPolicy: LEAST_REQUEST
           name: httproute/default/backend/rule/0
-          outlierDetection: {}
           perConnectionBufferLimitBytes: 32768
           type: EDS
       - cluster:
@@ -801,7 +800,6 @@ xds:
           ignoreHealthOnHostRemoval: true
           lbPolicy: LEAST_REQUEST
           name: grpcroute/default/backend/rule/0
-          outlierDetection: {}
           perConnectionBufferLimitBytes: 32768
           type: EDS
           typedExtensionProtocolOptions:
@@ -828,7 +826,6 @@ xds:
           ignoreHealthOnHostRemoval: true
           lbPolicy: LEAST_REQUEST
           name: tcproute/default/backend/rule/-1
-          outlierDetection: {}
           perConnectionBufferLimitBytes: 32768
           type: EDS
       - cluster:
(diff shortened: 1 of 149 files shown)


arkodg:
hey @lsjostro can you sign your commit/ commit amend and force push

arkodg:
can you also add a note in https://github.com/envoyproxy/gateway/blob/main/release-notes/current.yaml under the breaking changes section highlighting that this now disabled by default and If needed, needs to be explicitly set in https://gateway.envoyproxy.io/docs/api/extension_types/#backendtrafficpolicy

lsjostro:
> hey @lsjostro can you sign your commit/ commit amend and force push

trying out my workflow with [jj](https://github.com/martinvonz/jj) so let me lock into that..  so bear with me 😄
excellent challenge though! 😄

lsjostro:
for ref 😄  https://github.com/martinvonz/jj/issues/1399

zhaohuabing:
While I agree that passive health checks does not make sense when EG is used as the first tier of a proxy chain, I believe they can be beneficial in most scenarios where EG sits directly in front of services. Considering this, I see two potential approaches:

1. Enable passive health checks by default and allow users to explicitly disable them using a BTP.


2. Disable passive health checks by default and include a best practices section in the documentation to guide users on enabling them when needed.


Personally, I lean slightly toward the first option, as it provides a more resilient out-of-the-box experience. My assumptions may not be correct, I’d like to hear more practical opinions from users who are actively using EG in their environments.

arkodg:
If 1. is picked, a user setting active health will need to explicitly disable passive health check

zhaohuabing:
> If 1. is picked, a user setting active health will need to explicitly disable passive health check

Yeah, that's a good point. An implicit health check could be hard to be noticed.

We can go ahead with option 2 and add a best practices section in the documentation to ask users to set it if needed.

lsjostro:
agree option 2 is better otherwise we'll need to document why we choose to enable passive health by default and how that will benefit the user. I think it's better to let the user decide and don't have something enabled saying we know better than the user.

arkodg:
hey @lsjostro can you run `make testdata` as well ? and commit those changes

arkodg (approved):
LGTM thanks !
the scan failure seems unrelated to this PR

guydc (approved):

Issue #4861, opened 2024-12-06
Add ש resilience test suite to cover the expected behavior of the data plane for both existing and new proxy instances during specific edge conditions


alexwo:
*Description*:
Ensure we have a test that cover the expected behavior of the data plane for both existing and new proxy instances during specific edge conditions—such as API server unavailability, Envoy Gateway (EG) downtime, or EG leader failures.

alexwo:
we can also add a test case to cover https://github.com/envoyproxy/gateway/issues/4845#issuecomment-2524810183 once fixed

arkodg:
can this be closed @alexwo ?

PR #4862, opened 2024-12-06
feat: data plane & envoyproxy resilience test suite

alexwo:
**What type of PR is this?**
This pr adds a resilience test suite to cover the expected behavior of the data plane for both existing and new proxy instances during specific edge conditions.

**Why do we use Cilium?**
The default network plugin in Kubernetes-in-Docker (Kind) does not support enforcing network policies that restrict connectivity to local services, such as the Kubernetes API server. However, by disabling the default CNI and deploying Cilium as the CNI plugin, we can implement network policies that effectively control access to the API server. This approach allows for precise management of network traffic within the cluster, including the ability to block connectivity to the Kubernetes API server as needed.

**New test suite**
Having a new resilience test suite can be helpful due to the unique network setup and tests involved in resilience testing, which differ significantly from those of existing test suites.

For instance, we need a custom CNI only for this test case, other tests can still use the default.

Fault Injection: Resilience testing involves deliberately introducing failures to assess the system’s response and recovery capabilities. This approach is not typically encompassed by standard functional or performance test suites.

By maintaining a separate resilience test suite, we can effectively simulate and manage system stressors, and separate this concern from the other test cases we have.

Test suite is expected to pass once https://github.com/envoyproxy/gateway/pull/4767 is merged.

Issue:
https://github.com/envoyproxy/gateway/issues/4861


test/resilience/tests/envoyproxy.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+)
+
+func init() {
+	ResilienceTests = append(ResilienceTests, EPResilience)
+	zapLog, err := zap.NewDevelopment() // Use zap.NewProduction() for production
+	if err != nil {
+		panic(fmt.Sprintf("Failed to create logger: %v", err))
+	}
+	log.SetLogger(zapr.NewLogger(zapLog))
```
zirain, line 30: why we need this here?

tools/hack/create-cluster.sh
```diff
@@ -102,3 +138,29 @@ spec:
   ipAddressPools:
   - kube-services
 EOF
+}
+
+RETRY_INTERVAL=5  # seconds
+TIMEOUT=120        # seconds
+ELAPSED_TIME=0
+
+if [ "$CUSTOM_CNI" = "true" ]; then
```
zirain, line 147: can you share more details about why we need cilium?

zirain:
why we need a new test suite instead of reuse existing one?

test/resilience/tests/envoyproxy.go
alexwo: its Integrating the Kubernetes client (client-go) with the zap logging library enhances the visibility of internal operations and API interactions, facilitating more effective debugging and monitoring. so incase things fail, its helps to show some of internal client errors. however we can skip that.

tools/hack/create-cluster.sh
alexwo: The default network plugin in Kubernetes-in-Docker (Kind) does not support enforcing network policies that restrict connectivity to local services, such as the Kubernetes API server. However, by disabling the default CNI and deploying Cilium as the CNI plugin, we can implement network policies that effectively control access to the API server. This approach allows for precise management of network traffic within the cluster, including the ability to block connectivity to the Kubernetes API server as needed.

alexwo:
> Contributor

> why we need a new test suite instead of reuse existing one?
Having a dedicated resilience test suite is helpful due to the unique network setup and tests involved in resilience testing, which differ significantly from those of existing test suites. 

Fault Injection: Resilience testing involves deliberately introducing failures to assess the system’s response and recovery capabilities. This approach is not typically encompassed by standard functional or performance test suites.

By maintaining a separate resilience test suite, we can effectively simulate and manage system stressors, and separate this concern from the other test cases we have.

test/resilience/tests/envoyproxy.go
zirain: it should be in main or somewhere, better not to be in `init` like https://github.com/envoyproxy/gateway/blob/0f68219d191b2b1d5953d59945c5e174db229c1c/test/e2e/e2e_test.go#L30

tools/hack/create-cluster.sh
zirain: I recall kind support network policy as https://github.com/kubernetes-sigs/kind/issues/842 closed.

test/resilience/tests/envoyproxy.go
alexwo: indeed moved that.

test/resilience/tests/envoyproxy.go
```diff
@@ -0,0 +1,74 @@
...
+
+			require.NoError(t, err, "failed to add cleanup")
+
+			ns := "gateway-resilience"
+			routeNN := types.NamespacedName{Name: "backend", Namespace: ns}
+			gwNN := types.NamespacedName{Name: "all-namespaces", Namespace: ns}
+			gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+			resultCh := make(chan error, 1)
+			go func() {
+				err := suite.Kube().CheckConnectivityJob(fmt.Sprintf("http://%s/welcome", gwAddr), 10)
```
guydc, line 67: Can we use something similar to the conformance suite test util that these calls from outside the cluster and wait for x consecutive successes? Is there some limitation on using the metallb LB svc address when using cilium?

test/resilience/tests/envoyproxy.go
```diff
@@ -0,0 +1,74 @@
...
+				resultCh <- err // Send the error (or nil) to the channel
```
guydc, line 68: Is there some sort of envoy XDS cache invalidation (e.g. https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ttl) that we should consider here to determine the time when requests should succeed before failure begin? 
@arkodg

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,140 @@
...
+	Description: "Envoygateway resilience test",
+	Test: func(t *testing.T, suite *suite.ResilienceTestSuite) {
+		ap := kubernetes.Applier{
+			ManifestFS:     suite.ManifestFS,
+			GatewayClass:   suite.GatewayClassName,
+			ControllerName: "gateway.envoyproxy.io/gatewayclass-controller",
+		}
+		ap.MustApplyWithCleanup(t, suite.Client, suite.TimeoutConfig, "testdata/base.yaml", true)
+
+		t.Run("Envoyproxy reconciles missed resources and sync xds after api server connectivity is restored", func(t *testing.T) {
```
guydc, line 47: nit: EnvoyGateway reconciles

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,140 @@
...
+
+			// leader pod should go down, the standby remain
+			t.Log("Verifying deployment scales down to 1 replica")
+			err = suite.Kube().CheckDeploymentReplicas(ctx, envoygateway, namespace, 1, time.Minute)
+			require.NoError(t, err, "Deployment did not scale down")
+
+			t.Log("Monitoring logs for a new leader pod")
+			name, err = suite.Kube().MonitorDeploymentLogs(ctx, time.Now().Add(-time.Minute), namespace, envoygateway, targetString, timeout, false)
+			require.NoError(t, err, "Failed to monitor logs for new leader pod")
+			require.NotEmpty(t, name, "New leader pod name should not be empty")
```
guydc, line 137: Can we add an assertion that the new leader is actually doing some work besides being elected? e.g. add a new route and check that it's actually getting programmed on the proxies?

alexwo: Sure, i will add it.

.github/workflows/build_and_test.yaml
```diff
@@ -176,6 +176,21 @@ jobs:
...
     - name: Read Benchmark report
       run: cat test/benchmark/benchmark_report/benchmark_report.md
+  resilience-test:
+    runs-on: ubuntu-latest
+    if: ${{ ! startsWith(github.event_name, 'push') }}
+    needs: [build]
+    steps:
+    - uses: actions/checkout@v4.2.2
+    - uses: ./tools/github-actions/setup-deps
+    - name: Setup Graphviz
```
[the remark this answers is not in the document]
alexwo: not needed .. just a copy paste.

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,140 @@
...
+		ap.MustApplyWithCleanup(t, suite.Client, suite.TimeoutConfig, "testdata/base.yaml", true)
+
+		t.Run("Envoyproxy reconciles missed resources and sync xds after api server connectivity is restored", func(t *testing.T) {
+			err := suite.Kube().ScaleDeploymentAndWait(context.Background(), envoygateway, namespace, 0, time.Minute, false)
+			require.NoError(t, err, "Failed to scale deployment")
+			err = suite.Kube().ScaleDeploymentAndWait(context.Background(), envoygateway, namespace, 1, time.Minute, false)
+			require.NoError(t, err, "Failed to scale deployment")
+
+			t.Log("Monitoring logs to identify the leader pod")
+			name, err := suite.Kube().MonitorDeploymentLogs(context.Background(), time.Now(), namespace, envoygateway, targetString, timeout, false)
```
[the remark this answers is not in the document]
alexwo: In general we could use the lease resource to understand that there is a leader, however, we can't know the active leader pod. as the lease spec hold the deployment identify as holder "holderIdentity: envoy-gateway-control-plane"

As per:
`
spec:
  holderIdentity: envoy-gateway-control-plane
  leaseDurationSeconds: 40
  renewTime: "2024-12-09T22:19:14.325096Z"
`
For the above test its less interesting but we have a test that simulate a leader only connectivity to api server problem where this is more interesting to find the leading pod only.

We can add a metric to help us with that.

test/resilience/tests/envoyproxy.go
alexwo: Yes, this just made it more easy to run this suite from mac

tools/hack/create-cluster.sh
zirain: will https://github.com/kubernetes-sigs/kube-network-policies help?

alexwo: I will give it a try, maybe it can do the work :)

alexwo: @zirain I tested with kube-network-policies, but it was unable to block connectivity to the API server. In fact, I tried all the other available options as well. Currently, it seems this functionality is only achievable with the Cilium CNI.

test/resilience/tests/envoygateway.go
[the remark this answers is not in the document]
guydc: If all the pods think of themselves as having the same identity, how do they know that they are not the current leader when trying to acquire the lease? Is that based on some other meta field?

test/resilience/tests/envoyproxy.go
guydc: ah, yea, that's a general problem for all of our tests when developing on macos. So, I do prefer that we use the regular tooling.

test/resilience/tests/envoygateway.go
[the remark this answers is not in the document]
alexwo: hmm actually, holderIdentity seems to hold the elected pod identity. Not sure why i saw it as  envoy-gateway-control-plane. we can use that.

test/resilience/tests/envoyproxy.go
alexwo: done

tools/hack/create-cluster.sh
```diff
@@ -44,7 +52,38 @@ ${KIND_CFG}
 EOF
 fi
 fi
-
+if [ "$CUSTOM_CNI" = "true" ]; then
```
zhaohuabing, line 55: Nit, non-blocking: Calico also works for "normal" e2e tests?  We can simply remove the `CUSTOM_CNI` falg and install it by default?

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,248 @@
...
+
+			t.Log("Simulating API server connection failure for all pods")
+			err = suite.WithResCleanUp(ctx, t, func() (client.Object, error) {
+				return suite.Kube().ManageEgress(ctx, apiServerIP, namespace, policyName, true, map[string]string{})
+			})
+			require.NoError(t, err, "Failed to simulate API server connection failure")
+
+			// leader pod should go down, the standby remain
+			t.Log("Verifying deployment scales down to 2 replica")
+			err = suite.Kube().CheckDeploymentReplicas(ctx, envoygateway, namespace, 2, time.Minute)
```
zhaohuabing, line 77: Question: Even though the leader pod is done due to losing connection to the API server, the replicas will still reach 3 because the deployment has been scaled to 3 replicas?

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,248 @@
...
+			_, err = suite.Kube().GetElectedLeader(context.Background(), namespace, leaseName, metav1.Now(), timeout)
+			require.NoError(t, err, "unable to detect leader election")
+
+			t.Log("Simulating API server down for all pods")
+			err = suite.WithResCleanUp(context.Background(), t, func() (client.Object, error) {
+				return suite.Kube().ManageEgress(context.Background(), apiServerIP, namespace, policyName, true, map[string]string{})
+			})
+			require.NoError(t, err, "unable to block api server connectivity")
+
+			err = suite.Kube().WaitForDeploymentReplicaCount(context.Background(), "envoy-gatay", namespace, 0, time.Minute, false)
```
zhaohuabing, line 127: typo? envoy-gatay

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,248 @@
...
+				return suite.Kube().ManageEgress(ctx, apiServerIP, namespace, policyName, true, map[string]string{})
+			})
+			require.NoError(t, err, "Failed to simulate API server connection failure")
+
+			// leader pod should go down, the standby remain
+			t.Log("Verifying deployment scales down to 2 replica")
+			err = suite.Kube().CheckDeploymentReplicas(ctx, envoygateway, namespace, 2, time.Minute)
+			require.NoError(t, err, "Deployment did not scale down")
+
+			ns := "gateway-resilience"
```
zhaohuabing, line 80: Can we apply a new `HTTPRoute` and test it?
This ensures that the xDS comes from the newly eleceted EG instance.

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,248 @@
...
+			routeNN := types.NamespacedName{Name: "backend", Namespace: ns}
```
zhaohuabing, line 81: Can we create a new HTTPRoute and test it?
This ensures that the xDS comes from the newly eleceted EG instance.

zhaohuabing:
Overall, it looks great! This test will definitely improve the stability of EG. It seems the three Envoy Gateway tests can be merged into one.

zhaohuabing:
> I would prefer to make this as part of e2e if there's no special reason. Especailly don't want to introudce a whole new test suite framework.

@envoyproxy/gateway-maintainers please share your options.

It could be included in the "normal" test suite, but I don't have a strong opinion on this.

test/resilience/tests/envoygateway.go
alexwo: You’re absolutely right to consider the intermediate state during the transition, but it’s important to clarify how Kubernetes handles pods that lose connectivity to the API server. When a leader pod loses connectivity to the API server, Kubernetes does not immediately terminate or stop the pod. Instead, the pod itself detects the loss of connectivity, initiates a graceful teardown process, and restarts to attempt to reconnect to the API server. 

The replica count for the deployment remains at 3 throughout the process.

Kubernetes does not schedule a new pod to replace the one that lost connectivity because the existing pod is not considered failed from Kubernetes’ perspective. It’s the responsibility of the application running inside the pod (e.g., the leader election logic) to handle reconnection attempts or restart itself.

test/resilience/tests/envoygateway.go
alexwo: In this test, we simulate an API failure affecting all pods, preventing any pod from assuming the leader role. Consequently, a leader transition does not occur. However, if the xDS configuration for the standby pods was calculated correctly, we would expect the /welcome route to remain accessible after the Envoy proxy reconnects to the secondary nodes. At this is not the case, as xDS is not having correct state in the secondary eg pods.

tools/hack/create-cluster.sh
alexwo: For standard e2e tests, the default kindnet CNI is typically sufficient and offers a simpler, more resource-efficient setup. Introducing Calico can be reserved for scenarios that specifically require its advanced networking capabilities. 
At this point the Calico deployment adds an extra 2-3 min for each create-cluster.

test/resilience/tests/envoygateway.go
zhaohuabing: Thanks for the explanation!

tools/hack/create-cluster.sh
zhaohuabing: Yeah, an extra 2-3 minutes will slow down the tests, which already take a long time to finish.

zhaohuabing (dismissed):

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,242 @@
...
+
+			t.Log("Waiting for leader election")
+			// Ensure leadership was taken
+			name, err := suite.Kube().GetElectedLeader(context.Background(), namespace, leaseName, metav1.Now(), time.Minute*2)
+			require.NoError(t, err, "unable to detect leader election")
+
+			t.Log("Marking the identified pod as leader")
+			suite.Kube().MarkAsLeader(namespace, name)
+
+			t.Log("Simulating API server connection failure for all pods")
```
guydc, line 69: If all pods loose connectivity, don't we expect that they would all fail eventually once becoming leaders?

alexwo: Pods rely on connectivity to the API server to participate in leader election processes. Without this connectivity, they cannot become leaders. Pods should continue to operate using their last known good configuration (xDS) and serve traffic accordingly.

zhaohuabing:
@alexwo Could you please add a few lines of the comments for each test? (Like you explained in the response of this PR) This will help other contributors understand why we need the test and how the test works. Thanks!

zhaohuabing (approved):

zhaohuabing:
@alexwo I think this is not right:

```
Waiting for deployment "envoy-gateway": ready replicas 3/2
Waiting for deployment "envoy-gateway": ready replicas 3/2
Waiting for deployment "envoy-gateway": ready replicas 3/2
Deployment "envoy-gateway" reached 2 replicas as expected.
    helpers.go:609: 2024-12-11T21:54:16.694551571Z: Conditions matched expectations
```

The ready replicas should be 2 even though https://github.com/envoyproxy/gateway/pull/4767 is not merged. This step only tests the leader pod losing its status.

zhaohuabing (dismissed):

test/resilience/tests/envoygateway.go
```diff
@@ -0,0 +1,242 @@
...
+			err = suite.WithResCleanUp(ctx, t, func() (client.Object, error) {
+				return suite.Kube().ManageEgress(ctx, apiServerIP, namespace, policyName, true, map[string]string{})
```
zhaohuabing, line 71: This change will allow test Resilience test suite with #4767 pass.  See https://github.com/envoyproxy/gateway/pull/4904

Not sure the exact reason, it seems like a bug from the Calico CNI. 

```suggestion
				return suite.Kube().ManageEgress(ctx, apiServerIP, namespace, policyName, true, map[string]string{"app.kubernetes.io/name": "gateway-helm"})
```

alexwo: hmm i have merged the suggestion,  the Resilience suite still does not pass. 

Did this pass  for you when running locally? It’s possible that delays in resource cleanup caused the tests to behave inconsistently?

alexwo: Understood. Without this fix, the tests fail even with the xDS adjustments. However, this minor change ensures that everything functions as expected. right?

arkodg:
https://github.com/envoyproxy/gateway/pull/4767 has been merged, that should unblock this PR now ?

zhaohuabing (approved):

test/resilience/tests/envoygateway.go
zhaohuabing: @alexwo The Envoy wasn't able to connect to the standby EG instances with the egress rule even though only the API server IP was blocked. Adding a podselector solved that. 

This looks like a Calico bug to me.

zhaohuabing:
> #4767 has been merged, that should unblock this PR now ?

@arkodg I believe this PR is ready to be merged.

Should we also cherry-pick this to v1.2.4?

zhaohuabing (approved):

guydc (approved):

PR #4784, opened 2024-11-26
fix: btlsp section name doesn't support port name

zhaohuabing:
Fixes #4769

This PR uses port name as the `sectionName` for the targetRefs of the `BackendTLSPolicy`.

Release Notes: Yes

PR #4868, opened 2024-12-09
docs: update BackendTLSPolicy docs

zhaohuabing:
Update BackendTLSPolicy docs for https://github.com/envoyproxy/gateway/pull/4784

### DIFF
--- a/site/content/en/docs/tasks/security/backend-tls.md
+++ b/site/content/en/docs/tasks/security/backend-tls.md
@@ -17,7 +17,7 @@ Envoy Gateway supports the Gateway-API defined [BackendTLSPolicy][].
 
 ## TLS Certificates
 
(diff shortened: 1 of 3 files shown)


arkodg (approved):

arkodg (approved):

lsjostro:
This broke our setup when upgraded. Could we please update the release notes to highlight it as a breaking change? Just so that others not run into this causing downtime for their services.

arkodg:
are you using v0.0.0-latest @lsjostro ? this hasn't released yet https://github.com/envoyproxy/gateway/pull/4784/files

lsjostro:
Nope, the feature/fix was released with v1.2.4. (maybe not the docs?).

arkodg:
yeah you're right @lsjostro the fix was part of v1.2.4 and since the docs highlighted the example, it should be surfaced as a breaking change and not a bug fix

Issue #4845, opened 2024-12-04
Proxies connected to the secondary gateway do not receive configuration


cico78:
*Description*:

When a proxy is connected to a secondary gateway it does not receive configuration.
This is a new behaviour after the upgrade to `1.2.3` and was not happening with version `1.1.0`

*Repro steps*:
Have two gateways configured with leader election (default configuration, version 1.2.3). Scale up envoy deployment and check for proxies connected to the secondary gateway. Those proxies will not receive the configurations complaining that the initial timeout is failing (see logs).

Those proxies will be able to get the configuration once the primary gateway is deleted and the secondary becomes the new primary.

*Environment*:

Gateway version: `1.2.3`
Envoy version: `1.32.1`
Kubernetes: `EKS 1.28`

*Logs*:

Default log level is set to `debug`

logs from envoy PROXY-POD-NAME:

```
...
[2024-12-04 07:48:34.237][1][info][config] [source/server/configuration_impl.cc:124] loading 0 static secret(s)
[2024-12-04 07:48:34.237][1][info][config] [source/server/configuration_impl.cc:130] loading 3 cluster(s)
[2024-12-04 07:48:34.241][1][info][config] [source/server/configuration_impl.cc:138] loading 1 listener(s)
[2024-12-04 07:48:34.242][1][warning][misc] [source/extensions/filters/network/http_connection_manager/config.cc:88] internal_address_config is not configured. The existing default behaviour will trust RFC1918 IP addresses, but this will be changed in next release. Please explictily config internal address config as the migration step or config the envoy.reloadable_features.explicit_internal_address_config to true to untrust all ips by default
[2024-12-04 07:48:34.243][1][info][config] [source/server/configuration_impl.cc:154] loading stats configuration
[2024-12-04 07:48:34.285][1][info][main] [source/server/server.cc:990] starting main dispatch loop
[2024-12-04 07:48:34.296][1][info][runtime] [source/common/runtime/runtime_impl.cc:631] RTDS has finished initialization
[2024-12-04 07:48:34.296][1][info][upstream] [source/common/upstream/cluster_manager_impl.cc:245] cm init: initializing cds
[2024-12-04 07:48:49.295][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:130] gRPC config: initial fetch timed out for type.googleapis.com/envoy.config.cluster.v3.Cluster
[...]

PR #4878, opened 2024-12-09
fix: trigger a resync to avoid missing any status updates

zhaohuabing:
This PR does two things:
* Remove wait to make sure the status updater won't block the reconcile loop in a non-leader scenario.
* Trigger a reconciliation after starting the status updater to avoid missing any status updates.

Fixes: #4845

Release Note: Yes

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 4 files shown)


arkodg:
@zhaohuabing I dont think this will solve the original problem because the re sync will not trigger the status watchable subscribed handler to get any updates, because nothing would have changed

arkodg:
my suggestion would be to 
1. always push status updates via watchable, which was implemented in https://github.com/envoyproxy/gateway/pull/4767
and
2. we should block the status updater watchable subscribe goroutine similar to https://github.com/envoyproxy/gateway/blob/9ef758b1f044db4e02d05cbae1b160d91e81d598/internal/infrastructure/runner/runner.go#L70 and only enable it once its part of the controller thats the leader
wdyt @alexwo @guydc

alexwo:
> my suggestion would be to
> 
> 1. always push status updates via watchable, which was implemented in [fix: decoup gateway status updates from the reconciler #4767](https://github.com/envoyproxy/gateway/pull/4767)
>    and
> 2. we should block the status updater watchable subscribe goroutine similar to https://github.com/envoyproxy/gateway/blob/9ef758b1f044db4e02d05cbae1b160d91e81d598/internal/infrastructure/runner/runner.go#L70
>     and only enable it once its part of the controller thats the leader
>    wdyt @alexwo @guydc

+1 sounds good to me.

PR #4893, opened 2024-12-11
chore: organize go.mod require sections

shahar-h:
Since currently there are mixed direct and indirect dependencies inside `go.mod` require sections `go mod tidy` creates a new section each time. See [reference](https://github.com/golang/go/issues/56471#issuecomment-2383946087).

Before this change there were 4 require sections. now there are 2 - one with direct dependencies and one with indirect dependencies, as expected.

Release Notes: No


guydc (approved):
LGTM, thanks

arkodg (approved):

Issue #4746, opened 2024-11-20
BackendTrafficPolicy openapi schemas do not validate correctly


nicks:
*Description*:
I'm trying to use kubeconform to validate BackendTrafficPolicy

I get this error:

```
stdin - BackendTrafficPolicy repos-new is invalid: problem validating schema. Check JSON formatting: jsonschema: '/spec/healthCheck/active/interval' does not validate with https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/gateway.envoyproxy.io/backendtrafficpolicy_v1alpha1.json#/properties/spec/properties/healthCheck/properties/active/properties/interval/format: '3s' is not valid 'duration'
stdin - BackendTrafficPolicy admin-insights is invalid: problem validating schema. Check JSON formatting: jsonschema: '/spec/healthCheck/active/timeout' does not validate with https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/gateway.envoyproxy.io/backendtrafficpolicy_v1alpha1.json#/properties/spec/properties/healthCheck/properties/active/properties/timeout/format: '20ms' is not valid 'duration'
```

*Additional info*:
I think the openapi schemas that we're generating are wrong. Particularly this line:

https://github.com/envoyproxy/gateway/blob/6c6633c2a8bfc772415ea3bf68e14ef20064cb22/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_backendtrafficpolicies.yaml#L314

They say that `timeout` and `interval` are openapi durations. But they're not, they're a Kubernetes-specific duration format.

arkodg:
thanks for raising this, do we need to edit these to `string` instead ?
The CEL validation on the `metav1.Duration` should still kick in and allow acceptable values
https://github.com/envoyproxy/gateway/blob/6c6633c2a8bfc772415ea3bf68e14ef20064cb22/api/v1alpha1/healthcheck_types.go#L81

nicks:
ya, i poked around at how Cluster API is doing it, and i think they just leave the `kubebuilder` validation off entirely?
https://github.com/kubernetes-sigs/cluster-api/blob/781d1e4c39286d3d1e93c71557b5d02c2e78961b/api/v1beta1/machinehealthcheck_types.go#L101

and it will do the right thing with the existing `metav1.Duration` type?

arkodg:
thanks for helping us find this @nicks 

[...]

PR #4931, opened 2024-12-14
Use gwapiv1.Duration instead of metav1.Duration in the API

arkodg:
* Also rm `+kubebuilder:validation:Format=duration` and instead rely on the CEL validation of gwapiv1.Duration

Fixes: https://github.com/envoyproxy/gateway/issues/4746


arkodg:
closing this PR since its become inactive, @Ninir feel free to raise a new PR for this

Issue #4627, opened 2024-11-04
Support Lua Filters in EnvoyExtensionPolicy


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Allow users to configure Lua filters using the EnvoyExtensionPolicy https://gateway.envoyproxy.io/docs/api/extension_types/#envoyextensionpolicyspec

relates to https://github.com/envoyproxy/gateway/issues/2025

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zhaohuabing:
+1 Lua filter is essential for users migrating from the Nginx ecosystem.

arkodg:
Here's a workaround until this is implemented using EnvoyPatchPolicy which needs to be enabled https://gateway.envoyproxy.io/docs/tasks/extensibility/envoy-patch-policy/

This patch assumes the Gateway installed is the same as the one from the Quickstart https://gateway.envoyproxy.io/docs/tasks/quickstart/

```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: lua-header-patch-policy 
  namespace: default 
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
  type: JSONPatch
  jsonPatches:
    - type: "type.googleapis.com/envoy.config.listener.v3.Listener"
      # The listener name is of the form <GatewayNamespace>/<GatewayName>/<GatewayListenerName>
      name: default/eg/http
      operation:
        op: add
        path: "/default_filter_chain/filters/0/typed_config/http_filters/0"
        value:
          name: "envoy.filters.http.lua"
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
            inline_code: |
              function envoy_on_request(request_handle)
                print("Hello World")
              end
```  

You will need to also enable info logging in the proxy if testing this example
```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: envoyproxy-logging-config
spec:
  logging:
    level:
      default: info
```            


Note if linking this to a HTTPS listener, the `path` prefix will need to be updated to `/filter_chains/0` instead of `default_filter_chain`

arkodg:
go for it @rudrakhp !

would be great to think about how the Lua config can be validated before pushing it to the proxy 
[...]

PR #4932, opened 2024-12-14
api: lua support in EnvoyExtensionPolicy

rudrakhp:
**What type of PR is this?**

api: lua support in EnvoyExtensionPolicy

Enables EGW users to define Lua filters natively without a need to patch Envoy filters. 

Related #4627 

Release Notes: No


zirain:
for a general idea, I think we should support in only in EnvoyExtensionPolicy?

rudrakhp:
@zirain do you mean have no support for per route Lua filters at all? Or have only global filter for the first phase and support per route filter later?

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -54,11 +54,19 @@ type EnvoyExtensionPolicySpec struct {
...
 	// +optional
 	ExtProc []ExtProc `json:"extProc,omitempty"`
+
+	// LuaFilters is an ordered list of Lua filters
+	// that should be added to the envoy filter chain
+	//
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	// +notImplementedHide
+	LuaFilters []LuaFilter `json:"lua,omitempty"`
```
zhaohuabing, line 69: Prefer to keep the same naming pattern with Wasm and ExtProc.
```suggestion
	Lua []Lua `json:"lua,omitempty"`
```

zirain:
> @zirain do you mean have no support for per route Lua filters at all? Or have only global filter for the first phase and support per route filter later?

I means we can start from support it in EnvoyExtensionPolicy first.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,67 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package v1alpha1
+
+// LuaFilter defines a Lua extension that applies globally
```
zhaohuabing, line 8: The comment here could be misleading, as the EEP can apply at either the Gateway or HTTPRoute level.

zhaohuabing:
> @zirain do you mean have no support for per route Lua filters at all? Or have only global filter for the first phase and support per route filter later?

@rudrakhp Let's only support Lua in the EEP. EEP can be applied to either a Gateway or an HTTPRoute.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,67 @@
...
+	// using custom stat prefix to distinguish emitted statistics.
+	//
+	// +optional
+	StatPrefix *string `json:"statPrefix,omitempty"`
+}
+
+// LuaPerRouteFilter defines a Lua extension that applies to specific routes based on route match.
+// Only one of Disabled, Name or OverrideSource must be set.
+// +kubebuilder:validation:XValidation:rule="has(self.disabled) ? (!has(self.name) && !has(self.overrideSource)) : (has(self.name) ? !has(self.overrideSource) : has(self.overrideSource))",message="Exactly one of disabled, name, or overrideSource must be set."
+type LuaPerRouteFilter struct {
```
zhaohuabing, line 31: We should not need a `LuaPerRouteFilter` struct in the API. Let's define a universal API that can be used for both a Gateway or an HTTPRoute.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,67 @@
...
+	// DefaultSourceCode is the default LuaSource that will be executed
+	// unless overridden by LuaPerRouteFilter configured for the selected route
+	DefaultSourceCode *LuaSource `json:"defaultSourceCode"`
+	// StatPrefix is the additional prefix to use when emitting statistics.
+	// By default, metrics are emitted in .lua. namespace.
+	// If multiple lua filters are configured in a filter chain, the stats from each filter instance can be emitted
+	// using custom stat prefix to distinguish emitted statistics.
+	//
+	// +optional
+	StatPrefix *string `json:"statPrefix,omitempty"`
```
zhaohuabing, line 25: The StatPrefix is a low-level Envoy configuration and it can be generated by EG. We don't need to expose it to the EG API.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,67 @@
...
+	// OverrideSource is a LuaSource defined for this specific route
+	//
+	// +optional
+	OverrideSource *LuaSource `json:"overrideSource,omitempty"`
+}
+
+// LuaSource contains source code information for a user defined Lua script
+// Only one of FileName, InlineBytes, InlineString or EnvironmentVariable must be set
+// +kubebuilder:validation:XValidation:rule="has(self.fileName) ? (!has(self.inlineBytes) && !has(self.inlineString) && !has(self.environmentVariable)) : (has(self.inlineBytes) ? (!has(self.inlineString) && !has(self.environmentVariable)) : has(self.inlineString) ? !has(self.environmentVariable) : has(self.environmentVariable))",message="Exactly one of fileName, inlineBytes, inlineString, or environmentVariable must be set."
+type LuaSource struct {
```
zhaohuabing, line 50: I suggest to support an `Inline` for inline striing and  a `ValueRef` to a ConfigMap. You can find an example here:

https://github.com/envoyproxy/gateway/blob/a9f6cd2b414e9f2c25012fb251c88088e745d1cb/api/v1alpha1/shared_types.go#L716-L739

zhaohuabing:
@rudrakhp Thanks for working on this! The API looks quite similar to the Envoy lua configuration, which is not ideal for EG. The Envoy Lua configuration is primarily designed as a programming API for a control plane, rather than for direct use by end users. 

Please refer to the comments and the other EG API to redefine the Lua API. Thanks!

rudrakhp:
@zhaohuabing I was actually wondering if we need to opinionated about what we expose to users or be as transparent as possible. I will incorporate the changes recommended in the comments, thanks!
>EEP can be applied to either a Gateway or an HTTPRoute.

I think starting with only Gateway level support might be a good start.

**One question about the route level filter (to be worked upon in another PR):** 
If the lua filter is per route, envoy provides a way for the per route filter to point to lua defined in the global filter. Do we need to have similar functionality in GW as well, i.e, users can define global EEP and point to that Lua in the per route EEP? If yes, validation of EEP might become tricky - as per route EEPs will depend on the global EEP definition. Other option is we can opinionated about it and allow only inline overrides per route. 
Note: Per route Lua filter doesn't support file based definition like Global Lua filter.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,41 @@
...
+type Lua struct {
+	// Source is the LuaSource that will be executed for this extension.
+	//
+	// +optional
+	Source *LuaSource `json:"source"`
+	// Disabled is the status of the extension.
+	// Used for disabling Lua extensions at a route level.
+	//
+	// +optional
+	Disabled *bool `json:"disabled,omitempty"`
```
zhaohuabing, line 23: Do we need a `Disabled` knob here?

release-notes/current.yaml
```diff
@@ -17,6 +17,7 @@ new features: |
   Added support for trusted CIDRs in the ClientIPDetectionSettings API
   Added support for sending attributes to external processor in EnvoyExtensionPolicy API
   Added support for patching EnvoyProxy.spec.provider.kubernetes.envoyHpa and EnvoyProxy.spec.provider.kubernetes.envoyPDB
+  Added Lua support in EnvoyExtensionPolicy and HTTPRouteFilter APIs
```
zhaohuabing, line 20: Nit: We don't need this since the API is currently hiden. let's add release note after implementing the Lua filter.

api/v1alpha1/lua_types.go
rudrakhp: @zhaohuabing for usecases where Lua is defined at GW level but we want to exclude some routes?

release-notes/current.yaml
rudrakhp: Done!

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// +kubebuilder:validation:XValidation:rule="has(self.inline) ? !has(self.valueRef) : has(self.valueRef)",message="Exactly one of inline or valueRef must be set."
+type LuaSource struct {
+	// Inline contains the value as an inline string.
+	//
+	// +optional
+	Inline *string `json:"inline,omitempty"`
+	// ValueRef contains the contents of the body
+	// specified as a local object reference.
+	// Only a reference to ConfigMap is supported.
```
zirain, line 37: need cel validation for this `Only a reference to ConfigMap is supported.`

what about the configmap key?

rudrakhp: Added checks for object kind and default core group.

rudrakhp: Also added a mention of the key `lua` that will be used for fetching the value.

api/v1alpha1/lua_types.go
rudrakhp: >Besides, the EEP at the HTTPRoute level will override the one at the Gateway level.

Yes, but there should be a lever to turn it off altogether, rather than just override. WDYT?

rudrakhp: >I think this is a common issue for all the xPolicies, so prefer to solve it on the global level if needed

Removed disable for now, we can solve this later if needed once the bulk of Lua validation and implementation is done.

zhaohuabing (approved):
LGTM thanks!

zirain (approved):

zhaohuabing:
@rudrakhp @zirain Let's hold this until @arkodg is back from vacation as he may have more insights on the API.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,27 @@
...
+package v1alpha1
+
+import gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+// Lua defines a Lua extension
+// Only one of Inline or ValueRef must be set
+//
+// +kubebuilder:validation:XValidation:rule="has(self.inline) ? !has(self.valueRef) : has(self.valueRef)",message="Exactly one of inline or valueRef must be set."
+type Lua struct {
+	// Inline contains the source code as an inline string.
```
arkodg, line 15: also needs a `Type` enum here e.g. https://github.com/envoyproxy/gateway/blob/10a31f12f618a4781e6568196e923884067e8c23/api/v1alpha1/shared_types.go#L723

rudrakhp: @arkodg I was kind of making it work by allowing only one of both value types, but added a type as suggested.

api/v1alpha1/lua_types.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// +kubebuilder:validation:XValidation:rule="(self.type == 'Inline' && has(self.inline) && !has(self.valueRef)) || (self.type == 'ValueRef' && !has(self.inline) && has(self.valueRef))",message="Exactly one of inline or valueRef must be set with correct type."
+type Lua struct {
+	// Type is the type of method to use to read the Lua value.
+	// Valid values are Inline and ValueRef, default is Inline.
+	//
+	// +kubebuilder:default=Inline
+	// +unionDiscriminator
+	// +required
+	Type *LuaValueType `json:"type,omitempty"`
```
arkodg, line 33:
```suggestion
	Type LuaValueType `json:"type"`
```

api/v1alpha1/lua_types.go
arkodg: yeah we've followed https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md to implement unions in most places of the API to getter better schema for validation

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM thanks!

denniskniep:
> current approach prioritizes velocity and ease of review 

That makes totally sense 

Probably we need to distinguish between `release-build` and `latest-build` / `build from source`

Following behavior during a `release-build` would be optimal from my point of view if there is an `+notImplementedHide` annotation:
* not rendered into the CRDs
* not integrated into docs

I don´t know if there is appetite to change this into the proposed direction? If that´s too much effort then an "Not Implemented" in the docs would be better than nothing :)

PR #4936, opened 2024-12-16
feat(translator): extension server should fail close

### LINKED ISSUE #4155: "Fail Safe" control knob for Extension Server
Currently, the processing of [Extension Server](https://gateway.envoyproxy.io/docs/tasks/extensibility/extension-server/) logic (ie `Translate()` step) is "best effort".  If the Extension Server fails in some way (ie is not available, crashes during request handling), than the envoy configuration is not impacted.

In some situations, this could be a "bad thing".  For example, if the Extension Server is being used to add a default Authz filter to all Listeners, if the grpc call to the Extension Server fails, than the Listener will still be activated but will _not_ have the Authz filter (thus incorrectly exposing the resource without the desired protection). 

Ideally, similar to #3873, there would be an option added to [`ExtensionManager`](https://gateway.envoyproxy.io/docs/api/extension_types/#extensionmanager) which would allow either "fail open" (current behavior of best effort) or "fail closed" (alternate behavior of disabling the resource associated with the failed hook).

liorokman:
This PR changes the default behavior for extension servers to fail-close instead of fail open.

Fixes #4155 

Release Notes: Yes


release-notes/current.yaml
```diff
@@ -7,6 +7,7 @@ breaking changes: |
   An empty TLS ALPNProtocols list is now treated as user-defined disablement of the TLS ALPN extension.
   Outlier detection (passive health check) is now disabled by default.
   refer to https://gateway.envoyproxy.io/docs/api/extension_types/#backendtrafficpolicy for working with passive health checks.
+  Envoy Gateway treats errors in calls to an extension service as fail-closed by default. The previous behavior can be enabled by setting the `failOpen` field to `true` in the extension service configuration.
```
guydc, line 10: Maybe we can elaborate a bit here on what fail closed means. Basically, translation/connectivity errors would lead to replacement of proxy routing config with immediate responses.

internal/xds/translator/extensionserver_test.go
```diff
@@ -202,6 +202,17 @@ func (t *testingExtensionServer) PostHTTPListenerModify(_ context.Context, req *
 
 // PostTranslateModifyHook inserts and overrides some clusters/secrets
 func (t *testingExtensionServer) PostTranslateModify(_ context.Context, req *pb.PostTranslateModifyRequest) (*pb.PostTranslateModifyResponse, error) {
+	for _, cluster := range req.Clusters {
+		if edsConfig := cluster.GetEdsClusterConfig(); edsConfig != nil {
+			if strings.Contains(edsConfig.ServiceName, "fail-close-error") {
```
arkodg, line 207: would be good to have some comments in here explaining what this is simulating

internal/xds/translator/translator.go
```diff
@@ -500,6 +572,21 @@ func (t *Translator) addRouteToRouteConfig(
 		// If no extension exists (or it doesn't subscribe to this hook) then this is a quick no-op.
 		if err = processExtensionPostVHostHook(vHost, t.ExtensionManager); err != nil {
 			errs = errors.Join(errs, err)
+			if t.ExtensionManager != nil && !(*t.ExtensionManager).FailOpen() {
+				vHost.Routes = []*routev3.Route{
```
arkodg, line 576: would be good to have some comments in here explaining what this setting means, and why its added here, similar comment would be great for other snippets where xds is being modified to return a 500

arkodg:
thanks @liorokman , overall LGTM, added some non blocking comments around adding comments

internal/xds/translator/translator.go
```diff
@@ -179,11 +186,71 @@ func (t *Translator) notifyExtensionServerAboutListeners(
...
+	} else {
+		clearAllRoutes(hcm)
+		if err := replaceHCMInFilterChain(hcm, listener.DefaultFilterChain); err != nil {
+			errs = errors.Join(errs, err)
+		}
+	}
+	for _, filter := range listener.FilterChains {
+		hcm, err := findHCMinFilterChain(filter)
+		if err != nil {
+			// no HCM found, skip
```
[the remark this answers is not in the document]
liorokman: The listener hook will be called for non-HTTP(s) listeners.

But for non-HTTP(s) listeners, it's not possible to return an HTTP 500 response. What would "failing-close" look like here?

release-notes/current.yaml
liorokman: Done.

internal/xds/translator/translator.go
liorokman: Done

internal/xds/translator/extensionserver_test.go
liorokman: Done.

internal/xds/translator/translator.go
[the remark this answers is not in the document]
guydc: Maybe we can remove the filter chain such that envoy resets connections?

guydc (approved):

arkodg (approved):
LGTM thanks !

internal/xds/translator/translator.go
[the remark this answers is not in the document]
liorokman: OK, I'll do something like that.

PR #4941, opened 2024-12-17
chore: set go version for the osv scanner

zhaohuabing:


zirain (approved):

shahar-h:
@zhaohuabing installing go before running `osv-scanner-action` won't work since `osv-scanner` runs inside a container with pre-installed go version. Anyway the workflow is broken now after this change:
image

I opened an issue in `osv-scanner` repo that suggests to either release github-action image once a new go version is available or allow running osv-scanner on machine so one can install its own go version before running the scan.

Before this PR the scan failed on  `golang.org/x/crypto` vulnerability in `tools/src/buf/go.mod`. I wonder if we really need to perform a recursive scan or only scan root `go.mod`. WDYT?

zhaohuabing:
@shahar-h This PR has been merged too early - it was a draft.  :-(

shahar-h:
> > I opened an [google/osv-scanner#1453](https://github.com/google/osv-scanner/issues/1453) in osv-scanner repo that suggests to either release github-action image once a new go version is available or allow running osv-scanner on machine so one can install its own go version before running the scan.
> 
> @shahar-h Thanks for digging into this! This PR has been merged too early - it was a draft, and I was exploring the solutions.
> 
> > Before this PR the scan failed on golang.org/x/crypto vulnerability in tools/src/buf/go.mod. I wonder if we really need to perform a recursive scan or only scan root go.mod. WDYT?
> 
> I think we need a recursive scan to make sure all the dependencies's license are accurate and compliant.
> 
> OSV scanner is quite troblesome, it brings additional maintainnce efforts for nearly every dependent bot updates. Any idea to improve this ? @envoyproxy/gateway-maintainers

* Recursive scan is only done for [vulnerabilities scan](https://github.com/envoyproxy/gateway/blob/4cba2e2703847bb106b8d3bf5cb0eb355aabf9e1/.github/workflows/osv-scanner.yml#L47). For license scan only root `go.mod` is scanned since we don't care about licenses of dev packages.
I wonder if we also need recursive vulnerabilities scan. I added it back then to prevent supply chain attacks, but on the other the scan also fails because of CVEs like [this one ](https://osv.dev/vulnerability/GO-2024-3321) which is only relevant for packages that are compiled with the released binary, not for packages used in dev / CI/CD.
* Regarding the maintenance effort - we can omit the version [field](https://github.com/envoyproxy/gateway/blob/4cba2e2703847bb106b8d3bf5cb0eb355aabf9e1/tools/osv-scanner/license-scan-config.toml#L9) in `PackageOverrides` sections with the risk of package relicensing on version upgrade.
package relicensing can occur both on minor release(i.e. terraform) or major release(i.e. grafana).
* Currently call analysis doesn't work so there are many false positives in vulnerability scan since we use a newer go version than the one installed in `osv-scanner-action` image. One of `osv-scanner` maintainers [agreed](https://github.com/google/osv-scanner/issues/1453#issuecomment-2549906182) that it's a good idea to support running `osv-scanner` on host, thus allowing the consumer to pre-install its desired go version. Once it's supported we will have no false positives which will make the maintenance easier.

PR #4955, opened 2024-12-20
feat(translator): preserve route order in EnvoyProxy

### LINKED ISSUE #4943: Support user-defined route order
*Description*:
Currently, envoy gateway orders routes according to requirements in GW-API: https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule

>Proxy or Load Balancer routing configuration generated from HTTPRoutes MUST prioritize matches based on the following criteria, continuing on ties. Across all rules specified on applicable Routes, precedence must be given to the match having:
> - “Exact” path match.
> - “Prefix” path match with largest number of characters.
> - Method match.
> - Largest number of header matches.
> - Largest number of query param matches.
> - Note: The precedence of RegularExpression path matches are implementation-specific.

Some projects support user-defined route ordering:
- [istio](https://istio.io/latest/docs/concepts/traffic-management/):
> Each virtual service consists of a set of routing rules that are evaluated in order
- [gloo edge](https://docs.solo.io/gloo-edge/latest/reference/cli/glooctl_route_sort/): 
> The order of routes matters. A route is selected for a request based on the first matching route matcher in the virtual service’s list

The sorting of routes improves consistency and guards against end user mistakes (e.g. more specific prefix matchers being ignored because shorter matchers are specified first). However, advanced users and users migrating from other projects may need more control over the route order (e.g. determining the order of regex matchers not based on length).  

The default behavior of Envoy Gateway must remain backwards compatible and compliant with GW-API specs. However, EG can introduce a new flag in `EnvoyGateway` configmap or `ClientTrafficPolicy` CRD to support user-defined route ordering. When multiple HTTPRoute resources are attached to a GW, date-of-creation sorting can be used to determine the order of the route groups. 

Allowing users to define this behavior on a per `HTTPRoute` or even `HTTPRouteRule` level with `BackendTrafficPolicy`/`HTTPRouteFilter` would create significant implementation complexity and possible ambiguity.   

[optional *Relevant Links*:]

guydc:
**What type of PR is this?**

Add an option to `EnvoyProxy` to preserve the user-defined route order. 

Fixes #4943 

Release Notes: Yes/No


test/e2e/base/manifests.yaml
```diff
@@ -35,21 +35,21 @@ spec:
       allowedRoutes:
         namespaces:
           from: Same
----
-apiVersion: gateway.networking.k8s.io/v1
```
zhaohuabing, line 39: Why comments the all-namespaces Gateway?

zhaohuabing (approved):
LGTM thanks!

arkodg:
CTP doesn't feel like the right place for this, imo this is a global setting that should live in either EnvoyProxy or EnvoyGateway

internal/gatewayapi/sort.go
```diff
@@ -85,8 +85,10 @@ func (x XdsIRRoutes) Less(i, j int) bool {
 func sortXdsIRMap(xdsIR resource.XdsIRMap) {
 	for _, irItem := range xdsIR {
 		for _, http := range irItem.HTTP {
-			// descending order
-			sort.Sort(sort.Reverse(XdsIRRoutes(http.Routes)))
+			if !http.PreserveRouteOrder {
```
arkodg, line 88: should skip the entire sort at the top level instead of inside the sort
https://github.com/envoyproxy/gateway/blob/731d0d0722532b654b9d508fea641e37ce63f98d/internal/gatewayapi/translator.go#L228

internal/gatewayapi/listener.go
```diff
@@ -134,6 +134,7 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
 					// see more https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/gwapiv1.Listener.
 					irListener.Hostnames = append(irListener.Hostnames, "*")
 				}
+				irListener.PreserveRouteOrder = getPreserveRouteOrder(gateway.envoyProxy)
```
arkodg, line 137: probably not needed

arkodg (dismissed):
LGTM thanks !

internal/gatewayapi/sort.go
arkodg: the current approach is the right approach, because there may be an EP per GW, and so listeners within the entire XdsIRMap may have different values

arkodg (approved):

cnvergence (approved):

Issue #4756, opened 2024-11-21
Usage based Rate Limiting (Counting from response header values)


arkodg:

arkodg:
cc @zirain @missBerg

zirain:
xref: https://github.com/envoyproxy/ratelimit/pull/752

arkodg:
can we update the ratelimit filter in Envoy https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting#per-connection-or-per-http-request-rate-limiting to regenerate ratelimit actions attached to the route and scope it for responses (using the `stage` field) ?

mathetake:
so the problem for ai-gateway usecase with this getting use from header is; for streaming endpoints, the content is sent line by line and envoy shouldn't block the entire body but let it go through. OpenAI and AWS's streaming chat endpoint sends the usage stats at the very end of streaming, using a header as a communication medium is not suitable for that case. I think it's better to use dynamic_metadata as a communication between the producer (ai-gateway's case extproc which analyzes the streaming content) and the rate limit filter, and make the rate limit filter to subtract the usage from the budget on stream closure (not the header phase).

arkodg:
thanks for explaining that @mathetake , so we can use this GH issue to track 3 items
* reusing ratelimit filter in response path in Envoy Proxy
* supporting ratelimiting based on response In EG
* exposing metadata as a first class selector (like header and client ip) in EG

mathetake:
sounds good - i will open an issue in envoyproxy/envoy to start the discussion with maintainers (i think wbpcode is owner of ratelimit)

mathetake:
so the Envoy side PR has landed the main - next is EG

PR #4957, opened 2024-12-20
api: adds cost specifier to RateLimitRule

mathetake:
**What type of PR is this?**

This adds a new API into rate limit API.

This is for #4756. Basically, this adds the API to specify the hits_addend for the rate limit rule.
Especially, configuring the hits_addend in the response_path allows us to "reduce" the counter
based on the response content that affects the subsequent requests. This will enable the "token"
based request limits that are required fro AI gateway.

This is based on the Envoy's brand new features called apply_on_stream_done
and per-descriptor level hits_addend configuration introduced in 
* https://github.com/envoyproxy/envoy/pull/37548
* https://github.com/envoyproxy/envoy/pull/37684

https://github.com/envoyproxy/gateway/issues/4756

Release Notes: No (until the implementation is done)


api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,37 @@ type RateLimitRule struct {
...
+
+// RateLimitUsageSpecifier specifies the attributes within the request context from which
+// the Envoy retrieves the usage number to reduce the rate limit counters.
+//
+// By default, Envoy looks up the usage number from the `envoy.ratelimit.hits_addend` filter metadata.
+// If there's no such metadata or the number stored in the metadata is invalid, it will use the default
+// usage number of 1.
+//
+// This default behavior can be overridden by specifying one of the fields in this RateLimitUsageSpecifier.
+type RateLimitUsageSpecifier struct {
```
mathetake, line 115: this is basically https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-ratelimit-hitsaddend

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,37 @@ type RateLimitRule struct {
...
 	// the selected requests have reached the limit.
 	Limit RateLimitValue `json:"limit"`
+	// UsageSpecifier specifies the "usage" of each request that matches the rule.
+	// The usage is used to reduce the rate limit counters for the matching requests.
+	// The reduction happens after the request stream is complete, so the rate limit
+	// won't be enforced for the current request, but for the subsequent requests.
+	//
+	// This is optional and if not specified, the rate limit counters are not reduced.
+	//
+	// Currently, this is only supported for HTTP Global Rate Limits.
```
zirain, line 101: can you add cel validation for this?

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,37 @@ type RateLimitRule struct {
 	// 429 HTTP status code is sent back to the client when
 	// the selected requests have reached the limit.
 	Limit RateLimitValue `json:"limit"`
+	// UsageSpecifier specifies the "usage" of each request that matches the rule.
+	// The usage is used to reduce the rate limit counters for the matching requests.
+	// The reduction happens after the request stream is complete, so the rate limit
+	// won't be enforced for the current request, but for the subsequent requests.
+	//
+	// This is optional and if not specified, the rate limit counters are not reduced.
```
zhaohuabing, line 99: Not sure if I understand this correctly, but the default behaivor of "usage of each request" is one? We may need to rephrase this comment to make it more understandable.

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,37 @@ type RateLimitRule struct {
...
+
+// RateLimitUsageSpecifier specifies the attributes within the request context from which
+// the Envoy retrieves the usage number to reduce the rate limit counters.
+//
+// By default, Envoy looks up the usage number from the `envoy.ratelimit.hits_addend` filter metadata.
+// If there's no such metadata or the number stored in the metadata is invalid, it will use the default
+// usage number of 1.
+//
+// This default behavior can be overridden by specifying one of the fields in this RateLimitUsageSpecifier.
+type RateLimitUsageSpecifier struct {
```
zhaohuabing, line 115: Also a CEL validation: one and only one of `number` or `format` must be specified.

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,37 @@ type RateLimitRule struct {
...
+	// Number specifies the fixed usage number to reduce the rate limit counters.
+	//
+	// +optional
+	Number *uint64 `json:"number,omitempty"`
+	// Format specifies the format of the usage number. See the Envoy documentation for the supported formats:
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-ratelimit-hitsaddend
```
yuzisun, line 121: It is not clear to me from the envoy doc what's the supported substitution format string

mathetake:
feel like maybe calling it "weight" instead of usage maybe be better? idk

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,51 @@ type RateLimitRule struct {
...
+// By default, Envoy looks up the usage number from the `envoy.ratelimit.hits_addend` filter metadata.
+// If there's no such metadata or the number stored in the metadata is invalid, it will use the default
+// usage number of 1.
+//
+// This default behavior can be overridden by specifying one of the fields in this RateLimitUsage.
+type RateLimitUsage struct {
+	// Number specifies the fixed usage number to reduce the rate limit counters.
+	//
+	// +optional
+	// +notImplementedHide
```
mathetake, line 131: hidden until https://github.com/envoyproxy/ratelimit/pull/802

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +91,51 @@ type RateLimitRule struct {
...
+	Number *uint64 `json:"number,omitempty"`
+	// Format specifies the format of the usage number. See the Envoy documentation for the supported formats:
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-ratelimit-hitsaddend
+	//
+	// +optional
+	// +notImplementedHide
```
mathetake, line 137: hidden until https://github.com/envoyproxy/ratelimit/pull/802

mathetake:
will add the cel validation this afternoon (in japan)

mathetake:
maybe simply Usage->HitAddends would be consistent

api/v1alpha1/ratelimit_types.go
mathetake: done

api/v1alpha1/ratelimit_types.go
mathetake: done

api/v1alpha1/ratelimit_types.go
mathetake: https://github.com/envoyproxy/gateway/pull/4957/commits/efc5b8790add581d4e68ef3cbb49a12124d1e82c

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,62 @@ type RateLimitRule struct {
...
+	// Since the reduction happens after the request stream is complete, the rate limit
+	// won't be enforced for the current request, but for the subsequent matching requests.
+	//
+	// This is optional and if not specified, the rate limit counters are not reduced
+	// on the response path.
+	//
+	// Currently, this is only supported for HTTP Global Rate Limits.
+	//
+	// +optional
+	ResponseHitsAddend *RateLimitHitsAddend `json:"responseHitsAddend,omitempty"`
```
zhaohuabing, line 118: I guess RequestHitsAddend and ResponseHitsAddend can't coexist?

RequestHitsAddend and  ResponseHitsAddend should also be hidden?

mathetake: Well it can coexist

mathetake: cc @zirain has more contexts on how to implmente this

zirain: it could be co-exist, e.g. count 1 on request, and the (real usage -1) on response, this will give us better control of limitation.

mathetake:
@zirain @zhaohuabing ptal - I removed the implementation-hide as https://github.com/envoyproxy/ratelimit/pull/802 got merged

mathetake:
i will do the implementation in a subsequent PR

zhaohuabing:
> @zirain @zhaohuabing ptal - I removed the implementation-hide as [envoyproxy/ratelimit#802](https://github.com/envoyproxy/ratelimit/pull/802) got merged

@mathetake the implementation is for EG API, we normally hide the API at the API PR and unhide it at the implementation PR.

zhaohuabing (dismissed):
LGTM thanks!
Prefer to hide the API in the docs as it's not implemented yet.

mathetake:
ah ok - will do

mathetake:
@zhaohuabing hid the APIs!

zirain (commented):
LGTM, hold for a while as @arkodg may want to take a look.

zhaohuabing (dismissed):
LGTM defer to @arkodg

mathetake:
👍

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,64 @@ type RateLimitRule struct {
 	// 429 HTTP status code is sent back to the client when
 	// the selected requests have reached the limit.
 	Limit RateLimitValue `json:"limit"`
+	// RequestHitsAddend specifies the number to reduce the rate limit counters
```
arkodg, line 95: thanks for introducing this @mathetake !

thinking out loud, and trying to describe user intent 
until now, all we had was "I want to ratelimit at X requests per time unit", now we also have "I want to ratelimit requests based on a total count X, with each response adding to that count by a value (from a response header)"
Im unsure if `cost` is the best word to use here, `hitsAddend` is not intuitive enough imo, another option is `count` e.g. `countPerRequest`

```
costPerRequest: <val>.  // defaults to 1
costPerResonse:             // defaults to 0
  from:
    type: ResponseHeader / Header
    .......
```

cc @envoyproxy/gateway-reviewers @envoyproxy/gateway-maintainers

mathetake: > Im unsure if cost is the best word to use here, hitsAddend is not intuitive enough imo, another option is count e.g. countPerRequest or perRequestCount or simply requestCount

yeah agree not that intuitive, and costPer* sounds good to me

mathetake: btw, our primary usecase is not getting the cost from response header but from a dynamic metadata - so having the explicit from:type:.. won't be much helpful and I would rather expose the opaque "format" which supports both dynamic metadata and headers

mathetake:
btw the per-descriptor hits_addend only supported for per-route ratelimit policy config via per_filter_config either via virtual or route config according to #37684 cc @wbpcode 

is that an issue for EG?

api/v1alpha1/ratelimit_types.go
arkodg: can we map to format internally, and somehow use a user facing construct instead ?

arkodg:
> btw currently the per-descriptor hits_addend is only supported for per-route ratelimit policy config either via virtual or route config according to [envoyproxy/envoy#37684](https://github.com/envoyproxy/envoy/pull/37684) cc @wbpcode
> 
> is that an issue for EG?
> 
> edited: nvm, i think it's not an issue

yeah should be fine https://github.com/envoyproxy/gateway/blob/10a31f12f618a4781e6568196e923884067e8c23/internal/xds/translator/ratelimit.go#L141

api/v1alpha1/ratelimit_types.go
mathetake: gotcha - will work on it as well as renaming to costPer* stuff tomorrow (in japan 😉)

mathetake: done https://github.com/envoyproxy/gateway/pull/4957/commits/c0694f7c4d846079c037784c67111fcb25048e33

mathetake:
@arkodg PTAL

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,80 @@ type RateLimitRule struct {
...
+
+// RateLimitCostType specifies the source of the rate limit cost.
+// Valid RateLimitCostType values are "Number" and "DynamicMetadata".
+//
+// +kubebuilder:validation:Enum=Number;DynamicMetadata
+type RateLimitCostType string
+
+const (
+	// RateLimitCostTypeNumber specifies the rate limit cost to be a fixed number.
+	RateLimitCostTypeNumber RateLimitCostType = "Number"
```
arkodg, line 152: can we leave out `Number` for now from the API ? is there a use case for it ?
for e.g. if I want to ratelimit at 100 counts/sec, and each requests costs me 2, I can easily update the limit to 50 counts/sec

arkodg: oh is number used here, so we can set request rate limit cost to `0` ?

mathetake: yes

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,80 @@ type RateLimitRule struct {
...
+	// RateLimitCostTypeDynamicMetadata specifies the rate limit cost to be retrieved from the dynamic metadata.
+	RateLimitCostTypeDynamicMetadata RateLimitCostType = "DynamicMetadata"
```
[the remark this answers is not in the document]
mathetake: @arkodg i think you missed what it can be implemented in the upstream "Format" API of hits_addend https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit

It can be used for anything that is supported by formatter. It can also be get requests from Headers. What you suggested before was to abstract the formatter hence I added `RateLimitCostTypeDynamicMetadata RateLimitCostType = "DynamicMetadata"`. You can add `RateLimitCostTypeResponseHeader RateLimitCostType = "ResponseHeader"` later which is implemented `%REQP_...` 
image

api/v1alpha1/ratelimit_types.go
mathetake: I explicitly added that comment about 0 can be used to do "just checking"

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,80 @@ type RateLimitRule struct {
...
+	// TODO: add headers, etc. Anything that can be represented in "Format" can be added here.
+	// 	https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format
+)
+
+// RateLimitCostDynamicMetadata specifies the filter metadata to retrieve the usage number from.
+type RateLimitCostDynamicMetadata struct {
+	// Namespace is the namespace of the dynamic metadata.
+	//
+	// +kubebuilder:validation:Required
+	Namespace string `json:"filterNamespace"`
```
arkodg, line 164:
```suggestion
	Namespace string `json:"namespace"`
```

mathetake, line 164:
```suggestion
	Namespace string `json:"namespace"`
```

api/v1alpha1/ratelimit_types.go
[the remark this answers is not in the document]
mathetake: discussed offline :D

zirain (approved):

api/v1alpha1/ratelimit_types.go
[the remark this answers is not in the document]
arkodg: yeah thanks for clarifying this, looks like response header is not an option for AI Gateway, and doesnt need to be added right away

arkodg:
thanks for the multiple iteration @mathetake, high level, this PR looks good ! will bring this PR up in the tomorrow's community meeting to discuss minor naming options
1. using `cost` or `count` 
2. `perRequestCount` or `countPerRequest`
3. `metadata` vs `dynamicMetadata`

arkodg:
> thanks for the multiple iteration @mathetake, high level, this PR looks good ! will bring this PR up in the tomorrow's community meeting to discuss minor naming options
> 
> 1. using `cost` or `count`
> 2. `perRequestCount` or `countPerRequest`
> 3. `metadata` vs `dynamicMetadata`

@guydc @zhaohuabing and I discussed this in the community meeting today, the easiest one to find consensus on is 3. `metadata` is preferred over `dynamicMetadata`

reg 1 and 2., we have a new contender - `increment` - @guydc prefers `perRequestIncrement`, @zhaohuabing prefers `costPerRequest` , I prefer `count` or `increment` over `cost`

mathetake:
I would give +1 to `costPerRequest / perRequestCost` vs increment or count

zhaohuabing:
As a non-native speaker, `costPerRequest` sounds more intuitive to me.

mathetake:
yeah i would also prefer costPerRequest over perRequestCost

missBerg:
Just to add another idea into the mix.... what about:

```
cost:
  request:
      source: metadata
      metadata:
        namespace:
        key:
  response:
        source: metadata
      metadata:
        namespace:
        key:
```

then we get cleaner API field names, and we put all the cost configurations together.

arkodg:
> Just to add another idea into the mix.... what about:
> 
> ```
> cost:
>   request:
>       source: metadata
>       metadata:
>         namespace:
>         key:
>   response:
>         source: metadata
>       metadata:
>         namespace:
>         key:
> ```
> 
> then we get cleaner API field names, and we put all the cost configurations together.

+1 wfm

arkodg (dismissed):
LGTM thanks !

mathetake:
is this flak or what

api/v1alpha1/ratelimit_types.go
```diff
@@ -62,6 +62,7 @@ type LocalRateLimit struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:XValidation:rule="self.all(foo, !has(foo.cost.response))", message="response cost is not supported for Local Rate Limits"
```
zirain, line 65: @mathetake is this the cause of failed tests?

mathetake: huh

mathetake:
sorry the e2e tests logs are hard to grok ... will take a look

mathetake:
@zirain passed now

zirain (approved):

api/v1alpha1/ratelimit_types.go
```diff
@@ -91,6 +92,91 @@ type RateLimitRule struct {
 	// 429 HTTP status code is sent back to the client when
 	// the selected requests have reached the limit.
 	Limit RateLimitValue `json:"limit"`
```
yuzisun, line 94: It seems like it is still required to specify the number of requests limit, how we can handle the case when we only care of the cost not the number of the request limit?

arkodg (approved):

api/v1alpha1/ratelimit_types.go
arkodg: also with this new `cost` field, `limit.requests` doesnt quite fit anymore, something like `limit.total` feels better and can be an alias to `limit.requests` in the future to improve defining intent

Issue #4882, opened 2024-12-09
Define a default memory limit for shutdown-manager


ryanhristovski:
*Description*:
Currently the shutdown-manager sidecar does not have a default memory limit set, this can lead to uncontrolled memory usage, potentially causing node instability or OOMKilled errors. It's best practice to set the memory request equal to the memory limit to ensure consistent resource allocation and avoid unexpected scheduling issues.

I haven't yet benchmarked shutdown-managers memory usage to have a good idea of what to set the request/limit to if anyone has any suggestions.

arkodg:
requests == limit sgtm, wdyt @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

zhaohuabing:
+1 sgtm. The memory usage of the shutdown-manager shouldn't increase over time.

PR #4960, opened 2024-12-20
feat(default-memory-limits): set default memory limits 

ryanhristovski:
Fixes https://github.com/envoyproxy/gateway/issues/4882

- sets default memory limits for shutdown manager to 32Mi and default containers to 512Mi ([== to limit](https://github.com/envoyproxy/gateway/issues/4882#issuecomment-2529629806))
- applied [calculated max heap size](https://github.com/envoyproxy/gateway/blob/2a10d474fdfe2f3a9689266355fa6d89a042daac/internal/infrastructure/kubernetes/proxy/resource.go#L356-L369) envoy config to test data for new limit (512Mi)

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/testdata/daemonsets/override-labels-and-annotations.yaml
+++ b/internal/infrastructure/kubernetes/proxy/testdata/daemonsets/override-labels-and-annotations.yaml
@@ -215,6 +215,21 @@ spec:
               typed_config:
                 "@type": type.googleapis.com/envoy.extensions.resource_monitors.downstream_connections.v3.DownstreamConnectionsConfig
                 max_active_downstream_connections: 50000
+            - name: "envoy.resource_monitors.fixed_heap"
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.resource_monitors.fixed_heap.v3.FixedHeapConfig
+                max_heap_size_bytes: 429496729
+            actions:
+            - name: "envoy.overload_actions.shrink_heap"
+              triggers:
+              - name: "envoy.resource_monitors.fixed_heap"
+                threshold:
+                  value: 0.95
+            - name: "envoy.overload_actions.stop_accepting_requests"
+              triggers:
+              - name: "envoy.resource_monitors.fixed_heap"
+                threshold:
+                  value: 0.98
         - --log-level warn
         - --cpuset-threads
         - --drain-strategy immediate
@@ -255,6 +270,8 @@ spec:
           successThreshold: 1
           timeoutSeconds: 1
         resources:
+          limits:
+            memory: 512Mi
           requests:
             cpu: 100m
             memory: 512Mi
@@ -331,6 +348,8 @@ spec:
           successThreshold: 1
           timeoutSeconds: 1
         resources:
+          limits:
+            memory: 32Mi
           requests:
             cpu: 10m
             memory: 32Mi
--- a/api/v1alpha1/envoyproxy_helpers.go
+++ b/api/v1alpha1/envoyproxy_helpers.go
@@ -155,6 +155,9 @@ func DefaultShutdownManagerContainerResourceRequirements() *corev1.ResourceRequi
 			corev1.ResourceCPU:    resource.MustParse(DefaultShutdownManagerCPUResourceRequests),
 			corev1.ResourceMemory: resource.MustParse(DefaultShutdownManagerMemoryResourceRequests),
 		},
+		Limits: corev1.ResourceList{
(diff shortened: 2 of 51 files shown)


internal/infrastructure/kubernetes/proxy/testdata/daemonsets/override-labels-and-annotations.yaml
```diff
@@ -215,6 +215,21 @@ spec:
               typed_config:
                 "@type": type.googleapis.com/envoy.extensions.resource_monitors.downstream_connections.v3.DownstreamConnectionsConfig
                 max_active_downstream_connections: 50000
+            - name: "envoy.resource_monitors.fixed_heap"
```
zirain, line 218: why these changes happen?

ryanhristovski: It's set based on the memory limit by default, since memory limits weren't defined before they weren't being applied

Code here https://github.com/envoyproxy/gateway/blob/2a10d474fdfe2f3a9689266355fa6d89a042daac/internal/infrastructure/kubernetes/proxy/resource.go#L356-L369

zirain: thanks for clarify

zirain (approved):

cnvergence (approved):
LGTM

zirain:
looks like 32 cause OOM on master.

```
shutdown-manager:
    Container ID:    containerd://e43b953745ffdc5c2e86cf97c061d9e3560b7d800b15b625ca27d4ba315f9ece
    Image:           docker.io/envoyproxy/gateway-dev:latest
    Image ID:        docker.io/envoyproxy/gateway-dev@sha256:cb9737c83359588d3c10683a90ad835ae62696fb3ef8df20383425048aba1133
    Port:            <none>
    Host Port:       <none>
    SeccompProfile:  RuntimeDefault
    Command:
      envoy-gateway
    Args:
      envoy
      shutdown-manager
      --ready-timeout=15s
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Sat, 28 Dec 2024 14:15:42 +0800
      Finished:     Sat, 28 Dec 2024 14:15:42 +0800
    Ready:          False
    Restart Count:  5
    Limits:
      memory:  32Mi
    Requests:
      cpu:      10m
      memory:   32Mi
    Liveness:   http-get http://:19002/healthz delay=0s timeout=1s period=10s #success=1 #failure=3
    Readiness:  http-get http://:19002/healthz delay=0s timeout=1s period=10s #success=1 #failure=3
    Startup:    http-get http://:19002/healthz delay=0s timeout=1s period=10s #success=1 #failure=30
    Environment:
      ENVOY_GATEWAY_NAMESPACE:  envoy-gateway-system (v1:metadata.namespace)
      ENVOY_POD_NAME:           envoy-gateway-conformance-infra-all-namespaces-302def45-6brqfc7 (v1:metadata.name)
    Mounts:                     <none>
```

ryanhristovski:
@zirain hmm, it's generally not best practice to have memory request & limits not equal to each other - but given that this is a shutdown manager it _could_ make sense to have separate values.

What do you think:
1. set memory request to 32Mi and memory limit to something like 128Mi
or
2. set both request and limit to 128Mi

I didnt have time to fully benchmark it's resource use but it seems to sit idle around 25Mi and during shutdown (in my environment) it can hit around 70Mi.

I'm leaning #1 due to the nature of the shutdown manager - but this could cause resource issues for users in the future.

zirain:
can you take a look at #4978?

zirain:
@ryanhristovski this block CI, I didn't figure out why(test passed on my local).

let's revert it first, please resubmit later if you still want this, sorry for this.

sanposhiho:
my 2p for the future:
1. Memory consumption depends on Go's GC cycle. In short, if CPU is busy for other stuff, Go runtime cannot take time for GC and hence memory consumption could go up. There's some env parameters (GOMEMLIMIT, GOGC) that can adjust the behaviour though, AFAIK, there's nothing to define a _hard_ limit of memory.
2. CPU `10m` looks too small, and also it doesn't have any limit. No CPU limit isn't a bad thing (rather I'm no-cpu-limit lover), but you have to put CPU limit when benchmarking to measure memory consumption for this kind of change. The reason is (1); obviously, there'd be a difference in memory consumption when your shutdown manager can use 10 CPU vs when your shutdown manager can use only 10m CPU.

PR #4983, opened 2024-12-30
feat(infra): Add rateLimitHpa support in EnvoyGateway API

### LINKED ISSUE #4930: Add HPA Support for the Rate Limit Service
*Description*:
>Describe the issue.

Add support for rateLimitHPA in the EnvoyGateway API https://gateway.envoyproxy.io/docs/api/extension_types/#envoygatewaykubernetesprovider similar to what exists for EnvoyProxy https://gateway.envoyproxy.io/docs/api/extension_types/#envoyproxykubernetesprovider

[optional *Relevant Links*:]

keithfz:
**What type of PR is this?**

Adds HPA support for rate limit deployment under the EnvoyGateway API

Allows autoscaling of the rate limit deployment similar to what is done today for the envoy proxy deployment.

Fixes #4930

Release Notes: Yes

### DIFF
--- a/api/v1alpha1/envoygateway_types.go
+++ b/api/v1alpha1/envoygateway_types.go
@@ -202,6 +202,12 @@ type EnvoyGatewayKubernetesProvider struct {
 	// +optional
 	RateLimitDeployment *KubernetesDeploymentSpec `json:"rateLimitDeployment,omitempty"`
 
+	// RateLimitHpa defines the Horizontal Pod Autoscaler settings for Envoy ratelimit Deployment.
+	// If the HPA is set, Replicas field from RateLimitDeployment will be ignored.
+	//
+	// +optional
+	RateLimitHpa *KubernetesHorizontalPodAutoscalerSpec `json:"rateLimitHpa,omitempty"`
+
 	// Watch holds configuration of which input resources should be watched and reconciled.
 	// +optional
 	Watch *KubernetesWatchMode `json:"watch,omitempty"`
--- a/internal/infrastructure/kubernetes/ratelimit/resource_provider.go
+++ b/internal/infrastructure/kubernetes/ratelimit/resource_provider.go
@@ -42,6 +42,7 @@ type ResourceRender struct {
 
 	rateLimit           *egv1a1.RateLimit
 	rateLimitDeployment *egv1a1.KubernetesDeploymentSpec
+	rateLimitHpa        *egv1a1.KubernetesHorizontalPodAutoscalerSpec
 
 	// ownerReferenceUID store the uid of its owner reference.
 	ownerReferenceUID map[string]types.UID
@@ -53,6 +54,7 @@ func NewResourceRender(ns string, gateway *egv1a1.EnvoyGateway, ownerReferenceUI
 		Namespace:           ns,
 		rateLimit:           gateway.RateLimit,
 		rateLimitDeployment: gateway.GetEnvoyGatewayProvider().GetEnvoyGatewayKubeProvider().RateLimitDeployment,
+		rateLimitHpa:        gateway.GetEnvoyGatewayProvider().GetEnvoyGatewayKubeProvider().RateLimitHpa,
 		ownerReferenceUID:   ownerReferenceUID,
 	}
 }
@@ -288,6 +290,11 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		}
 	}
 
+	// omit the deployment replicas if HPA is being set
+	if r.rateLimitHpa != nil {
+		deployment.Spec.Replicas = nil
+	}
+
 	// apply merge patch to deployment
 	var err error
 	if deployment, err = r.rateLimitDeployment.ApplyMergePatch(deployment); err != nil {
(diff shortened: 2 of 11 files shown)


internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -288,6 +290,11 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 		}
 	}
 
+	// omit the deployment replicas if HPA is being set
+	if r.rateLimitHpa != nil {
```
arkodg, line 294: imo we shouldnt do this, its bad practice to set replicas and hpa, but we shouldnt override a user setting, and instead recommend a user to leave `replicas` unset when using `hpa`

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
```diff
@@ -309,11 +316,50 @@ func (r *ResourceRender) DaemonSet() (*appsv1.DaemonSet, error) {
...
 // HorizontalPodAutoscalerSpec returns the `HorizontalPodAutoscaler` sets spec.
 func (r *ResourceRender) HorizontalPodAutoscalerSpec() (*egv1a1.KubernetesHorizontalPodAutoscalerSpec, error) {
-	return nil, nil
+	return r.rateLimitHpa, nil
 }
 
 func (r *ResourceRender) HorizontalPodAutoscaler() (*autoscalingv2.HorizontalPodAutoscaler, error) {
-	return nil, nil
+	hpaConfig, err := r.HorizontalPodAutoscalerSpec()
+	if hpaConfig == nil {
```
arkodg, line 324: isnt `nil` allowed and the default ? should this be checking `err != nil` instead ?

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
arkodg: oh nvm youre following https://github.com/envoyproxy/gateway/blob/dd1c66fb024e1ce76f59fe1480900307e9ae480d/internal/infrastructure/kubernetes/proxy/resource_provider.go#L328

arkodg: i wonder why that is needed
cc @ardikabs

arkodg:
thanks @keithfz, this PR looks great ! added some minor non blocking comments

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
keithfz: yeah, honestly I kind of agree, but I just went for consistency between this and the `EnvoyProxy` API. If we choose _not_ to override a user setting here, we should probably have similar behavior in the `EnvoyProxy` HPA case as well.

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
keithfz: Yeah, `nil` is allowed, but if we only check `err != nil` then we miss the case of exiting the function when  `hpaConfig` is `nil` (and `err` being `nil`). This attempts to catch both -- if there is an error, `hpaConfig` will still be nil and we propagate the error message.

But if there is a preferred stylistic way to do it, happy to change that :)

internal/infrastructure/kubernetes/ratelimit/resource_provider.go
arkodg: yeah, lets wait for ardikabs' response, else keep it as is for now

ardikabs: It primarily prevents unwanted values from being set during HPA configuration, with the Envoy Gateway proactively ensuring such scenarios are avoided. This condition is already documented in the EG API Reference, highlighting when Replicas and HPA are being set. 

https://github.com/envoyproxy/gateway/blob/10a31f12f618a4781e6568196e923884067e8c23/api/v1alpha1/envoyproxy_types.go#L312-L316.

Historically, this behavior was implemented because EG previously set default replicas. As a result, even if users left the field empty (as recommended when using HPA), a default value of 1 was still applied.

But after this PR #2816, I think we can leave it, wdyt @arkodg?

arkodg: ah thanks for adding clarity here @ardikabs, this extra step can be cleaned up in a follow up commit

arkodg (approved):
LGTM thanks !

api/v1alpha1/envoygateway_types.go
```diff
@@ -202,6 +202,12 @@ type EnvoyGatewayKubernetesProvider struct {
 	// +optional
 	RateLimitDeployment *KubernetesDeploymentSpec `json:"rateLimitDeployment,omitempty"`
 
+	// RateLimitHpa defines the Horizontal Pod Autoscaler settings for Envoy ratelimit Deployment.
+	// If the HPA is set, Replicas field from RateLimitDeployment will be ignored.
+	//
+	// +optional
+	RateLimitHpa *KubernetesHorizontalPodAutoscalerSpec `json:"rateLimitHpa,omitempty"`
```
zirain, line 209: this' fine now, if we bring more fields in the future, pefer to something like:
```yaml
ratelimit:
  deployment:
     ...
  hpa:
    ...
```

zirain (approved):

api/v1alpha1/envoygateway_types.go
keithfz: would we want to consider the same for other APIs in the future as well? e.g. EnvoyProxy: https://github.com/envoyproxy/gateway/blob/7952936666f706f7b04e578036c89f875f5e3b14/api/v1alpha1/envoyproxy_types.go#L291-L328

arkodg: Both are making a good point, the extra hierarchy hides the component specific details, but the current style is 'envoyHpa' which is probably why the same was done for 'rateLimitDeployment'

PR #4986, opened 2024-12-31
feat: implement API Key Auth security policy

### LINKED ISSUE #2630: Support API Key Auth
> @zhaohuabing can we reuse the same Envoy Filter for API key auth ? 

The problem is, unlike HTTP Basic Auth, API key auth is much more flexible, the key can be in the header, path, or even body. I think we will need a dedicated envoy filter in Envoy if we want to support API key auth, we can follow the open API specs: https://swagger.io/docs/specification/authentication/api-keys/#:~:text=API%20keys%20are%20supposed%20to,mechanisms%20such%20as%20HTTPS%2FSSL.

_Originally posted by @zhaohuabing in https://github.com/envoyproxy/gateway/issues/1660#issuecomment-1814974031_

sanposhiho:
**What type of PR is this?**

implement API Key Auth security policy.

Fixes #2630

Release Notes: Yes


api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,69 @@
...
+	// +optional
+	KeySources []*KeySource `json:"keySources,omitempty"`
+
+	// AllowedClients is a list of clients that are allowed to access the route or vhost.
+	// The clients listed here should be subset of the clients listed in the `Credentials` to provide authorization control
+	// after the authentication is successful. If the list is empty, then all authenticated clients
+	// are allowed. This provides very limited but simple authorization.
+	//
+	// +optional
+	AllowedClients []string `json:"allowedClients,omitempty"`
```
zhaohuabing, line 42: Are there any use case where an API key is authenticated but not allowed to access the targeted Route/Gateway?

sanposhiho: The use case is:
1. the gateway has API Key Auth configured globally. e.g., `Credentials` has `admin: key1` and `user: key2`.
2. the route also has per-route API Key Auth, and you can define this route is only for `admin` with `AllowedClients`.

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,69 @@
...
+	Credentials gwapiv1.SecretObjectReference `json:"credentials,omitempty"`
+
+	// KeySources is where to fetch the key from the coming request.
+	// The value from the first source that has a key will be used.
+	// If this is attached to a HTTPRoute while the other SecurityPolicies with APIKeyAuth is attached to a Gateway,
+	// this key sources takes precedence.
+	// This field is optional, but it doesn't make sense that any attached APIKeyAuth for the route/gateway doesn't have key sources at all.
+	//
+	// +optional
+	KeySources []*KeySource `json:"keySources,omitempty"`
```
zhaohuabing, line 34: The `KeySources`  should not be optional.

The SecurityPolicy at HTTPRoute should override the SecurityPolicy at the Gateway as a whole.  The SecurityPolicy at the HTTPRoute should not use  the `KeySources` from the parent Gateway.

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,69 @@
...
+	// Credentials is the Kubernetes secret which contains the API keys.
+	// This is an Opaque secret.
+	// Each API key is stored in the key representing the client id,
+	// which can be used in AllowedClients to authorize the client in a simple way.
+	// If this is attached to a HTTPRoute while the other SecurityPolicies with APIKeyAuth is attached to a Gateway,
+	// this credentials takes precedence.
+	// This field is optional, but it doesn't make sense that any attached APIKeyAuth for the route/gateway doesn't have credentials at all.
+	//
+	// +optional
+	Credentials gwapiv1.SecretObjectReference `json:"credentials,omitempty"`
```
zhaohuabing, line 25: The `Credentials`  should be mandatory.

In EG, we don't "merge" SecurityPolicies at different levels. The SecurityPolicy at the HTTPRoute level should override the SecurityPolicy at the Gateway as a whole.  The SecurityPolicy at the HTTPRoute level should not use the `Credentials` from the parent Gateway.

api/v1alpha1/api_key_auth_types.go
zhaohuabing: In such case, this route should only has admin: key1 credential attached.

sanposhiho: I see, I missed that point. 
But, I'm not sure if it's the best interface for users that have to prepare secret per access group. Also, the same secret for each client would be spread across different secrets.

So, I still see the benefit of AllowedClients there; users can have only one secret as a source of credentials and they can select which key(s) is enabled on certain routes.

sanposhiho:
E2e test passed on my local and opened the PR for the review.

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -57,9 +57,28 @@ func validateSecurityPolicySpec(spec *egv1a1.SecurityPolicySpec) error {
 		errs = append(errs, err)
 	}
 
+	if err := ValidateAPIKeyAuth(spec.APIKeyAuth); err != nil {
```
sanposhiho, line 60: There's a bug that `ValidateSecurityPolicy` isn't called at all currently.
I sent another PR to patch that, https://github.com/envoyproxy/gateway/pull/4987.

api/v1alpha1/api_key_auth_types.go
zhaohuabing: > users can have only one secret as a source of keys and they can select which key(s) is enabled on certain routes with AllowedClients.

I think your argument kind of makes sense 🙂, but I’d prefer not to include the AllowedClients field in the API for the following reasons:

* **Increased Complexity**: Adding another field for users to configure complicates the API, making it less intuitive.
* **Potential Security Risk**: The shared secret may include API keys not authorized to access an HTTPRoute. This poses a security risk, as the owner of an HTTPRoute might inadvertently or intentionally add such keys to AllowedClients.


@envoyproxy/gateway-maintainers please weigh in.

sanposhiho: > Potential Security Risk

First of all, if a bad actor has an access to `HTTPRoute` or associated `SecurityPolicy`, they can just disable `APIKeyAuth` to allow access for unintended clients.

Also, even if we assume they _somehow_ have to keep `APIKeyAuth` but try to broaden the access of the route, both (with vs without `AllowedClients`) contain essentially the same level of risk.

Let's say there're two keys; one for `normal-users` and another for `admin-users`, and a bad owner is trying to allow access for the admin route to broader users.

- With AllowedClients: The owner of HTTPRoute can indicate the client id for `normal-users` in `AllowedClients`
  - The owner has to know which client-id is used for `normal-users`. (a read access for secret or other HTTPRoute)
- Without AllowedClients: The owner of HTTPRoute can indicate a API Key secret for `normal-users` in `Credentials`.
  - The owner has to know which secret is used for `normal-users` (a read access for secret or other HTTPRoute)

So,
> Increased Complexity

I believe whether we want to include `AllowedClients` or not only depends on this point; whether we believe it's useful for users, or not that useful, rather just making the API complex unnecessarily.

zhaohuabing: I have some reservation about the idea of reusing a shared secret globally. To align more closely with the EG policy design principles, I believe a better would be to create a SP targeting a Gateway and refrence that secret in that SP, and we can override it with a SP targeting an HTTPRoute for any more specific HTTPRoutes, sch as an "admin" route.

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,59 @@
...
+type APIKeyAuth struct {
+	// Credentials is the Kubernetes secret which contains the API keys.
+	// This is an Opaque secret.
+	// Each API key is stored in the key representing the client id,
+	// which can be used in AllowedClients to authorize the client in a simple way.
+	Credentials gwapiv1.SecretObjectReference `json:"credentials"`
+
+	// KeySources is where to fetch the key from the coming request.
+	// The value from the first source that has a key will be used.
+	KeySources []*KeySource `json:"keySources"`
```
arkodg, line 24: I like OpenAPI's definiton which uses `In`
https://swagger.io/docs/specification/v3_0/authentication/api-keys/
```
in:
  type: Header
  name: "x-api-id"
```

another existing design pattern

the `jwt` field uses `extractFrom` https://gateway.envoyproxy.io/docs/api/extension_types/#jwtprovider

api/v1alpha1/api_key_auth_types.go
arkodg: > First of all, if a bad actor has an access to HTTPRoute or associated SecurityPolicy, they can just disable APIKeyAuth to allow access for unintended clients

1. according to EG's design,  if the app dev service account can create a SecurityPolicy in the dev namespace where the HTTPRoute also lives, they should be able to override the auth, we are relying on users to set up kube RBAC based on their use case

2. my vote would be to make it look more like `basicAuth` https://gateway.envoyproxy.io/docs/api/extension_types/#basicauth
with one field `credentials` / `keyRefs` that can hold a list of secrets

3. we need to specify what the secret `key` and `value` should look like

api/v1alpha1/api_key_auth_types.go
sanposhiho: Well, generally speaking, I do prefer the current form in this kind of case at API design, especially for better future extensibility; it allows us to add a different structure later, e.g., let's say envoy supports a new key source fetching key from body (which Kong already supports), it might be like:

```go
type KeySource struct {
	Header *string 
	Query *string 
	Cookie *string

	Body *BodyKeySource
}

type BodyKeySource struct {
	BodyType string // yaml or json.
	Field string // where we get API Key from.
}
```

Also, `extractFrom` actually looks like very similar to `KeySource` here, are you suggesting renaming it from KeySources to ExtractFrom?

arkodg: yeah suggesting we pick a name thats already used in the API / or industry and also try and use a union discriminator i.e.
```
extractFrom:
  type: Header
  header: <val> 
```

e.g. https://gateway.envoyproxy.io/docs/api/extension_types/#ratelimitspec
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md

api/v1alpha1/api_key_auth_types.go
sanposhiho: Sorry, so what's your suggestion, whether removing `AllowedClients` vs not?

api/v1alpha1/api_key_auth_types.go
sanposhiho: Renamed to `extractFrom`.

api/v1alpha1/api_key_auth_types.go
zhaohuabing: > with one field credentials / keyRefs that can hold a list of secrets

@arkodg unlike BasicAuth, where the user credential always in the Authorization header, the APIKeyAuth needs users to specify where to get the API key from the request - it can be Authorization header, a custom header, or path parameter. So the `keySources` is required.

arkodg: yeah my vote is to rm it, I'll bring this API up in the community meeting tomorrow, to hear more thoughts from the community

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,59 @@
...
+
+const APIKeysSecretKey = "credentials"
+
+// APIKeyAuth defines the configuration for the API Key Authentication.
+type APIKeyAuth struct {
+	// Credentials is the Kubernetes secret which contains the API keys.
+	// This is an Opaque secret.
+	// Each API key is stored in the key representing the client id,
+	// which can be used in AllowedClients to authorize the client in a simple way.
+	Credentials gwapiv1.SecretObjectReference `json:"credentials"`
```
arkodg, line 20: prefer `CredentialRefs []gwapiv1.SecretObjectReference` we can start and only support a single entry in the list, but it helps us leave room for expansion

api/v1alpha1/api_key_auth_types.go
arkodg: @zhaohuabing and I discussed this in the community meeting today and we prefer if `AllowedClients` was removed from the initial API. In the future if we want to do API key based Authz, we could enhance then authorization API https://github.com/envoyproxy/gateway/blob/d23eab92f714c38fe8d11e0c37c68ffe85f12385/api/v1alpha1/authorization_types.go#L55

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,59 @@
...
+// Only one of header, query or cookie is supposed to be specified.
+//
+// Note: we intentionally don't add the validation for the only one of header, query or cookie is supposed to be specified with +kubebuilder:validation:XValidation:rule.
+// Instead, we add the validation in the controller reconciliation.
+// Technically we can define CEL, but the CEL estimated cost exceeds the threshold and it wouldn't be accepted.
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.header) || has(self.query) || has(self.cookie))",message="one of header, query or cookie must be specified"
+type ExtractFrom struct {
+	// Header is the name of the header to fetch the key from.
+	// This field is optional, but only one of header, query or cookie is supposed to be specified.
```
arkodg, line 45: needs a `Type` field - `Header`, `QueryParams`, `Cookie`

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,59 @@
...
+	//
+	// +optional
+	Header *string `json:"header,omitempty"`
+	// Query is the name of the query parameter to fetch the key from.
+	// This field is optional, but only one of header, query or cookie is supposed to be specified.
+	//
+	// +optional
+	Query *string `json:"query,omitempty"`
```
arkodg, line 53: prefer `QueryParams` here 
* envoy calls it `parameters` in some places and `queryParameters` in some places, 
* openAPI calls it query
https://swagger.io/docs/specification/v3_0/describing-parameters/#query-parameters
* gateway api uses `queryParams` https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRouteMatch

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const APIKeysSecretKey = "credentials"
+
+// APIKeyAuth defines the configuration for the API Key Authentication.
+type APIKeyAuth struct {
+	// Credentials is the Kubernetes secret which contains the API keys.
+	// This is an Opaque secret.
```
arkodg, line 17: we should highlight the format of the `Secret` here , its a map of key - value pairs, is the map key the `API Key` and the map value the `Client ID` or vice versa

api/v1alpha1/api_key_auth_types.go
sanposhiho: Why do you think so? I'm not sure if that's a common pattern of K8s APIs. Isn't that just a duplicate effort for users? because we can determine which to use by checking non-nil field. 

There're many examples just having `omitempty` fields for this kind of enum-ish structure:
- https://github.com/kubernetes/kubernetes/blob/8a1db6e8a5e255307e78e8f4df93a572c4bd72a7/pkg/apis/core/types.go#L58
- https://github.com/kubernetes/kubernetes/blob/8a1db6e8a5e255307e78e8f4df93a572c4bd72a7/pkg/apis/core/types.go#L232

within envoy gateway:
- https://github.com/envoyproxy/gateway/blob/e028254baf33858edc5b9e162993a105b6efc463/api/v1alpha1/backend_types.go#L50-L70

api/v1alpha1/api_key_auth_types.go
sanposhiho: Isn't the existing doc enough? (or is this explanation of mine unclear?

```go
 	// Each API key is stored in the key representing the client id,
	// which can be used in AllowedClients to authorize the client in a simple way.
```

api/v1alpha1/api_key_auth_types.go
sanposhiho: Sure, thanks for having a discussion for this!

api/v1alpha1/api_key_auth_types.go
arkodg: lets take a case that looks like
```
extractFrom:
  header: foo
  queryParam: bar
```

where should the API Gateway look for the API Key - in one if so which one, or both, if both, which has precedence ?

arkodg: sidenote: you can reduce the levels by using
```
extractFrom: Header
header: foo
```
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#summary uses VolumeSource as an example to describe an instance where `oneOf` semantics is implictly implemented

https://github.com/envoyproxy/gateway/blob/e028254baf33858edc5b9e162993a105b6efc463/api/v1alpha1/backend_types.go#L50 probably should have had a `Type` , only argument to not do it would be, an endpoint with `FQDN` and `IP` set is still valid

api/v1alpha1/api_key_auth_types.go
sanposhiho: > we can start and only support a single entry in the list

Any reason for this? It's not complicated to support the multiple secret refs from the first step.

arkodg: go for it, supporting multiple refs would be great ! didn't want to request another enhancement to this initial API

api/v1alpha1/api_key_auth_types.go
arkodg: sure, lets keep this for now, and can improve in the future

api/v1alpha1/api_key_auth_types.go
```diff
@@ -0,0 +1,51 @@
...
+	// Header is the name of the header to fetch the key from.
+	// This field is optional, but only one of header, queryParams or cookie is supposed to be specified.
+	//
+	// +optional
+	Header *string `json:"header,omitempty"`
+	// QueryParams is the name of the query parameter to fetch the key from.
+	// This field is optional, but only one of header, queryParams or cookie is supposed to be specified.
+	//
+	// +optional
+	QueryParams *string `json:"queryParams,omitempty"`
```
arkodg, line 45:
```suggestion
	QueryParam *string `json:"queryParam,omitempty"`
```

api/v1alpha1/api_key_auth_types.go
sanposhiho: Ok, I didn't know this just-`omitempty` pattern is considered problematic, needing [an improvement](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#summary). I'm convinced, will make a change like you suggested.

> only argument to not do it would be, an endpoint with FQDN and IP set is still valid

Actually, it's not considered valid right now by the way because of the CEL validation we have:

```go
// +kubebuilder:validation:XValidation:rule="(has(self.fqdn) || has(self.ip) || has(self.unix))",message="one of fqdn, ip or unix must be specified"
// +kubebuilder:validation:XValidation:rule="((has(self.fqdn) && !(has(self.ip) || has(self.unix))) || (has(self.ip) && !(has(self.fqdn) || has(self.unix))) || (has(self.unix) && !(has(self.ip) || has(self.fqdn))))",message="only one of fqdn, ip or unix can be specified"
```

arkodg: I reread this API definition and the extract from field is a list, which means that the API does support multiple key sources and first one wins, within the extract from struct if we give header priority over queryParam we could do without a Type like enum.
Cc @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

sanposhiho: > sidenote: you can reduce the levels by using

If we go with this, how do users specify multiple `extractFrom`s? Envoy allows setting multiple.

api/v1alpha1/api_key_auth_types.go
sanposhiho: oops

api/v1alpha1/api_key_auth_types.go
arkodg: sorry to add more confusion and churn, but also found https://github.com/envoyproxy/gateway/blob/542785400ed5b809285fe78894ffd195a0e54017/api/v1alpha1/jwt_types.go#L105 which keeps `extractFrom` as a scalar field, but has a list inside - `headers`, `params` , both these APIs are trying to achieve the same thing

sanposhiho: Sorry, looks like GitHub missed [this comment](https://github.com/envoyproxy/gateway/pull/4986#discussion_r1908107737) before I posted the last comment.

> within the extract from struct if we give header priority over queryParam we could do without a Type like enum.

Yeah, the envoy API allows multiple fields are set and has a priority Header > Query > Cookie for such case. But, that means, setting multiple fields is basically the mistake of users and they'd honor only one field from users' input. 
`Type` would help users preventing such a mistake, making it clear for users that it's supposed to have only one field filled like [the enhancement](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#summary) that you shared argued.

> but has a list inside - headers, params , both these APIs are trying to achieve the same thing

Yeah, but the envoy API doesn't accept multiple headers, params, etc on its own, hence we'd need a custom solution, if we want to support such, like translating this:

```yaml
extractFrom:
  - headers: ["header1", "header2"]
```

to this:

```yaml
        envoy.filters.http.api_key_auth:
          '@type': type.googleapis.com/envoy.extensions.filters.http.api_key_auth.v3.ApiKeyAuthPerRoute
          keySources:
          - header: header1
          - header: header2
```

Also, `JWTHeaderExtractor.ValuePrefix` cannot be supported since envoy API for APIKeyAuth doesn't have such feature. Do you think we still should do this?

arkodg: yes, suggesting exactly this so it maintains the look and feel of existing field, as you mentioned, we dont need an extra hierarchy for `Name` and `ValuePrefix`
```
# Envoy Gateway Security policy.
extractFrom:
  - headers: ["header1", "header2"]
to this:

# Envoy API
        envoy.filters.http.api_key_auth:
          '@type': type.googleapis.com/envoy.extensions.filters.http.api_key_auth.v3.ApiKeyAuthPerRoute
          keySources:
          - header: header1
          - header: header2
```

sanposhiho: > extractFrom doesnt need to be a list

In that case, how do we prioritize each entry then? Like, in the below example, which entry should we prioritize?

```yaml
extractFrom:
# should we prioritize one field over others...?
  headers: ["header1", "header2"]
  cookies: ["cookie1"]
  params: ["param1", "param2"]
```

Envoy API of `KeySource` has the priority Header > Query > Cookie, but as I said, it's for one `KeySource` entry of the `keySources` list that mistakenly has multiple fields filled. If we just followed such prioritization in `extractFrom`, users could never prioritize, for example, cookie1 over header1, which is possible if we go with the list like this:

```yaml
extractFrom:
# Envoy would look for the key in cookie1 and then in header1 if not found in the cookie. 
  - cookies: ["cookie1"]
  - headers: ["header1"]
  - ...
```

So, changing `extractFrom` to non-list makes it impossible. What do you think? Should we just ignore this use case and change `extractFrom` to non-list?

arkodg: I see, good point, lets go with the above style to account for this case ( I'm unsure how popular if is, but supporting it doesnt cost us much from a API readability or code complexity perspective)
```
extractFrom:
- cookies: ["cookie1"]
```

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -53,13 +58,28 @@ func validateSecurityPolicySpec(spec *egv1a1.SecurityPolicySpec) error {
...
-
 	return utilerrors.NewAggregate(errs)
 }
 
+func ValidateAPIKeyAuth(p *egv1a1.APIKeyAuth) error {
+	if p == nil {
+		return nil
+	}
+
+	for _, keySource := range p.ExtractFrom {
```
arkodg, line 72: this is a good case where this can be moved to CEL( no action needed here,  just wanted to highlight that)

arkodg (dismissed):
LGTM thanks !

arkodg:
ptal @zhaohuabing

api/v1alpha1/validation/securitypolicy_validate.go
arkodg: Ah thanks for clarifying

arkodg:
@sanposhiho looks like the e2e is failing consistently

sanposhiho:
Yeah, sorry for that. I'll look into it in this weekend.

internal/ir/xds.go
```diff
@@ -991,6 +993,47 @@ type BasicAuth struct {
 	Users PrivateBytes `json:"users,omitempty" yaml:"users,omitempty"`
 }
 
+// APIKeyAuth defines the schema for the API Key Authentication.
+//
+// +k8s:deepcopy-gen=true
+type APIKeyAuth struct {
+	// Name is a unique name for an APIKeyAuth configuration.
+	// The xds translator only generates one API key filter for each unique name.
+	Name string `json:"name" yaml:"name"`
```
zhaohuabing, line 1002: Nit: I think we don't need a Name field for APIKeyAuth ir, as only one APIKeyAuth filer in the filter chain.

zirain:
@sanposhiho can you rebase master and rerun `make -k gen-check`.

zhaohuabing (approved):
LGTM thanks!

arkodg (approved):

Issue #4858, opened 2024-12-06
Panic when adding healthcheck to gateway in standalone mode


hoyon:
*Description*:

When using  standalone mode, envoy gateway panics when applying a config which applies a `BackendTrafficPolicy` with a health check attached.

I've also tested this on macOS and there it causes a segfault rather than recovering from a panic.

I haven't been able to reproduce this issue when using Envoy Gateway in a local minikube cluster.

*Repro steps*:

`envoy-gateway server -c standalone.yaml`

standalone.yaml

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyGateway
gateway:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
provider:
  type: Custom
  custom:
    resource:
      type: File
      file:
        paths: ["/tmp/config.yaml"]
    infrastructure:
      type: Host
      host: {}
logging:
  level:
    default: info
extensionApis:
  enableBackend: true
```

config.yaml
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 8888
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: "gateway.envoyproxy.io"
          kind: Backend
          name: backend
      matches:
        - path:
            type: PathPrefix
            value: /
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: Backend
metadata:
  name: backend
spec:
  endpoints:
    - ip:
        address: 0.0.0.0
        port: 4000
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy
metadata:
  name: backend-policy
spec:
  targetRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: eg
# vvvvv all works if bit below here is removed
  healthCheck:
    active:
      type: HTTP
      http:
        path: "/"
        method: GET

```

*Environment*:

[...]

PR #4987, opened 2024-12-31
fix: validate SecurityPolicy on controller and `egctl translate`

sanposhiho:
**What type of PR is this?**

- `ValidateSecurityPolicy` exists, but isn't called like other validation functions (ValidateEnvoyProxy etc).
-  Fix the validation; Currently, all `SecurityPolicy`s without CORS or JWT are considered invalid and failed at the validation. 

Fixes #

Release Notes: No


internal/provider/kubernetes/controller.go
```diff
@@ -1067,6 +1067,11 @@ func (r *gatewayAPIReconciler) processSecurityPolicies(
 	}
 
 	for _, policy := range securityPolicies.Items {
+		if err := validation.ValidateSecurityPolicy(&policy); err != nil {
+			r.log.Error(err, "failed to process SecurityPolicy", "namespace", policy.Namespace, "name", policy.Name)
```
arkodg, line 1071: we dont want to run this here, we want to save this, run validation and attach the errors in the status (which happen in the next gateway api layer) and then flag that to the user

sanposhiho: You meant like this? 0e44d8258a369de382f0371fb0698a679fc2beb8

arkodg: yes, thanks !

internal/cmd/egctl/testdata/translate/out/invalid-securitypolicy.all.yaml
```diff
@@ -0,0 +1,115 @@
...
+  status:
+    ancestors:
+    - ancestorRef:
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: eg
+        namespace: envoy-gateway-system
+      conditions:
+      - lastTransitionTime: null
+        message: 'Invalid SecurityPolicy: policy is nil'
```
arkodg, line 111: curious why we are only seeing `policy is nil` here, shouldn't we also see `spec is nil` here ?

internal/gatewayapi/securitypolicy.go
```diff
@@ -147,6 +148,17 @@ func (t *Translator) ProcessSecurityPolicies(securityPolicies []*egv1a1.Security
 					continue
 				}
 
+				if err := validation.ValidateSecurityPolicy(policy); err != nil {
+					status.SetTranslationErrorForPolicyAncestors(&policy.Status,
+						parentGateways,
+						t.GatewayControllerName,
+						policy.Generation,
+						fmt.Sprintf("Invalid SecurityPolicy: %v", err),
```
arkodg, line 156: we should probably use `status.Error2ConditionMsg(err)` here as the rest of the file

internal/cmd/egctl/testdata/translate/out/invalid-securitypolicy.all.yaml
sanposhiho: Good catch. Looks like the existing implementation incorrectly overrides the validation error message. I fixed in:
https://github.com/envoyproxy/gateway/pull/4987/commits/25eeb9402285985740164236a84580c450d576be

arkodg (approved):
LGTM thanks !

zhaohuabing (dismissed):
LGTM.

IMO, the validation should be removed from the api package and moved into the gateway API translation package.  This can be addressed in a follow-up PR.

sanposhiho:
Another question around the validation is that, shouldn't we do it with a validation webhook instead of at the reconciliation? 
I mean, I know it's a tradeoff between "GW users can immediately/easily find the mistakes at applying" vs "admin has to maintain the webhook properly (e.g., autoscaling etc)" though.

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -53,10 +64,6 @@ func validateSecurityPolicySpec(spec *egv1a1.SecurityPolicySpec) error {
 		return utilerrors.NewAggregate(errs)
 	}
 
-	if err := ValidateJWTProvider(spec.JWT.Providers); err != nil {
```
sanposhiho, line 56: Yet another bug: It panics if spec.JWT is nil.

sanposhiho:
@arkodg @zhaohuabing Sorry I had missed the test was failing.
I just fixed them, please retake a look.

zhaohuabing:
> Another question around the validation is that, shouldn't we do it with a validation webhook instead of at the reconciliation? I mean, I know it's a tradeoff, for example, between "GW users can immediately/easily find the mistakes at applying" vs "admin has to maintain the webhook properly (e.g., autoscaling etc)" though.

I guess another reason probably is that Webhook can not be used for the host deployment mode. @arkodg should have more context on this.

zhaohuabing (approved):

shawnh2:
The PR looks good!

But it confuses me that why we have this SecurityPolicy validations here ? I can understand the existence of EnvoyGateway validations, and EnvoyProxy validations (which can be removed once the CEL expression is stable).

sanposhiho:
> But it confuses me that why we have this SecurityPolicy validations here ?

So, what would you propose instead?

arkodg:
outlining the steps EG performs - Receives Input (1) and Translates it (2)
a. We rely on OpenAPI validations to validate input in 1.  These are generated using kubebuilder tags and [CEL validations ](https://kubernetes.io/docs/reference/using-api/cel/)(which can represent complex validations) . Most implementations have moved their their webhook validation logic to CEL to simplify operations (out of process vs inline), so the config still gets rejected synchronously during apply.
b. For some cases where we cannot define CEL we use the Resource Status to flag any validations errors we catch during 2. e.g. targetRef of a policy specifying a non existant resource, which cannot be computed by CEL and requires more info.
c. EG's philosophy is we rely on these OpenAPI validations baked into the CRD which apply in 1. and we dont need to rewrite  these validations in 2. So a case of `spec` being `nil` is NA in Kubernetes because the Kube API server will reject such a config so it should never reach 2.
d. The issue however does exist for the File Provider case, where we read config from a file in 1. , the OpenAPI validations are not kicking in, this is being tracked with https://github.com/envoyproxy/gateway/issues/4858 and needs some more exploration

arkodg (approved):

PR #4990, opened 2025-01-02
docs: explain panic mode

zhaohuabing:
Explain the "panic mode" where failed endpoints exceeds 50% as users were asking why requests were sent to unhealth endpoints.

### DIFF
--- a/api/v1alpha1/healthcheck_types.go
+++ b/api/v1alpha1/healthcheck_types.go
@@ -9,6 +9,11 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 // HealthCheck configuration to decide which endpoints
 // are healthy and can be used for routing.
+//
+// Please note that Envoy load balancer may behave differently when lots of endpoints are unhealthy because of the "panic mode".
+// When the percentage of unhealthy endpoints exceeds 50%, Envoy will disregard health status and balance across all endpoints.
+// This is called "panic mode". It's designed to prevent a situation in which host failures cascade throughout the cluster
+// as load increases.
 type HealthCheck struct {
 	// Active health check configuration
 	// +optional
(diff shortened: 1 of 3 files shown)


zirain (approved):

api/v1alpha1/healthcheck_types.go
```diff
@@ -9,6 +9,11 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 // HealthCheck configuration to decide which endpoints
 // are healthy and can be used for routing.
+//
+// Please note that Envoy load balancer may behave differently when lots of endpoints are unhealthy because of the "panic mode".
```
arkodg, line 13: // **Note**: Once the overall health of the backendRef drops below 50% (e.g. a backendRef having 10 endpoints
// with more than 5 unhealthy endpoints), the health check is ignored for the remaining endpoints i.e. they are not
// removed from the load balancing pool. This prevents cascading failures and retry storms in the distributed system

zhaohuabing: > the health check is ignored for the remaining endpoints i.e. they are not
 removed from the load balancing pool.
 
The healtch check is not ignored - the failed endpoints are still failed, but the load balancer sends requests to all the endpoints, including both the failed and the health endpoints, once the percentage of the failed endpoints reach 50%.

api/v1alpha1/healthcheck_types.go
```diff
@@ -9,6 +9,11 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 // HealthCheck configuration to decide which endpoints
 // are healthy and can be used for routing.
+//
+// Note: Once the overall health of the backendRef drops below 50% (e.g. a backendRef having 10 endpoints
+// with more than 5 unhealthy endpoints), Envoy will disregard health status and balance across all endpoints.
```
arkodg, line 14:
```suggestion
// with more than 5 unhealthy endpoints), Envoy will disregard health status and balance across the remaining endpoints.
```

arkodg (approved):

zirain (approved):

PR #5002, opened 2025-01-03
fix: store one copy of HTTPRoute Extension Filters

### LINKED ISSUE #4989: Cannot reuse HTTPRouteFilter with urlRewrite for multiple HTTPRoute
*Description*:
When applying multiple `HTTPRoute` resources that reference the same `HTTPRouteFilter` with a `urlRewrite` configuration, an error occurs for all the `HTTPRoute` resources. The error message states:
Cannot configure multiple urlRewrite filters for a single HTTPRouteRule.

However, each `HTTPRoute` only applies one `urlRewrite`, so the error seems incorrect. This issue prevents reusing `HTTPRouteFilter` objects for multiple routes, which should be allowed.

*Repro steps*:
1. Follow the official installation guide as written here: https://gateway.envoyproxy.io/latest/tasks/quickstart/#installation
2. But instead of the regular `HTTPRoute`, apply these:
```yaml
---
kind: HTTPRouteFilter
metadata:
  name: route-filter
spec:
  urlRewrite:
    path:
      replaceRegexMatch:
        pattern: /
        substitution: /
      type: ReplaceRegexMatch

---
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
      - group: ""
        kind: Service
        name: backend
        port: 3000
        weight: 1
      filters:
      - extensionRef:
          group: gateway.envoyproxy.io
          kind: HTTPRouteFilter
          name: route-filter
        type: ExtensionRef
      matches:
        - path:
            type: PathPrefix
            value: /
---
kind: HTTPRoute
metadata:
  name: backend-2
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
      - group: ""
        kind: Service
        name: backend
        port: 3000
        weight: 1
      filters:
      - extensionRef:
          group: gateway.envoyproxy.io
          kind: HTTPRouteFilter
          name: route-filter
        type: ExtensionRef
      matches:
        - path:
            type: PathPrefix
            value: /
```
3. Observe the error message on all `HTTPRoute` resources:
```Cannot configure multiple urlRewrite filters for a single HTTPRouteRule.```

*Environment*:
Latest versions

### PR DESCRIPTION
**Pull request by guydc**
**What type of PR is this?**

Currently, EG accumulates multiple copies of extension filters (one for each reference). As a result, multiple matching filters can be found when processing routes, leading to the reported issue. 

Fixes #4989

Release Notes: Yes

### DIFF
--- a/internal/provider/kubernetes/resource.go
+++ b/internal/provider/kubernetes/resource.go
@@ -64,32 +64,36 @@ type resourceMappings struct {
 	// httpRouteFilters is a map of HTTPRouteFilters, where the key is the namespaced name,
 	// group and kind of the HTTPFilter.
 	httpRouteFilters map[utils.NamespacedNameWithGroupKind]*egv1a1.HTTPRouteFilter
+	// Set for storing HTTPRouteExtensions (Envoy Gateway or Custom) NamespacedNames referenced by various
+	// route rules objects.
+	allAssociatedHTTPRouteExtensionFilters sets.Set[utils.NamespacedNameWithGroupKind]
 }
 
 func newResourceMapping() *resourceMappings {
 	return &resourceMappings{
-		allAssociatedGateways:               sets.New[string](),
-		allAssociatedReferenceGrants:        sets.New[string](),
-		allAssociatedServiceImports:         sets.New[string](),
-		allAssociatedEndpointSlices:         sets.New[string](),
-		allAssociatedBackends:               sets.New[string](),
-		allAssociatedSecrets:                sets.New[string](),
-		allAssociatedConfigMaps:             sets.New[string](),
-		allAssociatedNamespaces:             sets.New[string](),
-		allAssociatedEnvoyProxies:           sets.New[string](),
-		allAssociatedEnvoyPatchPolicies:     sets.New[string](),
-		allAssociatedTLSRoutes:              sets.New[string](),
-		allAssociatedHTTPRoutes:             sets.New[string](),
-		allAssociatedGRPCRoutes:             sets.New[string](),
-		allAssociatedTCPRoutes:              sets.New[string](),
-		allAssociatedUDPRoutes:              sets.New[string](),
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: internal/provider/kubernetes/resource.go [hunk 1]
```diff
@@ -64,32 +64,36 @@ type resourceMappings struct {
 	// httpRouteFilters is a map of HTTPRouteFilters, where the key is the namespaced name,
 	// group and kind of the HTTPFilter.
 	httpRouteFilters map[utils.NamespacedNameWithGroupKind]*egv1a1.HTTPRouteFilter
```
arkodg, line 66: can this be removed ?

release-notes/current.yaml
```diff
@@ -21,6 +21,7 @@ new features: |
 # Fixes for bugs identified in previous versions.
 bug fixes: |
   Fixed a nil pointer error that occurs when a SecurityPolicy refers to a UDS backend
+  Fixed validation failure when multiple HTTPRoutes refer to the same extension filter
```
arkodg, line 24:
```suggestion
  Fixed a validation failure when multiple HTTPRoutes refer to the same extension filter
```

internal/provider/kubernetes/routes.go
```diff
@@ -416,12 +404,13 @@ func (r *gatewayAPIReconciler) processHTTPRoutes(ctx context.Context, gatewayNam
 
 					switch string(filter.ExtensionRef.Kind) {
 					case egv1a1.KindHTTPRouteFilter:
-						httpFilter, ok := resourceMap.httpRouteFilters[key]
-						if !ok {
+						httpFilter, err := r.getHTTPRouteFilter(ctx, key.Name, key.Namespace)
```
arkodg, line 407: lets wrap this inside `if r.hrfCRDExists {` ?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5011, opened 2025-01-06
feat: support BackendCluster for Remote JWKS

zhaohuabing:
This PR introduces support for representing JWT remote JWKS as `Backend` resources. With this enhancement, users can:
* **Specify a self-signed CA:** use `BackendTLSPolicy` to define a self-signed CA  for the remote JWKS
* **Configure retry policies**: use `rmoteJWKS.backendSettings.retry` to sepecify the retry policy for the remote JWKS

Implements: #3536
Release Notes: Yes


api/v1alpha1/validation/securitypolicy_validate.go
zhaohuabing, line 1: SecuritcyPolicy validation has been moved up from the xds translation to the API translation to catch the errors earlier and reflect the validation errors in the SecurityPolicy status.

internal/gatewayapi/securitypolicy.go
```diff
@@ -151,17 +152,6 @@ func (t *Translator) ProcessSecurityPolicies(securityPolicies []*egv1a1.Security
 					continue
 				}
 
-				if err := validation.ValidateSecurityPolicy(policy); err != nil {
```
zhaohuabing, line 154: We don't need to validate the policy again here because:
* Most of the validations are enforced by the k8s validator
* The extra validation for JWT Providers are enforced in the `validateJWTProvider` method

arkodg: this function also added some validation around api key auth that could not be represented in CEL due to the cost https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/api/v1alpha1/validation/securitypolicy_validate.go#L75
so I believe we cannot rm this function
cc @sanposhiho

zhaohuabing: IMO, this should be validated through the a CEL validation.

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -1,154 +0,0 @@
...
-	case spec.Authorization != nil:
-		sum++
-	case spec.APIKeyAuth != nil:
-		sum++
-	case spec.BasicAuth != nil:
-		sum++
-	case spec.ExtAuth != nil:
-		sum++
-	case spec.OIDC != nil:
-		sum++
```
zhaohuabing, line 58: Remove this empty spec check because other policy  types don't do this check. We can apply it to all the policy types in a follow-up PR if it's needed.

api/v1alpha1/validation/securitypolicy_validate.go
```diff
@@ -1,154 +0,0 @@
-// Copyright Envoy Gateway Authors
```
zhaohuabing, line 1: This file is deleted from the API package and the validations are moved into the gatewapi translator.

internal/gatewayapi/securitypolicy.go
sanposhiho: Specifically, I wanted to use the following CEL, but it hit this estimated cost problem:

```
((size(self.headers) > 0 ? 1 : 0) + (size(self.params) > 0 ? 1 : 0) + (size(self.cookies) > 0 ? 1 : 0)) == 1
```

which is actually very similar to the second issue above complainted.

zhaohuabing: Thanks for the explanation! @sanposhiho 

I kept the validation for APIKeyAuth in the Gateway API translator.

internal/cmd/egctl/testdata/translate/out/invalid-securitypolicy.all.yaml
```diff
@@ -108,8 +108,8 @@ securityPolicies:
         namespace: envoy-gateway-system
       conditions:
       - lastTransitionTime: null
-        message: 'Invalid SecurityPolicy: no security policy is specified.'
```
arkodg, line 111: does this test case need to be revisited ? its called `invalid-securitypolicy` but `Accepted=true`

zhaohuabing: Good catch.

zhaohuabing: Dleted this test file.

StephenRobin:
@zhaohuabing Thank you for working on this! It's been blocking us from adopting Envoy Gateway and I'm looking forward to seeing this merged.

I notice this PR doesn't update the documentation, it would be good to see an example added there. Also for using OIDC with an internal certificate, it took me a couple of hours to work out how to do that.

arkodg (approved):
LGTM thanks !

zhaohuabing:
@StephenRobin 

> I notice this PR doesn't update the documentation, it would be good to see an example added there. 

Will update the docs in a follow-up PR.

> Also for using OIDC with an internal certificate, it took me a couple of hours to work out how to do that.

You can find an example for OIDC here: https://gateway.envoyproxy.io/docs/tasks/security/oidc/#connect-to-an-oidc-provider-with-self-signed-certificate

zirain (approved):

Issue #4811, opened 2024-11-29
Improve "extension-server" documentation


Tharsanan1:
*Description*:
> Problem: In the [documentation](https://gateway.envoyproxy.io/docs/tasks/extensibility/extension-server/) under the "Configure Envoy Gateway to use the Extension Server" section, the instructions suggest modifying the configuration file. However, the linked "Configuration file" points to the EnvoyGateway API reference, which is misleading. Initially, it appears that changes should be made to a CR related to kind: EnvoyGateway, but in reality, the changes must be applied to the ConfigMap named envoy-gateway-config.
Proposed Solution: Update the documentation to explicitly state that the required changes should be made to the ConfigMap envoy-gateway-config

PR #4879, opened 2024-12-09
fix: Fix example documentation to include all the expected privileges for extension server policies

liorokman:
**Which issue(s) this PR fixes**:
Fixes #4812 

Release Notes: No

PR #5013, opened 2025-01-06
docs: fix extension-server instructions

Dean-Coakley:
Updates instructions to match where extension-server config is now loaded from.

Fixes https://github.com/envoyproxy/gateway/issues/4811

Release Notes: No


Dean-Coakley:
@zhaohuabing 

1. Should the fix for this look something like this? I had a look around but I don't see a reference in the code for the expected format in the configmap. 
These changes are a guess based on other docs such as https://github.com/envoyproxy/gateway/blob/main/site/content/en/docs/tasks/observability/gateway-observability.md#metrics

2. Should the same fix be copied over to versioned docs 1.1 and 1.2 or are the current docs correct for those versions?

site/content/en/latest/tasks/extensibility/extension-server.md
```diff
@@ -88,48 +88,55 @@ image name and tag.
 * Grant Envoy Gateway's `ServiceAccount` permission to access the extension server's CRD
 
   ```shell
-  kubectl create clusterrole listener-context-example-status-update \
```
liorokman, line 91: Don't remove this, it reverts #4879

site/content/en/latest/tasks/extensibility/extension-server.md
```diff
@@ -88,48 +88,55 @@ image name and tag.
...
-             --resource=ListenerContextExample/status
-
   kubectl create clusterrole listener-context-example-viewer \
              --verb=get,list,watch  \
              --resource=ListenerContextExample
 
   kubectl create clusterrolebinding envoy-gateway-listener-context \
              --clusterrole=listener-context-example-viewer \
              --serviceaccount=envoy-gateway-system:envoy-gateway
-
```
liorokman, line 102: Don't remove this, it reverts #4879

liorokman (changes requested):

liorokman (approved):
LGTM

arkodg (approved):

PR #5018, opened 2025-01-06
fix: enable ipv4 compat mode for dual stack cluster support

### LINKED ISSUE #5017: Single stack IPv6 on dual stack cluster not passing ready checks
*Description*:

I have a dual stack (IPv4-first) cluster and am trying to create a single stack gateway. I expect the gateway service to have `ipFamilies: ["IPv6"]` and `ipFamilyPolicy: SingleStack`. Using the proxy config below, the gateway pod never passes readiness checks.

The k8s event log for the gateway pod:

```
Events:
  Type     Reason     Age                  From               Message
  ----     ------     ----                 ----               -------
  Normal   Scheduled  3m1s                 default-scheduler  Successfully assigned envoy-gateway-system/envoy-gateway-eg-ipv6-d7ea7f64-59bb5cf866-bpfrf to talos-b8s-ysm
  Normal   Pulled     3m                   kubelet            Container image "docker.io/envoyproxy/envoy:distroless-v1.32.2" already present on machine
  Normal   Created    3m                   kubelet            Created container: envoy
  Normal   Started    3m                   kubelet            Started container envoy
  Normal   Pulled     3m                   kubelet            Container image "docker.io/envoyproxy/gateway:v1.2.4" already present on machine
  Normal   Created    3m                   kubelet            Created container: shutdown-manager
  Normal   Started    3m                   kubelet            Started container shutdown-manager
  Warning  Unhealthy  1s (x18 over 2m51s)  kubelet            Startup probe failed: Get "http://10.244.26.31:19001/ready": dial tcp 10.244.26.31:19001: connect: connection refused
```

*Repro steps*:

Create a gateway with this proxy config on a dual stack IPv4-first cluster:

```yaml
kind: EnvoyProxy
metadata:
  name: homelab-proxy-ipv6-config
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        replicas: 2
      envoyService:
        annotations:
          metallb.io/address-pool: ipv6-dynamic-pool
        patch:
          value:
            spec:
              ipFamilies:
                - IPv6
              ipFamilyPolicy: SingleStack
  ipFamily: IPv6
```

sharing.

*Environment*:

Gateway helm chart v1.2.4

*Logs*:

tekulvw:
**What type of PR is this?**

Bug fix.

Based on the [envoy docs](https://www.envoyproxy.io/docs/envoy/v1.22.11/api-v3/config/core/v3/address.proto#envoy-v3-api-field-config-core-v3-socketaddress-ipv4-compat) the `ipv4_compat` flag must be set anytime an envoy proxy is required to respond to IPv4 requests when the listening address is defined as `::`. There are two cases when this is required with the current gateway implementation: DualStack gateway and SingleStack IPv6 gateway on a DualStack IPv4-first cluster. The latter case is not handled without this, or a similar, change.

Reviewing a relevant Envoy proxy [PR](https://github.com/envoyproxy/envoy/pull/23616/files), this change should be safe to keep for SingleStack IPv6 config on a SingleStack IPv6 cluster.

This change _only_ affects the static resource listener for pod readiness checks. The data plane listener will continue to be IPv6 only.

Fixes #5017 

Release Notes: Yes/No

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
+++ b/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
@@ -81,6 +81,7 @@ spec:
                   address: '::'
                   port_value: 19001
                   protocol: TCP
+                  ipv4_compat: true
               filter_chains:
               - filters:
                 - name: envoy.filters.network.http_connection_manager
--- a/internal/xds/bootstrap/bootstrap.yaml.tpl
+++ b/internal/xds/bootstrap/bootstrap.yaml.tpl
@@ -68,7 +68,7 @@ static_resources:
         address: '{{ .ReadyServer.Address }}'
         port_value: {{ .ReadyServer.Port }}
         protocol: TCP
-        {{- if eq .IPFamily "DualStack"}}
+        {{- if eq .IPFamily "DualStack" "IPv6" }}
         ipv4_compat: true
         {{- end }}
     filter_chains:
--- a/internal/xds/bootstrap/testdata/render/ipv6.yaml
+++ b/internal/xds/bootstrap/testdata/render/ipv6.yaml
@@ -37,6 +37,7 @@ static_resources:
         address: '::'
         port_value: 19001
         protocol: TCP
+        ipv4_compat: true
     filter_chains:
     - filters:
       - name: envoy.filters.network.http_connection_manager


internal/xds/translator/listener.go
```diff
@@ -180,8 +180,9 @@ func buildXdsTCPListener(
 		},
 	}
 
-	if ipFamily != nil && *ipFamily == egv1a1.DualStack {
+	if ipFamily != nil && *ipFamily == egv1a1.DualStack || *ipFamily == egv1a1.IPv6 {
```
arkodg, line 183: based on https://github.com/envoyproxy/gateway/pull/4817/files we shouldnt need this, because it should listen on `::`

tekulvw: Only thing I'm aware of is that this block was added in that PR. If it was necessary to support `DualStack` family it follows (to me) that it would be necessary in a dual stack cluster. I don't know enough about xds/envoy to explain why.

tekulvw: Based on [envoy docs](https://www.envoyproxy.io/docs/envoy/v1.22.11/api-v3/config/core/v3/address.proto#envoy-v3-api-field-config-core-v3-socketaddress-ipv4-compat) this seems to be a required field for this use case (IPv6 single stack service on a dual stack IPv4 first cluster).

However, due to envoy defaults, this is likely only valid to set specifically for this usecase (IPv6 singlestack on dualstack cluster) and may break on a different usecase (IPv6 singlestack on IPv6 singlestack cluster).

> listener: the [ipv4_compat](https://www.envoyproxy.io/docs/envoy/v1.22.11/api-v3/config/core/v3/address.proto#envoy-v3-api-field-config-core-v3-socketaddress-ipv4-compat) flag can only be set on Ipv6 address and Ipv4-mapped Ipv6 address. A runtime guard is added envoy.reloadable_features.strict_check_on_ipv4_compat and the default is true.

[source](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22/v1.22.0.html)

This leads me to believe this should be a separate setting in the EnvoyProxy config. Thoughts?

tekulvw: Turns out that `listener: removed envoy.reloadable_features.strict_check_on_ipv4_compat and legacy code paths.` was removed from envoy on a [subsequent patch](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.25/v1.25.0.html).

It seems to me this is a safe change to make.

tekulvw: Relevant Envoy [PR](https://github.com/envoyproxy/envoy/pull/23616/files)

internal/xds/translator/listener.go
```diff
@@ -180,8 +180,9 @@ func buildXdsTCPListener(
 		},
 	}
 
-	if ipFamily != nil && *ipFamily == egv1a1.DualStack {
+	if ipFamily != nil && (*ipFamily == egv1a1.DualStack || *ipFamily == egv1a1.IPv6) {
```
zirain, line 183: will this make it impossible to generate a pure IPv6 listener?

tekulvw: To my understanding, this flag allows envoy to respond to IPv4 requests when the listening address is `::`. A pure IPv6 listener on a single stack IPv6 cluster should work, it seems the E2E tests confirm this though I haven't tested it personally.

Is a pure IPv6 listener possible on a dual stack cluster with this change? No. The alternative would be to introduce another parameter to the EnvoyProxy spec which allows this behavior.

tekulvw:
I was entirely incorrect about the source of the `ipv4_compat` flag. The proxy deployment template is the only required change necessary to fix the issue I'm encountering (single stack IPv6 service on dual stack IPv4-first cluster).

@zirain this still does not consider a possibility for a pure IPv6 listener.

arkodg (approved):
LGTM thanks !
can you add/move some more details from the comments into the PR description about the fix, so its easier to understand the commit in the future

arkodg:
@tekulvw the listener receiving traffic is still a pure IPv6 listener and will only traffic from IPv6 downstreams. The issue here is kublet is making a request via a IPv4 interface, so the `ipv4_compat` field is enabled for the health check / readiness port only to respond to those requests, not for the data plane traffic port

tekulvw:
> @tekulvw the listener receiving traffic is still a pure IPv6 listener and will only accept traffic from IPv6 downstreams. The issue here is kublet is making a request via a IPv4 interface, so the `ipv4_compat` field is enabled for the health check / readiness port only to respond to those requests, not for the data plane traffic port

You are correct, good call! I will update the description.

zirain (approved):

PR #5020, opened 2025-01-07
bump envoyproxy/go-control-plane

zirain:
fixes: https://github.com/envoyproxy/gateway/pull/5007


internal/xds/extensions/extensions.go
```diff
@@ -3,9 +3,12 @@
 // The full text of the Apache license is available in the LICENSE file at
 // the root of the repo.
 
+package extensions
+
 // Import all Envoy filter types so they are registered and deserialization does not fail
 // when using them in the "typed_config" attributes.
-package extensions
+// DO NOT REMOVE THE FOLLOWING IMPORTS
+import _ "github.com/envoyproxy/go-control-plane/contrib/envoy/extensions/compression/qatzip/compressor/v3alpha"
```
arkodg, line 11: curious why contrib is needed here

zirain: it's splited from `github.com/envoyproxy/go-control-plane` by default.

arkodg (approved):
LGTM thanks !

internal/xds/extensions/extensions.go
arkodg: @zirain and I discussed offline around whether we should include the `contrib` imports or not.
for now we should since we have docs that link to `contrib` images https://gateway.envoyproxy.io/docs/tasks/security/private-key-provider/
if image size becomes an issue, we can revisit this in the future

zhaohuabing (approved):

Issue #3993, opened 2024-08-03
support GEP-1731 Configurable Retries


zirain:
https://github.com/kubernetes-sigs/gateway-api/pull/3199

wait for gateway api v1.2

zirain:
@guydc @arkodg what is the relationship between https://gateway.envoyproxy.io/docs/api/extension_types/#retry and gep-1731?

arkodg:
its similar to timeouts, we'll have to support both (core gateway API as well as the extension field in BackendTrafficPolicy), but core setting overrides extension setting

PR #5024, opened 2025-01-08
feat: support GEP-1731

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3993


internal/xds/translator/route.go
```diff
@@ -573,8 +572,8 @@ func buildHashPolicy(httpRoute *ir.HTTPRoute) []*routev3.RouteAction_HashPolicy
 }
 
 func buildRetryPolicy(route *ir.HTTPRoute) (*routev3.RetryPolicy, error) {
-	rr := route.Traffic.Retry
-	any, err := protocov.ToAnyWithValidation(&previoushost.PreviousHostsPredicate{})
+	rr := route.GetRetry()
```
arkodg, line 575: looks clean ! 
cc @guydc

release-notes/current.yaml
```diff
@@ -19,6 +19,7 @@ new features: |
   Added support for patching EnvoyProxy.spec.provider.kubernetes.envoyHpa and EnvoyProxy.spec.provider.kubernetes.envoyPDB
   Added support for defining rateLimitHpa in EnvoyGateway API
   Added support for preserving the user defined HTTPRoute match order in EnvoyProxy API
+  Added support for GEP-1731
```
arkodg, line 22:
```suggestion
  Added support for GEP-1731 (HTTPRoute Retries)
```

internal/ir/xds.go
```diff
@@ -714,6 +714,21 @@ type HTTPRoute struct {
 	SessionPersistence *SessionPersistence `json:"sessionPersistence,omitempty" yaml:"sessionPersistence,omitempty"`
 	// Timeout is the time until which entire response is received from the upstream.
 	Timeout *metav1.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	// Retry defines the retry policy for the route.
+	// This's from core Gateway API, and should take precedence over Traffic.Retry.
```
arkodg, line 718:
```suggestion
	// This is derived from the core Gateway API, and should take precedence over Traffic.Retry.
```

internal/gatewayapi/route.go
```diff
@@ -271,6 +271,11 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 	return routeRoutes, nil
 }
 
+func processHTTPRouteRule(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
```
arkodg, line 274: suggest using a different name here, since we have `processHTTPRouteRules` https://github.com/envoyproxy/gateway/blob/00ecd08a789e3b443bc15561a10461f4d6816630/internal/gatewayapi/route.go#L183 which processes everything, something like `processRouteTrafficFeatures` maybe ?

internal/gatewayapi/route.go
```diff
@@ -293,6 +298,38 @@ func processRouteTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
...
+	}
+	if retry.Backoff != nil {
+		d, err := time.ParseDuration(string(*retry.Backoff))
+		if err == nil {
+			res.PerRetry = &ir.PerRetryPolicy{
+				BackOff: &ir.BackOffPolicy{
+					BaseInterval: ptr.To(metav1.Duration{Duration: d}),
+				},
+			}
+		}
```
arkodg, line 319: we should add another section here
`if rule.Timeout != nil and rule.timeouts.backendRequest != nil` we should set `perRetry.Timeout` to that val
https://gateway-api.sigs.k8s.io/geps/gep-1742/#timeout-values

internal/gatewayapi/route.go
```diff
@@ -751,10 +788,11 @@ func (t *Translator) processHTTPRouteParentRefListener(route RouteContext, route
 					IsHTTP2:               routeRoute.IsHTTP2,
 					SessionPersistence:    routeRoute.SessionPersistence,
 					Timeout:               routeRoute.Timeout,
+					Retry:                 routeRoute.Retry,
 				}
-				if routeRoute.Traffic != nil {
+				if retry := routeRoute.GetRetry(); retry != nil {
```
arkodg, line 793: @guydc @liorokman do you remember why this code is needed ? it doesnt feel like it should be here

arkodg:
thanks @zirain , this PR looks good, added some minor comments !

internal/gatewayapi/route.go
arkodg: cc @shawnh2 seems like this was introduced in https://github.com/envoyproxy/gateway/commit/6f6d8689d18f8b2011eaaf62c39ccf6930476e3e but the logic doesnt hold true because you seperate out route timeout and retries from BTP timeout and retries

arkodg: my vote would be to delete this block L800-803

zirain: removed, let's see how CI reacted.

arkodg (approved):
LGTM thanks !

arkodg (approved):
LGTM thanks !

arkodg (approved):

PR #5026, opened 2025-01-09
fix: check before setting cookie TTL in sessionPersistence

arkodg:
Fixes a null ptr exception when the cookie ttl is nil but was being accessed without checking if its valid or not

### DIFF
--- a/internal/xds/translator/session_persistence.go
+++ b/internal/xds/translator/session_persistence.go
@@ -17,6 +17,7 @@ import (
 	cookiev3 "github.com/envoyproxy/go-control-plane/envoy/extensions/http/stateful_session/cookie/v3"
 	headerv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/http/stateful_session/header/v3"
 	httpv3 "github.com/envoyproxy/go-control-plane/envoy/type/http/v3"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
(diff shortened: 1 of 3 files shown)


arkodg:
cc @sanposhiho

internal/xds/translator/session_persistence.go
```diff
@@ -61,28 +60,41 @@ func (s *sessionPersistence) patchHCM(mgr *hcmv3.HttpConnectionManager, irListen
...
 			configName = headerConfigName
-			sessionCfg = &headerv3.HeaderBasedSessionState{
+			sessionCfg := &headerv3.HeaderBasedSessionState{
 				Name: sp.Header.Name,
 			}
-		}
 
-		sessionCfgAny, err := anypb.New(sessionCfg)
-		if err != nil {
-			return fmt.Errorf("failed to marshal %s config: %w", egv1a1.EnvoyFilterSessionPersistence.String(), err)
```
sanposhiho, line 85: Any reason you moved this `sessionCfgAny, err := anypb.New(sessionCfg)` to each `case` of the above `switch`? Looks like just making a duplicated code.

sanposhiho (commented):
Thanks for a bug catch 🙇

internal/xds/translator/session_persistence.go
arkodg: ack, address this, ptal

zhaohuabing (approved):
LGTM thanks!

sanposhiho (approved):
Thanks.

PR #4938, opened 2024-12-16
fix: shutdown-manager not respecting security context of container spec

Dean-Coakley:
**What type of PR is this?**
Fixes setting custom security context in container spec

> **Note**: I have not taken any steps to test my changes outside of my unit test. If this is required please provide some guidance on the best way to do so

**Which issue(s) this PR fixes**:

Fixes https://github.com/envoyproxy/gateway/issues/4881

Release Notes: Yes

PR #4857, opened 2024-12-06
fix: use tls config from BTP when connecting to the OIDC provider's well-known endpoint.

zhaohuabing:
Fixes #4838

Release Notes: Yes

PR #5029, opened 2025-01-10
[release/v1.2] cherry pick for v1.2.5

zhaohuabing:
Release Notes: Yes


arkodg:
thanks @zhaohuabing 
should we also include some/all of these ?
* https://github.com/envoyproxy/gateway/pull/4857
* https://github.com/envoyproxy/gateway/pull/4803
* https://github.com/envoyproxy/gateway/pull/4938

arkodg:
@zhaohuabing lets rm #4803 from the patch release since its causing in another patch to be picked up resulting in a CRD change

arkodg (approved):

Issue #5028, opened 2025-01-09
docs: add a column to show defaults in API docs


arkodg:
*Description*:
>Describe the issue.

Add a column in the API docs https://gateway.envoyproxy.io/docs/api/extension_types/ to show the default value of the API field

Similar to https://github.com/envoyproxy/gateway/pull/2547

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
I will take a look at this.

PR #5031, opened 2025-01-10
docs: show API default value

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5028

only support maker like `	// +kubebuilder:default=1024`


arkodg (dismissed):
LGTM thanks !

zhaohuabing (approved):

arkodg (approved):

Issue #5033, opened 2025-01-10
Unable to run `envoy` binary inside `envoy-gateway` container image


dshatokhin:
*Description*:

I'm trying to run `envoy-gateway` in **standalone** mode in a container using
provided image `envoyproxy/gateway:v1.2.4`. I'm reproducing steps from article [Standalone Deployment Mode](https://gateway.envoyproxy.io/docs/tasks/operations/standalone-deployment-mode/) but instead of binary image mentioned above is used.

The `envoy` binary successfully got downloaded but can't start with `no such file or directory` error.

This is common error when trying to run dynamically linked binary in a distroless image.

Instead of an error `envoy` should be able to start inside running container.

*Repro steps*:

Here are steps to reproduce, requires installed `docker`:

```shell
# Put standalone.yaml file from the article in the docs (only 'paths' field changed)
$ cat << EOF > /tmp/standalone.yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyGateway
gateway:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
provider:
  type: Custom
  custom:
    resource:
      type: File
      file:
        paths: ["/tmp/envoy-gateway/config"]
    infrastructure:
      type: Host
      host: {}
logging:
  level:
    default: debug
extensionApis:
  enableBackend: true
EOF

# Create directory for configs and widen the permissions (needed to be accessible inside container)
$ mkdir -p /tmp/envoy-gateway/config
$ sudo chmod -R 777 /tmp/envoy-gateway/

# Generate certificates and put them in a mounted directory
$ docker run --rm -v /tmp/envoy-gateway:/tmp/envoy-gateway envoyproxy/gateway:v1.2.4 certgen --local
2025-01-10T21:51:13.559Z        INFO    cmd/server.go:106       No config file provided, using default parameters
2025-01-10T21:51:14.318Z        INFO    cmd/certgen.go:71       generated certificates  {"path": "/tmp/envoy-gateway/certs"}

# Download quickstart.yaml
$ wget -q https://raw.githubusercontent.com/envoyproxy/gateway/refs/heads/main/examples/standalone/quickstart.yaml -O /tmp/envoy-gateway/config/quickstart.yaml

# Start 'envoy-gateway' with created config
[...]

PR #5034, opened 2025-01-11
use distroless-base-nossl image

arkodg:
It includes glibc which allows us to run envoy-proxy inside the container image

Fixes: https://github.com/envoyproxy/gateway/issues/5033

### DIFF
--- a/tools/docker/envoy-gateway/Dockerfile
+++ b/tools/docker/envoy-gateway/Dockerfile
@@ -4,7 +4,7 @@ RUN mkdir -p /var/lib/eg
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot@sha256:6cd937e9155bdfd805d1b94e037f9d6a899603306030936a3b11680af0c2ed58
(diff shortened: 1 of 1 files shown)


zirain:
should we mention this in release notes or we ship both?

arkodg:
> should we mention this in release notes or we ship both?

I'd prefer if we used 1 image, and later created images with multiple tags if we really needed to reduce size (which would be breaking change that would affect `standalone` only, but im hoping we dont have to go there), will add a release note to this PR

zhaohuabing (approved):

zhaohuabing:
I wonder why we need to run envoy inside the Envoy Gateway docker container?

arkodg:
> I wonder why we need to run envoy inside the Envoy Gateway docker container?

@zhaohuabing when someone runs the envoy-gateway in standalone mode using the docker image, the envoy fleet that gets forked as a child process within the container and not on the host

zhaohuabing:
> > I wonder why we need to run envoy inside the Envoy Gateway docker container?
> 
> @zhaohuabing when someone runs the envoy-gateway in standalone mode using the docker image, the envoy fleet that gets forked as a child process within the container and not on the host

Thanks for the explanation!

zirain (approved):

arkodg:
these can be disabled using `--ignore-unfixed` or filtering by severity https://trivy.dev/v0.21.3/vulnerability/examples/filter/

PR #5037, opened 2025-01-13
docs: version URL and image support baseURL

zirain:
this make it possible to change baseURL to any value in `hugo.toml`.


arkodg:
hey @zirain can you elaborate a little more on the benefit of this PR

zirain:
this enable support for baseURL like `https://example.com/subproject`.

arkodg:
ah thanks
cc @missBerg

arkodg (approved):

PR #5045, opened 2025-01-13
feat(certgen): add flag for certificate overwrite

### LINKED ISSUE #4935: missing the envoy-gateaway-config for certgen command
*Description*:
returned, etc.

Currently, the certgen job does not mount the envoy-gateway-config, so it can not read the extension config from envoy-gateway-config.

*Repro steps*:

sharing.

*Environment*:

*Logs*:

guydc:
**What type of PR is this?**

Fixes #4935

Release Notes: Yes


charts/gateway-helm/templates/certgen.yaml
```diff
@@ -21,8 +21,12 @@ spec:
     spec:
       containers:
       - command:
+        {{- if .Values.certgen.job.command }}
+        {{- toYaml .Values.certgen.job.command | nindent 8 }}
```
arkodg, line 25: shouldnt this be `args` ?

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -92,6 +92,8 @@ kubernetesClusterDomain: cluster.local
 certgen:
   job:
     annotations: {}
+    args: []
+    command: []
```
arkodg, line 96: command isnt needed

internal/cmd/certgen.go
```diff
@@ -52,6 +57,10 @@ func certGen(local bool) error {
 	}
 	log := cfg.Logger
 
+	if overwriteControlPlaneCerts {
+		cfg.EnvoyGateway.Provider.Kubernetes.OverwriteControlPlaneCerts = ptr.To(true)
```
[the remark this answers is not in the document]
guydc: Well, previously, certgen would use the the result of `config.New` since there was never a path to a config file. `config.New` sets values that are relevant to correct certificate generation like the DNS domain and other things like log levels, so I'd rather keep it for compatibility.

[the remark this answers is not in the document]
arkodg: cool, thanks, makes sense

test/helm/gateway-helm/deployment-images-config.out.yaml
```diff
@@ -544,7 +544,8 @@ spec:
         app: certgen
     spec:
       containers:
-      - command:
+      - args:
```
arkodg, line 547: args should not be set by default

arkodg (approved):
LGTM thanks !

guydc:
cc @qicz , @zirain

zirain (approved):

PR #5066, opened 2025-01-15
fix panic when updating the envoy-gateway-config configMap

### LINKED ISSUE #5065: Envoy Gateway panic when updating the envoy-gateway-config  configMap
```
panic: sync: negative WaitGroup counter

goroutine 1123 [running]:
sync.(*WaitGroup).Add(0xc002ad2000?, 0xc002982000?)
	/opt/hostedtoolcache/go/1.23.4/x64/src/sync/waitgroup.go:64 +0xd8
sync.(*WaitGroup).Done(...)
	/opt/hostedtoolcache/go/1.23.4/x64/src/sync/waitgroup.go:89
github.com/envoyproxy/gateway/internal/provider/kubernetes.New.func1()
	/home/runner/work/gateway/gateway/internal/provider/kubernetes/kubernetes.go:115 +0x45
created by github.com/envoyproxy/gateway/internal/provider/kubernetes.New in goroutine 528
	/home/runner/work/gateway/gateway/internal/provider/kubernetes/kubernetes.go:113 +0x77f
```

zhaohuabing:
Fixes #5065

Release Notes: Yes


test/e2e/e2e_test.go
```diff
@@ -58,7 +58,7 @@ func TestE2E(t *testing.T) {
 		Debug:                *flags.ShowDebug,
 		CleanupBaseResources: *flags.CleanupBaseResources,
 		ManifestFS:           []fs.FS{Manifests},
-		RunTest:              *flags.RunTest,
+		RunTest:              tests.RateLimitBasedJwtClaimsTest.ShortName,
```
zirain, line 61: nit?

tools/make/kube.mk
```diff
@@ -170,16 +170,9 @@ e2e-prepare: prepare-ip-family ## Prepare the environment for running e2e tests
 
 .PHONY: run-e2e
 run-e2e: e2e-prepare ## Run e2e tests
-	@$(LOG_TARGET)
-ifeq ($(E2E_RUN_TEST),)
+
```
zirain, line 173: why change this?

zirain: if you want to run single e2e test locally, try `E2E_RUN_TEST=RateLimitCIDRMatch make run-e2e`.

zirain: if you want to run single e2e test locally, try `E2E_RUN_TEST=RateLimitCIDRMatch make run-e2e`.

internal/cmd/server.go
```diff
@@ -128,6 +128,10 @@ func getConfigByPath(cfgPath string) (*config.Server, error) {
 // setupRunners starts all the runners required for the Envoy Gateway to
 // fulfill its tasks.
 func setupRunners(ctx context.Context, cfg *config.Server) (err error) {
+	// The Elected channel is used to block the tasks that are waiting for the leader to be elected.
+	// It will be closed once the leader is elected in the controller manager.
+	cfg.Elected = make(chan struct{})
```
arkodg, line 133: we cannot use one channel to block multiple threads (status updater and infra runner)

arkodg:
@zhaohuabing can you help explain why can't we use a waitgroup anymore ?

internal/cmd/server.go
zhaohuabing: The channel is closed after an EG instance is elected, so all the blocked threads will be notified.

zhaohuabing:
> @zhaohuabing can you help explain why can't we use a waitgroup anymore ?

WaitGroup will block the goroutines until `WaitGroup.Done()` is called. The blocked goroutines can't be terminated properly after reloading the `envoy-gateway-config` ConfigMap, result in goroutine leaks.

arkodg:
thanks for the explaination @zhaohuabing , can we achieve the same by 
* calling `cfg.Elected.Done()` when shutting down https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/internal/cmd/server.go#L228, unblocking the go routines
* instantiating a new `cfg` object in the `hook` https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/internal/cmd/server.go#L59

cc @zirain

zhaohuabing:
> thanks for the explaination @zhaohuabing , can we achieve the same by
> 
> * calling `cfg.Elected.Done()` when shutting down https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/internal/cmd/server.go#L228
>   , unblocking the go routines

There will be two issues in this approach:
* We don't know if `cfg.Elected.Done()`  has been called or not, and calling it repeately will result in panic.
* Unblocking it will result in the gorotines to continue execution when they shouldn't.


> * instantiating a new `cfg` object in the `hook` https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/internal/cmd/server.go#L59
> 
> cc @zirain

arkodg (approved):
LGTM thanks !

arkodg:
cc @alexwo

zirain (approved):

PR #5035, opened 2025-01-11
feat(translator): implement ratelimit costs

mathetake:
**What type of PR is this?**

The API implementation

**What this PR does / why we need it**:

This is the follow up on #4957 and implement the API.

**Which issue(s) this PR fixes**:
Fixes #4756

Release Notes: Yes

PR #5072, opened 2025-01-15
chore: move ratelimit per-route config to typedPerFilterConfig

zhaohuabing:
Address the comment on https://github.com/envoyproxy/gateway/pull/5035#discussion_r1914449675
Release Notes: No


internal/xds/translator/ratelimit.go
```diff
@@ -136,27 +136,8 @@ func patchRouteWithRateLimit(route *routev3.Route, irRoute *ir.HTTPRoute) error
...
 	}
 	global := irRoute.Traffic.RateLimit.Global
-	rateLimits, costSpecified := buildRouteRateLimits(irRoute.Name, global)
-	if costSpecified {
-		// PerRoute global rate limit configuration via typed_per_filter_config can have its own rate routev3.RateLimit that overrides the route level rate limits.
-		// Per-descriptor level hits_addend can only be configured there: https://github.com/envoyproxy/envoy/pull/37972
-		// vs the "legacy" core route-embedded rate limits doesn't support the feature due to the "technical debt".
-		//
-		// This branch is only reached when the response cost is specified which allows us to assume that
-		// users are using Envoy >= v1.33.0 which also supports the typed_per_filter_config.
```
zhaohuabing, line 146: It's safe to use typed_per_filter_config for the "ordinay" ratelimit configuration, as the EG v1.3.0 will bump Envoy to v1.33.0, which supports this feature.

mathetake:
This will break the existing EG cluster running with the old Envoy I guess? The "limits" field of the global rate limit  typed_per_filter_condig became available only in the main and the next v1.33

internal/xds/translator/ratelimit.go
mathetake: > as EG v1.3.0 will bump Envoy to v1.33.0, which supports this feature.

This will break the existing EG cluster running with the old Envoy I guess? so are we suggest users upgrade envoy fleet first before upgrading EG?

zhaohuabing:
> with the old Envoy I guess? The "limits" field of the global rate limit typed_per_filter_condig became available only in the main and the next v1.33

Ha @mathetake you're absolutely right! Closed this PR, and raised an issue #5078 to track it - we can address in v1.4.0.

mathetake (approved):
👍

arkodg (approved):

PR #5077, opened 2025-01-16
feat(extension-manager): make grpc message size configurable

### LINKED ISSUE #5075: Support configuration of gRPC MaxMsgSize in Extension Manager
*Description*:
Extension manager currently doesn't support streaming: https://github.com/envoyproxy/gateway/blob/8b89dadfbd2cae74ab39055710688dae3c792512/proto/extension/service.proto#L20 

As a result, when there are many XDS resources, the message size may go over the gRPC default limit of 4mb. 

While streaming is the preferred solution, a short-term mitigation could be to allow users to configure the max message size to send/receive in the EnvoyGateway ExtensionManager config. 

[optional *Relevant Links*:]

guydc:
**What type of PR is this?**

Fixes #5075

Release Notes: Yes


api/v1alpha1/envoygateway_types.go
```diff
@@ -512,6 +512,14 @@ type ExtensionManager struct {
 	//
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// GRPCMaxMessageSizeBytes defines the maximum message size in bytes that can be
```
zirain, line 516: I assume that we didn't support http, so `MaxMessageSizeBytes` or `MaxMessageSize` is fine?

zirain (dismissed):

api/v1alpha1/envoygateway_types.go
```diff
@@ -512,6 +512,14 @@ type ExtensionManager struct {
...
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// MaxMessageSizeBytes defines the maximum message size in bytes that can be
+	// sent to or received from the Extension Service.
+	// Default: 4 MB
+	// +kubebuilder:validation:Minimum=1
+	//
+	// +optional
+	MaxMessageSizeBytes *int `json:"grpcMaxMessageSize,omitempty"`
```
arkodg, line 522: the json tag is different from the field name

also, can you check if we use the `bytes` suffix in other places of the API (e.g. buffer limit) or is it implied and part of the doc string ? would be good to keep it consistent

api/v1alpha1/envoygateway_types.go
```diff
@@ -512,6 +512,14 @@ type ExtensionManager struct {
...
 	// +optional
 	FailOpen bool `json:"failOpen,omitempty"`
+
+	// MaxMessageSize defines the maximum message size in bytes that can be
+	// sent to or received from the Extension Service.
+	// Default: 4 MB
+	// +kubebuilder:validation:Minimum=1
+	//
+	// +optional
+	MaxMessageSize *int `json:"maxMessageSize,omitempty"`
```
arkodg, line 522: should we use `resource.Quantity` here similar to https://github.com/envoyproxy/gateway/blob/0ee0a8af8e0a86bfbafbca4cee6d5432afa07830/api/v1alpha1/connection_types.go#L28?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5079, opened 2025-01-16
xds: move ready to dynamic listener

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/2810


zirain:
wait https://github.com/envoyproxy/gateway/pull/5081

internal/xds/bootstrap/bootstrap.go
```diff
@@ -86,6 +83,7 @@ type bootstrapParameters struct {
 	// EnablePrometheusCompression defines whether to enable HTTP compression on metrics endpoint for prometheus.
 	EnablePrometheusCompression bool
```
arkodg, line 84: should we delete fields no longer used ?

zirain: I recall I addressed this with a TODO for following refactor.

guydc:
Do we want to maybe keep prom stats in bootstrap config and serve readiness on a different port to avoid breaking scraping?
I think that metrics have some value even if the proxy is not ready, e.g. they can reveal issues in the proxy subscribing for XDS with EG, handling the bootstrap config, etc.

zirain:
> Do we want to maybe keep prom stats in bootstrap config and serve readiness on a different port to avoid breaking scraping? I think that metrics have some value even if the proxy is not ready, e.g. they can reveal issues in the proxy subscribing for XDS with EG, handling the bootstrap config, etc.

IIUC, now they share the listener.

Issue #5080, opened 2025-01-16
provider: EG should not provision if `parametersRef` not found


zirain:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg-daemonset
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      port: 80
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: All
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: eg-daemonset
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: eg-daemonset
  namespace: default
spec:
  ipFamily: IPv4
  provider:
    type: Kubernetes
    kubernetes:
      envoyDaemonSet:
        patch:
          type: StrategicMerge
          value:
            spec:
              template:
                spec:
                  containers:
                    - name: envoy
                      readinessProbe:
                        initialDelaySeconds: 5
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: foo-route
  namespace: default
spec:
  parentRefs:
    - name: eg-daemonset
  rules:
    - backendRefs:
        - name: infra-backend-v1
          port: 8080
      matches:
        - path:
            type: PathPrefix
            value: /foo

```

output:

```yaml
k get deployments.apps -n envoy-gateway-system
NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
envoy-default-eg-daemonset-aee8b7f7   1/1     1            1           16s
envoy-gateway                         1/1     1            1           38s

╭─ ~/Codes/servicemesh/gateway on xds-ready-listener !2                                         Ruby 9.2.6.0 at ○ kind-envoy-gateway at 16:58:09
╰─❯ k get ds -n envoy-gateway-system
NAME                                  DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
envoy-default-eg-daemonset-aee8b7f7   1         1         1       1            1           <none>          20s

```

zirain:
EG should not provision provider if `parametersRef` resource not found.
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg-daemonset
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      port: 80
[...]

PR #5081, opened 2025-01-16
provider: EG should not provision if `parametersRef` not found

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5080

### DIFF
--- a/internal/gatewayapi/listener.go
+++ b/internal/gatewayapi/listener.go
@@ -188,24 +188,29 @@ func (t *Translator) processProxyObservability(gwCtx *GatewayContext, xdsIR *ir.
 
 	xdsIR.AccessLog, err = t.processAccessLog(envoyProxy, resources)
 	if err != nil {
-		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+		status.SetGatewayAccepted(gwCtx.Gateway, false, gwapiv1.GatewayReasonInvalidParameters,
 			fmt.Sprintf("Invalid access log backendRefs: %v", err))
 		return
 	}
 
 	xdsIR.Tracing, err = t.processTracing(gwCtx.Gateway, envoyProxy, t.MergeGateways, resources)
 	if err != nil {
-		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+		status.SetGatewayAccepted(gwCtx.Gateway, false, gwapiv1.GatewayReasonInvalidParameters,
 			fmt.Sprintf("Invalid tracing backendRefs: %v", err))
 		return
 	}
 
 	xdsIR.Metrics, err = t.processMetrics(envoyProxy, resources)
 	if err != nil {
-		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+		status.SetGatewayAccepted(gwCtx.Gateway, false, gwapiv1.GatewayReasonInvalidParameters,
 			fmt.Sprintf("Invalid metrics backendRefs: %v", err))
 		return
 	}
+
+	// need explicit setting accepted status to true, as the status may have been set to false in previous loop,
+	// e.g. when access log backendRefs created later than the Gateway or EnvoyProxy.
+	status.SetGatewayAccepted(gwCtx.Gateway, true, gwapiv1.GatewayReasonAccepted,
+		"The Gateway has been scheduled by Envoy Gateway")
 }
 
 func (t *Translator) processInfraIRListener(listener *ListenerContext, infraIR resource.InfraIRMap, irKey string, servicePort *protocolPort, containerPort int32) {
--- a/internal/gatewayapi/status/gateway.go
+++ b/internal/gatewayapi/status/gateway.go
@@ -17,15 +17,14 @@ import (
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
-func UpdateGatewayListenersNotValidCondition(gw *gwapiv1.Gateway, reason gwapiv1.GatewayConditionReason, status metav1.ConditionStatus, msg string) *gwapiv1.Gateway {
-	cond := newCondition(string(gwapiv1.GatewayReasonListenersNotValid), status, string(reason), msg, time.Now(), gw.Generation)
+// SetGatewayAccepted set the status condition for the provided Gateway based on the accepted state.
+func SetGatewayAccepted(gw *gwapiv1.Gateway, accepted bool, reason gwapiv1.GatewayConditionReason, message string) *gwapiv1.Gateway {
+	status := metav1.ConditionFalse
+	if accepted {
+		status = metav1.ConditionTrue
+	}
+	cond := newCondition(string(gwapiv1.GatewayReasonAccepted), status, string(reason), message, time.Now(), gw.Generation)
 	gw.Status.Conditions = MergeConditions(gw.Status.Conditions, cond)
-	return gw
-}
-
-// UpdateGatewayStatusAcceptedCondition updates the status condition for the provided Gateway based on the accepted state.
-func UpdateGatewayStatusAcceptedCondition(gw *gwapiv1.Gateway, accepted bool) *gwapiv1.Gateway {
-	gw.Status.Conditions = MergeConditions(gw.Status.Conditions, computeGatewayAcceptedCondition(gw, accepted))
 	return gw
 }
 
@@ -114,20 +113,6 @@ func SetGatewayListenerStatusCondition(gateway *gwapiv1.Gateway, listenerStatusI
 		LastTransitionTime: metav1.NewTime(time.Now()),
 	}
 	gateway.Status.Listeners[listenerStatusIdx].Conditions = MergeConditions(gateway.Status.Listeners[listenerStatusIdx].Conditions, cond)
-}
-
-// computeGatewayAcceptedCondition computes the Gateway Accepted status condition.
-func computeGatewayAcceptedCondition(gw *gwapiv1.Gateway, accepted bool) metav1.Condition {
-	switch accepted {
-	case true:
-		return newCondition(string(gwapiv1.GatewayReasonAccepted), metav1.ConditionTrue,
-			string(gwapiv1.GatewayReasonAccepted),
-			"The Gateway has been scheduled by Envoy Gateway", time.Now(), gw.Generation)
-	default:
-		return newCondition(string(gwapiv1.GatewayReasonAccepted), metav1.ConditionFalse,
-			string(gwapiv1.GatewayReasonAccepted),
-			"The Gateway has not been scheduled by Envoy Gateway", time.Now(), gw.Generation)
-	}
 }
 
 const (
--- a/internal/gatewayapi/testdata/envoyproxy-accesslog-cel-with-invalid.out.yaml
+++ b/internal/gatewayapi/testdata/envoyproxy-accesslog-cel-with-invalid.out.yaml
@@ -18,9 +18,9 @@ gateways:
     conditions:
     - lastTransitionTime: null
       message: 'Invalid access log backendRefs: invalid CEL expression: )++++'
-      reason: Invalid
+      reason: InvalidParameters
       status: "False"
-      type: ListenersNotValid
+      type: Accepted
     listeners:
     - attachedRoutes: 0
       conditions:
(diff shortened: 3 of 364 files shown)


internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -406,11 +406,6 @@ func (i *Infra) deleteServiceAccount(ctx context.Context, r ResourceRender) (err
 
 // deleteDeployment deletes the Envoy Deployment in the kube api server, if it exists.
 func (i *Infra) deleteDeployment(ctx context.Context, r ResourceRender) (err error) {
-	// If deployment config is nil,ignore Deployment.
-	if deploymentConfig, er := r.DeploymentSpec(); deploymentConfig == nil {
```
zirain, line 410: You should not care whether the resource exists when deleting.

tools/make/kube.mk
```diff
@@ -138,10 +138,10 @@ kube-demo-undeploy: ## Uninstall the Kubernetes resources installed from the `ma
 #	tools/hack/run-kube-local.sh
 
 .PHONY: conformance
-conformance: create-cluster kube-install-image kube-deploy run-conformance delete-cluster ## Create a kind cluster, deploy EG into it, run Gateway API conformance, and clean up.
+conformance: create-cluster kube-install-image kube-deploy install-eg-addons run-conformance delete-cluster ## Create a kind cluster, deploy EG into it, run Gateway API conformance, and clean up.
```
zirain, line 141: GC(envoy-gateway) refernece EnvoyProxy(proxy-config), that make service `monitoring/otel-collector` is required.

internal/provider/kubernetes/controller.go
```diff
@@ -938,7 +938,12 @@ func (r *gatewayAPIReconciler) processGateways(ctx context.Context, managedGC *g
 
 		if err := r.processGatewayParamsRef(ctx, &gtw, resourceMap, resourceTree); err != nil {
 			r.log.Error(err, "failed to process infrastructure.parametersRef for gateway", "namespace", gtw.Namespace, "name", gtw.Name)
+			status.SetGatewayAccepted(&gtw, false, gwapiv1.GatewayReasonInvalidParameters, err.Error())
+			r.updateGatewayStatus(&gtw)
+			continue
 		}
+		status.SetGatewayAccepted(&gtw, false, gwapiv1.GatewayReasonAccepted, "The Gateway has not been scheduled by Envoy Gateway")
```
arkodg, line 945: why is it false here ?

internal/provider/kubernetes/status.go
```diff
@@ -568,8 +568,7 @@ func (r *gatewayAPIReconciler) updateStatusForGateway(ctx context.Context, gtw *
 		r.log.Info("failed to get Service for gateway",
 			"namespace", gtw.Namespace, "name", gtw.Name)
 	}
-	// update accepted condition
-	status.UpdateGatewayStatusAcceptedCondition(gtw, true)
```
arkodg, line 572: curious why this was removed

tools/make/kube.mk
arkodg: is this because we are using the same GC for conformance and e2e ? should we maintain 2 different GatewayClasses to keep each test suite more lightweight ?

internal/infrastructure/kubernetes/infra_resource.go
arkodg: was this a bug, uncovered while testing this change ?

internal/provider/kubernetes/controller.go
zirain: nit

tools/make/kube.mk
zirain: yes, we can do that later.

internal/infrastructure/kubernetes/infra_resource.go
zirain:
```
kind: GatewayClass
metadata:
  name: eg
spec:
  parametersRef:
    group: gateway.envoyproxy.io
    kind: EnvoyProxy
    name: eg-daemonset
    namespace: default
---
kind: Gateway
metadata:
  name: eg-daemonset
  namespace: default
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      port: 80
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: All
---
kind: EnvoyProxy
metadata:
  name: eg-daemonset
  namespace: default
spec:
  ipFamily: IPv4
  telemetry:
    accessLog:
      settings:
        - format:
            type: Text
            text: |
              [%START_TIME%] %METADATA(ROUTE:envoy-gateway:resources)% "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% "%REQ(X-FORWARDED-FOR)%" "%REQ(USER-AGENT)%" "%REQ(X-REQUEST-ID)%" "%REQ(:AUTHORITY)%" "%UPSTREAM_HOST%"
          matches:
            - "'x-envoy-logged' in request.headers"
          sinks:
            - type: OpenTelemetry
              openTelemetry:
                backendRefs:
                  - name: otel-collector
                    namespace: monitoring
                    port: 4317
                resources:
                  k8s.cluster.name: "envoy-gateway"
  provider:
    type: Kubernetes
    kubernetes:
      envoyDaemonSet:
        patch:
          type: StrategicMerge
          value:
            spec:
              template:
                spec:
                  containers:
                    - name: envoy
                      readinessProbe:
                        initialDelaySeconds: 5
---
kind: HTTPRoute
metadata:
  name: foo-route
  namespace: default
spec:
  parentRefs:
    - name: eg-daemonset
  rules:
    - backendRefs:
        - name: infra-backend-v1
          port: 8080
      matches:
        - path:
            type: PathPrefix
            value: /foo
```

with this demo, ds won't be removed.

internal/provider/kubernetes/status.go
zirain: I think we should respect the logic in `controller.go`.

test/config/gatewayclass.yaml
```diff
@@ -61,14 +61,6 @@ spec:
                     port: 4317
                 resources:
                   k8s.cluster.name: "envoy-gateway"
-        - sinks:
```
zirain, line 64: this's not working for a long time and should blocked gateway provision.

internal/provider/kubernetes/status.go
zirain: this made gateway always accepted, even the ref EnvoyProxy is not found.

internal/gatewayapi/listener.go
```diff
@@ -188,24 +188,28 @@ func (t *Translator) processProxyObservability(gwCtx *GatewayContext, xdsIR *ir.
...
 	if err != nil {
-		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+		status.SetGatewayAccepted(gwCtx.Gateway, false, gwapiv1.GatewayReasonInvalidParameters,
 			fmt.Sprintf("Invalid metrics backendRefs: %v", err))
 		return
 	}
+	
+	// need explicit setting accepted status to true, as the status may have been set to false in previous loop,
+	// e.g. when access log backendRefs created later than the Gateway or EnvoyProxy.
+	status.SetGatewayAccepted(gwCtx.Gateway, true, gwapiv1.GatewayReasonAccepted, "The Gateway has not been scheduled by Envoy Gateway")
```
arkodg, line 212: there are a few issues here
* we've added a new status condition that did not apply before
* accepted=true but the message says `The Gateway has not been scheduled by Envoy Gateway` , this is not useful for the user

zirain: my bad, it should be `The Gateway has not been scheduled by Envoy Gateway`.

internal/provider/kubernetes/controller.go
```diff
@@ -938,7 +938,13 @@ func (r *gatewayAPIReconciler) processGateways(ctx context.Context, managedGC *g
 
 		if err := r.processGatewayParamsRef(ctx, &gtw, resourceMap, resourceTree); err != nil {
 			r.log.Error(err, "failed to process infrastructure.parametersRef for gateway", "namespace", gtw.Namespace, "name", gtw.Name)
+			status.SetGatewayAccepted(&gtw, false, gwapiv1.GatewayReasonInvalidParameters, err.Error())
+			r.updateGatewayStatus(&gtw)
+			continue
 		}
+		// Set Gateway as accepted to passed test?
+		status.SetGatewayAccepted(&gtw, true, gwapiv1.GatewayReasonAccepted, "The Gateway has not been scheduled by Envoy Gateway")
```
zirain, line 946: @arkodg do you think we should remove this and update the test.
because now we reply the final result in translator.

tools/make/kube.mk
zirain: wait https://github.com/envoyproxy/gateway/pull/5091

internal/gatewayapi/testdata/envoyproxy-accesslog-cel-with-invalid.out.yaml
```diff
@@ -18,9 +18,9 @@ gateways:
     conditions:
     - lastTransitionTime: null
       message: 'Invalid access log backendRefs: invalid CEL expression: )++++'
-      reason: Invalid
+      reason: InvalidParameters
       status: "False"
-      type: ListenersNotValid
```
zirain, line 23: this seems not right, GatewayConditionType only allow `Accepted` and `Programmed`.

internal/gatewayapi/listener.go
```diff
@@ -188,24 +188,29 @@ func (t *Translator) processProxyObservability(gwCtx *GatewayContext, xdsIR *ir.
...
 	xdsIR.Metrics, err = t.processMetrics(envoyProxy, resources)
 	if err != nil {
-		status.UpdateGatewayListenersNotValidCondition(gwCtx.Gateway, gwapiv1.GatewayReasonInvalid, metav1.ConditionFalse,
+		status.SetGatewayAccepted(gwCtx.Gateway, false, gwapiv1.GatewayReasonInvalidParameters,
 			fmt.Sprintf("Invalid metrics backendRefs: %v", err))
 		return
 	}
+
+	// need explicit setting accepted status to true, as the status may have been set to false in previous loop,
+	// e.g. when access log backendRefs created later than the Gateway or EnvoyProxy.
```
zhaohuabing, line 211: This is not necessary as the Gateway status has always been set to nil in every translation loop.

https://github.com/envoyproxy/gateway/blob/126d5073aa3f47af3df93c365fd9105ce94d260c/internal/provider/kubernetes/controller.go#L1002-L1009

zhaohuabing: Addtionally, we should not set Gatway Accepted status solely based on the listener status.

internal/gatewayapi/status/gateway.go
```diff
@@ -17,18 +17,17 @@ import (
...
 
-func UpdateGatewayListenersNotValidCondition(gw *gwapiv1.Gateway, reason gwapiv1.GatewayConditionReason, status metav1.ConditionStatus, msg string) *gwapiv1.Gateway {
-	cond := newCondition(string(gwapiv1.GatewayReasonListenersNotValid), status, string(reason), msg, time.Now(), gw.Generation)
+// SetGatewayAccepted set the status condition for the provided Gateway based on the accepted state.
+func SetGatewayAccepted(gw *gwapiv1.Gateway, accepted bool, reason gwapiv1.GatewayConditionReason, message string) *gwapiv1.Gateway {
+	status := metav1.ConditionFalse
+	if accepted {
+		status = metav1.ConditionTrue
+	}
+	cond := newCondition(string(gwapiv1.GatewayReasonAccepted), status, string(reason), message, time.Now(), gw.Generation)
```
zhaohuabing, line 26:
```suggestion
	cond := newCondition(string(gwapiv1.GatewayConditionAccepted), status, string(reason), message, time.Now(), gw.Generation)
```

internal/provider/kubernetes/controller.go
```diff
@@ -940,7 +940,13 @@ func (r *gatewayAPIReconciler) processGateways(ctx context.Context, managedGC *g
 
 		if err := r.processGatewayParamsRef(ctx, &gtw, resourceMap, resourceTree); err != nil {
 			r.log.Error(err, "failed to process infrastructure.parametersRef for gateway", "namespace", gtw.Namespace, "name", gtw.Name)
+			status.SetGatewayAccepted(&gtw, false, gwapiv1.GatewayReasonInvalidParameters, err.Error())
+			r.updateGatewayStatus(&gtw)
```
zhaohuabing, line 944: `r.updateGatewayStatus` is used to  trigger a dummy Gateway Status update event to update the gateway IP address and can't be used for update Accepted status.

I suggest to add the Accepted status to the Gateway resource and pass it to the Gateway API translator to handle it.

internal/gatewayapi/listener.go
zirain: this's necessary during the test.

zhaohuabing:
@zirain I reponed https://github.com/envoyproxy/gateway/pull/5117 as I pefer the approach to handle the EnvoyProxy error in that PR.

PR #5082, opened 2025-01-16
fix certgen job race condition

ryanhristovski:
Envoy-gateway certgen job frequently goes into a crash loop when running a sync using ArgoCD in our environment with the following error:

```
envoy-gateway-gateway-helm-certgen-pxnd7 envoy-gateway-certgen failed to output certificates: failed to create or update secrets: failed to get secret infra-routing/envoy-gateway: failed to get API group resources: unable to retrieve the complete list of server APIs: v1: Unauthorized
```

Though, sometimes this job succeeds with no problems...

I believe this inconsistency is due to a race condition between the RBAC resources and the certgen job (both installed with `pre-install`). Since the RBAC resources do not exist continuously, and are only applied on installation and then removed, I believe the job Is trying to start before the permissions are available. After setting the `hook-weight` annotation in our environment we no longer see this problem.

Another solution could be to have the RBAC resources exist continuously without the `pre-hook` but I imagine this was a conscious security decision.

### DIFF
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -10,6 +10,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-1"
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
   {{- end }}
@@ -26,6 +27,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-1"
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
   {{- end }}
@@ -51,6 +53,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-1"
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
   {{- end }}
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -8,10 +8,8 @@ metadata:
   {{- if .Values.certgen.rbac.labels }}
   {{- toYaml .Values.certgen.rbac.labels | nindent 4 }}
   {{- end }}
-  annotations:
-    "helm.sh/hook": pre-install
-    "helm.sh/hook-weight": "-1"
   {{- if .Values.certgen.rbac.annotations }}
+  annotations:
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
   {{- end }}
 ---
@@ -25,10 +23,8 @@ metadata:
   {{- if .Values.certgen.rbac.labels }}
   {{- toYaml .Values.certgen.rbac.labels | nindent 4 }}
   {{- end }}
-  annotations:
-    "helm.sh/hook": pre-install
-    "helm.sh/hook-weight": "-1"
   {{- if .Values.certgen.rbac.annotations }}
+  annotations:
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
   {{- end }}
 rules:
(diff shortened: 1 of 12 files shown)


shahar-h:
The hook-weight should not make any difference since RBAC resources are installed by helm before `Job`. See: https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package.

ryanhristovski:
@shahar-h ArgoCD doesn't use `helm install` to apply its resources from my understanding, it applies yaml rendered with helm template - which I believe can cause this issue.

shahar-h:
> @shahar-h ArgoCD doesn't use `helm install` to apply its resources from my understanding, it applies yaml rendered with helm template - which I believe can cause this race condition.

According to argocd [install order](https://github.com/argoproj/gitops-engine/blob/master/pkg/sync/sync_tasks.go#L27) `Job` is installed after RBAC resources.

ryanhristovski:
@shahar-h setting this in draft while I run some tests on my side to replicate this. I'm not sure these same dependencies are applied on pre-install hooks. Thanks for reviewing

ryanhristovski:
@shahar-h I can't understand why this is failing, my only theory is that the order is not respected when it comes to hooks... regardless of being able to replicate the error, the more I thought about it, I don't think having a pre-install hook for RBAC makes sense.

The object persists throughout and they are just recreated each time a sync is triggered with nothing new added, I don't see any benefit in them being recreated each time theres a sync.

shahar-h:
> @shahar-h I was able to replicate my error with the following code, using a server-side apply... but its very inconsistent (it happens about 1/20 times, trying to understand why, and to see if the `hook-weight` even fixed this) 😅
> 
> ```
> ---
> apiVersion: v1
> kind: Namespace
> metadata:
>   name: test-namespace
>   labels:
>     name: test-namespace
> ---
> apiVersion: batch/v1
> kind: Job
> metadata:
>   name: test-job
>   namespace:test-namespace
>   annotations:
>     "helm.sh/hook": pre-install, pre-upgrade
> spec:
>   template:
>     spec:
>       serviceAccountName: test-sa
>       containers:
>         - name: test-container
>           # Using 'bitnami/kubectl' so we can run `kubectl get secrets`.
>           image: bitnami/kubectl:latest
>           command: ["/bin/sh", "-c"]
>           args:
>             - |
>               echo "Attempting to read secrets..."
>               kubectl get secrets
>               echo "If you see an RBAC or SA error above, the job ran before RBAC was ready."
>       restartPolicy: Never
> ---
> apiVersion: v1
> kind: ServiceAccount
> metadata:
>   name: test-sa
>   namespace: test-namespace
>   annotations:
>     "helm.sh/hook": pre-install
> ---
> apiVersion: rbac.authorization.k8s.io/v1
> kind: Role
> metadata:
>   name: test-role
>   namespace: test-namespace
>   annotations:
>     "helm.sh/hook": pre-install
> rules:
>   - apiGroups: [""]
>     resources: ["secrets"]
>     verbs: ["get", "list", "create", "update"]
> ---
> apiVersion: rbac.authorization.k8s.io/v1
> kind: RoleBinding
> metadata:
>   name: test-rolebinding
>   namespace: test-namespace
>   annotations:
>     "helm.sh/hook": pre-install
> roleRef:
>   apiGroup: rbac.authorization.k8s.io
>   kind: Role
>   name: test-role
> subjects:
>   - kind: ServiceAccount
>     name: test-sa
>     namespace: infra-routing
> ```
> 
> Setting to draft while I confirm this.

The order here is not aligned with helm/argocd since job should be installed after RBAC resources.

shahar-h:
> @shahar-h I can't understand why this is failing, my only theory is that the order is not respected when it comes to hooks... regardless of being able to replicate the error, the more I thought about it, I don't think having a pre-install hook for RBAC makes sense.
> 
> The object persists throughout and they are just recreated each time a sync is triggered with nothing new added, I don't see any benefit in them being recreated each time theres a sync. I updated this PR to remove the pre-install annotations on RBAC, let me know what you think.
> 
> Edit: will fix tests if I get positive consensus

You can't remove pre-install annotations from RBAC resources because they have to be installed before certgen job.

ryanhristovski:
> The order here is not aligned with helm/argocd since job should be installed after RBAC resources.

This is applying the resources via ArgoCD, so it follows the install order you pointed out [here](https://github.com/argoproj/gitops-engine/blob/master/pkg/sync/sync_tasks.go#L27). Which is the reason why it works ~19/20 times (otherwise in theory it would fail 100% of the time).

At the end of the day, this _is_ an error we are seeing while using ArgoCD but doesn't come up that frequently. Just wanted to flag it

ormandj:
We see this too, but it's almost every deployment with ArgoCD. We've tried 1.2.x and 1.3.x and have the same behavior with both. It takes quite some time to straighten itself out. The symptoms are the same as above. Just wanted to bring awareness that for some of us, it isn't intermittent and rare.

zhaohuabing:
I also ran into this while testing EG installation using Argo CD.

```
failed to output certificates: failed to create or update secrets: failed to get secret envoy-gateway-system/envoy-gateway: failed to get server groups: the server has asked for the client to provide credentials
```

arkodg:
closing this PR since its become inactive, this should be fixed in v1.4.0 with hook weights set to -1

PR #5100, opened 2025-01-20
feat(file-loader): Add recursive loading from sub-directories

### LINKED ISSUE #4689: Support recursive subdirectories for file-provider
*Description*:

Currently, the `paths` in file-provider does not support watch any recursive subdirectories:

https://github.com/envoyproxy/gateway/blob/2c986175211df64a269e19111b9ed480646c285a/api/v1alpha1/envoygateway_types.go#L308

It has been noted that each backend service related configs can be put in separate directory, but instead of watching all these directories, we can treat them as subdirectories of a root dir, and only watching this root dir.

[optional *Relevant Links*:]

Discussed in https://envoyproxy.slack.com/archives/C03E6NHLESV/p1731083390060779?thread_ts=1731035129.484099&cid=C03E6NHLESV

EshaanAgg:
**What type of PR is this?**
Adds support for recursive file loading. 

Fixes #4689 

Release Notes: Yes
```yaml
new features: |
  Added support for recursive subdirectories in the FileProvider API
```


internal/provider/file/resources.go
```diff
@@ -54,27 +55,48 @@ func loadFromFile(path string) (*resource.Resources, error) {
 	return resource.LoadResourcesFromYAMLBytes(bytes, false)
 }
 
-// loadFromDir loads resources from all the files under a specific directory excluding subdirectories.
 func loadFromDir(path string) ([]*resource.Resources, error) {
```
shawnh2, line 58: We may need to do some work under https://github.com/envoyproxy/gateway/blob/main/internal/provider/file/file.go#L148
to make sure that any changes happened in these subdirs will also trigger the resource reload

EshaanAgg:
Closing the PR due to corruption of commit history due to incorrect rebasing! Will open the new PR after making the required changes! Sorry for the trouble.

internal/provider/file/resources.go
shawnh2: Let's stick to the easy way at first, can resolve any performance issues later.

PR #5117, opened 2025-01-21
fix the gateway accepted condition

zhaohuabing:
This RP fixed these Gateway Status issues:

* To comply with the Gateway API spec, the condtion type must be `Accepted `, and the reason should be `InvalidParameters`. 

> 	// This reason is used with the "Accepted" condition when the
	// Gateway was not accepted because the parametersRef field
	// was invalid, with more detail in the message.
	GatewayReasonInvalidParameters GatewayConditionReason = "InvalidParameters"

* When the referenced EnvoyProxy is invalid, surface the error to the Gateway status

For example:

For a `Gatway` referencing a non-existing `EnvoyProxy`  `test`

```
spec:
  gatewayClassName: eg
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: test
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 8082
    protocol: HTTP
```

Before:
```
  conditions:
  - lastTransitionTime: "2025-01-21T07:42:26Z"
    message: The Gateway has been scheduled by Envoy Gateway
    observedGeneration: 12
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2025-01-21T07:42:26Z"
    message: Address assigned to the Gateway, 1/1 envoy replicas available
    observedGeneration: 12
    reason: Programmed
    status: "True"
    type: Programmed
  listeners:
  - attachedRoutes: 1
    conditions:
    - lastTransitionTime: "2025-01-21T07:42:16Z"
      message: Sending translated listener configuration to the data plane
      observedGeneration: 12
      reason: Programmed
      status: "True"
      type: Programmed
    - lastTransitionTime: "2025-01-21T07:42:16Z"
      message: Listener has been successfully translated
      observedGeneration: 12
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2025-01-21T07:42:16Z"
      message: Listener references have been resolved
      observedGeneration: 12
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    name: http
    supportedKinds:
      kind: HTTPRoute
      kind: GRPCRoute
  ```
  
  After:
  
  ```
  status:
  conditions:
  - lastTransitionTime: "2025-01-21T07:39:37Z"
    message: 'failed to find envoyproxy default/test: EnvoyProxy.gateway.envoyproxy.io
      "test" not found'
    observedGeneration: 12
    reason: InvalidParameters
    status: "False"
    type: Accepted
  ```
Related to: #2717

Release Notes: Yes

### CODE REVIEW

**Comment by zirain** [comment]
covered by https://github.com/envoyproxy/gateway/pull/5081?

**Comment by zhaohuabing** [comment]
> covered by #5081?

Ha, I found this while trying to fix https://github.com/envoyproxy/gateway/issues/2717 and haven't noticed that it's already covered in #5081.

**Comment by zhaohuabing** [comment]
Reopening this PR as I believe the approach taken here is preferable to the one in [PR #5081](https://github.com/envoyproxy/gateway/pull/5081).

File: internal/provider/kubernetes/controller.go [hunk 1]
```diff
@@ -999,9 +995,17 @@ func (r *gatewayAPIReconciler) processGateways(ctx context.Context, managedGC *g
...
+
 		// Discard Status to reduce memory consumption in watchable
 		// It will be recomputed by the gateway-api layer
 		gtw.Status = gwapiv1.GatewayStatus{}
+
+		if err := r.processGatewayParamsRef(ctx, &gtw, resourceMap, resourceTree); err != nil {
+			// Update the Gateway status to not accepted if there is an error processing the parametersRef
+			status.UpdateGatewayStatusNotAccepted(&gtw, gwapiv1.GatewayReasonInvalidParameters, err.Error())
+			r.log.Error(err, "failed to process infrastructure.parametersRef for gateway", "namespace", gtw.Namespace, "name", gtw.Name)
+		}
```
zirain, line 1007: I would prefer to add a `continue` here, otherwise the gateway instance will be provisoned?

zhaohuabing: This is tricky because:
* If we skip adding this to the `resourceTree`, the Gateway status won't be updated.
* If we update the status directly in the provider controller, it introduces a potential race condition: both the provider controller goroutine and the Gateway API translator goroutine will attempte to write to resources.GatewayStatus.

zhaohuabing: Skipped Gateways with invalid EnvoyProxy in https://github.com/envoyproxy/gateway/pull/5117/commits/8f51c9a6b131c54738cea8a4af9a60c942b439af

internal/gatewayapi/translator.go
```diff
@@ -156,7 +157,17 @@ func newTranslateResult(gateways []*GatewayContext,
 
 func (t *Translator) Translate(resources *resource.Resources) (*TranslateResult, error) {
 	// Get Gateways belonging to our GatewayClass.
-	gateways := t.GetRelevantGateways(resources)
+	allGateways := t.GetRelevantGateways(resources)
+
+	// Filter out Gateways that are not accepted.
+	// This will only happen if the Gateway references an invalid Envoyproxy, either directly or indirectly through GatewayClass.
+	gateways := make([]*GatewayContext, 0)
```
zirain, line 164:
```suggestion
	gateways := make([]*GatewayContext, 0, len(allGateways))
```

internal/gatewayapi/translator.go
```diff
@@ -241,7 +252,7 @@ func (t *Translator) Translate(resources *resource.Resources) (*TranslateResult,
 		}
 	}
 
-	return newTranslateResult(gateways, httpRoutes, grpcRoutes, tlsRoutes,
+	return newTranslateResult(allGateways, httpRoutes, grpcRoutes, tlsRoutes,
```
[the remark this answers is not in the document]
zhaohuabing: We should add allGateways to the translationResult to update the status for existing Gateways.

internal/gatewayapi/translator.go
```diff
@@ -156,7 +157,17 @@ func newTranslateResult(gateways []*GatewayContext,
 
 func (t *Translator) Translate(resources *resource.Resources) (*TranslateResult, error) {
 	// Get Gateways belonging to our GatewayClass.
-	gateways := t.GetRelevantGateways(resources)
+	allGateways := t.GetRelevantGateways(resources)
+
+	// Filter out Gateways that are marked as not accepted by the controller.
```
arkodg, line 162: this means a bad config on EnvoyProxy can take down all data plane xds config ?

arkodg: also, thoughts on changing the api of `GetRelevantGateways` to `acceptedGateways, failedGateways := GetRelevantGateways(res)` ?

zhaohuabing: > this means a bad config on EnvoyProxy can take down all data plane xds config ?

Yes, the current behavior is that the Gateway will be still accepted and the envoy fleet will be deployed.
We can keep the current behavior, but there is no other condition type to reflect a "partial error" status. The Gateway can only be either accepted or not.

internal/provider/kubernetes/status.go
```diff
@@ -568,10 +568,17 @@ func (r *gatewayAPIReconciler) updateStatusForGateway(ctx context.Context, gtw *
 		r.log.Info("failed to get Service for gateway",
 			"namespace", gtw.Namespace, "name", gtw.Name)
 	}
-	// update accepted condition
-	status.UpdateGatewayStatusAcceptedCondition(gtw, true)
-	// update address field and programmed condition
-	status.UpdateGatewayStatusProgrammedCondition(gtw, svc, envoyObj, r.store.listNodeAddresses()...)
+
+	// Clear the listeners status if the gateway is not accepted.
```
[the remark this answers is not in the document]
zhaohuabing: If translation is skipped, `listeners` list was created by not filled with correct status, so it has to be cleared.

internal/gatewayapi/translator.go
zhaohuabing: Or we could set `Accepted` as false but still provision the Envoy fleet and server the xDS.

arkodg: sounds good, lets document this in the breaking change that any issues with `EnvoyProxy` reference in a gateway resulting in a `Accepted=False` will result in no xDS being pushed to the fleet

internal/provider/kubernetes/status.go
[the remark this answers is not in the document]
arkodg: ah thanks

arkodg (dismissed):
LGTM thanks !

test/benchmark/config/gatewayclass.yaml
```diff
@@ -29,15 +29,6 @@ spec:
               memory: "256Mi"
               cpu: "500m"
   telemetry:
-    metrics:
-      prometheus: {}
-      sinks:
-        - type: OpenTelemetry
-          openTelemetry:
-            backendRefs:
-              - name: otel-collector
```
zhaohuabing, line 38: `otel-collector` is not installed in the bechmark test. Removing it form the EnvoyProxy, otherwise the gateway will not be accepted and the test will fial.

zirain (approved):

arkodg (approved):

PR #5118, opened 2025-01-21
Make panic threshold configurable for cluster

### LINKED ISSUE #4015: Panic threshold to be configurable
*Description*:
It would be good if the healthy_panic_threshold could be configurable via the Gateway API. Currently, it is only possible through EnvoyPatchPolicy.

see:
https://github.com/envoyproxy/gateway/discussions/3406

nezdolik:
feat(api): Make panic threshold configurable per backend/cluster
Fixes #4015 

Release Notes: TBD


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -355,6 +356,14 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(
 		err = perr.WithMessage(err, "ResponseOverride")
 		errs = errors.Join(errs, err)
 	}
+
+	if policy.Spec.CommonLbSettings != nil {
+		if lbSettings, err = buildCommonLbSettings(policy.Spec.CommonLbSettings); err != nil {
+			err = perr.WithMessage(err, "RateLimit")
```
nezdolik, line 362: oopsie

api/v1alpha1/shared_types.go
```diff
@@ -579,6 +579,22 @@ type ClusterSettings struct {
 	//
 	// +optional
 	HTTP2 *HTTP2Settings `json:"http2,omitempty"`
+
```
arkodg, line 582: my preference would be to add a `PanicThreshold unit32` field inside `HealthCheck`
https://github.com/envoyproxy/gateway/blob/0e52d06721737bdcbc9a1e8ffe969a4acd697a25/api/v1alpha1/healthcheck_types.go#L17

internal/ir/xds.go
```diff
@@ -2456,6 +2459,12 @@ func (h *HealthCheck) Validate() error {
 		}
 	}
 
+	if h.PanicThreshold != nil {
+		if *h.PanicThreshold < 0 || *h.PanicThreshold > 100 {
```
nezdolik, line 2463: realistically it cannot be `< 0` due to uint32 type

internal/ir/xds_test.go
```diff
@@ -1421,6 +1439,7 @@ func TestValidateHealthCheck(t *testing.T) {
 					},
 				},
 				&OutlierDetection{},
+				ptr.To[uint32](10),
```
nezdolik, line 1442: @arkodg not sure why it was needed to set default value in panic thresholds here and in all test below (since it is supposed to be an optional field). But otherwise it did not compile

internal/ir/xds.go
nezdolik:
```
Error: internal/ir/xds.go:2463:6: SA4003: no value of type uint32 is less than 0 (staticcheck)
		if *h.PanicThreshold < 0 || *h.PanicThreshold > 100 {
```
CI is smart :)

nezdolik:
there is a failure for conformance test ` --- FAIL: TestGatewayAPIConformance/HTTPRouteListenerPortMatching (72.08s)` with not much additional info

arkodg (approved):
LGTM thanks !
as a follow up can you add a test case in https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/xds-ir

shawnh2 (approved):
LGTM

internal/xds/translator/cluster.go
```diff
@@ -106,6 +107,13 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		PerConnectionBufferLimitBytes: buildBackandConnectionBufferLimitBytes(args.backendConnection),
 	}
 
+	// 50% is the Envoy default value for panic threshold. No need to explicitly set it in this case.
```
arkodg, line 110: why is this set twice - here and below ?

arkodg:
thanks @nezdolik can you also add a yaml test in https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/xds-ir and https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/out/xds-ir

internal/xds/translator/cluster.go
nezdolik: Good question, error while refactoring

arkodg (approved):
LGTM thanks !

zirain (approved):
Thanks, it would be better if you can add an e2e for this later.

Issue #4929, opened 2024-12-13
Add an option to configure HPA for the Envoy Gateway deployment in Helm


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Dean-Coakley:
Not sure I understand what is required here.

Would it just be a case of adding something like?:
```diff
diff --git a/charts/gateway-helm/values.tmpl.yaml b/charts/gateway-helm/values.tmpl.yaml
index cfcd9532..51517556 100644
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -80,6 +80,16 @@ config:
       controllerName: gateway.envoyproxy.io/gatewayclass-controller
     provider:
       type: Kubernetes
+      # envoyHpa:
+      #   minReplicas: 2
+      #   maxReplicas: 10
+      #   metrics:
+      #     - resource:
+      #         name: cpu
+      #         target:
+      #           averageUtilization: 60
+      #           type: Utilization
+      #       type: Resource
     logging:
       level:
         default: info
```
(Ref: https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/#customize-envoyproxy-horizontal-pod-autoscaler) 

And then I think those entries would get pulled in by the templating in https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/envoy-gateway-config.yaml 

Or am I misunderstanding?

arkodg:
the API would probably be at the top level https://github.com/envoyproxy/gateway/blob/fcdbafa5925482f705c2efd7ff8f54e348372d64/charts/gateway-helm/values.tmpl.yaml#L76 similar to the `deployment` and `service` field and would need a template similar to https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/envoy-gateway-deployment.yaml

Dean-Coakley:
Ah whoops, gotcha.

I've started working on this.


Based on the issue title "Add an option" I've put HPA behind a boolean.
The top level config looks like:
```
hpa:
  enable: false
  # minReplicas: 1
  maxReplicas: 1
  metrics: []
    # - resource:
    #     name: cpu
    #     target:
    #       averageUtilization: 80
    #       type: Utilization
    #   type: Resource
  behavior: {}
```

Does that look reasonable enough?

arkodg:
[...]

PR #5127, opened 2025-01-22
feat: support HPA in helm chart

Dean-Coakley:
**What type of PR is this?**

* Adds support for HorizontalPodAutoscaler support to helm chart

Fixes https://github.com/envoyproxy/gateway/issues/4929

Release Notes: Yes


arkodg:
thanks @Dean-Coakley this looks good ! can you add a test in https://github.com/envoyproxy/gateway/tree/main/test/helm

charts/gateway-helm/values.tmpl.yaml
```diff
@@ -74,6 +74,15 @@ deployment:
 service:
   annotations: {}
 
+hpa:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 1
+  targetCPU: 50
```
arkodg, line 81: thinking out loud, should there be a percentage suffix here ?

charts/gateway-helm/templates/envoy-gateway-hpa.yaml
```diff
@@ -0,0 +1,37 @@
...
+  {{- if .Values.hpa.minReplicas }}
+  minReplicas: {{ .Values.hpa.minReplicas }}
+  {{- end }}
+  maxReplicas: {{ .Values.hpa.maxReplicas }}
+  {{- if .Values.hpa.behavior }}
+  behavior:
+{{ toYaml .Values.hpa.behavior | indent 4 }}
+  {{- end }}
+  metrics:
+    {{- if .Values.hpa.targetCPU }}
```
guydc, line 21: Prefer to keep this entirely in values (similar to behavior) and not force end users to rely only on CPU/Mem. For example, custom metrics could be relevant here, e.g. size of the XDS cache, number of connected envoy proxies, etc.

charts/gateway-helm/values.tmpl.yaml
Dean-Coakley: Well manifests don't have/required it. I imagine you could, but it would be ineffectual.

Anyway, I think this will be removed in favour of: https://github.com/envoyproxy/gateway/pull/5127#discussion_r1927272320

arkodg: sure lets go with below comment

arkodg (dismissed):
LGTM thanks !

charts/gateway-helm/templates/envoy-gateway-hpa.yaml
```diff
@@ -0,0 +1,24 @@
...
+  namespace: '{{ $.Release.Namespace }}'
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: envoy-gateway
+  {{- if .Values.hpa.minReplicas }}
+  minReplicas: {{ .Values.hpa.minReplicas }}
+  {{- end }}
+  maxReplicas: {{ .Values.hpa.maxReplicas }}
```
sanposhiho, line 15:
```suggestion
  {{- if .Values.hpa.maxReplicas }}
  maxReplicas: {{ .Values.hpa.maxReplicas }}
  {{- end }}
```

missing?

Dean-Coakley: I can't seem to find the resource I read at the time but I did this because `maxReplicas` is a required value.

Just double checked:
```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: envoy-gateway
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: envoy-gateway
  # maxReplicas: 1
```


```
$ kubectl apply -f hpa.yaml 
The HorizontalPodAutoscaler "envoy-gateway" is invalid: 
* spec.maxReplicas: Invalid value: 0: must be greater than 0
```

I think it's possible to set a default value in gotmpl or document if you want.

Dean-Coakley: Or should I add the if as suggested and then I think helm should surface the error?

sanposhiho: can we use `required` then?
https://helm.sh/docs/howto/charts_tips_and_tricks/

sanposhiho: how do we implement other template values for required fields? (do we have such?)

Dean-Coakley: I don't see any other examples in this chart but yes that works. 
```shell
$ helm template .
Error: execution error at (gateway-helm/templates/envoy-gateway-hpa.yaml:15:18): .Values.hps.maxRetries entry is required

Use --debug flag to render out invalid YAML

```
And the `maxRetries` value is ignored if `hpa.enabled=false` as the template is not rendered.

zirain (approved):

arkodg:
Hey @Dean-Coakley can you rebase ? Let's prioritize getting this one in

arkodg (approved):

test/resilience/resilience_test.go
```diff
@@ -34,7 +34,7 @@ func TestResilience(t *testing.T) {
 		*flags.GatewayClassName,
 	)
 	if err != nil {
-		t.Fatalf("Failed to create the resillience test suit: %v", err)
+		t.Fatalf("Failed to create the resilience test suit: %v", err)
```
arkodg, line 37:
```suggestion
		t.Fatalf("Failed to create the resilience test suite: %v", err)
```

zirain (commented):
need changelog, otherwise LGTM.

arkodg:
Here's the link to the release notes https://github.com/envoyproxy/gateway/blob/main/release-notes/current.yaml

release-notes/current.yaml
```diff
@@ -14,6 +14,7 @@ new features: |
   Allow matchExpressions in TargetSelector
   Add defaulter for gateway-api resources loading from file to be able to set default values.
   Added support for defining Lua EnvoyExtensionPolicies
+  Added support for HorizontalPodAutoscaler to helm chart
```
Dean-Coakley, line 17: I assume this was just missed? I don't see any extra changes required here

zirain (approved):

release-notes/current.yaml
zirain: my bad, misreading.

arkodg (approved):

Issue #2629, opened 2024-02-16
Configure Envoy Proxy fleet in different namespaces


arkodg:
thinking out loud, potential API could look like
```
deploymentMode:
  watch:
    namespaces: .........
  deploy:
    envoy:
      namespaceType: GatewayNamespace  
```

* By default (when this field is unset), EG watches all resources in all namespaces, and depoys infra in the namespace where EG is running (`ControllerNamespace`)
* `watch.namespaces` can be used to set the specific namespaces to watch resources (Gateway, xRoute, Service etc)
* `deploy.envoy.namespaceType` can be used to specify to EG to deploy proxy infra to deploy envoy deployments in the namespace where the Gateway is created in. This option will need a helm knob too, to configure added permissions to be able to create deployments and services in other namespaces

_Originally posted by @arkodg in https://github.com/envoyproxy/gateway/issues/1117#issuecomment-1492387115_

cnvergence:
I would like to tackle this one

cnvergence:
It will also be necessary to move away from mTLS for control plane auth to sTLS + JWT Authn to validate the envoy.
cc @arkodg

arkodg:
yah @cnvergence, we can selectively do it for the case when the Gateway is running in a user namespace

kamalmarhubi:
I believe this would be solved once [GEP-1762: In Cluster Gateway Deployments](https://gateway-api.sigs.k8s.io/geps/gep-1762/) is implemented, which is tracked in #4330 as referenced just above. Or I know at least the use case that led me to this issue would be solved by it.

arkodg:
@kamalmarhubi can you share your use case ? 
we already support customizing gateway infrastructure https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/, the generated resources live in the `envoy-gateway-system` namespace, instead of the Gateway namespace

kamalmarhubi:
@arkodg thanks yes, I'm already customizing it! The reason I'd like the resources to be created in the Gateway namespace is that it allows cleaner RBAC policies. I'd like owners to be able to inspect their gateways' generated resources without necessarily giving them access to the gateway controller itself.

It also makes setting up network policies easier: with a default-deny setup, we need to add allow-egress policies in the `envoy-gateway-system` which seems messy.

Do you expect that Envoy Gateway will implement GEP-1762?

arkodg:
[...]

PR #5137, opened 2025-01-23
feat: support infra deployment in the gateway namespace

cnvergence:
**What type of PR is this?**

feat: support infra deployment in the gateway namespace

Enable deploying Infra envoy proxies in the namespace of related Gateway resources.
- Support `GatewayNamespaceMode` deploy mode in the gateway config map.
- Adjust infra resources to reflect non-default namespace.
- Set up the auth between xds-server and infra envoy proxy based on sTLS and JWT Auth, set up proper TLS credentials and interceptor in gRPC server.
- Update the related helm chart to set up necessary permissions for envoy gateway ServiceAccount.

JWT token are validated using k8s [TokenReview API ](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#schema-for-service-account-private-claims).

Fixes https://github.com/envoyproxy/gateway/issues/2629

Release Notes: Yes


zhaohuabing:
Hi @cnvergence Just checking in - any updates on this PR? Would love to see it land in 1.4.0.

cnvergence:
hey @zhaohuabing, getting back to this after Kubecon, I need to fix connection between control-plane and proxies, if everything will work again, I think it should be ready for 1.4.0 :)

zhaohuabing:
> hey @zhaohuabing, getting back to this after Kubecon, I need to fix connection between control-plane and proxies, if everything will work again, I think it should be ready for 1.4.0 :)

@cnvergence  a quick way to solve this issue is to copy the `envoy` secret into each namespace that contains a `Gateway` resource.  While not ideal, this would unblock some key use cases -- such as using EG as an Ambient Waypoint proxy, and allow us to get this PR into v1.4.0.

This is not the most optimal solution, since all Gateways share the same client cert. As the next step, we can have a piece of code in Envoy Gateway to automatically generate a unique cert secret per Gateway. This can be handled in a follow-up PR.

Would love your input on this, @envoyproxy/gateway-maintainers

arkodg:
> > hey @zhaohuabing, getting back to this after Kubecon, I need to fix connection between control-plane and proxies, if everything will work again, I think it should be ready for 1.4.0 :)
> 
> @cnvergence a quick way to solve this issue is to copy the `envoy` secret into each namespace that contains a `Gateway` resource. While not ideal, this would unblock some key use cases -- such as using EG as an Ambient Waypoint proxy, and allow us to get this PR into v1.4.0.
> 
> This is not the most optimal solution, since all Gateway infras share the same client cert. As the next step, we can introduce logic in Envoy Gateway to automatically generate a unique cert secret per Gateway. This can be handled in a follow-up PR.
> 
> Would love your input on this, @envoyproxy/gateway-maintainers

-1 to this, app teams have will have access to this, and can take advantage

zhaohuabing:
To move this forward, I agree that we can start without client cert and use the JWT token for client side validation for the namespaced mode, but please keep mTLS for the current mode.

We eventually will need mTLS for a better security gesture and compliance purpose. We may end up adding a xds proxy to exchange the client cert with the jwt token, and then upgrade the XDS connection from TLS to mTLS.

If @arkodg also feels this is the right approach, we can go ahead.

internal/xds/server/kubejwt/tokenreview.go
```diff
@@ -0,0 +1,55 @@
...
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+
+	return clientset, nil
+}
+
+func validateKubeJWT(ctx context.Context, clientset *kubernetes.Clientset, token string) (bool, error) {
+	tokenReview := &authenticationv1.TokenReview{
```
arkodg, line 35: are we validating that these token belongs to an envoy created by EG ?

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -284,34 +287,84 @@ func expectedShutdownPreStopCommand(cfg *egv1a1.ShutdownConfig) []string {
...
+		MountPath: "/certs",
+		ReadOnly:  true,
 	}
+	volumeMounts = append(volumeMounts, certsMount)
+
+	sdsMount := corev1.VolumeMount{
+		Name:      "sds",
+		MountPath: "/sds",
+	}
+	volumeMounts = append(volumeMounts, sdsMount)
```
zirain, line 303: Unnecessary changes?

internal/xds/server/kubejwt/tokenreview.go
cnvergence: Just by the service account, I was thinking about it, we will need to have an exact name of the envoy pod

arkodg: Can we leverage the IR keys in the snapshot cache which map to the fleet deployment name and namespace?

api/v1alpha1/envoygateway_helpers.go
```diff
@@ -96,6 +96,14 @@ func (e *EnvoyGateway) NamespaceMode() bool {
 		len(e.Provider.Kubernetes.Watch.Namespaces) > 0
 }
 
+// GatewayNamespaceMode returns true if controller uses gateway namespace mode for infra deployments.
+func (e *EnvoyGateway) GatewayNamespaceMode() bool {
```
zhaohuabing, line 100: Would prefer a clearer name to make it easier to read. For example: `IsInfraDeployedInGatewayNamespace`

internal/gatewayapi/translator.go
```diff
@@ -305,6 +308,10 @@ func (t *Translator) InitIRs(gateways []*GatewayContext) (map[string]*ir.Xds, ma
 		}
 
 		gwInfraIR.Proxy.Name = irKey
+		gwInfraIR.Proxy.Namespace = t.Namespace
```
zhaohuabing, line 311: Nit, unblocking: could we rename `t.Namespace`  to `t.ControllerNamespace` for better readability?

internal/xds/server/kubejwt/tokenreview.go
zhaohuabing: We can use `kubernetes.io.namespace` and `kubernetes.io.name` claims to verify the pod.

https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#schema-for-service-account-private-claims

zhaohuabing (changes requested):
Can we add an e2e test to verify the basic HTTP routing in the `GatewayNamespace` mode?

charts/gateway-helm/templates/namespaced-infra-manager-rbac.yaml
```diff
@@ -0,0 +1,66 @@
+{{ $watchedNamespaces := list }}
+{{ if .Values.config.envoyGateway.provider.kubernetes }}
+{{ $kube := .Values.config.envoyGateway.provider.kubernetes }}
+{{ if and ($kube.watch) ($kube.deploy) (eq $kube.deploy.type "GatewayNamespace") }}
```
zhaohuabing, line 4: Is `$kube.watch` mandatory? 

By default, EG should watch Gateways for all the namesapces.

internal/cmd/egctl/testdata/translate/out/quickstart.all.yaml
```diff
@@ -96,6 +96,7 @@ infraIR:
           gateway.envoyproxy.io/owning-gateway-name: eg
           gateway.envoyproxy.io/owning-gateway-namespace: envoy-gateway-system
       name: envoy-gateway-system/eg
+      namespace: ""
```
zhaohuabing, line 99: Is this correct?

internal/infrastructure/host/proxy_infra.go
```diff
@@ -74,7 +74,7 @@ func (i *Infra) CreateOrUpdateProxyInfra(ctx context.Context, infra *ir.Infra) e
 		StatsServerPort: ptr.To(int32(0)),
 	}
 
-	args, err := common.BuildProxyArgs(proxyInfra, proxyConfig.Spec.Shutdown, bootstrapConfigOptions, proxyName)
+	args, err := common.BuildProxyArgs(proxyInfra, proxyConfig.Spec.Shutdown, bootstrapConfigOptions, proxyName, i.EnvoyGateway.GatewayNamespaceMode())
```
zhaohuabing, line 77: I don't think the host mode works with the `GatewayNamespaceMode`. There's no JWT token in the host mode.

@shawnh2

charts/gateway-helm/templates/namespaced-infra-manager-rbac.yaml
cnvergence: I have added two RBAC modes to the helm chart:
- one is the default, when we have the GatewayNamespace Mode, it will create clusterrole for managing infra
- other is when we have selected just few namespaces to watch, it will setup only roles in the selected namespaces for managing infra

cnvergence:
> Can we add an e2e test to verify the basic HTTP routing in the `GatewayNamespace` mode?

I was planning to add it later on, but sure I can commit basic e2e test in this PR

zhaohuabing (changes requested):
Can we add a test to `internal/infrastructure/kubernetes/proxy/resource_provider_test.go` to verfiy the generated envoy deployment in `GatewayNamespace` mode?

internal/infrastructure/host/proxy_infra.go
zhaohuabing: This change is a bit misleading - we don't need the `GatewayNamespaceMode` parameter for host mode.

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -371,6 +411,24 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec) []corev
 			},
 		},
 	}
+	if gatewayNamespaceMode {
+		env = []corev1.EnvVar{
+			{
```
arkodg, line 416: curious whats this for ?

arkodg:
@cnvergence I prefer merging this into v1.4-rc.1 and dealing with the remaining piece before v1.4 releases. If we cant wrap it up, we can disable the feature, until SA authorization is complete

internal/xds/server/kubejwt/tokenreview.go
zhaohuabing: Tracked in this issue: https://github.com/envoyproxy/gateway/issues/5863

internal/infrastructure/kubernetes/proxy/resource.go
cnvergence: This is setting an Envoy-gateway namespace envVar, which defaults to metadata.namespace field of envoy infra pod.

zhaohuabing:
LGTM. Thanks for your patience!

cnvergence:
Thanks for the review, I will tackle open issues as the followups

internal/xds/server/kubejwt/tokenreview.go
cnvergence: I have added the basic check for deployment in snapshot cache, we will need to improve it and make it work with merged gateways mode

api/v1alpha1/envoygateway_helpers.go
cnvergence: I am okay with changing the name, however, I would like to keep a shorter name, just like `GatewayNamespaceMode`

zhaohuabing (approved):
LGTM. Thanks for the patience!

zirain (approved):

charts/gateway-helm/templates/infra-manager-rbac.yaml
```diff
@@ -1,3 +1,34 @@
+{{ if .Values.config.envoyGateway.provider.kubernetes }}
+{{ $kube := .Values.config.envoyGateway.provider.kubernetes }}
+{{ if and (not $kube.watch) ($kube.deploy) (eq $kube.deploy.type "GatewayNamespace") }}
```
zirain, line 3: I think we may need another controls for this, other wise it won't with following:
- install EG with default values with helm
- update configmap with kubectl

cnvergence: true, we might think how to add this rbac after installing

PR #5142, opened 2025-01-24
fix(translator): make OIDC and JWT authentication work together

### LINKED ISSUE #2496: Filters for OIDC authentication mechanisms
*Description*:

Currently, when OIDC and JWT authentication mechanisms are configured in the same `SecurityPolicy`, the OIDC is applied first. It ensures the presence of the bearer and refresh tokens in cookies, and adds the `Authorisation` header to the request. Then JWT is applied, validating the added header.

This setup works perfectly for browser requests. However, it prevents monitoring systems and regression tests, which provide a valid `Authorisation` header in the requests, from accessing the services on the same URLs. The OIDC mechanism kicks in first and redirects the requests to the login pages because of the missing cookies.

We suggest extending the configuration of, at least, the OIDC mechanism with the possibility of adding a header-based filter.
When implemented, the OIDC can be skipped if the `Authorization` header is present.

An example of the configuration:

```
  oidc:
    provider: {}
    clientID: ""
    scopes: []
    clientSecret: {}
    matches:
    - ifRequestHeader:
        name: Authorization
        valueRegex: Bearer.*
      onMatch: skip
```

Envoy Proxy supports Composite Filters, which support such decisions and therefore can be used to implement this functionality.

*Relevant Links*:

* The [Combining OIDC and JWT authentication](https://github.com/envoyproxy/gateway/discussions/2425) discussion
* Envoy [Composite Filters](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/composite_filter)

StephenRobin:
**What type of PR is this?**

Currently, when both OIDC and JWT authentication mechanisms are configured in
the same SecurityPolicy, the OIDC is applied first. It ensures the presence of
the tokens in the request's cookies, and adds the Authorisation header to the
request. Then JWT is applied, validating the added header.

This setup works perfectly for browser requests. However, it blocks requests
from clients that provide a valid "Authorization: Bearer..." header (normally
non-browser clients). The OIDC mechanism kicks in first and redirects the
requests to the login pages because of the missing cookies.

Envoy Proxy has a pass_through_matcher option that can be used to skip over the
OIDC filter when the request is going to be handled by the JWT filter later.
Add a new option passThroughAuthHeader on the OIDC config that will add this
option in xDS.

It would be an error to have the new passThroughAuthHeader enabled without also
having JWT config that will later check the validity of the token. If this is
not the case, the SecurityPolicy is rejected for safety.

Note that this change is designed to be entirely backwards compatible. The
system behaviour is unchanged unless the new option is switched on.

Fixes #2496

Release Notes: Yes


arkodg:
thanks @StephenRobin this looks good !
can you add a test in https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata & https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/xds-ir ?

StephenRobin:
Hi @arkodg, thanks, hadn't noticed all those tests. I have added some. Coupled with the unit tests, I believe this should give  good coverage, let me know if you want me to expand any of them.

arkodg (approved):
LGTM thanks !

arkodg:
cc @denniskniep

zhaohuabing (dismissed):
LGTM thanks!

internal/xds/translator/oidc.go
```diff
@@ -228,9 +230,48 @@ func oauth2Config(oidc *ir.OIDC) (*oauth2v3.OAuth2, error) {
...
+
 	return oauth2, nil
 }
 
+func buildHeaderMatchers(jwt *ir.JWT) []*routev3.HeaderMatcher {
+	// Bypass OIDC if a header that will be handled by JWT is passed.
+	passThroughMatchers := make([]*routev3.HeaderMatcher, 0, len(jwt.Providers))
+
+	for _, provider := range jwt.Providers {
+		if provider.ExtractFrom == nil {
```
denniskniep, line 246: Shouldn´t we do here:
`provider.ExtractFrom == nil || provider.ExtractFrom.Headers == nil`

Because ExtractFrom might contain the id-token-cookie from the OIDC Filter

StephenRobin:
Hi @arkodg Looks like feedback is positive - how do we progress this PR? Do you need anything further from me?

internal/xds/translator/oidc.go
arkodg: looks like this has been addressed in the preceding Gateway API layer
```

		hasValidJwtExtractor := false
		for _, provider := range jwt.Providers {
			// When ExtractFrom is not specified it falls back to looking at the "Authorization: Bearer ..." header
			if provider.ExtractFrom == nil || len(provider.ExtractFrom.Headers) > 0 {
				hasValidJwtExtractor = true
				break
			}
		}
		if !hasValidJwtExtractor {
			return errors.New("the OIDC.PassThroughAuthHeader setting must be used in conjunction with a JWT provider that is configured to read from a header")
		}
```

arkodg: oh nvm, I understand Dennis's question here, `if provider.ExtractFrom == nil` will not kick in if there's a cookie containing a JWT that has an authentication rule
so maybe lets add
```
provider.ExtractFrom == nil || provider.ExtractFrom.Headers == nil
```

to solve this @StephenRobin

zhaohuabing:
Hi @StephenRobin Can we test this in the e2e tests?

A test function can be added to https://github.com/envoyproxy/gateway/blob/main/test/e2e/tests/oidc.go to test a request with JWT token will not be redirected to the keycloak when `passThroughAuthHeader` is set to true.

thiagogcm:
Thanks for this! It is very helpful for me too
is it expected for next release?

zhaohuabing:
ping @StephenRobin

StephenRobin:
Hi @zhaohuabing. Sorry for the delay here, I've been busy with other things. I have added the E2E test as requested. Let me know if there are any further barriers to getting this merged.

api/v1alpha1/oidc_types.go
```diff
@@ -102,6 +102,14 @@ type OIDC struct {
 	// Note: this field is only applicable when the "refreshToken" field is set to true.
 	// +optional
 	DefaultRefreshTokenTTL *metav1.Duration `json:"defaultRefreshTokenTTL,omitempty"`
+
+	// Skips OIDC authentication when the request contains any header that will be extracted by the JWT
+	// filter, normally "Authorization: Bearer ...". This is typically used for non-browser clients that
+	// may not be able to handle OIDC redirects and wish to directly supply a token instead.
+	//
```
zhaohuabing, line 109: Nit - could we add a line in the comment to mention that JWT must be specified if this field is true, and why?

zhaohuabing:
@StephenRobin This PR looks good! I just have a minor comment on the API doc.

api/v1alpha1/oidc_types.go
StephenRobin: Thanks for the review! You actually don't need to specify the JWT header extractor. If it's not specified it looks for the "Authorization: Bearer ..." header.

https://gateway.envoyproxy.io/docs/api/extension_types/#jwtprovider

"ExtractFrom defines different ways to extract the JWT token from HTTP request. If empty, it defaults to extract JWT token from the Authorization HTTP request header using Bearer schema or access_token from query parameters."

StephenRobin: I have updated the comment, hopefully it is clearer now.

arkodg:
@StephenRobin @zhaohuabing I admit, I was the one pushing for a `bool` option, but there have been other users who raised issues around skipping OIDC if an API Key was present in the header (can't find the GH issue or slack thread)

Should we stick to a header based pass through approach ?

zhaohuabing:
Hi @arkodg Is this one that you were referring to? https://github.com/envoyproxy/gateway/pull/5455

zhaohuabing:
> > Hi @arkodg Is this the one that you were referring to? #5455
> 
> no, not referring to that

Oh I see, OIDC can be skipped if either a JWT token or an API key is present in the request header.

StephenRobin:
> OIDC can be skipped if either a JWT token or an API key is present in the request header, and we may have other similar use cases. Explicitly setting the skipped headers seems the most flexible way to handle this.

Definitely the most flexible, but perhaps not the most intuitive. Especially if the full complexity of the underlying pass_through_matcher in Envoy is to be replicated here.

The name of the bool `passThroughAuthHeader` is not specific to JWT. This change could be merged now and a subsequent PR could change the implementation to also pass through headers listed in the extractFrom option of API Key Auth, just as it passes through headers listed in the extractFrom of the JWT Provider.

arkodg:
sure then lets continue with this

StephenRobin:
Thanks to everyone who's helped me with this. What's the procedure for getting it merged - will one of the maintainers pick it up?

arkodg:
@StephenRobin requested reviewers to take another look, once approved, this can be merged, also seeing some conflicts, can you rebase ?

internal/xds/translator/oidc.go
zhaohuabing: I may not understand this correctly, but if we use `provider.ExtractFrom == nil || provider.ExtractFrom.Headers == nil` here,  and a JWT token cookie is in the request, `Authorization` will be added to the passthroughMatchers, which shouldn't because `Authorization` is only used by default when `provider.ExtractFrom == nil`.

zhaohuabing (dismissed):
Loos good to me, as long as I'm not misunderstanding [this part](https://github.com/envoyproxy/gateway/pull/5142#discussion_r2092173636). :-)

zhaohuabing (dismissed):

arkodg (dismissed):

arkodg:
@StephenRobin can you run `make testdata && make generate` and commit those changes ?

arkodg (approved):

itmwiw:
Awesome feature! Do you have an estimated release date? I noticed it's not included in version 1.4.1 yet.

arkodg:
@itmwiw should be part of v1.5 (end July)

Issue #5143, opened 2025-01-24
Support Cross-Namespace Policy attachment


guydc:
*Description*:

@dprotaso [writes](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1737742853191229): 
> anyone know off hand if an `EnvoyExtensionPolicy` can match all HTTPRoutes.
> 
> I've tried an empty label selector
>   targetSelectors:
>   - group: gateway.networking.k8s.io
>     kind: HTTPRoute
>     matchLabels: {} # Test all HTTPRoutes
>
>  [...]
> 
> I'm just trying to match all routes right now in the cluster. [...] I'm sidecaring an ext-proc server with the envoy proxy. But repeating the boilerplate config for each namespace is a bit much - do you recommend an alternative. I sorta consider this config as part of the 'gateway' deployment

GW-API currently only support Local Policy References in TargetRef(s). 

To support scalable cross-namespace policy attachment, EG can introduce a `NamespaceSelector` as part of `TargetSelectors` (A custom EG policy selection mechanism), specifying which namespaces should be considered for policy attachment. 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

akevdmeer:
I'm interested in this; can I ask what is the status? Is cross-namespace policy attachment expected to be delivered in 1.8?

zhaohuabing:
Close this as it's been implemented in #8676.

PR #5145, opened 2025-01-24
api: cross-namespace policy target selectors

guydc:
**What type of PR is this?**

- Adds the option to select targets across all namespaces 
- Inspired by https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.RouteNamespaces
- In the future, we can extend this mechanism to support:
  - List of named namespaces
  - Label selector to select namespaces  

Relates to #5143

Release Notes: Yes/No


api/v1alpha1/policy_helpers.go
```diff
@@ -38,6 +38,29 @@ type TargetSelector struct {
...
+	// +optional
+	// +notImplementedHide
+	Namespaces *TargetSelectorNamespaces `json:"namespaces,omitempty"`
+}
+
+type FromNamespaces string
+
+const (
+	// FromNamespacesAll indicates that the target selector should apply to targets from all namespaces
+	FromNamespacesAll FromNamespaces = "All"
```
zhaohuabing, line 55: Should we have a `same` defined here?

Issue #5147, opened 2025-01-25
Mistake in a BackendTrafficPolicy causes all routes to return 404


dghubble:
*Description*:

A colleage and I found that a subtle mistake in a single `BackendTrafficPolicy` can make envoy proxy instances return 404's for ALL routes.

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy
metadata:
  name: hellogo
  namespace: default
spec:
  targetRefs:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
      name: hello
  retry:
    numRetries: 1
    perRetry:
      backOff:
        baseInterval: 0s      # 0s breaks everything, 1s is ok
```

*Repro steps*:

Create a BackendTrafficPolicy as shown above. Nothing stops a developer setting `baseInterval: 0s`.

At first, nothing is wrong. Then, if you restart envoy proxies, you'll find ALL httproutes return 404s immediately.  Logs show `route_not_found` for all requests but no mention of why or which resources causs this. Inspecting the raw envoy config via the admin portal, the `dynamic_route_configs` section is never generated (usually its populated).

To find the offending resource, we had to delete resources until discovering the problematic thing was this one `BackendTrafficPolicy` and this one value within it. Pretty scary to us. Questions:

* What values should be allowed in `baseInterval`?
* What validations can be done to stop misconfigurations like this?
* Supposing there are other (perhaps future) resource misconfigs / validation issues, how can those be scoped to avoid breaking all routes?
* How can a user identify the problematic resources, either in envoy-gateway or the envoy proxy? Here we had to guess and test

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

envoy-gateway: v1.2.5

*Logs*:
>Include the access logs and the Envoy logs.

arkodg:
looks like this one managed to escape all the checks, here's the error from the envoy proxy
```
[...]

PR #5148, opened 2025-01-25
validate all xds resources before returning the translation result

zhaohuabing:
This PR enforces validation for all generated xDS resources before sending them to the Envoy fleet.

Related to: #5147 
Release Notes: No

```
2025-01-25T05:07:21.922Z	ERROR	xds-translator	runner/runner.go:85	failed to translate xds ir	{"runner": "xds-translator", "error": "invalid RouteConfiguration.VirtualHosts[0]: embedded message failed validation | caused by: invalid VirtualHost.Routes[0]: embedded message failed validation | caused by: invalid Route.Route: embedded message failed validation | caused by: invalid RouteAction.RetryPolicy: embedded message failed validation | caused by: invalid RetryPolicy.RetryBackOff: embedded message failed validation | caused by: invalid RetryPolicy_RetryBackOff.BaseInterval: value must be greater than 0s"}
```


internal/xds/translator/translator.go
```diff
@@ -131,6 +131,12 @@ func (t *Translator) Translate(xdsIR *ir.Xds) (*types.ResourceVersionTable, erro
 		}
 	}
 
+	// Validate all the xds resources in the table before returning
+	// This is necessary to catch any misconfigurations that might have been missed during translation
+	if err := tCtx.ValidateAll(); err != nil {
```
zhaohuabing, line 136: This is the key change, the others are minor refactors around the proto validation.

internal/xds/translator/basicauth.go
```diff
@@ -82,10 +82,7 @@ func buildHCMBasicAuthFilter(basicAuth *ir.BasicAuth) (*hcmv3.HttpFilter, error)
 			},
 		},
 	}
-	if err = basicAuthProto.ValidateAll(); err != nil {
```
zhaohuabing, line 85: Validation in individual filer translators are not required anymore, as they'll be validated inside the `proto.ToAnyWithValidation` function before converting to `any`.

guydc:
should we CP to 1.3.0 ?

arkodg:
@guydc , Haubing will be away this week, and we won't be able to get this not v1.3.0, we could get it into a patch release.
for now / v1.3.0 #5147 can we fixed with a CEL validation, I'll work on it to get it into v1.3.0

zhaohuabing:
@arkodg @guydc could you please take another look at this PR? Thanks!

internal/xds/translator/translator.go
arkodg: so are we validating twice ?

arkodg: because there's already a `proto.Validate` in resourceversiontable.go

zhaohuabing: Yes, this is the final check to ensure no validations are missed, it's a safe guard.

We may still want to retain the original validation to fail earlier for proper error handling.

arkodg: Can we do with just 1 ?

arkodg: Which we do in the end ?

zhaohuabing: It's common that [a xds resource is modified after being  added to ResourceVersionTable](https://github.com/envoyproxy/gateway/issues/5147#issuecomment-2613776366), and we don't have a good way to prevent this from happening again in future code changes, so I prefer to add a final check in the translator.

arkodg: can we then keep this one and rm the one in `AddXdsResource` ?

zhaohuabing: > can we then keep this one and rm the one in AddXdsResource ?

I think it's reasonable to move the validation out of the AddXdsResource and into a dedicate method, as the current validation in AddXdsResource is implicit and could be easily ignored.

Would you mind if I address this in a follow-up PR? AddXdsResource method has been used everywhere, so the modification would make this scope of this PR a bit too large, and this is not a blocking issue for this PR.

arkodg (approved):

internal/xds/types/resourceversiontable.go
```diff
@@ -150,6 +93,21 @@ func (t *ResourceVersionTable) AddXdsResource(rType resourcev3.Type, xdsResource
...
 }
 
+// ValidateAll validates all the xds resources in the ResourceVersionTable
+func (t *ResourceVersionTable) ValidateAll() error {
+	var errs error
+
+	for _, xdsResource := range t.XdsResources {
+		for _, resource := range xdsResource {
+			if err := proto.Validate(resource); err != nil {
+				errs = errors.Join(errs, err)
```
arkodg, line 103: duplicate lines

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5171, opened 2025-01-28
feat: implement Lua EnvoyExtensionPolicy

rudrakhp:
**What type of PR is this?**

 feat: implement Lua feature in EnvoyExtensionPolicy

Implement API introduced in #4932

Related #4627 

Release Notes: No


rudrakhp:
@arkodg @zhaohuabing Wanted to get an initial buy in regarding the approach while I add more unit tests and E2E, please do review. Thanks!

arkodg:
hey @rudrakhp the code looks great ! 

please feel free to continue with unit tests and e2e

api/v1alpha1/lua_types.go
```diff
@@ -41,8 +41,8 @@ type Lua struct {
 	// The value of key `lua` in the ConfigMap will be used.
 	// If the key is not found, the first value in the ConfigMap will be used.
 	//
-	// +kubebuilder:validation:XValidation:rule="self.kind == 'ConfigMap' && (!has(self.group) || self.group == '')",message="Only a reference to an object of kind ConfigMap belonging to default core API group is supported."
+	// +kubebuilder:validation:XValidation:rule="self.kind == 'ConfigMap' && (!has(self.group) || self.group == '')",message="Only a reference to an object of kind ConfigMap belonging to default API group is supported."
 	// +optional
 	// +unionMember
-	ValueRef *gwapiv1.LocalObjectReference `json:"valueRef,omitempty"`
+	ValueRef *gwapiv1.SecretObjectReference `json:"valueRef,omitempty"`
```
arkodg, line 47: any reason why this is secret ?

rudrakhp: `SecretObjectReference` is not only for `Secret`. It's for referring to any K8s object, by default it's Secret type. Here is it's documentation:
```
// SecretObjectReference identifies an API object including its namespace,
// defaulting to Secret.
//
// The API object must be valid in the cluster; the Group and Kind must
// be registered in the cluster for this reference to be valid.
//
// References to objects with invalid Group and Kind are not valid, and must
// be rejected by the implementation, with appropriate Conditions set
// on the containing object.
```

rudrakhp: Reason for preferring this is to reuse the function for processing config maps:
https://github.com/envoyproxy/gateway/blob/54ca51be82188c844d6674bd7a66e90f0e23928b/internal/provider/kubernetes/controller.go#L753-L764

arkodg: check out the code flow for https://github.com/envoyproxy/gateway/blob/36fdc0be354807a5bbf8bb4e756653bbe9583b10/api/v1alpha1/shared_types.go#L746

rudrakhp: @arkodg will do. One additional caveat here is `LocalObjectReference` let's you locate resources in same namespace only, do we want to have that restriction?
```
// LocalObjectReference contains enough information to let you locate the
// referenced object inside the same namespace.
```

arkodg: yes prefer that for now

rudrakhp: @arkodg done!

zhaohuabing (changes requested):
Hi @rudrakhp we need to set the default order for the Lua filters in the filter chain, similart to what have been done to the Wasm and other filters.

https://github.com/envoyproxy/gateway/blob/c4920899659e75ceb342df02665e1b484eb0ecbc/internal/xds/translator/httpfilters.go#L121

api/v1alpha1/lua_types.go
```diff
@@ -41,7 +41,7 @@ type Lua struct {
 	// The value of key `lua` in the ConfigMap will be used.
 	// If the key is not found, the first value in the ConfigMap will be used.
 	//
-	// +kubebuilder:validation:XValidation:rule="self.kind == 'ConfigMap' && (!has(self.group) || self.group == '')",message="Only a reference to an object of kind ConfigMap belonging to default core API group is supported."
+	// +kubebuilder:validation:XValidation:rule="self.kind == 'ConfigMap' && (self.group == 'v1' || self.group == '')",message="Only a reference to an object of kind ConfigMap belonging to default v1 API group is supported."
 	// +optional
 	// +unionMember
```
zhaohuabing, line 46: Can we add CEL validations to ensure that `inline` must be configured for `inline` type and `valueRef` must be configured for `valueRef` type?

Similar to: https://github.com/envoyproxy/gateway/blob/c4920899659e75ceb342df02665e1b484eb0ecbc/api/v1alpha1/shared_types.go#L730-L731

internal/ir/xds.go
```diff
@@ -2811,6 +2813,16 @@ type ExtProc struct {
...
 }
 
+// Lua holds the information associated with Lua extensions
+// +k8s:deepcopy-gen=true
+type Lua struct {
+	// Name is a unique name for the LUa configuration.
+	// The xds translator only generates one Lua filter for each unique name
+	Name string
+	// Body is the Lua source body
+	Body *string
```
zhaohuabing, line 2823: Nit, unblocking: `code` or `script` sounds more natural to me.

zhaohuabing (commented):
Lua filter has built-in perRoute support, so we won't need to create a seperate lua filter for each route.

I think it's fine to keep the current implmentation as it's a generic approach to support configuration per route. If we find any issues with the multple-filter approach, we can then switch to the built-in perRoute approach.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/lua/v3/lua.proto#envoy-v3-api-msg-extensions-filters-http-lua-v3-luaperroute

rudrakhp:
> gateway/internal/xds/translator/httpfilters.go

Thanks for the comments @zhaohuabing, wanted to understand the rationale behind the ordering of the filters. Although the comments explained the high level thought process, didn't get why there is a huge gap of order between ExtProc and WASM, considering these two (along with Lua) are similar filters. IMO it might make sense to execute Lua first as they are light weight compared to ExtProc or WASM, WDYT?

rudrakhp:
>Lua filter has built-in perRoute support, so we won't need to create a seperate lua filter for each route.

Yes I started with LuaPerRoute filter but finally followed the generic framework we have for WASM and ExtProc, adding all Lua filters and enabling them as required per route.

zhaohuabing:
> IMO it makes sense to execute Lua first as they are light weight compared to ExtProc or WASM, WDYT?

Agree. This order is adjustable if users want to change the default order in the filter chain.

zhaohuabing (changes requested):
Can you please also [inclue lua filter to the `EnvoyFilter` API CEL rule](https://github.com/envoyproxy/gateway/blob/662d3653d0e46847239e9454aa5d8592f2947819/api/v1alpha1/envoyproxy_types.go#L198) and [update the API doc for `FilterOrder`](https://github.com/envoyproxy/gateway/blob/662d3653d0e46847239e9454aa5d8592f2947819/api/v1alpha1/envoyproxy_types.go#L96-L131) ? Thanks!

api/v1alpha1/lua_types.go
rudrakhp: This is already there: https://github.com/envoyproxy/gateway/blob/c1e18b29c41f7cdeec7e50ea859d7101356f78af/api/v1alpha1/lua_types.go#L25

Xunzhuo (approved):
Thanks for working on it! Hope to have a user-guide docs for a following PR.

Issue #5144, opened 2025-01-24
Add a dedicated listener port for envoy proxy readiness


arkodg:
*Description*:
>Describe the issue.

Currently the envoy proxy readiness port is `19001` and is shared with prometheus. This has two issues
* there's no clean way to change the prometheus port w/o affecting the readiness port https://github.com/envoyproxy/gateway/issues/1703
* the readiness port cannot be separated from the prometheus port and be converted into a dynamic xds listener to add more accuracy into the  pod readiness https://github.com/envoyproxy/gateway/issues/2810#issuecomment-2583516275

If this proposal is accepted, a another port `19002` would be added and dedicated for readiness

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
ptal @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

shawnh2:
+1, 19001 should be readiness port, 19002 can be used for prometheus

arkodg:
@shawnh2 moving prom to 19002 will be a breaking change for existing users, so prefer keeping prom on 19001, and readiness on 19002, readiness port is used internally ideally only by kubelet so it shouldnt affect users

shawnh2:
> @shawnh2 moving prom to 19002 will be a breaking change for existing users, so prefer keeping prom on 19001, and readiness on 19002, readiness port is used internally ideally only by kubelet so it shouldnt affect users

SGTM

zirain:
I can take this, as I have finished part of the work.

PR #5193, opened 2025-02-03
api: make readiness and prometheus port configurable

zirain:
xref: https://github.com/envoyproxy/gateway/issues/5144


arkodg:
hey @zirain can we do this in 2 steps, where we create a fixed dedicated port for readiness in envoy via https://github.com/envoyproxy/gateway/issues/5144 and then make it configurable via an API ?

PR #5197, opened 2025-02-04
xds: use a dedicated listener port for envoy proxy readiness

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5144


internal/infrastructure/kubernetes/proxy/testdata/daemonsets/disable-prometheus.yaml
```diff
@@ -78,19 +78,11 @@ spec:
                 - name: envoy.filters.network.http_connection_manager
```
guydc, line 78: should this listener exist if prometheus is disabled (test case) and ready listener is moved to dynamic listeners?

guydc (commented):
Do we want to make this feature available under some sort of feature gate? It changes behavior, and I'm concerned that in some unforeseen situation, the previous behavior would be better.

internal/xds/bootstrap/bootstrap.go
```diff
@@ -42,10 +42,9 @@ const (
 	// DefaultWasmServerPort is the default listening port of the wasm HTTP server.
 	wasmServerPort = 18002
 
-	envoyReadinessAddressv4 = "0.0.0.0"
-	envoyReadinessAddressv6 = "::"
+	EnvoyStatsPort = 19001
 
-	EnvoyReadinessPort = 19001
+	EnvoyReadinessPort = 19003
```
arkodg, line 47: should this be 19002 or 19003 ?

tools/make/kube.mk
```diff
@@ -187,6 +187,9 @@ else
 		--run-test $(E2E_RUN_TEST)
 endif
 
+run-e2e-upgrade:
```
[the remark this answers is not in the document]
zirain: this's a new target for test upgrade only.

internal/xds/bootstrap/bootstrap.go
zirain: 19002 is used by shutdown manager

internal/xds/translator/translator.go
```diff
@@ -89,6 +89,10 @@ func (t *Translator) Translate(xdsIR *ir.Xds) (*types.ResourceVersionTable, erro
 	// to collect all errors and reflect them in the status of the CRDs.
 	var errs error
 
+	if err := t.processHTTPReadyListenerXdsTranslation(tCtx, xdsIR.ReadyListener); err != nil {
+		errs = errors.Join(errs, err)
+	}
+
 	if err := t.processHTTPListenerXdsTranslation(
```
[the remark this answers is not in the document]
zirain: there's a page told the port used by EG, so I assume it's fine.
Next, we will support user customization of these ports.

internal/xds/translator/translator.go
```diff
@@ -89,6 +89,10 @@ func (t *Translator) Translate(xdsIR *ir.Xds) (*types.ResourceVersionTable, erro
 	// to collect all errors and reflect them in the status of the CRDs.
 	var errs error
 
+	if err := t.processHTTPReadyListenerXdsTranslation(tCtx, xdsIR.ReadyListener); err != nil {
```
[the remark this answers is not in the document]
zirain: do you mean return error here directly?

guydc (dismissed):
I would've preferred some feature gate here, but I think that users can revert to the old behavior using bootstrap and deployment patches if something goes wrong. 
Other than that, LGTM.

internal/xds/translator/listener_ready.go
```diff
@@ -0,0 +1,104 @@
...
+	hcmFilters = append(hcmFilters, router)
+
+	hcm := &hcmv3.HttpConnectionManager{
+		StatPrefix: "eg-ready-http",
+		RouteSpecifier: &hcmv3.HttpConnectionManager_RouteConfig{
+			RouteConfig: &routev3.RouteConfiguration{
+				Name: "ready_route",
+				VirtualHosts: []*routev3.VirtualHost{
+					{
+						Name:    "ready_route",
```
arkodg, line 49: is the route needed ? my worry is, the response is not based off health

zirain: we cannot leave route_config nil, or I missed something?

arkodg: can we return `500` in the DR then ? wanna make sure only the good case works

zirain: updated, hope CI happy.

guydc (approved):

arkodg (approved):
LGTM thanks !

Issue #5188, opened 2025-02-02
extension-server doesn't build


dprotaso:
Make image gives a weird error because of  controller-gen setup
```sh
% make image
cd tools/src/controller-gen && GOOS= GOARCH= go build -o /Users/dave/work/envoy-gateway/examples/extension-server/tools/bin/controller-gen $(sed -En 's,^import _ "(.*)".*,\1,p' pin.go)
/Users/dave/go/pkg/mod/golang.org/x/net@v0.34.0/idna/idna10.0.0.go:25:2: missing go.sum entry for module providing package golang.org/x/text/secure/bidirule (imported by golang.org/x/net/idna); to add:
        go get golang.org/x/net/idna@v0.34.0
/Users/dave/go/pkg/mod/golang.org/x/net@v0.34.0/idna/idna10.0.0.go:26:2: missing go.sum entry for module providing package golang.org/x/text/unicode/bidi (imported by golang.org/x/net/idna); to add:
        go get golang.org/x/net/idna@v0.34.0
/Users/dave/go/pkg/mod/golang.org/x/net@v0.34.0/idna/idna10.0.0.go:27:2: missing go.sum entry for module providing package golang.org/x/text/unicode/norm (imported by golang.org/x/net/idna); to add:
        go get golang.org/x/net/idna@v0.34.0
/Users/dave/go/pkg/mod/sigs.k8s.io/controller-tools@v0.15.0/pkg/loader/loader.go:31:2: missing go.sum entry for module providing package golang.org/x/tools/go/packages (imported by sigs.k8s.io/controller-tools/pkg/genall); to add:
        go get sigs.k8s.io/controller-tools/pkg/genall@v0.15.0
/Users/dave/go/pkg/mod/github.com/mattn/go-isatty@v0.0.20/isatty_bsd.go:8:8: missing go.sum entry for module providing package golang.org/x/sys/unix (imported by github.com/mattn/go-isatty); to add:
        go get github.com/mattn/go-isatty@v0.0.20
make: *** [tools/bin/controller-gen] Error 1
```

dprotaso:
I went into `examples/externsion-server/tools/src/controller-gen` and ran the `go get` commands and it worked - but I'm wondering if there is a bigger maintenance issue here. eg. maybe other make targets for examples don't work etc.

dprotaso:
Also seems the example doesn't work - the delta config returned can't be applied

```
[...]

PR #5199, opened 2025-02-04
fix build extension-server

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5188


examples/extension-server/Makefile
```diff
@@ -6,6 +6,14 @@ tools/controller-gen     = $(tools.bindir)/controller-gen
 $(tools.bindir)/%: $(tools.srcdir)/%/pin.go $(tools.srcdir)/%/go.mod
 	cd $(<D) && GOOS= GOARCH= go build -o $(abspath $@) $$(sed -En 's,^import _ "(.*)".*,\1,p' pin.go)
 
+IMAGE_PREFIX ?= envoyproxy/gateway-
+APP_NAME ?= extension-server
+TAG ?= latest
+
+.PHONY: docker-buildx
+docker-buildx:
```
dprotaso, line 14: While you're here - unsure if you want to add multi-platform support.

I noticed on a mac arm machine the binary in the image doesn't match the default arch.

You could use `ko` to simplify the build https://ko.build/ - and you won't require docker as a tool here

zirain: this's just an example, and we won't push it to docker.io, so I won't spend time on supporting multi-platform.

if you're interesting, feel free to send PR in the future.

dprotaso (approved):

examples/extension-server/Makefile
arkodg: Multi arch scripts love in https://github.com/envoyproxy/gateway/blob/main/tools/make/image.mk in case they need to be perused for extension-server

guydc (approved):

arkodg (approved):

PR #5211, opened 2025-02-05
feat: add per_host_thresholds in CircuitBreakers

### LINKED ISSUE #5210: Support per_host_thresholds for Circuit Breakers
*Description*:
would be used.

I'd like to configure Circuit Breakers under `per_host_thresholds` so I can set connection limits per-host.

[optional *Relevant Links*:]
- Envoy docs: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking
- Envoy proto: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto

jukie:
Adds support for configuring per-host circuit breaker limits

Fixes #5210

Release Notes: Yes


api/v1alpha1/circuitbreaker_types.go
```diff
@@ -46,4 +46,20 @@ type CircuitBreaker struct {
 	// +kubebuilder:validation:Maximum=4294967295
 	// +optional
 	MaxRequestsPerConnection *int64 `json:"maxRequestsPerConnection,omitempty"`
+
+	// Defines per-host Circuit Breaker thresholds
+	// +optional
+	// +notImplementedHide
+	PerHostThresholds *PerHostCircuitBreakers `json:"perHostThresholds,omitempty"`
```
jukie, line 53: Should this be a separate field or something like `maxConnectionsPerHost` since that's the only field currently supported. In the future could add `maxParallelRequestsPerHost`, etc as more support is added.

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -46,4 +46,20 @@ type CircuitBreaker struct {
 	// +kubebuilder:validation:Maximum=4294967295
 	// +optional
 	MaxRequestsPerConnection *int64 `json:"maxRequestsPerConnection,omitempty"`
+
+	// Defines per-host Circuit Breaker thresholds
```
zirain, line 50:
```suggestion
	// PerHostThresholds defines per-host Circuit Breaker thresholds
```

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -46,4 +46,20 @@ type CircuitBreaker struct {
...
+	// +optional
+	// +notImplementedHide
+	PerHostThresholds *PerHostCircuitBreakers `json:"perHostThresholds,omitempty"`
+}
+
+// PerHostCircuitBreakers defines the per-host Circuit Breaker configuration.
+type PerHostCircuitBreakers struct {
```
zhaohuabing, line 57: Nit: 
It looks to me the field name `PerHostThresholds` and the type name`PerHostCircuitBreakers` are misaligned.

How about?
1. `PerHost` and  `PerHostThresholds`
2. `PerHost` and  `PerHostCircuitBreakers`

api/v1alpha1/circuitbreaker_types.go
zhaohuabing: A addtional layer seems more future-proof.

api/v1alpha1/circuitbreaker_types.go
jukie: I prefer option 2 and updated with that

zhaohuabing (dismissed):
LGTM. Thanks!

jukie:
Went ahead and added the implementation here as well

internal/ir/xds.go
```diff
@@ -2398,6 +2398,16 @@ type CircuitBreaker struct {
...
+
+	// PerHost defines per-host Circuit Breakers
+	PerHost *PerHostCircuitBreakers `json:"perHost,omitempty"`
+}
+
+// PerHostCircuitBreakers defines the per-host Circuit Breaker configuration.
+// +k8s:deepcopy-gen=true
+type PerHostCircuitBreakers struct {
+	// MaxConnections configures the maximum number of connections that Envoy will establish per-host to the referenced backend defined within a xRoute rule.
+	MaxConnections *uint32 `json:"maxConnections,omitempty"`
```
zirain, line 2410: nit? there's a default value in API, so I think we can use non point here, will that make thing easier?

jukie: Sure, should that also be changed for the non-PerHost CircuitBreaker field? I was aiming to align with the existing fields

zirain: we can do the refactor later.

zirain (approved):

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -46,4 +46,20 @@ type CircuitBreaker struct {
 	// +kubebuilder:validation:Maximum=4294967295
 	// +optional
 	MaxRequestsPerConnection *int64 `json:"maxRequestsPerConnection,omitempty"`
+
+	// PerHost defines per-host Circuit Breakers
+	//
+	// +optional
+	PerHost *PerHostCircuitBreakers `json:"perHost,omitempty"`
+}
+
```
arkodg, line 55: can we move all the existing thresholds (MaxConnections, Max ...) into 1 struct, embed that in the top level and reuse it here ?

arkodg: ignore my comment, only per host max connections is supported so this is fine

api/v1alpha1/circuitbreaker_types.go
```diff
@@ -46,4 +46,20 @@ type CircuitBreaker struct {
 	// +kubebuilder:validation:Maximum=4294967295
 	// +optional
 	MaxRequestsPerConnection *int64 `json:"maxRequestsPerConnection,omitempty"`
+
+	// PerHost defines per-host Circuit Breakers
+	//
+	// +optional
+	PerHost *PerHostCircuitBreakers `json:"perHost,omitempty"`
```
arkodg, line 53: should this be called perEndpoint ?

jukie: I prefer host as it aligns with the upstream field and multiple EG API fields refer to host as well but can change if that's preferred.

jukie: Is it okay as-is or would you prefer a change here @arkodg?

arkodg: cc @envoyproxy/ai-gateway-maintainers @envoyproxy/gateway-reviewers

guydc: +1 to endpoint

zhaohuabing: +1 to using endpoint as it aligns better with exiting EG API.

jukie: Sounds good, updated

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):
LGTM thanks!

PR #5212, opened 2025-02-05
Gateway API - support percentage-based-request-mirroring

LiorLieberman:
**What type of PR is this?**
feat(gateway-api) support percentage-based-request-mirroring

This is an initial implementation, if we have agreement on the direction, happy to add the tests and resolve the TODOs. (also seeking feedback on the TODOs I left there.

This PR adds support for Gateway API Percetage-based request-mirroring (https://github.com/kubernetes-sigs/gateway-api/issues/3171)

Release Notes: Yes

/cc @arkodg


internal/xds/translator/route.go
```diff
@@ -462,16 +462,39 @@ func buildXdsDirectResponseAction(res *ir.CustomResponse) *routev3.DirectRespons
...
+func mirrorPercentByPolicy(mirror *ir.MirrorPolicy) *corev3.RuntimeFractionalPercent {
+	switch {
+	case mirror.Percentage != nil:
+		if *mirror.Percentage > 0 {
+			return &corev3.RuntimeFractionalPercent{
+				DefaultValue: translatePercentToFractionalPercent(mirror.Percentage),
+			}
+		}
+		// If zero percent is provided explicitly, we should not mirror.
+		return nil
```
arkodg, line 491: looks like `nil` doesnt mean 0, and `nil` means 100 because previously `RuntimeFraction` was unset

internal/gatewayapi/filters.go
```diff
@@ -1065,3 +1073,12 @@ func (t *Translator) processInvalidHTTPFilter(filterType string, filterContext *
 		StatusCode: ptr.To(uint32(500)),
 	}
 }
+
+// TODO (liorlieberman) is there a better home for this func?
+// OrDefault returns *t if its non-nil, or else def.
+func OrDefault[T any](t *T, def T) T {
```
arkodg, line 1079: can you use existing ptr.Deref() instead

arkodg:
thanks @LiorLieberman left some comments
would be great if you can add a unit test in
* https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata
and
* https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata/in/xds-ir

internal/gatewayapi/filters.go
LiorLieberman: sure, thanks for pointing it out

internal/xds/translator/route.go
LiorLieberman: not sure I understand, we have a few cases: 
1. There is a mirror policy but `percentage` is unset - in this case we mirror 100% of the requests to the dst

2. There is mirror policy with `percentage` set, then we honor the percentage.

LiorLieberman (commented):
image

Any idea why I fail the tests, could not understand the diffs. You have the same configs in the unitests for faultInjection. Cant tell why the denominator is not happy. 

Also @arkodg mind referring to the TODOs I left there - I just need feedback there

arkodg:
you can run `make testdata` , and let the test tool write those files for you

internal/gatewayapi/testdata/httproute-with-mirror-filter-percentage-mirroring.in.yaml
```diff
@@ -0,0 +1,53 @@
...
+          percent: 20
+      - type: RequestMirror
+        requestMirror:
+          backendRef:
+            kind: Service
+            name: mirror-service
+            port: 8080
+          fraction:
+            numerator: 25
+            denominator: 50
```
arkodg, line 53: nit: can we change the denominator to make this percent less than 1% ?

internal/xds/translator/route.go
arkodg: ah I see you have a `mp != nil` check in the caller and we are now explicitly setting `100` for the default case

internal/xds/translator/fault.go
```diff
@@ -180,11 +180,20 @@ func (*fault) patchRoute(route *routev3.Route, irRoute *ir.HTTPRoute) error {
 	return nil
 }
 
-// translatePercentToFractionalPercent translates a v1alpha3 Percent instance
-// to an envoy.type.FractionalPercent instance.
+// TODO(liorlieberman) Below 2 functions should probably go somewhere else?
```
arkodg, line 183: can we combine these functions and move it into https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/utils.go ?

internal/ir/xds.go
```diff
@@ -1312,7 +1325,8 @@ func (h *HTTPRoute) Validate() error {
 	}
 	if h.Mirrors != nil {
 		for _, mirror := range h.Mirrors {
-			if err := mirror.Validate(); err != nil {
+			// TODO(liorlieberman) add validate on mirrorPolicy?
```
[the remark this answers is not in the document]
arkodg: if you think we can do more than what CEL does, lets add it, else not required

internal/gatewayapi/testdata/httproute-with-mirror-filter-percentage-mirroring.in.yaml
LiorLieberman: We should be. in this case 5000 would make that.  any complications in envoy that I am not aware of with less than 1%?

arkodg: Not that I'm aware of, just making sure the code can deal with less than 0 here and we'll need a conformance/e2e to make sure data plane can deal with it as well

internal/xds/translator/fault.go
LiorLieberman: moved. I think its cleaner to leave it as two functions.

internal/gatewayapi/testdata/httproute-with-mirror-filter-percentage-mirroring.in.yaml
LiorLieberman: Do you want more tests here? I can add tests in gateway upstream for less-than-one percent.

arkodg: Nah this is fine, hoping the upstream conformance tests for less than 1 percent

internal/xds/translator/route.go
LiorLieberman: by default case you mean mirrorPolicy without percentage? 
If yes, than yes its a 100 - to maintain compatibility. Without this feature, when a user specified a backend to mirror requests to, 100% of the requests would be mirrored.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #5194, opened 2025-02-03
Extend RateLimitFilter to support multiple ordered RateLimit rules


nezdolik:
*Description*:
In certain cases users may need to specify multiple ratelimit rules that should be applied in certain order. E.g. a global RL rule per ip and subsequent rule per ip per upstream/backend service. The Api could look like:
```
#first RL rule
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: RateLimitFilter
metadata:
  name: global-ip-rule
spec:
  type: Global
  priority: 0
  global:
    rules:
    - clientSelectors:
      - headers:
        - name: x-real-ip
      limit:
        requests: 3
        unit: Hour

#second RL rule
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: RateLimitFilter
metadata:
  name: per-service-per-ip-rule
spec:
  type: Global
  priority: 1
  global:
    rules:
    - clientSelectors:
      - headers:
        - name: x-real-ip
        - name: some-header-that-denotes-backend-name # destination cluster is calculated at router?
      limit:
        requests: 2
        unit: Hour
```
The config would translate into multiple Envoy filters ordered according to priority. Note, in my personal opinion is easier to operate with multiple envoy filters/multiple RL domains than try to order descriptors within single RL domain. This is due how envoy ratelimit service descriptor matcing algo works, e.g. match the longest descriptor first (which will not work for the example above). So Envoy filter chain config would look like:

```
- name: envoy.filters.http.ratelimit
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit
    domain: global-ip-rule
    ...
- name: envoy.filters.http.ratelimit
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimit
    domain: per-service-per-ip-rule
    ...
```
And RL service descriptors will be populated accordingly.
Looking forward to feedback, if this proposal makes sense.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

nezdolik:
cc @envoyproxy/gateway-maintainers

arkodg:
@nezdolik this API `RateLimitFilter` has been removed (although chatgpt still remembers it :) ), and RateLimit exists in BackendTrafficPolicy https://gateway.envoyproxy.io/docs/tasks/traffic/global-rate-limit/

rsaelim:
[...]

PR #5227, opened 2025-02-06
api: sharing global ratelimit buckets

ryanhristovski:
Related to https://github.com/envoyproxy/gateway/issues/5194#issuecomment-2640112695

Currently using `shared: bool` for setting a global rate limiting rule, another option discussed was `scope: gateway/httproute`.

Does anyone have a strong opinion on how rate limit policies should be labeled to have a global scope?


api/v1alpha1/ratelimit_types.go
```diff
@@ -22,6 +22,15 @@ type RateLimitSpec struct {
 	//
 	// +optional
 	Local *LocalRateLimit `json:"local,omitempty"`
+
+	// Shared determines whether this rate limit rule applies globally across the gateway.
```
arkodg, line 26:
```suggestion
	// Shared determines whether the rate limit rules apply across all the policy targets.
```

api/v1alpha1/ratelimit_types.go
```diff
@@ -22,6 +22,15 @@ type RateLimitSpec struct {
...
+	// If set to true, the rule is treated as a common bucket and is shared across all routes under the gateway.
```
arkodg, line 27:
```suggestion
	// If set to true, the rule is treated as a common bucket and is shared across all policy targets (xRoutes).
```

api/v1alpha1/ratelimit_types.go
```diff
@@ -22,6 +22,15 @@ type RateLimitSpec struct {
...
+	// Must have targetRef set to Gateway
+	// Default: false.
+	//
+	// +optional
+	// +kubebuilder:default=false
+	Shared bool `json:"shared,omitempty"`
```
arkodg, line 33: lets make this a ptr, since its optional

api/v1alpha1/ratelimit_types.go
```diff
@@ -22,6 +22,15 @@ type RateLimitSpec struct {
 	//
 	// +optional
 	Local *LocalRateLimit `json:"local,omitempty"`
+
+	// Shared determines whether this rate limit rule applies globally across the gateway.
+	// If set to true, the rule is treated as a common bucket and is shared across all routes under the gateway.
+	// Must have targetRef set to Gateway
+	// Default: false.
+	//
+	// +optional
```
arkodg, line 31: can you also add a `+notImplementedHide` marker

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -309,6 +309,12 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(
 		err, errs error
 	)
```
arkodg, line 311: can you rm all the changes in `internal` from this `api` PR, and add these in a subsequent PR

arkodg:
added some comments, but overall +1 with the feature 

the implementation will be tricky, since we cannot rely on per route based keys and descriptors, and find a way to share it, one idea is to create a key prefix in the gateway api layer and use it in the xds layer if its set

api/v1alpha1/ratelimit_types.go
arkodg: also I dont think it may be possible to support this feature for local ratelimit so this field needs to move inside `Global`

ryanhristovski: yeah i think you're right, I'll move that over

ryanhristovski:
@arkodg i fixed a test in this PR unrelated to my changes

arkodg (approved):
LGTM thanks !

api/v1alpha1/ratelimit_types.go
```diff
@@ -51,6 +51,16 @@ type GlobalRateLimit struct {
...
+
+	// Shared determines whether the rate limit rules apply across all the policy targets.
+	// If set to true, the rule is treated as a common bucket and is shared across all policy targets (xRoutes).
+	// Must have targetRef set to Gateway
+	// Default: false.
+	//
+	// +optional
+	// +notImplementedHide
+	// +kubebuilder:default=false
+	Shared *bool `json:"shared,omitempty"`
```
sanposhiho, line 63: Using `bool` is not recommended..
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

Any strong reason for that? Can we consider enum instead?

arkodg: thanks for linking that @sanposhiho , it does say to think twice before using `bool`, mainly because intent/field could be multidimensional in the future (which enums like `type` really come in handy).
the intent here is to say that this ratelimit bucket is a singleton or is shared across all targets, can recommendations for enums here ?

sanposhiho: You can make a final decision based on other API designs in EG, but we've mostly (or probably always) used enum for new APIs even if we first thought it was a two-value switching and likely won't get changed in the future when we designed them. So, if you ask me, I'd probably change it to enum and change the name from `Shared`. (some general name, like `ApplyPolicy` etc)

rudrakhp, line 63:
```suggestion
	BucketScope RateLimitBucketScope `json:"bucketScope,omitempty"`
```
You can define enums:
```go
type RateLimitBucketScope string

const (
	SharedBucket    RateLimitBucketScope = "shared"    // Shared across routes
	PerRouteBucket  RateLimitBucketScope = "perRoute" // Independent for each route, default
)
```

guydc: +1 to future proof this API for other scopes (e.g. vhost?)

arkodg: prefer just `scope`
and instead of `perRoute` it may help to define it as `perTarget`

arkodg: I like @rudrakhp's suggestion https://github.com/envoyproxy/gateway/pull/5227#discussion_r1957401918

rudrakhp: Just scope might be confusing with the rate limit scope, i.e, Global/Local. But I guess that could be clarified in the documentation.

ryanhristovski: I could see the confusion with `scope` being confused with `type` (Global/Local), `bucketScope` could provide more clarity. Also agree with  @arkodg  for `perTarget` rather than `perRoute`.

I'll push through whichever changes we agree upon here

arkodg: i was trying to avoid  the `bucket` prefix, because i was unsure if it should be `bucket`, `rule` (which is an existing noun in the API) or `somethingElse`
we can brainstorm a little more here

rudrakhp: How about "context" (values are shared, perTarget, perVHost, etc)? I am ok with "scope" as well since it's explicitly defined under GlobalRateLimit API, just that context feels less confusing.

ryanhristovski: though i see the potential confusion, `scope: shared/target/vHost` does seem most clear

can the `vHost` option be added in another patch? was hoping to just focus on shared/target

arkodg: Yeah we can't add vHost yet, because that noun doesn't exist in any user facing docs

zhaohuabing (commented):
Non-blocking: just think it loud, this can also apply to local ratelimit. Should we move the `shared` flag to the top level?

arkodg:
> Non-blocking: just think it loud, this can also apply to local ratelimit. Should we move the `shared` flag to the top level?

I dont think this is possible, the filter is per hcm and so the bucket cannot be shared across listeners for local ratelimit

api/v1alpha1/ratelimit_types.go
arkodg: lets try and unblock Ryan, sharing some options, please vote
@envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers 
a) `bucketScope: Gateway | PerTarget`
b) `scope: Gateway | PerTarget`
c) `context: Gateway | PerTarget`

rudrakhp: a

zhaohuabing:
> I dont think this is possible, the filter is per hcm and so the bucket cannot be shared across listeners for local ratelimit

I assume the "shared" global rate limit is going to be enforced at the Listener level, while the "normal" one is applied at the HTTPRoute level, as it is currently.

This can also extend to the local rate limit filter, though its use case may be less relevant since the "shared" bucket is generally  expected to apply across the entire Envoy fleet.

zhaohuabing:
Yes, for local rate limit, the bucket can only be shared within a single lisenter as there is no rate limit server for central bucket stroage.

api/v1alpha1/ratelimit_types.go
arkodg: +1 for Option b

ryanhristovski: +1 for b

I've been developing the feature just using shared=true field for now, can readjust once we have a decision - so i'm not blocked

ryanhristovski:
Circling back here, we we're discussing the following api options:
```
a) bucketScope: Gateway | PerTarget
b) scope: Gateway | PerTarget
c) context: Gateway | PerTarget
```

But that said, after developing the feature (https://github.com/envoyproxy/gateway/pull/5361) and being mostly read for review, I do believe the most clear would be the original:
`shared: true | false`

The reason I've changed my mind on this is that I didn't originally consider the use case of having multiple BTP targetRefs that need to share the bucket. For example:
```
kind: BackendTrafficPolicy 
metadata:
  name: shared-policy-example
spec:
  targetRefs:
    kind: HTTPRoute
    name: service-a
    kind: HTTPRoute
    name: service-b
  rateLimit:
    type: Global
    global:
      shared: true
      rules:
      ...
```

none of the options of `Gateway | PerTarget` would fit the case where multiple targets share the bucket with one another, but with `shared`, it doesnt matter what is in the targetRefs, they will all share the bucket.

let me know what you think

arkodg:
thanks @ryanhristovski "shared: true | false" lgtm

zhaohuabing (approved):

Xunzhuo (approved):
/lgtm

api/v1alpha1/ratelimit_types.go
nezdolik: +1 for b

Issue #5247, opened 2025-02-10
Breaking change: Backend cluster DNS lookup is V4_PREFERRED


guydc:
*Description*:
When originally introduced, `Backend` clusters had `V4_ONLY` dns lookup settings when FQDN endpoints were used. 

With introduction of IPv6, the XDS cluster translation now uses `V4_PREFERRED` as a default when IP family is not set for a certain destination. 

https://github.com/envoyproxy/gateway/blob/a58021e6b4ae79fc268249cffcab8e342b04171b/internal/xds/translator/cluster.go#L87

Destination setting created for `Service` resources have their IPFamily determined: 

https://github.com/envoyproxy/gateway/blob/a58021e6b4ae79fc268249cffcab8e342b04171b/internal/gatewayapi/route.go#L1358

The same does not happen for `Backend` resources, and as a result, `V4_PREFERRED` is now the default for clusters created from `Backend`.

Users that specify FQDN endpoints are likely referring to cluster external resources. To reach these endpoints, traffic may need to traverse the cluster network, NAT GWs, etc. So, while the endpoint itself may advertise an IPv6 address, there's no guarantee that Envoy Proxy can actually establish an IPv6 connection with it. If Envoy falls back to an IPv6 address when an IPv4 address is temporarily found, attempts to establish a connection may fail. 

To restore support for `V4_ONLY`, Envoy Gateway can:
- Use the current `EnvoyProxy.IPFamily` as a hint for backend DNS resolution strategy. If `IPv4` is set, backend clusters will use `V4_ONLY`, and otherwise - `V4_PREFERRED. 
- Introduce a new EnvoyProxy parameter for default backend IP resolution strategy
- Support per-backend definition of resolution strategy, using a structure similar to the Service resource `IPFamilies`
- a mix of the options mentioned above

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
>Include the access logs and the Envoy logs.

arkodg:
[...]

PR #5249, opened 2025-02-10
api: BackendTrafficPolicy DNSLookupFamily

guydc:
**What type of PR is this?**

- Makes it possible to define DNS lookup family for xRoute and non-xRoute BackendRefs.
- This is particularly useful when using the `Backend` resource, see #5247 
- These settings will override any default configuration that Envoy Gateway autogenerates based on Service `IpFamilies`. 

Fixes #

Release Notes: In the implementation PR


arkodg:
thanks @guydc, this looks good !
reg naming, does any other implementation have a similar setting

api/v1alpha1/dns_types.go
```diff
@@ -7,6 +7,27 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+// DNSLookupFamily defines the behavior of Envoy when resolving DNS for hostnames
+// +enum
+// +kubebuilder:validation:Enum=IPv4Only;IPv6Only;IPv4Preferred;IPv6Preferred;IPv4AndIPv6
+type DNSLookupFamily string
+
+const (
+	// IPv4Only means the DNS resolver will first perform a lookup for addresses in the IPv4 family.
```
arkodg, line 16: thoughts on just `IPv4` ? its less to type, but will it confuse the user who is aware of the envoyproxy api enums?

api/v1alpha1/dns_types.go
```diff
@@ -7,6 +7,27 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+// DNSLookupFamily defines the behavior of Envoy when resolving DNS for hostnames
+// +enum
+// +kubebuilder:validation:Enum=IPv4Only;IPv6Only;IPv4Preferred;IPv6Preferred;IPv4AndIPv6
```
zirain, line 12: here is `IPv4Only ` while below is `IPv4`, can you make them same?

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5254, opened 2025-02-11
URLRewrite filter on empty match prefix

sadovnikov:
**What type of PR is this?**

Adding a test to demonstrate a bug with the `URLRewrite` filter.
Not to be merged until #5256 is resolved.

Instead of `"/prefix-match/doc"`, the resulting request will be `"/prefix-matchdoc"`

Found in end-to-end test logs
```
not ready yet: expected path to be /prefix-match/doc, got /prefix-matchdoc (after 1.513µs)
```

Release Notes: No


zirain:
e2e failed as expected, @sadovnikov plan to send a fix?

sadovnikov:
> e2e failed as expected, @sadovnikov plan to send a fix?

@zirain, based on the comments in #5256, the fix is not trivial. I hope it will be provided by someone more familiar with the EG code. I'm happy to add more tests but using end-2-end for this seems too late. Are there unit tests?

Issue #3706, opened 2024-06-29
Request/Response Buffering API support


luvk1412:
## Feature
I would like a direct API in envoy gateway to enable request and response buffering. As of today this can be achieved by using either [Buffer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/buffer_filter) filter to enable request buffering or using [File System Buffer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/file_system_buffer_filter#file-system-buffer) filter to enable request and response buffering with help of [EnvoyPatchPolicy](https://gateway.envoyproxy.io/v1.0.2/tasks/extensibility/envoy-patch-policy/).

I don't wish to use [EnvoyPatchPolicy](https://gateway.envoyproxy.io/v1.0.2/tasks/extensibility/envoy-patch-policy/) in production hence would prefer an api in envoy gateway.

## Use Case
One of our main reasons for setting up envoy gateway is to have a layer between our upstream servers and aws ALB which can buffer both requests and responses to prevent slow DDoS attacks. The reasons are explained further in detail in point 1 and 2 of [What is it Good For section of File System Buffer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/file_system_buffer_filter#what-is-it-good-for) documentation as well : 
> 1. To shield a server from intentional or unintentional denial of service via slow requests. Normal requests open a connection and stream the request. If the client streams the request very slowly, the server may have its limited resources held by that connection for the duration of the slow stream. With one of the “always buffer” configurations for requests, the connection to the server is postponed until the entire request has been received by Envoy, guaranteeing that from the server’s perspective the request will be as fast as Envoy can deliver it, rather than at the speed of the client.
> 2. Similarly, to shield a server from clients receiving a response slowly. For this case, an “always buffer” configuration is not a requirement. The standard Envoy behaviour already implements a configurable memory buffer for this purpose, that will allow the server to flush until that buffer hits the “high watermark” that provokes a request for the server to slow down.

Some of our upstream servers run on a request per thread model, so above types of attacks become very critical for us as it can quickly lead to server downtimes.

arkodg:
isnt this already supported https://gateway.envoyproxy.io/docs/tasks/traffic/client-traffic-policy/#configure-downstream-per-connection-buffer-limit ?
cc @yaelSchechter

luvk1412:
[...]

PR #5257, opened 2025-02-11
api: add request/response buffering via Buffer filter in ClientTrafficPolicy/BackendTrafficPolicy

jukie:
**What type of PR is this?**
Adds support for Envoy's [Buffer Filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/buffer/v3/buffer.proto) which can be used for request/response buffering. This adds a new `maxRequestBytes` field to both BackendConnection and ClientConnection which would lead to applying the buffer limit at either the Cluster level (BackendTrafficPolicy) for response buffering or Listener level (ClientTrafficPolicy) for request buffering.

Adds support for request/response buffers

Xref: #3706

Release Notes: Yes/No


api/v1alpha1/connection_types.go
```diff
@@ -26,6 +26,16 @@ type ClientConnection struct {
 	// +kubebuilder:validation:Pattern="^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$"
 	// +optional
 	BufferLimit *resource.Quantity `json:"bufferLimit,omitempty"`
+	// MaxRequestBytes specifies the maximum allowed size in bytes for each incoming request.
```
guydc, line 29: I would also make it clear to users that:
1. requests are now buffered, which may lead to a degradation in performance 
2. they should adjust their connection buffer sizes, or they may experience failures when the buffer size is exceeded. 
3. they should consider adjusting memory allocation to the envoy process. 

Since envoy does support additional means of buffering (eg filesystem), maybe the API should leave room for that. 

Finally, I wonder if there's an option limit size without buffering at all? It was discussed here: https://github.com/envoyproxy/envoy/issues/36946. It's not the best UX on the server side, but could be nice for reducing memory pressure on the proxy. Maybe `MaxRequestBytes` doesn't have to translate to a buffering implementation. 

So, maybe the API should be more explicit about the method of enforcement in use, e.g.

```
requestSizeLimit:
  type: [buffer|filesystem-buffer|stream]
  size: 10M
```

jukie: That's a great suggestion, thank you

jukie:
Closing this out in favor of @markwinter's

Issue #5230, opened 2025-02-07
HTTPRoute weighted backends with filters attached sends traffic to the wrong backend


dprotaso:
*Description*:
When I add an weighted backend to an HTTPRoute I expect traffic to go to that backend with the correct filters applied

Instead I'm seeing traffic being sent to the wrong backend.

*Repro steps*:

1. I have a simple go server that prints request headers.

<details>

<summary> Go HTTP Sever </summary>

```go
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"
)

func main() {
	target := os.Getenv("TARGET")
	if target == "" {
		target = "World"
	}

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		// Log all request headers
		fmt.Println("Request Headers:")
		for name, values := range r.Header {
			for _, value := range values {
				fmt.Printf("%s: %s\n", name, value)
			}
		}

		// Send response
		fmt.Fprintf(w, "Hello %s!\n", target)
	})

	port := os.Getenv("PORT")
	if port == "" {
		port = "8080"
	}

	log.Printf("Starting server on port %s...\n", port)
	if err := http.ListenAndServe(":"+port, nil); err != nil {
		log.Fatal(err)
	}
}

```

</details>

Next deploy two workloads using this go application

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: second
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: second
  template:
    metadata:
      labels:
        app: second
    spec:
      containers:
      - name: helloworld
        image: test-image
        env:
        - name: TARGET
          value: "second"
        ports:
        - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: second
  namespace: default
spec:
  type: ClusterIP
  ports:
  - port: 80
    targetPort: 8080
    protocol: TCP
  selector:
    app: second
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: first
    namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: first
  template:
    metadata:
      labels:
        app: first
    spec:
      containers:
      - name: helloworld
        image: test-image
        env:
        - name: TARGET
          value: "first"
        ports:
        - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: first
  namespace: default
spec:
  type: ClusterIP
  ports:
  - port: 80
    targetPort: 8080
[...]

PR #5269, opened 2025-02-13
fix: create a cluster per destination setting for backends with filters

arkodg:
* create a unique name by adding a new destination setting field
* create a unique cluster per dest setting if the setting contains filters
* reuse that for locality region name as well

Fixes: https://github.com/envoyproxy/gateway/issues/5230


guydc:
CP to 1.3.1?

internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.json
```diff
@@ -393,7 +393,7 @@
                       ],
                       "loadBalancingWeight": 1,
                       "locality": {
-                        "region": "raw_githubusercontent_com_443/backend/0"
+                        "region": "raw_githubusercontent_com_443/backend/-1"
```
zirain, line 396: why there's `-1`, is it should be `0`?

internal/xds/translator/utils.go
```diff
@@ -85,6 +85,10 @@ func clusterName(host string, port uint32) string {
 	return fmt.Sprintf("%s_%d", strings.ReplaceAll(host, ".", "_"), port)
 }
 
+func destinationSettingName(destName string) string {
+	return fmt.Sprintf("%s/backend/-1", destName)
```
zirain, line 89: can you add comment for this `-1`?

internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.json
arkodg: I've used -1 as an index for all settings which are based off a singluar field like backendRef ( and not backendRefs)

internal/xds/translator/utils.go
arkodg: sure will do

guydc (approved):
LGTM

jukie (approved):

zirain (approved):
LGTM, can you fix the conflict?

release-notes/current.yaml
```diff
@@ -18,6 +18,8 @@ new features: |
   Added support for HorizontalPodAutoscaler to helm chart
 
 bug fixes: |
+
```
zirain, line 21: nit: redundant empty line

zirain (approved):

zhaohuabing (approved):

internal/gatewayapi/listener.go
```diff
@@ -370,17 +370,19 @@ func (t *Translator) processAccessLog(envoyproxy *egv1a1.EnvoyProxy, resources *
 					logName = fmt.Sprintf("%s/%s", envoyproxy.Namespace, envoyproxy.Name)
 				}
 
+				// TODO: rename this, so that we can share backend with tracing?
+				destName := fmt.Sprintf("accesslog_als_%d_%d", i, j)
+				settingName := irDestinationSettingName(destName, -1)
```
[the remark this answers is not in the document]
arkodg: Agree, I left it as is, can improve in follow up

PR #5166, opened 2025-01-27
chore: fix gen

zhaohuabing:
Update the expired certs in the test files.

https://github.com/envoyproxy/gateway/actions/runs/12982026711/job/36201260503

PR #5273, opened 2025-02-13
[release/v1.2] chore: fix gen (#5166)

arkodg:
fix gen

(cherry picked from commit 34db8afc21437647f3c163106d11e1d8310e0ef4)


internal/gatewayapi/testdata/clienttrafficpolicy-invalid-settings.in.yaml
```diff
@@ -0,0 +1,147 @@
+clientTrafficPolicies:
```
guydc, line 1: Do we need to add this test file. 

as long as it passes test, I guess that there's no harm...

arkodg: tests pass, lets keep it ?

guydc (approved):

PR #5283, opened 2025-02-14
feat: Add RequestID to ClientTrafficPolicy for controlling X-Request-ID header behavior

### LINKED ISSUE #5281: Allow disabling `generate_request_id`
*Description*:
would be used.

I'd like the ability to disable the behavior of generating request IDs without using an EnvoyPatchPolicy. Would it make sense to add this to `ClientTrafficPolicy` under HeaderSettings?

[optional *Relevant Links*:]

ClientTrafficPolicy.HeaderSettings spec: https://github.com/envoyproxy/gateway/blob/1ba03b643046cfe05ce1a500a4c1fc43235823db/api/v1alpha1/clienttrafficpolicy_types.go#L103-L144

```
kind: EnvoyPatchPolicy
metadata:
  name: patch-pol
spec:
  jsonPatches:
  - name: default/eg/https
    operation:
      op: add
      path: /filter_chains/0/filters/0/typed_config/generate_request_id
      value: false
    type: type.googleapis.com/envoy.config.listener.v3.Listener
```

jukie:
**What type of PR is this?**

feat: allow disabling `generate_request_id`

This allows for disabling Envoy's default behavior of generating an X-request-id header for every request. Generating a random UUID4 is expensive so in high throughput scenarios it's desirable to disable this.

Fixes #5281

Release Notes: Yes


jukie:
Need to adjust the true|false logic

internal/ir/xds.go
```diff
@@ -640,6 +640,11 @@ type HeaderSettings struct {
 	// It defaults to false.
 	PreserveXRequestID bool `json:"preserveXRequestID,omitempty" yaml:"preserveXRequestID,omitempty"`
 
+	// GenerateRequestID configures whether envoy will generate the x-request-id header if it does not exist.
+	// Generating a random UUID4 is expensive so in high throughput scenarios where this feature is not desired it can be disabled.
+	// It defaults to true
+	GenerateRequestID *bool `json:"generateRequestID" yaml:"generateRequestID,omitempty"`
```
guydc, line 646: json omitempty missing

internal/xds/translator/listener_test.go
```diff
@@ -101,3 +103,35 @@ func Test_buildTCPProxyHashPolicy(t *testing.T) {
 		})
 	}
 }
+
+func Test_buildGenerateRequestID(t *testing.T) {
```
guydc, line 107: Can you also add a test case using testdata for xds translator_test?

internal/ir/xds.go
```diff
@@ -640,6 +640,11 @@ type HeaderSettings struct {
 	// It defaults to false.
 	PreserveXRequestID bool `json:"preserveXRequestID,omitempty" yaml:"preserveXRequestID,omitempty"`
 
+	// GenerateRequestID configures whether envoy will generate the x-request-id header if it does not exist.
+	// Generating a random UUID4 is expensive so in high throughput scenarios where this feature is not desired it can be disabled.
+	// It defaults to true
```
guydc, line 645: I think that we try to avoid default-true boolean flags. However, not sure if there's a good naming option here for a default-false bool.

guydc (commented):
overall LGTM, not sure if we can improve the naming here.

internal/ir/xds.go
jukie: I can adjust this because it's probably not clear. The _Envoy API _ defaults to true for this feature and Envoy Gateway defaults to an empty field here

internal/xds/translator/listener_test.go
jukie: Sure, added

internal/ir/xds.go
jukie: @guydc does the updated description sound reasonable?

internal/gatewayapi/testdata/clienttrafficpolicy-mtls-forward-client-cert.out.yaml
```diff
@@ -711,6 +716,7 @@ xdsIR:
     http:
     - address: 0.0.0.0
       headers:
+        generateRequestID: null
```
zirain, line 719: why here is a `null`?

jukie: The desire there was to leave the field omitted when left empty, I'll fix that now

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -136,6 +136,12 @@ type HeaderSettings struct {
 	// +optional
 	PreserveXRequestID *bool `json:"preserveXRequestID,omitempty"`
 
+	// GenerateRequestID configures whether envoy will generate the x-request-id header if it does not exist.
+	// Generating a random UUID4 is expensive so in high throughput scenarios where this feature is not desired it can be disabled.
+	// This feature is enabled by default when empty
+	// +optional
+	GenerateRequestID *bool `json:"generateRequestID,omitempty"`
```
arkodg, line 143: thoughts on `DisableRequestID` ?

zirain: I'm fine with the name, but this should be opt-in, and didn't change previous behavior(even it's not quite right).

jukie: This wouldn't change existing behavior. By default Envoy enables this feature when left omitted and is the same approach taken here.
* Explicit enable - adds the field to the config but effectively a no-op
* Explicit disable - disables the feature
* Omitted - no-op for the resulting envoy config and is also omitted there

zirain: IIRC, there're 200+ files changed before, that's why I call it out.

arkodg: 3 options here
```
a) GenerateRequestId
b) DisableRequestId
c) RequestId: "Preserve" | "Overwrite" | "Disable"
```
cc @envoyproxy/gateway-reviewers

zirain: +1 on `c) RequestId: "Preserve" | "Overwrite" | "Disable"`

jukie: I think we may four options but I like that approach. What behavior would `Overwrite` lead to there?

What about:
```
a) "PreserveOrGenerate" (preserve if exists, generate otherwise) # naming could be better here
b) "Preserve" (preserve if exists, do not generate)
c) "Generate" (ignore external, always generate)
d) "Disabled" (do not preserve, do not generate)
```

zirain: in that case, option (c) is better choice.

jukie: Updated with the new field and a deprecation notice for PreserveXRequestID. I'll mark this as resolved.

release-notes/current.yaml
```diff
@@ -14,6 +14,7 @@ new features: |
   Allow matchExpressions in TargetSelector
   Add defaulter for gateway-api resources loading from file to be able to set default values.
   Added support for defining Lua EnvoyExtensionPolicies
+  Added support for disabling generate_request_id
```
guydc, line 17: let's add a not about the deprecation in the deprecations section.

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -160,6 +168,24 @@ const (
...
 )
 
+// RequestIDAction configures the action to take for populating or preserving the
+// `X-Request-ID` header for a request.
+//
+// +kubebuilder:validation:Enum=PreserveOrGenerate;Preserve;Generate;Disabled
+type RequestIDAction string
+
+const (
+	// Preserve `X-Request-ID` if already presenet or generate if empty
```
arkodg, line 178:
```suggestion
	// Preserve `X-Request-ID` if already present or generate if empty
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -160,6 +168,24 @@ const (
...
+	RequestIDActionPreserveOrGenerate RequestIDAction = "PreserveOrGenerate"
+	// Preserve `X-Request-ID` if already presenet, do not generate when empty
```
arkodg, line 180:
```suggestion
	// Preserve `X-Request-ID` if already present, do not generate when empty
```

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -132,10 +132,18 @@ type HeaderSettings struct {
 	// PreserveXRequestID configures Envoy to keep the X-Request-ID header if passed for a request that is edge
 	// (Edge request is the request from external clients to front Envoy) and not reset it, which is the current Envoy behaviour.
 	// It defaults to false.
+	// Deprecated: use RequestIDAction instead
```
arkodg, line 135:
```suggestion
	// Deprecated: use RequestID instead
```

arkodg (dismissed):
LGTM thanks !

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -160,6 +168,23 @@ const (
...
+const (
+	// Preserve `X-Request-ID` if already present or generate if empty
+	RequestIDActionPreserveOrGenerate RequestIDAction = "PreserveOrGenerate"
+	// Preserve `X-Request-ID` if already present, do not generate when empty
+	RequestIDActionPreserve RequestIDAction = "Preserve"
+	// Always generate `X-Request-ID` header, do not preserve `X-Request-ID`
+	// header if it exists. This is the default behavior.
+	RequestIDActionGenerate RequestIDAction = "Generate"
+	// Do not preserve or generate `X-Request-ID` header
+	RequestIDActionDisabled RequestIDAction = "Disabled"
```
arkodg, line 185: this should be `Disable` since all other actions are in present tense

arkodg (approved):
LGTM thanks !

api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -132,10 +132,18 @@ type HeaderSettings struct {
 	// PreserveXRequestID configures Envoy to keep the X-Request-ID header if passed for a request that is edge
 	// (Edge request is the request from external clients to front Envoy) and not reset it, which is the current Envoy behaviour.
 	// It defaults to false.
+	// Deprecated: use RequestID instead
```
sanposhiho, line 135: need to mention what happens when both are set. Or, we should validate only either of them is set at CEL validation.

sanposhiho: also it'd be more user friendly how they can use RequestID to configure the same. (i.e. set `Preserve`)

internal/ir/xds.go
```diff
@@ -638,8 +647,14 @@ type HeaderSettings struct {
 	// PreserveXRequestID configures whether Envoy will keep the x-request-id header if passed for a request that is edge
 	// (Edge request is the request from external clients to front Envoy) and not reset it, which is the current Envoy behaviour.
 	// It defaults to false.
+	// Deprecated: use RequestID instead
 	PreserveXRequestID bool `json:"preserveXRequestID,omitempty" yaml:"preserveXRequestID,omitempty"`
```
sanposhiho, line 651: Can we just delete this field and use only `RequestID` internally? Do we usually avoid breaking changes at IR level as well?

arkodg: Yah +1 to cleaning this up, can also be done in a follow up

zirain (approved):
LGTM, we can handle the deprecated things seperated.

zirain (approved):

arkodg (approved):

jukie:
Thanks I'll open a cleanup PR next week.

PR #5286, opened 2025-02-16
fix: sort the missing resources to fix the gen-check

shawnh2:
**What type of PR is this?**

fix the gen-check error: https://github.com/envoyproxy/gateway/actions/runs/13351974556/job/37289680958?pr=5285#step:4:587


zirain (approved):

Xunzhuo (approved):

internal/gatewayapi/resource/load.go
```diff
@@ -330,7 +331,10 @@ func loadKubernetesYAMLToResources(input []byte, addMissingResources bool) (*Res
 	}
 
 	if addMissingResources {
-		for ns := range requiredNamespaceMap {
+		// Sort the required namespace in order to keep the consistency for test.
```
arkodg, line 334: Should the sort be in the output test code instead of here ?

PR #5299, opened 2025-02-18
feat: zone aware routing

jukie:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No


jukie:
@arkodg mind taking a first-pass look? If possible I'd prefer to limit the scope here to enabling zone aware routing with future plans to implement something similar to Istio's locality [failover](https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/failover/) or [weighted distribution](https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/distribute/).

I'd also like to handle the implementation of `EnvoyProxySpec.ServiceZoneAutodiscovery` triggering the discussed initContainer logic with a followup PR.

internal/xds/translator/testdata/out/xds-ir/backend-priority.endpoints.yaml
```diff
@@ -24,9 +24,6 @@
       region: httproute/default/httproute-2/rule/0/backend/0
 - clusterName: envoyextensionpolicy/default/policy-for-http-route/0
   endpoints:
-  - loadBalancingWeight: 1
-    locality:
-      region: envoyextensionpolicy/default/policy-for-http-route/0/backend/0
```
jukie, line 29: I think this is a pre-existing bug where a locality could be created without any endpoints. Should that logic be kept?

arkodg:
hey hoping we can get https://github.com/envoyproxy/gateway/pull/5269 in to avoid large merge conflicts

jukie:
Will split the initContainer implementation to a separate PR

internal/xds/translator/cluster.go
```diff
@@ -100,9 +102,7 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 		Name:            args.name,
 		DnsLookupFamily: dnsLookupFamily,
 		CommonLbConfig: &clusterv3.Cluster_CommonLbConfig{
-			LocalityConfigSpecifier: &clusterv3.Cluster_CommonLbConfig_LocalityWeightedLbConfig_{
-				LocalityWeightedLbConfig: &clusterv3.Cluster_CommonLbConfig_LocalityWeightedLbConfig{},
-			},
+			LocalityConfigSpecifier: &clusterv3.Cluster_CommonLbConfig_ZoneAwareLbConfig_{},
```
jukie, line 105: The default LbConfig is being changed but if a backend service doesn't also enable zone aware routing the zone information isn't added to LB Endpoints and as a result routing logic is the same as today.

internal/xds/translator/testdata/out/xds-ir/http-route-weighted-invalid-backend.endpoints.yaml
```diff
@@ -10,6 +10,3 @@
     loadBalancingWeight: 1
     locality:
       region: first-route-dest/backend/0
-  - loadBalancingWeight: 1
-    locality:
-      region: first-route-dest/backend/1
```
jukie, line 15: Previously EnvoyGateway always created a locality regardless of whether the DestinationSetting had endpoints. My assumption is that this was a bug, or at the very least a non-issue to change this. If the preference is to keep it I can include something similar to below in cluster.go.

```
// Create locality even if there's no endpoints. This is probably a bug
// but keeping this behavior to avoid breaking anything
if len(zonalEndpoints) == 0 {
       locality := &endpointv3.LocalityLbEndpoints{
               Locality: &corev3.Locality{
                       Region: ds.Name,
               },
               LoadBalancingWeight: wrapperspb.UInt32(ptr.Deref(ds.Weight, 1)),
               Priority:            ptr.Deref(ds.Priority, 0),
       }
       localities = append(localities, locality)
}
```

jukie: Looks like this is needed for conformance tests - https://productionresultssa7.blob.core.windows.net/actions-results/6d24279e-0f10-4a58-850f-4e06dc80adde/workflow-job-run-b2f1d0f0-c0de-54f9-9f89-771cfcad8528/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-02-26T14%3A49%3A47Z&sig=bkDTGhlGBIJBg%2Fd2ZK9wjg7vzCmXw4%2FzMyQMeLskqY0%3D&ske=2025-02-26T23%3A15%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-02-26T11%3A15%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-02-26T14%3A39%3A42Z&sv=2025-01-05

internal/xds/translator/cluster.go
```diff
@@ -457,28 +456,38 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
+		// Create locality even if there's no endpoints. This is needed for Gateway API Conformance.
+		if len(zonalEndpoints) == 0 {
+			locality := &endpointv3.LocalityLbEndpoints{
+				Locality: &corev3.Locality{
+					Region: ds.Name,
+				},
+				LoadBalancingWeight: wrapperspb.UInt32(ptr.Deref(ds.Weight, 1)),
+				Priority:            ptr.Deref(ds.Priority, 0),
+			}
+			localities = append(localities, locality)
```
jukie, line 476: Ref https://github.com/envoyproxy/gateway/pull/5299#discussion_r1971652773

internal/xds/translator/testdata/out/xds-ir/ext-proc-with-traffic-settings.endpoints.yaml
```diff
@@ -24,9 +24,6 @@
       region: httproute/default/httproute-2/rule/0/backend/0
 - clusterName: envoyextensionpolicy/default/policy-for-http-route/0
   endpoints:
-  - loadBalancingWeight: 1
-    locality:
-      region: envoyextensionpolicy/default/policy-for-http-route/0/backend/0
```
jukie, line 29: Previously EnvoyGateway always created a locality regardless of whether the DestinationSetting had endpoints. My assumption is that this was a bug, or at the very least a non-issue to change this. If the preference is to keep it I can include something similar to below in cluster.go.
```
// Create locality even if there's no endpoints. This is probably a bug
// but keeping this behavior to avoid breaking anything
if len(zonalEndpoints) == 0 {
       locality := &endpointv3.LocalityLbEndpoints{
               Locality: &corev3.Locality{
                       Region: ds.Name,
               },
               LoadBalancingWeight: wrapperspb.UInt32(ptr.Deref(ds.Weight, 1)),
               Priority:            ptr.Deref(ds.Priority, 0),
       }
       localities = append(localities, locality)
}
```

jukie:
I'm finding some issues with this breaking weighted backends. @guydc or @arkodg could you have a look please?

From https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#config-endpoint-v3-localitylbendpoints:
```
load_balancing_weight
([UInt32Value](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#uint32value)) Optional: Per priority/region/zone/sub_zone weight; at least 1. The load balancing weight for a locality is divided by the sum of the weights of all localities at the same priority level to produce the effective percentage of traffic for the locality. The sum of the weights of all localities at the same priority level must not exceed uint32_t maximal value (4294967295).

Locality weights are only considered when [locality weighted load balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/locality_weight#arch-overview-load-balancing-locality-weighted-lb) is configured. These weights are ignored otherwise. If no weights are specified when locality weighted load balancing is enabled, the locality is assigned no load.
```

Which means the current logic of setting `locality.LoadBalancingWeight` to ds.Weight will be ignored. I then tried also setting each individual endpoint's load balancing weight to ds.Weight which did bring _some_ weighted backend routing back but the conformance tests still fail and I'm sure why yet.

I also found an open [Envoy issue](https://github.com/envoyproxy/envoy/issues/37275) which could be relevant regarding whether zone aware routing respects weights. I'm beginning to read through the [loadbalancer code](https://github.com/envoyproxy/envoy/blob/eb0bce8c4f71af817bb75a91ca5d623cab1368e1/source/extensions/load_balancing_policies/common/load_balancer_impl.cc#L450) to build a better understanding but would either of you happen to know why zone aware lb config breaks weighted backendRefs on xRoutes?

jukie:
Retest here worked but I think the problem would be more apparent with a larger set of endpoints

internal/xds/translator/cluster.go
zhaohuabing: Could you please also add this comments to the code? It would help the people come later to understand why zone aware routing is chosen as default.

internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - lbEndpoints:
+    - endpoint:
+        address:
+          socketAddress:
+            address: 1.1.1.1
+            portValue: 50001
+      loadBalancingWeight: 40
+    locality:
+      region: first-route-dest/backend/0
+      zone: us-west-2a
```
zhaohuabing, line 12: Nit: can we add more endpoints with same/different zones in this `lbEndpoints` to make this test more meaningful?

internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml
```diff
@@ -0,0 +1,63 @@
...
+  endpoints:
+  - lbEndpoints:
+    - endpoint:
+        address:
+          socketAddress:
+            address: 1.1.1.1
+            portValue: 50001
+      loadBalancingWeight: 40
+    locality:
+      region: first-route-dest/backend/0
```
zhaohuabing, line 11: Question: does envoy also take `region` into consideration when zone aware routing is enabled?

For example, are these two endpoints considered in the same zone or not?

``
address: 1.2.3.4
region: first-route-dest/backend/0
zone: us-west-2a

address: 3.4.5.6
region: first-route-dest/backend/1
zone: us-west-2a
```

Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml:41]: Setup a test environment to help with this and the current structure for region would break zone aware routing as it does an exact locality match so the envoy proxy instances can't simply omit the region field.

I think having a cluster per-destination setting would make things much easier and I'l begin working on that unless there's a different way you'd suggest handling this.

Why?
* Instead of needing to calculate per-endpoint weights when zone aware routing is enabled (this is because locality weights are [only supported with locality weighted lb config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#config-endpoint-v3-localitylbendpoints)) weights could be determined per-cluster under [route config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routeaction) via `weighted_clusters`.
2. Would allow for mixed zone aware and locality weighted configs and allow for future expansions to locality weighted configuration
3. Would allow for dropping the region field under locality without concern of mixing all endpoints

**Comment by zhaohuabing** [comment]
> Which means the current logic of setting `locality.LoadBalancingWeight` to ds.Weight will be ignored. I then tried also setting each individual endpoint's load balancing weight to ds.Weight which did bring _some_ weighted backend routing back but the conformance tests still fail and it's not clear to me why yet. The backends are receiving _close_ to the desired weights so something it's partially working but something is off.
> 

I believe the endpoints's load balancing weight is only calculated within its own `lbEndpints` group and not across multiple localities(backendRefs), that's why the WeightEqualBackend` E2E test failed.

Would it make sense to create a separate cluster for each backendRef instead of using a single cluster for the entire `HTTPRouteRule`? This approach would provide a clearer mapping between BackendRef weight and the xDS, and also lay the groundwork for implementing locality-weight balancing.

File: internal/xds/translator/cluster.go (hunk 2 above)
Comment (jukie) [reply internal/xds/translator/cluster.go]: Since updating with cluster per-DestinationSetting I changed the default back to locality weighted

File: internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml (hunk 6 above)
Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml:18]: Thanks, that was my original intention but added this wrong 😅

File: internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml (hunk 7 above)
Comment (zhaohuabing) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: [I also believe that having a separate cluster per BackendRef(DestinationSetting) should unblock this](https://github.com/envoyproxy/gateway/pull/5299#issuecomment-2693024278). However, it would be a significant change. It might be best to implement it in a separate PR once a decision is finalized.

Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: I'm not sure if it's possible to implement zone aware routing without having a cluster per BackendRef because otherwise the weight logic is _close_ to correct but would degrade accuracy and fails gateway API conformance/e2e tests >50% of the time (tolerance of 5%).

Comment (zhaohuabing) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: Yeah, agree. I meant that we probably need to raise another PR to implement cluster per BackendRef, and then continue working on this PR.

I did some initial experiments, and the change is non-trivial: https://github.com/envoyproxy/gateway/pull/5382

Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: I can if that's preferred but on the flip side the zone aware routing changes are somewhat trivial. Could we please handle both here?

Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: On second thought I've already discovered some misses here so will take your suggestion

Comment (jukie) [reply internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml]: This should be resolved now that we've abandoned cluster per-backendRef

**Comment by jukie** [comment]
@arkodg this is ready now

File: internal/xds/translator/translator.go [hunk 8]
```diff
@@ -585,7 +585,7 @@ func (t *Translator) addRouteToRouteConfig(
 			var err error
 			// If there are no filters in the destination settings we create
 			// a regular xds Cluster
-			if !hasFiltersInSettings(httpRoute.Destination.Settings) {
+			if !hasFiltersInSettings(httpRoute.Destination.Settings) && !hasMultiBackendsAndZoneAwareRouting(httpRoute.Destination.Settings) {
```
arkodg, line 588: can we simplify this by adding a wrapper func called `needsClusterPerSetting()`

internal/xds/translator/route.go
```diff
@@ -717,3 +717,12 @@ func hasFiltersInSettings(settings []*ir.DestinationSetting) bool {
 	}
 	return false
 }
+
+func hasMultiBackendsAndZoneAwareRouting(settings []*ir.DestinationSetting) bool {
```
arkodg, line 721: nit: prefer `hasZoneAwareRouting`

internal/xds/translator/route.go
```diff
@@ -224,7 +224,7 @@ func buildXdsStringMatcher(irMatch *ir.StringMatch) *matcherv3.StringMatcher {
 
 func buildXdsRouteAction(backendWeights *ir.BackendWeights, settings []*ir.DestinationSetting) *routev3.RouteAction {
 	// only use weighted cluster when there are invalid weights
-	if hasFiltersInSettings(settings) || backendWeights.Invalid != 0 {
+	if hasFiltersInSettings(settings) || backendWeights.Invalid != 0 || hasMultiBackendsAndZoneAwareRouting(settings) {
```
arkodg, line 227: similar reuse of https://github.com/envoyproxy/gateway/pull/5299/files#r1980217608

internal/xds/translator/cluster.go
```diff
@@ -461,45 +465,86 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
-			}
-			lbEndpoint := &endpointv3.LbEndpoint{
-				Metadata: metadata,
-				HostIdentifier: &endpointv3.LbEndpoint_Endpoint{
-					Endpoint: &endpointv3.Endpoint{
-						Address: buildAddress(irEp),
+		if ds.ZoneAwareRoutingEnabled {
+			localities = append(localities, buildZonalLocalities(metadata, ds)...)
+		} else {
+			endpoints := make([]*endpointv3.LbEndpoint, 0, len(ds.Endpoints))
```
arkodg, line 471: to keep the code uniform, can we move this into another func like `buildWeightedLocalities`

internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing.endpoints.yaml
```diff
@@ -0,0 +1,125 @@
...
+            address: 2.2.2.5
+            portValue: 50002
+      loadBalancingWeight: 1
+    - endpoint:
+        address:
+          socketAddress:
+            address: 2.2.2.6
+            portValue: 50002
+      loadBalancingWeight: 1
+    loadBalancingWeight: 10
```
arkodg, line 81: shouldnt these be `1`, since we are moving the load balancing weights to the route level

arkodg: resolving this, this is a no-op, there's one locality per cluster

arkodg (dismissed):
LGTM thanks !

jukie:
I'm still unable to recreate the failed test locally if anyone sees what's wrong with my testdata and why it's generating a diff. Maybe I should add sorting to the yaml render?

<details><summary>Show Output</summary>

```

            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,2 +2,11 @@
            	            	   endpoints:
            	            	+  - lbEndpoints:
            	            	+    - endpoint:
            	            	+        address:
            	            	+          socketAddress:
            	            	+            address: 1.2.3.5
            	            	+            portValue: 50000
            	            	+      loadBalancingWeight: 1
            	            	+    locality:
            	            	+      zone: us-east-1b
            	            	   - lbEndpoints:
            	            	@@ -11,11 +20,2 @@
            	            	       zone: us-east-1a
            	            	-  - lbEndpoints:
            	            	-    - endpoint:
            	            	-        address:
            	            	-          socketAddress:
            	            	-            address: 1.2.3.5
            	            	-            portValue: 50000
            	            	-      loadBalancingWeight: 1
            	            	-    locality:
            	            	-      zone: us-east-1b
            	            	 
            	Test:       	TestTranslateXds/http-route-zonal-routing
```

</details>

arkodg:
@jukie as a workaround can you
1. rm <out-file>
2. touch <out-file>
3. make testdata
4. commit those changes and push

arkodg (approved):

internal/xds/translator/cluster.go
```diff
@@ -461,45 +465,90 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
-			if irEp.Draining {
-				healthStatus = corev3.HealthStatus_DRAINING
-			}
-			lbEndpoint := &endpointv3.LbEndpoint{
-				Metadata: metadata,
-				HostIdentifier: &endpointv3.LbEndpoint_Endpoint{
-					Endpoint: &endpointv3.Endpoint{
-						Address: buildAddress(irEp),
-					},
+		if ds.ZoneAwareRoutingEnabled {
```
zhaohuabing, line 468: To help people come later understand why buildZonalLocalities is chosen over buildWeightedLocalities, can we add a few lines to explain the rations here?

It would be helpful to add links to the existing comments and issues on the problems of using `WeightedCluster` 
https://github.com/envoyproxy/gateway/issues/5307#issuecomment-2688767482

arkodg: this section isnt related to weighted cluster though

jukie: Was planning to add that with the zone discovery PR along with some docs but I'll add some code comments

jukie: Updated

internal/gatewayapi/route.go
```diff
@@ -1468,15 +1471,16 @@ func (t *Translator) processServiceDestinationSetting(
 		endpoints, addrType = getIREndpointsFromEndpointSlices(endpointSlices, servicePort.Name, servicePort.Protocol)
 	} else {
 		// Fall back to Service ClusterIP routing
-		ep := ir.NewDestEndpoint(service.Spec.ClusterIP, uint32(*backendRef.Port), false)
+		ep := ir.NewDestEndpoint(service.Spec.ClusterIP, uint32(*backendRef.Port), false, nil)
```
zirain, line 1474: we can optimize with new function named `ir. NewDestEndpointWithZone()`

internal/gatewayapi/route.go
```diff
@@ -1496,6 +1500,18 @@ func getBackendFilters(routeType gwapiv1.Kind, backendRefContext BackendRefConte
 	return nil
 }
 
+func isZoneAwareRoutingEnabled(svc *corev1.Service) bool {
+	if trafficDist := svc.Spec.TrafficDistribution; trafficDist != nil {
+		return *trafficDist == corev1.ServiceTrafficDistributionPreferClose
+	}
+
+	if val, ok := svc.Annotations[corev1.AnnotationTopologyMode]; ok {
+		return val == "Auto" || val == "auto"
```
zirain, line 1509: will someone use `aUto` or `AUTO`, or is that valid?

internal/gatewayapi/route.go
jukie: Currently only Kubernetes Services are supported but wanted to leave this open in case `Backend` or other backend types had a way to provide zone information in the future.

internal/gatewayapi/route.go
zirain: can you address this with comments.

internal/gatewayapi/route.go
zirain: it's not a blocker.

internal/gatewayapi/route.go
jukie: This just follows native Kubernetes logic, is that necessary?

zirain: you can do it later.

jukie: Updated

arkodg (approved):

internal/gatewayapi/route.go
zirain: Thanks!

zirain (approved):

zhaohuabing (approved):
Thanks for the patience!

zirain:
@jukie is there a flaky? https://github.com/envoyproxy/gateway/actions/runs/13677062142/job/38239825535

need sort to make a stable result?

internal/xds/translator/cluster.go
```diff
@@ -461,45 +465,96 @@ func buildXdsClusterLoadAssignment(clusterName string, destSettings []*ir.Destin
...
+			},
+			LoadBalancingWeight: wrapperspb.UInt32(1),
+			HealthStatus:        healthStatus,
 		}
 
+		zone := ptr.Deref(irEp.Zone, "")
+		zonalEndpoints[zone] = append(zonalEndpoints[zone], lbEndpoint)
+	}
+
+	for zone, endPts := range zonalEndpoints {
```
zirain, line 506: @jukie range on map(`zonalEndpoints`) result random order.

zirain: xref: https://github.com/envoyproxy/gateway/pull/5413

jukie: Thank you I'll follow up

arkodg: Can we avoid using a map to ensure stable prder

Issue #5276, opened 2025-02-14
Support for SPDY Upgrade in Envoy Gateway for Kubernetes Port Forwarding


ZhengXinwei-F:
*Description*:

I'm currently using Envoy Gateway and trying to perform Kubernetes `kubectl port-forward`, which relies on SPDY upgrade. However, I encountered the following error:

```
{"start_time":"2025-02-13T11:59:05.192Z","method":"POST","x-envoy-origin-path":"/clusters/ecp-debug/api/v1/namespaces/default/pods/grafana-55b788b4cd-p9hn4/portforward","protocol":"HTTP/1.1","response_code":"403","response_flags":"-","response_code_details":"upgrade_failed","connection_termination_details":"-","upstream_transport_failure_reason":"-","bytes_received":"0","bytes_sent":"0","duration":"0","x-envoy-upstream-service-time":"-","x-forwarded-for":"100.64.0.5","user-agent":"kubectl/v1.30.2 (linux/amd64) kubernetes/3968350","x-request-id":"b462263b-4491-4693-b401-479d70b7c4f9",":authority":"ecp.cn-sh-02.sensecore.dev","upstream_host":"-","upstream_cluster":"httproute/vcluster-ecp-debug/vcluster-ecp-debug/rule/0","upstream_local_address":"-","downstream_local_address":"10.119.6.29:10443","downstream_remote_address":"100.64.0.5:59184","requested_server_name":"-","route_name":"httproute/vcluster-ecp-debug/vcluster-ecp-debug/rule/0/match/0/ecp_cn-sh-02_sensecore_dev"}
```

```
response_code_details: "upgrade_failed"
response_code: 403
```
It seems that Envoy Gateway does not support SPDY upgrade at the moment. When I switch to kubectl 1.32.2, which uses WebSocket instead, everything works fine.

I’d like to ask if there are any plans to support SPDY upgrade in the future, or if there are known workarounds for this issue?

Thanks!

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

https://kubernetes.io/blog/2024/08/20/websockets-transition/

ZhengXinwei-F:
@zirain

zirain:
xref: https://github.com/envoyproxy/envoy/issues/11190, need a way to custmize `UpgradeType`, `EnvoyPatchPolicy` should be an option.

zirain:
duplicated with https://github.com/envoyproxy/gateway/issues/4859.

Issue #4859, opened 2024-12-06
support other `upgradeType`s then just `websocket`


foosinn:
*Description*:

Currently this section is automatically added to all dynamic https routes:

```
[...]
"upgradeConfigs": [
  {
    "upgradeType": "websocket"
  }
]
[...]
```

I was unable to find a way to modify that. For some tools (in my case https://github.com/juanfont/headscale) it's required to change this value.

It would be awesome to have that configurable. Thanks!

foosinn:
Found a workaround :tada:

1. Enable `EnvoyPatchPolicy`: https://gateway.envoyproxy.io/latest/tasks/extensibility/envoy-patch-policy/#enable-envoypatchpolicy

2. Apply the policy to the cluster (replace the Gateway in `targetRef` and the underscored domain in `jsonPath` with your values)
```
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: headscale
  namespace: envoy-gateway-system
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: default
  type: JSONPatch
  jsonPatches:
    - type: "type.googleapis.com/envoy.config.route.v3.RouteConfiguration"
      name: envoy-gateway-system/default/https
      operation:
        op: replace
        jsonPath: "..virtual_hosts[?match(@.name, '.*vpn_example_org')].routes[0].route.upgrade_configs[0]"
        path: upgrade_type
        value: tailscale-control-protocol
```

guydc:
Two additional use cases:
-  Support an opt-out option for websockets using this mechanism
- Also required by  #3663

zirain:
another user case on this: https://github.com/envoyproxy/gateway/issues/5276

zirain:
@guydc what about support this in btp?

Issue #3663, opened 2024-06-24
HTTP2 Connect support


denniskniep:
*Description*:
It should be possible to use the [HTTP Verb "CONNECT"](https://en.wikipedia.org/wiki/HTTP_tunnel) to start a tcp tunnel  through envoy to an upstream target.
This would have the benefit that one can use the http filters during initialization of  the tcp tunnel (i.e. oauth2 filter for authentication and ext_auth for authorization).

Here is a small PoC with pure envoy config:
https://github.com/denniskniep/envoy-http-connect-tunnel-demo

Assuming this example HTTPRoute:
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend
  namespace: default
spec:
  hostnames:
  - www.example.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
    matches:
    - path:
        type: PathPrefix
        value: /
```

This example will be translated to this VirtualHost:
```
{
       "virtual_hosts": [
              {
                     "name": "default/eg/http/www_example_com",
                     "domains": [
                            "www.example.com"
                     ],
                     "routes": [
                            {
                                   "match": {
                                          "prefix": "/"
                                   },
                                   "route": {
                                          "cluster": "httproute/default/backend/rule/0",
                                          "upgrade_configs": [
                                                 {
                                                        "upgrade_type": "websocket"
                                                 }
                                          ]
                                   },
                                   "name": "httproute/default/backend/rule/0/match/0/www_example_com"
                            }
                     ]
              }
       ]
}
```

To make this HTTPRoute able to handle `HTTP Connect` requests we need to configure the following:
```
  virtual_hosts:
  - name: default/eg/http/www_example_com
    domains:
    - "www.example.com"
    routes:
    - match:
        connect_matcher:
          {}
      route:
        cluster: "httproute/default/backend/rule/0",
        upgrade_configs:
        - upgrade_type: CONNECT
          connect_config:
            {}
```

We need to utilize the `connect_matcher` (see [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routematch)) for route matching.
Important is that the  `connect_matcher` can not be used together with `prefix` matcher! (Because of the precisely one requirement)

[...]

PR #5300, opened 2025-02-18
api: support customized UpgradeConfigs

zirain:
xref: https://github.com/envoyproxy/gateway/issues/4859 https://github.com/envoyproxy/gateway/issues/3663 https://github.com/envoyproxy/gateway/issues/5276


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,11 +73,26 @@ type BackendTrafficPolicySpec struct {
 	//
 	// +optional
 	ResponseOverride []*ResponseOverride `json:"responseOverride,omitempty"`
+	// UpgradeConfigs defines the configuration for protocol upgrades.
+	// If not specified, the default upgrade configuration(websocket) will be used.
+	//
+	// +optional
+	UpgradeConfigs []*BackendUpgradeConfig `json:"upgradeConfigs,omitempty"`
```
arkodg, line 80: prefer `upgrade` or `httpUpgrade`

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,11 +73,26 @@ type BackendTrafficPolicySpec struct {
...
 }
 
-// +kubebuilder:object:root=true
+type BackendUpgradeConfig struct {
+	// UpgradeType is the case-insensitive name of protocol upgrade.
+	//
+	// +kubebuilder:validation:Required
+	UpgradeType string `json:"upgradeType"`
```
arkodg, line 87: prefer `type`, since we've used `upgrade` earlier in the parent name

arkodg: we should also consider enum here

zirain: prefer to `backendProtocol` or `protocol` than type here.
Keep it as string will give us free to support new/unspported protocol in the future.

arkodg: does envoy proxy handle arbitrary string well ?

zirain: the problem is that envoy didn't list protocols it supported in doc.

arkodg:
hey @zirain lets also make sure we address https://github.com/envoyproxy/gateway/issues/3663 here
cc @denniskniep

denniskniep:
Hi @zirain, thanks for tackling that topic!

You added the new config property to the [BackendTrafficPolicy](https://gateway.envoyproxy.io/docs/api/extension_types/#backendtrafficpolicy). The docs are defining it as:

> BackendTrafficPolicy allows the user to configure the behavior of the connection between the Envoy Proxy listener and the backend service.

Isn´t http upgrade_type an HttpRoute specific thing? Just wondering if it makes sense to integrate it as implementation specific [HTTPRouteFilter](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRouteFilter) within [ExtensionRef](https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilterspec)
There is already a complex filter defined in the spec: [requestMirror](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRequestMirrorFilter)

```
kind: HTTPRoute
metadata:
  name: upgrade-example-com
spec:
  parentRefs:
  - name: eg
  hostnames:
  - "www.example.com"    
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /test
    filters:
    - type: ExtensionRef
      extensionRef:
        group: gateway.envoyproxy.io
        kind: HTTPRouteFilter
        name: upgrade-websockets
        
kind: HTTPRouteFilter
metadata:
  name: upgrade-websockets
spec:
  upgrade:
    protocol: websocket
    disabled: false
```


Can we also tackle [CONNECT_UDP ](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades.html#connect-udp-support) ?

> allows HTTP clients to create UDP tunnels through an HTTP proxy server. Unlike CONNECT, which is limited to tunneling TCP, CONNECT-UDP can be used to proxy UDP-based protocols such as HTTP/3.


Furthermore adding [ConnectConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction-upgradeconfig-connectconfig) as props would be nice. This can be specified if protocol is CONNECT or CONNECT-UDP
```
kind: HTTPRouteFilter
metadata:
  name: upgrade-connect
spec:
  upgrade:
    disabled: false
    protocol: CONNECT
    connect:
      proxyProtocol:
        version: V2
```

zirain:
IMO, `upgrade` is a thing per backend/port, not per route(even we will put the configuration on route, it's because we share the listener).

Is there a user case, we do the upgrade in one route and not in another route for the same backend?

denniskniep:
> Is there a user case, we do the upgrade in one route and not in another route for the same backend?

tbh, not absolutely sure about that. 

The property wouldn't have an effect if we use a tcp-route, right? So therefore I thought properties are added where the effect takes place.

Furthermore there is a related property per route that might be necessary to set in certain scenarios: `connect_matcher`

see here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routematch

zirain:
> Furthermore there is a related property per route that might be necessary to set in certain scenarios: connect_matcher

that's another feature, which is related more than protocol upgrade, in the future, we may need both to support connect matcher.

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,11 +73,28 @@ type BackendTrafficPolicySpec struct {
...
+	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty"`
 }
 
-// +kubebuilder:object:root=true
+type ProtocolUpgradeConfig struct {
+	// Protocol is the case-insensitive name of protocol upgrade.
+	// e.g. `websocket`, `CONNECT`, `spdy/3.1` etc.
+	//
+	// +kubebuilder:validation:Required
+	Protocol string `json:"protocol"`
```
arkodg, line 89: unsure here if we should be using `Protocol` or `Type` (for type of upgrade)

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,11 +73,28 @@ type BackendTrafficPolicySpec struct {
...
+	// Disabled indicates whether the upgrade is disabled.
+	// +optional
```
arkodg, line 91: we could add a `Connect` field in this API in the future
```
connect:
  forwardTCP: false | true // Supports TCP tunnels
  allowHTTP2: false | true
  allowHTTP3: false | true
```

arkodg (approved):
LGTM thanks !

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -74,11 +73,28 @@ type BackendTrafficPolicySpec struct {
...
+	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty"`
 }
 
-// +kubebuilder:object:root=true
+type ProtocolUpgradeConfig struct {
+	// Type is the case-insensitive type of protocol upgrade.
+	// e.g. `websocket`, `CONNECT`, `spdy/3.1` etc.
+	//
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
```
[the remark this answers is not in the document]
zirain: the problem is that envoy didn't list all the protocol types it supported in doc.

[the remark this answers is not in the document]
zirain: prefer to leave it as string.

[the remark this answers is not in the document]
rudrakhp: As a control plane we could be strict about what we support, but since [proxy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-upgradeconfig) allows arbitrary string, so it's fine I guess.

rudrakhp (approved):

arkodg:
hey @zirain rethinking think one, as @denniskniep suggested `HTTPRouteFilter` may be a better fit here especially if we add `upgrade.connect` to this struct in the future
wdyt @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

zirain:
> Is there a user case, we do the upgrade in one route and not in another route for the same backend?

I replied this before, and we may need both in the future.

arkodg (approved):

zirain:
From my point of view, a gateway may have more than 1 httproute.

Thinking of I have N httproute in , before EG support https://github.com/envoyproxy/gateway/issues/5351, you need to ref the route filter in every http route for that backend.

I'm not against copy this struct into `HTTPRouteFilter` so that we can customize for specified route.

ATM, you can ref BTP to Gateway or HTTPRoute, which sounds good to me.

I think this API solves https://github.com/envoyproxy/gateway/issues/4859, but for https://github.com/envoyproxy/gateway/issues/3663 we need more work on `connect_match`.

Issue #4661, opened 2024-11-07
Request Authorization based on Header Value


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Relates to https://github.com/envoyproxy/gateway/issues/2250 & https://github.com/envoyproxy/gateway/issues/4412 & https://github.com/envoyproxy/gateway/issues/2947

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
requires an enhancement 
* in the API https://github.com/envoyproxy/gateway/blob/d32481eb52a752e9779d30ef5b80eb5711d3aa45/api/v1alpha1/authorization_types.go#L55
* Translate this to IR
* Enhance RBAC Filter Matcher Predicate https://github.com/envoyproxy/gateway/blob/d32481eb52a752e9779d30ef5b80eb5711d3aa45/internal/xds/translator/authorization.go#L205

Issue #4660, opened 2024-11-07
Requestion Authorization based on HTTP Method


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Relates to https://github.com/envoyproxy/gateway/issues/2250

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #5310, opened 2025-02-19
api: HTTP header and method based authz

zhaohuabing:
API for #4660 and #4661


api/v1alpha1/authorization_types.go
```diff
@@ -72,6 +76,18 @@ type Principal struct {
...
+	// If multiple headers are specified, all headers must match for the rule to match.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	Headers []HeaderMatch `json:"headers,omitempty"`
+
+	// Methods authorize the request based on the HTTP methods.
+	// If multiple methods are specified, one of the methods must match for the rule to match.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	Methods []gwapiv1.HTTPMethod `json:"methods,omitempty"`
```
zhaohuabing, line 90: Methods could also be support as pseudo headers(:method) within the `Headers` field, however, providing a separate field would be more user-friendly and improve UI clarity.

api/v1alpha1/authorization_types.go
```diff
@@ -72,6 +76,20 @@ type Principal struct {
 	// JWT authentication in the same `SecurityPolicy`.
 	// +optional
 	JWT *JWTPrincipal `json:"jwt,omitempty"`
+
+	// Headers authorize the request based on the headers in the request.
```
arkodg, line 80: curious why we are choosing `AND` here
`OR` fits here better

or maybe we can't use `HeaderMatch` here and may need some other type

here is the use case, id like to deny traffic when `x-geoip-country` contains `foo` or `bar`
so maybe
```
headers:
- name: x-geo-ip
   values:
   - "foo"
   - "bar"  
```

fits better here
cc @nezdolik

api/v1alpha1/authorization_types.go
```diff
@@ -72,6 +76,20 @@ type Principal struct {
...
+	// +kubebuilder:validation:MinItems=1
+	// +notImplementedHide
+	Headers []HeaderMatch `json:"headers,omitempty"`
+
+	// Methods authorize the request based on the HTTP methods.
+	// If multiple methods are specified, one of the methods must match for the rule to match.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +notImplementedHide
+	Methods []gwapiv1.HTTPMethod `json:"methods,omitempty"`
```
arkodg, line 92: should methods live in the top level outside `principal` because its falls under `actions` / `operation` performed by the principal

zhaohuabing: Yeah, moving `methods` outside of `principal` looks better.

api/v1alpha1/authorization_types.go
arkodg:
```
    authorization:
      defaultAction: Allow
      rules:
      - action: Deny
        principal:
          headers:
          - name: x-geoip-country
            value: foo
      - action: Deny
        principal:
          headers:
          - name: x-geoip-country
            value: bar
```
is too lengthy, the common case is there will be hundreds of such entries, which requires 5x more lines

api/v1alpha1/authorization_types.go
arkodg: referencing istio here https://istio.io/latest/docs/reference/config/security/authorization-policy/#Operation
should we have a top level `operation` here ?
so we can have `operation.methods` and leave room for `operation.paths` in the future ?

api/v1alpha1/authorization_types.go
zhaohuabing: Ha, you're right! I was blind by trying to reuse the exiting HeaderMatch, instead, an array should be used for header value:

```
    authorization:
      defaultAction: Deny
      rules:
      - action: Allow
        principal:
          headers:
          - name: x-user-id
            values: [john, alice]
          - name: x-org-id
            values: sales
      - action: Deny
        principal:
          headers:
          - name: x-geoip-country
            values: [foo, bar]
```

This is similar to what we have done for JWT:

```
        principal:
          jwt:
            provider: example1
            scopes: [foo, bar]
```

api/v1alpha1/authorization_types.go
```diff
@@ -72,6 +96,36 @@ type Principal struct {
...
 	JWT *JWTPrincipal `json:"jwt,omitempty"`
+
+	// Headers authorize the request based on user identity extracted from custom headers.
+	// If multiple headers are specified, all headers must match for the rule to match.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=256
+	// +notImplementedHide
+	Headers []AuthorizationHeaderMatch `json:"headers,omitempty"`
```
zhaohuabing, line 107: A new `AuthorizationHeaderMatch` is introduced for authorization, as the existing `HeaderMatch` is not suitable for this use case.

api/v1alpha1/authorization_types.go
```diff
@@ -72,6 +96,36 @@ type Principal struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	Name string `json:"name"`
+
+	// Values are the values that the header must match.
+	// If multiple values are specified, the rule will match if any of the values match.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=256
+	Values []string `json:"values"`
```
zhaohuabing, line 125: Only exact matches are supported for now. It should be good enough for authorization use cases. If use cases for other matching types arise, we can add a MatchingType field here.

api/v1alpha1/ratelimit_types.go
```diff
@@ -220,20 +220,22 @@ type SourceMatch struct {
...
 	// Name of the HTTP header.
+	// The header name is case-insensitive unless PreserveHeaderCase is set to true.
+	// For example, "Foo" and "foo" are considered the same header.
+	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=256
 	Name string `json:"name"`
 
-	// Value within the HTTP header. Due to the
-	// case-insensitivity of header names, "foo" and "Foo" are considered equivalent.
```
zhaohuabing, line 236: This existing comment seems to be in wrong place. Moving it to the header comment.

api/v1alpha1/authorization_types.go
```diff
@@ -38,17 +42,35 @@ type AuthorizationRule struct {
 	// Action defines the action to be taken if the rule matches.
 	Action AuthorizationAction `json:"action"`
 
+	// Operation specifies the operation of a request, such as HTTP methods.
+	// If not specified, all operations are allowed or denied, based on the action of the rule.
```
arkodg, line 46:
```suggestion
	// If not specified, all operations are matched on.
```

api/v1alpha1/authorization_types.go
```diff
@@ -38,17 +42,35 @@ type AuthorizationRule struct {
...
 
+// Operation specifies the operation of a request.
+// +notImplementedHide
+type Operation struct {
+	// Methods are the HTTP methods of the request.
+	// If multiple methods are specified, all specified methods are allowed or denied, based on the action of the rule.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Methods []gwapiv1.HTTPMethod `json:"methods"`
```
arkodg, line 66: omitempty ? or we add that later ?

zhaohuabing: It can't be empty becuase no other fields inside the Operation right now.  It should be optional/omitempty after paths or hosts are supported.

arkodg (approved):
LGTM thanks !

zirain (approved):

nathanwang-ops:
@zhaohuabing does it support xfcc header?

zhaohuabing:
> @zhaohuabing does it support xfcc header?

@nathanwang-ops any arbitrary header in the HTTP request should be supported. 

I think xfcc may deserve its own first-class API as a `Principal` in the `SecurityPolicy`'s `Authorization` section.

Something like:

```
clientCert:
  Subject: foo.bar.com
  URI: http://foo.bar.com
  ...
 ```

nathanwang-ops:
@zhaohuabing thank you for quick response, is there a feat for SecurityPolicy's Principal improvement?

zhaohuabing:
> @zhaohuabing thank you for quick response, is there a feat for SecurityPolicy's Principal improvement?

#2250 did mentioned client cert auth, but it's marked as completed.

You can create one and linked it to https://github.com/envoyproxy/gateway/issues/2250

Issue #4757, opened 2024-11-21
Append a Request Header value from a Secret


arkodg:

arkodg:
there doesnt seem to be a way to read a header value from sds config source
today https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalueoption

will need something like https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/config_source.proto#config-core-v3-configsource to implement this

mathetake:
alternatively simply EG can read the secret and use the plain header mutation?

arkodg:
The value will be exposed in the config dump which is not desirable

mathetake:
yeah though i imagine people capable of dumping configs == people having access to secrets?

arkodg:
that may not be the case, app devs (xRoute creators) may also hold sensitive info they may not want to share with infra admins (managing EG infra) like basic auth keys, API keys

arkodg:
upstream issue https://github.com/envoyproxy/envoy/issues/37656

arkodg:
in envoy this can we solved using https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/credential_injector_filter

from an EG perspective, this can be added into HTTPRouteFIlter https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilterspec
cc @missBerg

maulik13:
Hi @arkodg, is there a way to use `credential_injector_filter` in Envoy Gateway using either a patch policy or extensions? (while we wait for having this accessible directly in the GW api?)

Some of our backend services have their own authentication.

arkodg:
Keeping this open to complete docs work

PR #5312, opened 2025-02-19
api: credential injection HTTPRouteFilter

zhaohuabing:
API for https://github.com/envoyproxy/gateway/issues/4757


mathetake:
will the secret updates be picked up?

mathetake:
like changing the content

arkodg:
> like changing the content

yah, we have existing logic for it, which is used in other places (TLS Certs, Basic Auth etc)

mathetake:
but at the moment the core logic is tied with AWS signing so this can only be used for API key auth 😞

maulik13:
> but at the moment the core logic is tied with AWS signing so this can only be used for API key auth 😞

I am not sure what this AWS signing has to do with credential injector, but in its current state credential injector will already be useful for us.

arkodg (approved):
LGTM thanks !

guydc (approved):

PR #5317, opened 2025-02-19
feat: support adding additional labels to dashboard configmap

fengxsong:
**What type of PR is this?**

* "feat(chart): Support adding additional labels to the dashboard ConfigMap."

If we use this chart along with a tool like https://github.com/kiwigrid/k8s-sidecar to automatically manage dashboards, this tool requires explicitly adding labels like 'grafana_dashboard: ""' to the ConfigMap.


zirain (approved):

charts/gateway-addons-helm/values.yaml
```diff
@@ -1,3 +1,6 @@
+dashboards:
```
arkodg, line 1: can we use `dashboard` here ? because the object is not an array/map type

fengxsong: although the generated object is a single ConfigMap, but it creates multiple grafana dashboards, doesn't it? so I think using the plural form is more accurate.

fengxsong: okay, it's been done.

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #5315, opened 2025-02-19
Support distinct match for local ratelimit


zhaohuabing:
EG can now add distinct match support for local ratelimit, as it has been implemented in Envoy.

https://github.com/envoyproxy/envoy/pull/36623

PR #5319, opened 2025-02-19
feat: distinct match support for local ratelimit

zhaohuabing:
This PR adds distinct header match and distinct source CIDR match to local ratelimit. Both are verified with e2e tests.

implement: #5315 

Release Notes: Yes


arkodg:
thanks @zhaohuabing ! this looks good
we'll also need to edit max_dynamic_descriptors, the default of 20 is too low, and later we can could consider a API for it
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto

internal/xds/translator/local_ratelimit.go
```diff
@@ -201,56 +199,66 @@ func buildRouteLocalRateLimits(local *ir.LocalRateLimit) (
 
 		// HeaderMatches
 		for mIdx, match := range rule.HeaderMatches {
+			var action *routev3.RateLimit_Action
```
arkodg, line 202: curious, if any of this code can be deduped with a function
https://github.com/envoyproxy/gateway/blob/5b6a35f3e6b6b31ad7479a9ceb5159b77cd68f7a/internal/xds/translator/ratelimit.go#L166

zhaohuabing:
> thanks @zhaohuabing ! this looks good
> we'll also need to edit max_dynamic_descriptors, the default of 20 is too low, and later we can could consider a API for it
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto

Setting a default value is tricky because "Actual number of dynamic descriptors will depend on the cardinality of unique values received from the http request for the omitted values", which is unpredictable. 

Would 100 be reasonable default for a listener?

arkodg:
> > thanks @zhaohuabing ! this looks good
> > we'll also need to edit max_dynamic_descriptors, the default of 20 is too low, and later we can could consider a API for it
> > https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto
> 
> Setting a default value is tricky because "Actual number of dynamic descriptors will depend on the cardinality of unique values received from the http request for the omitted values", which is unpredictable.
> 
> Would 100 be reasonable default for a listener?

I was thinking 10K, assuming at any given time, the route had 10K unique active users which need to be ratelimited
wdyt @wbpcode @vikaschoudhary16

zhaohuabing:
What happens if the cache limit is reached?  The oldest value will be removed and the limit for that value may not be accurate?

vikaschoudhary16:
Yeah, let's say the one which got removed, LRU algo, had its token 50% consumed. Now if again a request comes, new token bucket will be created and thus tokens will get refreshed unexpectedly

arkodg (dismissed):
LGTM thanks !

internal/xds/translator/local_ratelimit.go
zhaohuabing: Ha, may look into later, the rate limit code is pretty complex.

zirain (approved):

arkodg (approved):

Issue #5260, opened 2025-02-11
Support exclusion of listeners from overload manager


guydc:
*Description*:
Currently, EG generates overload manager settings that stop accepting requests when heap utilization is high:
https://github.com/envoyproxy/gateway/blob/c4920899659e75ceb342df02665e1b484eb0ecbc/internal/xds/bootstrap/bootstrap.yaml.tpl#L265

This may have an unexpected side effect on the Envoy proxy pod's readiness probes, leading to probe failure, e.g. as seen here: https://github.com/envoyproxy/envoy/issues/23843. Since liveness probes are not used by EG, the pod would not be restarted. 

However, even a readiness probe failure may in some conditions lead to wide service disruption. For example, if using `ExternalTrafficPolicy: Local` and a single envoy pod in every node, an unready pod would trigger a failure of the NLB Health Check (https://kubernetes.io/blog/2022/12/30/advancements-in-kubernetes-traffic-engineering/). Some NLBs, e.g. [AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html) NLB, are documented to reset all connections to that node in such cases: 

> If a target becomes unhealthy, the load balancer sends a TCP RST for packets received on the client connections associated with the target

EG Should:
1. make it possible to exclude readiness, stats and other system listeners from overload manager 
2. make it possible to exclude user-defined listeners from overload manager, in case custom HCs are implemented

Envoy supports excluding listeners from overload manager: https://www.envoyproxy.io/docs/envoy/v1.31.5/api-v3/config/listener/v3/listener.proto.html#envoy-v3-api-field-config-listener-v3-listener-bypass-overload-manager 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
+1 to defaulting exclusion of the stats and readiness listener

rudrakhp:
I want to give this a shot, quick question though:
> make it possible to exclude user-defined listeners from overload manager, in case custom HCs are implemented

Maybe I don't have the full picture, but was wondering how this can be achieved, given we are confirmed by the Gateway API spec. Would a label in the Gateway object do the job? For example, a label `gateway.envoyproxy.io/bypass-overload-manager-listeners` would expect a comma separated list of listener names that are to be bypassed by the overload manager.

guydc:
Hi @rudrakhp.

[...]

PR #5321, opened 2025-02-20
Bypass overload manager for stats and ready listeners

rudrakhp:
**What type of PR is this?**
Fix: bypass overload manager for stats and ready listeners

Related #5260


arkodg:
that was quick 😄

guydc:
LGTM, maybe just a release note under "behavior changes". IMHO, this is the correct default behavior, but maybe some folks want OM to fail checks so that the pods are excluded from the serving set by kube-proxy/externalLB, ...

arkodg (dismissed):
LGTM thanks !

zirain:
need changelog

arkodg (approved):

guydc (approved):

PR #5324, opened 2025-02-21
feat: adding support for DNSLookupFamily api

alexwo:
Makes it possible to define DNS lookup family for xRoute and non-xRoute BackendRefs.

Implements:
https://github.com/envoyproxy/gateway/pull/5249

https://github.com/envoyproxy/gateway/issues/5247


internal/xds/translator/cluster.go
```diff
@@ -85,14 +85,32 @@ func buildEndpointType(settings []*ir.DestinationSetting) EndpointType {
 
 func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 	dnsLookupFamily := clusterv3.Cluster_V4_PREFERRED
+	customDNSPTrafficPolicy := args.dns != nil && args.dns.LookupFamily != nil
+	// apply DNS lookup family if custom DNS traffic policy is set
+	if customDNSPTrafficPolicy {
```
guydc, line 90: `customDNSTrafficPolicy` ? is the `T` intentional?

internal/ir/xds.go
```diff
@@ -1452,9 +1454,12 @@ type DestinationSetting struct {
 	AddressType *DestinationAddressType `json:"addressType,omitempty" yaml:"addressType,omitempty"`
 	// IPFamily specifies the IP family (IPv4 or IPv6) to use for this destination's endpoints.
 	// This is derived from the backend service and endpoint slice information.
-	IPFamily *egv1a1.IPFamily    `json:"ipFamily,omitempty" yaml:"ipFamily,omitempty"`
-	TLS      *TLSUpstreamConfig  `json:"tls,omitempty" yaml:"tls,omitempty"`
-	Filters  *DestinationFilters `json:"filters,omitempty" yaml:"filters,omitempty"`
+	IPFamily *egv1a1.IPFamily `json:"ipFamily,omitempty" yaml:"ipFamily,omitempty"`
+	// DNSLookupFamily specifies a preferred resolution strategy for ip.
+	// This is derived from the backend service and endpoint slice information.
+	DNSLookupFamily *egv1a1.DNSLookupFamily `json:"DNSLookupFamily,omitempty" yaml:"ipFamily,omitempty"`
```
guydc, line 1460: We usually create another IR-level struct. It's kind of meaningless here, as the translation is 1:1, but would still be nice for consistency.

internal/xds/translator/cluster.go
```diff
@@ -85,14 +85,32 @@ func buildEndpointType(settings []*ir.DestinationSetting) EndpointType {
...
+			dnsLookupFamily = clusterv3.Cluster_V6_ONLY
+		case egv1a1.IPv6PreferredDNSLookupFamily:
+			dnsLookupFamily = clusterv3.Cluster_AUTO
+		case egv1a1.IPv4AndIPv6DNSLookupFamily:
+			dnsLookupFamily = clusterv3.Cluster_ALL
+		}
+	}
+
+	// Ensure to override if a specific IP family is set.
 	if args.ipFamily != nil {
```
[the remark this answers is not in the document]
alexwo: If the IP family is determined, using a custom policy may not be practical, as the destination is expected to support a specific protocol.

internal/ir/xds.go
alexwo: Right, but is not is the same for IPFamily?

internal/xds/translator/cluster.go
alexwo: Updated

internal/gatewayapi/clustersettings.go
```diff
@@ -475,8 +475,9 @@ func translateDNS(policy egv1a1.ClusterSettings) *ir.DNS {
 		return nil
 	}
 	return &ir.DNS{
-		RespectDNSTTL:  policy.DNS.RespectDNSTTL,
-		DNSRefreshRate: policy.DNS.DNSRefreshRate,
+		DNSLookupFamily: buildDNSLookupFamily(policy.DNS.LookupFamily),
```
zirain, line 478: if the struct named DNS, maybe we should remove the prefix to use LookupFamily.

internal/ir/xds.go
```diff
@@ -763,6 +763,8 @@ type DNS struct {
 	DNSRefreshRate *metav1.Duration `json:"dnsRefreshRate,omitempty"`
 	// RespectDNSTTL indicates whether the DNS Time-To-Live (TTL) should be respected.
 	RespectDNSTTL *bool `json:"respectDnsTtl,omitempty"`
+	// LookupFamily allows you to configure a preferred DNS lookup policy.
+	DNSLookupFamily *DNSLookupFamily `json:"lookupFamily,omitempty"`
```
zirain, line 767:
```suggestion
	LookupFamily *DNSLookupFamily `json:"lookupFamily,omitempty"`
```

zirain: cc @guydc , if this struct live in DNS struct, it's fine to me to use api struct directly.

internal/xds/translator/cluster.go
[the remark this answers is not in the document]
zhaohuabing: Nit:  I assume that the DNSLookupFamily settings is only relevant to the `Backend` FQDN endpoints? If so, should we mention this in the API? Otherwise, it could confuse users.

zhaohuabing (commented):
Nit: I assume that the DNSLookupFamily settings is only relevant to the Backend FQDN type endpoints? If so, should we mention this in the API? Otherwise, it could confuse users.

alexwo:
> Nit: I assume that the DNSLookupFamily settings is only relevant to the Backend FQDN type endpoints? If so, should we mention this in the API? Otherwise, it could confuse users.

Right, Destination setting created for Service resources have their IPFamily determined by: 

https://github.com/envoyproxy/gateway/blob/a58021e6b4ae79fc268249cffcab8e342b04171b/internal/gatewayapi/route.go#L1358

So right now it will apply for backends.

zhaohuabing:
> So right now it will apply for backends dns resolution.

Could you please add comment to the BackendTrafficPolicy API clarifying that the DNS Lookup Family setting only affects targeted Routes with an FQDN backend?  The current API doc is somewhat vague, and this clarification would help improve understanding.

```
	// LookupFamily determines how Envoy would resolve DNS for. If set, this configuration overrides other default
	// value that Envoy Gateway configures based on attributes of the backends, such Service resource IPFamilies.
	// +optional
	// +notImplementedHide
	LookupFamily *DNSLookupFamily `json:"lookupFamily,omitempty"`
```

internal/ir/xds.go
guydc: Yea, sure, we can drop the dns prefix

guydc (approved):

PR #5332, opened 2025-02-23
chore: cleanup after adding RequestID

jukie:
**What type of PR is this?**

Some cleanup after comments in https://github.com/envoyproxy/gateway/pull/5283
* Remove `PreserveXRequestID` from internal and map ClientTrafficPolicy `PreserveXRequestID` to RequestID=Preserve
* Mention `RequestID=Preserve` alternative in ClientTrafficPolicy description

Fixes #

Release Notes: No


api/v1alpha1/clienttrafficpolicy_types.go
```diff
@@ -131,8 +131,8 @@ type HeaderSettings struct {
 
 	// PreserveXRequestID configures Envoy to keep the X-Request-ID header if passed for a request that is edge
 	// (Edge request is the request from external clients to front Envoy) and not reset it, which is the current Envoy behaviour.
-	// It defaults to false.
-	// Deprecated: use RequestID instead
+	// Defaults to false and will be ignored if RequestID is also set.
+	// Deprecated: use RequestID=Preserve instead
```
zirain, line 135: can you add a cel validation for this?

zirain: yeah, that should be enough.

arkodg:
thanks !
can you also add a test in https://github.com/envoyproxy/gateway/tree/main/test/cel-validation to validate the CEL validation

jukie:
Rebased and updated with cel test

arkodg (approved):
LGTM thanks !

zirain (approved):

guydc:
@jukie, @arkodg - shouldn't the doc string say: Deprecated: use RequestID=Preserve**AndGenerate**? 
Envoy's default for GenerateRequestID is true. So, a user switching to the new struct would probably want to keep the previous behavior wrt. that field. 


> generate_request_id
> ([BoolValue](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#boolvalue)) Whether the connection manager will generate the [x-request-id]> (https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-request-id) header if it does not exist. This defaults to true. Generating a random UUID4 is expensive so in high throughput scenarios where this feature is not desired it can be disabled.

I might be misreading something though.

arkodg:
@guydc the doc string is saying, if you want the previous functionality of `preserveRequestID: true`, you can achieve that with `requestID: Preserve`

guydc:
> // Preserve `X-Request-ID` if already present, do not generate when empty
RequestIDActionPreserve RequestIDAction = "Preserve"

Previously, we didn't expose the `generate_request_id` option. #5832 was about making it possible to opt-out of generation, which is enabled by default. 

So, previously, if a user had `preserveRequestID: true`, he actually had the the equivalent of `RequestID=PreserveAndGenerate` in the new API,

Issue #5344, opened 2025-02-24
gateway controller overloads envoy with many xDS updates


dboslee:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: The expected value isn't returned, etc.

After upgrading to a newer version of the controller (v0.5.0 -> v1.2.4). We experienced an incident where a large number of pods being rolled concurrently over the course of a few hours caused envoy to get OOMd frequently (about every 10 minutes). We believe the primary cause of this was the gateway controller pushing many updates to envoy.

I think this was due to the addition of endpoint routing. Each update to an endpoint slice triggers an event which will cause a reconcile to run. We aren't using endpoint routing so any updates to endpoints should really be a noop on the config.

The controller should skip pushing updates in this case. Either by skipping events triggered by endpoint slices or by detecting that the envoy configuration has no changes.

I am curious to hear other ideas to address this issue as well. I am worried that there could be other cases where the controller could overload envoy with configuration updates. Maybe some sort of rate limiting or better detection of updates which cause no configuration change to envoy as I mentioned above would be a more robust solution.

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

- Configure an EnvoyProxy resource with `.spec.routingType: "Service"`
- Setup a route to a service with endpoints
- Roll the pods in the endpoint slice to create new endpoints
- Observe configuration updates pushed to envoy via controller logs or envoys lds metrics

*Environment*:
>Include the environment like gateway version, envoy version and so on.

gateway: v1.2.4

arkodg:
hey @dboslee if you have `routingType` set to `Service`, im curious why there were any updates being sent to envoy when pods were being rolling out, if this is happening, its a bug - since the EDS config hasn't changes (clusterIP is the same) and due to delta xDS, no new config should have been pushed

+1 on disabling endpointslice reconciles, or even saving them in watchable if `routingType` is set to `Service`

PR #5345, opened 2025-02-24
fix: skip reconciles on endpoint slices when endpoint routing not in use

dboslee:

As the title states this PR skips unnecessary reconciles on endpoint slice events when endpoint routing is not used by the given gateway.

This addresses the issue here however I wouldn't necessary close it with this fix as there are likely additional issues that could be addressed to improve the overall robustness of the controller.

https://github.com/envoyproxy/gateway/issues/5344

Release Notes: No

### DIFF
--- a/internal/provider/kubernetes/predicates.go
+++ b/internal/provider/kubernetes/predicates.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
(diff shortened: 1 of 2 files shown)


internal/provider/kubernetes/predicates.go
```diff
@@ -516,6 +517,142 @@ func (r *gatewayAPIReconciler) validateEndpointSliceForReconcile(obj client.Obje
...
+				Name:      string(ref.Name),
+				Namespace: namespace,
+			},
+		}
+		err := r.client.Get(ctx, client.ObjectKeyFromObject(&gw), &gw)
+		if err != nil {
+			r.log.Error(err, "unable to find associated gateway")
+			return false
+		}
+
```
arkodg, line 622: the `Gateway` resource can also have a `infrastructure.parameteresRef` field, which takes precedence over the one in GatewayClass, so if thats not nil, we should use that `EnvoyProxy` as source of truth

arkodg:
we we also save on memory by not inserting the endpointslice into watchable by skipping it here 
https://github.com/envoyproxy/gateway/blob/d3f95cc50af9e975c34df9639f67ac23862731f2/internal/provider/kubernetes/controller.go#L398

jukie:
Hey @dboslee could you fix the merge conflicts and add more test coverage?

PR #5348, opened 2025-02-25
fix: validate mixing of FQDNS addresses across destinations

### LINKED ISSUE #5050: Mixed FQDN and IP/UDS endpoints when using multiple backendRefs
*Description*:
Create `SecurityPolicy` or `EnvoyExtensionPolicy` with two `backendRefs` for ext-auth/proc:
-  An FQDN `Backend` 
- A K8s `Service` or `Backend` with UDS/IP endpoints

The policies are accepted, but envoy rejects the config: 
``
[2025-01-14 00:12:26.958][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment rejected: malformed IP address: foo.br. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
``

Similar to mixed-endpoint checks within a single destination, EG should verify that endpoints aggregated from multiple destinations are not mixed. 

*Repro steps*:

sharing.

*Environment*:

*Logs*:

rudrakhp:
**What type of PR is this?**
fix: validate mixing of FQDNS addresses across destinations

Fixes #5050 

Release Notes: No


guydc:
LGTM. Can you also add a gateway-api translator test, so that we can see the impact on status of relevant resources?

rudrakhp:
@guydc Looks like RD validation was not being invoked for external services, surprisingly existing E2Es were also invalid. Can you please verify if this is indeed the case or I maybe missing something. Thanks!

guydc:
yea, I think that those tests were wrong in the sense that they would not really work if Envoy would receive that XDS config. I would maybe change the test to have two different policies where each has a successful translation of a backend, to make sure that some of the original intent of the test is still enforced.

guydc:
Can we also cover the xRoute case (e.g. for HTTPRoute) in a test case? The translation flows are a bit different and I want to make sure that this validation happens for both technical backend refs and xRoute backendRefs

rudrakhp:
@guydc Added tests for:
* internal/gatewayapi/testdata/envoyextensionpolicy-with-extproc-with-mixed-backendrefs.in.yaml --> for checking policy backend refs (this was not being validated before)
* internal/gatewayapi/testdata/httproute-with-invalid-backend-ref-invalid-mixed-kind.in.yaml --> for checking xRoute backend refs (this was being validated before)

guydc (dismissed):
LGTM, Thanks!

internal/ir/xds.go
```diff
@@ -1429,10 +1430,17 @@ func (r *RouteDestination) Validate() error {
 	if len(r.Name) == 0 {
 		errs = errors.Join(errs, ErrDestinationNameEmpty)
 	}
+	routeHasAddressTypes := make(map[DestinationAddressType]bool)
```
shawnh2, line 1433: can we use utils/set instead?

rudrakhp: @shawnh2 right, forgot that's a thing in the new go versions. Modified the same in the xRoute validation as well.

shawnh2 (approved):
lgtm

cnvergence (approved):

PR #5352, opened 2025-02-26
Zone Aware Routing support

### LINKED ISSUE #1909: Locality Based Routing Support
*Description*:
Implement locality based routing support by default in EG. Now that we we can have individual endpoints as backend to EG. Can we support region/zone/subzone based routing based on EndpointSlice information, node labels etc.?

jukie:
**What type of PR is this?**

This is the second half to #5299 and adds support for Kubernetes Topology Aware Routing and Traffic Distribution via Envoy Zone Aware Routing.
* Adds a MutatingWebhookConfiguration which reacts to proxy pod `Binding` resource creations and injects underlying node topology information back to the pod.
* When Traffic Distribution or Topology Aware Routing are enabled for a Kubernetes Service, EnvoyGateway will inspect the endpoints and include zone information to LocalityLbEndpoints. EnvoyGateway will also then set the LocalityLbConfig to use Zone Aware Routing.

Fixes #1909

Release Notes: Yes


jukie:
Still WIP but this example "works". A challenge is the way that Envoy internally tries to keep an overall equal request level between other Proxy instances so I think we'll need to add the EnvoyProxy service as an EDS cluster in bootstrap and then have the endpoints populated afterwards.

AlexKrudu:
@jukie Hi! I am potentially considering using this feature for my edge-proxy envoy setup and I was wondering if you could give me some clarifications on how this feature's intent and perhaps some thoughts on your implementation.
As far as I understand envoy developer's intent - this feature is meant to be tracking ratio of both the callee and caller service (clusters) local workload to the total workload. And the result would be somewhat equal request rate to the all of upstream service instances in all of the localities. Your implementation considers Envoy itself as a originating cluster with the static size of 1. Isn't that the reason that this setup doesn't achieve the goal of leaving of requests in the local zone? Because from envoy's point of view local zone is the only zone requests are coming from and that's why it would try to spread it equally between all of the zones. You mentioned adding the EnvoyProxy service as an EDS cluster in bootstrap. Would you mind elaborating on how that would tackle the described problem? And thank you for you work!

AlexKrudu:
Because for now my thought are that this feature does not quite align with the edge proxy setup :)

arkodg:
@AlexKrudu , @jukie is working in upstream to change the behavior for ingress https://github.com/envoyproxy/envoy/pull/38756, which can be leveraged here

jukie:
Made some updates and everything should be in place here. Once the Envoy PR gets merged all that's needed will be to update this line: https://github.com/envoyproxy/gateway/pull/5352/files?file-filters%5B%5D=.go&file-filters%5B%5D=.json&file-filters%5B%5D=.md&show-viewed-files=true#diff-d0da16bff73f9d98d6346ffcf878a7bf6c689dbe11a8176598319ddb0b2a7a0dR174-R175

jukie:
Is there any preference for handling EnvoyProxy RBAC? Will need permissions to get nodes in the cluster for zone discovery

* A: Add Role/RoleBinding management permissions to EG
* B: All EnvoyProxy deployments use the same predefined ServiceAccount and RBAC is setup via helm
* C: Predefined Role/RoleBinding via helm chart with a ref target of `some-sa-prefix-*`

arkodg:
hey @jukie vote for B
we already create an SA per fleet https://github.com/envoyproxy/gateway/blob/61474e3601daace297cd15734a87d76862aa147e/internal/infrastructure/kubernetes/proxy/resource_provider.go#L72
we'll need to also create a Role and RoleBinding here going forward if zone aware routing is enabled in `EnvoyProxy`

internal/xds/translator/cluster.go
```diff
@@ -225,42 +237,80 @@ func buildXdsCluster(args *xdsClusterArgs) *clusterv3.Cluster {
 	// Set Load Balancer policy
 	//nolint:gocritic
 	if args.loadBalancer == nil {
-		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
+		leastRequest := &least_requestv3.LeastRequest{
```
arkodg, line 240: these changes are affecting xds config to change for the default case, curious why is this needed in this PR ?

internal/cmd/envoy/envoy_init.go
```diff
@@ -0,0 +1,160 @@
...
+
+	zone, exists := node.Labels[corev1.LabelTopologyZone]
+	if !exists {
+		return "", fmt.Errorf("zone label %q not found on node %q", corev1.LabelTopologyZone, node.Name)
+	}
+	return zone, nil
+}
+
+// writeConfig writes the locality information to the config file used for Envoy bootstrapping.
+func writeConfig(configPath, zone string) error {
```
arkodg, line 91: this is no longer needed, right ?

internal/xds/translator/cluster.go
jukie: My zone aware changes were added to LocalityLbConfig vs the legacy CommonLbConfig. We don't necessarily need to remove `cluster.LbPolicy` but `cluster.LoadBalancingPolicy` will supersede any value set here so thought I'd just clean up unused fields.

api/v1alpha1/envoyproxy_types.go
```diff
@@ -93,6 +93,11 @@ type EnvoyProxySpec struct {
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
 
+	// Init defines configuration for the envoy initialization process.
+	//
+	// +optional
+	Init *InitConfig `json:"init,omitempty"`
```
arkodg, line 99: can we start w/o this config ?

internal/cmd/envoy/envoy_init.go
jukie: This is still needed so that Envoy knows what the local zone is. The alternative would've been to grab _every_ node and make a fake locality endpoint to represent each zone.

api/v1alpha1/envoyproxy_types.go
jukie: From an earlier discussion it was suggested that there be an opt-in/out in EnvoyProxy, with the default being disabled upon initial release. Would you suggest a different way for handling that?

I was planning on using`ZoneOverride` for e2e tests but I imagine could handle that another way so wouldn't be opposed to dropping it.

I can't think of a need for changing `ConfigPath` so can remove that as well.

arkodg: I had assumed it would be something along the lines of `ZoneAwareRouting bool` , we could make it a struct so its easier to extend it

internal/cmd/envoy/envoy_init.go
arkodg: I had assumed that `--service-zone` would be enough with the new feature you added in the upstream

jukie: It makes things easier but it's still necessary unless there's a better idea. --service-zone still needs to have its value set but isn't known until _after_ the point of scheduling. Init container or EDS are the only options I'm aware of.

internal/xds/translator/cluster.go
jukie: Moved this logic into a separate, temporary func as discussed and can followup with a change for making `cluster.LoadBalancingPolicy` the default

jukie:
Hey @arkodg, this is ready for review now!

api/v1alpha1/envoygateway_types.go
```diff
@@ -225,6 +226,10 @@ type EnvoyGatewayKubernetesProvider struct {
 	// ShutdownManager defines the configuration for the shutdown manager.
 	// +optional
 	ShutdownManager *ShutdownManager `json:"shutdownManager,omitempty"`
+
+	// InitManager defines the configuration for the init manager.
```
arkodg, line 230: can we not expose this in the initial PR, and infer image name and pull policy from shutdown manager internally

api/v1alpha1/envoyproxy_types.go
```diff
@@ -93,6 +93,13 @@ type EnvoyProxySpec struct {
 	// +optional
 	Shutdown *ShutdownConfig `json:"shutdown,omitempty"`
 
+	// EnableZoneDiscovery defines whether to enable service zone discovery via
+	// topology.kubernetes.io/zone label on the underlying node
+	// Default: false
+	//
+	// +optional
+	EnableZoneDiscovery *bool `json:"enableZoneDiscovery,omitempty"`
```
arkodg, line 101: cc @envoyproxy/gateway-maintainers / @envoyproxy/gateway-reviewers to help with naming here

also this should probably be a struct and an empty struct to being with, so [minEndpointsPerZone](https://kubernetes.io/docs/concepts/services-networking/topology-aware-routing/#three-or-more-endpoints-per-zone) like settings can be added in the future

jukie: I think something like minEndpointsPerZone should live under load balancer settings personally since it would be configured per-backend rather than globally.

This field would solely be for enabling the feature in EnvoyProxy.

api/v1alpha1/envoygateway_types.go
jukie: Sure, but could you elaborate on the reasoning?

charts/gateway-helm/templates/infra-manager-rbac.yaml
```diff
@@ -57,3 +57,37 @@ subjects:
 - kind: ServiceAccount
   name: 'envoy-gateway'
   namespace: '{{ .Release.Namespace }}'
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
arkodg, line 61: when we chatted earlier, I failed to realize that creating a clusterRole programmatically would require us to acquire a clusterRole with a super power to create more clusterRoles 🙈 , this is not going to land well in any security audit process

arkodg: instead a common `envoy-gateway-node-reader` ClusterRole, ClusterRoleBinding and ServiceAccount is may be the way to go
are there any other alternatives ?

jukie: I was surprised by your suggestion as well 😃.
Should we revisit doing this via EDS again?

internal/cmd/envoy.go
```diff
@@ -69,3 +70,15 @@ func getShutdownManagerCommand() *cobra.Command {
 
 	return cmd
 }
+
+// getEnvoyInitCommand returns the envoy init cobra command to be executed.
+func getEnvoyInitCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Provides HTTP endpoint used in preStop hook to block until ready for pod shutdown.",
```
arkodg, line 78: needs to be changed

api/v1alpha1/envoygateway_types.go
arkodg: ideally dont want to confuse users with another setting they are unlikely to use, and this setting/implementation of using init containers will hopefully go away in the future

charts/gateway-helm/templates/infra-manager-rbac.yaml
arkodg: EDS will be tricky because it will be unique to every envoy pod, and we will need to manage an additional unique cache for this in xds server which will be tricky

another idea is to create a Secret with nodeName as key and zone as value (this can be done in the gateway-api layer and saved in the Infra IR)

so now we can either do something like
```
    env:
    - name: NODE_ZONE
      valueFrom:
        secretKeyRef:
          name: node-zone
          key: $(NODE_NAME)
```          
and if something like this isnt possible, we can mount the secret into the init container, and read it as a file

jukie: I've got an example mutating webhook working to inject the zone to a pod at `binding` (when the node is selected) and we could do something like this:

```
    env:
    - name : ZONE
      valueFrom :
        fieldRef :
          fieldPath : metadata.labels['topology.kubernetes.io/zone']
    args:
    - "--service-zone=$(ZONE)"
```

arkodg: sure that works

api/v1alpha1/envoygateway_types.go
jukie: Removed

internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
```diff
@@ -17,6 +17,11 @@ envoyProxyForGatewayClass:
             socket_address:
               address: 127.0.0.1
               port_value: 19000
+        cluster_manager:
+          local_cluster_name: local_cluster
+        node:
+          locality:
+            zone: $(ENVOY_SERVICE_ZONE)
```
jukie, line 24: This gets replaced with the container's env var or renders as an empty string leading to an empty locality. Still valid configuration and would be the same end result as today.

jukie:
@arkodg does this approach seem more reasonable?

charts/gateway-helm/templates/envoy-proxy-topology-mutator-webhook.yaml
```diff
@@ -0,0 +1,37 @@
...
+  labels:
+    app.kubernetes.io/component: topology-webhook
+  {{- include "eg.labels" . | nindent 4 }}
+webhooks:
+  - name: topology.webhook.envoyproxy.io
+    admissionReviewVersions: ["v1"]
+    sideEffects: NoneOnDryRun
+    timeoutSeconds: 5
+    clientConfig:
+      caBundle: {{ default "" (index $secret.data "ca.crt" ) | b64dec }}
```
jukie, line 20: I'll clean this up a bit more but we can't use the internal certgen because we need the ca cert data at deploy time. An alternative here would be to conditionally use cert-manager ca cert injection which is what a lot of other charts do.

charts/gateway-helm/templates/envoy-proxy-topology-mutator-deployment.yaml
```diff
@@ -0,0 +1,91 @@
+{{- if .Values.topologyWebhook.enabled }}
+apiVersion: apps/v1
```
arkodg, line 2: is it possible to reuse the EG deployment, and have 1 less moving part from an ops perspective

arkodg: https://github.com/envoyproxy/gateway/blob/c2075cd01f849269d7afd2a53654f8ab67e68ec8/internal/provider/kubernetes/kubernetes.go#L45

charts/gateway-helm/templates/envoy-proxy-topology-mutator-webhook.yaml
```diff
@@ -0,0 +1,37 @@
...
+      service:
+        name: {{ include "eg.webhookService" . }}
+        namespace: '{{ .Release.Namespace }}'
+        path: "/mutate-pod-topology"
+        port: 443
+    rules:
```
arkodg, line 26: is there a way to add a label to the proxy pod and filter/match here ?

internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -166,6 +171,23 @@ static_resources:
                 address: {{ $sink.Address }}
                 port_value: {{ $sink.Port }}
   {{- end }}
+  - connect_timeout: 10s
```
arkodg, line 174: is it safe to add this by default without a `{{- if .ZoneDiscoveryEnabled  }}` container  ? @nezdolik @mathetake

jukie: Yes the new local_cluster would just render as having zero endpoints and wouldn't change any existing behavior.

jukie: Yes the new local_cluster would just render as having zero endpoints and wouldn't change any existing behavior.

charts/gateway-helm/templates/envoy-proxy-topology-mutator-webhook.yaml
jukie: No because pod labels/annotations aren't copied to the relevant Binding resource so the filtering happens inside the hook.

charts/gateway-helm/templates/envoy-proxy-topology-mutator-webhook.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: envoy-gateway-topology-webhook
+  {{- with .Values.topologyWebhook.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  labels:
+    app.kubernetes.io/component: topology-webhook
+  {{- include "eg.labels" . | nindent 4 }}
+webhooks:
+  - name: topology.webhook.envoyproxy.io
```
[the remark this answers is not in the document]
jukie: Yeah I need to clean it up but that's exactly the plan!

charts/gateway-helm/templates/envoy-proxy-topology-mutator-deployment.yaml
jukie: Definitely, will update with that

charts/gateway-helm/templates/envoy-proxy-topology-mutator-webhook.yaml
arkodg: thanks, resolving this one

jukie:
@arkodg @guydc I've added the suggested changes from our discussion:
* Cert gen will handle the caBundle injection
  * This includes additional RBAC permissions to manage the webhook but is limited by `resourceNames` to only the intended webhook
  * The `MutatingWebhookConfiguration` adds a `hook-weight=-1` annotation to ensure it's created before the cert gen job executes.
* FailurePolicy=Ignore for the webhook so we fail open - this is intentional to avoid any limitations in creating the Proxy pods. We can later add a docs snippet informing users how to monitor webhook failures and/or envoy proxy stats for issues with zone aware routing.
* Webhook code moved to the internal k8s provider logic instead of a separate container
* Webhook name includes a suffix of the EG namespace to allow for multiple (such as per-namespace deployments)

jukie:
The pre-install hook will make this tricky

arkodg:
@jukie can we use same annotations as certgen rbac https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/certgen-rbac.yaml ?

jukie:
That's what it already had but it appears the ClusterRole wasn't being created until after the cert gen job. Should be good this time.

api/v1alpha1/envoygateway_types.go
```diff
@@ -89,6 +89,9 @@ type EnvoyGatewaySpec struct {
 	//
 	// +optional
 	ExtensionAPIs *ExtensionAPISettings `json:"extensionApis,omitempty"`
+
+	// TopologyInjector defines the configuration for topology injector MutatatingWebhookConfiguration
+	TopologyInjector *EnvoyGatewayTopologyInjector `json:"proxyTopologyInjector,omitempty"`
```
arkodg, line 94: since this is specific to k8s, should be moved under provider.Kubernetes

api/v1alpha1/envoygateway_types.go
```diff
@@ -610,6 +613,11 @@ type ShutdownManager struct {
 	Image *string `json:"image,omitempty"`
 }
 
+// EnvoyGatewayTopologyInjector defines the configuration for topology injector MutatatingWebhookConfiguration
+type EnvoyGatewayTopologyInjector struct {
+	Disable bool `json:"disabled,omitempty"`
```
arkodg, line 618:
```suggestion
	Disable *bool `json:"disabled,omitempty"`
```

charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -62,3 +62,63 @@ subjects:
 - kind: ServiceAccount
   name: '{{ include "eg.fullname" . }}-certgen'
   namespace: '{{ .Release.Namespace }}'
+---
+{{- if .Values.topologyInjector.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: '{{ include "eg.fullname" . }}-certgen:{{ .Release.Namespace }}'
```
arkodg, line 70: do we really need to append the namespace in the suffix ?
we dont do this for cert-gen or envoy-gateway deployments since they are unique to a namespce

arkodg: oh its cluster scoped, nvm

charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -62,3 +62,63 @@ subjects:
...
+  labels:
+  {{- include "eg.labels" . | nindent 4 }}
+  {{- if .Values.certgen.rbac.labels }}
+  {{- toYaml .Values.certgen.rbac.labels | nindent 4 }}
+  {{- end }}
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-1"
```
arkodg, line 78: can we rm the weight annotation ? dont see it on any certgen resources

charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -62,3 +62,63 @@ subjects:
...
+      - admissionregistration.k8s.io
+    resources:
+      - mutatingwebhookconfigurations
+    resourceNames:
+      - 'envoy-gateway-topology-injector.{{ .Release.Namespace }}'
+    verbs:
+      - update
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
arkodg, line 101: same comments as above

charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.topologyInjector.enabled }}
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: 'envoy-gateway-topology-injector.{{ .Release.Namespace }}'
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-1"
```
arkodg, line 8: can we rm the hook weight annotation ?

charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
```diff
@@ -0,0 +1,37 @@
...
+  {{- if .Values.topologyInjector.annotations }}
+    {{- toYaml .Values.topologyInjector.annotations | nindent 4 -}}
+  {{- end }}
+  labels:
+    app.kubernetes.io/component: topology-injector
+  {{- include "eg.labels" . | nindent 4 }}
+webhooks:
+  - name: topology.webhook.envoyproxy.io
```
arkodg, line 16:
```suggestion
  - name: topology.webhook.gateway.envoyproxy.io
```

charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.topologyInjector.enabled }}
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: 'envoy-gateway-topology-injector.{{ .Release.Namespace }}'
```
arkodg, line 5: imo since its unique to a ns, we dont need to suffix the ns

arkodg: oh nvm its cluster scoped

charts/gateway-helm/templates/certgen-rbac.yaml
arkodg: nvm we need this

jukie: Yes but as you said those are already unique to a namespace whereas cluster roles aren't. In order to support namespaced deployments we'll need unique ClusterRoles per-install.

arkodg, line 70: curious why  `:` was picked over `-` ?

charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
jukie: This is a cluster scoped resource though. The actions it takes is scoped to an individual namespace but there would be conflicts between namespaced deployments otherwise.

charts/gateway-helm/templates/certgen-rbac.yaml
jukie: 🤷 just a naming convention I've used personally and seen elsewhere for RBAC resources. I can change to `-` here.

charts/gateway-helm/templates/certgen-rbac.yaml
jukie: It's necessary here due to helm resource sorting. Otherwise the cert gen job executes before the ClusterRole is created.

internal/cmd/certgen.go
```diff
@@ -71,6 +80,9 @@ func certGen(ctx context.Context, logOut io.Writer, local bool) error {
 		if err = outputCertsForKubernetes(ctx, cli, cfg, overwriteControlPlaneCerts, certs); err != nil {
 			return fmt.Errorf("failed to output certificates: %w", err)
 		}
+		if err = patchTopologyInjectorWebhook(ctx, cli, cfg, certs.CACertificate); err != nil {
```
arkodg, line 83: lets track this as a follow up issue - there's a case where the user brings their own certs and drops it as a Secret, in that case we'll need to read the Secret, copy the ca cert from it and use it

test/e2e/tests/zone_aware_routing.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+				Namespace: ns,
+			}
+			req := http.MakeRequest(t, &expectedResponse, gwAddr, "HTTP", "http")
+
+			// The EnvoyProxy pods have a nodeSelector targeting zone "0" so all requests
+			// should be routed to upstreams there.
+			expected := map[string]int{
+				"0": sendRequests,
+				"1": 0,
```
arkodg, line 58: arent there multiple workers, do we need more entries here ?

charts/gateway-helm/templates/certgen-rbac.yaml
arkodg: its fine, lets use this name, I thought dots were not allowed, but if they are, this is fine

charts/gateway-helm/templates/certgen-rbac.yaml
arkodg: thinking out loud - why dont the other resources in this file cause the same issue

internal/provider/kubernetes/topology_injector.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:      binding.Name,
+	}
+
+	pod := &corev1.Pod{}
+	if err := m.Get(ctx, podName, pod); err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	// Skip non-proxy pods
+	if v, ok := pod.Labels["app.kubernetes.io/component"]; !ok || v != "proxy" {
```
arkodg, line 48: can we also add some more in here ?
app.kubernetes.io/name: envoy

arkodg:
this PR looks really good @jukie, thanks for adding an e2e in there
added some minor comment, overall LGTM with the approach !

internal/provider/kubernetes/kubernetes.go
```diff
@@ -87,11 +103,27 @@ func New(ctx context.Context, restCfg *rest.Config, svrCfg *ec.Server, resources
 			mgrOpts.Cache.DefaultNamespaces[watchNS] = cache.Config{}
 		}
 	}
+	if !ptr.Deref(svrCfg.EnvoyGateway.TopologyInjector, egv1a1.EnvoyGatewayTopologyInjector{}).Disable {
+		mgrOpts.WebhookServer = webhook.NewServer(webhook.Options{
```
arkodg, line 107: what is the impact of specifying a caCert here than doesnt match the caBundle in the mutating webhook config

jukie: No requests would be sent. The caBundle here must match the certificate being used in the webhook server and the certificate must also match the target ref in the MutatingWebhookConfiguration (e.g. `envoy-gateway.envoy-gateway-system.svc.local`)

test/e2e/tests/zone_aware_routing.go
jukie: If I'm reading the github action workflow configuration right it's only the control-plane node (NUM_WORKERS is unset) and I've now adjusted that to 1.

internal/cmd/certgen.go
jukie: Is that supported today in the normal cert gen flow? Can certainly be supported though and would just be whatever caBundle is determined above this code.

internal/cmd/certgen.go
```diff
@@ -117,7 +114,7 @@ func outputCertsForKubernetes(ctx context.Context, cli client.Client, cfg *confi
 }
 
 func patchTopologyInjectorWebhook(ctx context.Context, cli client.Client, cfg *config.Server, caBundle []byte) error {
-	if disableTopologyInjector {
```
arkodg: I think we still need the flag, the cfg is not available in the cert ten phase afaik
https://github.com/envoyproxy/gateway/issues/4935

test/e2e/tests/zone_aware_routing.go
arkodg: cool sgtm cc @zirain

internal/cmd/certgen.go
arkodg: its supported today, since we are so close to the rc release, dont want to block on this, we can get this into v1.5, the subset of users using BYO certs + wanting zone aware routing, may need to do some gymnastics in v1.4, but it will be temporary

arkodg: https://gateway.envoyproxy.io/docs/install/custom-cert/

charts/gateway-helm/templates/certgen-rbac.yaml
jukie: It's the helm resource sorting and in the test log you can see the clusterrol/rolebinding never get installed - https://github.com/envoyproxy/gateway/actions/runs/14654407193/job/41127299304?pr=5352

internal/cmd/certgen.go
jukie: Added back in

arkodg (dismissed):
LGTM thanks !

jukie:
@arkodg I fixed the e2e if you could give this another stamp please

arkodg (approved):
LGTM thanks !

go.sum
```diff
@@ -372,12 +372,12 @@ github.com/emicklei/go-restful/v3 v3.12.0/go.mod h1:6n3XBCmQQb25CM2LCACGz8ukIrRr
...
-github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250303232259-0c5cb90447bb h1:4oJf7zkbbNoBRGkeA0O8jB48FjM78qfYwGL6Q68AaH4=
-github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250303232259-0c5cb90447bb/go.mod h1:vggMdJSZaAGnQ5sf1pBl1OV7w3IFqt8qo3id+asttJw=
+github.com/envoyproxy/go-control-plane v0.13.5-0.20250408134212-157c26b62099 h1:Shucu2sY2a/KjKzO10Y0st2iN0Flzj5QsTEgMFtd0MY=
+github.com/envoyproxy/go-control-plane v0.13.5-0.20250408134212-157c26b62099/go.mod h1:Kf4hNGzgvzKhoKdlSXD+IZtG55h9r2SOpO1kRKLI03o=
+github.com/envoyproxy/go-control-plane/contrib v1.32.5-0.20250408134212-157c26b62099 h1:YCkhFxkFlN1h7QVwZTo3a14QlxXf0p1MrycEBuE3wJc=
+github.com/envoyproxy/go-control-plane/contrib v1.32.5-0.20250408134212-157c26b62099/go.mod h1:zD9QKAwQBnb9DWvRrkMa7cOuuMOt+ITrm4ZLoZ0aa30=
+github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250408134212-157c26b62099 h1:LFmrrIXbGktG4DrXsvFFyvZLe+HnimGn6Ed+xuiP6TU=
+github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250408134212-157c26b62099/go.mod h1:pY0vLp032ToSdejxKtzWZs/TzoAtdK0+50ELjTkXiYA=
```
jukie, line 380: Ref https://github.com/envoyproxy/envoy/pull/38756

jukie:
Because I've added multiple nodes to the E2E tests the UDS pods need to include affinity for the proxy pods

test/e2e/testdata/ext-proc-service.yaml
```diff
@@ -52,6 +52,21 @@ spec:
       labels:
         app: grpc-ext-proc
     spec:
+      affinity:
```
zirain, line 55: why we need this?

jukie: Because the UDS socket is only available on the local host. In order to test zone aware routing I've configured NUM_WORKERS=2 for e2e tests so the UDS pods need to run where the envoy proxy pods are.

That's what I was referring to here - https://github.com/envoyproxy/gateway/pull/5352#issuecomment-2832470367

zirain: It is a bit confusing that this test case has been affected。

internal/provider/kubernetes/topology_injector.go
```diff
@@ -0,0 +1,97 @@
...
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	var patch string
+	if zone, ok := node.Labels[corev1.LabelTopologyZone]; ok {
+		patch = fmt.Sprintf(`[{"op":"replace", "path":"/metadata/labels/%s", "value":"%s"}]`, jsonpointer.Escape(corev1.LabelTopologyZone), zone)
+	}
+
+	rawPatch := client.RawPatch(types.JSONPatchType, []byte(patch))
+	if err := m.Patch(ctx, pod, rawPatch); err != nil {
```
zirain, line 71: it would be better to add some webhook metrics in the following PR.

internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -166,6 +171,23 @@ static_resources:
...
+    lb_policy: ROUND_ROBIN
+    load_assignment:
+      cluster_name: local_cluster
+      endpoints:
+      - lb_endpoints:
+        - endpoint:
+            address:
+              socket_address:
+                address: 127.0.0.1
+                port_value: 10080
```
zirain, line 184: need update open ports in doc?

zirain (commented):
Gernally LGTM.

zirain (dismissed):

internal/xds/bootstrap/bootstrap.yaml.tpl
jukie: This doesn't actually open a port it's just a reference to a fake lb_endpoint to satisfy the internal envoy logic of at least 1 endpoint. I used 10080 from the default port 80 listener but can change this in a followup to be the readiness probe or another dummy port if it's confusing.

zirain: can you address this somewhere, this would be enough.

test/e2e/testdata/ext-proc-service.yaml
jukie: Previously there was only a single node in the kind cluster so all pods were already guaranteed to run on the same node. When using a unix domain socket Backend that must be mounted to the envoy proxy pods and thus scheduled to the same node. Now that the number of nodes in the kind cluster has been increased we need to explicitly configure that.

zirain: can you address these in the yaml?

jukie: Isn't that already the requirement described in the docs? This should already be the default configuration but was only working because of the single node kind cluster.

Ref: https://gateway.envoyproxy.io/docs/tasks/traffic/backend/#restrictions
> Envoy Gateway does not manage the lifecycle of unix domain sockets referenced by the Backend resource. Envoy Gateway admins are responsible for creating and mounting the sockets into the envoy proxy pod. The latter can be achieved by patching the envoy deployment using the [EnvoyProxy](https://gateway.envoyproxy.io/docs/api/extension_types#envoyproxy) resource.

internal/xds/bootstrap/bootstrap.yaml.tpl
jukie: Updated

test/e2e/testdata/ext-proc-service.yaml
zirain: can you address it in this test yaml?

internal/provider/kubernetes/topology_injector.go
jukie: Adding metrics over here - https://github.com/envoyproxy/gateway/pull/5835

Issue #5326, opened 2025-02-21
Endpoints disvovery problem in kubernetes during rollout updates/restarts


Hartigan:
*Description*:
I found significant drop of rps or inability of envoy proxies to connect to backends during rollout restarts or similar activity 

*Repro steps*:
It's reproducible on envoy-gateway v1.2.2 and v1.3.0. I setup simple deployment of nginx with static answer and HTTPRoute to this deployment. During the load test I'm trying to rollout restart nginx deployment(100 pods).

In result:

- Traffic drops to zero or significant for 1-2 minutes
- Slow recovering

*Workarounds*:
- Switch routing type to Service https://gateway.envoyproxy.io/docs/api/extension_types/#routingtype
- Set rolloutStrategy for deployment to `maxUnavailable: 0` and `maxSurge: 1` to slowdown updates of `EndpointSlices`

I tried to create custom build of `envoy-gateway` with disabled cache for `EndpointSlices` https://github.com/envoyproxy/gateway/blob/57c7e645188d8ff1b4fd6c9c52a3aad0eabcc76f/internal/provider/kubernetes/controller.go#L465 Controller starts to update cluster for envoy proxies by large batches (added ~20 removed ~20 endpoints) and problem was gone, but sometimes controller hits limit by QPS for kube-api. I guess there is some problem with cache invalidation for `EndpointSlices`.

*Logs*:
logs of envoy proxy with cluster updates:
```
[2025-02-20 19:12:56.300][1][debug][upstream] [source/common/upstream/cluster_manager_impl.cc:1679] membership update for TLS cluster httproute/default/http-echo-server-demo/rule/0 added 0 removed 0
[2025-02-20 19:12:56.300][13][debug][upstream] [source/common/upstream/cluster_manager_impl.cc:1679] membership update for TLS cluster httproute/default/http-echo-server-demo/rule/0 added 0 removed 0
[2025-02-20 19:13:06.840][1][debug][upstream] [source/common/upstream/cluster_manager_impl.cc:1679] membership update for TLS cluster httproute/default/http-echo-server-demo/rule/0 added 0 removed 1
[2025-02-20 19:13:06.840][13][debug][upstream] [source/common/upstream/cluster_manager_impl.cc:1679] membership update for TLS cluster httproute/default/http-echo-server-demo/rule/0 added 0 removed 1
[...]

PR #5358, opened 2025-02-26
Wasm: cache Wasm OCI image permission check results

zhaohuabing:
This PR introduce a caching mechanism for the Wasm OCI image permission check results.

**Background:**
Previously, to prevent unauthorized EEPs from accessing a private Wasm OCI image, EG would verify the pullSecrect against the OCI registry in the Gateway API translator, even if the OCI image has already been pulled and cached locally. Because this verification involved network access, it could block the translator, especially when a lots of endpoints are scaling up/down.

**What's changed:**
This PR adds a permission cache in the Wasm package.  A backgroud goroutine perodically checks the EEP's pullSecret against OCI registry and cache the check result. The Gateway API translator can now use the cached permission check result to tell if an EEP can access the OCI image.
 * Initial Image Pull: The first attempt to pull an OCI image will still block because neither the file cache nor the permission cache has any entry yet.
 * Subsequent Requests: After the initial pull, subsequent requests for the same image will not block because the permission check is served from the cache.
 
**Cache expiry:**
* The cached permissions are rechecked every hour in case the registry invalidates an existing pullSecret.
* If a chaced entry is not accessed in 24 hours, it will be removed from the cache.

These default values are now fixed consts, they can be exposed to EG configuration later if needed.

fix: #5326

### DIFF
--- a/internal/wasm/cache.go
+++ b/internal/wasm/cache.go
@@ -44,8 +44,8 @@ const (
 	ociURLPrefix = "oci://"
 	// sha256 scheme prefix
 	sha256SchemePrefix = "sha256:"
-	// permissionCacheTTL is the TTL for permission cache.
-	permissionCacheTTL = 10 * time.Minute
+	// permissionCacheRefreshInterval is the interval to recheck the permission for the cached permission entries.
+	permissionCacheRefreshInterval = 10 * time.Minute
 )
 
 // Cache models a Wasm module cache.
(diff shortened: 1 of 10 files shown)


internal/wasm/cache.go
```diff
@@ -44,6 +44,8 @@ const (
 	ociURLPrefix = "oci://"
 	// sha256 scheme prefix
 	sha256SchemePrefix = "sha256:"
+	// permissionCacheTTL is the TTL for permission cache.
+	permissionCacheTTL = 10 * time.Minute
```
zirain, line 48: how can we change this?

internal/wasm/cache.go
```diff
@@ -67,15 +69,67 @@ type localFileCache struct {
...
 
+	// permissionCheckCache is a cache for permission check for private OCI images.
+	permissionCheckCache *permissionCache
+
 	// logger
 	logger logging.Logger
 }
 
+// permissionCache is a cache for permission check for private OCI images.
+type permissionCache struct {
```
zirain, line 80: learn from https://github.com/istio/istio/pull/55168, should we use `otter` here?

zhaohuabing: I think `otter` is over-skilled for this very simple permission check result cache :-)

internal/wasm/premissioncache.go
```diff
@@ -0,0 +1,266 @@
...
+		return e.checkError == nil, e.checkError
+	}
+
+	e := &permissionCacheEntry{
+		image: image,
+		fetcherOption: &ImageFetcherOption{
+			Insecure:   insecure,
+			PullSecret: pullSecret,
+		},
+	}
```
arkodg, line 239: the drawback of this, is, if permissions change on the server side, I'll have to wait 1 hr to update my wasm image

arkodg: the primary issue here is - fetching a WASM Image including checking permissions blocks the route linked to this EEP as well as other routes and endpoints and other xds from being pushed to the data plane. Ideally we'd like this two tasks to be happening in parallel.
Can we fetch and check permissions in a separate go routine to unblock the translator ? If the wasm fetcher / checker go routine finds errors it could flag it in the status using watchable
Hoping the WASM URL is based off policy ns-name so data plane requests will fail in case of a deny

zhaohuabing: > Can we fetch and check permissions in a separate go routine to unblock the translator ? If the wasm fetcher / checker go routine finds errors it could flag it in the status using watchable
Hoping the WASM URL is based off policy ns-name so data plane requests will fail in case of a deny

We can't fetch the OCI image async. The sha256 checksum of the wasm module is a mandatory field for xDS Wasm remote code source, and we can't get the checksum without fetching the image.

zhaohuabing: > the drawback of this, is, if permissions change on the server side, I'll have to wait 1 hr to update my wasm image

Permission change actually won't invalidate cached wasm image, it only prevents the requests from unauthorized EEPs,  unauthorized EEP translation will fail.

The cached images are purged after not being touched for expiry duration, [which is handled in the wasm image chace](https://github.com/envoyproxy/gateway/blob/a79843856d3fc2c74ca7fc09fe2695699056984c/internal/wasm/cache.go#L457-L485).

arkodg (approved):
LGTM thanks !

guydc (approved):

PR #5361, opened 2025-02-26
feat: support global ratelimit buckets

ryanhristovski:
Core logic in `internal/xds/translator/ratelimit.go`

#### Descriptor/Domain Changes
- If the ratelimit is shared the following values are set
```
Descriptor Key = BackendTrafficPolicy.Name
Descriptor Value = BackendTrafficPolicy.Namespace
Domain = BackendTrafficPolicy.Name + "-" BackendTrafficPolicy.Namespace
```
This will allow the ratelimit to be shared across all xRoutes/Gateways that share this BackendTrafficPolicy (BTP).

- If the ratelimit is not shared, it will use the listener name as the domain and maintain previous descriptor key/value of the route name.

#### HTTPFilter Changes
- Since we are now potentially using multiple domains, we needed to create multiple Ratelimit HTTPFilters to associate with them, `buildRateLimitFilter` now builds a list of filters.

#### API Changes
- `GlobalRateLimit`, now includes `Shared` value responsible for enabling this logic
- `TrafficFeatures` IR, responsible for holding all information associated to BTP, now includes BTP metadata including BTP.name and BTP.namespace (this is used for setting domains & descriptors)

#### Runner changes
- Runner now iterates through a list of Ratelimit configuration now that there can be multiple

#### Testing
- Ratelimit-config translator has been updated to build multiple configurations separated by `---` in order to account for the multiple domains that are now being used.
- All tests have been updated to account for new traffic.BackendTrafficPolicy fields

#### TODO:
- ~~Cleanup debugging logs~~
- ~~Update shared=true logic to be whatever is decided https://github.com/envoyproxy/gateway/pull/5227#discussion_r1966046203~~
- ~~Write some updated testing~~
- ~~Add all updated tests (left out for now for easier reviewing)~~


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -309,6 +309,12 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(
 		err, errs error
 	)
 
+	// // **Validation: Shared rate limits can only be applied to Gateways**
+	if policy.Spec.RateLimit != nil  && ptr.Deref(policy.Spec.RateLimit.Global.Shared, false) {
+		return fmt.Errorf("rateLimit.Global.Shared=true is only allowed for Gateway targetRefs, not HTTPRoute: policy=%s/%s",
```
arkodg, line 314: do we have to add this limitation ? there maybe a case where the policy can target multiple routes using target selectors, and the user may want to share buckets across them

internal/xds/translator/testdata/out/ratelimit-config/global-shared-distinct-match.yaml
```diff
@@ -0,0 +1,20 @@
+name: first-listener
+domain: first-listener
```
arkodg, line 2: the domain is still the listener name, can you add multiple listeners in the input and generate the o/p again ?

imo the domain here should be the policy ns-name which is a unique value, and you can populate the domain value in the gateway api layer (by default it can be set to `nil`, and the ratelimit layer can then generate it, else reuse if it exists)

internal/gatewayapi/backendtrafficpolicy.go
ryanhristovski: from my understanding a btp can only target 1 route at a time - so i'm not sure of a usecase where this would work 🤔

if the (assuming we go with option b) `scope: gateway`, then why would we allow that to be attached an xRoute?

arkodg: its `scope: shared`  so either sharing aross all routes in a gateway or amongst all routes the BTP is attched to (via targetSelectors, we dont need to solve this now, can add this as a TODO)

internal/xds/translator/testdata/out/ratelimit-config/global-shared-distinct-match.yaml
ryanhristovski: this is some of the test cleanup that i refer to in the todo, but basically if it can't find the gateway name, it defaults to the listener name - maybe we want to change this
```
func getRateLimitDomain(irListener *ir.HTTPListener, isShared bool) string {
	if isShared && irListener.GatewayName != "" {
		return irListener.GatewayName
	}
	return irListener.Name
}
```

but by adding the gatewayName variable to the test file like so:
```
name: "first-listener"
gatewayName: "test-gateway"
address: "0.0.0.0"
port: 10080
...
```

it returns this:
```
name: test-gateway
domain: test-gateway
descriptors:
  - key: global-limit
    value: global-limit
    rate_limit: null
```


in respect to using the policy ns-name, i personally agree with Zhao's implementation example here, and is how i originally envisioned it as well: https://github.com/envoyproxy/gateway/pull/5227#issuecomment-2676886848. i think debugging the ratelimiting is more clear when the domain is the gateway name (since it applies to the entire gateway), and not the policy name which could be harder to decipher imo

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -509,6 +517,12 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(
 
 	ds = translateDNS(policy.Spec.ClusterSettings)
 
+	// Create BackendTrafficPolicy instance
```
arkodg, line 520: the IR structs use a `name` field as a differentiator https://github.com/envoyproxy/gateway/blob/main/internal/ir/xds.go
can we use that same style here ?

it can be generated with a helper similar to https://github.com/envoyproxy/gateway/blob/0c22094563ee1462223c97e867865c613b302b16/internal/gatewayapi/helpers.go#L394

internal/globalratelimit/runner/runner.go
```diff
@@ -163,12 +163,18 @@ func (r *Runner) subscribeAndTranslate(ctx context.Context) {
...
 
+	// Iterate over each HTTP listener
 	for _, listener := range xdsIR.HTTP {
-		cfg := translator.BuildRateLimitServiceConfig(listener)
-		if cfg != nil {
-			// Add to xDS Config resources.
-			if err := resourceVT.AddXdsResource(resourcev3.RateLimitConfigType, cfg); err != nil {
-				return nil, err
+		// Build the rate limit service config for this listener
+		configs := translator.BuildRateLimitServiceConfig(listener)
```
arkodg, line 169: curious how this works ? we are interacting across listeners that are returning multiple configs, that point to the same domain with the same descriptor keys and values ?

internal/xds/translator/testdata/in/ratelimit-config/global-shared-multiple-shared-policies.yaml
```diff
@@ -0,0 +1,64 @@
+name: "first-listener"
```
arkodg, line 1: is it possible to test same bucket across 2 listeners ?

internal/xds/translator/testdata/in/ratelimit-config/global-shared-multiple-shared-policies.yaml
```diff
@@ -0,0 +1,64 @@
...
+address: "0.0.0.0"
+port: 10080
+hostnames:
+- "*"
+path:
+  mergeSlashes: true
+  escapedSlashesAction: UnescapeAndRedirect
+routes:
+- name: "first-route"
```
arkodg, line 10: can we add 2 routes with the same bucket ?

internal/xds/translator/ratelimit.go
```diff
@@ -556,8 +770,17 @@ func getRateLimitServiceClusterName() string {
 	return "ratelimit_cluster"
 }
 
-func getRateLimitDomain(irListener *ir.HTTPListener) string {
-	// Use IR listener name as domain
+func getRateLimitDomainFilters(irListener *ir.HTTPListener, isShared bool, usedDomains map[string]bool) string {
+	// Iterate over the routes to find a domain that hasn't been used yet
+	for _, route := range irListener.Routes {
+		if hasValidBackendTrafficPolicy(route) {
+			domain := route.Traffic.BackendTrafficPolicy.Name + "-" + route.Traffic.BackendTrafficPolicy.Namespace
```
arkodg, line 777: suggest using `name` here

internal/xds/translator/ratelimit.go
```diff
@@ -556,8 +770,17 @@ func getRateLimitServiceClusterName() string {
...
+			if isShared && !usedDomains[domain] {
```
arkodg, line 778: curious, why do we need `usedDomains` here ?

internal/xds/translator/ratelimit.go
```diff
@@ -460,16 +662,28 @@ func buildRateLimitServiceDescriptors(global *ir.GlobalRateLimit) []*rlsconfv3.R
 
 		// Case when both header and cidr match are not set and the ratelimit
 		// will be applied to all traffic.
+		// 3) No Match (apply to all traffic)
 		if !rule.IsMatchSet() {
 			pbDesc := new(rlsconfv3.RateLimitDescriptor)
-			// GenericKey case
-			pbDesc.Key = getRouteRuleDescriptor(rIdx, -1)
-			pbDesc.Value = getRouteRuleDescriptor(rIdx, -1)
+
```
arkodg, line 668: would it be easier to reuse `getRouteRuleDescriptor` and return `traffic.name` if `shared: true` ?

internal/xds/translator/testdata/in/ratelimit-config/global-shared-multiple-shared-policies.yaml
ryanhristovski: I think we'd need to update the structures of the tests to accept a list of listeners in order to do this

internal/xds/translator/ratelimit.go
ryanhristovski: I personally find `domain` to be clearer when developing, why do you prefer `name`?

internal/xds/translator/ratelimit.go
ryanhristovski: To ensure only unique HTTPFilters are created, if there are routes that have the same ratelimiting domain they will be added twice. For example if you had 3 routes:
```
route-1:
	ratelimitDomain: example-1
route-2:
	ratelimitDomain: example-1
route-3:
	ratelimitDomain: example-2
```

without ensuring unique names it would generate 3 HTTPFilters, 2 `example-1` filters and 1 `example-2`

internal/xds/translator/testdata/in/ratelimit-config/global-shared-multiple-shared-policies.yaml
ryanhristovski: i've updated tests to input a list of listeners, this gives us a lot more accurate coverage

internal/globalratelimit/runner/runner.go
ryanhristovski: fixed this by accepting all listeners at once in the runner.go to be able to only allow unique domains so theres no replications

internal/xds/translator/testdata/in/ratelimit-config/global-shared-multiple-shared-policies.yaml
```diff
@@ -0,0 +1,81 @@
...
+              exact: "one"
+            limit:
+              requests: 5
+              unit: second
+  - name: "second-route"
+    traffic:
+      backendTrafficPolicy:
+        name: "test-policy-1"
+        namespace: "test-namespace"
+      rateLimit:
```
arkodg, line 31: should the `ratelimit` be same if the policy is same as shared for the same policy ?

also can we also add the same IRs from the test in `/in/xds-ir` to view the xDS descriptors etc

ryanhristovski: > should the ratelimit be same if the policy is same as shared for the same policy ?

yeah youre right

> also can we also add the same IRs from the test in /in/xds-ir to view the xDS descriptors etc

makes sense, added

test/e2e/testdata/ratelimit-global-shared-cidr-match.yaml
```diff
@@ -0,0 +1,59 @@
...
+  name: cidr-ratelimit-1
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+    - name: same-namespace
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /
```
arkodg, line 39: the routes look identical, can you update the path prefix for one to be `/foo` and other to be `/bar` and send requests to specific routes but check if the count/limit is shared ?

test/e2e/testdata/ratelimit-global-shared-gateway-header-match.yaml
```diff
@@ -0,0 +1,69 @@
...
+kind: HTTPRoute
+metadata:
+  name: header-ratelimit-2
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+    - name: eg-rate-limit
+  rules:
+    - matches:
+        - path:
```
arkodg, line 40: same as https://github.com/envoyproxy/gateway/pull/5361/files#r2004628119

internal/ir/xds.go
```diff
@@ -2956,3 +2967,11 @@ type ResourceMetadata struct {
 	// SectionName is the name of a section of a resource
 	SectionName string `json:"sectionName,omitempty" yaml:"sectionName,omitempty"`
 }
+
+// BackendTrafficPolicy represents the policy associated with backend traffic.
+type BackendTrafficPolicy struct {
```
arkodg, line 2972: can we rm this ?

api/v1alpha1/ratelimit_types.go
```diff
@@ -54,7 +54,6 @@ type GlobalRateLimit struct {
 
 	// Shared determines whether the rate limit rules apply across all the policy targets.
 	// If set to true, the rule is treated as a common bucket and is shared across all policy targets (xRoutes).
-	// Must have targetRef set to Gateway
```
zhaohuabing, line 57: I guess the use case is that the bucket is shared among multiple HTTPRoutes targets (but not all of them under one Gateway)?

internal/gatewayapi/testdata/backendtrafficpolicy-with-shared-ratelimit.in.yaml
```diff
@@ -0,0 +1,121 @@
...
+- apiVersion: gateway.envoyproxy.io/v1alpha1
+  kind: BackendTrafficPolicy
+  metadata:
+    namespace: default
+    name: policy-for-route
+  spec:
+    targetRef:
+      group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      name: httproute-1
```
zhaohuabing, line 102: Can we set multiple HTTPRoutes here? Targeting only one HTTPRoute for a shared Global ratelimit seem meaningless.

zhaohuabing: Also, can we add a test for this use case:
* All routes of a Gateway share a bucket, e.g. 100 R/S
* Define a more specific bucket for an HTTPRoute under that gateway, e.g. 10 R/S

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -696,15 +700,15 @@ func (t *Translator) buildGlobalRateLimit(policy *egv1a1.BackendTrafficPolicy) (
...
 	if !t.GlobalRateLimitEnabled {
 		return nil, fmt.Errorf("enable Ratelimit in the EnvoyGateway config to configure global rateLimit")
 	}
 
 	global := policy.Spec.RateLimit.Global
 	rateLimit := &ir.RateLimit{
 		Global: &ir.GlobalRateLimit{
-			Rules: make([]*ir.RateLimitRule, len(global.Rules)),
+			Rules:  make([]*ir.RateLimitRule, len(global.Rules)),
+			Shared: global.Shared,
```
zhaohuabing, line 711: I think we need  two ratLimits in the ir to allow two level of global ratelimiting:
* First level - a shared bucket applied across multiple routes
* Second level - a more specific bucket for a particular route

zhaohuabing: Of course, if this use case falls outside the scope of this PR, we can handle it in a follow-up PR.

api/v1alpha1/ratelimit_types.go
zhaohuabing: Nit: Can we add a validation that if the targeted type is xRoute, then the targets should at least be two?

A shared global ratelimit targeting at one xRoute is not useful.

internal/gatewayapi/testdata/backendtrafficpolicy-dns-lookup-family.out.yaml
```diff
@@ -361,6 +361,7 @@ xdsIR:
           dns:
             dnsRefreshRate: 30s
             lookupFamily: IPv4AndIPv6
+          name: ""
```
zhaohuabing, line 364: Why this changed?

internal/ir/xds.go
```diff
@@ -1997,6 +1999,15 @@ type GlobalRateLimit struct {
 
 	// Rules for rate limiting.
 	Rules []*RateLimitRule `json:"rules,omitempty" yaml:"rules,omitempty"`
+
+	// Shared determines whether this rate limit rule applies globally across the gateway.
```
zhaohuabing, line 2003: Nit: the rule can applies globally across the gateway, or multiple httproutes.

internal/xds/translator/testdata/out/xds-ir/ratelimit-global-shared.listeners.yaml
```diff
@@ -0,0 +1,64 @@
...
+      typedConfig:
+        '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+        commonHttpProtocolOptions:
+          headersWithUnderscoresAction: REJECT_REQUEST
+        http2ProtocolOptions:
+          initialConnectionWindowSize: 1048576
+          initialStreamWindowSize: 65536
+          maxConcurrentStreams: 100
+        httpFilters:
+        - name: envoy.filters.http.ratelimit
```
zhaohuabing, line 17: Can we add a suffix to ensure the filter name remain unique across the HTTP filter chain?

Similar to what we have done to wsam filter: envoy.filters.http.ratelimit/test-namespace/test-policy-1

internal/xds/translator/testdata/out/xds-ir/ratelimit-global-shared.routes.yaml
```diff
@@ -0,0 +1,99 @@
...
+        upgradeConfigs:
+        - upgradeType: websocket
+      typedPerFilterConfig:
+        envoy.filters.http.ratelimit:
+          '@type': type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimitPerRoute
+          rateLimits:
+          - actions:
+            - genericKey:
+                descriptorKey: test-policy-2
+                descriptorValue: test-namespace
```
zhaohuabing, line 68: Nit: non-blocking
This key and value should work, but can we keep the same matching logic of the genericKey for route and use the same string for key and value? This would make the ratlimit.go a bit easier to understand.
descriptorKey: test-namespace-test-policy-2
descriptorValue: test-namespace-test-policy-2

internal/xds/translator/ratelimit.go
```diff
@@ -163,26 +211,51 @@ func patchRouteWithRateLimitOnTypedFilterConfig(route *routev3.Route, rateLimits
...
+	// Determine if the rate limit is shared across multiple routes.
+	global := route.Traffic.RateLimit.Global
+	isShared := isSharedRateLimit(route)
+
+	// Set the descriptor key based on whether the rate limit is shared.
+	var descriptorKey, descriptorValue string
+	if isShared {
+		// For shared rate limits, use BTP name as key and namespace as value
+		descriptorKey = getTrafficPolicyName(route)
+		descriptorValue = getTrafficPolicyNamespace(route)
```
zhaohuabing, line 231: https://github.com/envoyproxy/gateway/pull/5361/files#r2004856877

internal/xds/translator/ratelimit.go
```diff
@@ -329,70 +406,187 @@ func GetRateLimitServiceConfigStr(pbCfg *rlsconfv3.RateLimitConfig) (string, err
...
+
+// addSharedRateLimitDescriptor adds shared rate limit descriptors to the domain descriptor map
+func addSharedRateLimitDescriptor(
+	route *ir.HTTPRoute,
+	serviceDescriptors []*rlsconfv3.RateLimitDescriptor,
+	domain string,
+	domainDescriptors map[string][]*rlsconfv3.RateLimitDescriptor,
+) {
+	sharedKey := getTrafficPolicyName(route)
+	sharedValue := getTrafficPolicyNamespace(route)
```
zhaohuabing, line 457: https://github.com/envoyproxy/gateway/pull/5361/files#r2004856877

internal/gatewayapi/testdata/backendtrafficpolicy-buffer-limit-out-of-range-error.out.yaml
```diff
@@ -282,6 +282,7 @@ xdsIR:
           namespace: default
         name: grpcroute/default/grpcroute-1/rule/0/match/-1/*
         traffic:
+          name: policy-for-gateway/envoy-gateway
```
zhaohuabing, line 285: Nit: prefer keeping the same format as other xDs naming:  namespace/policy

internal/gatewayapi/helpers.go
```diff
@@ -449,6 +449,11 @@ func irTLSCACertName(namespace, name string) string {
 	return fmt.Sprintf("%s/%s/%s", namespace, name, caCertKey)
 }
 
+// Helper function to format the policy name and namespace
+func irTrafficName(policy *egv1a1.BackendTrafficPolicy) string {
+	return fmt.Sprintf("%s/%s", policy.Name, policy.Namespace)
```
zhaohuabing, line 454: https://github.com/envoyproxy/gateway/pull/5361/files#r2004880933

internal/xds/translator/ratelimit.go
```diff
@@ -329,70 +406,187 @@ func GetRateLimitServiceConfigStr(pbCfg *rlsconfv3.RateLimitConfig) (string, err
...
+
+	// Check if we already have a descriptor for this key/value pair
+	for i, desc := range domainDescriptors[domain] {
+		if desc.Key == sharedKey && desc.Value == sharedValue {
+			// Found existing shared descriptor, merge new descriptors into it
+			domainDescriptors[domain][i].Descriptors = mergeUniqueDescriptors(
```
zhaohuabing, line 463: I think we don't need to merge here? 
Even though the shared global ratelimit has been added to every individual route, but they're the same. So merging is not necessary here, and it would confuse people who read this code.
```
for _, desc := range domainDescriptors[domain] {
		if desc.Key == sharedKey && desc.Value == sharedValue {
		        // We just skip repeated shared global ratelimit because the configuration on each route would be the same.
			return
		}
	}
```

internal/gatewayapi/testdata/backendtrafficpolicy-with-shared-ratelimit.in.yaml
zhaohuabing: we can handle it in a follow-up PR if it's not in the scope.

api/v1alpha1/ratelimit_types.go
ryanhristovski: It's not necessarily useful, but I do think it's valid - personally I don't think it should fail on a single xRoute as I think this could potentially lead to some confusion for the user. What do you think @arkodg

arkodg: Prefer if we didn't add that validation, because users may want to add or remove items from that list ( for prod/debugging use cases ) and we shouldn't reject valid config

internal/gatewayapi/testdata/backendtrafficpolicy-with-shared-ratelimit.in.yaml
arkodg: Yeah this is something we can achieve with mergeSemantics

internal/xds/translator/testdata/out/xds-ir/ratelimit-global-shared.routes.yaml
ryanhristovski: Right now with the domain also being: test-namespace-test-policy-2, your suggestion would make the keys look something like:
`test-namespace-test-policy-2-test-namespace-test-policy-2-test-namespace-test-policy-2`... which felt excessive while testing, so i reduced it by separating key and value. I found it to be slightly confusing personally

internal/gatewayapi/backendtrafficpolicy.go
ryanhristovski: Yeah this falls outside scope, I believe this will be covered under policy mergeSemantics

internal/xds/translator/ratelimit.go
ryanhristovski: You're right, this is a great find - thanks!

internal/xds/translator/testdata/out/xds-ir/ratelimit-global-shared.listeners.yaml
ryanhristovski: Updated this to set the suffix to be the equal to the route.traffic.name if the ratelimit is shared, if the ratelimit is not shared, I kept it = to the default `envoy.filters.http.ratelimit`. Adding the listener.name to the end would require a major refactor and isn't worth it, especially since there can only be 1 generic `envoy.filters.http.ratelimit` per listener.

zhaohuabing (approved):
LGTM. Thanks for adding this awesome feature!

internal/xds/translator/ratelimit.go
```diff
@@ -329,70 +410,153 @@ func GetRateLimitServiceConfigStr(pbCfg *rlsconfv3.RateLimitConfig) (string, err
...
+	configs := createRateLimitConfigs(domainDescriptors)
+	return configs
+}
+
+// addSharedRateLimitDescriptor adds shared rate limit descriptors to the domain descriptor map
+func addSharedRateLimitDescriptor(
+	route *ir.HTTPRoute,
+	serviceDescriptors []*rlsconfv3.RateLimitDescriptor,
+	domain string,
+	domainDescriptors map[string][]*rlsconfv3.RateLimitDescriptor,
```
arkodg, line 458: is this domain created to avoid creating duplicates when parsing multiple Route IRs that contain the same shared bucket ? would it be enough for the map to hold the domain name without the actual descriptors, and skip parsing if that domain was already created ?

internal/xds/translator/ratelimit.go
```diff
@@ -159,30 +211,55 @@ func patchRouteWithRateLimitOnTypedFilterConfig(route *routev3.Route, rateLimits
...
+
+	// Determine if the rate limit is shared across multiple routes.
+	global := route.Traffic.RateLimit.Global
+	isShared := isSharedRateLimit(route)
+
+	// Set the descriptor key based on whether the rate limit is shared.
+	var descriptorKey, descriptorValue string
+	if isShared {
+		// For shared rate limits, use BTP name as key and namespace as value
+		descriptorKey = getTrafficPolicyName(route)
```
arkodg, line 234: the current code sets the key and value to be the same value
```
				DescriptorKey:   getRouteDescriptor(descriptorPrefix),
				DescriptorValue: getRouteDescriptor(descriptorPrefix),
```

so it may be better to do the same for the shared case and use `traffic.Name` directly, we can optimize and improve debuggability and metric labels later

zhaohuabing: +1 this reduce confusion.

internal/xds/translator/testdata/out/xds-ir/ratelimit-global-shared.listeners.yaml
ryanhristovski: The route.traffic.name value is `namespace-name/policy-name` so this already works

ryanhristovski:
@arkodg @zhaohuabing I just want to highlight that these changes require proxy 1.33.0+ for release notes

PR #5363, opened 2025-02-27
fix: metric compressor not working as expected

zirain:
I found the configuration was broken for a long while, this patch fixed it, also add e2e test for it.

### DIFF
--- a/internal/xds/bootstrap/bootstrap.yaml.tpl
+++ b/internal/xds/bootstrap/bootstrap.yaml.tpl
@@ -86,6 +86,12 @@ static_resources:
             - name: prometheus_stats
               domains:
               - "*"
+              {{- if .EnablePrometheusCompression }}
+              typed_per_filter_config:
+                envoy.filters.http.compression:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.compressor.v3.CompressorPerRoute
+                  disabled: true
+              {{- end }}
               routes:
               - match:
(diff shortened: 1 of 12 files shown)


internal/xds/bootstrap/bootstrap.yaml.tpl
```diff
@@ -95,30 +95,24 @@ static_resources:
...
-                    {{- end }}
-                    {{- if eq .PrometheusCompressionLibrary "zstd"}}
-                    compressor_library:
-                      name: text_optimized
-                      typed_config:
-                        "@type": type.googleapis.com/envoy.extensions.compression.zstd.compressor.v3.Zstd
-                    {{- end }}
-                {{- end }}
           http_filters:
+          {{- if .EnablePrometheusCompression }}
```
arkodg, line 99: shouldnt this be scoped only to the prom route ?

zirain: we didn't share the listener now

zirain: is https://github.com/envoyproxy/gateway/pull/5363/commits/5e18fb6938985e7450bc18ae7c3f5f6bbcdb6e91 what you want?

arkodg: ah its a unique listener so this is safe, thanks

arkodg (approved):

PR #5391, opened 2025-03-03
chore: fix dependabot

zirain:
xref: https://github.com/envoyproxy/gateway/runs/37897802670


.github/dependabot.yml
```diff
@@ -22,13 +22,16 @@ updates:
...
+        patterns:
+          - "actions*"
   - package-ecosystem: gomod
     directories:
       - "/"
-      - "examples/extension-server"
     schedule:
       interval: weekly
+    ignore:
+      # skip to update cel-go, because it related to k8s.io/apiserver
```
arkodg, line 33: can we link any GH issue / link here, as a hint to figure out when to rm this section ?

zirain: we should let tie cel-go to k8s.io/apiserver.

arkodg (approved):

zhaohuabing (approved):

Issue #5307, opened 2025-02-18
Revisit Kubernetes Services mapping to Envoy v3.Cluster


dprotaso:
*Description*:
I was surprised to see that Kubernetes services weren't mapped to envoy v3.Clusters.  One thing I was hoping for was to create high level Gateway Policy resources that would then reconcile EnvoyPatchPolicy. I prefer not to build out an entire extension service.

I wanted my policy to target k8s services and then creating a patch could modify the v3.Cluster

Unfortunately,  it seems like Gateway HTTPRoute backends maps to envoy v3.Clusters (which has it's own side effects https://github.com/envoyproxy/gateway/issues/5230).

guydc:
In general, I agree that in some cases it could be beneficial to map Backends/Services to a single cluster. For example, users may want to apply limits such as circuit breakers on a per-proxy-per-backend basis, and not have these limits duplicated per-route-backendref. Other issues include excessive active health checking from duplication of clusters and their HC settings, inconsistent passive health check status.

arkodg:
agree with @guydc's points around setting per backend values. The reason to keep fields such as `circuitBreaker` and `healthCheck` in the BackendTrafficPolicy instead of a dedicated policy like BackendPolicy was to prioritize usability and reduce the number of place to author intent. We should revisit this design decision based off more user feedback

There are limitations with weighted clusters, adding the ones I can think of here
* Traffic Splitting + Session Affinity https://github.com/envoyproxy/envoy/issues/8167
* Traffic Splitting + Session Persistence https://github.com/envoyproxy/envoy/issues/24741
* Traffic Splitting + Consistent Hashing Loadbalancing (Maglev) https://github.com/envoyproxy/envoy/issues/21675
* Retries dont consider other hosts in other clusters https://github.com/envoyproxy/envoy/issues/18837#issuecomment-956946870

@mathetake @cpakulski @nezdolik are they any other limitations of weighted clusters you are aware of ?

muwaqar:
+1

Gateway API spec recently seems to be introducing policies on a per-`BackendReference` (`Service`, `ServiceImport`, `Backend`) basis as well.

For example:
- `gateway.networking.k8s.io/v1alpha3`: `BackendTLSPolicy` ([reference](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1alpha3.BackendTLSPolicy))
[...]

PR #5394, opened 2025-03-03
feat: weighted clusters and cluster per-backendref

jukie:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/default-resources.all.yaml
@@ -724,7 +724,7 @@ xds:
       dynamicEndpointConfigs:
       - endpointConfig:
           '@type': type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment
-          clusterName: httproute/default/backend/rule/0
+          clusterName: httproute/default/backend/rule/0/backend/0
           endpoints:
           - lbEndpoints:
             - endpoint:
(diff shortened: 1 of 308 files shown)


zhaohuabing:
cc @arkodg @guydc who may have more background on the pros and cons of using a per-HTTPRouteRule cluster versus a per-BackendRef cluster.

arkodg:
there are some drawbacks of weighted clusters outlined in https://github.com/envoyproxy/gateway/issues/5307#issuecomment-2688767482
I'll try and take a look at the zone aware PR today to better understand the problem

jukie:
The Locality LoadBalancingWeight field is only respected by LoacalityWeightedLBConfig and is ignored by other LB config types (such as zone aware routing) so the only way to implement would be to calculate per-endpoint weights and apply those. However, those can't be 100% accurate and will fail conformance test tolerance levels of 5% more than half the time (accurate to ~10%). We could possibly scale the backend weights to a common divisor across all backends but there's edge cases where that won't work and also makes the rendered config hard to reason about. Would love to discuss a workaround if anyone has ideas.

zhaohuabing:
> @arkodg and I spoke to discuss the limitations with weighted clusters (mostly retries and sticky sessions)

It could be extremly difficult to debug if users report such issues.  Probably EG should surface these limitations back to the status of the BackendTrafficPolicy when weighted clusters are used.

PR #5409, opened 2025-03-05
deps: upgrades some direct dependencies

mathetake:


mathetake:
@arkodg i tried to use the EG latest in AI Gateway and found that the cel-go library is so old that there is a incompatibility between the one used here vs the one used in AI gateawy. So I upgraded it along with some dependencies. hope this won't be any trouble

mathetake:
"so old" was an exaggeration sorry ;) but anyways

mathetake:
oh wait

mathetake:
https://github.com/kubernetes/apiserver/issues/116 yeah this ... looks like k8s/apiserver looks incompatible with the latest cel-go

mathetake:
ok for now i will downgrade the cel-go used in ai-gateway 🤷

PR #5410, opened 2025-03-05
release: v1.2.7

zhaohuabing:


api/v1alpha1/shared_types.go
```diff
@@ -30,7 +30,7 @@ const (
 	// DefaultShutdownManagerImage is the default image used for the shutdown manager.
 	DefaultShutdownManagerImage = "docker.io/envoyproxy/gateway-dev:latest"
 	// DefaultRateLimitImage is the default image used by ratelimit.
-	DefaultRateLimitImage = "docker.io/envoyproxy/ratelimit:49af5cca"
+	DefaultRateLimitImage = "docker.io/envoyproxy/ratelimit:ae4cee11"
```
arkodg, line 33: thanks, this needs to be updated in the helm chart too
https://github.com/envoyproxy/gateway/blob/1ef27f7c2764a18c4bf0f7330840c30397010de9/charts/gateway-helm/values.tmpl.yaml#L15

arkodg (approved):

PR #5416, opened 2025-03-05
standalone: propagate signal to Envoy procs

mathetake:
**What type of PR is this?**

Fixes standalone mode to handle signals correctly.

Previously, the signals to envoy-proxy CLI hadn't been propagated to Envoy processes. Therefore, they resulted in orphans.

Release Notes:No


arkodg:
thanks @mathetake, has this been manually tested - does killing the envoy-gateway process terminate envoy-proxy as well ?

mathetake:
it's not working ... i guess there must be a bug in func-e as well but i think the change here itself should be merged anyways

arkodg (approved):

mathetake:
i got the root cause

PR #5417, opened 2025-03-05
Propagate contexts from root to subcommands

mathetake:
**What type of PR is this?**

Refactors cmd packages

Previously, the context was locally created inside the subcommand, and the context set to the cobra.Cmd was ignored. This made it difficult to use the root command as a library since the only way to cancel the command is sending the real OS singals hence it was impossible to write a unit tests around it.

Release Notes: No

### DIFF
--- a/cmd/envoy-gateway/main.go
+++ b/cmd/envoy-gateway/main.go
@@ -9,11 +9,13 @@ import (
 	"fmt"
 	"os"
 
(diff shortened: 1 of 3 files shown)


arkodg (approved):

mathetake:
come on

mathetake:
good to go?

zirain (approved):

PR #5423, opened 2025-03-06
chore: fix examples gen-check

zirain:


zirain:
looks like `gen-check` is broken.

arkodg (approved):

shawnh2 (approved):

PR #5441, opened 2025-03-08
feat: support bypassing envoy overload manager for listeners

### LINKED ISSUE #5260: Support exclusion of listeners from overload manager

rudrakhp:
**What type of PR is this?**

feat: support bypassing envoy overload manager for listeners
Also has a minor refactor of Listener IR struct that resolves a TODO.

To enable user to bypass overload manager for listeners. See #5260 for more details.

Fixes #5260 

Release Notes: Yes/No


internal/ir/xds.go
```diff
@@ -1786,20 +1778,28 @@ func (s StringMatch) Validate() error {
...
 }
 
-// TCPListener holds the TCP listener configuration.
+// TCPBasedListenerDetails holds listener configuration for listeners based on the TCP protocol.
+// Eg: TCPListener and HTTPListener
+//
 // +k8s:deepcopy-gen=true
-type TCPListener struct {
+type TCPBasedListenerDetails struct {
 	CoreListenerDetails `json:",inline" yaml:",inline"`
```
rudrakhp, line 1786: Was wondering if we should have `CoreListenerDetails` nested under `TCPBasedListenerDetails`, hierarchically it's correct but might be confusing.

arkodg:
can you explain this use case @rudrakhp ? im not sure why its required ? if a proxy is overloaded, we do want the health checks enabled on the listener to fail so the fronting LB can send new requests to another proxy w/o affecting existing requests/connections

rudrakhp:
From the conversation in the original issue, I was under the impression we do support custom health check probes, that might require users to set this option in custom listeners for these probes as well. Do correct me if this is not the case.

rudrakhp:
Closing this, we can come back to this after discussing it's requirement in the issue.

Issue #5414, opened 2025-03-05
Validate ID Token when OIDC is set


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

The OIDC spec calls out verifying the ID Token https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

This be be enabled by default or opt in

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

markwinter:
Hey, I'm looking to start working on Envoy Gateway and wouldn't mind picking this one up

/assign

arkodg:
thanks for picking this one up @markwinter, recommend starting off with an API PR

zhaohuabing:
@arkodg @markwinter I believe this must be supported by Envoy first, then we can enforce the validation in EG.

arkodg:
@zhaohuabing can't we add jwt authn if this field is set ?

markwinter:
I've found some related discussion
https://github.com/envoyproxy/gateway/discussions/2425#discussioncomment-8722381
https://github.com/envoyproxy/envoy/issues/32805

I also took a look at envoyproxy to check that enabling forward_bearer_token puts the access token in the `Authorization` header
https://github.com/envoyproxy/envoy/blob/eb398e9e4e0e33c8d68d5b9f86db88f236fd57c4/source/extensions/filters/http/oauth2/filter.cc#L883

And I checked that the Oauth2 filter doesn't do any token validation, only the JWT filter.

So if we wanted to validate the id token, we could maybe use both Oauth2 Filter, JWT Filter, and also use jwtExtractor/from_cookies to get the idToken from the cookie?

Alternatively validation should be added upstream in the oidc filter

markwinter:
I've found some related discussion
https://github.com/envoyproxy/gateway/discussions/2425#discussioncomment-8722381
https://github.com/envoyproxy/envoy/issues/32805

I also took a look at envoyproxy to check that enabling forward_bearer_token puts the access token in the `Authorization` header
https://github.com/envoyproxy/envoy/blob/eb398e9e4e0e33c8d68d5b9f86db88f236fd57c4/source/extensions/filters/http/oauth2/filter.cc#L883

And I checked that the Oauth2 filter doesn't do any token validation, only the JWT filter.

So if we wanted to validate the id token, we could maybe use both Oauth2 Filter, JWT Filter, and also use jwtExtractor/from_cookies to get the idToken from the cookie?

[...]

PR #5443, opened 2025-03-08
api(oidc): Add validation fields to OIDCProvider

markwinter:

This PR is the start of adding ID token validation when using OIDC authentication.

It adds the `RemoteJWKS` and `Audience` fields to the OIDCProvider.

The future intention is that when `RemoteJWKS` is set, the `Issuer` `Audience` and `RemoteJWKS` fields will be used to configure a JWT Filter that will validate the ID Token (using IdToken cookie)

Reference #5414

Release Notes: No

### DIFF
--- a/api/v1alpha1/oidc_types.go
+++ b/api/v1alpha1/oidc_types.go
@@ -144,7 +144,7 @@ type OIDCProvider struct {
 	//
 	// If set, the ID token will be validated using the RemoteJWKS, Issuer, and Audiences fields
 	// +optional
-	RemoteJWKS RemoteJWKS `json:"remoteJWKS"`
+	RemoteJWKS *RemoteJWKS `json:"remoteJWKS,omitempty"`
 
 	// The OIDC Provider's [authorization endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
 	// If not provided, EG will try to discover it from the provider's [Well-Known Configuration Endpoint](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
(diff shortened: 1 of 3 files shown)


zirain:
Can we add CEL validation rule for these rules?

api/v1alpha1/oidc_types.go
```diff
@@ -125,10 +125,26 @@ type OIDCProvider struct {
...
+	// +kubebuilder:validation:MaxItems=8
+	// +optional
+	Audiences []string `json:"audiences,omitempty"`
+
+	// RemoteJWKS defines how to fetch and cache JSON Web Key Sets (JWKS) from a remote
+	// HTTP/HTTPS endpoint.
+	//
+	// If set, the ID token will be validated using the RemoteJWKS, Issuer, and Audiences fields
+	// +optional
+	RemoteJWKS *RemoteJWKS `json:"remoteJWKS,omitempty"`
```
zhaohuabing, line 147: I think the current `SecurityPolicy` already covers these in the JWT section. Dplicating these fields in two places within `SecurityPolicy` is confusing.

zhaohuabing: I'm -1 on this. More at https://github.com/envoyproxy/gateway/issues/5414#issuecomment-2709488131

arkodg: yeah, `enableIDTokenValidation/validateIDToken` seems like the right approach, does that sgty @zhaohuabing  ?

zhaohuabing: `enableIDTokenValidation/validateIDToken` LGTM from the API perspective.

Do we currently plan to do only partial validation on the issu and aud?

PR #5444, opened 2025-03-08
fix: loop over all endpointslices

jukie:
**What type of PR is this?**

fix: loop over all endpointslices

When listing EndpointSlices via client.List, the default limit is 100 items so in order to support more the `endpointSliceList.Continue` token should be checked and if non-empty continue looping.

Fixes #

Release Notes: No

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -462,19 +462,27 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
 				}),
 				client.InNamespace(*backendRef.Namespace),
 			}
-			if err := r.client.List(ctx, endpointSliceList, opts...); err != nil {
-				r.log.Error(err, "failed to get EndpointSlices", "namespace", string(*backendRef.Namespace),
-					backendRefKind, string(backendRef.Name))
-			} else {
-				for _, endpointSlice := range endpointSliceList.Items {
-					key := utils.NamespacedName(&endpointSlice).String()
-					if !resourceMappings.allAssociatedEndpointSlices.Has(key) {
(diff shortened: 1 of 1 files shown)


jukie:
I'll add a test

internal/provider/kubernetes/controller.go
```diff
@@ -462,20 +462,28 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
...
+						if !resourceMappings.allAssociatedEndpointSlices.Has(key) {
+							resourceMappings.allAssociatedEndpointSlices.Insert(key)
+							r.log.Info("added EndpointSlice to resource tree",
+								"namespace", endpointSlice.Namespace,
+								"name", endpointSlice.Name)
+							gwcResource.EndpointSlices = append(gwcResource.EndpointSlices, &endpointSlice)
+						}
 					}
 				}
+				if endpointSliceList.Continue != "" {
```
arkodg, line 482: can we simplify this logic to ?
```
continueToken = endpointSliceList.Continue
// return if all items have been listed
if continueToken == "" {
  break
}
```

arkodg:
hey @jukie can you share the reference where pagination is enabled by default with a limit of 100 for the EndpointSlice API ?
trying to better understand the defaults and whether it applies to all APIs or just EndpointSlice

jukie:
I can't find the docs on it now and it's unclear but in the past when I hit this in an unrelated project IIRC it was dependent on API (but in that case was EndpointSlices) as well as API server configuration.

I ran a manual test listing pods and endpointslices with a client using `sigs.k8s.io/controller-runtime/pkg/client` (same as EnvoyGateway) across a few of my clusters running in various providers (both cloud and on-prem) and I'm unable to reproduce this and it has no trouble listing more than 100 items. I'll close this, sorry for the noise.

PR #5445, opened 2025-03-09
docs: don't run `copy-current-release-docs` as part of `gen-check`

zirain:
key changes:
1. remove current `docs`
2. `docs` will be generated when run `make docs/docs-serve`
3. change `site/layouts/partials/page-meta-links.html` to replace the url for `edit this page` button


site/content/en/docs/_index.md
```diff
@@ -1,15 +0,0 @@
-+++
```
arkodg, line 1: why are these removed ?

zirain: we didn't nee these, make docs will copy them when release doc site.

arkodg: will the Hugo "Edit this Page" feature still work ?

zirain: I think it won't.

Xunzhuo: I recall that ths is added by @zhaohuabing, for accessing default path as the latest release version. Need a review from @zhaohuabing

zhaohuabing: If we could point the "Edit this Page" links in the docs to the latest released version (e.g. v1.3), then it would work.

arkodg: is that possible using hugo config @zhaohuabing @zirain  ? if so then this change is a great one

zirain: I'm not sure about it, I'm not a fan of that, will take a look.

zirain: seems https://www.docsy.dev/docs/adding-content/repository-links/#path_base_for_github_subdir-optional may help.

zhaohuabing (approved):
Great job!

Xunzhuo (approved):
Nice improves, LGTM

PR #5455, opened 2025-03-10
api(oidc): adds support in OIDC api to optionally deny redirects

sam-burrell:
Adds support in the OIDC api to optionally deny redirects that match a header/:path/:method.

The intention is to then implement this with the underlying envoy [oauth2 configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2) `deny_redirect_matcher` 
This is useful for ajax/machine paths that should not be directed to the authentication flow.

I have the changes ready for a second pull request to implement this if accepted.

Could be used as fix for? [2496](https://github.com/envoyproxy/gateway/issues/2496)

Release Notes: No


api/v1alpha1/oidc_types.go
```diff
@@ -143,6 +150,19 @@ type OIDCProvider struct {
...
+// +notImplementedHide
+type OIDCDenyRedirectMatcher struct {
+	// Specifies the name of the header in the request.
+	// The pseudo-headers ``:path`` and ``:method`` can be used to match the request path and method, respectively
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// If specified, header match will be performed based on the string match of the header value.
+	// Specifies how the header match will be performed to route the request.
+	// +optional
+	StringMatch StringMatch `json:"stringMatch,omitempty"`
```
zhaohuabing, line 163: It also worth to mention in the comment that the matcher will only check the presence of the header if the `stringMatcher` is not specified.
```suggestion
	StringMatch *StringMatch `json:"stringMatch,omitempty"`
```

api/v1alpha1/oidc_types.go
```diff
@@ -62,6 +62,13 @@ type OIDC struct {
 	// If not specified, uses the default redirect URI "%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback"
 	RedirectURL *string `json:"redirectURL,omitempty"`
 
+	// Any request that matches any of the provided matchers won’t be redirected to OAuth server when tokens are not valid.
+	// Automatic access token refresh will be performed for these requests, if enabled.
```
zhaohuabing, line 66: This comment is somewhat ambiguous.  The request can be denied with a 401 response in two conditions:
* User has logged in but tokens are expired.
* User has not logged in, and no tokens in the cookie.

zhaohuabing:
@sam-burrell Thanks for adding this useful feature. Overall, the API looks good. I just have a few nits.

api/v1alpha1/oidc_types.go
sam-burrell: I think its not optional. If a denyRedirectMatcher is specified it will ned to have a StringMatch of some type.
The whole denyRedirectMatcher is optional.

I will remove +optional from StringMatch

api/v1alpha1/oidc_types.go
sam-burrell: Agreed. Have changed this comment to 

```Any request that matches any of the provided matchers (with either tokens that are expired or missing tokens) will not be redirected to the OIDC Provider.```

**Comment by sam-burrell** [comment]
Thanks @zhaohuabing for the feedback. I've made some (hopeful) improvements!

**Comment by sam-burrell** [comment]
Do I wait for this to be merged before I do a pull request for the actual underlying code change?

**Comment by arkodg** [comment]
@sam-burrell will this make https://github.com/envoyproxy/gateway/pull/5142 redundant for the machine case ?

**Comment by zhaohuabing** [comment]
> @sam-burrell will this make #5142 redundant for the machine case ?

@arkodg I think they're used for different purposes:
* `deny_redirect_matcher` is used to not redirect the request to the IDP login page when the request is from non-human
* https://github.com/envoyproxy/gateway/pull/5142 is used to bypass OIDC authn when there is a token in the request

They may be used together for non-human requests.

**Comment by sam-burrell** [comment]
I agree with @zhaohuabing this is a slightly different usecase. This is mainly for non human requests

**Comment by arkodg** [comment]
hey @sam-burrell had a chat with @zhaohuabing offline, who mentioned that this could be solved with https://github.com/envoyproxy/gateway/pull/5142 + a jwt section in the SecurityPolicy authenticating the token.
Would that work for your use case ?

**Comment by sam-burrell** [comment]
Thanks for your thoughts all.

Thinking about this https://github.com/envoyproxy/gateway/pull/5142 could work if we are able to make different httproutes for the **same hostname**. 

We would like to deny redirects by path.
The use case I am considering is the following

https://app.foo/api -> we want 401 (no redirect)
https://app.foo/* -> we want redirects to normal login flow

Would https://github.com/envoyproxy/gateway/pull/5142 work for this?

**Comment by zhaohuabing** [comment]
> https://app.foo/api -> we want 401 (no redirect)
https://app.foo/* -> we want redirects to normal login flow

I think this PR can't be solved by https://github.com/envoyproxy/gateway/pull/5142 . `PassThroughAuthHeader` only skips OIDC if there's a token in the header. 
If a request to https://app.foo/api without OIDC header, it won't passthrough and will be redirected. @arkodg

**Comment by zhaohuabing** [comment]
Think again, this PR can't be solved by https://github.com/envoyproxy/gateway/pull/5142 . `PassThroughAuthHeader` only skips OIDC if there's a token in the header. 
If a request to https://app.foo/api without a token, it won't passthrough and will be redirected. @arkodg

**Comment by zhaohuabing** [comment]
Thought again, #5142 may not be able to slove this use case.

Let's say all AJAX requests contains a header `is-ajax-request`, and usr requests don't have this header,and we want to avoid redirection for AJAX requests.

How could we achieve this with `passThroughAuthHeader` ?

This configuration only allows requests with Authorization header passthrough without OIDC, it won't work for AJAX requests without tokens.

```yaml
    oidc:
      provider:
        issuer: "https://oauth.foo.com"
        authorizationEndpoint: "https://oauth.foo.com/oauth2/v2/auth"
        tokenEndpoint: "https://oauth.foo.com/token"
      clientID: "client.oauth.foo.com"
      clientSecret:
        name: "client-secret-1"
      redirectURL: "https://www.example.com/oauth2/callback"
      passThroughAuthHeader: true
    jwt:
      providers:
      - name: example1
        issuer: https://oauth.foo.com
        remoteJWKS:
          uri: https://oauth.foo.com/jwt/public-key/jwks.json
```

This configuration can skip AJAX request for OIDC, but obvious it's a hack on the SecurityPolicy, which is difficult to understand and maintain. This will also block normal user requests since there's no token in the `is-ajax-request` header.

```yaml
   oidc:
      provider:
        issuer: "https://oauth.foo.com"
        authorizationEndpoint: "https://oauth.foo.com/oauth2/v2/auth"
        tokenEndpoint: "https://oauth.foo.com/token"
      clientID: "client.oauth.foo.com"
      clientSecret:
        name: "client-secret-1"
      redirectURL: "https://www.example.com/oauth2/callback"
      passThroughAuthHeader: true
    jwt:
      providers:
      - name: example1
        issuer: https://oauth.foo.com
        remoteJWKS:
          uri: https://oauth.foo.com/jwt/public-key/jwks.json
        extractFrom:
          headers:
          - name: is-ajax-request
          - name: Authorization

**Comment by sam-burrell** [comment]
@zhaohuabing yup agreed, I guess this is the reason DenyRedirectMatcher is there in the underlying envoy [oauth2 configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2)

File: api/v1alpha1/oidc_types.go [hunk 3]
```diff
@@ -62,6 +62,12 @@ type OIDC struct {
 	// If not specified, uses the default redirect URI "%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback"
 	RedirectURL *string `json:"redirectURL,omitempty"`
 
+	// Any request that matches any of the provided matchers (with either tokens that are expired or missing tokens) will not be redirected to the OIDC Provider.
+	// This behavior can be useful for AJAX or machine requests.
+	// +optional
+	// +notImplementedHide
+	DenyRedirectMatcher []OIDCDenyRedirectMatcher `json:"denyRedirectMatcher,omitempty"`
```
arkodg, line 69: can we make this look more Gateway API like ?
one suggestion
```
denyRedirect:
  headers:
  - name: foo
     value: bar
```

zhaohuabing:
ping @sam-burrell

sam-burrell:
very sorry for the delayed response, i've been away for a bit, I agree @arkodg that makes sense.

I think the team has been reviewing this and it seems like this is something we need support
I will be picking this back up over the next few days

sam-burrell:
- we need the feature of automatic access token refresh without redirects, this is what this underlying feature of envoy enables us to do

sam-burrell:
I have now got the underlying changes ready to make this api change a feature in another pull request

arkodg:
hey @sam-burrell hoping to come back to this PR in 1.5 weeks after v1.4 is out

zhaohuabing:
> @zhaohuabing is there anything I need to do to get these final checks to pass?
> 
> * [Build and Test / conformance-test (v1.33.0, dual) (pull_request)](https://github.com/envoyproxy/gateway/actions/runs/14773182828/job/41477170476?pr=5455)Failing after 15m
> * [Build and Test / e2e-test (v1.32.3, ipv6) (pull_request)](https://github.com/envoyproxy/gateway/actions/runs/14773182828/job/41477170429?pr=5455)Failing after 29m

It's fine. These tests are flaky.

api/v1alpha1/oidc_types.go
```diff
@@ -62,6 +62,12 @@ type OIDC struct {
 	// If not specified, uses the default redirect URI "%REQ(x-forwarded-proto)%://%REQ(:authority)%/oauth2/callback"
 	RedirectURL *string `json:"redirectURL,omitempty"`
 
+	// Any request that matches any of the provided matchers (with either tokens that are expired or missing tokens) will not be redirected to the OIDC Provider.
+	// This behavior can be useful for AJAX or machine requests.
+	// +optional
+	// +notImplementedHide
+	OIDCDenyRedirect *OIDCDenyRedirect `json:"denyRedirect,omitempty"`
```
zhaohuabing, line 69: Can we rename the field to `DenyRedirect`? The `OIDC` prefix feels redundant - the outer structure is already named as `OIDC`.

sam-burrell: agreed, this makes sense

api/v1alpha1/oidc_types.go
```diff
@@ -143,6 +149,25 @@ type OIDCProvider struct {
...
+type OIDCDenyRedirect struct {
+	// Defines the headers to match against the request to deny redirect to the OIDC Provider.
+	Headers []OIDCDenyRedirectHeader `json:"headers,omitempty"`
+}
+
+// OIDCDenyRedirectHeader defines how a header is matched
+// +notImplementedHide
+type OIDCDenyRedirectHeader struct {
+	// Specifies the name of the header in the request.
+	// The pseudo-headers ``:path`` and ``:method`` can be used to match the request path and method, respectively
```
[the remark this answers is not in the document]
sam-burrell: agreed, will remove this

sam-burrell:
@zhaohuabing these two comments should now be resolved

zhaohuabing:
Hi @sam-burrell  You can fix the gen-check by runnig `make gen-check`.

sam-burrell:
@zhaohuabing this is now done, I needed to rebase my fork

zhaohuabing (dismissed):
LGTM thanks!

sam-burrell:
many thanks!!

api/v1alpha1/oidc_types.go
```diff
@@ -143,6 +149,24 @@ type OIDCProvider struct {
...
+
+// OIDCDenyRedirectHeader defines how a header is matched
+// +notImplementedHide
+type OIDCDenyRedirectHeader struct {
+	// Specifies the name of the header in the request.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// If specified, header match will be performed based on the string match of the header value.
+	// Specifies how the header match will be performed to route the request.
+	StringMatch StringMatch `json:"stringMatch"`
```
arkodg, line 167: this needs to be an embedded field so the api looks like
```
name: foo
value: bar
```

sam-burrell: doesnt the already defined StringMatch struct match more comfortably onto the underlying envoy denyRedirect method in envoy?

sam-burrell: giving the flexibility to use regex, prefix or suffix matching?

sam-burrell:
```yaml
  kind: SecurityPolicy
  metadata:
    namespace: envoy-gateway
    name: policy-for-gateway  # This policy should attach httproute-2
    uid: b8284d0f-de82-4c65-b204-96a0d3f258a1
  spec:
    targetRef:
      kind: Gateway
      name: gateway-1
    oidc:
      provider:
        issuer: "https://accounts.google.com"
      clientID: "client1.apps.googleusercontent.com"
      clientSecret:
        name: "client1-secret"
      redirectURL: "https://www.example.com/bar/oauth2/callback"
      logoutPath: "/bar/logout"
      denyRedirect:
        headers:
        - name: :path
          stringMatch:
            type: Prefix
            value: /api
        - name: X-No-Redirect
          stringMatch:
            type: RegularExpression
            value: ".*"
        - name: X-No-Redirect
          stringMatch:
            type: Exact
            value: "foobar"
        - name: :path
          stringMatch:
            type: Suffix
            value: "bar"
      cookieNames:
        idToken: "CustomIdTokenCookie"
        accessToken: "CustomAccessTokenCookie"
```

arkodg: @sam-burrell I was referring to avoiding the explicit `stringMatch` keyword
https://gobyexample.com/struct-embedding

sam-burrell: so more like? @arkodg 

```yaml
name: Foo
type: Exact
value: Bar
```

arkodg: Yah exactly

sam-burrell: @arkodg how does that look now?

api/v1alpha1/oidc_types.go
```diff
@@ -153,6 +159,27 @@ type OIDCProvider struct {
...
+// OIDCDenyRedirectHeader defines how a header is matched
+// +notImplementedHide
+type OIDCDenyRedirectHeader struct {
+	// Specifies the name of the header in the request.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// The type of match to perform on the header.
+	// Valid MatchType values are "Exact", "Prefix", "Suffix", "RegularExpression".
+	// +kubebuilder:validation:Enum=Exact;Prefix;Suffix;RegularExpression
+	Type *StringMatchType `json:"type"`
```
arkodg, line 178: instead of redefining these field you can achieve the same by defining it as
```
Name string `json:"name"`
StringMatch
```

sam-burrell: aahh 

```go
type OIDCDenyRedirectHeader struct {
	// Specifies the name of the header in the request.
	// +kubebuilder:validation:MinLength=1
	Name        string `json:"name"`
	StringMatch `json:",inline"`
}
```

does the trick

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

PR #5458, opened 2025-03-10
Add Docker as an adopter of envoy gateway

ryanhristovski:
Add Docker to adopters of Envoy Gateway

Description will be updated as we continue to migrate onto Envoy Gateway from our legacy ingress.Signed-off-by: Ryan Hristovski <ryan.hristovski@docker.com>


zirain (approved):

arkodg (approved):
Thank you for the support @ryanhristovski @nezdolik ❤️ !

PR #5459, opened 2025-03-10
Specify a possible valid root CAs on darwin

mathetake:
**What type of PR is this?**

Make BackendTLSPolicy work on macOS

Without this, BTP will never work on macOS via standalone mode.

### DIFF
--- a/internal/xds/translator/translator.go
+++ b/internal/xds/translator/translator.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/xds/translator/translator.go
```diff
@@ -1031,6 +1032,31 @@ const (
...
+//
+// This assumes the Envoy running in a very specific environment. For example, the default location of the system
+// trust store on Debian derivatives like the envoy-proxy image being used by the infrastructure controller.
+//
+// TODO: this might be configurable by an env var or EnvoyGateway configuration.
+var defaultCertificateName = func() string {
+	switch runtime.GOOS {
+	case "darwin":
+		// TODO: maybe automatically get the keychain cert? That might be macOS version dependent.
+		// For now, we'll just use the root cert installed by Homebrew: brew install ca-certificates.
```
mathetake, line 1045: i am trying to see if it's possible to use `x509` packag, extract a cert and write it into a temporary file. But that would result in the weird skew in the logic of this translator: the translator shouldn't be aware of the deployment mode: whether or not it's running on host or inside container shouldn't be a matter of translator? idk. In anyways, it looks tricky to get the []byte of a cert withe `x509` so no luck it seems

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

cnvergence (approved):
/approve

mathetake:
good to go?

zirain:
this make `make -k gen-check` generated a differnet output on mac with the CI runner(which is linux based).

mathetake:
Ah

zirain:
see https://github.com/envoyproxy/gateway/actions/runs/13802919554/job/38608503715

PR #5463, opened 2025-03-11
chore: support customize `API_SERVER_ADDRESS`

zirain:
1. support customize `API_SERVER_ADDRESS`
2. change default to `127.0.0.1` on mac


zhaohuabing:
Is this going to be used in IPV6 e2e tests?

zirain:
> Is this going to be used in IPV6 e2e tests?

you can always use `127.0.0.1`, this patch provide a better solution.

PR #5466, opened 2025-03-11
fix: Wasm cache directory on macOS host

mathetake:
**What type of PR is this?**

fix: permission error on startWasmCache

Previously, the cache directory for Wasm binaries is hardcoded to /var/lib/eg/wasm but it's not writable without root permission on macOS. This fixes it by making it default to ~/.eg/wasm.

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path"
 	"reflect"
 
 	"github.com/docker/docker/pkg/fileutils"
@@ -36,8 +37,6 @@ import (
 )
 
 const (
-	wasmCacheDir = "/var/lib/eg/wasm"
-
 	// Default certificates path for envoy-gateway with Kubernetes provider.
 	serveTLSCertFilepath = "/certs/tls.crt"
 	serveTLSKeyFilepath  = "/certs/tls.key"
@@ -96,10 +95,9 @@ func (r *Runner) startWasmCache(ctx context.Context) {
 		r.Logger.Error(err, "failed to start wasm cache")
 		return
 	}
-
-	// Create the file directory if it does not exist.
-	if err = fileutils.CreateIfNotExists(wasmCacheDir, true); err != nil {
-		r.Logger.Error(err, "Failed to create Wasm cache directory")
+	wasmDir, err := getWasmCacheDir()
+	if err != nil {
+		r.Logger.Error(err, "failed to start wasm cache")
 		return
 	}
 	r.wasmCache = wasm.NewHTTPServerWithFileCache(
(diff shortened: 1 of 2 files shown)


internal/gatewayapi/runner/runner.go
```diff
@@ -110,11 +108,24 @@ func (r *Runner) startWasmCache(ctx context.Context) {
...
 		}, r.Logger)
 	r.wasmCache.Start(ctx)
 }
 
+func getWasmCacheDir() (string, error) {
+	homedir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	wasmDir := path.Join(homedir, ".eg", "wasm")
```
arkodg, line 121: can we `switch` this based on `GOOS` to minimize the change here ?

mathetake: if that works works for me too

mathetake: done

arkodg (approved):
LGTM thanks !

zirain (approved):

zhaohuabing (approved):
LGTM, thanks!

mathetake:
@zhaohuabing @arkodg ` /var/lib/eg` seems like a bad choice on non default Envoy container so I think I would like to make it configurable via a command line arg like `--wasm-cache-dir`. wdyt

> 2025-03-12T17:13:58.812Z	ERROR	gateway-api	runner/runner.go:110	Failed to create Wasm cache directory	{"runner": "gateway-api", "error": "mkdir /var/lib/eg: permission denied"}

https://github.com/envoyproxy/ai-gateway/actions/runs/13817304087/job/38653772222

arkodg:
@zhaohuabing has more context on the design, can you help with this, relates to https://github.com/envoyproxy/gateway/blob/76c57ecc1ab9c78236361b2d2178d12806cc2857/tools/docker/envoy-gateway/Dockerfile#L3

mathetake:
or is it possible to disable the file cache at all when the dir is not available (it should be easy if the abstraction is right at the moment? )

zhaohuabing:
> @zhaohuabing has more context on the design, can you help with this, relates to

The design considerations are outlined in this PR description: https://github.com/envoyproxy/gateway/pull/3564#issue-2339441124  . Some of them definitely have room for improvement.

> or is it possible to disable the file cache at all when the dir is not available (it should be easy if the abstraction is right at the moment? )

Just disabling the cache will break the Wasm feature, it can not function without the cache. In the initial design, we don't want to put Wasm in memory because it can lead to unexpected OOM of EG.

Our choices are:
* Make the cache directory configurable, or
* Disable the Wasm feature as a whole

I prefer the first option, as we need to make this feature work in VM deployment where this directory may not exist.

mathetake:
Yeah let's make it configurable. Passing it as an argument of envoy-gateway CLI makes sense to you ?

mathetake:
If /tmo/eg works here then we won't need to add an option in the first place ?

mathetake:
Just to clarify that it's not only about Mac but also non root Linux env as well (see the CI failure I pasted above, it's  on Ubuntu GHA)

mathetake:
btw the whole point of the conversation here is to have a consistent output from `make -k gen-check`; is that resolved by the k8s/host swtich ?

mathetake:
(deleted the confusing comment)

zhaohuabing:
Should be ~/.eg/wasm as the cached Wasm files are not temp files.

PR #5467, opened 2025-03-11
fix: do not log error when wasm cache server properly closed

mathetake:
**What type of PR is this?**

fix: incorrect handling of err returned by ListenAndServe.

Previously, the error returned by `ListenAndServe` was unconditionally treated as an error hence a log saying like "failed to start server etc" was emitted when the envoy-gateway is properly shutdown. `ErrServerClosed` is always returned by ListenAndServe and its variant so this make the change to ignore that error.

Release Notes:No


zirain (approved):

arkodg (approved):
thanks !

PR #5469, opened 2025-03-11
refactor(make): use helm upgrade --install in e2e tests

Chever-John:
This change replaces all helm install commands with helm upgrade --install in the e2e testing process. Using helm upgrade --install provides idempotent deployment behavior, automatically handling both initial installations and subsequent updates without requiring manual namespace deletion between test runs.

Previously, failed e2e tests required manual cleanup of the envoy-gateway-system namespace before retrying, significantly slowing down the development workflow. This change streamlines the testing process, allowing for continuous make e2e runs without manual intervention.

Benefits of helm upgrade --install:
- Eliminates need to manually delete namespaces between test runs
- Properly handles both new installs and upgrades with a single command
- Maintains release history for easier debugging
- Handles failed installations more gracefully
- Significantly improves developer experience during iterative testing

**What type of PR is this?**

This PR replaces all helm install commands with helm upgrade --install in the e2e testing process. During development, when running make e2e repeatedly, the current implementation requires manual deletion of the envoy-gateway-system namespace between runs if tests fail. This creates unnecessary friction in the development workflow.

Using helm upgrade --install provides idempotent behavior that automatically handles both initial installations and updates, eliminating the need for manual namespace deletion between test runs. This significantly improves the developer experience by allowing continuous make e2e runs without manual intervention.

Fixes #

Release Notes: Yes

### DIFF
--- a/tools/make/kube.mk
+++ b/tools/make/kube.mk
@@ -93,19 +93,19 @@ endif
 .PHONY: kube-deploy
 kube-deploy: manifests helm-generate.gateway-helm ## Install Envoy Gateway into the Kubernetes cluster specified in ~/.kube/config.
 	@$(LOG_TARGET)
-	helm install eg charts/gateway-helm --set deployment.envoyGateway.imagePullPolicy=$(IMAGE_PULL_POLICY) -n envoy-gateway-system --create-namespace --debug --timeout='$(WAIT_TIMEOUT)' --wait --wait-for-jobs
+	helm upgrade --install eg charts/gateway-helm --set deployment.envoyGateway.imagePullPolicy=$(IMAGE_PULL_POLICY) -n envoy-gateway-system --create-namespace --debug --timeout='$(WAIT_TIMEOUT)' --wait --wait-for-jobs
 
 .PHONY: kube-deploy-for-benchmark-test
 kube-deploy-for-benchmark-test: manifests helm-generate ## Install Envoy Gateway and prometheus-server for benchmark test purpose only.
 	@$(LOG_TARGET)
 	# Install Envoy Gateway
-	helm install eg charts/gateway-helm --set deployment.envoyGateway.imagePullPolicy=$(IMAGE_PULL_POLICY) \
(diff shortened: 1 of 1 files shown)


zirain:
@Chever-John can you fix DCO?

zhaohuabing:
I think we want a clean environment for the e2e tests, and helm upgrade may cause unexpected result.

Chever-John:
@zhaohuabing 

Got u.

How about `--atomic` and `--cleanup-on-fail` flag?

and add `helm uninstall eg --ignore-not-found` before.

The whole command as following:

```shell
helm uninstall eg --ignore-not-found
helm upgrade --install eg charts/gateway-helm \
  --set deployment.envoyGateway.imagePullPolicy=$(IMAGE_PULL_POLICY) \
  -n envoy-gateway-system \
  --create-namespace \
  --debug \
  --timeout='$(WAIT_TIMEOUT)' \
  --wait \
  --wait-for-jobs \
  --atomic \
  --cleanup-on-fail
  ```

**Comment by Chever-John** [comment]
After a detailed discussion with @zhaohuabing , I've decided to cancel this PR submission.

Initially, I was following the `make e2e` command in the [documentation](https://gateway.envoyproxy.io/contributions/develop/#:~:text=Run%20make%20e2e%20to%20perform%20end%2Dto%2Dend%20testing.) locally, but encountered various issues specific to my development machine (network problems, missing libraries, etc.). The `make e2e` command would break frequently and require re-run locally. When attempting to rerun it, I would receive an error stating that "helm install eg" had already been installed, forcing me to execute helm uninstall or kubectl delete ns envoy-gateway-system to resolve the issue.

After thorough discussion with @zhaohuabing , we concluded that this approach could lead to an unpredictable CI runtime environment and potentially contaminate developers' local validation of e2e tests. We determined that a more predictable and consistent solution for e2e public testing was necessary.

Importantly, he suggested a better development workflow using the make run-e2e command.

First, prepare the environment by running:

```bash
make create-cluster kube-install-image kube-deploy install-ratelimit install-eg-addons kube-install-examples-image e2e-prepare
After the environment preparation is complete, execute the following command for development:
```

```bash
make run-e2e
This approach provides a more stable and reliable development environment for e2e testing.
```

Thks~

PR #5470, opened 2025-03-11
Make file provider errors useful

mathetake:
**What type of PR is this?**

Enhances error message with file provider.

Previously, the file provider doesn't log any information about which file is wrong, so this fixes it and adds the missing test coverage.


test/helm/gateway-addons-helm/default.out.yaml
mathetake, line 1: what

arkodg (approved):

Xunzhuo (approved):
Troubleshooting improves, thanks!

Issue #3560, opened 2024-06-06
Support Dynamic HTTP Forward Proxy


guydc:
*Description*:
In some cases, the backend address is not known in advance and must be inferred from request/connection attributes (such as a user-provided host header, jwt claim, TLS SNI, etc.). 

The Envoy Dynamic Forward proxy feature can be used to achieve dynamic routing. EG can support Dynamic Forward Proxy in the following manner:
- The EG Backend API can be extended to support a "dynamic" endpoint type, representing a dynamic envoy cluster. This is mostly required for BTLS policy attachment. 
- The EG BackendTrafficPolicy can be used to enable dynamic routing: 
  - Injecting the appropriate HCM filter
  - Defining the strategy for inferring the upstream address
  - Configuring a DNS cache policy

- [x]  Implement Dynamic Resolver
- [x]  E2E
- [ ]  Docs
- [ ]  Add warnings for Dynamic Resolver in docs
- [ ]  Support appProtocol for Dynamic Resolver

[optional *Relevant Links*:]
- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy
- https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/sni_dynamic_forward_proxy_filter
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/clusters/dynamic_forward_proxy/v3/cluster.proto#envoy-v3-api-msg-extensions-clusters-dynamic-forward-proxy-v3-clusterconfig

arkodg:
+1 to this

arkodg:
@guydc what would the UX of this be like ? a filter like `Redirect` and `DirectResponse` or a `Backend` with a special field set

rudrakhp:
Maybe a new [ `BackendEndpoint`](https://gateway.envoyproxy.io/docs/api/extension_types/#backendendpoint) type for dynamic usecases, something like `DynamicEndpoint`?

guydc:
Yea, `DynamicEndpoint` sound ok to me. We need the ability to attach BTLSP and BTP for various cluster-level options, so I think that a filter won't be enough. 

[...]

PR #5487, opened 2025-03-13
api: adds Type on Backend 

mathetake:
**What type of PR is this?**

api: adds Type on Backend 

To add support for the "dynamic forward proxy" #3560, we need a special backend that acts as a dns resolver from host header. That's the only thing we need from the next Envoy v1.34 as discussed in envoyproxy/envoy#38452.

Release Notes: No (after the implementation is done)


mathetake:
cc @arkodg

api/v1alpha1/backend_types.go
```diff
@@ -113,6 +113,13 @@ type UnixSocket struct {
 
 // BackendSpec describes the desired state of BackendSpec.
 type BackendSpec struct {
+	// Type defines the type of the backend. Defaults to "Endpoints"
+	//
+	// +kubebuilder:validation:Enum=Endpoints;DynamicForwardProxy
+	// +kubebuilder:default=Endpoints
+	// +optional
+	Type BackendType `json:"type"`
```
zirain, line 121: if this's optional, it should be an point and with `omitempty `.

mathetake: so not to introduce the breaking change, this needs to be optional. However, we haven't used a pointer for "type" elsewhere (i guess this will be the first time?). is that ok with you

mathetake: done

zirain:
do we need some cel rule for the conflict between type and other fields?

mathetake:
i guess when type=DyanmicForwardProxy, EG will effectively ignore every other existing field. Not sure if it's worth the validation or not.  At least we should documenta that. wdyt? @arkodg

api/v1alpha1/backend_types.go
```diff
@@ -135,6 +142,16 @@ type BackendSpec struct {
...
 }
 
+// BackendType defines the type of the Backend.
+type BackendType string
+
+const (
+	// BackendTypeEndpoints defines the type of the backend as Endpoints.
+	BackendTypeEndpoints BackendType = "Endpoints"
+	// BackendTypeDynamicForwardProxy defines the type of the backend as DynamicForwardProxy.
+	BackendTypeDynamicForwardProxy BackendType = "DynamicForwardProxy"
```
zhaohuabing, line 152: Nit : `DynamicForwardProxy` is the role of Envoy, should we call the Backend type `DynamicResolve` ?

mathetake: Yeah im open for any choice and agree DynamicForeqrdProxy is a bit confusing. How about DynamicResolver?

zhaohuabing: DynamicResolver SGTM

mathetake:
* Renamed DynamicForwardProxy to DynamicResolver
* Added the CEL validation.

arkodg:
thanks this looks good @mathetake !
adding.a comment around naming, also happy to pick up the implementation PR

api/v1alpha1/backend_types.go
```diff
@@ -135,6 +143,22 @@ type BackendSpec struct {
...
+	// BackendTypeEndpoints defines the type of the backend as Endpoints.
+	BackendTypeEndpoints BackendType = "Endpoints"
+	// BackendTypeDynamicResolver defines the type of the backend as DynamicResolver.
+	//
+	// When a backend is of type DynamicResolver, the Envoy will resolve the upstream
+	// ip address and port from the host header of the incoming request. If the ip address
+	// is directly set in the host header, the Envoy will use the ip address and port as the
+	// upstream address. If the hostname is set in the host header, the Envoy will resolve the
+	// ip address and port from the hostname using the DNS resolver.
+	BackendTypeDynamicResolver BackendType = "DynamicResolver"
```
arkodg, line 159: throwing some more options -
1)  "DynamicResolver"
2) "DynamicForwardProxy"
3) "HostResolver"
4) "ForwardProxy"

ptal @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

zhaohuabing: +1 DynamicResolver or HostResolver

Xunzhuo: +1 for DynamicResolver

guydc:
- Why are we not allowing appProtocols here? Is it just a matter of scope reduction, or is there a real limitation here?
- We should extend the Backend docs with a major security warning. With dynamic routing, the proxy can be misused in a variety of ways, depending on how it's configured. 
  - Right now, Backend can be used to route to cluster-internal endpoints that were explicitly enabled by the persona creating Backend CRs. With dynamic proxy, all endpoints are implicitly exposed. 
  - Sending requests with injected credentials to an attacker-controlled endpoint
  - Being used as part of a botnet 
  - Sending authenticated requests (e.g. using the proxy's client certificate) to protected resources on behalf of an attacker

arkodg:
> * Why are we not allowing appProtocols here? Is it just a matter of scope reduction, or is there a real limitation here?
> * We should extend the Backend docs with a major security warning. With dynamic routing, the proxy can be misused in a variety of ways, depending on how it's configured.
>   
>   * Right now, Backend can be used to route to cluster-internal endpoints that were explicitly enabled by the persona creating Backend CRs. With dynamic proxy, all endpoints are implicitly exposed.
>   * Sending requests with injected credentials to an attacker-controlled endpoint
>   * Being used as part of a botnet
>   * Sending authenticated requests (e.g. using the proxy's client certificate) to protected resources on behalf of an attacker

thanks for raising these,
+1 to adding these warning in the doc, I'll do post implementation
+1 to loosening the appProtocol support in the future, will raise a GH issue for it

api/v1alpha1/backend_types.go
guydc: I'm fine with DynamicResolver. I don't think that any of the options expresses well what this does, which is to use an endpoint from the incoming request. I don't have a good idea for a name that expresses this.

arkodg: yah same, lets pick DynamicResolver for now then

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

zirain (approved):

Xunzhuo (approved):

zhaohuabing:
> Why are we not allowing appProtocols here? Is it just a matter of scope reduction, or is there a real limitation here?
We should extend the Backend docs with a major security warning. With dynamic routing, the proxy can be misused in a variety of ways, depending on how it's configured.
Right now, Backend can be used to route to cluster-internal endpoints that were explicitly enabled by the persona creating Backend CRs. With dynamic proxy, all endpoints are implicitly exposed.
Sending requests with injected credentials to an attacker-controlled endpoint
Being used as part of a botnet
Sending authenticated requests (e.g. using the proxy's client certificate) to protected resources on behalf of an attacker

Just think it loud, should we add a global configuration knob similar to the ISTIO egress traffic policy to EG, default to `REGISTRY_ONLY`, and only allow `DynamicResolver` type Backend when egress traffic policy is set to `ALLOW_ANY`? 

This would give the cluster administrator control whether egress traffic should be loosed to any backend to mitigate the security risk.

@arkodg @guydc @mathetake

PR #5498, opened 2025-03-14
fix: otel sink json access logging without text field

### LINKED ISSUE #5481: OpenTelemetry access log sink failure with "otel.Text is nil" error (EG v1.3.1)
*Description*:
After upgrading our development clusters to v1.3.1 of Envoy Gateway, we noticed that none of our HTTPRoutes in the cluster were not registering to the proxy listeners, making all services unavailable. This was validated with `egctl config envoy-proxy listener` and `... routes`, showing that none of the configured routes were registered.

```
❯ curl https://<url> -vvv
* Host <url>:443 was resolved.
* IPv6: (none)
* IPv4: ...
*   Trying ...
* Connected to ... port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to ...:443
* Closing connection
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to ...:443 

❯ openssl s_client -connect ...:443
Connecting to ...
CONNECTED(00000005)
4088A9F501000000:error:0A000126:SSL routines::unexpected eof while reading:ssl/record/rec_layer_s3.c:693:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 0 bytes and written 343 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
Protocol: TLSv1.3
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
```

In the envoy logs, these log lines were visible:

```
2025-03-12T11:13:00.577Z	ERROR	xds-translator	runner/runner.go:85	failed to translate xds ir	{"runner": "xds-translator", "error": "otel.Text is nil\notel.Text is nil"}
2025-03-12T11:13:00.577Z	INFO	provider.demo-keyauth.demo-keyauth	kubernetes/status_updater.go:109	status unchanged, bypassing update	{"runner": "provider"}
2025-03-12T11:13:00.577Z	INFO	provider	kubernetes/status_updater.go:145	received a status update	{"runner": "provider", "namespace": "<namespace>", "name": "<name>"}
2025-03-12T11:13:00.577Z	ERROR	watchable	message/watchutil.go:80	observed an error	{"runner": "xds-translator", "error": "otel.Text is nil\notel.Text is nil"}
```

After disabling the `openTelemetry` sink from the proxy access log config, the routes were configured and services became accessible.

The lines generating the logs seems to originate here: https://github.com/envoyproxy/gateway/blob/df2bb96c32604554f50ec08702cbea834ce968f5/internal/xds/translator/accesslog.go#L339-L341

When downgrading back to v1.3.0, the existing telemetry sink configuration works again.

*Repro steps*:

Envoy proxy configmap includes following telemetry config:

```
...
  telemetry:
    accessLog:
      settings:
      - format:
          json:
            ...
          type: JSON
        sinks:
        - openTelemetry:
            host: telemetry-opentelemetry-collector.telemetry.svc.cluster.local
            port: 4317
            resources:
              k8s.cluster.name: <cluster-name>
          type: OpenTelemetry
        - file:
            path: /dev/stdout
          type: File
```

*Environment*:

`envoy-gateway v1.3.1`, templated and installed via latest Helm Chart.

*Logs*:

Included in description.

kraashen:
**What type of PR is this?**

This PR attempts to fix issue #5481 by passing only the JSON attributes to the otel formatter builder. It does this by leaving the text format empty when JSON type is specified but text is not.

As described in the linked issue, this leads to a situation where HTTPRoutes are not registering to listeners due to error returned in situations where JSON logging is used. This issue seemed to be rising when OpenTelemetry sink type is used without `text` field in the configuration.

Tested this in a local cluster calling the test API, and logs seemed to be produced in Loki when queried with `exporter=OTEL` label.
image

Fixes #5481

Release Notes: Yes

**Note**

I have little to none prior experience of Envoy Gateway development. Went through the [guidelines](https://gateway.envoyproxy.io/contributions/develop/) with best effort, but I'm a bit unsure if there are any possible other downstream consequences to this change. Any advice would be appreciated!


internal/xds/translator/accesslog.go
```diff
@@ -336,12 +336,10 @@ func buildXdsAccessLog(al *ir.AccessLog, accessLogType ir.ProxyAccessLogType) ([
 			ResourceAttributes: convertToKeyValueList(otel.Resources, false),
 		}
 
-		if otel.Text == nil {
-			return nil, errors.New("otel.Text is nil")
-		}
-		format := *otel.Text
+		var format string
```
zirain, line 339: In v1.3.0-, we use `EnvoyTextLogFormat` as default value when unset, cc @guydc

kraashen: Yep. As the `Text` format as default logging format was superceded by JSON in v1.3.1, it's default value was removed. That's where I presumed that text logging would require defining the logging format, thus having no preset value as default. Would it be better if a default format would still be provided?

zirain (commented):
LGTM, defer to @guydc

internal/xds/translator/accesslog.go
```diff
@@ -336,12 +336,10 @@ func buildXdsAccessLog(al *ir.AccessLog, accessLogType ir.ProxyAccessLogType) ([
...
+		if otel.Text != nil && *otel.Text != "" {
+			format = *otel.Text
```
kraashen, line 341: Should we also do a separate check for JSON logging? I.e. check if JSON attributes are defined or not, and in case not, default to `EnvoyJSONLogFields`?

Something along the lines

```go
if len(otel.Attributes) != 0 {
	al.Attributes = convertToKeyValueList(otel.Attributes, true)
	formatters = accessLogOpenTelemetryFormatters(format, otel.Attributes)
} else {
	// if there are no attributes, use the default EnvoyJSONLogFields
	al.Attributes = convertToKeyValueList(EnvoyJSONLogFields, true)
	formatters = accessLogOpenTelemetryFormatters(format, EnvoyJSONLogFields)
}
```

zirain: sounds goo to me, cc @guydc

kraashen: Sorry this took a bit of time. Went through the OpenTelemetry specs and documentation on the field. Implemented following:

- When no text or JSON type is set, default to `EnvoyJSONLogFields`, similar to previous implementation with the respective text format prior to v1.3.1/v1.2.7

The JSON fields are populated to the accessLog `Attributes` field as before with the JSON fields. Went through the [OpenTelemetry specification discussion](https://github.com/open-telemetry/opentelemetry-specification/issues/1613) on this and relevant [docs](https://github.com/dynatrace-oss-contrib/semantic-conventions/blob/f6af53c0c580f75ebcc7c373d51d92017d9ed2f1/specification/logs/data-model.md#field-attributes) and to me it seems that this should be an OK way to go.

guydc (approved):
LGTM, thanks for fixing this and adding a test!

internal/xds/translator/accesslog.go
```diff
@@ -349,9 +349,17 @@ func buildXdsAccessLog(al *ir.AccessLog, accessLogType ir.ProxyAccessLogType) ([
 			}
 		}
 
-		al.Attributes = convertToKeyValueList(otel.Attributes, true)
+		var attrs map[string]string
```
arkodg, line 352: doesnt this feel weird, if we have a format its part of `Body`, else `Attributes`

arkodg: can we move this to another GH issue and only have the fix around
```
		if otel.Text != nil && *otel.Text != "" {
			format = *otel.Text
```
in this PR so we can add it into v1.3.2 , releasing soon

kraashen: Can be done! Only concern is, how does this behave in case no output type or format/attributes are defined? As JSON logging is now default, does it render logs empty for users who have neither defined? Previously in v1.3.0, the `EnvoyTextLogFormat` was set as default and then it was checked if log format was defined. https://github.com/envoyproxy/gateway/blob/v1.3.0/internal/xds/translator/accesslog.go#L325-L336

arkodg (approved):
LGTM thanks !

internal/xds/translator/accesslog.go
arkodg: this looks good, sorry for the confusion

arkodg:
@kraashen can you fix DCO and force push

Issue #3213, opened 2024-04-17
refactor the kubernetes provider to be provider-insensitive


shawnh2:
*Description*:

For now, all the resources related process method in k8s provider are coupling deeply with k8s client, we should free these method from it.

https://github.com/envoyproxy/gateway/blob/e8b8074fc7e9a23e3e12d6ade55307b618f421fa/internal/provider/kubernetes/controller.go#L698

So I propose a refactor: **the basic idea is to separate how we process resources and how we retrieve resources.** 

By defining `Operations` interface and `ResourceProcessor` structure, each provider can maintain their own implemention of `Operations`, and different provider can share the same `ResourceProcessor`.

```go
type Operations interface {
    ListGateways()
    ListHTTPRoutes()
    FindReferenceGrant()
    ...
}

type ResourceProcessor struct {
    Operations
}

func (r *ResourceProcessor) ProcessGateways() {
    // use r.ListGateways() to retrieve all Gateway resources
}
func (r *ResourceProcessor) ProcessHTTPRoutes() {}
func (r *ResourceProcessor) findReferenceGrant() {}
...
```

we can

- let k8s provider only foucs on how to retrieve resources, so does file provider
- the k8s and file provider can share the same process logic for resources, and we don't need to maintain two places of code for each provider type

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Xunzhuo:
Make sense to me, it helps for multi-provider support.

shawnh2:
will try to first refactor the current existing k8s provider, and then implement for the new coming file provider.

sky92zwq:
Is it possible with this refactoring that part of the gateway resources come from k8s and another part from files ( Gateway api related)?

arkodg:
hey @sky92zwq we'll need to update the `Custom` provider API and add a `Aggregate` ish resource provider to handle the case you highlighted

shawnh2:
Found a simple yet effective way to let other provider (like file provider) to reuse the k8s reconcile logic.

Create a offline k8s controller structure, like:

```go
package kubernetes

import (
	"context"
	"fmt"

	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/apimachinery/pkg/util/sets"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/client/fake"
	"sigs.k8s.io/controller-runtime/pkg/reconcile"
[...]

PR #5507, opened 2025-03-15
feat: respect event type while handling events in file provider

shawnh2:
**What type of PR is this?**

respect event type while doing `HandleEvent` in file-provider:
- for remove event type, just reload
- for other type, reload only the file that has been changed
  - if the reload resource can be indexed (by the matching key of GatewayClass) in current all resources, let's update it
  - otherwise also reload all resources
  - doing this is under the assumption that the GatewayClass should not be changed very often

Release Notes: Yes


shawnh2:
closing this one, can be redone after https://github.com/envoyproxy/gateway/issues/3213#issuecomment-2816518741

Issue #5256, opened 2025-02-11
ReplacePrefixMatch in the URLRewrite filter does not handle empty match prefix correctly


sadovnikov:
*Description*:

In its [examples](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPPathModifier), the Gateway API uses "foo" in "Prefix Match". However, we can conclude that the `HTTPPathModifier` should also work on `/` as the path prefix match.

The PR #5254 PR demonstrates that `/` is not handled correctly.

*Repro steps*:

Follow the end-2-end test, which is added in #5254 

*Environment*:

In our environment, we use `docker.io/envoyproxy/gateway:v1.3.0`. The bug is reproducible in the `main` branch, and is reproducible in `v1.2.5`

arkodg:
relates to https://github.com/kubernetes-sigs/gateway-api/issues/3592

we've had some issues in this area 
* https://github.com/envoyproxy/gateway/issues/2976
* https://github.com/envoyproxy/gateway/issues/3807
* https://github.com/envoyproxy/gateway/issues/4331

hoping someone can help with ironing this out

arkodg:
cc @zhaohuabing @LiorLieberman

LiorLieberman:
Are we good to employ https://github.com/istio/istio/pull/54939?

Will also start a discussion in slack

sadovnikov:
@LiorLieberman, @arkodg, can we still expect the fix to be in 1.4?

sadovnikov:
Meanwhile, the #5254 PR with e2e tests got closed as stale

yoctozepto:
Since Gateway API is still working on common semantics, could we at least document that the handling of a lonely `/` is currently astonishing? [1]

[1] https://en.wikipedia.org/wiki/Principle_of_least_astonishment

wdoekes:
I'm not sure astonishment is the right word here.
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
spec:
  hostnames: [example.com]
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: https
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: some-service
      port: 443
    filters:
    - type: URLRewrite
      urlRewrite:
        path:
          # Needs extra slash! Something eats one char.
          replacePrefixMatch: /myprefix//
          type: ReplacePrefixMatch
[...]

PR #5516, opened 2025-03-17
xds: use regex rewrite when prefix match all

### LINKED ISSUE #4331: replacePrefixMatch broken in main
*Description*:
returned, etc.

with envoy gateway 1.1 I was using following spec

```
    - type: URLRewrite
      urlRewrite:
        path:
          type: ReplacePrefixMatch
          replacePrefixMatch: /docs/
```

So my goal is to add `/docs/` prefix to all url paths. So it will be modified like `/docs/<originalpath>`.

However, with latest main it will not add the ending slash anymore. I was forced to use 

```
    - type: URLRewrite
      urlRewrite:
        path:
          type: ReplacePrefixMatch
          replacePrefixMatch: /docs//
```

to get same behaviour working.

*Repro steps*:

sharing.

*Environment*:

eg 1.1 and latest main

*Logs*:

zirain:
Fixes: https://github.com/envoyproxy/gateway/issues/4331 https://github.com/envoyproxy/gateway/issues/5256

Use logic from https://github.com/istio/istio/pull/54939

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -421,16 +421,17 @@ func buildXdsURLRewriteAction(destName string, urlRewrite *ir.URLRewrite, pathMa
 			// An empty replace string does not seem to solve the issue so we are using
 			// a regex match and replace instead
 			// Remove this workaround once https://github.com/envoyproxy/envoy/issues/26055 is fixed
-			if useRegexRewriteForPrefixMatchReplace(pathMatch, *urlRewrite.Path.PrefixMatchReplace) {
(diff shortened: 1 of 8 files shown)


zirain:
should we wait for https://github.com/kubernetes-sigs/gateway-api/issues/3592?

howardjohn:
> should we wait for [kubernetes-sigs/gateway-api#3592](https://github.com/kubernetes-sigs/gateway-api/issues/3592)?

IMO this would be good, I have been waiting for this for Istio

Issue #5182, opened 2025-01-31
problem with all routes 404 when using incorrect filters


arkodg:
> Same problem with all routes 404 when using filters
> ```
> ---
> apiVersion: gateway.networking.k8s.io/v1
> kind: HTTPRoute
> metadata:
>   name: something
> spec:
>   parentRefs:
>     - name: eg
>       namespace: something
>       sectionName: something
>   hostnames:
>     - something.com
>   rules:
>     - backendRefs:
>         - group: ""
>           kind: Service
>           name: something
>           port: 123
>           weight: 1
>       filters:
>         - type: RequestHeaderModifier
>           requestHeaderModifier:
>             add:
>               - name: Host
>                 value: something.com
>               - name: X-Forwarded-Proto
>                 value: https
>               - name: X-Forwarded-Host
>                 value: something.com
>       matches:
>         - path:
>             type: PathPrefix
>             value: /
> ```
> I understand that filters are not needed here and maybe they are written incorrectly, I'm just making a report that the problem in one httproute affects all httroutes in the cluster.
>  

 _Originally posted by @dda104 in [#5147](https://github.com/envoyproxy/gateway/issues/5147#issuecomment-2626491121)_

surenraju:
Currently, in `snapshotcache.go` (line 332), when Envoy rejects an xDS update, we log it at DEBUG level:

https://github.com/envoyproxy/gateway/blob/main/internal/xds/cache/snapshotcache.go#L332

```go
if status := req.ErrorDetail; status != nil {
    // if Envoy rejected the last update log the details here.
    // TODO(youngnick): Handle NACK properly
    errorCode = status.Code
    errorMessage = status.Message
}
s.log.Debugf("handling v3 xDS resource request, response_nonce %s, nodeID %s, node_version %s, resource_names_subscribe %v, resource_names_unsubscribe %v, type_url %s, errorCode %d, errorMessage %s",
    req.ResponseNonce,
    nodeID, nodeVersion,
    req.ResourceNamesSubscribe, req.ResourceNamesUnsubscribe,
    req.GetTypeUrl(),
    errorCode, errorMessage)
```
We should log this as an error `s.log.Errorf` instead of `Debugf`, since a failed xDS update might indicate a critical configuration issue.

zhaohuabing:
@dda104 I didn't reproduce this with EG 1.3.0. Did I miss anything?

dda104:
[...]

PR #5533, opened 2025-03-18
fix: host header should not be allowed to modify

zhaohuabing:
Fix: https://github.com/envoyproxy/gateway/issues/5182

### DIFF
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -192,15 +192,9 @@ func (t *Translator) processURLRewriteFilter(
 	if filterContext.URLRewrite != nil {
 		if hasMultipleCoreRewrites(rewrite, filterContext.URLRewrite) ||
 			hasConflictingCoreAndExtensionRewrites(rewrite, filterContext.URLRewrite) {
-			routeStatus := GetRouteStatus(filterContext.Route)
-			status.SetRouteStatusCondition(routeStatus,
-				filterContext.ParentRef.routeParentStatusIdx,
-				filterContext.Route.GetGeneration(),
-				gwapiv1.RouteConditionAccepted,
-				metav1.ConditionFalse,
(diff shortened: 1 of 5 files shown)


guydc (approved):
LGTM

internal/gatewayapi/filters.go
```diff
@@ -625,28 +626,18 @@ func (t *Translator) processResponseHeaderModifierFilter(
...
-					filterContext.ParentRef.routeParentStatusIdx,
-					filterContext.Route.GetGeneration(),
-					gwapiv1.RouteConditionAccepted,
-					metav1.ConditionFalse,
-					gwapiv1.RouteReasonUnsupportedValue,
-					fmt.Sprintf("ResponseHeaderModifier Filter cannot set headers with a '/' or ':' character in them. Header: %q", string(addHeader.Name)),
+			if !isModifiableHeader(string(addHeader.Name)) {
+				updateRouteStatusForRHMFilter(
+					filterContext,
+					fmt.Sprintf("ResponseHeaderModifier Filter cannot set headers with a '/' or ':' character in them, or the host header. Header: %q",
```
arkodg, line 639:
```suggestion
					fmt.Sprintf("Header: %q. The ResponseHeaderModifier filter cannot set the Host header or headers with a '/' or ':' character in them. To modify the Host header use the URLRewrite or the HTTPRouteFilter filter.",
```

arkodg: can we use this in other places too ?

zhaohuabing: Do we want to add this?  The host header should not be modified in the response or be deleted. @arkodg

arkodg: ah this is response, yeah you can update that line, and use this line for requestHeaderModifier, my bad

zirain (approved):

PR #5537, opened 2025-03-18
api: Add BufferRequests field to BackendTrafficPolicy

markwinter:

This adds a new field to the `BackendTrafficPolicy` called `BufferRequests`.

This field will be used to configure the `Buffer` filter in Envoy https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/buffer/v3/buffer.proto#envoy-v3-api-msg-extensions-filters-http-buffer-v3-buffer

The buffer filter allows Envoy to buffer and fully receive requests from the client before passing them upstream to the backend, so that the backend only receives the full request and is protected from slow clients/attacks. It also imposes a maximum buffer/request size and a request will be rejected if it's too large.

In the implementation, it's planned that depending on whether the targetRef is a Gateway or a HTTPRoute, it will use the Buffer or BufferPerRoute filter.

I could see this not sitting directly in the `BackendTrafficPolicy` but as part of some sub-struct dealing with Requests, or Buffering instead.

Potentially introduce a `Type` field to support alpha/not-yet-implemented buffers in the future: [filesystem](https://github.com/envoyproxy/envoy/issues/19026) or [bufferless](https://github.com/envoyproxy/envoy/issues/36946)

We could also document that the equivalent in nginx is `proxy_request_buffering` and `client_body_buffer_size` if thats desired https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_request_buffering

Ref #3706

Release Notes: No


jukie:
CC @guydc I closed my PR in favor of this one

api/v1alpha1/shared_types.go
```diff
@@ -574,6 +574,22 @@ type ClusterSettings struct {
...
+	//
+	// When enabling this option, you should also configure your connection buffer size to account for these request buffers. There will also be an
+	// increase in memory usage for Envoy that should be accounted for in your deployment settings.
+	//
+	// When the target of this policy is a gateway the buffer will be applied at a listener level. When the target of the policy is a HTTPRoute it
+	// will be applied for that particular route only.
+	//
+	// +notImplementedHide
+	// +optional
+	BufferRequests *BufferRequests `json:"bufferRequests,omitempty"`
```
zhaohuabing, line 591: I think this should be put in `BackendTrafficPolicySpec` instead of `ClusterSettings`.

`BufferRequests` shouldn't be required for non-route backends such as ext aut, ext proc, oidc provider, OTL sink, etc.

markwinter: Ah that was my intention and saw ClusterSettings is inline, but if it's used for other purposes I'll move it directly on to `BackendTrafficPolicySpec` 👍

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -78,6 +79,22 @@ type BackendTrafficPolicySpec struct {
...
+	// BufferRequests allows the gateway to buffer and fully receive each request from a client before continuing to send the request
+	// upstream to the backends. This can be helpful to shield your backend servers from slow clients, and also to enforce a maximum size per request
+	// as any requests larger than the buffer size will be rejected.
+	//
+	// This can have a negative performance impact so should only be enabled when necessary.
+	//
+	// When enabling this option, you should also configure your connection buffer size to account for these request buffers. There will also be an
+	// increase in memory usage for Envoy that should be accounted for in your deployment settings.
+	//
+	// When the target of this policy is a gateway the buffer will be applied at a listener level. When the target of the policy is a HTTPRoute it
```
arkodg, line 92: lets leave this implementation detail out from the API doc string

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -90,6 +107,18 @@ type ProtocolUpgradeConfig struct {
...
+type BufferRequests struct {
+	// MaxBufferSize specifies the maximum allowed size in bytes for each incoming request buffer.
+	// If exceeded, the request will be rejected with HTTP 413 Content Too Large.
+	//
+	// Accepts values in resource.Quantity format (e.g., "10Mi", "500Ki").
+	//
+	// +kubebuilder:validation:XIntOrString
+	// +kubebuilder:validation:Pattern="^[1-9]+[0-9]*([EPTGMK]i|[EPTGMk])?$"
+	// +notImplementedHide
+	MaxBufferSize resource.Quantity `json:"maxBufferSize,omitempty"`
```
arkodg, line 119: the word `buffer` is used twice `bufferRequests.maxBufferSize`
can we do better here ? maybe `bufferRequests.Limit` taking some inspiration from https://gateway.envoyproxy.io/docs/api/extension_types/#clientconnection

api/v1alpha1/shared_types.go
zhaohuabing: Some features are put in the `ClusterSettings` because they can also be used for non-route backends, like  ext aut, ext proc, OIDC provider, or OTL sink.

BufferRequests can be used for non-route backends. Let's put it directly in the BackendTrafficPolicySpec.

api/v1alpha1/backendtrafficpolicy_types.go
markwinter: Yer I agree with this for some consistency. The word `Limit` is also perhaps more k8s native as well

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -78,6 +79,19 @@ type BackendTrafficPolicySpec struct {
...
+	// as any requests larger than the buffer size will be rejected.
+	//
+	// This can have a negative performance impact so should only be enabled when necessary.
+	//
+	// When enabling this option, you should also configure your connection buffer size to account for these request buffers. There will also be an
+	// increase in memory usage for Envoy that should be accounted for in your deployment settings.
+	//
+	// +notImplementedHide
+	// +optional
+	BufferRequests *BufferRequests `json:"bufferRequests,omitempty"`
```
arkodg, line 94: We've used plurals for list fields.
Is requestBuffer more appropriate here ?
cc @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

rudrakhp: The name `requestBuffer` suggests it's a buffer containing request data. `RequestBufferOptions` / `RequestBufferConfig`   makes more sense, WDYT?

markwinter: Perhaps `EnableRequestBuffer` ?

arkodg: > The name requestBuffer suggests it's a buffer containing request data. 
this is true, and also we're trying to implicitly say this is the `requestBufferConfiguration` without adding the prefix of `setting` / `options` / `config` , its 
* `requestBuffer` implies `enableRequestBuffer` imo

other examples in the same API https://gateway.envoyproxy.io/docs/api/extension_types/#backendtrafficpolicyspec
* `retry`
* `healthCheck`
* `rateLimit`
* `circuitBreaker`
* `timeout`

these are feature nouns, and can be enabled by setting them

markwinter: That makes sense to me

rudrakhp (approved):

zirain (approved):

Issue #5468, opened 2025-03-11
Determine how errors in xds-translator are handled


guydc:
## Description

Currently, the XDS Translator may encounter several types of failures that would lead to an error being returned from the Translate function:
- Failure to update Routes, Network/HTTP filters, Clusters and Secrets due to unexpected missing XDS resources or unexpected existing conflicting XDS resource (existing route config, filters, default filter chain, ... ). 
- Failure to translate IR config to XDS due to missing/unsupported IR input, which was not properly verified in earlier validation phases (CEL, gateway-api translator validations, ... ) 
- Failure to apply Envoy Patch Policies: errors in input validation of patches or proto validation of the patched resources
- Failures in validation of EG-generated envoy protos during translation, leading to resources being dropped from the xds resource table.
- Failures in validation of all XDS resources post-translation, leading to errors being reported without resources being dropped. 
- Failures in translation by EG Extension Server. 

The XDS translators works in a "best-effort" manner: when errors are encountered, some resources may be omitted/changed due to errors. However, the resulting configuration is still published and saved to the XDS cache. This configuration may be partial or invalid. As a result:
- The proxy may not reflect the user's intent.
- The proxy may reject some or all of the config. This may be due to partial configuration that contains invalid resource references or invalid resources. Rejection of config creates a risk that new proxy instances would not be programmed at all. 
- Users may not notice that translation is partially failing, if the impact is limited.
- There's no way for users to understand which part of their desired state is "really" applied.
- The proxies remain in an "incorrect" while users work to identify which issue/configuration change triggered this state.

In #4155, users requested the following:

> Ideally, similar to https://github.com/envoyproxy/gateway/issues/3873, there would be an option added to [ExtensionManager](https://gateway.envoyproxy.io/docs/api/extension_types/#extensionmanager) which would allow either "fail open" (current behavior of best effort) or "fail closed" (alternate behavior of disabling the resource associated with the failed hook). 

This option was implemented in #4936. The `FailClosed` option in #4936 provides greater reliability that traffic is either handled according to configuration provided by the extension server or rejected entirely. 

#4936 is a partial solution:
- It only addresses the extension server use case
- users may prefer to handle translation errors by pausing XDS publishes until the issue is resolved and the translator produces error-free xds resources. 
  - Existing proxies are serving a recent version of XDS config which was valid at the time of the translation. 
  - By persisting a valid XDS snapshot, there is a guarantee that new proxies can still be programmed. 

[...]

PR #5540, opened 2025-03-18
Skip snapshot updates on XDS translator error

guydc:
**What type of PR is this?**

See background in #5468.

This PR proposes to change the default behavior of xds translator. xds-translator would no longer work in "best effort" mode. Instead, when errors are encountered by the translator, the resulting xds resources are not published. This behavior is more consistent with other error handling in the xds-translator today, such as handling of panic recovery and empty snapshot. 

The impact of this change is most significant in the context of the extension manager. To align with this new operational mode, when extension server is configured with `failOpen: false` (default), the extension server error is propagated and resulting xds config is not published. As a result, the current behavior introduced in #4936 which replaces affected resources with immediate response resources becomes irrelevant and it is removed. When extension manager is configured with `failOpen: true`, errors are suppressed, and the generated config is stored in the XDS cache. 

This PR includes:
- A simple extension server that can create translation errors based on resource names
- An E2E test that verifies:
  - Pauses in XDS updates after translation failures
  - Proxies are able to consume the "paused" snapshot after they restart
  - If EG is restarted, proxies are not consuming the empty XDS cache when they connect to an EG instance without a cache
 - Update translator and runner tests to reflect the new pause behavior

Relates to #5468 

Release Notes: Yes


guydc:
CC @logan-hcg

If we support a configuration pause on error feature, would you still prefer to use the Extension Manager `FailClosed` feature and publish partial configuration (where some resources associated with the failed hook are replaced) instead?

zhaohuabing:
I think all the failure are caused by some kinds of bug in the EG code, not necessarily an error, right? 
Should we expose this knob to API?

guydc:
> I think all these failures in the xDS translator are caused by some kinds of bug in the EG code, not necessarily an error that can be recovered, right?

Some, like errors in the extension manager or envoy patch policies, can be recovered: users can correct bad patch policies, extension servers can recover from temporary unavailability, etc. 

Many of the errors are indeed EG issues that should have been avoided by better validation and IR construction. Users can maybe recover from them by reverting to a previous working configuration. 

The question here is if we want to allow such errors to create unpredictable impact for end-users by publishing an xds snapshot when we **know** that something went wrong when it was created. In this situation:
- Users may not notice that translation is partially failing, if the impact is limited. 
- There's no way for users to understand which part of their desired state is "really" applied.
- The proxies remain in an "incorrect" while users work to identify which issue/configuration change triggered this state. 

> Should we expose a configuration knob to the API for handling code bugs?

I propose to change the default behavior and stop publishing new xds snapshots when errors are seen. It's a significant change that would be backwards incompatible, so I want to make it possible for users to opt-out. We can decide to expose this as more of a feature gate and eventually eliminate the knob. I'm hesitant to make such a significant behavior change without creating some sort of opt-out option, at least for a limited period of time.

internal/xds/translator/runner/runner.go
```diff
@@ -121,7 +121,11 @@ func (r *Runner) subscribeAndTranslate(sub <-chan watchable.Snapshot[string, *ir
 				result.EnvoyPatchPolicyStatuses = nil
 
 				// Publish
-				r.Xds.Store(key, result)
+				if err == nil || r.Config.EnvoyGateway.XDS.UpdateSnapshotOnError {
```
arkodg, line 124: my preference is to make this the default case and not expose a configuration (yet)
by making this change
```
if err == nil {
  r.Xds.Store(key, result)
}
```
the behaviour looks like the panic recovery case, where we dont push the config to the xds server which holds the last good state

this change may surface bugs that should have been caught in CEL or gateway-api layer, but hoping those get caught in staging and fixed in patch releases

logan-hcg:
> CC @logan-hcg
> 
> If we support a configuration pause on error feature, would you still prefer to use the Extension Manager `FailClosed` feature and publish partial configuration (where some resources associated with the failed hook are replaced) instead?

I think, yes, `FailClosed` is still better, because I think the behavior in the following situation would be different:

1. Route is added without ExtensionManager; then
2. ExtensionManager is added but fails

With `FailClosed`, I think the Route would go "offline"?  But with "pause on error" the Route would remain published and "open"?

guydc:
> With FailClosed, I think the Route would go "offline"? But with "pause on error" the Route would remain published and "open"?

Correct. If the last translation result is "not secure", it would persist, instead of being replaced. 

An approach that we can consider here would be to treat the "FailClosed" mode as a non-pausing error. That way, these partial translation results would be published.

guydc:
In last community call, proposal to move forward is:
- don't publish translation results in case of error
- do not suppress errors in extension manager FailClosed mode
- propagate errors in extension manager FailOpen mode

internal/xds/translator/translator.go
```diff
@@ -126,13 +128,18 @@ func (t *Translator) Translate(xdsIR *ir.Xds) (*types.ResourceVersionTable, erro
...
 	// If no extension exists (or it doesn't subscribe to this hook) then this is a quick no-op
 	if err := processExtensionPostTranslationHook(tCtx, t.ExtensionManager); err != nil {
-		errs = errors.Join(errs, err)
-		// Setting the configuration to fail open will mean that Envoy Gateway ignores the error and keeps the resources
-		// as they were before the extension server was called.
-		if t.ExtensionManager != nil && !(*t.ExtensionManager).FailOpen() {
+		// If the extension server returns an error, and the extension server is not configured to fail open,
+		// then replace all of the routes in the virtual host with a single route that returns an InternalServerError result.
+		// The original extension server error is suppressed.
+		if !(*t.ExtensionManager).FailOpen() {
```
arkodg, line 134: fail open becomes useless if we return an error to the xds-translator, since those changes will not push to the data plane, we should log here and not return an error imo

on a different note, we should rethink how to want to express error scenario tradeoffs and fail closed / fail open may not be able to sum it up properly

api/v1alpha1/envoygateway_types.go
```diff
@@ -503,11 +503,13 @@ type ExtensionManager struct {
 	// FailOpen defines if Envoy Gateway should ignore errors returned from the Extension Service hooks.
 	// The default is false, which means Envoy Gateway will fail closed if the Extension Service returns an error.
 	//
-	// Fail-close means that if the Extension Service hooks return an error, the relevant route/listener/resource
-	// will be replaced with a default configuration returning Internal Server Error (HTTP 500).
+	// Fail-close means that if the Extension Service hooks return an error, the error is propagated to the
+	// xds-translator. As a result, the xds configuration produced by the translator for that IR key is not
```
nareddyt, line 507: nit: xds-translator and IR key may be internal implementation details that users are not familiar with. Or do we expect them to understand these details if they are running an extension server?

nareddyt (commented):
Sorry for the late reply. Overall the new behavior looks great. This is what we'd expect in production for `FailClosed` 👍 I also don't think a separate config option is needed for this behavior change to pre-existing field.

api/v1alpha1/envoygateway_types.go
arkodg: +1, maybe we can say xDS tied to a specific Envoy Proxy fleet

api/v1alpha1/envoygateway_types.go
```diff
@@ -503,11 +503,13 @@ type ExtensionManager struct {
 	// FailOpen defines if Envoy Gateway should ignore errors returned from the Extension Service hooks.
 	// The default is false, which means Envoy Gateway will fail closed if the Extension Service returns an error.
 	//
-	// Fail-close means that if the Extension Service hooks return an error, the relevant route/listener/resource
-	// will be replaced with a default configuration returning Internal Server Error (HTTP 500).
+	// Fail-close means that if the Extension Service hooks return an error, the error is propagated to the
```
arkodg, line 506: 1. should we rename fail-close to fail-safe ?
2. instead of saying,  error is propagated to the xds translator, can we say - the xDS update specific the envoy proxy fleet is skipped and the previous good state is preserved

internal/xds/translator/runner/runner.go
```diff
@@ -125,7 +126,11 @@ func (r *Runner) subscribeAndTranslate(sub <-chan watchable.Snapshot[string, *ir
 				result.EnvoyPatchPolicyStatuses = nil
 
 				// Publish
-				r.Xds.Store(key, result)
+				if err == nil {
+					r.Xds.Store(key, result)
+				} else {
+					r.Logger.Error(err, "xds resources are not published due to error in xds translator")
```
arkodg, line 132:
```suggestion
					r.Logger.Error(err, "skipped published xds resources")
```

release-notes/current.yaml
```diff
@@ -4,6 +4,9 @@ date: Pending
 breaking changes: |
   Use a dedicated listener port(19003) for envoy proxy readiness
   Uses the envoy JSON formatter for the default access log instead of text formatter.
+  Envoy Gateway would pause XDS snapshot updates in case of errors during XDS translation.
```
arkodg, line 7: prefer `skip` over `pause`

api/v1alpha1/envoygateway_types.go
guydc: > should we rename fail-close to fail-safe ?
fail safe can mean different things depending on the context. Maybe it's just better to not name the other failure state and just describe what it does. It's anyway not an enum in the API.

arkodg: Sure sgtm

arkodg (dismissed):
LGTM thanks !

api/v1alpha1/envoygateway_types.go
```diff
@@ -501,13 +501,15 @@ type ExtensionManager struct {
 	Service *ExtensionService `json:"service,omitempty"`
 
 	// FailOpen defines if Envoy Gateway should ignore errors returned from the Extension Service hooks.
-	// The default is false, which means Envoy Gateway will fail closed if the Extension Service returns an error.
 	//
-	// Fail-close means that if the Extension Service hooks return an error, the relevant route/listener/resource
-	// will be replaced with a default configuration returning Internal Server Error (HTTP 500).
+	// When set to false, Envoy Gateway does not ignore extension Service hook errors. As a result,
+	// xDS updates are skipped for the specific the envoy proxy fleet, the previous state is preserved.
```
zhaohuabing, line 506: typo ?
```suggestion
	// xDS updates are skipped for the envoy proxy fleet, the previous state is preserved.
```

internal/xds/translator/testdata/out/extension-xds-ir/http-route-extension-listener-error.listeners.yaml
```diff
@@ -21,18 +21,11 @@
...
-            - directResponse:
-                status: 500
-              match:
-                prefix: /
-              name: error_route
+        rds:
+          configSource:
+            ads: {}
+            resourceApiVersion: V3
+          routeConfigName: extension-post-xdslistener-hook-error
```
zhaohuabing, line 28: I don’t think we need this test, since the failed xDS wouldn’t be sent to the Envoy fleet anyway.
Would it make more sense to test the xDS output when `failOpen=true` instead?

internal/xds/translator/testdata/out/extension-xds-ir/http-route-extension-translate-error.listeners.yaml
```diff
@@ -21,18 +21,11 @@
...
-            - directResponse:
-                status: 500
-              match:
-                prefix: /
-              name: error_route
+        rds:
+          configSource:
+            ads: {}
+            resourceApiVersion: V3
+          routeConfigName: extension-post-xdstranslate-hook-error
```
zhaohuabing, line 28: Same here: https://github.com/envoyproxy/gateway/pull/5540/files#r2053717113

internal/xds/translator/testdata/out/extension-xds-ir/multiple-listeners-same-port-error.listeners.yaml
```diff
@@ -24,18 +24,11 @@
...
-            - directResponse:
-                status: 500
-              match:
-                prefix: /
-              name: error_route
+        rds:
+          configSource:
+            ads: {}
+            resourceApiVersion: V3
+          routeConfigName: first-listener-error
```
zhaohuabing, line 31: Same here: https://github.com/envoyproxy/gateway/pull/5540/files#r2053717113

nareddyt (approved):

internal/xds/translator/testdata/out/extension-xds-ir/http-route-extension-listener-error.listeners.yaml
guydc: Changed the original test to run in failOpen mode, and duplicated it to also assert errors are retruned in the fail closed option without checking translation results.

zhaohuabing (approved):
LGTM thanks!

PR #5543, opened 2025-03-19
Add SenseTime as an adopter of envoy gateway

ZhengXinwei-F:
**Add SenseTime as an adopter of envoy gateway**

Looking forward to more cooperation! 👋


shawnh2 (approved):
Thanks

Issue #5475, opened 2025-03-12
Support headers for responseOverride


jordangarside-kiwi:
It's common to want to do a redirect when auth fails, but that's currently not possible as `responseOverride` only supports rewriting the `statusCode` and `body`:
```
responseOverride: [
  {
    match: {
      statusCodes: [
        {
          type: 'Value',
          value: 401,
        },
        {
          type: 'Value',
          value: 403,
        },
      ],
    },
    response: {
      statusCode: 302,
      // not supported
      headers: [
        {
          name: 'Location',
          value: `https://${ssoHost}/oauth2/sign_in?rd={url}`,
        },
      ],
    },
  },
```

```
Kubernetes API server: failed to create typed patch object (oauth2-proxy/oauth2-proxy-response-override; gateway.envoyproxy.io/v1alpha1, Kind=BackendTrafficPolicy): .spec.responseOverride[0].response.headers: field not declared in schema
```

Same request:
https://github.com/envoyproxy/gateway/discussions/4254#discussion-7192030

Chever-John:
```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: auth-redirect-patch
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: your-http-route-name  # 替换为您的 HTTPRoute 名称
  patches:
    - type: JSONPatch
      value:
      - op: add
        path: /spec/virtualHosts/0/routes/0/typed_per_filter_config/envoy.filters.http.custom_response
        value:
          config:
            custom_response_matcher:
              matcher_list:
                matchers:
                - predicate:
                    single_predicate:
                      input:
                        name: response-code
                        typed_config:
                          '@type': type.googleapis.com/envoy.type.matcher.v3.HttpResponseStatusCodeMatchInput
                      value_match:
                        safe_regex:
                          google_re2: {}
                          regex: ^(401|403)$
                  on_match:
                    action:
                      name: redirect
                      typed_config:
                        '@type': type.googleapis.com/envoy.extensions.http.custom_response.redirect_policy.v3.RedirectPolicy
                        redirect_action:
                          host_redirect: "your-sso-host"
                          path_redirect: "/oauth2/sign_in"
                          scheme_redirect: "https"
                          strip_query: false
                          response_code: FOUND
                        status_code: 302
```

I tried to write a configuration, hopefully it helps!

[...]

PR #5549, opened 2025-03-19
api: support headers and redirects in custom response

rudrakhp:
**What type of PR is this?**

api: support headers and redirects in custom response

Extends the `CustomResponse` API to support adding response headers and redirect responses.

Related #5475


arkodg:
hey @rudrakhp sorry couldn't get to this sooner, since we are too close to v1.4, this will probably need to be included in v1.5, sorry about this

api/v1alpha1/shared_types.go
```diff
@@ -797,3 +800,57 @@ type Tracing struct {
...
+	// +optional
+	// +kubebuilder:validation:Enum=http;https
+	Scheme *string `json:"scheme,omitempty"`
+
+	// Hostname is the hostname to be used in the value of the `Location`
+	// header in the response.
+	// When empty, the hostname in the `Host` header of the request is used.
+	//
+	// +optional
+	Hostname *gwapiv1.PreciseHostname `json:"hostname,omitempty"`
```
rudrakhp, line 818: I am using predefined types from GW API, namely `Hostname` and `Port` to avoid duplications in validations. Is this a good idea though, didn't come across other instances of reuse, rather we seem to be defining our own, for example `HTTPPathModifier` which is also available in GW API.

arkodg: we'll need to use the upstream `HTTPPathModifier` since the one in this project is only tied to RegexReplace https://github.com/envoyproxy/gateway/blob/34a327b2c55ca72c10fb38285d4e200a6a4df6f9/api/v1alpha1/httproutefilter_types.go#L78

arkodg:
thanks for the updates @rudrakhp, apart from the one comment, LGTM from my side

zhaohuabing (approved):

PR #5483, opened 2025-03-12
Make configloader accept configured logOut

mathetake:
Follow up on #5476

PR #5476, opened 2025-03-12
Enables standalone library to configure log output

mathetake:
Release Notes: No

PR #5550, opened 2025-03-20
Removes inline os.Stdout as log output

mathetake:
**What type of PR is this?**

refactor: removes inline os.Stdout as log output

This removes the (almost) remaining places where the logger is configured to use os.Stdout instead of the configured output inherited from the cobra command. Without this change, the logging destination is scattered randomly from the downstream user's perspective. This is a follow up on #5476 #5483 and eliminates almost all places using os.Stdout except for the global loggers in a few places plus the stdout for the standalone mode's Envoy process. The implementation should be agnostic to any io.Writer in reality and there shouldn't be any necessity to directly use os.Stdout for all cases.

### DIFF
--- a/internal/cmd/certgen.go
+++ b/internal/cmd/certgen.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"path"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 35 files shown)


Xunzhuo:
Plz fix the CI failures @mathetake

mathetake:
yeah sorry done

mathetake:
great

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5557, opened 2025-03-20
fix: do not treat namespace as the missing resources

### LINKED ISSUE #5474: Weird inconsistent handling of namespace in Standalone mode
*Description*:

The existence of `namespace` in the resources passed to the standalone mode make the working resources not working only with the standalone mode. The log doesn't show any useful message, basically silencing any errors and it seems almost impossible to debug to reach the conclusion that namespace is the culprit.  The documentation in https://gateway.envoyproxy.io/docs/tasks/operations/standalone-deployment-mode/ not mentioning this behavior either. So, EG definitely should fix it by respecting namespaces or document this behavior as expected. 

*Repro steps*:

The following yam is not working (route not found) `curl localhost:8888/ --head` returns `404` due to route not found:

```yaml
---
kind: GatewayClass
metadata:
  name: aigw-run
spec:
---
kind: Gateway
metadata:
  name: aigw-run
  namespace: default
spec:
  gatewayClassName: aigw-run
  listeners:
    - name: http
      port: 8888
      protocol: HTTP
---
kind: Backend
metadata:
  name: openai
  namespace: default
spec:
  endpoints:
    - fqdn:
        hostname: api.openai.com
        port: 443
---
kind: HTTPRoute
metadata:
  name: aigw-run
  namespace: default
spec:
  parentRefs:
    - name: aigw-run
      namespace: default
  rules:
    - backendRefs:
        - group: gateway.envoyproxy.io
          kind: Backend
          name: openai
      matches:
        - path:
            value: /
```

On the other hand, the following yaml is working: the same curl gets 400 bad request from the OpenAI upstream:

```yaml
---
kind: GatewayClass
metadata:
  name: aigw-run
spec:
---
kind: Gateway
metadata:
  name: aigw-run
spec:
  gatewayClassName: aigw-run
  listeners:
    - name: http
      port: 8888
      protocol: HTTP
---
kind: Backend
metadata:
  name: openai
spec:
  endpoints:
    - fqdn:
        hostname: api.openai.com
        port: 443
---
kind: HTTPRoute
metadata:
  name: aigw-run
spec:
  parentRefs:
    - name: aigw-run
  rules:
    - backendRefs:
        - group: gateway.envoyproxy.io
          kind: Backend
          name: openai
      matches:
        - path:
            value: /
```

which basically only removes the namespace specifier from the resources.

shawnh2:
**What type of PR is this?**

Fixes #5474

Release Notes: Yes


internal/gatewayapi/resource/load.go
```diff
@@ -384,22 +385,21 @@ func loadKubernetesYAMLToResources(input []byte, addMissingResources bool) (*Res
...
-		sort.Strings(sortedRequiredNamespace)
-		for _, ns := range sortedRequiredNamespace {
-			if !providedNamespaceMap.Has(ns) {
-				namespace := &corev1.Namespace{
-					ObjectMeta: metav1.ObjectMeta{
-						Name: ns,
-					},
-				}
-				resources.Namespaces = append(resources.Namespaces, namespace)
+	// Sort the required namespace in order to keep the consistency for test.
```
zhaohuabing, line 388: Nit: a short comment here would help clarify why namespaces should always be added, even when `addMissingResources` is not set to true.

zhaohuabing (dismissed):

internal/gatewayapi/resource/load.go
shawnh2: Added

zhaohuabing (approved):

arkodg (approved):

PR #5527, opened 2025-03-17
Add Closer on Runner interface

mathetake:
**What type of PR is this?**

chore: add Close method interface on Runner interface

**What this PR does / why we need it**:

Previously, the runner "shudown" was not existent, so the main goroutine doesn't wait for the cleanup that is not synchronous to the context closure or channel. That includes virtually all the operations after the context cancelation running in non main goroutines. One notable example is that when running with the host provider, Envoy processes are spawned and managed from non-main goroutines, so without the "synchronous" closure hook on Runner, it is impossible to ensure the exit of Envoy processes. Without this fix, for example, standalone mode will leave all the Envoy processes as orphans after the CLI exits with SIGINT or SIGTERM. 

As a side effect, this also fixes a bug in the host provider which is that its CreateOrUpdateProxyInfra blocks forever until the Envoy process is running. That is not the expected semantics of infra.Manager interface.

Release Notes: No

PR #5560, opened 2025-03-20
Wait for the runners to close

mathetake:
This is a follow up on #5527 which was not sufficient to ensure Envoy processes to stop before existing the program. 

Release Notes:  No

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -62,12 +62,14 @@ func server(ctx context.Context, logOut io.Writer) error {
 		return err
 	}
 
+	runnersDone := make(chan struct{})
 	hook := func(c context.Context, cfg *config.Server) error {
-		cfg.Logger.Info("Setup runners")
-		if err := setupRunners(c, cfg); err != nil {
+		cfg.Logger.Info("Start runners")
+		if err := startRunners(c, cfg); err != nil {
 			cfg.Logger.Error(err, "failed to setup runners")
 			return err
 		}
+		runnersDone <- struct{}{}
 		return nil
 	}
 	l := loader.New(cfgPath, cfg, hook)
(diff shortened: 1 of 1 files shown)


internal/cmd/server.go
```diff
@@ -84,11 +86,14 @@ func server(ctx context.Context, logOut io.Writer) error {
 		return err
 	}
 
-	// Wait exit signal
+	// Wait for the context to be done, which usually happens the process receives a SIGTERM or SIGINT.
 	<-ctx.Done()
```
mathetake, line 90: note on why `<-runnersDone` is needed; Right after ctx is canceled, this main goroutine will call `os.Exit(0)` vs runners are not running on the main goroutine. so without blocking and waiting for it to return, this will highly likely leave some undesirable state left orphaned that is supposed to be cleaned up by non-main goroutine

arkodg (approved):
LGTM thanks !

internal/cmd/server.go
```diff
@@ -62,12 +62,14 @@ func server(ctx context.Context, logOut io.Writer) error {
 		return err
 	}
 
+	runnersDone := make(chan struct{})
 	hook := func(c context.Context, cfg *config.Server) error {
-		cfg.Logger.Info("Setup runners")
-		if err := setupRunners(c, cfg); err != nil {
+		cfg.Logger.Info("Start runners")
+		if err := startRunners(c, cfg); err != nil {
 			cfg.Logger.Error(err, "failed to setup runners")
```
shawnh2, line 69: failed to start runners

mathetake: done

arkodg (approved):

shawnh2 (approved):

Issue #5553, opened 2025-03-20
API documentation not up to date


aukevanleeuwen:
*Description*:
I've noticed that the documentation on https://gateway.envoyproxy.io/docs/api/extension_types/ is not (always) up to date. An example would be that https://gateway.envoyproxy.io/docs/tasks/observability/proxy-accesslog/ says that the attributes `disabled` is a required attribute. [This example however](https://gateway.envoyproxy.io/docs/tasks/observability/proxy-accesslog/#opentelemetry-sink) does not list the attribute. Also `kubectl explain envoyproxy` does not list the attributes as `- required -`:

```shell
❯ k explain envoyproxy --recursive | grep -A3 telemetry
    telemetry	<Object>
      accessLog	<Object>
        disable	<boolean>
        settings	<[]Object>
```

Another one would be `matches` in https://gateway.envoyproxy.io/docs/api/extension_types/#proxyaccesslogsetting.

On top of this it would be great if whenever `required` _is_ in fact `false`, the default value would be explicit as well, it's often not provided in the documentation.

PR #5564, opened 2025-03-21
api: access log disabled should be optional

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5553


arkodg (approved):

internal/gatewayapi/listener.go
```diff
@@ -266,7 +266,7 @@ func (t *Translator) processAccessLog(envoyproxy *egv1a1.EnvoyProxy, resources *
 	if envoyproxy == nil ||
 		envoyproxy.Spec.Telemetry == nil ||
 		envoyproxy.Spec.Telemetry.AccessLog == nil ||
-		(!envoyproxy.Spec.Telemetry.AccessLog.Disable && len(envoyproxy.Spec.Telemetry.AccessLog.Settings) == 0) {
+		(!ptr.Deref(envoyproxy.Spec.Telemetry.AccessLog.Disable, false) && len(envoyproxy.Spec.Telemetry.AccessLog.Settings) == 0) {
```
rudrakhp, line 269: Can we add/modify a gateway api test, where this is omitted from input testdata with no change in out?

zirain: sorry, I didn't get your point.

rudrakhp: @zirain We don't seem to have a test for this, can we have a test with `accessLog: {}`. Modifying an existing test with disabled access logs should work.
https://github.com/envoyproxy/gateway/blob/54efa343998cfcc0b8305721c94f04c9dc29610c/internal/gatewayapi/testdata/disable-accesslog.in.yaml#L9-L10

rudrakhp (approved):

arkodg (approved):

Issue #4380, opened 2024-10-01
Envoy Proxy deployment is not cleaned up when we customize the deployment name, replicas using EnvoyProxy resource


tejaboppana:
*Description*:
Envoy Proxy deployment is not cleaned up when we customize the deployment name, replicas etc using EnvoyProxy resource,

*Repro steps*:
1. Create a Gateway and use a [EnvoyProxy](https://gateway.envoyproxy.io/docs/api/extension_types/#envoyproxy) to set a custom name and increase the replicas. 
2. I tried deleting the Gateway and expected the envoy gateway to delete the envoy proxy deployment and service, but it did not. '
3. Next I created another Gateway with an EnvoyProxy resource. Then I deleted the EnvoyProxy, envoy gateway created a single replica deployment with the a default name but did not delete the older deployment. 
4. Now when I deleted the Gateway , the new deployment with default name and the service were deleted but the older deployment with custom name still exists.  

*Environment*:
Envoy Gateway version - 1.1.2 

*Logs*:
I could not see any logs in Envoy Gateway pods when I deleted the Gateway.

arkodg:
@tejaboppana thanks for flagging this
in step 1, did you create the `Gateway` resource first (a.) and then create a `EnvoyProxy` specifying the custom deployment name (b.) and then link the `EnvoyProxy` resource to the `Gateway` by editing it (c.) ?

For this case the I think the infra layer will create a new deployment with the new name but not delete the older deployment
When a custom name is set, we also need to delete the older deployment
https://github.com/envoyproxy/gateway/blob/0f4cb27f0110051f7811122a2dd73652c93a15b2/internal/infrastructure/kubernetes/proxy/resource_provider.go#L153

tejaboppana:
@arkodg I tried 2 things and below are my observations 

1. Steps that you mentioned i.e a) Create `Gateway` b) Create `EnvoyProxy` c) Edit `Gateway `to link it with `EnvoyProxy`
 In this case a new deployment is created but the older deployment is not deleted which is not ideal. In addition to that,  when I delete the `Gateway` and `EnvoyProxy`, I expected everything to be cleaned up. Only the older deployment and the service are deleted but the deployment with custom name is not deleted. 

2. a) Create `EnvoyProxy` b) Create `Gateway` and link it to the `EnvoyProxy`
[...]

PR #5569, opened 2025-03-21
infra: delete resource if exists when using customized name

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/4380

### DIFF
--- a/internal/infrastructure/kubernetes/infra.go
+++ b/internal/infrastructure/kubernetes/infra.go
@@ -20,6 +20,7 @@ import (
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 	"github.com/envoyproxy/gateway/internal/infrastructure/kubernetes/proxy"
 	"github.com/envoyproxy/gateway/internal/infrastructure/kubernetes/ratelimit"
+	"github.com/envoyproxy/gateway/internal/logging"
 )
 
 var _ ResourceRender = &proxy.ResourceRender{}
@@ -58,6 +59,8 @@ type Infra struct {
 
 	// Client wrap k8s client.
 	Client *InfraClient
+
+	logger logging.Logger
 }
 
 // NewInfra returns a new Infra.
@@ -67,6 +70,7 @@ func NewInfra(cli client.Client, cfg *config.Server) *Infra {
 		DNSDomain:    cfg.DNSDomain,
 		EnvoyGateway: cfg.EnvoyGateway,
 		Client:       New(cli),
+		logger:       cfg.Logger.WithName("infra-kubernetes"),
 	}
 }
 
--- a/internal/infrastructure/kubernetes/infra.go
+++ b/internal/infrastructure/kubernetes/infra.go
@@ -70,7 +70,7 @@ func NewInfra(cli client.Client, cfg *config.Server) *Infra {
 		DNSDomain:    cfg.DNSDomain,
 		EnvoyGateway: cfg.EnvoyGateway,
 		Client:       New(cli),
-		logger:       cfg.Logger.WithName("infra-kubernetes"),
+		logger:       cfg.Logger.WithName(string(egv1a1.LogComponentInfrastructureRunner)),
 	}
 }
 
--- a/internal/infrastructure/kubernetes/infra_client.go
+++ b/internal/infrastructure/kubernetes/infra_client.go
@@ -33,6 +33,15 @@ func (cli *InfraClient) ServerSideApply(ctx context.Context, obj client.Object) 
 	return nil
 }
 
+func (cli *InfraClient) DeleteIfExists(ctx context.Context, object client.Object) error {
+	if err := cli.Delete(ctx, object); err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil
+		}
+	}
+	return nil
+}
+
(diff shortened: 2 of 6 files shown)


internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -119,6 +119,16 @@ func (i *Infra) createOrUpdateDeployment(ctx context.Context, r ResourceRender)
 	}
 
 	defer func() {
+		// custom Deployment name is enabled, and the deployment name is different from the resource name.
```
arkodg, line 122: this takes into account when default name changes to custom name, what about the case when custom name is removed and we switch to default name ?
my suggestion would be instead do a `Client.List` and delete all that are not current

zirain: we should use the one which not the currnet one.

zirain: good catch!

zirain:
@arkodg should we highlight that EG does not guarantee with zero downtime when using customized name?

test/e2e/tests/envoyproxy.go
```diff
@@ -0,0 +1,315 @@
...
+		if dp.Status.ReadyReplicas <= 0 {
+			tlog.Logf(t, "Expected Deployment %s ready", dp.Name)
+		}
+
+		tlog.Logf(t, "Check Envoy proxy deployment pass, name: %s", dp.Name)
+		return true, nil
+	})
+}
+
+func checkEnvoyProxyService(t *testing.T, suite *suite.ConformanceTestSuite, gwNN types.NamespacedName, exceptName string) error {
```
shawnh2, line 249: can move these helper functions into utils.go?

test/e2e/tests/envoyproxy.go
```diff
@@ -0,0 +1,315 @@
...
+			err = suite.Client.Update(context.Background(), gw)
+			if err != nil {
+				t.Fatalf("Failed to update Gateway: %v", err)
+			}
+
+			err = checkEnvoyProxyDeployment(t, suite, gwNN, "deploy-custom-name")
+			if err != nil {
+				t.Fatalf("Failed to delete Gateway: %v", err)
+			}
+			err = checkEnvoyProxyService(t, suite, gwNN, "deploy-custom-name")
```
shawnh2, line 88: should we check the old deployments and services exist?

test/e2e/tests/envoyproxy.go
```diff
@@ -0,0 +1,315 @@
...
+				t.Fatalf("Failed to get Gateway: %v", err)
+			}
+			gw.Spec.Infrastructure = &gwapiv1.GatewayInfrastructure{}
+			err = suite.Client.Update(context.Background(), gw)
+			if err != nil {
+				t.Fatalf("Failed to update Gateway: %v", err)
+			}
+
+			// Make sure there's deployment for the gateway
+			err = checkEnvoyProxyDeployment(t, suite, gwNN, fmt.Sprintf("envoy-%s-%s", gwNN.Namespace, gwNN.Name))
```
shawnh2, line 109: ditto

test/e2e/tests/envoyproxy.go
zirain: maybe in the future, right now these functions ensure only 1 items exists.

test/e2e/tests/envoyproxy.go
zirain: we ensure there's only 1 item, see L229.

shawnh2 (dismissed):

internal/infrastructure/kubernetes/infra.go
```diff
@@ -67,6 +70,7 @@ func NewInfra(cli client.Client, cfg *config.Server) *Infra {
 		DNSDomain:    cfg.DNSDomain,
 		EnvoyGateway: cfg.EnvoyGateway,
 		Client:       New(cli),
+		logger:       cfg.Logger.WithName("infra-kubernetes"),
```
arkodg, line 73: won't be have `infrastructure` twice https://github.com/envoyproxy/gateway/blob/55fc01fecf2819af479aa1a3e5f4af37b737a947/api/v1alpha1/envoygateway_types.go#L148 ? should this be `kubernetes` instead ?

internal/infrastructure/kubernetes/infra_client.go
```diff
@@ -33,6 +34,41 @@ func (cli *InfraClient) ServerSideApply(ctx context.Context, obj client.Object)
 	return nil
 }
 
+// DeleteAllExcept delete all resources filter by ListOption except the one specified by key.
+func (cli *InfraClient) DeleteAllExcept(ctx context.Context, objList client.ObjectList, key client.ObjectKey, listOpts ...client.ListOption) error {
```
arkodg, line 38: nice !

internal/infrastructure/kubernetes/infra.go
zirain: fixed

arkodg (dismissed):
LGTM thanks !

arkodg (approved):

cnvergence (approved):

PR #5460, opened 2025-03-10
load EnvoyExtensionPolicy in standalone mode

mathetake:
Similar to https://github.com/envoyproxy/gateway/pull/5431

PR #5575, opened 2025-03-21
Secret validation error in TLS termination mode with Envoy Gateway Standalone 

arminabf:
**What type of PR is this?**

Fix: Fix secret validation error in TLS termination mode with Envoy Gateway Standalone.

Envoy Gateway Standalone fails to validate TLS secrets for HTTPS listeners configured with TLS termination. The validation logic at https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/validate.go#L501-L510 expects the type field to be set to _corev1.SecretTypeTLS_, but the type is not populated when loading KindSecret resources from the Gateway configuration.

*Example configuration*

```
---
kind: GatewayClass
metadata:
  name: eg
spec:
---
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: https
      protocol: HTTPS
      port: 11443
      tls:
        mode: Terminate
        certificateRefs:
          - name: gateway-cert-secret
            kind: Secret
---
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "eg-test.local"
  rules:
    - backendRefs:
        - group: "gateway.envoyproxy.io"
          kind: Backend
          name: backend
      matches:
        - path:
            type: PathPrefix
            value: /
---
kind: Backend
metadata:
  name: backend
spec:
  endpoints:
    - fqdn:
        hostname: httpbin.local
        port: 80
---
kind: Secret
apiVersion: v1
metadata:
  name: gateway-cert-secret
type: kubernetes.io/tls
data:
  tls.crt:  <ommited>
  tls.key: <ommited>
```

Release Notes: Yes


arkodg:
thanks, this is great !
1. can you sign your commit and force push to fix DCO 
2. can you add a test file similar to the one in https://github.com/envoyproxy/gateway/pull/5460

arkodg (approved):

internal/gatewayapi/resource/load.go
```diff
@@ -315,6 +315,9 @@ func loadKubernetesYAMLToResources(input []byte, addMissingResources bool) (*Res
 		case KindSecret:
 			typedData := data.Interface()
 			typedStringData := stringData.Interface()
+			secretType := kobjVal.FieldByName("Type")
+			typedSecretType := secretType.Interface()
```
rudrakhp, line 319: I think we need to check if value [`IsValid()`](https://pkg.go.dev/reflect#Value.IsValid) before converting to an interface to avoid panics. Good to have a test with empty type as well.
```
typedSecretType := "Opaque" // default secret type
secretType := kobjVal.FieldByName("Type")
if secretType.IsValid() {
	typedSecretType := secretType.Interface()
}
```

internal/gatewayapi/resource/testdata/all-resources.in.yaml
```diff
@@ -257,6 +257,15 @@ stringData:
   secret: "literal value"
 ---
 apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-with-type
+  namespace: default
+type: "type value"
```
rudrakhp, line 264: We can try omitting this field in another testcase to test https://github.com/envoyproxy/gateway/pull/5575#discussion_r2011296350

internal/gatewayapi/resource/load.go
arminabf: Thanks for the feedback!

`Good to have a test with empty type as well.`
Well, every already existing test case  has no secret type and the tests run without issue.

`I think we need to check if value IsValid() before converting to an interface to avoid panics`
I have changed the code according your advise... still, the test cases with no secret type do not show up the "Opaque" default type. For this reason I come to the conclusion that the valid check is not needed: as being ensured that the kind is "KindSecret" it is always given that "type" is an existing field that can be accessed. Otherwise, accessing other fields (like data or stringData) wouldn't be safe either.

rudrakhp: Ok I see the default might be an empty string, which is still a valid value. In that case can we could just do:
```suggestion
			secretType := kobjVal.FieldByName("Type").Interface()
```
But good to have the validation IMO. Either works for me.

arminabf: Variable name fixed, see ec05f63d008dc9980634b34f3c27090d11f61ec0. Thanks!

rudrakhp (approved):

arkodg (approved):

Issue #5571, opened 2025-03-21
Unable to specify `samplingFraction` in `EnvoyProxy#telemetry.tracing`


aukevanleeuwen:
*Description*:
I'm unable to specify a `samplingFraction` in `EnvoyProxy`. 

*Repro steps*:
```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: foo
spec:
  telemetry:
    tracing:
      provider:
        backendRefs:
          - group: ''
            kind: Service
            name: "otel-collector"
            namespace: "monitoring"
            port: 4317
        type: OpenTelemetry
      samplingFraction:
        numerator: 100
        denominator: 100
```

```shell
❯ k -n default apply -f envoy-proxy.yaml
The EnvoyProxy "foo" is invalid: spec.telemetry.tracing: Invalid value: "object": only one of SamplingRate or SamplingFraction can be specified
```

I'm hitting this validation: https://github.com/envoyproxy/gateway/blob/54efa343998cfcc0b8305721c94f04c9dc29610c/api/v1alpha1/tracing_types.go#L11

Due to this default:
https://github.com/envoyproxy/gateway/blob/54efa343998cfcc0b8305721c94f04c9dc29610c/api/v1alpha1/tracing_types.go#L22

So basically `samplingRate` defaults to `100`, which means that I can never specify a `samplingFraction`. 

When I update the custom resource definition and manually comment out the `default: 100` on `samplingRate`, I'm able to apply the manifest.

arkodg:
thanks for surfacing this, I didn't realize CEL rules apply after defaulting
another option is to delete the explicit default tag and set it in the gateway-api layer

Issue #5486, opened 2025-03-12
Support Per-Route Sampling Rates for Trace Generation


json-openai:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Currently, Envoy Gateway supports configuring a global sampling rate for trace generation, as documented in the [official observability guide](https://gateway.envoyproxy.io/latest/tasks/observability/proxy-trace/#sampling-rate). This rate applies to all routes uniformly, without the ability to configure different sampling behaviors for individual routes.

However, Envoy itself supports per-route random sampling for trace generation, as outlined in the [Envoy route configuration API](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-tracing). This feature enables more targeted observability by allowing different sampling rates for different routes.

In Kubernetes environments, Envoy Gateway is often shared across multiple services and teams. A single global sampling rate may not meet the diverse observability needs of each team. Some services may require high sampling rates for critical paths, while others may prefer minimal tracing to reduce overhead and cost.

Proposal:
Enable configuration of per-route sampling rates in Envoy Gateway to align with what Envoy natively supports. This would give teams the autonomy to tune observability per service or route, based on their own needs and priorities.

Would love to hear your thoughts on whether this aligns with the roadmap, and happy to help with a design or implementation plan if helpful.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.
- [Envoy Gateway Proxy Tracing Documentation](https://gateway.envoyproxy.io/latest/tasks/observability/proxy-trace/)
- [Envoy HTTP Route Configuration Documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-tracing)

arkodg:
@zirain had warned me earlier that this would be needed in the future 😄 

From https://gateway.envoyproxy.io/docs/api/extension_types/#proxytelemetry - accesLog and tracing are both applicable at a per route level

https://gateway.envoyproxy.io/docs/api/extension_types/#backendtrafficpolicy and https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilter apply at the route level, but neither feel ideal

hoping others with similar use cases can chime in here as well

arkodg:
linking some prior art
* Istio has a first class API - Telemetry https://istio.io/latest/docs/tasks/observability/telemetry/

zirain:
[...]

PR #5578, opened 2025-03-22
fix tracing SamplingFraction not working

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5571

### DIFF
--- a/api/v1alpha1/tracing_types.go
+++ b/api/v1alpha1/tracing_types.go
@@ -19,7 +19,6 @@ type ProxyTracing struct {
 	//
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=100
-	// +kubebuilder:default=100
 	// +optional
 	SamplingRate *uint32 `json:"samplingRate,omitempty"`
 	// SamplingFraction represents the fraction of requests that should be
--- a/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoyproxies.yaml
+++ b/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoyproxies.yaml
@@ -14633,7 +14633,6 @@ spec:
                         - message: numerator must be less than or equal to denominator
                           rule: self.numerator <= self.denominator
                       samplingRate:
-                        default: 100
                         description: |-
                           SamplingRate controls the rate at which traffic will be
                           selected for tracing if no prior sampling decision has been made.
--- a/internal/gatewayapi/testdata/tracing-sampling-fraction.in.yaml
+++ b/internal/gatewayapi/testdata/tracing-sampling-fraction.in.yaml
@@ -1,0 +1,90 @@
+envoyProxyForGatewayClass:
+  apiVersion: gateway.envoyproxy.io/v1alpha1
+  kind: EnvoyProxy
+  metadata:
+    namespace: envoy-gateway
+    name: test
+  spec:
+    mergeGateways: true
+    telemetry:
+      tracing:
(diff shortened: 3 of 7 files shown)


api/v1alpha1/tracing_types.go
```diff
@@ -25,8 +23,7 @@ type ProxyTracing struct {
 	// SamplingFraction represents the fraction of requests that should be
 	// selected for tracing if no prior sampling decision has been made.
 	//
-	// Only one of SamplingRate or SamplingFraction may be specified.
-	// If neither field is specified, all requests will be sampled.
+	// if both SamplingRate and SamplingFraction are set, SamplingRate will be ignored.
```
aukevanleeuwen, line 26:
```suggestion
	// If both SamplingRate and SamplingFraction are set, SamplingRate will be ignored.
```

internal/gatewayapi/listener.go
```diff
@@ -510,22 +510,25 @@ func (t *Translator) processTracing(gw *gwapiv1.Gateway, envoyproxy *egv1a1.Envo
...
 		if tracing.SamplingFraction.Denominator != nil {
 			denominator = float64(*tracing.SamplingFraction.Denominator)
 		}
 
-		rate = numerator / denominator
+		rate := numerator / denominator
 		// Identifies a percentage, in the range [0.0, 100.0]
 		rate = math.Max(0, rate)
 		rate = math.Min(100, rate)
+		return rate
```
aukevanleeuwen, line 524: Out of curiousity, this percentage is indeed specified as `[0.0, 100.0]` here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/v3/percent.proto#envoy-v3-api-msg-type-v3-percent

Does this really mean that the values such as `0.05` would not be allowed? I don't see any code here that does something to that effect. Hence the question. Also: all of the test cases show only really trivial cases such as 1/100 or 100/100 or something.

zirain: I think https://github.com/envoyproxy/gateway/issues/5486 will help on this.

api/v1alpha1/tracing_types.go
```diff
@@ -8,14 +8,12 @@ package v1alpha1
...
 // ProxyTracing defines the tracing configuration for a proxy.
-// +kubebuilder:validation:XValidation:message="only one of SamplingRate or SamplingFraction can be specified",rule="!(has(self.samplingRate) && has(self.samplingFraction))"
 type ProxyTracing struct {
 	// SamplingRate controls the rate at which traffic will be
 	// selected for tracing if no prior sampling decision has been made.
 	// Defaults to 100, valid values [0-100]. 100 indicates 100% sampling.
 	//
-	// Only one of SamplingRate or SamplingFraction may be specified.
-	// If neither field is specified, all requests will be sampled.
+	// if both SamplingRate and SamplingFraction are set, SamplingRate will be ignored.
```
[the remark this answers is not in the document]
zirain: `api/v1alpha1/tracing_types.go` is the source, others are generated from it.

internal/gatewayapi/listener.go
zirain: `0.05` is valid

aukevanleeuwen: 👍

arkodg:
my preference would be to remove the kube-builder tag,.so a user is alerted during apply time

aukevanleeuwen (approved):

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

PR #5582, opened 2025-03-23
api: MergeType in BackendTrafficPolicy

arkodg:
* Allows a user to merge a BackendTrafficPolicy targeting a child resourcce (xRoute) with a BackendTrafficPolicy targeting the parent resource (Gateway/Listener) using `JSONMerge` or `StrategicMerge`

Relates to https://github.com/envoyproxy/gateway/issues/1934


api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -46,6 +46,14 @@ type BackendTrafficPolicySpec struct {
 	PolicyTargetReferences `json:",inline"`
 	ClusterSettings        `json:",inline"`
 
+	// PatchType determines how this configuration is merged with existing BackendTrafficPolicy
+	// configurations targeting a parent resource. When set, this configuration will be merged
+	// into a parent BackendTrafficPolicy (i.e. the one targeting a Gateway or Listener).
+	// This field cannot be set when targeting a parent resource (Gateway).
+	// If unset, no patching occurs, and only the most specific configuration takes effect.
+	// +optional
+	PatchType *MergeType `json:"patchType,omitempty"`
```
arkodg, line 55: other names to consider
* `type`
* `mergeType`

ryanhristovski: +1 for `MergeType`

ryanhristovski (approved):
Looking forward to this feature!

api/v1alpha1/backendtrafficpolicy_types.go
zirain: MergeType feels better than PatchType here.

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -46,6 +46,14 @@ type BackendTrafficPolicySpec struct {
 	PolicyTargetReferences `json:",inline"`
 	ClusterSettings        `json:",inline"`
 
+	// PatchType determines how this configuration is merged with existing BackendTrafficPolicy
+	// configurations targeting a parent resource. When set, this configuration will be merged
+	// into a parent BackendTrafficPolicy (i.e. the one targeting a Gateway or Listener).
```
nareddyt, line 51: Thanks for putting together this API, I do think it's useful to support merging parent-level `BackendTrafficPolicy` with into the route-level ones.

However, I have concerns about RBAC with this merge policy. Specifically, my concern is that the field is set on the route-level resource instead of the parent-level resource.

In my team, the personas for `BackendTrafficPolicy` are as follows:
- Parent-level BTP attached to Gateway is configured by the _cluster operator_
- Route-level BTP attached to HTTPRoute/GRPCRoute is configured by _application developer_

With the current implementation, the _application developer_ can set the `MergeType` to be `JSONMerge`, giving them full control to **override** any lists set by the _cluster operator_.

Wouldn't it make more sense to define this `MergeType` at the parent resource level? That way, the _cluster operator_ has control over whether _application developers_ are allowed to `StrategicMerge` or `JSONMerge` lists, essentially preventing them from overriding some default policies.

arkodg: if the application developers have write access to the BTP resource, there are already overriding the BTP applied at the Gateway resource, since by default its a `Replace` . This API build on top of it by supporting a `Merge`.

`override` from the parent (Gateway) to child (Route) is a separate use case imo.

nezdolik: agree with @arkodg that ability to override form parent (gateway) to child (route) is a different use case. There are numerous configuration use cases when gateway level config needs to be overridden by route level config and it also aligns how envoy allows to specify filter level config and to override it with route level config for variety of filters. For a specific case where such override should not be allowed, one could utilise for example OPA and validate+reject route level btp if it does not satisfy the criteria.

nezdolik:
Docker is very interested in this feature and looking forward when it becomes available :)

api/v1alpha1/backendtrafficpolicy_types.go
zhaohuabing: +1 for `MergeType`

zhaohuabing:
Not directly API-related -- It would be hard for users to figure out what's the effective configuration if merging is used. To improve usability,  we should surface the merged spec in the status.

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: @nareddyt creating a new issue to track this

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: https://github.com/envoyproxy/gateway/issues/5746

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: https://github.com/envoyproxy/gateway/issues/5746

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: renamed to mergeType

arkodg:
cc @guicassolato

zhaohuabing (approved):

api/v1alpha1/backendtrafficpolicy_types.go
guicassolato: First and foremost, I’m glad to see more use of [Inherited Policies](https://gateway-api.sigs.k8s.io/geps/gep-2649/). Well done!

I thought regarding the API chosen…

In https://github.com/kubernetes-sigs/gateway-api/pull/3609, we’re proposing that the policy that is attached to parent dictates how the merge happens, and not the one attached to the child.

There, “merge policy” is called “merge strategy”, and there are a few ones specified, aiming to support the different use cases mentioned in the comments, including:
- **Atomic defaults:** the child policy fully (“atomically”) replaces the parent one
- **Merge defaults:** the child policy is merged into the parent one
- **Atomic overrides:** the parent policy fully replaces (supersedes) the child one
- **Merge overrides:** the parent policy merges into the child one, occasionally superseding only parts of it

The API I see proposed here is different. It pushes the responsibility of defining the merge strategy (“merge policy”) to the child. I personally find this approach hard to scale to support the _override_ strategies later on, without requiring another fine-grained validation method added on top.

I wonder how the API would look like in case a BTP resource would allow targeting yet a third level (e.g. Service or GatewayClass) and with both parent defaults and parent overrides supported. 🤔

cnvergence (approved):

Issue #5602, opened 2025-03-25
Flaky TestPermissionCache test


shahar-h:
*Description*:
```log
--- FAIL: TestPermissionCache (0.15s)
    --- FAIL: TestPermissionCache/Cached_permission_should_be_removed_after_expiry (0.02s)
        premissioncache_test.go:175: 
            	Error Trace:	/home/runner/work/gateway/gateway/internal/wasm/premissioncache_test.go:175
            	Error:      	Should be true
            	Test:       	TestPermissionCache/Cached_permission_should_be_removed_after_expiry
            	Messages:   	expired entry should be added after recheck
FAIL
```

https://github.com/envoyproxy/gateway/actions/runs/14049621057/job/39337429702
https://github.com/envoyproxy/gateway/actions/runs/14044002059/job/39320497190
https://github.com/envoyproxy/gateway/actions/runs/14027635760/job/39269018396
https://github.com/envoyproxy/gateway/actions/runs/14000738609/job/39206364526

PR #5603, opened 2025-03-25
fix: flaky wasm permission check tests

zhaohuabing:
fix: https://github.com/envoyproxy/gateway/issues/5602

### DIFF
--- a/internal/wasm/premissioncache.go
+++ b/internal/wasm/premissioncache.go
@@ -71,6 +71,9 @@ type permissionCache struct {
 
 	cache  map[string]*permissionCacheEntry
 	logger logging.Logger
+
+	// For tests
+	triggerChan chan struct{}
 }
 
 // permissionCacheEntry is an entry in the permission cache.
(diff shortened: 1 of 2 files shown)


internal/wasm/premissioncache.go
```diff
@@ -125,6 +128,7 @@ func newPermissionCache(options permissionCacheOptions, logger logging.Logger) *
 		cache:                  make(map[string]*permissionCacheEntry),
 		permissionCacheOptions: options,
 		logger:                 logger,
+		triggerChan:            make(chan struct{}),
```
arkodg, line 131: can this be avoided by using a really small checkInterval in tests ?

zhaohuabing: Probably, but no guarantee as it's inherently flaky.

zirain: can we use `waitUntil` in test?

zhaohuabing: > can we use waitUntil in test?
Unlike HTTP Get, these tests are stateful, so `waitUntil` would be tricky.

zhaohuabing: @arkodg you prefer adjusting the checkInterval ?

arkodg: yes prefer if we didn't have to add another field which is relevant only for tests, if possible

arkodg (approved):

rudrakhp (approved):

Issue #5351, opened 2025-02-26
Support HTTPRouteFilter as a Backend Filter


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Gateway API supports specifying filters per backendRefs https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPBackendRef which allow filter definitions to kick in after backend selection

Envoy Gateway has an extended filter called [HTTPRouteFilter](https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilter) . This will soon support a credentialInjection option https://github.com/envoyproxy/gateway/pull/5312.

If this filter can be applied per BackendRef we can enable some more use cases like injecting different API Keys based on the backendRef

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
cc @mathetake @missBerg @yuzisun

zhaohuabing:
@arkodg It looks like EG can't support Backend Filter. Envoy supports upstream filter per cluster, but EG creates Cluster per RouteRule, not per Backend.
To support Backend Filter, EG needs to create Cluster per Backend.

arkodg:
should be possible https://github.com/envoyproxy/gateway/blob/ce8b6bd56c87376f97986067fbb9632825568c60/internal/xds/translator/translator.go#L589

zhaohuabing:
So if this is enabled, then we create Cluster per backend?

zhaohuabing:
Docs are already available on this page: https://gateway.envoyproxy.io/latest/tasks/security/credential-injection/#injecting-credentials-into-backendref

PR #5607, opened 2025-03-25
feat: support CredentialInjector as a Backend Filter 

zhaohuabing:
This PR adds support for using the EG HTTPRouteFilter as a Backend Filter. Currently, only the credential injector is supported, which allows configuring different credentials for individual backends within an HTTPRoute.

Implement: https://github.com/envoyproxy/gateway/issues/5351
Release Notes: Yes


internal/xds/translator/testdata/out/xds-ir/credential-injection-backend-filter.routes.yaml
```diff
@@ -0,0 +1,40 @@
...
+            resources:
+            - kind: HTTPRoute
+              name: httproute-1
+              namespace: default
+      name: httproute/default/httproute-1/rule/1/match/0/*
+      route:
+        clusterNotFoundResponseCode: INTERNAL_SERVER_ERROR
+        upgradeConfigs:
+        - upgradeType: websocket
+        weightedClusters:
```
arkodg, line 30: this is generating `weightedClusters` logic which is not required
https://github.com/envoyproxy/gateway/blob/6863a17a99eec2a46623fd8b275a2dae186223db/internal/xds/translator/route.go#L724 needs to be updated

internal/ir/xds.go
```diff
@@ -2996,6 +2996,8 @@ type DestinationFilters struct {
 	AddResponseHeaders []AddHeader `json:"addResponseHeaders,omitempty" yaml:"addResponseHeaders,omitempty"`
 	// RemoveResponseHeaders defines a list of headers to be removed from response.
 	RemoveResponseHeaders []string `json:"removeResponseHeaders,omitempty" yaml:"removeResponseHeaders,omitempty"`
+	// CredentialInjection defines the credential injection configuration.
+	CredentialInjection *CredentialInjection `json:"credentialInjection,omitempty" yaml:"credentialInjection,omitempty"`
```
arkodg, line 3000: should the same field at Route level be deleted, and should we push down the IR to this level ? to reuse/consolidate code ?

internal/xds/translator/testdata/out/xds-ir/credential-injection-backend-filter.routes.yaml
arkodg: this is needed, and is working as expected, my bad

internal/ir/xds.go
arkodg: yeah, but I guess the downside is it will create a weighted backend, so lets stick with this

arkodg (approved):
LGTM thanks !

zirain (commented):
LGTM, should we change the PR title?

zhaohuabing:
> LGTM, should we change the PR title?

Done

zirain (approved):

PR #5611, opened 2025-03-25
fix: use rds instead of inline route config in fail closed mode

### LINKED ISSUE #5595: Partial config recovery after extension server failclose
*Description*:
Currently, when the extension server fail closed mode is used on Listener level, the HCM RDS reference is replaced with an inline `error_vhost` route config. When EG recovers and creates a new snapshot, proxies will request all HTTPRouteConfigs, as the HCM RDS references are returned to the listener. However, in some cases, the XDS snapshot will only return resources that have **actually** changed, causing all other RDS fetch to timeout. 

*Repro steps*:
- Setup a TLS listener with two http routes on distinct domains
- Create an extension server that returns Service Unavailable error on demand for the PostTranslate hook
- Trigger translation by changing http route 1 with extension server failure: note that all HCMs use an `error_vhost`
- Trigger translation by changing http route 1, and without an extension server failure
- Observe that only http route 1 recovers, while traffic to http route 2 returns 404.

The root cause may be the fact that routes are not really removed from the snapshot, only the reference to them is removed. And so, the cache doesn't identify a "delta" that needs to be synced to proxies. 

To solve this issue, we can change the route clearing logic to update the existing RDS to use an `error_vhost`, rather than creating additional HCM-inline route configs. 

*Environment*:
EG main

*Logs*:

After recovery, proxy requests both routes: `[default/eg/https default/eg/https-foo]`

```sh
2025-03-24T22:48:01.003Z	DEBUG	xds-server	cache/snapshotcache.go:336	handling v3 xDS resource request, response_nonce , nodeID envoy-default-eg-e41e7b31-69b7fb5bb6-hnm6j, node_version v1.34.0, resource_names_subscribe [default/eg/https default/eg/https-foo], resource_names_unsubscribe [], type_url type.googleapis.com/envoy.config.route.v3.RouteConfiguration, errorCode 0, errorMessage 

```

Cache responds only with the changed resource: `[default/eg/https-foo]`

```sh
2025-03-24T22:48:01.004Z	DEBUG	xds-server	v3/simple.go:594	node: envoy-default-eg-e41e7b31-69b7fb5bb6-hnm6j, sending delta response for typeURL type.googleapis.com/envoy.config.route.v3.RouteConfiguration with resources: [default/eg/https-foo] removed resources: [] with wildcard: false
```

Envoy times-out on initial fetch 15 seconds later (default fetch timeout):

```sh
[2025-03-24 22:48:15.993][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:130] gRPC config: initial fetch timed out for type.googleapis.com/envoy.config.route.v3.RouteConfiguration
```

guydc:
**What type of PR is this?**

Fixes #5595

Release Notes: Yes

### DIFF
--- a/internal/xds/translator/testdata/out/extension-xds-ir/http-route-extension-listener-error.listeners.yaml
+++ b/internal/xds/translator/testdata/out/extension-xds-ir/http-route-extension-listener-error.listeners.yaml
@@ -21,18 +21,11 @@
         mergeSlashes: true
         normalizePath: true
         pathWithEscapedSlashesAction: UNESCAPE_AND_REDIRECT
-        routeConfig:
-          name: error_route_configuration
-          virtualHosts:
-          - domains:
-            - '*'
-            name: error_vhost
-            routes:
-            - directResponse:
-                status: 500
-              match:
-                prefix: /
-              name: error_route
+        rds:
+          configSource:
+            ads: {}
+            resourceApiVersion: V3
+          routeConfigName: extension-post-xdslistener-hook-error
         serverHeaderTransformation: PASS_THROUGH
         statPrefix: http-10080
         useRemoteAddress: true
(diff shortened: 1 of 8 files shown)


guydc:
closing, as the behavior that replaces routes is being removed in #5540

Issue #5396, opened 2025-03-03
Extension Manager: support CA certificate reload


guydc:
*Description*:
Currently, extension manager loads the CA certificate when establishing GRPC connections:

https://github.com/envoyproxy/gateway/blob/5b6a35f3e6b6b31ad7479a9ceb5159b77cd68f7a/internal/extension/registry/extension_manager.go#L271

Connections are then cached and reused:

https://github.com/envoyproxy/gateway/blob/5b6a35f3e6b6b31ad7479a9ceb5159b77cd68f7a/internal/extension/registry/extension_manager.go#L219

At this time, golang doesn't provide a convenient way to reload CA certificates: https://github.com/golang/go/issues/64796. However, grpc-go's `advancedtls` library does support working with CA cert providers/callbacks: https://pkg.go.dev/google.golang.org/grpc/security/advancedtls#RootCertificateOptions. 

Envoy Gateway can implement dynamic reload of the extension manager CA certificate by:

- Using grpc-go's advancedtls
- Implementing a provider/callback that:
  - reads the CA cert from controller-runtime cache of k8s secret
  - reads the CA cert from a cache watching a mount of the K8s secret in Envoy Gateway's pod 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #5613, opened 2025-03-25
feat(extension manager): dynamically reload CA certificates

sapirpol:
**What type of PR is this?**
Dynamically reload CA certificates for GRPC connections on extension-manager.

This PR introduces dynamic reloading of CA certificates used by Envoy Gateway's extension manager when establishing GRPC connections. Previously, certificates were loaded once, making updates difficult without restarting the service. To resolve this, the PR uses grpc-go's advancedtls library, enabling dynamic certificate management via a custom callback that reloads the certificate.

**Comparison of TLS Settings: Before vs. After Update**
Setting | Before (Old Approach) | After (New Approach)
-- | -- | --
Min TLS Version | TLS 1.2 | TLS 1.2
Max TLS Version | Highest available | Highest available
Ciphers | Used safe default cipher list with explicitly forbidden RFC 7540 ciphers [grpc-go](https://github.com/grpc/grpc-go/blob/master/credentials/tls.go#L192). | Uses safe default cipher list [grpc-go](https://github.com/grpc/grpc-go/blob/master/security/advancedtls/advancedtls.go#L220-L222). RSA key exchange, 3DES, and CBC cipher suites removed in [Go 1.22](https://go.dev/doc/go1.22#crypto/tls) and [Go 1.23](https://go.dev/doc/go1.22#crypto/tls).
Dynamic Certificate Loading | No | Yes (Using GetRootCertificates)

This change was manually tested.

Fixes https://github.com/envoyproxy/gateway/issues/5396

Release Notes: Yes


guydc:
cc @nareddyt

guydc (dismissed):
LGTM

sapirpol:
@nareddyt Could you please review?

internal/extension/registry/extension_manager_test.go
```diff
@@ -150,3 +160,103 @@ func Test_setupGRPCOpts(t *testing.T) {
...
+	require.NoError(t, err)
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cert",
+			Namespace: "default",
+		},
+		Type: corev1.SecretTypeTLS,
+		Data: map[string][]byte{
+			corev1.TLSCertKey:       certData,
```
nareddyt, line 233: Unfortunately this doesn't seem right. `extension_manager.go` line 316 is reading CA cert from `corev1.TLSCertKey` (yes it is incorrectly named, but that is a legacy behavior we need to support).

If so, I'm not entirely sure how this test is working (?) Because the CA cert is not in the correct place.

internal/extension/registry/extension_manager.go
```diff
@@ -300,3 +285,41 @@ func setupGRPCOpts(ctx context.Context, client k8scli.Client, ext *egv1a1.Extens
...
+		RootOptions: advancedtls.RootCertificateOptions{
+			// A callback function that dynamically loads root CA certificates from secret
+			GetRootCertificates: createGetRootCertificatesHandler(ctx, client, ext, namespace),
+		},
+	})
+}
+
+func createGetRootCertificatesHandler(ctx context.Context, client k8scli.Client, ext *egv1a1.ExtensionManager, namespace string) func(*advancedtls.ConnectionInfo) (*advancedtls.RootCertificates, error) {
+	return func(params *advancedtls.ConnectionInfo) (*advancedtls.RootCertificates, error) {
+		cp, err := getCertPoolFromSecret(ctx, client, ext, namespace)
```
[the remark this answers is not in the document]
sapirpol: Done

internal/extension/registry/extension_manager.go
```diff
@@ -300,3 +285,41 @@ func setupGRPCOpts(ctx context.Context, client k8scli.Client, ext *egv1a1.Extens
...
+func getCertPoolFromSecret(ctx context.Context, client k8scli.Client, ext *egv1a1.ExtensionManager, namespace string) (*x509.CertPool, error) {
+	certRef := ext.Service.TLS.CertificateRef
+	secret, _, err := kubernetes.ValidateSecretObjectReference(ctx, client, &certRef, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to validate TLS certificate reference: %w", err)
+	}
+
+	caCertPEMBytes, ok := secret.Data[corev1.TLSCertKey]
+	if !ok {
+		return nil, errors.New("no TLS certificate found in CA secret")
```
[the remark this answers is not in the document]
sapirpol: I fixed the error message

sapirpol:
> Thanks for making this PR! A few minor comments on the production code. Though the test may need another look.

Thanks! I’ve made the changes and replied back.

internal/extension/registry/extension_manager_test.go
nareddyt: Wow, did not know that. Thanks for the explanation!

nareddyt: I recommend modifying the test to only pass in the exact necessary data. This makes it clearer to readers that the current naming is incorrect.

Specifically, change the secret to just this, the test should still pass:

```
Data: map[string][]byte{
  corev1.TLSCertKey:       caCert,
},
```

Otherwise the PR looks good.

guydc (dismissed):

sapirpol:
@nareddyt Could you please re-review the PR and trigger the tests?

nareddyt (approved):

arkodg:
Hey @sapirpol can you rebased the go.mod file

sapirpol:
Hi @arkodg, I rebased the branch. Could you please review the PR and retrigger the tests?

arkodg (approved):

Issue #5609, opened 2025-03-25
Remove Gateway API CRD from helm chart


NissesSenap:
*Description*:

I'm migrating from one Gateway API provider to envoy gateway. 
As a part of that, I already have the Gateway API CRDs installed in my cluster.

So when I install Envoy Gateway it starts to overwrite my existing Gateway API CRDs, which of course creates issues.

Sadly, helm is terrible at managing [CRDs](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds), and anything located in the CRD folder will get installed, unless you use skipCRDs flag, but this will of course make it so the envoy CRDs isn't installed.

Sadly helm don't support templating around CRDs and just grabs the pure yaml.

To solve this I would suggest just removing the Gateway API CRDs from the helm chart and point to the official upstream [yaml](https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml).

Another options is that you publish your own version of the Gateway API CRDs if you want them to be part of your specific release.
It's rather easy to setup, here you can find an example how we solved it in the [grafana-operator](https://github.com/grafana/grafana-operator/blob/2533d2c714f61a388ee10a2c64a713f6f9da8f7d/.github/workflows/release.yaml#L138-L139)

NissesSenap:
I'm happy to help out with this issue, just need some pointer how you want to implement it.

arkodg:
hey @NissesSenap, I dont have a good answer here, hoping these points help

* you can use `--skip-crds` to avoid installing the `crds/` but this would also skip the Envoy Gateway CRDs which is not ideal as you've highlighted
* since Helm doesnt update CRDs, you may not need to explicitly add `--skip-crds` in a cluster where Gateway API CRDs are already installed like GKE, it will be a no-op for Gateway API CRDs, and Envoy Gateway should easily be able to run with older CRD versions
* we'd ideally like to create a separate Helm Chart that templates Gateway API CRDs and Envoy Gateway CRDs so you can have selectively do things like `--set crds.GatewayAPI=disabled` , and reuse it for the upgrade case. This work is unfortunately blocked on a Helm limitation tracked with https://github.com/envoyproxy/gateway/issues/4001
* if upstream can move some these CRDS in tree, life will become much easier https://github.com/kubernetes-sigs/gateway-api/issues/3552

NissesSenap:
Thanks for your answer @arkodg , I will jump in to the issue that you linked.
[...]

Issue #3094, opened 2024-04-05
Update Install docs to show how to install EG where the infrastructure/cluster provider is installing Gateway API CRD


arkodg:
I looked into this. The gateway api maintainers [do not provide an upstream helm chart](https://github.com/kubernetes-sigs/gateway-api/issues/1590) that we can source as a dependency. 

Personally, I think that the envoy-gateway helm chart should leave the installation of these CRDs to the users: Just list it as a requirement, and link to their [installation docs](https://gateway-api.sigs.k8s.io/guides/#installing-gateway-api).

Due to this bundling of Gateway API  CRDs and envoy-gateway CRDs, using the helm chart gets a bit fiddly for us when installing into our Google Kubernetes Engine clusters:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: envoy-gateway-system

helmCharts:
  - name: gateway-helm
    releaseName: envoy-gateway
    # The helm chart includes Gateway API, which we already install through a GKE Addon
    includeCRDs: false
    # GKE uses an older version of Gateway API than the latest helm chart
    # https://gateway.envoyproxy.io/v1.0.0/install/matrix/
    version: v0.5.0
    repo: oci://registry-1.docker.io/envoyproxy
    valuesFile: ./values.yaml

resources:
  - https://raw.githubusercontent.com/envoyproxy/gateway/v0.5.0/charts/gateway-helm/crds/generated/config.gateway.envoyproxy.io_envoyproxies.yaml
  - https://raw.githubusercontent.com/envoyproxy/gateway/v0.5.0/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_authenticationfilters.yaml
  - https://raw.githubusercontent.com/envoyproxy/gateway/v0.5.0/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoypatchpolicies.yaml
  - https://raw.githubusercontent.com/envoyproxy/gateway/v0.5.0/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_ratelimitfilters.yaml
```

_Originally posted by @djfinnoy in https://github.com/envoyproxy/gateway/issues/1721#issuecomment-2035558087_

davem-git:
When i run the helm, I don't get the CRDS at all. Is that normal?

pschichtel:
I also noticed this, because I intent to run more than one gateway controller in my cluster.

pschichtel:
[...]

PR #5614, opened 2025-03-25
feat(helm) Make Gateway API CRDs a pre-req

### LINKED ISSUE #4001: CRDs upgrade support

NissesSenap:
**What type of PR is this?**
feature

Support upgrading EG CRDs and make it possible to install EG in a Kubernetes cluster that already got Gateway API CRDs installed.

Fixes #4001

Release Notes: Yes/No


NissesSenap:
I'm a bit too tired, so I just leave this here as a start.

I need to update the quickstart docs and probably some others install instructions. My updates should be in `site/content/en/latest` if I understand correctly.

Since this is a breaking change, we should do some kind of release note, but I can't find where to write it.
Or is it just a description in the PR? 

Another issues that the helm chart follow the version of EG, so it's kind of hard to make a breaking change of the helm chart since versions won't be bumped.

This is kind of normal in the Kubernets world, sadly I think the helm users just have to live with it.... We could of course try to fix it, but it's a pain to manage separate versions, that is why everyone does it this way.

Would love to get any other points on what I have missed, there has been some discussion around how to solve this in #4001, #3094, #5609, and I wanted to move this forward since this is a blocker for me. So I hope a PR will make the issue move forward.

tools/helm-docs/readme.gateway-helm.gotmpl
```diff
@@ -18,6 +18,14 @@
 [Helm](https://helm.sh) must be installed to use the charts.
 Please refer to Helm's [documentation](https://helm.sh/docs) to get started.
 
+### Pre-requisites
+
+Install Kubernetes Gateway API CRDs before installing this chart. You can install the CRDs using the following command:
+
+``` shell
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{ env "GATEWAY_API_VERSION" }}/experimental-install.yaml
```
daixiang0, line 26: Could we add this file into charts and make it as an option? That would be better for user.

NissesSenap:
This will be solved in https://github.com/envoyproxy/gateway/pull/5616/ instead.

PR #5606, opened 2025-03-25
Upgrade Gateway API to 1.3.0

zhaohuabing:
This PR:
* Bump Gateway api to 1.3.0-rc.2
* Implement HTTPCORSFilter

Part of https://github.com/envoyproxy/gateway/issues/5598
Release Notes: Yes

PR #5616, opened 2025-03-26
feature(helm): create gateway-crds-helm helm chart, to store CRDs trough template

NissesSenap:
What type of PR is this?
feature

What this PR does / why we need it:
Support upgrading EG CRDs and make it possible to install EG in a Kubernetes cluster that already got Gateway API CRDs installed.

Which issue(s) this PR fixes:

Fixes https://github.com/envoyproxy/gateway/issues/4001

Release Notes: Yes/No


.github/workflows/release.yaml
```diff
@@ -72,6 +72,11 @@ jobs:
       - name: Generate Release Artifacts
         run: IMAGE_PULL_POLICY=IfNotPresent make generate-artifacts IMAGE=envoyproxy/gateway TAG=${{ env.release_tag }} OUTPUT_DIR=release-artifacts
 
+      - name: Generate and Package CRDs
+        run: |
+          make manifests
+          cd charts/gateway-helm/crds/ && zip -r envoy-gateway-crds.zip generated/
```
zirain, line 78: why we need this?

.github/workflows/release.yaml
```diff
@@ -103,6 +108,7 @@ jobs:
           files: |
             release-artifacts/install.yaml
             release-artifacts/quickstart.yaml
+            release-artifacts/envoy-gateway-crds.yaml
```
zirain, line 111: How is this come?

zirain:
like gateway-helm, we need to release this as oci image.

.github/workflows/release.yaml
NissesSenap: Since helm can't handle CRD upgrades, users can have the option to apply the yaml trough, for example kustomize instead of having to manually clone this repo and grab the CRDs from the helm chart.

So this will especially help users that already got the current helm chart installed and don't feel like changing over to the new one.

.github/workflows/release.yaml
NissesSenap: I had screwed up my different branches when I fixed this. The file now gets generated on line 78 and gets published here.

NissesSenap: This is generated from the make command now. And it publishes the EG CRD to make it easy users of the current helm chart to apply the new CRD version.

NissesSenap:
@zirain 

> like gateway-helm, we need to release this as oci image.

I haven't been able to find where this is done, I was hoping this was solved in some makefile somewhere... Do you have a good idea what I need to update to fix that?

tools/helm-docs/readme.gateway-crds-helm.gotmpl
```diff
@@ -0,0 +1,42 @@
...
+
+If you want to manage the CRDs outside of the Envoy Gateway Helm chart, you can use this chart to install the CRDs separately.
+If you do, make sure that you don't install the CRDs again when installing the Envoy Gateway Helm chart, by using `--skip-crds` flag.
+
+### Install from DockerHub
+
+Once Helm has been set up correctly, install the chart from dockerhub:
+
+``` shell
+    helm install gateway-crds oci://docker.io/envoyproxy/gateway-crds-helm --version v0.0.0-latest -n monitoring --create-namespace
```
illrill, line 29:
```suggestion
    helm install gateway-crds oci://docker.io/envoyproxy/gateway-crds-helm --version v0.0.0-latest -n envoy-gateway-system --create-namespace
```

tools/helm-docs/readme.gateway-crds-helm.gotmpl
```diff
@@ -0,0 +1,42 @@
...
+```
+
+You can find all helm chart release in [Dockerhub](https://hub.docker.com/r/envoyproxy/gateway-crds-helm/tags)
+
+To uninstall the chart:
+
+``` shell
+    helm uninstall gateway-crds -n monitoring
```
illrill, line 37:
```suggestion
    helm uninstall gateway-crds -n envoy-gateway-system
```

charts/gateway-crds-helm/values.yaml
```diff
@@ -0,0 +1,5 @@
+crd:
+  apiGateway:
+    enabled: false
+  envoyGateway:
+    enabled: true
```
illrill, line 5: - `crd` -> `crds` since there are multiple
- `apiGateway` -> `gatewayApi` since that's the correct name
```suggestion
crds:
  gatewayApi:
    enabled: false
  envoyGateway:
    enabled: true
```

NissesSenap: Sure, I will make it enabled by default

NissesSenap:
I'm wondering how to pass the tests: 
image

I can't find any ignore folder setting ether.
I used `make -k gen-check` and I got a few changes, but it won't make `make lint-deps` what I can see.

Is there any command to run to automatically fix the yaml to follow the standard you want? I can't find anything around this in contributing.md ether.

NissesSenap:
> like gateway-helm, we need to release this as oci image.

My understanding that the publish of the helm chart is solved here: https://github.com/envoyproxy/gateway/blob/521b78b6efdcd13ebe3fe14c06c45681a84ef97c/tools/make/helm.mk#L29-L41

Together with: https://github.com/envoyproxy/gateway/blob/521b78b6efdcd13ebe3fe14c06c45681a84ef97c/.github/workflows/release.yaml#L75-L78

charts/gateway-crds-helm/templates/gatewayapi-crds.yaml
```diff
@@ -0,0 +1,14976 @@
+{{- if .Values.crd.apiGateway.enabled }}
```
arkodg, line 1: should be `gatewayApi`

arkodg:
thanks @NissesSenap this is looking good !
i had 2 comments
* imo should be `crd.gatewayAPI` 
* should the default value for `crd.gatewayAPI` be `false` ? this depends on what we think will be the most common deployment model for EG users
  * will they deploy in a cloud which will manage these CRDs for them ?
  * or will they prefer to install these CRDs themselves

arkodg (approved):
LGTM thanks for addressing the comments !

arkodg:
the gen check error can be avoided by committing the changes from `make generate && make manifests`
the lint error can be fixed by adding a line in https://github.com/envoyproxy/gateway/blob/dd1c66fb024e1ce76f59fe1480900307e9ae480d/tools/linter/yamllint/.yamllint#L3

NissesSenap:
Nice, thank you, @arkodg.
I updated the ignore and ran the make commands.

arkodg:
looks like there was another conflict, can you fix the `geo-check` error
@envoyproxy/gateway-maintainers can you help prioritize enabling CI, review and merging this PR, to avoid more conflicts

NissesSenap:
Is it just to run the make generate && make manifests commands again. Or something else?

arkodg:
yeah that or you need to revert the changes in test/helm/gateway-addons-helm/e2e.out.yaml

arkodg:
hey @zhaohuabing lets wait until this is merged, before we merge https://github.com/envoyproxy/gateway/pull/5606 to avoid more conflcts

charts/gateway-crds-helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+``` shell
+    helm uninstall gateway-crds -n envoy-gateway-system
+```
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| crds.envoyGateway.enabled | bool | `true` |  |
+| crds.gatewayApi.enabled | bool | `true` |  |
```
arkodg, line 36: also this should be `gatewayAPI` according to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

NissesSenap: Good catch. I have fixed that now

NissesSenap:
I have changed from `Api` to `API` and removed my changes to the e2e file. Hopefully this time the PR passes.

arkodg:
hey @NissesSenap there were new API changes that landed, can you rebase again and run make generate && make manifests

arkodg (approved):

zirain (approved):

arkodg:
yeah i like option 2, installing in 2 steps, so we can reduce maintenance of managing artifacts on our end yet user workflow looks like 1

arkodg:
also maybe best to disable by default so users dont hit the helm issue by default

NissesSenap:
And people wonder why none likes helm...
My understanding is that this really is an issue only for people using `helm upgrade/helm install`. Great for PoC, but I assume none is doing this for production. I assume upgrade is something only people do in production any way.

It doesn't help to install the CRDs separate in gateway-crds-helm. I tried it and I get the same error.
```
~envoy-gateway/charts/gateway-crds-helm on  main [!?] via 🐹 v1.24.2 at ☸️ kind-kind
➜ helm upgrade -i gateway . --set crds.gatewayAPI.enabled=true --set crds.envoyGateway.enabled=false
Release "gateway" does not exist. Installing it now.
Error: create: failed to create: Secret "sh.helm.release.v1.gateway.v1" is invalid: data: Too long: must have at most 1048576 bytes

~envoy-gateway/charts/gateway-crds-helm on  main [!?] via 🐹 v1.24.2 at ☸️ kind-kind
➜ helm upgrade -i gateway . --set crds.gatewayAPI.enabled=false --set crds.envoyGateway.enabled=true
Release "gateway" does not exist. Installing it now.
Error: create: failed to create: Secret "sh.helm.release.v1.gateway.v1" is invalid: data: Too long: must have at most 1048576 bytes
``` 

I assume it works with `gateway-helm` since it's inside the CRD folder and helm treats those files differently.

If you use ArgoCD you can use `argocd.argoproj.io/sync-options: ServerSideApply=true` or `argocd.argoproj.io/sync-options: Replace=true`
This is how we solve it in the [grafana-operator](https://grafana.github.io/grafana-operator/docs/installation/kustomize/#argocd). Instead of using `kubectl apply -f` you will have to use `kubectl create/replace -f`
In flux, you get server-side apply by default if my memory serves me correct, so it shouldn't be any issues from the start.

My suggestion would be to document out of this issue. There is no good solution due to the limitations of etcd/helm, and it's not much we can do about it.

arkodg:
yeah +1 to this, a workaround for helm / k8s yaml users is `helm template | kubectl apply `

PR #5655, opened 2025-04-02
api(oidc): Allow override of all OIDC CookieNames

### LINKED ISSUE #5152: Option to overwrite every OAuth2Filter Cookie via OIDCCookieNames
**FeatureRequest:**
Can we extend [SecurityPolicy.OIDC.CookieNames](https://gateway.envoyproxy.io/docs/api/extension_types/#oidccookienames) to be able to optionally overwrite every possible OAuth2Filter cookie name (adding: `oauth_hmac`, `oauth_expires`, `refresh_token`, `oauth_nonce`) [see here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2credentials-cookienames)? Currently you can only overwrite `bearer_token` and `id_token`.

**UseCase:**
I would like to be able to remove the internal OAuthFilter cookies before forwarding the request to upstream.

Happy to take this task.

coro:
**What type of PR is this?**

Allows for:

- removal of the internal OAuthFilter cookies before forwarding the request to upstream
- migrations from other OIDC providers with preemptible cookie names

Fixes #5152

Release Notes: No


zhaohuabing (commented):
@coro Thanks for picking this up. Could you please also add CodeVerifier  cookie?

reference: https://github.com/envoyproxy/envoy/pull/37849

zhaohuabing:
A second thought, we probably shouldn't expose these cookies to the users - not like ID and Access tokens, they're some trivial internal implementation details.

zhaohuabing:
@coro sorry for the confusion. I meant that the existing AccessToken and IDToken cookies are needed in the API because they're in the OAuth2/OIDC specs and some backend applications may need them for auth purposes/getting user identity, but all the other cookies should not be exposed to the user-facing API and just be deleted before sending to the Backend applications. We should fix this in the Envoy OAuth2 filter.

Does this make sense?

coro:
That makes sense. I'm happy for this to be closed. I'll leave a comment in the original issue about this.

zhaohuabing:
Tracked in this Enovy issue: https://github.com/envoyproxy/envoy/issues/39196

PR #5661, opened 2025-04-04
feat: Implement requestBuffer

### LINKED ISSUE #3706: Request/Response Buffering API support

markwinter:

This implements the new requestBuffer API using the Envoy's Buffer and BufferPerRoute filters.
API PR: https://github.com/envoyproxy/gateway/pull/5537

If the BTP has a Gateway as a target, then we will configure the Buffer filter on the Listener. If the BTP has a HTTPRoute as a target, then we will configure the BufferPerRoute filter for that particular route. The BufferPerRoute will take precedence - I've tested having both applied and the BufferPerRoute maxBytes won

Fixes #3706 

Release Notes: Yes


api/v1alpha1/envoyproxy_types.go
```diff
@@ -255,6 +255,9 @@ const (
 
 	// EnvoyFilterRouter defines the Envoy HTTP router filter.
 	EnvoyFilterRouter EnvoyFilter = "envoy.filters.http.router"
+
```
zirain, line 258: Since this is a new filter, I think you should pay attention to the `newOrderedHTTPFilter`, take https://github.com/envoyproxy/gateway/pull/5171 as example.

zhaohuabing:
> If the BTP has a Gateway as a target, then we will configure the Buffer filter on the Listener. If the BTP has a HTTPRoute as a target, then we will configure the BufferPerRoute filter for that particular route. The BufferPerRoute will take precedence - I've tested having both applied and the BufferPerRoute maxBytes won.

This behavior is different than other types of settings in the BackendTrafficPolicy, where the settings targeting Gateway apply to all xRoutes. This inconsistent behavior of BackendTrafficPolicy could confuse users. We also can't apply a default RequestBuffer for all the xRoutes under a Gateway. 

I prefer keeping the same "Override" behavior as the other settings, and adding a `Shared` field in the `RequestBuffer`, similar to what we have done with the shared global ratelimit. We'll need to implement the "Merge" mechanism to allow two-level request buffer settings. This is tracked in https://github.com/envoyproxy/gateway/issues/1934 .

arkodg:
hey @markwinter as @zhaohuabing suggested, can the implementation push all the intent to per Route, and we can enhance this in the future with a `shared` api field to be able to set common per gateway buffers ?

markwinter:
@arkodg @zhaohuabing Thanks for the feedback. On holiday for a week and will return to the PR afterwards 👍

arkodg:
hey @markwinter hope you had a good break, feature freeze for v1.4 is April 30th, curious if you youd be able to complete this by then or prefer moving this to v1.5 ?

markwinter:
@arkodg @zhaohuabing Sorry for the little bit of delay in getting back.

I've updated the request buffer xds patching.

Now the patchHCM will add the Buffer filter but in a disabled state. This is required to make any request buffering work at all, as BufferPerRoutes by themselves is not enough. And now in patchRoute it will add the BufferPerRoute settings which will enable Buffer for that route.

If the BTP target was a Gateway it gets added to all Routes, and if it was just a Route, it gets added for just that route.

I took inspiration for how BasicAuth filter works which also adds a HCM filter in Disabled state, but then overrides the per route config.

zhaohuabing (approved):
LGTM thanks!

arkodg (approved):
LGTM thanks for adding this ! this will be part of v1.4  😅

PR #5669, opened 2025-04-06
feat: support dynamic modules

Xunzhuo:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No


arkodg:
1) EEP is probably not the best place for this, since a bad module can take out the entire proxy, not just the route
2) Can you share a workflow of how a user will fetch a module into the pod

arkodg:
> Some options that I can think of:
> 
> * use a docker image to download dynamic modules. This docker image can be deployed as a side car in the Envoy pod and the lib can be loaded to Envoy via a shared volume.
> * use a customized envoy docker image provided by users. The libs for dynamic modules are already loaded to that image.

this experience is not seamless, so maybe lets first document this in docs post v1.4, and then get feedback, and then we can  eliminate the need for the EnvoyPatchPolicy with an API field (maybe in EnvoyProxy) in the future

mathetake:
FYI: Dynamic modules [require the exact version match (at least for now)](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules#compatibility), so for example, the modules working with Envoy v1.50.0 are not guaranteed to work with v1.51.0 (if the compatibility check fails, [then HTTP filter configuration xds request will be rejected](https://github.com/envoyproxy/envoy/blob/538c9bee2bea77ad5eec8a9e09a0cd92defbe38e/source/extensions/dynamic_modules/dynamic_modules.cc#L62-L65)). So, coupling Envoy images would be the safest way; either using the Envoy container itself or with init container downloading (with some version argument). I believe the same restriction applies to [golang filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/golang_filter#developing-a-go-plugin).

So either way, the life cycle of downloading into Envoy container should match the one of Envoy version itself. that's my 2p

zhaohuabing:
@mathetake Thanks for the explanation of the version compatibility between Envoy and dynamic modules. 

This would be a problem for upgrade - since the ABI version is defined in the abi_version.h, EG can't check if a dynamic module will still be compatible or not after upgrade. Users can only find it out after something is broken.

Is there a possible way that allows EG to detect the incompatibility between Envoy and a dynamic module?

mathetake:
There's no mechanism to check that without actually loading Envoy. Like i suggested, the current best way is to tie the Envoy container with modules as in [the examples](https://github.com/envoyproxy/dynamic-modules-examples/blob/main/Dockerfile) or EG introduces some module management init container like the container receives the Envoy version and downloads the specific dynamic modules for that version before Envoy running. To be clear, the same problem exists for golang filter already, and nothing specific to dynamic modules.

mathetake:
so i think we can go with this as-is leaving the lifecycle of dynamic modules to users for now, and then we can later add the "managed" installation of remote dynamic modules like in `EnvoyProxy` API or `EnvoyGateway` API maybe

```yaml
kind: EnvoyProxy
metadata:
  name: envoy-ai-gateway
spec:
  dynamic_modules:
  - name: mymodule
    location:
      url_template: https://foo.bar.com/path/to/${ENVOY_VERSION}/libmymodule.so
```

mathetake:
> but any guesstimate on average and max sizes of the module ?

shared library is almost the same as normal application binary, so the question would be how large a normal application would be; a few MB at least for Rust and Go, and tens of MB if it contains debug info:

```
~/dynamic-modules-examples/rust$ ls -hl target/debug/librust_module.so 
-rwxr-xr-x 2 mathetake mathetake 23M Apr 16 01:58 target/debug/librust_module.so
```

api/v1alpha1/dynamicmodule_types.go
```diff
@@ -0,0 +1,42 @@
...
+	// The search path is configured by the environment variable ENVOY_DYNAMIC_MODULES_SEARCH_PATH.
+	// The actual search path is ${ENVOY_DYNAMIC_MODULES_SEARCH_PATH}/lib${name}.so.
+	//
+	// +kubebuilder:validation:Required
+	Module string `json:"module"`
+
+	// Config is the configuration for the Dynamic Module extension.
+	// This configuration will be passed to the Dynamic Module extension.
+	// +optional
+	Config *apiextensionsv1.JSON `json:"config,omitempty"`
```
mathetake, line 33: i am not sure if we want to force JSON all the time. The rationale behind why the Envoy API is Any is that we don't want to force dynamic modules to pull in dependencies to parse the input. Simply making this to string should work i guess?

api/v1alpha1/envoyextensionypolicy_types.go
```diff
@@ -65,6 +65,13 @@ type EnvoyExtensionPolicySpec struct {
 	// +kubebuilder:validation:MaxItems=16
 	// +optional
 	Lua []Lua `json:"lua,omitempty"`
+
+	// DynamicModule is an ordered list of Dynamic Module filters
+	// that should be added to the envoy filter chain
+	//
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	DynamicModule []DynamicModule `json:"dynamicModule,omitempty"`
```
mathetake, line 74: I guess plural ?
```suggestion
	DynamicModules []DynamicModule `json:"dynamicModules,omitempty"`
```

api/v1alpha1/dynamicmodule_types.go
```diff
@@ -0,0 +1,42 @@
...
+
+// DynamicModule defines a Dynamic Module extension.
+type DynamicModule struct {
+	// Name is a unique name for this Dynamic Module extension. It is used to identify the
+	// Dynamic Module extension if multiple extensions are loaded.
+	// It's also used for logging/debugging.
+	// If not specified, EG will generate a unique name for the Dynamic Module extension.
+	//
+	// +optional
+	Name *string `json:"name,omitempty"`
```
mathetake, line 20: i would make it `ExtensionName` to clarify what is this for. I would also want to add the comments about one module can contain multiple extensions so this name is used to specify one out of many

api/v1alpha1/dynamicmodule_types.go
mathetake: Related to the comment above, I would make this `ExtensionConfig` but not that strong preference

Xunzhuo:
> so i think we can go with this as-is leaving the lifecycle of dynamic modules to users for now, and then we can later add the "managed" installation of remote dynamic modules like in `EnvoyProxy` API or `EnvoyGateway` API maybe
> 
> ```yaml
> apiVersion: gateway.envoyproxy.io/v1alpha1
> kind: EnvoyProxy
> metadata:
>   name: envoy-ai-gateway
> spec:
>   dynamic_modules:
>   - name: mymodule
>     location:
>       url_template: https://foo.bar.com/path/to/${ENVOY_VERSION}/libmymodule.so
> ```

Why not just directly add this to DM fields EEP API? Like what we did in WASM

arkodg:
> > so i think we can go with this as-is leaving the lifecycle of dynamic modules to users for now, and then we can later add the "managed" installation of remote dynamic modules like in `EnvoyProxy` API or `EnvoyGateway` API maybe
> > ```yaml
> > apiVersion: gateway.envoyproxy.io/v1alpha1
> > kind: EnvoyProxy
> > metadata:
> >   name: envoy-ai-gateway
> > spec:
> >   dynamic_modules:
> >   - name: mymodule
> >     location:
> >       url_template: https://foo.bar.com/path/to/${ENVOY_VERSION}/libmymodule.so
> > ```
> 
> Why not just directly add this to DM fields EEP API? Like what we did in WASM

That will allow a route owner to access another route or even the listener violating its scope of extension

mathetake:
So *currently* a dynamic module's capability is as same as Lua and Wasm so I don't have that concern right now. However yes indeed it can expand to have more broader capability plus a module won't be limited to HTTP filter (might end up having Load balancer extension etc). So loading/distribution of modules should be a separate API vs EEP which purely should focus on the configuration of HTTP filters. That's my 2c

api/v1alpha1/dynamicmodule_types.go
Xunzhuo: Yep, make sense, updated

api/v1alpha1/dynamicmodule_types.go
Xunzhuo: Updated

Xunzhuo:
> That will allow a route owner to access another route or even the listener violating its scope of extension

api/v1alpha1/dynamicmodule_types.go
```diff
@@ -0,0 +1,38 @@
...
+	// +optional
+	ExtensionConfig *string `json:"extensionConfig,omitempty"`
+
+	// DoNotClose prevents the module from being unloaded with dlclose.
+	// This is useful for modules that have global state that should not be unloaded.
+	// A module is closed when no more references to it exist in the process.
+	// For example, no HTTP filters are using the module (e.g. after configuration update).
+	//
+	// +optional
+	DoNotClose *bool `json:"doNotClose,omitempty"`
```
mathetake, line 37: Even though it's required at Envoy config level, i am not sure if this could be part of EEP - this feels like an option at the EnvoyProxy API level (distribution/loading) rather than filter. not strong opinion

arkodg: +1

jukie:
@Xunzhuo any plans on picking this back up? Would love to see dynamic modules be supported

arkodg:
@jukie the TODO with this one Is to figure out the delivery mechanism for the modules

Zheaoli:
> the TODO with this one Is to figure out the delivery mechanism for the modules

For us, it would be better to delivery the module with the base envoy proxy image. The dynamic modules is always binding with the certain version of envoy.

jukie:
Closing, https://github.com/envoyproxy/gateway/pull/8278 added this

PR #5672, opened 2025-04-07
api: support per-route tracing

zirain:
xref: https://github.com/envoyproxy/gateway/issues/5486


api/v1alpha1/envoyproxy_tracing_types.go
```diff
@@ -31,6 +31,8 @@ type ProxyTracing struct {
 	SamplingFraction *gwapiv1.Fraction `json:"samplingFraction,omitempty"`
 	// CustomTags defines the custom tags to add to each span.
 	// If provider is kubernetes, pod name and namespace are added by default.
+	//
```
arkodg, line 34: can we take SamplingRate, SamplingFraction and CustomTags, and put in into another struct which we can embed here and in BackendTelemetry ?

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -92,6 +93,33 @@ type BackendTrafficPolicySpec struct {
 	// +notImplementedHide
 	// +optional
 	RequestBuffer *RequestBuffer `json:"requestBuffer,omitempty"`
+	// Telemetry configures the telemetry settings for the backend.
```
arkodg, line 96: should it be backend or route ?

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -92,6 +93,33 @@ type BackendTrafficPolicySpec struct {
...
+	//
```
arkodg, line 97: needs a line here, mentioning that it overrides the setting in the EnvoyProxy resource

api/v1alpha1/envoyproxy_tracing_types.go
zirain: I think we should move to use SamplingFraction and deprecate SamplingRate in the future.

RDS support FractionalPercent directly, so I don't want another convertion here.

arkodg: cool, can you create a embedded struct inside `Tracing` so `SamplingFraction` and `CustomTags` can be reused

zirain: The reason I cannot use `Tracing` inline in `ProxyTracing` is because we need following comment:
```
	// Only one of SamplingRate or SamplingFraction may be specified.
	// If neither field is specified, all requests will be sampled.
```

arkodg: makes sense

arkodg (approved):
LGTM thanks !

shawnh2 (approved):
LGTM

PR #5719, opened 2025-04-11
feat(helm): Add global imageRegistry and imagePullSecrets overrides

### LINKED ISSUE #5718: Add global registry settings instead of per-chart ones
*Description*:

Hello! We're running Envoy Gateway in an environment with a private container registry available and preferable to use over docker.io.

It is possible to achieve, but is a little bit cumbersome with the current chart setup.

First, instead of overriding the registry just once, we have to override everything twice, like so:
```
gateway-helm:
  deployment:
    envoyGateway:
      image:
        repository: private.registry.example.com/envoyproxy/gateway
global:
  images:
    ratelimit:
      image: private.registry.example.com/envoyproxy/ratelimit:123abc
```

Second, the common global overrides as used with bitnami charts (e.g. nginx https://github.com/bitnami/charts/blob/0f3c1ccdeefe6e244c9ff9577e06b6c7e9b4bb15/bitnami/nginx/values.yaml#L12) and some others (e.g. Grafana agent, https://github.com/grafana/agent/blob/18b357cd9cd0bbc5946b73e2ca82ad237c2d52d3/operations/helm/charts/grafana-agent/values.yaml#L10) don't really work: the globals defined right now are specific to Envoy charts. It kind of defeats the purpose of Helm globals, which is to be defined once at the top level, propagate and be applied to any charts down the road.

What I would like to propose is to change the charts so it's possible to define the registry just once at the top level as `global.imageRegistry`, and invert the current overriding logic: right now the locals take precedence over the globals, which means the global overriding won't apply if specified. With this change, it'd be possible to specify the registry only once at the very top level, and have an air-gapped Helm environment get the images from an internal registry. Whoever uses bitnami charts (and similar) along Envoy Gateway gets private registries working for free.

The override would look like this, somewhere at the top chart level:
```
global:
  imageRegistry: private.registry.example.com
```

I will be posting an MR addressing this shortly.

*Relevant Links*
Similar Grafana agent issue: https://github.com/grafana/agent/issues/3929
Corresponding MR: https://github.com/grafana/agent/pull/4127

pasha-gurkov:
The PR enables global container registry and pullSecrets overrides which makes it easier to use Envoy Gateway with private container registries.

Fixes #5718

Release Notes: Yes

**Detailed description**:

Add global.imageRegistry and global.imagePullSecrets to gateway-helm chart values.

Invert override logic: globals take precedence if defined.

Decouple registry, repository and tag from each other.

Add new ENV vars: REPOSITORY, RATELIMIT_REPOSITORY, RATELIMIT_TAG -- REPOSITORY is added to not overcomplicate Makefiles with parsing logic that would be required to stay fully compatible with the current ways (like issuing `IMAGE=xyz make image`); RATELIMIT_REPOSITORY and RATELIMIT_TAG just felt logical to add to make things consistent and configurable in the same way.

Delete ENV vars: IMAGE_NAME, IMAGE -- these are replaced with REGISTRY and REPOSITORY, if default registry is used, just REPOSITORY suffices. This changes developer-side behaviour though, but I'm not sure what's better -- Makefile madness or this. Also, this maps better to `values.yaml`.

### DIFF
--- a/charts/gateway-helm/templates/_helpers.tpl
+++ b/charts/gateway-helm/templates/_helpers.tpl
@@ -65,36 +65,55 @@ Create the name of the service account to use
 The name of the Envoy Gateway image.
 */}}
 {{- define "eg.image" -}}
-{{- if .Values.deployment.envoyGateway.image.repository }}
-{{- .Values.deployment.envoyGateway.image.repository }}:{{ .Values.deployment.envoyGateway.image.tag | default .Values.global.images.envoyGateway.tag | default .Chart.AppVersion }}
-{{- else if .Values.global.images.envoyGateway.image }}
-{{- .Values.global.images.envoyGateway.image }}
-{{- else }}
-docker.io/envoyproxy/gateway:{{ .Chart.Version }}
-{{- end }}
-{{- end }}
+{{- $registryName := default .Values.deployment.envoyGateway.image.registry ((.global).imageRegistry) -}}
+{{- $repositoryName := .Values.deployment.envoyGateway.image.repository -}}
+{{- $termination := .Values.deployment.envoyGateway.image.tag -}}
+{{- printf "%s/%s:%s" $registryName $repositoryName $termination -}}
+{{- end -}}
 
 {{/*
 Pull policy for the Envoy Gateway image.
 */}}
 {{- define "eg.image.pullPolicy" -}}
-{{ .Values.deployment.envoyGateway.imagePullPolicy | default .Values.global.images.envoyGateway.pullPolicy | default "IfNotPresent" }}
+{{ .Values.deployment.envoyGateway.imagePullPolicy }}
 {{- end }}
 
 {{/*
 Pull secrets for the Envoy Gateway image.
 */}}
 {{- define "eg.image.pullSecrets" -}}
-{{- if .Values.deployment.envoyGateway.imagePullSecrets -}}
-imagePullSecrets:
+{{- if .Values.global.imagePullSecrets -}}
+{{ toYaml .Values.global.imagePullSecrets }}
+{{- else if .Values.deployment.envoyGateway.imagePullSecrets -}}
 {{ toYaml .Values.deployment.envoyGateway.imagePullSecrets }}
-{{- else if .Values.global.images.envoyGateway.pullSecrets -}}
-imagePullSecrets:
-{{ toYaml .Values.global.images.envoyGateway.pullSecrets }}
 {{- else -}}
-imagePullSecrets: []
+[]
 {{- end }}
 {{- end }}
+
+{{/*
+The name of the Envoy Ratelimit image.
+*/}}
+{{- define "eg.ratelimit.image" -}}
+{{- $registryName := default .Values.ratelimit.image.registry ((.global).imageRegistry) -}}
+{{- $repositoryName := .Values.ratelimit.image.repository -}}
+{{- $termination := .Values.ratelimit.image.tag -}}
+{{- printf "%s/%s:%s" $registryName $repositoryName $termination -}}
+{{- end -}}
+
+{{/*
+Pull secrets for the Envoy Ratelimit image.
+*/}}
+{{- define "eg.ratelimit.image.pullSecrets" -}}
+{{- if .Values.global.imagePullSecrets -}}
+{{ toYaml .Values.global.imagePullSecrets }}
+{{- else if .Values.ratelimit.imagePullSecrets -}}
+{{ toYaml .Values.ratelimit.imagePullSecrets }}
+{{- else -}}
+{{ toYaml list }}
+{{- end }}
+{{- end }}
+
 
 {{/*
 The default Envoy Gateway configuration.
@@ -105,17 +124,9 @@ provider:
   kubernetes:
     rateLimitDeployment:
       container:
-        {{- if .Values.global.images.ratelimit.image }}
-        image: {{ .Values.global.images.ratelimit.image }}
-        {{- else }}
-        image: "docker.io/envoyproxy/ratelimit:master"
-        {{- end }}
-      {{- with .Values.global.images.ratelimit.pullSecrets }}
+        image: {{ include "eg.ratelimit.image" . }}
       pod:
-        imagePullSecrets:
-        {{- toYaml . | nindent 10 }}
-      {{- end }}
-      {{- with .Values.global.images.ratelimit.pullPolicy }}
+        imagePullSecrets: {{- include "eg.ratelimit.image.pullSecrets" . | nindent 10 }}
       patch:
         type: StrategicMerge
         value:
@@ -124,8 +135,7 @@ provider:
               spec:
                 containers:
                 - name: envoy-ratelimit
-                  imagePullPolicy: {{ . }}
-      {{- end }}
+                  imagePullPolicy: {{ .Values.ratelimit.imagePullPolicy }}
     shutdownManager:
       image: {{ include "eg.image" . }}
 {{- end }}
--- a/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
@@ -84,7 +84,7 @@ spec:
         - mountPath: /certs
           name: certs
           readOnly: true
-      {{- include "eg.image.pullSecrets" . | nindent 6 }}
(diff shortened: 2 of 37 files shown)


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -1,33 +1,37 @@
-# The global settings for the Envoy Gateway Helm chart.
-# These values will be used if the values are not overridden in the other sections.
+# Global settings
+# If set, these override image-specific values: useful when installing the chart in a private registry environment.
+# Override image-specific values directly if a global override is not desired.
 global:
-  images:
```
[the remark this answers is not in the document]
pasha-gurkov: @arkodg I'll rework this in a backwards-compatible way. What about environment variables, would you be okay with changes or rather keep those as is?

pasha-gurkov:
@arkodg I pushed the backwards-compatible version. It now has some additional parsing inside `_helpers.tpl`, and the overriding logic became somewhat hairy -- if `global.imageRegistry` is defined, it overrides both `global.images` and `deployment`. Regarding my comments for ENV variables, if you'd rather return those to how they were, I'm not opposed to it - they no longer map to much.

pasha-gurkov:
Thanks for the comment @arkodg ,
> can we scope the changes in this PR to only helm charts, else it will take much longer for a review
Sure thing, I'll push it later today.

> should it be global.images.registry and global.images.pullSecrets
My 2 cents is it should be global.imageRegistry as that's what bitnami charts use. This should give the most win for the end-users (single override working everywhere), as bitnami charts are very popular.

> and global.images.envoyGateway.image should have higher precedence over global.images.registry since its most specific
This would render the PR useless. The point of this is to be able to override registry somewhere at the top level and let it propagate downstream, with minimal possible changes to the downstream charts. gateway-specific values taking precedence would mean one would still have to override `global.images.envoyGateway.image` or `deployment.envoyGateway.image`, since they're both more specific than global. 

Globals overriding specific values is a common practice for helm globals, please take a look at https://github.com/bitnami/charts/blob/0f3c1ccdeefe6e244c9ff9577e06b6c7e9b4bb15/bitnami/common/templates/_images.tpl#L13 or https://github.com/grafana/agent/blob/18b357cd9cd0bbc5946b73e2ca82ad237c2d52d3/operations/helm/charts/grafana-agent/templates/containers/_agent.yaml#L3 .

If end-user wants to opt-out, and e.g. use their private registry for every image _but_ envoy-gateway, they can override a chart-specific global:
```
gateway-helm:
  global:
    imageRegistry: ""
```

pasha-gurkov:
> > can we scope the changes in this PR to only helm charts, else it will take much longer for a review
> Sure thing, I'll push it later today.

Reverted all Makefile and other relevant changes.

arkodg:
thanks @pasha-gurkov, the reasoning for top level overriding the specific images makes sense (by default the specifics are set today in the helm chart)

release-notes/current.yaml
```diff
@@ -4,6 +4,7 @@ date: Pending
 breaking changes: |
   Use a dedicated listener port(19003) for envoy proxy readiness
   Uses the envoy JSON formatter for the default access log instead of text formatter.
+  Removed environment variables IMAGE, IMAGE_NAME used for building the image; use REGISTRY and REPOSITORY instead
```
arkodg, line 7: lets rm this

test/helm/gateway-helm/deployment-images-config.out.yaml
```diff
@@ -571,8 +573,8 @@ spec:
           seccompProfile:
             type: RuntimeDefault
       imagePullSecrets:
-      - name: secret1
-      - name: secret2
+        - name: secret1
```
arkodg, line 576: here

charts/gateway-helm/templates/envoy-gateway-deployment.yaml
```diff
@@ -84,7 +84,7 @@ spec:
         - mountPath: /certs
           name: certs
           readOnly: true
-      {{- include "eg.image.pullSecrets" . | nindent 6 }}
+      imagePullSecrets: {{- include "eg.image.pullSecrets" . | nindent 8 }}
```
arkodg, line 87: can we revert this ?
it seems to be changing the default indentation style in the file
https://github.com/envoyproxy/gateway/pull/5719/files#r2047900977

charts/gateway-helm/templates/_helpers.tpl
```diff
@@ -65,37 +65,85 @@ Create the name of the service account to use
...
+The name of the Envoy Ratelimit image.
+*/}}
+{{- define "eg.ratelimit.image" -}}
+{{-   $imageParts := splitn "/" 2 .Values.global.images.ratelimit.image -}}
+{{/*    if global.imageRegistry is defined, it takes precedence always */}}
+{{-   $registryName := default $imageParts._0 .Values.global.imageRegistry -}}
+{{-   $repositoryTag := $imageParts._1 -}}
+{{-   $repositoryParts := splitn ":" 2 $repositoryTag -}}
+{{-   $repositoryName := $repositoryParts._0 -}}
+{{-   $imageTag := default .Chart.AppVersion $repositoryParts._1 -}}
```
arkodg, line 129: the ratelimit image tag is not tied to the EG chart version but to short commit SHAs from the ratelimit repo
https://hub.docker.com/r/envoyproxy/ratelimit/tags
so this change will break release charts

release-notes/current.yaml
pasha-gurkov: Makes sense, done.

pasha-gurkov:
@arkodg thanks for the feedback! Please check the latest commit, should have addressed all of it.

test/helm/gateway-helm/deployment-priorityclass.out.yaml
```diff
@@ -47,6 +47,8 @@ data:
                     containers:
                     - imagePullPolicy: IfNotPresent
                       name: envoy-ratelimit
+          pod:
```
arkodg, line 50: can this change be avoided ? is it harmless ? for users moving to newer charts with an old values file, who may have `pod` set but `imagePullSecrets` will be `nil` for them, will this cause any regression ?

test/helm/gateway-helm/deployment-images-config.out.yaml
```diff
@@ -47,8 +47,10 @@ data:
                     containers:
                     - imagePullPolicy: IfNotPresent
                       name: envoy-ratelimit
+          pod:
+            imagePullSecrets: []
         shutdownManager:
-          image: private-hub/envoyproxy/gateway:abcdef12
+          image: envoyproxy/gateway:abcdef12
```
arkodg, line 53: this doesnt look right

charts/gateway-helm/templates/_helpers.tpl
```diff
@@ -65,37 +65,90 @@ Create the name of the service account to use
...
-{{- .Values.global.images.envoyGateway.image }}
-{{- else }}
+{{/* if deployment-specific repository is defined, it takes precedence */}}
+{{- if .Values.deployment.envoyGateway.image.repository -}}
+{{/*  if global.imageRegistry is defined, it takes precedence always */}}
+{{-   if .Values.global.imageRegistry -}}
+{{-     $repositoryParts := splitn "/" 2 .Values.deployment.envoyGateway.image.repository -}}
+{{-     $registryName := .Values.global.imageRegistry -}}
+{{-     $repositoryName := $repositoryParts._1 -}}
+{{-     $imageTag := default .Chart.AppVersion .Values.deployment.envoyGateway.image.tag -}}
```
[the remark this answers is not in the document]
arkodg: cool thanks for highlighting this, then this should not affect default behavior

test/helm/gateway-helm/deployment-images-config.out.yaml
pasha-gurkov: That's strange, could've sworn I checked in all the test outs. Should be fine in the next commit.

test/helm/gateway-helm/deployment-priorityclass.out.yaml
pasha-gurkov: I would imagine users would override the config as a whole anyway; but to be on the safe side, I've hidden this under pullSecrets being defined at all to not change the behaviour.

arkodg (approved):
LGTM thanks !

pasha-gurkov:
Thank you very much @arkodg!

zirain (approved):

PR #5731, opened 2025-04-14
Added metrics for restclient and workqueue

Inode1:
This PR introduces additional metrics to improve observability around the Envoy Gateway, specifically focusing on the REST client and the workqueue.

Motivation
There are currently several issues that limit our ability to effectively monitor and troubleshoot the system:

Lack of Metrics: The limited set of existing metrics makes it difficult to gain insights into the performance and health of the Envoy Gateway. This PR enhances observability and makes it easier to identify and diagnose performance issues—particularly those related to the workqueue.

Workqueue Metrics Conflict: Metrics for the workqueue are registered both by controller-runtime and k8s.io/component-base via init() functions. However, only one of these libraries can successfully register the metrics. In our setup, k8s.io/component-base wins, which results in the metrics from controller-runtime not being exposed at all.

This PR resolves these issues by explicitly adding and managing the necessary metrics, ensuring better visibility into queue behavior and client interactions.

Release Notes: Yes

### DIFF
--- a/internal/metrics/workqueue/metrics.go
+++ b/internal/metrics/workqueue/metrics.go
@@ -1,0 +1,151 @@
+// Copyright Envoy Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package workqueue
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// This file is copied and adapted from k8s.io/component-base/metrics/prometheus/workqueue
+// which registers metrics to the k8s legacy Registry. We require very
+// similar functionality, but must register metrics to a different Registry.
+
+// Metrics subsystem and all keys used by the workqueue.
+const (
+	WorkQueueSubsystem         = metrics.WorkQueueSubsystem
+	DepthKey                   = metrics.DepthKey
+	AddsKey                    = metrics.AddsKey
+	QueueLatencyKey            = metrics.QueueLatencyKey
+	WorkDurationKey            = metrics.WorkDurationKey
+	UnfinishedWorkKey          = metrics.UnfinishedWorkKey
+	LongestRunningProcessorKey = metrics.LongestRunningProcessorKey
+	RetriesKey                 = metrics.RetriesKey
+)
+
+var (
+	depth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: WorkQueueSubsystem,
+		Name:      DepthKey,
+		Help:      "Current depth of workqueue by workqueue and priority",
+	}, []string{"name", "controller", "priority"})
+
+	adds = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: WorkQueueSubsystem,
+		Name:      AddsKey,
+		Help:      "Total number of adds handled by workqueue",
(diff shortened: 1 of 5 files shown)


test/e2e/tests/metric.go
```diff
@@ -96,6 +96,54 @@ var MetricTest = suite.ConformanceTest{
...
+			// make sure listener is ready
+			httputils.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, expectedResponse)
+
+			// Check that workqueue and restclient metrics are registered
+			if err := wait.PollUntilContextTimeout(context.TODO(), time.Second, time.Minute, true,
+				func(_ context.Context) (done bool, err error) {
+					url, err := RetrieveURL(suite.Client, types.NamespacedName{
+						Namespace: "envoy-gateway-system",
+						Name:      "same-namespace-gw-metrics",
+					}, 19001, "/stats/prometheus")
```
zirain, line 119: isn't this's a controller metric? looks like you're trying to get metrics from dataplane?

Inode1: Yup, you are right. I couldn't run e2e tests locally, so I decided to run them in ci). It should work now, could you rerun tests?

Inode1: @zirain fix linter

test/e2e/tests/metric.go
```diff
@@ -99,6 +100,71 @@ var MetricTest = suite.ConformanceTest{
 	},
 }
 
+var MetricWorkqueueAndRestclientTest = suite.ConformanceTest{
```
zirain, line 103: QQ: can we simply query result from prometheus?

Inode1: What do you mean? Can you give more details?

zirain: can you checkout if [this](https://github.com/envoyproxy/gateway/blob/8a01dd68ae8c352c438292fdb93316d9badd35d8/test/utils/prometheus/prometheus.go#L31) help?

Inode1: I can switch the test to use prom-client if you really want, but honestly, I’m not seeing much of a difference here—what do you think?

zirain: yeah, better to keep things in same way.

Inode1: @zirain Done

Inode1:
@zirain, could you take a look?

internal/metrics/workqueue/metrics.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// This file is copied and adapted from k8s.io/component-base/metrics/prometheus/workqueue
```
arkodg, line 17: can you share how this is different from https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/internal/metrics/workqueue.go#L23

which is being set by default
https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/controller/priorityqueue/priorityqueue.go#L58

arkodg: reread the title description, trying to better under the conflict

arkodg: thanks

arkodg (approved):

PR #5732, opened 2025-04-14
Add nemlig as adopter of envoygateway in production

Havnevej:
Add [nemlig](https://nemlig.com) as an adopter of envoy gateway in production


arkodg:
thanks @Havnevej !
could you 
* sign your commit and force push
* also update this https://github.com/envoyproxy/gateway/blob/3dbf9cc00456f2566e6c55e6c2f58a150b04ee46/tools/make/docs.mk#L7 to include nemlig.com, else linkinator seems to be failing

Xunzhuo (approved):

Havnevej:
@arkodg  Should be good now

arkodg (approved):

PR #5734, opened 2025-04-14
api: adds mutationRules config on extproc

mathetake:
**What type of PR is this?**

api: adds headerMutationRules on extproc configuration

Without this, an external processor cannot modify, say, "host" header.


arkodg:
@mathetake there are some alternatives already implemented that solve this use case https://gateway.envoyproxy.io/docs/tasks/traffic/http-urlrewrite/#rewrite-url-host-name-by-header-or-backend

mathetake:
hah

mathetake:
ok i don't think i need this one then

PR #5741, opened 2025-04-16
feat(helm): Add management of extensionApis

sorrison:
Adds the ability to enable extension APIs with the helm chart

### DIFF
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -93,6 +93,7 @@ config:
     logging:
       level:
         default: info
+    extensionApis: {}
 
 createNamespace: false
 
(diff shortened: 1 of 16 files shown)


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -18,6 +18,7 @@ global:
       pullPolicy: IfNotPresent
       # List of secrets in the same namespace of the component that can be used to pull images from private repositories.
       pullSecrets: []
+  extensionApis: {}
```
arkodg, line 21: can this be added in L97 instead ? under `config.envoyGateway` ?

sorrison: ok I've done that but I think it has unintended consequences. At least when I added all the stuff from `make generate` but still learning all this stuff

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5749, opened 2025-04-17
chore: revert ratelimit per-rout config

zhaohuabing:
Revert https://github.com/envoyproxy/gateway/pull/5072

Some existing users may choose to keep using an old version of envoy while upgrading to new version of Envoy Gateway. To maintain compatibility, we're reverting this change so that rate limit configuration continues to work with older Envoy versions.

We can revisit this in a few versions, once we're confident that older Envoy versions are no longer in use.


arkodg (approved):
thanks @zhaohuabing

zirain (approved):

Issue #5662, opened 2025-04-04
docs: Add steps in troubleshooting docs to access admin interface


arkodg:
*Description*:
>Describe the issue.

Some steps live in https://gateway.envoyproxy.io/contributions/develop/#debugging-the-envoy-config which can be used

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #5758, opened 2025-04-17
docs: troubleshooting by accessing envoy proxy admin interface

arkodg:
Fixes: https://github.com/envoyproxy/gateway/issues/5662


zirain:
this can be simply by using egctl, should we point out that?

arkodg:
good point @zirain, can we do that in a follow up ? we'll need to add a section in https://gateway.envoyproxy.io/docs/tasks/operations/egctl/ and link it in this doc

zirain (approved):

site/content/en/latest/troubleshooting/envoy-proxy-admin-interface.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Note**: Application Developers may not have access to the namespace where the Envoy Proxy fleet is running and should rely on [exported telemetry](https://gateway.envoyproxy.io/docs/tasks/observability/) instead for troubleshooting.
+
+## Prerequisites
+
+{{< boilerplate prerequisites >}}
+
+### Access
+
+You will need to port-forward to the admin interface port (currently 19000) on the Envoy deployment that corresponds to a Gateway, since it only listens on the `localhost`
```
cnvergence, line 16: It might be worth to mention that this port needs to be exposed in the deployment and service of EG, as it is not exposed by default
wdyt?

arkodg: I don't think the port needs to be defined in the deployment,  if you check out the steps, it works w/o doing that

cnvergence: sorry, I thought this is about the EG deployment, my previous comment is invalid then
This is working fine :)

cnvergence (approved):

PR #5760, opened 2025-04-18
feat: log resource status for standalone mode

### LINKED ISSUE #5556: Add resources status for file-provider
*Description*:

File provider should be able to process, maintain and retrieve the status of each resource.

We can start a status server along with file provider, process and maintain the resources status in mem. Provide a API allow users to retrieve resource status, also would be better if we can utilise `egctl x status` for such thing.

This will be very helpful to debug in standalone mode.

shawnh2:
**What type of PR is this?**

image

Fixes #5556

Release Notes: Yes


zirain (approved):

arkodg (approved):
LGTM thanks !

we can improve on structure later

PR #5763, opened 2025-04-18
Revert the temporary bug simulator fuzzer.

sudiptob2:


arkodg (approved):

cnvergence (approved):

sudiptob2:
It would be beneficial to wait for the `OSS-Fuzz` **coverage build** to report some logs. Let's wait utill a coverage report it generated in OSS-Fuzz.

sudiptob2:
we can merge this one now. As we found out the root cause why coverage is not working. Therefore, we do not need the temporary bug simulator.

PR #5767, opened 2025-04-19
feat: implement offline kubernetes controller

### LINKED ISSUE #3213: refactor the kubernetes provider to be provider-insensitive

shawnh2:
**What type of PR is this?**

impl offline controller detailed in https://github.com/envoyproxy/gateway/issues/3213#issuecomment-2816518741 and let file-provider to be able to utilise offline controller.

fix #3213

Release Notes: Yes


zhaohuabing:
Hi @shawnh2 this PR adds the "offline controller," but it doesn’t include an example of how to actually use it. Since it's just a small change, could we also include the file provider modification as part of this PR?

shawnh2:
> Hi @shawnh2 this PR adds the "offline controller," but it doesn’t include an example of how to actually use it. Since it's just a small change, could we also include the file provider modification as part of this PR to verify it works?
> 
> 
> 
> Thanks!
> 
> 
> 
> 

sure, I was intend to include file provider in next PR. I will add it then.

arkodg (dismissed):
LGTM thanks

arkodg:
@shawnh2 can you rebase ?

zhaohuabing (approved):
LGTM thanks!

mathetake:
So this PR has a regression:

https://github.com/envoyproxy/ai-gateway/blob/87f98c6d5d08500bdd07aeec39a3d268c22e5574/cmd/aigw/run_test.go#L52

this test runs successfully before this PR and failing after. It looks like a Route is not properly reconciled as it seems Envoy starts returning "route not found" after this PR. On the other hand, the standalone tutorial runs successfully even after this PR, so I believe there's some edge case that is not covered by tests in EG.

Issue #5598, opened 2025-03-25
Update to Gateway API v1.3.0


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

RC just released https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.3.0-rc.1

- [x] Bump version

- [x]  Add OverlappingTLSConfig condition https://github.com/envoyproxy/gateway/issues/2675#issuecomment-2780706141

- [x]  Implement CORS https://gateway-api.sigs.k8s.io/geps/gep-1767/

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Issue #2675, opened 2024-02-22
Multiple Gateway listeners with different hostnames and same certificate not working in a browser session 


jaynis:
I have configured two listeners on a `Gateway` for two different (sub)domains, e.g. `foo.example.com` and `bar.example.com`: 

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: envoy-gateway-https
  namespace: default
spec:
  gatewayClassName: envoy
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    hostname: 'foo.example.k3d'
    name: example-foo
    port: 443
    protocol: HTTPS
    tls:
      certificateRefs:
      - group: ""
        kind: Secret
        name: example-wildcard-tls
        namespace: default
      mode: Terminate
  - allowedRoutes:
      namespaces:
        from: Same
    hostname: 'bar.example.k3d'
    name: example-bar
    port: 443
    protocol: HTTPS
    tls:
      certificateRefs:
      - group: ""
        kind: Secret
        name: example-wildcard-tls
        namespace: default
      mode: Terminate
```

If I try to reach `foo.example.com` or `bar.example.com` with a CLI tool such as `curl` both domains are working fine and the desired content is correctly served. However, if I try to do the same from within a browser session with Chrome or Firefox, only one of both is working and the other one responds with 404's. Which one is working and which one not depends on the order in which I attempt to reach it: If I open Chrome and navigate to `foo.example.com` the desired content is served correctly, but if I go to `bar.example.com` afterwards I receive a 404. If I close the browser and open it again and this time first navigate to `bar.example.com` it is correctly served and subsequent requests to `foo.example.com` receive 404's. Therefore I assume this might have something to do with connection sharing / reuse, but this is only a wild guess.

In the envoy debug logs I can see the failing request and the reason for the 404 seems to be a problem with the route matching:
```
[2024-02-22 15:32:32.711][26][debug][router] [source/common/router/router.cc:446] [Tags: "ConnectionId":"3","StreamId":"11246534096143794784"] no route match for URL '/'
```

If I remove the `hostname` constraint from the `Gateway` entirely, all domains are working fine as well as with `curl` as in a browser session.

akevdmeer:
Does this perhaps involve HTTP/2 connection coalescing by the browser?

arkodg:
hey @jaynis would help if you shared
*  the `HTTPRoute` attached to the `Gateway`
[...]

PR #5777, opened 2025-04-21
feat: set OverlappingTLSConfig condition for HTTPS listeners with conflicting hostnames or certs

zhaohuabing:
This PR adds support for `OverlappingTLSConfig` condition in GatewayStatus. This condition is set if there are overlapping hostnames or certificates between listeners. The ALPN protocol is set to HTTP/1.1 for the overlapping listeners to avoid [HTTP/2 Connection Coalescing](https://gateway-api.sigs.k8s.io/geps/gep-3567/).

Implements: #5598
Part of: #2675
Release Notes: Yes
Reference: https://gateway-api.sigs.k8s.io/geps/gep-3567/


internal/gatewayapi/contexts.go
```diff
@@ -73,6 +75,9 @@ type ListenerContext struct {
 	listenerStatusIdx int
 	namespaceSelector labels.Selector
 	tlsSecrets        []*corev1.Secret
+	tlsCertificates   []*x509.Certificate
```
arkodg, line 78: this is going to take up a lot of mem, can we instead only save DNSNames ?

arkodg: although its a pointer here, its avoiding gc of the underlying memory

internal/gatewayapi/listener.go
```diff
@@ -172,9 +176,172 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
...
+
+func checkOverlappingHostnames(gateway *GatewayContext) {
+	httpsListeners := []*ListenerContext{}
+	for _, listener := range gateway.listeners {
+		if listener.Protocol == gwapiv1.HTTPSProtocolType {
+			httpsListeners = append(httpsListeners, listener)
+		}
+	}
+
+	overlappingListeners := make([][2]string, len(httpsListeners)) // The listener name and hostname of the overlapping listener.
```
arkodg, line 202: nit: temp struct instead of [2] will improve readability here

internal/gatewayapi/listener.go
```diff
@@ -172,9 +176,172 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
...
+
+func checkOverlappingCertificates(gateway *GatewayContext) {
+	httpsListeners := []*ListenerContext{}
+	for _, listener := range gateway.listeners {
+		if listener.Protocol == gwapiv1.HTTPSProtocolType {
+			httpsListeners = append(httpsListeners, listener)
+		}
+	}
+
+	overlappingListeners := make([]*overlappingListener, len(httpsListeners)) // The listener name and san of the overlapping listener.
```
arkodg, line 259: same as above

internal/gatewayapi/listener.go
```diff
@@ -172,9 +176,172 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
...
+					san2: dns2,
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// isOverlappingHostname checks if two hostnames overlap.
+func isOverlappingHostname(hostname1, hostname2 *gwapiv1.Hostname) bool {
```
arkodg, line 335: can we reuse / dedup some of https://github.com/envoyproxy/gateway/blob/7b64768859d1221c039fe0e767fbf1bc41b42e5c/internal/gatewayapi/route.go#L820 ?

internal/gatewayapi/status/gateway.go
```diff
@@ -182,3 +182,11 @@ func updateGatewayProgrammedCondition(gw *gwapiv1.Gateway, envoyObj client.Objec
 		newCondition(string(gwapiv1.GatewayConditionProgrammed), metav1.ConditionFalse, string(gwapiv1.GatewayReasonNoResources),
 			messageNoResources, time.Now(), gw.Generation))
 }
+
+// GetGatewayListenerStatusConditions returns the status conditions for a specific listener in the gateway status.
+func GetGatewayListenerStatusConditions(gateway *gwapiv1.Gateway, listenerStatusIdx int) []metav1.Condition {
```
arkodg, line 187: is this being used ?

internal/gatewayapi/contexts.go
zhaohuabing: Yeah, good catch! Raw cert data can be huge.

internal/gatewayapi/status/gateway.go
zhaohuabing: It's used in the listener_test.go.

internal/gatewayapi/listener.go
zhaohuabing: I don't think we can. These are handling two different logics.

arkodg (approved):
LGTM thanks !

guydc:
Do we treat merge gateway correctly here and consider overlap across merged GWs? If not, we can do that as a further improvement.

internal/gatewayapi/listener.go
```diff
@@ -172,7 +175,182 @@ func (t *Translator) ProcessListeners(gateways []*GatewayContext, xdsIR resource
...
+
+	for i, listener := range httpsListeners {
+		if overlappingListeners[i] != nil {
+			status.SetGatewayListenerStatusCondition(gateway.Gateway,
+				listener.listenerStatusIdx,
+				gwapiv1.ListenerConditionOverlappingTLSConfig,
+				metav1.ConditionTrue,
+				gwapiv1.ListenerReasonOverlappingCertificates,
+				fmt.Sprintf(
+					"The certificate san %s overlaps with the certificate san %s in listener %s",
```
arkodg, line 313: can we add a line saying ALPN has been set to http/1.1 to avoid connection coalescing issues seen on h2 ?

arkodg (dismissed):

zhaohuabing:
> Do we treat merge gateway correctly here and consider overlap across merged GWs? If not, we can do that as a further improvement.

Track this separately here as we're trying to release v1.4 rc today. https://github.com/envoyproxy/gateway/issues/5860

arkodg (approved):

zirain (approved):

PR #5788, opened 2025-04-21
fix: allow users to set empty retriable status codes

### LINKED ISSUE #5691: retriable_status_codes Defaults to 503 Even if Not Configured in retry_on
We noticed that the `retriable_status_codes` field is being populated with [503] even when `retry_on.httpStatusCodes` is not explicitly configured.

Expected Behavior: `retriable_status_codes` should only be set when `retry_on.httpStatusCodes` is explicitly supplied.

Actual Behavior: If httpStatusCodes is unset, Envoy Gateway still sets the default 503 value in `retriable_status_codes`.

guydc:
**What type of PR is this?**

- When retryOn is set to an empty struct, existing defaults work as before.
- When users customize retryOn and leave retryOn.httpStatusCodes unset, the xds retriableStatusCodes will remain unset.

Fixes #5691

Release Notes: Yes

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -640,9 +640,7 @@ func buildRetryPolicy(route *ir.HTTPRoute) (*routev3.RetryPolicy, error) {
 			}
 		}
 
-		if len(rr.RetryOn.HTTPStatusCodes) > 0 {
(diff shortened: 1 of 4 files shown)


arkodg (approved):
LGTM thanks

zirain (approved):

cnvergence (approved):

Issue #5298, opened 2025-02-18
Implement gep-3388 (Retry Budget)


zirain:
xref: https://github.com/kubernetes-sigs/gateway-api/pull/3488

zirain:
We have a user case around retry budget, instead of introuduce new filed into BTP, prefer to implement XBackendTrafficPolicy.
cc @envoyproxy/gateway-maintainers

arkodg:
Can we support both ? 
I'm interested in retry budget in BTP

zirain:
Then let's start with BTP first and wait for BackendTrafficPolicy promote to standard channel.

ericdbishop:
Might be of interest for the Envoy Gateway BackendTrafficPolicy implementation: https://github.com/envoyproxy/envoy/pull/44147

PR #5790, opened 2025-04-22
feat: impl GEP-3388 (Retry Budget)

zirain:
xref: https://github.com/envoyproxy/gateway/issues/5298


internal/gatewayapi/testdata/xbackendtrafficpolicy-invalid-retry-budget.in copy.yaml
```diff
@@ -0,0 +1,50 @@
+gateways:
```
zhaohuabing, line 1: This file should be deleted?

internal/gatewayapi/testdata/xbackendtrafficpolicy-invalid-retry-budget.out.yaml
```diff
@@ -0,0 +1,225 @@
...
+        group: gateway.envoyproxy.io
+        kind: Backend
+        name: backend-not-exist
+      conditions:
+      - lastTransitionTime: null
+        message: Min retry rate interval is not supported.
+        reason: Invalid
+        status: "False"
+        type: Accepted
+      controllerName: gateway.envoyproxy.io/gatewayclass-controller
```
zhaohuabing, line 159: Should we align the status behavior with BackendTLSPolicy and use the Gateway as the ancestor?

zhaohuabing:
```
// Ancestors refer to objects that are either the Target of a policy or above it
// in terms of object hierarchy. For example, if a policy targets a Service, the
// Policy's Ancestors are, in order, the Service, the HTTPRoute, the Gateway, and
// the GatewayClass. Almost always, in this hierarchy, the Gateway will be the most
// useful object to place Policy status on, so we recommend that implementations
// SHOULD use Gateway as the PolicyAncestorStatus object unless the designers
// have a _very_ good reason otherwise.
```

zirain: sounds good

zhaohuabing:
@zirain Let's hold this until https://github.com/envoyproxy/gateway/pull/6153 is fixed.

PR #5794, opened 2025-04-23
API: TLS API for dynamic resolver Backend

zhaohuabing:
This PR introduces the TLS configuration API for the `DynamicResolver` Backend type.

We are not using the gateway API `BackendTLSPolicy` to configure TLS for `DynamicResolver`, because it requires a fixed hostname, which isn't compatible with the dynamic nature of these backends.

Release note:  the release note will be added in the implementation PR.

CEL validation for API will be added in the implementation PR.


api/v1alpha1/backend_types.go
```diff
@@ -141,6 +143,41 @@ type BackendSpec struct {
...
+
+	// TLS defines the TLS settings for the backend.
+	// Only supported for DynamicResolver backends.
+	//
+	// +optional
+	TLS *BackendTLSSettings `json:"tls,omitempty"`
+}
+
+// BackendTLSSettings holds the TLS settings for the backend.
+// Only used for DynamicResolver backends.
```
arkodg, line 155: // +kubebuilder:validation:XValidation:message="must not contain both CACertificateRefs and WellKnownCACertificates",rule="!(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 && has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"
// +kubebuilder:validation:XValidation:message="must specify either CACertificateRefs or WellKnownCACertificates",rule="(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 || has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"

zhaohuabing:
> minor comment around CEL, else LGTM, thanks for adding this !

I plan to add CEL and CEL tests in the implementation PR to focus this one on the API itself :-)

arkodg (dismissed):
LGTM thanks !

Issue #5745, opened 2025-04-16
Surface merged spec in the status.


arkodg:
Not directly API-related -- It would be hard for users to figure out what's the effective configuration if merging is used. To improve usability,  we should surface the merged spec in the status.

_Originally posted by @zhaohuabing in https://github.com/envoyproxy/gateway/issues/5582#issuecomment-2808038082_

PR #5795, opened 2025-04-23
feat: impl mergeType in BackendTrafficPolicy

zirain:
xref: https://github.com/envoyproxy/gateway/issues/1934

### DIFF
--- a/api/v1alpha1/backendtrafficpolicy_types.go
+++ b/api/v1alpha1/backendtrafficpolicy_types.go
@@ -85,8 +85,11 @@ type BackendTrafficPolicySpec struct {
 	// HTTPUpgrade defines the configuration for HTTP protocol upgrades.
 	// If not specified, the default upgrade configuration(websocket) will be used.
 	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	//
 	// +optional
-	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty"`
+	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty" patchMergeKey:"type" patchStrategy:"merge"`
 
 	// RequestBuffer allows the gateway to buffer and fully receive each request from a client before continuing to send the request
 	// upstream to the backends. This can be helpful to shield your backend servers from slow clients, and also to enforce a maximum size per request
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -69,12 +69,13 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 	handledPolicies := make(map[types.NamespacedName]*egv1a1.BackendTrafficPolicy)
 
 	gatewayPolicyMap := make(map[types.NamespacedName]*egv1a1.BackendTrafficPolicy)
+	gatewayPolicyMerged := sets.New[types.NamespacedName]()
 
 	// Translate
 	// 1. First translate Policies targeting xRoutes
 	// 2. Finally, the policies targeting Gateways
 
-	// Build gateway policy map
+	// Build gateway policy map, which is needed when processing the policies targeting xRoutes.
 	for _, currPolicy := range backendTrafficPolicies {
 		targetRefs := getPolicyTargetRefs(currPolicy.Spec.PolicyTargetReferences, gateways)
 		for _, currTarget := range targetRefs {
@@ -124,7 +125,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 				// policy overrides and populate its ancestor status.
 				parentRefs := GetParentReferences(route)
 				ancestorRefs := make([]gwapiv1a2.ParentReference, 0, len(parentRefs))
-				routeGateways := sets.New[types.NamespacedName]()
+				routeParents := sets.New[types.NamespacedName]()
 				for _, p := range parentRefs {
 					if p.Kind == nil || *p.Kind == resource.KindGateway {
 						namespace := route.GetNamespace()
@@ -135,7 +136,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 							Namespace: namespace,
 							Name:      string(p.Name),
 						}
-						routeGateways.Insert(gwNN)
+						routeParents.Insert(gwNN)
 
 						key := gwNN.String()
 						if _, ok := gatewayRouteMap[key]; !ok {
@@ -172,7 +173,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 					}
 				} else {
 					// merge with parent target policy if exists
-					for _, gwNN := range routeGateways.UnsortedList() {
+					for _, gwNN := range routeParents.UnsortedList() {
 						// find policy for Gateway
 						gwPolicy := gatewayPolicyMap[gwNN]
 						if gwPolicy == nil {
@@ -199,6 +200,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 							continue
 						}
 
+						gatewayPolicyMerged.Insert(gwNN)
 						status.SetConditionForPolicyAncestors(&policy.Status,
 							ancestorRefs,
 							t.GatewayControllerName,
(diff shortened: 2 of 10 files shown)


internal/utils/merge.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
+)
+
+func Merge[T client.Object](original T, patch T, mergeType egv1a1.MergeType) (T, error) {
```
arkodg, line 19: non blocking comment: would be good to dedup this and https://github.com/envoyproxy/gateway/blob/645b65fa16c6ed924bcb5c9abbdcd4a60e64e787/api/v1alpha1/kubernetes_helpers.go#L177 in a follow up

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -96,6 +124,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 				// policy overrides and populate its ancestor status.
 				parentRefs := GetParentReferences(route)
 				ancestorRefs := make([]gwapiv1a2.ParentReference, 0, len(parentRefs))
+				routeGateways := sets.New[types.NamespacedName]()
```
arkodg, line 127: nit: can we call this `routeParents` to make it easier to understand ?

arkodg: can't we reuse parentRefs ? (trying to think of ways to reduce men)

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -68,10 +68,38 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
...
+					Name:      string(currTarget.Name),
+					Namespace: currPolicy.Namespace,
+				}
+				gateway, ok := gatewayMap[key]
+				if !ok {
+					continue
+				}
+
+				// Check if policy for gateway exists
+				gatewayNN := utils.NamespacedName(gateway)
```
arkodg, line 93: this is a tricky line, this logic here is a pre processing step where we are building some state
however, we haven't yet computed which policies targeting a gateway have been accepted,
is this logic of `if exists, skip` safe here ? thinking out loud - it may be since we are sorting based on timestamp on L45

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -68,10 +68,38 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 
 	handledPolicies := make(map[types.NamespacedName]*egv1a1.BackendTrafficPolicy)
 
+	gatewayPolicyMap := make(map[types.NamespacedName]*egv1a1.BackendTrafficPolicy)
+
 	// Translate
 	// 1. First translate Policies targeting xRoutes
 	// 2. Finally, the policies targeting Gateways
 
+	// Build gateway policy map
```
arkodg, line 77: hey can an extra comment here be adding highlighting the why - to make it easier to access the parent policy targeting the gateway that also directs affects the child resource

arkodg:
thanks @zirain, this looks good ! added some minor comments
ptal @guydc @liorokman @zhaohuabing

internal/gatewayapi/backendtrafficpolicy.go
zirain: if all the parents are gateway, we can reuse it.

internal/gatewayapi/backendtrafficpolicy.go
zirain: the first loop is very similar to the third one(the gateway one), but it's hard to merge them into one AFAICT.

internal/gatewayapi/backendtrafficpolicy.go
arkodg: got it, let's keep the extra var

internal/gatewayapi/backendtrafficpolicy.go
arkodg: Yeah I understand, if you can add a comment here around why it's correct/okay, should help the next person looking at this code

internal/utils/merge.go
zirain: https://github.com/envoyproxy/gateway/issues/5849

internal/gatewayapi/backendtrafficpolicy.go
zirain: we may need to review/refactor the whole thing in `v1.5`.

internal/gatewayapi/testdata/backendtrafficpolicy-strategic-merge.out.yaml
```diff
@@ -0,0 +1,278 @@
...
+    - ancestorRef:
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: gateway-1
+        namespace: envoy-gateway
+        sectionName: http
+      conditions:
+      - lastTransitionTime: null
+        message: Merged with policy envoy-gateway/policy-for-gateway
+        reason: Overridden
```
zhaohuabing, line 27: Should we use `Merged` instead of `Overridden` ?

zirain: I have no strong opinion.

internal/gatewayapi/testdata/backendtrafficpolicy-strategic-merge.out.yaml
```diff
@@ -0,0 +1,278 @@
...
+        kind: Gateway
+        name: gateway-1
+        namespace: envoy-gateway
+      conditions:
+      - lastTransitionTime: null
+        message: Policy has been accepted.
+        reason: Accepted
+        status: "True"
+        type: Accepted
+      controllerName: gateway.envoyproxy.io/gatewayclass-controller
```
zhaohuabing, line 66: Should we also add `Merged` condition in the BTP for Gateway?

internal/gatewayapi/testdata/backendtrafficpolicy-strategic-merge.out.yaml
zhaohuabing: I think `Merged` is a better fit here, as `Overriden` doesn't accurately describe the behavior.

arkodg: +1 to Merged, and we can improve this later in v1.5 as part of https://github.com/envoyproxy/gateway/issues/5745

zirain: https://github.com/envoyproxy/gateway/pull/5795/commits/da4ee6cc598e97bfba09a89a87c68c37780978e1

internal/gatewayapi/testdata/backendtrafficpolicy-strategic-merge.out.yaml
```diff
@@ -0,0 +1,314 @@
...
+          these routes: [default/httproute-1]'
+        reason: Merged
+        status: "True"
+        type: Merged
+      - lastTransitionTime: null
+        message: 'This policy is being overridden by other backendTrafficPolicies
+          for these routes: [default/grpcroute-1]'
+        reason: Overridden
+        status: "True"
+        type: Overridden
```
zhaohuabing, line 105: We don't need this anymore?

zirain: this's another BTP without `merge`.

internal/gatewayapi/testdata/backendtrafficpolicy-strategic-merge.in.yaml
```diff
@@ -0,0 +1,92 @@
...
+    metadata:
+      namespace: default
+      name: merged-policy-for-route
+    spec:
+      targetRef:
+        group: gateway.networking.k8s.io
+        kind: HTTPRoute
+        name: httproute-1
+      mergeType: StrategicMerge
+      connection:
```
arkodg, line 79: can we add 2 other fields in here to highlight the following -
* scalar field on route policy will replace gateway policy field
* for list fields, route values will come after gateway values

zirain: updated, also add UT in merge.go

arkodg (approved):
LGTM thanks, lets track the list merging piece as a separate issue

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -85,8 +85,11 @@ type BackendTrafficPolicySpec struct {
 	// HTTPUpgrade defines the configuration for HTTP protocol upgrades.
 	// If not specified, the default upgrade configuration(websocket) will be used.
 	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	//
 	// +optional
-	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty"`
+	HTTPUpgrade []*ProtocolUpgradeConfig `json:"httpUpgrade,omitempty" patchMergeKey:"type" patchStrategy:"merge"`
```
zirain, line 92: @arkodg we need this tag for every slice we want to merge.

zhaohuabing (approved):

Issue #5805, opened 2025-04-25
Options to configure client-side rate limiter


dghubble:
*Description*:

Running 3 envoy-gateway controllers with leader election, I'm often seeing the gateway's rate limit themselves ("client rate limiter"). There does not seem to be a way to configure the client side qps and burst like might be found in other controllers.

```
envoy-gateway 2025-04-25T01:15:12.148Z    ERROR    provider    leaderelection/leaderelection.go:429    Failed to update lock optimistically: client rate limiter Wait returned an error: context deadline exceeded, falling back to slow path    {"runner": "provider"}                                                     │
│ envoy-gateway 2025-04-25T01:15:12.149Z    ERROR    provider    leaderelection/leaderelection.go:436    error retrieving resource lock ingress/5b9825d2.gateway.envoyproxy.io: client rate limiter Wait returned an error: context deadline exceeded    {"runner": "provider"}                                               │
│ envoy-gateway 2025-04-25T01:15:12.149Z    INFO    provider    leaderelection/leaderelection.go:297    failed to renew lease ingress/5b9825d2.gateway.envoyproxy.io: context deadline exceeded    {"runner": "provider"}
```

Another option might be to extend the leader election leases so they don't have to hit the apiserver so often I suspect. 

```
leaderElection:
  leaseDuration: 15s
  renewDeadline: 10s
  retryPeriod: 2
```

With the defaults, leader election seems to rate limit itself often enough to cause gateway pod restarts and churn.  Is running with multiple replicas, but no election valid? It's unfortunate that the main cause of gateways going down is their own leader election mechanism it seems.

>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

arkodg:
cc @alexwo

dghubble:
Extending the leases does seem to help (to force the clients to request less often) reduce frequency of leader changes, but with the tradeoff of less resiliency in outages I guess.

```
                leaseDuration = "90s"
                renewDeadline = "60s"
                retryPeriod   = "5s"
```

Issue #5089, opened 2025-01-18
Allow tracing sampleRate smaller than 1%


dghubble:
*Description*:

Envoy Gateway's `EnvoyProxy` CustomResourceDefinition expects the tracing field's `sampleRate` to be an integer, allowing 0 to 100.

I'd like to keep tracing on, but sample perhaps 0.01% of traffic. This does not seem to be directly possible, although envoy-proxy itself [seems to support](https://github.com/envoyproxy/gateway/blob/0ee0a8af8e0a86bfbafbca4cee6d5432afa07830/internal/ir/xds.go#L2210) float64 sample rates.

```
# desired
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
spec:
  telemetry:
    tracing:
      sampleRate: 0.01
...
```

[optional *Relevant Links*:]

https://gateway.envoyproxy.io/latest/api/extension_types/#proxytelemetry

> SamplingRate controls the rate at which traffic will be
selected for tracing if no prior sampling decision has been made.
Defaults to 100, valid values [0-100]. 100 indicates 100% sampling

arkodg:
cc @zirain , we may have been trying to avoid using floats in the CRD in the past, but this is a legit use case, and recently the faultInjection API used floats https://github.com/envoyproxy/gateway/blob/dd1c66fb024e1ce76f59fe1480900307e9ae480d/api/v1alpha1/fault_injection.go#L37

zirain:
TBH, is that real meaningful in real world?

arkodg:
gateway-api's mirroring filter introduced an addition `fraction` field to solve this https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRequestMirrorFilter, we could have something similar here

dghubble:
> is that real meaningful in real world?

With a lot of traffic, 1% is a ton of traces. A fraction would help because even with 0.01% is doesn't take long to find traces with errors or p99s worth looking into.

zirain:
you may not get help from traing with a very low sample rate, you may need CEL based sampler.
With CEL based sampler, you're able to sample only `response.code != 200`.

dghubble:
Im just looking to sample a smaller percentage of all traffic, not do expressions. Success traces still have value to me too

zirain:
sgtm

dghubble:
Amazing, thanks!

PR #5815, opened 2025-04-25
feat(controller): make Kubernetes client QPS and Burst configurable

alexwo:
Adds an option to configure the k8s client rate limiter.

Fixes: https://github.com/envoyproxy/gateway/issues/5805


api/v1alpha1/envoygateway_types.go
```diff
@@ -91,6 +91,23 @@ type EnvoyGatewaySpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	QPS float32 `json:"qps,omitempty"`
+
+	// Burst defines the burst limit for the Kubernetes client.
+	// If unspecified, defaults to 100. Min value must be greater than or equal to 1.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	Burst int `json:"burst,omitempty"`
```
arkodg, line 108:
```suggestion
	Burst *int `json:"burst,omitempty"`
```

api/v1alpha1/envoygateway_types.go
```diff
@@ -91,6 +91,23 @@ type EnvoyGatewaySpec struct {
...
 }
 
+// KubernetesClientRateLimit defines the rate limit settings for the Kubernetes client.
+type KubernetesClientRateLimit struct {
+	// QPS defines the queries per second (QPS) limit for the Kubernetes client.
+	// If unspecified, defaults to 50. Min value must be greater than or equal to 1.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	QPS float32 `json:"qps,omitempty"`
```
arkodg, line 101: can we make this `*int` ?

guydc, line 101: I think that we avoided floats so far in the API, as mentioned here: https://github.com/envoyproxy/gateway/issues/5089. 

In this case, is it really meaningful to use a float over an int?

arkodg:
thanks for builidng this out so quickly @alexwo 🎉 
adding some comments around field types, overall LGTM !

api/v1alpha1/envoygateway_types.go
```diff
@@ -225,6 +242,11 @@ type EnvoyGatewayKubernetesProvider struct {
 	// ShutdownManager defines the configuration for the shutdown manager.
 	// +optional
 	ShutdownManager *ShutdownManager `json:"shutdownManager,omitempty"`
+
+	// ClientRateLimit defines the rate limit for the Envoy Gateway Kubernetes client.
+	// If it's not set, defaults to 50 QPS and 100 Burst.
+	// +optional
+	ClientRateLimit *KubernetesClientRateLimit `json:"clientRateLimit,omitempty"`
```
arkodg, line 249: another thought here is - `Client.RateLimit` vs `ClientRateLimit`, would there be more things worth adding under `Client` in the future

alexwo: Indeed , i have updated it.

api/v1alpha1/envoygateway_types.go
```diff
@@ -91,6 +91,28 @@ type EnvoyGatewaySpec struct {
...
+	RateLimit *KubernetesClientRateLimit `json:"rateLimit,omitempty"`
+}
+
+// KubernetesClientRateLimit defines the rate limit settings for the Kubernetes client.
+type KubernetesClientRateLimit struct {
+	// QPS defines the queries per second (QPS) limit for the Kubernetes client.
+	// If unspecified, defaults to 50. Min value must be greater than or equal to 1.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
arkodg, line 105: adding the 50 and 100 default tags here will show up in the docs under the default coloumn

arkodg:
the client settings should also be applied to the infra client https://github.com/envoyproxy/gateway/blob/a2d39738d9fcfd7562d9109cb24e79d1e6d3292f/internal/infrastructure/manager.go#L59

arkodg (dismissed):
LGTM thanks !

alexwo:
> the client settings should also be applied to the infra client
> 
> https://github.com/envoyproxy/gateway/blob/a2d39738d9fcfd7562d9109cb24e79d1e6d3292f/internal/infrastructure/manager.go#L59

good catch, updated

arkodg (approved):

zirain (approved):

PR #5649, opened 2025-04-01
build: fix permission issue

daixiang0:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #5648

Release Notes: Yes/No

PR #5819, opened 2025-04-25
fix: adapt permissions for certgen

miltalex:
**What type of PR is this?**

This PR introduces a fix following the permissions changes in the build Dockerfile as part of the https://github.com/envoyproxy/gateway/pull/5649 

Fixes #

Release Notes: No


arkodg:
thanks for finding this @miltalex, can't figure out why CI was working fine these few weeks after the PR was merged

also I dont remember why `65534` was set for cert-gen so setting it `65532` and keep is consistent with the `envoy-gateway` deployment feels right

the third issue here is, do we need to rewrite the docker file so the `User` can be overwritten by k8s, I dont want to block this PR to tackle this issue though

arkodg (approved):
LGTM thanks !

miltalex:
> thanks for finding this @miltalex, can't figure out why CI was working fine these few weeks after the PR was merged
> 
> also I dont remember why `65534` was set for cert-gen so setting it `65532` and keep is consistent with the `envoy-gateway` deployment feels right
> 
> the third issue here is, do we need to rewrite the docker file so the `User` can be overwritten by k8s, I dont want to block this PR to tackle this issue though

Thanks happy to tackle the third issue in a seperate PR

arkodg:
@miltalex raised https://github.com/envoyproxy/gateway/pull/5820 to tackle the third issue

zirain (approved):

arkodg:
waiting on #5352 before merging this one to avoid conflicts

Issue #5561, opened 2025-03-20
Gateway context in Translation hook


muwaqar:
*Description*:
The current [xDS Translation hook](https://gateway.envoyproxy.io/contributions/design/extending-envoy-gateway/#post-xds-translation-modify-hook) has an empty context.

Sometimes some mutation needs to be done on Clusters based on Gateway policies. For example, if a Gateway supports mTLS via SPIFFE, we want to be able to add a SPIRE agent as an upstream cluster.

Currently it is not possible to know which Gateway xDS is being mutated so this conditional mutation is not possible.

EG already does a great job of custom Gateway policy attachments that it passes to the [Listener modification hook](https://gateway.envoyproxy.io/contributions/design/extending-envoy-gateway/#http-listener-modification-hook). The proposal here is to pass the `ExtensionResource`(s) attached for the Gateway to the Translation hook as well.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

guydc:
cc @liorokman

guydc:
Hi @muwaqar. WDYT about adding a cluster streaming hook that will have extension `HTTPRouteFilter` context, similar to how route rule extension filters drive xds route mutation in extension service?

guydc:
After carefully re-reading your request, this does not solve the problem of adding the SPIRE cluster itself. Rather, it facilitates conditional and contextual mutation of specific clusters that use SPIRE.

muwaqar:
closing this since https://github.com/envoyproxy/gateway/pull/5914 is completed

Issue #5523, opened 2025-03-17
Envoy xDS metadata for `Cluster` and `LocalityLbEndpoints` resources


muwaqar:
*Description*:
Currently EnvoyGateway propagates xDS metadata for Route and VirtualHost resources, but not for Cluster and LbEndpoints.

Note that the [main documentation page](https://gateway.envoyproxy.io/contributions/design/metadata/) for this feature currently says the following:

> [Service](https://kubernetes.io/docs/concepts/services-networking/service/), [ServiceImport](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/#serviceimport-and-endpointslices) and [Backend](https://gateway.envoyproxy.io/latest/api/extension_types/#backend) metadata and port name are propagated to envoy [cluster](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto) metadata.

This is incorrect, as evident from the [code](https://github.com/envoyproxy/gateway/blob/292894057fd4083b1c4dce691d510c1a6bc53073/internal/xds/translator/cluster.go#L86) since `buildXdsCluster()` never sets any metadata.

The proposal in this ticket is to:

1) `Route` metadata be propagated to `Cluster` xDS.

Each `*RouteRule` creates a distinct xDS cluster which encapsulates upstream endpoints of its `backendRefs`.

Accordingly, maybe we can put the `*Route` metadata into the `Cluster` xDS?

For example:

For a HTTPRoute as:
```
kind: HTTPRoute
apiVersion: gateway.networking.k8s.io/v1
metadata:
  annotations:
    gateway.envoyproxy.io/foo: bar
  name: myroute
  namespace: myns
spec:
  rules:
    matches:
    - path:
        type: PathPrefix
        value: /mypath
```

we get:
```
name: httproute/myns/myroute/rule/0
type: CDS
[...]
metadata:
  filter_metadata:
    envoy-gateway:
      resources:
        - namespace: myns
          groupVersion: gateway.networking.k8s.io/v1
          kind: HTTPRoute
          annotations:
            foo: bar
          name: myroute
``` 

2) `BackendReference` (`Service`, `ServiceImport`, `Backend`) metadata be propagated to `LocalityLbEndpoints` xDS. 

If a `*RouteRule` contains multiple `BackendReference`, each of these are modeled as separate `LocalityLbEndpoints`. 

[...]

PR #5821, opened 2025-04-25
api: extension post cluster modify hook

guydc:
**What type of PR is this?**

Adds an option for extension server to mutate clusters with context derived from GRPCRoute/HTTPRoute BackendRefs ExtensionRef HTTPRouteFilters. 

Since Upstream HTTP Filters are mapped to cluster-level [HTTPProtocolOptions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto#extensions-upstreams-http-v3-httpprotocoloptions) in envoy, it makes most sense to aggregate BackendRef ExternalRefs and provide them as context for cluster mutation (as opposed to CLA/Endpoint mutation). 

Relates to #5561 

Release Notes: Yes/No


proto/extension/context.proto
```diff
@@ -54,3 +54,11 @@ message PostTranslateExtensionContext {
 message ExtensionResource {
     bytes unstructured_bytes = 1;
 }
+
+
+// PostClusterExtensionContext provides resources introduced by an extension and watched by Envoy Gateway
+// additional context information can be added to this message as more use-cases are discovered
+message PostClusterExtensionContext {
+    // Resources introduced by the extension that were used as extensionRefs in an HTTPRoute/GRPCRoute BackendRefs
```
TheRushingWookie, line 62: Hi! Quinn here, we met last week.

Will one of these ExtensionResources be the gateway CR that is currently used in the xDS translation?

The backrefs by themselves give us the HTTPRoutes but the parentsRefs of those may be all the gateways instead of the current gateway.

proto/extension/service.proto
```diff
@@ -103,3 +111,17 @@ message PostTranslateModifyResponse {
...
+message PostClusterModifyRequest {
+    envoy.config.cluster.v3.Cluster cluster = 1;
+    PostClusterExtensionContext post_cluster_context = 2;
+}
+
+
+// PostClusterModifyResponse is the expected response from an extension and contains a modified version of the Cluster that was sent
+// If an extension returns a nil Cluster then it will not be modified
+message PostClusterModifyResponse {
+    envoy.config.route.v3.Route route = 1;
```
TheRushingWookie, line 126: Should this return  a cluster?

guydc: yes! my bad.

proto/extension/context.proto
guydc: To support additional use cases (besides extension server), I was thinking of propagating the GW context in XDS Metadata. That way, the information is also available to dataplane filters and extensions. The extension server can also parse this metadata. 
So, first step would be to complete #5523, and then add an opt-in option for including metadata from additional related resources (GW, Route, ... ).

Issue #5547, opened 2025-03-19
docs: Add Install and Upgrade instructions for ArgoCD


arkodg:
*Description*:
>Describe the issue.

Add steps for Installing EG https://gateway.envoyproxy.io/docs/install/ using Argo CD, to avoid issues like https://github.com/envoyproxy/gateway/issues/5546

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Issue #5546, opened 2025-03-19
Envoy Gateway Installation with Argo CD Fails Due to Annotation Size Limit


djbulsink:
*Description*:
When installing Envoy Gateway with Argo CD, the following error occurs:  

`error when patching "/dev/shm/3535441978": CustomResourceDefinition.apiextensions.k8s.io "envoyproxies.gateway.envoyproxy.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes`

This error is caused by the large size of the manifest. When applying the manifest using `kubectl apply`, the limit of **262144 bytes** is not reached. However, when using **Argo CD** or `kubectl apply` with additional annotations, the manifest exceeds this limit.

*Repro steps*:

###  Using Argo CD  
Create an **Application** resource with the following configuration:  

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: envoy-gateway
  namespace: operators
spec:
  destination:
    namespace: envoy-gateway
    server: https://kubernetes.default.svc
  project: default
  source:
    chart: gateway-helm
    helm:
      valuesObject: {}
    repoURL: docker.io/envoyproxy
    targetRevision: v1.3.1
  syncPolicy:
    syncOptions:
    - CreateNamespace=true
    - RespectIgnoreDifferences=true
```

### Using kubectl

The issue is also reproducible by applying the EnvoyProxies Custom Resource Definition (CRD) directly using kubectl apply:

```bash
kubectl apply -f https://raw.githubusercontent.com/envoyproxy/gateway/refs/heads/release/v1.3/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_envoyproxies.yaml
```

This results in the following error:
```
The CustomResourceDefinition "envoyproxies.gateway.envoyproxy.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
```

*Environment*:
Helm chart version: v1.3.1

*Logs*:

```
error when patching "/dev/shm/3535441978": CustomResourceDefinition.apiextensions.k8s.io "envoyproxies.gateway.envoyproxy.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
```

arkodg:
@djbulsink you'll also need to add `— ServerSideApply=true`
let us know if you hit any other issues 

I'll raise another issue to add support for installing with Argo CD

Issue #5223, opened 2025-02-06
job/eg-gateway-helm-certgen


zhujiangtao123:
helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.3.0 -n envoy-gateway-system --create-namespace
Pulled: docker.io/envoyproxy/gateway-helm:v1.3.0
Digest: sha256:9f142766a8388ebaa27d1afa9cc9ffb717d6a6fc96c6118d71642ba6522100dd
Error: INSTALLATION FAILED: failed pre-install: 1 error occurred:
	* timed out waiting for the condition

job/eg-gateway-helm-certgen  exec is failed

owenhaynes:
I see this issue with v1.3.0 

````
2025-02-06T10:24:20.133Z        INFO    cmd/certgen.go:65       generated certificates
Error: failed to output certificates: failed to create or update secrets: failed to get secret envoy-gateway-system/envoy-gateway: failed to get server groups: the server has asked for the client to provide credentials
Usage:
  envoy-gateway certgen [flags]

Flags:
  -h, --help        help for certgen
  -l, --local       Generate all the certificates locally.
  -o, --overwrite   Updates the secrets containing the control plane certs.

failed to output certificates: failed to create or update secrets: failed to get secret envoy-gateway-system/envoy-gateway: failed to get server groups: the server has asked for the client to provide credentials
```

arkodg:
cc @guydc

guydc:
Clean installation on a fresh cluster seems to work:

> helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.3.0 -n envoy-gateway-system --create-namespace
> 
> Pulled: docker.io/envoyproxy/gateway-helm:v1.3.0
> Digest: sha256:9f142766a8388ebaa27d1afa9cc9ffb717d6a6fc96c6118d71642ba6522100dd
> NAME: eg
> LAST DEPLOYED: Thu Feb  6 16:12:59 2025
> NAMESPACE: envoy-gateway-system
> STATUS: deployed
> REVISION: 1
> TEST SUITE: None
> NOTES:
> **************************************************************************
> *** PLEASE BE PATIENT: Envoy Gateway may take a few minutes to install ***
> **************************************************************************
> 
> Envoy Gateway is an open source project for managing Envoy Proxy as a standalone or Kubernetes-based application gateway.
> 
> Thank you for installing Envoy Gateway! 🎉
> [...]

[...]

PR #5824, opened 2025-04-26
docs: install EG via Argo CD

zhaohuabing:
Closes: #5547
Related: #5223 #5546


site/content/en/latest/install/install-argocd.md
```diff
@@ -0,0 +1,164 @@
...
+  namespace: argocd
+spec:
+  project: default
+  source:
+    helm:
+      valuesObject:
+        certgen:
+          rbac:
+            annotations:
+              argocd.argoproj.io/sync-wave: "-1"  # Ensure rbac is created before the certgen job. This is a workaround for
```
zhaohuabing, line 117: I might be wrong, but according to my test, OCI helm repos don't work with the `valueFiles` option.

site/content/en/latest/install/install-yaml.md
```diff
@@ -55,3 +51,30 @@ kubectl apply --force-conflicts --server-side -f ./gateway-helm/crds/generated
 ```shell
 helm upgrade eg oci://docker.io/envoyproxy/gateway-helm --version {{< yaml-version >}} -n envoy-gateway-system
 ```
+
+## Open Ports
```
zirain, line 55: Not a must, can you make this into `boilerplates`?

site/content/en/latest/install/install-yaml.md
```diff
@@ -11,23 +11,19 @@ installation, it is recommended that you use helm.
...
-* The `kubectl` command-line tool
-
 {{% alert title="Compatibility Matrix" color="warning" %}}
 Refer to the [Version Compatibility Matrix](/news/releases/matrix) to learn more.
 {{% /alert %}}
 
-## Install with YAML
-
-Envoy Gateway is typically deployed to Kubernetes from the command line. If you don't have Kubernetes, you should use `kind` to create one.
+Envoy Gateway is typically deployed in a Kubernetes cluster.
```
zirain, line 18: Not a must, can you make this into boilerplates?

zhaohuabing: Let's keep it for now?  It's just two simple sentences.  If this becomes more complex in the future, then we can move it to a bolierplate.

zirain (dismissed):

site/content/en/latest/install/install-argocd.md
illrill: Check this out: https://argo-cd.readthedocs.io/en/latest/user-guide/multiple_sources/#helm-value-files-from-external-git-repository

site/content/en/latest/install/install-argocd.md
```diff
@@ -0,0 +1,139 @@
...
+  source:
+    chart: gateway-helm
+    repoURL: docker.io/envoyproxy
+    targetRevision: {{< helm-version >}}
+    helm:
+      valuesObject:
+        certgen:
+          rbac:
+            annotations:
+              argocd.argoproj.io/sync-wave: "-1"  # Ensure rbac is created before the certgen job.
```
arkodg, line 42: https://github.com/envoyproxy/gateway/pull/5352 hit this issue and set `"helm.sh/hook-weight": "-1"`, can we add that first into cert gen and avoid this `patch` setting in argo install docs ?

zirain:
is this possible to add some test for argocd?

zirain:
A quick idea: we can run one of our conformance/e2e with installation of ArgoCD, WDYT?

zhaohuabing:
> A quick idea: we can run one of our conformance/e2e with installation of ArgoCD, WDYT?

It may not be necessary. Argo CD should be able to install any helm chart, EG is not special here.

site/content/en/latest/install/install-argocd.md
```diff
@@ -0,0 +1,128 @@
...
+spec:
+  project: default
+  source:
+    chart: gateway-helm
+    repoURL: docker.io/envoyproxy
+    targetRevision: {{< helm-version >}}
+  destination:
+    namespace: envoy-gateway-system
+    server: https://kubernetes.default.svc
+  syncPolicy:
```
shamusx, line 40: The following does the same as `argocd app sync envoy-gateway`, but not all users have the `cli`
```suggestion
  syncPolicy:
    automated: {}
```

site/content/en/v1.3/install/install-yaml.md
```diff
@@ -38,9 +29,9 @@ Refer to the [Developer Guide](../../contributions/develop) to learn more.
 
    Envoy Gateway should now be successfully installed and running, but in order to experience more abilities of Envoy Gateway, you can refer to [Tasks](/latest/tasks).
 
-## Upgrading from v1.2
+## Upgrading from a previous version
```
arkodg, line 32: the actual previous version here matters, we dont garuntee upgrades for N-2 to N yet

site/content/en/latest/install/install-argocd.md
arkodg: Chatgpt enhanced this suggestion with
```
    automated:
      prune: true
      selfHeal: true
```

is that needed ?

site/content/en/v1.3/install/install-yaml.md
zhaohuabing: What about `the previous version`?  Then we don't have to keep updating the doc after each release.

site/content/en/latest/install/install-argocd.md
zhaohuabing: I think it's up to the user to decide how they want to sync Envoy Gateway application - they may have a consistent sync policy for all their applications.

We can use automated sync as an example here, as it better aligns with the GitOps workflow.

shamusx (approved):

cnvergence (approved):

PR #5829, opened 2025-04-27
chore: exclue go-jose/v4

zirain:


.github/workflows/build_and_test.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574  # v5.4.0
       with:
-        fail_ci_if_error: true
+        fail_ci_if_error: false  # don't fail the CI if codecov fails
```
shahar-h, line 60: codecov-action [v5.4.2](https://github.com/codecov/codecov-action/releases/tag/v5.4.2) fixes the oidc issue so you should just merge https://github.com/envoyproxy/gateway/pull/5773 instead of changing this.

PR #5833, opened 2025-04-27
Use cluster.loadBalancingPolicy instead of commonLbConfig

### LINKED ISSUE #5834: Use cluster.loadBalancingPolicy to configure load balancer instead of commonLbConfig
*Description*:
CommonLbConfig only exposes the "legacy" implementation for LocalityLbConfig and can't be configured with all available options (for example Zone Aware Lb Config and `ForceLocalityDirectRouting`). The loadBalancingPolicy field is more feature-rich and should be the default.

Common config for Zone Aware Lb:
* https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-msg-config-cluster-v3-cluster-commonlbconfig-zoneawarelbconfig
Locality Lb config for Zone Aware Lb:
* https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#envoy-v3-api-msg-extensions-load-balancing-policies-common-v3-localitylbconfig-zoneawarelbconfig

jukie:
**What type of PR is this?**

Uses cluster.loadBalancingPolicy instead of commonLbConfig for access to all available fields during load balancer configuration.

Fixes #5834

Release Notes: No


zirain (approved):

zirain (dismissed):

jukie:
I'll reopen after some other pending MRs

arkodg:
thanks @jukie, lets get this PR into v1.5

jukie:
This doesn't depend on any upstream changes so we could still get it into 1.4.x but just planning until after some of the other planned changes are in to avoid conflicts

jukie:
Would like to start implementation for LoadBalancer.ZoneAware from https://github.com/envoyproxy/gateway/pull/6154 but wanted to get this in first.

internal/xds/translator/testdata/out/xds-ir/http-route-dynamic-resolver.clusters.yaml
```diff
@@ -9,11 +9,18 @@
...
   dnsLookupFamily: V4_PREFERRED
   lbPolicy: CLUSTER_PROVIDED
+  loadBalancingPolicy:
+    policies:
+    - typedExtensionConfig:
+        name: envoy.load_balancing_policies.least_request
+        typedConfig:
+          '@type': type.googleapis.com/envoy.extensions.load_balancing_policies.least_request.v3.LeastRequest
+          localityLbConfig:
+            localityWeightedLbConfig: {}
```
jukie: Hey @zhaohuabing do you know why dynamic-resolver isn't compatible with `cluster.loadBalancingPolicy`? Looking at git blame I see you added this exclusion - https://github.com/envoyproxy/gateway/blame/main/internal/xds/translator/cluster.go#L341

jukie: I think I see the issue, I need to set loadBalancingPolicy back to `CLUSTER_PROVIDED`

jukie: That did it!

zirain (dismissed):

internal/xds/translator/cluster.go
```diff
@@ -228,45 +225,110 @@ func buildXdsCluster(args *xdsClusterArgs) (*buildClusterResult, error) {
...
+					TypedConfig: typedRoundRobin,
+				},
+			}},
+		}
 	case args.loadBalancer.Random != nil:
 		cluster.LbPolicy = clusterv3.Cluster_RANDOM
+		random := &randomv3.Random{
+			LocalityLbConfig: localityLbConfig,
+		}
+		typeRandom, _ := anypb.New(random)
```
arkodg: can we use https://github.com/envoyproxy/gateway/blob/c455a6e6158667f5d8b2036416860c45da479371/internal/utils/proto/proto.go#L49 which adds an additional validation step

internal/xds/translator/cluster.go
```diff
@@ -228,45 +225,110 @@ func buildXdsCluster(args *xdsClusterArgs) (*buildClusterResult, error) {
...
+	// Set default localityLbConfig
+	localityLbConfig := &commonv3.LocalityLbConfig{
+		LocalityConfigSpecifier: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig_{
+			LocalityWeightedLbConfig: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig{},
+		},
+	}
+
+	// Override LocalityWeightedLbConfig if zone aware routing is enabled.
+	// Zone aware enabled backendRefs always have a single DestinationSetting per-cluster.
+	if len(args.settings) == 1 && args.settings[0].ZoneAwareRouting != nil {
```
arkodg, line 237: not seeing `len(args.settings) == 1` in the existing logic, is this needed ?

jukie: Previously this was a for loop but achieved the same result - https://github.com/envoyproxy/gateway/pull/5833/files/aca60518e46305027c12444967a9216eee7e0ef1?file-filters%5B%5D=.go&show-viewed-files=true#diff-d0da16bff73f9d98d6346ffcf878a7bf6c689dbe11a8176598319ddb0b2a7a0dL347-L352

internal/xds/translator/cluster.go
jukie: Fixed

PR #5835, opened 2025-04-27
Add metrics for topologyInjector webhook

jukie:
**What type of PR is this?**

Folllow-up to https://github.com/envoyproxy/gateway/pull/5352

Fixes #

Release Notes: No


jukie:
CC @zirain

zirain (dismissed):

shawnh2 (commented):
It would be better if you can add this metric in a new section under https://github.com/envoyproxy/gateway/blob/main/site/content/en/latest/tasks/observability/gateway-exported-metrics.md

jukie:
Not sure how to go about adding tests for the metrics change. Is that required for merging?

zirain (approved):

zirain:
> Not sure how to go about adding tests for the metrics change. Is that required for merging?

it would be better to have, but not a blocker.

it shouldn't be very hard to add in the existing zone e2e.

PR #5848, opened 2025-04-29
Ensure rbac is created before the certgen job when using ArgoCD

zhaohuabing:
Fixes: #5223

### DIFF
--- a/charts/gateway-helm/templates/certgen-rbac.yaml
+++ b/charts/gateway-helm/templates/certgen-rbac.yaml
@@ -10,7 +10,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install
-    "argocd.argoproj.io/sync-wave": "-1"  # Ensure rbac is created before the certgen job when using ArgoCD.
+    "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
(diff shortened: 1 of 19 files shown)


charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -110,6 +114,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
     "helm.sh/hook-weight": "-1"
+    "argocd.argoproj.io/sync-wave": "-1"  # Ensure rbac is created before the certgen job when using ArgoCD.
```
arkodg, line 117: instead, can you apply hook-weigh: -1 to the other certgen rbac ?

zhaohuabing: Does Argo CD honor the "helm.sh/hook-weight" annotation?

zhaohuabing: Oh NVM, it does.

charts/gateway-helm/templates/certgen-rbac.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install
```
arkodg, line 12: piggying another request in this PR, can we add `pre-upgrade` to the above line for consistency

arkodg (approved):

zirain (approved):

PR #5868, opened 2025-04-30
Support client cert validation (spki, hash, san)

kinolaev:
**What type of PR is this?**

api: add `publicKeyPins`, `certificateHashes` and `subjectAltNames` fields to ClientTrafficPolicy.spec.tls.clientValidation
feat(translator): translate these fields to tls validation context

This PR allows to validate a client certificate against a list of SKPI hashes, certificate hashes or SANs using `ClientTrafficPolicy`

https://github.com/envoyproxy/gateway/issues/5909

Release Notes: Yes


arkodg:
hey hoping to get to this one in 1.5 weeks post v1.4 release

kinolaev:
Thanks @arkodg , sounds good! I fixed crds and will add some tests to make Codecov happy)

kinolaev:
Hi @arkodg , sorry for bothering you, but I've added tests, could you please rerun the pipeline? I'd like to be sure that Codecov is happy)

Additionally I've created an issue https://github.com/envoyproxy/gateway/issues/5909 with a design proposal for this PR.
I believe it fully complies with your contribution guide now)

kinolaev:
`gen-check` test was broken in the main branch. I rebased the PR, the test must pass now.
I also added a release note.

api/v1alpha1/tls_types.go
```diff
@@ -141,6 +141,52 @@ type ClientValidationContext struct {
 	// +kubebuilder:validation:MaxItems=8
 	// +optional
 	CACertificateRefs []gwapiv1.SecretObjectReference `json:"caCertificateRefs,omitempty"`
+
+	// An optional list of base64-encoded SHA-256 hashes. If specified, Envoy will
+	// verify that the SHA-256 of the DER-encoded Subject Public Key Information
+	// (SPKI) of the presented certificate matches one of the specified values.
+	// +optional
+	PublicKeyPins []string `json:"publicKeyPins,omitempty"`
```
zhaohuabing, line 149: Should we name it `PublicKeyHashes`? 

PublicKeyPins may be better here though.  I'm just speaking from the perspective of a non-native English reader.

zhaohuabing (commented):
Nit: Prefer to add yaml tests to the [gateway API translator](https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata) and [xDS translator](https://github.com/envoyproxy/gateway/tree/main/internal/xds/translator/testdata) instead of individual go tests files.

zhaohuabing (commented):
It's not a must, but it would be great if we could add an e2e test for client cert validation.

api/v1alpha1/tls_types.go
kinolaev: I've chosen the name `PublicKeyPins` because the format `base64(sha256(public key from cert in DER))` comes from [Public Key Pinning Extension for HTTP rfc7469](https://datatracker.ietf.org/doc/html/rfc7469) where they have `Public-Key-Pins` HTTP header. But let me know if you still want me to rename it to `PublicKeyHashes`, I won't argue)

kinolaev:
@zhaohuabing I've added testdata as you requested to:
- internal/gatewayapi/testdata/clienttrafficpolicy-mtls-client-verification.in.yaml
- internal/xds/translator/testdata/in/xds-ir/mutual-tls-san.yaml

Do I have to delete all or some of go tests from the files below?
- internal/gatewayapi/clienttrafficpolicy_test.go
- internal/gatewayapi/helpers_test.go
- internal/xds/translator/listener_test.go

api/v1alpha1/tls_types.go
zhaohuabing: I'd go with `PublicKeyHashes` or `SPKIHashes` because:
* It's easier to understand at a glance, without requiring background knowledge of [Public Key Pinning Extension for HTTP rfc7469](https://datatracker.ietf.org/doc/html/rfc7469). 
* Even though we use the same format of `Public-Key-Pins` HTTP header here, the purpose is different.

cc @envoyproxy/gateway-maintainers please vote for
1. PublicKeyHashes
2. SPKIHashes
3. PublicKeyPins

My vote is either 1 or 2

zhaohuabing:
> @zhaohuabing I've added testdata as you requested to:
> 
> * internal/gatewayapi/testdata/clienttrafficpolicy-mtls-client-verification.in.yaml
> * internal/xds/translator/testdata/in/xds-ir/mutual-tls-san.yaml
> 
> Do I have to delete all or some of go tests from the files below?
> 
> * internal/gatewayapi/clienttrafficpolicy_test.go
> * internal/gatewayapi/helpers_test.go
> * internal/xds/translator/listener_test.go

I think we wouldn't need clienttrafficpolicy_test.go and listener_test.go, but it's fine to keep them.

api/v1alpha1/tls_types.go
kinolaev: I do like SPKIHashes, I've renamed the field. Ready to do it again if the vote suggest a different option)

kinolaev:
> we wouldn't need clienttrafficpolicy_test.go and listener_test.go

Agree, these were duplicates of testdata, I've removed them.

api/v1alpha1/tls_types.go
```diff
@@ -141,6 +141,52 @@ type ClientValidationContext struct {
...
+	// URIs matchers
+	// +optional
+	URIs []StringMatch `json:"uris,omitempty"`
+
+	// Other names matchers
+	// +optional
+	OtherNames []OtherNameMatch `json:"otherNames,omitempty"`
+}
+
+type OtherNameMatch struct {
```
zhaohuabing: Nit: consider renaming it to `OtherSANNameMatch` or`OtherSANMatch` ?
All API types are in the same namespace, a more specific name would be better.

kinolaev: SAN type `OtherName` is defined in [section 4.2.1.6 Subject Alternative Name of rfc5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6). [Envoy](https://github.com/envoyproxy/envoy/blob/v1.34.1/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L340) and [cert-manager](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.OtherName) follow this definition. That is why I'd prefer to not rename the field.

zhaohuabing: OK, OtherNameMatch works for me.

zhaohuabing:
LGTM.

@kinolaev Can you resolve the conflicts and fix the lint?

api/v1alpha1/tls_types.go
kinolaev: Oh, I'm sorry, only now I understand that you asked to rename the type and not the field! I agree I should definitely add "SAN" to the type name, I'll do it shortly.

kinolaev: Renamed to `OtherSANMatch`

kinolaev:
Thank you @zhaohuabing, the conflicts are resolved and CI is green now)

zhaohuabing (approved):
LGTM thanks!

rudrakhp (approved):

zhaohuabing (approved):

Issue #5863, opened 2025-04-30
validate the identity in the Envoy JWT token


zhaohuabing:
Originally posted by @arkodg at https://github.com/envoyproxy/gateway/pull/5137#discussion_r2061000785

zhaohuabing:
cc @cnvergence

PR #5871, opened 2025-04-30
feat: validate JWT token and use projected token

cnvergence:
**What type of PR is this?**

- Validate JWT token against intercepted nodeId and check if the service account name matches irkey stored in the snapshot cache.
- Use the projected service account token and validate against EG audience

Fixes https://github.com/envoyproxy/gateway/issues/5863
Followup to the https://github.com/envoyproxy/gateway/pull/5137

Release Notes: Yes

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
+++ b/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
@@ -239,10 +239,7 @@ spec:
         - envoy
         env:
         - name: ENVOY_GATEWAY_NAMESPACE
-          valueFrom:
-            fieldRef:
-              apiVersion: v1
-              fieldPath: metadata.namespace
+          value: envoy-gateway-system
         - name: ENVOY_SERVICE_ZONE
           valueFrom:
             fieldRef:
@@ -326,10 +323,7 @@ spec:
         - envoy-gateway
         env:
         - name: ENVOY_GATEWAY_NAMESPACE
-          valueFrom:
-            fieldRef:
-              apiVersion: v1
-              fieldPath: metadata.namespace
+          value: envoy-gateway-system
         - name: ENVOY_SERVICE_ZONE
           valueFrom:
             fieldRef:
--- a/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
+++ b/internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
@@ -164,20 +164,6 @@ spec:
                       connection_keepalive:
                         interval: 30s
                         timeout: 5s
-                  http_filters:
-                  - name: envoy.filters.http.credential_injector
-                    typed_config:
-                      "@type": type.googleapis.com/envoy.extensions.filters.http.credential_injector.v3.CredentialInjector
-                      credential:
-                        name: envoy.http.injected_credentials.generic
-                        typed_config:
-                          "@type": type.googleapis.com/envoy.extensions.http.injected_credentials.generic.v3.Generic
-                          credential:
-                            name: jwt-sa-bearer
-                      overwrite: true
-                  - name: envoy.extensions.filters.http.upstream_codec.v3.UpstreamCodec
-                    typed_config:
-                      "@type": type.googleapis.com/envoy.extensions.filters.http.upstream_codec.v3.UpstreamCodec
               name: xds_cluster
               type: STRICT_DNS
               transport_socket:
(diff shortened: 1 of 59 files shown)


internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
```diff
@@ -36,7 +36,6 @@ spec:
         gateway.envoyproxy.io/owning-gateway-name: default
         gateway.envoyproxy.io/owning-gateway-namespace: default
     spec:
-      automountServiceAccountToken: false
```
arkodg, line 39: can we do this only for the gateway mode case ?

internal/xds/server/kubejwt/tokenreview.go
```diff
@@ -56,17 +56,18 @@ func (i *JWTAuthInterceptor) validateKubeJWT(ctx context.Context, token string)
...
 		podName := tokenReview.Status.User.Extra[authPodNameKey]
 		if podName[0] == "" {
 			return fmt.Errorf("pod name not found in token review response")
 		}
-		parts := strings.Split(podName[0], "-")
-		irKey := fmt.Sprintf("%s/%s", parts[1], parts[2])
 
-		if !i.cache.SnapshotHasIrKey(irKey) {
-			return fmt.Errorf("pod %s not found in cache", podName)
+		if podName[0] != proxyMetadata.nodeID {
```
arkodg, line 65: these can also be spoofed, 
instead in the interceptor, Upton successful verification can be add 
ctx := context.WithValue(r.Context(), "envoy-namespace", namespaceInClaim)

and perform a GET later inside https://github.com/envoyproxy/gateway/blob/7afa94ae3c25f3a006c7fb678b11f22cee1f85d8/internal/xds/cache/snapshotcache.go#L177 and validate with `Node.Cluster`  https://github.com/envoyproxy/gateway/blob/7afa94ae3c25f3a006c7fb678b11f22cee1f85d8/internal/infrastructure/common/proxy_args.go#L54 or some other node metadata 
https://github.com/envoyproxy/envoy/blob/9cd454d18227dd9b7b13d5b3051304e4fdd219fb/api/envoy/api/v2/core/base.proto#L132

zhaohuabing:
@cnvergence Talked with @arkodg offline, and he mentioned that we don't need to support `gatewaynamespace` mode for merged gateway. In that case, the `irKey` will always follow the "gateway-name/gateway-namesapce" format, so we should be able to verify the service account in the jwt claim against the `irKey`.

Example:

JWT claim: "username": "system:serviceaccount:default:envoy-default-eg-e41e7b31"
IrKey: "default/eg"

We can construct the expected service account using the `IrKey`, and compare it with the `username` claim.

@arkodg feel free to add anything if I haven't captured our discussion correctly.

internal/infrastructure/kubernetes/proxy/testdata/deployments/ipv6.yaml
arkodg: thanks

internal/xds/server/kubejwt/tokenreview.go
```diff
@@ -56,19 +55,21 @@ func (i *JWTAuthInterceptor) validateKubeJWT(ctx context.Context, token string)
...
+	if len(parts) != 4 {
+		return fmt.Errorf("invalid username format: %s", tokenReview.Status.User.Username)
+	}
+	sa := parts[3]
 
-		if !i.cache.SnapshotHasIrKey(irKey) {
-			return fmt.Errorf("pod %s not found in cache", podName)
+	irKeys := i.cache.GetIrKeys()
+	for _, irKey := range irKeys {
+		if proxy.ExpectedResourceHashedName(irKey) == sa {
```
zhaohuabing, line 69: Nit: Could we add a local `irKeyToServiceAccountName` method in this file instead? I think this would be better because:
* It’s more straightforward and easier to read
* I'd prefer to avoid importing the infra package into the xds package if possible

cnvergence: we can add it, I will add the comment where it came from

zirain (approved):

zhaohuabing (dismissed):
LGTM.

Discussed with @arkodg ,  to prevent an envoy from requesting xDS not belong to it, we also need to enforce authz, which can be addressed in a follow-up PR.

Or you can simply revert b0748a066c9f6a41920df95f728ace5f84ed1acb to add pod check back @cnvergence

zhaohuabing (approved):

zirain (approved):

zhaohuabing (approved):

PR #5883, opened 2025-05-01
skip TCP listener if it has no routes

zhaohuabing:
This PR skips the TCP listener if it has no routes and the port is the hbone port 15008.

Now that EG can deploy Envoy in the Gateway namespace, this is the only remaining blocker for EG to work with Istio Ambient mode as a waypoint proxy.

Reference: https://github.com/istio/istio/blob/9104ca90df74067c7cc9a2bfe4a9ea0190ec6a43/tests/integration/ambient/waypoint_test.go#L273-L277

### DIFF
--- a/internal/xds/translator/translator.go
+++ b/internal/xds/translator/translator.go
@@ -660,6 +660,17 @@ func (t *Translator) processTCPListenerXdsTranslation(
 	// errors and return them at the end.
 	var errs, err error
 	for _, tcpListener := range tcpListeners {
+		// Skip the fake hbone port 15008.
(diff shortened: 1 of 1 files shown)


arkodg:
doesnt feel like the right fix, it breaks other use cases

zirain:
you can achieve this with a patch on service.
i can share a poc demo when i'm back.

zhaohuabing:
> doesnt feel like the right fix, it breaks other use cases

To avoid breaking other use cases, only skip TCP listener without routes if the port is 15008.

This change will be reverted after it's been fixed in Istio.

PR #5894, opened 2025-05-01
disable settings by default in gateway-crds-helm

arkodg:
* These settings dont work by default in the way `helm` works and this helm chart is now mainly used as a package artifact to be consumed by CI tools like Argo, so changed the default settings to disable by default, so users are opting into specific CRDs they want.

Relates to https://github.com/envoyproxy/gateway/pull/5616#issuecomment-2840245509


arkodg:
cc @shahar-h @NissesSenap

arkodg:
@zirain any idea how I can make linter happy ?

arkodg:
> https://github.com/envoyproxy/gateway/tree/9eff541254221963dcacadb02e18b6ac7cb53c73/tools/linter/codespell

thanks i'll use this work around for now, but why arent we hitting this for the `gateway-helm` tests ?

zirain (approved):

cnvergence (approved):

PR #5902, opened 2025-05-02
fix(chart): passing root context to template

hansselvig:
**What type of PR is this?**

* fix(chart): passing the root context to template so that e.g. .`Values.topologyInjector.enabled` is possible to resolve

gateway-helm chart fails without this fix

Relates https://github.com/envoyproxy/gateway/issues/5903

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-rbac.yaml
@@ -19,7 +19,7 @@ metadata:
   name: {{ include "eg.fullname" $ }}-envoy-gateway-role
   namespace: {{ $ns | quote }}
 rules:
-{{ include "eg.rbac.namespaced" . }}
+{{ include "eg.rbac.namespaced" $ }}
 ---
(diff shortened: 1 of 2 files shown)


arkodg (approved):

arkodg:
thanks @hansselvig !
would be great if you could also add a test in https://github.com/envoyproxy/gateway/tree/main/test/helm/gateway-helm

cnvergence (approved):
I have encountered this recently, thank you for fixing this!

hansselvig:
Done.
Committed the output from `make generate`

arkodg:
@hansselvig can you fix DCO, by signing your commit and force pushing

hansselvig:
Commit should be signed now

arkodg (approved):

arkodg:
@hansselvig there's the same issue with signing, you may need to squash your commits

hansselvig:
Maybe now...
Thanks of the pointers!

arkodg (approved):

Issue #5906, opened 2025-05-03
helm lint errors are not failing CI


arkodg:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: The expected value isn't returned, etc.

Seeing lint errors in CI

```
1 chart(s) linted, 0 chart(s) failed
# The jb does not support self-assigned jsonnetfile, so entering working dir before executing jb.
echo -e "\033[36m===========> "Run helm template for chart: gateway-helm!"\033[0m";
===========> Run helm template for chart: gateway-helm!
Error: template: gateway-helm/templates/envoy-gateway-rbac.yaml:22:3: executing "gateway-helm/templates/envoy-gateway-rbac.yaml" at <include "eg.rbac.namespaced" .>: error calling include: template: gateway-helm/templates/_rbac.tpl:12:14: executing "eg.rbac.namespaced" at <.Values.topologyInjector.enabled>: can't evaluate field Values in type string
```

https://github.com/envoyproxy/gateway/actions/runs/14802116890/job/41565245925

These are also being reported by users https://github.com/envoyproxy/gateway/issues/5903

these should be caught in PR checks

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
>Include the access logs and the Envoy logs.

PR #5908, opened 2025-05-03
ci: make helm-generate should failed as expected

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5906


zirain:
wait https://github.com/envoyproxy/gateway/pull/5902

arkodg (approved):

PR #5910, opened 2025-05-03
docs: Add docs for request buffering

### LINKED ISSUE #3706: Request/Response Buffering API support

markwinter:

This is adding docs for the new request buffer feature added in #5661

First docs PR so hopefully didn't miss any other steps

Fixes #3706

Release Notes: No

### DIFF
--- a/site/content/en/latest/tasks/traffic/request-buffering.md
+++ b/site/content/en/latest/tasks/traffic/request-buffering.md
@@ -36,7 +36,7 @@ spec:
     kind: HTTPRoute
     name: foo
   requestBuffer:
-    limit: 4Ki
+    limit: 4 # Supports SI units e.g. 4Ki, 1Mi
 ---
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
--- a/site/content/en/latest/tasks/traffic/request-buffering.md
+++ b/site/content/en/latest/tasks/traffic/request-buffering.md
@@ -1,0 +1,207 @@
+---
+title: "Request Buffering"
+---
+
+The [Envoy buffer filter] is used to stop filter iteration and wait for a fully buffered complete request. This is useful in different situations including protecting some applications from having to deal with partial requests and high network latency.
+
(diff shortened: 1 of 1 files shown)


arkodg (approved):
LGTM thanks ! 
cc @melsal13

rudrakhp (approved):

PR #5911, opened 2025-05-03
fix: fix topology injector bug

### LINKED ISSUE #5890: Topology injector blocking envoy proxy from starting
*Description*:

Noticed envoy-proxy pods were not starting

```
Conditions:
  Type           Status  Reason
  ----           ------  ------
  Available      False   MinimumReplicasUnavailable
  Progressing    True    ReplicaSetUpdated
```

in the envoy gateway logs, I noticed topology injector failures

```
2025-05-01T15:58:03.840Z	ERROR	provider	kubernetes/topology_injector.go:78	error decoding patch: unexpected end of JSON inputpatch pod failedpodenvoy-gateway-system/envoy-default-eg-e41e7b31-654ccf5f87-mw6hk	{"runner": "provider"}
```

*Repro steps*:

sharing.

*Environment*:

*Logs*:

jukie:
**What type of PR is this?**

fix: fix topology injector bugs

* Fix caBundle
* Fix patch - patches binding resource instead and uses annotation env var

Fixes #5890

Release Notes: No


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -402,7 +402,7 @@ func expectedContainerEnv(containerSpec *egv1a1.KubernetesContainerSpec, gateway
 			ValueFrom: &corev1.EnvVarSource{
 				FieldRef: &corev1.ObjectFieldSelector{
 					APIVersion: "v1",
-					FieldPath:  fmt.Sprintf("metadata.labels['%s']", corev1.LabelTopologyZone),
+					FieldPath:  fmt.Sprintf("metadata.annotations['%s']", corev1.LabelTopologyZone),
```
zirain, line 405: why change from labels to annotations?

jukie: Annotations from `Binding` will get copied to the pod unlike labels so it saves an extra resource to handle during webhook execution.

arkodg: ah cool !

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #5912, opened 2025-05-04
docs(rate-limit): minor fix in 'Distinct Users Except Admin' section

tomas-rojo:
**What type of PR is this?**
This is a documentation update. It improves clarity and correctness in the BackendTrafficPolicy example for global rate limiting. Specifically, it adds a missing header match condition (x-user-id: admin with invert: true) that is required to make the "Distinct Users Except Admin" use case work as described.

This PR fixes a small but critical gap in the documentation for the **Global Rate Limit Distinct Users Except Admin** feature. The missing header configuration for excluding a specific user (e.g., x-user-id: admin) was causing the documented example to not behave as intended. This PR adds those missing lines to help users correctly configure rate limiting for distinct users while excluding admin traffic.

Fixes: N/A (no issue filed)

Release Notes: No

### DIFF
--- a/site/content/en/latest/tasks/traffic/global-rate-limit.md
+++ b/site/content/en/latest/tasks/traffic/global-rate-limit.md
@@ -493,6 +493,9 @@ spec:
         - headers:
           - type: Distinct
             name: x-user-id
+          - name: x-user-id
+            value: admin
+            invert: true
         limit:
           requests: 3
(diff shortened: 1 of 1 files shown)


arkodg (approved):
nice catch thanks !

zirain (approved):

Issue #5859, opened 2025-04-30
BTP: make pacth merge work on slice type


zirain:
@arkodg we need this tag for every slice we want to merge.

_Originally posted by @zirain in https://github.com/envoyproxy/gateway/pull/5795#discussion_r2067706879_

zhaohuabing:
Is this a blocker for v1.4?

arkodg:
This will be solved with #5944

zhaohuabing:
> This will be solved with [#5944](https://github.com/envoyproxy/gateway/pull/5944)

I may not understand it correctly, but two slices can only combine, the member can't be merged because they have different names(namesapce/policy/rule/index)

PR #5915, opened 2025-05-04
feat: allow merge rate limit rule in BTP

zirain:
xref: https://github.com/envoyproxy/gateway/issues/5859


charts/gateway-helm/crds/generated/gateway.envoyproxy.io_backendtrafficpolicies.yaml
```diff
@@ -947,6 +947,11 @@ spec:
                               - requests
                               - unit
                               type: object
+                            name:
```
arkodg, line 950: guessing there should be more changes in the CRD, tied to the `mergeKey` ?

zirain: `patchMergeKey` won't affect CRD files.

internal/utils/testdata/backendtrafficpolicy_ratelimitrule_merge.strategicmerge.out.yaml
```diff
@@ -0,0 +1,47 @@
...
+            number: 0
+          response:
+            from: Metadata
+            metadata:
+              key: request_cost_set_by_ext_proc
+              namespace: io.envoyproxy.gateway.e2e
+        limit:
+          requests: 21
+          unit: Hour
+        name: two
```
arkodg, line 26: is order arbitrary or deterministic ?

arkodg: I.e. two before one / patch before original

arkodg (approved):
LGTM thanks ! 
cc @ryanhristovski

api/v1alpha1/ratelimit_types.go
```diff
@@ -69,15 +71,23 @@ type LocalRateLimit struct {
...
 }
 
 // RateLimitRule defines the semantics for matching attributes
 // from the incoming requests, and setting limits for them.
 type RateLimitRule struct {
+	// Name is the name of the rule. This is used to identify the rule
+	// in the Envoy configuration and as a unique identifier for merging.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
```
zhaohuabing, line 90: A unique name will be generated if not provide by the user?

zhaohuabing (approved):
I just have a minor comment on the new `name` field.

api/v1alpha1/ratelimit_types.go
zirain: it won't happen.

zhaohuabing:
This PR is reverted by #5944

Issue #5909, opened 2025-05-03
Support client cert validation (spki, hash, san)


kinolaev:
*Description*:
I'd like to allow a TLS handshake based on SNI and client certificate fields. For example, to allow only handshakes from a client with SAN URI `spiffe://example.com/client` to `server.example.com` service.

In the context of Envoy I need to configure [CertificateValidationContext](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext) in a corresponding listener's filter chain.

To archive this goal I propose to add `spkis`, `hashes` and `subjectAltNames` fields to `ClientValidationContext` of `ClientTrafficPolicy`. For example:
```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: ClientTrafficPolicy
metadata:
  namespace: default
  name: mtls-authentication
spec:
  targetRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: eg
      sectionName: server
  tls:
    clientValidation:
      caCertificateRefs:
        - kind: Secret
          name: root-ca
      spkis:
        - NvqYIYSbgK2vCJpQhObf77vv+bQWtc5ek5RIOwPiC9A=
      hashes:
        - df6ff72fe9116521268f6f2dd4966f51df479883fe7037b39f75916ac3049d1a
      subjectAltNames:
        dnsNames:
          - { type: Exact, value: client.example.com }
        emailAddresses:
          - { type: Suffix, value: @example.com }
        ipAddresses:
          - { type: Prefix, value: 192.168. }
        uris:
          - { value: spiffe://example.com/client }
        otherNames:
          - { oid: 1.3.6.1.4.1.311.20.2.3, value: client }
```

An alternative would be to introduce a new type of [Principal](https://gateway.envoyproxy.io/docs/api/extension_types/#principal) to `AuthorizationRule` of `SecurityPolicy`.
But we can only support `Allow` action (there are only allowlists in Envoy's CertificateValidationContext) on a Gateway target (because it affects all routes) and return an error otherwise. Additionally [SecurityPolicy doesn't support sectionName](https://github.com/envoyproxy/gateway/blob/v1.4.0-rc.2/api/v1alpha1/securitypolicy_types.go#L43), so for now it's not possible to attach a policy to a specific listener and propagate rules to the right filter chain. It must be fixed first.

[...]

PR #5916, opened 2025-05-04
feat: support listener name in SecurityPolicy target

kinolaev:
**What type of PR is this?**

feat: support listener name in SecurityPolicy target

It allows to target only a specific listener in a Security Policy. One usage example is attaching mtls authorization as described in https://github.com/envoyproxy/gateway/issues/5909

Release Notes: No

### DIFF
--- a/api/v1alpha1/securitypolicy_types.go
+++ b/api/v1alpha1/securitypolicy_types.go
@@ -37,10 +37,10 @@ type SecurityPolicy struct {
 //
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.group == 'gateway.networking.k8s.io' : true", message="this policy can only have a targetRef.group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute'] : true", message="this policy can only have a targetRef.kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind == 'Gateway' || !has(self.targetRef.sectionName) : true",message="this policy supports the sectionName field only for gateways"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind == 'Gateway' || !has(self.targetRef.sectionName) : true",message="this policy supports the sectionName field only for kind Gateway"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.group == 'gateway.networking.k8s.io') : true ", message="this policy can only have a targetRefs[*].group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute']) : true ", message="this policy can only have a targetRefs[*].kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind == 'Gateway' || !has(ref.sectionName)) : true",message="this policy supports the sectionName field only for gateways"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind == 'Gateway' || !has(ref.sectionName)) : true",message="this policy supports the sectionName field only for kind Gateway"
 // +kubebuilder:validation:XValidation:rule="(has(self.authorization) && has(self.authorization.rules) && self.authorization.rules.exists(r, has(r.principal.jwt))) ? has(self.jwt) : true", message="if authorization.rules.principal.jwt is used, jwt must be defined"
 //
 // SecurityPolicySpec defines the desired state of SecurityPolicy.
--- a/api/v1alpha1/securitypolicy_types.go
+++ b/api/v1alpha1/securitypolicy_types.go
@@ -37,10 +37,10 @@ type SecurityPolicy struct {
 //
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.group == 'gateway.networking.k8s.io' : true", message="this policy can only have a targetRef.group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute'] : true", message="this policy can only have a targetRef.kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? !has(self.targetRef.sectionName) : true",message="this policy does not yet support the sectionName field"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind == 'Gateway' || !has(self.targetRef.sectionName) : true",message="this policy supports the sectionName field only for gateways"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.group == 'gateway.networking.k8s.io') : true ", message="this policy can only have a targetRefs[*].group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute']) : true ", message="this policy can only have a targetRefs[*].kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, !has(ref.sectionName)) : true",message="this policy does not yet support the sectionName field"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind == 'Gateway' || !has(ref.sectionName)) : true",message="this policy supports the sectionName field only for gateways"
 // +kubebuilder:validation:XValidation:rule="(has(self.authorization) && has(self.authorization.rules) && self.authorization.rules.exists(r, has(r.principal.jwt))) ? has(self.jwt) : true", message="if authorization.rules.principal.jwt is used, jwt must be defined"
 //
 // SecurityPolicySpec defines the desired state of SecurityPolicy.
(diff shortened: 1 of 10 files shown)


kinolaev:
Test is fixed, `make go.test.cel` runs without any error now

internal/gatewayapi/securitypolicy.go
```diff
@@ -568,6 +568,9 @@ func (t *Translator) translateSecurityPolicyForGateway(
 		if t.MergeGateways && gatewayName != policyTarget {
 			continue
 		}
+		if target.SectionName != nil && string(*target.SectionName) != h.Name[strings.LastIndex(h.Name, "/")+1:] {
```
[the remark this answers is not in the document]
kinolaev: Thanks for pointing these issues out! I've addressed all of them in 3f7c7bb6454040b667b4fe4d370500bed1c865e3

zhaohuabing:
Hi @kinolaev Could you clarify in the description how the overriding/merging strategy works between Gateway/Listener/Route?

zhaohuabing:
Do we still need this for https://github.com/envoyproxy/gateway/issues/5909?  Cert validation will be done in ClientTrafficPolicy at https://github.com/envoyproxy/gateway/pull/5868, and fine-grained control using SAN as a type of [Principal](https://gateway.envoyproxy.io/docs/api/extension_types/#principal) can be done at Gateway/Route level.

kinolaev:
Sorry for long reply(

> Could you clarify in the description how the overriding/merging strategy works

A Listener attached security policy applies only to Routes that doesn't have a security policy attached to them directly. Could you please be more specific about where exactly you want me to add this explanation?

> do we need this for https://github.com/envoyproxy/gateway/issues/5909

It depends on your decision) In the issue I've described two possibilities of mTLS authorization - using ClientTrafficPolicy (during handshake) or using SecurityPolicy (after handshake). ClientTrafficPolicy is a good place to add SAN allowlist into envoy's CertificateValidationContext and we don't need this PR for that. But if you only want to authorize SANs using envoy's RBAC then SecurityPolicy might be a better place and consequently we need this PR as a prerequisite.

So, yeah, I made steps in both directions because I didn't want to guess what you would decide) But as you already mentioned, this PR could be useful even without relation to the mTLS authorization issue.

arkodg:
requesting @zirain to review, who implemented something similar for BTP but per route rule

internal/gatewayapi/testdata/securitypolicy-override-replace.out.yaml
```diff
@@ -234,14 +377,78 @@ securityPolicies:
...
+        sectionName: listener-2
+      conditions:
+      - lastTransitionTime: null
+        message: Policy has been accepted.
+        reason: Accepted
+        status: "True"
+        type: Accepted
+      - lastTransitionTime: null
+        message: 'This policy is being overridden by other securityPolicies for these
+          routes: [default/httproute-3 default/httproute-5]'
```
zirain, line 450: Is this wording correct? Do we only need to tell the end users on which routes the current policy takes effect?
cc @arkodg

zhaohuabing:
> A Listener attached security policy applies only to Routes that doesn't have a security policy attached to them directly. Could you please be more specific about where exactly you want me to add this explanation?

Yeah, this makes sense as we need to apply SAN based authz for the entire listener.

api/v1alpha1/securitypolicy_types.go
```diff
@@ -37,10 +37,10 @@ type SecurityPolicy struct {
 //
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.group == 'gateway.networking.k8s.io' : true", message="this policy can only have a targetRef.group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute'] : true", message="this policy can only have a targetRef.kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? !has(self.targetRef.sectionName) : true",message="this policy does not yet support the sectionName field"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind == 'Gateway' || !has(self.targetRef.sectionName) : true",message="this policy supports the sectionName field only for gateways"
```
arkodg:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.targetRef) ? self.targetRef.kind == 'Gateway' || !has(self.targetRef.sectionName) : true",message="this policy supports the sectionName field only for kind Gateway"
```

api/v1alpha1/securitypolicy_types.go
```diff
@@ -37,10 +37,10 @@ type SecurityPolicy struct {
...
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.group == 'gateway.networking.k8s.io') : true ", message="this policy can only have a targetRefs[*].group of gateway.networking.k8s.io"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute']) : true ", message="this policy can only have a targetRefs[*].kind of Gateway/HTTPRoute/GRPCRoute"
-// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, !has(ref.sectionName)) : true",message="this policy does not yet support the sectionName field"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind == 'Gateway' || !has(ref.sectionName)) : true",message="this policy supports the sectionName field only for gateways"
```
arkodg:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) ? self.targetRefs.all(ref, ref.kind == 'Gateway' || !has(ref.sectionName)) : true",message="this policy supports the sectionName field only for kind Gateway"
```

internal/gatewayapi/testdata/securitypolicy-override-replace.out.yaml
arkodg: this is fine, this is consistent with overrides today https://github.com/envoyproxy/gateway/blob/28424b012665450a3ee00c3948508dfb56798ff9/internal/gatewayapi/testdata/backendtrafficpolicy-status-conditions.out.yaml#L129

test/cel-validation/securitypolicy_test.go
```diff
@@ -177,8 +177,8 @@ func TestSecurityPolicyTarget(t *testing.T) {
 							{
 								LocalPolicyTargetReference: gwapiv1a2.LocalPolicyTargetReference{
 									Group: gwapiv1a2.Group("gateway.networking.k8s.io"),
-									Kind:  gwapiv1a2.Kind("Gateway"),
-									Name:  gwapiv1a2.ObjectName("eg"),
+									Kind:  gwapiv1a2.Kind("HTTPRoute"),
```
arkodg, line 218: can we have one with `Gateway` so we also tst posiitive case

internal/gatewayapi/securitypolicy.go
```diff
@@ -77,17 +77,18 @@ func (t *Translator) ProcessSecurityPolicies(securityPolicies []*egv1a1.Security
 	gatewayMap := map[types.NamespacedName]*policyGatewayTargetContext{}
 	for _, gw := range gateways {
 		key := utils.NamespacedName(gw)
-		gatewayMap[key] = &policyGatewayTargetContext{GatewayContext: gw}
+		gatewayMap[key] = &policyGatewayTargetContext{GatewayContext: gw, attachedToListeners: make(sets.Set[string])}
 	}
 
 	// Map of Gateway to the routes attached to it
-	gatewayRouteMap := make(map[string]sets.Set[string])
+	gatewayRouteMap := make(map[string]map[string]sets.Set[string])
```
arkodg, line 86: does L83 comment need to be updated ?

internal/gatewayapi/securitypolicy.go
```diff
@@ -564,10 +647,14 @@ func (t *Translator) translateSecurityPolicyForGateway(
 
 	policyTarget := irStringKey(policy.Namespace, string(target.Name))
 	for _, h := range x.HTTP {
-		gatewayName := h.Name[0:strings.LastIndex(h.Name, "/")]
+		gatewayNameEnd := strings.LastIndex(h.Name, "/")
+		gatewayName := h.Name[0:gatewayNameEnd]
 		if t.MergeGateways && gatewayName != policyTarget {
 			continue
 		}
+		if target.SectionName != nil && string(*target.SectionName) != h.Name[gatewayNameEnd+1:] {
```
arkodg, line 680: can you add a comment for this ? can't understand this code

api/v1alpha1/securitypolicy_types.go
kinolaev: Updated in f1a1fd5262b6609b7617e39edee602514d3c4595

api/v1alpha1/securitypolicy_types.go
kinolaev: Updated in f1a1fd5262b6609b7617e39edee602514d3c4595

test/cel-validation/securitypolicy_test.go
kinolaev: Added in f1a1fd5262b6609b7617e39edee602514d3c4595

internal/gatewayapi/securitypolicy.go
kinolaev: Added "The routes are grouped by sectionNames of their targetRefs" in f1a1fd5262b6609b7617e39edee602514d3c4595

internal/gatewayapi/securitypolicy.go
kinolaev: "A HTTPListener name has the format namespace/gatewayName/listenerName" and "If specified the sectionName must match the listenerName part of the HTTPListener name" added in f1a1fd5262b6609b7617e39edee602514d3c4595

arkodg:
the test is failing due to a difference in string
```
--- FAIL: TestSecurityPolicyTarget (0.15s)
    --- FAIL: TestSecurityPolicyTarget/sectionName_disabled_until_supported_for_kind_xRoute_-_targetRef (0.00s)
        securitypolicy_test.go:1358: Unexpected response while creating SecurityPolicy; got err=
            SecurityPolicy.gateway.envoyproxy.io "sp-1749864680953051291" is invalid: spec: Invalid value: "object": this policy supports the sectionName field only for kind Gateway
            ;missing strings within error=["spec: Invalid value: \"object\": this policy supports the sectionName field only kind Gateway"]
```

you should be able to repro this locally using `make kube-test`

kinolaev:
Oh, sorry, forgot "for" in "only kind Gateway"( fixed

zirain (approved):

PR #5933, opened 2025-05-06
fix: add validation for header values

### LINKED ISSUE #5901: Issue with validation with incorrect header values
While migrating from Gloo to Envoy Gateway I discovered that a single malformed header can stall the controller for every other host on the same Gateway.
If a ResponseHeaderModifier filter sets a header using the YAML pipe (|), the resulting multi-line value slips through validation, but afterwards no further config changes are reconciled. New routes/hosts return 404 because the Gateway is stuck on the bad object.

Envoy Gateway: 1.3.2
Kubernetes: v1.31.7

*Steps to reproduce:*
```
rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    filters:
    - type: ResponseHeaderModifier
      responseHeaderModifier:
        set:
        - name: Content-Security-Policy
          value: |
            default-src 'none'; connect-src https://*.example.com
```

1. Confirm the route shows Accepted: true.
2. Add another HTTPRoute (or edit the same one) to introduce a new host.
3. Curl the new host → 404.

gavinkflam:
**What type of PR is this?**

A single malformed header can stall the entire controller. We need to validate header values according to RFC 7230.

The proposed changes catch invalid header values and flag them in status messages.

Fixes #5901

Release Notes: Yes

### DIFF
--- a/internal/gatewayapi/clienttrafficpolicy.go
+++ b/internal/gatewayapi/clienttrafficpolicy.go
@@ -952,7 +952,12 @@ func translateEarlyRequestHeaders(headerModifier *gwapiv1.HTTPHeaderFilter) ([]i
 			}
 			// Per Gateway API specification on HTTPHeaderName, : and / are invalid characters in header names
 			if strings.ContainsAny(string(addHeader.Name), "/:") {
-				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders Filter cannot set headers with a '/' or ':' character in them. Header: %q", string(addHeader.Name)))
+				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot add a header with a '/' or ':' character in them. Header: '%q'", string(addHeader.Name)))
+				continue
+			}
+			// Gateway API specification allows only valid value as defined by RFC 7230
+			if !HeaderValueRegexp.MatchString(addHeader.Value) {
+				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot add a header with an invalid value. Header: '%q'", string(addHeader.Name)))
 				continue
 			}
 			// Check if the header is a duplicate
@@ -992,7 +997,12 @@ func translateEarlyRequestHeaders(headerModifier *gwapiv1.HTTPHeaderFilter) ([]i
 			}
 			// Per Gateway API specification on HTTPHeaderName, : and / are invalid characters in header names
 			if strings.ContainsAny(string(setHeader.Name), "/:") {
-				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot set headers with a '/' or ':' character in them. Header: '%s'", string(setHeader.Name)))
+				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot set a header with a '/' or ':' character in them. Header: '%q'", string(setHeader.Name)))
+				continue
+			}
+			// Gateway API specification allows only valid value as defined by RFC 7230
+			if !HeaderValueRegexp.MatchString(setHeader.Value) {
+				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot set a header with an invalid value. Header: '%q'", string(setHeader.Name)))
 				continue
 			}
 
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -66,6 +66,9 @@ type HTTPFilterIR struct {
 	ExtensionRefs []*ir.UnstructuredRef
 }
 
+// Header value pattern according to RFC 7230
+var HeaderValueRegexp = regexp.MustCompile(`^[!-~]+([\t ]?[!-~]+)*$`)
+
 // ProcessHTTPFilters translates gateway api http filters to IRs.
 func (t *Translator) ProcessHTTPFilters(parentRef *RouteParentContext,
 	route RouteContext,
@@ -376,6 +379,7 @@ func (t *Translator) processRequestHeaderModifierFilter(
 			emptyFilterConfig = false
 		}
 		for _, addHeader := range headersToAdd {
+
 			emptyFilterConfig = false
 			if addHeader.Name == "" {
 				updateRouteStatusForFilter(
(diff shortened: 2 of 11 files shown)


gavinkflam:
> for v1.4, we should probably catch this in the gateway-api layer and flag it in a status

@arkodg is this what you are thinking about?

gavinkflam:
@arkodg 

I believe we want the same validation for RequestHeaderModifier as well. I’ll add it tonight and push an update.

As for CRD validation patterns, I suggest to hold off on that for now since it would be a breaking change. A soft rejection with a status message seems like a better approach at this stage. We can align with the Gateway API’s schedule to introduce those validation rules.

Pipeline - I believe the e2e tests are just flaky.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -938,6 +939,9 @@ func translateEarlyRequestHeaders(headerModifier *gwapiv1.HTTPHeaderFilter) ([]i
 	var AddRequestHeaders []ir.AddHeader
 	var RemoveRequestHeaders []string
 
+	// Header value pattern according to RFC 7230
```
arkodg, line 942: can this be defined once maybe in filters.go ?

internal/gatewayapi/testdata/clienttrafficpolicy-headers-empty-header-names-error.in.yaml
```diff
@@ -12,14 +12,14 @@ clientTrafficPolicies:
       earlyRequestHeaders:
         add:
         - name: ""
-          value: "empty"
-        - name: "invalid"
-          value: ":/"
+          value: "dummy"
```
arkodg, line 15: curious why these were changed ?

arkodg:
thanks @gavinkflam !

rejection with a status message is the best we can do for requestHeaderModifier and responseHeaderModifier, since the CRD is defined upstream

we have a similar issue for earlyRequestHeaders which is also using the upstream definition
https://github.com/envoyproxy/gateway/blob/95b4f9757817bd6211158251cad6b3a9d663701d/api/v1alpha1/clienttrafficpolicy_types.go#L153

internal/gatewayapi/clienttrafficpolicy.go
gavinkflam: sure. this is moved to filters.go

internal/gatewayapi/testdata/clienttrafficpolicy-headers-empty-header-names-error.in.yaml
gavinkflam: there was only one test for `earlyRequestHeaders` header validations, and it only covered empty header names. I added three more test files to cover all relevant validations.

the header values don’t necessarily need to be changed, as the test results would remain the same. I was doing all the empty header test cases at the same time and used similar values across the test cases.

internal/gatewayapi/clienttrafficpolicy.go
```diff
@@ -952,7 +952,12 @@ func translateEarlyRequestHeaders(headerModifier *gwapiv1.HTTPHeaderFilter) ([]i
 			}
 			// Per Gateway API specification on HTTPHeaderName, : and / are invalid characters in header names
 			if strings.ContainsAny(string(addHeader.Name), "/:") {
-				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders Filter cannot set headers with a '/' or ':' character in them. Header: %q", string(addHeader.Name)))
+				errs = errors.Join(errs, fmt.Errorf("EarlyRequestHeaders cannot add a header with a '/' or ':' character in them. Header: '%q'", string(addHeader.Name)))
+				continue
+			}
+			// Gateway API specification allows only valid value as defined by RFC 7230
+			if !HeaderValueRegexp.Match([]byte(addHeader.Value)) {
```
arkodg, line 959: should https://pkg.go.dev/regexp#Regexp.MatchString be used instead ?

internal/gatewayapi/testdata/clienttrafficpolicy-headers-empty-header-names-error.in.yaml
arkodg: ah got it , thanks

internal/gatewayapi/clienttrafficpolicy.go
gavinkflam: good catch. fixed

internal/gatewayapi/testdata/clienttrafficpolicy-headers-empty-header-names-error.in.yaml
gavinkflam: actuallyit’s probably cleaner to merge all these tests into a single file, since `earlyRequestHeaders` can report all validation errors of the policy. I've combined them into one file now.

arkodg (dismissed):
LGTM thanks !

gavinkflam:
thanks I can rebase! just a conflict in the release notes file

gavinkflam:
@arkodg rebased

guydc:
Envoy supports format specifiers for headers using symbols like `%():.-` : https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalue. Users can create expressions, rather than literal values, that allow adding dynamic header values. 

Is this functionality impacted this change? If so, can we perhaps add a flag in EnvoyGateway to allow "unsafe" header values?

arkodg (approved):

arkodg:
> Envoy supports format specifiers for headers using symbols like `%():.-` : https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalue. Users can create expressions, rather than literal values, that allow adding dynamic header values.
> 
> Is this functionality impacted this change? If so, can we perhaps add a flag in EnvoyGateway to allow "unsafe" header values?

`%():.-` should be okay to still include in `value` as far as i can tell rn

guydc (approved):

gavinkflam:
Yes, these are still valid characters.

I tried the examples in formatting string documents as well, they are valid values with the pattern.

PR #5935, opened 2025-05-06
docs: dynamic resolver backend

zhaohuabing:


shawnh2 (approved):

arkodg:
thanks @zhaohuabing ! can we add this as a section within https://gateway.envoyproxy.io/docs/tasks/traffic/backend/ ?
Also can we refer to this as Dynamic Resolver, and refer to the under the hood implementation using Envoy's Dynamic Forward Proxy in the overview

zhaohuabing:
@arkodg Not a strong opinion, but since the tasks are arranged by features, I slightly prefer keeping "dynamic forward proxy" as a standalone task. The `DynamicResolver` type of `Backend` is the way this feature is implemented in Envoy Gateway. 

Having it as a standalone task also makes it easier for users to discover this powerful feature.

zhaohuabing:
Hi @arkodg moved it to the backend main page. I still use the "Dynamic Forward Proxy" as the title for this section since I feel it's easier to understand. If you prefer "Dynamic Resolver" as the title, I can change it.

shawnh2 (approved):

PR #5936, opened 2025-05-06
feat: support app protocol for dynamic resolver backend

zhaohuabing:
part of #3560


test/e2e/tests/httproute_with_dynamic_resolver_backend.go
```diff
@@ -63,6 +63,28 @@ var DynamicResolverBackendTest = suite.ConformanceTest{
...
+				Request: http.Request{
+					Host: "httpbin.org",
+					Path: "/status/200",
+				},
+				ExpectedRequest: &http.ExpectedRequest{
+					Request: http.Request{
+						Host: "",
+					},
+				},
+				Response: http.Response{
```
rudrakhp, line 81: Can we not verify if it's an response from `gateway.envoyproxy.io/status` other than the status code?

zhaohuabing: I couldn’t find a reliable way to verify it beyond the status code. That said, I think it’s safe to assume the response is from httpbin.org—a random server is unlikely to expose a /status/200 endpoint.

rudrakhp (approved):

guydc (approved):

PR #5937, opened 2025-05-06
feat: add controller namespace field to infrastructure render

cnvergence:
**What type of PR is this?**

Added new infra resource render field - controller namespace, this helps to distinguish namespace names in the infra processing.

Fixes https://github.com/envoyproxy/gateway/issues/2629
Followup to the https://github.com/envoyproxy/gateway/pull/5137

Release Notes: No

### DIFF
--- a/internal/infrastructure/kubernetes/infra.go
+++ b/internal/infrastructure/kubernetes/infra.go
@@ -45,8 +45,8 @@ type ResourceRender interface {
 // Infra manages the creation and deletion of Kubernetes infrastructure
 // based on Infra IR resources.
 type Infra struct {
-	// Namespace is the Namespace used for managed infra.
-	Namespace string
+	// ControllerNamespace is the namespace where Envoy Gateway is deployed.
+	ControllerNamespace string
 
 	// DNSDomain is the dns domain used by k8s services. Defaults to "cluster.local".
 	DNSDomain string
@@ -65,11 +65,11 @@ func NewInfra(cli client.Client, cfg *config.Server) *Infra {
 	return &Infra{
 		// Always set infra namespace to cfg.ControllerNamespace,
 		// Otherwise RateLimit resource provider will failed to create/delete.
-		Namespace:    cfg.ControllerNamespace,
-		DNSDomain:    cfg.DNSDomain,
-		EnvoyGateway: cfg.EnvoyGateway,
-		Client:       New(cli),
-		logger:       cfg.Logger.WithName(string(egv1a1.LogComponentInfrastructureRunner)),
+		ControllerNamespace: cfg.ControllerNamespace,
+		DNSDomain:           cfg.DNSDomain,
+		EnvoyGateway:        cfg.EnvoyGateway,
+		Client:              New(cli),
+		logger:              cfg.Logger.WithName(string(egv1a1.LogComponentInfrastructureRunner)),
 	}
 }
 
--- a/internal/infrastructure/kubernetes/infra.go
+++ b/internal/infrastructure/kubernetes/infra.go
@@ -48,9 +48,6 @@ type Infra struct {
 	// Namespace is the Namespace used for managed infra.
 	Namespace string
 
-	// ControllerNamespace is the Namespace used for Envoy Gateway controller.
-	ControllerNamespace string
-
 	// DNSDomain is the dns domain used by k8s services. Defaults to "cluster.local".
 	DNSDomain string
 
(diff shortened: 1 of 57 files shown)


internal/infrastructure/kubernetes/infra.go
```diff
@@ -47,6 +47,9 @@ type Infra struct {
 	// Namespace is the Namespace used for managed infra.
 	Namespace string
 
+	// ControllerNamespace is the Namespace used for Envoy Gateway controller.
```
arkodg, line 50: shouldnt this be available in Runner Config  ?

cnvergence: Runner config calls the NewInfra, I'm adding this field to be able to differentiate the namespace used for Infra and Controller

arkodg: referring to https://github.com/envoyproxy/gateway/blob/ef507181e89219c04faf7fbe67e4bcce898b751e/internal/infrastructure/runner/runner.go#L27

cnvergence: controller namespace will be fetched from the Config

arkodg: so, we dont have access to the runner context anymore ? is that the issue ?

zhaohuabing: @arkodg runner context is not in the Infra, instead, only the needed fields are passed down to the Infra structure.

```
func NewInfra(cli client.Client, cfg *config.Server) *Infra {
	var infraNamespace string
	if !cfg.EnvoyGateway.GatewayNamespaceMode() {
		infraNamespace = cfg.ControllerNamespace
	}
	return &Infra{
		Namespace:           infraNamespace,
		ControllerNamespace: cfg.ControllerNamespace,    # cfg is runner config
		DNSDomain:           cfg.DNSDomain,
		EnvoyGateway:        cfg.EnvoyGateway,
		Client:              New(cli),
		logger:              cfg.Logger.WithName(string(egv1a1.LogComponentInfrastructureRunner)),
	}
}
```

internal/infrastructure/kubernetes/infra.go
```diff
@@ -47,6 +47,9 @@ type Infra struct {
 	// Namespace is the Namespace used for managed infra.
 	Namespace string
```
zhaohuabing, line 48: Hi @cnvergence do we need to change the namespace used by the ratelimit ResourceRender? It should use controllernamespace instead of namespace.

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -135,7 +131,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 			ImagePullPolicy:          corev1.PullIfNotPresent,
 			Command:                  []string{"envoy"},
 			Args:                     args,
-			Env:                      expectedContainerEnv(containerSpec, gatewayNamespaceMode),
+			Env:                      expectedContainerEnv(containerSpec, egNamespace),
```
zhaohuabing, line 134: Nit, non-blocking: Can we use a consistent name for `controllerNamespace/egNamespace`. Two names for the same concept might confuse future readers.

I slightly prefer `EGNamespace` as it feels a bit more straightforward.

arkodg: API calls it ControllerNamespace

zhaohuabing: ControllerNamespace is fine.

internal/infrastructure/kubernetes/infra.go
Xunzhuo: Yep, +1 for this, clearer to understand the differences

zhaohuabing:
Can we also rename the `Namespace` field in the `Infra` structure to `ControllerNamesapce` to avoid confusion?

```
type Infra struct {
	// Namespace is the Namespace used for managed infra.
	Namespace string

	// DNSDomain is the dns domain used by k8s services. Defaults to "cluster.local".
	DNSDomain string

	// EnvoyGateway is the configuration used to startup Envoy Gateway.
	EnvoyGateway *egv1a1.EnvoyGateway

	// Client wrap k8s client.
	Client *InfraClient

	logger logging.Logger
}
```

zhaohuabing (approved):
LGTM thanks!

zirain (approved):

PR #5955, opened 2025-05-08
fix: add validation for duplicated API keys

### LINKED ISSUE #5729: Listener config rejected is shown as warning instead of error
*Description*:

When having duplicate API keys in the secret that is used by the security policy of a route, the config is listed as "accepted" in the manifest, but the logs of the gateway show a warning that contains an error:

```
[2025-04-09 02:29:35.230][1][warning][config] [source/extensions/config_subscription/grpc/grpc_subscription_impl.cc:138] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) dm-gateway/dm-gateway/internal-gateway-https: Duplicated credential key: '...'
```

So this brings envoy in an invalid state where the config is actually not applied, but you think it does.
In our case we restarted envoy to debug something and it couldn't handle any message anymore due to the faulty config.

The above should be an error that blocks the accepting of the config.

*Repro steps*:

1. Create a security policy with a duplicate api key
2. Try doing a request: nothing will have changed (apart from the warnings in the logs)
3. Restart envoy
4. Now you get internal error on each request.

*Environment*:

envoy gateway `v1.3.0` and `v1.33.0` running on Kubernetes.

*Logs*:

(see above)

gavinkflam:
**What type of PR is this?**

Bug fix

Duplicate API keys in apiKeyAuth can cause the entire controller to stall.

The proposed changes address this by rejecting the affected SecurityPolicy and reporting the error through a status message.

Fixes #5729

Release Notes: Yes

### DIFF
--- a/internal/gatewayapi/securitypolicy.go
+++ b/internal/gatewayapi/securitypolicy.go
@@ -1124,6 +1124,8 @@ func (t *Translator) buildAPIKeyAuth(
 	}
 
 	credentials := make(map[string]ir.PrivateBytes)
+	seenKeys := make(sets.Set[string])
+
 	for _, ref := range policy.Spec.APIKeyAuth.CredentialRefs {
 		credentialsSecret, err := t.validateSecretRef(
 			false, from, ref, resources)
@@ -1134,6 +1136,13 @@ func (t *Translator) buildAPIKeyAuth(
 			if _, ok := credentials[clientid]; ok {
 				continue
 			}
+
+			keyString := string(key)
+			if seenKeys.Has(keyString) {
+				return nil, errors.New("duplicated API key")
+			}
+
+			seenKeys.Insert(keyString)
 			credentials[clientid] = key
 		}
 	}
--- a/test/e2e/testdata/api-key-auth.yaml
+++ b/test/e2e/testdata/api-key-auth.yaml
@@ -10,6 +10,15 @@ data:
   # key2
   client2: "a2V5Mg=="
 ---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: gateway-conformance-infra
+  name: api-key-auth-users-secret-2
+data:
+  # key2 - duplicate client id should be ignored
+  client1: "a2V5Mg=="
+---
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
@@ -76,6 +85,7 @@ spec:
       - headers: ["X-API-KEY"]
     credentialRefs:
       - name: "api-key-auth-users-secret-1"
+      - name: "api-key-auth-users-secret-2"
 ---
 apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: SecurityPolicy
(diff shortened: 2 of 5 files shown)


internal/gatewayapi/securitypolicy.go
```diff
@@ -1134,6 +1136,13 @@ func (t *Translator) buildAPIKeyAuth(
 			if _, ok := credentials[clientid]; ok {
 				continue
 			}
+
+			keyString := string(key)
```
arkodg, line 1140: shouldnt this be solved In L1136, instead of `continue` we flag the error and return

gavinkflam: L1136 handles duplicate client IDs by retaining only the first API key. For instance, in `{client1: "key1", client1: "key2"}`, key2 is dropped. This behavior is documented in the [APIKeyAuth documentation](https://gateway.envoyproxy.io/docs/api/extension_types/#apikeyauth).

The issue reported in #5729, however, involves duplicate API keys, such as `{client1: "key1", client2: "key1"}`. In this case, downstream gives an error and the controller stalls because duplicate API keys are not allowed.

arkodg: thanks for explaining, makes sense

internal/gatewayapi/securitypolicy.go
```diff
@@ -1134,6 +1136,13 @@ func (t *Translator) buildAPIKeyAuth(
...
+			if _, ok := seenKeys[keyString]; ok {
+				return nil, fmt.Errorf("duplicated API key, client id: %s", clientid)
```
arkodg, line 1142: lets avoid adding the client id in the status for now, to avoid any sort of data leak

gavinkflam: done and rebased

arkodg (approved):

internal/gatewayapi/securitypolicy.go
```diff
@@ -1124,6 +1124,8 @@ func (t *Translator) buildAPIKeyAuth(
 	}
 
 	credentials := make(map[string]ir.PrivateBytes)
+	seenKeys := make(map[string]struct{})
```
rudrakhp, line 1127: Use utils.Set instead?

arkodg: +1

gavinkflam: Thanks. This is much nicer

test/e2e/testdata/api-key-auth.yaml
```diff
@@ -10,6 +10,15 @@ data:
...
 ---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: gateway-conformance-infra
+  name: api-key-auth-users-secret-2
+data:
+  # key2 - duplicate client id should be ignored
+  client1: "a2V5Mg=="
+---
```
gavinkflam, line 21: Enhanced APIKeyAuth e2e test to cover duplicate client IDs. Only the first API key should be retained. For instance, in {client1: "key1"} + {client1: "key2"}, key2 is dropped.

This behavior is documented in the [APIKeyAuth documentation](https://gateway.envoyproxy.io/docs/api/extension_types/#apikeyauth).

rudrakhp (approved):

arkodg (approved):

Issue #5941, opened 2025-05-07
Support channel: stable for gateway-crds-helm chart


arkodg:
*Description*:
>Describe the issue.

The gateway-crds-helm chart only installs gateway API CRDs from the experimental channel today

Add support to opt into installing CRDs from the stable channel

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zirain:
fixed by https://github.com/envoyproxy/gateway/pull/5958?

PR #5958, opened 2025-05-08
helm: support standard channel

zhaohuabing:
This PR adds an option to install the Gateway API standard channel in the gateway-crds-helm.

implements: #5941


zhaohuabing:
Do we also want to add the standard channel option to the gateway-helm? @arkodg

charts/gateway-crds-helm/values.tmpl.yaml
```diff
@@ -1,5 +1,6 @@
 crds:
   gatewayAPI:
     enabled: false
+    channel: experimental
```
arkodg, line 4: can we add a doc string to mention the option of `standard` is available

arkodg:
thanks @zhaohuabing, added a minor comment around doc string

arkodg (approved):
LGTM thanks !

zirain (approved):

maodahua:
> Do we also want to add the standard channel option to the gateway-helm? @arkodg

I am very keen to have this. But annother question come in to my brain is: Does the different channel have any impact of the envoy gateway? I want to use the `gateway-helm` to manage all the resources, include CRDs.

PR #5959, opened 2025-05-08
Revert "e2e: test for dynamic resolver backend using system ca for TLS"

zirain:
Reverts envoyproxy/gateway#5932


zhaohuabing:
It only fails in IPv6, or fail in both IPv4 and IPv6?

zirain:
looks like only IPv6.

zhaohuabing:
If it's just occasionally flaky and not blocking, prefer to not revert it since it tests the system ca for dynamic resolver backend.

Xunzhuo:
> Can we just disable it for IPv6 before we figure out why it's flaky in IPv6?

+1

zirain:
> Can we just disable it for IPv6 before we figure out why it's flaky in IPv6?

you can send another PR.

zhaohuabing:
Interesting, looks like this only fails in IPV6 with k8s 1.32

https://github.com/envoyproxy/gateway/actions/runs/14900411202/job/41851628532?pr=5958
https://github.com/envoyproxy/gateway/actions/runs/14898729273/job/41846659658?pr=5955
https://github.com/envoyproxy/gateway/actions/runs/14898348190/job/41848750435?pr=5954

zhaohuabing:
> > Can we just disable it for IPv6 before we figure out why it's flaky in IPv6?
> 
> you can send another PR.

https://github.com/envoyproxy/gateway/pull/5964

PR #5967, opened 2025-05-08
fix btp merge not working when there's multi parent refs on router

zirain:
this patch fix an issue while merging a route with multiple parent refs.

### DIFF
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -418,6 +418,7 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(
 
 	// Apply IR to all relevant routes
 	for key, x := range xdsIR {
+		// if gatewayNN is not nil, only apply to the specific gateway
 		if gatewayNN != nil && key != t.IRKey(*gatewayNN) {
 			// Skip if not the gateway wanted
 			continue
--- a/internal/gatewayapi/backendtrafficpolicy.go
+++ b/internal/gatewayapi/backendtrafficpolicy.go
@@ -165,7 +165,7 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 
 				if policy.Spec.MergeType == nil {
 					// Set conditions for translation error if it got any
-					if err := t.translateBackendTrafficPolicyForRoute(policy, route, xdsIR, resources); err != nil {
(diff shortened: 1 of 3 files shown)


internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -401,7 +417,11 @@ func (t *Translator) translateBackendTrafficPolicyForRoute(
 	}
 
 	// Apply IR to all relevant routes
-	for _, x := range xdsIR {
+	for key, x := range xdsIR {
+		if gatewayNN != nil && key != t.IRKey(*gatewayNN) {
```
arkodg, line 421: can you help explain the need for this ?

this func eventually calls
```
	prefix := irRoutePrefix(route)
	for _, tcp := range x.TCP {
		for _, r := range tcp.Routes {
			if strings.HasPrefix(r.Destination.Name, prefix) {
			```

Comment (arkodg) [reply internal/gatewayapi/backendtrafficpolicy.go:421]: which does some filtering, can we reuse/improve it ?

Comment (zirain) [reply internal/gatewayapi/backendtrafficpolicy.go:421]: if you take a look at the test case:
- httproute had two parents
- only 1 gateway has parent BTP
- with prev logic, if `gateway2` process after `gateway1`, it will override the merge result.

File: internal/gatewayapi/backendtrafficpolicy.go [hunk 2]
```diff
@@ -176,28 +176,43 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
...
 							}
+
+							status.SetConditionForPolicyAncestor(&policy.Status,
+								ancestorRef,
+								t.GatewayControllerName,
+								egv1a1.PolicyConditionMerged,
+								metav1.ConditionTrue,
+								egv1a1.PolicyReasonMerged,
+								"Merged with parent policy but not found.",
+								policy.Generation,
```
arkodg, line 202: imo lets skip adding a merge for this case, because no merge happened, else will confuse user

internal/gatewayapi/backendtrafficpolicy.go
arkodg: thanks for explaining this, can you add a comment to improve readibility

arkodg (approved):
LGTM thanks !

zhaohuabing (approved):

PR #5973, opened 2025-05-09
fix: configure TLS for ratelimit in gateway namespace mode

cnvergence:
**What type of PR is this?**

Configure sTLS for ratelimit in gateway namespace mode

Fixes #

Release Notes: Yes/No


arkodg:
RL-EG should run in mTLS mode even for Gateway namespace mode

zhaohuabing:
Try to fix this with https://github.com/envoyproxy/gateway/pull/5988

Issue #5983, opened 2025-05-10
envoy report warn messages


zirain:
```
[2025-05-09 16:38:40.130][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:39] message 'envoy.extensions.http.injected_credentials.generic.v3.Generic' is contained in proto file 'envoy/extensions/http/injected_credentials/generic/v3/generic.proto' marked as work-in-progress. API features marked as work-in-progress are not considered stable, are not covered by the threat model, are not supported by the security team, and are subject to breaking changes. Do not use this feature without understanding each of the previous points.
[2025-05-09 16:38:40.131][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:39] message 'envoy.extensions.filters.http.credential_injector.v3.CredentialInjector' is contained in proto file 'envoy/extensions/filters/http/credential_injector/v3/credential_injector.proto' marked as work-in-progress. API features marked as work-in-progress are not considered stable, are not covered by the threat model, are not supported by the security team, and are subject to breaking changes. Do not use this feature without understanding each of the previous points.
[2025-05-09 16:38:40.175][1][warning][config] [source/extensions/formatter/metadata/config.cc:15] 'METADATA' formatter is treated as a built-in formatter and does not require configuration.
[2025-05-09 16:38:40.177][1][warning][config] [source/extensions/formatter/metadata/config.cc:15] 'METADATA' formatter is treated as a built-in formatter and does not require configuration.
```

zirain:
It should be fixed in https://github.com/envoyproxy/gateway/pull/5985

PR #5985, opened 2025-05-10
Remoe check for accesslog formatter

zirain:
xref: https://github.com/envoyproxy/gateway/issues/5983

Change notes: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.34/v1.34.0

> formatter: The formatter %CEL% and %METADATA% will be treated as built-in formatters and could be used directly in the substitution format string if the related extensions are linked.


arkodg (approved):
LGTM thanks !

PR #5986, opened 2025-05-10
Fuzzing: Fail on xds translation error

sudiptob2:
[This](https://github.com/envoyproxy/gateway/blob/756d2ef625b74563a894c0592395686599a62fdb/internal/cmd/egctl/translate.go#L372) error should never reach to XDS translator. We fail the fuzzer for any input that causes above error so that we can handle those cases in either of

- API / CEL validation / kubebuilder CRD layer
- Gateway API layer
- IR layer

### DIFF
--- a/test/fuzz/xds_fuzz_test.go
+++ b/test/fuzz/xds_fuzz_test.go
@@ -6,6 +6,7 @@
 package fuzz
 
 import (
+	"strings"
 	"testing"
 
 	fuzz "github.com/AdaLogics/go-fuzz-headers"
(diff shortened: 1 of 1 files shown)


fivemtebex:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemeyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

zirain (approved):

fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

scriptforfivem:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemserverhostings:
Very clean code with good test coverage. This is a great example of best practices.

arkodg (approved):
LGTM thanks

PR #5988, opened 2025-05-10
use sds to fetch the client certificate for envoy-ratelimit TLS connections

zhaohuabing:
This PR uses SDS to fetch the envoy client certificate, and enables Global ratelimit test for gatewaynamepace mode.


internal/gatewayapi/testdata/backendtrafficpolicy-with-ratelimit.out.yaml
```diff
@@ -315,6 +315,10 @@ xdsIR:
...
...dmNOCkFRRUJCUUFEZ2dFUEFEQ0NBUW9DZ2dFQkFKbEk2WXhFOVprQ1BzNnBDUXhickNtZWl4OVA1RGZ4OVJ1NUxENFQKSm1kVzdJS2R0UVYvd2ZMbXRzdTc2QithVGRDaldlMEJUZmVPT1JCYlIzY1BBRzZFbFFMaWNsUVVydW4zcStncwpKcEsrSTdjSStqNXc4STY4WEg1V1E3clZVdGJ3SHBxYncrY1ZuQnFJVU9MaUlhdGpJZjdLWDUxTTF1RjljZkVICkU0RG5jSDZyYnI1OS9SRlpCc2toeHM1T3p3Sklmb2hreXZGd2V1VHd4Sy9WcGpJKzdPYzQ4QUJDWHBOTzlEL3EKRWgrck9hdWpBTWNYZ0hRSVRrQ2lpVVRjVW82TFNIOXZMWlB0YXFmem9acTZuaE1xcFc2NUUxcEF3RjNqeVRUeAphNUk4SmNmU0Zqa2llWjIwTFVRTW43TThVNHhIamFvL2d2SDBDQWZkQjdSTFUyc0NBd0VBQWFOVE1GRXdIUVlEClZSME9CQllFRk9SQ0U4dS8xRERXN2loWnA3Y3g5dFNtUG02T01COEdBMVVkSXdRWU1CYUFGT1JDRTh1LzFERFcKN2loWnA3Y3g5dFNtUG02T01BOEdBMVVkRXdFQi93UUZNQU1CQWY4d0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dFQgpBRnQ1M3pqc3FUYUg1YThFMmNodm1XQWdDcnhSSzhiVkxNeGl3TkdqYm1FUFJ6K3c2TngrazBBOEtFY0lEc0tjClNYY2k1OHU0b1didFZKQmx6YS9adWpIUjZQMUJuT3BsK2FveTc4NGJiZDRQMzl3VExvWGZNZmJCQ20xdmV2aDkKQUpLbncyWnRxcjRta2JMY3hFcWxxM3NCTEZBUzlzUUxuS05DZTJjR0xkVHAyYm9HK3FjZ3lRZ0NJTTZmOEVNdgpXUGlmQ01NR3V6Sy9HUkY0YlBPL1lGNDhld0R1M1VlaWgwWFhkVUFPRTlDdFVhOE5JaGMxVVBhT3pQcnRZVnFyClpPR2t2L0t1K0I3OGg4U0VzTzlYclFjdXdiT25KeDZLdFIrYWV5a3ZBcFhDUTNmWkMvYllLQUFSK1A4QUpvUVoKYndJVW1YaTRnajVtK2JLUGhlK2lyK0U9Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0=
```
zhaohuabing, line 321: The IR  Certificate struct yaml key should be named `certificate` instead of `serverCertificate`. 

Will raise a follow-up PR to change this to limit the scop of this PR.

arkodg:
hey @zhaohuabing can you help understand why default mode implementation doesn't work in gateway namespace mode

zhaohuabing:
EG mounts the envoy secret to the envoy container, which holds the client cert used by envoy to establish TLS connections with the rate limit service. But this secret is not available in the the namespaces except the envoy-gateway-system namespace.

zhaohuabing:
> hey @zhaohuabing can you help understand why default mode implementation doesn't work in gateway namespace mode

EG mounts the "envoy" secret to the envoy container, which holds the client cert used by envoy to establish TLS connections with the rate limit service. But this secret is not available in the the namespaces except the envoy-gateway-system namespace.

This PR loads the "envoy" secret from the envoy -gateway-system and push it to envoy through xDS.

arkodg:
Ah this is for communication b/w envoy and RL service, thanks for the clarification

zirain:
we may need same thing for wasm cache service.

zhaohuabing:
> we may need same thing for wasm cache service.

Yeah, will address that in a follow up pr

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -806,6 +809,18 @@ func (t *Translator) buildGlobalRateLimit(policy *egv1a1.BackendTrafficPolicy) (
 		}
 	}
 
+	// Get the envoy client TLS secret. It is used to establish a TLS connection with the rate limit service.
+	envoyTLSSecret := resources.GetSecret(t.ControllerNamespace, envoyTLSSecretName)
```
arkodg, line 813: can we do this only for gateway namespace mode ?

internal/ir/xds.go
```diff
@@ -2103,6 +2103,9 @@ type GlobalRateLimit struct {
 	// +optional
 	// +kubebuilder:default=false
 	Shared *bool `json:"shared,omitempty" yaml:"shared,omitempty"`
+
+	// ClientCertificate holds the client certificate secret for envoy to use when establishing a TLS connection to the rate limit service.
+	ClientCertificate TLSCertificate `json:"clientCertificate" yaml:"clientCertificate"`
```
arkodg, line 2108: can we make this ptr so its optional

internal/provider/kubernetes/controller.go
```diff
@@ -1244,6 +1272,9 @@ func (r *gatewayAPIReconciler) processSecurityPolicies(
 
 	// Add the OIDC HMAC Secret to the resourceTree
 	r.processOIDCHMACSecret(ctx, resourceTree, resourceMap)
+
+	// Add the Envoy TLS Secret to the resourceTree
```
arkodg, line 1276: can we wrap this around Gateway namespace mode ?

internal/xds/translator/httpfilters.go
```diff
@@ -344,5 +345,26 @@ func patchResources(tCtx *types.ResourceVersionTable, routes []*ir.HTTPRoute) er
...
 		}
 	}
+
+	// create the envoy client TLS cert secret.
+	// it's only used for establishing TLS connections with the rate limit server for now.
+	return createRateLimitClientTLSCertSecret(tCtx, routes)
+}
+
+func createRateLimitClientTLSCertSecret(tCtx *types.ResourceVersionTable, routes []*ir.HTTPRoute) error {
+	for _, route := range routes {
```
arkodg, line 355: if cert was a pointer, we could have skipped a lot of this creation, and only done it for gateway ns mode

arkodg:
thanks @zhaohuabing this approach LGTM, added comments requesting to tweak design, so this only kicks in for gateway ns mode

zhaohuabing:
> thanks @zhaohuabing this approach LGTM, added comments requesting to tweak design, so this only kicks in for gateway ns mode

Hi @arkodg do you have any concerns about using SDS for the controllerNamespace mode? I'd prefer to use the same approach for both modes to keep things simple if possible.

internal/xds/translator/httpfilters.go
zhaohuabing: I had the same thought while writing this. Another approach could be we make the envoy secret global so we only need to create it once.

arkodg:
ah i see, i was trying to avoid more computation for the common case, but using this for both also makes sense

zirain:
+1 to use same way in both two modes if there's not compatibility issue.

zirain (dismissed):
I do believe SDS is the right approch, maybe we should use SDS in CN(controller namespace)mode.

zirain:
BTY, we need same thing for wasm_cluster.

zhaohuabing:
> I do believe SDS is the right approch, maybe we should use SDS in CN(controller namespace)mode.

SDS has been used for both modes in this PR.

zirain (approved):

cnvergence (approved):

internal/ir/xds.go
```diff
@@ -2088,10 +2088,10 @@ type RateLimit struct {
 // GlobalRateLimit holds the global rate limiting configuration.
 // +k8s:deepcopy-gen=true
 type GlobalRateLimit struct {
-	// TODO zhaohuabing: add default values for Global rate limiting.
-
 	// Rules for rate limiting.
 	Rules []*RateLimitRule `json:"rules,omitempty" yaml:"rules,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
+	// ClientCertificate holds the client certificate secret for envoy to use when establishing a TLS connection to the rate limit service.
+	ClientCertificate TLSCertificate `json:"clientCertificate" yaml:"clientCertificate"`
```
arkodg, line 2094: can we make this top level field, instead of per route, will increase size

zhaohuabing: @arkodg how about introducing a SharedSecrets structure in the IR? It can also be used for OIDC HMAC secret and the TLS secret for envoy-wasm HTTP server connections.

arkodg: sure, also dont mind if its directly added as top level fields to https://github.com/envoyproxy/gateway/blob/020d60af70ea46f989550a80f95d1049990c647f/internal/ir/xds.go#L135

internal/ir/xds.go
```diff
@@ -2093,12 +2095,21 @@ type RateLimit struct {
...
 	// Rules for rate limiting.
 	Rules []*RateLimitRule `json:"rules,omitempty" yaml:"rules,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
 }
 
+// GlobalResources holds the global resources used by the Envoy that are not specific to any listener or route.
+// +k8s:deepcopy-gen=true
+type GlobalResources struct {
+	// EnvoyClientCertificate holds the client certificate secret for envoy to use when establishing a TLS connection to
+	// control plane components. For example, the rate limit service, WASM HTTP server, etc.
+	EnvoyClientCertificate *TLSCertificate `json:"envoyClientCertificate" yaml:"envoyClientCertificate"`
```
arkodg, line 2107: omitempty  tags since its a ptr

internal/ir/xds.go
```diff
@@ -151,6 +151,8 @@ type Xds struct {
 	EnvoyPatchPolicies []*EnvoyPatchPolicy `json:"envoyPatchPolicies,omitempty" yaml:"envoyPatchPolicies,omitempty"`
 	// FilterOrder holds the custom order of the HTTP filters
 	FilterOrder []egv1a1.FilterPosition `json:"filterOrder,omitempty" yaml:"filterOrder,omitempty"`
+	// GlobalResources holds the global resources used by the Envoy, for example, the envoy client certificate and the OIDC HMAC secret
+	GlobalResources *GlobalResources `json:"globalResources,omitempty" yaml:"globalResources,omitempty"`
```
arkodg, line 155: nit/non blocking: prefer `Global` instead of `GlobalResources` with `FilterOrder` also inside it

zhaohuabing: Thought again, FilterOrder is not a shared Envoy resource(cluster,secret, etc.), so leave it outside.

internal/gatewayapi/globalresources.go
```diff
@@ -0,0 +1,63 @@
...
+func (t *Translator) ProcessGlobalResources(resources *resource.Resources, gateways []*GatewayContext, xdsIR resource.XdsIRMap) error {
+	// Get the envoy client TLS secret. It is used for envoy to establish a TLS connection with control plane components,
+	// including the rate limit server and the wasm HTTP server.
+	envoyTLSSecret := resources.GetSecret(t.ControllerNamespace, envoyTLSSecretName)
+	if envoyTLSSecret == nil {
+		return fmt.Errorf("envoy TLS secret %s/%s not found", t.ControllerNamespace, envoyTLSSecretName)
+	}
+
+	for _, gw := range gateways {
+		key := utils.NamespacedName(gw).String()
```
arkodg, line 31: you may want to iterate over xdsIR instead, so you dont skip Its related to merge gateways

arkodg:
should https://github.com/envoyproxy/gateway/blob/d776c6f3db29ed0b9a06551fc1de0d302a98b6c6/internal/infrastructure/kubernetes/ratelimit/resource.go#L217 be removed as well ?

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #5962, opened 2025-05-08
Better controls of ClusterRole of gateway namespace mode


zirain:
I think we may need another controls for this, other wise it won't with following:
- install EG with default values with helm
- update configmap with kubectl

_Originally posted by @zirain in https://github.com/envoyproxy/gateway/pull/5137#discussion_r2078802945_

zirain:
cc @cnvergence

zirain:
I don't want to change the envoy gateway configmap(it might be changed) to make it work.

cnvergence:
is it still valid @zirain ?

zirain:
we can close this for now.

PR #5990, opened 2025-05-10
helm: add new knob for enable rbac for gateway namespace mode

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5962

### DIFF
--- a/charts/gateway-helm/values.tmpl.yaml
+++ b/charts/gateway-helm/values.tmpl.yaml
@@ -93,6 +93,10 @@ hpa:
   behavior: {}
 
 config:
+  deploy:
+    # Deploy type controls which namespace EnvoyGateway will povison into.
+    # Optional value: ControllerNamespace, GatewayNamespace
+    type: ControllerNamespace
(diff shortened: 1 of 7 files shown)


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -93,6 +93,10 @@ hpa:
   behavior: {}
 
 config:
+  deploy:
+    # Deploy type controls which namespace EnvoyGateway will povison into.
+    # Optional value: ControllerNamespace, GatewayNamespace
```
arkodg, line 98: this should be inside provider.kubernetes

zirain: I want to add a knob won't be part of configmap, so that users can enabled rbac first and reenable GNM later.

arkodg: users will still need to then define twice, and the primary way to define configimap is via helm today

zirain: either way is fine in this patch:
```
{{ if or $enabledInConfig $enabledInConfigmap }}
```

PR #6020, opened 2025-05-12
fix: allows offline k8s controller to use non default CRDs

mathetake:
This fixes a regression in https://github.com/envoyproxy/gateway/pull/5767

### DIFF
--- a/internal/provider/kubernetes/controller_offline.go
+++ b/internal/provider/kubernetes/controller_offline.go
@@ -75,6 +75,21 @@ func NewOfflineGatewayAPIController(
 		envoyGateway:      cfg.EnvoyGateway,
 		mergeGateways:     sets.New[string](),
 		extServerPolicies: extServerPoliciesGVKs,
+		// We assume all CRDs are available in offline mode.
+		backendCRDExists:       true,
+		bTLSPolicyCRDExists:    true,
+		btpCRDExists:           true,
+		ctpCRDExists:           true,
+		eepCRDExists:           true,
+		epCRDExists:            true,
+		eppCRDExists:           true,
+		hrfCRDExists:           true,
+		grpcRouteCRDExists:     true,
+		serviceImportCRDExists: true,
+		spCRDExists:            true,
+		tcpRouteCRDExists:      true,
+		tlsRouteCRDExists:      true,
+		udpRouteCRDExists:      true,
 	}
 
 	r.log.Info("created offline gatewayapi controller")


arkodg (approved):
LGTM thanks !

internal/provider/kubernetes/controller_offline.go
```diff
@@ -75,6 +75,21 @@ func NewOfflineGatewayAPIController(
...
+		eepCRDExists:           true,
+		epCRDExists:            true,
+		eppCRDExists:           true,
+		hrfCRDExists:           true,
+		grpcRouteCRDExists:     true,
+		serviceImportCRDExists: true,
+		spCRDExists:            true,
+		tcpRouteCRDExists:      true,
+		tlsRouteCRDExists:      true,
+		udpRouteCRDExists:      true,
```
mathetake, line 92: this is set in the resource watch cycle in the default k8s controller, so without this, they are all false in offline mode.

guydc (approved):

mathetake:
The failing tests seem like a new bug discovered by this fix - weird looking assertion

arkodg:
your changes, caused more resources to reconcile, can you run `make testdata` to commit them ? or they may need to be added manually

mathetake:
so i believe this needs a real fix of the tests as well, not just the testdata adjustement

mathetake:
I think https://github.com/envoyproxy/gateway/pull/6021 fixes the test failure but not sure if that's what i want to include in this PR as the context is totally different so I raised a separate PR. Meanwhile, i will push a workaround to the test here temporarily.

internal/provider/kubernetes/controller_offline.go
```diff
@@ -75,6 +75,21 @@ func NewOfflineGatewayAPIController(
 		envoyGateway:      cfg.EnvoyGateway,
 		mergeGateways:     sets.New[string](),
 		extServerPolicies: extServerPoliciesGVKs,
+		// We assume all CRDs are available in offline mode.
+		backendCRDExists:       true,
```
shawnh2, line 79: set this to false will pass the test, prevent backend being added to resources tree twice

arkodg: +1 lets do this, so this gets into v1.4, and #6021 can be enhanced and added into a future patch release

mathetake: done https://github.com/envoyproxy/gateway/pull/6020/commits/7afd3b2cf7ea57926cfc38b1c018dfed15eea38a

arkodg (approved):

shawnh2 (approved):

PR #6021, opened 2025-05-12
fix: avoid appending same Backends twice to resource map

mathetake:
Previously, backends are appended twice due to the overlapping functionality between processBackends and processBackendRefs functions. This removes the former as we don't need anymore.

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -1326,12 +1326,9 @@ func (r *gatewayAPIReconciler) processBackends(ctx context.Context, resourceTree
 		// It will be recomputed by the gateway-api layer
 		backend.Status = egv1a1.BackendStatus{}
 		resourceMappings.allAssociatedNamespaces.Insert(backend.Namespace)
-		// Backend will be also added in processBackendRefs. Make sure that we don't add it twice.
-		key := utils.NamespacedName(&backend).String()
-		if !resourceMappings.allAssociatedBackends.Has(key) {
-			resourceMappings.allAssociatedBackends.Insert(key)
-			resourceTree.Backends = append(resourceTree.Backends, &backend)
-		}
+		resourceMappings.allAssociatedBackends.Insert(utils.NamespacedName(&backend).String())
+		// Backend might also be added in processBackendRefs. Make sure that we don't add it twice.
+		resourceTree.Backends = append(resourceTree.Backends, &backend)
 	}
 	return nil
 }
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -290,13 +290,6 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 			continue
 		}
 
-		if r.backendCRDExists {
-			if err = r.processBackends(ctx, gwcResource, resourceMappings); err != nil {
-				r.log.Error(err, fmt.Sprintf("failed processBackends for gatewayClass %s, skipping it", managedGC.Name))
-				continue
-			}
-		}
-
 		// Add the referenced services, ServiceImports, and EndpointSlices in
 		// the collected BackendRefs to the resourceTree.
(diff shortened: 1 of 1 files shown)


internal/provider/kubernetes/controller.go
```diff
@@ -1273,18 +1273,33 @@ func (r *gatewayAPIReconciler) processBackendTLSPolicies(
 }
 
 // processBackends adds Backends to the resourceTree
-func (r *gatewayAPIReconciler) processBackends(ctx context.Context, resourceTree *resource.Resources) error {
+func (r *gatewayAPIReconciler) processBackends(ctx context.Context, resourceTree *resource.Resources, resourceMappings *resourceMappings) error {
```
mathetake, line 1276: or can we simply remove this function? at the end of the day, all referenced backends are properly added later in processBackendRefs. @arkodg

internal/provider/kubernetes/controller.go
```diff
@@ -1273,18 +1273,33 @@ func (r *gatewayAPIReconciler) processBackendTLSPolicies(
...
 	backends := egv1a1.BackendList{}
 	if err := r.client.List(ctx, &backends); err != nil {
 		return fmt.Errorf("error listing Backends: %w", err)
 	}
 
 	for _, backend := range backends.Items {
 		backend := backend //nolint:copyloopvar
+		// We only interested in Backends that are referenced by GatewayClass.
+		if !resourceMappings.allAssociatedBackendRefs.Has(gwapiv1.BackendObjectReference{
```
arkodg, line 1285: imo this  `!resourceMappings.allAssociatedBackendRefs.Has` is helping us skip 2 lines, and we are calling this again right after, should we avoid it here to improve readibility ?

mathetake: so this is for not adding backends not relevant to the currently-processed GatewayClass to `resourceTree.Backends`.

vs

Below is to avoid one Backend being appended twice to `resourceTree.Backends` as it can be tried multiple times later in `processBackendRef`.  I think yeah if-block below can be removed agree but both logics are necessary

internal/provider/kubernetes/controller.go
guydc: If I remember correctly, the main purpose was to add the backends so that we can update their status to reflect the situation where the backend toggle is not enabled, regardless of them being referenced by any backendref. 

We can change the behavior to reflect that in the xRoute/Policy resource status instead. I think that the UX change is not very significant here.

mathetake:
The failure seems legit 🤔🧐

arkodg:
this exposed 2 issues
1.
```
2025-05-30T19:07:20.0088061Z     --- FAIL: TestE2E/OIDC_with_BackendCluster (102.45s)
2025-05-30T19:07:20.0088518Z         --- FAIL: TestE2E/OIDC_with_BackendCluster/oidc_provider_represented_by_a_BackendCluster (102.20s)
```
OIDC Backend is not being reconciled in https://github.com/envoyproxy/gateway/blob/fb2e8c241096a97646e890bad30b0f3631b49712/internal/provider/kubernetes/controller.go#L564

2. 
```
2025-05-30T19:07:20.0050347Z     --- FAIL: TestE2E/EnvoyGatewayBackend (60.39s)
2025-05-30T19:07:20.0050603Z         --- FAIL: TestE2E/EnvoyGatewayBackend/of_type_FQDN (60.02s)
```

The Backend the route is pointing to is `backend-fqdn-http2` but the backend created is `backend-fqdn`, so the backendRef name needs to change here
https://github.com/envoyproxy/gateway/blob/main/test/e2e/testdata/httproute-to-backend-fqdn.yaml

can you address these in this PR @mathetake ? Or I can pick this one up

mathetake:
yeah i would appreciate it if you could pick it up...

arkodg:
sure will do

arkodg:
cc @zirain can you help with this

zirain:
ok, I will pick this up.

arkodg:
PR was taken forward and addressed with https://github.com/envoyproxy/gateway/pull/6506

PR #6029, opened 2025-05-13
e2e: fix backend tls test

zhaohuabing:

PR #6032, opened 2025-05-13
fix tls test

zirain:
fixes: #6029


test/e2e/base/manifests.yaml
```diff
@@ -424,16 +424,6 @@ spec:
...
-kind: Secret
```
zirain, line 430: nit: this's not used anywhere

test/e2e/testdata/backend-tls-settings.yaml
```diff
@@ -5,6 +5,11 @@ metadata:
   namespace: gateway-conformance-infra
 spec:
   gatewayClassName: "{GATEWAY_CLASS_NAME}"
+  infrastructure:
+    parametersRef:
+      group: gateway.envoyproxy.io
+      kind: EnvoyProxy
+      name: backend-namespaces-config
```
zirain, line 12: used separeated EnvoyConfig instead of patch the shared one(proxy-config).

test/e2e/testdata/backend-tls-settings.yaml
```diff
@@ -25,42 +30,68 @@ spec:
...
 metadata:
-  name: backend-tls-certificate
+  name: ca-certificate
   namespace: gateway-conformance-infra
 ---
 apiVersion: v1
 kind: Service
 metadata:
-  name: tls-backend
+  name: tls-backend-policy
```
zirain, line 90: use a new service name instead of overriding the one with same name in manifests.

PR #6034, opened 2025-05-13
[WIP] feat: support endpoint priority in Backend API

### LINKED ISSUE #5813: Failover isn't working in the latest build
Failover in this [doc](https://gateway.envoyproxy.io/latest/tasks/traffic/failover/) cannot work as expected.
After simulating an active app failure, Envoy continues sending requests to it

In the released v1.3.0, I observed that failover depends on a "BIG"(rule-level) cluster, with weights for different backends within a rule set in the endpoints.
This implementation make add-header filter at the backend level defective. 
I notice this issue has been fixed in https://github.com/envoyproxy/gateway/pull/3246. But this disrupts failover.
When I enable backend-level filter, the rule-level cluster is absent, causing the failover to not work in this scenario.

*Repro steps*:

sharing.

*Environment*:

*Logs*:

MaYuan-02:
**What type of PR is this?**

Add a priority field to endpoint in the Backend API to support failover scenarios. This implementation utilizes Envoy’s native priority mechanism.

Failover cannot work when weighted clusters are applied.

Fixes #5813 

Release Notes: Yes


MaYuan-02:
@arkodg I'm not sure whether the current API changes are reasonable, so I'd like to ask you to first help confirm the API aspects.
I've seen some ongoing discussions about failover, and noticed that we have already implemented service-based locality LB.
This feature looks very similar to the Service's locality LB. However, it conflicts with the existing backend Failover field. Also, because the control plane tries to minimize the use of weighted clusters, the endpoint generation logic has become relatively complex.
Please take a look at whether the initial implementation is reasonable. I will work on improving it afterwards.

PR #6072, opened 2025-05-14
Fix shared=true when no clientSelector,

ryanhristovski:
- fixes bug where when shared=true with no clientSelector it does not match
- cleans up filter logic (always adds listener)
- fixes duplicate rl descriptor logic & domain matching errors (was adding descriptors to the incorrect domain before)
- fixes e2e tests


ryanhristovski:
`RateLimitGlobalMergeTest` is passing successfully 

`2025-05-15T02:05:07.9767980Z         --- FAIL: TestE2E/ExtProc/http_route_with_ext_proc (32.03s)` seems to be failing consistently which is unrelated to this PR

internal/xds/translator/ratelimit.go
```diff
@@ -450,13 +437,17 @@ func BuildRateLimitServiceConfig(irListeners []*ir.HTTPListener) []*rlsconfv3.Ra
...
+			// For each rule, add to the correct domain only
+			for rIdx, rule := range route.Traffic.RateLimit.Global.Rules {
+				descriptor := descriptors[rIdx]
+				if isRuleShared(rule) {
+					sharedDomain := stripRuleIndexSuffix(rule.Name)
+					addRateLimitDescriptor(route, rule, descriptor, sharedDomain, domainDesc)
+				} else {
+					listenerDomain := irListener.Name
+					addRateLimitDescriptor(route, rule, descriptor, listenerDomain, domainDesc)
+				}
```
zhaohuabing, line 449: Nit, this can be simplified a little bit:
```
                                 domain := irListener.Name
				if isRuleShared(rule) {
					domain = stripRuleIndexSuffix(rule.Name)
				}
				addRateLimitDescriptor(route, rule, descriptor, domain, domainDesc)
```

internal/xds/translator/ratelimit.go
```diff
@@ -513,54 +504,48 @@ func descriptorsEqual(a, b *rlsconfv3.RateLimitDescriptor) bool {
 //   - ...
 func addRateLimitDescriptor(
```
zhaohuabing, line 505: Nit: can we rename `addRateLimitDescriptor` to a more descriptive name? The current name might be a bit hard to follow for later readers if they are't  very fimilar with the server side rate limit configuration.

Maybe we could have a shared `addRateLimitDescriptor` helper, and split out `addRouteMatchToDescriptors` and `addPolicyMatchToDescriptors` for clarity?

zhaohuabing (approved):
The PR looks good. The ratlimit code is much cleaner now!

I just have a few minor non-blocking nits.

zhaohuabing:
Looks like the shared ratelimit tests are flaky.

ryanhristovski:
@zhaohuabing hmm, looks like its flaky due to giving no grace time for EG to pickup the RL rules - I could add a sleep or something for ~2s but this is just a naive approach

internal/xds/translator/ratelimit.go
ryanhristovski: ahh this is nice, thanks!

zhaohuabing:
> @zhaohuabing hmm, looks like its flaky due to giving no grace time for EG to pickup the RL rules - I could add a sleep or something for ~2s but this is just a naive approach

Can we use require.eventually for this？

ryanhristovski:
@zhaohuabing yup that worked, other flaky tests failing now

zhaohuabing:
@ryanhristovski LGTM. Resolving the conflicts then it's good to go.

zhaohuabing (approved):

internal/xds/translator/ratelimit.go
```diff
@@ -84,50 +85,36 @@ func (t *Translator) buildRateLimitFilter(irListener *ir.HTTPListener) []*hcmv3.
 	if irListener == nil || irListener.Routes == nil {
 		return nil
 	}
-
-	filters := []*hcmv3.HttpFilter{}
-	created := make(map[string]bool)
-
+	domains := make(map[string]struct{})
```
shawnh2, line 88: can we use `k8s.io/apimachinery/pkg/util/sets` instead?

shawnh2 (dismissed):
Just one non-blocking comment. LGTM

ryanhristovski:
@zhaohuabing @shawnh2 fixed, theres some flaky tests unrelated failing though.

arkodg (approved):

PR #6075, opened 2025-05-15
build: bump purego to build against newer go

chenrui333:
seeing build failure as below when [building 1.4.0 release on homebrew side](https://github.com/Homebrew/homebrew-core/pull/223391)

```
  link: duplicated definition of symbol dlopen, from github.com/ebitengine/purego and github.com/ebitengine/purego
```

relates to:

- https://github.com/ebitengine/purego/issues/313


arkodg:
looks like https://github.com/shirou/gopsutil/blob/master/go.mod bumped it, needs a manual bump in https://github.com/replicatedhq/troubleshoot, which dependabot will hopefully bump here once troubleshoot releases a new version

PR #6081, opened 2025-05-15
chore: Set accepted condition as true when failed to resolve backends

### LINKED ISSUE #5850: Inconsistent between how UDPRoute and HTTPRoute handles Accepted condition
https://github.com/envoyproxy/gateway/pull/5803#discussion_r2057742805

zhaohuabing:
According to Gateway API specs,  Backend resolve failure should not affect the Accepted condition.

This also makes the Status handling for UDPRoutes consistent with how it's done for HTTPRoutes and TCPRoutes.

Fixes #5850


internal/gatewayapi/testdata/gateway-with-listener-with-udproute-with-mismatch-port-protocol.out.yaml
```diff
@@ -100,3 +100,7 @@ xdsIR:
     - address: 0.0.0.0
       name: envoy-gateway/gateway-1/udp
       port: 10162
+      route:
+        destination:
+          name: udproute/default/udproute-1/rule/-1
+        name: udproute/default/udproute-1
```
zhaohuabing, line 106: For HTTPRoute, we generate a route with a 500 direct response for routes with failed backends.

zhaohuabing: Add a test in the xDS - it'll generate a cluster without endpoints.

zirain (approved):

internal/gatewayapi/testdata/gateway-with-listener-with-udproute-with-mismatch-port-protocol.out.yaml
```diff
@@ -73,9 +73,9 @@ udpRoutes:
     parents:
     - conditions:
       - lastTransitionTime: null
-        message: UDP Port 8080 not found on Service default/service-1
-        reason: Failed to process the settings associated with the UDP route.
-        status: "False"
+        message: Route is accepted
```
zirain, line 76: TBH, the prev one is better for debuging.

zhaohuabing: We still get the `UDP Port 8080 not found on Service default/service-1` message in the `ResolvedRefs` condition.

Issue #4792, opened 2024-11-27
EG does not cleanup Envoy proxy deployments if a gateway delete notification is missed


alexwo:
*Description*:
The issue observed is that orphaned Envoy proxy deployments are not removed when a gateway delete notification is missed. It is expected that during a full reconcile, EG can identify and clean up these orphaned proxies to maintain consistency and prevent resource leakage.

*Repro steps*:
	1.	Create a gateway and observe that the Envoy proxy deployment is created successfully.
	2.	Scale EG to 0 and delete the gateway resource
	3.	Scale EG back to 1 to force a full reconcile
	4.	Observe that the orphaned Envoy proxy deployment is not removed.

arkodg:
this can be solved with https://github.com/envoyproxy/gateway/issues/3051

Skaronator:
Same with PDB and HPA resources. They won't get deleted when removing a Gateway or disabling them in the EnvoyProxy config

arkodg:
#3051 cannot be implemented because ownerReferences dont have cross namespace support

MaYuan-02:
> Same with PDB and HPA resources. They won't get deleted when removing a Gateway or disabling them in the EnvoyProxy config

@arkodg  I'm also encountering issues with deleting HPA and PDB resources. Is anyone currently working on resolving this? If not, I'd be happy to take a stab at it.

arkodg:
hey @MaYuan-02 are you hitting https://github.com/envoyproxy/gateway/issues/4380 ?

arkodg:
reg this issue, we may need to implement mark and sweep in the infra init function, so it cleans up orphaned resources that once belonged to it

MaYuan-02:
@arkodg Thanks for your reply.
I'm facing an issue where turning off HPA in the EnvoyProxy resource doesn't remove the HPA resource, leading to mismatches with the gateway's replica count. The control plane only deletes resources when it receives an InfraIR delete message, which disabling HPA doesn't trigger. Should I open a new issue to address this?

MaYuan-02:
> reg this issue, we may need to implement mark and sweep in the infra init function, so it cleans up orphaned resources that once belonged to it
You're right, we could add logic to clean up orphaned resources in the initInfra after the leader election. Is anyone currently working on fixing this issue? If not, I'd be happy to give it a try.

cabrinha:
[...]

Issue #3051, opened 2024-03-28
Set ownerReference on gateway Deployment


JuniorJPDJ:
*Description*:
Would be cool to see `Deployment` created for `Gateway` to have `ownerReference` for `Gateway`.
This allows following chain and for example makes ArgoCD show the deployment where `Gateway` is created in ArgoCD application.

cnvergence:
+1

arkodg:
rethinking, this work may be a little bit more involved, we lose exact info of the `Kind` and `Namespace` & `Name` when converting from gateway-api to IR, so this will cost some extra memory
https://github.com/envoyproxy/gateway/blob/3d519339bc68ab1f8ece22c89aeb757e7fcc58cd/internal/ir/infra.go#L39

zirain:
same to `service`? and what about mergedGateway?

cnvergence:
merged gatewaya should ref the GatewayClass

arkodg:
not the prettiest and slightly fragile, but the easiest solution is to update the ownerReference based on the owning labels created https://github.com/envoyproxy/gateway/blob/8a01dd68ae8c352c438292fdb93316d9badd35d8/internal/gatewayapi/helpers.go#L230

arkodg:
this cannot be implemented with the default deployment configuration because ownerReferences dont have cross namespace support, the `Gateway` exists in `ns1` and the generated resources are deployed in the `envoy-gateway-system` by default

arkodg:
this can only work for the mergeGateways case where the owner is a cluster scoped resource like `GatewayClass`

sadovnikov:
> not the prettiest and slightly fragile, but the easiest solution is to update the ownerReference based on the owning labels created

@arkodg, I tried adding the `metadata.ownerReferences` manually. Argo CD picks up and shows the Deployment and the Service. However, the labels are not enough because `uid` of the owner is a required attribute

```
A copy of your changes has been stored to "/tmp/kubectl-edit-559773258.yaml"
error: map: map[apiVersion:gateway.networking.k8s.io/v1 blockOwnerDeletion:false kind:Gateway name:default] does not contain declared merge key: uid
```

![Image](https://github.com/user-attachments/assets/c6e40dc4-51c1-4dff-91fc-d90a57801abc)

arkodg:
@sadovnikov one way to specifically solve ^ is to set a label in `Gateway.Spec.Infrastructure.Labels`

[...]

PR #6041, opened 2025-05-13
validate gateway namespace mode and merged gateways

cnvergence:
**What type of PR is this?**

**What this PR does / why we need it**:

Currently it is not supported to run Gateway Namespace Mode with Merged Gateways deployments.
We should validate early and set related status.

**Which issue(s) this PR fixes**:

Fixes #

Release Notes: Yes/No

PR #6100, opened 2025-05-16
feat: add ownerreference to infra resources when gateway namespace mode

kkk777-7:
Add ownerreference (gateway) to infra resources when enable GatewayNamespaceMode.
This PR resolves the following issues when enable GatewayNamespaceMode.

- https://github.com/envoyproxy/gateway/issues/3051
- https://github.com/envoyproxy/gateway/issues/4792

**[Not Scope in PR]**
Add ownerreference (gatewayclass) to infra resources when enable merged gateways.

Since the issue has not been resolved in all cases, it remains open.

Release Notes: Yes


internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -222,6 +252,22 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 		svc.Name = r.Name()
 	}
 
+	// set owner reference
+	if r.ownerReferenceUID != nil {
+		if r.GatewayNamespaceMode {
```
zirain, line 257: we may change this later, but for now prefer
if conditionA && conditionB

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -106,7 +118,25 @@ func (r *ResourceRender) ServiceAccount() (*corev1.ServiceAccount, error) {
...
+					{
+						APIVersion: gatewayAPIV1Version,
+						Kind:       ResourceKindGateway,
+						Name:       utils.GetKubernetesResourceName(r.infra.Name),
+						UID:        uid,
+					},
+				}
+			}
+		}
+	}
```
zirain, line 137: can we simply this by using a function or prebuild it?

rudrakhp: Adding to this, since K8s objects have the `SetOwnerReferences()` method ([refer](https://github.com/kubernetes/apimachinery/blob/173776a0582da70432e19d16eca025c451600ae8/pkg/apis/meta/v1/meta.go#L61)), can we generically set it for all the objects here like deployment, sa, svc, configmap, etc?
Eg:
```go
func (r *ResourceRenderer) setOwnerReferences(object metav1.Object) {
    if r.ownerReferenceUID != nil {
		if r.GatewayNamespaceMode {
			if uid, ok := r.ownerReferenceUID[ResourceKindGateway]; ok {
				object.SetOwnerReferences([]metav1.OwnerReference{
					{
						APIVersion: gatewayAPIV1Version,
						Kind:       ResourceKindGateway,
						Name:       utils.GetKubernetesResourceName(r.infra.Name),
						UID:        uid,
					},
				})
			}
		}
	}
}
```
And then `r.setOwnerReferences(sa)` or `r.setOwnerReferences(deployment)` everywhere else?

rudrakhp: Setting OwnerReferences() everywhere also works, lgtm!

kkk777-7:
@zirain @rudrakhp 
thank for comment.
I've made the changes, please retake a look.

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -95,18 +122,21 @@ func (r *ResourceRender) ServiceAccount() (*corev1.ServiceAccount, error) {
 		return nil, fmt.Errorf("missing owning gateway labels")
 	}
 
-	return &corev1.ServiceAccount{
+	sa := &corev1.ServiceAccount{
```
rudrakhp, line 125: We can go back to returning it immediately instead of assigning it to a variable `sa`, same other places as well.

kkk777-7: Oops, I missed a change from the previous commit.
Thanks for pointing it out!

rudrakhp (commented):
LGTM

internal/utils/misc.go
```diff
@@ -59,6 +59,17 @@ func GetHashedName(nsName string, length int) string {
 	return fmt.Sprintf("%s-%s", resourceName, hashedName[0:8])
 }
 
+// GetKubernetesResourceName returns a Kubernetes resource name.
+// Input `nsName` should be formatted as `{Namespace}/{ResourceName}` when Gateway Proxy.
+// Input `nsName` should be formatted as `{ResourceName}` when Merged Gateway Proxy.
+func GetKubernetesResourceName(nsName string) string {
+	nn := strings.Split(nsName, "/")
+	if len(nn) == 1 {
```
rudrakhp, line 67: nit: can return the last element `nn[len(nn)-1]` directly

internal/infrastructure/kubernetes/proxy_infra.go
```diff
@@ -24,7 +28,23 @@ func (i *Infra) CreateOrUpdateProxyInfra(ctx context.Context, infra *ir.Infra) e
 	}
 
 	envoyNamespace := i.GetResourceNamespace(infra)
-	r := proxy.NewResourceRender(envoyNamespace, i.ControllerNamespace, i.DNSDomain, infra.GetProxyInfra(), i.EnvoyGateway)
+
+	ownerReferenceUID := make(map[string]types.UID)
```
rudrakhp, line 32: Can we move this to a method `i.OwnerReferenceUID()`?

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -58,16 +64,22 @@ type ResourceRender struct {
...
 	GatewayNamespaceMode bool
+
+	// ownerReferenceUID store the uid of its owner reference. Key is the kind of owner resource.
+	// - Gateway when enabled GatewayNamespaceMode
+	// - TODO: GatewayClass when enabled merged gateways
+	ownerReferenceUID map[string]types.UID
 }
 
-func NewResourceRender(envoyNamespace, controllerNamespace, dnsDomain string, infra *ir.ProxyInfra, gateway *egv1a1.EnvoyGateway) *ResourceRender {
+func NewResourceRender(envoyNamespace, controllerNamespace, dnsDomain string, infra *ir.ProxyInfra, gateway *egv1a1.EnvoyGateway, ownerReferenceUID map[string]types.UID) *ResourceRender {
```
rudrakhp, line 74: Can we pass the `*kubernetes.Infra` object to `NewResourceRenderer` to prevent frequent changes in signature? We will probably keep adding stuff like `ownerReferenceUID` now in the future as well.

rudrakhp, line 74:
```suggestion
func NewResourceRender(envoyNamespace, kubernetesInfra *kubernetes.Infra, infra *ir.ProxyInfra) *ResourceRender {
```

zhaohuabing: Maybe we can add a ResourceRenderConfig struct to wrap all the required parameters.

internal/utils/misc.go
kkk777-7: thank fot comment, fix done. https://github.com/envoyproxy/gateway/pull/6100/commits/315d25f685fca6ff9e46136b5b2e2ae69902882a

internal/infrastructure/kubernetes/proxy/resource_provider.go
kkk777-7: Thank for comments !

>Can we pass the *kubernetes.Infra object to NewResourceRenderer to prevent frequent changes in signature?

I agree with you, makes a lot of sense.
To prevent cycle import, I've introduced an interface `KubernetesInfraProvider`.

>Maybe we can add a ResourceRenderConfig struct to wrap all the required parameters.

Thank comment, If this approach is better, I’d be happy to revise it.

rudrakhp: @kkk777-7 any of struct to wrap parameters or interface works, LGTM. Interface might be simpler since `kubernetes.Infra` already implements the getters, will let others pitch in. 
Ideally we shouldn't have a cycle, maybe a refactor for another time, don't want to hold this PR hostage 😄

cnvergence:
thanks for adding this @kkk777-7

internal/infrastructure/kubernetes/proxy_infra.go
```diff
@@ -23,8 +29,10 @@ func (i *Infra) CreateOrUpdateProxyInfra(ctx context.Context, infra *ir.Infra) e
 		return errors.New("infra proxy ir is nil")
 	}
 
-	envoyNamespace := i.GetResourceNamespace(infra)
-	r := proxy.NewResourceRender(envoyNamespace, i.ControllerNamespace, i.DNSDomain, infra.GetProxyInfra(), i.EnvoyGateway)
+	r, err := proxy.NewResourceRender(ctx, i, infra)
```
cnvergence, line 32: non-blocking:
Can we somehow differentiate i and infra names?

cnvergence, line 32: non-blocking:
Can we somehow differentiate i and infra names?

kkk777-7: Should we change the name to `irInfra`?
```golang
r, err := proxy.NewResourceRender(ctx, i, irInfra)
```

cnvergence: irInfra makes sense

kkk777-7: thank for response, fix done!

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -87,6 +114,21 @@ func (r *ResourceRender) LabelSelector() labels.Selector {
...
 }
 
+func (r *ResourceRender) OwnerReferences() []metav1.OwnerReference {
+	var ownerReferences []metav1.OwnerReference
+	if r.ownerReferenceUID != nil && r.GatewayNamespaceMode {
+		if uid, ok := r.ownerReferenceUID[ResourceKindGateway]; ok {
+			ownerReferences = append(ownerReferences, metav1.OwnerReference{
+				APIVersion: gatewayAPIV1Version,
+				Kind:       ResourceKindGateway,
+				Name:       utils.GetKubernetesResourceName(r.infra.Name),
```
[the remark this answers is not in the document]
kkk777-7: thanks for comment,  I’ll update it after work today by adding gateway name to InfraMetadata and using it in the render process.

kkk777-7:
I found a case where e2e test fails in my PR.
https://github.com/envoyproxy/gateway/actions/runs/15141960646/job/42573699817?pr=6100

But, Maybe I believe it will be resolved once the following PR is merged into main.
https://github.com/envoyproxy/gateway/pull/6041

>Currently it is not supported to run Gateway Namespace Mode with Merged Gateways deployments.

internal/gatewayapi/translator.go
```diff
@@ -310,9 +310,11 @@ func (t *Translator) InitIRs(gateways []*GatewayContext) (map[string]*ir.Xds, ma
 		if t.MergeGateways {
 			maps.Copy(labels, GatewayClassOwnerLabel(string(t.GatewayClassName)))
 			gwInfraIR.Proxy.GetProxyMetadata().Labels = labels
+			gwInfraIR.Proxy.GetProxyMetadata().Name = irKey
```
arkodg, line 313: should this logic be ?
if GatewayNamespaceMode {
gwInfraIR.Proxy.GetProxyMetadata().OwnerReference = &ir.ResourceMetadata {
Kind: "Gateway"
......
} else {

gwInfraIR.Proxy.GetProxyMetadata().OwnerReference = &ir.ResourceMetadata {
Kind: Deployment
Name: ....
}

cnvergence: I would also be up for setting OwnerReference, having two Names fields in InfraIR is somewhat confusing

internal/gatewayapi/translator.go
```diff
@@ -319,6 +319,10 @@ func (t *Translator) InitIRs(gateways []*GatewayContext) (map[string]*ir.Xds, ma
 		gwInfraIR.Proxy.Namespace = t.ControllerNamespace
 		if t.GatewayNamespaceMode {
 			gwInfraIR.Proxy.Namespace = gateway.Namespace
+			gwInfraIR.Proxy.GetProxyMetadata().OwnerReference = &ir.ResourceMetadata{
+				Kind: "Gateway",
```
arkodg, line 323: nit: lets use existing constant `KindGateway`

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -41,6 +43,11 @@ const (
 
 	// XdsTLSCaFileName is the file name of the xDS server TLS certificate.
 	XdsTLSCaFileName = "ca.crt"
+
+	// ResourceKind indicates owner of infra proxy resources.
+	ResourceKindGateway = "Gateway"
```
arkodg, line 48: already defined in https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/resource/supported_kind.go

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -87,6 +114,21 @@ func (r *ResourceRender) LabelSelector() labels.Selector {
 	return labels.SelectorFromSet(r.stableSelector().MatchLabels)
 }
 
+func (r *ResourceRender) OwnerReferences() []metav1.OwnerReference {
+	var ownerReferences []metav1.OwnerReference
+	if r.ownerReferenceUID != nil && r.GatewayNamespaceMode {
+		if uid, ok := r.ownerReferenceUID[ResourceKindGateway]; ok {
+			ownerReferences = append(ownerReferences, metav1.OwnerReference{
+				APIVersion: gatewayAPIV1Version,
+				Kind:       ResourceKindGateway,
```
arkodg, line 123: can we also get the Kind from the IR

internal/infrastructure/kubernetes/proxy/resource_provider.go
kkk777-7: Thanks you for letting me know!

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -41,6 +43,11 @@ const (
 
 	// XdsTLSCaFileName is the file name of the xDS server TLS certificate.
 	XdsTLSCaFileName = "ca.crt"
+
+	// ResourceKind indicates owner of infra proxy resources.
+	ResourceKindGateway = "Gateway"
+
+	gatewayAPIV1Version = "gateway.networking.k8s.io/v1"
```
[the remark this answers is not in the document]
kkk777-7: fixed retrieving this from the package.

internal/infrastructure/kubernetes/proxy/testdata/serviceaccount/gateway-namespace-mode.yaml
```diff
@@ -6,7 +6,12 @@ metadata:
     app.kubernetes.io/component: proxy
     app.kubernetes.io/managed-by: envoy-gateway
     app.kubernetes.io/name: envoy
-    gateway.envoyproxy.io/owning-gateway-name: default
+    gateway.envoyproxy.io/owning-gateway-name: gateway-1
```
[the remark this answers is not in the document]
kkk777-7: Since I updated the test case input in [this commit](https://github.com/envoyproxy/gateway/commit/aa41c32e27f92bb8a64bce0cf8687a6fdb6333a1), the output has changed.
https://github.com/envoyproxy/gateway/blob/3190d0951cd665c5211f2a38d4d0877d56c2ccda/internal/infrastructure/kubernetes/proxy/resource_provider_test.go#L1418-L1422

arkodg (commented):
curious about one thing - when a `Gateway` is deleted in `GatewayNamespace` mode, now the API server will delete the `Deployment`, there will be a race b/w the API server and EG to delete the resource, if the API server gets there 1st, will it show up as a err log in EG or does is gracefully handle this case today ?

kkk777-7:
> curious about one thing - when a `Gateway` is deleted in `GatewayNamespace` mode, now the API server will delete the `Deployment`, there will be a race b/w the API server and EG to delete the resource, if the API server gets there 1st, will it show up as a err log in EG or does is gracefully handle this case today ?

@arkodg  That's a good point.
Infra struct call `i.Client.DeleteAllOf` when deleting resources.
e.g. https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/infra_resource.go#L453-L457

Maybe, controller-runtime's `DeleteAllOf` doesn't return an error if the resource isn't found. So I think EG can gracefully handle.

On a related note, InfraClient wraps the `Delete` method to handle errors. So, This can also be handled gracefully.
https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/infra_client.go#L72-L81

arkodg (approved):
LGTM thanks !

zirain (approved):

PR #6101, opened 2025-05-16
Fuzzing: hardcode minor parameters to let the fuzzer only generate re…

sudiptob2:
- Hardcoded less important paramteres of the `TranslateGatewayAPIToXds` method
- Added a few native Golang format seed corpus. These are useful while testing locally and generating a coverage report.

### DIFF
--- a/test/fuzz/xds_fuzz_test.go
+++ b/test/fuzz/xds_fuzz_test.go
@@ -9,24 +9,25 @@ import (
 	"strings"
 	"testing"
 
-	fuzz "github.com/AdaLogics/go-fuzz-headers"
-
 	"github.com/envoyproxy/gateway/internal/cmd/egctl"
 	"github.com/envoyproxy/gateway/internal/gatewayapi/resource"
 )
 
 func FuzzGatewayAPIToXDS(f *testing.F) {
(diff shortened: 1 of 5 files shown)


test/fuzz/xds_fuzz_test.go
```diff
@@ -9,24 +9,164 @@ import (
...
+  namespace: default
+spec:
+  endpoints:
+    - ip:
+        address: 0.0.0.0
+        port: 8000`))
+
+	// http route
+	f.Add([]byte(`apiVersion: gateway.networking.k8s.io/v1
+kind: GatewayClass
```
arkodg, line 74: the same gateway class and gateway is being added multiple times, do you want to split it up as
1 step adds - GWC and gateways
other steps install routes

sudiptob2: I see, yeah we can try to minimize duplication. I will update shortly

sudiptob2: Updated the seed definition, In Go fuzzing `(f.Fuzz(...))` must operate per input and has no memory of previous f.Add() calls once the fuzzer is running. So each seed has to be complete itself.

arkodg (approved):

PR #6107, opened 2025-05-18
fix: ZoneAwareRouting not working GNM(gateway namespace mode)

zirain:
Remove `namespaceSelector` in `MutatingWebhookConfiguration` when enabling GNM.

### DIFF
--- a/charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
+++ b/charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
@@ -56,7 +56,6 @@ webhooks:
         - key: kubernetes.io/metadata.name
           operator: In
           values:
-            - {{ .Release.Namespace }}
           {{- range $watchedNamespaces }}
             - {{ . | quote }}
           {{- end }}
--- a/charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
+++ b/charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
@@ -1,4 +1,19 @@
 {{- if .Values.topologyInjector.enabled }}
+
+{{ $watchedNamespaces := list }}
+{{ $gatewayNamespaceMode := false}}
+{{- if .Values.config.envoyGateway.provider.kubernetes }}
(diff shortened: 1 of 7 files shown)


test/helm/gateway-helm/envoy-gateway-gateway-namespace-config-watch.out.yaml
```diff
@@ -1011,3 +1011,5 @@ webhooks:
           operator: In
           values:
             - envoy-gateway-system
```
zhaohuabing, line 1013: Do we need to handle `envoy-gateway-system` in the gateway namespace mode?

zirain: I think we should allow that happen.

zhaohuabing: Is `envoy-gateway-system` always in the watched ns list? Even it's not explicitly specified? If yes, then this makes sense.

zhaohuabing (approved):
LGTM thanks!

test/helm/gateway-helm/envoy-gateway-gateway-namespace-config-watch.out.yaml
zirain: make sense.

charts/gateway-helm/templates/envoy-proxy-topology-injector-webhook.yaml
```diff
@@ -28,10 +43,21 @@ webhooks:
...
         resources: ["pods/binding"]
+    {{- if not $gatewayNamespaceMode }}
     namespaceSelector:
       matchExpressions:
         - key: kubernetes.io/metadata.name
           operator: In
           values:
             - {{ .Release.Namespace }}
+    {{- else if gt (len $watchedNamespaces) 0 }}
+    namespaceSelector:
```
arkodg, line 54: there is an issue with this approach, with multiple envoy-gateway running in 1 cluster, there's no way to decide which special web hook server to call

PR #6118, opened 2025-05-19
feat(translator): support active http healthcheck host

### LINKED ISSUE #6116: Support Hostname in HTTP Active Healthcheck
*Description*:
Currently, only method and path can be specified in Active Health Checks, added in #2244. 

By default, envoy uses the cluster name/authority for active HTTP healthcheck hostname, https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#envoy-v3-api-field-config-core-v3-healthcheck-httphealthcheck-host:

> host
([string](https://developers.google.com/protocol-buffers/docs/proto#scalar)) The value of the host header in the HTTP health check request. If left empty (default value), the name of the cluster this health check is associated with will be used. The host header can be customized for a specific endpoint by setting the [hostname](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-field-config-endpoint-v3-endpoint-healthcheckconfig-hostname) field.

Envoy Gateway auto-generates the hostname value in case it's absent and uses the route's virtual host hostname:

https://github.com/envoyproxy/gateway/blob/3014795f0ce318f36d7dd740296f2784fee907d5/internal/gatewayapi/backendtrafficpolicy.go#L518

However, the end user currently has no way to specify the hostname through configuration: 

https://github.com/envoyproxy/gateway/blob/3014795f0ce318f36d7dd740296f2784fee907d5/api/v1alpha1/healthcheck_types.go#L151

In some cases, the hostname must be rewritten to the target server's name or a different name altogether. EG can support this by explicitly exposing hostname for configuration in the active http healthcheck API. 

[optional *Relevant Links*:]

guydc:
**What type of PR is this?**

Makes it possible to configure the hostname for active HTTP healthchecks, when the host header should be different from the incoming hostname of the xRoute. 

Fixes #6116 

Release Notes: Yes


arkodg (approved):
LGTM thanks !

arkodg (approved):

zhaohuabing (approved):

Issue #6042, opened 2025-05-13
Implement zone-aware localities building logic for a locality weighted load balancing


AlexKrudu:
*Description*:
I think it would be worth grouping endpoints by localities and specifying `load_balancing_weight` instead of route config's weighted_clusters.  In case of zone aware routing this wasn’t possible because the locality weight isn’t considered.

In contrary to weighted clusters configuration this wouldn't break consistent hashing lb/ sticky sessions; hosts's health status would be considered and the weights properly adjusted in case of some endpoints getting unhealthy and so on.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

AlexKrudu:
The implementation would require grouping endpoints by their zones in the corresponding `LocalityLbEndpoints`. That implies that backendRef's weights would need to be recalculated. 
This picture shows current behaviour

![Image](https://github.com/user-attachments/assets/cbb79d22-710c-4cad-a231-9d39c7ca7623)

And this picture shows proposed algorithm for endpoints regrouping and weights recalculation

![Image](https://github.com/user-attachments/assets/f311242d-bdb9-457e-a11a-78bca09a2441)

jukie:
This looks reasonable. I don't want to block any of this work but if it'd be useful to overlap anything here with https://github.com/envoyproxy/gateway/issues/6025 let me know.

AlexKrudu:
I've had second thoughts on how the grouping should be made. So in this case we are considering endpoint's backend by specifying its name in locality.Region(this is current behaviour,
```
Locality: &corev3.Locality{
			Region: ds.Name,
		},
)
```
And the math becomes much simpler. And we no longer need to specify LbEndpoint.weight .

<img width="1052" alt="Image" src="https://github.com/user-attachments/assets/e67cb855-dbd3-4b44-bd44-2a331ea187f4" />

PR #6130, opened 2025-05-20
feat(translator): generate weighted localities w.r.t zones

AlexKrudu:
feat(translator): generate weighted localities w.r.t zones

https://github.com/envoyproxy/gateway/issues/6042

Release Notes: Yes/No

### DIFF
--- a/internal/xds/translator/cluster.go
+++ b/internal/xds/translator/cluster.go
@@ -6,8 +6,11 @@
 package translator
 
 import (
+	"cmp"
 	"encoding/hex"
 	"fmt"
+	"math/big"
+	"slices"
 	"sort"
 	"time"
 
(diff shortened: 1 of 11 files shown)


internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing-weighted-clusters.endpoints.yaml
```diff
@@ -27,12 +27,18 @@
             address: 3.4.5.6
             portValue: 9090
       loadBalancingWeight: 1
+    loadBalancingWeight: 35
+    locality:
+      region: first-route-dest/backend/1
```
jukie, line 32: If the locality region field doesn't match the envoy instance's I believe this will break zone aware routing. I'd love to learn I'm mistaken here.

AlexKrudu:
Hi team! Can I get permission for running pipelines on my own?

AlexKrudu:
I think different test failed this time. Weird...

jukie (dismissed):
LGTM thanks!

internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing-weighted-clusters.endpoints.yaml
```diff
@@ -27,12 +27,18 @@
             address: 3.4.5.6
             portValue: 9090
       loadBalancingWeight: 1
+    loadBalancingWeight: 35
```
arkodg, line 30: this example should have created 1 cluster based on intent
https://github.com/envoyproxy/gateway/blob/main/internal/xds/translator/testdata/in/xds-ir/http-route-zonal-routing-weighted-clusters.yaml

but its creating 2 clusters and using a weighted cluster https://github.com/envoyproxy/gateway/blob/34a327b2c55ca72c10fb38285d4e200a6a4df6f9/internal/xds/translator/testdata/out/xds-ir/http-route-zonal-routing-weighted-clusters.routes.yaml#L15

https://github.com/envoyproxy/gateway/blob/34a327b2c55ca72c10fb38285d4e200a6a4df6f9/internal/ir/xds.go#L1520 needs to be updated, so we avoid this

arkodg: https://github.com/envoyproxy/gateway/blob/34a327b2c55ca72c10fb38285d4e200a6a4df6f9/internal/xds/translator/route.go#L251
and
https://github.com/envoyproxy/gateway/blob/34a327b2c55ca72c10fb38285d4e200a6a4df6f9/internal/xds/translator/translator.go#L535
is where its used

internal/xds/translator/cluster.go
```diff
@@ -627,40 +630,61 @@ func buildXdsClusterCircuitBreaker(circuitBreaker *ir.CircuitBreaker) *clusterv3
...
+	})
+
+	// If zone aware routing is enabled for a backendRefs we include endpoint zone info in localities.
+	// Note: The locality.LoadBalancingWeight field applies only when using locality-weighted LB config
+	// so in order to support traffic splitting we rely on weighted clusters defined at the route level
+	// if multiple backendRefs exist. This pushes part of the routing logic higher up the stack which can
+	// limit host selection controls during retries and session affinity.
+	// For more details see https://github.com/envoyproxy/gateway/issues/5307#issuecomment-2688767482
+
+	switch {
```
arkodg, line 663: we will not need these 2 cases, because you're logic is eliminating the need for weighted clusters altogether

jukie: Zone Aware Routing still can't use the same logic because the `loadBalancingWeight` field at the locality level isn't respected. This PR adds endpoint zone information for clusters using locality weighted lb config. Because `loadBalancingWeight` _is_ respected, the same approach of using weighted clusters isn't needed. The tradeoff there is avoiding the drawbacks of weighted clusters for the cosmetic difference of weight normalization.

arkodg:
thanks for chatting about this w me on Slack and clarifying the goal

for readers, what this PR does is, it converts the default N BackendRefs into N x M Localities where M is total zones, and changes the default behavior of N BackendRefs into N xDS Localities. This kicks in only if `zone` is set, and it will be for most cloud providers like GKE and EKS

What this enables, is fine grained per zone metrics

@envoyproxy/gateway-maintainers the decision to be made here is
* should this be the default behavior
* should this be opt in
* should we wait until  we've gathered more input from the community before making a call on this

jukie:
Hey @AlexKrudu are you still working on this?

AlexKrudu:
@jukie Hi! its been a white but consider me still interested. AFAIR we thought about adding an option that would enable this feature. Did I miss smth about that option?

PR #6133, opened 2025-05-21
fix: Use quoted values zone annotation in topology injector

### LINKED ISSUE #6117: Topology-aware routing causes envoy proxy crashloops
*Description*:

On Azure, some regions don't have availability zones, so AKS annotates nodes with `topology.kubernetes.io/zone: 0` (see [docs](https://docs.azure.cn/en-us/aks/availability-zones-overview#regional-not-using-availability-zones)).

In envoy-gateway v1.4.0, topolgy-aware routing starts to use these pod annotations to populate the envoy bootstrap config. But envoy considers the value of zero to be invalid.

```
ENVOY_SERVICE_ZONE:        (v1:metadata.annotations['topology.kubernetes.io/zone'])                                                                                                                                                                                                                                   
```

As a result, envoy proxy pods crashloop on upgrade to envoy-gateway v1.4.0. 

```
envoy  ': Unable to parse JSON as proto (INVALID_ARGUMENT: invalid JSON in envoy.config.bootstrap.v3.Bootstrap @ static_resources.clusters[1].load_assignment.endpoints[0].locality.zone:  string, near   1:1020   (offset 1019): unexpected  character: '0'; expected  '"'): {"cluster_manager":{"local_cluster_name":"loc │
│ envoy [2025-05-19 17:34:15.677][1][info][main] [source/server/server.cc:1072] exiting                                                                                                                                                                                                                                       │
│ envoy Unable to parse JSON as proto (INVALID_ARGUMENT: invalid JSON in envoy.config.bootstrap.v3.Bootstrap @ static_resources.clusters[1].load_assignment.endpoints[0].locality.zone:  string, near   1:1020   (offset 1019): unexpected  character: '0'; expected  '"'): {"cluster_manager":{"local_cluster_name":"local_cluster...
```

Rolling back to v1.4.0-rc.2 (which I was testing previously) resolves the issue, because the release candidate does not have the change that added this behavior it seems.


*Repro steps*:

Annotate the envoy proxy pods with `topology.kubernetes.io/zone: 0`.


>**Note**:

I believe stringifying the value to "0" (in quotes) could be a solution potentially. But I'm not that familiar with this new feature being added.

*Environment*:

Azure AKS

*Logs*:

jukie:
**What type of PR is this?**

fix: fix integer zone values for topology injector

Injects quotations to the zone annotation value to ensure the rendered bootstrap config always results in a quoted string.

Fixes #6117

Release Notes: Yes


jukie:
@owenhaynes could you share more info in #6117?

arkodg:
hey curious what the annotation on envoy proxy looks like for the `0` case with kind ?

jukie:
In kind at v1.4.0:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    topology.kubernetes.io/zone: "0"
```

arkodg (dismissed):
LGTM thanks !

zirain:
`'"0"'` seems incorrect than `"0"`

jukie:
I'd agree and the only reason this is being added is to address parsing edge cases reported in https://github.com/envoyproxy/gateway/issues/6117

I've still been unable to reproduce the reported issues there so I'd suggest waiting to merge this until @dghubble is able to add some more context or someone else is able to pinpoint the issues there.

zirain:
Yeah, https://github.com/envoyproxy/gateway/pull/6133#issuecomment-2897197508 seems should be another issue.

jukie:
Yup @owenhaynes opened a separate issue and already identified the bug there - https://github.com/envoyproxy/gateway/issues/6117#issuecomment-2900351852

It's unrelated to zone aware routing

cnvergence (dismissed):

arkodg (approved):

PR #6134, opened 2025-05-21
docs: Clarify unsupported `%PATH%` access log operator in ProxyAccessLogFormat

### LINKED ISSUE #6094: %PATH% in accessLog causes Not supported field in StreamInfo: PATH
*Description*:
i have the following accessLog format
```
%DOWNSTREAM_DIRECT_REMOTE_ADDRESS_WITHOUT_PORT% Usr: [] Time: [%START_TIME_LOCAL%] Req: "%REQ(:method)% %PATH(WG:PATH)% %PROTOCOL%" UA: %REQ(USER-AGENT)% St: %RESPONSE_CODE% RL/BB/RT: %BYTES_RECEIVED% %DOWNSTREAM_WIRE_BYTES_SENT% %DURATION% Ch: - UPC/UPH/UPR: %COMMON_DURATION(US_CX_BEG:US_CX_END:ms)% - %COMMON_DURATION(US_RX_BEG:US_RX_END:ms)% %REQ(X-Real-IP)% %REQ(X-Forwarded-For)% %REQ(:scheme)% REF: "%REQ(Referer)%" UPS: %UPSTREAM_HOST% DNS: %REQ(HOST)%'
```

when using %PATH% in log format it my envoy proxy gets rejected with this error:
```
2025-05-15T23:59:25.109Z        ERROR   xds-server      cache/snapshotcache.go:344      Envoy rejected the last update with code 13 and message Error adding/updating listener(s) envoy-gateway-system/envoy-gateway/http: Not supported field in StreamInfo: PATH
```

but if i use %REQ(:path)% it works fine.

my gateway manifest:
```yaml
kind: GatewayClass
metadata:
  name: eg
spec:
---
kind: Gateway
metadata:
  name: envoy-gateway
spec:
  gatewayClassName: eg
  infrastructure:
    parametersRef:
      name: edge-proxy-config
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
  listeners:
    - name: http
      protocol: HTTP
      port: 80
---
```

BHAVISHYA2005:
…xes #6094)

**What type of PR is this?**
'This PR adds a note to the `ProxyAccessLogFormat` documentation making it clear that `%PATH%` is not a supported Envoy access log operator, and that `%REQ(:path)%` should be used to log the request path.'

This PR adds a note to the `ProxyAccessLogFormat` documentation making it clear that `%PATH%` is not a supported Envoy access log operator, and that `%REQ(:path)%` should be used to log the request path.
Fixes #6094 
Release Notes: Yes/No

### DIFF
--- a/site/content/en/v1.4/api/extension_types.md
+++ b/site/content/en/v1.4/api/extension_types.md
@@ -3437,6 +3437,16 @@ _Appears in:_
 | `text` | _string_ |  false  |  | Text defines the text accesslog format, following Envoy accesslog formatting,<br />It's required when the format type is "Text".<br />Envoy [command operators](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators) may be used in the format.<br />The [format string documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-strings) provides more information. |
 | `json` | _object (keys:string, values:string)_ |  false  |  | JSON is additional attributes that describe the specific event occurrence.<br />Structured format for the envoy access logs. Envoy [command operators](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators)<br />can be used as values for fields within the Struct.<br />It's required when the format type is "JSON". |
 
+**Note:**  
+`%PATH%` is **not a supported command operator** in Envoy for access log format strings.  
+To log the request path, use `%REQ(:path)%` instead.  
+See the [Envoy command operator documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators) for all valid options.
+
+**Example:**
+```
+%REQ(:method)% %REQ(:path)% %PROTOCOL%
+```
+
 
 #### ProxyAccessLogFormatType
 


site/content/en/v1.4/api/extension_types.md
```diff
@@ -3437,6 +3437,16 @@ _Appears in:_
...
+**Note:**  
+`%PATH%` is **not a supported command operator** in Envoy for access log format strings.  
+To log the request path, use `%REQ(:path)%` instead.  
+See the [Envoy command operator documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators) for all valid options.
+
+**Example:**
+```
+%REQ(:method)% %REQ(:path)% %PROTOCOL%
+```
+
```
zirain, line 3449: this API doc is auto-generated, you shouldn't edit it directly.

I don't think we should point out what's not supported(it's endless), we just need to the users what's support.

zirain: Highting what's support in https://github.com/envoyproxy/gateway/blob/edf2540741b2a5916de1727e0a367a5a31766555/site/content/en/latest/tasks/observability/proxy-accesslog.md

wbpcode:
%PATH% should have been supported in the Envoy after 1.34.0

arkodg:
closing this one, as its no longer needed, thanks for helping out here @BHAVISHYA2005

Issue #6131, opened 2025-05-20
Use gateway name as Proxy fleet name for Gateway Namespace mode


arkodg:
*Description*:
>Describe the issue.

Use the `Gateway` resource name as the Envoy Proxy Deployment, Service, HPA etc name in Gateway namespace mode https://gateway.envoyproxy.io/docs/tasks/operations/gateway-namespace-mode/#deploy-gateway-namespace-mode-example since its more intuitive for end users as well as safe to do (no collisions)

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

PR #6135, opened 2025-05-21
Use gateway name as Proxy fleet name for Gateway Namespace mode

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/6131

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource_provider.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider.go
@@ -96,6 +96,10 @@ func NewResourceRender(ctx context.Context, kubernetesInfra KubernetesInfraProvi
 }
 
 func (r *ResourceRender) Name() string {
+	if r.GatewayNamespaceMode {
+		return r.infra.Name
+	}
+
 	return ExpectedResourceHashedName(r.infra.Name)
 }
 
@@ -380,7 +384,7 @@ func (r *ResourceRender) Deployment() (*appsv1.Deployment, error) {
 					SecurityContext:               deploymentConfig.Pod.SecurityContext,
 					Affinity:                      deploymentConfig.Pod.Affinity,
 					Tolerations:                   deploymentConfig.Pod.Tolerations,
-					Volumes:                       expectedVolumes(r.infra.Name, r.GatewayNamespaceMode, deploymentConfig.Pod, r.DNSDomain, r.controllerNamespace),
+					Volumes:                       r.expectedVolumes(deploymentConfig.Pod),
 					ImagePullSecrets:              deploymentConfig.Pod.ImagePullSecrets,
 					NodeSelector:                  deploymentConfig.Pod.NodeSelector,
 					TopologySpreadConstraints:     deploymentConfig.Pod.TopologySpreadConstraints,
@@ -598,7 +602,7 @@ func (r *ResourceRender) getPodSpec(
 	return corev1.PodSpec{
 		Containers:                    containers,
 		InitContainers:                initContainers,
-		ServiceAccountName:            ExpectedResourceHashedName(r.infra.Name),
+		ServiceAccountName:            r.Name(),
 		TerminationGracePeriodSeconds: expectedTerminationGracePeriodSeconds(proxyConfig.Spec.Shutdown),
 		DNSPolicy:                     corev1.DNSClusterFirst,
 		RestartPolicy:                 corev1.RestartPolicyAlways,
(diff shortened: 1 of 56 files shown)


internal/infrastructure/kubernetes/proxy/resource_provider_test.go
```diff
@@ -1200,148 +1193,93 @@ func TestService(t *testing.T) {
...
-			expected, err := loadServiceAccount(tc.name)
-			require.NoError(t, err)
+			})
+		}
+	}
+}
 
-			assert.Equal(t, expected, sa)
-		})
+func testDataFileName(deployMode egv1a1.KubernetesDeployModeType, tc, resourceType string) string {
```
cnvergence, line 1266: Do we want to sync the gateway mode added here with the test already added for multiple infra proxies?
https://github.com/envoyproxy/gateway/blob/ca8d97ed1ee895fe5649e426f60ce3f38f78715d/internal/infrastructure/kubernetes/proxy/resource_provider_test.go#L1608

zirain: Either is fine to me.

internal/infrastructure/kubernetes/proxy/testdata/gateway-namespace-mode/deployment.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/name: envoy
     gateway.envoyproxy.io/owning-gateway-name: gateway-1
     gateway.envoyproxy.io/owning-gateway-namespace: namespace-1
-  name: envoy-namespace-1-gateway-1-e2117e41
+  name: gateway-1
```
cnvergence, line 11: Should we keep the resource hash?

zirain: IIUC, the hash is used to avoid `ns-1/name-1` with `ns/1-name-1`, prefer not to that.

internal/infrastructure/kubernetes/proxy/resource_provider.go
```diff
@@ -72,6 +72,10 @@ func NewResourceRender(envoyNamespace, controllerNamespace, dnsDomain string, in
 }
 
 func (r *ResourceRender) Name() string {
+	if r.GatewayNamespaceMode {
+		return GetGatewayName(r.infra)
```
arkodg, line 76: imo, this is something that should be set in the gateway api layer which should directly set the `Name` field within `ProxyInfra`

internal/infrastructure/kubernetes/proxy/testdata/gateway-namespace-mode/deployment.yaml
zirain: it will be conflict.

internal/infrastructure/kubernetes/proxy/resource_provider.go
zirain: Right now, the infra name is actually the irKey.
Change that will be a huge mass, which should be do in a follow PR.

arkodg: only for this case though https://github.com/envoyproxy/gateway/blob/0815cdc695568e1b9fb66976df7daa87ed3bd1df/internal/gatewayapi/translator.go#L320

zirain: hope https://github.com/envoyproxy/gateway/pull/6135/commits/426d8d98e47191fd723fd43e9438a9d9045512b6 is what you want.

internal/infrastructure/kubernetes/proxy/testdata/gateway-namespace-mode/deployment.yaml
cnvergence: if not adding hash, maybe just keep the envoy prefix

zirain: @envoyproxy/gateway-maintainers wdyt?

zirain: what won't help, e.g. gateway name with gw, and there's a deployment named `envoy-gw`.

arkodg (approved):
LGTM thanks !
added one comment

internal/infrastructure/kubernetes/proxy/testdata/configmap/default.yaml
```diff
@@ -1,14 +1,15 @@
 apiVersion: v1
+data:
```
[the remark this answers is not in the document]
zirain: Before, it's hand made. Now it's part of `make go.testdata.complete`, which is better IMO.

internal/infrastructure/kubernetes/proxy/testdata/gateway-namespace-mode/deployment.yaml
zirain: @envoyproxy/gateway-maintainers PTAL before merging this.

arkodg: +1 with PRs approach

PR #6136, opened 2025-05-21
feat: implementing the ability to optionally deny redirects in security policies

sam-burrell:

Implements the accepted api change from merged pull request https://github.com/envoyproxy/gateway/pull/5455 

Implement this change with the underlying envoy [oauth2 configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/oauth2/v3/oauth.proto#envoy-v3-api-msg-extensions-filters-http-oauth2-v3-oauth2) deny_redirect_matcher
This is useful for ajax/machine paths that should not be directed to the authentication flow.

Release Notes: Yes


zhaohuabing:
Hi @sam-burrell We will need a release not for this because it's an API change for OIDC.

sam-burrell:
Makes sense, do I need add anything specific?

zhaohuabing:
> Makes sense, do I need add anything specific?

Just fix the lint and pass all the checks. :-)

api/v1alpha1/oidc_types.go
```diff
@@ -160,14 +159,12 @@ type OIDCProvider struct {
 }
 
 // OIDCDenyRedirect defines headers to match against the request to deny redirect to the OIDC Provider.
-// +notImplementedHide
 type OIDCDenyRedirect struct {
 	// Defines the headers to match against the request to deny redirect to the OIDC Provider.
 	Headers []OIDCDenyRedirectHeader `json:"headers"`
```
zhaohuabing, line 164: Nit: length must be longer than 0?

zhaohuabing, line 164: Nit: while it's technically valid, an empty header array doesn't seem meaningful here.
Also, consider setting a max limit for the headers - most other APIs define one.
```suggestion
        // +kubebuilder:validation:MinItems=1
	// +kubebuilder:validation:MaxItems=16
	Headers []OIDCDenyRedirectHeader `json:"headers"`
```

PR #6139, opened 2025-05-21
use SDS for the client cert for connections with Wasm HTTP server

zhaohuabing:
This PR uses SDS to fetch the envoy client certificate for envoy-wasm HTTP server TLS connection, and enables Wasm e2e tests for gatewaynamepace mode.


arkodg (dismissed):
LGTM thanks !

arkodg:
since this change also affects non gateway mode users, maybe its safer to do this as a minor release than patch release

zirain (approved):

cnvergence (approved):

cnvergence (approved):

PR #6137, opened 2025-05-21
feat(api): add zone field to BackendEndpoint

AlexKrudu:
**What type of PR is this?**
feat(api): add zone field to BackendEndpoint

**What this PR does / why we need it**:
Provides the opportunity for specifying endpoint's zone not only from the `EndpointSlice` but also from `Backend`

**Which issue(s) this PR fixes**:
Fixes https://github.com/envoyproxy/gateway/issues/6017

Release Notes: Yes/No

PR #6148, opened 2025-05-22
chore: fix CI

zirain:


internal/gatewayapi/testdata/backend-with-endpoint-zones.out.yaml
```diff
@@ -171,6 +179,10 @@ xdsIR:
             - host: 2.2.2.2
               port: 3001
               zone: zone2
+            metadata:
+              kind: Backend
+              name: backend-2
+              namespace: default
```
zhaohuabing, line 185: Do we need to change this file?

zirain: yeah, missed in https://github.com/envoyproxy/gateway/pull/6137.

zirain: https://github.com/envoyproxy/gateway/actions/runs/15175136959/job/42673862376

zhaohuabing (approved):
LGTM thanks!

arkodg (approved):

PR #6149, opened 2025-05-22
change default component log level to `misc:error`

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/5983


zhaohuabing (approved):
LGTM thanks!

cnvergence (approved):

internal/infrastructure/common/proxy_args.go
```diff
@@ -67,6 +67,9 @@ func BuildProxyArgs(
 
 	if componentsLogLevel := logging.GetEnvoyProxyComponentLevel(); componentsLogLevel != "" {
 		args = append(args, fmt.Sprintf("--component-log-level %s", componentsLogLevel))
+	} else {
+		// Default to error level for misc components if not set.
```
arkodg, line 71: Should this be moved to https://github.com/envoyproxy/gateway/blob/c09c42ca34a2eba9d7699e800c218c293f69d5fc/api/v1alpha1/envoyproxy_helpers.go#L132 ?

arkodg: sure that would be great if we can keep the source of truth of defaults inside `GetEnvoyProxyComponentLevel` rather than in the caller

PR #6150, opened 2025-05-22
fix: add unix domain socket path length validation

### LINKED ISSUE #6145: Limit UDS Path length to 108 with CEL
*Description*:
>Describe the issue.

https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars 

https://github.com/envoyproxy/gateway/blob/eb7266c042f4573d7e67c9ea02a19e0deff6e17f/api/v1alpha1/backend_types.go#L112

[optional *Relevant Links*:]

mukeshmahato17:
This PR adds validation to limit Unix domain socket path length to 108 characters, which is the standard limit across Unix systems. This limitation exists for historical reasons related to the kernel's mbuf (memory buffer) data structure in early BSD implementations.

Fixes #6145 

Release Notes: Yes

### DIFF
--- a/api/v1alpha1/backend_types.go
+++ b/api/v1alpha1/backend_types.go
@@ -115,6 +115,9 @@ type FQDNEndpoint struct {
 // https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-pipe
 type UnixSocket struct {
 	// Path defines the unix domain socket path of the backend endpoint.
+	// The path length must not exceed 108 characters.
+	//
+	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=108
 	Path string `json:"path"`
 }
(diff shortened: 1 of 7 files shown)


api/v1alpha1/backend_types.go
```diff
@@ -110,6 +110,8 @@ type FQDNEndpoint struct {
 // https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#config-core-v3-pipe
 type UnixSocket struct {
 	// Path defines the unix domain socket path of the backend endpoint.
+	// +kubebuilder:validation:MaxLength=108
+	// +kubebuilder:validation:XValidation:rule="self.size() <= 108",message="Unix domain socket path length must not exceed 108 characters"
```
rudrakhp, line 114:
```suggestion
	
```
Do we need this validation after `+kubebuilder:validation:MaxLength=108`?

zhaohuabing:
@mukeshmahato17 run `make gen-check` locally and submit the changes will fix the gen-check.

test/cel-validation/backend_test.go
```diff
@@ -276,28 +275,20 @@ func TestBackend(t *testing.T) {
 			wantErrors: []string{"DynamicResolver type cannot have endpoints and appProtocols specified"},
 		},
 		{
-			desc: "tls settings on non-dynamic resolver",
```
rudrakhp, line 279: can we have a new test instead of replacing an existing one?

mukeshmahato17: actually this was merge conflict. i resolved the conflict only

PR #6152, opened 2025-05-22
fix: zone-aware-routing e2e falkiness

### LINKED ISSUE #6119: Flaky: TestE2E/ZoneAwareRouting
--- FAIL: TestE2E/ZoneAwareRouting (2.38s)
        --- FAIL: TestE2E/ZoneAwareRouting/only_local_zone_should_get_requests (2.09s)

https://github.com/envoyproxy/gateway/actions/runs/15121989153/job/42507080550

jukie:
**What type of PR is this?**

Fixes e2e flakiness after a refactoring

Fixes #6119

Release Notes: No

### DIFF
--- a/test/e2e/tests/weighted_backend.go
+++ b/test/e2e/tests/weighted_backend.go
@@ -66,7 +66,7 @@ func runWeightedBackendTest(t *testing.T, suite *suite.ConformanceTestSuite, rou
 		suite.ControllerName,
 		kubernetes.NewGatewayRef(SameNamespaceGateway), weightEqualRoute)
 
-	// make sure all backend resources are ready
+	// Make sure all test resources are ready
 	kubernetes.NamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, []string{ConformanceInfraNamespace})
 
 	expectedResponse := http.ExpectedResponse{
--- a/test/e2e/tests/weighted_backend.go
+++ b/test/e2e/tests/weighted_backend.go
@@ -66,6 +66,9 @@ func runWeightedBackendTest(t *testing.T, suite *suite.ConformanceTestSuite, rou
 		suite.ControllerName,
 		kubernetes.NewGatewayRef(SameNamespaceGateway), weightEqualRoute)
 
+	// make sure all backend resources are ready
+	kubernetes.NamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, []string{ConformanceInfraNamespace})
+
 	expectedResponse := http.ExpectedResponse{
 		Request: http.Request{
 			Path: path,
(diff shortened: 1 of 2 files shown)


test/e2e/tests/weighted_backend.go
```diff
@@ -66,6 +66,9 @@ func runWeightedBackendTest(t *testing.T, suite *suite.ConformanceTestSuite, rou
 		suite.ControllerName,
 		kubernetes.NewGatewayRef(SameNamespaceGateway), weightEqualRoute)
 
+	// Make sure all test resources are ready
+	kubernetes.NamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, []string{ConformanceInfraNamespace})
+
```
jukie, line 71: @zhaohuabing @zirain I narrowed this down to being a race condition if `NamespacesMustBeReady()` executes prior to `GatewayAndHTTPRoutesMustBeAccepted()` and that's what was changed with the refactor.

The manifest apply action happens before test execution but the client cache seems to still be outdated and as a result the list actions in `NamespacesMustBeReady()` are empty and the intended wait action never happens.

I was able to reproduce the failure locally ~50% of the time but after moving `NamespacesMustBeReady()` into `runWeightedBackendTest()` _after_ `GatewayAndHTTPRoutesMustBeAccepted()` execution it consistently suceeds.

zirain: I'm good with this chagne, it's harmless.
BTY, can you explain more about this? 
`NamespacesMustBeReady()` make sure that all the pods in the namespace should be ready, it should be more strict than `GatewayAndHTTPRoutesMustBeAccepted()`, why move it after solve this? 
Need more time to refresh the pod list? if so, prefer to bring back 
```
WaitForPods(t, suite.Client, ns, map[string]string{"app": "zone-aware-backend"}, corev1.PodRunning, podReady)
```

jukie: I don't have any real proof about the issue being client cache but that's a theory.
I suspect that by first calling `GatewayAndHTTPRoutesMustBeAccepted()` that implicitly ensures the proxy pods are also up and running. In the case of zone aware routing tests, this is needed due to the affinity rules.

It seems useful that for all tests we'd want to wait for all test resources to be completely ready before proceeding. Limiting to just Pods allows for other edge cases in the future so a comprehensive function like `NamespacesMustBeReady()` (tbf it currently just checks gateways and pods but could be expanded further) would be my preference but if you'd like to instead bring back `WaitForPods()` specifically for zone-aware-routing I can make that update.

zirain: You're using `same-namespace` gateway, so it's always there.

Let us test with, we can bring `WaitForPods()` back if needed in the future.

zirain (approved):
Thanks for debugging!

arkodg (approved):

Issue #6125, opened 2025-05-20
Policy with targetSelector targeting multiple gateways does not show both ancestors


BadLiveware:
*Description*:
When targeting resources across multiple gateways with targetSelectors, the status of the policy will only contain one gateway, and it will continously rotate between the targeted gateways

*Repro steps*:
Create two gateways, attach two httproutes to the gateways(one per), and use targetSelector to target one policy(We see it with BackendTrafficPolicy and SecurityPolicy) to both httproutes

```
GW-1 <- HTTPRoute-1
              \ 
               BackendTrafficPolicy
              /
GW-2 <- HTTPRoute-2
```
*Environment*:
Seeing it both on 1.3.0 and 1.4.0

We did previously see a similar issue attaching a HTTPRoute to multiple gateways which was solved in https://github.com/envoyproxy/gateway/issues/4264

zhaohuabing:
@BadLiveware I just tested with the below setup and it looks good. Can you share the configurations that you had problems with?

```yaml
backendTrafficPolicies:
- apiVersion: gateway.envoyproxy.io/v1alpha1
  kind: BackendTrafficPolicy
  metadata:
    creationTimestamp: null
    name: foo
    namespace: default
  spec:
    targetSelectors:
    - kind: HTTPRoute
      matchLabels:
        foo: bar
    timeout:
      http:
        connectionIdleTimeout: 21mib
      tcp:
        connectTimeout: 20s
  status:
    ancestors:
    - ancestorRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: gateway-1
        namespace: default
      conditions:
      - lastTransitionTime: null
        message: 'Timeout: invalid ConnectionIdleTimeout value 21mib.'
        reason: Invalid
        status: "False"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
    - ancestorRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: gateway-2
        namespace: default
      conditions:
      - lastTransitionTime: null
        message: 'Timeout: invalid ConnectionIdleTimeout value 21mib.'
        reason: Invalid
        status: "False"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
gateways:
- apiVersion: gateway.networking.k8s.io/v1
  kind: Gateway
  metadata:
    creationTimestamp: null
    name: gateway-1
    namespace: default
  spec:
[...]

Issue #774, opened 2022-12-03
Stale Route Parent condition present in Route object


chauhanshubham:
*Description*:
When the parent reference in a Route object (on both HTTPRoute and TLSRoute), is updated
to point to a non-existing gateway, the Route parent condition is not updated properly

*Repro steps*:
1. apply `examples/kubernetes/quickstart.yaml` - ensure that Gateway has proper Status conditions, with 1 listener condition
showing `attachedRoutes: 1`. Ensure that the HTTPRoute also has proper Parent Condition present. The httproute parent condition might look something like this
```yaml
status:
    parents:
    - conditions:
      - lastTransitionTime: "2022-12-03T07:23:33Z"
        message: Route is accepted
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
      parentRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
```
2. Now edit the HTTPRoute to point to a non-existing Gateway i.e.
```yaml
parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: eg1
```

*Observation*
While the Gateway status properly specifies the `attachedRoutes: 0`, the HTTPRoute on the other hand shows - what appears to be a stale parent condition. i.e.
```yaml
status:
    parents:
    - conditions:
      - lastTransitionTime: "2022-12-03T07:23:33Z"
        message: Route is accepted
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      controllerName: gateway.envoyproxy.io/gatewayclass-controller
      parentRef:
        group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
```
Which is the same as seen earlier.
Note that the HTTPRoute now has no mention of Gateway `eg` as we changed the reference to `eg1`, but still the Condition shows a condition that was filled earlier, while it was connected to Gateway `eg`.

arkodg:
thanks for raising this issue -
* today, we only process `HTTPRoute`s associated with an accepted `Gateway` . This is most likely the reason we are not reprocessing this `HTTPRoute` since it is no longer attached to an accepted `Gateway` .
@youngnick from a spec perspective, do we need to re update the status on such a route, or is that something that is good to have ?

arkodg:
[...]

PR #6153, opened 2025-05-22
preserve policy status cross gc boundary

zhaohuabing:
fixes: #6125


internal/gatewayapi/runner/runner.go
```diff
@@ -310,6 +317,70 @@ func (r *Runner) subscribeAndTranslate(sub <-chan watchable.Snapshot[string, *re
...
+			policy.Status = *policyAncestorsFromOtherGatewayClasses(policyStatus, policy.Namespace, resources)
+		}
+	}
+}
+
+func policyAncestorsFromOtherGatewayClasses(policyStatus *gwapiv1a2.PolicyStatus, policyNamespace string, resources *resource.Resources) *gwapiv1a2.PolicyStatus {
+	AncestorsFromOtherGatewayClasses := make([]gwapiv1a2.PolicyAncestorStatus, 0)
+	for _, ancestor := range policyStatus.Ancestors {
+		if ancestor.AncestorRef.Group == nil ||
+			*ancestor.AncestorRef.Group == "gateway.networking.k8s.io" {
```
zhaohuabing, line 366: This assumes all xPolicy types use `Gateway` as the status ancestor, which is recommended by the Gateway API spec.
We may need to clean up the xPolicy status in a follow-up PR.

zhaohuabing:
This is a pretty major change on the policy status, so let's hold off for a bit before cherry-picking it into 1.4 and 1.3, just to be safe.

arkodg:
@zhaohuabing dont xRoutes status have the same issue ? I believe we solved those in the provider status updater layer ?

zhaohuabing:
> @zhaohuabing dont xRoutes status have the same issue ? I believe we solved those in the provider status updater layer ?

@arkodg Yes, prefer this solution, more in this comment: https://github.com/envoyproxy/gateway/pull/6153#issuecomment-2901627044

arkodg:
hey @zhaohuabing thanks for sharing the past history :) , this looks like two issues
1) around cleaning up old status condition entries. can we solve this with a separate util function similar to `mergeRouteParentStatus` that removes any `condition`  (linked to EG controller) whose `ObservedGeneration` doesnt match the current `ObservedGeneration` of the client Object ?
2 ) Regarding this specific issue of multiple Gateways linked to different GatewayClasses, this feels like we need to change the ancestor to `GatewayClass` instead of `Gateway` and merge properly in the status update layer ?

zhaohuabing:
> around cleaning up old status condition entries. can we solve this with a separate util function similar to mergeRouteParentStatus that removes any condition (linked to EG controller) whose ObservedGeneration doesnt match the current ObservedGeneration of the client Object ?

Thought it again - this can't be solved by ObservedGeneration.
ObservedGeneration comes form the Generation of the xPolicy resource, however, the xPolicy status can change without a bump of its Generation. 
For example:
a xPolicy targets a Gateway gw. After gw is deleted, it's still in the status of xPolicy, and this status can't be cleaned because the ObservedGeneration of status is the same as the policy.

```yaml
kind: BackendTrafficPolicy
metadata:
  generation: 11
  name: test
spec:
  connection:
    bufferLimit: 100M
  targetRefs:
    kind: Gateway
    name: gw
status:
  ancestors:
  - ancestorRef:
      kind: Gateway
      name: gw
      namespace: default
    conditions:
    - lastTransitionTime: "2025-06-25T00:55:28Z"
      message: Policy has been accepted.
      observedGeneration: 11
      reason: Accepted
      status: "True"
      type: Accepted
```

arkodg:
@zhaohuabing same is true for xRoute and linked resources ( backendRef)

zhaohuabing:
> with Gateway as the ancestor, we ca

You mean this issue https://github.com/envoyproxy/gateway/issues/774 ? I'm planning to work on that after fixing the xPolicy status.

PR #6154, opened 2025-05-22
api: add requestdistribution config

jukie:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No


jukie:
Hey @arkodg @guydc could I have a review on this when you get the chance?

jukie:
Could I get a review here please?

arkodg:
hey will review this API PR this week, sorry for the delay

api/v1alpha1/loadbalancer_types.go
```diff
@@ -135,3 +142,46 @@ type SlowStart struct {
...
+	// Enabled causes Envoy to route all requests to the local zone if there are at least "minZoneSize" healthy hosts
+	// available.
+	//
+	// +optional
+	// +notImplementedHide
+	Enabled *bool `json:"enabled"`
+
+	// MinZoneSize is the minimum number of upstream hosts in the local zone required to honor the forceLocalZone
+	// override. Defaults to 1 if not specified.
+	//
```
rudrakhp:
```suggestion
    // +kubebuilder:default=1
	//
```
Please add CEL validations wherever required.

api/v1alpha1/loadbalancer_types.go
rudrakhp, line 1: Can we simplify to have a single type `ZoneAwareRouting`? Eg:
```go
// this needs to be configured for enabling zone aware routing, not configured would mean zone aware routing not enabled
type ZoneAwareRouting struct {
  // this is the minimum number of hosts in cluster across zones to enable zone aware routing
  // default 6?
  MinClusterSize *int
  // this is the minimum number of hosts in cluster in this envoy's local zone to force all traffic to this zone
  // If not configured force local zone is not enabled
  MinForceLocalZoneSize *int
}
```

api/v1alpha1/loadbalancer_types.go
jukie: The default is 1 for the envoy API and isn't being injected by EnvoyGateway so it's an optional field.

api/v1alpha1/loadbalancer_types.go
jukie: Not opposed to it but I think my current implementation provides some future proofing if more features are added upstream to Envoy ForceLocalZone.

rudrakhp: I was wondering if we can simplify the control plane API instead of exposing the [data plane API](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_runtime#zone-aware-load-balancing) as is. I have not give much thought on the exact API (above one is just an example) but currently we have 2 layers of nesting that might be unnecessary. WDYT think about collapsing these layers while keeping required options for future proofing the API?

api/v1alpha1/loadbalancer_types.go
rudrakhp: Ah got it!

api/v1alpha1/loadbalancer_types.go
jukie: Yeah I'll tweak this a bit. The current `preferLocalZone` --> `forceLocalZone` also isn't great.

api/v1alpha1/loadbalancer_types.go
```diff
@@ -34,6 +35,12 @@ type LoadBalancer struct {
 	//
 	// +optional
 	SlowStart *SlowStart `json:"slowStart,omitempty"`
+
+	// RequestDistribution defines the configuration related to the distribution of requests between localities.
+	//
+	// +optional
+	// +notImplementedHide
+	RequestDistribution *RequestDistribution `json:"requestDistribution,omitempty"`
```
arkodg: sharing my preference for naming, please lmk what you think

```
zoneAware:
  preferLocal:
    force:
      minEndpointCount: 1
    minEndpointCount: 3
```
```

Comment (rudrakhp) [reply api/v1alpha1/loadbalancer_types.go]: This looks like a good middle ground without the enable booleans. A minor modification to @arkodg's suggestion to make it clear min endpoints in zone vs all endpoints:
```
zoneAware:
  preferLocal:
    force:
      minEndpointsInZoneCount: 1
    minEndpointCount: 3
```

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go:43]: I agree. Will update with that!

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: I don't think we need an extra step from zoneAware --> preferLocal but I agree overall.
RequestDistribution or similar should still be the top-level key because I plan on adding weighted zones after this one.

How about one of these? Can add the "Count" suffix too if that's preferable but I think "minEndpoints" sufficiently indicates that.

A)
```
requestDistribution:
  zoneAware:
    forceLocal:
      minEndpointsInZone: 1
    minEndpoints: 3
```

B)
```
requestDistribution:
  preferLocal:
    force:
      minEndpointsInZone: 1
    minEndpoints: 3
```

File: api/v1alpha1/loadbalancer_types.go
Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go:1]: Continuing the discussion in other thread

File: api/v1alpha1/loadbalancer_types.go (hunk 2 above)
Comment (rudrakhp) [reply api/v1alpha1/loadbalancer_types.go]: I prefer A, the data plane `upstream.zone_routing.min_cluster_size` sounds closer to `zoneAware.minEndpoints` than `preferLocal.minEndpoints`.

Comment (rudrakhp) [reply api/v1alpha1/loadbalancer_types.go]: Do we need the top level `requestDistribution` though?

Comment (rudrakhp) [reply api/v1alpha1/loadbalancer_types.go]: >I plan on adding weighted zones after this one

If this is planned, having the complete API in this PR might give us a better picture to review maybe. We can implement them in phases later.

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: That's what I started with but @arkodg  requested doing it like this instead.

Comment (arkodg) [reply api/v1alpha1/loadbalancer_types.go]: @jukie here is what I was solving/accounting for
1. 
```
loadBalancer:
  requestDistribution:
    zoneAware:
```

feels repetitive, which is why I also prefer not having the top level

2. I think there's still room for weighted zones even if we skip top level
```
loadBalancer:
  zoneAware:
    distribution:
```

with a CEL validation saying that `distribution` and `preferLocal` cannot be set at the same time

3.. adding some way to say that the feature kicks in only if the threshold/count is reached explicitly is helpful which is why i suggested the `count` suffix, but its not a strong opinion

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: Having a top-level field also helps communicate the intent and keeps the `loadBalancer` struct a bit cleaner in my opinion vs flattening. When reading docs for `requestDistribution` I think it'd be easier for a user to reason about configuration options vs having separate fields under `loadBalancer`.

To reiterate `requestDistribution` maps to Envoy LocalityLbConfig where either `zoneAwareLbConfig` or `localityWeightedLbConfig` can be used, each requiring unique configuration. The weighted zone configuration wouldn't go under `zoneAware` unless you're proposing those be flattened and if we do that, we'll most likely want something similar to the `LoadBalancerType` field.

If you'd still prefer to flatten it I can @arkodg @rudrakhp

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: Updated with removal of `requestDistribution` and now looks like:

```
loadBalancer:
  zoneAware:
    forceLocal:
      minEndpointsInZoneCount: 1
    minEndpointsCount: 6
```

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: Thanks for the discussion today @arkodg. Does this sound reasonable?

Ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#extensions-load-balancing-policies-common-v3-localitylbconfig
```
loadBalancer:
  zoneAware: # Equivalent to LocalityLbConfig
    preferLocal: # Equivalent to ZoneAwareLbConfig
      force:
        minEndpointsInZoneCount: 1
      minEndpointsCount: 6
```

Not included here but would also have something like `loadbalancer.zoneAware.weightedZones` for  LocalityWeightedLbConfig and the weights would get applied to the localities during cluster/endpoint rendering.

Comment (jukie) [reply api/v1alpha1/loadbalancer_types.go]: Updated with ^

Comment (arkodg) [reply api/v1alpha1/loadbalancer_types.go]: sgtm thanks

File: api/v1alpha1/loadbalancer_types.go [hunk 3]
```diff
@@ -135,3 +142,39 @@ type SlowStart struct {
...
+	//
+	// +optional
+	// +notImplementedHide
+	Force *ForceLocalZone `json:"force,omitempty"`
+
+	// MinEndpointsCount is the minimum number of total upstream endpoints across all zones required to enable zone-aware routing.
+	//
+	// +optional
+	// +notImplementedHide
+	MinEndpointsCount *uint64 `json:"minEndpointsCount,omitempty"`
```
arkodg: thoughts on MinEndpointsThreshold ?

jukie: Sounds good to me

arkodg: wdyt @envoyproxy/gateway-maintainers

rudrakhp: For consistency, IMO we should align the naming of `minEndpointsCount` and `minEndpointsInZoneCount`. If we're switching to Threshold, lets apply to both? Fine with both 👍

jukie: Yeah that's what I had assumed and changed both to use "Threshold"

api/v1alpha1/loadbalancer_types.go
```diff
@@ -135,3 +142,39 @@ type SlowStart struct {
...
+	//
+	// +optional
+	// +notImplementedHide
+	Force *ForceLocalZone `json:"force,omitempty"`
+
+	// MinEndpointsThreshold is the minimum number of total upstream endpoints across all zones required to enable zone-aware routing.
+	//
+	// +optional
+	// +notImplementedHide
+	MinEndpointsThreshold *uint64 `json:"minEndpointsCount,omitempty"`
```
arkodg, line 168: is the uint type tied to envoy types ?

jukie: Yes that maps to `min_cluster_size` and uses uint64 but can adjust to something else if preferred.

Ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#extensions-load-balancing-policies-common-v3-localitylbconfig-zoneawarelbconfig

jukie: And uint32 for minEndpointsInZoneThreshold/`min_size` - https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#envoy-v3-api-msg-extensions-load-balancing-policies-common-v3-localitylbconfig-zoneawarelbconfig-forcelocalzone

jukie:
@rudrakhp could you take another look here?

rudrakhp (approved):
LGTM!

rudrakhp:
@jukie I missed this the first time, but looks like the json name for the attribute has not been updated here:
https://github.com/envoyproxy/gateway/blob/7b7052c80ed1e729556d5955dce665769a659573/api/v1alpha1/loadbalancer_types.go#L168
Can you please raise a quick fix for this, thanks!

jukie:
Good catch, will do!

PR #6160, opened 2025-05-22
add unix domain socket path length validation

mukeshmahato17:
**What type of PR is this?**
This PR adds validation to limit Unix domain socket path length to 108 characters, which is the standard limit across Unix systems. This limitation exists for historical reasons related to the kernel's mbuf (memory buffer) data structure in early BSD implementations.

Fixes #6150 

Release Notes: Yes


arkodg (approved):
LGTM thanks !

mukeshmahato17:
hey @arkodg the TestEGUpgrade test failed in CI during the upgrade sequence with resource cleanup operations, while the Envoy shutdown portion passes successfully (might be).  Could someone please re-run the test?

rudrakhp (approved):

PR #6169, opened 2025-05-22
fix: return early from buildwasms

### LINKED ISSUE #6159: WASM cache initialization failure causes all routes with EEP to fail
*Description*:

see here https://github.com/envoyproxy/gateway/issues/6124#issuecomment-2901233079 :

> The "Translator" calls buildWasms twice, once in translateEnvoyExtensionPolicyForRoute and once in translateEnvoyExtensionPolicyForGateway both treat a nil as an error

Seems like buildWasms will default all routes with an EEP, regardless of if the EEP even has wasms: 

It's always called when processing EEPs: 
https://github.com/envoyproxy/gateway/blob/c09c42ca34a2eba9d7699e800c218c293f69d5fc/internal/gatewayapi/envoyextensionpolicy.go#L302

It returns an error before checking if wasm cache is really needed:
https://github.com/envoyproxy/gateway/blob/c09c42ca34a2eba9d7699e800c218c293f69d5fc/internal/gatewayapi/envoyextensionpolicy.go#L618

In other words, if wasm cache fails to start - all routes with EEPs are zeroed-out. 

cc @zhaohuabing

*Repro steps*:

*Environment*:

*Logs*:

guydc:
**What type of PR is this?**

Fixes #6159 

Release Notes: Yes

### DIFF
--- a/internal/gatewayapi/envoyextensionpolicy.go
+++ b/internal/gatewayapi/envoyextensionpolicy.go
@@ -614,11 +614,15 @@ func (t *Translator) buildWasms(
 	policy *egv1a1.EnvoyExtensionPolicy,
 	resources *resource.Resources,
 ) ([]ir.Wasm, error) {
+	var wasmIRList []ir.Wasm
+
+	if len(policy.Spec.Wasm) == 0 {
+		return wasmIRList, nil
+	}
+
 	if t.WasmCache == nil {
 		return nil, fmt.Errorf("wasm cache is not initialized")
 	}
-
-	var wasmIRList []ir.Wasm
 
 	if policy == nil {
 		return nil, nil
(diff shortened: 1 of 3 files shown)


zhaohuabing (approved):
LGTM thanks!

internal/gatewayapi/envoyextensionpolicy.go
```diff
@@ -614,12 +614,16 @@ func (t *Translator) buildWasms(
 	policy *egv1a1.EnvoyExtensionPolicy,
 	resources *resource.Resources,
 ) ([]ir.Wasm, error) {
+	var wasmIRList []ir.Wasm
+
+	if len(policy.Spec.Wasm) == 0 {
+		return wasmIRList, nil
```
arkodg, line 620: does the caller expect `nil` or empty list ?

guydc: nil slice wasmIRList was previosuly returned if `range policy.Spec.Wasm` was empty. So, should be safe.

arkodg (approved):

PR #6176, opened 2025-05-23
feat: add option to disable lua validation

### LINKED ISSUE #6172: Option to disable Lua validation at proxy level
*Description*:

Provide an option in the EnvoyProxy API to disable Lua validation, which must be enabled by default.
https://github.com/envoyproxy/gateway/blob/0815cdc695568e1b9fb66976df7daa87ed3bd1df/api/v1alpha1/envoyproxy_types.go#L34

*Relevant Links*:

Slack [thread](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1747915582046419)

rudrakhp:
**What type of PR is this?**

feat: add option to disable lua validation

We need this for user to be able to optionally disable Lua validation. Related slack [thread](https://envoyproxy.slack.com/archives/C03E6NHLESV/p1747915582046419).

Fixes #6172 

Release Notes: Yes/No


api/v1alpha1/envoyproxy_types.go
```diff
@@ -157,6 +157,10 @@ type EnvoyProxySpec struct {
 	//
 	// +optional
 	PreserveRouteOrder *bool `json:"preserveRouteOrder,omitempty"`
+
+	// ExtensionPolicySettings defines settings for EnvoyExtensionPolicies to be applied on the EnvoyProxy fleet.
+	// +optional
+	ExtensionPolicySettings *ExtensionPolicySettings `json:"extensionPolicySettings,omitempty"`
```
aukevanleeuwen, line 163: Out of curiousity: why would you it on the `EnvoyGateway` kind?

I would have expected it here to be honest:

  ```yaml
  kind: EnvoyExtensionPolicy
  metadata:
    creationTimestamp: null
    name: policy-for-http-route
    namespace: default
  spec:
    lua:
    - inline: <lua code>
      validate: false
      type: Inline
    targetRef:
      kind: HTTPRoute
      name: httproute-1
      ```
      
 I didn't think too hard about the naming, could be `validation: "Disabled"` as well. But either way it would allow you to specify different settings for different items in the array for example and it's more local to where you enable the extension in the first place?

rudrakhp: Thanks for the comment! The only valid reason why anyone would skip Lua validation (that comes to my mind) is they are using an API that's not part of the Envoy proxy's standard StreamHandle API ,in the standard case validation is expected to work and we do not want validation to be disabled in any scenario. And non-standard API is possible only if they are using a custom proxy image, in which case it makes sense to disable validation across the proxy fleet since validation is not applicable to this image.
See the related issue and slack thread linked for more.

rudrakhp: Although I can think of a scenario where you might want to have validations in place for some of your policies while skipping others, I am trying to think of a better way to discourage disabling validations.

aukevanleeuwen: it's not _only_ a custom image of course. If you're loading files with `dofile(...)` from a config map mount (as I do) you get the same thing I suppose. This file cannot be found yet, because it's not mounted yet.

rudrakhp: Yeah in both cases (custom image vs mounting a file to envoy), it's a fleet wide change.

arkodg: one big reason to add it in `EnvoyProxy` is the platform engineer persona needs to enable/authorize this for the app dev (EEP creator)

api/v1alpha1/envoyproxy_types.go
```diff
@@ -169,6 +173,14 @@ const (
 	EndpointRoutingType RoutingType = "Endpoint"
 )
 
+// ExtensionPolicySettings defines settings for EnvoyExtensionPolicies to be applied on the EnvoyProxy fleet.
+type ExtensionPolicySettings struct {
+	// DisableLuaValidation disables the Lua script validation for Lua EnvoyExtensionPolicies
+	// Default: false
+	// +optional
+	DisableLuaValidation *bool `json:"disableLuaValidation,omitempty"`
```
arkodg, line 181: imo, id put this directly at top level, can't think of which other fields go here

rudrakhp:
retest

PR #6183, opened 2025-05-25
Fix translate xds udp

### LINKED ISSUE #6156: Gateway not creating UDP Listeners (with DaemonSet?)
*Description*

A listener is not being created for a UDP port on the Envoy Proxy, whereas TCP/HTTP listeners work as expected. This may be related to the use of a `DaemonSet` (the configuration has not yet been tested with a `Deployment`). Notably, the Envoy Proxy `Service` is updated with the correct UDP port, which suggests that Envoy Gateway is processing the configuration and recognizing the port.

There are no errors in the logs from either Envoy Gateway or Envoy Proxy. Based on this and the fact that the UDP listener is absent from the config dump and there are no errors logged, it appears that Envoy Gateway is not even attempting to configure a listener for this UDP port.

*Reproduction Steps*

Create a `Gateway` and `UDPRoute` as shown below:

```yaml
kind: Gateway
metadata:
  name: gateway
spec:
  listeners:
    - name: 21110-udp
      port: 21110
      protocol: UDP
      allowedRoutes:
        kinds:
          - kind: UDPRoute
        namespaces:
          from: Same
```

*Environment*

Envoy Gateway Version (Latest): `docker.io/envoyproxy/gateway-dev@sha256:e8beabddf486ca0878bb8a7e935c716e046e6e7560ed1c8cf4dac224a3ae557f`
Deployment Type: DaemonSet (not tested with Deployment)

*Observed Behavior*

- The `Service` for Envoy Proxy includes the UDP port as expected.
- The Envoy admin interface (`/config_dump`) does **not** show a UDP listener under `type.googleapis.com/envoy.admin.v3.ListenersConfigDump`.
- There are **no errors** in the logs of either Envoy Gateway or Envoy Proxy that reference the UDP listener or port.

*Expected Behavior*

A UDP listener should be created and appear in the listener config dump. If there is a reason the listener is not being created (e.g. a misconfiguration or unsupported scenario), an error or warning should be emitted by Envoy Gateway.

kkk777-7:
Fixed issue which UDP listeners were not created in the Envoy proxy config when Gateway was created.

Fixes #6156 

Release Notes: Yes

**Check Behavior:**
Verified that generates the UDP listener in the Envoy config when creating the following Gateway in local KIND environment.

```yaml
kind: Gateway
metadata:
  name: udp-gateway
spec:
  gatewayClassName: eg
  listeners:
  - allowedRoutes:
      kinds:
        kind: UDPRoute
      namespaces:
        from: Same
    name: udp
    port: 21110
    protocol: UDP
```

The following shows a partial output of `config_dump`.
[udp-listener-config.json](https://github.com/user-attachments/files/20429957/udp-listener-config.json)


internal/xds/translator/translator.go
```diff
@@ -1124,3 +1133,10 @@ func buildXdsUpstreamTLSSocketWthCert(tlsConfig *ir.TLSUpstreamConfig) (*corev3.
 		},
 	}, nil
 }
+
+func getClusterNameForUDPListener(listener *ir.UDPListener) string {
```
kkk777-7, line 1137: There might be a more appropriate name for the function.
Also, there might be a more appropriate place to implement the method.

internal/xds/translator/translator.go
```diff
@@ -770,29 +770,38 @@ func processUDPListenerXdsTranslation(
...
+				ClusterDiscoveryType: &clusterv3.Cluster_Type{Type: clusterv3.Cluster_STATIC},
+			}
+			if findXdsCluster(tCtx, emptyClusterName) == nil {
+				if err := tCtx.AddXdsResource(resourcev3.ClusterType, emptyRouteCluster); err != nil {
+					errs = errors.Join(errs, err)
+				}
+			}
+		}
+
+		xdsListener, err := buildXdsUDPListener(getClusterNameForUDPListener(udpListener), udpListener, accesslog)
```
rudrakhp, line 795: I think we should create an empty UDP route and set it in `listener.Route` before building the UDP listener here. That way we don't need `getClusterNameForUDPListener` and can always do `udpListner.Route.Destination.Name`. Check this from TCP:
https://github.com/envoyproxy/gateway/blob/f497360b2c15d46c20c54deb8a0bac306107cadf/internal/xds/translator/translator.go#L745-L750

internal/gatewayapi/testdata/gateway-with-multiple-https-listeners-with-overlapping-hostnames-and-certs-merged-gateways.out.yaml
```diff
@@ -191,12 +191,21 @@ xdsIR:
       port: 10443
       routes:
       - destination:
+          metadata:
```
rudrakhp, line 194: Curious why we added metadata to these tests which don't seem to be relevant to UDP listener translation.

rudrakhp (approved):

Issue #6184, opened 2025-05-25
CEL tests failing doesn't fail coverage tests


rudrakhp:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: The expected value isn't returned, etc.

CEL tests failing is not failing the coverage tests in CI

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

https://github.com/envoyproxy/gateway/actions/runs/15192435467/job/42728985493
PR with breaking test: #6160 

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

CI

*Logs*:
>Include the access logs and the Envoy logs.

```
--- FAIL: TestBackend (2.04s)
    --- FAIL: TestBackend/Invalid_Unix_socket_path_length (0.00s)
        backend_test.go:346: Unexpected response while creating Backend; got err=
            Backend.gateway.envoyproxy.io "backend-1747933880420698097" is invalid: [spec.endpoints[0].unix.path: Too long: may not be longer than 108, <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
            ;missing strings within error=["spec.endpoints[0].unix.path: Too long: may not be more than 108 bytes"]
FAIL
FAIL	github.com/envoyproxy/gateway/test/cel-validation	12.502s
FAIL
Run CEL Validation on k8s 1.32.0
ok  	github.com/envoyproxy/gateway/test/cel-validation	12.876s
===========> Running go.test.coverage ... 
```

PR #6185, opened 2025-05-25
ci: fix go.test.cel

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/6184

### DIFF
--- a/api/v1alpha1/backend_types.go
+++ b/api/v1alpha1/backend_types.go
@@ -117,7 +117,7 @@ type UnixSocket struct {
 	// Path defines the unix domain socket path of the backend endpoint.
 	// The path length must not exceed 108 characters.
 	//
-	// +kubebuilder:validation:MaxLength=108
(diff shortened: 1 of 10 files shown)


api/v1alpha1/backend_types.go
```diff
@@ -117,7 +117,7 @@ type UnixSocket struct {
 	// Path defines the unix domain socket path of the backend endpoint.
 	// The path length must not exceed 108 characters.
 	//
-	// +kubebuilder:validation:MaxLength=108
+	// +kubebuilder:validation:XValidation:rule="size(self) <= 108",message="unix domain socket path must not exceed 108 characters"
```
zirain, line 120: Prev implement cause unexpected error:
```console
The Backend "backend-unix" is invalid:
* spec.endpoints[0].unix.path: Too long: may not be more than 108 bytes
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

Issue #6181, opened 2025-05-23
Support host and priority predicates in Retry config


cmaddalozzo:
*Description*:

Currently the retry config[1] exposes a subset of the available options available on the underlying Envoy retry policy[2]. We'd like to expand this to include the host and priority retry predicates[3]. These enable users to configure host selection and priority selection on retries. For reference, Istio allows configuring these policies via the `retry_ignore_previous_hosts` and `retry_remote_localities` options[4], which configure the underlying predicates. I think this approach works well as the naming of the underlying Envoy config is a bit esoteric.

**Note:** The Istio config does not provide a way to configure the `update_frequency` on the previous priorities predicate but uses a default value that likely works for most use cases (including ours).  

*Relevant Links*

[1] https://gateway.envoyproxy.io/docs/api/extension_types/#retry
[2] https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy
[3] https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#retry-plugin-configuration
[4] https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.proto#L1315

cmaddalozzo:
@Inode1 this works for me. Although `update_frequency` should be set to `1` on the underlying config to ensure that we try a new priority for every request. Setting it to `2` would mean we only exclude priority levels after 2 requests.

Inode1:
> Although update_frequency should be set to 1 on the underlying config to ensure that we try a new priority for every request

Good point. But it seems like we want more customization here, and a simple boolean might not be enough. Maybe we should use a value and add an additional constraint that it must be less than or equal to attempts.
```
spec:
  retry:
    retryRemoteLocalities: 1 # default: disabled
```

@arkodg Hello, what do you thing about it?

arkodg:
@cmaddalozzo @Inode1 looking at the underlying config https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/retry/priority/previous_priorities/v3/previous_priorities_config.proto#extension-envoy-retry-priorities-previous-priorities

Can this (shifting traffic from P0 to P1 if P0 is unhealthy for the same request using retry) be achieved by using passive health check / outlier detection + retry instead ? https://gateway.envoyproxy.io/docs/tasks/traffic/failover/

arkodg:
[...]

PR #6204, opened 2025-05-28
feat(translator): support setting previous priorities retry predicate

Inode1:
**What type of PR is this?**

Makes it possible to configure the previous priorities retry for Route.

Fixes https://github.com/envoyproxy/gateway/issues/6181

Release Notes: Yes/No


api/v1alpha1/retry_types.go
```diff
@@ -18,6 +18,12 @@ type Retry struct {
 	// +kubebuilder:default=2
 	NumRetries *int32 `json:"numRetries,omitempty"`
 
+	// NumRetriesToAcrossPriorities defines the number of requests (initial attempt + retries)
+	// that should be sent to the same priority before switching to a different one.
+	//
+	// +optional
+	NumRetriesToAcrossPriorities *int32 `json:"numRetriesToAcrossPriorities,omitempty"`
```
cmaddalozzo: Can we change this to `NumAttemptsPerPriority`? We also need to document that omitting this (i.e. value is `0`) means there is no limit to the number of attempts per priority.

arkodg: 1) piggy backing off @cmaddalozzo's suggestion, if there are more setting related to `perPriority`
this could be represented as
```
perPriority:
  numAttempts: 2

```

else `numAttemptsPerPriority` makes sense

Inode1: > numAttemptsPerPriority makes sense

it sounds reasonable. fixed

api/v1alpha1/retry_types.go
```diff
@@ -18,6 +18,13 @@ type Retry struct {
 	// +kubebuilder:default=2
 	NumRetries *int32 `json:"numRetries,omitempty"`
 
+	// NumAttemptsPerPriority defines the number of requests (initial attempt + retries)
+	// that should be sent to the same priority before switching to a different one.
+	// If not specified or set to 0, all requests are sent to the highest priority.
```
arkodg:
```suggestion
	// If not specified or set to 0, all requests are sent to the highest priority that is healthy.
```

arkodg:
thanks @Inode1 the PR looks good, added a minor comment 
can you also add a test in https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata ?

Inode1:
> thanks @Inode1 the PR looks good, added a minor comment can you also add a test in https://github.com/envoyproxy/gateway/tree/main/internal/gatewayapi/testdata ?

Done

arkodg (approved):
LGTM thanks

zirain (approved):

PR #6206, opened 2025-05-28
fix: ratelimit cost only can use operation

Xunzhuo:
**What type of PR is this?**

fix: ratelimit cost only can use operation

Release Notes: Yes/No


arkodg:
is https://github.com/envoyproxy/gateway/blob/845b36b38f98b6c9cfd388e309ed2fd377574e1b/api/v1alpha1/ratelimit_types.go#L139 not working as expected ?

PR #6210, opened 2025-05-28
decouple Gateway API version with go mod

zirain:
this allow EG adopt changes for conformace test before a offical release.


zhaohuabing (approved):
Do you also want to get the latest commit and say if this works?

Issue #3415, opened 2024-05-17
Add a hook for failed test suite


shawnh2:
*Description*:

The conformance test and e2e test is unstable recently, like https://github.com/envoyproxy/gateway/issues/3290 & https://github.com/envoyproxy/gateway/issues/2269.

For these test failures, they are very hard to reproduce and debug. So by adding a hook for testSuite, executed when is failed, dumped all the configs and resources status etc. could be very helpful for developers to debug.

---

can also be mentioned via https://github.com/envoyproxy/gateway/issues/2855

shawnh2:
xref: https://github.com/kubernetes-sigs/gateway-api/issues/3131

zirain:
xref: https://github.com/kubernetes-sigs/gateway-api/pull/3786

PR #6215, opened 2025-05-28
e2e: use ConformanceOptions.Hook

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/3415


arkodg (approved):
LGTM thanks for driving this all the way from upstream till here !

zhaohuabing (approved):

Issue #4595, opened 2024-10-31
TLS Backend cert validation types


lsjostro:
*Description*:
I'm trying to configure TLS Backend validation of a service (out of my control and out side k8s). The cert on that external service is not using SAN DNS. Only Common name. Is there a better way other then creating a patch like this to remove validation `san_type: DNS`?
   
```
# foo-svc cert doesn't use SubjectAltName DNS so we need to patch away SAN DNS validation.
# CA cert validation is working, just not using SubjectAltName DNS which I'm patching away for now
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyPatchPolicy
metadata:
  name: foo-svc-delete-san-patch
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: GatewayClass
    name: eg
  type: JSONPatch
  jsonPatches:
    - type: "type.googleapis.com/envoy.config.cluster.v3.Cluster"
      name: "httproute/external-legacy-services/foo-svc/rule/0"
      operation:
        op: remove
        path: "/transport_socket_matches/0/transport_socket/typed_config/common_tls_context/combined_validation_context/default_validation_context/match_typed_subject_alt_names/0"
```

arkodg:
glad you were able to find the workaround @lsjostro 
from a spec perspective we made the SAN check mandatory to implement this section of the GEP https://gateway-api.sigs.k8s.io/geps/gep-1897/
```
The Hostname field is required and is to be used to configure the SNI the Gateway should use to connect to the backend. Implementations must validate that at least one name in the certificate served by the backend matches this field. We originally proposed using a list of allowed Subject Alternative Names, but determined that this was [not needed in the first round](https://github.com/kubernetes-sigs/gateway-api/pull/2113#issuecomment-1696127092), but may be added in the future.
```
With https://github.com/envoyproxy/gateway/issues/4381, we should be able to implement the newly added SAN field, hopefully for v1.3, from the spec https://gateway-api.sigs.k8s.io/geps/gep-3155/#sans-on-backendtlspolicy its not clear how we can disable this

once GEP-3155 is implemented, should we add this check as an opt in instead ? @guydc

guydc:
Some background:

- As per [RFC 2818](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1) [envoy doesn't support CN validation](https://github.com/envoyproxy/envoy/issues/6763)
[...]

PR #6222, opened 2025-05-29
feat(api): add insecureSkipVerify flag to Backend TLS settigns

AlexKrudu:
feat(api): add insecureSkipVerify flag to Backend TLS settigns

provides an opportunity for skipping upstream certificate validation 

Fixes https://github.com/envoyproxy/gateway/issues/4595

Decided to add the flag to `Backend.BackendSpec.TLS` instead of `EnvoyProxySpec.BackendTLSConfig` because chosen option provides better granularity – there may be cases where the user wants to disable validation selectively (skip validation only for some of Backends, not for all of them). Also, certificate validation settings within Envoy are already at a level equivalent to the Backend (the setting in Envoy is at the Cluster level, and in EG Backend, after transformations, it maps to the Cluster).

Release Notes: Yes/No


api/v1alpha1/backend_types.go
```diff
@@ -187,8 +187,18 @@ type BackendTLSSettings struct {
...
 	// +optional
 	WellKnownCACertificates *gwapiv1a3.WellKnownCACertificatesType `json:"wellKnownCACertificates,omitempty"`
+
+	// InsecureSkipVerify indicates whether the upstream's certificate verification
+	// should be skipped. Defaults to "false".
+	//
+	// +kubebuilder:default=false
+	// +optional
+	// +notImplementedHide
+	InsecureSkipVerify *bool `json:"insecureSkipVerify,omitempty"`
```
arkodg, line 201: im not a fan of this term, but its an industry wide term

arkodg (approved):
LGTM thanks !

api/v1alpha1/backend_types.go
arkodg: @envoyproxy/gateway-maintainers thoughts on `insecureSkipVerify` vs `skipValidation` ?

AlexKrudu:
I messed up - pushed implementation in the same branch and rolled it back

api/v1alpha1/backend_types.go
```diff
@@ -187,8 +187,18 @@ type BackendTLSSettings struct {
 	// must be specified with at least one entry for a valid configuration. Only one of
 	// CACertificateRefs or WellKnownCACertificates may be specified, not both.
 	//
+	// Only used for DynamicResolver backends.
```
arkodg, line 190: I dont think will work, we'll need to enable this TLS section for all backend, lets raise a GH issue to track it

zhaohuabing (approved):

PR #6249, opened 2025-06-02
fix: don't render .spec.replicas if HPA is enabled

y-rabie:
**What type of PR is this?**
fix: don't render .spec.replicas if HPA is enabled

The current chart template will render `.spec.replicas` even if autoscaling is enabled. This causes a conflict in ownership of the `.spec.replicas` field between the HPA and whoever is installing/reconciling the manifests.

The effect of this depends on the tool used to install the manifests:

Helm CLI -> the `.spec.replicas` is reset to the `.Values.deployment.replicas` value everytime Helm Install/Update is invoked.

AgroCD (or any GitOps tool) -> `.spec.replicas` is reset on each Sync... which can result into a constant back and forth between ArgoCD and the HPA that technically eliminate the benefits of the HPA.

### DIFF
--- a/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
+++ b/charts/gateway-helm/templates/envoy-gateway-deployment.yaml
@@ -7,7 +7,9 @@ metadata:
     control-plane: envoy-gateway
   {{- include "eg.labels" . | nindent 4 }}
 spec:
+  {{- if not .Values.hpa.enabled }}
   replicas: {{ .Values.deployment.replicas }}
+  {{- end }}
   selector:
     matchLabels:
       control-plane: envoy-gateway
--- a/test/helm/gateway-helm/horizontal-pod-autoscaler.out.yaml
+++ b/test/helm/gateway-helm/horizontal-pod-autoscaler.out.yaml
@@ -383,7 +383,6 @@ metadata:
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
 spec:
-  replicas: 1
   selector:
     matchLabels:
       control-plane: envoy-gateway


arkodg:
thanks for flagging this @y-rabie, instead of disabling replicas, can we add templating to unset it ? and add a doc strin in the hpa section to recommend users to do so ?

y-rabie:
@arkodg I'm not exactly sure what you mean by adding templating to unset it 😅 Do you mean leaving things as is and recommending to the users setting the replicas value to null when HPA is enabled?

arkodg:
if we added a doc string in the hpa section to comment out replicas
```
deployment:
  envoyGateway:
  .......
  # replicas: 1
```

could something like this be used in the template to avoid setting the replica ?

dberkerdem:
Any updates here?

sherifabdlnaby:
> if we added a doc string in the hpa section to comment out replicas
> 
> ```
> deployment:
>   envoyGateway:
>   .......
>   # replicas: 1
> ```
> 
> could something like this be used in the template to avoid setting the replica ?

If I understand the comment correctly then no, not with the current template. If we set `replicas: null` in the helm values it will still render, which means ArgoCD (and other GitOps tool) will track and own this field. We have to not render it at all so that the HPA can own the field without triggering an out-of-sync and fight the GitOps tool.

arkodg:
@sherifabdlnaby will something like this work ?
```
+  {{- if hasKey (default (dict) .Values.deployment) "replicas" }}
   replicas: {{ .Values.deployment.replicas }}
+  {{- end }}
```

proximix:
Did this ever get merged into a release?  I am still seeing this issue with Argo.  Any update on this?

zirain:
@y-rabie are you still working on this? if not, @proximix feel free to send another.

PR #6250, opened 2025-06-02
Allow for headless envoy services

ryanhristovski:
When clusterip on service is set to None it causes a not critical error (everything still functions as expected):

```
envoy-gateway-65589cc498-8q4gl envoy-gateway 2025-06-02T20:10:54.589Z   ERROR   provider    kubernetes/status_updater.go:119    unable to update status {"runner": "provider", "name": "internal-gateway", "namespace": "infra-routing", "error": "Gateway.gateway.networking.k8s.io "internal-gateway" is invalid: [status: Invalid value: "": "addresses[0]" must validate one and only one schema (oneOf). Found none valid, status: Invalid value: "": "addresses[0].value" must validate at least one schema (anyOf), status.addresses[0].value: Invalid value: "None": addresses[0].value in body must be of type ipv4: "None", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]"}
```

Caused by this:
```
apiVersion: v1
kind: Service
metadata:
  name: envoy-headless
  namespace: <names
spec:
  type: ClusterIP
  clusterIP: None # This makes it a headless service
  ports:
    - name: http
      port: 80
      targetPort: 10080
      protocol: TCP
...
```

This solution simply adds valid cluster IP's, and ignores None or "".

Signed-off-by: Ryan Hristovski <ryan.hristovski@docker.com>

### DIFF
--- a/internal/gatewayapi/status/gateway_test.go
+++ b/internal/gatewayapi/status/gateway_test.go
@@ -12,7 +12,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmp/cmpopts"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"


arkodg:
thanks @ryanhristovski 
can you add a test in https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/status/gateway_test.go
as well as a comment, highlighting that the step skips adding `None` which represents headless services

arkodg (dismissed):

arkodg:
test is failing @ryanhristovski

arkodg (approved):

shawnh2 (approved):

PR #6253, opened 2025-06-03
feat(translator): impl tls insecure-skip-verify

AlexKrudu:
**What type of PR is this?**
feat(translator): impl tls insecure-skip-verify

Implements upstream TLS certificate verification skipping

Fixes https://github.com/envoyproxy/gateway/issues/4595

Release Notes: Yes/No

### DIFF
--- a/api/v1alpha1/backend_types.go
+++ b/api/v1alpha1/backend_types.go
@@ -153,7 +153,8 @@ type BackendSpec struct {
 	Fallback *bool `json:"fallback,omitempty"`
 
 	// TLS defines the TLS settings for the backend.
-	// Only supported for DynamicResolver backends.
+	// TLS.CACertificateRefs and TLS.WellKnownCACertificates can only be specified for DynamicResolver backends.
+	// TLS.InsecureSkipVerify can be specified for any Backends
 	//
 	// +optional
 	TLS *BackendTLSSettings `json:"tls,omitempty"`
@@ -161,7 +162,7 @@ type BackendSpec struct {
 
 // BackendTLSSettings holds the TLS settings for the backend.
 // +kubebuilder:validation:XValidation:message="must not contain both CACertificateRefs and WellKnownCACertificates",rule="!(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 && has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"
-// +kubebuilder:validation:XValidation:message="must specify either CACertificateRefs or WellKnownCACertificates",rule="(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 || has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"
+// +kubebuilder:validation:XValidation:message="must not contain either CACertificateRefs or WellKnownCACertificates when InsecureSkipVerify is enabled",rule="!((has(self.insecureSkipVerify) && self.insecureSkipVerify) && ((has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0) || (has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")))"
 type BackendTLSSettings struct {
 	// CACertificateRefs contains one or more references to Kubernetes objects that
 	// contain TLS certificates of the Certificate Authorities that can be used
@@ -197,7 +198,6 @@ type BackendTLSSettings struct {
 	//
 	// +kubebuilder:default=false
 	// +optional
-	// +notImplementedHide
 	InsecureSkipVerify *bool `json:"insecureSkipVerify,omitempty"`
 }
 
--- a/internal/gatewayapi/backendtlspolicy.go
+++ b/internal/gatewayapi/backendtlspolicy.go
@@ -38,9 +38,15 @@ func (t *Translator) applyBackendTLSSetting(
 	// and not from the BackendTLSPolicy. This is because the BackendTLSPolicy requires a valid hostname, and
 	// dynamic resolvers's hostname is not fixed.
 	if isDynamicResolver {
-		if tlsBundle, err = t.processBackendTLSConfig(backendRef, backendNamespace, resources); err != nil {
+		upstreamConfig, err := t.processDynamicResolverBackendTLSConfig(backendRef, backendNamespace, resources)
+		if err != nil {
 			return nil, err
 		}
+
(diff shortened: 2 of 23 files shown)


internal/gatewayapi/backendtlspolicy.go
```diff
@@ -49,22 +55,48 @@ func (t *Translator) applyBackendTLSSetting(
...
+func (t *Translator) processDynamicResolverBackendTLSConfig(
+	backendRef gwapiv1.BackendObjectReference,
+	backendNamespace string,
+	resources *resource.Resources,
+) (*ir.TLSUpstreamConfig, error) {
+	backend := resources.GetBackend(backendNamespace, string(backendRef.Name))
+	if backend == nil {
+		return nil, fmt.Errorf("backend %s not found", backendRef.Name)
+	}
+	if backend.Spec.TLS == nil || (len(backend.Spec.TLS.CACertificateRefs) == 0 && backend.Spec.TLS.WellKnownCACertificates == nil) {
```
arkodg, line 99: since the CEL was relaxed, we'll need to return an err if both `CACertificateRefs` & `WellKnownCACertificates ` is empty, and it should show up in the status

api/v1alpha1/backend_types.go
```diff
@@ -161,7 +161,6 @@ type BackendSpec struct {
 
 // BackendTLSSettings holds the TLS settings for the backend.
 // +kubebuilder:validation:XValidation:message="must not contain both CACertificateRefs and WellKnownCACertificates",rule="!(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 && has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"
-// +kubebuilder:validation:XValidation:message="must specify either CACertificateRefs or WellKnownCACertificates",rule="(has(self.caCertificateRefs) && size(self.caCertificateRefs) > 0 || has(self.wellKnownCACertificates) && self.wellKnownCACertificates != \"\")"
```
zhaohuabing, line 164: I think these two combinations are not allowed and should be checked with CEL:
*  (InsecureSkipVerify == nil || InsecureSkipVerify == false) && WellKnownCACertificates == nil && caCertificateRefs == nil 
* InsecureSkipVerify == true && (WellKnownCACertificates != nil || caCertificateRefs != nil)

internal/gatewayapi/backendtlspolicy.go
zhaohuabing: Better verify it with CEL, if not possible, then we can surface errors in the status. https://github.com/envoyproxy/gateway/pull/6253/files#r2128672386

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -86,6 +118,8 @@ func (t *Translator) processBackendTLSConfig(
 		Certificate: caCert,
 		Name:        fmt.Sprintf("%s/%s-ca", backend.Name, backend.Namespace),
 	}
+	tlsBundle.InsecureSkipVerify = ptr.Deref(backend.Spec.TLS.InsecureSkipVerify, false)
```
zhaohuabing: CA certificate and InsecureSkipVerify == true should not co-exist.

```

  tlsBundle.InsecureSkipVerify = ptr.Deref(backend.Spec.TLS.InsecureSkipVerify, false)

  if !tlsBundle.InsecureSkipVerify {
    // set CACertificate
    tlsBundle := &ir.TLSUpstreamConfig{
      UseSystemTrustStore: ptr.Deref(backend.Spec.TLS.WellKnownCACertificates, "") == gwapiv1a3.WellKnownCACertificatesSystem,
    }
    if tlsBundle.UseSystemTrustStore {
      tlsBundle.CACertificate = &ir.TLSCACertificate{
        Name: fmt.Sprintf("%s/%s-ca", backend.Name, backend.Namespace),
      }
    } else {
      caCert, err := getCaCertsFromCARefs(backend.Spec.TLS.CACertificateRefs, resources)
      if err != nil {
        return nil, err
      }
      tlsBundle.CACertificate = &ir.TLSCACertificate{
        Certificate: caCert,
        Name:        fmt.Sprintf("%s/%s-ca", backend.Name, backend.Namespace),
      }
    }
  }
```

internal/gatewayapi/backendtlspolicy.go
```diff
@@ -49,22 +55,48 @@ func (t *Translator) applyBackendTLSSetting(
...
+
 	backend := resources.GetBackend(backendNamespace, string(backendRef.Name))
 	if backend == nil {
 		return nil, fmt.Errorf("backend %s not found", backendRef.Name)
 	}
 	if backend.Spec.TLS == nil {
+		return tlsConfig, nil
+	}
+
+	tlsConfig.InsecureSkipVerify = ptr.Deref(backend.Spec.TLS.InsecureSkipVerify, false)
```
zhaohuabing, line 86: ~~The TLS setting in the Backend is supposed to be used by the `DynamicResolver` type only. For other Backend Types, we should always use BackendTLSPolicy for consistency~~.

I've learned some background from the issue: https://github.com/envoyproxy/gateway/issues/4595

If we would like to use this for other types, then the comment in the API should be modified: `UseSystemTrustStore` and `CACertificate` can only be used for DynamicResolver, but `InsecureSkipVerify` can be used for any type.

```
          // TLS defines the TLS settings for the backend.
	// Only supported for DynamicResolver backends.
	//
	// +optional
	TLS *BackendTLSSettings `json:"tls,omitempty"`
```

api/v1alpha1/backend_types.go
AlexKrudu: > InsecureSkipVerify == true && (WellKnownCACertificates != nil || caCertificateRefs != nil)

that's reasonable

> (InsecureSkipVerify == nil || InsecureSkipVerify == false) && WellKnownCACertificates == nil && caCertificateRefs == nil

`InsecureSkipVerify == false && WellKnownCACertificates == nil && caCertificateRefs == nil` I think this case makes sense and shouldn't be prohibited, doesn't it? For example, it could be specifed as explicit setting for not skipping TLS verification for `Endpoints` Backend type

AlexKrudu: `Type=DynamicResolver && InsecureSkipVerify == false && WellKnownCACertificates == nil && caCertificateRefs == nil` I think this case should be prohibited

AlexKrudu:
All fixed, I think

arkodg (approved):
LGTM thanks

zhaohuabing (approved):
LGTM thanks！

PR #6267, opened 2025-06-05
Hotfix headless service programmed status flapping

ryanhristovski:
When you have a headless service for envoy the programmed status of the gateway can flap from True<->False. This patch favours service definitions with valid IP's and if thats not the case then it defaults to whatever is available.

The flapping does not cause any actual effect on the traffic, just causes a false negative in the gateway status.

### DIFF
--- a/internal/provider/kubernetes/predicates_test.go
+++ b/internal/provider/kubernetes/predicates_test.go
@@ -6,6 +6,7 @@
 package kubernetes
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"testing"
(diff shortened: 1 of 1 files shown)


internal/provider/kubernetes/predicates.go
```diff
@@ -642,6 +642,16 @@ func (r *gatewayAPIReconciler) envoyServiceForGateway(ctx context.Context, gatew
 	if len(services.Items) == 0 {
 		return nil, nil
 	}
+
+	// Prefer non-headless services over headless services for Gateway address assignment
```
arkodg, line 646: can you elaborate on this change ?
how can we have multiple generated services that map to the same gateway ?

ryanhristovski: We have the main clusterip service generated via:
```
kind: EnvoyProxy
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        name: internal-envoy
        type: ClusterIP
```

But due to the way were migrating, we still require a headless service for our legacy ingress stack to point to (once we complete the migration we will no longer have a headless svc). We manage this via a simple service definition in our helm chart, something like:
```
# Source: envoy-proxy/templates/headless-service.yaml
apiVersion: v1
kind: Service
metadata:
spec:
  type: ClusterIP
  clusterIP: None 
  ports:
    - name: http
      port: 80
      targetPort: 10080
      protocol: TCP
```

So this is quite a niche use-case and I don't think many others will run into it - but it's a requirement for us to have both of these service definitions simultaneously.

ryanhristovski:
The usecase for this is too specific and adds unneeded complexity. Closing

PR #6271, opened 2025-06-08
feat: support original dst based routing

### LINKED ISSUE #6234: support endpoint picker based on header/dynamic metadata
*Description*:
would be used.

AIBrix:

https://github.com/vllm-project/aibrix/blob/main/config/gateway/gateway.yaml looks like aibrix need to apply envoypatchpolicy in every new Gateway instance, thoughts on replacing it with other approaches?

1. Route to extension server only when path /v1 + header route-strategy 
2. Route to endpoint based on target-pod header

Gateway-api inference extension also needs to pick endpoint based on header or metadata.

Based on Envoy Override Host Load Balancer Policy https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/override_host/v3/override_host.proto which supports a fallback IP that be used during a retry

Xunzhuo:
**What type of PR is this?**

feat: support host override based routing

support choose the ep by specific patterns: header, metadata.

Scenarios like llm inference endpoint picker

Fixes https://github.com/envoyproxy/gateway/issues/6234

Release Notes: Yes


api/v1alpha1/backend_types.go
```diff
@@ -157,6 +158,13 @@ type BackendSpec struct {
 	//
 	// +optional
 	TLS *BackendTLSSettings `json:"tls,omitempty"`
+
+	// StaticResolver defines the configuration for StaticResolver backend type.
+	// This allows endpoint selection based on request headers or dynamic metadata.
+	// Only supported for StaticResolver backends.
+	//
+	// +optional
+	StaticResolver *StaticResolverSettings `json:"staticResolver,omitempty"`
```
arkodg: any other name recommendations @envoyproxy/gateway-maintainers @envoyproxy/gateway-reviewers

arkodg:
we discussed this in the community meeting yesterday and the preference is, instead of adding a new `Type`, this can be represented as a IPEndpoint https://github.com/envoyproxy/gateway/blob/e78e2684d7b57526344f5f420be3dac78cbb2c15/api/v1alpha1/backend_types.go#L81 field called Ref or something similar whose type would be something like `OverrideHostSource`

api/v1alpha1/backend_types.go
rudrakhp: Since we expect IP:Port addresses and don't actually resolve the address, does `BackendTypeHostOverride BackendType = "HostOverride"`? More in line with the simple `Endpoints` we have but specifies we allow overriding hosts (from request)?
```
type: hostOverride
hostOverrideSettings:
  overrideHostSources:
    header: x-session-host
```

Xunzhuo: Looks good, updated.

Xunzhuo:
Here is how it looks like now:

```
  kind: Backend
  metadata:
    name: backend-routing-based-on-header
    namespace: default
  spec:
    type: HostOverride
    hostOverrideSettings:
      overrideHostSources:
      - header: target-pod
```

api/v1alpha1/backend_types.go
```diff
@@ -123,10 +123,11 @@ type UnixSocket struct {
 
 // BackendSpec describes the desired state of BackendSpec.
 // +kubebuilder:validation:XValidation:rule="self.type != 'DynamicResolver' || !has(self.endpoints)",message="DynamicResolver type cannot have endpoints specified"
+// +kubebuilder:validation:XValidation:rule="self.type != 'HostOverride' || !has(self.endpoints)",message="HostOverride type cannot have endpoints specified"
```
rudrakhp, line 126: nit: validation to ensure hostOverrideSettings is set?

internal/gatewayapi/backend.go
```diff
@@ -115,3 +137,33 @@ func validateBackend(backend *egv1a1.Backend) status.Error {
 	}
 	return nil
 }
+
+func validateHostOverrideSettings(settings *egv1a1.HostOverrideSettings) error {
```
rudrakhp: Are we validating something here that we are not/cannot do in CEL?

internal/xds/translator/testdata/in/xds-ir/static-resolver.yaml
rudrakhp, line 1: nit: rename to host-override.yaml to make searching easier?

rudrakhp (commented):
Minor comments rest LGTM!

internal/xds/translator/testdata/in/xds-ir/static-resolver.yaml
Xunzhuo: Done

internal/gatewayapi/backend.go
Xunzhuo: Done

api/v1alpha1/backend_types.go
Xunzhuo: Done

rudrakhp:
@Xunzhuo can you fix the lint check (failing due to extra line)?

Xunzhuo:
@rudrakhp working on it, thanks!

Xunzhuo:
PTAL @rudrakhp thanks!

rudrakhp (dismissed):
LGTM!

internal/gatewayapi/backend.go
```diff
@@ -59,17 +64,31 @@ func validateBackend(backend *egv1a1.Backend) status.Error {
...
+				fmt.Errorf("HostOverride type cannot have endpoints specified"),
 				status.RouteReasonInvalidBackendRef,
 			)
 		}
-		if len(backend.Spec.TLS.CACertificateRefs) > 0 {
+	default: // BackendTypeEndpoints
+		if backend.Spec.TLS != nil {
+			if backend.Spec.TLS.WellKnownCACertificates != nil {
+				return status.NewRouteStatusError(
+					fmt.Errorf("TLS.WellKnownCACertificates settings can only be specified for DynamicResolver backends"),
```
zhaohuabing: TLS should also be used for `HostOverride` because the host for `HostOverride` type is also not fixed - we can't use BackendTLSPolicy to set TLS for it.

zhaohuabing (commented):
Can we add an e2e test to verify this works?

internal/gatewayapi/backend.go
Xunzhuo: Good point, updated.

Xunzhuo:
I am verifying the e2e test failures today, and had a discussion with @wbpcode, this current API looks not complete:

The HostOverride lbPolicy requires the fallback upstream, so the previous approach is not correct, should be done by adding at least one endpoint in Backend when type is HostOverride

```
kind: Backend
metadata:
  name: backend-routing-based-on-header
  namespace: default
spec:
  type: HostOverride
  hostOverrideSettings:
    overrideHostSources:
    - header: target-pod
  endpoints:
    - fqdn:
        hostname: fallback.default.svc.cluster.local
        port: 80
```

This can be achieved another feature like, we do best-match for the pod in cluster, but if no best-match pod selected, use the endpoint in Backend to fallback, even it can be out-of-cluster.

arkodg:
@Xunzhuo if overrideHost LB Policy + Original Dst cluster dont work together in Envoy proxy, we'll have to use Option 1 https://github.com/envoyproxy/gateway/issues/6234#issuecomment-2951260617

Xunzhuo:
how envoy impl the host override(source/extensions/load_balancing_policies/override_host/load_balancer.cc):
```
OverrideHostLoadBalancer::LoadBalancerImpl::chooseHost(LoadBalancerContext* context) {
  if (!context || !context->requestStreamInfo()) {
    // If there is no context or no request stream info, we can't use the
    // metadata, so we just return a host from the fallback picker.
    return fallback_picker_lb_->chooseHost(context);
  }

  OverrideHostFilterState* override_host_state = nullptr;
  if (override_host_state =
          context->requestStreamInfo()->filterState()->getDataMutable<OverrideHostFilterState>(
              OverrideHostFilterState::kFilterStateKey);
      override_host_state == nullptr) {
    auto state_ptr = std::make_shared<OverrideHostFilterState>(getSelectedHosts(context));
    override_host_state = state_ptr.get();

    context->requestStreamInfo()->filterState()->setData(
        OverrideHostFilterState::kFilterStateKey, std::move(state_ptr),
        StreamInfo::FilterState::StateType::Mutable);
  }

  if (override_host_state->empty()) {
    ENVOY_LOG(trace, "No overriden hosts were found. Using fallback LB policy.");
    return fallback_picker_lb_->chooseHost(context);
  }

  if (HostConstSharedPtr host = getEndpoint(*override_host_state); host != nullptr) {
    return {host};
  }

  // If some endpoints were found, but none of them are available in
  // the cluster endpoint set, or the number of retries in the retry policy
  // exceeds the number of fallback endpoints, then use to the fallback LB
  // policy.
  ENVOY_LOG(trace, "Failed to find any endpoints from metadata in the cluster. "
                   "Using fallback LB policy.");
  return fallback_picker_lb_->chooseHost(context);
}

```

forward to an endpoint which is not in the cluster(I manually add 3 endpoint to the cluster 10.96.93.71:8080, 10.96.166.130:8080, 10.96.82.175:8080):
```
[2025-07-03 03:23:36.542][35][debug][http] [source/common/http/conn_manager_impl.cc:1168] [Tags: "ConnectionId":"334","StreamId":"14294559941569293744"] request end stream timestamp recorded
[2025-07-03 03:23:36.542][35][debug][connection] [./source/common/network/connection_impl.h:99] [Tags: "ConnectionId":"334"] current connecting state: false
[2025-07-03 03:23:36.542][35][debug][router] [source/common/router/router.cc:542] [Tags: "ConnectionId":"334","StreamId":"14294559941569293744"] cluster 'httproute/gateway-conformance-infra/httproute-host-override-header/rule/0' match for URL '/header-override'
[2025-07-03 03:23:36.542][35][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:245] Selected endpoints: 10.244.0.22:3000
[2025-07-03 03:23:36.542][35][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:260] Looking up 10.244.0.22:3000 in 10.96.93.71:8080, 10.96.166.130:8080, 10.96.82.175:8080
[2025-07-03 03:23:36.542][35][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:281] Number of attempts has exceeded the number of override hosts.
[2025-07-03 03:23:36.542][35][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:182] Failed to find any endpoints from metadata in the cluster. Using fallback LB policy.
[2025-07-03 03:23:36.542][35][debug][router] [source/common/router/router.cc:802] [Tags: "ConnectionId":"334","StreamId":"14294559941569293744"] router decoding headers:
':authority', '172.18.0.203'
':path', '/header-override'
':method', 'GET'
':scheme', 'http'
'user-agent', 'curl/8.7.1'
'accept', '*/*'
'target-pod', '10.244.0.22:3000'
'x-forwarded-for', '172.18.0.1'
'x-forwarded-proto', 'http'
'x-envoy-external-address', '172.18.0.1'
'x-request-id', '1823981f-9ccf-4c1b-aa18-fc3f45f6fd2a'
```



forward to an endpoint which in the cluster:
```
[2025-07-03 03:21:30.301][22][debug][router] [source/common/router/router.cc:542] [Tags: "ConnectionId":"291","StreamId":"3232757434578693721"] cluster 'httproute/gateway-conformance-infra/httproute-host-override-header/rule/0' match for URL '/header-override'
[2025-07-03 03:21:30.301][22][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:245] Selected endpoints: 10.96.82.175:8080
[2025-07-03 03:21:30.301][22][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:260] Looking up 10.96.82.175:8080 in 10.96.93.71:8080, 10.96.166.130:8080, 10.96.82.175:8080
[2025-07-03 03:21:30.301][22][trace][upstream] [source/extensions/load_balancing_policies/override_host/load_balancer.cc:274] Selected endpoint: 10.96.82.175:8080
[2025-07-03 03:21:30.301][22][debug][router] [source/common/router/router.cc:802] [Tags: "ConnectionId":"291","StreamId":"3232757434578693721"] router decoding headers:
':authority', '172.18.0.203'
':path', '/header-override'
':method', 'GET'
':scheme', 'http'
'user-agent', 'curl/8.7.1'
'accept', '*/*'
'target-pod', '10.96.82.175:8080'
'x-forwarded-for', '172.18.0.1'
'x-forwarded-proto', 'http'
'x-envoy-external-address', '172.18.0.1'
'x-request-id', '2083f99f-e6c8-425f-9808-afb4d8ca0b94
```

In summary, if we want to pick an endpoint with the host override policy, it has to be in-range of the endpoints belong to the cluster, so we cannot use the dummy endpoint to hack, it will always go to the fallback policy.

Xunzhuo:
The Backend API may only fit for the original dst implementation:

Header:
```
  kind: Backend
  metadata:
    name: backend-routing-based-on-header
    namespace: default
  spec:
    type: OriginalDestination
      originalDestinationSetting:
         header: target-pod
```

Metadata:
```
  kind: Backend
  metadata:
    name: backend-routing-based-on-header
    namespace: default
  spec:
    type: OriginalDestination
      originalDestinationSetting:
        metadata:
          key: "envoy.lb"
          path:
          - key: "override-host"
```

arkodg:
originalDestination is not a great name imo 🤷

Xunzhuo:
@arkodg we need some naming genius to help us 😄

PR #6279, opened 2025-06-09
fix: add optional flag for hostname fix #6278

### LINKED ISSUE #6278: `hostname` should not be required in `HTTPActiveHealthChecker`
*Description*:
`hostname` is marked as required in the `HTTPActiveHealthChecker`: https://gateway.envoyproxy.io/latest/api/extension_types/#httpactivehealthchecker

this should be optional, as the doc mentions the default behavior:
> Default: HTTPRoute or GRPCRoute hostname.

*Repro steps*:

sharing.

*Environment*:

*Logs*:

yyadavalli:
**What type of PR is this?**

Fixes #6278 

Release Notes: Yes/No

### DIFF
--- a/api/v1alpha1/healthcheck_types.go
+++ b/api/v1alpha1/healthcheck_types.go
@@ -155,6 +155,7 @@ type HTTPActiveHealthChecker struct {
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +optional
 	Hostname *string `json:"hostname,omitempty" yaml:"hostname,omitempty"`
 	// Path defines the HTTP path that will be requested during health checking.
 	// +kubebuilder:validation:MinLength=1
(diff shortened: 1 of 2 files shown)


arkodg:
@yyadavalli above tag may not be used in docs, you may need to use `// +optional` to populate the docs column

api/v1alpha1/healthcheck_types.go
```diff
@@ -152,6 +152,7 @@ type HTTPActiveHealthChecker struct {
 	// Hostname defines the HTTP host that will be requested during health checking.
 	// Default: HTTPRoute or GRPCRoute hostname.
 	//
+	// +kubebuilder:validation:Optional
```
muwaqar: use `// +optional` as the rest of places in this file do for optional fields

yyadavalli:
> @yyadavalli above tag may not be used in docs, you may need to use `// +optional` to populate the docs column

Ack, will use the other tag

api/v1alpha1/healthcheck_types.go
muwaqar: you need to run `make kube-generate` to reflect these changes in the CRD (the `git status` would show a modified generated file)

muwaqar (changes requested):

arkodg (approved):
LGTM thanks

Issue #5733, opened 2025-04-14
ResponseHeaderModifier does not permit to set a content with commas


saez0pub:
*Description*:
>I cannot set `Cache-Control: foo,bar` with ResponseHeaderModifier

*Repro steps*:

1. Deploy the quickstart example
2. Apply the following HTTPRoute:
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend
  namespace: default
spec:
  hostnames:
  - www.example.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
      weight: 1
    filters:
    - responseHeaderModifier:
        set:
        - name: Cache-Control
          value: private, no-store
      type: ResponseHeaderModifier
    matches:
    - path:
        type: PathPrefix
        value: /
```

3. Run the following curl command:

curl -i http://${ENVOY_GATEWAY_IP}/-H 'Host: www.example.com' -H 'X-Echo-Set-Header: Cache-control: value1' | grep ^cache-control
You will see This:
```
cache-control:  no-store
```

Expected output:
```
cache-control: private, no-store
```

Proxy config extract:
```
          "response_headers_to_add": [
           {
            "header": {
             "key": "Cache-Control",
             "value": "private"
            },
            "append_action": "OVERWRITE_IF_EXISTS_OR_ADD"
           },
           {
            "header": {
             "key": "Cache-Control",
             "value": " no-store"
            },
            "append_action": "OVERWRITE_IF_EXISTS_OR_ADD"
           }
          ],
```

saez0pub:
Seems linked to #4031

I'm kinda puzzled about this: https://gateway-api.sigs.k8s.io/reference/spec/#httpheaderfilter
> Configuration to set or add multiple values for a header must use RFC 7230 header value formatting, separating each value with a comma.

If I look at the RFC, it says
> A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).

I think the specification was talking more about the former which is more of a comma-separated join than generating an array of values from a comma-separated list.

[...]

PR #6281, opened 2025-06-09
fix: handle multi-value response headers as comma-separated per RFC 7230

patrostkowski:
**What type of PR is this?**

fix

This change updates the translation logic of ResponseHeaderModifier so that when multiple values are specified for a header, they are joined into a single, comma-separated value as required by RFC 7230. As a result, the generated Envoy config now emits one HeaderValueOption per header, with all values joined by commas.

Tested using following config:

```
kind: HTTPRoute
metadata:
  name: backend
  namespace: default
spec:
  hostnames:
  - www.example.com
  parentRefs:
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: backend
      port: 3000
      weight: 1
    filters:
    - responseHeaderModifier:
        set:
        - name: Cache-Control
          value: private, no-store
      type: ResponseHeaderModifier
    matches:
    - path:
        type: PathPrefix
        value: /
```

```bash
# curl latest image
root@ubuntu:/$ curl -i http://10.96.20.98/ -H 'Host: www.example.com' -H 'X-Echo-Set-Header: Cache-control: value1' | grep ^cache-control
cache-control:  no-store

# curl e4c8f89 image
root@ubuntu:/$ curl -i http://10.96.20.98/ -H 'Host: www.example.com' -H 'X-Echo-Set-Header: Cache-control: value1' | grep ^cache-control
cache-control: private, no-store
```

Fixes https://github.com/envoyproxy/gateway/issues/5733

Release Notes: Yes/No

### DIFF
--- a/internal/xds/translator/route.go
+++ b/internal/xds/translator/route.go
@@ -522,6 +522,10 @@ func mirrorPercentByPolicy(mirror *ir.MirrorPolicy) *corev3.RuntimeFractionalPer
 }
 
 func buildXdsAddedHeaders(headersToAdd []ir.AddHeader) []*corev3.HeaderValueOption {
+	if len(headersToAdd) == 0 {
+		return nil
+	}
+
 	headerValueOptions := []*corev3.HeaderValueOption{}
 
 	for _, header := range headersToAdd {
--- a/internal/gatewayapi/filters.go
+++ b/internal/gatewayapi/filters.go
@@ -617,7 +617,7 @@ func (t *Translator) processResponseHeaderModifierFilter(
 			newHeader := ir.AddHeader{
 				Name:   headerKey,
 				Append: true,
-				Value:  strings.Split(addHeader.Value, ","),
+				Value:  []string{addHeader.Value},
 			}
 
 			filterContext.AddResponseHeaders = append(filterContext.AddResponseHeaders, newHeader)
@@ -672,7 +672,7 @@ func (t *Translator) processResponseHeaderModifierFilter(
 			newHeader := ir.AddHeader{
 				Name:   string(setHeader.Name),
 				Append: false,
-				Value:  strings.Split(setHeader.Value, ","),
+				Value:  []string{setHeader.Value},
 			}
 
 			filterContext.AddResponseHeaders = append(filterContext.AddResponseHeaders, newHeader)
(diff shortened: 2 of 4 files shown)


internal/xds/translator/testdata/out/xds-ir/http-route-request-headers.routes.yaml
```diff
@@ -11,10 +11,7 @@
       requestHeadersToAdd:
       - header:
           key: some-header-multi-value
-          value: some-value
-      - header:
-          key: some-header-multi-value
-          value: some-additional-value
```
zirain: will the prev way still work?

patrostkowski: I had to change the test because it was failing. I think this change might be breaking, since anyone using this feature and expecting the same header specified multiple times with different values would not be RFC 7230 compliant. I could add a feature flag to control this behavior if needed.

internal/xds/translator/route.go
```diff
@@ -541,16 +545,15 @@ func buildXdsAddedHeaders(headersToAdd []ir.AddHeader) []*corev3.HeaderValueOpti
 				KeepEmptyValue: true,
 			})
 		} else {
-			for _, val := range header.Value {
```
arkodg: this will break the logic for requestHeaders

for responseHeaders, is the fix to avoid the split in the gateway-api layer https://github.com/envoyproxy/gateway/blob/32d61c492167a37311315ab04bed383fb6504b03/internal/gatewayapi/filters.go#L620 ?

please add a test case in the gateway-api and xds/translator layer

arkodg:
hey @patrostkowski we use YAML based testcases for better readability 

* e.g. for gateway-api lib - internal/gatewayapi/testdata/grpcroute-with-request-header-modifier.in.yaml
* for xds translator - internal/xds/translator/testdata/in/xds-ir/http-route-request-headers.yaml

if you run `make testdata` it will even generate the out files for you

patrostkowski:
@arkodg sure, let me adjust the tests then

arkodg:
@patrostkowski still working on this ?

jasmin-terrien:
@patrostkowski any news/time to finish your fix ?

VonNao:
This still seems to be worked on right?

arkodg:
@jasmin-terrien @VonNao feel free to pick this one up, there's been inactivity for a long time

arkodg:
fixed with [#7436](https://github.com/envoyproxy/gateway/pull/7436)

PR #6295, opened 2025-06-11
feat: support custom redirects

### LINKED ISSUE #5475: Support headers for responseOverride

rudrakhp:
**What type of PR is this?**

feat: support custom redirects in BackendTrafficPolicy

To support setting redirect responses in custom response overrides.

Fixes #5475 
Related #5549 

Release Notes: Yes/No


internal/ir/xds.go
```diff
@@ -591,7 +591,9 @@ type ResponseOverrideRule struct {
 	// Match configuration.
 	Match CustomResponseMatch `json:"match"`
 	// Response configuration.
-	Response CustomResponse `json:"response"`
+	Response *CustomResponse `json:"response"`
```
arkodg: can you add a `omitempty` to both the fields, so they dont show up as `null` in test file

rudrakhp: nice catch added it in the existing Redirect struct as well

zirain (approved):

PR #6296, opened 2025-06-11
feat(api): allow pinning custom image repository for Envoy Proxy.

### LINKED ISSUE #6066: Allow custom image repository for Envoy Proxy image
*Description*:

We're running a caching AWS ECR for all 3rd party images and would like Envoy Proxy to pull its image from it, without having to pin the image tag/version, and let Envoy Gateway decide on the tag/version. 

Currently it's possible to configure the Envoy Proxy image, but only as a single repository + tag combination.
The "KubernetesContainerSpec" has an "image" string parameter, which we set like:

```
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        container:
          image: 1234.dkr.ecr.eu-west-1.amazonaws.com/dockerhub/envoyproxy/envoy:v1.33.1
```

As you can see, the "image" must include the tag. There is no separate fields for repository and tag. 

Being able to specify repository and tag separately, is a common pattern.

It would be great if following Envoy Proxy configuration would be supported:

```
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        container:
          image:
            repository: 1234.dkr.ecr.eu-west-1.amazonaws.com/dockerhub/envoyproxy/envoy
            tag: v1.33.1
```

sudiptob2:
This PR allows users to specify a custom image repository for **EnvoyProxy** without needing to provide a tag.

Fixes #6066

Release Notes: No

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -208,7 +208,7 @@ type KubernetesPodSpec struct {
 }
 
 // KubernetesContainerSpec defines the desired state of the Kubernetes container resource.
-// +kubebuilder:validation:XValidation:rule="!has(self.image) || !has(self.imageRepository)",message="image and imageRepository are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!has(self.image) || !has(self.imageRepository)",message="Either image or imageRepository can be set."
 type KubernetesContainerSpec struct {
 	// List of environment variables to set in the container.
 	//
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -208,6 +208,7 @@ type KubernetesPodSpec struct {
 }
 
 // KubernetesContainerSpec defines the desired state of the Kubernetes container resource.
+// +kubebuilder:validation:XValidation:rule="!has(self.image) || !has(self.imageRepository)",message="image and imageRepository are mutually exclusive"
 type KubernetesContainerSpec struct {
 	// List of environment variables to set in the container.
 	//
@@ -227,10 +228,18 @@ type KubernetesContainerSpec struct {
 	// +optional
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
 
-	// Image specifies the EnvoyProxy container image to be used, instead of the default image.
+	// Image specifies the EnvoyProxy container image to be used including a tag, instead of the default image.
+	// This field is mutually exclusive with ImageRepository.
 	//
 	// +optional
 	Image *string `json:"image,omitempty"`
+
+	// ImageRepository specifies the container image repository to be used without specifying a tag.
+	// The default tag will be used.
+	// This field is mutually exclusive with Image.
+	//
+	// +optional
+	ImageRepository *string `json:"imageRepository,omitempty"`
 
 	// VolumeMounts are volumes to mount into the container's filesystem.
 	// Cannot be updated.
--- a/internal/infrastructure/kubernetes/proxy/resource.go
+++ b/internal/infrastructure/kubernetes/proxy/resource.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/containers/image/v5/docker/reference"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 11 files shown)


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -479,3 +485,41 @@ func expectedShutdownManagerSecurityContext(containerSpec *egv1a1.KubernetesCont
...
 }
+
+func resolveProxyImage(containerSpec *egv1a1.KubernetesContainerSpec) (string, error) {
+	if containerSpec == nil {
+		return "", fmt.Errorf("containerSpec is nil")
+	}
+
+	repo := ptr.Deref(containerSpec.ImageRepository, "")
+	if repo != "" {
+		tag, err := getImageTag(egv1a1.DefaultEnvoyProxyImage)
```
rudrakhp, line 496: Shouldn't this be `*containerSpec.Image`? Why is it always default?

api/v1alpha1/shared_types.go
```diff
@@ -208,6 +208,7 @@ type KubernetesPodSpec struct {
 }
 
 // KubernetesContainerSpec defines the desired state of the Kubernetes container resource.
+// +kubebuilder:validation:XValidation:rule="!has(self.image) || !has(self.imageRepository)",message="image and imageRepository are mutually exclusive"
```
rudrakhp: Can you help me understand how a user wanting to use `imageRepository` will set tag if only one of `imageRepository` or `image` is allowed to be set?

sudiptob2: Thanks for the review! Referring to https://github.com/envoyproxy/gateway/issues/6066#issuecomment-2951233503, here is the summary:
- If the user wants to specify the image tag explicitly, they should use the format: `image: repo:tag`. This has been the standard format across all versions up to this PR.
- If the user prefers to define only the image repository and allow Envoy Gateway to determine the appropriate tag, they should use the format: `imageRepository: repo_url`.

internal/infrastructure/kubernetes/proxy/resource.go
sudiptob2: We want to extract the default tag pinned here with  the`envoy-gateway` release https://github.com/envoyproxy/gateway/blob/76e256d243d4aaba83ea7eb2502e541c1900929c/api/v1alpha1/shared_types.go#L27

This line will only execute when `ImageRepository` is set. So technically, `*containerSpec.Image` would contain the same string as `egv1a1.DefaultEnvoyProxyImage`.

api/v1alpha1/shared_types.go
rudrakhp: Can we change this message to something more meaningful like "only one of image or imageRepository can be set"

rudrakhp: Can we add a cel test for this?

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -479,3 +485,41 @@ func expectedShutdownManagerSecurityContext(containerSpec *egv1a1.KubernetesCont
...
+		if err != nil {
+			return "", err
+		}
+		return fmt.Sprintf("%s:%s", repo, tag), nil
```
rudrakhp, line 500: Should we have any validation in place to ensure user is not passing a tag in imageRepository?

api/v1alpha1/shared_types.go
sudiptob2: Updated with simpler wording.

sudiptob2: Added test for CEL validation

internal/infrastructure/kubernetes/proxy/resource.go
sudiptob2: Yeah, a validation could be useful, but in my opinion, it's redundant. There are many ways an `imageRepository` string can be invalid—not just tag. Also, the `image` field doesn’t currently validate whether a tag is present. So, I think we can skip this validation in this case as well.

arkodg: good to track this with a GH issue, so we dont forget about general image validation

rudrakhp (approved):

PR #6297, opened 2025-06-11
Dynamic variable rendering for btp responseOverride

ryanhristovski:
- dynamic variable rendering for BTP custom responses (such as access logging variables seen here: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage, allowing us to display headers and client data)

What this looks like:
```
kind: BackendTrafficPolicy
metadata:
  name: response-override-enhanced
spec:
  targetRef:
    kind: HTTPRoute
    name: response-override-enhanced
  responseOverride:
    - match:
        statusCodes:
          - type: Value
            value: 429
      response:
        body:
          type: Inline
		  inline: '%REQ(x-user-id)% is ratelimited.'
        contentType: "application/json"
```

This is particularly useful in our use case to give users more information on why theyre being ratelimited when we have multiple ratelimits in a single BTP.

### DIFF
--- a/site/content/en/latest/api/extension_types.md
+++ b/site/content/en/latest/api/extension_types.md
@@ -41,9 +41,9 @@ _Appears in:_
 
 | Value | Description |
 | ----- | ----------- |
-| `http/1.0` | HTTPProtocolVersion1_0 specifies that HTTP/1.0 should be negotiable with ALPN<br /> |
-| `http/1.1` | HTTPProtocolVersion1_1 specifies that HTTP/1.1 should be negotiable with ALPN<br /> |
-| `h2` | HTTPProtocolVersion2 specifies that HTTP/2 should be negotiable with ALPN<br /> |
+| `http/1.0` | HTTPProtocolVersion1_0 specifies that HTTP/1.0 should be negotiable with ALPN<br /> | 
+| `http/1.1` | HTTPProtocolVersion1_1 specifies that HTTP/1.1 should be negotiable with ALPN<br /> | 
+| `h2` | HTTPProtocolVersion2 specifies that HTTP/2 should be negotiable with ALPN<br /> | 
 
 
 #### ALSEnvoyProxyAccessLog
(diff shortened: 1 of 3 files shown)


arkodg:
reg `responseHeadersToAdd` @rudrakhp were you planning on picking that one up https://github.com/envoyproxy/gateway/issues/5475, or should @ryanhristovski continue here

hey @ryanhristovski prefer if we break up the 2 features into separate PRs - reg formatter for `body`, do we need an API for this, or can this be taken care of under the hood ?

ryanhristovski:
@arkodg formatter for body is taken care of underhood with no api change iirc and requires minimal changes, thats a good idea ill break that up into a couple prs:

1. var formatter for body
2. responseHeadersToAdd (assuming @rudrakhp isnt doing that)
3. json field for direct response

how's that sound? (json field isnt super necessary i just find it a lot nicer of a ux)

arkodg:
yeah sg, feel free to start with 1. in this PR, until we get a response for 2.
the body content can be put under body format if a formatter exists in the string which is similar UX with what we have in header modifier where a formatter can be added to a `value` field directly

rudrakhp:
@arkodg yes I was planning to do it but @ryanhristovski can continue here. One thing I wanted to mention though - response headers to add is relevant to redirect response as well. Can we move it to the top level response override API instead? You can choose to implement it only for direct response and we can implement it for redirect later.

internal/xds/translator/custom_response.go
```diff
@@ -369,10 +369,12 @@ func (c *customResponse) buildStatusCodeCELMatcher(codeRange ir.StatusCodeRange)
 
 func (c *customResponse) buildAction(r ir.ResponseOverrideRule) (*matcherv3.Matcher_OnMatch_Action, error) {
 	response := &policyv3.LocalResponsePolicy{}
+
 	if r.Response.Body != nil && *r.Response.Body != "" {
-		response.Body = &corev3.DataSource{
-			Specifier: &corev3.DataSource_InlineString{
-				InlineString: *r.Response.Body,
+		bodyContent := *r.Response.Body
```
arkodg: any reason this extra stack variable is needed ?

arkodg:
looks good, added a minor comment to reduce some mem

internal/xds/translator/custom_response.go
ryanhristovski: nope, removed

arkodg:
thanks @ryanhristovski, can we also update the doc string in https://github.com/envoyproxy/gateway/blob/4dfb4ef90779caa2a69023bf9ab5783865d29ea2/api/v1alpha1/shared_types.go#L734 to mention that string formatters mentioned in https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators are supported as well

arkodg (approved):
LGTM thanks !

Issue #6284, opened 2025-06-10
Errors during reconcile causes a creation of partial xDS configuraion


shahar-h:
*Description*:
After upgrading from v1.3.2 to v1.4.0 we started to see `filter_chain_not_found` and `cluster_not_found` errors in envoy access logs.
The errors disappeared after few hours.
The errors started with a specific xDS update and stopped after another update.
We discovered a single error log exactly on the time the errors started (filter chain not found):

"ERROR  provider  kubernetes/controller.go:237  failed processGateways for gatewayClass eg, skipping it  {"runner": "provider", "error": "failed to list \*\*\*, Kind=\*\*\*: etcdserver: leader changed"}"

We noticed #5953 which replaced all `return reconcile.Result{}, err` statements with `continue` statements in `Reconcile` function. 
This explains the behavior we observed since gateway resources are stored [here](https://github.com/envoyproxy/gateway/blob/598c813bd34ea7be45733e63802df73ab0cce6c1/internal/provider/kubernetes/controller.go#L360) regardless of errors during reconcile process which can lead to a storage of a partial state like in our case.

Furthermore, `reconcile.Result{}, nil` is always returned from `Reconcile` function regardless of any errors in the process, which means controller-runtime backoff mechanism will not be triggered. In case of a transient error(e.g. etcd error like in our case) we'll have to wait for some watch event so `Reconcile` will be called again.

*Repro steps*:
Produce some transient error in the middle of Reconcile process.

*Environment*:
EG v1.4.0

zirain:
cc @arkodg

guydc:
cc @arkodg . 
This was discussed in yesterday's community meeting. Main points were:
- Explore handling transient errors differently from other (validation, ...) errors, e.g. retry at least in case of transient errors
- Check if we can store configs only for successfully GWCs without wiping out existing config for errored GWCs

arkodg:
yeah, we can semi revert the logic, to return early with an err from the `Reconcile` if we get any Kube client errors not tied to bad config ( like a SP Ext proc config pointing to an invalid backend )

patrostkowski:
@arkodg @guydc would it be sufficient if controller checked err against set of https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors errors? like e.g.:
```
func isTransientError(err error) bool {
	return kerrors.IsServerTimeout(err) ||
		kerrors.IsTimeout(err) ||
		kerrors.IsTooManyRequests(err) ||
[...]

PR #6299, opened 2025-06-12
fix: retry reconcile on transient errors during reconcile 

patrostkowski:
**What type of PR is this?**

fix: #6284 

This PR:
* Introduces isTransientError to detect transient Kubernetes errors and enable proper reconciliation retries.
* Ignores non-transient errors from the provider layer and publishes them to the translation layer - the errors mostly come from missing referenced resources or misconfigurations that can be caught later in the Gateway API translator and surfaced in the resource status. 

Description updated by @zhaohuabing on Jul 3 2025.

Fixes https://github.com/envoyproxy/gateway/issues/6284

Release Notes: Yes

### DIFF
--- a/internal/provider/kubernetes/controller.go
+++ b/internal/provider/kubernetes/controller.go
@@ -230,6 +230,7 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 
 		// Process the parametersRef of the accepted GatewayClass.
 		// This should run before processGateways and processBackendRefs
+		failToProcessGCParamsRef := false
 		if managedGC.Spec.ParametersRef != nil && managedGC.DeletionTimestamp == nil {
 			if err := r.processGatewayClassParamsRef(ctx, managedGC, gwcResourceMapping, gwcResource); err != nil {
 				if isTransientError(err) {
@@ -245,6 +246,7 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 					string(gwapiv1.GatewayClassReasonInvalidParameters),
 					msg)
 				r.resources.GatewayClassStatuses.Store(utils.NamespacedName(gc), &gc.Status)
+				failToProcessGCParamsRef = true
 			}
 		}
 
@@ -262,9 +264,10 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 				string(gwapiv1.GatewayClassReasonAccepted),
 				fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err))
 			r.resources.GatewayClassStatuses.Store(utils.NamespacedName(gc), &gc.Status)
+			failToProcessGCParamsRef = true
 		}
 
-		if _, ok := r.resources.GatewayClassStatuses.Load(utils.NamespacedName(managedGC)); !ok {
+		if !failToProcessGCParamsRef {
 			// GatewayClass is valid so far, mark it as accepted.
 			gc := status.SetGatewayClassAccepted(
 				managedGC.DeepCopy(),


test/helm/gateway-addons-helm/e2e.out.yaml
patrostkowski, line 1: not sure if I should have this in my PR... it was generated after I did `make generate`

internal/provider/kubernetes/controller.go
```diff
@@ -221,6 +230,10 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 		// This should run before processGateways and processBackendRefs
 		if managedGC.Spec.ParametersRef != nil && managedGC.DeletionTimestamp == nil {
 			if err := r.processGatewayClassParamsRef(ctx, managedGC, resourceMappings, gwcResource); err != nil {
+				if isTransientError(err) {
```
rudrakhp, line 236: nit: move repeating block to defer?

test/helm/gateway-addons-helm/e2e.out.yaml
rudrakhp: yes `*.out.yaml` files are autogenerated

rudrakhp: `address: ${env:MY_POD_IP}:8888` is fishy though, try rebasing and running `make -k gen-check`

patrostkowski: @rudrakhp I tried to rebase against main, to me it looks like some file has not been pushed to remote after codegen? https://github.com/envoyproxy/gateway/issues/6303

guydc:
@arkodg, @patrostkowski 
My question here would be: In case of a non-transient error, we will still store inconsistent resources (partial? invalid?) that may lead to corruption of XDS. Is there a way to avoid this, e.g. by using previously-stored resources for the affected GWC if they exist? This would be more aligned and compatible with previous behavior, where ALL XDS is essentially not updated in case of ANY error in this area of provider?

test/helm/gateway-addons-helm/e2e.out.yaml
rudrakhp: Looks like it's due to older opentelemetry chart version, have you tried doing a `helm repo remove` or force delete local helm cache to ensure you get the latest charts?

guydc:
Notes from the community meeting:
- Provider layer should publish an accurate state-of-the-world. 
- If transient errors make it impossible to build an accurate snapshot, the reconciliation should be retried by returning early with an error
- If validation errors are encountered, state should be published, so that status can be updated and user notified about issues with resources. 
- If users want to implement alternative strategies for handling invalid configuration (e.g. pausing XDS updates), that should be implemented in other layers.

arkodg:
can we also log the transient error

arkodg:
@patrostkowski still working on this ? we'd like to get this into the patch release scheduled to be out next week

zhaohuabing:
It seems @patrostkowski is currently unavailable. I’ll take over from here since we need this in the upcoming patch.

patrostkowski:
thanks @zhaohuabing and sorry, I am very busy time atm with other things 😞

zhaohuabing:
> thanks @zhaohuabing and sorry, I am very busy time atm with other things 😞 

No worries. I'll take care of it.

internal/provider/kubernetes/controller.go
zhaohuabing: It can't be properly handled in a defer block.

internal/provider/kubernetes/controller.go
```diff
@@ -217,101 +226,170 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+				errs = append(errs, err)
 				r.log.Error(err, fmt.Sprintf("failed processBackends for gatewayClass %s, skipping it", managedGC.Name))
-				continue
 			}
 		}
 
 		// Add the referenced services, ServiceImports, and EndpointSlices in
 		// the collected BackendRefs to the resourceTree.
 		// BackendRefs are referred by various Route objects and the ExtAuth in SecurityPolicies.
+		// TODO: zhaohuabing handle errors during processing of BackendRefs
```
zhaohuabing: I believe we should handle errors from the API Gateway translation the same way - by skipping the IRs,  allowing the Envoy fleet to continue serving traffic using the last good state. This ensures consistent behavior of configuration error handling from the user’s perspective.

If this makes sense, we can address it in a follow-up PR.

WDYT ? @guydc @envoyproxy/gateway-maintainers

internal/provider/kubernetes/controller.go
```diff
@@ -325,32 +418,35 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+	            // Log the error and continue processing.
 				r.log.Error(err, fmt.Sprintf("failed adding finalizer to gatewayClass %s",
 					managedGC.Name))
-				continue
 			}
 		}
+
+		// Set HasErrors to true if there are any non-transient errors during processing of the Gateway and Envoy Gateway resources.
+		// This will be used to determine whether the xDS cache should be updated or not.
+		gwcResource.HasErrors = (len(errs) > 0)
```
zhaohuabing, line 449: Should we surface the non-transient errors from provider layer to the GatewayClass status?

internal/provider/kubernetes/controller.go
cnvergence: +1

internal/gatewayapi/resource/resource.go
```diff
@@ -33,6 +33,11 @@ type (
 // resources that the translators needs as inputs.
 // +k8s:deepcopy-gen=true
 type Resources struct {
+	// HasErrors indicates whether the resources contain errors that may affect xDS generation.
+	// For example, if some resources are missing due to errors in the Provider layer, EG should not update the xDS cache.
+	// This ensures the xDS cache is not updated with incomplete or erroneous data and the Envoy Proxy can continue to serve requests
+	// with the last known good configuration.
+	HasErrors bool `json:"hasErrors,omitempty" yaml:"hasErrors,omitempty"`
```
arkodg: im a -1 on this logic
the goal of the provider layer is to an accurate snapshot of the resources, the only reason it may be unable to do so is because of any faulty I/O, in that case we skip a `publish` . In the future we can improve this to skip public per GC, which is a bigger design change

zhaohuabing: This PR skips publishing per GC when non-retriable errors occur in the provider layers, and we may also need to handle errors from the API Gateway translation the same way. 

I agree that this is a big design decision - it's risky to cherry-pick it into a previous release. Raised  https://github.com/envoyproxy/gateway/issues/6448 to collect more feedback.

internal/provider/kubernetes/controller.go
```diff
@@ -180,6 +180,15 @@ func byNamespaceSelectorEnabled(eg *egv1a1.EnvoyGateway) bool {
...
 }
 
+func isTransientError(err error) bool {
+	return kerrors.IsServerTimeout(err) ||
+		kerrors.IsTimeout(err) ||
+		kerrors.IsTooManyRequests(err) ||
+		kerrors.IsServiceUnavailable(err) ||
+		kerrors.IsStoreReadError(err) ||
+		kerrors.IsUnexpectedServerError(err)
+}
```
shahar-h: * What about [IsInternalError](https://pkg.go.dev/k8s.io/apimachinery@v0.33.2/pkg/api/errors#IsInternalError)?
* Since all checked errors are either `429` or `5xx` maybe it will be better to check error status code instead?
However [IsUnexpectedServerError](https://github.com/kubernetes/apimachinery/blob/852f12619baa477c3169095bfa608adcc562c984/pkg/api/errors/errors.go#L760) still needs to be checked.

internal/provider/kubernetes/controller.go
```diff
@@ -243,21 +283,31 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
 
 		// Add all Gateways, their associated Routes, and referenced resources to the resourceTree
 		if err = r.processGateways(ctx, managedGC, gwcResourceMapping, gwcResource); err != nil {
+			if isTransientError(err) {
+				r.log.Error(err, "transient error processing gateways", "gatewayClass", managedGC.Name)
+				return reconcile.Result{}, err
+			}
 			r.log.Error(err, fmt.Sprintf("failed processGateways for gatewayClass %s, skipping it", managedGC.Name))
-			continue
```
zhaohuabing, line 247: If one of the resources fail, continue processing remaining resources and publish them as-is. This avoid skipping valid resources and publishing partial resources.

internal/provider/kubernetes/controller.go
```diff
@@ -233,6 +247,32 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+			}
+
+			r.log.Error(err, fmt.Sprintf("failed process TLS SecretRef for EnvoyProxy for gatewayClass %s, skipping it", managedGC.Name))
+			gc := status.SetGatewayClassAccepted(
+				managedGC.DeepCopy(),
+				false,
+				string(gwapiv1.GatewayClassReasonAccepted),
+				fmt.Sprintf("%s: %v", status.MsgGatewayClassInvalidParams, err))
+			r.resources.GatewayClassStatuses.Store(utils.NamespacedName(gc), &gc.Status)
+			continue
```
zhaohuabing: Skip the processing GatewayClass if failed to process its parameter - should we continue processing the resources for an invalid GatewayClass and publish them?

internal/provider/kubernetes/controller.go
zhaohuabing: Added IsInternalError.  Thanks for catching that!

> Since all checked errors are either 429 or 5xx maybe it will be better to check error status code instead?

I prefer using the current `IsxxxError` approach for its explicit error semantics

internal/provider/kubernetes/controller.go
guydc: What's the downside? Will any of these resources actually translate to Infra/XDS, or would they only be updated with a status that reflect problems (if they have any) ?

internal/provider/kubernetes/controller.go
```diff
@@ -243,21 +284,31 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
 
 		if r.ctpCRDExists {
 			// Add all ClientTrafficPolicies and their referenced resources to the resourceTree
 			if err = r.processClientTrafficPolicies(ctx, gwcResource, gwcResourceMapping); err != nil {
+				if isTransientError(err) {
+					r.log.Error(err, "transient error processing ClientTrafficPolicies", "gatewayClass", managedGC.Name)
+					return reconcile.Result{}, err
+				}
 				r.log.Error(err, fmt.Sprintf("failed processClientTrafficPolicies for gatewayClass %s, skipping it", managedGC.Name))
 				continue
```
arkodg: shouldnt this `continue` be removed as well ?

arkodg: all except the one associated with gatewayclass

internal/provider/kubernetes/controller.go
```diff
@@ -439,6 +520,9 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
 			service := new(corev1.Service)
 			err := r.client.Get(ctx, types.NamespacedName{Namespace: string(*backendRef.Namespace), Name: string(backendRef.Name)}, service)
 			if err != nil {
+				if isTransientError(err) {
```
arkodg, line 537: is going to hard to remember to add this for every client call

internal/provider/kubernetes/controller.go
zhaohuabing: >  Will any of these resources actually translate to Infra/XDS

Yes, all resources of this GatewayClass will be translated into Infra/XDS if we continue processing the failed GatewayClass, but the resulting configuration are incorrect - for example, the `EnvoyProxy` resource may not work as expected, including client cert, additional sidecars, custom bootstrap, etc.

> they only be updated with a status that reflect problems (if they have any) ?

The previous change took this approach, but as [this comment](https://github.com/envoyproxy/gateway/pull/6299#discussion_r2179142075) suggested, it's a big design decision and it may be risky to cherry-pick it into v1.4 branch. So I reverted it and track it at https://github.com/envoyproxy/gateway/issues/6448 . We can discuss this in the next EG meeting and do it in the main branch.

~This PR skips processing for the failed GatewayClass. As a result, the Envoy fleet associated with all Gateways under that GatewayClass will be deleted because [the Gateway API runner deleted infra IRs not in the published resources](https://github.com/envoyproxy/gateway/blob/3aee7a660f72db2a59c5b2eb76ac2c78da59cac4/internal/gatewayapi/runner/runner.go#L298-L304).~

~If we consider that to be a higher risk than deploying a potentially misconfigured Envoy fleet, we could alternatively choose to continue processing and publishing the resources for the failed GatewayClass. cc @guydc @arkodg~

Update: this PR continues processing failed GatewayClass and publish its resources - this may publish Infra IR and xDS IR with misconfiguration, but avoids taking down the whole Gateway with a single user error from an `EnvoyProxy` resource.

internal/provider/kubernetes/controller.go
zhaohuabing: Yeah, we can figure out a better approach to handle this later.

internal/provider/kubernetes/controller.go
```diff
@@ -233,90 +248,166 @@ func (r *gatewayAPIReconciler) Reconcile(ctx context.Context, _ reconcile.Reques
...
+		r.resources.GatewayClassStatuses.Store(utils.NamespacedName(gc), &gc.Status)
+
 		// it's safe here to append gwcResource to gwcResources
 		gwcResources = append(gwcResources, gwcResource)
 		// process global resources
 		// add the OIDC HMAC Secret to the resourceTree
-		r.processOIDCHMACSecret(ctx, gwcResource, gwcResourceMapping)
+		if err = r.processOIDCHMACSecret(ctx, gwcResource, gwcResourceMapping); err != nil {
+			if isTransientError(err) {
+				r.log.Error(err, "transient error processing OIDC HMAC Secret", "gatewayClass", managedGC.Name)
```
arkodg, line 286: missing return

arkodg (dismissed):
LGTM thanks

internal/gatewayapi/resource/resource.go
zhaohuabing: > In the future we can improve this to skip public per GC, which is a bigger design change

Oh, I see what you mean - we don't need a flag in the resource then, we just don't publish the resources for the failed GatewayClass.

arkodg (approved):

guydc (approved):

PR #6307, opened 2025-06-13
fix wrong enum on SubjectAltNameType

zirain:
Valid values for `SubjectAltNameType` as following:

```golang
// SubjectAltNameType is the type of the Subject Alternative Name.
// +kubebuilder:validation:Enum=Hostname;URI
type SubjectAltNameType string
```


zirain:
@arkodg need cherry-pick?  at least for v1.4?

arkodg:
yeah we need to, will add the label to it, thanks @zirain

PR #6308, opened 2025-06-13
Add responseHeadersToAdd functionality to httproutefilter and btp

ryanhristovski:
Adds response headers to direct response in btp and httproutefilter

Example BTP:
```
kind: BackendTrafficPolicy
spec:
  responseOverride:
    - match:
        statusCodes:
          - type: Value
            value: 429
      response:
        responseHeaderModifier:
          add:
            - name: X-Response-Header-Added
              value: "true"
          set:
            - name: X-Response-Header-Set
              value: "true"     
		...
```

Example HTTPRouteFilter:
```
kind: HTTPRouteFilter
spec:
  directResponse:
    responseHeaderModifier:
        add:
          - name: X-Response-Header-Added
            value: "true"
        set:
          - name: X-Response-Header-Set
            value: "true"     
    statusCode: 200
    body:
      ...
```

Todo in another PR:
- Response headers to redirects Signed-off-by: Ryan Hristovski <ryan.hristovski@docker.com>


api/v1alpha1/shared_types.go
```diff
@@ -741,6 +741,30 @@ type CustomResponse struct {
...
+	// ResponseHeadersToAdd defines headers to add to the response.
+	// This allows the response policy to append, add or override headers
+	// of the original response before it is sent to a downstream client.
+	//
+	// +optional
+	ResponseHeadersToAdd []ResponseHeaderToAdd `json:"responseHeadersToAdd,omitempty"`
+}
+
+// ResponseHeaderToAdd defines a header to add to the response.
+type ResponseHeaderToAdd struct {
```
rudrakhp: WDYT about just HeaderValue instead? Can be reused in RequestHeadersToAdd as well.

api/v1alpha1/shared_types.go
```diff
@@ -741,6 +741,30 @@ type CustomResponse struct {
...
+	// Value is the value of the header to add.
+	Value string `json:"value"`
+
+	// Append specifies the action to take when the header already exists.
+	// If true, the value will be appended to the existing header value.
+	// If false, the value will overwrite the existing header value.
+	// Default is false.
+	//
+	// +optional
+	Append *bool `json:"append,omitempty"`
```
rudrakhp: Can we have an optional enum `AppendAction` instead of this bool as exposed by the data plane [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-headervalueoption-headerappendaction).

api/v1alpha1/shared_types.go
```diff
@@ -741,6 +741,30 @@ type CustomResponse struct {
 	//
 	// +optional
 	StatusCode *int `json:"statusCode,omitempty"`
+
+	// ResponseHeadersToAdd defines headers to add to the response.
+	// This allows the response policy to append, add or override headers
+	// of the original response before it is sent to a downstream client.
+	//
+	// +optional
+	ResponseHeadersToAdd []ResponseHeaderToAdd `json:"responseHeadersToAdd,omitempty"`
```
rudrakhp: Can we move this to top level `ResponseOverride` since it will be relevant for `CustomRedirect` as well.

rudrakhp (changes requested):
Added some comments regarding API. Please rebase before addressing them. Thanks!

api/v1alpha1/shared_types.go
arkodg: also is `set` and `add` the only 2 options here, is `remove` valid too ?
can we make this more gateway api like https://gateway-api.sigs.k8s.io/reference/spec/#httpheaderfilter ?

rudrakhp: +1 on `remove` operation, I do see `response_headers_to_remove` in the data plane APIs so it is valid. We can choose to rename the `AppendAction` enum as `Action` instead to capture this usecase as well.

rudrakhp:
@ryanhristovski just a quick follow up - is this still being worked on?

api/v1alpha1/httproutefilter_types.go
```diff
@@ -72,6 +72,13 @@ type HTTPDirectResponseFilter struct {
 	// If unset, defaults to 200.
 	// +optional
 	StatusCode *int `json:"statusCode,omitempty"`
+
+	// ResponseHeadersToAdd defines headers to add to the response.
+	// This allows the response policy to append, add or override headers
+	// of the original response before it is sent to a downstream client.
+	//
+	// +optional
+	ResponseHeadersToAdd []ResponseHeaderToAdd `json:"responseHeadersToAdd,omitempty"`
```
zhaohuabing: Since this field is inside the `HTTPDirectResponseFilter` struct, should we just name it `Headers`?

api/v1alpha1/httproutefilter_types.go
```diff
@@ -72,6 +72,13 @@ type HTTPDirectResponseFilter struct {
 	// If unset, defaults to 200.
 	// +optional
 	StatusCode *int `json:"statusCode,omitempty"`
+
+	// ResponseHeadersToAdd defines headers to add to the response.
+	// This allows the response policy to append, add or override headers
+	// of the original response before it is sent to a downstream client.
```
zhaohuabing: I might be misunderstanding something, but there wouldn't be an "original response" if this is a direct response from Envoy, right?

rudrakhp: +1 it should be more like "final response". Whether it's "direct" or "redirect" will depend on whether this is used with redirect or direct response action.

ryanhristovski:
@rudrakhp yeah hoping to finish this up by EOW, sorry for the delay

ryanhristovski:
Because im using the existing ResponseHeadersModifier from upstream, the updated generated btp manifest includes `remove`, but unfortunately the logic is done in [LocalResponsePolicy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/custom_response/local_response_policy/v3/local_response_policy.proto) which doesnt support removal of headers, only adding. 

I've added a CEL as discussed w @arkodg to explicitly not allow it

ryanhristovski:
Tests are failing due to a broken link: [CVE-2021-25740]: https://nvd.nist.gov/vuln/detail/CVE-2021-25740
cc: @zhaohuabing

zhaohuabing:
> Tests are failing due to a broken link: [[CVE-2021-25740](https://github.com/advisories/GHSA-vw47-mr44-3jf9)]: https://nvd.nist.gov/vuln/detail/CVE-2021-25740 cc: @zhaohuabing

@ryanhristovski You can add `nvd.nist.gov` to [LINKINATOR_IGNORE](https://github.com/envoyproxy/gateway/blob/64f3576cffa892acdf3af396b51b29b7a28cd4cc/tools/make/docs.mk#L7) if it keeps failing.

zhaohuabing:
> 2025-07-11T13:52:39Z	DEBUG	controller-runtime.test-env	installing CRD	{"crd": "backendtrafficpolicies.gateway.envoyproxy.io"}
panic: Failed to start testenv: unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "backendtrafficpolicies.gateway.envoyproxy.io": CustomResourceDefinition.apiextensions.k8s.io "backendtrafficpolicies.gateway.envoyproxy.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[responseOverride].items.properties[response].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.048576x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)

This may solve it:

```
!(has(self.responseHeaderModifier) && 
  has(self.responseHeaderModifier.remove) && 
  size(self.responseHeaderModifier.remove) != 0)
```

internal/ir/xds.go
```diff
@@ -638,6 +639,13 @@ type CustomResponse struct {
 
 	// StatusCode will be used for the response's status code.
 	StatusCode *uint32 `json:"statusCode,omitempty"`
+
+	// ResponseHeaderModifier defines headers to add, set or remove from the response.
+	// This allows the response policy to append, add or override headers
+	// of the original response before it is sent to a downstream client.
+	//
+	// +optional
+	ResponseHeaderModifier *gwapiv1.HTTPHeaderFilter `json:"responseHeaderModifier,omitempty"`
```
arkodg: can we keep the logic and data types consistent b/w gateway, ir and xds with how they are on Route
https://github.com/envoyproxy/gateway/blob/55bd448cba1be22d8fcaecc3aba40b729c93e50d/internal/ir/xds.go#L775

tools/make/docs.mk
```diff
@@ -4,7 +4,7 @@ RELEASE_VERSIONS ?= $(foreach v,$(wildcard ${ROOT_DIR}/docs/*),$(notdir ${v}))
 #       find a way to remove github.com from ignore list
 # TODO: example.com is not a valid domain, we should remove it from ignore list
 # TODO: https://www.gnu.org/software/make became unstable, we should remove it from ignore list later
-LINKINATOR_IGNORE := "opentelemetry.io github.com jwt.io githubusercontent.com example.com github.io gnu.org _print canva.com sched.co sap.com httpbin.org nemlig.com verve.com"
+LINKINATOR_IGNORE := "opentelemetry.io github.com jwt.io githubusercontent.com example.com github.io gnu.org _print canva.com sched.co sap.com httpbin.org nemlig.com verve.com nvd.nist.gov gateway-api.sigs.k8s.io"
```
arkodg: were these failing ?

ryanhristovski: yeah, i'll probably revert though - i imagine it was a known issue and is intentionally skipped while a patch is made (?)
image

jgoodall:
Any updates on this? Anything we can do to help?

ryanhristovski:
@jgoodall our team has deprioritized this, but I can try to get it out for next patch release.

There's just a couple API comments here if you're interested in getting it across the line though! The logic is already ready

zhaohuabing:
> @jgoodall our team has deprioritized this, but I can try to get it out for next patch release.
> 
> There's just a couple API comments here if you're interested in getting it across the line though! The logic is already ready

Hi @ryanhristovski I'll take this forward as we need this feature. Really appreciate all the work you've put in!

api/v1alpha1/httproutefilter_types.go
```diff
@@ -72,6 +72,13 @@ type HTTPDirectResponseFilter struct {
 	// If unset, defaults to 200.
 	// +optional
 	StatusCode *int `json:"statusCode,omitempty"`
+
+	// ResponseHeaderModifier defines headers to add, set or remove from the response.
+	// This allows the response policy to append, add or override headers
+	// of the final response before it is sent to a downstream client.
+	//
+	// +optional
+	ResponseHeaderModifier *gwapiv1.HTTPHeaderFilter `json:"responseHeaderModifier,omitempty"`
```
zhaohuabing: Should we just call it `Header` or `Headers` instead of `ResponseHeaderModifier`? @envoyproxy/gateway-maintainers

api/v1alpha1/shared_types.go
```diff
@@ -766,6 +767,14 @@ type CustomResponse struct {
...
 	// +optional
 	StatusCode *int `json:"statusCode,omitempty"`
+
+	// ResponseHeaderModifier defines headers to add, set or remove from the response.
+	// This allows the response policy to append, add or override headers
+	// of the final response before it is sent to a downstream client.
+	// Note: Header removal is not supported for responseOverride.
+	//
+	// +optional
+	ResponseHeaderModifier *gwapiv1.HTTPHeaderFilter `json:"responseHeaderModifier,omitempty"`
```
zhaohuabing: Should we just call it Header instead of ResponseHeaderModifier?

jgoodall:
> > @jgoodall our team has deprioritized this, but I can try to get it out for next patch release.
> > There's just a couple API comments here if you're interested in getting it across the line though! The logic is already ready
> 
> Hi @ryanhristovski I'll take this forward as we need this feature. Really appreciate all the work you've put in!

Thanks @zhaohuabing ! Actually, for our use case, this PR has superseded: https://github.com/envoyproxy/gateway/pull/6851

api/v1alpha1/httproutefilter_types.go
rudrakhp: @zhaohuabing I think `ResponseHeaders` at least would make sense. We would want to make it clear that these are headers to be added to the response only. 
In case of redirect, it is possible to add request headers as well so it will help maintain a clear distinction between the two.

zhaohuabing: Header/Headers is clear enough in the context of DirectResponse, and it also stays  consistent with the existing Body field naming.

rudrakhp: Doesn't the data plane API have `body` and `response_headers_to_add`? I would lean towards having response in the name, but like you said in case of direct response, headers make sense in response scenario only, so either works for me.

zhaohuabing: I feel that the dataplane API name could be clearer in the EG direct response/override response API. Since both are already clearly related to responses, the additional response prefix seems a bit redundant.

arkodg: sure `responseOverride.response.responseHeaderModifier` is too verbose, `header` or `headerModifier` works

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -1105,6 +1105,24 @@ func buildResponseOverride(policy *egv1a1.BackendTrafficPolicy, resources *resou
...
+				}
+				for h := range rhm.Set {
+					response.AddResponseHeaders = append(response.AddResponseHeaders, ir.AddHeader{
+						Name:   string(rhm.Set[h].Name),
+						Append: false,
+						Value:  []string{rhm.Set[h].Value},
+					})
+				}
+			}
+
```
arkodg, line 1125: no remove ?

arkodg (approved):
LGTM thanks !

zirain (approved):

Issue #6190, opened 2025-05-26
Add support for route-level downstream metrics in Envoy Gateway


ctausendfreund:
*Description*:
Currently, Envoy Gateway provides downstream metrics (e.g., request count, response codes) at the listener or virtual host level. However, as far as I know there is no support for collecting metrics per route, such as request counts, latencies, or error rates segmented by route match (e.g., path or prefix).

This could be achieved by:
* Exposing route identifiers (e.g., route name, path, prefix) as tags/labels on downstream metrics.
* Providing optional configuration in the Gateway API to enable per-route metric collection.
* Ensuring that Prometheus-compatible metrics (e.g., envoy_http_downstream_rq_total and envoy_http_downstream_rq_time_sum) can be filtered or grouped by route.

Example desired outcome:
```
envoy_http_downstream_rq_total{route="/api/foo"} 123
envoy_http_downstream_rq_total{route="/api/bar"} 5
```

Why is this needed?
* Enables better debugging and performance monitoring per endpoint.
* Allows fine-grained alerting (e.g., high error rate on /api/payment only).

Currently, as far as I know, this level of detail is only achievable by manually inspecting logs or instrumenting the upstream applications.

arkodg:
@guydc this would be an additional opt in feature, similar to how we have enableVirtualCluster stats

guydc:
Right, I didn't notice that Envoy has a route stat_prefix. That would make it much simpler to implement!

If we make it possible to configure this in BTP, that would allow reasonable granularity without inflating stats for the entire system. 

+1.

Inode1:
Hello, we’d like to be able to expose metrics for each route. What do you think about adding a new field statPrefix to BTP, this field would be passed down to the corresponding Envoy route as [stat_prefix](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-route). This would help collect more detailed [per-route metrics](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router-vcluster-stats). 
```
spec:
  statPrefix: "my-awesome-route"
```
@arkodg @guydc WDY?

arkodg:
@Inode1 are you looking to specify absolute values or templated like https://github.com/envoyproxy/gateway/pull/5898, if templated, my suggestion would be to add this in the Envoy Proxy API similar to above PR, once its merged

Inode1:
[...]

PR #6310, opened 2025-06-13
feat: route stat name

Inode1:
**What type of PR is this?**

Add support for determining route stat names in BackendTrafficPolicy

Supported: xRoute
Supported Formatters: %ROUTE_KIND%, %ROUTE_NAME%, %ROUTE_NAMESPACE%, %ROUTE_RULE_NAME%, %ROUTE_RULE_NUMBER%

Fixes https://github.com/envoyproxy/gateway/issues/6190

Release Notes: Yes


arkodg:
seeing the API being added to BTP, is the use case to enable this only on some routes or all ? If all is the most common use case, , then lets start with a field in `EnvoyProxy` and later one in BTP

Inode1:
> seeing the API being added to BTP, is the use case to enable this only on some routes or all ? If all is the most common use case, , then lets start with a field in `EnvoyProxy` and later one in BTP

Hi, I thought we agreed to add stats only in BTP), i.e., only for a specific route using the template. I don't quite understand how the settings of EnvoyProxy and BTP will work together. It seems that in the [discussion](https://github.com/envoyproxy/gateway/issues/6190#issuecomment-2914431630) it was decided that this should be opt-in, as enabling it for all routes through EnvoyProxy could lead to a large number of metrics, which might affect the performance of Envoy (and it's unclear how to disable the option at the EnvoyProxy level for a specific route). Additionally, with BTP, you can specify more specific things, such as the name of the route (routeStatPrefix: name=/route), as well as more concrete business-related aspects, like tenant, for example.

arkodg:
okay, lets start with BTP first

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -115,6 +115,25 @@ type BackendTelemetry struct {
...
+	// RouteStatName defines the value of the Route stat_prefix, determining how the route stats are named.
+	// For more details, see envoy docs: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-route
+	// The supported operators for this pattern are:
+	// %ROUTE_NAME%: name of Gateway API xRoute resource
+	// %ROUTE_NAMESPACE%: namespace of Gateway API xRoute resource
+	// %ROUTE_KIND%: kind of Gateway API xRoute resource
+	// %ROUTE_RULE_NAME%: name of the Gateway API xRoute section
+	// %ROUTE_RULE_NUMBER%: sequence number of the Gateway API xRoute resource
+	// Example: %ROUTE_KIND%/%ROUTE_NAMESPACE%/%ROUTE_NAME%/rule/%ROUTE_RULE_NUMBER% => httproute/my-ns/my-route/rule/0
+	//
```
arkodg, line 144:
```suggestion
	// Disabled by default.
	//
```

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -115,6 +115,25 @@ type BackendTelemetry struct {
...
+	// Metrics defines metrics configuration for the backend or Route.
+	//
+	// +optional
+	Metrics *BackendMetrics `json:"metrics,omitempty"`
+}
+
+type BackendMetrics struct {
+	// RouteStatName defines the value of the Route stat_prefix, determining how the route stats are named.
+	// For more details, see envoy docs: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-route
+	// The supported operators for this pattern are:
```
arkodg, line 138: routes can be attached to multiple gateways/parentRefs which will need to be represented here in case the user doesnt want to aggregate them into 1, and prefers per parent per route metrics
https://gateway-api.sigs.k8s.io/reference/spec/#httproutespec

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -115,6 +115,25 @@ type BackendTelemetry struct {
...
+	// %ROUTE_NAME%: name of Gateway API xRoute resource
+	// %ROUTE_NAMESPACE%: namespace of Gateway API xRoute resource
+	// %ROUTE_KIND%: kind of Gateway API xRoute resource
+	// %ROUTE_RULE_NAME%: name of the Gateway API xRoute section
+	// %ROUTE_RULE_NUMBER%: sequence number of the Gateway API xRoute resource
+	// Example: %ROUTE_KIND%/%ROUTE_NAMESPACE%/%ROUTE_NAME%/rule/%ROUTE_RULE_NUMBER% => httproute/my-ns/my-route/rule/0
+	//
+	// +optional
+	RouteStatName string `json:"routeStatName,omitempty"`
```
arkodg: must be a ptr since its optional

internal/gatewayapi/testdata/gateway-with-two-listeners-with-tcproutes-with-sectionname.out.yaml
```diff
@@ -188,6 +188,10 @@ xdsIR:
             name: tcproute/default/tcproute-1/rule/-1/backend/0
             protocol: TCP
             weight: 1
+        metadata:
```
arkodg, line 216: its expensive to add this from a mem usage perspective, why this this needed

internal/ir/xds.go
```diff
@@ -1958,10 +1958,14 @@ type TCPRoute struct {
...
 	// Proxy Protocol Settings
 	ProxyProtocol *ProxyProtocol `json:"proxyProtocol,omitempty" yaml:"proxyProtocol,omitempty"`
+	// Metadata is used to enrich envoy route metadata with user and provider-specific information
+	Metadata *ResourceMetadata `json:"metadata,omitempty" yaml:"metadata,omitempty"`
 	// settings of upstream connection
 	BackendConnection *BackendConnection `json:"backendConnection,omitempty" yaml:"backendConnection,omitempty"`
 	// DNS is used to configure how DNS resolution is handled for the route
 	DNS *DNS `json:"dns,omitempty" yaml:"dns,omitempty"`
+	// Metrics defines the schema for metric stat configuration.
+	Metrics *egv1a1.BackendMetrics `json:"metrics,omitempty" yaml:"metrics,omitempty"`
```
arkodg: my suggestion would be to convert BackendMetrics into RouteStat name in the gateway api layer similar to what ClusterStatName is doing, add add that as a sub field here

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: Apply

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: Done, but can you explain why it's like that? controllergen automatically makes string fields with omitempty optional. I understand that we might want to distinguish between a field being unset and being set to an empty "" in order to apply some default, but that's not our case.

arkodg: We don't want to generate additional stats if it's unset, so making it a ptr makes it opt in

arkodg: Yeah "", may not always means empty, sometimes it could be a valid value

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: As far as I understand your point, I thought this [function](https://github.com/envoyproxy/gateway/blob/44d329fb74176d2f7adbd4f9b96a76b039b67a64/internal/gatewayapi/backendtrafficpolicy.go#L484) was supposed to set the stats for the route in the gateways.

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: Ok

internal/ir/xds.go
Inode1: I see your point, but Routes are processed before BTP is handled — see [this line](https://github.com/envoyproxy/gateway/blob/44d329fb74176d2f7adbd4f9b96a76b039b67a64/internal/gatewayapi/translator.go#L192). So at the Route processing stage, I can't create RouteStats, and instead I store all the necessary route information in metadata. I could generate RouteStats in [this place](https://github.com/envoyproxy/gateway/blob/44d329fb74176d2f7adbd4f9b96a76b039b67a64/internal/gatewayapi/backendtrafficpolicy.go#L484), but I still need additional route-related data like SectionName and RouteIndex, which are only available during Route processing.

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/testdata/httproute-with-multiple-gateways-from-different-ns.out.yaml should help understand how 1 route can attach to multiple gateways
and this introduces two uses cases for metrics granularity that is mentioned in https://github.com/envoyproxy/gateway/pull/6310#discussion_r2152979819

internal/gatewayapi/testdata/gateway-with-two-listeners-with-tcproutes-with-sectionname.out.yaml
Inode1: I explained below why the metadata is needed.

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: Do we want to check and validate this somehow?)

internal/ir/xds.go
arkodg: recommend using a `StatName` field in `HTTPRoute` to address this

Inode1: > recommend using a `StatName` field in `HTTPRoute` to address this

Sorry, I didn't understand you. Should StatName already contain the fully parsed and built stat_prefix field, or will it contain information about the route (SectionName, RuleIndex)? If it's the former, then we can do it that way, but we'll have to drop SectionName and RuleIndex—in that case, the metadata won't be needed either. Are you okay with that?

arkodg: you should have access to RULE_NAME and RULE_NUMBER during route processing
but yes, any values coming from the policy cannot be used (today, we can do better in the future with some post processing if really needed)

api/v1alpha1/backendtrafficpolicy_types.go
Inode1: > https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/testdata/httproute-with-multiple-gateways-from-different-ns.out.yaml should help understand how 1 route can attach to multiple gateways and this introduces two uses cases for metrics granularity that is mentioned in [#6310 (comment)](https://github.com/envoyproxy/gateway/pull/6310#discussion_r2152979819)

It seems I understand what this is about. In the function https://github.com/envoyproxy/gateway/blob/26f9b20a7d5f3060f841787e677c0ed127adc759/internal/gatewayapi/backendtrafficpolicy.go#L664, there is processing for the Gateway, but here we have the IR, and we need to somehow map it to a resource to get the template parameters. We could do this via "resources", but that would be expensive — we'd have to go through all resources for every route, which is O(n²). Maybe we should reconsider the metadata field for all routes?)

Inode1:
@arkodg PTAL

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -430,6 +431,16 @@ func (t *Translator) translateBackendTrafficPolicyForRouteWithMerge(
 		return fmt.Errorf("error merging policies: %w", err)
 	}
 
+	if policy.Spec.Telemetry != nil && policy.Spec.Telemetry.Metrics != nil {
```
arkodg: looks like this needs to be removed

internal/ir/xds.go
```diff
@@ -739,6 +739,8 @@ type HTTPClientTimeout struct {
 type HTTPRoute struct {
 	// Name of the HTTPRoute
 	Name string `json:"name" yaml:"name"`
+	// StatName is the name of the route used for statistics and metrics.
+	StatName string `json:"statName,omitempty" yaml:"statName,omitempty"`
```
arkodg: should be a ptr since its optional

api/v1alpha1/backendtrafficpolicy_types.go
arkodg: you should have access to it indirectly via the IR listener name
https://github.com/envoyproxy/gateway/blob/26f9b20a7d5f3060f841787e677c0ed127adc759/internal/gatewayapi/helpers.go#L377

arkodg: thinking out loud, it may help to start off with an API within EnvoyProxy which is easier to build out its not tied to policy translation

internal/ir/xds.go
Inode1: Done

internal/gatewayapi/backendtrafficpolicy.go
Inode1: Yup

internal/gatewayapi/testdata/gateway-with-two-listeners-with-tcproutes-with-sectionname.out.yaml
guydc: can we do it conditionally only if the feature is enabled? Maybe that also related to Arko's other comment about EnvoyProxy level feature as opposed to BTP.

jukie:
Hey @Inode1 are you still working on this?

Inode1:
> Hey @Inode1 are you still working on this?

As far as I understand, @arkodg doesn’t agree to pass the fields into the metadata and suggests making this configuration at the EnvoyProxy level.

arkodg:
hey @Inode1 do you have a use case for using `%ROUTE_RULE_NUMBER%` instead of `%ROUTE_SECTION_NAME%` which is already available in the IR today ? my preference would be to avoid adding `RuleIndex` to the IR unless we absolutely need it

Inode1:
> hey @Inode1 do you have a use case for using `%ROUTE_RULE_NUMBER%` instead of `%ROUTE_SECTION_NAME%` which is already available in the IR today ? my preference would be to avoid adding `RuleIndex` to the IR unless we absolutely need it

No, to be honest, I don't need formatters tags at all, just a custom stats name string).

Done, delete ROUTE_RULE_NUMBER

arkodg:
thanks @Inode1, will take a look at this PR next week, after v1.6 is out

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -955,6 +973,7 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(
 			setIfNil(&r.TCPKeepalive, tf.TCPKeepalive)
 			setIfNil(&r.Timeout, tf.Timeout)
 			setIfNil(&r.DNS, tf.DNS)
+			r.StatName = buildRouteStatName(routeStatName, r.Metadata)
```
arkodg: shouldnt this be setIfNil ?

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -1002,6 +1021,7 @@ func (t *Translator) translateBackendTrafficPolicyForGateway(
 				continue
 			}
 
+			r.StatName = buildRouteStatName(routeStatName, r.Metadata)
```
arkodg: shouldnt this use setIfNIl ?

internal/gatewayapi/backendtrafficpolicy.go
Inode1: r.StatName is a value, not a pointer

arkodg: the PR sets `setIfNil(&r.StatName, buildRouteStatName(routeStatName, r.Metadata))` in `applyTrafficFeatureToRoute`

Inode1: Yup, you are right. Done

arkodg (approved):
LGTM thanks

zirain:
@Inode1 can you fix the DCO?

PR #6327, opened 2025-06-17
feat: load certs fallback to first key

zirain:
Fallback to the first key when `ca.crt` is not found, similar to what we did in custom response body.


internal/gatewayapi/helpers.go
```diff
@@ -673,3 +673,37 @@ func getPreserveRouteOrder(envoyProxy *egv1a1.EnvoyProxy) bool {
...
 	return false
 }
+
+func getCaCertFromConfigMap(cm *corev1.ConfigMap) (string, bool) {
+	var data string
+	data, exits := cm.Data[caCertKey]
+	switch {
+	case exits:
+		return data, true
+	case len(cm.Data) == 1: // Fallback to the first key if response.body is not found
```
arkodg:
```suggestion
	case len(cm.Data) == 1: // Fallback to the first key if ca.crt key is not found
```

internal/gatewayapi/helpers.go
```diff
@@ -673,3 +673,37 @@ func getPreserveRouteOrder(envoyProxy *egv1a1.EnvoyProxy) bool {
...
+	}
+}
+
+func getCaCertFromSecret(s *corev1.Secret) ([]byte, bool) {
+	var data []byte
+	data, exits := s.Data[caCertKey]
+	switch {
+	case exits:
+		return data, true
+	case len(s.Data) == 1: // Fallback to the first key if response.body is not found
```
arkodg:
```suggestion
	case len(s.Data) == 1: // Fallback to the first key if ca.crt is not found
```

arkodg (approved):

PR #6330, opened 2025-06-17
api: remove enum validation on ALPNProtocol

zirain:
This will allow us to config that ALPNProtocol when trying to integrate other OSS project(e.g. Istio).


guydc:
If we don't have a validation, can we still keep a comment outlining the "typical" values `http/1.0;http/1.1;h2` that most users should choose from?

zirain:
> typical

it's there:
```console
        // ALPNProtocols supplies the list of ALPN protocols that should be
	// exposed by the listener or used by the proxy to connect to the backend.
	// Defaults:
	// 1. HTTPS Routes: h2 and http/1.1 are enabled in listener context.
	// 2. Other Routes: ALPN is disabled.
	// 3. Backends: proxy uses the appropriate ALPN options for the backend protocol.
	// When an empty list is provided, the ALPN TLS extension is disabled.
	// Typical Supported values are:
	// - http/1.0
	// - http/1.1
	// - h2
	//
	// +optional
```

Issue #6322, opened 2025-06-16
Add Console for Envoy Gateway Admin


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Similar to Envoy Proxy's Web UI for the Admin Server https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/admin#envoy-admin-web-ui,it would be great to add a similar Web Ui for the Envoy Gateway admin server
https://github.com/envoyproxy/gateway/blob/main/internal/admin/server.go to get information around
* pprof
* resources
* status
* stats

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Xunzhuo:
## Overview
![Image](https://github.com/user-attachments/assets/fbd97289-8aff-4faf-b308-5e170d4f6d76)

## Resources
![Image](https://github.com/user-attachments/assets/54860d28-1776-4e35-9ec2-3351712c868e)

## Stats
![Image](https://github.com/user-attachments/assets/9f0db2c3-b049-4c56-b623-41e33b7e592f)

## Profile
![Image](https://github.com/user-attachments/assets/5a34100a-2751-40a5-97a4-d4880266ed18)

Xunzhuo:
Add a egctl command: `egctl x dashboard envoy-gateway` as a follow-up

zirain:
@Xunzhuo we want this land this in v1.5, It can be improved through iterations.

Xunzhuo:
@zirain sure do

PR #6331, opened 2025-06-17
feat: add admin console support

Xunzhuo:
**What type of PR is this?**

Fixes https://github.com/envoyproxy/gateway/issues/6322

Release Notes: Yes/No


internal/admin/handlers.go
```diff
@@ -0,0 +1,573 @@
...
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		http.Error(w, fmt.Sprintf("Failed to encode response: %v", err), http.StatusInternalServerError)
+	}
+}
+
+// handleStats returns statistics information
+func handleStats(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	var memStats runtime.MemStats
```
arkodg: nit: can be a follow up, https://pkg.go.dev/runtime/metrics seems to expose more, which can be used to also export to prom

internal/admin/handlers.go
```diff
@@ -0,0 +1,573 @@
...
+			"go_version": runtime.Version(),
+		},
+	}
+
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		http.Error(w, fmt.Sprintf("Failed to encode response: %v", err), http.StatusInternalServerError)
+	}
+}
+
+// Helper functions to get resources
```
arkodg: can we read from watchable instead ?
it will be single Load, and we already have a schema for it https://github.com/envoyproxy/gateway/blob/6a9019506ee49e222e9e1a4f6cf713fb351e4ebe/internal/gatewayapi/resource/resource.go#L35

arkodg: this will also work for non k8s / file provider case

internal/admin/server.go
```diff
@@ -6,22 +6,30 @@
...
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/envoyproxy/gateway/internal/envoygateway/config"
 )
 
-func Init(cfg *config.Server) error {
+//go:embed webui.html logo.svg
```
arkodg: can we use the stacked EG logo instead :) 
https://github.com/cncf/artwork/blob/main/examples/graduated.md#envoy-gateway-logos

internal/admin/server.go
```diff
@@ -6,22 +6,30 @@
...
+var webuiFS embed.FS
+
+func Init(cfg *config.Server, k8sClient client.Client) error {
 	if cfg.EnvoyGateway.GetEnvoyGatewayAdmin().EnableDumpConfig {
 		spewConfig := spew.NewDefaultConfig()
 		spewConfig.DisableMethods = true
 		spewConfig.Dump(cfg)
 	}
 
+	// Set the k8s client for handlers
```
arkodg: can we rm k8s dependancy
https://github.com/envoyproxy/gateway/pull/6331/files#r2198490409

internal/admin/server.go
```diff
@@ -33,6 +41,23 @@ func start(cfg *config.Server) error {
...
 	adminLogger.Info("starting admin server", "address", address, "enablePprof", enablePprof)
 
+	// Serve the web UI
+	handlers.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/":
+			handleWebUI(w, r)
+		case "/logo.svg":
+			handleSVG(w, r, "logo.svg")
+		default:
```
arkodg: does this break other paths added via HandleFunc ?

internal/cmd/egctl/utils.go
```diff
@@ -79,3 +85,36 @@ func newK8sClient() (client.Client, error) {
...
+		signals := make(chan os.Signal, 1)
+		signal.Notify(signals, os.Interrupt)
+		defer signal.Stop(signals)
+		<-signals
+		fw.Stop()
+	}()
+}
+
+// openBrowser opens the given URL in the default browser
+func openBrowser(url string, writer io.Writer) {
```
arkodg: can we reuse from https://github.com/envoyproxy/gateway/blob/main/internal/cmd/egctl/dashboard_envoy.go

internal/envoygateway/config/config.go
```diff
@@ -29,26 +29,33 @@ const (
 // Server wraps the EnvoyGateway configuration and additional parameters
 // used by Envoy Gateway server.
 type Server struct {
+	// ServerConfiguration is the configuration used to startup Envoy Gateway.
+	ServerConfiguration
```
arkodg: imo this isnt needed, to display config we can show Server.EnvoyGateway

internal/admin/server.go
```diff
@@ -33,6 +41,23 @@ func start(cfg *config.Server) error {
...
+			http.NotFound(w, r)
+		}
+	})
+
+	// API endpoints
+	handlers.HandleFunc("/admin/api/status", handleStatus(cfg))
```
arkodg: thinking out loud
* should we rm `/admin/api/resources` ? and use `/resources` directly
* should we call it `/config_dump` instead of `/resources`
* should we express runtime memory metrics using `/memory`
* should we replace `/status` with `/server_info`
* should `/stats` emit CP stats https://gateway.envoyproxy.io/docs/tasks/observability/gateway-exported-metrics/

arkodg:
made some comments, really excited to see this land in v1.5

internal/cmd/egctl/utils.go
Xunzhuo: Actually, it is reusing that, but move it into utils, since two places need this func

internal/envoygateway/config/config.go
Xunzhuo: but we will miss two fields:
```
	// ControllerNamespace is the namespace that Envoy Gateway runs in.
	ControllerNamespace string
	// DNSDomain is the dns domain used by k8s services. Defaults to "cluster.local".
	DNSDomain string
```

Xunzhuo:
FireShot Capture 001 - Envoy Gateway Admin Console - Envoy Gateway Admin -  localhost


FireShot Capture 002 - Server Information - Envoy Gateway Admin -  localhost


FireShot Capture 003 - Configuration Dump - Envoy Gateway Admin -  localhost

FireShot Capture 005 - Statistics - Envoy Gateway Admin -  localhost


FireShot Capture 004 - Performance Profiling - Envoy Gateway Admin -  localhost

internal/admin/console/api.go
```diff
@@ -0,0 +1,561 @@
...
+// handleAPIMetrics handles requests for metrics using the Prometheus registry
+func (h *Handler) handleAPIMetrics(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Create a Prometheus handler using the controller-runtime registry
+	// This includes all Envoy Gateway control plane metrics
+	handler := createCombinedMetricsHandler()
```
arkodg, line 511: is this another metrics server in addition to the one created by the metrics runner ?

Xunzhuo: No, just redirecting from the existing regisry

arkodg (approved):
LGTM thanks !

Xunzhuo:
Follow up: user guide for admin console

arkodg (approved):

PR #6332, opened 2025-06-17
fix: fix bug in hostname overlap detection

### LINKED ISSUE #6321: Bug in logic detecting overlapping cert SANs
*Description*:

Some logic was added in #5777 to detect overlapping certificate SANs in listeners. I've noticed the following in the gateway logs:

```
The certificate SAN testing-api.foo.dev overlaps with the certificate SAN api.foo.dev in listener api.foo.dev. ALPN will default to HTTP/1.1 to prevent HTTP/2 connection coalescing, unless explicitly configured via ClientTrafficPolicy
```

This looks like a bug, since the hostname `testing-api.foo.dev` does not actually overlap `api.foo.dev`.

*Repro steps*:

1. Add a listener for `testing-api.foo.dev` and `api.foo.dev`. 
2. Use cert-manager to issue certs for them (note that a wildcard certificate should not be used)
3. Look in the logs to see a warning about overlapping SANs

*Environment*:

Envoy Gateway version 1.4.1

*Logs*:

```
{"type":"OverlappingTLSConfig","status":"True","observedGeneration":1,"lastTransitionTime":"2025-06-06T14:33:26Z","reason":"OverlappingCertificates","message":"The certificate SAN test-api.foo.dev overlaps with the certificate SAN api.foo.dev in listener api.foo.dev. ALPN will default to HTTP/1.1 to prevent HTTP/2 connection coalescing, unless explicitly configured via ClientTrafficPolicy"}
```

rudrakhp:
**What type of PR is this?**

fix: fix bug in hostname overlap detection

Fixes #6321

Release Notes: Yes/No


zirain (approved):

zirain:
need a release note for backport?

rudrakhp:
Added release note

zirain (approved):

Issue #4085, opened 2024-08-20
Support GEP-995


guydc:
*Description*:
https://github.com/kubernetes-sigs/gateway-api/pull/2985 

Wait for GW-API v1.2 

- Propagate route rule names to xds as route metadata
- Support attachment of policies to route rules

kkk777-7:
Hello @arkodg , @guydc 
Do you know if anyone is working on this?
If not, can I try to work on it?

arkodg:
got for it @kkk777-7 !

kkk777-7:
@arkodg 
I'm not sure about the scope in this PR should cover, so I'd like to hear your thoughts.

---

It seems that `targetRef` with a `sectionName` is recommended only for use with **Direct** policies.
>LocalPolicyTargetReferenceWithSectionName identifies an API object to apply a direct policy to.

https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1alpha2/policy_types.go#L84

From what I understand, policies like **BackendTrafficPolicy (BTP)** and **SecurityPolicy (SP)** are considered **Inherited** policies, which can target either Gateway or HTTPRoute.
- https://gateway.envoyproxy.io/contributions/design/backend-traffic-policy/#design-decisions
- https://gateway.envoyproxy.io/contributions/design/security-policy/#design-decisions
- https://gateway-api.sigs.k8s.io/geps/gep-2649

Therefore, I'm considering the following options for the scope of implementation — which do you think would be the most appropriate?

---

**Option 1: Support only the most specific resource scope**

- (Inherited) Gateway/HTTPRoute, attached to Gateway – **Not Supported**
- (Inherited) Gateway/HTTPRoute, attached to HTTPRoute – **Supported**
- (Direct) Gateway, attached to Gateway – **Supported**

※ In this option, the mergeType feature in BTP would not be supported.

This approach is simple and treats inherited policies as effectively direct policies by limiting them to the most specific target.
However, it would not allow applying policies at the Gateway listener level for BTP/SP.

---

**Option 2: Full support**

Policies would be applied in the following order of precedence:

**xRoute sectionName > xRoute > Gateway sectionName > Gateway**

If `sectionName` is specified on a Gateway, the policy is applied only to the `xRoutes` attached to the corresponding listener.

This approach is more complex and diverges from the standard conventions of the Gateway API.
However, users can specify sectionName on all resources.

kkk777-7:
[...]

PR #6335, opened 2025-06-17
feat: support route rule in SecurityPolicy target

kkk777-7:
It allows to target only a specific route rule in a Security Policy.

According to the Gateway API specification: https://gateway-api.sigs.k8s.io/geps/gep-2648/#section-names

- If the target resource does not have the section name specified by the policy, the policy attachment will fail.
- If multiple policies are attached to the same section name, they are processed according to the standard conflict resolution rules.
- When multiple Policies of the same type target the same object, one with a sectionName and one without, the more specific policy will have its entire spec applied to the named section. The less specific policy will have its spec applied to all other sections of the target that are not targeted by any other more specific policies.

### overriding/merging strategy
overrides are applied in the following order of precedence:
**Route Rule > xRoute > Listener > Gateway**
(Due to the third rule of the Gateway API specification above, the combination of Route Rule/xRoute and Listener/Gateway can also be interpreted as a merge.)

## Related Issue
#4085
remain issue open as all policies have not been addressed.


zhaohuabing:
Hi @kkk777-7 thanks for working on this - this PR looks great!

Could you please also add an e2e test to verify the route rule level SecurityPolicy? (you could update an existing SP test, such as basic auth or API key auth, to set the SP at route rule level)

zhaohuabing (dismissed):

kkk777-7:
@zhaohuabing 
thanks for review!
I've added e2e test, please check when you have a moment.

internal/gatewayapi/securitypolicy.go
```diff
@@ -88,107 +88,47 @@ func (t *Translator) ProcessSecurityPolicies(securityPolicies []*egv1a1.Security
 	handledPolicies := make(map[types.NamespacedName]*egv1a1.SecurityPolicy)
 
 	// Translate
-	// 1. First translate Policies targeting xRoutes
-	// 2. Then translate Policies targeting Listeners
-	// 3. Finally, the policies targeting Gateways
+	// 1. First translate Policies targeting RouteRules
```
arkodg, line 91: if we dont support 4 levels of replacement for other policies like BTP and EEP, can we raise a GH issue to track that work ?

arkodg:
thanks @kkk777-7, would be great if we can have 1 testdata showing 4 level replace

internal/gatewayapi/securitypolicy.go
zhaohuabing: @arkodg  👉 https://github.com/envoyproxy/gateway/issues/4085

PR #6337, opened 2025-06-17
feat: support user provided name for generated HPA and PDB

shahar-h:
Add support for configuring user provided name for generated HPA and PDB resources, similar to Deployment, DaemonSet and Service.

Release Notes: Yes


arkodg (approved):
LGTM thanks !

zirain:
do we still need this as we support gateway namespace mode?

arkodg:
> do we still need this as we support gateway namespace mode?

Yes for default case ( non GNM case )

zirain (approved):

zirain:
it would be better to update the `EnvoyProxyCustomName` test to make sure this worked as expected.

Issue #4762, opened 2024-11-22
Support custom backendRefs via extensions


muwaqar:
*Description*:
Currently Envoy Gateway implements support for the following [BackendReference](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendObjectReference).
- `Service` (support: [Core](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendObjectReference))
- `ServiceImport` (support: [Experimental](https://gateway-api.sigs.k8s.io/geps/gep-1748/#serviceimport-as-a-backend))
- `Backend` (support: [Envoy Gateway custom](https://gateway.envoyproxy.io/latest/tasks/traffic/backend/))

However, vendors may need to support specific backends (S3, EC2, Lambda, ...) which EG does not support natively. In that case, the ask is to enhance the [Extensions](https://gateway.envoyproxy.io/v0.4/design/extending-envoy-gateway/) sub-system to invoke a custom extension which can program the filter chain, route and upstream cluster for the configured backend correctly.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

I could find some previous discussion in #4423, specifically the comment at https://github.com/envoyproxy/gateway/issues/4423#issuecomment-2412658619

Envoy Gateway [extensions doc](https://gateway.envoyproxy.io/v0.4/design/extending-envoy-gateway/) also mentions support for custom `backendRefs` is expected in future, but I could not find an issue tracking this, hence raising this.

> Allow Extension Developers to introduce their own custom resources for extending the Gateway-API via [ExtensionRefs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.LocalObjectReference), [policyAttachments](https://gateway-api.sigs.k8s.io/references/policy-attachment/?h=policy) (future) and [backendRefs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendObjectReference) (future).

arkodg:
+1 @muwaqar, @guydc and @zhaohuabing have already +1'd this idea in #4423 
would you be interested in driving this work ?

arkodg:
[...]

PR #6342, opened 2025-06-18
feat: support custom backendRef via extension server

Xunzhuo:

Fixes https://github.com/envoyproxy/gateway/issues/4762

Release Notes: Yes

This PR added support for custom backend, it is used widely in vendor specific Backend. Also it is used in Envoy AI Gateway:

```
apiVersion: inference.networking.x-k8s.io/v1alpha2
kind: InferencePool
metadata:
  name: vllm-llama3-8b-instruct
spec:
  targetPortNumber: 8000
  selector:
    app: vllm-llama3-8b-instruct
  extensionRef:
    name: vllm-llama3-8b-instruct-epp
---
kind: HTTPRoute
metadata:
  name: backend
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - backendRefs:
        - group: inference.networking.x-k8s.io/v1alpha2
          kind: InferencePool
          name: vllm-llama3-8b-instruct
      matches:
        - path:
            type: PathPrefix
            value: /
```


internal/gatewayapi/route.go
```diff
@@ -1378,21 +1383,22 @@ func (t *Translator) processDestination(name string, backendRefContext BackendRe
...
-			return nil, err
-		}
-	}
-
 	// Skip processing backends with 0 weight
 	if weight == 0 {
 		return nil, nil
 	}
 
+	backendNamespace := NamespaceDerefOr(backendRef.Namespace, route.GetNamespace())
```
arkodg, line 1389: why was this moved, this makes an invalid backend with weight 0 be considered as valid and will probably not publish status for it

arkodg: recommend instead keeping the code as is, and adding a switch based on `t.isCustomBackendResource`

internal/xds/translator/translator.go
```diff
@@ -563,10 +582,6 @@ func (t *Translator) addRouteToRouteConfig(
 					}
 				}
 			}
-
```
arkodg, line 566: why is this removed ?

internal/ir/xds.go
```diff
@@ -779,6 +779,8 @@ type HTTPRoute struct {
 	CredentialInjection *CredentialInjection `json:"credentialInjection,omitempty" yaml:"credentialInjection,omitempty"`
 	// ExtensionRefs holds unstructured resources that were introduced by an extension and used on the HTTPRoute as extensionRef filters
 	ExtensionRefs []*UnstructuredRef `json:"extensionRefs,omitempty" yaml:"extensionRefs,omitempty"`
+	// BackendExtensionRefs holds unstructured resources that were introduced by an extension and used on the backendRef as a dynamic backend
+	BackendExtensionRefs []*UnstructuredRef `json:"backendExtensionRefs,omitempty" yaml:"backendExtensionRefs,omitempty"`
```
arkodg: curious why is this and UnstructuredRef are needed ?

internal/provider/kubernetes/controller.go
```diff
@@ -516,6 +523,31 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
...
+					},
+					GroupKind: schema.GroupKind{
+						Group: string(*backendRef.Group),
+						Kind:  backendRefKind,
+					},
+				}
+				if !resourceMappings.allAssociatedBackendRefExtensionFilters.Has(key) {
+					resourceMappings.allAssociatedBackendRefExtensionFilters.Insert(key)
+					if extRefFilter, exists := resourceMappings.extensionRefFilters[key]; !exists {
+						resourceMappings.extensionRefFilters[key] = extRefFilter
```
arkodg, line 661: curious why extRefFilter is needed here

arkodg: oh nvm, these are filters associated with these backendRefs, nice, you went straight for bonus points :)

internal/xds/translator/translator.go
```diff
@@ -532,9 +532,28 @@ func (t *Translator) addRouteToRouteConfig(
 			}
 
 			var err error
+			var clusters []*clusterv3.Cluster
+			switch {
+			case len(httpRoute.BackendExtensionRefs) > 0:
+				clusters, err = processExtensionBackendPostHook(xdsRoute, vHost, httpRoute, t.ExtensionManager)
```
arkodg: we cannot call the hook twice, server may not garuntee idempotency + 2 calls will increase latency / route convergence

we'll have to reuse L503

internal/xds/translator/translator.go
```diff
@@ -532,9 +532,28 @@ func (t *Translator) addRouteToRouteConfig(
...
+				if err != nil {
+					// If the extension server returns an error, and the extension server is not configured to fail open,
+					// then propagate the error
+					if !(*t.ExtensionManager).FailOpen() {
```
arkodg: null ptr deref ?

proto/extension/service.proto
```diff
@@ -59,6 +59,7 @@ message PostRouteModifyRequest {
 // If an extension returns a nil Route then it will not be modified
 message PostRouteModifyResponse {
     envoy.config.route.v3.Route route = 1;
+    repeated envoy.config.cluster.v3.Cluster clusters = 2;
```
arkodg: we can support EDS Clusters in the future, by introducing the ability to add CLA, lets track this with a GH issue

guydc:
Question: Do we prefer to have EG generate the cluster and the extension server mutate it?

This way, we have better control over naming of clusters (impacting metrics, collisions), linking them to routes and allowing to leverage from BTP, BTLSP etc. that attach to the route.

internal/ir/xds.go
Xunzhuo: removed

internal/provider/kubernetes/controller.go
Xunzhuo: Yep

internal/xds/translator/translator.go
Xunzhuo: Changed

examples/extension-server/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	k8s.io/apimachinery v0.34.0-alpha.0
 	sigs.k8s.io/controller-runtime v0.21.0
 	sigs.k8s.io/gateway-api v1.3.1-0.20250527223622-54df0a899c1c
+	sigs.k8s.io/gateway-api-inference-extension v0.3.0
```
mathetake: nit: v0.4.0 is out

proto/extension/service.proto
```diff
@@ -39,6 +39,12 @@ service EnvoyGatewayExtension {
     // in order to not make any changes to it.
     rpc PostHTTPListenerModify(PostHTTPListenerModifyRequest) returns (PostHTTPListenerModifyResponse) {};
 
+	// PostClusterModify provides a way for extensions to modify clusters generated by Envoy Gateway for custom backends.
```
guydc, line 42: nit: indentation looks a bit off

proto/extension/context.proto
```diff
@@ -29,6 +29,14 @@ message PostVirtualHostExtensionContext {
 }
 
 
+// PostClusterExtensionContext provides context information for cluster modification
+// additional context information can be added to this message as more use-cases are discovered
+message PostClusterExtensionContext {
+    // Resources introduced by the extension that were used as custom backendRefs
+    repeated ExtensionResource extension_resources = 1;
```
guydc: I think that in the future, we may have a use case where clusters are also modified because of HTTPRoute Backend ExtensionFilterRefs.

It's also possible that we would want to provide all the context at once: both custom backendsRefs and custom HTTPRotueFilterRefs. 

So, maybe for future-proofing we can call it `backend_extension_resources` ?

internal/xds/translator/translator.go
```diff
@@ -897,6 +837,162 @@ func findXdsEndpoint(tCtx *types.ResourceVersionTable, name string) *endpointv3.
...
+	if existingCluster := findXdsCluster(tCtx, modifiedCluster.Name); existingCluster == nil {
+		if err := tCtx.AddXdsResource(resourcev3.ClusterType, modifiedCluster); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// buildClusterForHTTPRoute builds a cluster for HTTPRoute but doesn't add it to the resource table
+func (t *Translator) buildClusterForHTTPRoute(tCtx *types.ResourceVersionTable, name string, settings []*ir.DestinationSetting, route clusterArgs, extras *ExtraArgs, metadata *ir.ResourceMetadata) (*clusterv3.Cluster, error) {
```
guydc: Can we try to avoid the duplication with the existing `addXdsCluster` func, without introducing additional cluster lookups in the resource table for the purpose of mutation?

internal/provider/kubernetes/controller.go
```diff
@@ -68,6 +68,7 @@ type gatewayAPIReconciler struct {
 	resources            *message.ProviderResources
 	extGVKs              []schema.GroupVersionKind
 	extServerPolicies    []schema.GroupVersionKind
+	extBackendGVKs       []schema.GroupVersionKind
```
guydc, line 71: In general, do we need to make some changes in predicates and indexers to ensure that a change in the custom backend would trigger a reconcile?

internal/extension/types/hooks.go
```diff
@@ -26,6 +26,12 @@ type XDSHookClient interface {
 	// that uses extension resources as externalRef filters.
 	PostRouteModifyHook(route *route.Route, routeHostnames []string, extensionResources []*unstructured.Unstructured) (*route.Route, error)
 
+	// PostClusterModifyHook provides a way for extensions to modify a cluster generated by Envoy Gateway for custom backends.
+	// This allows extensions to modify cluster configurations for custom backend types while letting Envoy Gateway
+	// control cluster naming and basic configuration. This hook is called when custom backend resources are used
+	// in HTTPRoute or GRPCRoute backendRefs. The hook is called for each individual cluster after it's built.
+	PostClusterModifyHook(cluster *cluster.Cluster, extensionResources []*unstructured.Unstructured) (*cluster.Cluster, error)
```
mathetake, line 37: maybe also clarify what's the difference between modifying cluster here vs PostTranslateModify and who will be called first etc.

proto/extension/context.proto
Xunzhuo: Good point : )

proto/extension/service.proto
Xunzhuo: Nope, i might be show incorrectly in github page, the indent in file is correct

internal/xds/translator/translator.go
Xunzhuo: Done

proto/extension/context.proto
Xunzhuo: Done

examples/extension-server/go.mod
Xunzhuo: Done

internal/extension/types/hooks.go
Xunzhuo: Done

internal/provider/kubernetes/controller.go
Xunzhuo: extBackendGVKs currently follows the same implmentation logics in provider with extServerPolicies and extGVKs

internal/gatewayapi/route.go
```diff
@@ -252,7 +255,11 @@ func (t *Translator) processHTTPRouteRules(httpRoute *HTTPRouteContext, parentRe
 
 		// process each ir route
 		for _, irRoute := range ruleRoutes {
+			if len(backendCustomRefs) > 0 {
+				irRoute.ExtensionRefs = append(irRoute.ExtensionRefs, backendCustomRefs...)
+			}
 			destination := &ir.RouteDestination{
+				Name:     destName,
```
arkodg: curious why its needed
its set below

```
		default:
				destination.Name = destName
				destination.Settings = allDs
				irRoute.Destination = destination
```

internal/gatewayapi/route.go
arkodg: still seeing the same issue here

internal/gatewayapi/route.go
Xunzhuo: just want to make sure the destination to have a name

internal/gatewayapi/route.go
Xunzhuo: fixed

internal/xds/translator/translator.go
```diff
@@ -897,6 +837,169 @@ func findXdsEndpoint(tCtx *types.ResourceVersionTable, name string) *endpointv3.
...
+}
+
+// buildAndAddCluster builds a cluster for HTTPRoute, applies extension hooks, and adds it to the resource table
+func (t *Translator) buildAndAddCluster(tCtx *types.ResourceVersionTable, name string, settings []*ir.DestinationSetting, route clusterArgs, extras *ExtraArgs, metadata *ir.ResourceMetadata, extensionResources []*unstructured.Unstructured) error {
+	cluster, err := t.buildClusterForHTTPRoute(tCtx, name, settings, route, extras, metadata)
+	if err != nil || cluster == nil {
+		return err
+	}
+
+	// Apply extension hook to modify the cluster if needed
```
arkodg: can we add this under if t.ExtensionManager

Xunzhuo: Fixed it in the latest commit

arkodg (dismissed):
LGTM thanks !

mathetake (approved):

internal/xds/translator/translator.go
```diff
@@ -564,9 +577,6 @@ func (t *Translator) addRouteToRouteConfig(
 				}
 			}
 
-			if err != nil {
```
arkodg, line 567: can we add this back

Xunzhuo: nope, this is duplicated err append, plz see the above lines, err has append already.

Xunzhuo: One error will append twice, it is a hidden bug.

arkodg: ha thanks

arkodg (approved):

guydc (approved):
LGTM, thanks!

In general, we need to prioritize investment in E2Es for the extensions server. A lot of users and projects are increasingly relying on this feature which is currently not E2E tested.

arkodg (approved):

arkodg (approved):

PR #6343, opened 2025-06-18
feat: add validation of section name for Gateway listener

kkk777-7:
**What type of PR is this?**

Folllow-up to #5916 
Added validation that the section name exists in the gateway's listeners.

Fixes #

Release Notes: No

### DIFF
--- a/internal/gatewayapi/securitypolicy.go
+++ b/internal/gatewayapi/securitypolicy.go
@@ -400,21 +400,12 @@ func resolveSecurityPolicyGatewayTargetRef(
 
 	// If sectionName is set, make sure its valid
 	if target.SectionName != nil {
-		found := false
-		for _, l := range gateway.listeners {
-			if l.Name == *target.SectionName {
-				found = true
-				break
-			}
-		}
-		if !found {
-			message := fmt.Sprintf("No section name %s found for Gateway %s",
-				string(*target.SectionName), key.String())
-
-			return gateway.GatewayContext, &status.PolicyResolveError{
-				Reason:  gwapiv1a2.PolicyReasonTargetNotFound,
-				Message: message,
-			}
+		if err := validateGatewayListenerSectionName(
+			*target.SectionName,
+			key,
+			gateway.listeners,
+		); err != nil {
+			return gateway.GatewayContext, err
 		}
 	}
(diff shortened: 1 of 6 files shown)


internal/gatewayapi/securitypolicy.go
```diff
@@ -398,6 +398,26 @@ func resolveSecurityPolicyGatewayTargetRef(
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			message := fmt.Sprintf("No section name %s found for Gateway %s",
+				string(*target.SectionName), key.String())
+
+			return gateway.GatewayContext, &status.PolicyResolveError{
+				Reason:  gwapiv1a2.PolicyReasonInvalid,
```
arkodg: nice !
can we use `PolicyReasonTargetNotFound` instead ?

arkodg: https://github.com/kubernetes-sigs/gateway-api/blob/b4794ff3cfee5e49910aa1ba55eb1b8336695ce9/apis/v1alpha2/policy_types.go#L141

kkk777-7: thanks for comment! I've updated by [5724cc2](https://github.com/envoyproxy/gateway/pull/6343/commits/5724cc2453a1cf16dc167408ea5292dd5e491fe7).
(Also, ClientTrafficPolicy in the same case)

internal/gatewayapi/securitypolicy.go
```diff
@@ -398,6 +398,26 @@ func resolveSecurityPolicyGatewayTargetRef(
 		return nil, nil
 	}
 
+	// If sectionName is set, make sure its valid
+	if target.SectionName != nil {
```
shawnh2, line 402: this section name checking logic is same as in client-traffic-policy, is there any chance we can wrap them up in a common function ?

kkk777-7: thanks for comment. refactored by https://github.com/envoyproxy/gateway/pull/6343/commits/2212d947e3c911b69a35b15396afe1f89fab2a2e

arkodg (approved):
LGTM thanks !

PR #6347, opened 2025-06-18
default on SameSite attribute unset for Oauth2 cookies

arkodg:
Revisits https://github.com/envoyproxy/gateway/pull/6289 which had set SameSite=Strict. This may cause some issues for specific flows

I had misinterpresented the meaning of `Disabled` earlier, it means `Unset`, and is separate from `Samesite=None`


arkodg:
cc @vibe

zirain (dismissed):

vibe:
Totally reasonable to default back to `disabled` / unset. 

Thanks @arkodg

zirain (approved):

shawnh2 (approved):

PR #6354, opened 2025-06-19
chore(installation): early validate presence of tag in `image` and `imageRepository`

### LINKED ISSUE #6350: Validate proxy image and repository
*Description*:

Validate proxy image and imageRepository.

[optional *Relevant Links*:]

https://github.com/envoyproxy/gateway/pull/6296#discussion_r2155815222

> Should we have any validation in place to ensure user is not passing a tag in imageRepository?

sudiptob2:
PR [https://github.com/envoyproxy/gateway/pull/6296](https://github.com/envoyproxy/gateway/pull/6296) added support for customizing the **EnvoyProxy** image using the `image` and `imageRepository` fields. In summary:

- If users want to explicitly specify the image tag, they should use the format: `image: repo:tag`. This has been the standard approach in all versions up to this PR.
- If users prefer to define only the image repository and let Envoy Gateway decide the appropriate tag, they should use the format: `imageRepository: repo_url`.

To ensure these fields are used correctly, it's helpful to validate the presence or absence of a tag depending on which field is set.

~This PR improves the `resolveProxyImage` function by adding the following validations:~ (https://github.com/envoyproxy/gateway/pull/6354/commits/5a85f82009ca66ae3e6b3b85e855082862876bbb)
This PR adds following CEL validation:

- If `imageRepository` is set, it **must not** include a tag.
- If `image` is set, it **must** include a valid tag.

_Idea for more validation logic is appreciated._

Fixes #6350


arkodg:
can this be solved with CEL ? so the user hit this error during apply and won't have to find the error in logs

sudiptob2:
> can this be solved with CEL ? So the user hit this error during apply and won't have to find the error in logs

Updated with CEL validation

api/v1alpha1/shared_types.go
```diff
@@ -209,6 +209,8 @@ type KubernetesPodSpec struct {
 
 // KubernetesContainerSpec defines the desired state of the Kubernetes container resource.
 // +kubebuilder:validation:XValidation:rule="!has(self.image) || !has(self.imageRepository)",message="Either image or imageRepository can be set."
+// +kubebuilder:validation:XValidation:rule="has(self.image) ? self.image.matches('^[^:]+:[^:]+$') : true",message="Image must include a tag (e.g., 'image:tag')."
```
rudrakhp:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.image.matches('^[a-zA-Z0-9._/-]+:[a-zA-Z0-9._-]+$') : true",message="Image must include a tag and allowed characters only (e.g., 'image:tag')."
```
Added some suggestions to make regex stricter, let me know what you think! Also can we remove `has` checks and move them to the respective fields, so kube validations apply on non-null values of the fields only?

api/v1alpha1/shared_types.go
```diff
@@ -209,6 +209,8 @@ type KubernetesPodSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.imageRepository) ? !self.imageRepository.contains(':') : true",message="ImageRepository must not include a tag or any colons."
```
rudrakhp:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.imageRepository.matches('^[a-zA-Z0-9._/-]+$')",message="ImageRepository must include allowed characters only and not include a tag or any colons."
```

api/v1alpha1/shared_types.go
sudiptob2: Stricter regex sounds good!


> can we remove has checks and move them to the respective fields

Yeah, moving validation to the respective field will be a cleaner approach, but we wont be able to provide a custom message. But a generic message would be fine, I guess.

sudiptob2: Well, the generic message does not look very good. Let me know your thougts

```
EnvoyProxy.gateway.envoyproxy.io "proxy-1750699777609469000" is invalid: spec.provider.kubernetes.envoyDeployment.container.image: Invalid value: "envoyproxy/envoy": spec.provider.kubernetes.envoyDeployment.container.image in body should match '^[a-zA-Z0-9._/-]+:[a-zA-Z0-9._-]+$'
```

rudrakhp: @sudiptob2 can we do something like this?
https://github.com/envoyproxy/gateway/blob/ddc9a5c169877f13e0ce6b0ece3fe820df83ee25/api/v1alpha1/shared_types.go#L339

sudiptob2: updated

shawnh2 (approved):

PR #6360, opened 2025-06-20
api: allow reuse exiting service account

zirain:
With this, users are able to use an existing sa for the proxy which contains a sidecar required additional role/permission.


api/v1alpha1/envoyproxy_types.go
```diff
@@ -373,6 +373,18 @@ type EnvoyProxyKubernetesProvider struct {
...
 	EnvoyPDB *KubernetesPodDisruptionBudgetSpec `json:"envoyPDB,omitempty"`
+
+	// EnvoyServiceAccount defines the desired state of the Envoy service account resource.
+	EnvoyServiceAccount *KubernetesServiceAccountSpec `json:"envoyServiceAccount,omitempty"`
+}
+
+type KubernetesServiceAccountSpec struct {
+	// Name is the name of the service account.
+	// If specified, the controller will use the provided service account
+	// instead of creating a new one.
```
arkodg:
```suggestion
type KubernetesServiceAccountSpec struct {
	// Name of the Service Account.
	// When unset, this defaults to an autogenerated name.
	//
```

PR #6361, opened 2025-06-20
e2e: simply the Datadog e2e test

zirain:
we should create the service/container port in the helm


rudrakhp (approved):

charts/gateway-addons-helm/values.yaml
```diff
@@ -814,6 +814,12 @@ opentelemetry-collector:
       hostPort: 9000
       protocol: TCP
       appProtocol: grpc
+    datadog:
+      enabled: true
```
arkodg, line 818: shouldnt this be opt in only for e2e?

zirain: the receiver is alway on.

arkodg: ah thanks

Issue #6348, opened 2025-06-18
Tracing doesn't work with UDS


zhaohuabing:
*Description*:

When deploying datadog agent as a sidecar and using UDS for communication between the datadog agent and envoy proxy, this error is reported by Envoy:

```
[2025-06-13 01:25:24.945][13][error][tracing] [source/extensions/tracers/datadog/logger.cc:37] Datadog [error 44]: Unexpected error submitting telemetry event: Failed to create request.
```

The reason is that `STRICT_DNS` is used for UDS endpoint:

arkodg:
https://github.com/envoyproxy/gateway/blob/e78e2684d7b57526344f5f420be3dac78cbb2c15/internal/xds/translator/tracing.go#L135 needs to be read from the IR

PR #6370, opened 2025-06-21
fix: use buildEndpointType for access and tracing

zirain:
fix: https://github.com/envoyproxy/gateway/issues/6348


internal/xds/translator/testdata/out/xds-ir/tracing-zipkin.clusters.yaml
```diff
@@ -22,22 +22,13 @@
...
-              portValue: 9411
-        loadBalancingWeight: 1
-      loadBalancingWeight: 1
-      locality:
-        region: tracing-0/backend/0
   name: tracing-0
   perConnectionBufferLimitBytes: 32768
-  respectDnsTtl: true
-  type: STRICT_DNS
+  type: EDS
```
zhaohuabing: Should this be `STRICT_DNS` for "zipkin.default.svc.cluster.local" ?

zirain: it's because IR is wrong.

zhaohuabing: Can we use the correct IR for xDS test input?

zirain: fixed.

zhaohuabing (approved):
LGTM Thanks!

internal/gatewayapi/testdata/envoyproxy-tracing-backend-uds.out.yaml
```diff
@@ -0,0 +1,282 @@
...
+      port: 19003
+    tracing:
+      destination:
+        metadata:
+          kind: EnvoyProxy
+          name: test
+          namespace: envoy-gateway-system
+        name: tracing
+        settings:
+        - addressType: IP
```
arkodg, line 210: should ideally be `UDS`
looks like this doesnt impact the final translation ?

zirain: valid values list as following:
```
// DestinationAddressType describes the address type state for a group of DestinationEndpoint
type DestinationAddressType string

const (
	IP    DestinationAddressType = "IP"
	FQDN  DestinationAddressType = "FQDN"
	MIXED DestinationAddressType = "Mixed"
)
```

arkodg: dont want to block this bug fix, can we track this with a  follow up ?

zirain: https://github.com/envoyproxy/gateway/issues/6408

arkodg (approved):

PR #6378, opened 2025-06-23
feat: Add Support for Setting Certgen Pod Labels/Annotations

### LINKED ISSUE #6377: CertGen Supports Pod Labels/Annotations
*Description*:

 I would like to be able to configure pod labels and annotations for Certgen job especially when using Envoy Gateway with Linkerd I want to be able to inject all the pods in the namespace but make it easy to use certgen by disabling injection from the job.

blakeromano:
**What type of PR is this?**

This PR adds support for configuring pod labels or annotations for Certgen job on the helm chart in order to make it easier to work with service mesh providers like Linkerd.

Fixes #6377

Release Notes: No


charts/gateway-helm/values.tmpl.yaml
```diff
@@ -113,6 +113,8 @@ certgen:
   job:
     annotations: {}
     args: []
+    podLabels: {}
```
arkodg: prefer
```
pod:
  labels: {}
  annotations: {}
```

blakeromano: Updated to this model instead

arkodg:
thanks can you also add tests in https://github.com/envoyproxy/gateway/tree/main/test/helm

arkodg (approved):
LGTM thanks

zirain (approved):

arkodg:
@blakeromano can you fix the gen issues by running `make gen-check` locally and noticing the differences ?
you can generate the diff using `make generate && make manifests`

zirain (approved):

arkodg (approved):

PR #6390, opened 2025-06-23
fix: create IPv4 Service with ipFamily IPv4 on dual-stack, IPv6-first cluster

### LINKED ISSUE #6389: Setting `ipFamily: IPv4` on a dual stack, IPv6-first cluster results in an IPv6 Service
*Description*:
Setting `ipFamily: IPv4` on a dual stack, IPv6-first cluster should result in an IPv4 Service. Instead `envoy-gateway` creates a single stack, IPv6 `Service`.

There's [a comment in the code](https://github.com/envoyproxy/gateway/blob/4140f616451faa1840cc4fc3f774ef1da46ca991/internal/infrastructure/kubernetes/proxy/resource_provider.go#L265):

```
	if ipFamily != nil {
		// SingleStack+IPv4 is default behavior from K8s and so is omitted
		switch *ipFamily {
```

This isn't true according to [the Kubernetes docs](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services):

> The address family of a Service defaults to the address family of the first service cluster IP range (configured via the --service-cluster-ip-range flag to the kube-apiserver).

Unfortunately [simply setting the family of the Service isn't enough](https://github.com/envoyproxy/gateway/pull/6390#issuecomment-2998040677). The health listener doesn't listen on IPv4 if the family is set on the EnvoyProxy but kubernetes tries the first IP of the Pod which is IPv6 in this case.

*Repro steps*:
Set up an EnvoyProxy with `spec.ipFamily: IPv4` on a dual-stack, IPv6-first cluster.

*Environment*:
Envoy gateway v1.4.1

*Logs*:
None relevant

michaelbeaumont:
**What type of PR is this?**
fix: create IPv4 Service with ipFamily IPv4 on dual-stack, IPv6-first cluster

The deleted comment isn't accurate. From [Kubernetes docs](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services):

> The address family of a Service defaults to the address family of the first service cluster IP range (configured via the --service-cluster-ip-range flag to the kube-apiserver).

Fixes #6389

Release Notes: Yes

### DIFF
--- a/internal/infrastructure/kubernetes/proxy/resource_provider.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider.go
@@ -251,8 +251,10 @@ func (r *ResourceRender) Service() (*corev1.Service, error) {
 	// Set IP family policy and families based on proxy config request
 	ipFamily := r.infra.GetProxyConfig().Spec.IPFamily
 	if ipFamily != nil {
-		// SingleStack+IPv4 is default behavior from K8s and so is omitted
 		switch *ipFamily {
+		case egv1a1.IPv4:
+			serviceSpec.IPFamilies = []corev1.IPFamily{corev1.IPv4Protocol}
+			serviceSpec.IPFamilyPolicy = ptr.To(corev1.IPFamilyPolicySingleStack)
 		case egv1a1.IPv6:
 			serviceSpec.IPFamilies = []corev1.IPFamily{corev1.IPv6Protocol}
 			serviceSpec.IPFamilyPolicy = ptr.To(corev1.IPFamilyPolicySingleStack)
--- a/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
+++ b/internal/infrastructure/kubernetes/proxy/resource_provider_test.go
@@ -1864,8 +1864,8 @@ func TestIPFamilyPresentInSpec(t *testing.T) {
 		{
 			"ipv4 specified",
 			ptr.To(egv1a1.IPv4),
-			nil,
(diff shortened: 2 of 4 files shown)


arkodg:
thanks @michaelbeaumont, can you confirm if the fix and the configuration (ipFamily: IPv4) works in a IPv6 first cluster i.e. can the kubelet communicate successfully with the envoy pod, and set Ready=True ?

michaelbeaumont:
Hmm, @arkodg no unfortunately. There's an issue with the health check not being accessible on IPv6 afaict. It seems like 

https://github.com/envoyproxy/gateway/blob/72c9065a5e691c0c383f12399da33972e05a4487/internal/gatewayapi/listener.go#L447-L452

needs to be tweaked in this case. Maybe it should always listen on all ip families allocated to the Pod, not just the ip family set in the EnvoyProxy

arkodg:
@michaelbeaumont that makes sense
@zirain sgty ?

michaelbeaumont:
For the record, https://github.com/kubernetes/kubernetes/issues/101324 makes this tricky to implement the obvious way

michaelbeaumont:
Unfortunately I'm not sure how to figure out whether the Pod will have an IPv6 address so that we can decide whether to listen on `::` or `0.0.0.0`.

michaelbeaumont:
Full disclosure I won't have time to work on this further

cnvergence:
I will be happy to take a look :)

Issue #5086, opened 2025-01-17
Use Redis Sentinel for Global Ratelimiting


nebed:
*Description*:

I am trying to use a redis sentinel cluster for global rate limiting. I see it is supported by envoyproxy here
[EnvoyProxy Ratelimiting](https://github.com/envoyproxy/ratelimit/tree/main?tab=readme-ov-file#redis-type)]
but when i try this for the redis config in envoy-gateway it doesn't work.

I give it a redis config like - "mymaster,envoy-redis.redis.cluster.local:26379"

and I get this error that it cannot find the host, which i guess is a parsing error
`time="2025-01-17T12:27:54Z" level=error msg="creating redis connection error : dial tcp: lookup mymaster,envoy-redis.redis.cluster.local: no such host"
panic: dial tcp: lookup mymaster,envoy-redis.redis.cluster.local: no such host
goroutine 1 [running]:
github.com/envoyproxy/ratelimit/src/redis.checkError(...)
	/ratelimit/src/redis/driver_impl.go:69
github.com/envoyproxy/ratelimit/src/redis.NewClientImpl({0x102f930, 0xc0001c85d0}, 0x0, {0x0, 0x0}, {0xc00004602a, 0x3}, {0xcd1d94, 0x6}, {0xc00004205a, ...}, ...)
	/ratelimit/src/redis/driver_impl.go:138 +0xae5
github.com/envoyproxy/ratelimit/src/redis.NewRateLimiterCacheImplFromSettings({0x0, {0xcc552f, 0x7}, 0x1f90, {0xcd2dea, 0x7}, 0x17b6, {0x0, 0x0}, {0xcd0d59, ...}, ...}, ...)
	/ratelimit/src/redis/cache_impl.go:25 +0x3a5
github.com/envoyproxy/ratelimit/src/service_cmd/runner.createLimiter({_, _}, {0x0, {0xcc552f, 0x7}, 0x1f90, {0xcd2dea, 0x7}, 0x17b6, {0x0, ...}, ...}, ...)
	/ratelimit/src/service_cmd/runner/runner.go:87 +0x454
github.com/envoyproxy/ratelimit/src/service_cmd/runner.(*Runner).Run(0xc0002bc508)
	/ratelimit/src/service_cmd/runner/runner.go:151 +0x565
main.main()
	/ratelimit/src/service_cmd/main.go:10 +0xbc`

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

[...]

PR #6394, opened 2025-06-24
fix: fix validation errors that occur when the redis url contains valid sentinel or cluster config

keithfz:
**What type of PR is this?**

Fixes validation errors that occur when the Redis URL contains "valid" sentinel or cluster configurations.

There are three modes of deployment for Redis that are also compatible with the rate limit service. They each have a slightly different format under the url configuration field.

- **Standalone**: A single hostname. The only current configuration considered validation by the validation code
- **Sentinel**: the name of the primary node followed by a comma separated list of Redis Sentinel nodes. The primary node can contain alphanumeric characters, `-`, `.`, and `_`.
- **Cluster**: comma separated list of all hostnames in the Redis cluster.

This splits the url string by commas, and then validates all returned items in the list. This way, Sentinel and Cluster configurations are also considered valid, and do not need [work-arounds that obfuscate the actual configuration](https://github.com/envoyproxy/gateway/issues/5086#issuecomment-2703917823).

Fixes https://github.com/envoyproxy/gateway/issues/5086

Release Notes: Yes


arkodg (dismissed):
LGTM thanks !

zirain (approved):
Thanks!

arkodg (approved):

PR #6395, opened 2025-06-24
bugfix: fix hpa/pdb deletion issue

zirain:
follow up: https://github.com/envoyproxy/gateway/pull/6337

- add e2e for custom HPA/PDB name.
- fix labels on HPA/PDB which cause deletePDB/HPA not working.


internal/infrastructure/kubernetes/infra_resource.go
```diff
@@ -324,8 +338,30 @@ func (i *Infra) createOrUpdateHPA(ctx context.Context, r ResourceRender) (err er
...
+				Namespace:     hpa.Namespace,
+				LabelSelector: r.LabelSelector(),
+			})
+			if deleteErr != nil {
+				i.logger.Error(deleteErr, "failed to delete all except HorizontalPodAutoscaler",
+					"name", r.Name(), "namespace", r.Namespace())
+			}
+		}
+
+		if err == nil {
```
arkodg: should probably be first above L342

zirain: it's fine, because we use another `deleteErr`.

arkodg: was mostly for timing/metrics perspective

zirain: make sense, done!

Issue #5905, opened 2025-05-02
bug: 500 on all routes when adding dependency on service that isn't deployed


artgon-openai:
*Description*:

What we've observed is when we add configuration for a service that is not yet deployed in k8s, envoy-gateway returns 500 on all routes for that service. We've observed this with both ext_proc (EnvoyExtensionPolicy) and standard HttpRoute backends. The reason this was surprising is because in the ext_proc case it was set to `failOpen: true`, and in the HttpRoute case the weight was set to 0.

Another observation is that it did not trigger our alerts for envoy 5xx metrics, we only saw the errors in the edge/clients.

*Repro steps*:

Create an EnvoyExtensionPolicy or HttpRoute for a service that does not exist.

*Environment*:

Envoy Gateway version 1.3.2

*Logs*:

The logs indicated a 500 with a direct response.

```
{'response_code': 500, 'response_code_detail':	'direct_response'}
```

arkodg:
hey can you go through https://gateway.envoyproxy.io/docs/troubleshooting/configuration/ which should tell you why this is happening, in the status of the related resource

arkodg:
you're most likely hitting the `BackendNotFound` reason, there's an upstream discussion here on why the spec says to return `500` for all the backendRefs for the rule
https://github.com/kubernetes-sigs/gateway-api/discussions/1764

artgon-openai:
That thread is specifically about weighting backends, which I agree is a bit trickier -- although if the weight is set to 0, it should be pretty obvious what the user intent is. 

However, that aside, it's much more clear-cut with ext_proc. If it runs on all routes, and the config is incorrectly modified for some reason, then all your traffic returns 500 even if the ext_proc is set to fail open.

The bigger issue IMO, is that none of this was reflected in our envoy metrics. None of our envoy alerts went off because of this modification.

arkodg:
the envoy metrics are tied to traffic, these issues here are tied to configuration, recommend monitoring kube-state-matrics to catch these early https://gateway.envoyproxy.io/docs/tasks/observability/gateway-api-metrics/

arkodg:
the direct_response filter with a response of `500` is added by Envoy Gateway to flag bad configuration
https://gateway-api.sigs.k8s.io/geps/gep-1364/?h=500#accepted, which is followed in Gateway-API APIs as well as Gateway API Extensions (Envoy Gateway CRDs)

arkodg:
[...]

Issue #6124, opened 2025-05-20
Observability into defaulted xDS configuration


DoroNahari:
*Description*:
We've encountered some cases where our Envoy Proxy got default xDS configurations from Envoy Gateway with 
```
direct_response": {
  status": 500
} 
```
We didn't see anything on the Envoy Gateway pods logs indicating the reason for updating the xDS with default "direct_response 500" configuration.
We wanted to raise the log level to maybe see an higher level log indicating the reason but then, after restarting the Envoy Gateway pods, the problem resolved.

It would be good to have some indication why Envoy Gateway decided to put the default configurations on the xDS.

Default configuration example:

```
{
  "configs": [
    {
      "dynamic_listeners": [
        {
          "active_state": {
            "listener": {
              "filter_chains": [
                {
                  "filters": [
                    {
                      "name": "envoy.filters.network.http_connection_manager",
                      "typed_config": {
                        "route_config": {
                          "virtual_hosts": [
                            {
                              "routes": [
                                {
                                  "direct_response": {
                                    "status": 500
                                  }
                                }
                              ]
                            }
                          ]
                        },
                        "http_filters": [
                          {
                            "name": "envoy.filters.http.router"
                          }
                        ]
                      }
                    }
                  ]
                }
              ]
            }
          }
        }
      ]
    }
  ]
}

```

Ido-Itz:
When running Envoy Gateway in an HA (multi-pod) deployment only one will update the resources, so if the issue is with the other pod, it wouldn't be propagated to the relevant resource.

I think it would be worthwhile to log an error when defaulting the configuration due to real erroneous cases (that aren't due to user configuration). E.g. K8S API server was non-responsive, etc.

arkodg:
what we could do here is add logging whenever we are insert a `500` direct response

PR #6401, opened 2025-06-25
fix: skip failed EEP when it's failopen

zhaohuabing:
fixes: #5905 
release note: yes


internal/gatewayapi/envoyextensionpolicy.go
```diff
@@ -330,10 +330,14 @@ func (t *Translator) translateEnvoyExtensionPolicyForRoute(
...
 				for _, r := range irListener.Routes {
 					if strings.HasPrefix(r.Name, prefix) {
-						// return 500 and do not configure EnvoyExtensions in this case
+						// if FailOpen is false, return 500 error directly and not forward the request to the backend.
+						// if FailOpen is true, skip the policy application and requests will be processed as if the
+						// policy was not present. https://github.com/envoyproxy/gateway/issues/5905
 						if errs != nil {
-							r.DirectResponse = &ir.CustomResponse{
-								StatusCode: ptr.To(uint32(500)),
+							if !ShouldFailOpen(errs) {
```
guydc: Can we perhaps add a log here, warning probably, that this is happening? Should we add some sort of informational status condition for this case?

guydc (commented):
In general, do we want to apply this to other similar policies? e.g. ext-auth failopen here: https://gateway.envoyproxy.io/docs/api/extension_types/#extauth

zhaohuabing:
> In general, do we want to apply this to other similar policies? e.g. ext-auth failopen here: https://gateway.envoyproxy.io/docs/api/extension_types/#extauth

Yes, I plan to apply the same approach to other similar places after we agree on this one - will raise a follow-up PR for that.

internal/gatewayapi/envoyextensionpolicy.go
zhaohuabing: Added a log for failopen with translation errors.

We already have a condition in the status in the policy surfacing the translation errors.

guydc (approved):
LGTM!

internal/gatewayapi/envoyextensionpolicy.go
```diff
@@ -330,10 +330,18 @@ func (t *Translator) translateEnvoyExtensionPolicyForRoute(
...
+								r.DirectResponse = &ir.CustomResponse{
+									StatusCode: ptr.To(uint32(500)),
+								}
+							} else {
+								t.Logger.Error(
```
arkodg: imo it should be `Info`, not `Error`

internal/gatewayapi/error.go
```diff
@@ -0,0 +1,92 @@
...
+	if err == nil {
+		return false
+	}
+
+	var policyErr *PolicyTranslationError
+	if errors.As(err, &policyErr) {
+		return policyErr.FailOpen
+	}
+
+	// Check if error can be unwrapped and recurse
```
arkodg: can we move to for loop to improve readability ?

internal/gatewayapi/error.go
```diff
@@ -0,0 +1,92 @@
...
+
+// ShouldFailOpen checks if the error is a PolicyTranslationError and returns its FailOpen status.
+func ShouldFailOpen(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Check if error implements interface{ Unwrap() []error }
+	if unwrapper, ok := err.(interface{ Unwrap() []error }); ok {
+		// FailOpen only if all errors in the slice are PolicyTranslationError with FailOpen set to true
```
arkodg: arent all errors PolicyTransationErrors ?

internal/gatewayapi/envoyextensionpolicy.go
zhaohuabing: IMO, this is an error that requires user intervention to resolve. Even though the errors are already in the status of xPolicy, logging as an error helps surface the issue more clearly — it can show up in dashboards, logging systems, or even trigger alerts, ensuring it doesn’t go unnoticed.

internal/gatewayapi/error.go
zhaohuabing: We don't have to wrap all errors to `PolicyTransationError`.

It's only needed for translation errors for xPolicy with FailOpen options. For example, the translation errors from Lua and Wasm don't have to be wrapped.

That said, I'm happy to wrap these errors to `PolicyTransationError` as well if we prefer consistency.

zhaohuabing: `PolicyTranslationError` is introduced to help simplify the handling of the following logic in `ShouldFailOpen`:

- If all errors come from ext_proc extensions and all of them are configured with failOpen: true, the EEP is skipped as if it doesn't exist.
- If all errors come from ext_proc extensions and one of them is configured with failOpen: false, we return a 500 direct response.
- If there are errors from multiple types of extensions, we return a 500 direct response—regardless of the failOpen settings on any ext_proc extension.

This approach keeps the error handling logic cleaner and also makes it easier to extend failOpen behavior to other extension or policy types in the future, for example, supporting the same controller behavior for the [failOpen option in the Wasm extension](https://github.com/envoyproxy/gateway/blob/b8ca95827a1f9800ce32e4b05b38abcbc65b73a4/api/v1alpha1/wasm_types.go#L49-L58).

It's also possible to handle this without introducing `PolicyTranslationError`, but it might be more fragile and harder to maintain. 

The unit test and testdata included in this PR could help clarify why it's needed.

zhaohuabing: I've gone ahead and handled the `failOpen` for Wasm extensions in the same way.

internal/gatewayapi/envoyextensionpolicy.go
zhaohuabing: Discussed with @arkodg offline: removed the log message in this PR, and we can handle it more properly in the future across the entire code base - tracked in https://github.com/envoyproxy/gateway/issues/6124.

internal/cmd/egctl/testdata/translate/out/backend-endpoint.all.yaml
```diff
@@ -60,6 +60,7 @@ envoyExtensionPolicies:
         kind: Service
         name: grpc-ext-proc
         port: 9002
+      failOpen: false
```
arkodg, line 63: umm, what happened here ?

internal/gatewayapi/envoyextensionpolicy.go
```diff
@@ -490,22 +511,32 @@ func getLuaBodyFromLocalObjectReference(valueRef *gwapiv1.LocalObjectReference,
...
+		errs          error
+	)
 
 	if policy == nil {
-		return nil, nil
+		return nil, nil, failOpen
 	}
 
+	// If any failed ExtProcs are not fail open, the whole policy is not fail open.
+	failOpen = true
```
arkodg: to maintain idiomaticity we should default to false, and use the val to set to true

arkodg:
thanks @zhaohuabing, added minor comments, else LGTM 🚀

internal/cmd/egctl/testdata/translate/out/backend-endpoint.all.yaml
zhaohuabing: I believe it's caused by this  `kubebuilder:default=false`. This is added to align with the default value of the `FailOpen` field in the `Wasm` struct. 

```
	// +optional
	// +kubebuilder:default=false
```

arkodg (approved):

PR #6406, opened 2025-06-25
feat(translator): add store metric

guydc:
**What type of PR is this?**

Adds a new metric for publish events: `watchable_publish_total`. 

EG currently records metrics for the subscriber side. The watchable queues may coalesce and drop certain updates sent to subscribers. Users may also want to track the count of store events, in order to understand if certain layers are publishing updates successfully. 

Release Notes: Yes

### DIFF
--- a/internal/gatewayapi/runner/runner.go
+++ b/internal/gatewayapi/runner/runner.go
@@ -124,7 +124,7 @@ func (r *Runner) startWasmCache(ctx context.Context) {
 }
 
 func (r *Runner) subscribeAndTranslate(sub <-chan watchable.Snapshot[string, *resource.ControllerResources]) {
-	message.HandleSubscription(message.Metadata{Runner: r.Name(), Message: "provider-resources"}, sub,
+	message.HandleSubscription(message.Metadata{Runner: r.Name(), Message: message.ProviderResourcesMessageName}, sub,
 		func(update message.Update[string, *resource.ControllerResources], errChan chan error) {
 			r.Logger.Info("received an update")
 			val := update.Value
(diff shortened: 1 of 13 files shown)


internal/gatewayapi/runner/runner.go
```diff
@@ -188,7 +188,11 @@ func (r *Runner) subscribeAndTranslate(sub <-chan watchable.Snapshot[string, *re
 						r.Logger.Error(err, "unable to validate infra ir, skipped sending it")
 						errChan <- err
 					} else {
-						r.InfraIR.Store(key, val)
+						message.HandleStore(message.Metadata{
+							Runner:  string(egv1a1.LogComponentGatewayAPIRunner),
```
arkodg: can we use `r.Name()` instead ?

reg Message, can we add a Name helper to https://github.com/envoyproxy/gateway/blob/main/internal/message/types.go ?

internal/message/watchutil.go
```diff
@@ -100,3 +100,8 @@ func HandleSubscription[K comparable, V any](
 		}
 	}
 }
+
+func HandleStore[K comparable, V any](meta Metadata, key K, value V, publish *watchable.Map[K, V]) {
+	publish.Store(key, value)
+	watchablePublishTotal.WithSuccess(meta.LabelValues()...).Increment()
```
Xunzhuo, line 106: Here it is, overall looks good to me! Thanks for the code polish as well

Xunzhuo:
Looks like there are some conflicts..

arkodg (approved):

Issue #6035, opened 2025-05-13
Flaky: TestE2E/BackendUpgrade


zirain:
--- FAIL: TestE2E/BackendUpgrade (11.09s)
        --- FAIL: TestE2E/BackendUpgrade/All_requests_must_succeed (11.03s)

https://github.com/envoyproxy/gateway/actions/runs/14994434957/job/42126092844

zirain:
```console
=== RUN   TestE2E/BackendUpgrade
    conformance.go:72: 2025-05-16T01:01:10.601578127Z: Applying testdata/backend-upgrade.yaml
    apply.go:275: 2025-05-16T01:01:10.618929358Z: Creating backend-upgrade-example BackendTrafficPolicy
    apply.go:275: 2025-05-16T01:01:10.639546823Z: Creating http-backend-upgrade HTTPRoute
=== RUN   TestE2E/BackendUpgrade/All_requests_must_succeed
    helpers.go:611: 2025-05-16T01:01:10.664560053Z: Route gateway-conformance-infra/http-backend-upgrade expected 1 Parents got 0
    helpers.go:611: 2025-05-16T01:01:11.657792071Z: Conditions matched expectations
    helpers.go:611: 2025-05-16T01:01:11.657847644Z: Route gateway-conformance-infra/http-backend-upgrade Parents matched expectations
    backend_upgrade.go:45: 2025-05-16T01:01:11.66026429Z: Making GET request to http://172.18.0.203/backend-upgrade
    roundtripper.go:219: 2025-05-16T01:01:11.660416113Z: Sending Request:
        < GET /backend-upgrade HTTP/1.1
        < Host: 172.18.0.203
        < User-Agent: Go-http-client/1.1
        < X-Echo-Set-Header: 
        < Accept-Encoding: gzip
        < 
        < 
        
        
    roundtripper.go:235: 2025-05-16T01:01:11.66209054Z: Received Response:
        < HTTP/1.1 200 OK
        < Connection: close
        < Content-Length: 572
        < Content-Type: application/json
        < Date: Fri, 16 May 2025 01:01:11 GMT
        < X-Content-Type-Options: nosniff
        < 
        < {
        <  "path": "/backend-upgrade",
        <  "host": "172.18.0.203",
        <  "method": "GET",
        <  "proto": "HTTP/1.1",
        <  "headers": {
        <   "Accept-Encoding": [
        <    "gzip"
        <   ],
        <   "User-Agent": [
[...]

Issue #6009, opened 2025-05-12
Flaky: TestEGUpgrade/EGUpgrade


zirain:
--- FAIL: TestEGUpgrade/EGUpgrade (337.13s)
        --- FAIL: TestEGUpgrade/EGUpgrade/Upgrade_from_an_older_eg_release_should_succeed (335.81s)

https://github.com/envoyproxy/gateway/actions/runs/14964409304/job/42032171338

zirain:
TestEGUpgrade/EGUpgrade/Upgrade_from_an_older_eg_release_should_succeed (336.13s)
https://github.com/envoyproxy/gateway/actions/runs/15058426044/job/42329393011

PR #6409, opened 2025-06-26
e2e: use appropriate values for upgrade test

zirain:
fixes: https://github.com/envoyproxy/gateway/issues/6035 https://github.com/envoyproxy/gateway/issues/6009


Xunzhuo:
Looks like a regression here, can you elaborate why the different values will cause e2e failures?

zirain:
flaky happen frequently when qps ≥ 4000.

PR #6410, opened 2025-06-26
feat: add support for prometheus annotations

crhuber:

Adds support to specify annotations for prometheus metrics. 

for example:

```
  telemetry:
    metrics:
      prometheus:
        annotations:
          prometheus.io/scrape: "false"
          custom.annotation: "custom-value"
```

This is needed because some environments such as in Datadog need annotation  `prometheus.io/scrape=false`

Release Notes: Yes


arkodg:
@crhuber have you tried setting it via https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/ or https://gateway-api.sigs.k8s.io/guides/infrastructure/?h=infrast ?
my understanding is you dont want to disable prometheus, you just want to edit the annotation ?

crhuber:
@arkodg 

Yes, so with this tested configuration:

```
kind: EnvoyProxy
spec:
  provider:
    kubernetes:
        pod:
          annotations:
            prometheus.io/scrape: "false"
  telemetry:
    metrics:
      prometheus:
        disable: false
```

still results in a pod with these annotations

```
│     prometheus.io/path: /stats/prometheus                                                                                                                                                                                                                                                                                                                                                                                       │
│     prometheus.io/port: "19001"                                                                                                                                                                                                                                                                                                                                                                                                 │
│     prometheus.io/scrape: "true"
```

Yes the intention is to to still have the container ports 19001 exposed and the prometheus metrics being generated , just wait to edit annotations 

 The reason for this is that it causes the datadog agent "openmetrics" check to kick in https://docs.datadoghq.com/containers/kubernetes/prometheus/?tab=kubernetesadv2#advanced-configuration and start scraping the metrics (costing us money).  Instead we want the native `envoy` integration checks which are free https://docs.datadoghq.com/integrations/envoy/ but requires us to turn off the scrape annotation  prometheus.io/scrape: "false"

I also tried 

```
kind: Gateway
spec:
  gatewayClassName: envoy-gateway
  infrastructure:
    annotations:
      prometheus.io/scrape: "false"
```

but it did not regenerate a new replicaset

crhuber:
closing in favor of https://github.com/envoyproxy/gateway/pull/6430

PR #6412, opened 2025-06-26
fix: implement new subscribe pattern to prevent race condition with close

rudrakhp:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No


internal/message/types.go
```diff
@@ -18,11 +20,17 @@ import (
 	xdstypes "github.com/envoyproxy/gateway/internal/xds/types"
 )
 
+// Subscription for Gateway API Resources
+type GatewayAPIResources struct {
+	watchable.Map[string, *resource.ControllerResources]
+	Subscription <-chan watchable.Snapshot[string, *resource.ControllerResources]
```
arkodg: doesnt this limit to 1 subscriber, instead of N (which doesnt exist today afaik, but may be needed in the future)

rudrakhp: @arkodg yes check here `xdsIR` where we are already having 2 usages:
```
func NewSubscribedXdsIR(ctx context.Context) *XdsIR {
	x := new(XdsIR)
	x.SubscriptionForTranslator = x.Subscribe(ctx)
	x.SubscriptionForGlobalRateLimit = x.Subscribe(ctx)
	return x
}
```

rudrakhp: I can rename other single `Subscription` instances to be more specific about where they are used so they are not used unknowingly elsewhere.

rudrakhp: Updated the access pattern to support N subscriptions. Count has to be incremented if more subscribers are needed after a code change. @arkodg @mathetake

mathetake:
sorry i got caught up in lots of other things atm... hopefully i can give a quick review this week but cannot guarantee (Friday is also a us holiday)

internal/gatewayapi/runner/runner_test.go
```diff
@@ -114,9 +114,10 @@ func TestGetIRKeysToDelete(t *testing.T) {
 
 func TestDeleteStatusKeys(t *testing.T) {
 	// Setup
-	pResources := new(message.ProviderResources)
-	xdsIR := new(message.XdsIR)
-	infraIR := new(message.InfraIR)
+	ctx := context.Background()
```
mathetake, line 117: nit; maybe simply use `t.Context()`?

internal/message/subscription.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+// GetNextAvailable returns the next available subscription channel that hasn't been consumed yet.
+// This method is thread-safe and atomically increments the last used index.
+func (c *SubscriptionList[K, V]) GetNextAvailable() <-chan watchable.Snapshot[K, V] {
+	c.mutex.Lock()
+	nextIndex := c.lastUsedIndex + 1
+	if nextIndex >= len(c.subscriptions) {
+		c.mutex.Unlock()
```
mathetake: instead of manually placing UnLock in two places, simply defer c.mutext.Unloc() right after the Lock on L44? Deferred function calls are guaranteed to run.

internal/message/subscription.go
```diff
@@ -0,0 +1,53 @@
...
+		panic("no available subscriptions")
```
mathetake: maybe put code comment on why this code path is not unexpected for the future reference?

internal/message/types.go
```diff
@@ -93,47 +174,139 @@ func (s *GatewayAPIStatuses) Close() {
...
+	s.TCPRouteStatuses.Subscriptions = NewSubscriptionList(ctx, &s.TCPRouteStatuses, subscriptionCountTCPRouteStatuses)
+	s.UDPRouteStatuses.Subscriptions = NewSubscriptionList(ctx, &s.UDPRouteStatuses, subscriptionCountUDPRouteStatuses)
+}
+
 type NamespacedNameAndGVK struct {
 	types.NamespacedName
 	schema.GroupVersionKind
 }
 
+// Subscriptions for Policy Statuses
```
mathetake: i would define one generic struct like 
```go
type PlicyStatusSubscriptions[p any] struct {
	watchable.Map[types.NamespacedName, p]
	Subscriptions *SubscriptionList[types.NamespacedName, p]
}
```

and instantiates the type inline inside the definition of `PolicyStatuses` below like the original code on the main branch:
```go
type PolicyStatuses struct {
    ClientTrafficPolicyStatuses PlicyStatusSubscriptions[*gwapiv1a2.PolicyStatus]
    // ....
```

The difference is that you won't need to unnecessarily export many struct types from this package. Let's limit the context of each package and help reduce the noice of the callsite of the package.

This will also help us reduce the status related construct per CRD.

internal/message/subscription.go
```diff
@@ -0,0 +1,53 @@
...
+	lastUsedIndex int
+}
+
+// Subscribable represents an object that can be subscribed to for snapshots
+type Subscribable[K comparable, V any] interface {
+	Subscribe(ctx context.Context) <-chan watchable.Snapshot[K, V]
+}
+
+// NewSubscriptionList creates a new subscription list by calling Subscribe count times on the subscribable object
+func NewSubscriptionList[K comparable, V any](ctx context.Context, subscribable Subscribable[K, V], count int) *SubscriptionList[K, V] {
```
mathetake: could you please make this function unexported?

internal/message/types.go
mathetake: this is nit though; you can ignore my comment if you like

internal/message/subscription.go
```diff
@@ -0,0 +1,53 @@
...
+type SubscriptionList[K comparable, V any] struct {
+	subscriptions []<-chan watchable.Snapshot[K, V]
+	mutex         sync.Mutex
+	lastUsedIndex int
+}
+
+// Subscribable represents an object that can be subscribed to for snapshots
+type Subscribable[K comparable, V any] interface {
+	Subscribe(ctx context.Context) <-chan watchable.Snapshot[K, V]
+}
```
mathetake: if this is needs to be exported, then I am not sure if we are on the right track. What we want to do is to prevent this thingy to be called from various places outside the main grootuine, but as long as this implementation and interface are exported, I am not sure how this rule can be exposed. Would it be possible to un-export this interface and put all the implementations contained in this message package as well?

mathetake (commented):
I am struggling to understand why do you want to have a slice and keep having the reference to it. 

> No Subscribe() calls should exist outside subscription.go, enforcing strict control over subscription points.

At the end of the day, this is only thing we want to achieve right? Why do you want the list to keep track the channels as well as I am not sure how the fixed size on the slice helps.

mathetake:
my comments are mostly the question rather than blocking -- feel free to ignore them. I would appreciate you taking a shot at this bug anyways, thank you @rudrakhp !

internal/message/types.go
rudrakhp: @mathetake yeah that makes sense, in fact all these structs can be private, no need to export these as we need to restrict subscription points.

rudrakhp:
@mathetake Thanks for the comments 🙌  
One major comment I ack is that the new structs/methods must be private to restrict subscription points outside. I am working on that. This was a draft proposal for getting some initial buy in on this approach.
> Why do you want the list to keep track the channels as well as I am not sure how the fixed size on the slice helps.

We will need to fix the size of the list and know how many subscriptions we need to pre-subscribe. Dynamically increasing the list would basically mean not pre-subscribing that we were doing already today. I don't see major advantages of subscribing to snapshots in runtime as we should know required subscriptions as part of our code change/compile time.
> maybe put code comment on why this code path is not unexpected for the future reference?

Like you correctly mentioned we should never hit this path if we have correctly identified the required number of subscriptions in the controllers/runners. 
Do let me know what you think ...

internal/message/subscription.go
mathetake: > we should never hit this path if we have correctly identified the required number of subscriptions in the controllers/runners.

@rudrakhp let's put your comment here in the source code or even in the panic message...

arkodg:
hey @rudrakhp, if the goal is to move subscription to the main goroutine, can we instead move the logic into runner creation /  `New` https://github.com/envoyproxy/gateway/blob/df5a5e9725cbaebcab60ddb6c4baae1ea228576f/internal/cmd/server.go#L168
from `runner.Start`

rudrakhp:
Simpler fix in #6566

PR #6421, opened 2025-06-27
fix: Do not set backendRequestTimeout when Retries are set

### LINKED ISSUE #6414: Dont set backendRequestTimeout to Request Timeout when Retries are set
*Description*:

https://github.com/envoyproxy/gateway/blob/1376cae998ad5cea54bae765914908ee08a6121e/internal/gatewayapi/route.go#L328 should only be set when Retries are unset
https://github.com/envoyproxy/gateway/blob/1376cae998ad5cea54bae765914908ee08a6121e/internal/gatewayapi/route.go#L358

*Repro steps*:

sharing.

*Environment*:

*Logs*:

sudiptob2:
Fixes #6414

### DIFF
--- a/internal/gatewayapi/route.go
+++ b/internal/gatewayapi/route.go
@@ -327,7 +327,7 @@ func processRouteTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
 
 		// Also set the IR Route Timeout to the backend request timeout
 		// until we introduce retries, then set it to per try timeout
-		if rule.Timeouts.BackendRequest != nil {
+		if rule.Timeouts.BackendRequest != nil && rule.Retry == nil {
 			d, err := time.ParseDuration(string(*rule.Timeouts.BackendRequest))
 			if err != nil {
 				d, _ = time.ParseDuration(HTTPRequestTimeout)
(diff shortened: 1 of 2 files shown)


internal/gatewayapi/route.go
```diff
@@ -327,7 +327,7 @@ func processRouteTimeout(irRoute *ir.HTTPRoute, rule gwapiv1.HTTPRouteRule) {
 
 		// Also set the IR Route Timeout to the backend request timeout
 		// until we introduce retries, then set it to per try timeout
-		if rule.Timeouts.BackendRequest != nil {
+		if rule.Timeouts.BackendRequest != nil && rule.Retry == nil {
```
rudrakhp, line 331: Should we simply skip setting timeout or set something in status so it's clear that this value is being disregarded? Or a log maybe?

arkodg: the logic based on the spec, so it shouldnt confuse user imo, its a bug

arkodg: @sudiptob2 can you update the comment on L320, and update it so its valid e.g. // Only set Route timeout to BackendRequest when Retries are not set

sudiptob2: Thanks for the review 🚀 Comment updated.

zirain (approved):

Issue #6425, opened 2025-06-27
Guide for deployment in airgap 


ion-elgreco:
*Description*:
The current guides don't describe that without configuring a custom EnvoyProxy with an image specification it will automatically use the distroless image from docker.io.

Example from slack:

```yaml
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: custom-proxy-config
  namespace: default
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyDeployment:
        container:
          image: envoyproxy/envoy:v1.25-latest 
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: custom-proxy-config
  listeners:
 name: http
      protocol: HTTP
      port: 80
```

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

zhaohuabing:
It would be great to add some guidance on how to deploy Envoy Gateway in a air-gapped environment. Key points to cover could include:
* Customizing the Helm chart to set a local image registry for Envoy Gateway and RateLimit image by specifying the `global.imageRegistry` and `global.imagePullSecrets`.
* Creating an `EnvoyProxy` resource with a custom Envoy image and necessary imagePullSecrets, and attaching it to the `Gateway` or `GatewayClass`.

Tasks - Operations is a good home for this guidance.

@ion-elgreco since you’ve already got this working  — would you be up for creating a PR?

ion-elgreco:
@zhaohuabing sure, I'll take a look over the weekend to add this:)

PR #6429, opened 2025-06-28
docs: add airgap deployment guide

ion-elgreco:
- closes https://github.com/envoyproxy/gateway/issues/6425

### DIFF
--- a/site/content/en/v1.4/tasks/operations/airgap-deployment.md
+++ b/site/content/en/v1.4/tasks/operations/airgap-deployment.md
@@ -1,15 +1,20 @@
 ---
-title: "Deploy Envoy Gateway in Air-Gapped Environments"
+title: 'Deploy Envoy Gateway in Air-Gapped Environments'
 ---
 
-Deploying the Envoy Gateway in an air-gapped environment using a Helm chart requires careful configuration of the `values.yaml` file as well as adjustments when deploying a Gateway resource.
+Deploying the Envoy Gateway in an air-gapped environment using a Helm chart
+requires careful configuration of the `values.yaml` file as well as adjustments
+when deploying a Gateway resource.
 
-You will need to specify custom image repositories for the following components in the Helm chart:
+You will need to specify custom image repositories for the following components
+in the Helm chart. This can be done on a global level or image level.
 
 - Gateway
 - Ratelimit
 
 ## Gateway – `values.yaml` Configuration
+
+Example done in image level:
 
 ```yaml
 deployment:
@@ -19,7 +24,19 @@ deployment:
       tag: v1.4.1
 ```
 
+It's also possible to define the registry on a global level:
+
+```yaml
+# Global settings
+global:
+  # If set, these take highest precedence and change both envoyGateway and ratelimit's container registry and pull secrets.
+  # -- Global override for image registry
+  imageRegistry: 'custom-cr.internal.io'
+```
+
 ## Ratelimit - `values.yaml` Configuration
+
+Example done on global level:
 
 ```yaml
 global:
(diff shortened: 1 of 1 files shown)


site/content/en/v1.4/tasks/operations/airgap-deployment.md
```diff
@@ -0,0 +1,102 @@
+---
+title: "Deploy Envoy Gateway in Air-Gapped Environments"
+---
+
+Deploying the Envoy Gateway in an air-gapped environment using a Helm chart requires careful configuration of the `values.yaml` file as well as adjustments when deploying a Gateway resource.
+
+You will need to specify custom image repositories for the following components in the Helm chart:
```
zhaohuabing: There're two levels of image customization:
1. global level setting if you just need to use a different registry, but the same image name

2. image level setting if you need to customize both the registry and image name

https://github.com/envoyproxy/gateway/blob/2d4fba60ec1f9007368f24c07255c43887df9577/charts/gateway-helm/values.tmpl.yaml#L1-L27

site/content/en/v1.4/tasks/operations/airgap-deployment.md
```diff
@@ -0,0 +1,102 @@
...
+```yaml
+global:
+  images:
+    ratelimit:
+      image: custom-cr.internal.io/envoyproxy/ratelimit:master
+```
+
+## Gateway Requires a Custom EnvoyProxy Reference
+
+The Gateway must reference a custom EnvoyProxy resource that explicitly specifies the location of the distroless Envoy container image. Without this, the image will be pulled implicitly from Docker Hub.
```
zhaohuabing: EnvoyProxy can be associated with either the `GatewayClass` to customize the envoy image for all the Gateways of a specific GatewayClass, or a `Gateway` for that Gateway only.

Can we show examples for both?

site/content/en/v1.4/tasks/operations/airgap-deployment.md
```diff
@@ -0,0 +1,102 @@
...
+    protocol: HTTPS
+    tls:
+      certificateRefs:
+      - name: example-tls
+      mode: Terminate
+```
+
+## Default LoadBalancer Service Type
+
+By default, Envoy uses a Service of type `LoadBalancer`. Depending on your Kubernetes environment, you might need to add custom annotations. For example, when deploying in Azure, you can configure the service as follows:
```
zhaohuabing: Is the LoadBalancer setting related to air-gapped envs?

site/content/en/v1.4/tasks/operations/airgap-deployment.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Ratelimit - `values.yaml` Configuration
+
+```yaml
+global:
+  images:
+    ratelimit:
+      image: custom-cr.internal.io/envoyproxy/ratelimit:master
+```
```
zhaohuabing, line 46: Can we also mention the imagePullSecrets setting? ImagePullSecrets would be needed if the registry is private.

zhaohuabing:
Hi @ion-elgreco would you mind addressing these comments when you get a chance?

ion-elgreco:
@zhaohuabing I'll see if I can manage some spare time this weekend!

site/content/en/v1.4/tasks/operations/airgap-deployment.md
ion-elgreco: It can be, but it's definitely a footgun when you don't expect this default, since it was tricky to track down.

ion-elgreco:
@zhaohuabing I've pushed the changes now

zhaohuabing (approved):
LGTM thanks!

site/content/en/v1.4/tasks/operations/airgap-deployment.md
```diff
@@ -0,0 +1,160 @@
...
+spec:
+  controllerName: gateway.envoyproxy.io/gatewayclass-controller
+  parametersRef:
+    group: gateway.envoyproxy.io
+    kind: EnvoyProxy
+    name: custom-envoy-proxy
+    namespace: default
+```
+
+## Default LoadBalancer Service Type
```
arkodg, line 137: this is something that lives in https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/
can we add a line instead that says "for further customizations such as .... go here ...."

zirain (approved):

site/content/en/v1.4/tasks/operations/airgap-deployment.md
zhaohuabing: Address this in a follow-up PR.

PR #6433, opened 2025-06-29
fix: By default, httproutes are sorted alphabetically based on names

### LINKED ISSUE #6140: CPU spikes result from inconsistent route order after each pod restarts
*Description*:

We noticed that envoy CPU spikes whenever a pod restarts. This worsens when multiple deployments roll out together. The perf output shows most of these CPU spent on parsing and compiling CORS regex. This spike persists even zero traffic sent to the envoy.

Image

We can re-produce the same issue in a test environment with a gateway and multiple Httproutes configured in different namespaces (see details below). After digging into a little deeper, we identified the following issues/facts.

1. Every pod, with at least an Httproute, restarts triggers a full route configuration reload. It also re-compiles paths and cores domains with regex patterns.
2. We can confirm this by watching `Envoy::Regex::Utility::parseRegex` calls. see screenshots.
3. We also noticed that, after every pod restarts, the order of the route list changes. _See attached screenshots; 1 = before first restart, 2 = after first pod restart, 3 = after second pod restarts._
4. Applying Patchpolicy consistently results in this behavior (See the attached sample config). 
5. I can see this behaviour without patch policy when envoy proxy is newly started. But it is reproducible only one or two times. Not sure if too many pod churn can produce this.

I think this is the reason envoy spending too much CPU in the CORS re-compile. In our setup, we have 400+ routes and most of them are prefix or domain match. But we have 15+ regex based global cors domains. This means, 6000+ regex compilation on every pod restarts.

Image

Image

*Repro steps*:

1. Apply a patch policy on `type.googleapis.com/envoy.config.listener.v3.Listener`
2. Create a single Gateway config and allow cross-namespace reference.
3. Create multiple http routes in multiple namespaces with no domain names(i.e. *). Note: It should have regex pattern in few routes(or a global cors with wildcard match) to watch via bpftrace.
4. Take config dump
5. Restart one of the pod from any namespaces.
6. Wait till new pod is back and endpoint is updated.
7. Take second config dump
8. Compare route order in config dumps.

**How to watch via bpftrace?**

You can run following onliner in host machine to list total `Envoy::Regex::Utility::parseRegex` call made after a pod restarts.

`bpftrace --unsafe  -e ‘BEGIN{print(“Probing Envoy::Regex::Utility::parseRegex call”)} uprobe:/proc/<pid>/root/usr/local/bin/envoy:0x2087a00 {  @total = count();print(“call”)}’ 2>/dev/null`

Note that, the production image of envoy has a stripped binary and you need to calculate offset of fucntion manually from corresponding debug image. The offset `0x2087a00` may only work with `envoyproxy/envoy:contrib-v1.32.5`.

Alternatively, you can use debug image itself as proxy image and run `bpftrace -l <path>|grep parseRegex` to list the mangled symbol.

*Environment*
Gateway version: v1.2.7
Proxy version/image: envoyproxy/envoy:contrib-v1.32.5

*Sample config*:

[patchpolicy.txt](https://github.com/user-attachments/files/20366272/patchpolicy.txt)

fayizk1:
**What type of PR is this?**

fix: By default, httproutes are sorted alphabetically based on names

This PR ensures consistent route ordering by including route names in the default sorting logic, preventing the inconsistent reordering that may be triggered by unrelated updates (i.e. pod churn, etc.). After all sorting conditions, ties exist across multiple routes, causing this.

Requirements for the GW-API under these circumstances:

>  - The Route appearing first in alphabetical order by "{namespace}/{name}"

So sort routes by route names, which is a combination of namespace, route name, route index and domain.

Fixes #6140

Release Notes: No

### DIFF
--- a/internal/gatewayapi/testdata/httproute-default-order-by-route-name.in.yaml
+++ b/internal/gatewayapi/testdata/httproute-default-order-by-route-name.in.yaml
@@ -75,10 +75,24 @@ httpRoutes:
       backendRefs:
       - name: service-1
         port: 8080
+    - matches:
+      - path:
+          type: PathPrefix
+          value: '/foobar'
+      backendRefs:
+      - name: service-2
+        port: 8080
+    - matches:
+      - path:
+          type: PathPrefix
+          value: '/barfoo'
+      backendRefs:
+      - name: service-3
+        port: 8080
 - apiVersion: gateway.networking.k8s.io/v1
   kind: HTTPRoute
   metadata:
-    namespace: test-ns 
+    namespace: test-ns
     name: httproute-2
   spec:
     hostnames:
@@ -96,36 +110,36 @@ httpRoutes:
       - name: test-service
         port: 8080
 namespaces:
-  - apiVersion: v1
-    kind: Namespace
-    metadata:
-      name: test-ns
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    name: test-ns
 services:
-  - apiVersion: v1
-    kind: Service
-    metadata:
-      namespace: test-ns
-      name: test-service
-    spec:
-      ports:
-        - port: 8080
-          name: http
-          protocol: TCP
-endpointSlices:
-  - apiVersion: discovery.k8s.io/v1
-    kind: EndpointSlice
-    metadata:
-      name: endpointslice-test-service
-      namespace: test-ns
-      labels:
-        kubernetes.io/service-name: test-service
-    addressType: IPv4
+- apiVersion: v1
+  kind: Service
+  metadata:
+    namespace: test-ns
+    name: test-service
+  spec:
     ports:
-      - name: http
-        protocol: TCP
-        port: 8080
-    endpoints:
-      - addresses:
-          - 8.8.8.8
-        conditions:
-          ready: true
+    - port: 8080
+      name: http
+      protocol: TCP
+endpointSlices:
+- apiVersion: discovery.k8s.io/v1
+  kind: EndpointSlice
+  metadata:
+    name: endpointslice-test-service
+    namespace: test-ns
+    labels:
+      kubernetes.io/service-name: test-service
+  addressType: IPv4
+  ports:
+  - name: http
+    protocol: TCP
+    port: 8080
+  endpoints:
+  - addresses:
+    - 8.8.8.8
+    conditions:
+      ready: true
--- a/internal/gatewayapi/testdata/httproute-default-order-by-route-name.out.yaml
+++ b/internal/gatewayapi/testdata/httproute-default-order-by-route-name.out.yaml
@@ -140,6 +140,20 @@ httpRoutes:
       - path:
           type: PathPrefix
           value: /route1
+    - backendRefs:
+      - name: service-2
+        port: 8080
+      matches:
+      - path:
+          type: PathPrefix
+          value: /foobar
+    - backendRefs:
+      - name: service-3
+        port: 8080
+      matches:
+      - path:
+          type: PathPrefix
+          value: /barfoo
   status:
     parents:
     - conditions:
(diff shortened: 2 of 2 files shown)


internal/gatewayapi/testdata/httproute-default-order-by-route-name.in.yaml
```diff
@@ -0,0 +1,131 @@
...
+    namespace: default
+    name: httproute-1
+  spec:
+    hostnames:
+    - gateway.envoyproxy.io
+    parentRefs:
+    - namespace: envoy-gateway
+      name: gateway-1
+      sectionName: http
+    rules:
```
arkodg, line 70: can you also add an example with multiple rules, same path prefix lengh, but different chars

fayizk1: Done. Updated test example to include more rules in HTTPRoute

arkodg:
by doing this we are losing creationTimestamp order, because that is how the HTTPRoutes are originally processed
and now rely on naming order



the pro is this adds consistent ordering even when timestamp is same

the con is
* doesnt align with how ties should be broken according to https://gateway-api.sigs.k8s.io/reference/spec/#httprouterule
* users who rely on timestamp order will need to rely on naming order

fayizk1:
> by doing this we are losing creationTimestamp order, because that is how the HTTPRoutes are originally processed and now rely on naming order
> 
> the pro is this adds consistent ordering even when timestamp is same
> 
> the con is
> 
>     * doesnt align with how ties should be broken according to https://gateway-api.sigs.k8s.io/reference/spec/#httprouterule
> 
>     * users who rely on timestamp order will need to rely on naming order
> 
> 
> wdyt @envoyproxy/gateway-maintainers

@arkodg 
Another option is, include a sort logic [here](https://github.com/envoyproxy/gateway/blob/v1.4.1/internal/gatewayapi/sort.go#L99) based on namespace and name of the `HTTPRoute` if their creation dates are equal. This exactly match specs, but I have not checked if this order break later based on route rules. I can test it today and let you know.

arkodg:
> > by doing this we are losing creationTimestamp order, because that is how the HTTPRoutes are originally processed and now rely on naming order
> > the pro is this adds consistent ordering even when timestamp is same
> > the con is
> > ```
> > * doesnt align with how ties should be broken according to https://gateway-api.sigs.k8s.io/reference/spec/#httprouterule
> > 
> > * users who rely on timestamp order will need to rely on naming order
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > wdyt @envoyproxy/gateway-maintainers
> 
> @arkodg Another option is, include a sort logic [here](https://github.com/envoyproxy/gateway/blob/v1.4.1/internal/gatewayapi/sort.go#L99) based on namespace and name of the `HTTPRoute` if their creation dates are equal. This exactly match specs, but I have not checked if this order break later based on route rules. I can test it today and let you know.

@fayizk1 to do this, we'd also have to save creationTimestamp in the IR, which would increase mem, today the sort happens before translation
https://github.com/envoyproxy/gateway/blob/c455a6e6158667f5d8b2036416860c45da479371/internal/gatewayapi/route.go#L58

arkodg:
hey @fayizk1 rethinking this one, this can be solved in https://github.com/envoyproxy/gateway/blob/c455a6e6158667f5d8b2036416860c45da479371/internal/gatewayapi/route.go#L60 by converting it into a Sort Func and creating a two level match within it (timestamp and alphabetical name), it should have the same desired testdata result

fayizk1:
> hey @fayizk1 rethinking this one, this can be solved in https://github.com/envoyproxy/gateway/blob/c455a6e6158667f5d8b2036416860c45da479371/internal/gatewayapi/route.go#L60 by converting it into a Sort Func and creating a two level match within it (timestamp and alphabetical name), it should have the same desired testdata result

Exactly, this is what I wanted to test, and I had a working version for sometime but did not get time to add a broader test case. We may also need to move to a stable sort in other sort logic to avoid order change when equal case. I will open another PR as soon as possible.

_(Just realized that I put wrong long in previous comment :) )_

arkodg:
closing in favor of https://github.com/envoyproxy/gateway/pull/6550

PR #6438, opened 2025-06-30
feat: Add ExtProc support to BackendTrafficPolicy for backend-specific external processing

### LINKED ISSUE #5881: Support Ext Proc that executes per Backend
*Description*:
would be used.

Ext Proc https://gateway.envoyproxy.io/docs/api/extension_types/#extproc can be enhanced with a `phase`/`stage` field so it can be executed in the `Route` / `Backend` or `All` phases/stages

[optional *Relevant Links*:]

shiavm006:
Summary
Implements backend-specific external processing by adding ExtProc support to BackendTrafficPolicy. This enables different external processing logic per backend after backend selection occurs.

>Changes
- Add `extProc` field to `BackendTrafficPolicySpec` 
- Implement `buildBackendExtProc` function that reuses existing ExtProc translation logic
- Add ExtProc support to IR TrafficFeatures for backend-specific processing
- Update generated code and documentation

Fixes: Support Ext Proc that executes per Backend 
issue fixed #5881
what r ur views on it ?

### DIFF
--- a/internal/gatewayapi/envoyextensionpolicy.go
+++ b/internal/gatewayapi/envoyextensionpolicy.go
@@ -549,6 +549,12 @@ func (t *Translator) buildExtProc(
 		Destination: *rd,
 		Traffic:     traffic,
 		Authority:   authority,
+		Stage:       string(egv1a1.ExtProcStageRoute), // Default to Route
+	}
+
+	// Set stage if specified
+	if extProc.Stage != nil {
+		extProcIR.Stage = string(*extProc.Stage)
 	}
 
 	if extProc.MessageTimeout != nil {
--- a/internal/gatewayapi/testdata/backendtrafficpolicy-dns-lookup-family.out.yaml
+++ b/internal/gatewayapi/testdata/backendtrafficpolicy-dns-lookup-family.out.yaml
@@ -434,6 +434,7 @@ xdsIR:
                 protocol: GRPC
                 weight: 1
             name: envoyextensionpolicy/default/policy-for-httproute/extproc/0
+            stage: Route
             traffic:
               dns:
                 dnsRefreshRate: 5s
--- a/internal/gatewayapi/testdata/envoyextensionpolicy-override-replace.out.yaml
+++ b/internal/gatewayapi/testdata/envoyextensionpolicy-override-replace.out.yaml
@@ -324,6 +324,7 @@ xdsIR:
                 protocol: GRPC
                 weight: 1
             name: envoyextensionpolicy/default/policy-for-route-1/extproc/0
+            stage: Route
         hostname: gateway.envoyproxy.io
         isHTTP2: false
         metadata:
@@ -372,6 +373,7 @@ xdsIR:
                 protocol: GRPC
                 weight: 1
             name: envoyextensionpolicy/envoy-gateway/policy-for-gateway-1/extproc/0
+            stage: Route
         hostname: gateway.envoyproxy.io
         isHTTP2: false
         metadata:
--- a/internal/gatewayapi/testdata/envoyextensionpolicy-with-extproc-with-backendtlspolicy.out.yaml
+++ b/internal/gatewayapi/testdata/envoyextensionpolicy-with-extproc-with-backendtlspolicy.out.yaml
@@ -372,6 +372,7 @@ xdsIR:
             - connection.requested_server_name
             responseBodyProcessingMode: Streamed
             responseHeaderProcessing: true
+            stage: Route
         hostname: www.foo.com
         isHTTP2: false
         metadata:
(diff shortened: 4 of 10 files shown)


guydc:
Hi @shiavm006. Thanks for looking into this feature request. 
In the linked issue, the proposal for API was: 
> Ext Proc https://gateway.envoyproxy.io/docs/api/extension_types/#extproc can be enhanced with a phase/stage field so it can be executed in the Route / Backend or All phases/stages

In other words, the proposal was to enhance the existing EnvoyExtensionPolicy and make it possible to configure the scope of the ext-proc filter: downstream (listener HCM filter), upstream (cluster filter) or both. 

I support that API proposal:
- It keeps control over extensions in the hands of a privileged user. BackendTrafficPolicy can be used (typically) by application developers (owners of XRoutes), who are not as trusted when it comes to injecting extensions. 
- It's consistent with the current API: 
  - The current ext-proc is not in ClientTrafficPolicy
  - Ext-proc is the same CRD as all other extensions (lua, wasm, ... )

shiavm006:
> Hi @shiavm006. Thanks for looking into this feature request. In the linked issue, the proposal for API was:
> 
> > Ext Proc https://gateway.envoyproxy.io/docs/api/extension_types/#extproc can be enhanced with a phase/stage field so it can be executed in the Route / Backend or All phases/stages
> 
> In other words, the proposal was to enhance the existing EnvoyExtensionPolicy and make it possible to configure the scope of the ext-proc filter: downstream (listener HCM filter), upstream (cluster filter) or both.
> 
> I support that API proposal:
> 
> * It keeps control over extensions in the hands of a privileged user. BackendTrafficPolicy can be used (typically) by application developers (owners of XRoutes), who are not as trusted when it comes to injecting extensions.
> * It's consistent with the current API:
>   
>   * The current ext-proc is not in ClientTrafficPolicy
>   * Ext-proc is the same CRD as all other extensions (lua, wasm, ... )

hi @guydc Thank you for the clarification! You're absolutely right about the security model concerns.
I see now that my BackendTrafficPolicy approach has architectural issues
but i have some question i would apprecaite if u clarify that 
1. Field name preference: phase, stage, or executionPoint?
2. Default value for backwards compatibility (I assume "downstream")?
Thanks for your patience and the clear guidance!

arkodg:
@shiavm006 i recommended raising a API only PR for EEP, you could use `stage` , the orig issue has possible enum options, `Route` is the default

shiavm006:
> @shiavm006 i recommended raising a API only PR for EEP, you could use `stage` , the orig issue has possible enum options, `Route` is the default

Hi  ,
Thank you for the detailed feedback and clear guidance! I've updated the PR to follow the approach you recommended.
I have removed the previous implementation from BackendTrafficPolicy and instead enhanced EnvoyExtensionPolicy by adding a stage field to the ExtProc configuration. This new field allows setting the execution phase to Route (the default), Backend, or All.
This aligns perfectly with your points on security and API consistency and provides a solid foundation for the full implementation.

Let me know what you think of the new direction!

shiavm006:
> @shiavm006 i recommended raising a API only PR for EEP, you could use `stage` , the orig issue has possible enum options, `Route` is the default

i would love to know if u review it @arkodg @guydc

api/v1alpha1/backendtrafficpolicy_types.go
```diff
@@ -72,11 +72,13 @@ type BackendTrafficPolicySpec struct {
...
 
-	// The compression config for the http streams.
-	//
-	// +optional
+	// Compression settings for HTTP Response
 	Compression []*Compression `json:"compression,omitempty"`
 
+	// ExtProc for backend-specific external processing
+	// +optional
+	ExtProc []ExtProc `json:"extProc,omitempty"`
```
zhaohuabing, line 80: We don't need to touch BTP. We can still use EEP, just add a stage field to the extProc struct.

api/v1alpha1/ext_proc_types.go
```diff
@@ -24,6 +24,18 @@ const (
 	BufferedPartialExtBodyHeaderProcessingMode ExtProcBodyProcessingMode = "BufferedPartial"
 )
 
+// +kubebuilder:validation:Enum=Route;Backend;All
+type ExtProcStage string
```
zhaohuabing, line 28: Should we use a more common name? It could be used for other filters as well.
* ExecutionStage
* ExecutionPhase
* Stage
...

api/v1alpha1/backendtrafficpolicy_types.go
shiavm006: i have went through ur point and worked on the solution u can ceck it now

PR #6444, opened 2025-07-01
feat: api key auth forwarding setting

zhaohuabing:
This is required when
* the client identity needs to be used in a downstream filter or the backend service.
* we want to avoid exposing the API key to the backend service.

fixes: https://github.com/envoyproxy/gateway/issues/6461


api/v1alpha1/api_key_auth_types.go
```diff
@@ -45,4 +45,24 @@ type ExtractFrom struct {
 	//
 	// +optional
 	Cookies []string `json:"cookies,omitempty"`
+
+	// Forwarding defines how to forward the client identity to the backend service.
+	// If not specified, the client identity will not be forwarded.
+	//
+	// +optional
+	Forwarding *APIKeyAuthHeaderForwarding `json:"forwarding,omitempty"`
```
arkodg: imo this should live at the top level in `APIKeyAuth`
and would be good to keep it consistent with Basic Auth
https://github.com/envoyproxy/gateway/blob/fc2766b1c794926ef2c0946fbbbb1cf0e3fce252/api/v1alpha1/basic_auth_types.go#L36

zhaohuabing: How about this?

```go
// APIKeyAuth defines the configuration for the API Key Authentication.
type APIKeyAuth struct {
	// CredentialRefs is the Kubernetes secret which contains the API keys.
	// This is an Opaque secret.
	// Each API key is stored in the key representing the client id.
	// If the secrets have a key for a duplicated client, the first one will be used.
	CredentialRefs []gwapiv1.SecretObjectReference `json:"credentialRefs"`

	// ExtractFrom is where to fetch the key from the coming request.
	// The value from the first source that has a key will be used.
	ExtractFrom []*ExtractFrom `json:"extractFrom"`

	// ForwardClientIDHeader is the name of the header to forward the client identity to the backend
	// service. The header will be added to the request with the client id as the value.
	//
	// +optional
	ForwardClientIDHeader *string `json:"forwardClientIDHeader,omitempty"`

	// SanitizeAPIKey indicates whether to remove the API key from the request
	// before forwarding it to the backend service.
	//
	// +optional
	SanitizeAPIKey *bool `json:"sanitizeAPIKey,omitempty"`
}
```

api/v1alpha1/api_key_auth_types.go
```diff
@@ -46,3 +58,17 @@ type ExtractFrom struct {
 	// +optional
 	Cookies []string `json:"cookies,omitempty"`
 }
+
+type APIKeyAuthHeaderForwarding struct {
```
arkodg: is this needed ?

api/v1alpha1/api_key_auth_types.go
```diff
@@ -22,6 +22,18 @@ type APIKeyAuth struct {
...
+	// service. The header will be added to the request with the client id as the value.
+	//
+	// +optional
+	ForwardClientIDHeader *string `json:"forwardClientIDHeader,omitempty"`
+
+	// SanitizeAPIKey indicates whether to remove the API key from the request
+	// before forwarding it to the backend service.
+	//
+	// +optional
+	SanitizeAPIKey *bool `json:"sanitizeAPIKey,omitempty"`
```
arkodg: any prior art in this area ?
hard to make a choice b/w this and `SuppressCredentials`

zhaohuabing: Chose `Sanitize` because it's already used in EG API for similar purposes, and it sounds more straightforward and intuitive than `Suppress`.

```
const (
	// XFCCForwardModeSanitize removes the XFCC header from the request. This is the default mode.
	XFCCForwardModeSanitize XFCCForwardMode = "Sanitize"

	// The existing certificate information in the XFCC header is removed.
	XFCCForwardModeSanitizeSet XFCCForwardMode = "SanitizeSet"
)
```

arkodg: sure Sanitize works, is `apiKeyAuth.sanitizeAPIKey` too wordy ? 
should it be `apiKeyAuth.sanitize` instead ?

zhaohuabing: Yah, `sanitize ` sounds good to me.

internal/gatewayapi/testdata/envoypatchpolicy-invalid-target-kind-merge-gateways.in.yaml
```diff
@@ -20,7 +20,7 @@ envoyPatchPolicies:
       name: gateway-1
     jsonPatches:
     - type: "type.googleapis.com/envoy.config.listener.v3.Listener"
-      name: "envoy-gateway-gateway-1-http"
+      name: "0.0.0.0-80"
```
arkodg: diff from another PR ?

arkodg:
thanks @zhaohuabing ! there's a typo, else LGTM

internal/gatewayapi/testdata/envoypatchpolicy-invalid-target-kind-merge-gateways.in.yaml
zhaohuabing: Ha ...

arkodg (approved):
LGTM thanks !

Xunzhuo (approved):
/lgtm

PR #6445, opened 2025-07-01
feat: aggregate ancestor status

### LINKED ISSUE #5228: BackendTLSPolicy status updates fail when there are more than 16 ancestors
*Description*:
There is no effective limit on the number of Ancestors of BackendTLSPolicy:
- Gateway (Route BackendRef), EEP (Ext-Proc BackendRef), SP (Ext-Auth BackendRef) are considered Ancestors of a BTLSP
- A common backend/auxiliary-service within a cluster may be referenced by an unlimited amount of GWs/Policies

When the AncestorRef list is too long, CEL validations reject the status update operation. As a result, the actual status is not reflected, which could lead to:
- Failures in policy processing not being reported
- End-user automation that relies on status to reach a healthy state for all ancestors would fail

*Repro steps*:

Create:
-  17 GWs with routes referencing the same backend service
- BTLSP for the backend service

Observe EG Logs: 

> {"runner": "provider", "name": "policy-btls", "namespace": "gateway-conformance-infra", "error": "BackendTLSPolicy.gateway.networking.k8s.io \"policy-btls\" is invalid: [ancestors: Too many: 17: must have at most 16 items, <nil>: Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]"}

*Environment*:

EG 1.3.0

kkk777-7:
Aggregate policy ancestor status when there are more than 16 ancestors.
Define **Aggregated** condition to explicitly indicate that it is aggregated.

~~If this approach is suitable, apply it to all policies in next follow up pr.~~
This pr covers all policies.

Fixes #5228

Release Notes: Yes


internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
```diff
@@ -0,0 +1,2727 @@
...
+  status:
+    ancestors:
+    - ancestorRef:
+        group: gateway.networking.k8s.io
+        kind: BackendTLSPolicy
+        name: policy-btls
+        namespace: envoy-gateway
+      conditions:
+      - lastTransitionTime: null
+        message: This policy has 18 ancestors with Accepted condition in True status
```
zhaohuabing: It would be useful if we could include the names of Gateways in the message as well.

internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
```diff
@@ -0,0 +1,2727 @@
...
+      - group: ""
+        kind: ConfigMap
+        name: ca-cmap
+      hostname: example.com
+  status:
+    ancestors:
+    - ancestorRef:
+        group: gateway.networking.k8s.io
+        kind: BackendTLSPolicy
+        name: policy-btls
```
zhaohuabing: Would it be confusing if we use BackendTLSPolicy as the parent? How about using one of the Gateway as the parent and list all parents in the Aggregated condition message?


```yaml
    ancestors:
    - ancestorRef:
        kind: Gateway
        name: gateway1
        namespace: envoy-gateway
      conditions:
      - lastTransitionTime: null
        message: This policy has 18 ancestors with the Accepted condition set to True.
        reason: Accepted
        status: "True"
        type: Accepted
      - lastTransitionTime: null
        message: Ancestors have been aggregated because the number of policy ancestors exceeds 16. Aggregated ancestors: gateway1, gateway2, gateway3, gateway4. ...  (truncated to fit max message length, e.g., 1000 chars)
        reason: Aggregated
        status: "True"
        type: Aggregated
  ```

Comment (zhaohuabing) [reply internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml]: Or we can try our best effort to list the first 16 ancestors and add an Aggregated condition message to show the rest.

File: internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml (hunk 1 above)
Comment (kkk777-7) [reply internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml]: thanks for comment, I've Added by the following commit.
https://github.com/envoyproxy/gateway/pull/6445/commits/71704961b82a0f443f85c5ad489b1442f6a793cd

File: internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml (hunk 2 above)
Comment (kkk777-7) [reply internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml]: You're right! it would be better to use the actual ancestor.
so I’ve fixed it by using one of the Gateways as the parent, and listing all parents in each condition message.

Comment (kkk777-7) [reply internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml]: I'd like to ask for your opinion.
While the type validation allows up to 32,768, very long messages might be hard to handle in practice.
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1630-L1635

Should we truncate messages beyond a certain length?

File: internal/gatewayapi/translator.go [hunk 3]
```diff
@@ -247,6 +247,9 @@ func (t *Translator) Translate(resources *resource.Resources) (*TranslateResult,
 		errs = errors.Join(errs, err)
 	}
 
+	// Update status of Backend TLS Policies after translating all resources
```
arkodg, line 250: can we add a `Status` suffix to the func name to make it more explicit

arkodg: ProcessBackendTLSPolicyStatus

internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
arkodg: +1 to Huabing's suggestion of
* showing the first 15 ancestors as is
* showing the 16th ancestor as is with an `Accepted` condition + adding a generic `Aggregated` condition to it (we dont need to refer to 17th, 18th....)

arkodg:
thanks @kkk777-7, added a comment, can you also add a test case for BTP

internal/gatewayapi/translator.go
kkk777-7: fixed [this commit](https://github.com/envoyproxy/gateway/pull/6445/commits/77fcd03e7243b4bdd015e7539dbca53732e86e02).

internal/gatewayapi/backendtrafficpolicy.go
```diff
@@ -328,6 +328,12 @@ func (t *Translator) ProcessBackendTrafficPolicies(resources *resource.Resources
 		}
 	}
 
+	for _, policy := range res {
```
arkodg, line 331: we'll need something similar for client traffic policy, envoy extension policy and security policy as well

internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
kkk777-7: @arkodg 
thanks for comment. 
I’ve updated it again, could you check whether it matches what you had in mind? 
Also prioritizing the display of ancestors that have either an Overridden or Conflicted condition.

internal/gatewayapi/backendtrafficpolicy.go
kkk777-7: thanks for the comment! I'll address it all in the next commit.

internal/gatewayapi/status/policy.go
```diff
@@ -90,3 +96,47 @@ func SetConditionForPolicyAncestor(policyStatus *gwapiv1a2.PolicyStatus, ancesto
...
+		controllerName,
+		egv1a1.PolicyConditionAggregated,
+		metav1.ConditionTrue,
+		egv1a1.PolicyReasonAggregated,
+		"Ancestors have been aggregated because the number of policy ancestors exceeds 16.",
+		generation,
+	)
+}
+
+// sortRankForPolicyAncestor returns an integer sort key.
```
arkodg, line 145: nice !

arkodg (dismissed):
LGTM thanks !

internal/gatewayapi/status/policy.go
```diff
@@ -90,3 +96,47 @@ func SetConditionForPolicyAncestor(policyStatus *gwapiv1a2.PolicyStatus, ancesto
...
+
+// TruncatePolicyAncestors trims PolicyStatus.Ancestors down to at most 16 entries.
+// The first 15 ancestors are shown as is.
+// The last 16th ancestor is shown as is and add Aggregated condition.
+func TruncatePolicyAncestors(policyStatus *gwapiv1a2.PolicyStatus, controllerName string, generation int64) {
+	if len(policyStatus.Ancestors) <= 16 {
+		return
+	}
+
+	slices.SortStableFunc(policyStatus.Ancestors, func(a, b gwapiv1a2.PolicyAncestorStatus) int {
```
arkodg, line 116: nit: would be great to add a comment here saying we need to truncate because of the limit, so we are choosing to preserve the 16 most important ancestors .....

internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
```diff
@@ -0,0 +1,3021 @@
...
+        namespace: envoy-gateway
+      conditions:
+      - lastTransitionTime: null
+        message: Policy has been accepted.
+        reason: Accepted
+        status: "True"
+        type: Accepted
+      - lastTransitionTime: null
+        message: Ancestors have been aggregated because the number of policy ancestors
+          exceeds 16.
```
zhaohuabing: Can we list the aggregated Gateways in the message?

Like:

```
Ancestors have been aggregated because the number of policy ancestors exceeds 16. The aggregated ancestors: Gateway-17, Gateway-18, Gateway-19.
```

kkk777-7: I also think it would be useful to list the aggregated gateways in the message.
I'm only concerned about the following point in [this comment](https://github.com/envoyproxy/gateway/pull/6445#discussion_r2190673966) - how should we decide the max length of message?

internal/gatewayapi/status/policy.go
kkk777-7: Added! [88be5e1](https://github.com/envoyproxy/gateway/pull/6445/commits/88be5e141f1559f330de0d7f01398c503f43e8b4)

internal/gatewayapi/testdata/backendtlspolicy-status-conditions-truncated.out.yaml
zhaohuabing: Can we use the max allowed lenght of the message feild?  Do you have any specific concers on this?

arkodg (approved):

zhaohuabing (dismissed):
LGTM! One small (non-blocking) suggestion: would be great to include the aggregated Gateways in the message too.

kkk777-7:
@zhaohuabing 
>LGTM! One small (non-blocking) suggestion: would be great to include the aggregated Gateways in the message too.

Sorry for the late response.
I've added it to the message, please check when you have a moment.

zhaohuabing (approved):

arkodg (approved):

PR #6447, opened 2025-07-02
Add default=1 for MaxAcceptPerSocketEvent

jukie:
**What type of PR is this?**

Fixes #

Release Notes: Yes/No


internal/xds/translator/testdata/out/xds-ir/max-conn-per-socket-event.listeners.yaml
```diff
@@ -42,9 +42,8 @@
     - name: envoy.filters.network.tcp_proxy
       typedConfig:
         '@type': type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
-        cluster: tcp-route-dest
+        cluster: max-accept-disabled
         statPrefix: tcp-10081
-    name: tcp-route-dest
-  maxConnectionsToAcceptPerSocketEvent: 3
```
rudrakhp, line 48: I was expecting all other XDS translator tests to have `maxConnectionsToAcceptPerSocketEvent: 1` in their listeners set to 1 by default, or am I missing something 🤔

rudrakhp: Not talking about this case as we have explicitly set it to `0` to disable

internal/xds/translator/listener.go
```diff
@@ -233,7 +233,7 @@ func buildPerConnectionBufferLimitBytes(connection *ir.ClientConnection) *wrappe
 }
 
 func buildMaxAcceptPerSocketEvent(connection *ir.ClientConnection) *wrapperspb.UInt32Value {
-	if connection != nil && connection.MaxAcceptPerSocketEvent != nil {
+	if connection != nil && connection.MaxAcceptPerSocketEvent != nil && *connection.MaxAcceptPerSocketEvent > 0 {
```
rudrakhp: I think if `connection != nil && connection.MaxAcceptPerSocketEvent == nil` we need to set it to `1`.  This is to ensure translator reflects the default behavior, without kube validation. Ideally this should not happen since +kubebuilder:default=1 ensures it.

jukie: Thanks I was working through that once I realized I hadn't updated the other testdata. Should be good now.

internal/xds/translator/listener.go
```diff
@@ -52,7 +52,8 @@ const (
 	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-connection-window-size
 	http2InitialConnectionWindowSize = 1048576 // 1 MiB
 	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/connection_limit/v3/connection_limit.proto
-	networkConnectionLimit = "envoy.filters.network.connection_limit"
+	networkConnectionLimit             = "envoy.filters.network.connection_limit"
+	maxAcceptConnectionsPerSocketEvent = 1
```
rudrakhp: nit
```suggestion
	defaultMaxAcceptConnectionsPerSocketEvent = 1
```

PR #5023, opened 2025-01-07
feat(translator): support ext-proc dynamic metadata options

guydc:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #3170, #4160

Release Notes: Yes

PR #6453, opened 2025-07-02
feat(translator): support ext-auth dynamic metadata options (Envoy to auth server)

nareddyt:

Added support for specifying dynamic metadata namespaces that External Authorization services can access in SecurityPolicy API 

This follows the structure and CRD naming conventions established in https://github.com/envoyproxy/gateway/pull/5023. 

NOTE: This PR only introduces fields for "sending metadata to ext-authz server". It does NOT handle the reverse operation of "parsing ext-authz HTTP headers and turning them into dynamic metadata on Envoy" (which https://github.com/envoyproxy/gateway/issues/4163 was initially opened for). However, this PR structures the CRD changes in a way that allows for the reverse operation to be easily added.

Not aware of any pre-existing issue describing this specific use case. https://github.com/envoyproxy/gateway/issues/4163 is related but for the reverse operation.

Release Notes: Yes


charts/gateway-crds-helm/templates/generated/gateway.envoyproxy.io_securitypolicies.yaml
```diff
@@ -518,6 +518,40 @@ spec:
...
+                          type: string
+                        type: array
+                      routeTypedNamespaces:
+                        description: RouteTypedNamespaces are route-level typed metadata
+                          namespaces that are sent to the external authorization server
+                          as context.
+                        items:
+                          type: string
+                        type: array
+                      typedNamespaces:
```
TheRushingWookie, line 548: nit: this chunk would be better organized if it was directly after namespaces as it is in the struct

TheRushingWookie (commented):
Just a nit

charts/gateway-crds-helm/templates/generated/gateway.envoyproxy.io_securitypolicies.yaml
nareddyt: That would be nice, but unfortunately this is an auto-generated file. Generated by `controller-gen`.

I checked `controller-gen crd -ww` but could not find any options to preserve ordering. Looks like `controller-gen` puts it in alphabetical order.

internal/xds/translator/testdata/in/xds-ir/ext-auth.yaml
```diff
@@ -118,6 +118,14 @@ http:
           extAuth:
             name: securitypolicy/default/policy-for-gateway-1
             failOpen: true
+            metadataContextNamespaces:
+              - envoy.filters.http.custom1
```
wenfengp, line 122: Do you know where we define the `filters.http.custom{1...4}`?

nareddyt: It is a dummy value, doesn't exist anywhere.
In real production use cases, it would be a value from https://www.envoyproxy.io/docs/envoy/latest/configuration/advanced/well_known_dynamic_metadata

api/v1alpha1/ext_auth_types.go
```diff
@@ -106,3 +116,27 @@ type BodyToExtAuth struct {
...
+type ExtAuthAccessibleMetadata struct {
+	// Namespaces are metadata namespaces that are sent to the external authorization server as context.
+	//
+	// +optional
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// TypedNamespaces are typed metadata namespaces that are sent to the external authorization server as context.
+	//
+	// +optional
+	TypedNamespaces []string `json:"typedNamespaces,omitempty"`
```
guydc, line 131: Do we need both typed an untyped? I think that for ext-proc, we only support untyped at the moment.

api/v1alpha1/ext_auth_types.go
```diff
@@ -106,3 +116,27 @@ type BodyToExtAuth struct {
...
+
+	// RouteNamespaces are route-level metadata namespaces that are sent to the external authorization server as context.
+	//
+	// +optional
+	RouteNamespaces []string `json:"routeNamespaces,omitempty"`
+
+	// RouteTypedNamespaces are route-level typed metadata namespaces that are sent to the external authorization server as context.
+	//
+	// +optional
+	RouteTypedNamespaces []string `json:"routeTypedNamespaces,omitempty"`
```
guydc, line 141: same question regarding typed vs untyped

api/v1alpha1/ext_auth_types.go
```diff
@@ -54,6 +55,15 @@ type ExtAuth struct {
 	//
 	// +optional
 	RecomputeRoute *bool `json:"recomputeRoute,omitempty"`
+
+	// AccessibleMetadata defines options related to the sending of dynamic metadata.
```
guydc: dynamic and static (xds) metadata

test/cel-validation/securitypolicy_test.go
```diff
@@ -747,6 +747,31 @@ func TestSecurityPolicyTarget(t *testing.T) {
 			},
 			wantErrors: []string{"BackendRefs only supports Service and Backend kind."},
 		},
+		{
```
guydc, line 861: maybe a positive test as well for the grpc case

release-notes/current.yaml
```diff
@@ -30,6 +30,7 @@ new features: |
   Added support for client certificate validation (SPKI, hash, SAN) in ClientTrafficPolicy.
   Added support for OIDC RP initialized logout. If the end session endpoint is explicitly specified or discovered from the issuer's well-known url, the end session endpoint will be invoked when the user logs out.
   Added support for specifying deployment annotations through the helm chart.
+  Added support for specifying dynamic metadata namespaces that External Authorization services can access in SecurityPolicy API.
```
guydc: and static

api/v1alpha1/ext_auth_types.go
nareddyt: Hmm, I prefer if we can support both, as Confluent makes use of https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/set_metadata_filter to set some typed metadata. I see other OSS filters also set typed metadata (such as `gcp_authn`) as well as a few of Confluent's custom filters.

Supporting typed metadata essentially comes for free, though I understand the user experience might be confusing. WDYT @guydc?

api/v1alpha1/ext_auth_types.go
nareddyt: See https://github.com/envoyproxy/gateway/pull/6453#discussion_r2192866991

test/cel-validation/securitypolicy_test.go
nareddyt: Added in latest commit

nareddyt:
@guydc PTAL, addressed all review comments. Would like your input on our idea to expose both typed and untyped namespaces, as Confluent has a use case for typed namespaces

guydc (dismissed):
LGTM. 
Would be nice to extend the e2es to cover this feature, e.g. have grcp ext-auth consume metadata and add it to the request headers through the Check Response.

examples/envoy-ext-auth/main.go
```diff
@@ -99,19 +99,32 @@ func (s *authServer) Check(
...
+				routeMetadata = "unknown"
+			}
+
 			return &envoy_service_auth_v3.CheckResponse{
 				HttpResponse: &envoy_service_auth_v3.CheckResponse_OkResponse{
 					OkResponse: &envoy_service_auth_v3.OkHttpResponse{
 						Headers: []*envoy_api_v3_core.HeaderValueOption{
 							{
-								Append: &wrappers.BoolValue{Value: false},
+								AppendAction: envoy_api_v3_core.HeaderValueOption_OVERWRITE_IF_EXISTS_OR_ADD,
```
nareddyt, line 113: nit: field was deprecated, replaced with newer one

test/e2e/tests/ext_auth_grpc_service.go
```diff
@@ -66,6 +66,34 @@ var GRPCExtAuthTest = suite.ConformanceTest{
...
+				Response: http.Response{
+					StatusCode: 200,
+				},
+				ExpectedRequest: &http.ExpectedRequest{
+					Request: http.Request{
+						Host: "www.example.com",
+						Path: "/myapp",
+						Headers: map[string]string{
+							"x-current-user":             "user1",
+							"x-e2e-conformance-metadata": "static-untyped-route",
```
nareddyt, line 87: this is the new header we are expecting

nareddyt:
> LGTM. Would be nice to extend the e2es to cover this feature, e.g. have grcp ext-auth consume metadata and add it to the request headers through the Check Response.

Good idea, done. I tested static untyped route metadata in the e2e test. IMO there is no additional value is testing the other 3 fields (typed route metadata, typed + untyped listener metadata).

arkodg:
hey @nareddyt can we limit this feature to route metadata ? 
filters in the data plane may map to different resources which are tied to different actors, so prefer avoiding it until we've thought through all the implications

nareddyt:
@arkodg we do have a few use cases at Confluent where we need to propagate filter metadata from other OSS filters. We had a similar discussion in https://github.com/envoyproxy/gateway/pull/6453#discussion_r2192866991

Are there any specific implications you are worried about? I can limit it to route metadata if really required, but then we can no longer use this feature and will still need an extension server modification.

arkodg:
would A CEL validtion be tricky be add for this case ?
* allow filter metdata if target is Gateway, else dont

nareddyt:
> would A CEL validtion be tricky be add for this case ?
> * allow filter metdata if target is Gateway, else dont

Just trying to understand what that would solve?

IIRC we have precedent for filter metadata in EnvoyExtensionPolicy via ExtProc. https://gateway.envoyproxy.io/docs/api/extension_types/#extprocmetadata . This supports untyped filter metadata. And ExtProc can target both a HTTPRoute or a Gateway.

I didn't see any sort of similar validations in ExtProc CRD, so just trying to understand if they are necessary.

nareddyt:
Just want to clear up some confusion: Envoy allows attaching xDS metadata to multiple resources, such as routes, clusters, listeners. Envoy filters also can modify per-request dynamic metadata.

The ext_authz callout should allow sending all these types of metadata, regardless of whether the `SecurityPolicy` is targeting a Gateway vs HTTPRoute. That should be orthogonal

arkodg:
> Just want to clear up some confusion: Envoy allows attaching xDS metadata to multiple resources, such as routes, clusters, listeners. Envoy filters also can modify per-request dynamic metadata.
> 
> The ext_authz callout should allow sending all these types of metadata, regardless of whether the `SecurityPolicy` is targeting a Gateway vs HTTPRoute. That should be orthogonal

if a `SecurityPolicy` can target a `Gateway` its in the same namespace as the `Gateway`, and implies same persona is creating them.

Envoy Gateway does not gives direct access to xds filters, except in the `EnvoyProxy` resource to order xds filters which is applied at the `Gateway` level

There's `ExtensionServer` and `EnvoyExtensionPolicy` which is the back door enabled by the admin in the startup `EnvoyGateway` config

nareddyt:
Thanks @arkodg , I understand the concern now. Unfortunately this will not work for Confluent's use case. Let me close this PR and rethink it. Perhaps I will go with a custom extension server mutation on our end for the time being.

Issue #6456, opened 2025-07-03
support endpoint picker based on host override policy


Xunzhuo:
*Description*:

Support Endpoint Picker Based on the Host Override Policy:  https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/override_host/v3/override_host.proto

Host Override Policy selects the endpoint from the header or metadata, and lookup in the cluster endpoints. If there is no available endpoint selected, it will use the fallback policy to select an endpoint in the cluster endpoints.

Introducing a new loadbalancer type HostOverride in BackendTrafficPolicy to make this happen.

```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy
metadata:
  name: host-override-policy
  namespace: default
spec:
  targetRefs:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
      name:  host-override-route
  loadBalancer:
    type: HostOverride
    hostOverrideSettings:
      overrideHostSources:
      - header: target-pod
      - metadata:
          key: "envoy.lb"
          path:
          - key: "override-host"
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: static-resolver-route
  namespace: default
spec:
  parentRefs:
    - name: eg
  hostnames:
    - "www.example.com"
  rules:
    - 
      backendRefs:
        - kind: Service
           name: inference-pool-svc
           port: 3000
```

PR #6458, opened 2025-07-03
feat: support endpoint override policy based routing

Xunzhuo:
**What type of PR is this?**

feat: support host override policy based routing

Support host override policy based routing, a typical scenario is the LLM Endpoint Picker.

Fixes https://github.com/envoyproxy/gateway/issues/6456

Release Notes: Yes

Use in this way:

```
  kind: BackendTrafficPolicy
  metadata:
    namespace: default
    name: policy-for-header-override
  spec:
    targetRef:
      kind: HTTPRoute
      name: httproute
    loadBalancer:
      type: RoundRobin
      endpointOverride:
        extractFrom:
        - header: "x-gateway-destination-endpoint"
- apiVersion: gateway.networking.k8s.io/v1
  kind: HTTPRoute
  metadata:
    namespace: default
    name: httproute
  spec:
    hostnames:
    - gateway.envoyproxy.io
    parentRefs:
    - namespace: envoy-gateway
      name: inference-gateway
      sectionName: http
    rules:
    - matches:
      - path:
          value: "/v1"
      backendRefs:
      - name: fallback-inference-service
        port: 8080
```


Xunzhuo (commented):
Image

CI passed locally.

api/v1alpha1/loadbalancer_types.go
```diff
@@ -178,3 +188,68 @@ type ForceLocalZone struct {
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	OverrideHostSources []OverrideHostSource `json:"overrideHostSources"`
+
+	// FallbackPolicy defines the child LB policy to use in case neither header nor metadata with selected hosts is present.
+	// If not specified, defaults to LeastRequest.
+	//
+	// +optional
+	// +kubebuilder:default="LeastRequest"
+	FallbackPolicy *LoadBalancerType `json:"fallbackPolicy,omitempty"`
```
wbpcode: IMO, I will perfer the HostOverrideSettings as a common enhancement to all exist lb policy. It self needn't to be treated as an independent lb policy in the control plane.

That's say, we needn't the `FallbackPolicy` and the `HostOverride` policy. If the optional `HostOverrideSettings` is configured, then we can think the override lb policy of data plane is used, and the original `LoadBalancerType` , etc. could be used to construct the fallback policy of data plane. This should could simplify our API and avoid the users to configure nested override host lb.

I am not expert of the the gateway/control plane, only a comment from the point of the API.

cc @Xunzhuo @arkodg

arkodg: sure I like this approach

Xunzhuo: +1 for this

arkodg: piggybacking off this
```
endpointOverride:
 - header: x-custom-host
```

or 

```
endpointOverride:
  from:
   - header: x-custom-host
```

which is similar to 
* https://github.com/envoyproxy/gateway/blob/173707832719749dbaf91bc6d26ec8cb14a6b8d7/api/v1alpha1/api_key_auth_types.go#L24
* https://github.com/envoyproxy/gateway/blob/173707832719749dbaf91bc6d26ec8cb14a6b8d7/api/v1alpha1/jwt_types.go#L86

arkodg: we use `endpoint` instead of `host` in the API, so also suggested that change

Xunzhuo: then i am ok with option 2

arkodg: cool lets use `extractFrom` to keep the API surface similar

Xunzhuo: Working on it

api/v1alpha1/loadbalancer_types.go
```diff
@@ -178,3 +186,57 @@ type ForceLocalZone struct {
...
+	// The IPv6 address is enclosed in square brackets.
+	//
+	// +optional
+	Header *string `json:"header,omitempty"`
+
+	// Metadata defines the metadata key to get the override endpoint addresses from the request dynamic metadata.
+	// If set this field then it will take precedence over the header field.
+	//
+	// +optional
+	Metadata *EndpointOverrideMetadataKey `json:"metadata,omitempty"`
```
arkodg: can we reuse https://github.com/envoyproxy/gateway/blob/f8054b15bf592832d7a3687cc32e8823817e2232/api/v1alpha1/ratelimit_types.go#L172

Xunzhuo: should move it into shared types to reuse

Xunzhuo: Emmm I checked that, they are in different cases: endpoint override is to build metadatav3.MetadataKey

Ratelimit is to build the routev3.RateLimit_HitsAddend, they have different format input

arkodg: hey @wbpcode @mathetake curious why 2 different metadata types exist ?

arkodg: and do we need to add this to the API ? 
will ext proc write this value ? if so extProc's API in EG https://github.com/envoyproxy/gateway/blob/64f3576cffa892acdf3af396b51b29b7a28cd4cc/api/v1alpha1/ext_proc_types.go#L113

uses the term `writableNamespaces`

api/v1alpha1/loadbalancer_types.go
Xunzhuo: Clipboard_Screenshot_1752206423

wanna make sure the logics in envoy, take the picture above as an example: how does envoy pick the endpoint, from the index 0 to the end, if found then break the selection?

cc @wbpcode @yanavlasov

Xunzhuo:
## Endpoint Picker General Implementation Logics

This is how we generally implemented the EPP logics in Envoy based API Gateway, no matter the control plane is Envoy Gateway, Istio, or KGateway:

1. Control Plane: it tells envoy how to route (host override lbpolicy or original cluster) and tells envoy how to connect to the epp ext-proc (http ext-proc filter + route level epp ext-proc config override, if the extproc need to read/write the metadata we should also set receiving_namespaces/forwarding_namespaces at ext-proc config)
2. Data Plane: epp ext proc selects the endpoint and adding it to metadata or header. Envoy routes to that endpoint based on the control plane sent rules.

## Envoy Original Dst Cluster vs Host Override LbPolicy

[Original Dst Cluster](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/original_dst): It is easy to implement and don't need the real cluster endpoints. But it does not support fallback, which means if the selection is failed, the routing will fail immediately.

[Host Override LbPolicy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/override_host#arch-overview-load-balancing-override-host): It is a bit complexer than original dst cluster to implement, it requires the real cluster endpoints, and the selected endpoint should be in the endpoints, otherwise it will fallback. So when Gateway implements the InferencePool with host override lbpolicy, we usually need a real service selects the inference workload endpoints, and the host override lbpolicy is working on the kubernetes service, and the endpoint selection logics in EndpointPicker should also select the endpoints in the same endpoints (Istio creates a service with the same labels selectors with the InferencePool Selectors)

## How to implement the Endpoint Picker logics in Envoy Gateway?

Different AI Gateway based on Envoy Gateway has different approaches to reach the above goal:

### Envoy AI Gateway: use Envoy Gateway Extension Server. 

It edits cluster, route, listener to make this work, this is quite challenging since it is a complex work, which need to work well with the existing config. This is not suitable for adopters like AIBrix.

The default EPP implementation is GIE.

### AIBrix Inference Gateway: use Envoy Gateway CRD configuration.

The EPP imlementation is AIBrix Gateway Plugin.(Similar to GIE, it provides intelligently endpoint picker)

1. v1 (currently): use envoy gateway EnvoyPatchPolicy (patch original cluster config) + EnvoyExtensionPolicy (add epp ext-proc config to gateway), this is static and not easy to maintain or orchestrate.
2. v2 (planning): use envoy gateway btp (add host override lb policy) + eep (add epp ext-proc config to gateway, also add receiving/forwarding ns with 'envoy.lb' if needed), this can largely improve UX, and also add fallback abilities to the GW.
4. v3 (after v2): use controller to automatically do what we configure manually in v2, and support InferencePool API. Simplify UX and can adopt GIE conformance test.

api/v1alpha1/loadbalancer_types.go
zhaohuabing: The term `key` is quite confusing as it can mean two different things in EG and Envoy API.
* The `namespace` of the metadata, usually the filter name or sub-system of envoy.
* A `path` segment in the metadata value struct.

I suggest we make it clearer by using the terms `namespace` and `path`.

We should unify how value is extracted from the dynamic metadata using a common structure:

```go
type Metadata struct {

	Namespace string `json:"namespace"`

	Path []string `json:"path,omitempty" yaml:"path,omitempty"`

	// Deprecated, please use Path instead
	Key string `json:"key" yaml:"key"`
}
```

For example, to extract value from this dynamic metata:

```json
"com.test.my_filter": 
{
  "test_key": "foo", 
  "test_object":
  {
    "inner_key": "bar"
  }
}
````

1.

```go
Metadata{
	Namespace: " com.test.my_filter"
	Path: "test_object"
}
```

value: {"inner_key": "bar"}

2.

```go
Metadata{
	Namespace: " com.test.my_filter"
	Path: {"test_object", "inner_key"}
}
```

value: "bar"

Reference: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-dynamic-metadata

internal/ir/xds.go
```diff
@@ -2496,6 +2496,11 @@ type LoadBalancer struct {
 	Random *Random `json:"random,omitempty" yaml:"random,omitempty"`
 	// ConsistentHash load balancer policy
 	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty" yaml:"consistentHash,omitempty"`
+	// EndpointOverride defines the configuration for endpoint override.
+	// When specified, the load balancer will attempt to route requests to endpoints
+	// based on the override information extracted from request headers or metadata.
+	// If no valid override endpoint is found, the configured load balancer policy will be used as fallback.
```
zhaohuabing: Nit:
```suggestion
	//  If the override endpoints are not available, the configured load balancer policy will be used as fallback.
```

api/v1alpha1/loadbalancer_types.go
```diff
@@ -29,6 +29,14 @@ type LoadBalancer struct {
 	// +optional
 	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty"`
 
+	// EndpointOverride defines the configuration for endpoint override.
+	// When specified, the load balancer will attempt to route requests to endpoints
+	// based on the override information extracted from request headers or metadata.
+	// If no valid override endpoint is found, the configured load balancer policy will be used as fallback.
```
zhaohuabing: Nit:
```suggestion
	//  If the override endpoints are not available, the configured load balancer policy will be used as fallback.
```

api/v1alpha1/loadbalancer_types.go
Xunzhuo: So should we move a shared type of Metadata, which is using both in RL and host override? And add a cel for not configuring key and path at the same time? (key = strings.Join(path,"."))

```
type Metadata struct {

	Namespace string `json:"namespace"`

	Path []string `json:"path,omitempty" yaml:"path,omitempty"`

	// Deprecated, please use Path instead
	Key string `json:"key" yaml:"key"`
}
```

zhaohuabing: > (key = strings.Join(path,"."))

`key = strings.Join(path,":"))` when used with %DYNAMIC_METADATA%

> %DYNAMIC_METADATA(com.test.my_filter:test_object:inner_key)% will log: bar

Xunzhuo: Got it.

arkodg: @zhaohuabing does `path` need to be a `[]` ?

api/v1alpha1/loadbalancer_types.go
```diff
@@ -178,3 +186,57 @@ type ForceLocalZone struct {
...
+// EndpointOverride defines the configuration for endpoint override.
+// This allows endpoint picking to be implemented based on request headers or metadata.
+// It extracts selected override endpoints from the specified sources (request headers, metadata, etc.).
+// If no valid endpoint in the override list, then the configured load balancing policy is used as fallback.
+type EndpointOverride struct {
+	// ExtractFrom defines the sources to extract endpoint override information from.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
+	ExtractFrom []EndpointOverrideExtractFrom `json:"extractFrom"`
```
wbpcode, line 199: Will EndpointOverrideSources (field name) and `EndpointOverrideSource` (type name) better? Then the naming will align with the data plane.

api/v1alpha1/loadbalancer_types.go
wbpcode: > hey @wbpcode @mathetake curious why 2 different metadata types exist ?

The `RateLimitCostMetadata` should be a control plane abstractino that used to construct the `hits_addend`. The Envoy only provide a `MetadataKey` to access the `Metadata`.

But as @zhaohuabing said, `key` is confusing. The Envoy use `key` for both metadata namespace and metadata key/path. So, sound good to me to reuse the Envoy `MetadataKey` or create a new shared variant for EG.

wbpcode: > @zhaohuabing does `path` need to be a `[]` ?

As shared `MetadataKey`. I think `[]` will provide more flexibility.

zhaohuabing: > @zhaohuabing does path need to be a [] ?

Not necessarily, but it's semantically clearer than something like `a:b:c:d`.

api/v1alpha1/loadbalancer_types.go
arkodg: the word `endpoint` is repeated  `endpointOverride.endpointOverrideSources` which shouldnt be required since the parent structure already has it

zhaohuabing:
ping @Xunzhuo

api/v1alpha1/loadbalancer_types.go
```diff
@@ -12,7 +12,7 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 //
 // +kubebuilder:validation:XValidation:rule="self.type == 'ConsistentHash' ? has(self.consistentHash) : !has(self.consistentHash)",message="If LoadBalancer type is consistentHash, consistentHash field needs to be set."
 // +kubebuilder:validation:XValidation:rule="self.type in ['Random', 'ConsistentHash'] ? !has(self.slowStart) : true ",message="Currently SlowStart is only supported for RoundRobin and LeastRequest load balancers."
-// +kubebuilder:validation:XValidation:rule="self.type == 'ConsistentHash' ? !has(self.zoneAware) : true ",message="Currently ZoneAware is only supported for LeastRequest, Random, and RoundRobin load balancers."
+// +kubebuilder:validation:XValidation:rule="self.type in ['ConsistentHash'] ? !has(self.zoneAware) : true ",message="Currently ZoneAware is only supported for LeastRequest, Random, and RoundRobin load balancers."
```
arkodg: is this needed in this PR?

arkodg: cc  @jukie

internal/xds/translator/cluster.go
```diff
@@ -232,103 +233,122 @@ func buildXdsCluster(args *xdsClusterArgs) (*buildClusterResult, error) {
...
 	//nolint:gocritic
-	switch {
-	case args.loadBalancer == nil:
-		cluster.LbPolicy = clusterv3.Cluster_LEAST_REQUEST
-		leastRequest := &least_requestv3.LeastRequest{
-			LocalityLbConfig: localityLbConfig,
-		}
-		typedLeastRequest, err := proto.ToAnyWithValidation(leastRequest)
+	// Check if EndpointOverride is specified
+	if args.loadBalancer != nil && args.loadBalancer.EndpointOverride != nil {
```
zirain: why we add `if-else` here instead of adding a new switch-case closure?

test/e2e/tests/load_balancing.go
```diff
@@ -358,3 +361,159 @@ var ConsistentHashCookieLoadBalancingTest = suite.ConformanceTest{
...
+			req := http.MakeRequest(t, &expectedResponse, gwAddr, "HTTP", "http")
+
+			// Make multiple requests and verify fallback behavior (just check 200 response)
+			for i := 0; i < sendRequests; i++ {
+				cReq, cResp, err := suite.RoundTripper.CaptureRoundTrip(req)
+				require.NoError(t, err, "failed to get expected response")
+				require.NoError(t, http.CompareRequest(t, &req, cReq, cResp, expectedResponse), "failed to compare request and response")
+
+				// For invalid override host, we just verify the response is 200 (fallback works)
+				// No need to check specific pod routing since it should use fallback policy
```
zirain, line 490: should we verify that lb policy fallback to RoundRobin?

test/e2e/tests/load_balancing.go
```diff
@@ -358,3 +361,159 @@ var ConsistentHashCookieLoadBalancingTest = suite.ConformanceTest{
...
+			req := http.MakeRequest(t, &expectedResponse, gwAddr, "HTTP", "http")
+
+			// Make multiple requests and verify fallback behavior (just check 200 response)
+			for i := 0; i < sendRequests; i++ {
+				cReq, cResp, err := suite.RoundTripper.CaptureRoundTrip(req)
+				require.NoError(t, err, "failed to get expected response")
+				require.NoError(t, http.CompareRequest(t, &req, cReq, cResp, expectedResponse), "failed to compare request and response")
+
+				// For missing header, we just verify the response is 200 (fallback works)
+				// No need to check specific pod routing since it should use fallback policy
```
zirain, line 517: should we verify that lb policy fallback to RoundRobin?

Xunzhuo: we could do that when adding more e2e tests coverage, currently we covered if the core fallback workes

api/v1alpha1/loadbalancer_types.go
jukie: Isn't necessary for this PR but effectively a no-op so I don't feel too strongly about dropping it.

test/e2e/tests/load_balancing.go
zirain: we can track this in a separated issue.

zirain (approved):

test/e2e/tests/load_balancing.go
Xunzhuo: https://github.com/envoyproxy/gateway/issues/6634

arkodg (approved):

arkodg:
seeing consistent test failures @Xunzhuo 
```
    --- FAIL: TestE2E/EndpointOverrideLoadBalancing (4.26s)
        --- FAIL: TestE2E/EndpointOverrideLoadBalancing/header-based_endpoint_override_with_valid_pod_IP_should_route_to_specific_pod (1.02s)
 ```
 
 can we skip this for now, and track it with a GH issue

**Comment by Xunzhuo** [comment]
@arkodg it happened only when ipv6?

**Comment by arkodg** [comment]
> @arkodg it happened only when ipv6?

yeah look like it, but it passes on dual

File: test/e2e/tests/load_balancing.go [hunk 10]
```diff
@@ -358,3 +361,159 @@ var ConsistentHashCookieLoadBalancingTest = suite.ConformanceTest{
...
+		require.NoError(t, err, "failed to get service")
+		require.NotEmpty(t, svc.Spec.Ports, "service should have ports")
+		servicePort := svc.Spec.Ports[0].TargetPort.IntValue()
+
+		t.Logf("Found %d valid pods: %v, service port: %d", len(validPodIPs), podIPToName, servicePort)
+
+		t.Run("header-based endpoint override with valid pod IP should route to specific pod", func(t *testing.T) {
+			// Use the first valid pod IP as override host
+			targetPodIP := validPodIPs[0]
+			overrideHost := fmt.Sprintf("%s:%d", targetPodIP, servicePort)
```
zirain: can you change this to [%s]:%d in IPv6?

zirain:
```
  /**
   * @return a human readable string for the address that represents the
   * physical/resolved address. (This will not necessarily include port
   * information, if applicable, since that may not be resolved until bind()).
   *
   * This string will be compatible with the following example formats:
   * For IPv4 addresses: "1.2.3.4:80"
   * For IPv6 addresses: "[1234:5678::9]:443"
   * For pipe addresses: "/foo"
   */
  virtual const std::string& asString() const PURE;
```

zirain:
@Xunzhuo please fix gen-check

Issue #6459, opened 2025-07-03
EnvoyProxy with host/port not working


zirain:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: eg
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: eg
spec:
  gatewayClassName: eg
  listeners:
    - name: http
      protocol: HTTP
      port: 80
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: eg-gtw-config
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: eg-gtw-config
spec:
  telemetry:
    accessLog:
      settings:
        - format:
            type: JSON
            json:
              start_time: "%START_TIME%"
              method: "%REQ(:METHOD)%"
              "x-envoy-origin-path": "%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%"
              protocol: "%PROTOCOL%"
              response_code: "%RESPONSE_CODE%"
              response_flags: "%RESPONSE_FLAGS%"
              response_code_details: "%RESPONSE_CODE_DETAILS%"
              upstream_transport_failure_reason: "%UPSTREAM_TRANSPORT_FAILURE_REASON%"
              request_headers_bytes: "%REQUEST_HEADERS_BYTES%"
              bytes_received: "%BYTES_RECEIVED%"
              bytes_sent: "%BYTES_SENT%"
              duration: "%DURATION%"
              upstream_duration: "%COMMON_DURATION(US_TX_BEG:US_RX_END:ms)%"
              "x-envoy-upstream-service-time": "%RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)%"
              "x-forwarded-for": "%REQ(X-FORWARDED-FOR)%"
              "user-agent": "%REQ(USER-AGENT)%"
              "x-request-id": "%REQ(X-REQUEST-ID)%"
              ":authority": "%REQ(:AUTHORITY)%"
              authorization: "%REQ(AUTHORIZATION)%"
              upstream_host: "%UPSTREAM_HOST%"
              upstream_cluster: "%UPSTREAM_CLUSTER%"
              upstream_local_address: "%UPSTREAM_LOCAL_ADDRESS%"
[...]

PR #6460, opened 2025-07-03
fix telemetry with host port not working

zirain:
follow up: https://github.com/envoyproxy/gateway/pull/6370, so we don't need another release notes.

fixes: https://github.com/envoyproxy/gateway/issues/6459


zirain:
cc @shawnh2

shawnh2 (approved):

Issue #6457, opened 2025-07-03
support endpoint picker based on original dst cluster


Xunzhuo:
*Description*:

Support Endpoint Picker Based on the Original Dst Cluster: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/original_dst#arch-overview-load-balancing-types-original-destination

Original Dst Cluster selects the endpoint from the header or metadata, and directly foward it to the Endpoint.

Introducing a new backend type OriginalDestination in Backend to make this happen.

header:
```
 apiVersion: gateway.envoyproxy.io/v1alpha1
  kind: Backend
  metadata:
    name: backend-routing-based-on-header
    namespace: default
  spec:
    type: OriginalDestination
      originalDestinationSetting:
        metadata:
          key: "envoy.lb"
          path:
          - key: "override-host"
```

metadata:
```
 apiVersion: gateway.envoyproxy.io/v1alpha1
  kind: Backend
  metadata:
    name: backend-routing-based-on-header
    namespace: default
  spec:
    type: OriginalDestination
      originalDestinationSetting:
         header: target-pod
```

shiavm006:
i have worked on implementing Original Destination Backend support for Envoy Gateway, and I believe this addresses a critical gap in the project's current capabilities that aligns perfectly with the stated goals in the design documentation.would like to know ur vies on it

arkodg:
@shiavm006 this is already being worked on by @Xunzhuo with #6234

shiavm006:
> [@shiavm006](https://github.com/shiavm006) this is already being worked on by [@Xunzhuo](https://github.com/Xunzhuo) with [#6234](https://github.com/envoyproxy/gateway/issues/6234)

so can i work with him ?

Xunzhuo:
You can pick this one : ) If you are interested in.

shiavm006:
> You can pick this one : ) If you are interested in.
thanks 

for this one like i had post a pr i would love if u review and and suggest me if its not right .... #6474

Xunzhuo:
The only thing left for #6234 is the API changes, so I am not sure if you need to redo it from the scratch.

So my suggestion is to pick another issue for u, if you are interested.

And suggest that if you want to pick one issue, maybe should check if it has been assigned, in case you spend some same time with others.

shiavm006:
ok sure will make sure of that 
thanks for the acknowlegdement i will look for another issue than

PR #6474, opened 2025-07-07
feat: add Original Destination Backend support

shiavm006:
This PR implements support for Original Destination Backends in Envoy Gateway, enabling header-based dynamic routing with security controls.

issue being fixed :  #6457

implemented Original Destination Backend support in Envoy Gateway to unlock dynamic routing capabilities that were previously impossible with static backend configurations. This feature directly addresses the project's core goal of making Envoy's advanced capabilities accessible through simple, expressive APIs.

## Testing

- [x] All existing tests pass
- [x] New functionality thoroughly tested
- [x] Manual testing with example configurations
- [x] XDS cluster generation validated

## Related Documentation

- [Envoy Original Destination Load Balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/original_dst)
- [Envoy Gateway Backend API](https://gateway.envoyproxy.io/latest/api/extension_types/#backend)

what r ur views on it


api/v1alpha1/backend_types.go
```diff
@@ -123,41 +123,41 @@ type UnixSocket struct {
...
+	// +kubebuilder:validation:Enum=Endpoints;DynamicResolver;OriginalDestination
 	// +kubebuilder:default=Endpoints
 	// +optional
 	Type *BackendType `json:"type,omitempty"`
 
 	// Endpoints defines the endpoints to be used when connecting to the backend.
 	//
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:MaxItems=4
```
wbpcode: Why this change is necessary?

api/v1alpha1/backend_types.go
```diff
@@ -123,41 +123,41 @@ type UnixSocket struct {
...
 	Endpoints []BackendEndpoint `json:"endpoints,omitempty"`
 
-	// AppProtocols defines the application protocols to be supported when connecting to the backend.
+	// FQDN defines the FQDN used to contact the backend.
 	//
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 	// +optional
-	AppProtocols []AppProtocolType `json:"appProtocols,omitempty"`
+	FQDN *string `json:"fqdn,omitempty"`
```
wbpcode: I am not very familiar to the gateway, but I think this is a break change to the API?

api/v1alpha1/backend_types.go
shiavm006: Original destination routing resolves endpoints dynamically from request headers, making 64 static endpoints contradictory to its core purpose. Testing revealed this created unnecessary performance overhead without serving realworld use cases, which typically need only 2 4 endpoints for fallbacks. This constraint reflects good engineering  refining the API based on practical testing rather than theoretical assumptions 
hope u find it well

api/v1alpha1/backend_types.go
shiavm006: ya u were right about it but now i have fixed it whats ur point on it .?

PR #6478, opened 2025-07-07
feat(translator): remove absent healthy endpoints

### LINKED ISSUE #6463: Ignore Host Health for DNS clusters
*Description*:

Currently, if an upstream has active HCs enabled and the HCs succeed for a specific endpoint, that endpoint would not be removed, even if it's no longer returned by the discovery mechanism. See discussion here: https://github.com/envoyproxy/envoy/pull/3302. 

EG disables this behavior for EDS clusters following this issue: https://github.com/envoyproxy/gateway/issues/4564. 

This option is also beneficial for DNS clusters. For example, users may want to steer traffic towards a specific endpoint by making changes to DNS records, without the need to also actively fail health checks. Other users may prefer the current behavior, which is better suited for handling issues in eventually consistent discovery. 

EG BackendTrafficPolicy can expose this option, possibly in the HealthChecks struct.
 

[optional *Relevant Links*:]

guydc:
**What type of PR is this?**

Removes absent (no longer discovered) endpoints, without waiting for their health to fail as well. 

Fixes #6463 

Release Notes: Yes

### DIFF
--- a/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.json
+++ b/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.json
@@ -386,6 +386,7 @@
                 "connectTimeout": "10s",
                 "dnsLookupFamily": "V4_PREFERRED",
                 "dnsRefreshRate": "30s",
+                "ignoreHealthOnHostRemoval": true,
                 "lbPolicy": "LEAST_REQUEST",
                 "loadAssignment": {
                   "clusterName": "raw_githubusercontent_com_443",
--- a/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.all.yaml
@@ -218,6 +218,7 @@ xds:
           connectTimeout: 10s
           dnsLookupFamily: V4_PREFERRED
           dnsRefreshRate: 30s
+          ignoreHealthOnHostRemoval: true
           lbPolicy: LEAST_REQUEST
           loadAssignment:
             clusterName: raw_githubusercontent_com_443
--- a/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.cluster.yaml
+++ b/internal/cmd/egctl/testdata/translate/out/jwt-single-route-single-match-to-xds.cluster.yaml
@@ -44,6 +44,7 @@ xds:
         connectTimeout: 10s
         dnsLookupFamily: V4_PREFERRED
         dnsRefreshRate: 30s
+        ignoreHealthOnHostRemoval: true
         lbPolicy: LEAST_REQUEST
         loadAssignment:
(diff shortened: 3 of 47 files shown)


api/v1alpha1/healthcheck_types.go
```diff
@@ -132,6 +132,16 @@ type ActiveHealthCheck struct {
 	// It's optional, and can only be used if the specified type is GRPC.
 	// +optional
 	GRPC *GRPCActiveHealthChecker `json:"grpc,omitempty" yaml:"grpc,omitempty"`
+
+	// RemoveAbsentHealthyEndpoints determines if endpoints that were previously discovered and are now absent from
```
arkodg: can we do without an API field ?
are there use cases where the API has removed the endpoint, but we want the data plane to keep using it ?

guydc: The envoy rationale for this (default) behavior is outlined [here](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#on-eventually-consistent-service-discovery): 

>**Host absent / health check OK:**
Envoy will route to the target host. This is very important since the design assumes that the discovery service can fail at any time. If a host continues to pass health check even after becoming absent from the discovery data, Envoy will still route. Although it would be impossible to add new hosts in this scenario, existing hosts will continue to operate normally. When the discovery service is operating normally again the data will eventually re-converge.

When envoy depends on successful DNS resolution rather than a control plane to discover endpoints, there's a greater risk, and some users may prefer to route to stale endpoints, as long as they are known to be healthy, as mitigation to discovery failures.

arkodg: if the discovery service is not providing endpoints but DNS is, how will this setting help ?

guydc: Discovery works differently in these cases. XDS-based discovery is inherently more resilient (control plane is typically a local component, xds has a bunch of built-in guardrails like caching). There are several examples of DNS-based resolution failures that lead to traffic drops:
- https://github.com/envoyproxy/envoy/issues/37676 
- https://github.com/envoyproxy/envoy/issues/2691 

I agree that the envoy default is not intuitive. We can enable this by default and see if users are interested in the other option as time goes on.

guydc: I tested out envoy's behavior when DNS resolution fails due to network connectivity issues. When IgnoreHealthOnHostRemoval is true, endpoints are not being dropped. I'm not sure if this covers all possible edge-cases for DNS resolution (e.g. NXDOMAIN, NODATA results), but I feel fine with only changing the default for the time being.

guydc: Tested out a few behaviors:
- NXDOMAIN: in both cases endpoints are removed (even when considering health)
- Timeout to DNS: in both cases endpoints are retained (even when ignoring health)

I'm fine with changing the default behavior for STRICT_DNS as well.

arkodg: thanks for testing this out @guydc 

+1 on setting this by default, and considering adding a field to opt if if users really need it

arkodg (approved):
LGTM thanks

PR #6482, opened 2025-07-08
DRAFT: Implement ZoneAware loadbalancing

jukie:
Breaking this up into multiple PRs which can be reviewed/merged in any order:
* https://github.com/envoyproxy/gateway/pull/6485 - Implements internal logic to support explicit configuration for Envoy LocalityLbConfig.
* https://github.com/envoyproxy/gateway/pull/6497 - Updates the bootstrap local_cluster name to non-static value based on infra name.
* TODO: Render the EnvoyProxy cluster via xDS


internal/ir/xds.go
```diff
@@ -2170,9 +2180,13 @@ type GlobalResources struct {
...
 	// control plane components. For example, the rate limit service, WASM HTTP server, etc.
 	EnvoyClientCertificate *TLSCertificate `json:"envoyClientCertificate,omitempty" yaml:"envoyClientCertificate,omitempty"`
+
 	// HMACSecret holds the HMAC Secret used by the OIDC.
 	// TODO: zhaohuabing move HMACSecret here
 	// HMACSecret PrivateBytes
+
+	// ProxyInfraCluster holds cluster information for the ProxyInfra pods
+	ProxyInfraCluster *ProxyInfraCluster `json:"proxyInfraCluster,omitempty" yaml:"proxyInfraCluster,omitempty"`
 }
```
jukie, line 2213: Should this be under GlobalResources or inside a dedicated field?

Is `ProxyInfraCluster` a reasonable name? This aligns with the `ProxyInfra` naming convention but open to other ideas. It's not a user-facing API field but would like it to make sense for other contributors.

Issue #6025, opened 2025-05-13
Explicit settings for Zone Aware Routing


arkodg:
> I'd also like to expose the configuration for zone aware lb and weighted localities under BTP load balancer settings and have a [WIP PR](https://github.com/envoyproxy/gateway/pull/5938) I plan to bring up at the contributors call.
> 
> Something like this is what I'd like to have available:
> ```
> apiVersion: gateway.envoyproxy.io/v1alpha1
> kind: BackendTrafficPolicy
> metadata:
>   name: btp
> spec:
>   loadBalancer:
>     requestDistribution:
>       type: PreferLocalZone|WeightedZone
>       preferLocalZoneConfig: # maps to ZoneAwareLbConfig and overrides Service config (e.g. TrafficDistribution)
>         forceLocalZone: true
>         forceLocalZoneMinZoneSize: 1
>         minClusterSize: 1
>       weightedZoneConfig: # maps to LocalityWeightedLbConfig and weights will be considered during buildXdsClusterLoadAssignment() 
>         weights:
>           zone1: 25
>           zone2: 50
>           zone3: 25
> ```
> 
> I have two main use cases in my environment:
> * Support for the default zone aware routing config - We run Kubernetes in physical datacenters which can span hundreds of racks/zones. I'd like to set a preference that traffic stays local but I don't want to require running an Envoy pod on every rack in order to serve traffic to backends there. In most cases this will be true but I'd like the Envoy dataplane pods to have awareness of other locality zones and be able to distribute load evenly. In my PR I've added support to track the Envoy ProxyInfra and render an XDS cluster for this to work.
> * Ability to directly set weights per locality zone - In situations with uneven incoming traffic between zones or a separate set of loadbalancing infra is present that Envoy isn't aware of it'd be great if as a user I was able to adjust the zone weights accordingly. 

 _Originally posted by @jukie in [#1909](https://github.com/envoyproxy/gateway/issues/1909#issuecomment-2855166707)_

jukie:
Opened a PR for the API side with some tweaks:
```
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: BackendTrafficPolicy
metadata:
  name: btp
spec:
  loadBalancer:
    requestDistribution: # only one of preferLocalZone|weightedLocality can be set
      preferLocalZone: # maps to ZoneAwareLbConfig and overrides Service config (e.g. TrafficDistribution)
        forceLocalZone:
          enabled: true
          minZoneSize: 1
[...]

PR #6485, opened 2025-07-08
 feat: Implement ZoneAware loadbalancing - ZoneAware Lb Config (splitup #6482)

jukie:
**What type of PR is this?**

feat: Implement ZoneAware Lb Config

Implementation for `LoadBalancer.ZoneAware` API (Ref: https://github.com/envoyproxy/gateway/pull/6154)

This adds the internal logic to support explicit configuration for Envoy LocalityLbConfig. Leaving the API marked as hidden/not-implemented because this will depend on a followup for tracking the EnvoyProxy Endpoints in xDS but splitting those changes out since it's a fairly large change (See https://github.com/envoyproxy/gateway/pull/6482).

This also replaces `DestinationSetting.ZoneAwareRouting` with `DestinationSetting.PreferLocal` to consolidate under the new API field.

xRef #6025 

Release Notes: No
 (will add release note for whichever PR gets merged last)


jukie:
CC @rudrakhp @arkodg as you reviewed the API PR

internal/ir/xds.go
```diff
@@ -2513,6 +2521,8 @@ type LoadBalancer struct {
 	Random *Random `json:"random,omitempty" yaml:"random,omitempty"`
 	// ConsistentHash load balancer policy
 	ConsistentHash *ConsistentHash `json:"consistentHash,omitempty" yaml:"consistentHash,omitempty"`
+	// ZoneAware defines the configuration related to the distribution of requests between locality zones.
+	ZoneAware *ZoneAware `json:"zoneAware,omitempty" yaml:"zoneAware,omitempty"`
```
arkodg: is this field needed in this iteration ?
trying to understand why are we not consolidating the two fields - `PreferLocal` and `ZoneAware.PreferLocal`

internal/xds/translator/cluster.go
```diff
@@ -439,6 +421,51 @@ func buildXdsCluster(args *xdsClusterArgs) (*buildClusterResult, error) {
...
+		LocalityConfigSpecifier: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig_{
+			LocalityWeightedLbConfig: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig{},
+		},
+	}
+
+	// Check for LoadBalancer.ZoneAware configuration or if backendRef enables
+	// PreferLocal (such as Topology Aware Routing or Traffic Distribution)
+	if zoneAware := ptr.Deref(args.loadBalancer, ir.LoadBalancer{}).ZoneAware; zoneAware != nil {
+		// Will later include additional case for weighted localities
+		if zoneAware.PreferLocal != nil {
```
arkodg: can we combine these two into 1 ?

internal/ir/xds.go
jukie: Could you elaborate? Consolidating these two isn't really an option.
```
zoneAware: # maps to Envoy LocalityLbConfig
  preferLocal: {} # maps to Envoy ZoneAwareLbConfig
  # not implemented yet
  weighted: {} # maps to localityWeightedLbConfig
```

jukie: Context: https://github.com/envoyproxy/gateway/pull/6154#discussion_r2127236959

internal/xds/translator/cluster.go
jukie: Will use the other discussion thread but no they cannot

internal/ir/xds.go
arkodg: I meant removing `preferLocal` in favor of `zoneAware.preferLocal` and `zoneAware.Weighted` in the future

jukie: Specifically in the internal ir logic? Nope I can remove that if preferred but it makes the future implementation of `weighted` easier to approach.

I'd rather not change the API field twice though

arkodg: Yeah specifically for IR, you can start reusing the newer umbrella field

jukie: Updated

internal/xds/translator/cluster.go
```diff
@@ -439,6 +421,48 @@ func buildXdsCluster(args *xdsClusterArgs) (*buildClusterResult, error) {
...
+	// Default to LocalityWeightedLbConfig
+	localityLbConfig := &commonv3.LocalityLbConfig{
+		LocalityConfigSpecifier: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig_{
+			LocalityWeightedLbConfig: &commonv3.LocalityLbConfig_LocalityWeightedLbConfig{},
+		},
+	}
+
+	// Check for LoadBalancer.PreferLocal configuration or if backendRef enables
+	// PreferLocal (such as Topology Aware Routing or Traffic Distribution)
+	if preferLocal := ptr.Deref(args.loadBalancer, ir.LoadBalancer{}).PreferLocal; preferLocal != nil {
```
arkodg, line 434: looks like priority is given to BTP settings over Service settings ?
currently in the project, most specific setting wins, so in this case, Service settings > BTP settings

jukie: Yes, but in my view was thinking the opposite where BTP would be most specific. Shouldn't the EnvoyProxy CRD configurations always take precedence? That's a Gateway/Route owner explicitly configuring the loadbalancer behavior vs inheritance from the backendRef.

jukie: There's definitely use-cases where multiple loadbalancer implementations might target the same backendRef or even generic Kubernetes routing. In those situations, if the different loadbalancers have different owners and want different configurations then using the loadbalancer implementation's configuration would be the right place in my opinion.

arkodg: yeah those are good arguments, lets get this in so the feature goes in, and revisit this later

PR #6488, opened 2025-07-09
feat: admin and metrics server hot reloading on config change

### LINKED ISSUE #6340: Enabling pprof requires rollout restart to be reflected
*Description*:
When enabling pprof by changing `admin.enablePprof` to `true` in EG configmap a rollout restart is needed in order for the change to be reflected.

*Repro steps*:
* Deploy EG with default configuration into k8s cluster - pprof is disabled.
* Enable pprof by patching EG configmap with `admin.enablePprof=true`. 
* Port-forward to EG pod admin server at port 19000.
* Try to access pprof endpoint at `http://localhost:19000/debug/pprof/` - not accessible.
* Perform rollout restart to EG deployment - now  `http://localhost:19000/debug/pprof/` will be accessible.

Looks like the root cause of this issue is the fact that admin server is only [initialized](https://github.com/envoyproxy/gateway/blob/main/internal/cmd/server.go#L81) once when EG server starts and is not included in the [loader hook func](https://github.com/envoyproxy/gateway/blob/e7f58d26ae2d91797ce61dd4276f9b961f006067/internal/cmd/server.go#L66-L74).
The same applies for [metrics server initialization](https://github.com/envoyproxy/gateway/blob/e7f58d26ae2d91797ce61dd4276f9b961f006067/internal/cmd/server.go#L85). 

*Environment*:
latest EG.

sudiptob2:
fixes #6340

### DIFF
--- a/internal/admin/server.go
+++ b/internal/admin/server.go
@@ -6,6 +6,7 @@
 package admin
 
 import (
+	"context"
 	"net/http"
 	"net/http/pprof"
 	"time"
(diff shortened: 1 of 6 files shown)


arkodg:
@Xunzhuo this should help with https://github.com/envoyproxy/gateway/pull/6331

zirain (approved):

PR #6489, opened 2025-07-09
helm: clean up helm hook resources

gebinic:
**What type of PR is this?**

- helm: clean up helm hook resources

When `helm uninstall` is executed, some resources of the envoy gateway remain in the k8s cluster. Some resources contain Helm hooks and are no longer managed by Helm, so they must be cleaned up manually.

Release Notes: No

### DIFF
--- a/charts/gateway-helm/README.md
+++ b/charts/gateway-helm/README.md
@@ -117,4 +117,5 @@ helm uninstall eg -n envoy-gateway-system
 | service.trafficDistribution | string | `""` |  |
 | topologyInjector.annotations | object | `{}` |  |
 | topologyInjector.enabled | bool | `true` |  |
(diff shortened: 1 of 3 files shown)


zirain:
I didn't quite understand when will this happen?
I deploy the chart very often, I didn't notice that resources remained except CRDs.

gebinic:
This happens because `helm uninstall` does not delete resources created by hooks. All resources from the [certgen-rbac.yaml](https://github.com/envoyproxy/gateway/blob/main/charts/gateway-helm/templates/certgen-rbac.yaml) are created by a Helm hook and won't be deleted when `helm uninstall` is executed.

https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

PR #5451, opened 2025-03-10
feat: Add support for BackendTLSPolicy to target ServiceImport

nareddyt:
## Problem

BackendTLSPolicy only supports Service and Backend targetRefs, it doesn't support ServiceImport.

## Fix

This PR adds support for that case.

## Testing Done

Added a new translator test case that verifies `tls` settings appear in the xDS IR for ServiceImports.

## Additional Info

Fixes https://github.com/envoyproxy/gateway/issues/5376

Release Notes: Yes

PR #6495, opened 2025-07-09
feat(ext_service): Add support for SecurityPolicy and EnvoyExtensionPolicy to target ServiceImport

nareddyt:
## Problem

Similar to https://github.com/envoyproxy/gateway/pull/5451

SecurityPolicy and EnvoyExtensionPolicy only supports Service and Backend backendRefs, it doesn't support ServiceImport.

## Fix

This PR adds support for that case.

## Testing Done

Added a few new translator test cases that copy pre-existing ones but use ServiceImport for the ext-authz / ext-proc / oidc / jwks clusters.

## Additional Info

Release Notes: Yes


nareddyt:
> ## [Codecov](https://app.codecov.io/gh/envoyproxy/gateway/pull/6495?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=envoyproxy) Report
> Attention: Patch coverage is `44.82759%` with `16 lines` in your changes missing coverage. Please review.
> 


Looks like it's impossible to fix this coverage due to the comment here: https://github.com/envoyproxy/gateway/blob/5b6262655d6c1e29d4c457b469f6f44ce756f601/internal/gatewayapi/validate.go#L961-L962

zhaohuabing:
While it's technically possible, I'm curious if there are any real use cases for using a service from another cluster as the backend for ext auth or ext proc extensions? The best practice should be running the service as a sidecar in the envoy pod.

nareddyt:
> While it's technically possible, I'm curious if there are any real use cases for using a service from another cluster as the backend for ext auth or ext proc extensions? The best practice should be running the service as a sidecar in the envoy pod.

@zhaohuabing this is my first time hearing of this best practice. At Confluent, we have an ext_authorization server running per region in a separate cluster, and we'd like to call out to that via `ServiceImport`. I've also handled a very similar architecture during my time at Google, so I think it's fairly common. 

I'd also like to point out that ext_authz, jwt, oidc all support an EnvoyGateway `Backend` resource already. There are many test configs in this repo for this scenario. If `Backend` to an "external domain name" is already supported, IMO a `ServiceImport` to a k8s service in an "external cluster" should also be supported.

If you'd like me to limit `ServiceImport` to only be supported for ext_authz, I'm happy to do that, though it would be an artificial validation check and make the code a little more complicated.

internal/gatewayapi/validate.go
```diff
@@ -960,15 +960,12 @@ func (t *Translator) validateExtServiceBackendReference(
 ) error {
 	// These are sanity checks, they should never happen because the API server
 	// should have caught them
-	if backendRef.Group != nil && *backendRef.Group != "" && *backendRef.Group != egv1a1.GroupName {
-		return errors.New(
-			"group is invalid, only the core API group (specified by omitting" +
-				" the group field or setting it to an empty string) and the" +
-				" gateway.envoyproxy.io API group are supported")
+	if backendRef.Group != nil && *backendRef.Group != "" && *backendRef.Group != GroupMultiClusterService && *backendRef.Group != egv1a1.GroupName {
```
arkodg, line 963: should it be || instead ?

nareddyt: I don't think so. This only errors out when group name matches none of them. The many test cases I added would catch bugs here BTW

arkodg: Ah thanks

arkodg (approved):
LGTM thanks

zhaohuabing:
> > While it's technically possible, I'm curious if there are any real use cases for using a service from another cluster as the backend for ext auth or ext proc extensions? The best practice should be running the service as a sidecar in the envoy pod.
> 
> @zhaohuabing this is my first time hearing of this best practice. At Confluent, we have an ext_authorization server running per region in a separate cluster, and we'd like to call out to that via `ServiceImport`. I've also handled a very similar architecture during my time at Google, so I think it's fairly common.
> 
> I'd also like to point out that ext_authz, jwt, oidc all support an EnvoyGateway `Backend` resource already. There are many test configs in this repo for this scenario. If `Backend` to an "external domain name" is already supported, IMO a `ServiceImport` to a k8s service in an "external cluster" should also be supported.
> 
> If you'd like me to limit `ServiceImport` to only be supported for ext_authz, I'm happy to do that, though it would be an artificial validation check and make the code a little more complicated.

Hi @nareddyt — just curious about the use cases because the most common use case I heard about on this is sidecar + Backend with UDS endpoint. Really appreciate the detailed explanations!

zhaohuabing (approved):

PR #6497, opened 2025-07-10
feat: Implement ZoneAware loadbalancing - Use non-static local_cluster name (splitup #6482)

jukie:
**What type of PR is this?**
feat: Implement ZoneAware Lb Config

Relates to the implementation for `LoadBalancer.ZoneAware` API (Ref: https://github.com/envoyproxy/gateway/pull/6154)

Adjusts the bootstrap local_cluster name to be modeled after the proxy infra name. This will later be used to track  EnvoyProxy endpoints and render them via xDS.

Functionality-wise this is a no-op an changes the bootstrap cluster name from `local_cluster` to a non-static value generated based on the computed infra name.

xRef #6025 

Release Notes: No
(will add release note for whichever PR gets merged last)


internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -33,6 +33,8 @@ const (
 	envoyPodEnvVar = "ENVOY_POD_NAME"
 	// envoyZoneEnvVar is the Envoy pod locality zone name
 	envoyZoneEnvVar = "ENVOY_SERVICE_ZONE"
+	// envoyInfraNameEnvVar is the name of the envoy proxy infrastructure
+	envoyInfraNameEnvVar = "ENVOY_PROXY_INFRA_NAME"
```
arkodg: instead of using an env, my recommendation is to pass a value to the bootstrap template
and it could look like the serviceCluster field https://github.com/envoyproxy/gateway/blob/954c858fb42b00c08bcfee6ae22e7c9034b7dc6c/internal/infrastructure/common/proxy_args.go#L56 that is unique per infrastructure fleet, if that is the goal

arkodg: and also lets make sure we default to `local_cluster` in case no value is passed, to also account for the non k8s case

jukie: Updated with bootstrap arg and defaults to `local_cluster`

internal/infrastructure/kubernetes/proxy/resource.go
```diff
@@ -109,6 +109,7 @@ func expectedProxyContainers(infra *ir.ProxyInfra,
 		},
 		MaxHeapSizeBytes: maxHeapSizeBytes,
 		XdsServerHost:    ptr.To(fmt.Sprintf("%s.%s.svc.%s", config.EnvoyGatewayServiceName, controllerNamespace, dnsDomain)),
+		LocalClusterName: ptr.To(fmt.Sprintf("%s-%s", config.EnvoyPrefix, utils.GetHashedName(infra.Name, 64))),
```
arkodg: my preference would be to reuse the same naming scheme as service cluster to avoid introducing a new name here

jukie: Updated

guydc (approved):

Issue #6280, opened 2025-06-09
Add hostname to BackendEndpoint


Ri7ay:
*Description*:
Currently, the BackendEndpoint resource in Envoy Gateway supports only an address IP and port, but does not include a distinct hostname field. This means that when an endpoint is specified by IP, Envoy’s auto_host_rewrite feature cannot rewrite the HTTP Host header, because there is no hostname to rewrite to.
We would like to add an optional hostname field to BackendEndpoint, so that users can:
Define an upstream by its IP address (for example, in cases where the service has no stable DNS name or uses private IPs).
Specify a hostname to be used as the HTTP Host header when auto_host_rewrite is enabled on the Envoy cluster.
With this addition, Envoy Gateway will:
Continue to route traffic to the specified IP address.
When auto_host_rewrite is set to true, inject the configured hostname into the outgoing HTTP request’s Host header.

It is proposed to add a hostname field to BackendEndpoint, similar to the proto type Envoy config.endpoint.v3.Endpoint.

[optional *Relevant Links*:]
> https://gateway.envoyproxy.io/docs/api/extension_types/#backendendpoint
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto
> https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto

arkodg:
hey @Ri7ay you can use the `fqdn` field (short for fully qualified domain name) to represent a hostname, closing this for now, feel free to reopen if you have more questions

arkodg:
added this to the community meeting agenda item for today

arkodg:
we discussed this in the community meeting this week, and adding an optional `Hostname` field in the `Backend` API is reasonable addition, adding the `help-wanted` tag to this

Ri7ay:
I will take this

Inode1:
> we discussed this in the community meeting this week, and adding an optional `Hostname` field in the `Backend` API is reasonable addition, adding the `help-wanted` tag to this

The feature is not finished yet, we really want it. While BackendTLSPolicy allows specifying an SNI hostname, it does not influence the HTTP Host header. This creates a gap in functionality where: TLS SNI can be set, but the HTTP Host header still defaults to the original request or is left empty.
This feature will close that gap and provide a complete solution for controlling both TLS SNI and HTTP Host header independently or in tandem.

[...]

PR #6503, opened 2025-07-10
Add hostname to backend

Ri7ay:
**What type of PR is this?**

* "feat: add hostname to endpoint"

Relates to the implementation for **Add hostname to BackendEndpoint** (Ref: #6280)

xRef: #6280 

Release Notes: No


internal/ir/xds.go
```diff
@@ -1683,6 +1683,8 @@ const (
 // DestinationEndpoint holds the endpoint details associated with the destination
 // +kubebuilder:object:generate=true
 type DestinationEndpoint struct {
+	// Hostname refers to the endpoint's hostname
+	Hostname *string `json:"hostname,omitempty" yaml:"hostname,omitempty"`
```
arkodg, line 1687: should this be per DestinationSetting instead ?

arkodg:
the PR looks good @Ri7ay 
* can you sign your commit and force push
* added one comment

internal/ir/xds.go
Ri7ay: I think DestinationEndpoint is a more logical and convenient place because the hostname itself is located inside backend.Spec.Endpoints, which allows us to specify different hostnames for different endpoints. At the same time, in DestinationSetting we will have 1 hostname for the entire Endpoints array (https://github.com/envoyproxy/gateway/blob/main/internal/ir/xds.go#L1644). What do you think about this?

arkodg: although in xds, its per endpoint, from the input API perspective, its per Setting, by doing this we also save space for cases with many endpoints

Ri7ay: Should I add a check to make sure all hostnames are the same in the backend into validateBackend?

arkodg: Nah not needed imo

Ri7ay: Done, I rewrote it in DestinationSetting.

charts/gateway-addons-helm/tmpcharts/alloy-0.9.2.tgz
shawnh2, line 1: unrelated changes ?

Ri7ay: No, it's kind of like a backend_types.go file gen

Ri7ay: sorry no, its tmpchart, I will delete this

Ri7ay: Done, thanks to saw this

internal/gatewayapi/route.go
```diff
@@ -1535,6 +1536,7 @@ func (t *Translator) processServiceImportDestinationSetting(
 	if !t.IsEnvoyServiceRouting(envoyProxy) {
 		endpointSlices := resources.GetEndpointSlicesForBackend(backendNamespace, string(backendRef.Name), resource.KindServiceImport)
 		endpoints, addrType = getIREndpointsFromEndpointSlices(endpointSlices, servicePort.Name, servicePort.Protocol)
+		hostname = getIREndpointHostname(endpointSlices)
```
arkodg: prefer if we did this separately, im not sure how the Hostname field is populated in endpointslice, and since this PR is for Backend, lets scope it to only Backend

Ri7ay: Yes, I agree, I deleted this code.

zirain (commented):
LG, would be better to add a test in gatwayapi layer.

internal/xds/translator/cluster_test.go
```diff
@@ -22,8 +22,9 @@ import (
 )
 
 const (
-	envoyGatewayXdsServerHost = "envoy-gateway"
-	xdsClusterName            = "xds_cluster"
+	envoyGatewayXdsServerHostname = "envoy-gateway-hostname"
```
arkodg: where is this used

internal/ir/xds.go
```diff
@@ -1640,7 +1640,9 @@ type DestinationSetting struct {
 	// Lower priority endpoints will be used only if higher priority levels are unavailable.
 	Priority *uint32 `json:"priority,omitempty"`
 	// Protocol associated with this destination/port.
-	Protocol  AppProtocol            `json:"protocol,omitempty" yaml:"protocol,omitempty"`
+	Protocol AppProtocol `json:"protocol,omitempty" yaml:"protocol,omitempty"`
+	// Hostname refers to the endpoint's hostname
```
arkodg: apologies for just noticing this
the `Hostname` is per `Endpoint` in the API, so it should be under `Endpoint` where it was before

Ri7ay: Okay, back to the original version.

Ri7ay:
Passed make -k gen-chek so that ci passed, can you approve it again please.
@arkodg @shawnh2

shawnh2 (approved):

PR #6506, opened 2025-07-11
fix: avoid appending same Backends twice to resource map

zirain:
supersede https://github.com/envoyproxy/gateway/pull/6021


internal/provider/kubernetes/controller.go
```diff
@@ -588,21 +578,26 @@ func (r *gatewayAPIReconciler) processBackendRefs(ctx context.Context, gwcResour
 			endpointSliceLabelKey = mcsapiv1a1.LabelServiceName
 
 		case egv1a1.KindBackend:
+			if !r.backendCRDExists {
+				r.log.Info("Backend CRD does not exist, skipping processing BackendRef")
```
jukie: Do we need/want this log output?

zirain: I think it's fine to change level to 6, since we have status.

jukie (approved):
My approvals don't help much yet but LGTM!

zirain:
> My approvals don't help much yet but LGTM!

your approvals is helpful, we need 1 from reviewer/mantainer, and 1 from maintainer.

internal/provider/kubernetes/controller.go
```diff
@@ -747,6 +744,22 @@ func (r *gatewayAPIReconciler) processSecurityPolicyObjectRefs(
 					"failed to process OIDC SecretRef for SecurityPolicy",
 					"policy", policy, "secretRef", oidc.ClientSecret)
 			}
+
+			var backendRef gwapiv1.BackendObjectReference
+			if oidc.Provider.BackendRef != nil {
+				backendRef = *oidc.Provider.BackendRef
+			}
+			if len(oidc.Provider.BackendRefs) > 0 {
+				backendRef = oidc.Provider.BackendRefs[0].BackendObjectReference
```
arkodg: why arent we inserting all ?

arkodg: this bug also exists in ext auth :( , lets tackle it separately

arkodg: this change however breaks functionality, because before we maybe incorrectly processing here, but with processbackendRefs we were still processing all and they would be available in the gateway-api layer.

now since we have limited processing, its going to break those multi backendRef cases

zirain: All the test case is showing about use just one backend, but I change the logic to keep same as before, we can fix it later.

zirain: cc @zhaohuabing

